Issue #12005, opened 2024-03-13
Port XXDecomposer to Rust


mtreinish:
In the default 2q unitary synthesis plugin we use the `TwoQubitBasisDecomposer` and the `XXDecomposer` based on the target (both the basis and fidelities).  To finish #8774 and have a parallel rust implementation of the default synthesis plugin we need to port the `XXDecomposer` class and likely the entire `qiskit/synthesis/two_qubit/xx_decompose` subpackage to rust.

mtreinish:
Now that #12004 has been implemented the XX decomposer is potentially the slowest part of optimization level 3 transpile. I ran a 100 qubit QV with a depth of 30 and generated this profile using main:

![Screenshot_2024-03-28_17-23-21](https://github.com/Qiskit/qiskit/assets/2447371/61eb810a-e28e-4f59-a49b-b7fce89f4da4)

The highlighted section that takes 46.6 seconds is the time spent in the XXDecomposer and it's taking ~50% of the total transpile time.

I generated that profile by running the following under cprofile:

```python
import time
from qiskit import transpile
from qiskit.transpiler import CouplingMap
from qiskit.providers.fake_provider import GenericBackendV2
from qiskit.circuit.library import QuantumVolume

cmap = CouplingMap.from_heavy_hex(19)

backend = GenericBackendV2(len(cmap.graph), coupling_map=cmap, basis_gates=['cx', 'rz', 'sx', 'x', 'id'])

qc = QuantumVolume(100, depth=30)
qc.measure_all()
start = time.perf_counter()
transpile(qc, backend, optimization_level=3)
stop = time.perf_counter()
print(stop - start)
```

levbishop:
The xxdecomposer should be sped up, but for that basis set it seems like we shouldn't be calling it at all since the only xx class gate available is cx and the twoqubitbasisdecomposer should be giving optimal decomposition for that already?

mtreinish:
Hmm, that's a good point let me look at the `UnitarySynthesis` pass's default plugin and see why it's running the XX decomposer at all. There very likely is a logic bug in the pass for figuring out which decomposer to run in the `Target` path. The `basis_gates` list path is very simple it only runs `XXDecomposer` if  `rzx` is in the basis gates.

levbishop:
Just in general it would probably be nice `UnitarySynthesis` would do some logging about what decomposers it's trying and why.

PR #12151, opened 2024-04-06
Reduce number of decomposers used during UnitarySynthesis default plugin

mtreinish:
### Summary

This commit reduces the number of decomposers we run in the default synthesis plugin when we're in known targets. Previously the unitary synthesis plugin was trying to product of all 1q basis and 2q basis for every 2q pair that is being synthesized. This would result in duplicated work for several 1q bases where there were potential subsets available as two different target euler bases, mainly U321 and U3 or ZSX and ZSXX if the basis gates for a qubit where U3, U2, U1 or Rz, SX, and X respectively. This reuses the logic from Optimize1qGatesDecomposition to make the euler basis selection which does the deduplication. Similarly, in the presence of known 2q gates we can skip the XXDecomposer checks (or potentially running the XXDecomposer) which should speed up both the selection of decomposers and also reduce the number of decomposers we run.

### Details and comments

### DIFF
--- a/qiskit/transpiler/passes/synthesis/unitary_synthesis.py
+++ b/qiskit/transpiler/passes/synthesis/unitary_synthesis.py
@@ -772,7 +772,7 @@ class DefaultUnitarySynthesis(plugin.UnitarySynthesisPlugin):
             decomposers.append(decomposer)
 
         # If our 2q basis gates are a subset of cx, ecr, or cz then we know TwoQubitBasisDecomposer
-        # is an ideal deocmposition and there is no need to bother calculating the XX embodiments
+        # is an ideal decomposition and there is no need to bother calculating the XX embodiments
         # or try the XX decomposer
         if {"cx", "cz", "ecr"}.issuperset(available_2q_basis):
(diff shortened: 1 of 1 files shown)


mtreinish:
With this running the script from: https://github.com/Qiskit/qiskit/issues/12005#issuecomment-2026165687 goes from taking ~37.7 sec on main to ~13.1 sec.

qiskit/transpiler/passes/synthesis/unitary_synthesis.py
```diff
@@ -777,6 +771,13 @@ def is_controlled(gate):
             )
             decomposers.append(decomposer)
 
+        # If our 2q basis gates are a subset of cx, ecr, or cz then we know TwoQubitBasisDecomposer
+        # is an ideal deocmposition and there is no need to bother calculating the XX embodiments
```
jlapeyre:
```suggestion
        # is an ideal decomposition and there is no need to bother calculating the XX embodiments
```

jlapeyre (approved):
Looks like a great optimization. PR looks good.

PR #11855, opened 2024-02-21
Remove historical versions from release notes

Eric-Arellano:
See https://github.com/Qiskit/qiskit/pull/11840 for context.

Note that we don't include release notes for dev docs in docs.quantum.ibm.com, so these release notes won't hit production. This is instead intended to help Qiskit developers check that release notes are valid well in-advance of release prep.

PR #12153, opened 2024-04-07
Fix bad release-notes section

jakelishman:
### Summary

This causes failures in the docs build since #11855 merged (which just exposes them - it wasn't the cause), though I'm slightly confused as to why it doesn't _always_ seem to cause failures, nor why it didn't fail the docs build for 1.0.


sbrandhsn (approved):
LGTM

ElePT (commented):
Strange that it didn't fail before, I agree, but the change LGTM.

ElePT:
Haha @sbrandhsn beat me.

PR #12158, opened 2024-04-09
Add enhancement to BitArray

t-imamichi:
### Summary

Add more utility methods to BitArray

- [x] transpose
- [x] concatenate along existing axis
- [x] concatenate along shots axis
- [x] concatenate along bits axis
- [x] slicing along existing axis
- [x] slicing along shots axis
- [x] slicing along bits axis (this helps marginalize counts/bitstrings)
- [x] compute expectation values of diagonal observables
- [x] reno

### DIFF
--- a/qiskit/primitives/containers/bit_array.py
+++ b/qiskit/primitives/containers/bit_array.py
@@ -440,10 +440,8 @@ class BitArray(ShapedMixin):
         arr, num_bits = _pack(arr)
         return BitArray(arr, num_bits)
 
-    def expectation_value(
-        self, operator: str | Pauli | SparsePauliOp, loc: int | tuple[int, ...] | None = None
-    ) -> float:
-        """Compute the expectation value of an operator.
+    def expectation_values(self, operator: str | Pauli | SparsePauliOp) -> NDArray[np.float64]:
+        """Compute the expectation value of the operator for every entry of this bit array.
 
         .. note::
 
@@ -452,15 +450,16 @@ class BitArray(ShapedMixin):
             :func:`~.sampled_expectation_value`.
 
         Args:
-            operator: The operator for the expectation value
-            loc: Which entry of this array to compute the expectation value.
-                If ``None``, counts from all positions in this array are unioned together.
+            operator: The operator to take the expectation value of.
 
         Returns:
-            The expectation value.
+            An array of expectation values whose shape matches this bit array's :attr:`~shape`.
         """
-        counts = self.get_counts(loc)
-        return sampled_expectation_value(counts, operator)
+        arr = np.empty(self.shape, dtype=float)
+        for loc in np.ndindex(self.shape):
+            counts = self.get_counts(loc)
+            arr[loc] = sampled_expectation_value(counts, operator)
+        return arr
 
     @staticmethod
     def concatenate(bit_arrays: Sequence[BitArray], axis: int = 0) -> BitArray:
--- a/qiskit/primitives/containers/bit_array.py
+++ b/qiskit/primitives/containers/bit_array.py
@@ -533,6 +533,8 @@ class BitArray(ShapedMixin):
         num_bits = bit_arrays[0].num_bits
         num_shots = bit_arrays[0].num_shots
         ndim = bit_arrays[0].ndim
+        if ndim == 0:
+            raise ValueError("Zero-dimensional bit arrays cannot be concatenated")
         for i, ba in enumerate(bit_arrays):
             if ba.num_bits != num_bits:
                 raise ValueError(
--- a/qiskit/primitives/containers/bit_array.py
+++ b/qiskit/primitives/containers/bit_array.py
@@ -442,7 +442,7 @@ class BitArray(ShapedMixin):
 
     def expectation_values(self, observables: ObservablesArrayLike) -> NDArray[np.float64]:
         """Compute the expectation values of the provided observables, broadcasted against
-        this bit arrays.
+        this bit array.
 
         .. note::
 
--- a/qiskit/primitives/containers/bit_array.py
+++ b/qiskit/primitives/containers/bit_array.py
@@ -24,6 +24,7 @@ from typing import Callable, Iterable, Literal, Mapping, Sequence
 import numpy as np
 from numpy.typing import NDArray
 
+from qiskit.exceptions import QiskitError
 from qiskit.result import Counts, sampled_expectation_value
 
 from .observables_array import ObservablesArray, ObservablesArrayLike
@@ -440,6 +441,29 @@ class BitArray(ShapedMixin):
         arr, num_bits = _pack(arr)
         return BitArray(arr, num_bits)
 
+    def slice_shots(self, indices: int | Sequence[int]) -> "BitArray":
+        """Return a bit array sliced along the shots axis of some indices of interest.
+
+        Args:
+            indices: The shots positions of interest to slice along.
+
+        Returns:
+            A bit array sliced along the shots axis.
+
+        Raises:
+            ValueError: If there are any invalid indices of the shots axis.
+        """
+        if isinstance(indices, int):
+            indices = (indices,)
+        for index in indices:
+            if index < 0 or index >= self.num_shots:
+                raise ValueError(
+                    f"index {index} is out of bounds for the number of shots {self.num_shots}."
+                )
+        arr = self._array
+        arr = arr[..., indices, :]
+        return BitArray(arr, self.num_bits)
+
     def expectation_values(self, observables: ObservablesArrayLike) -> NDArray[np.float64]:
         """Compute the expectation values of the provided observables, broadcasted against
         this bit array.
@@ -451,25 +475,36 @@ class BitArray(ShapedMixin):
             :func:`~.sampled_expectation_value`.
 
         Args:
-            observables: The observable(s) to take the expectation value of. Must have a shape
-            broadcastable with with this bit array.
+            observables: The observable(s) to take the expectation value of.
+            Must have a shape broadcastable with with this bit array and
+            the same number of qubits as the number of bits of this bit array.
+            The observables must be diagonal (I, Z, 0 or 1) too.
 
         Returns:
-            An array of expectation values whose shape is the broadcast shape of ``operator``
+            An array of expectation values whose shape is the broadcast shape of ``observables``
             and this bit array.
 
         Raises:
-            ValueError: If the provided operator does not have a shape broadcastable with this bit array.
+            ValueError: If the provided observables does not have a shape broadcastable with
+                this bit array.
+            ValueError: If the provided observables does not have the same number of qubits as
+                the number of bits of this bit array.
+            ValueError: If the provided observables are not diagonal.
         """
         observables = ObservablesArray.coerce(observables)
         arr_indices = np.fromiter(np.ndindex(self.shape), dtype=object).reshape(self.shape)
         bc_indices, bc_obs = np.broadcast_arrays(arr_indices, observables)
+        counts = {}
         arr = np.zeros_like(bc_indices, dtype=float)
         for index in np.ndindex(bc_indices.shape):
             loc = bc_indices[index]
             for pauli, coeff in bc_obs[index].items():
-                counts = self.get_counts(loc)
-                expval = sampled_expectation_value(counts, pauli)
+                if loc not in counts:
+                    counts[loc] = self.get_counts(loc)
+                try:
+                    expval = sampled_expectation_value(counts[loc], pauli)
+                except QiskitError as ex:
+                    raise ValueError(ex.message) from ex
                 arr[index] += expval * coeff
         return arr
 
@@ -596,6 +631,9 @@ class BitArray(ShapedMixin):
                     f"but the bit array at index 0 has shape {shape} "
                     f"and the bit array at index {i} has shape {ba.shape}."
                 )
+        # This implementation introduces a temporary 8x memory overhead due to bit
+        # unpacking. This could be fixed using bitwise functions, at the expense of a
+        # more complicated implementation.
         data = np.concatenate([_unpack(ba) for ba in bit_arrays], axis=-1)
         data, num_bits = _pack(data)
         return BitArray(data, num_bits)
--- a/qiskit/primitives/containers/bit_array.py
+++ b/qiskit/primitives/containers/bit_array.py
@@ -24,19 +24,13 @@ from typing import Callable, Iterable, Literal, Mapping, Sequence
 import numpy as np
 from numpy.typing import NDArray
 
-from qiskit.result import Counts
+from qiskit.quantum_info import Pauli, SparsePauliOp
+from qiskit.result import Counts, sampled_expectation_value
 
 from .shape import ShapedMixin, ShapeInput, shape_tuple
 
-from qiskit._accelerate.sampled_exp_val import sampled_expval_float, sampled_expval_complex
-from qiskit.quantum_info import Pauli, SparsePauliOp
-
-
 # this lookup table tells you how many bits are 1 in each uint8 value
 _WEIGHT_LOOKUP = np.unpackbits(np.arange(256, dtype=np.uint8).reshape(-1, 1), axis=1).sum(axis=1)
-
-# A list of valid diagonal operators
-DIAG_OPERATORS = {"Z", "I", "0", "1"}
 
 
 def _min_num_bytes(num_bits: int) -> int:
@@ -410,16 +404,21 @@ class BitArray(ShapedMixin):
         arr = arr[..., indices[::-1]]
         num_bits = len(indices)
         pad_size = -num_bits % 8
-        pad_width = [(0, 0)] * (arr.ndim - 1) + [(pad_size, 0)]
-        arr = np.pad(arr, pad_width, constant_values=0)
+        if pad_size > 0:
+            pad_width = [(0, 0)] * (arr.ndim - 1) + [(pad_size, 0)]
+            arr = np.pad(arr, pad_width, constant_values=0)
         arr = np.packbits(arr, axis=-1, bitorder="big")
         return BitArray(arr, num_bits)
 
-    def expectation_values(self, operator: str | Pauli | SparsePauliOp) -> NDArray[np.float64]:
+    def expectation_value(
+        self, operator: str | Pauli | SparsePauliOp, loc: int | tuple[int, ...] | None = None
+    ) -> NDArray[np.float64]:
         """_summary_
 
         Args:
-            operators (str | Pauli | SparsePauliOp): _description_
+            operators: _description_
+            loc: Which entry of this array to return an expectation value for.
+                If ``None``, counts from all positions in this array are unioned together.
 
         Returns:
             ndarray: _description_
@@ -427,43 +426,18 @@ class BitArray(ShapedMixin):
         Raises:
             ValueError: if...
         """
-        if isinstance(operator, str):
-            oper_strs = [operator.upper()]
-            coeffs = np.asarray([1.0])
-        elif isinstance(operator, Pauli):
-            oper_strs = [operator.to_label()]
-            coeffs = np.asarray([1.0])
-        elif isinstance(operator, SparsePauliOp):
-            oper_strs = operator.paulis.to_labels()
-            coeffs = np.real_if_close(operator.coeffs)
-            # workaround to avoid "TypeError: The given array is not contiguous"
-            # by sampled_expval_float
-            coeffs = np.array(coeffs)
-        else:
-            raise ValueError(f"Invalid operator type {type(operator)}")
-        for op in oper_strs:
-            if not set(op).issubset(DIAG_OPERATORS):
-                raise ValueError(f"Input operator {op} is not diagonal")
-            if len(op) != self.num_bits:
-                raise ValueError(
-                    f"The size {len(op)} of operator {op} differs from the number of bits {self.num_bits}."
-                )
-        if coeffs.dtype == np.dtype(complex).type:
-            sampled_expval = sampled_expval_complex
-        else:
-            sampled_expval = sampled_expval_float
-        ret = np.empty(self.shape)
-        for index in np.ndindex(self.shape):
-            counts = self.get_counts(index)
-            ret[index] = sampled_expval(oper_strs, coeffs, counts)
-        return ret
+        counts = self.get_counts(loc)
+        return sampled_expectation_value(counts, operator)
 
 
 def concatenate(bitarrays: Sequence[BitArray], axis: int = 0) -> BitArray:
     """Join a sequence of bit arrays along an existing axis.
 
     Args:
-        bitarrays: The sequence of bit arrays.
+        bitarrays: The bit arrays must have (1) the same number of bits,
+            (2) the same number of shots, and
+            (3) the same shape, except in the dimension corresponding to axis
+            (the first, by default).
         axis: The axis along which the arrays will be joined. Default is 0.
 
     Returns:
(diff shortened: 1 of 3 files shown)


qiskit/primitives/containers/bit_array.py
```diff
@@ -347,3 +351,133 @@ def reshape(self, *shape: ShapeInput) -> "BitArray":
...
+                `numpy.transpose
+                <https://numpy.org/doc/stable/reference/generated/numpy.transpose.html>`_
+                for the details.
+
+        Returns:
+            BitArray: a bit array with axes permuted.
+
+        Raises:
+            ValueError: if ``axes`` don't match this bit array.
+            ValueError: if ``axes`` includes any indices that are out of bounds.
```
ihincks: Minor convention consistency. Qiskit does some of both, but I think the V2 primitives lean a bit more toward the cap convention at this point.

```suggestion
            axes: Tuple or list of ints, optional. See
                `numpy.transpose
                <https://numpy.org/doc/stable/reference/generated/numpy.transpose.html>`_
                for the details.

        Returns:
            BitArray: A bit array with axes permuted.

        Raises:
            ValueError: If ``axes`` don't match this bit array.
            ValueError: If ``axes`` includes any indices that are out of bounds.
```

qiskit/primitives/containers/bit_array.py
```diff
@@ -347,3 +351,133 @@ def reshape(self, *shape: ShapeInput) -> "BitArray":
...
             raise ValueError("Cannot change the size of the array.")
         return BitArray(self._array.reshape(shape), self.num_bits)
+
+    def transpose(self, axes: tuple[int, ...] | list[int] | None = None) -> "BitArray":
+        """Return a bit array with axes transposed.
+
+        Args:
+            axes: tuple or list of ints, optional. See
+                `numpy.transpose
+                <https://numpy.org/doc/stable/reference/generated/numpy.transpose.html>`_
```
ihincks: Good idea pointing directly at NumPy. I'm wondering whether we should point at the ndarray method instead? The main difference is that the method form also accepts `*args`. Do you think that's something people would find helpful, or is it just a burden to have two different ways of specifying the transpose?

qiskit/primitives/containers/bit_array.py
```diff
@@ -347,3 +351,133 @@ def reshape(self, *shape: ShapeInput) -> "BitArray":
...
+        Raises:
+            ValueError: if there are any invalid indices to marginalize over.
+        """
+        if isinstance(indices, int):
+            indices = (indices,)
+        for index in indices:
+            if index < 0 or index >= self.num_bits:
+                raise ValueError(
+                    f"index {index} is out of bounds for the number of bits {self.num_bits}."
+                )
```
ihincks, line 435:
```suggestion
                )
        # This implementation introduces a temporary 8x memory overhead due to bit 
        # unpacking. This could be fixed using bitwise functions, at the expense of a
        # more complicated implementation.
```

qiskit/primitives/containers/bit_array.py
```diff
@@ -347,3 +368,200 @@ def reshape(self, *shape: ShapeInput) -> "BitArray":
...
+            if i >= self.ndim or self.ndim + i < 0:
+                raise ValueError(
+                    f"axis {i} is out of bounds for bit array of dimension {self.ndim}."
+                )
+        axes = tuple(i if i >= 0 else self.ndim + i for i in axes) + (-2, -1)
+        return BitArray(self._array.transpose(axes), self.num_bits)
+
+    def marginalize(self, indices: int | Sequence[int]) -> "BitArray":
+        """Return a bit array marginalized over some indices of interests.
+
```
ihincks:
```suggestion
        """Return a bit array marginalized over some indices of interest.

        .. note::
            The convention used by this method is that the index ``0`` corresponds to the least-significant 
            bit in the :attr:`~array`, or equivalently the right-most bitstring entry as returned by 
            :meth:`~get_counts` or :meth:`~get_bitstrings`, etc.

            If this bit array was produced by a sampler, then an index ``i`` corresponds to the   
            :class:`~.ClassicalRegister` location ``creg[i]]`.
```

qiskit/primitives/containers/bit_array.py
```diff
@@ -347,3 +368,200 @@ def reshape(self, *shape: ShapeInput) -> "BitArray":
...
+        for i in axes:
+            if i >= self.ndim or self.ndim + i < 0:
+                raise ValueError(
+                    f"axis {i} is out of bounds for bit array of dimension {self.ndim}."
+                )
+        axes = tuple(i if i >= 0 else self.ndim + i for i in axes) + (-2, -1)
+        return BitArray(self._array.transpose(axes), self.num_bits)
+
+    def marginalize(self, indices: int | Sequence[int]) -> "BitArray":
+        """Return a bit array marginalized over some indices of interests.
```
ihincks: I'm second guessing whether ``marginalize`` is the right name for this method, because this class doesn't actually represent a distribution. I'm wondering if ``bit_slice`` might be better, or something along those lines. What do you or other reviewers think?

qiskit/primitives/containers/bit_array.py
```diff
@@ -347,3 +368,200 @@ def reshape(self, *shape: ShapeInput) -> "BitArray":
...
+
+        Args:
+            indices: The bit positions of interest to marginalize over.
+
+        Returns:
+            A bit array marginalized to bits of interests.
+
+        Raises:
+            ValueError: if there are any invalid indices to marginalize over.
```
ihincks:
```suggestion
            ValueError: If there are any invalid indices to marginalize over.
```

qiskit/primitives/containers/bit_array.py
```diff
@@ -347,3 +368,200 @@ def reshape(self, *shape: ShapeInput) -> "BitArray":
...
+            if index < 0 or index >= self.num_bits:
+                raise ValueError(
+                    f"index {index} is out of bounds for the number of bits {self.num_bits}."
+                )
+        arr = _unpack(self)
+        arr = arr[..., indices]
+        arr, num_bits = _pack(arr)
+        return BitArray(arr, num_bits)
+
+    def expectation_value(
```
ihincks: In the meeting we had discussed an array-valued version of this method. I'm not sure if it's possible to introduce such a thing in this PR because it would be blocked by making ObservableArray a public class. @chriseclectic may have advice.

qiskit/primitives/containers/bit_array.py
```diff
@@ -110,6 +128,9 @@ def __repr__(self):
         desc = f"<shape={self.shape}, num_shots={self.num_shots}, num_bits={self.num_bits}>"
         return f"BitArray({desc})"
 
+    def __getitem__(self, indices):
+        return BitArray(self._array[indices], self.num_bits)
```
ihincks, line 137: This could run into issues if indices slices the last axis because it would change the number of bits.

```suggestion
        if isinstance(indices, tuple) and len(indices) >= self.ndim + 2:
            raise ValueError("BitArrays cannot be sliced along the bits axis, see marginalize() instead.")
        return BitArray(self._array[indices], self.num_bits)
```

ihincks (commented):
This looks very nice, thank you @t-imamichi

qiskit/primitives/containers/bit_array.py
t-imamichi: You are right. I like the name `bit_slice`.

qiskit/primitives/containers/bit_array.py
t-imamichi: Thank you for pointing out `*args`. I forgot about it. I updated this method to accept integers in addition to a tuple of integers. As for the link, I like that of the ndarray method. I updated it too.

t-imamichi:
There are some topics to discuss as follows.

1. Names of methods and functions.
  - `concatenate`: join BitArray along BitArray's axes. I named it because the behavior is similar to `np.concatenate`.
  - `stack_shots`: join BitArray along shot axis.
  - `stack_bits`: join BitArray along bit axis.
  - `marginalize`: slice data along bit axis. I like Ian's suggestion `bit_slice` https://github.com/Qiskit/qiskit/pull/12158#discussion_r1571614531.
 
2. Location of the three functions to join BitArrays. I currently make them static methods of BitArray. So, users need to call `BitArray.concatenate`, for example. There is another option to make a module such as `qiskit.primitives.containers.utils`.

3. How to extend `expectation_value` method to deal with array of operators. https://github.com/Qiskit/qiskit/pull/12158#discussion_r1571619070

qiskit/primitives/containers/bit_array.py
t-imamichi: Since `sample_expval_{float,complex}` returns a single expval, we need to extend it to allow multiple operators and return multiple expvals. It looks technically easy, but it might be good to be a separate PR.

As the first step, it might be nice to extend this `expectation_value` method to handle `ObservableArray` itself or a list generated by `ObservableArray.tolist()` without changing the rust code, i.e., run a loop in `expectation_value` method.

As the second step, we can update the rust code optimized for our input of `expectaion_value` method and remove the loop in `expectation_value` method.

https://github.com/Qiskit/qiskit/blob/5fb343f5f2faf5578dc0afcd6a3393fbaa0e0c17/crates/accelerate/src/sampled_exp_val.rs#L58-L70

qiskit/primitives/containers/bit_array.py
```diff
@@ -347,3 +372,216 @@ def reshape(self, *shape: ShapeInput) -> "BitArray":
...
+            axes = tuple(reversed(range(self.ndim)))
+        if len(axes) == 1 and isinstance(axes[0], Sequence):
+            axes = axes[0]
+        if len(axes) != self.ndim:
+            raise ValueError("axes don't match bit array")
+        for i in axes:
+            if i >= self.ndim or self.ndim + i < 0:
+                raise ValueError(
+                    f"axis {i} is out of bounds for bit array of dimension {self.ndim}."
+                )
```
ihincks, line 403: There might be a couple of missing edge cases here that we should catch. For example, in the second conditional, axes[0] might not exist. Or, if the axes are not a permutation of range(self.ndim), ie something bis repeated. Maybe the intention was to rely on numpy.transpose for this last one?

qiskit/primitives/containers/bit_array.py
```diff
@@ -347,3 +372,216 @@ def reshape(self, *shape: ShapeInput) -> "BitArray":
...
+        axes = tuple(i if i >= 0 else self.ndim + i for i in axes) + (-2, -1)
+        return BitArray(self._array.transpose(axes), self.num_bits)
+
+    def marginalize(self, indices: int | Sequence[int]) -> "BitArray":
+        """Return a bit array marginalized over some indices of interest.
```
ihincks: This is unlikely to make a difference, but I think it's probably _slightly_ more clear that we'll be keeping the provided indices, rather than throwing them away.
```suggestion
        """Return a bit array marginalized to some indices of interest.
```

qiskit/primitives/containers/bit_array.py
```diff
@@ -347,3 +372,216 @@ def reshape(self, *shape: ShapeInput) -> "BitArray":
...
+
+        .. note::
```
ihincks, line 410:
```suggestion

        This method is called ``marginalize`` because its contents typically represent
        randomly sampled bitstrings, in which case the output of this function represents
        samples from the marginal distribution. However, it could just as well have been 
        called ``slice_bits``.

        .. note::
```

qiskit/primitives/containers/bit_array.py
ihincks: One way to interpret the current implementation is that it's implementing the trivial broadcast rule. Ie, the "shape" of the input observable is `()`. We could look to extend this method by, in a future PR, accepting a non-trivial shape of observables, like an ObservablesArray or something coercible thereto.

With this in mind, what do you think about removing the `loc` argument, and instead returning the output of `sampled_expectation_value` for _every_ index of the shape. For example, if a bit array has shape (2,5), then the output of this method will be a float array of shape (2,5)? The extension to multiple observables would be to demand broadcastability between the observables array and the bit array.

Please let me know if you want an example.

qiskit/primitives/containers/bit_array.py
```diff
@@ -347,3 +372,216 @@ def reshape(self, *shape: ShapeInput) -> "BitArray":
...
+        return sampled_expectation_value(counts, operator)
+
+    @staticmethod
+    def concatenate(bitarrays: Sequence[BitArray], axis: int = 0) -> BitArray:
+        """Join a sequence of bit arrays along an existing axis.
+
+        Args:
+            bitarrays: The bit arrays must have (1) the same number of bits,
+                (2) the same number of shots, and
+                (3) the same shape, except in the dimension corresponding to axis
```
ihincks, line 518:
```suggestion
                (3) the same shape, except in the dimension corresponding to ``axis``
```

qiskit/primitives/containers/bit_array.py
```diff
@@ -347,3 +372,216 @@ def reshape(self, *shape: ShapeInput) -> "BitArray":
...
+                (the first, by default).
+            axis: The axis along which the arrays will be joined. Default is 0.
+
+        Returns:
+            BitArray: The concatenated bit array.
```
ihincks:
```suggestion
            The concatenated bit array.
```

qiskit/primitives/containers/bit_array.py
```diff
@@ -347,3 +372,216 @@ def reshape(self, *shape: ShapeInput) -> "BitArray":
...
+                    f"and the bit array at index {i} has {ba.num_shots} shots."
+                )
+            if ba.ndim != ndim:
+                raise ValueError(
+                    "All bit arrays must have same number of dimensions, "
+                    f"but the bit array at index 0 has {ndim} dimension(s) "
+                    f"and the bit array at index {i} has {ba.ndim} dimension(s)."
+                )
+        if axis is None:
+            raise ValueError(f"axis {axis} must be a non-negative integer.")
```
ihincks: Not obvious to me why None gets special treatment here. Perhaps left-over from a previous iteration? 

```suggestion
```

qiskit/primitives/containers/bit_array.py
```diff
@@ -347,3 +372,216 @@ def reshape(self, *shape: ShapeInput) -> "BitArray":
...
+        if axis < 0 or axis >= ndim:
+            raise ValueError(f"axis {axis} is out of bounds for bit array of dimension {ndim}.")
+        data = np.concatenate([ba.array for ba in bitarrays], axis=axis)
+        return BitArray(data, num_bits)
+
+    @staticmethod
+    def stack_shots(bitarrays: Sequence[BitArray]) -> BitArray:
+        """Join a sequence of bit arrays along shots.
```
ihincks:
```suggestion
        """Join a sequence of bit arrays along the shots axis.
```

qiskit/primitives/containers/bit_array.py
```diff
@@ -347,3 +372,216 @@ def reshape(self, *shape: ShapeInput) -> "BitArray":
...
+
+        Args:
+            bitarrays: The bit arrays must have (1) the same number of bits,
+                and (2) the same shape.
+
+        Returns:
+            BitArray: The stacked bit array.
```
ihincks:
```suggestion
            The stacked bit array.
```

qiskit/primitives/containers/bit_array.py
```diff
@@ -347,3 +372,216 @@ def reshape(self, *shape: ShapeInput) -> "BitArray":
...
+                    "All bit arrays must have same shape, "
+                    f"but the bit array at index 0 has shape {shape} "
+                    f"and the bit array at index {i} has shape {ba.shape}."
+                )
+        data = np.concatenate([ba.array for ba in bitarrays], axis=-2)
+        return BitArray(data, num_bits)
+
+    @staticmethod
+    def stack_bits(bitarrays: Sequence[BitArray]) -> BitArray:
+        """Join a sequence of bit arrays along bits.
```
ihincks:
```suggestion
        """Join a sequence of bit arrays along the bits axis.
```

qiskit/primitives/containers/bit_array.py
```diff
@@ -347,3 +372,216 @@ def reshape(self, *shape: ShapeInput) -> "BitArray":
...
+
```
ihincks, line 604:
```suggestion

.. note::
    This method is equivalent to per-shot bitstring concatenation.

```

qiskit/primitives/containers/bit_array.py
```diff
@@ -347,3 +372,216 @@ def reshape(self, *shape: ShapeInput) -> "BitArray":
...
+        Args:
+            bitarrays: The bit arrays must have (1) the same number of shots,
+                and (2) the same shape.
```
ihincks:
```suggestion
            bitarrays: Bit arrays that have (1) the same number of shots,
                and (2) the same shape.
```

qiskit/primitives/containers/bit_array.py
```diff
@@ -347,3 +372,216 @@ def reshape(self, *shape: ShapeInput) -> "BitArray":
...
+
+        Returns:
+            BitArray: The stacked bit array.
```
ihincks:
```suggestion
            The stacked bit array.
```

qiskit/primitives/containers/bit_array.py
```diff
@@ -347,3 +372,216 @@ def reshape(self, *shape: ShapeInput) -> "BitArray":
...
+                If ``None``, counts from all positions in this array are unioned together.
+
+        Returns:
+            The expectation value.
+        """
+        counts = self.get_counts(loc)
+        return sampled_expectation_value(counts, operator)
+
+    @staticmethod
+    def concatenate(bitarrays: Sequence[BitArray], axis: int = 0) -> BitArray:
```
ihincks: nit: to match CamelCase to snake_case because the class is `BitArray`. This happens a few times, but I'll just make the suggestion once.
```suggestion
    def concatenate(bit_arrays: Sequence[BitArray], axis: int = 0) -> BitArray:
```

qiskit/primitives/containers/bit_array.py
```diff
@@ -347,3 +372,216 @@ def reshape(self, *shape: ShapeInput) -> "BitArray":
...
+                    f"and the bit array at index {i} has {ba.num_shots} shots."
+                )
+            if ba.shape != shape:
+                raise ValueError(
+                    "All bit arrays must have same shape, "
+                    f"but the bit array at index 0 has shape {shape} "
+                    f"and the bit array at index {i} has shape {ba.shape}."
+                )
+        data = np.concatenate([_unpack(ba) for ba in bitarrays], axis=-1)
+        data, num_bits = _pack(data)
```
ihincks: Obvious to you I'm sure, but we can, in the future, work to eliminate the temporary 8x memory overhead here.

ihincks (commented):
Thanks, these are all looking good. I've gone through it carefully and I don't expect to make any more significant comments.

qiskit/primitives/containers/bit_array.py
t-imamichi: Thank you for writing the note. I like your suggestion `bit_slice`. So, if there is no objection, I will rename this method name. Then, we can skip this note.

t-imamichi:
@ihincks Thank you for reviewing the codes. I will update the codes following your suggestions.
Btw, I'm wondering what are the most intuitive names for various concatenating functions `concatenate`, `stack_bits`, `stack_shots` and slicing function `slice_bits` or `bit_slice`. Do you have suggestions?

qiskit/primitives/containers/bit_array.py
t-imamichi: Retuning an array of expval with the same shape makes sense to me. If users want to compute expvals of a subset of an array, they can slice off a part of their bit array using the new methods introduced in this PR.
Which idea do you have in mind, returning an array of expvals or a data bin of evs and stds as same as Estimator V2 output?

ihincks:
> @ihincks Thank you for reviewing the codes. I will update the codes following your suggestions. Btw, I'm wondering what are the most intuitive names for various concatenating functions `concatenate`, `stack_bits`, `stack_shots` and slicing function `slice_bits` or `bit_slice`. Do you have suggestions?

For a consistent look-feel, I'd like to see:
```
stack_bits -> concatenate_bits
stack_shots -> concatenate_shots
marginalize -> slice_bits
```

This renaming suggests we are additionally missing (?)
```
slice_shots
```
This last one is is low priority, but would be nice if there's time.

qiskit/primitives/containers/bit_array.py
ihincks: I would say just the array. The stds could be derived from the array in the usual way (`1-ev^2` divided by shots etc etc), if the user requires it. Perhaps this could be mentioned in the docstring?


```python
def expectation_values(self, operator: ObservablesLike) -> NDArray[np.float64]:
    """Compute the expectation value the operator for every entry of this bit array.

    Args:
         operator: The operator to take the expectation value of. 
    
    Returns:
        An array of expectation values whose shape matches this bit array's :attr:`~shape`.
    """
```

with the goal of moving to

```python
def expectation_values(self, operator: ObservablesArrayLike) -> NDArray[np.float64]:
    """Compute the expectation values of the provided observables, broadcasted against this bit array.

    Args:
         operator: The operator(s) to take the expectation value of. Must have a shape broadcastable with
            with this bit array.
    
    Returns:
        An array of expectation values whose shape is the broadcast shape of ``operator`` and this bit array.
        
    Raises:
        ValueError: If the provided operator does not have a shape broadcastable with this bit array.
    """
```

I wonder if we could get this second version into 1.1? Probably not enough time.

qiskit/primitives/containers/bit_array.py
t-imamichi: I added this treatment because `np.concatenate` allows `axis=None`, which flattens the array.
https://numpy.org/doc/stable/reference/generated/numpy.concatenate.html

`BitArray.concatenate` behaves almost same as `np.concatenate` except `axis=None`.
I didn't include the behavior because `BitArray.ravel` can do it already and I think it's nice to simplify `axis` by allowing only integers.
But I don't have a strong opinion. I'm fine to remove this treatment.

qiskit/primitives/containers/bit_array.py
t-imamichi: Yes, we need to think of the way to reduce the memory overhead. If it is hard to do it with numpy, rust would be an option.

qiskit/primitives/containers/bit_array.py
t-imamichi: As for the second conditional, I think `axes[0]` exists because `len(axes) == 1`.
If `axes[0]` does not exits, the first conditional catches it.
As for non-permutation case, I intended to rely on `np.transpose`.

t-imamichi:
Thank you for your suggestions of the names. I have a question. There is `concatenate` function w/o suffix (along a bit array axis). Do you think we need to add any suffix?

I renamed method names following Ian's suggestions and updated `expectation_values` to return all expectation values (the first suggestion in https://github.com/Qiskit/qiskit/pull/12158#discussion_r1585330101). I'm thinking of how to extend it to `ObservableArray`.

t-imamichi:
The previous commit c144db63 takes a single observable and  the latest commit a189ce4 allows an observable array.

qiskit/primitives/containers/bit_array.py
t-imamichi: I implemented the observable array version in the latest commit. Could you take a look at it?

t-imamichi:
I will think of slice_shots and write reno tomorrow.

qiskit/primitives/containers/bit_array.py
```diff
@@ -347,3 +372,230 @@ def reshape(self, *shape: ShapeInput) -> "BitArray":
...
+        # unpacking. This could be fixed using bitwise functions, at the expense of a
+        # more complicated implementation.
+        arr = _unpack(self)
+        arr = arr[..., indices]
+        arr, num_bits = _pack(arr)
+        return BitArray(arr, num_bits)
+
+    def expectation_values(self, observables: ObservablesArrayLike) -> NDArray[np.float64]:
+        """Compute the expectation values of the provided observables, broadcasted against
+        this bit arrays.
```
ihincks:
```suggestion
        this bit array.
```

ihincks (commented):
The new expectation values looks good.

t-imamichi:
I added `slice_shots` and reno

qiskit/primitives/containers/bit_array.py
t-imamichi: I made a mistake in the previous comment. I noticed that `BitArray` does not have `ravel` method.
If necessary, I will update `transpose` to accept None or add `ravel` method.

t-imamichi: Ian told me a right way to ravel a bit array. So ignore my previous comments.

ihincks (dismissed):

ihincks (approved):

PR #12159, opened 2024-04-09
warning users about `qiskit-terra` support

1ucian0:
### Summary

The page in https://pypi.org/project/qiskit-terra/ should warning users that `qiskit-terra` is not going to be updated beyond 0.* end-of-life.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,6 @@
 # Qiskit
 
-[![Release](https://img.shields.io/github/v/release/Qiskit/qiskit?sort=semver&filter=0.*&logo=Qiskit&label=qiskit-terra)](https://github.com/Qiskit/qiskit/releases?q=tag%3A0)
+[![Release](https://img.shields.io/github/v/release/Qiskit/qiskit-terra.svg?style=flat&filter=0.*&logo=Qiskit)](https://github.com/Qiskit/qiskit/releases?q=tag%3A0)
 [![License](https://img.shields.io/github/license/Qiskit/qiskit-terra.svg?)](https://opensource.org/licenses/Apache-2.0) <!--- long-description-skip-begin -->
 [![Downloads](https://img.shields.io/pypi/dm/qiskit-terra.svg)](https://pypi.org/project/qiskit-terra/)
 [![Coverage Status](https://coveralls.io/repos/github/Qiskit/qiskit-terra/badge.svg?branch=main)](https://coveralls.io/github/Qiskit/qiskit-terra?branch=main)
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1,13 +1,19 @@
...
+
+[![Release](https://img.shields.io/github/v/release/Qiskit/qiskit-terra.svg?style=flat&filter=0.*&logo=Qiskit)](https://github.com/Qiskit/qiskit/releases?q=tag%3A0)
 [![License](https://img.shields.io/github/license/Qiskit/qiskit-terra.svg?)](https://opensource.org/licenses/Apache-2.0) <!--- long-description-skip-begin -->
-[![Release](https://img.shields.io/github/release/Qiskit/qiskit-terra.svg)](https://github.com/Qiskit/qiskit-terra/releases)
 [![Downloads](https://img.shields.io/pypi/dm/qiskit-terra.svg)](https://pypi.org/project/qiskit-terra/)
 [![Coverage Status](https://coveralls.io/repos/github/Qiskit/qiskit-terra/badge.svg?branch=main)](https://coveralls.io/github/Qiskit/qiskit-terra?branch=main)
-![PyPI - Python Version](https://img.shields.io/pypi/pyversions/qiskit)
+![PyPI - Python Version](https://img.shields.io/pypi/pyversions/qiskit-terra)
 [![Minimum rustc 1.64.0](https://img.shields.io/badge/rustc-1.64.0+-blue.svg)](https://rust-lang.github.io/rfcs/2495-min-rust-version.html)
 [![Downloads](https://static.pepy.tech/badge/qiskit-terra)](https://pepy.tech/project/qiskit-terra)<!--- long-description-skip-end -->
```
jakelishman, line 9: These particular badges don't appear in the PyPI package for older releases (before we moved the package configuration to be static in `pyproject.toml`), so they won't be showing on the PyPI page for `qiskit-terra`.  We used to have dynamic logic that stripped everything between `<!-- long-description-skip-begin -->` and its corresponding "end" when attaching the metadata to the Python distribution.

That said, it's probably fine to change them for the legacy stable branch anyway.

README.md
```diff
@@ -1,13 +1,19 @@
...
 [![DOI](https://zenodo.org/badge/161550823.svg)](https://zenodo.org/badge/latestdoi/161550823)
 
+> [!IMPORTANT]
+> **The package `qiskit-terra` is not going to be updated after August 15th, 2024**. Since `qiskit-terra>=0.25` (released on July 27th, 2023) the package was fully replaced by `qiskit>=0.44`. 
+> If you are installing or depending on `qiskit-terra`, consider changing that to `qiskit`. 
+> [Read more](https://docs.quantum.ibm.com/api/migration-guides/qiskit-1.0-installation#the-old-qiskit-structure).
```
jakelishman: It's not really correct to say that `qiskit>=0.44` _replaced_ `qiskit-terra`: for versions before 1.0, `qiskit-terra` was still absolutely installed.  I'm not sure which of the following two things you're trying to convince people to do:
- change dependencies from `qiskit` to `qiskit-terra`
- upgrade to `qiskit>=1.0`

If "change dependencies", possibly we want to say something like "a dependency on `qiskit-terra` can cause pip to build failed environments and you should swap to depending on `qiskit>=0.x,<1.0`".  If "upgrade to `qiskit>=1.0`" we should say that more explicitly, I guess.

1ucian0: fixed in https://github.com/Qiskit/qiskit/pull/12159/commits/bd2d2b4fab3275d70342346f28c85d2562ae4874

jakelishman (approved):
This looks like a good improvement to the messaging in the legacy README to me, thanks.

PR #11461, opened 2023-12-27
Update target from_configuration

MozammilQ:
### Summary
This PR updates logic of method:`from_configuration()` of class:`Target` in `qiskit/transpiler/target.py`

Credit:
    @nkanazawa1989 , this logic is heavily influenced by  https://github.com/Qiskit/qiskit-ibm-provider/pull/413
@to24toro  , for all the kind help and motivation :)
fixes: #10168

PR #12161, opened 2024-04-09
introduce `default_passmanager` to replace `generate_preset_pass_manager`

1ucian0:
### Summary

This PR introduces `default_passmanager` which is a cleaner version of `generate_preset_pass_manager`, designated to replace `transpile` in 2.0

### Notes:

 * the broken test could be because https://github.com/Qiskit/qiskit/pull/11461 ?


mtreinish:
I'm going to close this PR as I don't think we're pursuing this direction anymore. We re-exported `generate_preset_pass_manager` from the root `qiskit` module and also set 2 as the default value for optimization level. So users can do:

```
from qiskit import generate_preset_pass_manager
generate_preset_pass_manager(backend)
```

or

```
from qiskit import generate_preset_pass_manager
generate_preset_pass_manager(backend.target)
```

which I think covers the use case this was trying to address which was to lower the barrier to constructing a preset pass manager using a good default.

PR #12162, opened 2024-04-09
Fix bad indentation and missing hyphen on `vqd.py`

arnaucasau:
This little PR fixes the docs of two attributes of `qiskit/algorithms/eigensolvers/vqd.py`

The attribute `optimizer` had bad indentation and was split between two bullet points in the list of attributes.

Captura desde 2024-04-08 18-56-13

The attribute `initial_point` was missing a hyphen.

Screenshot 2024-04-09 at 14 40 10

This issue is tracked by **https://github.com/Qiskit/documentation/issues/1120**

### DIFF
--- a/qiskit/algorithms/eigensolvers/vqd.py
+++ b/qiskit/algorithms/eigensolvers/vqd.py
@@ -94,7 +94,7 @@ class VQD(VariationalAlgorithm, Eigensolver):
                 overlap estimation as indicated in the VQD paper.
             ansatz (QuantumCircuit): A parameterized circuit used as ansatz for the wave function.
             optimizer(Optimizer | Sequence[Optimizer]): A classical optimizer or a list of optimizers,
-            one for every k-th eigenvalue. Can either be a Qiskit optimizer or a callable
+                one for every k-th eigenvalue. Can either be a Qiskit optimizer or a callable
                 that takes an array as input and returns a Qiskit or SciPy optimization result.
             k (int): the number of eigenvalues to return. Returns the lowest k eigenvalues.
             betas (list[float]): Beta parameters in the VQD paper.
@@ -102,7 +102,7 @@ class VQD(VariationalAlgorithm, Eigensolver):
                 These hyper-parameters balance the contribution of each overlap term to the cost
                 function and have a default value computed as the mean square sum of the
                 coefficients of the observable.
-            initial point (Sequence[float] | Sequence[Sequence[float]] | None): An optional initial
+            initial_point (Sequence[float] | Sequence[Sequence[float]] | None): An optional initial
                 point (i.e. initial parameter values) or a list of initial points
                 (one for every k-th eigenvalue) for the optimizer.
                 If ``None`` then VQD will look to the ansatz for a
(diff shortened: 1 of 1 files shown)


Eric-Arellano (dismissed):

qiskit/algorithms/eigensolvers/vqd.py
```diff
@@ -94,15 +94,15 @@ class VQD(VariationalAlgorithm, Eigensolver):
...
+                one for every k-th eigenvalue. Can either be a Qiskit optimizer or a callable
                 that takes an array as input and returns a Qiskit or SciPy optimization result.
             k (int): the number of eigenvalues to return. Returns the lowest k eigenvalues.
             betas (list[float]): Beta parameters in the VQD paper.
                 Should have length k - 1, with k the number of excited states.
                 These hyper-parameters balance the contribution of each overlap term to the cost
                 function and have a default value computed as the mean square sum of the
                 coefficients of the observable.
-            initial point (Sequence[float] | Sequence[Sequence[float]] | None): An optional initial
+            initial_point (Sequence[float] | Sequence[Sequence[float]] | None): An optional initial
```
ElePT: Now that the initial point's attribute name is correct, sphinx complains that it shows up both here and as a property. You can just remove lines 170 - 178.

ElePT (commented):
Wow, I had forgotten that qiskit.algorithms was still there in qiskit 0.46. I should probably check and see if the same issue is present in the `VQD` class of `qiskit-algorithms`.  I found a way to fix the lint complaint, but qiskit-neko is always a bit of a mess with 0.46. I think that this might need to be fixed on the neko side, it might be worth looking into how other PRs against 0.46 handle it (Edit: or we can just move on, given that it's not part of the branch protection rules).

qiskit/algorithms/eigensolvers/vqd.py
Eric-Arellano: @ElePT this was a bad change. It's removing actual code that will impact runtime behavior. The better fix is to delete this line of the docs and not mess with the code.

qiskit/algorithms/eigensolvers/vqd.py
```diff
@@ -102,11 +102,6 @@ class VQD(VariationalAlgorithm, Eigensolver):
                 These hyper-parameters balance the contribution of each overlap term to the cost
                 function and have a default value computed as the mean square sum of the
                 coefficients of the observable.
-            initial_point (Sequence[float] | Sequence[Sequence[float]] | None): An optional initial
```
arnaucasau: Docs for the attribute removed :+1: We can still find the same description under the `Args` section (line 140) and we shouldn't have the warning. Thank you both for the review!

arnaucasau:
@ElePT it seems like `qiskit-algorithms` has the same [issue](https://github.com/qiskit-community/qiskit-algorithms/blob/b8315a218e8712dcb14df100e4c4977f39d50966/qiskit_algorithms/eigensolvers/vqd.py#L97). I can fix it there too when this PR is merged :+1:

Eric-Arellano (approved):

ElePT (approved):
Nice, thanks a lot for the effort, LGTM.

PR #12164, opened 2024-04-10
Remove pointless statement exclusion in tests

joesho112358:
### Summary
Noticed that there were some pylint exclusion statements in the tests that check for unused statements. Update this to store in an unused variable and can then remove the exclusion making it a little cleaner.

### Details and comments
Attempt to make things a little cleaner


1ucian0 (approved):
less `pylint: disable` is always good news. Thanks!

PR #12177, opened 2024-04-14
Extend Clifford class docs

### LINKED ISSUE #11330: Constructing a Clifford from an Instruction discards phase
### Environment

- **Qiskit Terra version**: `0.25.3`
- **Python version**: `3.10.12`
- **Operating system**: macOS Ventura 13.6

### What is happening?

If you make a `Clifford` from an `Instruction` which contains a phase, the phase is discarded.

```python
from qiskit.quantum_info import Pauli, Clifford

Clifford(Pauli("-I").to_instruction()) == Clifford(Pauli("I").to_instruction())
>>> True
```

### What should happen?

I am not sure if this quite qualifies as a bug, but it seems like one of these would be helpful:
* Return `False`.
* Raise an error when the phase of the input instruction is non-zero.
* Display a warning when the phase of the input instruction is non-zero.
* Include a statement about this in the documentation somewhere (unless I missed it!).

### Any suggestions?

_No response_

ShellyGarion:
### Summary

Close #11330.
Extend Clifford class docs, to emphasize that a Clifford operator is defined up to a global phase,


iheuze (commented):
This is great but I was thinking maybe add something like this after the equation:
U is a unitary operator in the unitary group U(2^N ), representing operations on N qubits 

Or is U already defined in the doc?

ShellyGarion:
> This is great but I was thinking maybe add something like this after the equation: U is a unitary operator in the unitary group U(2^N ), representing operations on N qubits
> 
> Or is U already defined in the doc?

Thanks @iheuze for your comment! I've updated the docs accordingly.

qiskit/quantum_info/operators/symplectic/clifford.py
```diff
@@ -37,7 +37,21 @@
...
+    An N-qubit unitary operator from the Clifford group.
+
+    An N-qubit Clifford operator takes Paulis to Paulis via conjugation
+    (up to a global phase). More precisely, the Clifford group :math:`\mathcal{C}_N`
+    is defined as
+
+     .. math::
+
+        \mathcal{C}_N = \{ U \in U(2^N) | U \mathcal{P}_N U^{\dagger} = \mathcal{P}_N \} / U(1)
+
```
jakelishman, line 50: What does the ${}/ U(1)$ mean in this context?  I'm familiar with ${}\setminus U(1)$ to mean "subtract U(1)".  From the definition, I'm guessing it's meant to represent normalisation over global phases?

It's totally fine if it's standard notation - I'm probably just not familiar with it, since set theory is not my strong suit.

jakelishman (commented):
Thanks Shelly.

qiskit/quantum_info/operators/symplectic/clifford.py
ShellyGarion: the meaning of this symbol is taking the quotient group, and U(1) represents the scalar unitary matrices, so it generally means "up to a global phase". I'll add some further explanation.

jakelishman: Ah cool, if it's standard notation that's all fine by me.  I'd guessed that it was _something_ to do with the phase-fixing (since $U(1)$ was clearly just the unitary scalars), but I don't always know the set-theory symbols.

jakelishman (approved):

jakelishman:

qiskit/quantum_info/operators/symplectic/clifford.py
ShellyGarion: it's group theory and not set theory :)
as a mathematician one should be very precise :)

PR #12109, opened 2024-03-31
Add support for returning a DAGCircuit to TwoQubitBasisDecomposer

mtreinish:
### Summary

This commit adds a new flag, use_dag, to the constructor for the TwoQubitBasisDecomposer. When set to True, the __call__ method will return a DAGCircuit instead of a QuantumCircuit. This is useful when the two qubit basis decomposer is called from within a transpiler context, as with the UnitarySynthesis pass, to avoid an extra conversion step.

### Details and comments

PR #12179, opened 2024-04-14
Avoid intermediate DAGCircuit construction in 2q synthesis

mtreinish:
### Summary
 
This commit builds on #12109 which added a dag output to the two qubit
decomposers that are then used by unitary synthesis to add a mode of
operation in unitary synthesis that avoids intermediate dag creation.
To do this efficiently this requires changing the UnitarySynthesis pass
to rebuild the DAG instead of doing a node substitution.

### Details and comments

This depends-on #12109 and will need to be rebased after that merges

### DIFF
--- a/qiskit/transpiler/passes/synthesis/unitary_synthesis.py
+++ b/qiskit/transpiler/passes/synthesis/unitary_synthesis.py
@@ -39,6 +39,7 @@ from qiskit.synthesis.two_qubit.xx_decompose import XXDecomposer, XXEmbodiments
 from qiskit.synthesis.two_qubit.two_qubit_decompose import (
     TwoQubitBasisDecomposer,
     TwoQubitWeylDecomposition,
+    GATE_NAME_MAP,
 )
 from qiskit.quantum_info import Operator
 from qiskit.circuit import ControlFlowOp, Gate, Parameter
@@ -293,7 +294,7 @@ class UnitarySynthesis(TransformationPass):
         natural_direction: bool | None = None,
         synth_gates: list[str] | None = None,
         method: str = "default",
-        min_qubits: int = None,
+        min_qubits: int = 0,
         plugin_config: dict = None,
         target: Target = None,
     ):
@@ -499,27 +500,55 @@ class UnitarySynthesis(TransformationPass):
                 ]
             )
 
-        for node in dag.named_nodes(*self._synth_gates):
-            if self._min_qubits is not None and len(node.qargs) < self._min_qubits:
-                continue
-            synth_dag = None
-            unitary = node.op.to_matrix()
-            n_qubits = len(node.qargs)
-            if (plugin_method.max_qubits is not None and n_qubits > plugin_method.max_qubits) or (
-                plugin_method.min_qubits is not None and n_qubits < plugin_method.min_qubits
-            ):
-                method, kwargs = default_method, default_kwargs
+        out_dag = dag.copy_empty_like()
+        for node in dag.topological_op_nodes():
+            if node.op.name == "unitary" and len(node.qargs) >= self._min_qubits:
+                synth_dag = None
+                unitary = node.op.to_matrix()
+                n_qubits = len(node.qargs)
+                if (
+                    plugin_method.max_qubits is not None and n_qubits > plugin_method.max_qubits
+                ) or (plugin_method.min_qubits is not None and n_qubits < plugin_method.min_qubits):
+                    method, kwargs = default_method, default_kwargs
+                else:
+                    method, kwargs = plugin_method, plugin_kwargs
+                if method.supports_coupling_map:
+                    kwargs["coupling_map"] = (
+                        self._coupling_map,
+                        [qubit_indices[x] for x in node.qargs],
+                    )
+                synth_dag = method.run(unitary, **kwargs)
+                if synth_dag is None:
+                    out_dag.apply_operation_back(node.op, node.qargs, node.cargs, check=False)
+                    continue
+                if isinstance(synth_dag, DAGCircuit):
+                    qubit_map = dict(zip(synth_dag.qubits, node.qargs))
+                    for node in synth_dag.topological_op_nodes():
+                        out_dag.apply_operation_back(
+                            node.op, (qubit_map[x] for x in node.qargs), check=False
+                        )
+                    out_dag.global_phase += synth_dag.global_phase
+                else:
+                    node_list, global_phase, gate = synth_dag
+                    qubits = node.qargs
+                    for (
+                        op_name,
+                        params,
+                        qargs,
+                    ) in node_list:
+                        if op_name == "USER_GATE":
+                            op = gate
+                        else:
+                            op = GATE_NAME_MAP[op_name](*params)
+                        out_dag.apply_operation_back(
+                            op,
+                            (qubits[x] for x in qargs),
+                            check=False,
+                        )
+                    out_dag.global_phase += global_phase
             else:
-                method, kwargs = plugin_method, plugin_kwargs
-            if method.supports_coupling_map:
-                kwargs["coupling_map"] = (
-                    self._coupling_map,
-                    [qubit_indices[x] for x in node.qargs],
-                )
-            synth_dag = method.run(unitary, **kwargs)
-            if synth_dag is not None:
-                dag.substitute_node_with_dag(node, synth_dag)
-        return dag
+                out_dag.apply_operation_back(node.op, node.qargs, node.cargs, check=False)
+        return out_dag
 
 
 def _build_gate_lengths(props=None, target=None):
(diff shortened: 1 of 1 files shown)


mtreinish:
I ran some quick asv benchmarks and it's showing a nice modest performance improvement over #12179 

```

       before           after         ratio
     [99c66b44]       [cdaf174a]
     <dag-2q-synth>       <no-dag-2q-synth>
-         311±2ms          281±4ms     0.90  transpiler_levels.TranspilerLevelBenchmarks.time_quantum_volume_transpile_50_x_20(1)
-         107±1ms       96.7±0.7ms     0.90  transpiler_levels.TranspilerLevelBenchmarks.time_schedule_qv_14_x_14(0)
-      86.0±0.4ms         76.4±1ms     0.89  transpiler_levels.TranspilerLevelBenchmarks.time_schedule_qv_14_x_14(1)
-        79.4±2ms       70.3±0.3ms     0.89  transpiler_levels.TranspilerLevelBenchmarks.time_transpile_qv_14_x_14(0)
-      65.8±0.3ms      57.4±0.09ms     0.87  transpiler_levels.TranspilerLevelBenchmarks.time_transpile_qv_14_x_14(1)

       before           after         ratio
     [99c66b44]       [cdaf174a]
     <dag-2q-synth>       <no-dag-2q-synth>
        225±0.9ms        227±0.4ms     1.01  transpiler_levels.TranspilerLevelBenchmarks.time_transpile_from_large_qasm(2)
          229±2ms          230±1ms     1.01  transpiler_levels.TranspilerLevelBenchmarks.time_transpile_from_large_qasm_backend_with_prop(2)
       54.2±0.2ms       54.4±0.1ms     1.00  transpiler_levels.TranspilerLevelBenchmarks.time_transpile_from_large_qasm(0)
             2565             2565     1.00  transpiler_levels.TranspilerLevelBenchmarks.track_depth_quantum_volume_transpile_50_x_20(0)
             1403             1403     1.00  transpiler_levels.TranspilerLevelBenchmarks.track_depth_quantum_volume_transpile_50_x_20(1)
             1403             1403     1.00  transpiler_levels.TranspilerLevelBenchmarks.track_depth_quantum_volume_transpile_50_x_20(2)
             1296             1296     1.00  transpiler_levels.TranspilerLevelBenchmarks.track_depth_quantum_volume_transpile_50_x_20(3)
             2705             2705     1.00  transpiler_levels.TranspilerLevelBenchmarks.track_depth_transpile_from_large_qasm(0)
             2005             2005     1.00  transpiler_levels.TranspilerLevelBenchmarks.track_depth_transpile_from_large_qasm(1)
             2005             2005     1.00  transpiler_levels.TranspilerLevelBenchmarks.track_depth_transpile_from_large_qasm(2)
                7                7     1.00  transpiler_levels.TranspilerLevelBenchmarks.track_depth_transpile_from_large_qasm(3)
             2705             2705     1.00  transpiler_levels.TranspilerLevelBenchmarks.track_depth_transpile_from_large_qasm_backend_with_prop(0)
             2005             2005     1.00  transpiler_levels.TranspilerLevelBenchmarks.track_depth_transpile_from_large_qasm_backend_with_prop(1)
             2005             2005     1.00  transpiler_levels.TranspilerLevelBenchmarks.track_depth_transpile_from_large_qasm_backend_with_prop(2)
                7                7     1.00  transpiler_levels.TranspilerLevelBenchmarks.track_depth_transpile_from_large_qasm_backend_with_prop(3)
              323              323     1.00  transpiler_levels.TranspilerLevelBenchmarks.track_depth_transpile_qv_14_x_14(0)
              336              336     1.00  transpiler_levels.TranspilerLevelBenchmarks.track_depth_transpile_qv_14_x_14(1)
              336              336     1.00  transpiler_levels.TranspilerLevelBenchmarks.track_depth_transpile_qv_14_x_14(2)
              272              272     1.00  transpiler_levels.TranspilerLevelBenchmarks.track_depth_transpile_qv_14_x_14(3)
          114±2ms        114±0.2ms     1.00  transpiler_levels.TranspilerLevelBenchmarks.time_transpile_from_large_qasm(1)
       56.3±0.2ms       56.2±0.3ms     1.00  transpiler_levels.TranspilerLevelBenchmarks.time_transpile_from_large_qasm_backend_with_prop(0)
          116±1ms        116±0.4ms     1.00  transpiler_levels.TranspilerLevelBenchmarks.time_transpile_from_large_qasm_backend_with_prop(1)
        128±0.6ms          127±2ms     0.99  transpiler_levels.TranspilerLevelBenchmarks.time_transpile_from_large_qasm(3)
        129±0.5ms        128±0.2ms     0.99  transpiler_levels.TranspilerLevelBenchmarks.time_transpile_from_large_qasm_backend_with_prop(3)
          1.43±0s       1.40±0.01s     0.97  transpiler_levels.TranspilerLevelBenchmarks.time_quantum_volume_transpile_50_x_20(2)
          273±1ms        266±0.8ms     0.97  transpiler_levels.TranspilerLevelBenchmarks.time_transpile_qv_14_x_14(3)
        244±0.7ms          236±1ms     0.97  transpiler_levels.TranspilerLevelBenchmarks.time_transpile_qv_14_x_14(2)
       1.82±0.04s          1.75±0s     0.96  transpiler_levels.TranspilerLevelBenchmarks.time_quantum_volume_transpile_50_x_20(3)
          815±3ms          784±7ms     0.96  transpiler_levels.TranspilerLevelBenchmarks.time_quantum_volume_transpile_50_x_20(0)

PERFORMANCE INCREASED.
```

mtreinish:
The other question for me is whether we want to have logic to default to the old mode. I did a quick benchmark comparing the runtime of a 1000 gate 2q circuit (I probably should do a deeper example as this isn't super realistic) where it was either a cx gate or a random unitary and did a sweep over the percentages. It looks like there is an crossover point at slightly less than 2%:

comparison

It wouldn't be very hard to compute the percentage of unitary gates in the circuit and use the old code (without this PR in it's current form) if it's `< 0.02` or some other threshold we decide on. That might give us the best of both worlds depending depending on how many unitary gates are in the circuit.

qiskit/synthesis/two_qubit/two_qubit_decompose.py
```diff
@@ -612,26 +648,40 @@ def __call__(
             _num_basis_uses=_num_basis_uses,
         )
         q = QuantumRegister(2)
-        circ = QuantumCircuit(q, global_phase=sequence.global_phase)
-        for name, params, qubits in sequence:
-            try:
-                getattr(circ, name)(*params, *qubits)
-            except AttributeError as exc:
+        if use_dag:
+            from qiskit.dagcircuit.dagcircuit import DAGCircuit
```
ElePT: This was added to avoid the cyclic import complaints, right? In https://github.com/Qiskit/qiskit/pull/12203 complaints I had too many of those and I opted for moving the synthesis imports in `qiskit.circuit.library` to import at runtime.

qiskit/synthesis/two_qubit/two_qubit_decompose.py
```diff
@@ -612,26 +648,40 @@ def __call__(
...
+                        circ.append(self.gate, qubits)
+                    elif name == "u3":
+                        gate = U3Gate(*params)
+                        circ.append(gate, qubits)
+                    elif name == "u2":
+                        gate = U2Gate(*params)
+                        circ.append(gate, qubits)
+                    elif name == "u1":
+                        gate = U1Gate(*params)
+                        circ.append(gate, qubits)
```
ElePT: Is there any reason why you didn't use `GATE_NAME_MAP` here too?

ElePT (commented):
Mostly LGTM, I just left one small comment in the code, and as mentioned in #12195, I like the idea of using the old code to gain a bit of speed on the cases with fewer unitaries. I think it's a use case that might come up when someone builds a benchmark, but if there's no time before the release deadline I also think it's fine to do it as a follow-up.

qiskit/synthesis/two_qubit/two_qubit_decompose.py
mtreinish: This was actually left over from https://github.com/Qiskit/qiskit/pull/12109 . I've rebased the PR now and this isn't here anymore. TBH, I'm not sure why I didn't use the dict here too. I guess my thinking at the time was for the `QuantumCircuit` path we can use the gate methods on the `QuantumCircuit` class except for these 4 cases. But, also I was also likely trying to minimize the diff in #12109 and not really touch the circuit path. I can go back and follow up and deduplicate this in a separate PR if you'd like though.

qiskit/synthesis/two_qubit/two_qubit_decompose.py
mtreinish: I think moving the synthesis imports to runtime make the most sense. I wrote this for #12109 so I don't really remember the exact cause of the import cycle. But I wouldn't have used an import runtime unless I needed to avoid a cycle.

ElePT (approved):
LGTM. The on hold label can be removed now, right?

mtreinish:
Yep, my bad it was leftover from when this was waiting for #12109 to merge.

PR #12181, opened 2024-04-15
Add reverse permutation for LNN connectivity

### LINKED ISSUE #9036: Synthesis algorithms for LNN connectivity
### What should we add?

There are several synthesis algorithms for linear nearest neighbor (LNN) connectivity (that can be embedded into heavy-hex connectivity). These algorithms include:

- [x] Linear (CX) circuits with depth 5n (Kutin et al) #9098 
- [x] Permutations with depth 3n (Kutin et al) #9082 
- [x] CZ circuits with 2q depth 2n+2 (Maslov et al) #9450
- [x] Clifford circuits synthesis by layers (Bravyi and Maslov) #9220 
- [x] CX-CZ circuits with 2q depth 5n (Maslov and Yang) & Clifford synthesis in 2q depth 7n-4 #9932
- [x] StabilizerState synthesis requires only the layers H2 - S1 - CZ1 - H1 - so it has 2q depth 2n+2 (depends on #9543)
- [x] QFT circuits (Fowler et al, Fig 2.b from https://arxiv.org/abs/quant-ph/0402196 ) #11236 
- [x] Reverse permutations in depth 2n+2 (Kutin et al, Section 5, See QFT LNN code) #12181 

These algorithms should be added to `qiskit.synthesis`. 

For completeness, it would be useful to transfer Clifford and CNOT-Dihedral synthesis from `qiskit.quantum_info.synthesis` to `qiskit.synthesis`. 

Further advanced synthesis methods could be added to qiskit ecosystem, and used via the High Level Synthesis plugin, such as the ones suggested in:
https://github.com/Qiskit/qiskit-terra/pull/8548#issuecomment-1255199400
https://github.com/Qiskit/qiskit-terra/issues/8769

ShellyGarion:
### Summary

See #9036 

Add reverse permutations in depth 2n+2 based on Section 5 of Kutin et al. (https://arxiv.org/abs/quant-ph/0701194)

### DIFF
--- a/qiskit/synthesis/permutation/permutation_reverse_lnn.py
+++ b/qiskit/synthesis/permutation/permutation_reverse_lnn.py
@@ -34,6 +34,35 @@ def _append_cx_stage2(qc, n):
     return qc
 
 
+def _append_reverse_permutation_lnn_kms(qc: QuantumCircuit, num_qubits: int) -> None:
+    """
+    Append reverse permutation to a QuantumCircuit for linear nearest-neighbor architectures
+    using Kutin, Moulton, Smithline method.
+
+    Synthesis algorithm for reverse permutation from [1], section 5.
+    This algorithm synthesizes the reverse permutation on :math:`n` qubits over
+    a linear nearest-neighbor architecture using CX gates with depth :math:`2 * n + 2`.
+
+    Args:
+        qc: The original quantum circuit.
+        num_qubits: The number of qubits.
+
+    Returns:
+        The quantum circuit with appended reverse permutation.
+
+    References:
+        1. Kutin, S., Moulton, D. P., Smithline, L.,
+           *Computation at a distance*, Chicago J. Theor. Comput. Sci., vol. 2007, (2007),
+           `arXiv:quant-ph/0701194 <https://arxiv.org/abs/quant-ph/0701194>`_
+    """
+
+    for _ in range((num_qubits + 1) // 2):
+        _append_cx_stage1(qc, num_qubits)
+        _append_cx_stage2(qc, num_qubits)
+    if (num_qubits % 2) == 0:
+        _append_cx_stage1(qc, num_qubits)
+
+
 def synth_permutation_reverse_lnn_kms(num_qubits: int) -> QuantumCircuit:
     """
     Synthesize reverse permutation for linear nearest-neighbor architectures using
@@ -56,10 +85,6 @@ def synth_permutation_reverse_lnn_kms(num_qubits: int) -> QuantumCircuit:
     """
 
     qc = QuantumCircuit(num_qubits)
-    for _ in range((num_qubits + 1) // 2):
-        qc = _append_cx_stage1(qc, num_qubits)
-        qc = _append_cx_stage2(qc, num_qubits)
-    if (num_qubits % 2) == 0:
-        qc = _append_cx_stage1(qc, num_qubits)
+    _append_reverse_permutation_lnn_kms(qc, num_qubits)
 
     return qc
--- a/qiskit/synthesis/permutation/permutation_reverse_lnn.py
+++ b/qiskit/synthesis/permutation/permutation_reverse_lnn.py
@@ -1,0 +1,65 @@
+# This code is part of Qiskit.
+#
+# (C) Copyright IBM 2024
+#
+# This code is licensed under the Apache License, Version 2.0. You may
+# obtain a copy of this license in the LICENSE.txt file in the root directory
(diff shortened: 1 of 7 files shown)


qiskit/synthesis/qft/qft_decompose_lnn.py
```diff
@@ -65,10 +65,7 @@ def synth_qft_line(
     if not do_swaps:
         # Add a reversal network for LNN connectivity in depth 2*n+2,
         # based on Kutin at al., https://arxiv.org/abs/quant-ph/0701194, Section 5.
-        for _ in range((num_qubits + 1) // 2):
-            qc = _append_cx_stage1(qc, num_qubits)
-            qc = _append_cx_stage2(qc, num_qubits)
-        if (num_qubits % 2) == 0:
-            qc = _append_cx_stage1(qc, num_qubits)
+        qc_rev = synth_permutation_reverse_lnn_kms(num_qubits)
+        qc = qc.compose(qc_rev)
```
henryzou50: This looks good, but I think we can do this more efficiently. While the current implementation using compose ensures correctness, it might not be the most efficient in terms of performance and memory usage, especially for larger circuits.

Instead of composing two quantum circuits, we could consider enhancing efficiency by adding a function that directly appends the reverse permutation LNN to an existing circuit. This approach would modify the circuit in place, avoiding the overhead associated with creating and composing a new circuit instance. Here's a conceptual sketch of how this could be implemented:

``` python

def append_reverse_permutation_lnn_kms(qc: QuantumCircuit, num_qubits: int) -> None:
    for _ in range((num_qubits + 1) // 2):
        _append_cx_stage1(qc, num_qubits)
        _append_cx_stage2(qc, num_qubits)
    if (num_qubits % 2) == 0:
        _append_cx_stage1(qc, num_qubits)

File: qiskit/synthesis/permutation/permutation_reverse_lnn.py [hunk 2]
```diff
@@ -0,0 +1,65 @@
...
+    return qc
+
+
+def _append_cx_stage2(qc, n):
+    """A single layer of CX gates."""
+    for i in range(n // 2):
+        qc.cx(2 * i + 1, 2 * i)
+    for i in range((n + 1) // 2 - 1):
+        qc.cx(2 * i + 1, 2 * i + 2)
+    return qc
```
henryzou50, line 34: I noticed `_append_cx_stage1` and `_append_cx_stage2` functions were moved from `qiskit/synthesis/linear_phase/cz_depth_lnn.py` to `qiskit/synthesis/permutation/permutation_reverse_lnn.py`. Could you share the reasoning behind this change?

While I understand these are private functions and the move does not impact their functionality,  I am curious if the new location makes more sense and better aligns with their specific use.

test/python/synthesis/test_permutation_synthesis.py
```diff
@@ -108,6 +112,26 @@ def test_synth_permutation_depth_lnn_kms(self, width):
             synthesized_pattern = LinearFunction(qc).permutation_pattern()
             self.assertTrue(np.array_equal(synthesized_pattern, pattern))
 
+    @data(4, 5, 10, 15, 20)
+    def test_synth_permutation_reverse_lnn_kms(self, num_qubits):
```
henryzou50: Great work on the implementation and testing. The current tests effectively assess the basic functionality and correct usage scenarios. However, I suggest adding tests that focus on:
- Edge Cases: Like scenarios with 0 or 1 qubits.
- Invalid Input: To ensure the system handles invalid inputs as expected.

henryzou50 (commented):
Overall LGTM, but for some improvement in efficiency for `synth_qft_line`, we could directly append the circuit rather than using `compose`, and for the tests we could add some tests for edge cases / invalid input to help ensure robustness of the new feature.

qiskit/synthesis/qft/qft_decompose_lnn.py
ShellyGarion: that's a good suggestion, I added it as a helper function.

qiskit/synthesis/permutation/permutation_reverse_lnn.py
ShellyGarion: These functions are helper functions to construct a reverse permutation, and this is the reason that they have been moved. CZ circuit synthesis for LNN is basically constructing a reverse permutation with phase gates in between.

test/python/synthesis/test_permutation_synthesis.py
ShellyGarion: I added test cases for 1,2 and 3 qubits.

henryzou50 (approved):
LGTM

Issue #9256, opened 2022-12-06
Migrate to `Target` as only backend input for transpiler passes


mtreinish:
### What should we add?

Right now we have a mix of the legacy approach to modeling a backend in the transpiler (`basis_gates`, `CouplingMap`, etc) and the new approach, the `Target` class. This causes friction when adding new passes or maintaining existing ones because we have to support dual paths for everything. We unify all of the internals on the new model because in almost every case we can convert from the legacy model to the target without issue.

- [ ] Add support for `Target` to all transpiler passes that need backend constraints (even if it's just to get a legacy view, like a `CouplingMap` from the target)
    - [ ] #7113 tracked a subset of passes
- [ ] Document that moving forward all new passes only will get a `Target` (both contributor guidelines along with preset pass managers)
- [ ] Update preset pass manager construction to only pass a `Target` to passes
    - Stage plugins initialization will need to fill out all the fields in the `PassManagerConfig`  from the `Target` for backwards compatibility. We can stop doing this as we deprecate non-target constraint fields in `PassManagerConfig`
- [ ] Update `transpile()` to convert `BackendV1` inputs to `BackendV2` with `BackendV2Converter`
- [ ] Update `transpile()` to convert loose input of constraints to a `Target` with `Target.from_configuration()`
    - [ ] constructor added in: https://github.com/Qiskit/qiskit-terra/pull/9255
- [ ] Deprecate anything we previously could represent that is lost in conversion to `Target` (or alternatively augment `Target`'s data model to support these edge cases). The cases currently known where we're unable to convert from the legacy model to the `Target` are:
    -  `coupling_map` without `basis_gates` (represent solely a connectivity constraint).
    
After the above is done and released then we can start the deprecation process for the legacy inputs to the transpiler passes along with the `PassManager` config.

wshanks:
I am trying to understand what the end state is here, based on the description and looking at #12850. I was surprised that #12850 does not deprecate `coupling_map`, `basis_gates`, or `dt`. Is it planned that it will remain possible to pass those parameters to `generate_preset_pass_manager` with or without a `Target`? I can see how it could be convenient to pass them rather than build a target but I don't see why it is necessary if the goal is to move everything to using a target.

mtreinish:
[...]

PR #12183, opened 2024-04-15
Rework handling of instruction durations in preset pass managers

### LINKED ISSUE #10449: Rework handling of instruction durations in preset pass managers
In #10291 we added a hack to serialize the execution of `transpile()` with multiple circuits if a custom `InstructionDurations` or `dt` was specified as arguments to `transpile()`. This was to workaround the behavior that the `InstructionDurations` object constructed by `transpile()` was a function of the input circuit to factor in the durations of custom calibrations in the circuit to that custom `InstructionDurations` object. A better long term fix is to revise the passes using `InstructionDurations` to do the custom circuit calibrations handling inline as needed instead of doing it in `transpile()`.

I do wonder if an implementation of #9256 would alleviate some of those concerns. Because then the passes will only get a target and we're not using `InstructionDurations` anywhere (except for the plugin interface). But I'll convert this comment into an issue for 0.26 and we can track it there.

_Originally posted by @mtreinish in https://github.com/Qiskit/qiskit-terra/pull/10291#discussion_r1268332806_

ElePT:
### Summary
This PR fixes #10449 by removing the "hack" in `transpile` that handled the priority order of different sources of `InstructionDurations` in scheduling passes. The passes now analyze the dag calibrations and set the relevant value themselves, which allows for a much easier use of the standalone passes.

I believe that the change in line 412 (where `num_processes` wasn't used) was a bug that also got fixed in this PR, unless I am missing something. 

### Details and comments
Connected to #9256.

### DIFF
--- a/qiskit/compiler/transpiler.py
+++ b/qiskit/compiler/transpiler.py
@@ -330,7 +330,7 @@ def transpile(  # pylint: disable=too-many-return-statements
 
     _skip_target = False
     _given_inst_map = bool(inst_map)  # check before inst_map is overwritten
-    # If a target is specified have it override any implicit selections from a backend
+    # If a target is specified, have it override any implicit selections from a backend
     if target is not None:
         if coupling_map is None:
             coupling_map = target.build_coupling_map()
@@ -347,7 +347,7 @@ def transpile(  # pylint: disable=too-many-return-statements
         if backend_properties is None:
             backend_properties = target_to_backend_properties(target)
     # If target is not specified and any hardware constraint object is
-    # manually specified then do not use the target from the backend as
+    # manually specified, do not use the target from the backend as
     # it is invalidated by a custom basis gate list, custom coupling map,
     # custom dt or custom instruction_durations
     elif (
@@ -372,6 +372,7 @@ def transpile(  # pylint: disable=too-many-return-statements
     _check_circuits_coupling_map(circuits, coupling_map, backend)
 
     timing_constraints = _parse_timing_constraints(backend, timing_constraints)
+    instruction_durations = _parse_instruction_durations(backend, instruction_durations, dt)
 
     if _given_inst_map and inst_map.has_custom_gate() and target is not None:
         # Do not mutate backend target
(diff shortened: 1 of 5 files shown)


qiskit/transpiler/passes/scheduling/time_unit_conversion.py
```diff
@@ -51,6 +51,7 @@ def __init__(self, inst_durations: InstructionDurations = None, target: Target =
         self.inst_durations = inst_durations or InstructionDurations()
         if target is not None:
             self.inst_durations = target.durations()
+        self._durations_provided = inst_durations is not None or target is not None
```
ElePT, line 54: I had to add this extra variable because unlike the other passes, this instantiates an empty `InstructionDurations`.

releasenotes/notes/rework-inst-durations-passes-28c78401682e22c0.yaml
```diff
@@ -0,0 +1,11 @@
+---
+features_transpiler:
```
mtreinish: Is this more of a fix than a feature? At a high level this change doesn't feel like a user facing feature to document on the release. Or are you thinking that for using the individual passes in isolation this is a new feature that you can have the pass handle the durations in the circuit on their own?

qiskit/compiler/transpiler.py
```diff
@@ -453,7 +409,7 @@ def callback_func(**kwargs):
         optimization_method=optimization_method,
         _skip_target=_skip_target,
     )
-    out_circuits = pm.run(circuits, callback=callback)
+    out_circuits = pm.run(circuits, callback=callback, num_processes=num_processes)
```
mtreinish, line 412: Do you think we should have a fix release note for this? This argument totally wouldn't ever work before, we were setting the flag in the single process path and we weren't setting the argument when the call could be multiprocess.

mtreinish (commented):
Overall this LGTM, it's a pretty straightforward fix just inlining the logic from `transpile()` in the passes that actually need it. I just had some questions about release notes.

qiskit/compiler/transpiler.py
ElePT: Added in [250ff77](https://github.com/Qiskit/qiskit/pull/12183/commits/250ff7744f76c3d928748672d549a0865123e74c)

releasenotes/notes/rework-inst-durations-passes-28c78401682e22c0.yaml
ElePT: I can see that, I found the change somewhere in between a feature and a bugfix, it does change the behavior of the individual passes but probably to the behavior that would have been expected anyways, so I'm fine changing it to a fix.

mtreinish (approved):
LGTM, thanks for the quick updates.

PR #11996, opened 2024-03-12
Update `transpile()` to convert `BackendV1` inputs to `BackendV2` with `BackendV2Converter`

ElePT:
### Summary
This PR is a step forward in unifying the transpiler internals to use the target-base model, as specified in #9256.
I had to fix the behavior of `transpile` with custom scheduling constraints when given `BackendV2` inputs, and extracted this into a separate PR, which should be merged before this one: #12042 [edit: merged]. ~Adding the on hold label again for this reason.~

### Details and comments
The conversion to the target-based model is hidden to the user but this PR includes a couple of small user-facing changes required for the pipeline to work:

- a change in `convert_to_target` that now allows to process calibrations for instructions without registered `InstructionProperties`, instead of skipping these instructions (I believe that this was not the expected behavior)
- a change in `BackendV2Converter` that now looks for `num_qubits` instead of the legacy `n_qubits` in the backend configuration
- allowing backends with no description in `BackendV2Converter`

Note that BackendV2 does not have any internal tracking whether the backend is a simulator, so any "simulator-based test" now uses `BasicSimulator` directly (`test/python/transpiler/test_1q.py`).

In addition to this, some unit tests were using "loopholes" in the `BackendV1` model that wouldn't (and shouldn't) translate to a target properly, such as giving a coupling map that didn't match the number of qubits of the backend (`test/python/transpiler/test_preset_passmanagers.py`) or an instruction schedule map that contained non registered basis gates (`test/python/transpiler/test_pulse_gate_pass.py`). These tests have been adapted to be coherent with the new model.

Issue #12052, opened 2024-03-20
Add `Target` representation to rust


mtreinish:
### What should we add?

An increasing amount of the transpiler is code is now living in rust and with that are core algorithms that need to reason about the various constraints of backends. So far this is just connectivity or various error rates, and we've been handling this by constructing various data containers in rust that contain the subset of information necessary for a particular algorithm. For example `OneQubitGateErrorMap`: https://github.com/Qiskit/qiskit/blob/main/crates/accelerate/src/euler_one_qubit_decomposer.rs#L37 in 1q peephole optimization,  the neighbor table and distance matrix in sabre, and `ErrorMap` for vf2 scoring. In the near future we'll be adding additional algorithms where we'll need different error rate information from the target in rust too (see #11659 and #8774). All of this is basically just working around the lack of a native rust representation of the `Target`. So instead of continuing to add ad-hoc data structures like this instead we should work on making the core data structures of the base `Target` class exist in rust, so that whenever a target is created we automatically have the data in rust.

It's probably not feasible to replace the entire target class with a rust implementation, but we can have the internal data structures be rust structs that contain the data and the python class wraps those with the existing API. This would probably mean the `InstructionProperties` class, and then the rust side of `Target` would replace at least the `._gate_name_map` (the `Operation` objects are still python so this will just contain the raw `PyObject`), `_gate_map`, `_global_operations`, and `_qarg_gate_map` attributes and then provide similar helper methods to update/append to the target and query the data contained in it. This would let us reuse the same target models in both python and rust space and then remove some of these custom data structures we're using for individual algorithms.

nkanazawa1989:
Given the direction of future development Qiskit probably we can purge the calibration property from `InstructionProperties` class in Qiskit2.0? This information is not used unless user wants to convert the entire circuit into a pulse schedule. One of such use cases is the circuit simulation at device Hamiltonian level with Qiskit Dynamics, but today's size of quantum computer is too large to handle with such simulators. I think removal of the calibration property allows you to represent entire Target with Rust object.

Probably, we can revive optional calibration storage like instruction schedule map, but in more efficient fashion (separation of template schedule and parameters as we did in Qiskit Experiments).

raynelfss:
I'd like to take a look at this, could I be assigned?

Issue #11897, opened 2024-02-26
generate_preset_pass_manager as a user-facing function


1ucian0:
Originally, `generate_preset_pass_manager` was meant to be a kinda internal API. However, with the increasing importance of PassManagers construction, the function should be promoted to the prefer way to use the transpiler.

This epic tracks the transformation of `generate_preset_pass_manager` into a more user-friendly function, closer to the `transpile` API.

- [ ] Come up with a better name (`transpiler`? `build_pm`?). It needs to convey what `transpile` is today
- [x] Extend the API to be as close as possible to `transpile`, including accepting build-in objects as parameters.
   - [ ] `default_passmanager` https://github.com/Qiskit/qiskit/pull/12161
- [ ] Update the documentation as the prefer way use the Qiskit transpiler

kevinsung:
The following arguments are present in `transpile` but not `generate_preset_pass_manager` (excluding arguments that are present in `PassManager.run` instead):
- `dt`
- `ignore_backend_supplied_default_methods`

Do we need to add these arguments to `generate_preset_pass_manager`?

It's possible that `dt` has been subsumed by other arguments, such as `target` and/or `timing_constraints`; is this correct?

As for `ignore_backend_supplied_default_methods`, I couldn't quite understand its description from the documentation. Is it still needed?

1ucian0:
> Come up with a better name (`transpiler`? `build_pm`?). It needs to convey what `transpile` is today

For this item, I suggest to slowly replace `StagedPassManager` for `Transpiler` object, which can have a constructor with the `generate_preset_pass_manager` parameters.

Currently, with `generate_preset_pass_manager`: 
```python
from qiskit.transpiler.preset_passmanagers import generate_preset_pass_manager
from qiskit_ibm_runtime.fake_provider import FakeAlmadenV2 

backend = FakeAlmadenV2()
 
pm = generate_preset_pass_manager(backend=backend, optimization_level=1)
type(pm)  # qiskit.transpiler.passmanager.StagedPassManager

pm.run(....)
```

Suggestion:
```python
from qiskit import Transpiler
from qiskit_ibm_runtime.fake_provider import FakeAlmadenV2 

backend = FakeAlmadenV2()
 
transpiler = Transpiler(backend=backend, optimization_level=1)
type(transpiler)  # qiskit.transpiler.passmanager.Transpiler

transpiler.run(...)
```

jakelishman:
[...]

PR #12185, opened 2024-04-16
Update `transpile()` and `generate_preset_pass_manager` to convert loose input of constraints to a `Target` with `Target.from_configuration()`

ElePT:
### Summary
This PR is a step towards a full target-based transpiler pipeline as specified in #9256. It builds up on #11996 and #12183 (should be merged first). The main difficulty of this PR is handling the edge cases found in the legacy model that don't translate well to the target-based model, documented below. For these edge cases, backwards compatibility in the near term is ensured with the "`_skip_target`" flag, but long term support would require in many cases changes in the target model. The cases are documented in detail for future reference.

#### 1. No basis gates but a coupling map
- Current Workaround: `_skip_target`
- Examples:
1. `test_symmetric_coupling_map` in `test/python/transpiler/test_preset_passmanagers.py`:

#### 2. No coupling map but basis gates
- Current Workaround: allow the `Target` to skip the coupling map build if num_qubits is `None`, instead of failing ([here](https://github.com/Qiskit/qiskit/pull/12185/files#diff-00276708f86c726ae04a2c670e74177eaad6aad971e017951bb5792501d13bf5R430) and [here](https://github.com/Qiskit/qiskit/pull/12185/files#diff-00276708f86c726ae04a2c670e74177eaad6aad971e017951bb5792501d13bf5R990-R991)).
- Examples:
1. `test_no_coupling_map` in `test/python/transpiler/test_preset_passmanagers.py`:
2. `test_layout_3239` in `test/python/transpiler/test_preset_passmanagers.py`:
3. `test_unitary_is_preserved_if_in_basis` in `test/python/transpiler/test_preset_passmanagers.py`:

#### 3. No coupling map but basis gates
- Current Workaround: `_skip_target`.
- Examples:
1. `test_no_coupling_map_with_sabre` in `test/python/transpiler/test_preset_passmanagers.py`
2. `test_no_basis_gates` in `test/python/transpiler/test_preset_passmanagers.py`
3. `test_level0_keeps_reset` in `test/python/transpiler/test_preset_passmanagers.py`:
4. `test_initial_layout_fully_connected_cm` in `test/python/transpiler/test_preset_passmanagers.py`:
5. `test_partial_layout_fully_connected_cm` in `test/python/transpiler/test_preset_passmanagers.py`:

#### 1. Unitary synthesis with parametrized gates

- Description: Unitary synthesis has separate methods for when target is provided/not provided. When target is provided, it requires gate re-parametrization, but so far it only re-parametrizes RXX and RZX gates.
- Workaround: This issue was a consequence of the previously proposed workaround. "Fixed" after implementation of `_skip_target`.

   (snippet from unitary_synthesis.py line 693)
   ```python
    # 2q gates sent to 2q decomposers must not have any symbolic parameters.  The
    # gates must be convertable to a numeric matrix. If a basis gate supports an arbitrary
    # angle, we have to choose one angle (or more.)
    def _replace_parameterized_gate(op):
        if isinstance(op, RXXGate) and isinstance(op.params[0], Parameter):
          op = RXXGate(pi / 2)
        elif isinstance(op, RZXGate) and isinstance(op.params[0], Parameter):
          op = RZXGate(pi / 4)
        return op
   ```

- Examples: 

`test_translation_method_synthesis` in `test/python/compiler/test_transpiler.py`:

   ```python
    out = transpile(
      qc,
      translation_method="synthesis",
      basis_gates=basis_gates,
      optimization_level=optimization_level,
      seed_transpiler=42,
      )
   ```
   
   Output:
   
   ```shell
     File "/Users/ept/qiskit_workspace/qiskit/qiskit/transpiler/passes/synthesis/unitary_synthesis.py", line 869, in run
    decomposers2q = self._decomposer_2q_from_target(
     File "/Users/ept/qiskit_workspace/qiskit/qiskit/transpiler/passes/synthesis/unitary_synthesis.py", line 753, in _decomposer_2q_from_target
       supercontrolled_basis = {
     File "/Users/ept/qiskit_workspace/qiskit/qiskit/transpiler/passes/synthesis/unitary_synthesis.py", line 754, in <dictcomp>
       k: v for k, v in available_2q_basis.items() if is_supercontrolled(v)
     File "/Users/ept/qiskit_workspace/qiskit/qiskit/transpiler/passes/synthesis/unitary_synthesis.py", line 738, in is_supercontrolled
       operator = Operator(gate)
     File "/Users/ept/qiskit_workspace/qiskit/qiskit/quantum_info/operators/operator.py", line 97, in __init__
       self._data = self._init_instruction(data).data
     File "/Users/ept/qiskit_workspace/qiskit/qiskit/quantum_info/operators/operator.py", line 700, in _init_instruction
       return Operator(np.array(instruction, dtype=complex))
     File "/Users/ept/qiskit_workspace/qiskit/qiskit/circuit/library/standard_gates/rx.py", line 125, in __array__
       cos = math.cos(self.params[0] / 2)
     File "/Users/ept/qiskit_workspace/qiskit/qiskit/circuit/parameterexpression.py", line 415, in __float__
       raise TypeError(
   TypeError: ParameterExpression with unbound parameters (dict_keys([Parameter(ϴ)])) cannot be cast to a float.
   ```

- Similar cases:

   - `test_translate_ecr_basis` in `test/python/compiler/test_transpiler.py`
     ```python
        res = transpile(
            circuit,
            basis_gates=["u", "ecr"],
            optimization_level=optimization_level,
            seed_transpiler=42,
        )
      ```
     
   - `test_unitary_is_preserved_if_basis_is_None` in `test/python/transpiler/test_preset_passmanagers.py`:
       ```python
       qc = QuantumCircuit(2)
       qc.unitary(random_unitary(4, seed=4242), [0, 1])
       qc.measure_all()
       result = transpile(
         qc, 
         basis_gates=None, 
         optimization_level=level)
       ```
   - `test_unitary_is_preserved_if_basis_is_None_synthesis_transltion` in `test/python/transpiler/test_preset_passmanagers.py`:

       ```python
          qc = QuantumCircuit(2)
          qc.unitary(random_unitary(4, seed=42424242), [0, 1])
          qc.measure_all()
          result = transpile(
            qc, 
            basis_gates=None, 
            optimization_level=level, 
            translation_method="synthesis"
          )
      ```
#### 2. Custom pulse gates
   
- Description: Custom 2q+ pulse gates that are directional may be flipped during the layout step if gate errors are provided (i.e. backend properties), triggering errors in `HighLevelSynthesis` or `GateDirection` (as it doesn't know how to flip them back).

- Workaround: `_skip_target` and not providing backend properties when we are skipping the target.

-  Examples: 

`test_backend_and_custom_gate` in `test/python/compiler/test_transpiler.py`:
    
   ```python
      backend = GenericBackendV2(
                  num_qubits=5,
                  coupling_map=[[0, 1], [1, 0], [1, 2], [1, 3], [2, 1], [3, 1], [3, 4], [4, 3]],
              )
     inst_map = InstructionScheduleMap()
     inst_map.add("newgate", [0, 1], pulse.ScheduleBlock())
     newgate = Gate("newgate", 2, [])
     circ = QuantumCircuit(2)
     circ.append(newgate, [0, 1])
     tqc = transpile(
         circ,
         backend,
         inst_map=inst_map,
         basis_gates=["newgate"],
         optimization_level=opt_level,
         seed_transpiler=42,
     )
   ```
   
   Output:

   ```shell
   qiskit.transpiler.exceptions.TranspilerError: "HighLevelSynthesis was unable to synthesize Instruction(name='newgate', num_qubits=2, num_clbits=0, params=[])."   ```
   ```

#### 3. [FIXED] Control Flow:

- Description: Should we always add control flow to created target?
- Workaround: always add control flow instructions

- Examples:

`test_transpile_control_flow_no_backend` in `test/python/compiler/test_transpiler.py`:
    
   ```python
        qc = QuantumCircuit(QuantumRegister(1, "q"), ClassicalRegister(1, "c"))
        qc.h(0)
        qc.measure(0, 0)
        with qc.if_test((qc.clbits[0], False)):
            qc.x(0)
        with qc.while_loop((qc.clbits[0], True)):
            qc.x(0)
        with qc.for_loop(range(2)):
            qc.x(0)
        with qc.switch(qc.cregs[0]) as case:
            with case(case.DEFAULT):
                qc.x(0)
        qc.measure(0, 0)

        transpiled = transpile(qc, optimization_level=opt_level)
   ```
   
   Output:

   (raised by passmanager)
   ```shell
   qiskit.transpiler.exceptions.TranspilerError: "The control-flow constructs ['for_loop', 'if_else', 'while_loop', 'switch_case'] are not supported by the backend."
   ```

#### 4. Scheduling with no coupling map:

- Description: `Target.from_configuration() `currently skips instruction durations if no coupling map is provided, which causes issues with some scheduling tests.
- Workaround: `_skip_target`

- Examples:

1. `test_circuit_with_delay` in `test/python/compiler/test_transpiler.py`:

```python
 qc = QuantumCircuit(2)
        qc.h(0)
        qc.delay(500, 1)
        qc.cx(0, 1)

  out = transpile(
      qc,
      scheduling_method="alap",
      basis_gates=["h", "cx"],
      instruction_durations=[("h", 0, 200), ("cx", [0, 1], 700)],
      optimization_level=optimization_level,
      seed_transpiler=42,
  )
```

Output:
```sh
qiskit.transpiler.exceptions.TranspilerError: 'Duration of cx on qubits [0, 1] is not found.'
```

#### 5. Backend with custom target & asymmetric coupling map

- Description: The current way we resolve the custom constraints vs. backend priorities require extracting backend (target) properties, resolving the priorities, and re-building a new target. Some special constraints like asymmetric coupling maps are lost in this conversion.
- Workaround: added `_no_loose_constraints` flag, if `True`, the target isn't built from config but directly taken from the backend.

- Examples:

`test_transpile` in `test/python/providers/test_backend_v2.py`:

```python
    self.backend = GenericBackendV2(num_qubits=2, seed=42, basis_gates=["rx", "u"])
    cx_props = {
        (0, 1): InstructionProperties(duration=5.23e-7, error=0.00098115),
    }
    self.backend.target.add_instruction(CXGate(), cx_props)
    ecr_props = {
        (1, 0): InstructionProperties(duration=4.52e-9, error=0.0000132115),
    }
    self.backend.target.add_instruction(ECRGate(), ecr_props)
```

```python
    tqc = transpile(qc, self.backend, optimization_level=opt_level)

### CODE REVIEW

**Comment by sbrandhsn** [comment]
I have comments on 1. and 4. :-)
1.: I think the semantics on 1. would be to only route without touching the gates in the circuit. This would be violated with the current workaround when the input circuit uses a custom gate set. 

4.: I think if we can schedule with the provided user-input, we should just schedule it. :-) This may be relevant for work flows where an user uses external compilers to generate a physical circuit and then just wants to use qiskit for scheduling and so on. If I understand it correctly, we can schedule if the `instruction_durations` is complete even in the absence of a coupling map.

**Comment by ElePT** [comment]
Thanks @sbrandhsn. I see your that your points come from a "transpiler pipeline usability" perspective, which I also believe is important. I guess that my question here is how could we realize these behaviors given the current `Target` object design.

1.: How could we actually implement this behavior? The initial way I see is to allow for `basis_gates==None` in the target, and have the pass manager interpret that as a signal to skip all stages except for routing. However, having basis gates is a fundamental assumption of the `Target` model. The target doesn't have a way to independently store a `coupling_map`, the internal data model uses a `gate_map` that contains both gate and connectivity data. This is allows for more flexibility in the target design, because we can specify connectivity at a gate level (open case 5 shows how a coupling map can't always reflect the connectivity details). With no basis gates we have no gate map, and thus no way to store connectivity. This is an issue I honestly am not sure how to overcome. 

4.: Unlike the previous example, I think that this one could be implemented by navigating the gray area created by `target.num_qubits=None`, it's just a matter of whether we want to do it, and potentially have an input like `instruction_durations` dominate the rest of the inputs in the target construction step (and I understand why @mtreinish has reservations).

The way things are now, we can create a target with no connectivity constraints, that is not an issue. However, this results in the target's `num_qubits` to be `None`, the basis gates' `qargs` to be `None` and all instructions (basis gates) to be considered global. When an instruction is global, **it cannot contain instruction properties (duration, error...)**, because the definition of these properties requires the qubits the instruction is acting on. So let's say we have the example listed above:

```
  out = transpile(
      qc,
      scheduling_method="alap",
      basis_gates=["h", "cx"],
      instruction_durations=[("h", 0, 200), ("cx", [0, 1], 700)],
      optimization_level=optimization_level,
      seed_transpiler=42,
  )
```

Your resulting target will contain 2 instructions (pseudocode): `Instruction(name=h, qargs=None, duration=None)` and `Instruction(name=cx, qargs=None, duration=None)`. So far, the instructions are global, they exist but have no connectivity or properties. To apply instruction durations, the way it's currently implemented in `Target.from_configuration()`, you do the following (pseudocode, the actual code tells apart 1q and 2q gates):

```
# only showing the 1q gate loop for simplicity
for gate in basis_gates:
    for qubit in range(num_qubits):
          ...
          if instruction_durations is not None:
              try:
                  duration = instruction_durations.get(gate, qubit, unit="s")
              except TranspilerError:
                  duration = None

              # Note that to define instruction properties (including durations) you need the qargs!!!
              gate_properties[(qubit,)] = InstructionProperties(
                            duration=duration, error=error, calibration=calibration
                        )

          target.add_instruction(name_mapping[gate], properties=gate_properties, name=gate)
```

As you can see, if `num_qubits` is `None`, you don't even get to run the second loop, and nothing gets added to the target. In fact, there is a previous check that completely skips the whole loop if `coupling_map=None` ([here](https://github.com/Qiskit/qiskit/blob/main/qiskit/transpiler/target.py#L1319)). Now, the only way I could see to implement your suggestion would be to, in the case where `coupling_map is None`,  give `instruction_durations` priority in the target construction loop, inspect it first, construct a `coupling_map` out of it (in this example, the coupling_map would be `[0,1]`) and use this new coupling map (and consequently, new `num_qubits`) in the rest of the target construction logic. 

I wonder how that would affect cases where targets are constructed to deliberately contain global instructions, and for example, what would we want to do in cases where information doesn't match, let's say `basis_gates=['x', 'rx', 'ecr']` but `instruction_durations=[("h", 0, 200), ("cx", [0, 1], 700)],`. Because this solution would give priority to `instruction_durations` .

**Comment by sbrandhsn** [comment]
Sure! I see your point. :-) I guess we could also fail the user when we detect an underspecified `Target` and point them to specialized passmanagers in that error message. For instance, something along the lines of:
```Coupling map without basis gates specified in Target, please use defaultroutingpassmanager if you intended to only route your input circuit(s) or specify basis gates in the input Target.```
I can imagine that a user would appreciate such an early fail more than us accidentally performing transpilation steps they did not intend to be performed (but I'm not too sure about this). :-) `defaultroutingpassmanager` should then do exactly what the routing stage is doing in default passmanager...

File: qiskit/compiler/transpiler.py [hunk 1]
```diff
@@ -335,73 +356,92 @@ def callback_func(**kwargs):
...
-        coupling_map=coupling_map,
-        instruction_durations=instruction_durations,
-        backend_properties=backend_properties,
-        timing_constraints=timing_constraints,
+        basis_gates=basis_gates if _skip_target else None,
+        coupling_map=coupling_map if _skip_target else None,
+        instruction_durations=instruction_durations if _skip_target else None,
+        backend_properties=backend_properties if _skip_target else None,
+        timing_constraints=timing_constraints if _skip_target else None,
+        inst_map=inst_map if _skip_target else None,
```
ElePT: This is not strictly necessary but I added it for debugging (to make sure that the target was actually being used when I wanted it to) and then thought it didn't hurt to have as an additional check. Let me know if it doesn't convince you.

test/python/transpiler/test_sabre_swap.py
```diff
@@ -1329,9 +1329,9 @@ def setUpClass(cls):
         super().setUpClass()
         cls.backend = Fake27QPulseV1()
         cls.backend.configuration().coupling_map = MUMBAI_CMAP
+        cls.backend.configuration().basis_gates += ["for_loop", "while_loop", "if_else"]
```
ElePT, line 1332: I realized that this test wasn't actually updating the backend basis gates, and this made the new target-based path fail with a "control flow gate no found" error. One of the oversights from the old model.

qiskit/transpiler/target.py
```diff
@@ -985,7 +987,8 @@ def instruction_properties(self, index):
 
     def _build_coupling_graph(self):
         self._coupling_graph = rx.PyDiGraph(multigraph=False)
-        self._coupling_graph.add_nodes_from([{} for _ in range(self.num_qubits)])
+        if self.num_qubits is not None:
+            self._coupling_graph.add_nodes_from([{} for _ in range(self.num_qubits)])
```
ElePT, line 991: I added these changes when I was trying to support the edge case with no `coupling_map`. This is no longer necessary to support the edge cases because we are skipping the target, but given that `Target` does technically suport `num_qubits=None`, should we keep these extra checks?

qiskit/compiler/transpiler.py
```diff
@@ -335,73 +356,92 @@ def callback_func(**kwargs):
...
         target.update_from_instruction_schedule_map(inst_map)
 
-    if not ignore_backend_supplied_default_methods:
-        if scheduling_method is None and hasattr(backend, "get_scheduling_stage_plugin"):
-            scheduling_method = backend.get_scheduling_stage_plugin()
-        if translation_method is None and hasattr(backend, "get_translation_stage_plugin"):
-            translation_method = backend.get_translation_stage_plugin()
-
+    # Edge cases require using the old model (loose constraints) instead of building a target,
+    # byt we don't populate the passmanager config with loose constraints unless it's one of
```
raynelfss:
```suggestion
    # but we don't populate the passmanager config with loose constraints unless it's one of

File: test/python/transpiler/test_sabre_swap.py (hunk 2 above)
Comment (raynelfss) [reply test/python/transpiler/test_sabre_swap.py:1332]: Good catch!

**Review by raynelfss — DISMISSED** [review]
LGTM! Thank you for explaining the new behavior with detail and the additions to check python-esque behavior (In the cases where we deal with `NoneType`) are much welcome.

File: qiskit/compiler/transpiler.py (hunk 1 above)
Comment (mtreinish) [reply qiskit/compiler/transpiler.py]: I feel like this would be simpler if we moved this logic down into `generate_preset_pass_manager()` itself. We already expose skip_target there so internally we can change that function to generate the target instead of doing it in transpile. (it might also require moving #11996 to `generate_preset_pass_manager()` too).

File: qiskit/compiler/transpiler.py [hunk 5]
```diff
@@ -40,6 +50,17 @@
 
 _CircuitT = TypeVar("_CircuitT", bound=Union[QuantumCircuit, List[QuantumCircuit]])
 
+CUSTOM_MAPPING = {
+    "if_else": IfElseOp,
```
mtreinish: The control flow instructions aren't handled by `Target.from_configuration()` already? That is arguably a bug in `from_configuration()`.

qiskit/compiler/transpiler.py
```diff
@@ -40,6 +50,17 @@
...
+    "for_loop": ForLoopOp,
+    "while_loop": WhileLoopOp,
+    "switch_case": SwitchCaseOp,
+    "continue": ContinueLoopOp,
+    "break": BreakLoopOp,
+    "unitary": UnitaryGate,
+    "qsd2q": UnitaryGate,
```
mtreinish: Where was this coming up? I'm assuming there was a quantum shannon decomposition test where we passed `qsd2q` to test the intermediate synthesis via `transpile()`?

This actually raises an interesting backwards compatibility question for me. Do we have any tests for the edge case someone does something like:

```python

qc = QuantumCircuit(4)
qc.cx(0, 1)
qc.cz(0, 2)
qc.append(MyCustomGate(), [0, 1, 2, 3])
transpile(qc, basis_gates=["cz", "u", "my_custom_gate"])
```

where `my_custom_gate` is the `name` attribute of `MyCustomGate`? I feel like this is a breaking change caused by this PR because we have no mechanism to do this without providing the custom mapping which we don't have an API for. I'm wondering if we want a fallback for when `Target.from_configuration()` fails with the error because there is no mapping for a gate name to handle this.

mtreinish (commented):
This is looking good, I'm very excited to have this. When coupled with #12052 this will mean we can unify most of the logic from `transpile()`. All the work on mapping the edge cases also gives us an easy way to deprecate these paths for Qiskit 2.0.

I left a few comments inline, but the high level things are I'm thinking we should move the logic down a layer to `generate_preset_pass_manager()` where we already expose `_skip_target` as an argument. This has the advantage of unifying the code paths for target construction. The other thing is that I think there is another edge case of a potential API breakage caused by custom gates with `basis_gates`.

qiskit/compiler/transpiler.py
ElePT: Adding these gates to the custom gate map had been my first attempt to handle these cases, but then I realized exactly what you mentioned (see edge case 2 of open edge cases, lots of examples in the pulse gate pass test too) and I added an edge case for handling custom gates in general (https://github.com/Qiskit/qiskit/pull/12185/files#diff-aad14fa2a652af98a9ab74e163db4992c88481e92ec4211bfcfd0b0c62a80c0aR393-R394). I just forgot to remove them from the name mapping.

ElePT:
> I feel like this would be simpler if we moved this logic down into generate_preset_pass_manager() itself. We already expose skip_target there so internally we can change that function to generate the target instead of doing it in transpile. (it might also require moving https://github.com/Qiskit/qiskit/pull/11996 to generate_preset_pass_manager() too).

@mtreinish Did you mean to move just the `Target.from_configuration()` logic to generate_preset_pass_manager(), or were you also thinking of moving all of the loose constraint parsing too? If we only move `Target.from_configuration()`, it will break when `generate_preset_pass_manager()` is used as a standalone function for all of the documented edge cases (131 unit tests fail immediately).

If we move all of the constraint parsing and target building logic, we will be translating the behavior of `transpile` to `generate_preset_pass_manager()` (which didn't do any input handling before). Given that this is going to be the "new interface", I guess that could make sense. Another small detail is that `generate_preset_pass_manager()` doesn't have a `dt` input argument, so that would have to be added.

mtreinish:
Yeah, I was thinking of the former case as a step towards doing the latter. But it makes sense that we would need to move everything over in one step to fix all the edge case handling. Maybe we should wait on this final PR until 1.2 so we can look at this in a bit more depth?

qiskit/compiler/transpiler.py
mtreinish: Ah ok that makes sense. That wasn't clear to me at first, because this would also potentially be the case in the absence of pulse gates too. Well if we've got handling for it now then there is nothing to worry about on that front. But we should remove this dictionary if it's not being used anymore.

qiskit/compiler/transpiler.py
```diff
@@ -451,13 +492,83 @@ def _log_transpile_time(start_time, end_time):
     logger.info(log_msg)
 
 
+def _parse_basis_gates(basis_gates, backend, inst_map, skip_target):
+    name_mapping = CUSTOM_MAPPING
+    standard_gates = get_standard_gate_name_mapping()
+    control_flow_gates = {
```
mtreinish: There is a static set defined for these names here: https://github.com/Qiskit/qiskit/blob/main/qiskit/circuit/controlflow/__init__.py#L27 Although we shouldn't need to match on "continue" or "break" they only appear inside control flow blocks and we shouldn't ever see them at this level.

qiskit/compiler/transpiler.py
ElePT: True, this is actually not necessary with the current implementation.

ElePT:
> Yeah, I was thinking of the former case as a step towards doing the latter. But it makes sense that we would need to move everything over in one step to fix all the edge case handling. Maybe we should wait on this final PR until 1.2 so we can look at this in a bit more depth?

Yes, unfortunately I don't think it's easy to move things from transpile incrementally because of all of the constraint handling that takes place (plus I'd add it makes it harder for people to understand what is done where, there are a few not so intuitive default generations and it's easier if all the logic relies on either transpile or generate_preset_pm). I have pushed a proposal [d176a1f](https://github.com/Qiskit/qiskit/pull/12185/commits/d176a1fe44899f1f44562e26fccb0d46bc8fbddb) where I moved all the target-building and constraint-parsing parts to `generate_preset_pass_manager` (it took me a bit to make it work). 

On one hand, I think this change would make `generate_preset_pass_manager` more robust to different input combinations, but I am still not 100% sure that we are not changing the previous behavior in some of the edge cases. Especially when a backend and loose constraints are used (I think that the target-based case is ok). Unit tests pass with no issues, but we don't have as many unit tests using `generate_preset_pass_manager` directly as we have using `transpile`. In principle the change works, but I am ok with pushing this PR to 1.2 to make sure we are not overseeing anything important regarding backwards compatibility. Maybe add some extra unit tests using `generate_preset_pass_manager` directly.

I think this PR is in a way now related to #11897.

mtreinish (dismissed):
This LGTM now. I like how we've moved all the input handling to one place. I think we should be good from a backwards compat PoV, but if there is another edge case from the extra input normalization in `generate_preset_pass_manager` we can handle that as part of the normal 1.2 cycle.

mtreinish (approved):

PR #10673, opened 2023-08-20
Adding h, p and u to basic aer

jaygambetta:
### Summary

While statevector in quantum-info replaces most of the need of basic aer it does not replace QasmSimulator nor should it. I personally think we should remove the unitary and statevector from basicaer but that is a longer term issue. I know that we can use the transpile to rewrite into u1, u2, and u3 these simulations but I am giving some beginner demonstrators and not having h, p and u seems a little strange to require an extra line for using these (transpile). Furthermore since we want to depreciate u1,u2, and u3 it would be good to include all basis [h, p], u and [sx, rz] in basic aer. 

### Details and comments

@blakejohnson I can here you while I'm in the plane see this is why we need simulators :-)

PR #12186, opened 2024-04-16
Extend the basis gates of BasicSimulator

### LINKED ISSUE #10852: extend BasicSimulator with standard gates
As part of the discussion in https://github.com/Qiskit/qiskit/pull/10673, we resolved that every standard gate with a fixed amount of qubits can be in BasicSimulator.

1ucian0:
Fixes #10852 

The simulator `BasicSimulator` is python-based simulator included in `qiskit.providers.basic_provider`. After the discussion in https://github.com/Qiskit/qiskit/pull/10673, Qiskit resolved that every standard gate, up to 3 qubits, can be in `BasicSimulator`.

The new basis support was extended to: , ``ccx``, ``ccz``, ``ch``, ``cp``, ``crx``, ``cry``, ``crz``, ``cs``, ``csdg``, ``cswap``, ``csx``, ``cu``, ``cu1``, ``cu3``, ``cx``, ``cy``, ``cz``, ``dcx``, ``delay``, ``ecr``, ``global_phase``, ``h``, ``id``, ``iswap``, ``measure``, ``p``, ``r``, ``rccx``, ``reset``, ``rx``, ``rxx``, ``ry``, ``ryy``, ``rz``, ``rzx``, ``rzz``, ``s``, ``sdg``, ``swap``, ``sx``, ``sxdg``, ``t``, ``tdg``, ``u``, ``u1``, ``u2``, ``u3``, ``unitary``, ``x``, ``xx_minus_yy``, ``xx_plus_yy``, ``y``, and ``z``.


ElePT (commented):
Thanks for the PR @1ucian0, would it make sense to add a few tests to check that the extended support works?

qiskit/circuit/library/standard_gates/x.py
```diff
@@ -584,8 +584,8 @@ def __init__(
 
         Args:
             label: An optional label for the gate [Default: ``None``]
-            ctrl_state: control state expressed as integer,
-                string (e.g.``'110'``), or ``None``. If ``None``, use all 1s.
+            ctrl_state: control state expressed as integer, string (e.g.``"110"``), or ``None``.
```
1ucian0: urelated, just noticed the formatting issue in https://docs.quantum.ibm.com/api/qiskit/qiskit.circuit.library.C3XGate#methods

Screenshot 2024-04-21 at 13 57 55

qiskit/providers/basic_provider/basic_simulator.py
```diff
@@ -617,24 +659,41 @@ def run_experiment(self, experiment: QasmQobjExperiment) -> dict[str, ...]:
                             value >>= 1
                         if value != int(operation.conditional.val, 16):
                             continue
-                # Check if single  gate
                 if operation.name == "unitary":
                     qubits = operation.qubits
                     gate = operation.params[0]
                     self._add_unitary(gate, qubits)
+                elif operation.name in ("id", "u0", "delay"):
```
1ucian0, line 666: I'm not fully sure that `delay` is the same as `id` in `BasicSimulator`

ElePT: Given that `BasicSimulator` doesn't understand pulses, I think that the choice is between accepting `delay` and treating it as an identity or failing if a circuit contains a `delay`. I would vote for accepting it but raising a user warning to err on the side of caution, given that users often don't understand the capabilities of each simulator.

qiskit/providers/basic_provider/basic_simulator.py
```diff
@@ -617,24 +659,41 @@ def run_experiment(self, experiment: QasmQobjExperiment) -> dict[str, ...]:
...
+                    pass
```
ElePT, line 667: What about a message like this one?

```suggestion
                    if operation.name == "delay":
                        warnings.warn(
                            "BasicSimulator cannot simulate gate durations, "
                            "treating delay operation as an identity.",
                            UserWarning,
                            stacklevel=2,
                        )
                    pass
```

ElePT (commented):
Thanks @1ucian0 for adding the tests and handling the edge cases. The PR looks good to me, I just have two minor suggestions. One is an inline suggestion about raising a user warning with delay, and the other one would be to include a transpilation step in the run tests, just to make sure that the pass manager runs successfully in case somebody chooses a BasicSimulator as target. I checked out your PR locally and confirmed that it works so this can also be left to a follow-up.

qiskit/providers/basic_provider/basic_simulator.py
1ucian0: @jakelishman what do you think? Should be `delay` be `id`, or warning, or not accepted?

1ucian0:
> the other one would be to include a transpilation step in the run tests, just to make sure that the pass manager runs successfully in case somebody chooses a BasicSimulator as target.

In e1ad7cb03a1cf975d948fa264f392384f714d112 I added the conversion to target. I feel that checking transpilation is checking too much. If the backend can convert to a target and transpile does something wrong with it, it is a different issue.

ElePT (dismissed):
https://github.com/Qiskit/qiskit/pull/12186#issuecomment-2072744854 testing the target seems fair. I am also not too worried about `delay` because I think that treating it as `id` is the closest we can get with a statevector simulator (so in any case it isn't wrong in my opinion), so in the interest of time I am going to approve.

qiskit/providers/basic_provider/basic_simulator.py
jakelishman: The basic simulator is noiseless, and in noiseless quantum-state evolution a time-delay is the identity op. I think it's fine / correct to accept delay and do nothing.

qiskit/circuit/library/standard_gates/x.py
```diff
@@ -584,8 +584,8 @@ def __init__(
...
+            If ``None``, use all 1s.
```
jakelishman: This will have broken the formatting because the continuation line is unindented, and I don't think it'll have actually fixed the problem, which is that there's no space betwen `e.g.` and the opening ` `` `.

qiskit/providers/basic_provider/basic_provider_tools.py
```diff
@@ -40,42 +63,93 @@ def single_gate_matrix(gate: str, params: list[float] | None = None) -> np.ndarr
...
+    elif gate == "u2":
+        gc = gates.U2Gate
     elif gate == "u3":
         gc = gates.U3Gate
     elif gate == "h":
         gc = gates.HGate
-    elif gate == "u":
-        gc = gates.UGate
     elif gate == "p":
         gc = gates.PhaseGate
```
jakelishman: If we're going to add this many new gates, please can we switch to using a dict lookup to retrieve the gate classes?

qiskit/providers/basic_provider/basic_simulator.py
jakelishman: I think we're fine without a warning; it's completely valid to simulate a delay as an identity if there's no noise in the system.

releasenotes/notes/fixes_10852-e197344c5f44b4f1.yaml
```diff
@@ -0,0 +1,45 @@
+---
+upgrade_providers:
+  - |
+    The :class:`.BasicSimulator` python-based simulator included in :mod:`qiskit.providers.basic_provider`
+    now includes all the standard gates (:mod:`qiskit.circuit.library .standard_gates`) up to 3 qubits. The new basis support includes:
```
jakelishman: This is maybe a feature rather than an upgrade?  Fwiw, I probably wouldn't bother listing all the gates below - "includes all the standard gates in `` :mod:`qiskit.circuit.library` ``" is descriptive enough.

qiskit/circuit/library/standard_gates/x.py
1ucian0: let's try in c35b9cfb4444053df943cb08ba1472fd1046d136

qiskit/providers/basic_provider/basic_provider_tools.py
1ucian0: done in 5cddee7572e4a369a0f26c5d0aaa4fe0c6dd5d36

releasenotes/notes/fixes_10852-e197344c5f44b4f1.yaml
1ucian0: done in e81740e8e26ebecd71043406f1ac33a9c7ed5d47

ElePT (approved):
I think that you have implemented all of @jakelishman's comments, so I will approve again and wait to see if he's got any final remark.

PR #11999, opened 2024-03-12
Finalise support for Numpy 2.0

jakelishman:
### Summary

This commit brings the Qiskit test suite to a passing state (with all optionals installed) with Numpy 2.0.0b1, building on previous commits that handled much of the rest of the changing requirements:

- gh-10890
- gh-10891
- gh-10892
- gh-10897
- gh-11023

Notably, this commit did not actually require a rebuild of Qiskit, despite us compiling against Numpy; it seems to happen that the C API stuff we use via `rust-numpy` (which loads the Numpy C extensions dynamically during module initialisation) hasn't changed.

Close #10889

### Details and comments

The Numpy team are still planning a 2.0.0rc1, so there's still the possibility of a couple more changes to come, but given the tiny magnitude of changes since a few months ago when I last checked the compatibility, it's looking pretty promising for us.

I think there's potentially more latent issues in Qiskit around the changed behaviour of `numpy.array(copy=False)` - that used to mean "only copy if required", but in Numpy 2.0 that means "_do not_ copy", and `copy=None` means "only copy if required".  There aren't any instances of us hitting that in the test suite beyond the one fixed in this PR, but I'd be surprised if there aren't more ways we _could_ hit it.

Marking "on hold" until all our core dependencies have released compatible versions.  Mostly, that's Numpy itself and Scipy, but we _may_ want to consider Matplotlib and Pandas (used transitively by seaborn) since they're used in the `visualization` extra, which is very commonly installed.

I verified this locally by building installing `numpy==2.0.0b1`, then building Scipy, scikit-learn, matplotlib and pandas from source against Numpy 2.0.

I'll probably need to make a separate PR to verify the 0.46.x series works with Numpy 2.0 too.

PR #11023, opened 2023-10-16
Fix deprecated Numpy logic in `NormalizeRXAngles`

jakelishman:
### Summary

This new pass added in gh-10634 uses some deprecated Numpy properties and has some slightly fragile exception-based logic when the required properties can be directly tested.  This code issues warnings with Numpy 1.25+, which is currently not used by CI due to gh-10305.

### Details and comments

This should fix the common errors in #11020.  There's still a Linux / Numpy 1.26.1 failure in unitary synthesis that's new, and it's not immediately clear to me if that's related or not.

I submitted this as a separate PR to #11020 because I'd like the history / CI to record that we're not sacrificing compatibility with Numpy 1.24 to add compatibility with 1.26.

PR #10897, opened 2023-09-25
Fix casting rules changing in Numpy 2

jakelishman:
### Summary

This fixes several places in `quantum_info` that will be affected by the casting-rule changes coming in Numpy 2.0[^1].  There are several other places where the casting rules will change the output types of certain arrays due to the removal of Python-scalar value-dependent casting, but those do not appear to have an observable effect on our behaviour, so I have left them.

It is also becoming invalid to set a scalar entry in a Numpy array with an array-like with a non-zero number of dimensions, even if the sequence is of length 1.

[^1]: https://numpy.org/neps/nep-0050-scalar-promotion.html

### Details and comments

For reference, I found these by running the test suite with Numpy 1.26.0 and the environment variable `NPY_PROMOTION_STATE=weak` set.  I also ran with that set to `weak_and_warn` to see if there were other places that looked suspicious, but that is (predictably) very noisy, and none of the related changes jumped out at me as particularly problematic.

Running with Numpy 1.26.0 (on macOS) runs afoul of #10305, so it's possible I missed a couple of things, but I don't think so.  Unfortunately, a bug in Numpy 1.24.4 means that `np.complex128(1j) * opflow.Z` (as a MWE) segfaults the Python interpreter when running in `NPY_PROMOTION_STATE=weak`, so it's pretty hard work to disentangle the test suite from that well enough to use the old Numpy version with the new casting rules to find errors (plus they've been changing a little over time).

PR #10892, opened 2023-09-22
Replace use of `numpy.lib.scimath` with preferred `numpy.emath`

jakelishman:
### Summary

`numpy.emath` is the preferred alias to `numpy.lib.scimath`, and direct access to the latter is scheduled for removal in Numpy 2.

PR #12189, opened 2024-04-16
[0.46] Finalise support for Numpy 2.0 

jakelishman:
### Summary

This commit brings the Qiskit test suite to a passing state (with all optionals installed) with Numpy 2.0.0b1, building on previous commits that handled much of the rest of the changing requirements:

- gh-10890
- gh-10891
- gh-10892
- gh-10897
- gh-11023

Notably, this commit did not actually require a rebuild of Qiskit, despite us compiling against Numpy; it seems to happen that the C API stuff we use via `rust-numpy` (which loads the Numpy C extensions dynamically during module initialisation) hasn't changed.

The main changes are:

- adapting to the changed `copy=None` and `copy=False` semantics in `array` and `asarray`.
- making sure all our implementers of `__array__` accept both `dtype` and `copy` arguments.

Co-authored-by: Lev S. Bishop <18673315+levbishop@users.noreply.github.com>

---

Update `__array__` methods for Numpy 2.0 compatibility

As of Numpy 2.0, implementers of `__array__` are expected and required to have a signature

    def __array__(self, dtype=None, copy=None): ...

In Numpys before 2.0, the `copy` argument will never be passed, and the expected signature was

    def __array__(self, dtype=None): ...

Because of this, we have latitude to set `copy` in our implementations to anything we like if we're running against Numpy 1.x, but we should default to `copy=None` if we're running against Numpy 2.0.

The semantics of the `copy` argument to `np.array` changed in Numpy 2.0.  Now, `copy=False` means "raise a `ValueError` if a copy is required" and `copy=None` means "copy only if required".  In Numpy 1.x, `copy=False` meant "copy only if required".  In _both_ Numpy 1.x and 2.0, `ndarray.astype` takes a `copy` argument, and in both, `copy=False` means "copy only if required".  In Numpy 2.0 only, `np.asarray` gained a `copy` argument with the same semantics as the `np.array` copy argument from Numpy 2.0.

Further, the semantics of the `__array__` method changed in Numpy 2.0, particularly around copying.  Now, Numpy will assume that it can pass `copy=True` and the implementer will handle this.  If `copy=False` is given and a copy or calculation is required, then the implementer is required to raise `ValueError`.  We have a few places where the `__array__` method may (or always does) calculate the array, so in all these, we must forbid `copy=False`.

With all this in mind: this PR sets up all our implementers of `__array__` to either default to `copy=None` if they will never actually need to _use_ the `copy` argument within themselves (except perhaps to test if it was set by Numpy 2.0 to `False`, as Numpy 1.x will never set it), or to a compatibility shim `_numpy_compat.COPY_ONLY_IF_NEEDED` if they do naturally want to use it with those semantics.  The pattern

    def __array__(self, dtype=None, copy=_numpy_compat.COPY_ONLY_IF_NEEDED):
        dtype = self._array.dtype if dtype is None else dtype
        return np.array(self._array, dtype=dtype, copy=copy)

using `array` instead of `asarray` lets us achieve all the desired behaviour between the interactions of `dtype` and `copy` in a way that is compatible with both Numpy 1.x and 2.x.

### Details and comments

This is the spiritual backport of #11999.


raynelfss (approved):
LGTM, thank you for working on this.

PR #12190, opened 2024-04-16
Add build requirements to `requirements-dev.txt`

jakelishman:
### Summary

We previously have not added the build-only requirements `setuptools` and `setuptools-rust` to `requirements-dev.txt` because the majority of contributors would not to build the Rust components outside of the regular build isolation provided by `pip` or other PEP-517 compatible builders.

Since we are accelerating our use of Rust, and it's more likely that more users will need to touch the Rust components, this adds the build requirements to the developer environment, so it's easier for people to do our recommended

        python setup.py build_rust --inplace --release

for the cases that they want to test against optimised versions of the Rust extensions while still maintaining a Python editable installation.

### Details and comments

We talk about re-building Rust outside the build-system isolation in the contributing guide, and in practice, those of us who do a bunch of Rust development for Qiskit already do this tonnes.  This just makes it a bit easier for contributors to make sure they've got an environment where they can do the same stuff as us.

### DIFF
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -5,6 +5,10 @@
 # never become actual package requirements, but still try to be as relaxed as
 # possible so it's easy to develop multiple packages from the same venv.
 
+# Build Rust directly
+setuptools
+setuptools-rust
+
 # Style
 black[jupyter]~=24.1
 
--- a/tox.ini
+++ b/tox.ini
@@ -25,7 +25,6 @@ passenv =
   QISKIT_IN_PARALLEL
   QISKIT_DOCS_GITHUB_BRANCH_NAME
 deps =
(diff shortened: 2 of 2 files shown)


mtreinish (approved):
I'm fine with this change I guess. As you said in the commit message we had resisted doing this in the past because `setuptools-rust` should be pulled in via the PEP517 compliant builder, and depending on the install build path, the `requirement-dev.txt` never came into play. But realistically this is needed for local development. When I'm working on rust code I typically am running `python setup.py build_rust --inplace` super frequently; and we document this as a pattern in the contributing guide. So it's definitely worth clogging up our dev venvs with an extra 26.6kb wheel (setuptools should already always be installed) even if it's not strictly necessary.

PR #12191, opened 2024-04-16
Removing _name attribute from Parameter

### LINKED ISSUE #10880: Remove `_name` attribute of Parameter
### What should we add?

It's redundant with the `_symbol_expr` attribute. See https://github.com/Qiskit/qiskit/pull/10875#discussion_r1333369933

joesho112358:
### Summary

Attempt to resolve #10880 Remove _name attribute of Parameter

### Details and comments

Removed _name attribute from Parameter and made sure the currently implemented tests passed for the new implementation.


qiskit/circuit/parameter.py
```diff
@@ -155,7 +154,7 @@ def _hash_key(self):
         # expression, so its full hash key is split into `(parameter_keys, symbolic_expression)`.
         # This method lets containing expressions get only the bits they need for equality checks in
         # the first value, without wasting time re-hashing individual Sympy/Symengine symbols.
-        return (self._name, self._uuid)
```
1ucian0, line 158: This change might affect code using picking. Would it be possible to keep the tuple shape the same?

qiskit/circuit/parameter.py
```diff
@@ -165,10 +164,10 @@ def __hash__(self):
     # operation attempts to put this parameter into a hashmap.
 
     def __getstate__(self):
-        return (self._name, self._uuid, self._symbol_expr)
+        return (self._uuid, self._symbol_expr)
```
1ucian0: This change might affect code using picking. Would it be possible to keep the tuple shape the same?

joesho112358: Yes, 1 moment- will fix, run the tests, and hopefully update with no problems

joesho112358: should be good for another look

1ucian0 (approved):
Thanks a lot for taking care of this! Let's merge it!

joesho112358:
@1ucian0 sorry this one fell off my radar. not sure where it failed, but it is updated and might be good for another go

PR #11387, opened 2023-12-08
Add star to linear pre-routing pass

mtreinish:
### Summary

This commit adds a new transpiler pass `StarPreRouting` that adds a dedicated transformation that finds a star graph connectivity subcircuit and replaces it with a linear routing equivalent. This makes the circuit easier for a routing pass to work with.

### Details and comments

TODO:

- [x] Add more testing
- [x] Add final layout tracking and composition with routing.

Co-Authored-By: Sebastian Brandhofer <148463728+sbrandhsn@users.noreply.github.com>

PR #12111, opened 2024-04-01
Add ElidePermutations pass to optimization level 3

alexanderivrii:
### Summary

This is a follow-up of #9523 that actually adds the `ElidePermutations` and the accompanying `FinalizeLayouts` passes to optimization level 3, and removes from there the `OptimizeSwapBeforeMeasure` as it's really just a special case of `ElidePermutations` (moreover, `OptimizeSwapBeforeMeasure` does **not** update the "final_layout" and breaks unitary equivalence between the original and the transpiled circuits).

### Details and comments

Should be rebased on top of #9523 (and in fact, separated from #9523 into a separate PR).

PR #12196, opened 2024-04-17
Add StarPrerouting to the Default Passmanagers

sbrandhsn:
### Summary

Adds the star to linear prerouting pass to optimization level 2 and 3 passmanagers (see #11387)


releasenotes/notes/adding-starprerouting-to-default-passmanager-0fb9c4a86ac2ad0d.yaml
```diff
@@ -0,0 +1,7 @@
+---
+features:
+  - |
+    Added the :class:`StarPreRouting` pass to the default optimization level 2+3 passmanager. `StarPreRouting`
+    runs before a layout is fixed (i.e. on abstract circuits) and transforms a circuit with inner
+    'star'-connectivities to a linear connectivity before continuing to other transpilation stages, e.g.
+    the layouting and routing stage.
```
mtreinish:
```suggestion
    the layout and routing stage.
```

mtreinish (dismissed):
This LGTM in it's current form (leaving an approval here just to indicate this now, but this can't merge until its dependencies do). There will be a merge conflict with #12111 but that should be simple to fix and I'm pretty sure we want to run this pass after `ElidePermutations` since this inserts permutations. :)

releasenotes/notes/adding-starprerouting-to-default-passmanager-0fb9c4a86ac2ad0d.yaml
```diff
@@ -0,0 +1,7 @@
+---
+features:
```
mtreinish:
```suggestion
features_transpiler:
```

sbrandhsn:
Sounds good, thanks! :-)

mtreinish (dismissed):

mtreinish:
I ran some quick asv benchmarks with this, we might want to hold off on adding this to the preset passmanagers for now. It seems to have a pretty big negative impact on performance. It helps on QFT's quality significantly, but for the other circuits it has a big impact on runtime and/or output depth.

```

| Change   |   Before [43c065fc] <add-star-prerouting~3^2> |   After [46aa5cfb] <add-star-prerouting> |   Ratio | Benchmark (Parameter)                                       |
|----------|-----------------------------------------------|------------------------------------------|---------|-------------------------------------------------------------|
| -        |                                          2582 |                                     1540 |    0.6  | utility_scale.UtilityScaleBenchmarks.track_qft_depth('cz')  |
| -        |                                          2582 |                                     1540 |    0.6  | utility_scale.UtilityScaleBenchmarks.track_qft_depth('ecr') |
| -        |                                          2582 |                                     1531 |    0.59 | utility_scale.UtilityScaleBenchmarks.track_qft_depth('cx')  |

| Change   | Before [43c065fc] <add-star-prerouting~3^2>   | After [46aa5cfb] <add-star-prerouting>   | Ratio   | Benchmark (Parameter)                                                         |
|----------|-----------------------------------------------|------------------------------------------|---------|-------------------------------------------------------------------------------|
|          | 2.42±0.01s                                    | 3.18±0.01s                               | ~1.31   | utility_scale.UtilityScaleBenchmarks.time_qaoa('ecr')                         |
|          | 21.5±0.03s                                    | 25.9±0.1s                                | ~1.20   | utility_scale.UtilityScaleBenchmarks.time_qft('cx')                           |
|          | 24.4±0.04s                                    | 26.2±0.1s                                | 1.07    | utility_scale.UtilityScaleBenchmarks.time_qft('cz')                           |
|          | 24.4±0.05s                                    | 25.4±0.1s                                | 1.04    | utility_scale.UtilityScaleBenchmarks.time_qft('ecr')                          |
|          | 9.11±0.05ms                                   | 9.12±0.04ms                              | 1.00    | utility_scale.UtilityScaleBenchmarks.time_parse_qaoa_n100('cz')               |
|          | 98.8±0.2ms                                    | 98.8±0.5ms                               | 1.00    | utility_scale.UtilityScaleBenchmarks.time_parse_qft_n100('cz')                |
|          | 99.1±0.3ms                                    | 99.2±0.7ms                               | 1.00    | utility_scale.UtilityScaleBenchmarks.time_parse_qft_n100('ecr')               |
|          | 9.17±0.03ms                                   | 9.10±0.2ms                               | 0.99    | utility_scale.UtilityScaleBenchmarks.time_parse_qaoa_n100('cx')               |
|          | 9.17±0.04ms                                   | 9.09±0.03ms                              | 0.99    | utility_scale.UtilityScaleBenchmarks.time_parse_qaoa_n100('ecr')              |
|          | 99.9±0.5ms                                    | 99.2±0.3ms                               | 0.99    | utility_scale.UtilityScaleBenchmarks.time_parse_qft_n100('cx')                |
|          | 32.3±0.07ms                                   | 32.1±0.1ms                               | 0.99    | utility_scale.UtilityScaleBenchmarks.time_parse_square_heisenberg_n100('cx')  |
|          | 32.2±0.3ms                                    | 32.0±0.04ms                              | 0.99    | utility_scale.UtilityScaleBenchmarks.time_parse_square_heisenberg_n100('cz')  |
|          | 32.5±0.2ms                                    | 32.2±0.1ms                               | 0.99    | utility_scale.UtilityScaleBenchmarks.time_parse_square_heisenberg_n100('ecr') |
|          | 444                                           | 414                                      | 0.93    | utility_scale.UtilityScaleBenchmarks.track_square_heisenberg_depth('cx')      |
|          | 444                                           | 414                                      | 0.93    | utility_scale.UtilityScaleBenchmarks.track_square_heisenberg_depth('cz')      |
|          | 444                                           | 414                                      | 0.93    | utility_scale.UtilityScaleBenchmarks.track_square_heisenberg_depth('ecr')     |

Benchmarks that have got worse:

| Change   | Before [43c065fc] <add-star-prerouting~3^2>   | After [46aa5cfb] <add-star-prerouting>   |   Ratio | Benchmark (Parameter)                                              |
|----------|-----------------------------------------------|------------------------------------------|---------|--------------------------------------------------------------------|
| +        | 1.28±0.02s                                    | 4.13±0.02s                               |    3.22 | utility_scale.UtilityScaleBenchmarks.time_square_heisenberg('cx')  |
| +        | 1.78±0.01s                                    | 4.61±0.03s                               |    2.59 | utility_scale.UtilityScaleBenchmarks.time_square_heisenberg('ecr') |
| +        | 1.93±0.04s                                    | 4.80±0.01s                               |    2.49 | utility_scale.UtilityScaleBenchmarks.time_square_heisenberg('cz')  |
| +        | 1.13±0.01s                                    | 1.87±0.01s                               |    1.66 | utility_scale.UtilityScaleBenchmarks.time_qaoa('cx')               |
| +        | 1607                                          | 2331                                     |    1.45 | utility_scale.UtilityScaleBenchmarks.track_qaoa_depth('cx')        |
| +        | 1622                                          | 2331                                     |    1.44 | utility_scale.UtilityScaleBenchmarks.track_qaoa_depth('cz')        |
| +        | 1622                                          | 2331                                     |    1.44 | utility_scale.UtilityScaleBenchmarks.track_qaoa_depth('ecr')       |
| +        | 2.92±0.02s                                    | 3.76±0.02s                               |    1.29 | utility_scale.UtilityScaleBenchmarks.time_qaoa('cz')               |

PERFORMANCE DECREASED.
```

qiskit/transpiler/passes/optimization/elide_permutations.py
```diff
@@ -105,8 +105,10 @@ def _apply_mapping(qargs):
             self.property_set["original_qubit_indices"] = input_qubit_mapping
 
         new_layout = Layout({dag.qubits[out]: idx for idx, out in enumerate(qubit_mapping)})
-        if current_layout := self.property_set["virtual_permutation_layout"] is not None:
-            self.property_set["virtual_permutation_layout"] = current_layout.compose(new_layout)
+        if current_layout := self.property_set["virtual_permutation_layout"]:
+            self.property_set["virtual_permutation_layout"] = new_layout.compose(
+                current_layout.inverse(dag.qubits, dag.qubits), dag.qubits
+            )
```
alexanderivrii, line 111: Oh right, this seems to be the correct order for composing final permutations, thanks for fixing this! If our circuit is `C -- P`, with `P` implicitly tracked as a "final permutation", and then we further rewrite `C` as `D -- Q`, then the original circuit becomes `D -- Q -- P`, i.e. we now want to implicitly track the "final permutation" for `Q -- P`.

qiskit/transpiler/preset_passmanagers/builtin_plugins.py
```diff
@@ -778,6 +783,8 @@ def _swap_mapped(property_set):
                 skip_routing=pass_manager_config.routing_method is not None
                 and pass_manager_config.routing_method != "sabre",
             )
+            layout.append(ConditionalController(StarPreRouting(), condition=_vf2_match_not_found))
+            layout.append(ConditionalController(choose_layout_0, condition=_star_prerouting_check))
```
alexanderivrii, line 787: Yeah, very good point that `StarPreRouting` may in cases make the circuit harder to route (and in cases to destroy the perfect match).

alexanderivrii (approved):
I am approving this, though looking at the ASV results it is indeed not clear that we want the `StarPreRouting` pass to run by default. 

It does seem that a more thorough investigation of when and how to run the pass, e.g.  maybe the pass would be beneficial if we only rewrite the "large stars", or only for low-degree coupling maps. 

But in any case, it would be good to include the fix to composing virtual pre-layouts.

qiskit/transpiler/passes/optimization/elide_permutations.py
alexanderivrii: This leads to a very interesting general question. If we consider SWAPs in the "final permutation" part of the circuit to be free (come at no cost), then what is the best way to "massage" a quantum circuit (e.g. inserting/deleting SWAPs and suitably changing the following gates) to make it easier to route.

mtreinish:
Given the open performance questions on this PR I think it's prudent for us to just defer this until the next release. The depth improvements for qft and also the square heisenberg circuit look really nice. But the qaoa circuit is concerning enough that we shouldn't rush into doing this. I'll split out the layout composition fix into a separate PR that we can include in 1.1.0 and in parallel lets investigate adding it to the preset pass managers.

ShellyGarion:
> Given the open performance questions on this PR I think it's prudent for us to just defer this until the next release. The depth improvements for qft and also the square heisenberg circuit look really nice. But the qaoa circuit is concerning enough that we shouldn't rush into doing this. I'll split out the layout composition fix into a separate PR that we can include in 1.1.0 and in parallel lets investigate adding it to the preset pass managers.

I think that QAOA has a good swap strategies algorithm by https://arxiv.org/abs/2202.03459.
I think it was implemented in Qiskit as part of the evolution gates synthesis (but not in general).

mtreinish:
I'm going to close this PR because I don't think we're planning to do this any time soon. Linear routing a star interaction doesn't seem to be an improvement in all cases at least in the approach this and if we want to integrate this into the preset pass manager we'll need to do more discovery around how to best do this.

PR #12197, opened 2024-04-17
Oxidize the numeric code in the Isometry gate class

mtreinish:
### Summary

This commit ports the numeric portion of the Isometry gate class to rust. While this will likely improve the performance slightly this move is more to make isolate this code from blas/lapack in numpy. We're hitting some stability issues on arm64 mac in CI and moving this code to rust should hopefully fix this issue. As this is more for functional reasons no real performance tuning was done on this port, there are likely several opportunities to improve the runtime performance of the code.

### Details and comments


crates/pyext/src/lib.rs
```diff
@@ -15,10 +15,11 @@ use pyo3::wrap_pymodule;
...
-    optimize_1q_gates::optimize_1q_gates, pauli_exp_val::pauli_expval, results::results,
-    sabre::sabre, sampled_exp_val::sampled_exp_val, sparse_pauli_op::sparse_pauli_op,
-    stochastic_swap::stochastic_swap, two_qubit_decompose::two_qubit_decompose, utils::utils,
+    error_map::error_map, euler_one_qubit_decomposer::euler_one_qubit_decomposer,
+    isometry::isometry, nlayout::nlayout, optimize_1q_gates::optimize_1q_gates,
+    pauli_exp_val::pauli_expval, results::results, sabre::sabre, sampled_exp_val::sampled_exp_val,
+    sparse_pauli_op::sparse_pauli_op, stochastic_swap::stochastic_swap,
+    two_qubit_decompose::two_qubit_decompose, uc_gate::uc_gate, utils::utils,
     vf2_layout::vf2_layout,
 };
```
jakelishman, line 24: This kind of stuff really makes me appreciate `black`'s approach to lists as "do what'll minimise git diffs" haha.

crates/accelerate/src/isometry.rs
```diff
@@ -0,0 +1,366 @@
...
+        if !abs_diff_eq!(
+            gate.mapv(|x| x * global_phase),
+            aview2(&ONE_QUBIT_IDENTITY),
+            epsilon = 1e-8 // Default tolerance from numpy for allclose()
+        ) {
+            return false;
+        }
+    }
+    true
+}
```
jakelishman, line 291: This is pre-existing, but `global_phase` is _not_ the global phase until its `abs` happens to be 1.  This test is actually testing whether the gate is close to a _scaled_ identity.  Same with` diag_is_identity_up_to_global_phase` below.

crates/accelerate/src/isometry.rs
```diff
@@ -0,0 +1,366 @@
...
+            [state[0].conj() / r, state[1].conj() / r],
+            [-state[1] / r, state[0] / r],
+        ]
+    } else {
+        array![
+            [-state[1] / r, state[0] / r],
+            [state[0].conj() / r, state[1].conj() / r],
+        ]
+    }
+}
```
jakelishman: In these functions, `state` is guaranteed to always be compatible with `[Complex64; 2]`.  It's probably worth encoding that in the signatures to a) avoid heap allocations coming from Python and b) allow more aggressive loop-unrolling in the calls to `l2_norm` (even if that stays as `&[Complex64]`).

No particular thoughts on pass-by-value `[Complex64; 2]` or pass-by-reference; feels like it's a borderline case at 32 bytes wide.

jakelishman: There's classic micro-optimisation stuff around `a/r, b/r, c/r, d/r` being slower than `rinv = 1/r; a*rinv, b*rinv, c*rinv, d*rinv` which is an invalid floating-point transformation for an optimising compiler but mathematically sound for all cases we care about, but there's larger fish to fry in the whole isometry implementation if we were going that route.

crates/accelerate/src/isometry.rs
```diff
@@ -0,0 +1,366 @@
...
+
+#[pyfunction]
+pub fn find_squs_for_disentangling(
```
jakelishman, line 76:
```suggestion
/// This method finds the single-qubit gates for a UCGate to disentangle a qubit:
/// we consider the n-qubit state v[:,0] starting with k zeros (in the computational basis).
/// The qubit with label n-s-1 is disentangled into the basis state k_s(k,s).

#[pyfunction]
pub fn find_squs_for_disentangling(
```
(copying in the comment from above this function in Python space)

crates/accelerate/src/isometry.rs
```diff
@@ -0,0 +1,366 @@
...
+        let diag_index = action_qubit_labels
+            .iter()
+            .fold(0_usize, |acc, i| (acc << 1) + state[*i] as usize);
+        let i = bin_to_int(&state);
+        for j in 0..num_col {
+            m[[i, j]] = diag[diag_index] * m[[i, j]]
+        }
+    }
+    m.into_pyarray_bound(py).into()
+}
```
jakelishman: I think we can change `diag` to take a Numpy array by reference - the Python-space code only ever calls this one, and it's with a Numpy array except that it calls `tolist` on it for no reason.

Same comment for `apply_diagonal_gate_to_diag` below.

crates/accelerate/src/isometry.rs
```diff
@@ -0,0 +1,366 @@
...
+    let num_qubits = shape[0].ilog2();
+    let num_col = shape[1];
+    for state in std::iter::repeat([0_u8, 1_u8])
+        .take(num_qubits as usize)
+        .multi_cartesian_product()
+    {
+        let diag_index = action_qubit_labels
+            .iter()
+            .fold(0_usize, |acc, i| (acc << 1) + state[*i] as usize);
+        let i = bin_to_int(&state);
```
jakelishman, line 162: Completely pre-existing, but holy moly.  This is a _wild_ way of spelling `for i in 0..2.pow(num_qubits) {}` (and swapping to corresponding bit-shift logic on `diag_index`).

No need to change it here unless you felt like it - the whole implementation needs a bit more TLC than that.

crates/accelerate/src/isometry.rs
```diff
@@ -0,0 +1,366 @@
...
+            e2 <<= 1;
+            e2 += 1;
+        } else {
+            assert!(j <= 1);
+            e1 <<= 1;
+            e1 += state_free[j] as usize;
+            e2 <<= 1;
+            e2 += state_free[j] as usize;
+            j += 1
+        }
```
jakelishman: Possibly for clarity it would help to pull the `e1 <<= 1; e2 <<= 1` outside the `if/elif` chain so it's more obvious that the difference in paths is just about the last digit.

crates/accelerate/src/isometry.rs
```diff
@@ -0,0 +1,366 @@
...
+    mut control_labels: Vec<usize>,
+    target_label: usize,
+    gate: PyReadonlyArray2<Complex64>,
+) -> PyObject {
+    let mut m = m.as_array().to_owned();
+    let gate = gate.as_array();
+    let shape = m.shape();
+    let num_qubits = shape[0].ilog2();
+    let num_col = shape[1];
+    control_labels.par_sort();
```
jakelishman: I don't think there's any reason to sort the control labels - all we do is turn them into a set, which tbh we probably ought to just do once in this function rather than inside `construct_basis_states`.

crates/accelerate/src/isometry.rs
```diff
@@ -0,0 +1,366 @@
...
+                .take(2)
+                .collect();
+            m[[e1, i]] = temp[0];
+            m[[e2, i]] = temp[1];
+        }
+        return m.into_pyarray_bound(py).into();
+    }
+    for state_free in std::iter::repeat([0_u8, 1_u8])
+        .take(free_qubits)
+        .multi_cartesian_product()
```
jakelishman: A bit inconvenient that the product of "empty iterator" doesn't produce the unit tuple in this form - I think Python's `itertools` feels more mathematically natural here.

If we really wanted to avoid the large code duplication we _could_ pull out the block of the if into a
```rust
fn apply<T: IntoIterator<Item = [usize]>>(...) {}
```
(or whatever the right iterator item is) and call it twice, but I don't really think it's worth bothering.

qiskit/circuit/library/generalized_gates/isometry.py
```diff
@@ -341,15 +334,8 @@ def inv_gate(self):
...
-    r = np.linalg.norm(state)
-    if r < epsilon:
-        return np.eye(2, 2)
-    if basis_state == 0:
-        m = np.array([[np.conj(state[0]), np.conj(state[1])], [-state[1], state[0]]]) / r
-    else:
-        m = np.array([[-state[1], state[0]], [np.conj(state[0]), np.conj(state[1])]]) / r
-    return m
+    res = isometry_rs.reverse_qubit_state(state, basis_state, epsilon)
+    return res
```
jakelishman: Ought we to just replace the call sites of this Python function with the Rust one?

Comment repeats for the next few functions.

crates/accelerate/src/uc_gate.rs
```diff
@@ -0,0 +1,160 @@
...
+    let mut diag: Array1<Complex64> = Array1::ones(2_usize.pow(num_qubits));
+    let num_controls = num_qubits - 1;
+    for dec_step in 0..num_controls {
+        let num_ucgs = 2_u32.pow(dec_step);
+        // The decomposition works recursively and the followign loop goes over the different
+        // UCGates that arise in the decomposition
+        for ucg_index in 0..num_ucgs {
+            let len_ucg = 2_u32.pow(num_controls - dec_step);
+            for i in 0..len_ucg / 2 {
+                let shift = ucg_index * len_ucg;
```
jakelishman: Super super minor, but I think `len_ucg`, `i` and `shift` are only ever used as indices.  If so, why is `len_ucg` a `u32` here, needing casts to `usize` everywhere?

crates/accelerate/src/uc_gate.rs
```diff
@@ -0,0 +1,160 @@
...
+                // Now we decompose the gates D as described in Figure 4 in
+                // https://arxiv.org/pdf/quant-ph/0410066.pdf and merge some of the gates
+                // into the UCGates and the diagonal at the end of the circuit
+
+                // Remark: The Rz(pi/2) rotation acting on the target qubit and the Hadamard
+                // gates arising in the decomposition of D are ignored for the moment (they will
+                // be added together with the C-NOT gates at the end of the decomposition
+                // (in the method dec_ucg()))
+                let rz_11 = (-Complex64::new(0., 0.5 * PI2)).exp();
+                let rz_00 = Complex64::new(0., 0.5 * PI2).exp();
```
jakelishman: These numbers are actually just $-i$ and $i$, which makes it all the funnier to me that the Python-space version defined a full "get the RZ matrix" method lol

jakelishman (commented):
As best as I could tell, this is a totally faithful port.  I highlighted a couple of places where we could probably improve the calling conventions, but I mostly just left suggestions at that (and  a couple of places where I couldn't resist commenting on stuff), since I think really the whole implementation could do with a rather more complete reworking, and it's not a worthwhile use of time to try that as part of this porting PR.

qiskit/circuit/library/generalized_gates/isometry.py
mtreinish: Oh, that's a good point, I wasn't exactly sure how much of these were going to end up in rust when I was developing this. Having the python side was good for debugging test failures because I could inject prints that went to the python stdout buffer, but yeah let's drop the extra call overhead , I'll remove all these.

crates/accelerate/src/uc_gate.rs
mtreinish: Lol, oh I didn't even see that I just was in mechanical porting mode. Do you think it's worth making that explicit here, or just leave it for LLVM to optimize?

crates/accelerate/src/uc_gate.rs
mtreinish: There was somewhere I thought it was being used as a u32 and why I picked that. But I might have gotten it mixed up with another integer. I'll try to standardize on usize and remove the casts.

crates/accelerate/src/uc_gate.rs
jakelishman: I'm fine leaving all the little things like this - larger-scale _real_ performance improvements in this would come from a more complete refactor of the implementation as a whole, and as you say, the compiler can probably constant-fold this anyway.

crates/accelerate/src/uc_gate.rs
mtreinish: Fixed in: https://github.com/Qiskit/qiskit/pull/12197/commits/3676469a1000c5e980ae4013c97981f3c6bbf0bf

qiskit/circuit/library/generalized_gates/isometry.py
mtreinish: Removed in: https://github.com/Qiskit/qiskit/pull/12197/commits/6e86efe470ba617e014aabd4c350c543ef5ecaf5

crates/accelerate/src/isometry.rs
mtreinish: Done in: https://github.com/Qiskit/qiskit/pull/12197/commits/366a934f9a33c4528bb375dedbf6749be2409b70

crates/accelerate/src/isometry.rs
mtreinish: Done in: https://github.com/Qiskit/qiskit/pull/12197/commits/29ffc994fd51c2f5d4fa8162bfa3a9966bbc70e5

crates/accelerate/src/isometry.rs
mtreinish: Done in: https://github.com/Qiskit/qiskit/pull/12197/commits/091d229957271de68904f5b3141364915585f8d1 there were some cases where the `diag` type was a `f64` instead of a complex (not sure where it came from), but I just added a conversion python side.

crates/accelerate/src/isometry.rs
mtreinish: Done in: https://github.com/Qiskit/qiskit/pull/12197/commits/4843ef5c41652afa6488d49d60cd75cfbd5a52e1 and I couldn't resist the micro optimization while I was touching this function in: https://github.com/Qiskit/qiskit/pull/12197/commits/d26bd4aadbac6001317ed8ecbb3e16c648cd2483

crates/accelerate/src/isometry.rs
jlapeyre: Maybe `global_phase` should be renamed `global_scale` if it can't be constrained always to have magnitude `1`. And likewise, this function should be renamed. But maybe these can wait for a rewrite.

crates/accelerate/src/isometry.rs
```diff
@@ -0,0 +1,367 @@
...
+    basis_state: usize,
+    epsilon: f64,
+) -> PyObject {
+    reverse_qubit_state_inner(&state, basis_state, epsilon)
+        .into_pyarray_bound(py)
+        .into()
+}
+
+#[inline(always)]
+fn l2_norm(vec: &[Complex64]) -> f64 {
```
jlapeyre, line 43: I'd prefer that this were somehow called the `2-norm` or `p-norm` for `p=2`  ("somehow" with a valid identifier).  `l2` is a space, usually a sequence space. But upon googling for uses of p norm, l2 norm, etc. it seems that, as far as the internet is concerned, especially machine-learning internet, lp, Lp, can mean whatever you want them to mean.  [EDIT. looks like this is translated from `np.linalg.norm`, which, along with Julia's `LinearAlgebra`, calls this the 2-norm)]

We might want a function for p norms with p as a parameter, with 2 as the default, and make sure constants are propagated, or whatever it takes to get the correct norm at compile time. But, unless we have an existing use for `p != 2` that can be postponed indefinitely.

In any case, at some point, we will wish that functions like this had been collected in a repo-wide module. (probably not in the ten thousandth "utils.xxx"). And in that case it's worth having somewhat more general trait bounds. Not repeating this function could be important for correctness and performance. For example, the [Julia 2-norm function](https://github.com/JuliaLang/julia/blob/68da780b93518204d874410307791702d5200e29/stdlib/LinearAlgebra/src/generic.jl#L468-L494) appears to sometimes scale the elements to avoid overflow. However, the function `mynorm(v) = sqrt(sum(x -> x^2, v));` is more than 3 times faster than top-level entry point `norm`.  We'd want an api and implementation that reflects our common uses, which might be almost all 2- and 4-element vectors.  Indeed, an optimization, which we don't need to make at the moment, would take advantage the length of the vector, which is known here at compile time to be two.

The main point for now is to consider at least collecting these somewhere where they are visible so that the questions of performance and correctness can be more easily addressed at some point

crates/accelerate/src/isometry.rs
```diff
@@ -0,0 +1,367 @@
...
+    vec.iter()
+        .fold(0., |acc, elem| acc + elem.abs().powi(2))
```
jlapeyre, line 45:
```suggestion
        .fold(0., |acc, elem| acc + elem.norm_sqr())
```
This is more idiomatic and performant.

crates/accelerate/src/isometry.rs
```diff
@@ -0,0 +1,367 @@
...
+        .sqrt()
```
jlapeyre:
```suggestion
        .fold(0., |acc, elem| acc + elem.norm_sqr())
```
This is more idiomatic and performant.

jlapeyre:
> Add back `sqrt()` accidentally removed by inline suggestion

Sorry. My mistake. I originally did not include the `sqrt()` in the edit. But then I went back and "corrected" it.

jlapeyre:
This looks good to me. The only thing I would change is these
```rust
                let rz_11 = (-Complex64::new(0., 0.5 * PI2)).exp();
                let rz_00 = Complex64::new(0., 0.5 * PI2).exp();
```
I know we are leaving a lot of things that could be improved, but this is such an easy and reasonable change, I don't see why not.

crates/accelerate/src/isometry.rs
mtreinish: I don't want to get into a bikeshedding conversation about naming, but `l2_norm` or `lp_norm` is a pretty standard terminology for this. If you look at the comment in the Julia code below what you linked they use it there too:

https://github.com/JuliaLang/julia/blob/68da780b93518204d874410307791702d5200e29/stdlib/LinearAlgebra/src/generic.jl#L496

```julia
# Compute L_p norm ‖x‖ₚ = sum(abs(x).^p)^(1/p)
```

That being said the response to the main comment here is that we can look at creating a dedicated mathematical functions module, or even a separate crate if it's generic enough for these kind of things in a follow up PR. Ideally we'd contribute this to something like ndarray or faer imo though assuming a generic implementation. I only did it like this because it was so small and simple and the usage was very minimal. So far the only function I think that falls into this category is the 2x2 determinant function which is used in this PR, the two qubit decomposer, and the one qubit decomposer (where it was originally added). But again it's just a single line so I didn't feel like it was worth creating a dedicate module for it.

mtreinish:
> This looks good to me. The only thing I would change is these
> 
> ```rust
>                 let rz_11 = (-Complex64::new(0., 0.5 * PI2)).exp();
>                 let rz_00 = Complex64::new(0., 0.5 * PI2).exp();
> ```
> 
> I know we are leaving a lot of things that could be improved, but this is such an easy and reasonable change, I don't see why not.

I updated this in: https://github.com/Qiskit/qiskit/pull/12197/commits/fcf587f661fe83970a1313e712aa6ccb189aac72 it turns out we all missed that this was `0.5 * pi / 2` not `0.5 * pi` so it's not just _-i_ and _i_. But I moved it to a constant for what the actual value is.

jakelishman (approved):
This looks as right to me as I think I'd be able to spot, given the original source.  Thanks for doing all this!

(will leave unmerged til tomorrow in case John wants to look too)

jlapeyre:
> it turns out we all missed that this was `0.5 * pi / 2` not `0.5 * pi`

On the one hand, oh duh. On the other hand, the fact that we all missed it suggests that `PI2` might not be a great name.
For example `FRAC_1_SQRT_2` might look verbose, but it's fairly easy to understand.

Issue #9614, opened 2023-02-17
Incrementally enable new Pylint lints


Eric-Arellano:
Our previous Pylint version was 2.8, and I recently upgraded it to 2.16 in .... To facilitate landing it, I disabled a bunch of checks. But most of them would be good to activate.

Working on this PR means:

1. Choose one of the checks you want to enable from here:
2. Delete it from `.pylintrc`
3. Run `tox -e lint` and fix all the issues until `tox -e lint` succeeds.

To make it easier for code reviewers, it's a good idea to activate 1 or 2 checks at a time.

lambda-knight:
Fixing consider-using-f-string (C0209) is worth trying ?

Eric-Arellano:
Yes, @lambda-knight consider-using-f-string looks useful to me. Note though that we are transitioning to Ruff instead of Pylint. So it may be worth double checking that Ruff has a similar lint. 

Even if Ruff doesn't have it, I still think it's valuable to fix all our historical instances regardless of if we introduce future "regressions" or not. Let's say we have 200 issues now, and we introduce 50 in the next 60 months. Better to have 200/250 fixed than 0/250.

lambda-knight:
OK, @Eric-Arellano ,
Enabling consider-using-f-string, 192 files and 520 warnings are reported.  I am fixing code of this warning and will issue PR.

joesho112358:
@lambda-knight hi, can i help out with `consider-using-f-string`?

PR #12198, opened 2024-04-18
Updates for removing non-ascii-name

joesho112358:
### Summary

### Details and comments

Only used for 3 variables in 1 file. If those variables are preferred, then we can close this and I can create a new PR to move it above the decision line. 

Also, open to renaming `lambda_p` and such :|

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -225,7 +225,6 @@ disable = [
     "nested-min-max",
     "no-member",
     "no-value-for-parameter",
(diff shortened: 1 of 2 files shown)


jakelishman (approved):
Thanks for looking at these lints!  I'm fine with `lambda_p` and friends in the test file to avoid the naming conflict - it's just one test file, and just using Greek letters (whether the letter or the name) is lazy-scientist naming on our part already, so no worries.

PR #12199, opened 2024-04-18
Updates for removing use-maxsplit-arg

joesho112358:
### Summary

### Details and comments

Only used in one place. It makes sense if the line only cares about the extension. On the other hand, is also very much a nit in this case since most image files don't have multiple `.` in them and makes the line a bit longer.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -241,7 +241,6 @@ disable = [
     "use-dict-literal",
     "use-list-literal",
     "use-implicit-booleaness-not-comparison",
-    "use-maxsplit-arg",
 ]
 
 enable = [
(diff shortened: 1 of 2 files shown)


jakelishman (approved):
This one also looks good, thanks.

Fwiw, I think there's a few in the "temporarily disabled" list that might not necessarily be the best style and might want moving into the "properly disabled" list, but it's been a long time since I looked at them, so I don't really remember.  For example, the one above this one (`use-implicit-booleaness-not-comparison`) is imo not a universally good idea and probably shouldn't be enabled as a lint requirement.

joesho112358:
> This one also looks good, thanks.
> 
> Fwiw, I think there's a few in the "temporarily disabled" list that might not necessarily be the best style and might want moving into the "properly disabled" list, but it's been a long time since I looked at them, so I don't really remember. For example, the one above this one (`use-implicit-booleaness-not-comparison`) is imo not a universally good idea and probably shouldn't be enabled as a lint requirement.

Yeah, figured these were pretty quick hitters... so might as well go ahead and just do it! No biggie to throw the branch away and go the other route! :D 

I can take a look at `use-implicit-booleaness-not-comparison`

PR #12201, opened 2024-04-18
Have `QuantumCircuit.store` do integer-literal type promotion

jakelishman:
### Summary

`QuantumCircuit.store` does convenience lifting of arbitary values to `expr.Expr` nodes.  Similar to the binary-operation creators, it's convenient to have `QuantumCircuit.store` correctly infer the required widths of integer literals in the rvalue.

This isn't full type inference, just a small amount of convenience.

### Details and comments

`QuantumCircuit.store` is not released yet, so no changelog for this behavioural change to it.

### DIFF
--- a/qiskit/circuit/quantumcircuit.py
+++ b/qiskit/circuit/quantumcircuit.py
@@ -2657,7 +2657,13 @@ class QuantumCircuit:
             :meth:`add_var`
                 Create a new variable in the circuit that can be written to with this method.
         """
-        return self.append(Store(expr.lift(lvalue), expr.lift(rvalue)), (), (), copy=False)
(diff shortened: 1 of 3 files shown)


jakelishman:
Just marking as "on hold" for a sec because I realised that we also have `add_var` that performs a similar role and I want to make sure that that does the same thing.

jakelishman:
Ok, I've added the same support to `add_var`.

kevinhartman (dismissed):
This looks good to me.

Is it worth it/possible to have a test to ensure that `add_var` doesn't widen bools?

Feel free to merge if not.

jakelishman:
Sure, done in 5603e5b310.

kevinhartman (approved):

PR #12203, opened 2024-04-18
Add support for high level synthesis plugins to return a DAGCircuit

### LINKED ISSUE #11681: Add support for high level synthesis plugins to return a DAGCircuit
### What should we add?

Right now the `HighLevelSynthesisPlugin` interface only allows for a plugin to return a `QuantumCircuit`, however this needlessly forces a conversion when `HighLevelSynthesis` is run as when the pass is being used we're always in a dag context. We can't change the interface, but we should expand it to support both a dag and circuit return from a plugin and just normalize to a dagcircuit inside the pass. So for passes that use a dagcircuit we avoid the conversion.

ElePT:
### Summary

This PR addresses #11681 by adding a `use_dag` flag to the high level synthesis plugins `run` method (and synthesis functions) to internally use and return a `DAGCircuit` instead of a `QuantumCircuit`. The flag is set to `True` when the plugins are called inside the `HighLevelSynthesis` pass to avoid back and forth conversions between circuit and dag.

This PR makes sure that the interface can handle `DAGCircuit` plugin outputs and implements the full end-to-end `DAGCircuit` pipeline for most of the synthesis functions/plugins (see table below). The only function that I didn't think made much sense to re-implement with `DAGCircuit` was `synth_clifford_layers`, because it works by constructing independent circuits that are then treated as instruction blocks, and I think that `QuantumCircuit` is better suited for this particular workflow. So I just convert the circuit to dag during the function return.

Other small changes that I included in this PR are:

- minor docstring/typing fixes
- setting `QuantumCircuit.append(.., copy=False)` where possible inside the synthesis functions (should be faster)
- I created an independent test file for clifford synthesis functions, which used to be tested in `test/python/quantum_info/operators/symplectic/test_clifford.py`. I migrated those tests to a new `test/python/synthesis/test_clifford_sythesis.py` file.
- I had to resolve some cyclic import conflicts to be able to import `DAGCircuit` in the synthesis functions, I did this by making all `qiskit.synthesis` imports in `qiskit.circuit` only imported at runtime. I think that this is a coherent move and should reduce the risk of cyclic imports in the future, but I am a bit weary of creating further issues so let me know if you see anything suspicious in this change.

### Details and comments

Because I found the structure of the plugins a bit convoluted at first, I have compiled a table where I have been keeping track of the modified plugins, related synthesis functions, whether they are able to use the dag end-to-end and where one can find the function tests (which have been modified to test `use_dag`):

| plugin                             | function(s)                       |   end-to-end dag? | independent function test?                                                     |
|------------------------------------|-----------------------------------|-------------------|--------------------------------------------------------------------------------|
| `DefaultSynthesisClifford`         |   `synth_clifford_full`           | yes               | [new!] test/python/synthesis/test_clifford_sythesis.py |
| `AGSynthesisClifford`              | `synth_clifford_ag`               | yes               |[new!] test/python/synthesis/test_clifford_sythesis.py |
| `BMSynthesisClifford`              | `synth_clifford_bm`               | yes               | [new!] test/python/synthesis/test_clifford_sythesis.py |
| `GreedySynthesisClifford`          | `synth_clifford_greedy`           | yes               | [new!] test/python/synthesis/test_clifford_sythesis.py |
| `LayerSynthesisClifford`           | `synth_clifford_layers`           | no                | test/python/synthesis/test_clifford_decompose_layers.py                        |
| `LayerLnnSynthesisClifford`        | `synth_clifford_depth_lnn`       | no                | test/python/synthesis/test_clifford_decompose_layers.py                        |
| `DefaultSynthesisLinearFunction`   | pmh linear function               | -                 | -                                                                              |
| `KMSSynthesisLinearFunction`       | `synth_cnot_depth_line_kms`       | yes                | test/python/synthesis/test_linear_synthesis.py                                 |
| `PMHSynthesisLinearFunction`       | `synth _cnot_count_full_pmh`      | yes                | test/python/synthesis/test_linear_synthesis.py                                 |
| `KMSSynthesisPermutation`          | `synth_permutation_depth_lnn_kms` | yes                | test/python/synthesis/test_permutation_synthesis.py                            |
| `BasicSynthesisPermutation`        | `synth_permutation_basic`         | yes                | test/python/synthesis/test_permutation_synthesis.py                            |
| `ACGSynthesisPermutation`          | `synth_permutation_acg`           | yes                | test/python/synthesis/test_permutation_synthesis.py                            |
| `TokenSwapperSynthesisPermutation` |  no                                | yes               | no (but test for plugin in test/python/transpiler/test_high_level_synthesis.py) |


qiskit/circuit/library/generalized_gates/uc.py
```diff
@@ -34,13 +34,9 @@
 from qiskit.circuit.exceptions import CircuitError
 from qiskit.exceptions import QiskitError
 
-# pylint: disable=cyclic-import
-from qiskit.synthesis.one_qubit.one_qubit_decompose import OneQubitEulerDecomposer
-
 from .diagonal import Diagonal
 
 _EPS = 1e-10  # global variable used to chop very small numbers to zero
-_DECOMPOSER1Q = OneQubitEulerDecomposer("U3")
```
ElePT, line 43: This was never used, I removed it to make the cyclic import handling easier but if it is necessary, I can bring it back.

qiskit/synthesis/linear_phase/cx_cz_depth_lnn.py
```diff
@@ -185,37 +194,63 @@ def _apply_phase_to_nw_circuit(n, phase_schedule, seq, swap_plus):
...
+            if use_dag:
+                cir.apply_operation_back(SGate(), (qreg[i],), check=False)
+            else:
+                cir.s(i)
 
     return cir
 
 
-def synth_cx_cz_depth_line_my(mat_x: np.ndarray, mat_z: np.ndarray) -> QuantumCircuit:
+def synth_cx_cz_depth_line_my(
```
ElePT, line 251: Added `use_dag` to the function because it's used in  `synth_clifford_depth_lnn`.

qiskit/synthesis/linear_phase/cz_depth_lnn.py
```diff
@@ -132,7 +150,7 @@ def _create_patterns(n):
     return pats
 
 
-def synth_cz_depth_line_mr(mat: np.ndarray) -> QuantumCircuit:
+def synth_cz_depth_line_mr(mat: np.ndarray, use_dag: bool = False) -> QuantumCircuit | DAGCircuit:
```
ElePT, line 153: Added `use_dag` to the function because it's used in  `synth_clifford_depth_lnn`.

qiskit/circuit/library/generalized_gates/uc.py
alexanderivrii: I am completely in favor of removing unused imports.

qiskit/synthesis/clifford/clifford_decompose_ag.py
```diff
@@ -53,7 +58,13 @@ def synth_clifford_ag(clifford: Clifford) -> QuantumCircuit:
...
 
     # Compose a circuit which we will convert to an instruction
-    circuit = QuantumCircuit(clifford.num_qubits, name=str(clifford))
+    qreg = QuantumRegister(clifford.num_qubits)
+    if use_dag:
+        circuit = DAGCircuit()
+        circuit.name = str(clifford)
+        circuit.add_qreg(qreg)
+    else:
+        circuit = QuantumCircuit(qreg, name=str(clifford))
```
alexanderivrii, line 67: Generally this looks fine, yet I do find that this introduces code duplication and that the new code a bit harder to read and to maintain. 

How much time do we save (here and in other synthesis functions) by directly constructing a `DAGCircuit` rather than constructing a `QuantumCircuit` and converting it to `DAGCircuit` using `dag_to_circuit`? 

For instance, If in practice this time is negligible, then maybe we don't really need the `DAGCircuit` version of the code and we can just let `HighLevelSynthesis` convert it to `DAGCircuit` (when needed).

And if this time is not negligible, then what exactly causes the slowdown? For instance, would it be possible to create some kind of "list of gates" instead of (the presumably more expensive) `QuantumCircuit`/`DAGCircuit`, and only construct the right object from this list at the very end? 

Alternatively, maybe we can split `synth_clifford_ag` into two functions like `_synth_clifford_ag_circuit`  and `_synth_clifford_ag_dag` that produce `QuantumCircuit` and `DAGCircuit` respectively? (Unfortunately this would introduce even more code duplication but imho would make each individual function more readable).

Or maybe we can have some generic functions like `_append_cx` or `_append_h`, etc. that under the hood use the correct method of adding a gate to either a `DAGCircuit` or a `QuantumCircuit` (and I think we might already have something of this kind).

Also this somewhat relates to the question/task of whether we eventually want to port these algorithms to Rust.

What do you think?

qiskit/synthesis/clifford/clifford_decompose_layers.py
```diff
@@ -137,34 +139,36 @@ def synth_clifford_layers(
...
         # cx_cz_synth_func is not None
-        layeredCircuit.append(CX_circ, qubit_list)
-
-    layeredCircuit.append(H2_circ, qubit_list)
-    layeredCircuit.append(S1_circ, qubit_list)
-    layeredCircuit.append(CZ1_circ, qubit_list)
+        layered_circuit.append(CX_circ, qubit_list, copy=False)
+    layered_circuit.append(H2_circ, qubit_list, copy=False)
+    layered_circuit.append(S1_circ, qubit_list, copy=False)
+    layered_circuit.append(CZ1_circ, qubit_list, copy=False)
```
alexanderivrii, line 154: Good idea (not copying)! I now seem to remember that either `_append` or `compose` should be even more performant (not sure).

qiskit/transpiler/passes/synthesis/high_level_synthesis.py
```diff
@@ -757,7 +823,29 @@ def run(self, high_level_object, coupling_map=None, target=None, qubits=None, **
...
+                    )
+                decomposition = transposed_circ
+            if use_inverted:
+                inverted_circ = decomposition.copy_empty_like()
+                for node in decomposition.topological_op_nodes():
+                    inverted_circ.apply_operation_front(
+                        node.op.inverse(), node.qargs, node.cargs, check=False
+                    )
+                decomposition = inverted_circ
+            return decomposition
```
alexanderivrii, line 848: This is another place where (imho) supporting both `QuantumCircuit` and `DAGCircuit` makes the code quite a bit more complex. I am wondering how much performance improvement we get in practice.

test/python/synthesis/test_clifford_sythesis.py
```diff
@@ -0,0 +1,282 @@
...
+def random_clifford_circuit(num_qubits, num_gates, gates="all", seed=None):
+    """Generate a pseudo random Clifford circuit."""
+
+    qubits_1_gates = ["i", "x", "y", "z", "h", "s", "sdg", "sx", "sxdg", "v", "w"]
+    qubits_2_gates = ["cx", "cz", "cy", "swap", "iswap", "ecr", "dcx"]
+    if gates == "all":
+        if num_qubits == 1:
+            gates = qubits_1_gates
+        else:
+            gates = qubits_1_gates + qubits_2_gates
```
alexanderivrii, line 91: This function still exists in the `test_cliffords` file, right? Maybe it would be nice to move it (and also `random_linear_circuit`) into the main `qiskit` repository. I don't know why we have not done this already, but more than once I wished we had this in the main codebase.

test/python/synthesis/test_clifford_sythesis.py
```diff
@@ -0,0 +1,282 @@
...
+# This code is licensed under the Apache License, Version 2.0. You may
+# obtain a copy of this license in the LICENSE.txt file in the root directory
+# of this source tree or at http://www.apache.org/licenses/LICENSE-2.0.
+#
+# Any modifications or derivative works of this code must retain this
+# copyright notice, and modified files need to carry a notice indicating
+# that they have been altered from the originals.
+
+# pylint: disable=invalid-name
+"""Tests for Clifford synthesis functions."""
```
alexanderivrii, line 14: I very much like that this functionality was moved to a separate file, thanks!

alexanderivrii (commented):
Thanks @ElePT! In general, the changes look fine to me, however I do have several comments and would like to discuss this a bit more before approving. 

From the user perspective, when writing a synthesis plugin, having to return both the `QuantumCircuit` version (when `use_dag` is `False`) and the `DAGCircuit` version (when `use_dag` is `True`) seems a bit too much, and is technically not backwards-compatible. Yet I like the idea of passing the flag `use_dag` to indicate which of the two versions is preferred, when one can create both. In other words, the user may ignore this flag and return only one version of the circuit, the `HighLevelSynthesis` pass is then responsible to convert it to the right format when needed. But for performance-critical applications the user may want to support both versions and return the preferred one based on the flag. What do you think? 

Speaking of this, the two mostly used clifford synthesis functions are currently `synthesize_clifford_greedy` and `synthesize_clifford_bm` (the latter is for at most 3-qubit cliffords). While it may make sense to optimize these two specific functions, it probably makes less sense to optimize e.g. `synthesize_clifford_ag`, @ShellyGarion, what do you think?(Also cc @mtreinish).

It would be nice to have some idea how much performance improvement we can get by directly constructing the `DAGCircuit` vs. constructing a `QuantumCircuit` and running `dag_to_circuit`.

Please also see my comment on possible ideas on avoiding code duplication. 

There is also a question on how this relates to eventually porting (at least some of the performance-critical) code to Rust.

What do you think?

qiskit/synthesis/clifford/clifford_decompose_ag.py
ElePT: I see your point @alexanderivrii, and I agree that the current code is less readable. I am working on adding a HLS plugin benchmark to our ASV suite to be able to test if we are actually getting relevant performance improvements or not, but I haven't had time to finish it yet. I think that the best course of action would be to finish the benchmark and make a decision based off the results. I also considered the option to have a full dag-based function as an alternative, but it's a lot of functions to migrate and I wanted to get some feedback on the current state first.

ElePT: I have made a first attempt at benchmarking and I am getting mixed results, some tests improve with the PR but others show a performance regression (marked with an asterisk). The benchmark is done with the code in (https://github.com/Qiskit/qiskit/pull/12329), which is still WIP, so they may also not be the best benchmarks.

[Edit]: took a look offline and some of the plugins might not be running as expected, so I wouldn't take the results too seriously.

```
                                                                                Before      After
transpiler_hls_plugins.HLSPluginsSuite.time_clifford_ag_large_circuit           909±4ms     967±30ms (*)
transpiler_hls_plugins.HLSPluginsSuite.time_clifford_ag_rand_benchmarking       2.20±0.01s  2.00±0.08s
transpiler_hls_plugins.HLSPluginsSuite.time_clifford_bm_large_circuit           936±20ms    960±30ms (*)
transpiler_hls_plugins.HLSPluginsSuite.time_clifford_bm_rand_benchmarking       5.93±0.09s  2.50±0.06s
transpiler_hls_plugins.HLSPluginsSuite.time_clifford_default_large_circuit      923±30ms    922±20ms
transpiler_hls_plugins.HLSPluginsSuite.time_clifford_default_rand_benchmarking  6.25±0.1s   6.41±0.2s (*)
transpiler_hls_plugins.HLSPluginsSuite.time_clifford_greedy_large_circuit       926±20ms    969±30ms (*)
transpiler_hls_plugins.HLSPluginsSuite.time_clifford_greedy_rand_benchmarking   8.86±0.2s   9.81±0.06s (*)
transpiler_hls_plugins.HLSPluginsSuite.time_linear_func_kms                     3.11±0.07s  2.64±0.03s
transpiler_hls_plugins.HLSPluginsSuite.time_linear_func_pmh                     1.84±0.08s  1.66±0.04s
```

ElePT:
Given the ongoing efforts to port different HLS components to Rust, and how outdated this PR is, I am in favor of closing it and opening a new one when we have a clearer picture of what should be done in this regard. It might be more practical than bumping it infinitely to the next release milestone. What do you think @alexanderivrii?

alexanderivrii:
Thanks @ElePT. I am fine with either option. Given the recent (and the additionally planned) changes to the HighLevelSynthesis transpiler pass, it might indeed be much simpler to reimplement the PR from scratch, but I would definitely like to keep this PR as a reference of what needs to be done.

alexanderivrii:
The original (and a very important) motivation for this PR was to reduce the (huge) number of dag-to-circuit and circuit-to-dag conversions done within the HighLevelSynthesis transpiler pass. The reason for that is that the recursion within HighLevelSynthesis is done on the level of DAGCircuits, while many of the internal methods (circuit definitions for custom gates, circuit blocks for control-flow statements, and synthesized circuits produced by high-level-synthesis plugins) produce QuantumCircuits. However, with PR #13605, all of the recursion is done on the level of QuantumCircuits as well (with the exception of a single pair of conversions on the top-level) and this PR is not strictly required. @ElePT, what do you think -- should we close this?

ElePT:
I agree with closing the PR, I think the current developments have diverged enough.

PR #12204, opened 2024-04-18
Add representation of `expr.Var` to `DAGCircuit`

### LINKED ISSUE #10927: Add equality comparisons for manual variables to `DAGCircuit.__eq__`
Ensure that `DAGCircuit.__eq__` (and by extension `QuantumCircuit.__eq__`) works in the presence of manual variables.  It is not immediately clear whether any sort of variable equivalence should be checked by default, but for the sake of conservatism in the initial implementation, we can say that only `Var` instances that directly compare equal will be considered equal (despite the `expr.structurally_equivalent` checker giving us the option to expand this).  With `expr.Var` expected to contain a UUID field to facilitate comparisons between differently serialised variables (#10923), the main reason to allow a more semantic-equality checker is far less prominent.

### LINKED ISSUE #10960: Support classical variables in `DAGCircuit.copy` and `copy_empty_like`

jakelishman:
### Summary

This adds basic support for `Var`s to `DAGCircuit`, enabling the base representation using the same wire structure used for clbits.  This is known to be an inefficient representation of the dataflow for clbits, which can be read multiple times without implying an order between those operations (unlike qubits for which reads and writes are more naturally linked).  We're using this simpler representation to make a better initial MVP; optimising the data-flow representation would come as part of a larger effort within the `DAGCircuit`.

This commit adds support in `DAGCircuit` for:

- representation of `Var` nodes
- appending all operations that might contain `Var` nodes to the DAG and updating the wire structure (including control-flow ops and stores)
- equality checking of DAGs with `Var`s (and so enables `QuantumCircuit.__eq__` as well)
- `DAGCircuit.copy_empty_like` with `Var`s
- the DAG/circuit converters

The other methods in `DAGCircuit` that might need to be aware of `Var` nodes will be handled separately.

### Details and comments

Close #10927 
Close #10960

### DIFF
--- a/qiskit/dagcircuit/dagcircuit.py
+++ b/qiskit/dagcircuit/dagcircuit.py
@@ -22,10 +22,12 @@ directly from the graph.
 """
 from __future__ import annotations
 
+import copy
+import enum
+import itertools
+import math
 from collections import OrderedDict, defaultdict, deque, namedtuple
 from collections.abc import Callable, Sequence, Generator, Iterable
-import copy
-import math
 from typing import Any
 
 import numpy as np
@@ -39,7 +41,9 @@ from qiskit.circuit import (
     SwitchCaseOp,
     _classical_resource_map,
     Operation,
+    Store,
 )
+from qiskit.circuit.classical import expr
 from qiskit.circuit.controlflow import condition_resources, node_resources, CONTROL_FLOW_OP_NAMES
 from qiskit.circuit.quantumregister import QuantumRegister, Qubit
 from qiskit.circuit.classicalregister import ClassicalRegister, Clbit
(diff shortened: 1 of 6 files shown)


qiskit/dagcircuit/dagcircuit.py
```diff
@@ -122,7 +147,11 @@ def __init__(self):
     @property
     def wires(self):
         """Return a list of the wires in order."""
-        return self.qubits + self.clbits
+        return (
+            self.qubits
+            + self.clbits
+            + [var for vars in self._vars_by_type.values() for var in vars]
```
mtreinish, line 153: Are you worried about the runtime impact of this? I guess the tradeoff would be storing this as a separate cached field at the dagcircuit level which doesn't seem worth it since the information is duplicative and realistically `Var`s won't be a common workflow to start.

mtreinish: Actually thinking about this more would it make more sense to do something like `return list(self.input_map)`?

test/python/dagcircuit/test_dagcircuit.py
```diff
@@ -1822,6 +1840,187 @@ def test_semantic_expr(self):
...
+            # c
+            (dag.input_map[c], op_store, c),
+            (op_store, op_switch, c),
+            (op_switch, dag.output_map[c], c),
+            # d
+            (dag.input_map[d], op_switch, d),
+            (op_switch, dag.output_map[d], d),
+        }
+        self.assertEqual(set(dag.edges()), expected_edges)
+
```
mtreinish, line 2029: You don't really have any tests covering the negative path with `DAGCircuit` construction with adding vars. Adding inputs with captures, duplicate var names, non-standalone vars, etc. I think it'd be good to have explicit tests for the failure modes on those too.

mtreinish (commented):
Overall this LGTM, changing the scope of this to just use one wire per `Var` makes this fairly straightforward to fit into the `DAGCircuit` structure. I just left one minor musing inline, but otherwise I didn't see anything odd. The only reason I held off on approving is because I left an inline comment about some test coverage gaps on the error path around vars. The tests only cover the valid use cases so none of the error handling/checking is being fully exercised.

qiskit/dagcircuit/dagcircuit.py
jakelishman: Honestly, I was kind of giving up on the runtime-performance of this particular property for this PR because if it would be a perf problem, it already _is_ because of the recalculation.

`list(self.input_map)` would be better, I was mostly concerned about it being a different iteration order to the existing form and that having knock-on effects on some of our tracked metrics.  I can swap it over to that and we can see, if you prefer?

test/python/dagcircuit/test_dagcircuit.py
jakelishman: whoops yep, thanks - I got wrapped up in writing the tests for the wire structures.  I need to get more into the habit of checking the coverage report.  Done in ff088b9dc3.

qiskit/dagcircuit/dagcircuit.py
mtreinish: It's fine changing the iteration order is a valid reason to leave this. If it becomes an issue we can revisit this in the future. Looking through the code the only thing that uses this is `CommutationAnalysis` and `CommutativeCancellation` to get an iterator over bits (I think it only cares about qubits I'll look at a fix PR for that).

mtreinish (approved):

PR #12207, opened 2024-04-18
Reject vars in `QuantumCircuit.to_instruction` and `to_gate`

jakelishman:
### Summary

`QuantumCircuit.to_gate` may be able to support `input` variables that are angle-typed at some point in the future, but that's very limited and not immediately on the roadmap.

`QuantumCircuit.to_instruction` does have a natural way to support both `input` vars (they become function arguments) and declared vars (since they're completely internal to the function body), but those are currently rejected because we don't have a way of specifying function signatures / calling functions yet and we don't have a neat complete story around unrolling variables into circuits yet (where there may be naming clashes), especially through transpilation.  This opts to raise the errors immediately, with alternatives where possible, rather than letting things _kind of_ work.

### Details and comments

I will add support for `QuantumCircuit.compose`, `DAGCircuit.compose` and the other `DAGCircuit` replacement passes in a follow-up, with some minimum support for inlining classical variables (though probably not yet with renaming, except maybe in `QuantumCircuit.compose`).

### DIFF
--- a/qiskit/converters/circuit_to_instruction.py
+++ b/qiskit/converters/circuit_to_instruction.py
@@ -148,8 +148,6 @@ def circuit_to_instruction(circuit, parameter_map=None, equivalence_library=None
     data.map_ops(fix_condition)
 
     qc = QuantumCircuit(*regs, name=out_instruction.name)
-    for var in circuit.iter_declared_vars():
-        qc.add_uninitialized_var(var)
     qc._data = data
     qc._parameter_table = ParameterTable(
         {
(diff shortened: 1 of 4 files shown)


qiskit/converters/circuit_to_instruction.py
```diff
@@ -126,6 +148,8 @@ def fix_condition(op):
     data.map_ops(fix_condition)
 
     qc = QuantumCircuit(*regs, name=out_instruction.name)
+    for var in circuit.iter_declared_vars():
+        qc.add_uninitialized_var(var)
```
mtreinish: If we raise above, does this loop ever do anything?

mtreinish (commented):
Overall this LGTM, just one inline question

qiskit/converters/circuit_to_instruction.py
jakelishman: Oh whoops, it's probably mistakenly left in from before I pivoted and thought it was better to reject-by-default. Fixed in f29163b.

mtreinish (approved):
LGTM now, thanks for the quick update.

PR #12214, opened 2024-04-18
Document generate_preset_pass_manager supports list for initial_layout

### LINKED ISSUE #11690: generate_preset_pass_manager should take initial_layout as a list
### What should we add?

Now that https://github.com/Qiskit/qiskit/pull/10344 is merged, we should make the `generate_preset_pass_manager` function take an integer list for layout. This is a user entry point and should be easy to use. The Layout object is a pain to use all around (I would love to completely eliminate it, since it doesn't contain any more information than a list of int. The circuit qubits have an order).

mtreinish:
### Summary

This commit updates the documentation for the
generate_preset_pass_manager() function to clearly indicate that the function will accept a integer list for the initial_layout field. This already was supported but it wasn't documented. As it's now a documented part of the API a unittest is added to ensure we don't regress this functionality in the future.

### Details and comments

Fixes #11690

### DIFF
--- a/test/python/transpiler/test_preset_passmanagers.py
+++ b/test/python/transpiler/test_preset_passmanagers.py
@@ -1436,6 +1436,38 @@ class TestGeneratePresetPassManagers(QiskitTestCase):
         # Ensure the DAGs from both methods are identical
         self.assertEqual(transpiled_circuit_list, transpiled_circuit_object)
 
+    @data(0, 1, 2, 3)
+    def test_generate_preset_pass_manager_with_list_initial_layout(self, optimization_level):
+        """Test that generate_preset_pass_manager can handle list based initial layouts."""
+        coupling_map_list = [[0, 1]]
+
+        # Circuit that doesn't fit in the coupling map
+        qc = QuantumCircuit(2)
+        qc.h(0)
+        qc.cx(0, 1)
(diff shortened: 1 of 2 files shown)


test/python/transpiler/test_preset_passmanagers.py
```diff
@@ -1436,6 +1436,38 @@ def test_generate_preset_pass_manager_with_list_coupling_map(self):
...
+            coupling_map=coupling_map_list,
+            basis_gates=["u", "cx"],
+            seed_transpiler=42,
+            initial_layout=Layout.from_intlist([1, 0], *qc.qregs),
+        )
+        tqc_list = pm_list.run(qc)
+        tqc_obj = pm_list.run(qc)
+        self.assertIsInstance(pm_list, PassManager)
+        self.assertIsInstance(pm_object, PassManager)
+        self.assertEqual(tqc_list, tqc_obj)
```
jakelishman, line 1469: We don't actually formally compare layouts in `QuantumCircuit.__eq__` because we immediately convert to `DAGCircuit`, which doesn't have a layout field, but we do compare qubit positions, so this should be a fine test.

jakelishman (approved):
Looks fine to me, and good to test/document the expected behaviour of the int list.

PR #12215, opened 2024-04-19
Support `Var` in circuit-substitution methods

### LINKED ISSUE #10932: Support manual variables in `QuantumCircuit.compose`
For simplicity in the initial implementation, we can likely reject any attempts to compose two circuits that use the same variable instances or have naming clashes; this is functionality we can design and add later if it transpires to be necessary.

See discussion in https://github.com/Qiskit/RFCs/pull/50/files#r1328805042 and https://github.com/Qiskit/RFCs/pull/50/files#r1329111336 for more.

### LINKED ISSUE #10933: Support manual variables in `DAGCircuit.compose`
Work on this will likely be subject to the same constraints discussed in the `QuantumCircuit.compose` counterpart of #10932.

### LINKED ISSUE #10934: Support manual variables in `DAGCircuit.replace_block_with_op`

### LINKED ISSUE #10935: Support manual variables in `DAGCircuit.substitute_node`

### LINKED ISSUE #10936: Support manual variables in `DAGCircuit.substitute_node_with_dag`

jakelishman:
### Summary

This PR will add support for `Var` to all the `QuantumCircuit` and `DAGCircuit` substitution methods:
- `QuantumCircuit.compose`
- `DAGCircuit.compose`
- `DAGCircuit.replace_block_with_op`
- `DAGCircuit.substitute_node`
- `DAGCircuit.substitute_node_with_dag`

I'll do it all in one PR, because the idea of substitution/inlining is quite common among all of them.

~At the point of pushing, the PR only does `QuantumCircuit.compose` (which I expected to be the hardest), and it's not got a test suite, though my local testing of it was successful.~ Now done.

### Details and comments

~Depends on #12204.~

#### Support `Var` in `QuantumCircuit.compose`

Close #10932
This converts the operation-rewriting step of `QuantumCircuit.compose` into one that can recurse through control-flow operations rewriting variables (to avoid conflicts). This initial commit adds two ways of combining the variables:

1. variables in the `other` are disjoint from those in `self`, and are simply added to them.  This makes it easy to join unrelated circuits.  The `var_remap` argument is to make it easier to combine existing circuits that may already have collisions outside of the joiner's control.
2. variables marked as "captures" in `other` can be inlined onto existing variables in `self`.  This makes it possible to build up a circuit with variables layer-by-layer.

In support of objective 2, I also taught `copy_empty_like` a way to produce a new base layer with all the declared variables converted to "captures" to make it easier to produce new base layers.

I deliberately did not include any _automatic_ variable renaming because the usability of that seemed very hard to do well; while the circuit can easily be created in a unique way, the user would then be hard-pressed to actually retrieve the new `Var` nodes afterwards.  Asking the user to manually break naming collisions guarantees that they'll be able to find their variables again afterwards.

#### Support `Var` in `DAGCircuit.compose`

Close #10933.

This similarly adds support for a `vars_mode` argument to `DAGCircuit.copy_empty_like` to make using this more convenient.  It threads the same argument through some of the `DAGCircuit` methods that build up layers of DAGs, since this is more common here.

Unlike `QuantumCircuit.compose`, `DAGCircuit.compose` does not (yet?) offer the ability to remap variables during the composition, because the use-cases for direct `DAGCircuit` composition are typically less about building up many circuits from scratch and more about rebuilding a DAG from itself.  Not offering the option makes it simpler to implement.

#### Support `Var` in `DAGCircuit.replace_block_with_op`

Close #10934.

This is straightforwards, since the wire structure is the same; there's not actually any changes needed except for a minor comment explaining that it works automatically.

#### Support `Var` in `DAGCircuit.substitute_node_with_dag`

Close #10936.

This is marginally trickier, and to avoid encoding performance problems for ourselves in the public API, we want to avoid any requirement for ourselves to recurse.  The expectation is that use-cases of this will be building the replacement DAG themselves, where it's easy for them to arrange for the `Var`s to be the correct ones immediately.

This also allows nice things like substitutions that contract wire use out completely, over all wire types, not just qubits.

#### Support `Var` in `DAGCircuit.substitute_node`

Close #10935.

The easiest way to do this is actually to simplify the code a whole bunch using the wire helper methods.  It would have been automatically supported, had those methods already been in use.

### DIFF
--- a/crates/circuit/src/circuit_data.rs
+++ b/crates/circuit/src/circuit_data.rs
@@ -459,21 +459,14 @@ impl CircuitData {
         clbits: Option<&Bound<PyAny>>,
     ) -> PyResult<()> {
         let mut temp = CircuitData::new(py, qubits, clbits, None, 0)?;
-        if temp.qubits_native.len() < self.qubits_native.len() {
-            return Err(PyValueError::new_err(format!(
-                "Replacement 'qubits' of size {:?} must contain at least {:?} bits.",
-                temp.qubits_native.len(),
-                self.qubits_native.len(),
-            )));
-        }
-        if temp.clbits_native.len() < self.clbits_native.len() {
-            return Err(PyValueError::new_err(format!(
-                "Replacement 'clbits' of size {:?} must contain at least {:?} bits.",
-                temp.clbits_native.len(),
-                self.clbits_native.len(),
-            )));
-        }
         if qubits.is_some() {
+            if temp.qubits_native.len() < self.qubits_native.len() {
+                return Err(PyValueError::new_err(format!(
+                    "Replacement 'qubits' of size {:?} must contain at least {:?} bits.",
+                    temp.qubits_native.len(),
+                    self.qubits_native.len(),
+                )));
+            }
             std::mem::swap(&mut temp.qubits, &mut self.qubits);
             std::mem::swap(&mut temp.qubits_native, &mut self.qubits_native);
             std::mem::swap(
@@ -482,6 +475,13 @@ impl CircuitData {
             );
         }
         if clbits.is_some() {
+            if temp.clbits_native.len() < self.clbits_native.len() {
+                return Err(PyValueError::new_err(format!(
+                    "Replacement 'clbits' of size {:?} must contain at least {:?} bits.",
+                    temp.clbits_native.len(),
+                    self.clbits_native.len(),
+                )));
+            }
             std::mem::swap(&mut temp.clbits, &mut self.clbits);
             std::mem::swap(&mut temp.clbits_native, &mut self.clbits_native);
             std::mem::swap(
--- a/test/python/circuit/test_circuit_operations.py
+++ b/test/python/circuit/test_circuit_operations.py
@@ -485,6 +485,39 @@ class TestCircuitOperations(QiskitTestCase):
         self.assertEqual({b, d}, set(copied.iter_captured_vars()))
         self.assertEqual({b}, set(qc.iter_captured_vars()))
 
+    def test_copy_empty_variables_alike(self):
+        """Test that an empty copy of circuits including variables copies them across, but does not
+        initialise them.  This is the same as the default, just spelled explicitly."""
+        a = expr.Var.new("a", types.Bool())
+        b = expr.Var.new("b", types.Uint(8))
+        c = expr.Var.new("c", types.Bool())
+        d = expr.Var.new("d", types.Uint(8))
+
+        qc = QuantumCircuit(inputs=[a], declarations=[(c, expr.lift(False))])
+        copied = qc.copy_empty_like(vars_mode="alike")
+        self.assertEqual({a}, set(copied.iter_input_vars()))
+        self.assertEqual({c}, set(copied.iter_declared_vars()))
+        self.assertEqual([], list(copied.data))
+
+        # Check that the original circuit is not mutated.
+        copied.add_input(b)
+        copied.add_var(d, 0xFF)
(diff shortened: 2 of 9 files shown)


jakelishman:
Ok, I think I've got all the relevant `QuantumCircuit` and `DAGCircuit` methods converted now.  For the most part, this should be the majority of the work needed to support `Var` end-to-end through `transpile`, though I think there's a couple of places in `ApplyLayout` and potentially passes that do DAG rebuilds that need to take a little more care.

Ready for review, barring the "on hold" because of the dependency on #12204 which should be reviewed first.

jakelishman:
Now rebased over #12204 and ready for review.

crates/circuit/src/circuit_data.rs
```diff
@@ -459,21 +459,14 @@ impl CircuitData {
...
-            )));
-        }
         if qubits.is_some() {
+            if temp.qubits_native.len() < self.qubits_native.len() {
+                return Err(PyValueError::new_err(format!(
+                    "Replacement 'qubits' of size {:?} must contain at least {:?} bits.",
+                    temp.qubits_native.len(),
+                    self.qubits_native.len(),
+                )));
+            }
```
mtreinish, line 469: Was it coming up somewhere that we were trying to do this check when qubits was none? I would assume `qubits_native.len()` would just eval to 0 then and we'd only trigger the error if there was a mismatch. It doesn't really matter I guess though since this is also obviously correct too.

test/python/circuit/test_circuit_operations.py
```diff
@@ -485,6 +485,36 @@ def test_copy_empty_variables(self):
         self.assertEqual({b, d}, set(copied.iter_captured_vars()))
         self.assertEqual({b}, set(qc.iter_captured_vars()))
 
+    def test_copy_empty_variables_to_captures(self):
```
mtreinish, line 521: Do you want a test case validating the `alike` mode case?

mtreinish (commented):
Overall this LGTM. I might have gotten a bit lost in some of the internal logic of `QuantumCircuit.compose()` but I couldn't spot anything that looked suspect to me. The only questions I have here are whether there are any test coverage gaps here (I didn't look at the coverage report yet). Also how were you planning to handle release notes. You're adding several new options to existing methods and it's probably best to document them in situ just so we don't lose them.

crates/circuit/src/circuit_data.rs
jakelishman: There was some point at which we were replacing either qubits or clbits but not both - I don't remember which, but it was causing a hard failure because the previous logic was wrong in its checks in that case.

test/python/circuit/test_circuit_operations.py
jakelishman: `alike` is the default, so that's tested by the `test_copy_empty_variables` that's immediately above this diff.  I can pass it explicitly in that, if you prefer?

jakelishman:
For release notes: I was probably planning to do it all in one big go at the end, post rc1.

For testing gap: from the coverage report, it looks like I'm missing a test of `DAGCircuit.compose` for the case that nobody uses - composing two completely unrelated `DAGCircuit`s together, but that's it - the other missed lines are pre-existing missing logic, just moved in the diff.

test/python/circuit/test_circuit_operations.py
jakelishman: Explicit test added in a4a9cea96b.

jakelishman:
Explicit test for the other bits of `DAGCircuit.compose` added in a4a9cea96b.

mtreinish (approved):

PR #12217, opened 2024-04-19
Removing lint rules that are unused

joesho112358:
### Summary

#9614 removing some of the rules that currently do not result in an error

### Details and comments

Removing
`unknown-option-value`
`unnecessary-direct-lambda-call`
`unnecessary-ellipsis`
`use-list-literal`
because they do not appear to currently trigger. Figure they can be looked at again if they ever do cause an issue. Otherwise, I can close this and push them above for long term removal. 

It does appear that `use-list-literal` vs `use-dict-literal` is inconsistent:
https://github.com/Qiskit/qiskit/blob/ff7ba3f1a61b6e474c00055abb62fd99dd393a99/test/python/result/test_result.py#L59-L61

Line 59 defines a dictionary with `dict(...)` and line 61 defines a list with `[...]`

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -227,18 +227,14 @@ disable = [
     "no-value-for-parameter",
     "not-context-manager",
     "superfluous-parens",
-    "unknown-option-value",
     "unexpected-keyword-arg",
     "unnecessary-dict-index-lookup",
-    "unnecessary-direct-lambda-call",
     "unnecessary-dunder-call",
-    "unnecessary-ellipsis",
     "unnecessary-lambda-assignment",
     "unnecessary-list-index-lookup",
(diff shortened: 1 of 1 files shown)


jakelishman (approved):
Thanks!

That discrepancy you highlighted in `dict(**mapping)` was likely overzealous find-and-replace than a particular _intention_ to use the `dict` constructor.  The `Result` code used to use a very heavy validation library that made _everything_ into objects with expensive schemas backing it, so there was loads of stuff like `RandomObject(**deserialised)` to try and stuff data back into the validators.  Some of those just became raw `dict` instances, so we probably accidentally ended up with this funny `dict(**dict)` stuff, which at the time there was no lint against.

PR #12129, opened 2024-04-03
Add mapping-like features to data_bin

t-imamichi:
### Summary

Adds mapping-like features to `DataBin` (`__iter__`, `__getitem__`, `__contains__`, `keys`, `values`, `items`).
It also revises the return type of `make_data_bin` so that IDE can recognize the new methods.

PR #12219, opened 2024-04-19
Update make_data_bin() to return DataBin proper

ihincks:
### Summary

This PR updates `DataBin` with the intention of making it easier to construct directly, avoiding the `make_data_bin` class.

I have concerns that this breaks the class because it drops support for positional arguments in the constructor, but am opening up the PR anyways. There's also no way to properly infer shape at instance construction time, but there is (as previously implemented) at dynamic type construction time.

### Details and comments

This PR attempts to keep legacy support for a bunch of behaviour that is technically private, but that has become widespread due to lack of a way to achieve certain basic tasks (like find namespace names) in a simple way.


qiskit/primitives/containers/data_bin.py
```diff
@@ -15,78 +15,78 @@
...
+
+    @property
+    def _SHAPE(self) -> tuple[int, ...] | None:  # pylint: disable=invalid-name
+        return None
+
+
+# pylint: disable=unused-argument
 def make_data_bin(
     fields: Iterable[tuple[str, type]], shape: tuple[int, ...] | None = None
 ) -> DataBinMeta:
```
t-imamichi, line 76: How about changing the return type to `type[DataBin]` as I did in #12129?
#12129 adds some methods to `DataBin`. But since they are not part of `DataBinMeta`, they are not recognized by IDE.

t-imamichi:
Thank you for overhauling DataBin. It gets simpler now.
Do you plan to add mapping methods `keys`, `values`, `items`, etc. in #12129 ?
It can be done easily by calling `self.__dict__.keys()` for example.
If necessary, you can copy the tests from #12129. I don't mind closing #12129.

ihincks:
Yes, I plan to include the methods added in #12129. I'm hoping to merge 12129 first because it doesn't have any of the concerns that are mentioned in the summary of this PR.

qiskit/primitives/containers/data_bin.py
```diff
@@ -15,78 +15,78 @@
...
 from __future__ import annotations
 
-from collections.abc import Iterable
-from dataclasses import make_dataclass
+from typing import Any, Iterable
 
 
 class DataBinMeta(type):
-    """Metaclass for :class:`DataBin` that adds the shape to the type name.
+    """This metaclass no longer has a purpose and will be removed."""
```
chriseclectic: Is there any existing code that uses this class that prevents removing it now? The only thing I can think to check is JSON serialization in the runtime

qiskit/primitives/containers/data_bin.py
```diff
@@ -15,78 +15,78 @@
...
-    """
 
-    def __repr__(cls):
-        name = cls.__name__
-        if cls._SHAPE is None:
-            return name
-        shape = ",".join(map(str, cls._SHAPE))
-        return f"{name}<{shape}>"
+class DataBin(metaclass=DataBinMeta):
+    """Namespace for storing data.
```
chriseclectic, line 33: Should this inherit from `Mapping` now? Also its now both a namespace and a mapping container (just needs an iter to complete the mapping abc i think)

qiskit/primitives/containers/data_bin.py
```diff
@@ -15,78 +15,78 @@
...
 
     _RESTRICTED_NAMES = ("_RESTRICTED_NAMES", "_SHAPE", "_FIELDS", "_FIELD_TYPES")
-    _SHAPE: tuple[int, ...] | None = None
-    _FIELDS: tuple[str, ...] = ()
-    """The fields allowed in this data bin."""
-    _FIELD_TYPES: tuple[type, ...] = ()
-    """The types of each field."""
+
+    def __init__(self, **kwargs):
+        self.__dict__.update(kwargs)
```
chriseclectic: Is it an ok thing to update the internal dict representation of an object? should we instead be doing something like
```python
for key, val in kwargs.items():
    super().__setattr__(key, val)
```
or 
```
self._data = kwargs
```
and use slots to prevent adding extra fields, and update getitem and getattr to look in `self._data`?

I suppose one reason for setting actual attributes was for auto-complete but does that work with dynamically set attributes?

qiskit/primitives/containers/data_bin.py
```diff
@@ -15,78 +15,78 @@
...
+
+    def __setattr__(self, key, value):
+        raise NotImplementedError
 
     def __len__(self):
-        return len(self._FIELDS)
+        return len(self.__dict__)
```
chriseclectic: Is this safe? there won't be any extra fields in the classes dict format (eg if we added a shape /ndim to it which I would find very useful in majority of the post-processing i do with these classes)

chriseclectic (commented):
Couple questions

qiskit/primitives/containers/data_bin.py
ihincks: I looked at all of our repos, public and private, and I didn't find any usage. I have therefore removed the metaclass altogether in this PR.

qiskit/primitives/containers/data_bin.py
ihincks: The iter already exists from a different PR. I'd like to control our own namespace, so I think I'm happy to be a mapping only via protocol.

qiskit/primitives/containers/data_bin.py
ihincks: It should be safe enough and prevents an extra dictionary from existing. This last point is only relevant if you have a bajillion of them. Anyways, I moved stuff to `_data` because `_shape` was added, so it just made things easier.

Some type checkers search objects dynamically, and some search statically. It's impossible to appease the ones that search statically in the way we use this class. I hope it currently works for the dynamic ones.

qiskit/primitives/containers/data_bin.py
ihincks: See a different answer.

qiskit/primitives/containers/data_bin.py
```diff
@@ -15,110 +15,138 @@
...
-        return len(self._FIELDS)
+        return len(self._data)
+
+    def __setattr__(self, *_):
+        raise NotImplementedError
 
     def __repr__(self):
-        vals = (f"{name}={getattr(self, name)}" for name in self._FIELDS if hasattr(self, name))
-        return f"{type(self)}({', '.join(vals)})"
+        vals = (f"{name}={val}" for name, val in self.items())
```
t-imamichi: This might be off-topic but I have a concern about repr of ndarray. Both `str(ndarray)` and `repr(ndarray)` print all matrix elements and it may break `repr(data_bin)` if the array size is large. On the other hand, BitArray displays a concise text such as `BitArray(<shape=(), num_shots=2, num_bits=4>)`.

It might be nice to introduce a specialized text format only for ndarray, e.g.,
- `f"ndarray(<shape={array.shape}, dtype={array.dtype}>)"`
  - `ndarray(<shape=(2, 3), dtype=bool>)`.

t-imamichi:
Thank you for overhauling data bin. It looks good overall. Could you write a reno?

qiskit/primitives/containers/data_bin.py
ihincks: Done in [d049863](https://github.com/Qiskit/qiskit/pull/12219/commits/d049863b601ec50a1e440c195d4fe9971d0ac9ab)

t-imamichi: Thanks. It looks great.
```
db = DataBin(a=np.empty((1, 2, 3)), shape=(2, 3))
print(db)
# DataBin(a=np.ndarray(<shape=(1, 2, 3), dtype=float64>), shape=(1, 2))
```

t-imamichi:
I noticed that we can simplify the code slightly. I wrote a suggestion https://github.com/ihincks/qiskit-terra/pull/2.

t-imamichi (approved):
LGTM

PR #12284, opened 2024-04-19
Add QPY format version table to the docs

### LINKED ISSUE #12283: Add format version mapped to releases table in QPY format docs
### What should we add?

One thing that is currently tricky to investigate without looking at the historical release notes is which QPY format versions were available in each release. The QPY format is backwards compatible but not forwards compatible, so there is a often a need to know which versions of QPY are supported for reading with different versions of Qiskit. In general you can always check `qiskit.qpy.QPY_VERSION` (or `qiskit.qpy.common.QPY_VERSION` on Qiskit < 1.0.0) to see what the maximum supported QPY version for `qpy.load()` is but being able to look this data up in the docs would be helpful.

We should add a small table to the qpy module docs that just enumerates the value of `QPY_VERSION` for each release.

mtreinish:
### Summary

This commit adds a new table to the QPY module docs that outlines all the supported QPY format versions for every qiskit/qiskit-terra release that had qpy. This information is potentially useful when trying to navigate generating QPY files across versions.

### Details and comments

Fixes #12283


qiskit/qpy/__init__.py
```diff
@@ -111,6 +111,135 @@
...
+     - 1
+   * - 0.18.3
+     - 1
+     - 1
+   * - 0.18.2
+     - 1
+     - 1
+   * - 0.18.1
+     - 1
+     - 1
```
mtreinish, line 238:
```suggestion
   * - 0.18.2
     - 1
     - 1
   * - 0.18.1
     - 1
     - 1
   * - 0.18.0
     - 1
     - 1
```

jakelishman (approved):
This is a good addition, thanks!

PR #12285, opened 2024-04-20
Removing unnecessary-list-index-lookup from lint exclusions and updates

joesho112358:
### Summary

#9614 - removing unnecessary-list-index-lookup

### Details and comments

Removing unnecessary-list-index-lookup from lint exclusions and updates


levbishop (approved):
Looks  good. Thanks for this.

PR #12286, opened 2024-04-20
Removing consider-using-enumerate from lint exclusions and updates

joesho112358:
### Summary

#9614 - removing consider-using-enumerate

### Details and comments

Updates for removing this rule if wanted. Open to suggestions on variable names


joesho112358:
@mtreinish as a heads up, the latest commit probably needs your eyes. specifically in the `passmanager.py` file. i made a change there that i think is in line with the intention, although a bit weird in implementation.

qiskit/transpiler/passmanager.py
```diff
@@ -121,8 +121,8 @@ def _finalize_layouts(self, dag):
         # Ordered list of original qubits
         original_qubits_reverse = {v: k for k, v in original_qubit_indices.items()}
         original_qubits = []
-        for i in range(len(original_qubits_reverse)):
-            original_qubits.append(original_qubits_reverse[i])
+        for v, k in original_qubits_reverse.items():
+            original_qubits.append(k)
```
jakelishman: This looks like a behavioural change to me.  Dictionary iteration order is insertion order, but the previous form was specifically looking up in _numerical_ order of the `v` items.

The `v`s are ints here (from `0` to `n-1`) and the `k` are `Qubit` instances.

joesho112358: ah, i see it now. ok, i'll have to update this

qiskit/transpiler/passmanager.py
```diff
@@ -119,10 +119,9 @@ def _finalize_layouts(self, dag):
             t_final_layout = Layout(dict(enumerate(t_qubits)))
 
         # Ordered list of original qubits
-        original_qubits_reverse = {v: k for k, v in original_qubit_indices.items()}
-        original_qubits = []
-        for i in range(len(original_qubits_reverse)):
-            original_qubits.append(original_qubits_reverse[i])
+        original_qubits = [None] * len(original_qubit_indices)
+        for k, v in original_qubit_indices.items():
+            original_qubits[v] = k
```
mtreinish: I'm mildly worried about this logic in the failure mode here. If somehow `original_qubit_indices`  was incorrectly constructed and missing an index this wouldn't catch that. We'd never know that there was an entry in the list which was `None` and I just checked that ` virtual_permutation_layout.inverse()` doesn't error if it gets `None` on an input element (which itself seems like a bug).. I liked the previous logic because it would error if you `original_qubit_indices` was incorrectly constructed.

While it's not critical and I think this would not be an issue in practice, I'm just wondering if we need to make this change here? Like if pylint flagging this I'm wondering if a local exclude makes more sense in this case.

mtreinish (commented):
Overall this LGTM, just one question/comment inline about the changes in `passmanager.py`

qiskit/transpiler/passmanager.py
joesho112358: Would it be sufficient to add a check after regenerating `original_qubits`?
```
        if None in original_qubits:
            raise ValueError("original_qubit_indices is missing an index")
```

mtreinish: That would work from a functional PoV, but the issue is that this would be an `O(n)` lookup to check for `None` and we've already done that iteration once so it would be more efficient to do the checking while we're doing the data conversion.

joesho112358: yeah, i guess it ends up just needing 2 loops either way. will add the comment and revert as it is preferred

mtreinish (approved):
Lgtm thanks for doing this.

PR #12288, opened 2024-04-21
Removing consider-using-dict-items from lint exclusions and updates

joesho112358:
### Summary

#9614 test for removing `consider-using-dict-items`

### Details and comments

Updates for removing the consider-using-dict-items rule if wanted. Open to suggestions on variable names


ihincks (dismissed):
LGTM. The primitive object in question only gained `.items()` the other day, so thanks for catching!

qiskit/transpiler/target.py
```diff
@@ -979,7 +981,7 @@ def instruction_properties(self, index):
             InstructionProperties: The instruction properties for the specified instruction tuple
         """
         instruction_properties = [
-            inst_props for op in self._gate_map for _, inst_props in self._gate_map[op].items()
+            inst_props for _, qargs in self._gate_map.items() for _, inst_props in qargs.items()
```
jakelishman: Rather than `for _, x in y.items()`, we might consider using `for x in y.values()`?

qiskit/visualization/circuit/latex.py
```diff
@@ -213,10 +213,9 @@ def _initialize_latex_array(self):
...
 
         # display the bit/register labels
-        for wire in self._wire_map:
+        for wire, index in self._wire_map.items():
             if isinstance(wire, ClassicalRegister):
                 register = wire
-                index = self._wire_map[wire]
             else:
                 register, bit_index, reg_index = get_bit_reg_index(self._circuit, wire)
                 index = bit_index if register is None else reg_index
```
jakelishman: I'm ok to approve this change - it's probably just stylistic - but there's certainly a possible argument that the previous form was clearer, since `index` is overwritten in the `else` branch.

(Same comment in the text drawer too, but no need to change either.)

jakelishman (commented):
Thanks for looking at these.

ihincks:
I should have been clear in my review description: I was looking mainly at the primitives module and was glad for the change. I did look elsewhere for general correctness, but not with an eye to decide whether the change was a net benefit.

qiskit/visualization/circuit/latex.py
joesho112358: I will have to revisit this and give it some thought

jakelishman (commented):
This has gained a couple of minor merge conflicts, but otherwise I think it's good to go.

t-imamichi:
LGTM

jakelishman (approved):
This looks good, thanks for the changes!  (Not sure if the typo fixes in `SECURITY.md` were meant for this PR, but it's fine - they're obviously correct and just minor.)

PR #12292, opened 2024-04-23
Adding `Target` and `InstructionProperties` representations to rust

### LINKED ISSUE #12052: Add `Target` representation to rust

raynelfss:
### Summary

Fixes #12052

~~**Currently blocked by #12459**~~ thanks @mtreinish!

As we continue moving more parts of the transpiler to Rust, it is important that some of the important data about the devices we're compiling for is easily accessible from the Rust side of things. These commits aim to implement representations of  both the `Target` and `InstructionProperties` classes, allowing most of their data to live inside of Rust but still be usable in Python.

### Details and comments

`qiskit._accelerate` now has a `target` module which includes both the `BaseTarget` and `BaseInstructionProperties` structs.
Here's a more detailed description of what has changed:

#### Rust

- `BaseTarget` will include the main data that the `Target` has in python. This will include all of the publicly available attributes of the existing Target class, except for the following:
   - `_coupling_graph`: of type `rustworkx.PyDiGraph`
   - `_instruction_durations`: of type `InstructionDurations`
   - `_instruction_schedule_map`: of type `InstructionScheduleMap`

   Due to these properties are not usable through Rust without using python api's, they were left in python.

   - This representation is purely additive through Python and non mutable through Rust.

- `BaseInstructionProperties` represents the main data of `InstructionProperties, that being:
   - Error: `f64`,
   - Duratlon: `f64`

#### Python:
- `Target`: this is a subclass of `BaseTarget` and has the missing attributes + functions from the previous version.
   - In addition it stores a copy of the `_gate_map` attribute from the rust side for easy access in python without conversions from rust.
   - We ensure that the data from `Target._gate_map` is identical to the data in `super().gate_map` by calling the `super()`'s equivalent function when adding or updating an instruction.
- `InstructionProperties`: is a subclass of `BaseInstructionProperties`
    - The attribute of `calibration` to store the instruction's calibration object is present in this representation along with it's custom setter and getter.

#### Possible breaking changes:
~~- `InstructionProperties` is now PyO3 class which means it cannot be subclassed by overloading `__init__()` alone. The new architecture would require users to extend both `__new__` and `__init__` to add extra properties. These wouldn't necessarily be visible in rust.~~ **Fixed!**

#### Broken tests
~~A procedure performed in Aer that attempts to erase the `Target`'s internal data is no longer possible due to the these attributes being hidden from python.~~ Solved by Qiskit/qiskit-aer#2142

#### Blocks
- [x] raynelfss#2
- [x] #12697
- [x]  #12698

#### Open to review!
Any suggestions, comments, feedback, or any request for changes are encouraged and welcome!


raynelfss:
I wonder if I should wait for #11461  for this to continue, since it introduces a change to the way ``Target.from_configuration()`` works.

crates/accelerate/src/target_transpiler/instruction_properties.rs
```diff
@@ -0,0 +1,78 @@
...
+
+This class provides the optional properties that a backend can provide
+about an instruction. These represent the set that the transpiler can
+currently work with if present. However, if your backend provides additional
+properties for instructions you should subclass this to add additional
+custom attributes for those custom/additional properties by the backend.
+*/
+#[pyclass(subclass, module = "qiskit._accelerate.target")]
+#[derive(Clone, Debug)]
+pub struct BaseInstructionProperties {
```
mtreinish: A suggestion on naming, from Rust space I think we should call this `InstructionProperties` because it's a rust representation of the instruction properties, there isn't anything building off of it in rust space so `Base` is more confusing. We can set a python space name in the `pyclass` macro to make it `BaseInstructionProperties` for python where the name makes more sense.

```suggestion
#[pyclass(subclass, name = "BaseInstructionProperties", module = "qiskit._accelerate.target")]
#[derive(Clone, Debug)]
pub struct InstructionProperties {
```

crates/accelerate/src/target_transpiler/mod.rs
```diff
@@ -0,0 +1,1238 @@
...
+use self::exceptions::TranspilerError;
+
+mod exceptions {
+    use pyo3::import_exception_bound;
+    import_exception_bound! {qiskit.exceptions, QiskitError}
+    import_exception_bound! {qiskit.transpiler.exceptions, TranspilerError}
+}
+
+// Custom types
+type Qargs = SmallVec<[PhysicalQubit; 4]>;
```
mtreinish: Why `4`? In general most targets will have 1 and 2 qubit qargs. So I would set this to `2` to optimize for the common case.

crates/accelerate/src/target_transpiler/mod.rs
```diff
@@ -0,0 +1,1238 @@
...
+type GateMap = IndexMap<String, PropsMap>;
+type PropsMap = IndexMap<Option<Qargs>, Option<BaseInstructionProperties>>;
```
mtreinish: Should you set the hasher here to be ahash?

crates/accelerate/src/target_transpiler/mod.rs
```diff
@@ -0,0 +1,1238 @@
...
+)>;
+
+/// Temporary interpretation of Gate
+#[derive(Debug, Clone)]
+pub struct GateRep {
+    pub object: PyObject,
+    pub num_qubits: Option<usize>,
+    pub label: Option<String>,
+    pub params: Option<Vec<Param>>,
+}
```
mtreinish: Is this temporary in that you're waiting for #12459 to merge to finalize it? What were you thinking this will look like in it's final form?

crates/accelerate/src/target_transpiler/mod.rs
```diff
@@ -0,0 +1,1238 @@
...
+        self.object
+    }
+}
+
+// Temporary interpretation of Param
+#[derive(Debug, Clone, FromPyObject)]
+pub enum Param {
+    Float(f64),
+    ParameterExpression(PyObject),
+}
```
mtreinish: I assume this is based on what I had in #12459, if so I did have to change definition here to be a 3 entry `enum` that also accounted for other types of parameters (since non-standard gates can contain anything as a parameter). I also needed to custom implement `FromPyObject` for this.

crates/accelerate/src/target_transpiler/mod.rs
```diff
@@ -0,0 +1,1238 @@
...
+    a new``Target`` object.
+
+.. note::
+
+    This class only supports additions of gates, qargs, and qubits.
+    If you need to remove one of these the best option is to iterate over
+    an existing object and create a new subset (or use one of the methods
+    to do this). The object internally caches different views and these
+    would potentially be invalidated by removals.
+ */
```
mtreinish: Do we need this docstring still? I'm wondering if a rust native docstring would make more sense, since while this is the base class for `Target` it's not actually something user facing in python. The advantage of having this base class live in rust is that we can get at the data from Rust now. So maybe documenting how this works in the rust domain will make it easier for us to use in Rust.

crates/accelerate/src/target_transpiler/mod.rs
```diff
@@ -0,0 +1,1238 @@
...
+    #[pyo3(get, set)]
+    pub qubit_properties: Option<Vec<PyObject>>,
+    #[pyo3(get, set)]
+    pub concurrent_measurements: Vec<Vec<usize>>,
+    gate_map: GateMap,
+    #[pyo3(get)]
+    _gate_name_map: IndexMap<String, GateRep>,
+    global_operations: IndexMap<usize, HashSet<String>>,
+    variable_class_operations: IndexSet<String>,
+    qarg_gate_map: IndexMap<Option<Qargs>, Option<HashSet<String>>>,
```
mtreinish: I'd probably use ahash for the hasher on all of these, just to speed up the lookups a bit.

crates/accelerate/src/target_transpiler/mod.rs
```diff
@@ -0,0 +1,1238 @@
...
+        self.non_global_basis = state.get_item(13)?.extract::<Option<Vec<String>>>()?;
+        self.non_global_strict_basis = state.get_item(14)?.extract::<Option<Vec<String>>>()?;
+        Ok(())
+    }
+}
+
+// Rust native methods
+impl BaseTarget {
+    /// Generate non global operations if missing
+    fn generate_non_global_op_names(&mut self, strict_direction: bool) -> &Vec<String> {
```
mtreinish: It's typical more idomatic to return a slice like `&[String]` instead of a borrowed Vec. Is there are reason you needed to return the full Vec?

crates/accelerate/src/target_transpiler/mod.rs
```diff
@@ -0,0 +1,1238 @@
...
+    #[pyo3(get, set)]
+    pub granularity: i32,
+    #[pyo3(get, set)]
+    pub min_length: usize,
+    #[pyo3(get, set)]
+    pub pulse_alignment: i32,
+    #[pyo3(get, set)]
+    pub acquire_alignment: i32,
+    #[pyo3(get, set)]
+    pub qubit_properties: Option<Vec<PyObject>>,
```
mtreinish, line 169: You left this as a `PyObject` to avoid having to define a struct for this in Rust? I think this is fine for a temporary thing but we'll want to represent this in Rust eventually (it's not critical for this PR). Can you add a `TODO` comment here to track a follow on work item to move this to rust.

crates/accelerate/src/target_transpiler/mod.rs
```diff
@@ -0,0 +1,1238 @@
...
+                    .entry(qarg_obj)
+                    .and_modify(|e| {
+                        if let Some(e) = e {
+                            e.insert(name.clone());
+                        }
+                    })
+                    .or_insert(Some(HashSet::from([name.clone()])));
+            }
+        }
+        // Modify logic once gates are in rust.
```
mtreinish: Just so we have a common keyword to search for.
```suggestion
        // TODO: Modify logic once gates are in rust.
```

crates/accelerate/src/target_transpiler/mod.rs
```diff
@@ -0,0 +1,1238 @@
...
+    }
+
+    /// Checks if a key exists in the Target
+    pub fn contains_key(&self, key: &String) -> bool {
+        self.gate_map.contains_key(key)
+    }
+}
+
+// To access the Target's gate map by gate name.
+impl Index<&String> for BaseTarget {
```
mtreinish: Can you use `&str` here to make this a bit more generic? Or does it have to be `&String`?

crates/accelerate/src/target_transpiler/mod.rs
```diff
@@ -0,0 +1,1238 @@
...
+        if strict_direction {
+            // Build search set
+            for qarg_key in self.qarg_gate_map.keys().cloned() {
+                search_set.insert(qarg_key);
+            }
+        } else {
+            for qarg_key in self.qarg_gate_map.keys().flatten() {
+                if qarg_key.len() != 1 {
+                    let mut vec = qarg_key.clone();
+                    vec.sort();
```
mtreinish: You can probably use `sort_unstable()` here since there will never be duplicates.

```suggestion
                    vec.sort_unstable();
```

crates/accelerate/src/target_transpiler/mod.rs
```diff
@@ -0,0 +1,1238 @@
...
+        self.non_global_strict_basis = state.get_item(14)?.extract::<Option<Vec<String>>>()?;
+        Ok(())
+    }
+}
+
+// Rust native methods
+impl BaseTarget {
+    /// Generate non global operations if missing
+    fn generate_non_global_op_names(&mut self, strict_direction: bool) -> &Vec<String> {
+        let mut search_set: HashSet<Option<Qargs>> = HashSet::new();
```
mtreinish: With `None` there can't be any other constraints on it because that means it's globally defined. This probably can be simplified if you changed search set to just be HashSet<Qargs> and you detect and  `None` and skip it.

crates/accelerate/src/target_transpiler/mod.rs
```diff
@@ -0,0 +1,1238 @@
...
+    }
+
+    // Magic methods:
+
+    fn __len__(&self) -> PyResult<usize> {
+        Ok(self.gate_map.len())
+    }
+
+    fn __getstate__(&self, py: Python<'_>) -> PyResult<Py<PyList>> {
+        let result_list = PyList::empty_bound(py);
```
mtreinish: Since there are so many fields here I'd probably use a dict, so that you can access by name. With 15 elements tracking which index maps to which field seems quite error prone when we need to update this.

mtreinish (commented):
Thanks for the hard work on this, I've done a quick first pass through most of the code, still have a bit more to get through, but figured I'd leave me comments so far so you can see my thoughts at this point.

I also have a couple of high level questions I have is how were you planning to work with this in conjunction with #12459? Were you think go forward with the temporary structures you've defined and integrate after the fact, or wait until #12459 merges and rebase on top of that? I feel like whatever we end up with in #12459 will inform your design here.

The other musing I have in the back of my head is thinking about whether this hybrid Python and Rust structure is the approach we want to go with. I've been thinking for some time it might make sense to just make a Rust native Target struct that contains the information we need from a target natively in Rust.  Unlike for gates/circuit and the dag the target is unique that when we're working with it in Rust we never need to mutate it. So doing this split approach where we have a read-only rust space view of the data from a Python target would be fine. We could build this struct from Python on demand and cache it in the python object so that we only build it once (and invalidate  the cached item on subsequent target modification in Python) The reason I'm thinking this is while this hybrid approach works, I worry we're tying a few too many Python-isms into this struct. For example, `instruction_supported()` is returning a `PyResult` and from a rust perspective I feel like this method should be infallible, a target either supports a given instruction or not. What we really need here is to have a convenient and ergonomic way for transpiler passes to query the constraints from a target directly in rust without having to worry about Python.  That being said if we went with what I'm suggesting we'd end up with duplicate implementations so it's not a 100% clear cut decision for me.

crates/accelerate/src/target_transpiler/mod.rs
raynelfss: Alright, I'll just wait for the one used in your PR. Since it requires certain imports that are only present there.

crates/accelerate/src/target_transpiler/mod.rs
raynelfss: Also temporary, this makes it easier to access properties of the gate without having to extract them every time. But it will adapt better to whatever changes your PR includes.

crates/circuit/src/operations.rs
```diff
@@ -184,6 +184,43 @@ impl ToPyObject for Param {
...
+        match (self, other) {
+            (Param::Float(s), Param::Float(other)) => s == other,
+            (Param::ParameterExpression(one), Param::ParameterExpression(other)) => {
+                Self::compare(one, other)
+            }
+            (Param::Obj(one), Param::Obj(other)) => Self::compare(one, other),
+            _ => false,
+        }
+    }
+}
```
mtreinish: I'd be worried about the overhead here. If we need to do this across thread boundaries the `with_gil()` in `compare()` will serialize execution and be generally really slow. We can probably keep `compare()` and give it a `py: Python` argument so we can control how the gil is handled. But we should be making it explicit here.  I'm a bit worried about this with the `Operation` trait too, and was planning to investigate potentially rework it once we start having parallelism in passes. But for something like `PartialEq` which is in the stdlib I feel the risk is a lot higher because we might not even realize this.

crates/circuit/src/operations.rs
```diff
@@ -184,6 +184,43 @@ impl ToPyObject for Param {
...
+            match self {
+                Param::ParameterExpression(obj) => obj.call_method0(py, "__repr__")?.extract(py),
+                Param::Float(float_param) => Ok(format!("Parameter({})", float_param)),
+                Param::Obj(obj) => obj.call_method0(py, "__repr__")?.extract(py),
+            }
+        })
+        .unwrap_or("None".to_owned());
+        write!(f, "{}", display_name)
+    }
+}
```
mtreinish: Is this needed for anything? If not lets save it for a standalone PR and discuss it in that.

qiskit/transpiler/target.py
```diff
@@ -220,38 +238,19 @@ class Target(Mapping):
...
-        "granularity",
-        "min_length",
-        "pulse_alignment",
-        "acquire_alignment",
-        "_non_global_basis",
-        "_non_global_strict_basis",
-        "qubit_properties",
-        "_global_operations",
-        "concurrent_measurements",
-    )
```
mtreinish: Are slots incompatible with being based on a rust pyclass? If not I would suggest leaving the slots for the python space attributes defined in this class so that the python space component of the class is slotted still.

qiskit/transpiler/target.py
```diff
@@ -295,37 +294,45 @@ def __init__(
...
         """
-        self.num_qubits = num_qubits
-        # A mapping of gate name -> gate instance
-        self._gate_name_map = {}
+
+        # In case a number is passed as first argument, assume it means num_qubits.
+        if description is not None:
+            if num_qubits is None and isinstance(description, int):
+                num_qubits = description
+                description = None
```
mtreinish: I see the intent here and it's to correct a lot of longstanding behavior around the API implications of this. While I get that a lot of the tests are doing `Target(2)` or something similar (especially since I wrote those) this is changing behavior. Like what would happen if a user did `Target(2, num_qubits=4)` then what would you do? I'd say hold off on doing this just for the sake of compatibility. If we want to make this change we should do it in isolation so we can discuss the backwards compatibility implications of doing this.

qiskit/transpiler/target.py
```diff
@@ -525,7 +509,7 @@ def update_from_instruction_schedule_map(self, inst_map, inst_name_map=None, err
                         continue
                 try:
                     # Update gate error if provided.
-                    props.error = error_dict[inst_name][qargs]
+                    setattr(props, "error", error_dict[inst_name][qargs])
```
mtreinish: Is there a reason to change this syntax? Using `setattr()` is fine, but I don't see any difference with this since error always has to exist right?

crates/accelerate/src/target_transpiler/instruction_properties.rs
```diff
@@ -0,0 +1,76 @@
...
+            output.push_str("duration=None, ");
+        }
+
+        if let Some(error) = self.error {
+            output.push_str("error=");
+            output.push_str(error.to_string().as_str());
+            output.push_str(", ");
+        } else {
+            output.push_str("error=None, ");
+        }
```
mtreinish: While not performance critical I would try to do this as a single `format!()` call because each `push_str()` will potentially result in another allocation to expand the string. You can just move the if statements to the top of the method and generate the str to substitute into `format!()`.

crates/accelerate/src/target_transpiler/mod.rs
```diff
@@ -0,0 +1,1285 @@
...
+// Custom types
+type Qargs = SmallVec<[PhysicalQubit; 2]>;
+type GateMap = IndexMap<String, PropsMap, RandomState>;
+type PropsMap = IndexMap<Option<Qargs>, Option<InstructionProperties>, RandomState>;
+type GateMapState = Vec<(String, Vec<(Option<Qargs>, Option<InstructionProperties>)>)>;
+
+#[derive(Debug, Clone, FromPyObject)]
+pub enum TargetOperation {
+    Normal(NormalOperation),
+    Variable(VariableOperation),
```
mtreinish: For `Variable` here are you referring to operations which can take a variable number of qubits? If so the normal term we use for that is "variadic" (https://en.wikipedia.org/wiki/Variadic_function)

crates/accelerate/src/target_transpiler/mod.rs
```diff
@@ -0,0 +1,1285 @@
...
+            Self::Variable(variable) => variable.to_object(py),
+        }
+    }
+}
+
+impl TargetOperation {
+    fn num_qubits(&self) -> u32 {
+        match &self {
+            Self::Normal(normal) => normal.operation.num_qubits(),
+            Self::Variable(_) => 0,
```
mtreinish: Is this correct? Or are you using 0 as a sentinel value here? Maybe I don't understand what `Variable` is here.

crates/accelerate/src/target_transpiler/mod.rs
```diff
@@ -0,0 +1,1285 @@
...
+            TargetOperation::Variable(_) => &[],
+        }
+    }
+}
+
+#[derive(Debug, Clone)]
+pub struct NormalOperation {
+    operation: OperationType,
+    params: SmallVec<[Param; 3]>,
+    internal_name: String,
```
mtreinish: Is `internal_name` the name as defined in the target? I was a little confused by "internal" here, like I wasn't sure if you meant the name from the `OperationType` object or internal to the target.

crates/accelerate/src/target_transpiler/mod.rs
```diff
@@ -0,0 +1,1285 @@
...
+    #[pyo3(get, set)]
+    pub dt: Option<f64>,
+    #[pyo3(get, set)]
+    pub granularity: i32,
+    #[pyo3(get, set)]
+    pub min_length: usize,
+    #[pyo3(get, set)]
+    pub pulse_alignment: i32,
+    #[pyo3(get, set)]
+    pub acquire_alignment: i32,
```
mtreinish: Is there a reason you used `i32` here? These fields can't really be negative as they're just an integer from from 1 to n which are a multiple of `dt`. I would use `u32` for all of these, but we can probably get away with `u8` or `u16` realistically since I don't think this will typically be > 255 or >65k. But maybe @nkanazawa1989 has some thoughts on this.

crates/accelerate/src/target_transpiler/mod.rs
```diff
@@ -0,0 +1,1285 @@
...
+    #[pyo3(get, set)]
+    // TODO: Port to Rust.
+    pub qubit_properties: Option<Vec<PyObject>>,
+    #[pyo3(get, set)]
+    pub concurrent_measurements: Vec<Vec<usize>>,
```
mtreinish: This might make more sense, since it's a grouping of qubit indices.
```suggestion
    pub concurrent_measurements: Vec<Vec<PhysicalQubit>>,
```

crates/accelerate/src/target_transpiler/mod.rs
```diff
@@ -0,0 +1,1285 @@
...
+        Ok(Target {
+            description,
+            num_qubits,
+            dt,
+            granularity: granularity.unwrap_or(1),
+            min_length: min_length.unwrap_or(1),
+            pulse_alignment: pulse_alignment.unwrap_or(1),
+            acquire_alignment: acquire_alignment.unwrap_or(0),
+            qubit_properties,
+            concurrent_measurements: concurrent_measurements.unwrap_or(Vec::new()),
```
mtreinish: Was there any checking or semantics around the usage of this with `None`? Or is an empty list treated the same way in the python side?

crates/accelerate/src/target_transpiler/mod.rs
```diff
@@ -0,0 +1,1285 @@
...
+                                if let Some(e) = e {
+                                    e.insert(name.clone());
+                                }
+                            })
+                            .or_insert(Some(HashSet::from([name.clone()])));
+                    }
+                }
+            }
+        }
+        // TODO: Modify logic once gates are in rust.
```
mtreinish: Gates are in rust now, is this still relevant?

crates/accelerate/src/target_transpiler/mod.rs
```diff
@@ -0,0 +1,1285 @@
...
+    ///         names. Backends can differentiate between different
+    ///         parameterization of a single gate by providing a unique name for
+    ///         each (e.g. `"rx30"`, `"rx60", ``"rx90"`` similar to the example in the
+    ///         documentation for the :class:`~qiskit.transpiler.Target` class).
+    /// Raises:
+    ///     AttributeError: If gate is already in map
+    ///     TranspilerError: If an operation class is passed in for ``instruction`` and no name
+    ///         is specified or ``properties`` is set.
+    #[pyo3(signature = (instruction, name, properties=None))]
+    fn add_instruction(
```
mtreinish, line 292: I'm wondering if we want to make this different somehow. I was initially confused because the signature here is different from the python space methods. and I thought at first this was being exposed to end users (which would have been a backwards incompatible change). This might just be an issue with my review since I'm trying to keep too much state in my head at once.

crates/accelerate/src/target_transpiler/mod.rs
```diff
@@ -0,0 +1,1285 @@
...
+        name: String,
+        properties: Option<PropsMap>,
+    ) -> PyResult<()> {
+        if self.gate_map.contains_key(&name) {
+            return Err(PyAttributeError::new_err(format!(
+                "Instruction {:?} is already in the target",
+                name
+            )));
+        }
+        let mut qargs_val: PropsMap = PropsMap::default();
```
mtreinish: Could we use `with_capacity()` here since  this will be unused or have `properties.len()`?

crates/accelerate/src/target_transpiler/mod.rs
```diff
@@ -0,0 +1,1285 @@
...
+                                        if acc > x.index() {
+                                            acc
+                                        } else {
+                                            x.index()
+                                        }
+                                    }) + 1,
+                                ));
+                            qarg_obj = Some(qarg.clone())
+                        }
+                        qargs_val.insert(qarg_obj.to_owned(), properties[qarg].clone());
```
mtreinish: This clone seems a bit weird, since we own `properties` I would expect us to be able to just move the data into `gate_map` directly.

crates/accelerate/src/target_transpiler/mod.rs
```diff
@@ -0,0 +1,1285 @@
...
+
+    /// Get the qargs for a given operation name
+    ///
+    /// Args:
+    ///     operation (str): The operation name to get qargs for
+    /// Returns:
+    ///     set: The set of qargs the gate instance applies to.
+    #[pyo3(name = "qargs_for_operation_name")]
+    fn py_qargs_for_operation_name(&self, operation: String) -> PyResult<Option<Vec<Qargs>>> {
+        match self.qargs_for_operation_name(&operation) {
```
mtreinish: I think this should work: 
```suggestion
    fn py_qargs_for_operation_name(&self, operation: &str) -> PyResult<Option<Vec<Qargs>>> {
        match self.qargs_for_operation_name(operation) {
```

crates/accelerate/src/target_transpiler/mod.rs
```diff
@@ -0,0 +1,1285 @@
...
+    /// Args:
+    ///     instruction (str): The instruction name to get the
+    ///         :class:`~qiskit.circuit.Instruction` instance for
+    /// Returns:
+    ///     qiskit.circuit.Instruction: The Instruction instance corresponding to the
+    ///     name. This also can also be the class for globally defined variable with
+    ///     operations.
+    #[pyo3(name = "operation_from_name")]
+    fn py_operation_from_name(&self, instruction: String) -> PyResult<TargetOperation> {
+        match self.operation_from_name(&instruction) {
```
mtreinish:
```suggestion
    fn py_operation_from_name(&self, instruction: &str) -> PyResult<TargetOperation> {
        match self.operation_from_name(instruction) {
```

crates/accelerate/src/target_transpiler/mod.rs
```diff
@@ -0,0 +1,1285 @@
...
+        self.gate_map.iter().flat_map(move |(op, props_map)| {
+            props_map
+                .keys()
+                .map(move |qargs| (&self._gate_name_map[op], qargs.as_ref()))
+        })
+    }
+    /// Returns an iterator over the operation names in the target.
+    pub fn operation_names(&self) -> impl Iterator<Item = &String> {
+        self.gate_map.keys()
+    }
```
mtreinish: `&String` is a bit weird I'd expect `&str` 
```suggestion
    pub fn operation_names(&self) -> impl Iterator<Item = &str> {
        self.gate_map.keys()
    }
```

But I'm not sure the compiler is smart enough to map that, so you might need:

```suggestion
    pub fn operation_names(&self) -> impl Iterator<Item = &str> {
        self.gate_map.keys().map(|x| x.as_str())
    }
```

crates/accelerate/src/target_transpiler/mod.rs
```diff
@@ -0,0 +1,1285 @@
...
+            }
+            *size_dict.entry(qarg.len()).or_insert(0) += 1;
+        }
+        for (inst, qargs_props) in self.gate_map.iter() {
+            let mut qarg_len = qargs_props.len();
+            let qargs_keys: IndexSet<&Option<Qargs>> = qargs_props.keys().collect();
+            let qarg_sample = qargs_keys.iter().next().cloned();
+            if let Some(qarg_sample) = qarg_sample {
+                if !strict_direction {
+                    let mut qarg_set: HashSet<SmallVec<[PhysicalQubit; 2]>, RandomState> =
```
mtreinish: I might rename this `deduplicated_qargs` because it wasn't entirely clear what the purpose was here. I assume it's for deduplication because otherwise `qarg_len` will always be `qargs_keys.len()`. That being said I wonder if we can move this definition outside the loop and reuse a single allocation for each instruction.

crates/accelerate/src/target_transpiler/mod.rs
```diff
@@ -0,0 +1,1285 @@
...
+            return self.non_global_basis.as_deref();
+        }
+        return Some(self.generate_non_global_op_names(strict_direction));
+    }
+
+    /// Gets all the operation names that use these qargs. Rust native equivalent of ``BaseTarget.operation_names_for_qargs()``
+    pub fn operation_names_for_qargs(
+        &self,
+        qargs: &Option<Qargs>,
+    ) -> Result<HashSet<&String, RandomState>, TargetKeyError> {
```
mtreinish: I would use `&str` here instead of `&String` also if you're using hashbrown you don't need to set `RandomState` since that's the default (which is a big part of why we use it).

crates/accelerate/src/target_transpiler/mod.rs
```diff
@@ -0,0 +1,1285 @@
...
+                qargs
+            )));
+        }
+        Ok(res)
+    }
+
+    /// Returns rust-native operation instances present in the Target that affect the provided qargs.
+    pub fn ops_from_qargs(
+        &self,
+        qargs: &Option<Qargs>,
```
mtreinish: It's a bit more idiomatic to use `Option<&Qargs>` because you can convert from `&Option<T>` -> `Option<&T>` with `Option.as_ref()` but you can't convert the other way.

crates/accelerate/src/target_transpiler/mod.rs
```diff
@@ -0,0 +1,1285 @@
...
+                })
+                .collect()),
+            Err(e) => Err(e),
+        }
+    }
+
+    /// Gets all the qargs used by the specified operation name. Rust native equivalent of ``BaseTarget.qargs_for_operation_name()``
+    pub fn qargs_for_operation_name(
+        &self,
+        operation: &String,
```
mtreinish: I would use `&str` here

crates/accelerate/src/target_transpiler/mod.rs
```diff
@@ -0,0 +1,1285 @@
...
+            return None;
+        }
+        Some(qargs)
+    }
+
+    pub fn instruction_supported(
+        &self,
+        operation_name: &String,
+        qargs: &Option<Qargs>,
+        parameters: &Option<Vec<Param>>,
```
mtreinish: Same comments here about argument typing from the earlier methods.

crates/accelerate/src/target_transpiler/mod.rs
```diff
@@ -0,0 +1,1285 @@
...
+
+    /// Retreive all the gate names in the Target
+    pub fn keys(&self) -> Keys<String, PropsMap> {
+        self.gate_map.keys()
+    }
+
+    /// Retrieves an iterator over the property maps stored within the Target
+    pub fn values(&self) -> Values<String, PropsMap> {
+        self.gate_map.values()
+    }
```
mtreinish: Instead of using `Keys` and `Values` here is there a reason to not just use `impl Iterator` like you've done for the other APIs?

mtreinish (commented):
I did a more complete pass through the code, I think this is getting closer thanks for sticking with this.

crates/circuit/src/operations.rs
raynelfss: This will be removed based on Jake's comments of #12697, I am making updates to this.

crates/circuit/src/operations.rs
raynelfss: This will also be removed as per #12697

qiskit/transpiler/target.py
raynelfss: Got it, it also reminds me of what @1ucian0 said about changing the behavior based on typing. This will be removed in future commits.

qiskit/transpiler/target.py
raynelfss: Yes, there should be no difference, this was old code that behaved based on having `InstructionProperties` be wrapped in a `Py` reference. This will be promptly moved back to its original syntax.

Edit: Okay, so the actual reason this was done was more because of formatting, PyLint was complaining about invisible attributes since Python does not know of the existence of the rust-only attributes and I had to make it this way.

crates/accelerate/src/target_transpiler/mod.rs
raynelfss: So `Variable` is when we provide a class instead of a class instance because it could have a variable amount of qubits (I believe we only do this for `ControlFlowOps`). And so if you call for `num_qubits` it will return a `@property`. We typically wouldn't have a need to call for this on a `VariableOp` as the `Target` operates on them differently, so I could just make this `unreachable!()`.

crates/accelerate/src/target_transpiler/mod.rs
raynelfss: It's a little different from the python one because this function is overloaded in the python side. Whereas in python providing a name could be skipped, here it won't be needed due to Python already taking care of that already. Since we spoke of just making the Target only modifiable through Python and just having its data be accessible in rust. The `add_instruction` and `update_instruction_properties` methods are only exposed to python and not rust. And there are extra checks performed in python before they're eventually sent here.

qiskit/transpiler/target.py
raynelfss: I will add slots for those attributes that live in Python space, it seems that extending a `PyClass` does not accept slots for that class.

crates/accelerate/src/target_transpiler/mod.rs
raynelfss: This will be removed, it was a way of comparing instructions by class names instead of by instance, in the case we had to.

crates/accelerate/src/target_transpiler/mod.rs
raynelfss: I'll set it to `u32` fore now, but will wait on further conversation on this.

crates/accelerate/src/target_transpiler/mod.rs
raynelfss: Seems to work the same with None as it does with an empty vec. I will change it to None as it was in the original `Target`.

crates/accelerate/src/target_transpiler/mod.rs
raynelfss: I will use `IndexMap::swap_remove()` to avoid cloning.

crates/accelerate/src/target_transpiler/mod.rs
raynelfss: Will rename, and I think it is already a single allocation for each instruction as these deduplicated `qargs` are those that are part of the instruction's property map.

crates/accelerate/src/target_transpiler/mod.rs
raynelfss: This will be a bit tricky to implement as it will require us to perform cloning inside of the function. I will leave this item open for further discussion.

raynelfss: I implemented an alternative structure that wraps around the `IndexMap` and can process `Option<&K>` as an acceptable index.

crates/accelerate/src/target_transpiler/mod.rs
```diff
@@ -0,0 +1,1298 @@
...
+    #[pyo3(get, set)]
+    pub concurrent_measurements: Option<Vec<Vec<PhysicalQubit>>>,
+    gate_map: GateMap,
+    #[pyo3(get)]
+    _gate_name_map: IndexMap<String, TargetOperation, RandomState>,
+    global_operations: IndexMap<u32, HashSet<String>, RandomState>,
+    variable_class_operations: IndexSet<String, RandomState>,
+    qarg_gate_map: NullableIndexMap<Qargs, Option<HashSet<String>>>,
+    non_global_strict_basis: Option<Vec<String>>,
+    non_global_basis: Option<Vec<String>>,
```
mtreinish, line 179: This is more a comment for a future performance optimization to investigate in a follow up PR (we can convert this to a tracking issue after the PR merges)

All the `String` entries here are all the same set of strings. Ideally we should have only one owned string copy as the canonical name and the others are all just `&str` references to that single string. We might have issues with lifetimes on this as a pyclass, but we could do something like have a `gate_names: Vec<String>` and then replace all the strings by a usize index.

crates/accelerate/src/target_transpiler/mod.rs
```diff
@@ -0,0 +1,1298 @@
...
+            Self::Variadic(variable) => variable.clone_ref(py),
+        }
+    }
+}
+
+impl TargetOperation {
+    fn num_qubits(&self) -> u32 {
+        match &self {
+            Self::Normal(normal) => normal.operation.num_qubits(),
+            Self::Variadic(_) => 0,
```
mtreinish: I think we should either use `u32::MAX` here or `unreachable!()`. This 0 still seems off to me since from the target perspective there are no qubit constraints on a variadic operation.

crates/accelerate/src/target_transpiler/mod.rs
```diff
@@ -0,0 +1,1298 @@
...
+
+The intent of the ``Target`` object is to inform Qiskit's compiler about
+the constraints of a particular backend so the compiler can compile an
+input circuit to something that works and is optimized for a device. It
+currently contains a description of instructions on a backend and their
+properties as well as some timing information. However, this exact
+interface may evolve over time as the needs of the compiler change. These
+changes will be done in a backwards compatible and controlled manner when
+they are made (either through versioning, subclassing, or mixins) to add
+on to the set of information exposed by a target.
```
mtreinish: Can you put a description of how `BaseTarget` is defined and used here? I think it's important to document what this class is and not just repeat the Python class's docstring.

crates/accelerate/src/target_transpiler/mod.rs
mtreinish: Ok, that makes sense. My comment was more about the method name and the docstring identical to the python one. This is really a different method from the python space `add_instruction` it's a subset of the functionality for handling the rust data but everything is showing it looks like the python space.

crates/accelerate/src/target_transpiler/mod.rs
```diff
@@ -0,0 +1,1298 @@
...
+    fn py_get_non_global_operation_names(
+        &mut self,
+        py: Python<'_>,
+        strict_direction: bool,
+    ) -> PyObject {
+        self.get_non_global_operation_names(strict_direction)
+            .to_object(py)
+    }
+
+    // Class properties
```
mtreinish:
```suggestion
    // Instance attributes
```

or `properties` are fine too. But they're not class properties because they're defined for each object and not at the class level.

crates/accelerate/src/target_transpiler/nullable_index_map.rs
```diff
@@ -0,0 +1,455 @@
...
+/// storage of `Option<K>` key values as `K`` and keep an extra slot reserved only for the
+/// `None` instance. There are some upsides to this including:
+///
+/// The ability to index using Option<&K> to index a specific key.
+/// Store keys as non option wrapped to obtain references to K instead of reference to Option<K>.
+///
+/// **Warning:** This is an experimental feature and should be used with care as it does not
+/// fully implement all the methods present in `IndexMap<K, V>` due to API limitations.
+#[derive(Debug, Clone)]
+pub struct NullableIndexMap<K, V>
```
mtreinish: With this warning do we want to make this fully public, or can we constrain this to `pub(crate)`? This is really only used internally for the Target right, so we may return it to python (but that doesn't matter for this) and we can avoid exposing this directly for accessing the target from rust right? (although all the usage right now is potentially coming from the same crate)

qiskit/transpiler/target.py
```diff
@@ -22,15 +22,19 @@
...
-from collections import defaultdict
 import datetime
 import io
 import logging
 import inspect
 
 import rustworkx as rx
 
+# import target class from the rust side
+from qiskit._accelerate.target import (  # pylint: disable=unused-import
```
mtreinish: Was pylint actually flagging these as unused?

qiskit/transpiler/target.py
```diff
@@ -135,8 +149,16 @@ def __repr__(self):
...
+        return (super().__getstate__(), self.calibration, self._calibration)
+
+    def __setstate__(self, state: tuple):
+        super().__setstate__(state[0])
+        self.calibration = state[1]
+        self._calibration = state[2]
+
 
-class Target(Mapping):
+class Target(BaseTarget):
```
mtreinish, line 162: I feel like dropping `Mapping` here might be a breaking api change. Can we add this as a metaclass or something so we've registered the class as a mapping in the same way.

mtreinish (commented):
Overall I think this is about ready to merge. I think there will definitely be an interative process where we refine this as we start using it more. But this has all the pieces we need to start and only will cost a bit extra memory for a rust space copy of the properties. I left a few inline comments, but the big things I think we must fix before moving forward here are the docstrings for `BaseTarget` should describe `BaseTarget`'s functionality and what/how it is to be used. Even though it's a private class and only for Qiskit internals having the docstrings read like the python space target is confusing.

The other big one is I think dropping `Mapping` from the Python space `Target` is a breaking api change and we should be able to avoid that, I think `metaclass=Mapping` or something similar will work to fix that and let you inherit from `BaseTarget` and still register the class as a `Mapping`.

crates/accelerate/src/target_transpiler/nullable_index_map.rs
raynelfss: Well, with this one it's a little complicated. Even though it is currently being used only in `Target`, it might be usable for other things in the future where we have to store a nullable value inside of a dict-like structure. And maybe someone could build on top of it.

However, the main reason why I made it public is that if we call `Target.values` the value that is returned is an instance of `NullableIndexMap` and, if we wanted to avoid that, we'd have to reconstruct the `IndexMap` entirely. Of course, there might be a better way to approach this so I'm curious as to which ideas you might have.

crates/accelerate/src/target_transpiler/mod.rs
```diff
@@ -0,0 +1,1253 @@
...
+            TargetOperation::Variadic(_) => &[],
+        }
+    }
+}
+
+#[derive(Debug, Clone)]
+pub(crate) struct NormalOperation {
+    pub operation: OperationType,
+    pub params: SmallVec<[Param; 3]>,
+    op_object: PyObject,
```
mtreinish, line 108: Isn't this largely duplicative with the other fields? Is there something we gain by storing this too? Is it just so you don't have to construct it on the fly again if a user requests the operation object for a name?

crates/accelerate/src/target_transpiler/mod.rs
```diff
@@ -0,0 +1,1253 @@
...
+        Ok(())
+    }
+}
+
+// Rust native methods
+impl Target {
+    /// Returns an iterator over all the instructions present in the `Target`
+    /// as pair of `&OperationType`, `&SmallVec<[Param; 3]>` and `Option<&Qargs>`.
+    // TODO: Remove once `Target` is being consumed.
+    #[allow(dead_code)]
```
mtreinish, line 908: Another note for a follow-up we should add rust space tests for these, since they're meant to be consumed from rust we should be able to test them without python.

qiskit/transpiler/target.py
```diff
@@ -994,25 +740,32 @@ def _build_coupling_graph(self):
...
                     return
                 continue
             for qarg, properties in qarg_map.items():
                 if qarg is None:
-                    if self._gate_name_map[gate].num_qubits == 2:
-                        self._coupling_graph = None
+                    if self.operation_from_name(gate).num_qubits == 2:
+                        self._coupling_graph = (
+                            None  # pylint: disable=attribute-defined-outside-init
+                        )
```
mtreinish: Is this pylint disable still needed? https://github.com/Qiskit/qiskit/pull/12292/files#diff-00276708f86c726ae04a2c670e74177eaad6aad971e017951bb5792501d13bf5R338 seems to define it in `__init__`

qiskit/transpiler/target.py
```diff
@@ -1188,10 +893,12 @@ def __str__(self):
                 prop_str_pieces = [f"\t\t{qarg}:\n"]
                 duration = getattr(props, "duration", None)
                 if duration is not None:
-                    prop_str_pieces.append(f"\t\t\tDuration: {duration} sec.\n")
+                    prop_str_pieces.append(
+                        f"\t\t\tDuration: {duration if duration > 0 else 0} sec.\n"
```
mtreinish: When is `duration < 0`?

qiskit/transpiler/target.py
```diff
@@ -1188,10 +893,12 @@ def __str__(self):
...
+                    )
                 error = getattr(props, "error", None)
                 if error is not None:
-                    prop_str_pieces.append(f"\t\t\tError Rate: {error}\n")
+                    prop_str_pieces.append(f"\t\t\tError Rate: {error if error > 0 else 0}\n")
```
mtreinish: Same question here when is `error < 0`

crates/accelerate/src/target_transpiler/mod.rs
```diff
@@ -0,0 +1,1253 @@
...
+}
+
+/**
+The base class for a Python ``Target`` object. Contains data representing the
+constraints of a particular backend.
+
+The intent of this struct is to contain data that can be representable and
+accessible through both Rust and Python, so it can be used for rust-based
+transpiler processes.
+ */
```
mtreinish: The only thing I think we should mention here is around where we have copies of data or not and how we keep things in sync. Just so people will understand the split in responsibilities between the python space and the rust space. Otherwise this is much better.

mtreinish (commented):
I think this is ready to go now. I have a few very small nits inline. But for a 2k line change I think I can continue to look at this and find something forever. so after these are fixed I'll approve and we can merge this. I think there are some potential cleanups we'll want to do down the road, but this is a great start and gets us the data model for the target in rust to start leveraging in various places.

This will conflict with @jakelishman's PR #12730 I think because that removes `OperationType`  but it shouldn't be too hard to update this with that or vice versa if this merges first.

qiskit/transpiler/target.py
raynelfss: This is so that Python doesn't display the floating point number at 0. Of course there might be a better way of doing this.

qiskit/transpiler/target.py
raynelfss: > This is so that Python doesn't display the floating point number at 0. Of course there might be a better way of doing this.

Same here.

crates/accelerate/src/target_transpiler/mod.rs
raynelfss: Yes, that is exactly the purpose of this. So that we don't have to reconstruct the `Operation` object. However, I understand that Jake's changes in #12730 might reshape the way this is done. So storing the `op_object` might not be necessary after that merges, so maybe we should wait for that PR.

mtreinish (approved):
LGTM, thanks for all the hard work on this @raynelfss. I think this is ready to merge. There are some small things I outlined in my earlier reviews that we can fix in follow ups. This is a new rust space representation so just like with everything we're migrating over there will likely be a longish tail of follow-ups we'll make to make incremental improvements as we start to use it more. But this is definitely in a state where I think it's good to start working with.

This all being said I'm going to hold off on pushing the automerge button right now, because I think we should wait until we branch 1.2.0 for this. Since nothing in 1.2.0 will use this I think it's better to just include this in 1.3.0.

IvanaGyro:
@raynelfss Hi, is making `BaseTarget.num_qubits` immutable intended? [Many tests in Qiskit Dynamics](https://github.com/qiskit-community/qiskit-dynamics/actions/runs/12726242543/job/36024767691?pr=368) failed because they tried to set the value of `num_qubits`. I am not sure if I have to fix all tests to let the tiny fix: https://github.com/qiskit-community/qiskit-dynamics/pull/368 be merged.

code of `BaseTarget.num_qubits`: https://github.com/Qiskit/qiskit/blame/e9ccd3f374fd5424214361d47febacfa5919e1e3/crates/accelerate/src/target_transpiler/mod.rs#L151-L152

PR #12294, opened 2024-04-24
Removing modified-iterating-list from lint exclusions and updates

joesho112358:
### Summary

#9614 test for removing `modified-iterating-list`

### Details and comments

Updated the single rule break


jakelishman (approved):
Thanks for this - if nothing else, for highly commutative graphs this is potentially a performance improvement too.

PR #12295, opened 2024-04-24
Removing nested-min-max from lint exclusions and updates

joesho112358:
### Summary

#9614 test for removing `nested-min-max`

### Details and comments

Flattening nested `max`'s

### DIFF
--- a/qiskit/transpiler/passes/layout/vf2_utils.py
+++ b/qiskit/transpiler/passes/layout/vf2_utils.py
@@ -129,7 +129,7 @@ def score_layout(
 ):
     """Score a layout given an average error map."""
     if layout_mapping:
-        size = max(max(layout_mapping), max(layout_mapping.values()))
+        size = max(layout_mapping, layout_mapping.values())
     else:
         size = 0
     nlayout = NLayout(layout_mapping, size + 1, size + 1)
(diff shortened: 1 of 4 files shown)


qiskit/transpiler/passes/layout/vf2_utils.py
```diff
@@ -129,7 +129,7 @@ def score_layout(
 ):
     """Score a layout given an average error map."""
     if layout_mapping:
-        size = max(max(layout_mapping), max(layout_mapping.values()))
+        size = max(layout_mapping, layout_mapping.values())
```
jakelishman, line 132: These don't quite mean the same thing - `max` in one-argument form consumes the iterable and with two-or-more arguments it returns the max argument without iterating any individual argument.

If pylint is suggesting this, it's a bug, because they're not the same thing.  (If it _is_ a bug, there's a good chance it's fixed on pylint main - we're not using the newest version.)

joesho112358: Righto! Looks like this will need to be closed and the `nested-min-max` should be moved up to permanent with a note. I can do that later on. Thanks for the look!

joesho112358:
This looks to be necessary and will need a new PR for commenting why `nested-min-max` is needed.

PR #12296, opened 2024-04-25
Adding note for nested-min-max for linting

joesho112358:
### Summary

#9614 adding `nested-min-max` to the more permanent section

### Details and comments

Discovered in #12295 that this lint rule is necessary for the current versions


jakelishman (approved):
thanks!

PR #12297, opened 2024-04-25
Removing use-dict-literal from lint exclusions and updates

joesho112358:
### Summary

### Details and comments

Removing `use-dict-literal` from lint exclusions and updates


test/python/result/test_result.py
```diff
@@ -56,7 +56,7 @@ def test_counts_no_header(self):
         no_header_processed_counts = {
             bin(int(bs[2:], 16))[2:]: counts for (bs, counts) in raw_counts.items()
         }
-        data = models.ExperimentResultData(counts=dict(**raw_counts))
+        data = models.ExperimentResultData(counts={**raw_counts})
```
jakelishman: In all places in this file, `raw_counts` is already a `dict`.  We can either directly pass `raw_counts` as the argument, or, if a copy _needs_ to be made (I don't think it does), pass `raw_counts.copy()`.

jakelishman (approved):
Ace, thanks!

PR #9064, opened 2022-11-02
Split out NLayout rust class into standalone Python module

mtreinish:
### Summary

This commit moves the NLayout rust class out of the stochastic swap python module into a new standalone nlayout module in qiskit._accelerate. The NLayout class was originally added with the stochastic swap rust code, but since then it's started being used by other rust code including SabreSwap (and soon to be VF2Layout and VF2PostLayout scoring in #9026).

### Details and comments

PR #12298, opened 2024-04-25
Fix pickling of `NLayout`

jakelishman:
### Summary

Previously the module of this was set incorrectly (stemming from its move in gh-9064), at which point the `__getstate__`/`__setstate__` pickling wouldn't work correctly any more.  Also, however, there was no `__getnewargs__` and `new` didn't have a zero-argument form, so this wouldn't have worked either.

### Details and comments

No changelog because this is a private class, and it doesn't actually seem to be causing us problems.

### DIFF
--- a/crates/accelerate/src/nlayout.rs
+++ b/crates/accelerate/src/nlayout.rs
@@ -91,7 +91,7 @@ impl VirtualQubit {
 ///         physical qubit index on the coupling graph.
 ///     logical_qubits (int): The number of logical qubits in the layout
 ///     physical_qubits (int): The number of physical qubits in the layout
-#[pyclass(module = "qiskit._accelerate.stochastic_swap")]
+#[pyclass(module = "qiskit._accelerate.nlayout")]
 #[derive(Clone, Debug)]
 pub struct NLayout {
     virt_to_phys: Vec<PhysicalQubit>,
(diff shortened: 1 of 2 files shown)


jakelishman:

crates/accelerate/src/nlayout.rs
```diff
@@ -91,7 +91,7 @@ impl VirtualQubit {
 ///         physical qubit index on the coupling graph.
 ///     logical_qubits (int): The number of logical qubits in the layout
 ///     physical_qubits (int): The number of physical qubits in the layout
-#[pyclass(module = "qiskit._accelerate.stochastic_swap")]
+#[pyclass(module = "qiskit._accelerate.nlayout")]
```
mtreinish, line 94: I wonder how many of these mistakes we've had after the multiple code refactors. It only will come up if we actually pickle an object though.

mtreinish (approved):
LGTM, this is a straightforward fix. Although I assume we're never actually pickling/deepcopying nlayout objects I think the only place we're using them are inside the `.run()` methods for the sabre passes (and maybe stochastic swap) which won't pass a process boundary or anything so I'm not sure this comes up. I'm not sure why/when we added the `__getstate__` and `__setstate__` to the pyclass (looking at the code I suspect I'm responsible) but  if we have the hook points defined we should make sure they work.

PR #12304, opened 2024-04-29
Moving the unsupported-assignment-operation lint rule

joesho112358:
### Summary

#9614 Moving up and comment `unsupported-assignment-operation`

### Details and comments

This looks like it should be fixed if/when it gets updated.


pyproject.toml
```diff
@@ -213,6 +213,7 @@ disable = [
     "docstring-first-line-empty", # relax docstring style
     "import-outside-toplevel", "import-error", # overzealous with our optionals/dynamic packages
     "nested-min-max", # this gives false equivalencies if implemented for the current lint version
+    "unsupported-assignment-operation", # one instance needed for the dragon: qiskit/providers/options.py:232
```
jakelishman: We can put a `# pylint: disable=unsupported-assignment-operation` with a comment on the offending line, if it's just one place.

We might even be able to get away with not doing that if we put a manual type hint
```python
class Options(Mapping):
    __slots__ = ("_fields", "validator")
    _fields: dict[str, Any]
    validator: dict[str, Any]
```
in the class body - with any luck, that should get pylint to infer the proper types of those fields, even though it can't reason about the `super().__setattr__` stuff we do.

joesho112358: the typing was unfortunately insufficient to convince the linter. updates for the single line were added

jakelishman: Ah, that's a shame.  The skip is fine, though - it's not uncommon to have very specific lines of code where it makes sense to disable one particular linter rule, which is why the skip syntax is there.

jakelishman (approved):
Thanks!

PR #12307, opened 2024-04-29
Add minimal support for standalone `Var` to visualisers

### LINKED ISSUE #10937: Add support for visualising classical variables
We will eventually want support for visualising the new `Store` instruction and the data-flow of classical variables described in https://github.com/Qiskit/RFCs/pull/50.  This may prove tricky in the near term because `Store` will appear as a 0q/0c operation in general, which the drawers have no way of handling yet (they just don't appear).

The text-based expression writing being added in #10869 will likely automatically work, because the variables will have fixed names.

jakelishman:
### Summary

This adds best-effort only support to the visualisers for handling stand-alone `Var` nodes.  Most of the changes are actually in `qasm3`, since the visualisers use internal details of that to handle the nodes.

This commit decouples the visualisers _slightly_ more from the inner workings of the OQ3 exporter by having them manage their own variable-naming contexts and using the encapsulated `_ExprBuilder`, rather than poking into random internals of the full circuit exporter. This is necessary to allow the OQ3 exporter to expand to support these variables itself, and also for the visualisers, since variables may now be introduced in inner scopes.

This commit does not attempt to solve many of the known problems around zero-operand "gates", of which `Store` is one, just leaving it un-drawn. Printing to OpenQASM 3 is possibly a better visualisation strategy for large dynamic circuits for the time being.

### Details and comments

Close #10937 (enough, at least, not to crash - it's not perfect by any stretch).

@enavarro51: would you be able to / like to review the changes to the visualisers that I made?


qiskit/qasm3/ast.py
```diff
@@ -123,13 +123,24 @@ class FloatType(ClassicalType, enum.Enum):
     OCT = 256
 
 
+class BoolType(ClassicalType):
+    """Type informatino for a Boolean."""
```
mtreinish:
```suggestion
    """Type information for a Boolean."""
```

qiskit/visualization/circuit/matplotlib.py
```diff
@@ -482,18 +484,41 @@ def _get_layer_widths(self, node_data, wire_map, outer_circuit, glob_data, build
...
+                            internal symbol table, and return an OQ3-like identifier."""
+                            # We don't attempt to disambiguate anything like register/var naming
+                            # collisions; we already don't really show classical variables.
+                            if isinstance(var, expr.Var):
+                                return ast.Identifier(var.name)
+                            if isinstance(var, ClassicalRegister):
+                                return ast.Identifier(var.name)
+                            # Single clbit.  This is not actually the correct way to lookup a bit on
+                            # the circuit (it doesn't handle bit bindings fully), but the text
+                            # drawer doesn't completely track inner-outer _bit_ bindings, only
```
mtreinish: This is the matplotlib drawer's code, although I realize this block is just copy-pasted from the text drawer which has the same exact block:

```suggestion
                            # the circuit (it doesn't handle bit bindings fully), but the mpl
                            # drawer doesn't completely track inner-outer _bit_ bindings, only
```

mtreinish (commented):
Besides a couple of typos this looks fine to me. I'm not a huge fan how coupled the visualizations and the oq3 exported code is though. It sort of reminds me of a light version of the oq2 parser and dagcircuit integration. But that's outside the scope of this PR.

jakelishman:
Yeah, the coupling in the expr renderers isn't ideal, but hopefully this PR decouples them rather more.  There's still a single private-class access, but that's also a component that's deliberately encapsulated away from the rest of the OQ3 exporter.

The coupling was a bit of a compromise - we don't have the resources to maintain the drawers as much as we'd like, and expression unparsing/formatting/whatever is not entirely trivial when it comes to determining when brackets are necessary in nested expressions.  The OQ3 exporter already has to do that, so in the choice between duplicating additional code into the visualisers that'd need to be maintained and having a not-entirely-ideal coupling between the two, we tried out the coupling.

mtreinish (approved):
I'm fine with this as is. Code wise this looks fine and it unblocks the other PRs on the stack for var support. The piece that I think is missing from the visualizers functionally is the input and store side. Looking at the output visualizations just showing the variables in the expressions is potentially confusing because you don't know where `a` or `b`  come from or what the values or types are.  But also at a certain point I don't think graphical visualization makes sense anymore and we'll likely just want to dump qasm or something instead to reason about the contents of a circuit in a "human" readable way. So this is not a regression in anyway and it's better than crashing or actively erroring with a var in an expression. We can follow up here with more improvements later.

PR #12308, opened 2024-04-29
Support standalone `Var`s in OQ3 exporter

### LINKED ISSUE #10929: Support manual variables in OpenQASM 3
The OpenQASM 3 exporter must be able to correctly output all the necessary variable declarations (including any potential `input` modifiers) and assignments from `Store` instructions.  It must be able to look up variables in its symbol table in order to write them out during classical expressions.

jakelishman:
### Summary

This adds the remaining support needed for the OpenQASM 3 exporter to support standalone variables.  The way the circuit model handles closures over these variables makes it much easier to support these than it was to handle free-form control-flow operations.

This PR somewhat refactors small parts of the exporter to better isolate the "top-level program" statement construction and analysis from the "build a scoped set of instructions" logic, which makes it rather easier to handle things like declaring IO variables only in the global scope, but locally declared variables in _all_ relevant scopes.

### Details and comments

Close #10929

Depends on #12307 because of the need to decouple the visualisers more from OQ3.

### DIFF
--- a/qiskit/qasm3/exporter.py
+++ b/qiskit/qasm3/exporter.py
@@ -353,8 +353,8 @@ class QASM3Builder:
         self.includeslist = includeslist
         # `_global_io_declarations` and `_global_classical_declarations` are stateful, and any
         # operation that needs a parameter can append to them during the build.  We make all
-        # classical declarations global because the IBM QSS stack (our initial consumer of OQ3
-        # strings) prefers declarations to all be global, and it's valid OQ3, so it's not vendor
+        # classical declarations global because the IBM qe-compiler stack (our initial consumer of
+        # OQ3 strings) prefers declarations to all be global, and it's valid OQ3, so it's not vendor
         # lock-in.  It's possibly slightly memory inefficient, but that's not likely to be a problem
         # in the near term.
         self._global_io_declarations = []
@@ -477,7 +477,7 @@ class QASM3Builder:
         # Similarly, QuantumCircuit qubits/registers are only new variables in the global scope.
         quantum_declarations = self.build_quantum_declarations()
         # This call has side-effects - it can populate `self._global_io_declarations` and
-        # `self._global_classical_declarations` as a courtesy to the QSS stack that prefers our
+        # `self._global_classical_declarations` as a courtesy to the qe-compiler that prefers our
         # hacky temporary `switch` target variables to be globally defined.
         main_statements = self.build_current_scope()
 
@@ -1072,8 +1072,8 @@ def _infer_variable_declaration(
 
     This is very simplistic; it assumes all parameters are real numbers that need to be input to the
     program, unless one is used as a loop variable, in which case it shouldn't be declared at all,
-    because the ``for`` loop declares it implicitly (per the Qiskit/QSS reading of the OpenQASM
-    spec at Qiskit/openqasm@8ee55ec).
+    because the ``for`` loop declares it implicitly (per the Qiskit/qe-compiler reading of the
+    OpenQASM spec at openqasm/openqasm@8ee55ec).
 
     .. note::
 
--- a/qiskit/qasm3/exporter.py
+++ b/qiskit/qasm3/exporter.py
@@ -33,6 +33,7 @@ from qiskit.circuit import (
     Qubit,
     Reset,
     Delay,
+    Store,
 )
 from qiskit.circuit.bit import Bit
 from qiskit.circuit.classical import expr, types
@@ -62,7 +63,6 @@ from .printer import BasicPrinter
 _RESERVED_KEYWORDS = frozenset(
     {
         "OPENQASM",
-        "U",
         "angle",
         "array",
         "barrier",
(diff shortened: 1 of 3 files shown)


qiskit/qasm3/exporter.py
```diff
@@ -441,15 +445,66 @@ def build_header(self):
...
+        self.hoist_classical_register_declarations()
+        # We hoist registers before new-style vars because registers are an older part of the data
+        # model (and used implicitly in PrimitivesV2 outputs) so they get the first go at reserving
+        # names in the symbol table.
+        self.hoist_classical_io_var_declarations()
+
+        # Similarly, QuantumCircuit qubits/registers are only new variables in the global scope.
+        quantum_declarations = self.build_quantum_declarations()
+        # This call has side-effects - it can populate `self._global_io_declarations` and
+        # `self._global_classical_declarations` as a courtesy to the QSS stack that prefers our
```
mtreinish: I realize it's just a code comment, but maybe say "qe-compiler" instead of "QSS stack" as it's the name of the open source project you're targeting with this right? Unless you meant something else here.

qiskit/qasm3/exporter.py
```diff
@@ -760,21 +808,34 @@ def build_aliases(self, registers: Iterable[Register]) -> List[ast.AliasStatemen
...
+
+        In addition to everything literally in the circuit's ``data`` field, this also includes
+        declarations for any local :class:`.expr.Var` nodes.
+        """
+        scope = self.current_scope()
+
+        # We forward-declare all local variables; it's much easier to ensure that we don't introduce
+        # use-before-definition errors in the OQ3 output this way, if the user has side-stepped the
+        # `QuantumCircuit` API protection to produce a circuit that uses an uninitialised variable.
+        # (It would naturally be easier to see the def/use chain if we were using `DAGCircuit`.)
```
mtreinish: This reads a bit awkwardly to me. Just to confirm the intent of forward declaring all the local variables is to protect against an error if the user manually generated a circuit which uses an uninitialized var? If so I wonder if just erroring is appropriate if that's not a valid program.

test/python/qasm3/test_export.py
```diff
@@ -948,7 +948,7 @@ def test_old_alias_classical_registers_option(self):
 
     def test_simple_for_loop(self):
         """Test that a simple for loop outputs the expected result."""
-        parameter = Parameter("x")
+        parameter = Parameter("my_x")
```
mtreinish, line 951: Is there a reason you changed the name here? I assume it's because of the conflict between the name `x` and the gate. I saw the code above handling this, but is this a behavior change we should account for in a release note?

mtreinish (commented):
I took a look at the second commit while we wait for #12307 to merge. Overall it looks fine to me, but I'm also not super familiar with the internals of the oq3 exporter so I don't have the most critical eye while looking at this code. I just had some quick questions inline.

qiskit/qasm3/exporter.py
jakelishman: The alternative to forwards-declare (`uint[8] x; x = 12;`) is to attempt to find the first use of the variable, and merge the declaration with its initialisation (`uint[8] x = 12;`).  It would be much more pleasant to do the latter, but there's like 200 subtleties in the model of `QuantumCircuit` to try and detect that without having a battle-tested def-use chain data structure / analysis that we can re-use from somewhere efficiently.  We do have that in `DAGCircuit`, but it would be too costly to pull the circuit into a DAG just to do that.

Edge-case examples: the variable might be declared in the outermost scope, but its first write is in both branches of an `if`/`else` - we still have to forward-declare.  Alternatively, there's an unintialised variable, so we have to have handling for them if we run off the end of the export and haven't put a declaration for them in.  I just kept coming up with more edge cases as I started writing this, so it's way easier just to forward-declare all the variables at the top and then not worry about it again.

test/python/qasm3/test_export.py
jakelishman: `x` is the name of a gate in `stdgates.inc` (same for `y` and `t`, also renamed), and the OQ3 exporter was falsely allowing the shadowing previously, which got fixed when I formalised the idea of variables being in the symbol table.  I should add an explicit test and a bugfix note for it for these parameters, yeah.

jakelishman: Test and release note added in 5fbeabacba.

qiskit/qasm3/exporter.py
jakelishman: It's not an invalid program to have an uninitialised variable, it's just undefined behaviour in the spec.  For the control-flow scope thing as well - it's totally possible to have a variable that it guaranteed to be initialised before it's first read from, but has no single definition source point.

I clarified the comment in 3b4188f57b.

qiskit/qasm3/exporter.py
jakelishman: Yeah, I apparently haven't got my head round the new name yet.  It's `qe-compiler` I meant.  Removed all references to QSS in e17337902e.

mtreinish (approved):

PR #12291, opened 2024-04-22
Improve performance of BackendSamplerV2 and EstimatorV2

t-imamichi:
### Summary

Fixes #12290 
Addresses https://github.com/Qiskit/qiskit-ibm-runtime/issues/1631 (need to port this PR to qiskit-ibm-runtime to close it)

TODO
- [x] BackendSamplerV2
- [x] BackendEstimatorV2
- [x] reno

### Sampler
```python
from timeit import timeit

from qiskit_aer import AerSimulator
from qiskit_ibm_runtime.fake_provider import FakeSherbrooke

from qiskit import QuantumCircuit
from qiskit.primitives import BackendSampler, BackendSamplerV2
from qiskit.transpiler.preset_passmanagers import generate_preset_pass_manager

backend = AerSimulator.from_backend(FakeSherbrooke())
shots = 10000
num_copies = 20


def gen_circuit(num_qubits: int, reps: int):
    qc = QuantumCircuit(num_qubits)
    for _ in range(reps):
        qc.h(range(num_qubits))
        for i in range(0, num_qubits - 1, 2):
            qc.cx(i, i + 1)
    qc.measure_all()
    return qc


def bench_sampler_v1(qc: QuantumCircuit):
    print("\nBackendSamplerV1")
    sampler = BackendSampler(backend)
    print(f"{timeit(lambda: sampler.run([qc] * num_copies, shots=shots).result(), number=1)} sec")


def bench_sampler_v2(qc: QuantumCircuit):
    print("\nBackendSamplerV2")
    sampler = BackendSamplerV2(backend=backend)
    print(f"{timeit(lambda: sampler.run([qc] * num_copies, shots=shots).result(), number=1)} sec")


qc = gen_circuit(5, 5)
pm = generate_preset_pass_manager(optimization_level=2, backend=backend)
qc2 = pm.run(qc)
bench_sampler_v1(qc2)
bench_sampler_v2(qc2)
```

```
# main
BackendSamplerV1
0.714542875000916 sec

BackendSamplerV2
9.41767649999747 sec
```

```
# this PR
BackendSamplerV1
0.7391576250010985 sec

BackendSamplerV2
0.7517977079987759 sec
```

-----------
### Estimator
```python
from math import sqrt
from timeit import timeit

from qiskit_aer import AerSimulator
from qiskit_ibm_runtime.fake_provider import FakeSherbrooke

from qiskit import QuantumCircuit
from qiskit.primitives import BackendEstimator, BackendEstimatorV2
[...]

PR #12309, opened 2024-04-30
Fix a type hint of an private method `BackendSamplerV2._run` (follow-up #12291)

t-imamichi:
### Summary

I noticed that `BackendSamplerV2` modified by #12291 has an incorrect type hint. This PR fixed it.
This PR does not change any behavior of the existing codes.
Because #12291 has not been released yet, I don't think it requires reno.

This change is related to the input type of `_run` method, which is guaranteed to be a list thanks to the preprocessing in `run` method.

I also noticed that there are no tests whether Primitives V2 work fine given an iterable  of pub-likes instead of a sequence of pub-like.

Follow-up #12291

### DIFF
--- a/qiskit/primitives/backend_sampler_v2.py
+++ b/qiskit/primitives/backend_sampler_v2.py
@@ -142,7 +142,7 @@ class BackendSamplerV2(BaseSamplerV2):
                     UserWarning,
                 )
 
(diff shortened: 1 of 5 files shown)


ihincks (approved):
:+1:

PR #12317, opened 2024-05-01
Update BitArray.get_counts() to support zero classical registers

### LINKED ISSUE #12062: `BitArray.get_counts()` fails given an empty classical register
### Environment

- **Qiskit version**: main
- **Python version**: 3.11
- **Operating system**: Linux

### What is happening?

`BitArray` is not fully tested with empty classical registers. For instance, `get_counts()` fails during the reshape if the output has zero bits.  It appears that `get_int_counts()` will have the same problem, too.

```python
import numpy as np

data = BitArray(np.zeros([1024, 0], dtype=np.uint8), 0)
data.get_counts()
```

results in

```
  File "/home/garrison/Qiskit/empty-creg-counts.py", line 7, in <module>
    data.get_counts()
  File "/home/garrison/serverless/.direnv/python-3.11.0/lib/python3.11/site-packages/qiskit/primitives/containers/bit_array.py", line 292, in get_counts
    return self._get_counts(loc=loc, converter=converter)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/garrison/serverless/.direnv/python-3.11.0/lib/python3.11/site-packages/qiskit/primitives/containers/bit_array.py", line 146, in _get_counts
    arr = self._array.reshape(-1, self._array.shape[-1]) if loc is None else self._array[loc]
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
ValueError: cannot reshape array of size 0 into shape (0)
```

### What should happen?

I expect a sensible result, such as `{"": 1024}` for `get_counts()` or `{0: 1024}` for `get_int_counts()`.

### Any suggestions?

_No response_

atharva-satpute:
### Summary
Fixes #12062
Added support for zero classical registers in `BitArray.get_counts()`,  `BitArray.get_int_counts()` and  `BitArray.get_bitstring()`

### Details and comments
1. With this change, if `BitArray` is initialised with zero classical registers it will return `{'': <count>}` for `get_counts()` and `{0: <count>}` for `get_int_counts()`
2. Noticed a case where `loc` is not `None` and `arr` is a **number** (`arr = self._array[loc]`), this would throw a `TypeError` on [line](https://github.com/Qiskit/qiskit/blob/f2b874b83c30df6bd08aa2c13399f9f8020b7a9a/qiskit/primitives/containers/bit_array.py#L149), to handle this converted `arr` to `np.array` and along with this, checking `arr.size == 0` in case `get_bitstring()`


ihincks:
Thanks for taking this on @atharva-satpute . I think we will also need to make sure that from_bitstrings, from_int also support the num_bits=0 case, because I suspect that they don't right now.

atharva-satpute:
Sure

ewinston:
@ihincks It seems BitString doesn't support from_bitstrings or from_int. The only reference I see to from_bitstrings is in the documentation of the DataBin class.

However, @atharva-satpute , I think it should probably support `from_mapping({'', 1024})`

atharva-satpute:
What would `from_mapping()` method do?

t-imamichi:
The discussion has suspended for months. Could you restart it, @ewinston, @ihincks? It would be nice to fix the issue with zero classical register.

PR #12320, opened 2024-05-01
Bump pylint version used to latest release

mtreinish:
### Summary

The pinned version of pylint we're running in CI right now doesn't support Python 3.12. This is an issue if you're developing with Python 3.12 because you couldn't run pylint. This commit bumps to the latest version as of this commit and fixes the additional checks flagged by the newer version.

### Details and comments

### DIFF
--- a/qiskit/passmanager/flow_controllers.py
+++ b/qiskit/passmanager/flow_controllers.py
@@ -45,8 +45,7 @@ class FlowControllerLinear(BaseController):
         return list(self.tasks)
 
     def iter_tasks(self, state: PassManagerState) -> Generator[Task, PassManagerState, None]:
-        for task in self.tasks:
-            state = yield task
+        yield from self.tasks
 
 
 class DoWhileController(BaseController):
@@ -78,8 +77,7 @@ class DoWhileController(BaseController):
     def iter_tasks(self, state: PassManagerState) -> Generator[Task, PassManagerState, None]:
         max_iteration = self._options.get("max_iteration", 1000)
         for _ in range(max_iteration):
-            for task in self.tasks:
-                state = yield task
+            yield from self.tasks
             if not self.do_while(state.property_set):
                 return
             # Remove stored tasks from the completed task collection for next loop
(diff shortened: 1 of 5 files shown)


qiskit/passmanager/flow_controllers.py
```diff
@@ -78,8 +77,7 @@ def passes(self) -> list[Task]:
     def iter_tasks(self, state: PassManagerState) -> Generator[Task, PassManagerState, None]:
         max_iteration = self._options.get("max_iteration", 1000)
         for _ in range(max_iteration):
-            for task in self.tasks:
-                state = yield task
+            yield from self.tasks
             if not self.do_while(state.property_set):
                 return
```
jakelishman, line 82: This is a logic change because the updated `state` isn't received back from `generator.send` any more.  It might happen to work if the state is only mutated in place, but in general it's not a valid transformation.

qiskit/transpiler/passmanager.py
```diff
@@ -31,6 +31,7 @@
 from .exceptions import TranspilerError
 from .layout import TranspileLayout
 
+# pylint: disable=invalid-name
 _CircuitsT = Union[List[QuantumCircuit], QuantumCircuit]
```
jakelishman, line 35: I don't actually think this is a false positive, I think it's indicative of the typing construct being wrong.  This defines an alias, but the way it's used is actually intended as a typevar generic like:
```python
from typing import TypeVar, Union, List

_CircuitsT = TypeVar("_CircuitsT", bound=Union[List[QuantumCircuit], QuantumCircuit])
```

mtreinish: Well it was complaining about the name, not the type. You had to infer the name didn't match the rules for type aliases and figure out it's because it wasn't using the right thing here. But yeah I'll update to use this instead.

mtreinish:
Marking as on hold until after 1.1.0rc1 so we don't disrupt pending PRs

qiskit/transpiler/passmanager.py
jakelishman: judging by my local branch that did this and half-remembered confusion, I think I went down the same rabbit hole as you a couple of months ago yeah

PR #12322, opened 2024-05-01
Support standalone `Var` throughout transpiler

### LINKED ISSUE #10931: Add testing for manual variables to `transpile`
Once all the `DAGCircuit` methods are updated to support manual variable storage and loads, `transpile` should automatically work, since everything largely just calls `DAGCircuit` methods and the logic should thread its way through.

However, we still need to add integration testing across `transpile` calls and post-compilation serialisation (+ deserialisation in the case of QPY) roundtripping.

jakelishman:
### Summary

This adds the missing pieces to fully support standalone `Var` nodes through every part of the transpiler (that I could detect).  It's quite possible there's some problem in a more esoteric non-preset pass somewhere, but I couldn't spot them.

For the most part there were very few changes needed to the actual passes, and only one place in `QuantumCircuit` that had previously been missed.  Most of the commit is updating passes to correctly pass `inline_captures=True` when appropriate for dealing with `DAGCircuit.compose`, and making sure that any place that needed to build a raw `DAGCircuit` for a rebuild _without_ using `DAGCircuit.copy_empty_like` made sure to correctly add in the variables.

This commit adds specific tests for every pass that I touched, plus the general integration tests that we have for the transpiler to make sure that OQ3 and QPY serialisation work afterwards.

### Details and comments

Close #10931.

This is the top of the stack and depends on #12215 and #12308 (the latter of which depends on #12307, though those could be merged together).

I'll write the full release note as a full PR after the tagging of rc1.

### DIFF
--- a/qiskit/circuit/store.py
+++ b/qiskit/circuit/store.py
@@ -59,6 +59,9 @@ class Store(Instruction):
     :class:`~.circuit.Measure` is a primitive for quantum measurement), and is not safe for
     subclassing."""
 
+    # This is a compiler/backend intrinsic operation, separate to any quantum processing.
+    _directive = True
+
     def __init__(self, lvalue: expr.Expr, rvalue: expr.Expr):
         """
         Args:
--- a/test/python/circuit/test_circuit_operations.py
+++ b/test/python/circuit/test_circuit_operations.py
@@ -1013,9 +1013,9 @@ class TestCircuitOperations(QiskitTestCase):
         qc.h(0)
         qc.cx(0, 1)
         with qc.if_test(a):
-            # We don't really comment on what should happen to control-flow operations in this
-            # method, and Sabre doesn't care (nor use this method, in the fast paths), so this is
-            # deliberately using a body of length 1 (a single `Store`).
+            # We don't really comment on what should happen within control-flow operations in this
+            # method - it's not really defined in a non-linear CFG.  This deliberately uses a body
+            # of length 1 (a single `Store`), so there's only one possibility.
             qc.add_var(c, 12)
 
         expected = qc.copy_empty_like()
--- a/test/python/circuit/test_circuit_operations.py
+++ b/test/python/circuit/test_circuit_operations.py
@@ -1002,6 +1002,31 @@ class TestCircuitOperations(QiskitTestCase):
 
         self.assertEqual(qc.reverse_ops(), expected)
 
+    def test_reverse_with_standlone_vars(self):
+        """Test that instruction-reversing works in the presence of stand-alone variables."""
+        a = expr.Var.new("a", types.Bool())
+        b = expr.Var.new("b", types.Uint(8))
+        c = expr.Var.new("c", types.Uint(8))
+
+        qc = QuantumCircuit(2, inputs=[a])
+        qc.add_var(b, 12)
+        qc.h(0)
+        qc.cx(0, 1)
+        with qc.if_test(a):
+            # We don't really comment on what should happen to control-flow operations in this
+            # method, and Sabre doesn't care (nor use this method, in the fast paths), so this is
+            # deliberately using a body of length 1 (a single `Store`).
+            qc.add_var(c, 12)
+
+        expected = qc.copy_empty_like()
+        with expected.if_test(a):
+            expected.add_var(c, 12)
+        expected.cx(0, 1)
+        expected.h(0)
+        expected.store(b, 12)
+
+        self.assertEqual(qc.reverse_ops(), expected)
+
     def test_repeat(self):
         """Test repeating the circuit works."""
         qr = QuantumRegister(2)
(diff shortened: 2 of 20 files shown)


jakelishman:
Now rebased over #12308 (speculatively assuming the merge-queue commit passes CI).

qiskit/circuit/store.py
```diff
@@ -59,6 +59,9 @@ class Store(Instruction):
     :class:`~.circuit.Measure` is a primitive for quantum measurement), and is not safe for
     subclassing."""
 
+    # This is a compiler/backend intrinsic operation, separate to any quantum processing.
+    _directive = True
```
mtreinish, line 63: I'm fine with this, but this basically means we're treating this as always supported through the transpiler regardless of what the target says. Is that the correct behavior for this? I'm mostly wondering for backends that don't have any classical logic support, and how we intend to indicate this to backends. For other control flow operations they have to be explicitly in the target to work. Is the answer just documentation?

test/python/circuit/test_circuit_operations.py
```diff
@@ -1002,6 +1002,31 @@ def test_reverse(self):
...
+        c = expr.Var.new("c", types.Uint(8))
+
+        qc = QuantumCircuit(2, inputs=[a])
+        qc.add_var(b, 12)
+        qc.h(0)
+        qc.cx(0, 1)
+        with qc.if_test(a):
+            # We don't really comment on what should happen to control-flow operations in this
+            # method, and Sabre doesn't care (nor use this method, in the fast paths), so this is
+            # deliberately using a body of length 1 (a single `Store`).
```
mtreinish: I'm a bit confused where is the sabre discussion coming from in this test? This is validating that `reverse_ops()` works correctly with variables.

test/python/compiler/test_transpiler.py
```diff
@@ -2350,6 +2422,21 @@ def test_qasm3_output_control_flow_expr(self, optimization_level):
...
+        can be dumped into OpenQASM 3."""
+        transpiled = transpile(
+            self._standalone_var_circuit(),
+            backend=GenericBackendV2(num_qubits=13, control_flow=True),
+            optimization_level=optimization_level,
+            seed_transpiler=2024_05_01,
+        )
+        # TODO: There's not a huge amount we can sensibly test for the output here until we can
+        # round-trip the OpenQASM 3 back into a Terra circuit.  Mostly we're concerned that the dump
+        # itself doesn't throw an error, though.
```
mtreinish, line 2437: You could assert the stores are preserved and present in the qasm string independent of the transpilers optimization or something I guess. But yeah it probably just makes more sense to wait for the parser to gain support for this.

mtreinish (commented):
This LGTM, this was much easier to integrate than I expected. Before we merge though I had one question about treating store as a directive for backends that don't have any dynamic circuits capabilities and won't support Vars at all.

test/python/compiler/test_transpiler.py
jakelishman: This comment is basically just copied from the tests above from when we added the enhanced conditions to control flow.

qiskit/circuit/store.py
jakelishman: Short of adding more magic fields to `supported_instruction` `Target`, which could never really capture all the constraints of dynamic-circuits requirements already, or some sort of global flag on `Target`/`BackendV2` that we haven't really prepared to put in place, I think documentation is the answer.

Dynamic circuits have always been tricky through these tests, because there's _so much_ that's never really been possible to represent as a restriction to the transpiler (nesting can only be to `n` levels, max of `m` blocks, switches must have a `default`, etc etc).

test/python/circuit/test_circuit_operations.py
jakelishman: `QuantumCircuit.reverse_ops` is only (meaningfully) used by the slow-path `SabreLayout` using an explicit routing pass, but thinking more, the idea of "Sabre doesn't care what happens within control-flow ops" only applies the same way to the fast path, so the Sabre bit is irrelevant.  Changed in 4dedd6cd38.

jakelishman:
I'll try closing and re-opening this to see if that triggers GH to recalculate the merge base.

qiskit/circuit/store.py
mtreinish: Let just go with documenting this for now as part of the 1.1.x final documentation, I'll convert this into an issue to track. Both as an upgrade note in the final release notes and also as a guidance in the writing a backend section of the docs.

I think for 1.2.x we can look at adding a global flag in the target for something like supports store, etc to provide a high level indicator for this kind of capability above the per instruction look ups. We'd probably need to do this into an opt-in way (so it defaults true) but I think having the capability to express this in the target is what we'll want longer term.

mtreinish (approved):
This LGTM, I'm still mildly concerned about how we handling `Store` for the target/backends, but I'm fine documenting it. Realistically I feel like this will be an uncommon use case in practice and the only people using it to start will know the backend is capable of supporting it. We can look at evolving our target model to better account for this in 1.2.0

Issue #12225, opened 2024-04-19
Port `synth_permutation_basic` to Rust


mtreinish:

jpacold:
Hello -- I've been looking for open issues to work on and am potentially interested in this one.

It looks as if [synth_permutation_basic](https://github.com/Qiskit/qiskit/blob/d10f9a0fe1f73cde5b66e49af880a6f7816f131f/qiskit/synthesis/permutation/permutation_full.py#L27) just decomposes a permutation into transpositions by calling [_get_ordered_swap](https://github.com/Qiskit/qiskit/blob/d10f9a0fe1f73cde5b66e49af880a6f7816f131f/qiskit/synthesis/permutation/permutation_utils.py#L16), and then builds a circuit from the list of swaps.

Adding a Rust implementation of `_get_ordered_swap` to the accelerate crate is pretty straightforward. Looking at the top-level issue #12212, I'm not sure there is much more that can be done with this issue for now since the actual circuit has to be built in Python.

Is that an accurate assessment, and if so is it okay to work on this?

alekkemeny:
@jpacold Sounds good, you can go ahead!

ShellyGarion:
can we close this issue since #12327 is closed ?

jpacold:
It looks like #12605 closes this.

ShellyGarion:
closed by #12605

PR #12327, opened 2024-05-02
Move utility functions _inverse_pattern and _get_ordered_swap to Rust

jpacold:
### Summary
This PR starts addressing #12225. I moved the implementations of two functions from [qiskit/synthesis/permutation/permutation_utils.py](https://github.com/Qiskit/qiskit/blob/d10f9a0fe1f73cde5b66e49af880a6f7816f131f/qiskit/synthesis/permutation/permutation_utils.py) to Rust.

### Details and comments
* Created [permutation.rs](https://github.com/Qiskit/qiskit/compare/main...jpacold:qiskit:main#diff-791e0107a3002abbb1b7fad2dddd59e6f0ce84dbea6c8e67c9a2f77e15cd403d)
    * I added some checks to ensure that inputs are actually permutations. Otherwise, `_get_ordered_swap` uses the same logic as it did in Python.
    * Performance is better than Python, but sensitive to input type (more notes below)

* Added a module `qiskit._accelerate.permutation` and a submodule `qiskit._accelerate.permutation.permutation_utils` (following the structure of the existing Python modules). The wrapped functions are in the `permutation_utils` submodule.

* Replaced the function bodies in `permutation_utils.py` with calls to the Rust functions.

* Added tests.

### Performance

As expected, this implementation of `_get_ordered_swap` is faster than the Python version. However, the input type is expected to be an array of `int64` and there is a performance cost for converting from any other type. On my machine:
| Input type  | Speedup over Python  |
| --- | --- |
| Numpy array (`int64`)  | ~15x  |
| Numpy array (`int32`, `uint32`, or `uint64`) | ~4x |
| Python list | ~7x |

For now, I don't think this matters much, because the bottleneck in `synth_permutation_basic` is constructing the circuit, but it is something that might need to be optimized later on.


raynelfss:
Is this still a work in progress? @jpacold if it's not, it would be helpful to remove the [WIP] from the title as it seems to me this has been completed.

jpacold:
@raynelfss Done, thanks for pointing that out.

crates/accelerate/src/permutation.rs
```diff
@@ -0,0 +1,122 @@
...
+        (permutation[ii], permutation[jj]) = (permutation[jj], permutation[ii]);
+        index_map[val] = jj;
+        index_map[ii] = ii;
+    }
+
+    swaps[..].reverse();
+    swaps
+}
+
+///Finds inverse of a permutation pattern.
```
raynelfss:
```suggestion
/// Finds inverse of a permutation pattern.
```

crates/accelerate/src/permutation.rs
```diff
@@ -0,0 +1,122 @@
...
+
+///Sorts the input permutation by iterating through the permutation list
+///and putting each element to its correct position via a SWAP (if it's not
+///at the correct position already). If ``n`` is the length of the input
+///permutation, this requires at most ``n`` SWAPs.
+///
+///More precisely, if the input permutation is a cycle of length ``m``,
+///then this creates a quantum circuit with ``m-1`` SWAPs (and of depth ``m-1``);
+///if the input  permutation consists of several disjoint cycles, then each cycle
+///is essentially treated independently.
```
raynelfss:
```suggestion
/// Sorts the input permutation by iterating through the permutation list
/// and putting each element to its correct position via a SWAP (if it's not
/// at the correct position already). If ``n`` is the length of the input
/// permutation, this requires at most ``n`` SWAPs.
///
/// More precisely, if the input permutation is a cycle of length ``m``,
/// then this creates a quantum circuit with ``m-1`` SWAPs (and of depth ``m-1``);
/// if the input  permutation consists of several disjoint cycles, then each cycle
/// is essentially treated independently.

**Review by raynelfss — DISMISSED** [review]
LGTM, just a small correction on the docstrings for your rust functions. Thank you for working on this!

File: crates/accelerate/src/permutation.rs [hunk 3]
```diff
@@ -0,0 +1,122 @@
...
+    py: Python,
+    permutation_in: PyArrayLike1<i64, AllowTypeChange>,
+) -> PyResult<PyObject> {
+    let view = permutation_in.as_array();
+    validate_permutation(&view)?;
+    Ok(get_ordered_swap(&view).to_object(py))
+}
+
+#[pymodule]
+pub fn permutation_utils(m: &Bound<PyModule>) -> PyResult<()> {
```
ShellyGarion: why do you need two `[pymodule]`s here? I see that in other cases only one was used.

jpacold: It's not really necessary. I copied the structure from the Python `synthesis` module:

```
qiskit
    |-- synthesis
        |-- permutation
            | -- permutation_full
            | -- permutation_lnn
            | -- permutation_reverse_lnn
            | -- permutation_utils
```

since I was thinking the idea is to eventually move all of it to Rust.

ShellyGarion: you are right that the idea is to eventually move all of the synthesis code to rust,
however, it seems that currently the structure is flat, see e.g.
https://github.com/Qiskit/qiskit/blob/6e3bc2c292d697eb180f75e9943f54839c553038/qiskit/__init__.py#L83
https://github.com/Qiskit/qiskit/blob/6e3bc2c292d697eb180f75e9943f54839c553038/qiskit/__init__.py#L70

jpacold: Got it, I made another commit to flatten this as well.

crates/accelerate/src/permutation.rs
```diff
@@ -0,0 +1,115 @@
...
+    swaps[..].reverse();
+    swaps
+}
+
+/// Finds inverse of a permutation pattern.
+#[pyfunction]
+#[pyo3(signature = (pattern))]
+fn _inverse_pattern(py: Python, pattern: PyArrayLike1<i64, AllowTypeChange>) -> PyResult<PyObject> {
+    let view = pattern.as_array();
+    validate_permutation(&view)?;
```
Cryoris, line 85: I'm not sure this is the best place to validate the permutation (same for the ordered swap), as (1) it's a private utility function without detailed docs and (2) explicit verification is probably done at a different level, if required. It's not super important but it would be slightly faster and the previous implementation also didn't have it 🙂

Cryoris (commented):
Just on comment otherwise LGTM as well 👍🏻

crates/accelerate/src/permutation.rs
```diff
@@ -0,0 +1,115 @@
...
+///
+/// More precisely, if the input permutation is a cycle of length ``m``,
+/// then this creates a quantum circuit with ``m-1`` SWAPs (and of depth ``m-1``);
+/// if the input  permutation consists of several disjoint cycles, then each cycle
+/// is essentially treated independently.
+#[pyfunction]
+#[pyo3(signature = (permutation_in))]
+fn _get_ordered_swap(
+    py: Python,
+    permutation_in: PyArrayLike1<i64, AllowTypeChange>,
```
ShellyGarion: why do we need  `AllowTypeChange` ?

crates/accelerate/src/permutation.rs
jpacold: I could go either way on this. It has some overhead (about 10% of the wall-clock time for `_get_ordered_swap` on my machine). On the other hand, if the input is invalid I would rather raise a Python exception than have something go wrong in the Rust layer.

crates/accelerate/src/permutation.rs
jpacold: From reading the numpy crate docs I assumed that I needed it to handle other integer types. But from a quick test it looks as if an implicit conversion happens without it (maybe in pyo3?). I will remove it in a minute.

crates/accelerate/src/permutation.rs
Cryoris: Being a non-user-facing, private method I'd say performance is more important and the developer should make sure the input makes sense. We could expose your validation function e.g. as `_validate_permutation` to Python and use it in the appropriate places.

jpacold: Sounds fine, I moved it into a separate pyfunction.

Cryoris (commented):
LGTM thanks for the updates! If @ShellyGarion and @raynelfss are good with it I think we can merge 👍🏻 

Edit: both are good with the changes 🙂

Cryoris (approved):

PR #12328, opened 2024-05-02
Set version number to 1.1.0rc1 for first release candidate

mtreinish:
### Summary

For the 1.1.0 release we're going to push release candidates prior to the final to enable testing before we cut the final release. In preparation for tagging the first release candidate this commit updates the version string to indicate it's a release candidate. This commit should be the final commit on main for 1.1.0rc1 and what gets tagged as 1.1.0rc1 post-merge.


sbrandhsn (approved):
LGTM, thanks!

PR #12329, opened 2024-05-02
[WIP] Add HLS plugin benchmarks

ElePT:
### Summary
This is a first attempt at adding a file to test potential performance improvements in HLS plugins.

### DIFF
--- a/test/benchmarks/transpiler_hls_plugins.py
+++ b/test/benchmarks/transpiler_hls_plugins.py
@@ -51,6 +51,7 @@ class VGate(Gate):
         qc.h(0)
         self.definition = qc
 
+
 class WGate(Gate):
     """W Gate used in Clifford synthesis."""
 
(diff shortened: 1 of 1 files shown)


test/benchmarks/transpiler_hls_plugins.py
```diff
@@ -0,0 +1,227 @@
...
+        self.num_samples = 20
+
+        # separate transpile from build
+        # track: depth, num. 2q gates, 2q depth
+
+    # Clifford Synthesis Plugins
+    def time_clifford_ag_large_circuit(self):
+        """Test A&G synthesis for set of {num_qubits}-qubit Cliffords"""
+        hls_config = HLSConfig(clifford=["ag"])
+        pm = PassManager([HighLevelSynthesis(hls_config=hls_config)])
```
mtreinish: I would probably move this to `setUp()` to avoid this skewing the measurements. The entire contents of this function is what ends up being measured.

test/benchmarks/transpiler_hls_plugins.py
```diff
@@ -0,0 +1,227 @@
...
+
+        for _ in range(self.num_samples):
```
mtreinish: We don't need to do this explicitly the asv harness handles taking multiple samples to get the statistical sampling. We probably want to move the random clifford construction to setup too though to avoid randomization impacting the measurements.

ElePT:
Thanks for the suggestions. I have moved the circuit and pm construction to the setup and fixed the random seeds, and applied a few offline suggestions from @alexanderivrii. So far the benchmarks run but I wonder if the circuit I am currently building makes sense for the `bm_wide_circuit` test, because the run seems to be too fast (range of `ms` instead of `s`), so it's probably not doing any synthesis? The difference between ag/bm and default/greedy is also quite stark in the `wide_circuit` tests (maybe the first two are also not synthesizing). I will add tests that monitor depth/gate count to try to get more information:

```
[55.00%] ··· transpiler_hls_plugins.HLSPluginsSuite.time_clifford_ag_rand_benchmarking           3.34±0.08s                                                                                                       
[60.00%] ··· transpiler_hls_plugins.HLSPluginsSuite.time_clifford_ag_wide_circuit                1.24±0.08s                                                                                             
[65.00%] ··· transpiler_hls_plugins.HLSPluginsSuite.time_clifford_bm_rand_benchmarking           9.98±0.09s                                                                                                       
[70.00%] ··· transpiler_hls_plugins.HLSPluginsSuite.time_clifford_bm_wide_circuit                2.02±0.07ms                                                                                                   
[75.00%] ··· transpiler_hls_plugins.HLSPluginsSuite.time_clifford_default_rand_benchmarking      9.47±0.2s                                                                                         
[80.00%] ··· transpiler_hls_plugins.HLSPluginsSuite.time_clifford_default_wide_circuit           26.1±0.9s                                                                                                     
[85.00%] ··· transpiler_hls_plugins.HLSPluginsSuite.time_clifford_greedy_rand_benchmarking       9.22±0.2s                                                                                                   
[90.00%] ··· transpiler_hls_plugins.HLSPluginsSuite.time_clifford_greedy_wide_circuit            25.8±0.5s                                                                                               
[95.00%] ··· transpiler_hls_plugins.HLSPluginsSuite.time_linear_func_kms                         1.86±0.1s                                                                                                   
[100.00%] ··· transpiler_hls_plugins.HLSPluginsSuite.time_linear_func_pmh                        1.14±0.06s                                                                                                           
```

test/benchmarks/transpiler_hls_plugins.py
alexanderivrii: Should we nonetheless consider _multiple_ different random linear functions, random cliffords, etc. to avoid overtuning synthesis algorithms on specific circuits?

PR #12331, opened 2024-05-02
Link to `qiskit.synthesis.unitary.aqc` on API index page

### LINKED ISSUE #1272: Fix unnecessary h1 header in module pages generated by autosummary
## Problem

Examples:

- https://docs.quantum.ibm.com/api/qiskit-ibm-runtime/qiskit_ibm_runtime.transpiler.passes.scheduling
- https://docs.quantum.ibm.com/api/qiskit/qiskit.synthesis.unitary.aqc

Note how the h1 is the module path, and then there is an h2 we wrote. That's the only h2: all other headers are h3.

Screenshot 2024-05-01 at 1 10 49 PM
Screenshot 2024-05-01 at 1 11 21 PM

The h2 comes from our docstring:

https://github.com/Qiskit/qiskit/blob/a41690d6075ac9438ff01b2346a35aa21dc3a568/qiskit/synthesis/unitary/aqc/__init__.py#L13-L23

Whereas the h1 comes from the autosummary template:

https://github.com/sphinx-doc/sphinx/blob/c41aab829d667a6fb5c9e6e8daf2d4b47b7ddebe/sphinx/ext/autosummary/templates/autosummary/module.rst?plain=1#L1-L3

We don't want that h1 to come from the autosummary template. Because of the way RST handles headings, setting the header in the template results in all subsequent headers being one heading level lower, like h1->h2. Instead, we should be treating the header in the docstring as the h1, which is how it works when we have manually defined module pages like 

https://github.com/Qiskit/qiskit/blob/a41690d6075ac9438ff01b2346a35aa21dc3a568/docs/apidoc/circuit.rst?plain=1#L1-L6

This issue makes https://github.com/Qiskit/documentation/issues/942 worse.

## attempted solutions so far

* Remove the h1 from the module docstring so that we only have its h2s (multiple of them).
* Override the module template to not add the h1.

Sphinx crashes with both of these.

Eric-Arellano:
Before, `qiskit.synthesis.unitary.aqc` was autogenerated by `autosummary` rather than being manually created like we normally do in `apidoc/`. This had two issues:

1. The module does not show up in the API index page https://docs.quantum.ibm.com/api/qiskit, even though it shows up as top-level in the doc app's left table of contents
2. The header hierarchy is messed up due to an autosummary quirk that I could not figure out how to fix https://github.com/Qiskit/documentation/issues/1272

Using a dedicated page fixes both of these problems.

I named the file `qiskit.synthesis.unitary.aqc` to preserve the previous URL we had for the module, so that we don't need redirects. (Technically, Sphinx now puts the HTML file in the `apidoc/` folder rather than `stubs/`, but the docs app doesn't care.)

All the other modules were fine.


ShellyGarion (approved):
LGTM, thanks

mtreinish:

PR #12335, opened 2024-05-03
Fix composition of multiple virtual permutation layouts

mtreinish:
### Summary

This commit fixes the composition of multiple virtual permutation layouts in StarPreRouting and ElidePermutations. Both of these passes set a virtual permutation layout and if both running a pipeline the layouts need to be composed together to track the permutation. Previously the logic for doing this was incorrect (and also had a syntax error). This commit fixes this so that you can run the two passes together (along with any other future passes that do a similar thing). This was spun out from #12196 as a standalone bugfix.

### Details and comments


ElePT (approved):
LGTM, I think it's a good idea to merge the fix separately from #12196

PR #12339, opened 2024-05-04
Fix Rustfmt and clippy with latest rust stable

mtreinish:
### Summary

Running `cargo fmt` with a newer version of rust than what we use in CI (1.77.2 and 1.78.0 locally for me) is triggering the formatting updates in this commit. To ensure developers don't have to worry about commiting an accidental formatting change in their commits this commit proactively makes the change. Similarly the recent Rust 1.78 release included new clippy rules which are flagging some small issues that our MSRV of clippy doesn't have. This commit also fixes these as the suggestions are good and are compatible with our MSRV of 1.70.

### Details and comments


Eric-Arellano (dismissed):

crates/qasm3/src/circuit.rs
```diff
@@ -16,6 +16,7 @@ use pyo3::types::{PyList, PyString, PyTuple, PyType};
 use crate::error::QASM3ImporterError;
 
 pub trait PyRegister {
+    #[allow(dead_code)]
```
jakelishman: We could just remove this whole part of the trait method and its impl, really - if it's not used, it's not used.  I guess I probably used it at some point while writing the exporter, but never noticed that I'd cut out the last uses since the lint wasn't firing.

mtreinish: Sure, I removed this method in : https://github.com/Qiskit/qiskit/pull/12339/commits/964b77fe0bf195188b77ef47fe34c6c02b1ced91

raynelfss (approved):
LGTM, both `cargo fmt` and `cargo clippy` are working as intended locally too. Thank you for the quick updates!

PR #12343, opened 2024-05-05
Hyper-efficient unitary diamond distance

### LINKED ISSUE #12341: Hyper efficient and more accurate diamond norm implementation
## Preliminaries
The diamond norm [1] is a commonly used metric in quantum information theory for calculating the distance between two quantum channels. It has a number of useful properties making it the gold standard [2] in the field for several applications. Qiskit currently implements the diamond norm using the semi-definite program formulation in [3]. Although elegant, this approach to calculating the diamond norm is very inefficient. Unfortunately, no alternative method has been discovered for calculating the diamond norm in the general case of CPTP channels.

However, in the special case where we are trying to calculate the difference between two unitary channels, a very efficient implementation exists. This makes use of an unproved theorem on page 29 of [1]. I have proved this theorem and elaborated an efficient algorithm to calculate the diamond distance between two unitaries as part of my masters thesis. 

The implementation is very simple - the hardest step involves diagonalising a unitary. Although time complexity is still exponential in the number of qubits, this implementation is far more efficient than Qiskit's current more general implementation. First, I do not use the Choi representation of the quantum channel, whose size increases faster than the unitary matrix representation. Second, there is no need to solve a complicated semi-definite program (meaning I can do away with the `cvxpy` dependency).

This method has the additional advantage of being more accurate than current implementation. I have tested the implementation for several cases where an analytical formula of the diamond norm exists. On average I have found the qiskit implementation has an average absolute error of 5.9168E-07 compared to 2.7645E-10.

## Empirical testing
Results of empirical testing on my machine are reported below

|                                | 1 qubit | 2 qubit | 3 qubit | 4 qubit  |
|--------------------------------|---------|---------|---------|----------|
| qiskit implementation          | 8.14 ms | 54.5 ms | 5.22 s  | 3min 21s |
| hyper-efficient implementation | 534 µs  | 696 µs  | 924 µs  | 1.11 ms  |

## Proposition
Given the popularity of the circuit model and unitary-based quantum computation, I believe this implementation of the diamond distance for unitaries would be incredibly valuable for the research community. Given the minimal amount of effort required for such an implementation (40 lines of code), I think it would be a simple and worthwhile addition to qiskit.

I would love feedback and comments on how best to approach this implementation in qiskit (as I am unfamiliar with how the library is structured).

## Citations

[1] D. Aharonov, A. Kitaev, and N. Nisan, “Quantum circuits with mixed states,” in Proceedings of the thirtieth annual ACM symposium on Theory of computing, pp. 20–30, 1998.
[2] A. Gilchrist, N. K. Langford, and M. A. Nielsen, “Distance measures to compare real and ideal quantum processes,” Physical Review A, vol. 71, no. 6, p. 062310, 2005
[3] J. Watrous, “Simpler semidefinite programs for completely bounded norms,” arXiv preprint arXiv:1207.5726, 2012.

owenagnel:
### Summary
Adds a new method `unitary_diamond_distance` to `qiskit.quantum_info.operators.measures`. This method implements a trick discussed in [1] for calculating the diamond norm (completely bounded trace norm) of a difference of unitary channels (also known as the diamond distance). The implementation is composed of three steps:
- Find the eigenvalues of $U^† V$ where `channel1` = $U \cdot U^†$ and `channel2` = $V \cdot V^†$
- Find the distance $d$ between origin and convex hull of eigenvalues
- Plug this distance into $2\sqrt{1-d^2}$

Refer to issue <https://github.com/Qiskit/qiskit/issues/12341> for more details.

### Citations
[1] D. Aharonov, A. Kitaev, and N. Nisan, “Quantum circuits with mixed states,” in Proceedings of the thirtieth annual ACM symposium on Theory of computing, pp. 20–30, 1998.

fixes #12341


kevinsung:
Is this ready for review? If not, please convert it to a draft PR.

owenagnel:
Hi Kevin, I'm pretty happy with it as is. Went through the CONTRIBUTING.md document and hopefully checked off everything. Would be great if someone with more experience contributing could go through it quickly (only a couple changes). All tests and lifting are passing on my machine using tox.

test/python/quantum_info/operators/test_measures.py
```diff
@@ -184,6 +186,17 @@ def test_diamond_norm(self, num_qubits):
         except cvxpy.SolverError:
             self.skipTest("CVXPY solver failed.")
 
+    def test_unitary_diamond_distance(self):
+        """Test the unitary_diamond_distance function"""
+        # Test unitary diamond distance for random unitaries
```
kevinsung: This comment does not describe what the test is actually doing (testing RZGates for a specific set of angles). However, I agree that testing random unitaries is a good idea. Please add a test that does that, comparing the answer with the one obtained by the existing `diamond_norm` function.

owenagnel: That sounds good. I will make the change. As a side note, the probability that the distance between two Harr random unitaries equals 2 increases exponentially with the number of qubits. Practically, testing random 3-qubit unitaries almost always returns a distance of 2. Additionally, it takes a while to run `diamond_norm` for 3 qubit channels (about 4 seconds). 

Would running the test for several random 1 and 2-qubit unitaries make for a better test?

test/python/quantum_info/operators/test_measures.py
```diff
@@ -184,6 +187,37 @@ def test_diamond_norm(self, num_qubits):
...
+        except ImportError:
+            # Skip test if CVXPY not installed
+            self.skipTest("CVXPY not installed.")
+        op1 = random_unitary(2**num_qubits)
+        op2 = random_unitary(2**num_qubits)
+        choi1 = Choi(op1)
+        choi2 = Choi(op2)
+        delta_choi = choi1 - choi2
+        try:
+            target = diamond_norm(delta_choi)
```
kevinsung: No need to define so many variables.
```python
diamond_norm(Choi(op1) - Choi(op2))
```
is readable enough.

test/python/quantum_info/operators/test_measures.py
```diff
@@ -184,6 +187,37 @@ def test_diamond_norm(self, num_qubits):
...
+            self.assertAlmostEqual(unitary_diamond_distance(op1, op2), target, places=4)
+        except cvxpy.SolverError:
+            self.skipTest("CVXPY solver failed.")
```
kevinsung: Remove the `try ... except`. Instead, use a seeded random number generator (e.g. `np.random.default_rng(1234)`) and choose the seed so that cvxpy succeeds every time.

test/python/quantum_info/operators/test_measures.py
```diff
@@ -184,6 +187,37 @@ def test_diamond_norm(self, num_qubits):
...
+            # Skip test if CVXPY not installed
+            self.skipTest("CVXPY not installed.")
+        op1 = random_unitary(2**num_qubits)
+        op2 = random_unitary(2**num_qubits)
+        choi1 = Choi(op1)
+        choi2 = Choi(op2)
+        delta_choi = choi1 - choi2
+        try:
+            target = diamond_norm(delta_choi)
+            self.assertAlmostEqual(unitary_diamond_distance(op1, op2), target, places=4)
```
kevinsung: Can we increase the `places`?

owenagnel: I don't think we can go for more precision actually. The semi-definite formulation is quite inaccurate. Running tests on my machine for different absolute tolerances (I assume this is what places refers to) I get the following results over 1000 random tests:
|              | 2 qubit random unitary test failures |
|--------------|--------------------------------------------|
| 1e-4 abs tol | 0.0%                                       |
| 1e-5 abs tol | 20.9%                                      |
| 1e-6 abs tol | 70.20%                                     |

Also note that the existing test for `diamond_norm`, `test_diamond_norm` uses `places=4`.

test/python/quantum_info/operators/test_measures.py
owenagnel: It may be worth leaving the try catch statement. Looking at how QuTiP implements their `d_norm` tests, we can see the following.

First they comment:
```
# dnorm tests have always been slightly flaky; in some cases, cvxpy will fail
# to solve the problem, and this can cause an entire test-suite failure.  As
# long as we are using random tests (perhaps not ideal), this will happen
# occasionally.  This isn't entirely a bug, it's just a reality of using a
# one-size-fits-all solver; we've historically assumed users who come up
# against this sort of thing will be accepting of the fact that dnorm
# calculation is nontrivial, and isn't always entirely feasible.
#
# To deal with it, we allow each test to be rerun twice, using
# pytest-rerunfailures.  This should forbid pathological cases where the test
# is failing every time, but not penalise one-off failures.  As far as we know,
# the failing tests always involve a random step, so triggering a re-run will
# have them choose new variables as well.
```

Also they skip tests if the platform used is windows:
```
@pytest.mark.skipif("Windows" in platform.system(),
                    reason="Skipping dnorm tests -- cvxpy and cvxopt can be"
                    " installed on Windows but fail to correctly solve"
                    " the given optimization problems.")
```

Finally, note that in the existing implementation of `test_diamond_norm`, there is a similar try catch statement for `cvxpy.SolverError` (see line 167). Maybe this is in place for the reasons discussed above. Either way, we should make sure both tests have consistent implementations.

kevinsung: Hmm ok, let's leave it as is then.

jakelishman: Fwiw I wrote that comment 3 years ago before I started at IBM, and I didn't really know very much about testing strategy back then. Assuming (as my comment implies) `cvxpy` has some uncontrollable internal randomness in its algorithm, it's perhaps worth leaving this skip, but if we can make the algorithm fully deterministic across Python runs by seeding everything, it'll be safer to do that.

jakelishman: Also about Windows: cvxpy updated its solvers _a lot_ since that test skip in QuTiP was written - it may not be accurate anymore.

owenagnel: Wow! What a coincidence. In that case, I'll just seed everything and remove the try catch. It looks like the CI pipeline tests in windows anyways. 

What's the best approach here? 
- Merge the PR *without* the try catch and patch it if we get unexpected test failures 
- Or keep the try catch in

Sadly, I can't seem to install cvxpy on my machine (unless I'm using anaconda) so not sure how much experimentations I can do on my end.

owenagnel: On an unrelated note, do you know how I can join the qiskit slack? I think IBM must have updated something because I swear I used to be part of the workspace

kevinsung: >Merge the PR without the try catch and patch it if we get unexpected test failures

Let's go with this.

Re: Slack. Can you try logging in at http://qiskit.slack.com/?

owenagnel: Sounds good, have made the change. 

Note: as we're no longer using the cvxpy import (for the cvxpy solver error), I've changed the try/catch `import cvxpy` to use `importlib.util.find_spec` instead. Does thus sound good?

qiskit/quantum_info/operators/measures.py
```diff
@@ -350,6 +350,89 @@ def cvx_bmat(mat_r, mat_i):
     return sol
 
 
+def unitary_diamond_distance(channel1: Operator, channel2: Operator) -> float:
+    r"""Return the diamond distance between two unitary operators.
+
+    This function computes the completely-bounded trace-norm (often
+    referred to as the diamond-norm) of the difference of two unitary channels
+    (or unitary operators). This is often used as a distance metric in quantum
+    information This method is a more specialised implementaion of
```
kevinsung: spelling "implementation"

qiskit/quantum_info/operators/measures.py
```diff
@@ -350,6 +350,89 @@ def cvx_bmat(mat_r, mat_i):
     return sol
 
 
+def unitary_diamond_distance(channel1: Operator, channel2: Operator) -> float:
```
kevinsung: From a quick look at `_input_formatter`, I think the type annotation can be generalized to `Operator | QuantumChannel`, right?

qiskit/quantum_info/operators/measures.py
```diff
@@ -350,6 +350,89 @@ def cvx_bmat(mat_r, mat_i):
...
+        channel2 (Operator): a unitary operator.
+
+    Returns:
+        float: The completely-bounded trace norm :math:`\|U - V\|_{\diamond}`.
+
+    Raises:
+        QiskitError: if the input operators do not have the same dimensions or aren't unitary.
+
+    Additional Information:
+        The implementation uses an unproved result in Aharonov et al. (1998). Geometrically,
```
kevinsung: You proved it though, right? If your thesis will be published soon, perhaps we can wait and cite it here? Also fine with adding it later.

qiskit/quantum_info/operators/measures.py
```diff
@@ -350,6 +350,89 @@ def cvx_bmat(mat_r, mat_i):
...
+        D. Aharonov, A. Kitaev, and N. Nisan. “Quantum circuits with
+        mixed states” in Proceedings of the thirtieth annual ACM symposium
+        on Theory of computing, pp. 20-30, 1998.
+    """
+    op1 = _input_formatter(channel1, Operator, "unitary_diamond_distance", "channel1")
+    op2 = _input_formatter(channel2, Operator, "unitary_diamond_distance", "channel2")
+
+    # Check operators are unitary and have same dimension
+    if not op1.is_unitary():
+        raise QiskitError(
```
kevinsung: Raise ValueError instead, it's more informative.

qiskit/quantum_info/operators/measures.py
```diff
@@ -350,6 +350,89 @@ def cvx_bmat(mat_r, mat_i):
...
+        )
+    if not op2.is_unitary():
+        raise QiskitError(
+            "Invalid operator supplied to channel2 of unitary_diamond_distance"
+            "operators must be unitary."
+        )
+    if op1.dim != op2.dim:
+        raise QiskitError(
+            "Input quantum channel and target unitary must have the same "
+            "dimensions ({op1.dim} != {op2.dim})."
```
kevinsung: Missed f-string?

qiskit/quantum_info/operators/measures.py
```diff
@@ -350,6 +350,89 @@ def cvx_bmat(mat_r, mat_i):
...
+        )
+
+    # Compute the diamond norm
+    op1 = op1.data
+    op2 = op2.data
+    pre_diag = np.matmul(np.conj(op1).T, op2)
```
kevinsung: Prefer `@` for matrix multiplication.

qiskit/quantum_info/operators/measures.py
```diff
@@ -350,6 +350,89 @@ def cvx_bmat(mat_r, mat_i):
...
+def _find_poly_distance(eigenvals: np.ndarray) -> float:
+    """Function to find the distance between origin and the convex hull of eigenvalues."""
+    phases = np.angle(eigenvals)
+    pos_max = phases.max()
+    neg_min = phases.min()
+    pos_min = np.where(phases > 0, phases, np.inf).min()
+    neg_max = np.where(phases <= 0, phases, -np.inf).max()
+
+    if neg_min > 0:  # all eigenvalues have positive phase, so hull is above x axis
+        return np.cos((pos_max - pos_min) / 2)
```
kevinsung: Normally we put comments in their own line. Example:
```
    if neg_min > 0:
        # all eigenvalues have positive phase, so hull is above x axis
        return np.cos((pos_max - pos_min) / 2)
```

test/python/quantum_info/operators/test_measures.py
```diff
@@ -178,11 +180,32 @@ def test_diamond_norm(self, num_qubits):
...
+    @combine(num_qubits=[1, 2, 3])
+    def test_unitary_diamond_distance_random(self, num_qubits):
+        """Tests the unitary_diamond_distance for random unitaries.
+        Compares results with semi-definite program."""
+        if importlib.util.find_spec("cvxpy") is None:
+            # Skip test if CVXPY not installed
+            self.skipTest("CVXPY not installed.")
+
+        op1 = random_unitary(2**num_qubits, seed=660477)
+        op2 = random_unitary(2**num_qubits, seed=765720)
```
kevinsung: Better practice is to seed an RNG object and pass it. e.g.

```python
        rng = np.random.default_rng(1234)
        op1 = random_unitary(2**num_qubits, seed=rng)
        op2 = random_unitary(2**num_qubits, seed=rng)
```

kevinsung (changes requested):

qiskit/quantum_info/operators/measures.py
kevinsung: I think it's fine as is, but feel free to submit a PR to cite your work once it's published.

qiskit/quantum_info/operators/measures.py
owenagnel: Hmm, I'm not super familiar with the `QuantumChannel` data type. If the channel is unitary, will running `channel.data` always return an n by n unitary matrix? Do I have to add 
```
if isinstance(op1, QuantumChannel):
            op1  = op1.to_operator()
```
Or does everything already work nicely

qiskit/quantum_info/operators/measures.py
owenagnel: Unfortunately I don't think it will be published super soon (maybe on arxiv at some point). The thesis is submitted for an examination at Oxford. I can send you the proof directly, but it shouldn't be circulated for the time being.  If it does get published, I'll make sure to open a new PR.

qiskit/quantum_info/operators/measures.py
kevinsung: Hmm on further thought let's just keep it to Operator because that's the class meant to store unitary objects. Maybe add QuantumCircuit, but I don't feel strongly about it.

test/python/quantum_info/operators/test_measures.py
```diff
@@ -178,11 +180,33 @@ def test_diamond_norm(self, num_qubits):
...
+    def test_unitary_diamond_distance_random(self, num_qubits):
+        """Tests the unitary_diamond_distance for random unitaries.
+        Compares results with semi-definite program."""
+        if importlib.util.find_spec("cvxpy") is None:
+            # Skip test if CVXPY not installed
+            self.skipTest("CVXPY not installed.")
+
+        rng = np.random.default_rng(1234)
+        op1 = random_unitary(2**num_qubits, seed=rng)
+        op2 = random_unitary(2**num_qubits, seed=765720)
```
kevinsung:
```suggestion
        op2 = random_unitary(2**num_qubits, seed=rng)
```

qiskit/quantum_info/operators/measures.py
kevinsung: Ok sounds good.

qiskit/quantum_info/operators/measures.py
```diff
@@ -350,6 +350,94 @@ def cvx_bmat(mat_r, mat_i):
     return sol
 
 
+def unitary_diamond_distance(channel1: Operator, channel2: Operator) -> float:
```
kevinsung: Rename these inputs to something like `op1` and `op2`. We don't want to call them `channel` if the input shouldn't be a QuantumChannel.

owenagnel: Sounds good!

kevinsung (dismissed):

qiskit/quantum_info/operators/measures.py
```diff
@@ -350,6 +350,94 @@ def cvx_bmat(mat_r, mat_i):
     return sol
 
 
+def unitary_diamond_distance(op1: Operator, op2: Operator) -> float:
```
chriseclectic: An idea is this could be implemented as a general function
```suggestion
def diamond_distance(op1: BaseOperator, op2: BaseOperator) -> float:
```
where if the operator type are unitary you can have special cases for implementing it eg this current case for Operators, and in general this can call diamond norm for `op1 - op2` appropriately normalized.

There is also an efficient expression for the diamond distance between Pauli channels that could also be added as a special case later as well.

Probably the trick used here could also be special cased for Pauli's and Cliffords to have an efficient version for those that works for large numbers of qubits?

qiskit/quantum_info/operators/measures.py
```diff
@@ -350,6 +350,94 @@ def cvx_bmat(mat_r, mat_i):
...
+    (or unitary operators). This is often used as a distance metric in quantum
+    information This method is a more specialised implementation of
+    :meth:`~qiskit.quantum_info.diamond_norm`.
+
+    Args:
+        op1 (Operator): a unitary operator.
+        op2 (Operator): a unitary operator.
+
+    Returns:
+        float: The completely-bounded trace norm :math:`\|U - V\|_{\diamond}`.
```
chriseclectic: It would be nice to include the equation `d(E, F) = ||E-F||_diamond` in the main body rather than in the return docs.

qiskit/quantum_info/operators/measures.py
```diff
@@ -350,6 +350,94 @@ def cvx_bmat(mat_r, mat_i):
...
+
+    Raises:
+        ValueError: if the input operators do not have the same dimensions or aren't unitary.
+
+    Additional Information:
+        The implementation uses an unproved result in Aharonov et al. (1998). Geometrically,
```
chriseclectic: If changed to a general function as suggested, this would need to be updated to say if both input operators are unitary this implementation uses ...

(the fact it says "unproved" makes it sound super sketchy though!)

test/python/quantum_info/operators/test_measures.py
```diff
@@ -178,11 +180,33 @@ def test_diamond_norm(self, num_qubits):
...
+    @combine(num_qubits=[1, 2, 3])
+    def test_unitary_diamond_distance_random(self, num_qubits):
+        """Tests the unitary_diamond_distance for random unitaries.
+        Compares results with semi-definite program."""
+        if importlib.util.find_spec("cvxpy") is None:
+            # Skip test if CVXPY not installed
+            self.skipTest("CVXPY not installed.")
+
+        rng = np.random.default_rng(1234)
+        op1 = random_unitary(2**num_qubits, seed=rng)
```
chriseclectic: If these tests aren't too slow it it might be nice to add a loop here to run over a bunch of random unitaries rather than just 1.

test/python/quantum_info/operators/test_measures.py
```diff
@@ -178,11 +180,33 @@ def test_diamond_norm(self, num_qubits):
...
+        angles = np.linspace(0, 2 * np.pi, 10, endpoint=False)
+        for angle in angles:
+            op1 = Operator(RZGate(angle))
+            op2 = Operator.from_label("I")
+            d2 = np.cos(angle / 2) ** 2  # analytical formula for hull distance
+            target = np.sqrt(1 - d2) * 2
+            self.assertAlmostEqual(unitary_diamond_distance(op1, op2), target, places=7)
+
+    @combine(num_qubits=[1, 2, 3])
+    def test_unitary_diamond_distance_random(self, num_qubits):
```
chriseclectic: It would also be good to add some random unit tests for comparing random cliffords, and random paulis for the new function to the diamond norm.

chriseclectic (changes requested):
I look forward to seeing the proof of this result.

In general I think it makes more sense to make this function a more general `diamond_distance` function with a specialized efficient case for two unitary inputs (like we have specialized cases for fidelity and other functions with unitary inputs), and the general case can be a call to the existing `diamond_norm` function.

owenagnel:
> I look forward to seeing the proof of this result.
> 
> In general I think it makes more sense to make this function a more general `diamond_distance` function with a specialized efficient case for two unitary inputs (like we have specialized cases for fidelity and other functions with unitary inputs), and the general case can be a call to the existing `diamond_norm` function.

Hi Chris! Thanks for the comments, and I completely agree, it makes more sense to generalise the function to all CPTP channels. Regarding the proof, I'm quite happy to send it to you, however, it is part of a dissertation submitted for examination and shouldn't be circulated for the time being. Shall I send it to your IBM email?

I'll make the changes you outlined ASAP.

test/python/quantum_info/operators/test_measures.py
owenagnel: On my machine it takes around 5.22s to evaluate the diamond norm for 3 qubit channels. Maybe we can run over a bunch of unitaries when num_qubits < 3 and only run one test otherwise?

test/python/quantum_info/operators/test_measures.py
owenagnel: I'm curious what this adds at the moment. Perhaps if optimisations are added in the future for Pauli channels or Cliffords this would make sense... But to be honest, I can't think of a simple optimisation for Clifford circuits (please let me know if you have something in mind).

qiskit/quantum_info/operators/measures.py
kevinsung: Yeah there's no need to say "unproved" here. They stated it as a fact and "left the proof to the reader." You're welcome to link your proof once it's published, of course.

test/python/quantum_info/operators/test_measures.py
chriseclectic: Running the random tests for 1 and 2 qubit only sounds fine to me

test/python/quantum_info/operators/test_measures.py
chriseclectic: The main reason is to have tests that Pauli and Clifford inputs work correctly (via coercion to Operator inside the function). If it is possible to add an optimized implementation for these later, then this would also be a test for that.

For an optimized implementation, I am wondering if for Pauli's you can use the fact that U^dg.V is a Pauli, and hence you can get all the eigen vecs and vals easily, does that make it easy to analytically compute the distance `d`. For Cliffords I'm not sure if there is any structure you can exploit or not, maybe its as hard as general unitaries?

owenagnel: Hi, I have made an optimisation for Paulis (turns out if the operators aren't equal, the diamond norm is automatically 2 since eigenvalues of all Paulis are -1, and we need all the Paulis to cancel out. Global phase is ignored.)

kevinsung (approved):

PR #12347, opened 2024-05-06
Add standalone test file for Clifford synthesis functions

ElePT:
### Summary
This PR is a spin-off from #12203 with a few minor improvements to the synthesis functions:

- Moving Clifford synthesis tests to a standalone file in `test/python/synthesis` instead of `test/python/quantum_info`
- A proposal to avoid cyclic import errors in synthesis files (this allows to freely import `DAGCircuit` inside `qiskit.synthesis` in #12203, which would normally raise a cyclic import error)
- Adding `copy=False` to `circuit.append`

### DIFF
--- a/test/python/synthesis/test_clifford_sythesis.py
+++ b/test/python/synthesis/test_clifford_sythesis.py
@@ -15,26 +15,7 @@
 
 import numpy as np
 from ddt import ddt
-from qiskit.circuit import Gate, QuantumCircuit, QuantumRegister
-from qiskit.circuit.library import (
-    CXGate,
-    CYGate,
-    CZGate,
-    DCXGate,
-    ECRGate,
-    HGate,
-    IGate,
-    SdgGate,
-    SGate,
-    SXGate,
-    SXdgGate,
-    SwapGate,
-    XGate,
-    YGate,
-    ZGate,
-    iSwapGate,
-)
-from qiskit.converters.dag_to_circuit import dag_to_circuit
+from qiskit.circuit.random import random_clifford_circuit
 from qiskit.quantum_info.operators import Clifford
 from qiskit.synthesis.clifford import (
     synth_clifford_full,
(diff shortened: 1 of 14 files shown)


test/python/synthesis/test_clifford_sythesis.py
```diff
@@ -0,0 +1,218 @@
...
+        "v": (VGate(), 1),
+        "w": (WGate(), 1),
+        "cx": (CXGate(), 2),
+        "cy": (CYGate(), 2),
+        "cz": (CZGate(), 2),
+        "swap": (SwapGate(), 2),
+        "iswap": (iSwapGate(), 2),
+        "ecr": (ECRGate(), 2),
+        "dcx": (DCXGate(), 2),
+    }
```
alexanderivrii: It would be nice to avoid duplicating this code. Actually, this is a useful function, can we move it to `qiskit/circuit/random/utils.py`? Let's also remove support for `VGate` and `WGate` in this function, I don't think we really need random clifford circuits to contain these gates (or else these gates would have been added to the circuit library).

tools/pgo_scripts/test_utility_scale.py
```diff
@@ -16,6 +16,7 @@
 
 import os
 
+# pylint: disable=cyclic-import
 from qiskit.transpiler.preset_passmanagers import generate_preset_pass_manager
```
alexanderivrii: I remember that #12203 had some problems with the cyclic imports. But do we really need this pylint directive now? I am completely in favor of all the other pylint-related changes.

alexanderivrii (commented):
Thank you very much -- this makes the code cleaner and more efficient (in particular avoiding to copy the subcircuits in clifford synthesis algorithms). My only real suggestion was to move the random clifford generation circuit directly to Qiskit (and remove its support for V and W gates that are not available in the standard library). Of course, that also means that you would need the release notes :).

tools/pgo_scripts/test_utility_scale.py
ElePT: True, removed in [0399e03](https://github.com/Qiskit/qiskit/pull/12347/commits/0399e033b223d7473d67c9ffab6bf09adfb59dc0), it can be re-added when necessary.

test/python/synthesis/test_clifford_sythesis.py
ElePT: Done in [df9e6fd](https://github.com/Qiskit/qiskit/pull/12347/commits/df9e6fdb4bd89e44adcab962804b4025954c7242). Removing the VGate and WGate also means these gates are not tested in the quantum info Clifford tests, but I agree that if these gates aren't used we don't necessarily need to go out of our way to test them.

alexanderivrii (approved):
Thank you, LGTM!

PR #12356, opened 2024-05-07
Removing broad-exception-raised lint rule and updates

joesho112358:
### Summary

#9614 removing `broad-exception-raised`

### Details and comments

Updates for removing the `broad-exception-raised` rule


qiskit/visualization/bloch.py
```diff
@@ -290,7 +290,7 @@ def set_label_convention(self, convention):
             self.zlabel = ["$\\circlearrowleft$", "$\\circlearrowright$"]
             self.xlabel = ["$\\leftrightarrow$", "$\\updownarrow$"]
         else:
-            raise Exception("No such convention.")
+            raise NotImplementedError("No such convention.")
```
jakelishman: This is maybe a `ValueError` because it refers to being given an incorrect string.

qiskit/visualization/bloch.py
```diff
@@ -396,7 +396,7 @@ def add_annotation(self, state_or_vector, text, **kwargs):
         if isinstance(state_or_vector, (list, np.ndarray, tuple)) and len(state_or_vector) == 3:
             vec = state_or_vector
         else:
-            raise Exception("Position needs to be specified by a qubit " + "state or a 3D vector.")
+            raise RuntimeError("Position needs to be specified by a qubit state or a 3D vector.")
```
jakelishman: This is maybe a `TypeError` because we'd have been given a completely incorrect type object.

qiskit/quantum_info/quaternion.py
```diff
@@ -43,7 +43,7 @@ def __mul__(self, r):
             out_data[3] = r(0) * q(3) - r(1) * q(2) + r(2) * q(1) + r(3) * q(0)
             return Quaternion(out_data)
         else:
-            raise Exception("Multiplication by other not supported.")
+            raise RuntimeError("Multiplication by other not supported.")
```
jakelishman: This should probably `return NotImplemented` rather than manually raise an exception.

qiskit/visualization/transition_visualization.py
```diff
@@ -72,10 +72,10 @@ def __mul__(self, b):
             return self._multiply_with_quaternion(b)
         elif isinstance(b, (list, tuple, np.ndarray)):
             if len(b) != 3:
-                raise Exception(f"Input vector has invalid length {len(b)}")
+                raise RuntimeError(f"Input vector has invalid length {len(b)}")
             return self._multiply_with_vector(b)
         else:
-            raise Exception(f"Multiplication with unknown type {type(b)}")
+            raise TypeError(f"Multiplication with unknown type {type(b)}")
```
jakelishman: This should likely `return NotImplemented` - Python will manually raise a suitable `TypeError` if the mirrored multiplication (`type(b).__rmul__(b, self)`) fails as well, but raising an exception means the mirror multiplication never gets attempted.

qiskit/visualization/transition_visualization.py
```diff
@@ -72,10 +72,10 @@ def __mul__(self, b):
             return self._multiply_with_quaternion(b)
         elif isinstance(b, (list, tuple, np.ndarray)):
             if len(b) != 3:
-                raise Exception(f"Input vector has invalid length {len(b)}")
+                raise RuntimeError(f"Input vector has invalid length {len(b)}")
```
jakelishman: This is possibly a `ValueError`?

test/benchmarks/utils.py
```diff
@@ -71,7 +71,7 @@ def random_circuit(
         Exception: when invalid options given
     """
     if max_operands < 1 or max_operands > 3:
-        raise Exception("max_operands must be between 1 and 3")
+        raise RuntimeError("max_operands must be between 1 and 3")
```
jakelishman: This one's a `ValueError`

jakelishman (commented):
Thanks for looking at this - this is _definitely_ a good lint to get enabled.  I left a couple of comments about avoiding `RuntimeError` - we usually try and avoid that one as much as possible because there's generally some tighter bound to apply.  It's all a little bit subjective of course, but we (at least I!) generally try to keep `RuntimeError` for exceptions in defensive programming bits amongst ourselves, like if an internal private invariant has somehow become violated.

(Fwiw, it matters rather less what we use in tests, since if we're ever raising an `Exception` in tests, it's usually just to indicate a test failure or something.)

joesho112358:
> Thanks for looking at this - this is _definitely_ a good lint to get enabled. I left a couple of comments about avoiding `RuntimeError` - we usually try and avoid that one as much as possible because there's generally some tighter bound to apply. It's all a little bit subjective of course, but we (at least I!) generally try to keep `RuntimeError` for exceptions in defensive programming bits amongst ourselves, like if an internal private invariant has somehow become violated.
> 
> (Fwiw, it matters rather less what we use in tests, since if we're ever raising an `Exception` in tests, it's usually just to indicate a test failure or something.)

seems fair to me! updates are in for another look

test/python/quantum_info/test_quaternions.py
```diff
@@ -92,12 +92,12 @@ def test_mul_by_quat(self):
...
         """Quaternions cannot be multiplied with an array."""
         other_array = np.array([0.1, 0.2, 0.3, 0.4])
-        self.assertRaises(Exception, self.quat_unnormalized.__mul__, other_array)
+        self.assertEqual(NotImplemented, self.quat_unnormalized.__mul__(other_array))
 
     def test_mul_by_scalar(self):
         """Quaternions cannot be multiplied with a scalar."""
         other_scalar = 0.123456789
-        self.assertRaises(Exception, self.quat_unnormalized.__mul__, other_scalar)
+        self.assertEqual(NotImplemented, self.quat_unnormalized.__mul__(other_scalar))
```
jakelishman: This is pre-existing, stemming from the previous implementation of these magic methods being incorrect, but in general we don't want to call `__mul__` directly - `NotImplemented` is just a sentinel value that's properly interpreted by Python.  The normal forms of these tests would be
```python
with self.assertRaises(TypeError):
    _ = self.quat_unnormalized * other_array
```
and similar for the other one.

This actually means that Python will also try the `__rmul__` methods on `other_array` and `other_scalar` as well before raising the `TypeError`, in case either of those classes is something that can explicitly handle multiplication by `Quaternion`, though in this case, both will fail.  The Numpy array test will actually _start_ to succeed (Numpy will attempt to broadcast the multiplication), but then it'll fail the internal scalar multiplication.

jakelishman (commented):
Thanks for the changes! One comment about a pre-existing test that ought to be written in a different way, otherwise this is good to merge.

test/python/quantum_info/test_quaternions.py
joesho112358: i see. yeah, i was thinking this was more of a unit test vs integration test. updates on the way!

jakelishman (approved):
Ace, thanks for the changes!

PR #12369, opened 2024-05-08
fix edge coloring bug in plot_coupling_map

### LINKED ISSUE #12354: plot_circuit_layout colors edges incorrectly
### Environment

- **Qiskit version**: 
- **Python version**: 
- **Operating system**: 

### What is happening?

I believe the intent is for the black edges to connect the black qubits.

```python
from qiskit.circuit.library import EfficientSU2
from qiskit.visualization import plot_circuit_layout

# Construct circuit
n_qubits = 10
circuit = EfficientSU2(n_qubits, entanglement="pairwise")

# Initialize simulator backend
coupling_map = CouplingMap.from_heavy_hex(3)
backend = GenericBackendV2(
    coupling_map.size(),
    basis_gates=["ecr", "id", "rz", "sx", "x"],
    coupling_map=coupling_map,
    seed=1234
)

# Transpile circuit
pass_manager = generate_preset_pass_manager(
    optimization_level=3, backend=backend, seed_transpiler=1234
)
isa_circuit = pass_manager.run(circuit)

plot_circuit_layout(isa_circuit, backend, view="physical")
```
image

### What should happen?

Black edges should connect black qubits.

### Any suggestions?

_No response_

kevinsung:
### Summary

Fixes #12354 .

### Details and comments

The conversion to an undirected graph changes the edge ordering, causing it not to match the color ordering. This change uses a dictionary to recover the correct ordering.

### DIFF
--- a/qiskit/visualization/gate_map.py
+++ b/qiskit/visualization/gate_map.py
@@ -1039,7 +1039,9 @@ def plot_coupling_map(
         graph = CouplingMap(coupling_map).graph
 
     if not plot_directed:
+        line_color_map = dict(zip(graph.edge_list(), line_color))
         graph = graph.to_undirected(multigraph=False)
+        line_color = [line_color_map[edge] for edge in graph.edge_list()]
 
     for node in graph.node_indices():
         graph[node] = node
--- a/releasenotes/notes/plot-circuit-layout-5935646107893c12.yaml
+++ b/releasenotes/notes/plot-circuit-layout-5935646107893c12.yaml
@@ -1,0 +1,5 @@
+---
+fixes:
+  - |
+    Fixed a bug in :func:`plot_coupling_map` that caused the edges of the coupling map to be colored incorrectly.
+    See https://github.com/Qiskit/qiskit/pull/12369 for details.


mtreinish (commented):
This looks good to me. It relies on the internal implementation detail about how the `to_undirected` method works in rustworkx (internally iterates over the edges in id order). But while undocumented it is unlikely to ever change so it's fine. The only thing is could you add a release note to document this fix?

kevinsung:
> This looks good to me. It relies on the internal implementation detail about how the `to_undirected` method works in rustworkx (internally iterates over the edges in id order). But while undocumented it is unlikely to ever change so it's fine. The only thing is could you add a release note to document this fix?

Sorry, I don't understand. I think the iteration order doesn't matter because `line_color_map` is a dictionary.

mtreinish:
Sorry that was mistake I misread the `edge_list()` method as `edge_indices()` somehow and thought you were just doing a mapping of explicit index to the index in a list. Which would work (and actually would be slightly faster) but would depend on the internal behavior. What you have here is more robust and a better approach.

mtreinish (approved):

Issue #12205, opened 2024-04-18
Add a rust representation of operations, instructions, and gates


mtreinish:
### What should we add?

As part of the ongoing effort to be able to more of Qiskit's operations in rust for performance and efficiency the last major piece (besides the `DAGCircuit` in #11721) are the operations object. This issue is to track the development of rust data structures that represent the the circuit's operations. 

The main complexity here is that the python side of Qiskit exposes a very flexible and dynamic interface for defining operations. This includes fully custom operation objects that are defined in Python and generate their contents dynamically. We want to find a model where we retain this functionality and capability from Python as it's quite widely used, but also ensure we're storing an efficient representation in rust as the source of truth. The key goal of this work though is to enable working with gates from rust without any involvement from Python.

mtreinish:
In my ideal model for the standard library our internal representation would be an `enum` that has a variant for each standard library gate and trait implementations for all of the data access. This would minimize what we store in the circuit structure to a minimal size. Something like:

```rust
#[derive(Clone)]
#[pyclass]
pub enum StandardGate {
    ZGate,
    YGate,
    XGate,
    CZGate,
    CYGate,
    CXGate,
    CCXGate,
    RXGate,
    RYGate,
    RZGate,
    ECRGate,
    SwapGate,
    SXGate,
    GlobalPhaseGate,
    IGate,
    HGate,
    ....
}
```

The trait impls for this enum aren't the most ergonomic to work with, but the efficiency of the structure seems worth it (especially as we don't modify the standard gate library too often)

PR #12372, opened 2024-05-08
[DAGCircuit Oxidation] Refactor bit management in `CircuitData`

kevinhartman:
### Summary

This is part of the work we're doing to port `DAGCircuit` to Rust. It's work I've pulled out of a local branch where I'm doing that port, in an effort to reduce the size of that eventual PR.

The primary improvement that this brings is the introduction of `BitData`, which encapsulates keeping Python and native bit indices in sync. It provides methods like `map_indices` and `map_bits` which improve the ergonomics of translating between native and Python bits at PyO3 boundaries.

### Details and comments
- Added zero-cost struct types for our native `Qubit` and `Clbit` representation.
- Added a new struct `BitData` which encapsulates maintaining a mapping between native Rust bit types and Python bit types. This significantly cleans up the code in `CircuitData` and makes it reusable for `DAGCircuit`.
- Made `InternContext` into a more generic `IndexedInterner` data structure, which implements the new `Interner` trait with slotted interning for any `T` which is `Hash` and `Eq`.
- Added `Interner` trait, for defining interner semantics for containers like `IndexedInterner`.

The one caveat to this PR is that we now no longer share interning between `Qubit` and `Clbit` indices, since this is more difficult from a typing perspective. If we want to share a single `IndexedInterner` for both `Qubit` and `Clbit` operand lists, this would be possible, but we should revisit it. For now I wanted to just focus on keeping things strongly typed so that our Rust code doesn't get overly messy at this early stage.


crates/circuit/src/bit_data.rs
```diff
@@ -0,0 +1,163 @@
...
+/// to determine equality. If these are not equal, only then does
+/// it call `repr()` on both sides, which has a significant
+/// performance advantage.
+#[derive(Clone, Debug)]
+struct BitAsKey {
+    /// Python's `hash()` of the wrapped instance.
+    hash: isize,
+    /// The wrapped instance.
+    bit: PyObject,
+}
```
mtreinish, line 37: I'm wondering if we maybe should remove one of the layers of abstraction here. It's something I've been thinking about in #12205 is that the source of truth for bits is still python but for interaction from rust space this isn't super ergonomic. I'm wondering if instead of playing games like this we try flattening the two representations. I guess the complexity is things like register and index (although https://github.com/Qiskit/qiskit/pull/11596 will take care of that)

jakelishman: I'm not sure if this is what you were getting at, but I was thinking longer term that we'll want/need to make it possible to construct circuits from Rust space without creating _any_ Python objects at all.  We might not be at that point yet, but it might to be thinking about it.

mtreinish: Yeah, that's basically what I was getting at, having a pattern for making a circuit without any python interaction is what we'll want longer term. The example I have for making circuits from rust space right now still needs a `py` handle because we need to instantiate `Qubit` objects.

kevinhartman: I'd agree as well. `BitAsKey` isn't new here, though. I've only moved it from `CircuitData`.

Also, the source of truth for bits is `CircuitData`. It owns its bits, and `QuantumCircuit` simply "references" them.

kevinhartman:
I'm rethinking a bit of this PR (specifically the stuff with packing and interning) so I'm going to put this into draft for now.

The `BitData` stuff is likely to stay, but I'm likely to reduce the traiting stuff around packing and interning a bit.

crates/circuit/src/bit_data.rs
```diff
@@ -0,0 +1,192 @@
...
+
+impl BitAsKey {
+    pub fn new(bit: &Bound<PyAny>) -> Self {
+        BitAsKey {
+            // This really shouldn't fail, but if it does,
+            // we'll just use 0.
+            hash: bit.hash().unwrap_or(0),
+            bit: bit.into_py(bit.py()),
```
mtreinish: Is there a reason to not use `unbind()` here?

crates/circuit/src/circuit_instruction.rs
```diff
@@ -63,15 +63,37 @@ pub struct CircuitInstruction {
...
+            Some(&PyTuple::new_bound(py, clbits)),
+        )
+    }
+}
+
 #[pymethods]
 impl CircuitInstruction {
     #[new]
-    pub fn new(
+    pub fn py_new(
```
mtreinish, line 90: Oh, I like this pattern, this is something we need to do a lot and have a dedicated python space constructor like this works well.

crates/circuit/src/circuit_instruction.rs
```diff
@@ -63,15 +63,37 @@ pub struct CircuitInstruction {
     pub clbits: Py<PyTuple>,
 }
 
+impl CircuitInstruction {
+    pub fn new<T1, T2, U1, U2>(
+        py: Python,
+        operation: PyObject,
+        qubits: impl IntoIterator<Item = T1, IntoIter = U1>,
+        clbits: impl IntoIterator<Item = T2, IntoIter = U2>,
+    ) -> PyResult<Py<Self>>
```
mtreinish: If this is the rust space constructor why are we putting it in a `Py<>`?

crates/circuit/src/lib.rs
```diff
@@ -28,6 +31,36 @@ pub enum SliceOrInt<'a> {
     Slice(Bound<'a, PySlice>),
 }
 
+type BitType = u32;
+#[derive(Copy, Clone, Debug, Hash, Ord, PartialOrd, Eq, PartialEq)]
+struct Qubit(BitType);
+#[derive(Copy, Clone, Debug, Hash, Ord, PartialOrd, Eq, PartialEq)]
+struct Clbit(BitType);
```
mtreinish: Should these be public?

mtreinish (commented):
This looks fine to me it's mostly moving code around, just a couple of questions inline.

crates/circuit/src/bit_data.rs
kevinhartman: We can, it didn't exist back when I wrote this code originally (I just moved it here).

crates/circuit/src/circuit_instruction.rs
kevinhartman: It's an interesting point. It'd probably be a better factoring to just return `Self`, but the only callers right now stuff it into a `Py` so I did it here for less code.

I think there's a much larger conversation we should have around the separation we'd ultimately want between the Python and Rust (model) hierarchies. Especially with the foresight that we'll probably want a pure Rust API at some point, which would likely mean having core Rust-only structs and then PyO3 wrappers around those.

Personally, I'd like the `qiskit-circuit` crate to not depend on PyO3 at all, and then to have a new crate `qiskit-pyo3` that depends on PyO3 and `qiskit-circuit` (and other core Rust-only crates) to bridge everything together. Before we get to that point, I'm not sure it matters too much for `new` versus `py_new` to be entirely Python versus entirely Rust, since I'd think we'd want to end up with two separate structs for each case, anyway.

crates/circuit/src/lib.rs
kevinhartman: At some point, yes. But, they aren't used in the public interface of the crate at the moment.

I also want to call out that I'm not especially keen on us implementing `From<BitType>` for `Qubit` and `Clbit` and `From<Qubit|Clbit>` for `BitType`, as I've done here. To me, that exposes what could be an implementation detail of `BitData` (how the bits are mapped between the native type and Python instances).

I have a separate branch where I explored removing those conversions. I thought it would be nice if `Qubit` and `Clbit` were defined in `bit_data.rs`, and the implementation of `BitData` encapsulated the awareness of these conversions to and from a `u32` (`BitType`), and then we simply exposed `Qubit` and `Clbit` opaquely in `lib.rs`. I decided against this for now, since Python users already expect that bits are contiguous and ordered by insertion in a circuit (and hence convertible to and from an index). But I'd be curious to hear your thoughts, and perhaps what @jakelishman has to say when he's back from holiday.

kevinhartman: Made public in 964d6fc550674e91ee2198ece946eb54e84a3602.

Though I do suspect we may want to make the underlying representation more private. I'll likely explore that as part of porting `DAGCircuit`.

crates/circuit/src/bit_data.rs
kevinhartman: Fixed in d4ef317f15add9a286158bd08ed2a71e9b6df47e

crates/circuit/src/circuit_instruction.rs
kevinhartman: Done in 84fad3e252395154a3d47d3555cfdab84e6f4293.

mtreinish (approved):
This LGTM now, thanks for making the updates.

PR #12373, opened 2024-05-09
Removing unnecessary-dict-index-lookup lint rule and updates

joesho112358:
### Summary

#9614 remove `unnecessary-dict-index-lookup`

### Details and comments

Removing `unnecessary-dict-index-lookup` and updates


jakelishman (approved):
This just seems obviously good to get enabled, thanks!

PR #12375, opened 2024-05-09
Fix a corner case of `SparsePauliOp.apply_layout`

t-imamichi:
### Summary

`SparsePauliOp.apply_layout` raises an error if a zero-qubit operator is given.
This PR fixes the case.

```python

op = SparsePauliOp("")
print(op.apply_layout(None, 3))
```

main branch
```
SparsePauliOp([''],
              coeffs=[1.+0.j])
  File "/Users/ima/tasks/4_2024/qiskit/terra/tmp/layout.py", line 5, in <module>
    print(op.apply_layout(None, 3))
  File "/Users/ima/tasks/4_2024/qiskit/terra/qiskit/quantum_info/operators/symplectic/sparse_pauli_op.py", line 1167, in apply_layout
    return new_op.compose(self, qargs=layout)
  File "/Users/ima/tasks/4_2024/qiskit/terra/qiskit/quantum_info/operators/symplectic/sparse_pauli_op.py", line 349, in compose
    q = np.logical_and(z1[:, np.newaxis], x2).reshape((-1, num_qubits))
ValueError: cannot reshape array of size 0 into shape (0)
```

this PR
```
SparsePauliOp(['III'],
              coeffs=[1.+0.j])
```

Although `Pauli.apply_layout` works correctly, I added the zero-qubit test cases to ensure it.

### DIFF
--- a/releasenotes/notes/fix-sparse-pauli-op-apply-layout-zero-43b9e70f0d1536a6.yaml
+++ b/releasenotes/notes/fix-sparse-pauli-op-apply-layout-zero-43b9e70f0d1536a6.yaml
@@ -1,3 +1,3 @@
 fixes:
   - |
-    Fixed :meth:`~.SparsePauliOp.apply_layout` to work correctly with zero-qubit operators.
+    Fixed :meth:`.SparsePauliOp.apply_layout` to work correctly with zero-qubit operators.
--- a/releasenotes/notes/fix-sparse-pauli-op-apply-layout-zero-43b9e70f0d1536a6.yaml
+++ b/releasenotes/notes/fix-sparse-pauli-op-apply-layout-zero-43b9e70f0d1536a6.yaml
@@ -1,0 +1,3 @@
+fixes:
+  - |
+    Fixed :meth:`~.SparsePauliOp.apply_layout` to work correctly with zero-qubit operators.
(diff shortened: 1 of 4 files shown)


t-imamichi:
I noticed that I have to fix Pauli as well.

t-imamichi:
I confirmed that the current Pauli.apply_laytout works correctly. So, we don't have to update it. I just added tests to ensure it.

ewinston:
Looks good although maybe could be made briefer using `@data` like `test_zero_qubit_pauli_construction` above.

t-imamichi:
Thank you for your suggestion, Erick. I simplified this PR with `combine`.
I also applied isort to remove `pylint: disable=wrong-import-order`

releasenotes/notes/fix-sparse-pauli-op-apply-layout-zero-43b9e70f0d1536a6.yaml
```diff
@@ -0,0 +1,3 @@
+fixes:
+  - |
+    Fixed :meth:`.SparsePauliOp.apply_layout` to work correctly with zero-qubit operators.
```
mtreinish, line 3: An example here would be really good. Something like:

```suggestion
    Fixed :meth:`.SparsePauliOp.apply_layout` to work correctly with zero-qubit operators.
    For example, if you previously created a 0 qubit and applied a layout like::
    
        op = SparsePauliOp("")
        op.apply_layout(None, 3)

    this would have previously raised an error. Now this will correctly return an operator of the form:
    ``SparsePauliOp(['III'], coeffs=[1.+0.j])``
```

mtreinish (approved):

releasenotes/notes/fix-sparse-pauli-op-apply-layout-zero-43b9e70f0d1536a6.yaml
t-imamichi: Thank you for revising reno.

PR #12376, opened 2024-05-09
Fix runtime scaling of StarPreRouting pass

mtreinish:
### Summary

This commit fixes a runtime performance scaling issue with the new StarPreRouting pass. If there are any stars identified by the pass when the pass goes to pre-route those star connectivity blocks it specifies a custom lexicographical topological sort key to ensure the stars are kept together in the sort. However the mechanism by which this sort key was generated scaled quadratically with the number of DAG nodes in the identified stars. This ended up being a large runtime performance bottleneck. This commit fixes this issue by pre-computing the sort key for all nodes in the stars and putting that in a dictionary so that when we call rustworkx to perform the topological sort the sort key callback does not become the bottleneck for the entire pass.

### Details and comments


kevinhartman (approved):
Nice, looks good to me.

PR #12380, opened 2024-05-09
[DAGCircuit Oxidation] Port `DAGNode` to Rust

kevinhartman:
### Summary
Ports `DAGNode` and subclasses to Rust. This is precursory to writing efficient ports of methods on `DAGCircuit` in Rust, since we can avoid a lot of interpreter interaction.

### Details and comments
Everything is ported over with the exception of `semantic_eq`, since its implementation has a lot of interplay with components that we currently must access through the Python interpreter (e.g. `Expr`, `Parameter`, registers, etc.).


crates/circuit/src/circuit_instruction.rs
```diff
@@ -129,7 +129,7 @@ impl CircuitInstruction {
         )
     }
 
-    fn __getstate__(&self, py: Python<'_>) -> PyObject {
+    pub fn __getstate__(&self, py: Python<'_>) -> PyObject {
```
jakelishman: It probably doesn't matter, but this makes me nervous - it feels very weird to have the `__getstate__` methods be officially marked "public access".  They're not typically even meant to be called directly in Python space.  Do we ever need to call them from Rust space?

Also, this is definitely pre-existing, but I don't actually think that this class needs `__getstate__` or `__setstate__` methods at all - `__getnewargs__` is already doing all the stateful work.  It shouldn't have much of an effect, since `pickle` will only be pickling the args one time each anyway, but it's still just a little more inefficient than we need.

(If you want to remove them in a separate PR so we can see if there's any `copy` improvement in `QuantumCircuit`, I'm also cool with it.)

crates/circuit/src/dag_node.rs
```diff
@@ -0,0 +1,329 @@
...
+// copyright notice, and modified files need to carry a notice indicating
+// that they have been altered from the originals.
+
+use crate::circuit_instruction::CircuitInstruction;
+use pyo3::prelude::*;
+use pyo3::sync::GILOnceCell;
+use pyo3::types::{PyDict, PyList, PySequence, PyString, PyTuple};
+use pyo3::{intern, PyObject, PyResult};
+
+static SEMANTIC_EQ_PYTHON: GILOnceCell<PyResult<PyObject>> = GILOnceCell::new();
```
jakelishman: This kind of violates the idea of Python's virtual dispatch for the method - the method exists on a module, but it's no longer possible to meaningfully monkey-patch it in a testing context or anything like that.

That's probably not so important, but is there any meaningful performance hit from just re-importing it from a suitable module each time it's used?

crates/circuit/src/dag_node.rs
```diff
@@ -0,0 +1,329 @@
...
+
+/// Parent class for DAGOpNode, DAGInNode, and DAGOutNode.
+#[pyclass(module = "qiskit._accelerate.circuit", subclass)]
+#[derive(Clone, Debug)]
+pub struct DAGNode {
+    #[pyo3(get, set)]
+    pub _node_id: isize,
+}
```
jakelishman, line 24: Can we type `_node_id` as the `petgraph::NodeIndex` that it represents, so it's well-typed from within Rust space?  I imagine we can't define the PyO3 conversion traits on it because of the coherence rules, but we could manually implement the Python-space getters and setters?

crates/circuit/src/dag_node.rs
```diff
@@ -0,0 +1,329 @@
...
+        DAGNode { _node_id: nid }
+    }
+
+    fn __getstate__(&self) -> isize {
+        self._node_id
+    }
+
+    fn __setstate__(&mut self, nid: isize) {
+        self._node_id = nid;
+    }
```
jakelishman, line 40: Rather than setting `nid = -1`, can we make this `Option<T>` if we _need_ it to be undefined, with a view to removing the `Option` once this moves properly to Rust space?

We also don't need `__getstate__` and `__setstate__` here if we define `__getnewargs__` to return `(self._node_id,)`, which will make copy/pickle a bit more efficient.

crates/circuit/src/dag_node.rs
```diff
@@ -0,0 +1,329 @@
...
+    fn semantic_eq<'py>(py: Python<'py>, args: &Bound<PyTuple>) -> PyResult<Bound<'py, PyAny>> {
+        let semantic_eq_fn = SEMANTIC_EQ_PYTHON
+            .get_or_init(py, || {
+                let module = PyModule::import_bound(py, "qiskit.dagcircuit.dagnode")?;
+                Ok(module.getattr("_semantic_eq")?.unbind())
+            })
+            .as_ref()
+            .map_err(|e| e.clone_ref(py))?;
+        semantic_eq_fn.bind(py).call1(args)
+    }
```
jakelishman: Possibly could we make `semantic_eq` a `#[classattr]` that takes only `py` and returns `semantic_eq_fn`? That way we don't need to apply a docstring or keep the args in sync - we'd just be attaching the outer function as a `staticmethod`, rather than nesting Python calls.

crates/circuit/src/dag_node.rs
```diff
@@ -0,0 +1,329 @@
...
+}
+
+/// Object to represent an Instruction at a node in the DAGCircuit.
+#[pyclass(module = "qiskit._accelerate.circuit", extends=DAGNode)]
+pub struct DAGOpNode {
+    pub instruction: CircuitInstruction,
+    #[pyo3(get)]
+    pub sort_key: PyObject,
+}
```
jakelishman, line 65: What kind of interface do you have in mind for the future when these are true Rust-space classes and only accessed from within Rust?  We might be needing to access `_node_id` all the time, which is quite inconvenient to do from the Python-space forms of this.

(I'm assuming that this form of `DAGNode` doesn't really survive into Qiskit 2.0.)

crates/circuit/src/dag_node.rs
```diff
@@ -0,0 +1,329 @@
...
+            DAGNode { _node_id: -1 },
+        ))
+    }
+
+    fn __getstate__(slf: PyRef<Self>, py: Python) -> PyObject {
+        (
+            (slf.instruction.__getstate__(py), &slf.sort_key),
+            (slf.as_ref()._node_id),
+        )
+            .into_py(py)
```
jakelishman: How come we're needing to call `__getstate__` manually here?  `pickle` will recurse safely by default, and if we do this manually, we stymie `copy.copy`'s shallow copies.

crates/circuit/src/dag_node.rs
```diff
@@ -0,0 +1,329 @@
...
+
+    /// Returns a representation of the DAGOpNode
+    fn __repr__(&self, py: Python) -> PyResult<String> {
+        Ok(format!(
+            "DAGOpNode(op={}, qargs={}, cargs={})",
+            self.instruction.operation.bind(py).str()?,
+            self.instruction.qubits.bind(py).str()?,
+            self.instruction.clbits.bind(py).str()?
+        ))
+    }
```
jakelishman: It's possibly marginally better to defer to Python-space formatting here if we can, especially if there are potentially any strings hiding in inner objects.

jakelishman (commented):
Overall this looks good to me, thanks for doing it!  I mostly had comments about our pickle handling through these cases - I think there's a fair amount of places where we could entirely elide the `__getstate__` and `__setstate__` calls and do it all through `__getnewargs__` (and some places where we're duplicating things already).

The other _super minor_ point is whether you think it's worth deferring to Python-space formatting in `__repr__` methods.  It's not as fast, but if anything is bottlenecked on `__repr__` we should probably make sure it _isn't_, and using the Python-space formatting means everything will be consistent no matter how deep or shallow in the object hierarchy a `__repr__` is called.

crates/circuit/src/dag_node.rs
kevinhartman: Are you suggesting that `DAGNode` store an `Option<petgraph::NodeIndex>` instead of an `isize`?

The reason I didn't attempt to do this for now is because technically `-1` as the default is part of the public API for `DAGNode` and all subclasses. What'd be your preference here?

crates/circuit/src/dag_node.rs
kevinhartman: `DAGNode` and all of its subclasses provide an object-oriented model of working with a DAG, and this is inherently at odds with the design of `petgraph` (and largely Rust, IMO).

I'd think that in the future, we'd want to migrate to working with node IDs only for any graph manipulations, and require Rust callers to look up node weights given their IDs. So, we'd remove `_node_id` from the `DAGNode` struct altogether.

crates/circuit/src/dag_node.rs
kevinhartman: We talked about this briefly offline, but I'd accidentally used `str` here instead of `repr`.

Resolved in 041f3419b06b6f3a8c1a9cb0cc1db0021d357d7c

crates/circuit/src/dag_node.rs
kevinhartman: Resolved in 041f3419b06b6f3a8c1a9cb0cc1db0021d357d7c, where I've opted to instead implement `__reduce__`.

crates/circuit/src/dag_node.rs
kevinhartman: To avoid the import and indirection, I pulled all of this out of Rust and instead just assign the static function as a class attribute of `DAGNode` within `dagnode.py`.

041f3419b06b6f3a8c1a9cb0cc1db0021d357d7c

Any objections to that approach?

crates/circuit/src/dag_node.rs
kevinhartman: This specific concern at least should be resolved with 041f3419b06b6f3a8c1a9cb0cc1db0021d357d7c.

crates/circuit/src/dag_node.rs
jakelishman: My main concern was to avoid encoding the `isize` into Rust space, since Rust has much better ways to represent that in safe typing.  That said, longer term, I also feel like we won't want to be storing the `_node_id` inside the weights of our graph object at all, because the graph weights won't be accessible from Python space without first getting unpacked and wrapped into Python objects from Rust space, at which point we could manually push a valid `_node_id` in.

So really, I guess that means I don't care too much about `DAGNode` here, because it oughtn't to stay our internal Rust data anyway.

jakelishman: When we're at the point of doing that, we can make this a proper `NodeIndex` or just `u32` for exposure to Python space, since Python space won't be constructing these things manually anyway (with suitable deprecation warnings ahead of 2.0).

crates/circuit/src/dag_node.rs
jakelishman: Nah, that seems fine to me for now.  One day long in the future that'll want to be a Rust-space method, but for now that seems like a good way of doing things.

crates/circuit/src/dag_node.rs
kevinhartman: It sounds like we don't mind too much if this stays an `isize` since we'll remove it completely in the future, so I'll mark this resolved.

crates/circuit/src/circuit_instruction.rs
kevinhartman: Addressed in 84835e3ba3966e362e568749eabcbf172219d778.

crates/circuit/src/dag_node.rs
```diff
@@ -0,0 +1,283 @@
...
+        DAGNode { _node_id: nid }
+    }
+
+    fn __getstate__(&self) -> isize {
+        self._node_id
+    }
+
+    fn __setstate__(&mut self, nid: isize) {
+        self._node_id = nid;
+    }
```
jakelishman, line 40: Strictly these two could also be replaced with a `__getnewargs__` as well, but it's unimportant - I wouldn't be surprised if they're never called at all, since I don't think we ever construct bare `DAGNode`.

jakelishman (approved):
This looks fine to merge to me, thanks for the changes.

Just for visibility for anyone else looking: the three-tuple return from `__reduce__` here where the callable is the `Self` type object can also be written as a `__getnewargs__`/`__getstate__`/`__setstate__` triple, but the current way also works totally fine.

PR #11354, opened 2023-11-30
Run logical optimization passes in init stage

mtreinish:
### Summary

This commit adds Optimize1qGatesDecomposition and InverseCancellation to the init stage for optimization levels 1, 2, and 3. The optimizations these passes run are completely valid on abstract circuits and do not require target qubits to be selected to run
(`Optimize1qGatesDecomposition` is noise aware and will perform target aware optimizations if one is provided, but if no target information is reported it will operate correctly for any circuit)

### Details and comments

PR #12384, opened 2024-05-09
Add RemoveIdentityEquivalent transpiler pass

### LINKED ISSUE #8654: Transpiler pass to remove gates with negligible effects
### What should we add?

This pass would be very useful to ensure that negligible gates are not needlessly applied. In Cirq, whether a gate is negligible is determined by its implementation of the trace distance bound protocol: see https://github.com/quantumlib/Cirq/blob/3e3d6f7d9368badeafd0f2ca3ee0929ab0fdaa25/cirq-core/cirq/transformers/drop_negligible_operations.py. I'm not sure yet what would be the best way to implement this in Terra.

### LINKED ISSUE #13344: `qiskit.transpile` with `optimization_level=3` does not optimize away an identity operation like `cp` with phase 0
### Environment

- **Qiskit version**: 1.2.4
- **Python version**: 3.11.3
- **Operating system**: Ubuntu Linux

### What is happening?

Why does qiskit.transpile() function with optimization level 3 not remove a cp (controlled phase) gate when its parameter is 0? Since a cp(0) gate is equivalent to an identity operation, it should be optimized away.

```python
import qiskit
import qiskit_aer
circuit = qiskit.QuantumCircuit(2)
circuit.cp(0, 1, 0)
circuit = qiskit.transpile(circuit, qiskit_aer.AerSimulator(), optimization_level=3)
circuit.data
```
Output: 
```
[CircuitInstruction(operation=Instruction(name='cp', num_qubits=2, num_clbits=0, params=[0.0]), qubits=(Qubit(QuantumRegister(2, 'q'), 1), Qubit(QuantumRegister(2, 'q'), 0)), clbits=())]
```

### What should happen?

The gate should be removed.

### Any suggestions?

_No response_

kevinsung:
### Summary

This adds a new transpiler pass `RemoveIdentityEquivalent`, which is used to remove any `Gate` in a circuit that is
equivalent to identity and doesn't have an impact on the circuit. This pass works for any gate that defines a matrix and only those that have numeric parameters (a symbolic parameter precludes having a matrix). It computes the average gate fidelity of the the operation for an identity target and if it's below the the specified tolerance the gate will be removed.

### Details and comments

Fixes #8654

Co-authored-by: Matthew Treinish <mtreinish@kortar.org>
TODO:

- [x] Expand test coverage


qiskit/transpiler/passes/optimization/drop_negligible.py
```diff
@@ -0,0 +1,70 @@
...
+            atol: Absolute numerical tolerance for determining whether a gate's effect
+                is negligible.
+        """
+        self.atol = atol
+        super().__init__()
+
+    def run(self, dag: DAGCircuit) -> DAGCircuit:
+        for node in dag.op_nodes():
+            if not isinstance(node.op, DROP_NEGLIGIBLE_GATE_CLASSES):
+                continue
```
mtreinish: Since you're using a fixed set of gates in this PR one thing you should do here is before looping is check if the names of the gate classes intersect with `dag.count_ops()`. That'll be much faster in the negative case.

qiskit/transpiler/passes/optimization/drop_negligible.py
```diff
@@ -0,0 +1,70 @@
...
+            if not all(isinstance(param, (int, float, complex)) for param in node.op.params):
+                continue
+            if np.allclose(node.op.params, 0, atol=self.atol):
+                dag.remove_op_node(node)
```
mtreinish: I'm wondering if we want to generalize this for any `Gate` object, we could do something similar to this but instead compare `node.op.to_matrix()` against an identity matrix with the tolerance. 

That being said doing this will obviously be slower though. If we're concerned about runtime performance I'd probably suggest moving this away from using `np.allclose()` and just do something like:

```suggestion
            if all(math.isclose(x, 0, atol=self.atol) for x in node.op.params):
                dag.remove_op_node(node)
```

because I expect the numpy overhead of converting params to an array and all the checking involved with that will be slower that just checking each value in params manually, especially as in general there won't be very many parameters for a gate.

test/python/transpiler/test_drop_negligible.py
```diff
@@ -0,0 +1,112 @@
...
+        circuit.append(RYYGate(1e-5), [a, b])
+        circuit.append(RYYGate(1e-8), [a, b])
+        circuit.append(RZZGate(1e-5), [a, b])
+        circuit.append(RZZGate(1e-8), [a, b])
+        circuit.append(XXPlusYYGate(1e-5, 1e-8), [a, b])
+        circuit.append(XXPlusYYGate(1e-8, 1e-8), [a, b])
+        circuit.append(XXMinusYYGate(1e-5, 1e-8), [a, b])
+        circuit.append(XXMinusYYGate(1e-8, 1e-8), [a, b])
+        pass_manager = PassManager([DropNegligible()])
+        transpiled = pass_manager.run(circuit)
```
mtreinish:
```suggestion
        transpiled = DropNegligible()(circuit)
```

mtreinish (commented):
I think having a pass like this in the library makes a lot of sense. We don't want to run it by default, but if you know it makes sense in your use case then having this available is useful. We could even look at adding it to the end of the init stage in the preset pass managers and tying the `atol` value to the `approximation_degree` value somehow.

levbishop:
I like the general idea to drop negligible gates but think this implementation is not optimal. `TwoQubitBasisDecomposer` will already drop all negligible 2q gates, not just the ones in a hardcoded list; will do it with a controlled approximation (the average-gate-fidelity of the approximation, rather than just a parameter value); and will optimize away 2q gates while keeping an optimal 1q approximation. Because of how the default passmanagers work, this `TwoQubitBasisDecomposer` only runs after routing, losing most of the benefit of dropping negligible 2q gates: I've been meaning for a while to write up some notes on a better default passmanager flow - I'll see if I can get something down for discussion over the weekend.

mtreinish:
The concern I have with just relying on the `TwoQubitBasisDecomposer` for this is the same issue we had with trying to do something similar with the 1q case in #11354 . As soon as we added `Optimize1qGatesDecomposition` to the init stage there users testing against main (mainly qiskit-experiments) reported issues that we were breaking transpilation for targets with a discrete basis. So we very quickly reverted this in #11360. In general I really like the idea of doing the peephole optimization early in the pipeline but we just have to be careful around how we do it. Regardless I look forward to seeing your proposal on an improved pipeline.

kevinsung:
`TwoQubitBasisDecomposer` is not appropriate for some of the uses cases I have in mind. As an example, ffsim defines a high-level gate [DiagonalCoulombJW](https://qiskit-community.github.io/ffsim/api/ffsim.qiskit.html#ffsim.qiskit.DiagCoulombEvolutionJW). This gate is defined by a matrix, and it has a [gate decomposition](https://qiskit-community.github.io/ffsim/explanations/qiskit-gate-decompositions.html#Diagonal-Coulomb-evolution) into CPhaseGates based on that matrix. Sometimes you know that some entries are close to zero, so you just want to get rid of them. Or, you might want to get rid of some gates based on a threshold just to make it easier to map to hardware. This workflow would not use `TwoQubitBasisDecomposer` and you might want to do this independently from the standard "preset staged pass manager" transpilation workflow.

qiskit/transpiler/passes/optimization/drop_negligible.py
kevinsung: Doesn't `dag.count_ops` also need to iterate through the nodes of the dag?

test/python/transpiler/test_drop_negligible.py
kevinsung: Done.

qiskit/transpiler/passes/optimization/drop_negligible.py
kevinsung: Good idea to use the built-ins instead of Numpy, done.

I don't think we should try to construct a gate's matrix, due to the performance overhead, and also because it's not feasible for all gates (cost is exponential in the number of qubits). However, I do think we should let the user pass additional gate classes to check.

qiskit/transpiler/passes/optimization/drop_negligible.py
mtreinish: `DAGCircuit` maintains a dictionary of counts that it updates whenever an op is added or removed from the circuit. This is explicitly to support cases like this to do constant time checks for the instructions presence in the circuit. The only exceptions are if you set `recurse=True` (which is the default) and there are any control flow ops in the circuit.

levbishop:
```python
try:
    decomp = TwoQubitWeylDecomposition(node.op, fidelity=fid,  _specialization=_specializations.IdEquiv)
except QiskitError:
    pass
else:
    <replace dag node with 1q gates from decomp>
```

qiskit/transpiler/passes/optimization/drop_negligible.py
kevinsung: I've added an argument for passing additional gate types.

qiskit/transpiler/passes/optimization/drop_negligible.py
kevinsung: `dag.count_ops()` stores strings, but I'm using the gate class itself to check the gate type. Any advice? Should I also just use strings? That doesn't seem robust because any gate can have any arbitrary name. But I guess there's many places in Qiskit which assume to know the type of a gate based on its name.

kevinsung:
> ```python
> try:
>     decomp = TwoQubitWeylDecomposition(node.op, fidelity=fid,  _specialization=_specializations.IdEquiv)
> except QiskitError:
>     pass
> else:
>     <replace dag node with 1q gates from decomp>
> ```

Doing linear algebra is too expensive when the aim is just to drop gates with small angles.

mtreinish:
I've reworked this PR to compute the identity equivalence numerically for all gates with a matrix defined and to operate purely in Rust for performance. While these gates would be likely caught in later stages between the synthesis peephole optimization and/or `Split2QUnitaries` this should be a lot faster to compute and will speed up those later stages which are doing more investigation. I think we'll want to add this pass to the `init` stage and the `optimization` stage (inside the loop) for levels 2 and 3 in a follow up PR. This should give us pretty good coverage for removing gates that don't have a measurable impact on the circuit.

The only todo here is to expand the test coverage which I need to expand to cover all the combinations of inputs in the pass now.

crates/accelerate/src/remove_identity_equiv.rs
```diff
@@ -0,0 +1,113 @@
...
+    for op_node in dag.op_nodes(false) {
+        let inst = dag.dag()[op_node].unwrap_operation();
+        match inst.op.view() {
+            OperationRef::Standard(gate) => {
+                if let Some(matrix) = gate.matrix(inst.params_view()) {
+                    let error = get_error_cutoff(inst);
+                    let dim = matrix.shape()[0] as f64;
+                    let trace: Complex64 = matrix.diag().iter().sum();
+                    let f_pro = (trace / dim).abs().powi(2);
+                    let gate_fidelity = (dim * f_pro + 1.) / (dim + 1.);
```
Cryoris: This would save 1 flop (I don't remember the HPC courses enough though to remember if mult and div have the same cost) 😄 
```suggestion
                    let trace_sq = trace.abs().powi(2);
                    let gate_fidelity = (trace_sq / dim + 1.) / (dim + 1.);
```

crates/accelerate/src/remove_identity_equiv.rs
```diff
@@ -0,0 +1,113 @@
...
+                }
+                None => f64::EPSILON,
+            },
+        }
+    };
+
+    for op_node in dag.op_nodes(false) {
+        let inst = dag.dag()[op_node].unwrap_operation();
+        match inst.op.view() {
+            OperationRef::Standard(gate) => {
```
Cryoris, line 80: How about adding a fast-path for Pauli rotations? We can compute their traces as 
```python
Tr(exp(-i th/2 P)) = Tr(B exp(-i th/2 Z) B^\dagger) = Tr(exp(-i th/2 Z)) = exp(-i th/2) + exp(i th / 2) = 2 cos(th/2)
```
For 2-qubit Pauli rotations we'd have ZZ instead and 4 terms, so that should be `4 cos(th/2)`.

crates/accelerate/src/remove_identity_equiv.rs
```diff
@@ -0,0 +1,121 @@
...
+                                .iter()
+                                .map(|x| PhysicalQubit::new(x.0))
+                                .collect();
+                            let error_rate = target.get_error(inst.op.name(), qargs.as_slice());
+                            match error_rate {
+                                Some(err) => err * degree,
+                                None => degree,
+                            }
+                        }
+                        None => degree,
```
Cryoris: I'm a bit confused here: shouldn't the error cutoff be inversely proportional to the approximation degree? E.g. `approximation_degree=1` implies `cutoff=f64::EPSILON`, as expected. But if I'm reading this correctly `approximation_degree=0.99` and no target would give `cutoff=0.99`? 🤔 

Shouldn't this be something like `cutoff = max(f64::EPSILON, 1 - approximation_degree)` in the case of no target?

qiskit/transpiler/passes/optimization/remove_identity_equiv.py
```diff
@@ -0,0 +1,56 @@
...
+from qiskit.transpiler.target import Target
+from qiskit.transpiler.basepasses import TransformationPass
+from qiskit._accelerate.remove_identity_equiv import remove_identity_equiv
+
+
+class RemoveIdentityEquivalent(TransformationPass):
+    """Remove gates with negligible effects.
+
+    Removes gates whose effect is close to an identity operation, up to the specified
+    tolerance.
```
Cryoris: This should mention that `GlobalPhaseGate` and other 0-q gates are ignored 🙂

qiskit/transpiler/passes/optimization/remove_identity_equiv.py
```diff
@@ -0,0 +1,56 @@
...
+    """
```
Cryoris, line 41: It would be nice to be concrete about when a gate is cutoff, e.g. something along the lines of 
```suggestion
    For a cutoff fidelity :math:`f`, this pass removes gates whose average 
    gate fidelity with respect to the identity is below :math:`f`. Concretely, 
    a gate :math:`G` is removed if :math:`\bar F < f` where 
    
    .. math::

        \bar{F} = \frac{1 + F_{\text{process}}{1 + d} 
        
        F_{\text{process}} = \frac{|\mathrm{Tr}(G)|^2}{d^2}
        
    where :math:`d = 2^n` is the dimension of the gate for :math:`n` qubits.
    """
```

qiskit/transpiler/passes/optimization/remove_identity_equiv.py
```diff
@@ -0,0 +1,56 @@
...
+    def __init__(
+        self, *, approximation_degree: float | None = 1.0, target: None | Target = None
+    ) -> None:
+        """Initialize the transpiler pass.
+
+        Args:
+            approximation_degree: The degree to approximate for the equivalence check. This can be a
+                floating point value between 0 and 1, or ``None``. If the value is 1 this does not
+                approximate above floating point precision. For a value < 1 this is used as a scaling
+                factor for the target fidelity. If the value is ``None`` this approximates up to the
```
Cryoris: In case we use the terminology "cutoff fidelity" in the docstring above it would be nice to use it here too
```suggestion
                factor for the cutoff fidelity. If the value is ``None`` this approximates up to the
```

Cryoris (commented):
Some minor comments, otherwise LGTM 🙂

crates/accelerate/src/remove_identity_equiv.rs
mtreinish: Ah yeah, good call this logic is incorrect. I'll update these

crates/accelerate/src/remove_identity_equiv.rs
mtreinish: In hardware in most cases division is more expensive. For example, on my workstation's CPU architecture f64 division is 13 cycles while multiplication is 5 cycles:

https://www.anandtech.com/show/16214/amd-zen-3-ryzen-deep-dive-review-5950x-5900x-5800x-and-5700x-tested/6

but the compiler knows this and will try to optimize away division or use different instructions for the same output if it can so it's not always clear cut. The only way to know for sure is to try it both ways and benchmark it.

jakelishman: Compilers often can't optimise away floating-point divisions because of rounding differences. `a / b` isn't the same as `a * (1. / b)` in general (though it is if `b` is a power of two that won't overflow on reciprocation)  - for example, `3.0 / 5.0` isn't the same as `(1.0/5.0) * 3.0` due to rounding - and most transformations suffer similar problems to intermediate rounding.

It's what `-ffast-math` is for in gcc and clang, but we don't run with the equivalent of that in Rust (nor should we).

test/python/transpiler/test_drop_negligible.py
jakelishman, line 1: This test file and class name in it still use the old name - can we make sure they're the same so we don't lose the test file in the future?

crates/accelerate/src/remove_identity_equiv.rs
mtreinish: Done in: https://github.com/Qiskit/qiskit/pull/12384/commits/835cd35575bad8d635f95e37d4ca49f5dc288641

test/python/transpiler/test_drop_negligible.py
mtreinish: Done in: https://github.com/Qiskit/qiskit/pull/12384/commits/1568105bc87b9e8ab02a220cf80daf965d3013d1

qiskit/transpiler/passes/optimization/remove_identity_equiv.py
mtreinish: Done in: https://github.com/Qiskit/qiskit/pull/12384/commits/4e3a76be3b19bab199078fddc551ac1e7adef355

crates/accelerate/src/remove_identity_equiv.rs
mtreinish: Done in: https://github.com/Qiskit/qiskit/pull/12384/commits/09b7c4769eb18698a65299432b686aee8985c164

Cryoris (approved):

PR #12385, opened 2024-05-10
Fix {Pauli,SparsePauliOp}.apply_layout to raise an error with negative or duplicate indices

t-imamichi:
### Summary

{Pauli,SparsePauliOp}.apply_layout currently allow negative or duplicate indices.
The behavior is not always comprehensive.

1. If we give duplicate indices, some coefficients or phases may change. 
2. If we give negative indices that are out of bounds, it raises IndexError while positive indices raises QiskitError.

This PR raises QiskitError with duplicate indices or negative indices (even within bounds for simplicity).
I think duplicate indices are not valid as a layout. But, if we need to support the pattern, we might need to update the code not to change the coefficients.

I noticed this situation when I review https://github.com/Qiskit/qiskit/pull/12221#issuecomment-2101839871

```python
from qiskit.quantum_info import Pauli, SparsePauliOp

op = Pauli("XY")
op2 = SparsePauliOp("XY")

print("duplicate indices")
try:
    print(op.apply_layout([0, 0]))
except Exception as ex:
    print(ex)
try:
    print(op2.apply_layout([0, 0]))
except Exception as ex:
    print(ex)

print("\nnegative indices not out of bounds")
try:
    print(op.apply_layout([-1, 0]))
except Exception as ex:
    print(ex)
try:
    print(op2.apply_layout([-1, 0]))
except Exception as ex:
    print(ex)

print("\nnegative indices out of bounds")
try:
    print(op.apply_layout([-3, 0]))
except Exception as ex:
    print(ex)
try:
    print(op2.apply_layout([-3, 0]))
except Exception as ex:
    print(ex)

print("\npositive indices out of bounds")
try:
    print(op.apply_layout([3, 0]))
except Exception as ex:
    print(ex)
try:
    print(op2.apply_layout([3, 0]))
except Exception as ex:
    print(ex)
```

main
```
duplicate indices
-iIX
SparsePauliOp(['IX'],
              coeffs=[0.-1.j])

negative indices not out of bounds
YX
SparsePauliOp(['YX'],
              coeffs=[1.+0.j])

negative indices out of bounds
index -3 is out of bounds for axis 1 with size 2
index -3 is out of bounds for axis 1 with size 2

positive indices out of bounds
'Provided layout contains indices outside the number of qubits.'
'Provided layout contains indices outside the number of qubits.'
```

this PR
```
duplicate indices
'Provided layout contains duplicate indices.'
'Provided layout contains duplicate indices.'

negative indices not out of bounds
'Provided layout contains indices outside the number of qubits.'
'Provided layout contains indices outside the number of qubits.'

negative indices out of bounds
'Provided layout contains indices outside the number of qubits.'
'Provided layout contains indices outside the number of qubits.'

positive indices out of bounds
'Provided layout contains indices outside the number of qubits.'
'Provided layout contains indices outside the number of qubits.'
```

### DIFF
--- a/releasenotes/notes/fix-apply-layout-duplicate-negative-indices-cf5517921fe52706.yaml
+++ b/releasenotes/notes/fix-apply-layout-duplicate-negative-indices-cf5517921fe52706.yaml
@@ -2,5 +2,5 @@
 fixes:
   - |
     Fixed :meth:`.SparsePauliOp.apply_layout` and :meth:`.Pauli.apply_layout`
-    to raise :exc:`QiskitError` if duplicate indices or negative indices are provided
+    to raise :exc:`.QiskitError` if duplicate indices or negative indices are provided
     as part of a layout.
--- a/releasenotes/notes/fix-apply-layout-duplicate-negative-indices-cf5517921fe52706.yaml
+++ b/releasenotes/notes/fix-apply-layout-duplicate-negative-indices-cf5517921fe52706.yaml
@@ -1,6 +1,6 @@
 ---
 fixes:
   - |
-    Fixed :meth:`~.SparsePauliOp.apply_layout` and :meth:`~.Pauli.apply_layout`
-    to raise ``QiskitError`` if duplicate indices or negative indices are provided
+    Fixed :meth:`.SparsePauliOp.apply_layout` and :meth:`.Pauli.apply_layout`
+    to raise :exc:`QiskitError` if duplicate indices or negative indices are provided
     as part of a layout.
--- a/releasenotes/notes/fix-apply-layout-duplicate-negative-indices-cf5517921fe52706.yaml
+++ b/releasenotes/notes/fix-apply-layout-duplicate-negative-indices-cf5517921fe52706.yaml
@@ -1,0 +1,6 @@
+---
+fixes:
+  - |
+    Fixed :meth:`~.SparsePauliOp.apply_layout` and :meth:`~.Pauli.apply_layout`
+    to raise ``QiskitError`` if duplicate indices or negative indices are provided
+    as part of a layout.
--- a/qiskit/quantum_info/operators/symplectic/pauli.py
+++ b/qiskit/quantum_info/operators/symplectic/pauli.py
@@ -736,8 +736,11 @@ class Pauli(BasePauli):
             n_qubits = num_qubits
         if layout is None:
             layout = list(range(self.num_qubits))
-        elif any(x >= n_qubits for x in layout):
-            raise QiskitError("Provided layout contains indices outside the number of qubits.")
+        else:
+            if any(x < 0 or x >= n_qubits for x in layout):
+                raise QiskitError("Provided layout contains indices outside the number of qubits.")
+            if len(set(layout)) != len(layout):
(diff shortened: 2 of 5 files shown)


jakelishman (commented):
Thanks, this seems like a good idea to me.  Could you add a brief bug-fix release note mentioning the change?

releasenotes/notes/fix-apply-layout-duplicate-negative-indices-cf5517921fe52706.yaml
```diff
@@ -0,0 +1,6 @@
+---
+fixes:
+  - |
+    Fixed :meth:`~.SparsePauliOp.apply_layout` and :meth:`~.Pauli.apply_layout`
+    to raise ``QiskitError`` if duplicate indices or negative indices are provided
```
jakelishman:
```suggestion
    Fixed :meth:`.SparsePauliOp.apply_layout` and :meth:`.Pauli.apply_layout`
    to raise :exc:`QiskitError` if duplicate indices or negative indices are provided
```

The leading `~` strips all but the rightmost part of the path, so `` :meth:`~.SparsePauliOp.apply_layout` `` and `` :meth:`~.Pauli.apply_layout` `` would both appear in the text as `apply_layout()` - dropping the `~` means we'll see the class as well.

jakelishman (dismissed):
Looks good, thanks for the quick change and the fix!

releasenotes/notes/fix-apply-layout-duplicate-negative-indices-cf5517921fe52706.yaml
```diff
@@ -0,0 +1,6 @@
+---
+fixes:
+  - |
+    Fixed :meth:`.SparsePauliOp.apply_layout` and :meth:`.Pauli.apply_layout`
+    to raise :exc:`QiskitError` if duplicate indices or negative indices are provided
```
jakelishman:
```suggestion
    to raise :exc:`.QiskitError` if duplicate indices or negative indices are provided
```

jakelishman (approved):

releasenotes/notes/fix-apply-layout-duplicate-negative-indices-cf5517921fe52706.yaml
t-imamichi: Thank you for the information about `~`. I didn't know the effect and use it always.

jakelishman: It's usually what we want for classes, so we use it a lot - it lets you use more prefix parts in the lookup to disambiguate things, e.g. `` :class`~.circuit.Instruction` `` and `` :class:`~qiskit.pulse.Instruction` `` will both display only `Instruction`, but will refer to different objects (and for those two, it's clear from context which is meant).  It mostly only starts to be confusing when where talking about methods from outside of the class description itself (like here).  Sometimes we do say things like

> The `` :meth:`~.SparsePauliOp.to_matrix` `` and `` :meth:`~.SparsePauliOp.from_operator` `` methods of `` :class:`.SparsePauliOp` `` [...]

and use the `~` with the methods because we're mentioning the class separately.

PR #12392, opened 2024-05-12
Avoid lossing precision when scaling frequencies

### LINKED ISSUE #12359: Some unit tests crash when using symengine
### Environment

- **Qiskit version**: 1.1.0rc1 (but also in previous stable versions)
- **Python version**:  3.12.3
- **Operating system**: Arch Linux

### What is happening?

Some unit tests fail due to symengine crashes. For example, the following five tests from `test_circuit_load_from_qpy.py` crash on my system:
- `test_parameter_expression`
- `test_evolutiongate_param_expr_time`
- `test_parameter_expression_global_phase`
- `test_parameter_vector_element_in_expression`
- `test_symengine_full_path`

From the first test, the exact line when the the test crashes is this:
https://github.com/Qiskit/qiskit/blob/6ff0eb54f88cdbe98275872b8ec9d65f76948bb4/qiskit/qpy/binary_io/value.py#L58

Which produces a `src/tcmalloc.cc:304] Attempt to free invalid pointer 0x6146ecca5b60`.

Here is a small self-contained script that can replicate the crash described above.

```python
from qiskit.circuit.parameterexpression import ParameterExpression


theta = Parameter("theta")
phi = Parameter("phi")
sum_param = theta + phi

sum_param._symbol_expr.__reduce__()
# src/tcmalloc.cc:304] Attempt to free invalid pointer 0x645c8e3eaa30 
# [1]    987079 IOT instruction (core dumped)  python
```

The following running on the same system works:

```python
import symengine

theta = symengine.Symbol("theta")
phi = symengine.Symbol("phi")
sum_param = theta + phi

sum_param.__reduce__()
```

### What should happen?

Ideally, all unit tests would pass in a clean chroot env on Arch Linux.

### Any suggestions?

_No response_

iyanmv:
### Summary
Classes in `pulse_instruction.py` scale frequency values to GHz by multipliying `ParameterExpression` with float `1e9`. This can lead to numerical errors on some systems due to `symengine` "rounding" errors. Instead, this scaling can be done multiplying by integer `10**9`.

### Details and comments

In this unit test
https://github.com/Qiskit/qiskit/blob/235e581b1f76f29add0399989ed47f47a4e98bb8/test/python/qobj/test_pulse_converter.py#L343

The `frequency` string `"f / 1000"` gets converted to `ParameterExpression(1000000.0*f)` after `ParameterExpression(f/1000)` is multiplied by `1e9`. For some unknown reason, when the symbol `f` is later substituted with the value `3.14`, and the `RealDouble` is converted to `float`, an error is introduced that can't be fixed by
https://github.com/Qiskit/qiskit/blob/235e581b1f76f29add0399989ed47f47a4e98bb8/qiskit/pulse/utils.py#L71-L74

This fixes: https://github.com/Qiskit/qiskit/issues/12359#issuecomment-2104426621

Upstream issue: https://github.com/symengine/symengine.py/issues/476

### DIFF
--- a/qiskit/qobj/converters/pulse_instruction.py
+++ b/qiskit/qobj/converters/pulse_instruction.py
@@ -234,7 +234,7 @@ class InstructionToQobjConverter:
             "name": "setf",
             "t0": time_offset + instruction.start_time,
             "ch": instruction.channel.name,
-            "frequency": instruction.frequency / 1e9,
+            "frequency": instruction.frequency / 10**9,
         }
         return self._qobj_model(**command_dict)
 
@@ -257,7 +257,7 @@ class InstructionToQobjConverter:
             "name": "shiftf",
             "t0": time_offset + instruction.start_time,
             "ch": instruction.channel.name,
-            "frequency": instruction.frequency / 1e9,
+            "frequency": instruction.frequency / 10**9,
         }
         return self._qobj_model(**command_dict)
 
@@ -746,7 +746,7 @@ class QobjToInstructionConverter:
         .. note::
 
             We assume frequency value is expressed in string with "GHz".
-            Operand value is thus scaled by a factor of 1e9.
+            Operand value is thus scaled by a factor of 10^9.
 
         Args:
             instruction: SetFrequency qobj instruction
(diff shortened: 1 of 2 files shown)


jakelishman (approved):
Thanks so much for all the work to track this down!  This seems like an easy change to make - in 99% of cases, Python's going to immediately convert the int to a float, but in the case that the pulse is symbolic, it's good not to introduce the floating-point error into the calculation.

I pushed up a release note, but otherwise this is good to go.

1ucian0:

PR #11744, opened 2024-02-07
Fix bad indentation on quantumcircuit.py

arnaucasau:
# Summary
This PR fixes a bad indentation of two code blocks that was causing the page to not render correctly those snippets.

## Screenshots before this PR

**First code block:**

Captura desde 2024-02-07 21-54-51

**Second code block:**
Captura desde 2024-02-07 21-55-08

## Screenshots applying the change
**First code block:**
Captura desde 2024-02-07 21-55-30

**Second code block:**
Captura desde 2024-02-07 21-55-18

PR #12394, opened 2024-05-13
Fix `qiskit.circuit` method header and broken cross-reference

arnaucasau:
Part of https://github.com/Qiskit/documentation/issues/1275

The `__array__` method has a wrong header in the API docs (as we can see in the following screenshot) because Sphinx is not able to set the correct id in the generated HTML for methods starting with double underscore. Instead, Sphinx truncates the name to `array__`.

Captura desde 2024-05-13 14-31-08

This PR changes the method to be instead `builtins.__array__`, and it also fixes a broken reference for `Instruction.params`. The cross-reference was broken due to the` .. currentmodule:: None` directive.

This is the result of Sphinx after this PR:

Captura desde 2024-05-13 14-27-06

The API docs will have the correct header because we only use the part after the last dot in the id.

The PR will need backport to `stable/1.1`

### DIFF
--- a/qiskit/circuit/__init__.py
+++ b/qiskit/circuit/__init__.py
@@ -832,10 +832,10 @@ Subclasses of :class:`Gate` will also likely wish to override `the Numpy array-p
 method <https://numpy.org/devdocs/user/basics.interoperability.html#the-array-method>`__,
 ``__array__``.  This is used by :meth:`Gate.to_matrix`, and has the signature:
 
-.. currentmodule:: None
+.. currentmodule:: builtins
 .. py:method:: __array__(dtype=None, copy=None)
 
-    Return a Numpy array representing the gate.  This can use the gate's :attr:`~Instruction.params`
+    Return a Numpy array representing the gate.  This can use the gate's :attr:`~qiskit.circuit.Instruction.params`
     field, and may assume that these are numeric values (assuming the subclass expects that) and not
     :ref:`compile-time parameters <circuit-compile-time-parameters>`.
 
--- a/qiskit/circuit/__init__.py
+++ b/qiskit/circuit/__init__.py
@@ -832,8 +832,8 @@ Subclasses of :class:`Gate` will also likely wish to override `the Numpy array-p
 method <https://numpy.org/devdocs/user/basics.interoperability.html#the-array-method>`__,
 ``__array__``.  This is used by :meth:`Gate.to_matrix`, and has the signature:
 
-.. currentmodule:: object
-.. py:method:: __array__(dtype=None, copy=None)
+.. currentmodule:: None
+.. py:method:: object.__array__(dtype=None, copy=None)
 
     Return a Numpy array representing the gate. This can use the gate's
     :attr:`~qiskit.circuit.Instruction.params` field, and may assume that these are numeric
--- a/qiskit/circuit/__init__.py
+++ b/qiskit/circuit/__init__.py
@@ -835,8 +835,9 @@ method <https://numpy.org/devdocs/user/basics.interoperability.html#the-array-me
 .. currentmodule:: object
 .. py:method:: __array__(dtype=None, copy=None)
 
-    Return a Numpy array representing the gate.  This can use the gate's :attr:`~qiskit.circuit.Instruction.params`
-    field, and may assume that these are numeric values (assuming the subclass expects that) and not
+    Return a Numpy array representing the gate. This can use the gate's
+    :attr:`~qiskit.circuit.Instruction.params` field, and may assume that these are numeric
+    values (assuming the subclass expects that) and not
     :ref:`compile-time parameters <circuit-compile-time-parameters>`.
 
     For greatest efficiency, the returned array should default to a dtype of :class:`complex`.
--- a/qiskit/circuit/__init__.py
+++ b/qiskit/circuit/__init__.py
@@ -832,7 +832,7 @@ Subclasses of :class:`Gate` will also likely wish to override `the Numpy array-p
 method <https://numpy.org/devdocs/user/basics.interoperability.html#the-array-method>`__,
 ``__array__``.  This is used by :meth:`Gate.to_matrix`, and has the signature:
 
-.. currentmodule:: builtins
+.. currentmodule:: object
 .. py:method:: __array__(dtype=None, copy=None)
 
     Return a Numpy array representing the gate.  This can use the gate's :attr:`~qiskit.circuit.Instruction.params`


qiskit/circuit/__init__.py
```diff
@@ -832,10 +832,10 @@
 method <https://numpy.org/devdocs/user/basics.interoperability.html#the-array-method>`__,
 ``__array__``.  This is used by :meth:`Gate.to_matrix`, and has the signature:
 
-.. currentmodule:: None
+.. currentmodule:: builtins
```
jakelishman: Why is this changed?  `__array__` is not a Python builtin and isn't something that's part of the `builtins` module - it doesn't have a module, which is spelled `currentmodule:: None` in Sphinx.

Sphinx might be generating a modified `id` to avoid HTML problems, but if that's causing problems, it's on the scripts that are parsing the output of Sphinx to fix that, not to change the semantics of the documentation.  The Sphinx-generated HTML entries do correctly name this method `__array__`, and it appears correctly in `objects.inv`.

If it's _really_ impossible to fix the conversion scripts to use Sphinx data (such as using `objects.inv` as a source of truth) rather than trying to scrape it from escaped HTML, then potentially the cleaner workaround is to document this as being a potential method on `object` (which _is_ in `builtins`) or somewhat lying and documenting it as part of `Instruction`.

jakelishman (commented):
Thanks for the catching the `Instruction.params` thing.

The other problem here is not Sphinx, it's the scripts that are scraping the HTML output.  I know it's not super popular, but these sorts of problems will keep cropping up and needing work for as long as the API generation scripts keep scraping the HTML rather than using Sphinx's data directly, such as by writing a Sphinx `Builder` that directly generates the target MDX.

Eric-Arellano:
> then potentially the cleaner workaround is to document this as being a potential method on object (which is in builtins) 

Ah, good suggestion. @arnaucasau, pardon me leading you astray with `builtins` rather than `object`.

> The other problem here is not Sphinx, it's the scripts that are scraping the HTML output.

Imo the problem is Sphinx. Sphinx is generating the HTML `id` as `=array__` rather than `__array__`.  Sphinx is specifically doing this when the currentmodule is set to `None`. Compare it to https://docs.quantum.ibm.com/api/qiskit/0.19/qiskit.transpiler.Layout#methods, where the HTML ID is correctly set by Sphinx as `qiskit.transpiler.Layout.__getitem__`. If we set `currentmodule` here to another value like `object`, then the ID gets set to `object.__array__` correctly.

>  I know it's not super popular, but these sorts of problems will keep cropping up and needing work for as long as the API generation scripts keep scraping the HTML rather than using Sphinx's data directly, such as by writing a Sphinx Builder that directly generates the target MDX.

A non-trivial portion of the rendering bugs have been due to RST issues, rather than the conversion script: https://github.com/Qiskit/documentation/issues/1200, https://github.com/Qiskit/documentation/issues/1199, https://github.com/Qiskit/documentation/issues/1275, https://github.com/Qiskit/qiskit/pull/11744, https://github.com/Qiskit/qiskit/pull/12162. Those would still be a problem.

The script-caused issues have been due to not knowing about generic edge cases—like aliases (`Sampler` vs `SamplerV1`). Now that those edge cases are fixed and tested, we haven't been seeing them regress.

It's always worth keeping that rewrite to a Sphinx builder in our back-pocket, but I don't think that's a good investment at the time.

qiskit/circuit/__init__.py
arnaucasau: I have been testing the use of `objects.inv` to determine the headers instead of reading the HTML and it would be more tricky than it seems. For example, for some classes, we have duplicated IDs that reference a rust 'class' and a python class:

```
py:class
    qiskit._accelerate.qasm3.CustomGate      apidoc/qasm3.html#qiskit.qasm3.CustomGate
    qiskit.qasm3.CustomGate                  apidoc/qasm3.html#qiskit.qasm3.CustomGate
```

I'm not sure if we can always differentiate between them, so maybe it's better to go with the workaround as `__array__` is the only instance of this problem that we have found. In [15f9842](https://github.com/Qiskit/qiskit/pull/12394/commits/15f9842f17276d59a48f78d9dc3b41dd6a913d4b), I changed `builtins` to `object` as you suggested, let me know if you think it's okay :+1:

jakelishman:
> Sphinx is generating the HTML id as `=array__` rather than `__array__`

Sure, but the `id` HTML attribute is meant to be purely opaque and not carry semantics - it's not part of Sphinx's data output other than a programmatic anchor point that's a little easier for humans to read than a random string. [It's typically recommended that an `id` tag start with an ASCII letter](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/id), which is probably why Sphinx is dropping the underscores here - it's a type of escaping.

> A non-trivial portion of the rendering bugs have been due to RST issues

Oh 100% - I'm not questioning that at all. I didn't mean to suggest that every problem on the docs is because of conversion script.  Actually, mostly I'm not as concerned about the final rendering in the long term; rendering is the "easy" bit, in the sense that you can do almost anything with CSS if your structure is good, which you guys have been showing by making big improvements to the looks of the pages. It's the semantics of the generated page I'm concerned about, and that's where I think scraping the HTML is a scary avenue for errors and fragile to updates in Sphinx.  At some point, I think a HTML scraping program is already doing as much work as a Sphinx builder on output (since you've got to convert your in-memory representation to MDX), it's just got an _awful_ lot harder job on input, because it doesn't get given easily structured data, and has to guess what the original inputs were from that.

---

Anyway, it's not a practical change right now, and I'm sorry for being overly grumpy about the proposed PR.

qiskit/circuit/__init__.py
jakelishman: Fwiw, the `objects.inv` format also includes a "priority" field, which might be what distinguishes the canonical reference?  I'm not super sure.  For example (the `printf` is just appending the magic numbers for zlib compression, and the `tail` cuts out the Sphinx comment header):
```bash
{ printf "\x1f\x8b\x08\x00\x00\x00\x00\x00" ; tail -n+5 docs/_build/html/objects.inv ; } | gzip -d | grep 'CustomGate'
```
```text
qiskit._accelerate.qasm3.CustomGate py:class -1 apidoc/qasm3.html#qiskit.qasm3.CustomGate -
qiskit.qasm3.CustomGate py:class 1 apidoc/qasm3.html#$ -
```
it's the -1 / 1 that I'm talking about.  But that might be something different too.

Anyway - thanks for the change.  What I meant was ever so slightly different: leaving `.. currentmodule:: None`, but documenting the method as `object.__array__`.  I've updated the PR to be like that, and I checked that it appears like that in the HTML's `id` tag.

jakelishman: The generated HTML is this, for reference:
```html
<dl class="py method">
<dt class="sig sig-object py" id="object.__array__">
<span class="sig-prename descclassname"><span class="pre">object.</span></span><span class="sig-name descname"><span class="pre">__array__</span></span><span class="sig-paren">(</span><em class="sig-param"><span class="n"><span class="pre">dtype</span></span><span class="o"><span class="pre">=</span></span><span class="default_value"><span class="pre">None</span></span></em>, <em class="sig-param"><span class="n"><span class="pre">copy</span></span><span class="o"><span class="pre">=</span></span><span class="default_value"><span class="pre">None</span></span></em><span class="sig-paren">)</span><a class="headerlink" href="#object.__array__" title="Permalink to this definition">¶</a></dt>
<dd><p>Return a Numpy array representing the gate. This can use the gate’s
<a class="reference internal" href="../stubs/qiskit.circuit.Instruction.html#qiskit.circuit.Instruction.params" title="qiskit.circuit.Instruction.params"><code class="xref py py-attr docutils literal notranslate"><span class="pre">params</span></code></a> field, and may assume that these are numeric
values (assuming the subclass expects that) and not
<a class="reference internal" href="#circuit-compile-time-parameters"><span class="std std-ref">compile-time parameters</span></a>.</p>
<p>For greatest efficiency, the returned array should default to a dtype of <a class="reference external" href="https://docs.python.org/3/library/functions.html#complex" title="(in Python v3.12)"><code class="xref py py-class docutils literal notranslate"><span class="pre">complex</span></code></a>.</p>
</dd></dl>
```

jakelishman (approved):

PR #12399, opened 2024-05-14
Remove not-context-manager lint rule

joesho112358:
### Summary

#9614 removing `not-context-manager`

### Details and comments

This is the only way i could find to get the typing to not be an issue other than removing it completely

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -221,7 +221,6 @@ disable = [
     "consider-using-f-string",
     "no-member",
     "no-value-for-parameter",
-    "not-context-manager",
     "unexpected-keyword-arg",
     "unnecessary-dunder-call",
     "unnecessary-lambda-assignment",
--- a/qiskit/circuit/quantumcircuit.py
+++ b/qiskit/circuit/quantumcircuit.py
@@ -5029,29 +5029,15 @@ class QuantumCircuit:
             else:
                 self._parameter_table[atomic_parameter] = new_entries
 
-    @typing.overload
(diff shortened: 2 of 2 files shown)


jakelishman (commented):
Thanks for looking at this.  I'm not super keen to merge this one as-is, because I think the problem is more likely a bug in Astroid's inference.  I do think there's typing problems in the currently spelling, but when I fixed those locally (so mypy is happy with things), it didn't fix the inference that pylint/astroid were pulling out - they were detecting the return type from the context-manager form as `None`, which was definitely false.

I'm roughly keen to work out why pylint's getting it wrong, because I think there should be a version of the `typing.overload` hinting that works - I think removing it entirely will cause Mypy to fail (not that it passes on Qiskit at the moment) because it would then think that it's undecideable whether the ersult .  I think at the moment it's probably passing pylint only _because_ it can't tell, so it errs on the safe side.

joesho112358:
@jakelishman not surprised about not being keen on it- i wasn't thrilled about the possible mismatching
of types i introduced with this, but figured something that works was a starting point! removing the typing did appear to work for me (not sure for Mypy), but figured keeping the possible mismatch of typing would be preferred over removal. i'll look to try a couple more things with the `typing.overload`

joesho112358:
looks like it's better to close this as the base has changed and the current changes are not inline with expectations. a fresh start is best i think, but i will probably have to bow out as well from this since i haven't made much progress

PR #12403, opened 2024-05-14
Write complete manual `QuantumCircuit` documentation

jakelishman:
### Summary

This writes huge tracts of new `QuantumCircuit` API documentation, linking together alike methods and writing explanatory text for how all the components fit together.  There's likely an awful lot more that could go into this too, but this hopefully should impose a lot more order on the huge `QuantumCircuit` documentation page, and provide a lot more explanation for how the class works holistically.

In particular, the section on the control-flow builder interface could do with a lot more exposition and examples right now.

### Details and comments

I ought to come back and write more about the control-flow builder interface with more examples, but this is already pushing it for 1.1 and I have other things that also need my attention right now.

@Eric-Arellano: at the moment, I've added `QuantumCircuit` to the toctree directly, especially because I don't want it to get generated with the standard templates we use for that.  If that's going to be a problem / not great, we can look at alternatives.

### DIFF
--- a/docs/apidoc/qiskit.circuit.QuantumCircuit.rst
+++ b/docs/apidoc/qiskit.circuit.QuantumCircuit.rst
@@ -1,0 +1,17 @@
+.. _qiskit-circuit-quantumcircuit:
+
+==============================
+:class:`.QuantumCircuit` class
+==============================
+
+..
+   This is so big it gets its own page in the toctree, and because we
+   don't want it to use autosummary.
+
+.. currentmodule:: qiskit.circuit
+
+.. autoclass:: qiskit.circuit.QuantumCircuit
+   :no-members:
+   :no-inherited-members:
+   :no-special-members:
+   :class-doc-from: class
--- a/qiskit/circuit/quantumcircuit.py
+++ b/qiskit/circuit/quantumcircuit.py
@@ -421,12 +421,23 @@ class QuantumCircuit:
             The methods for adding new :class:`~.expr.Var` variables to a circuit after
             initialization.
 
+    You can retrive a :class:`~.expr.Var` instance attached to a circuit by using its variable name
+    using :meth:`get_var`, or check if a circuit contains a given variable with :meth:`has_var`.
+
     .. automethod:: get_var
     .. automethod:: has_var
+
+    There are also several iterator methods that you can use to get the full set of variables
+    tracked by a circuit.  At least one of :meth:`iter_input_vars` and :meth:`iter_captured_vars`
+    will be empty, as inputs and captures are mutually exclusive.  All of the iterators have
+    corresponding dynamic properties on :class:`QuantumCircuit` that contain their length:
+    :attr:`num_vars`, :attr:`num_input_vars`, :attr:`num_captured_vars` and
+    :attr:`num_declared_vars`.
+
+    .. automethod:: iter_vars
+    .. automethod:: iter_input_vars
     .. automethod:: iter_captured_vars
     .. automethod:: iter_declared_vars
-    .. automethod:: iter_input_vars
-    .. automethod:: iter_vars
 
 
     .. _circuit-adding-operations:
--- a/qiskit/circuit/quantumcircuit.py
+++ b/qiskit/circuit/quantumcircuit.py
@@ -3301,25 +3301,36 @@ class QuantumCircuit:
 
     def depth(
         self,
-        filter_function: Callable[..., int] = lambda x: not getattr(
+        filter_function: Callable[[CircuitInstruction], bool] = lambda x: not getattr(
             x.operation, "_directive", False
         ),
     ) -> int:
         """Return circuit depth (i.e., length of critical path).
 
         Args:
-            filter_function (callable): A function to filter instructions.
-                Should take as input a tuple of (Instruction, list(Qubit), list(Clbit)).
-                Instructions for which the function returns False are ignored in the
-                computation of the circuit depth.
-                By default filters out "directives", such as barrier or snapshot.
+            filter_function: A function to decide which instructions count to increase depth.
+                Should take as a single positional input a :class:`CircuitInstruction`.
+                Instructions for which the function returns ``False`` are ignored in the
+                computation of the circuit depth.  By default filters out "directives", such as
+                :class:`.Barrier`.
 
         Returns:
             int: Depth of circuit.
 
-        Notes:
-            The circuit depth and the DAG depth need not be the
-            same.
+        Examples:
+            Simple calculation of total circuit depth::
+
+                from qiskit.circuit import QuantumCircuit
+                qc = QuantumCircuit(4)
+                qc.h(0)
+                qc.cx(0, 1)
+                qc.h(2)
+                qc.cx(2, 3)
+                assert qc.depth() == 2
+
+            Modifying the previous example to only calculate the depth of multi-qubit gates::
+
+                assert qc.depth(lambda instr: len(instr.qubits) > 1) == 1
         """
         # Assign each bit in the circuit a unique integer
         # to index into op_stack.
--- a/qiskit/circuit/quantumcircuit.py
+++ b/qiskit/circuit/quantumcircuit.py
@@ -3653,6 +3653,11 @@ class QuantumCircuit:
         """Clear all instructions in self.
 
         Clearing the circuits will keep the metadata and calibrations.
+
+        .. seealso::
+            :meth:`copy_empty_like`
+                A method to produce a new circuit with no instructions and all the same tracking of
+                quantum and classical typed data, but without mutating the original circuit.
         """
         self._data.clear()
         self._parameter_table.clear()
@@ -3886,6 +3891,28 @@ class QuantumCircuit:
 
         Measurements and barriers are considered final if they are
         followed by no other operations (aside from other measurements or barriers.)
+
+        .. note::
+            This method has rather complex behavior, particularly around the removal of newly idle
+            classical bits and registers.  It is much more efficient to avoid adding unnecessary
+            classical data in the first place, rather than trying to remove it later.
+
+        .. seealso::
+            :class:`.RemoveFinalMeasurements`
+                A transpiler pass that removes final measurements and barriers.  This does not
+                remove the classical data.  If this is your goal, you can call that with::
+
+                    from qiskit.circuit import QuantumCircuit
+                    from qiskit.transpiler.passes import RemoveFinalMeasurements
+
+                    qc = QuantumCircuit(2, 2)
+                    qc.h(0)
+                    qc.cx(0, 1)
+                    qc.barrier()
+                    qc.measure([0, 1], [0, 1])
+
+                    pass_ = RemoveFinalMeasurements()
+                    just_bell = pass_(qc)
 
         Args:
             inplace (bool): All measurements removed inplace or return new circuit.
--- a/qiskit/circuit/quantumcircuit.py
+++ b/qiskit/circuit/quantumcircuit.py
@@ -106,113 +106,859 @@ ClbitSpecifier = Union[
 BitType = TypeVar("BitType", Qubit, Clbit)
 
 
+# NOTE:
+#
+# If you're adding methods or attributes to `QuantumCircuit`, be sure to update the class docstring
+# to document them in a suitable place.  The class is huge, so we do its documentation manually so
+# it has at least some amount of organisational structure.
+
+
 class QuantumCircuit:
-    """Create a new circuit.
+    """Core Qiskit representation of a quantum circuit.
 
-    A circuit is a list of instructions bound to some registers.
+    .. note::
+        For more details setting the :class:`QuantumCircuit` in context of all of the data
+        structures that go with it, how it fits into the rest of the :mod:`qiskit` package, and the
+        different regimes of quantum-circuit descriptions in Qiskit, see the module-level
+        documentation of :mod:`qiskit.circuit`.
 
-    Args:
-        regs (list(:class:`~.Register`) or list(``int``) or list(list(:class:`~.Bit`))): The
-            registers to be included in the circuit.
+    Circuit attributes
+    ==================
 
-            * If a list of :class:`~.Register` objects, represents the :class:`.QuantumRegister`
-              and/or :class:`.ClassicalRegister` objects to include in the circuit.
+    :class:`QuantumCircuit` has a small number of public attributes, which are mostly older
+    functionality.  Most of its functionality is accessed through methods.
 
-              For example:
+    A small handful of the attributes are intentionally mutable, the rest are data attributes that
+    should be considered immutable.
 
-                * ``QuantumCircuit(QuantumRegister(4))``
-                * ``QuantumCircuit(QuantumRegister(4), ClassicalRegister(3))``
-                * ``QuantumCircuit(QuantumRegister(4, 'qr0'), QuantumRegister(2, 'qr1'))``
+    ========================= ======================================================================
+    Mutable attribute         Summary
+    ========================= ======================================================================
+    :attr:`global_phase`      The global phase of the circuit, measured in radians.
+    :attr:`metadata`          Arbitrary user mapping, which Qiskit will preserve through the
+                              transpiler, but otherwise completely ignore.
+    :attr:`name`              An optional string name for the circuit.
+    ========================= ======================================================================
 
-            * If a list of ``int``, the amount of qubits and/or classical bits to include in
-              the circuit. It can either be a single int for just the number of quantum bits,
-              or 2 ints for the number of quantum bits and classical bits, respectively.
+    ========================= ======================================================================
+    Immutable data attribute  Summary
+    ========================= ======================================================================
+    :attr:`ancillas`          List of :class:`AncillaQubit`\\ s tracked by the circuit.
+    :attr:`calibrations`      Custom user-supplied pulse calibrations for individual instructions.
+    :attr:`cregs`             List of :class:`ClassicalRegister`\\ s tracked by the circuit.
 
-              For example:
+    :attr:`clbits`            List of :class:`Clbit`\\ s tracked by the circuit.
+    :attr:`data`              List of individual :class:`CircuitInstruction`\\ s that make up the
+                              circuit.
+    :attr:`duration`          Total duration of the circuit, added by scheduling transpiler passes.
 
-                * ``QuantumCircuit(4) # A QuantumCircuit with 4 qubits``
-                * ``QuantumCircuit(4, 3) # A QuantumCircuit with 4 qubits and 3 classical bits``
+    :attr:`layout`            Hardware layout and routing information added by the transpiler.
+    :attr:`num_ancillas`      The number of ancilla qubits in the circuit.
+    :attr:`num_clbits`        The number of clbits in the circuit.
+    :attr:`num_captured_vars` Number of captured real-time classical variables.
 
-            * If a list of python lists containing :class:`.Bit` objects, a collection of
-              :class:`.Bit` s to be added to the circuit.
+    :attr:`num_declared_vars` Number of locally declared real-time classical variables in the outer
+                              circuit scope.
+    :attr:`num_input_vars`    Number of input real-time classical variables.
+    :attr:`num_parameters`    Number of compile-time :class:`Parameter`\\ s in the circuit.
+    :attr:`num_qubits`        Number of qubits in the circuit.
+
+    :attr:`num_vars`          Total number of real-time classical variables in the outer circuit
+                              scope.
+    :attr:`op_start_times`    Start times of scheduled operations, added by scheduling transpiler
+                              passes.
+    :attr:`parameters`        Ordered set-like view of the compile-time :class:`Parameter`\\ s
+                              tracked by the circuit.
+    :attr:`qregs`             List of :class:`QuantumRegister`\\ s tracked by the circuit.
+
+    :attr:`qubits`            List of :class:`Qubit`\\ s tracked by the circuit.
+    :attr:`unit`              The unit of the :attr:`duration` field.
+    ========================= ======================================================================
+
+    The core attribute is :attr:`data`.  This is a sequence-like object that exposes the
+    :class:`CircuitInstruction`\\ s contained in an ordered form.  You generally should not mutate
+    this object directly; :class:`QuantumCircuit` is only designed for append-only operations (which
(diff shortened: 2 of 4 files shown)


docs/apidoc/qiskit.circuit.QuantumCircuit.rst
```diff
@@ -0,0 +1,17 @@
+.. _qiskit-circuit-quantumcircuit:
+
+==============================
+:class:`.QuantumCircuit` class
```
Eric-Arellano, line 4: Some notes if we stick with this rather than the simpler `QuantumCircuit` (no link):

1. We're not consistent with all the other class pages, e.g. https://docs.quantum.ibm.com/api/qiskit/qiskit.circuit.AncillaRegister
2. I don't think the link is very useful. On https://qiskit-docs-preview-pr-1380.1799mxdls7qz.us-south.codeengine.appdomain.cloud/api/qiskit/dev/qiskit.circuit.QuantumCircuit, it simply takes you to the class definition which is the first part of the page anyways
3. The styling is messed up for code in headers, tracked by https://github.ibm.com/IBM-Q-Software/iqp-channel-docs/issues/1037. If we stick with this approach, we'll try to fix the issue in the next week

Despite those concerns, I think I agree with your approach here, particularly because it makes the index page more readable with `QuantumCircuit` being escaped as code and having the word `class`. I tried to get a custom title to work in `index.rst` like '\`\`QuantumCircuit\`\` class <qiskit.circuit.QuantumCircuit>', but RST doesn't like the inline formatting with \`.

Eric-Arellano (commented):
Nice! Reviewers, check out https://qiskit-docs-preview-pr-1380.1799mxdls7qz.us-south.codeengine.appdomain.cloud/api/qiskit/dev/qiskit.circuit.QuantumCircuit to see how this renders.

--

> @Eric-Arellano: at the moment, I've added QuantumCircuit to the toctree directly, especially because I don't want it to get generated with the standard templates we use for that. If that's going to be a problem / not great, we can look at alternatives.

I think it's good. FYI it results in the index page and ToC looking like this

Screenshot 2024-05-14 at 3 55 10 PM

I'm fine with the index page showing the class since it's so important. And I'm glad the left ToC is still sorted alphabetically. 

Thanks for naming the file `qiskit.circuit.QuantumCircuit` so that we don't need redirects! It would be good to always use that naming scheme moving forward for modules too so that our URLs are more predictable: the import path.

docs/apidoc/qiskit.circuit.QuantumCircuit.rst
jakelishman: I mostly did this out of habit because I almost always put in cross-references when possible.  If there's a way that works a little better, I can do that.

From my side, I don't mind the styling not being perfect immediately, if that's the remaining blocker (I'd seen it locally, too). It's a styling bug that we probably need to get fixed anyway, because I'd be surprised if it's not in use in at least some other places - I'm quite likely to have done it somewhere else in the API docs too - but it's still totally legible in the meantime, it just looks slightly odd.

Eric-Arellano: If we could get the `index.rst` to properly render '``QuantumCircuit`` class <qiskit.circuit.QuantumCircuit>', then I would ask us to change to simply `QuantumCircuit`. But I couldn't get that working. So I think what you have is the best choice because it makes the index page more readable.

> From my side, I don't mind the styling not being perfect immediately, if that's the remaining blocker

Indeed, it's my biggest concern. It's always been a problem though, and we'll bump https://github.ibm.com/IBM-Q-Software/iqp-channel-docs/issues/1037 up in priority. It's a short-term bug, so I agree it's not a blocker.

jakelishman: You mean a link where the inline code block and the word "class" are both the link text? If so: nested syntax (inline code inside a link) is (in)famously hard to do in rST. It's just about possible to hack together in some cases, but the syntax is at best arcane.

Sphinx-specific stuff is easier because that produces the nested elements by manipulating the `docutils` tree directly.

Eric-Arellano: > You mean a link where the inline code block and the word "class" are both the link text? If so: nested syntax (inline code inside a link) is (in)famously hard to do in rST. It's just about possible to hack together in some cases, but the syntax is at best arcane.

Yep, exactly. So, I think we should stick with how you have things.

jakelishman:
> Thanks for naming the file `qiskit.circuit.QuantumCircuit` so that we don't need redirects!

Oh, I missed this comment under the picture, but yeah no worries - I'm relatively sure I asked you and Arnau what I should name the file several weeks ago, so I'm just following instructions haha.

qiskit/circuit/quantumcircuit.py
```diff
@@ -106,113 +106,859 @@
...
+
+    The :attr:`name` of a circuit becomes the name of the :class:`~.circuit.Instruction` or
+    :class:`.Gate` resulting from :meth:`to_instruction` and :meth:`to_gate` calls, which can be
+    handy for visualizations.
+
+    .. autoattribute:: name
+
+    You can attach arbitrary :attr:`metadata` to a circuit.  No part of core Qiskit will inspect
+    this or change its behavior based on metadata, but it will be faithfully passed through the
+    transpiler, so you can tag your circuits yourself.
```
mtreinish: Do you want to link to the qpy dump docs here? There are some considerations for qpy circuit serialization around how this gets serialized to json internally.

qiskit/circuit/quantumcircuit.py
```diff
@@ -106,113 +106,859 @@
...
+                              passes.
+    :attr:`parameters`        Ordered set-like view of the compile-time :class:`Parameter`\\ s
+                              tracked by the circuit.
+    :attr:`qregs`             List of :class:`QuantumRegister`\\ s tracked by the circuit.
+
+    :attr:`qubits`            List of :class:`Qubit`\\ s tracked by the circuit.
+    :attr:`unit`              The unit of the :attr:`duration` field.
+    ========================= ======================================================================
+
+    The core attribute is :attr:`data`.  This is a sequence-like object that exposes the
```
mtreinish, line 178: This whole structure is formatted a bit weirdly in: https://qiskit-docs-preview-pr-1380.1799mxdls7qz.us-south.codeengine.appdomain.cloud/api/qiskit/dev/qiskit.circuit.QuantumCircuit the vertical alignment between the sections is weird. I'm wondering if we need to add horizontal lines to break it up, or maybe a lot of subsections. Because of the autoattribute formatting and spacing it looks like the text description below the autoattribute lines are associated with that attribute instead of the text block above it.

(it also does in the rst to me, I kept leaving comments on the wrong spot but figured it would render better)

qiskit/circuit/quantumcircuit.py
```diff
@@ -106,113 +106,859 @@
...
+    You can attach arbitrary :attr:`metadata` to a circuit.  No part of core Qiskit will inspect
+    this or change its behavior based on metadata, but it will be faithfully passed through the
+    transpiler, so you can tag your circuits yourself.
+
+    .. autoattribute:: metadata
+
+    :class:`QuantumCircuit` exposes data attributes tracking its internal quantum and classical bits
+    and registers.  These appear as Python :class:`list`\\ s, but you should treat them as
+    immutable; changing them will *at best* have no effect, and more likely will simply cause
+    :class:`QuantumCircuit`'s internal data to become incoherent.
```
mtreinish: I might reword this as corrupting the structure to make it a bit stronger.

```suggestion
    immutable; changing them will *at best* have no effect, and more likely will simply corrupt
    the internal data of the :class:`QuantumCircuit` object.
```

qiskit/circuit/quantumcircuit.py
```diff
@@ -106,113 +106,859 @@
...
 
-            This argument is intended for convenient circuit initialization when you already have a
-            set of created variables.  The variables used here will be directly passed to
-            :meth:`add_var`, which you can use directly if this is the first time you are creating
-            the variable.
+    Registers are added to the circuit with :meth:`add_register`.  In this method, it is not an
+    error if some of the bits are already present in the circuit.  In this case, the register will
+    be an "alias" over the bits.  This is not generally well-supported by any hardware backends;
+    while it is valid in the OpenQASM 3 abstract data model, it is probably best to stay away from
+    relying on it.  The registers a given bit is in are part of the return of :meth:`find_bit`.
```
mtreinish: I'm not sure we can make an absolute statement about this for all backends. I assume you were referring to IBM backends. Although the non-ibm backends I've worked with don't really model registers at all and just do the tracking in qiskit.

I also don't think we want to couple this with a discussion of the qasm3 data model because it's largely independent from that except if you're exporting the circuit to qasm3.

```suggestion
    be an "alias" over the bits.  This is not generally well-supported by most hardware backends;
    it is probably best to stay away from relying on it.  The registers a given bit is in are part of
     the return of :meth:`find_bit`.
```

qiskit/circuit/quantumcircuit.py
```diff
@@ -106,113 +106,859 @@
...
+
+    Working with bits and registers
+    -------------------------------
+
+    A :class:`.Bit` instance is, on its own, just a unique handle for circuits to use in their own
+    contexts.  If you have got a :class:`.Bit` instance and a cirucit, just can find the contexts
+    that the bit exists in using :meth:`find_bit`, such as its integer index in the circuit and any
+    registers it is contained in.
+
+    .. automethod:: find_bit
```
mtreinish, line 375: I know you haven't had any code examples in the text here. But having an example of using `find_bit` to get the positional index of a bit object either in the ext or part of the automethod (I checked it's not there currently) would be really useful here I think. Doing `qc.find_bit(foo).index` is something I see people needing to do all the time (and I see a lot of people doing `bit._index`)

qiskit/circuit/quantumcircuit.py
```diff
@@ -106,113 +106,859 @@
...
+        :ref:`circuit-adding-data-objects`
+            The methods for adding new :class:`~.expr.Var` variables to a circuit after
+            initialization.
+
+    .. automethod:: get_var
+    .. automethod:: has_var
+    .. automethod:: iter_captured_vars
+    .. automethod:: iter_declared_vars
+    .. automethod:: iter_input_vars
+    .. automethod:: iter_vars
```
mtreinish: Did you want any explainer text between the seealso box and the automethods here?

qiskit/circuit/quantumcircuit.py
```diff
@@ -106,113 +106,859 @@
...
+    :class:`~.circuit.Gate` instances.
+
+    .. seealso::
+        :ref:`circuit-operations-instructions`
+            The :mod:`qiskit.circuit`-level documentation on the different interfaces that Qiskit
+            uses to define circuit-level instructions.
+
+    .. _circuit-append-compose:
+
+    Driver methods to add operations
```
mtreinish: What does driver mean in this context? I would probably have just said "Methods to add operations"

qiskit/circuit/quantumcircuit.py
```diff
@@ -106,113 +106,859 @@
...
+    (:meth:`compose`).
+
+    A single instruction or operation appears as a single entry in the :attr:`data` of the circuit,
+    and as a single box when drawn in the circuit visualizers (see :meth:`draw`).  A single
+    instruction is the "unit" that a hardware backend might be defined in terms of (see
+    :class:`.Target`).  An :class:`~.circuit.Instruction` can come with a
+    :attr:`~.circuit.Instruction.definition`, which is one rule the transpiler (see
+    :mod:`qiskit.transpiler`) will be able to fall back on to decompose it for hardware, if needed.
+    An :class:`.Operation` can only be decomposed if it has some associated high-level synthesis
+    method registered for it (see :mod:`qiskit.transpiler.passes.synthesis.plugin`).
```
mtreinish: I think we should be a bit more clear here, because an `Instruction` and `Gate` object are also `Operation`s. There is also the complexity around opaque instructions and gates which don't include a definition. Which is covered by the previous sentence but not really expanded upon.

```suggestion
    A direct implementer of the :class:`.Operation` interface can only be decomposed if it has some associated high-level synthesis
    method registered for it (see :mod:`qiskit.transpiler.passes.synthesis.plugin`) as it does not include a default
    decomposition via :attr`.Instruction.definition`.
```

qiskit/circuit/quantumcircuit.py
```diff
@@ -106,113 +106,859 @@
...
+    If you are trying to join two circuits that will apply to completely disjoint qubits and clbits,
+    :meth:`tensor` is a convenient wrapper around manually adding bit objects and calling
+    :meth:`compose`.
+
+    .. automethod:: tensor
+
+    As some rules of thumb:
+
+    * If you have a single :class:`.Operation`, :class:`~.circuit.Instruction` or :class:`.Gate`,
+      you should definitely use :meth:`append`.
```
mtreinish: ?
```suggestion
      you should definitely use :meth:`append` or :meth:`_append`.
```

qiskit/circuit/quantumcircuit.py
```diff
@@ -106,113 +106,859 @@
...
+    * :meth:`tensor` is wanted far more rarely than either :meth:`append` or :meth:`compose`.
+      Internally, it is mostly a wrapper around :meth:`add_bits` and :meth:`compose`.
+
+    Some potential pitfalls to beware of:
+
+    * Even if you re-use a custom :class:`~.circuit.Instruction` during circuit construction, the
+      transpiler will generally have to "unroll" each invocation of it to its inner decomposition
+      before beginning work on it.  This should not prevent you from using the
+      :meth:`to_instruction`-plus-:meth:`append` pattern, as the transpiler will also become
+      cleverer about this over time.
```
mtreinish:
```suggestion
      :meth:`to_instruction`-plus-:meth:`append` pattern, as the transpiler will improve in this regard
      over time.
```

qiskit/circuit/quantumcircuit.py
```diff
@@ -106,113 +106,859 @@
...
+    .. automethod:: measure_active
+    .. automethod:: measure_all
+
+    There are two "subtractive" methods on :class:`QuantumCircuit` as well.  This is not a use-case
+    that :class:`QuantumCircuit` is designed for; typically you should just look to use
+    :meth:`copy_empty_like` in place of :meth:`clear`, and run :meth:`remove_final_measurements` as
+    its transpiler-pass form :class:`.RemoveFinalMeasurements`.
+
+    .. automethod:: clear
+    .. automethod:: remove_final_measurements
```
mtreinish, line 803: I think showing examples somewhere of the preferred way to do this would be good probably in the automethod docstring.

qiskit/circuit/quantumcircuit.py
```diff
@@ -106,113 +106,859 @@
...
+    directed acyclic graph (DAG).  However, such a definition is a bit hard to grasp, even for
+    experts.  Fortunately, the depth of a circuit can be easily understood by anyone familiar
+    with playing `Tetris <https://en.wikipedia.org/wiki/Tetris>`_.  Lets see how to compute this
+    graphically:
+
+    .. image:: /source_images/depth.gif
+
+    We can verify our graphical result using :meth:`QuantumCircuit.depth`::
+
+       assert qc.depth() == 9
```
mtreinish: I realize this is mostly copied over from before, but I wonder if we want to show an example of 2q gate depth and how to compute that using `depth()`. Either here or in the method docstring.

qiskit/circuit/quantumcircuit.py
```diff
@@ -106,113 +106,859 @@
...
+    which are mostly useful for quickly unwrapping some inner instructions or reversing the
+    :ref:`qubit-labelling conventions <circuit-conventions>` in the drawing.  For more general
+    mutation, including basis-gate rewriting, you should use the transpiler
+    (:mod:`qiskit.transpiler`).
+
+    .. automethod:: decompose
+    .. automethod:: reverse_bits
+
+    ..
+        Deliberately not documented; these methods shouldn't be public.
```
mtreinish: Should we explicitly deprecate them? Since they're in the public api docs we need to support them, but we can make new private methods and switch internal usage over to those and raise a deprecation warnings on the "public" interface.

mtreinish (commented):
Overall this looks great, there is a lot much needed detail being added to the docs here which will be a huge help. I have left a few inline comments on some specific minor details. The only big thing I'm concerned about is how this ends up getting formatted on the hosted docs page. The way headings get set for the auto attributes makes it quite confusing, for example:

Screenshot_2024-05-15_13-34-45

All the explainer text looks like it's in the wrong section because of this.

Eric-Arellano:
> All the explainer text looks like it's in the wrong section because of this.

Fwiw, @arnaucasau and I realized this is because the margin-top for h3s is quite large. I started talking to one of the designers if we can reduce it. There's a chance we can reduce it a bit, but they aren't convinced yet we should.

Related context: we went back-and-forth on removing those auto-generated headers and decided to keep them. They're there to populate the right ToC and allow users to copy an anchor link. 

So, the best improvement the docs app could make here is reducing margin-top for h3.

jakelishman:
Eric: I'd say that the styling is actually correct here, the problem is that semantically, a function name etc shouldn't be a document heading. A heading implies that everything after it should be part of that title (which is what the CSS is currently set up to indicate), but that's not semantically what the documentation actually means; the function names are just things to be defined, and that doesn't impact the overall grouping structure of the document.

Sphinx's HTML output is semantically "definition lists" for these, which is (imo) more correct. I know that we currently need headers to make the right sidebar work, but I think that the most proper fix here is to fix the HTML output and then fix whatever knock-on effects that has.

Eric-Arellano:
> Eric: I'd say that the styling is actually correct here, the problem is that semantically, a function name etc shouldn't be a document heading.

You're spot on. I agree, and I appreciate you taking the time to think that through. The problem indeed becomes much more apparent with these custom-written module and class pages, compared to autosummary.

I opened https://github.com/Qiskit/documentation/issues/1395. I still need to get designer sign-off, and it certainly won't be implemented by tomorrow's release, but I hope we can get to it within the next month. We're trying to finish this revamp of API docs before other priorities come up. 

So, from the docs perspective, I wouldn't change anything about this PR re: the weird margins. The docs are written correctly and we'll try to fix this on the app side of things.

qiskit/circuit/quantumcircuit.py
mtreinish: I'm going to mark this as resolved because of @Eric-Arellano's comment here https://github.com/Qiskit/qiskit/pull/12403#issuecomment-2113575766 and the PR: https://github.com/Qiskit/documentation/issues/1395 which should fix this on the docs hosting side.

qiskit/circuit/quantumcircuit.py
jakelishman: Oh yes, thanks - I must have grouped these into one heading and then forgotten to write the text that goes with them.

qiskit/circuit/quantumcircuit.py
jakelishman: Imo we should (also, this comment was mostly to myself while I was grouping things before I started writing the text), but at this point it's probably best to do that in 1.2.

qiskit/circuit/quantumcircuit.py
jakelishman: I did copy this (near) verbatim.  I think the text of the section is meant to be read as a whole thing, including the image - the `qc` circuit created up top is the same one as shown in the gif, and it does have a depth of 9.

qiskit/circuit/quantumcircuit.py
jakelishman: I mostly did this by explicit see-alsos, but mostly done in f6a2e50f37.

qiskit/circuit/quantumcircuit.py
jakelishman: Done in 956475957f.

qiskit/circuit/quantumcircuit.py
jakelishman: Done in 956475957f.

qiskit/circuit/quantumcircuit.py
jakelishman: Done in 956475957f.

qiskit/circuit/quantumcircuit.py
jakelishman: I think I meant `append` is the "driver" of `rx`, `cx`, etc.  I'll change it to "methods to add general operations" to contrast with "methods to add standard instructions" below it.  Done in 956475957f.

qiskit/circuit/quantumcircuit.py
jakelishman: Done in bb5456e517.

qiskit/circuit/quantumcircuit.py
jakelishman: Added in 36b58ff043.

qiskit/circuit/quantumcircuit.py
jakelishman: Done in 956475957f.

qiskit/circuit/quantumcircuit.py
jakelishman: Done in 956475957f.

qiskit/circuit/quantumcircuit.py
jakelishman: Oh good call - done in 5cc79d724c.

qiskit/circuit/quantumcircuit.py
jakelishman: In the meantime: I added them back to the documentation in 36b53f4010, and I'll prepare a deprecation PR for 1.2.

mtreinish: I'll open a tracking issue for this so we don't lose the thread

mtreinish: Done here: https://github.com/Qiskit/qiskit/issues/12417

qiskit/circuit/quantumcircuit.py
mtreinish: I was thinking more back to some offline conversations we were having with about having a documented way to do 2 qubit depth. But I think this probably belongs as an example in the `depth` method's docstring on how to use the filter arg.

jakelishman: Done in 32a766964d.

mtreinish (approved):
This LGTM now, thanks for the quick updates.

PR #12222, opened 2024-04-19
Make `SamplerPub` and `EstimatorPub` public

ihincks:
### Summary

This PR is a reversion of #11678. 

### Details and comments

This PR is blocked by #11594.

PR #12408, opened 2024-05-15
export SamplerPub and EstimatorPub

kevinsung:
### Summary

I think these should be exported too.


t-imamichi:
It will be 1.2.0. See https://github.com/Qiskit/qiskit/pull/12222#issuecomment-2089266735
This PR is a duplicate of #12222

PR #12410, opened 2024-05-15
Fix unnecessary serialisation of `PassManager` in serial contexts

jakelishman:
### Summary

This exposes the interal decision in `parallel_map` of whether to actually run in serial or not.  If not, there's no need for `PassManager` to side-car its `dill` serialisation onto the side of the IPC (we use `dill` because we need to pickle lambdas), which can be an unfortunately huge cost for certain IBM pulse-enabled backends.

### Details and comments

@sbrandhsn: as the release manager for 1.1.0, it's up to you whether you want to accept this PR in this form for 1.1.0, or defer to 1.1.1 (you can tag the milestone appropriately).  I could also suppress exporting `should_run_in_parallel` into the public namespace if you're not on board with a new public feature this late into the release cycle.

### DIFF
--- a/qiskit/utils/__init__.py
+++ b/qiskit/utils/__init__.py
@@ -43,8 +43,9 @@ Multiprocessing
 
 .. autofunction:: local_hardware_info
 .. autofunction:: is_main_process
+.. autofunction:: should_run_in_parallel
 
-A helper function for calling a custom function with python
+A helper function for calling a custom function with Python
 :class:`~concurrent.futures.ProcessPoolExecutor`. Tasks can be executed in parallel using this function.
 
 .. autofunction:: parallel_map
@@ -70,7 +71,7 @@ from .lazy_tester import LazyDependencyManager, LazyImportTester, LazySubprocess
 
 from . import optionals
 
-from .parallel import parallel_map
+from .parallel import parallel_map, should_run_in_parallel
 
 __all__ = [
     "LazyDependencyManager",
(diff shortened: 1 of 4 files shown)


qiskit/utils/__init__.py
```diff
@@ -70,7 +71,7 @@
 
 from . import optionals
 
-from .parallel import parallel_map
+from .parallel import parallel_map, should_run_in_parallel
```
mtreinish, line 73: For backport I was thinking we didn't expose this as part of the public api. Right now it's really only needed for internal decision making.

mtreinish (commented):
To make this eligible for backport I was thinking that we leave `should_run_in_parallel` as an internal function for now. I agree having it in the public api makes sense longer term, but if we wanted to include this fix in 1.1.x we don't need to expose it in this PR.

qiskit/utils/__init__.py
jakelishman: Does it work for you if I leave this PR like this for 1.2, then suppress the documentation and release note of the new function during the backport to 1.1 stable?

mtreinish: If you don't mind doing a manual backport that's fine with me. Personally I'd make the diff on main adding the docs and release note because that seems less error prone imo. But doing it that way works fine

jakelishman: Let me do that, then - it'll be clearer to see what we did when if we look back in a few months' time.

jakelishman: Done in 5fc08aea3c.

mtreinish (approved):

1ucian0:
Do you think it makes sense to backport this to stable/0.46?

1ucian0:

PR #12412, opened 2024-05-15
Refactor parallelism utilities for public API

jakelishman:
### Summary

`should_run_in_parallel` was added in a stable manner to enable backport to the 1.1 series, but from 1.3 onwards, we want this to be part of the public interface so that others can rely on it too.

As part of this, the parallelisation configuration was made more robust and controllable with context managers.  This is convenient beyond just for users - it makes it far easier to control the parallelism during the test suite runs.  Several instances where different parts of Qiskit and its test suite reached into deep internals of the parallelism utilities and made significant assumptions about the internal logic are refactored to use public interfaces to achieve what they wanted to.

The multiprocessing detection is changed from making OS-based assumptions about what Python does to simply querying the module for its configuration.  This makes it more robust to changes in Python's handling (especially important since 3.14 will change the default start method on Unix).  In the future, we may want to change to making these assumptions only if the user hasn't configured the `multiprocessing` start method themselves.

### Details and comments

This was elided from #12410 to make that PR backwards compatible.  This PR exposes the feature as part of the public API, so will be new for 1.2.

Depends on #12410.


jakelishman:
Now rebased over #12410.

releasenotes/notes/parallel-check-public-7faed5f3e20e1d03.yaml
```diff
@@ -0,0 +1,7 @@
+---
+features_misc:
+  - |
+    A new function, :func:`qiskit.utils.should_run_in_parallel`, is available to determine whether
+    :func:`.parallel_map` will choose to run in parallel or degrade to running in serial.  This
+    decision is dependent on how many CPUs are available to Qiskit, what the :mod:`multiprocessing`
+    start method is, how many processes were requested.
```
mtreinish: I mean sort of, we don't explicitly check what multiprocessing is set to (ie we ignore if the user explicitly sets this to spawn), but the value for `PARALLEL_DEFAULT` is based on what the OS default start method is.

jakelishman: Yeah, I think I was being more aspirational than correct here. We _should_ do the check based on the `multiprocessing` start method if that's what we care about, but we don't. I can change the wording.

mtreinish: Maybe we should just do this, it seems totally in scope to update the logic in the function to call ` multiprocessing.get_start_method()` and get rid of the OS based logic here for 1.2.

releasenotes/notes/parallel-check-public-7faed5f3e20e1d03.yaml
```diff
@@ -0,0 +1,7 @@
+---
+features_misc:
+  - |
+    A new function, :func:`qiskit.utils.should_run_in_parallel`, is available to determine whether
+    :func:`.parallel_map` will choose to run in parallel or degrade to running in serial.  This
```
kevinhartman: The word `degrade` seems a bit strange to me here. Maybe `fall back` would be more appropriate? I see that this word is also used in the docstring for `should_run_in_parallel` and would need to be updated as well.

releasenotes/notes/parallel-check-public-7faed5f3e20e1d03.yaml
kevinhartman: > Maybe we should just do this, it seems totally in scope to update the logic in the function to call multiprocessing.get_start_method() and get rid of the OS based logic here for 1.2.

@mtreinish is this still something we ought to do prior to merging this for 1.3?

jakelishman: Yeah, we still intended to do this

jakelishman:
I've force-pushed a major new commit that properly refactored a bunch of the parallelism utilities to better support `should_run_in_parallel` at the level of robustness we'd expect from the public interface.  This then removes a bunch of our own library code and test code that was reaching in the belly of the internals of the parallel code and layering assumptions on top of assumptions, and wraps the necessary components in safe interfaces.

I also added a `QISKIT_IGNORE_USER_SETTINGS` environment variable and configured it to be used in all tox and CI runs by default, which lets us isolate the test suite from the environment - that should make the new tests of what I added reliable, but it's probably something we should have had before anyway.

jakelishman:
Given how major the changes I made here were, and that we're after the feature freeze deadline, I'm fine if we choose to leave this for Qiskit 2.0.

releasenotes/notes/parallel-check-public-7faed5f3e20e1d03.yaml
jakelishman: Completely changed this release note during the force push.

releasenotes/notes/parallel-check-public-7faed5f3e20e1d03.yaml
jakelishman: I did this, which then of course broke a bunch of weird other assumptions we were making thirty miles away in the library, so I ended up _really_ rejigging the public API of the parallelism utilities so we didn't need to make internal assumptions across file boundaries.

I'm not entirely sold on the way I wrote the `multiprocessing` test (see the commit message) - I was trying to match the spirit of the previous assumption-heavy code, but I'm not certain we couldn't do a shade better.

qiskit/utils/parallel.py
```diff
@@ -50,101 +54,227 @@
...
-    CPU_COUNT = int(os.getenv("QISKIT_NUM_PROCS"))
-else:
-    CPU_COUNT = CONFIG.get("num_process", local_hardware_info()["cpus"])
+    In order of priority (highest to lowest), the return value will be:
 
+    1. The ``QISKIT_NUM_PROCS`` environment variable, if set.
+    2. The ``num_processes`` key of the Qiskit user configuration file, if set.
+    3. Half of the logical CPUs available to this process, if this can be determined.  This is a
+       proxy for the number of physical CPUs, assuming two-fold simultaneous multithreading (SMT);
+       emperically, multiprocessing performance of Qiskit seems to be worse when attempting to use
```
kevinhartman:
```suggestion
       empirically, multiprocessing performance of Qiskit seems to be worse when attempting to use
```

Unless this is a British spelling I'm not aware of 😄.

mtreinish:
I agree, given the scope of the changes and that we're only one day out from rc1 lets defer this to 2.0. We can move forward with it pretty soon after rc1 though.

jakelishman:
This is rebased over `main` now.

kevinhartman:
Do we want to resurrect this for 2.0, @jakelishman ?

jakelishman:
Should be resurrected now: I already rebased it a couple of weeks ago, so the last conflicts here were just super minor ones in a couple of test files that changed.

qiskit/utils/parallel.py
```diff
@@ -50,101 +54,227 @@
...
+       SMT cores.
+    4. 1, if all else fails.
+    """
+    if (env_num_processes := os.getenv("QISKIT_NUM_PROCS")) is not None:
+        return int(env_num_processes) or 1
```
kevinhartman: Should we guard against the case where the user may have (for whatever reason) specified a negative number of processes through either the env var or config file?

kevinhartman (commented):
Looks good to me. Just one minor comment.

qiskit/utils/parallel.py
jakelishman: The config file enforces it at load time already, I think, but I added it in for both to be sure in 27ed78c4f2.

kevinhartman (approved):

PR #12415, opened 2024-05-15
Update CITATION.bib file to point to Qiskit whitepaper

mtreinish:
### Summary

This commit updates the official citation for the Qiskit project to point to the overview paper "Quantum computing with Qiskit" which was recently published on arxiv:

https://arxiv.org/abs/2405.08810

### Details and comments

### DIFF
--- a/CITATION.bib
+++ b/CITATION.bib
@@ -1,6 +1,8 @@
-@misc{Qiskit,
-    author = {{Qiskit contributors}},
-    title = {Qiskit: An Open-source Framework for Quantum Computing},
-    year = {2023},
-    doi = {10.5281/zenodo.2573505}
+@misc{javadiabhari2024quantum,
+      title={Quantum computing with Qiskit},
+      author={Ali Javadi-Abhari and Matthew Treinish and Kevin Krsulich and Christopher J. Wood and Jake Lishman and Julien Gacon and Simon Martiel and Paul D. Nation and Lev S. Bishop and Andrew W. Cross and Blake R. Johnson and Jay M. Gambetta},
+      year={2024},
+      eprint={2405.08810},
+      archivePrefix={arXiv},
+      primaryClass={quant-ph}
 }
--- a/CITATION.bib
+++ b/CITATION.bib
@@ -1,7 +1,8 @@
-@misc{javadiabhari2024quantum,
-      title={Quantum computing with Qiskit},
-      author={Ali Javadi-Abhari and Matthew Treinish and Kevin Krsulich and Christopher J. Wood and Jake Lishman and Julien Gacon and Simon Martiel and Paul D. Nation and Lev S. Bishop and Andrew W. Cross and Blake R. Johnson and Jay M. Gambetta},
+@misc{qiskit2024,
+      title={Quantum computing with {Q}iskit},
+      author={Javadi-Abhari, Ali and Treinish, Matthew and Krsulich, Kevin and Wood, Christopher J. and Lishman, Jake and Gacon, Julien and Martiel, Simon and Nation, Paul D. and Bishop, Lev S. and Cross, Andrew W. and Johnson, Blake R. and Gambetta, Jay M.},
       year={2024},
+      doi={10.48550/arXiv.2405.08810},
       eprint={2405.08810},
       archivePrefix={arXiv},
       primaryClass={quant-ph}


CITATION.bib
```diff
@@ -1,6 +1,8 @@
-@misc{Qiskit,
-    author = {{Qiskit contributors}},
-    title = {Qiskit: An Open-source Framework for Quantum Computing},
-    year = {2023},
-    doi = {10.5281/zenodo.2573505}
+@misc{javadiabhari2024quantum,
+      title={Quantum computing with Qiskit},
+      author={Ali Javadi-Abhari and Matthew Treinish and Kevin Krsulich and Christopher J. Wood and Jake Lishman and Julien Gacon and Simon Martiel and Paul D. Nation and Lev S. Bishop and Andrew W. Cross and Blake R. Johnson and Jay M. Gambetta},
```
jakelishman: I don't remember to what degree it matters (might depend on the `.bst` file), but BibTeX typically prefers `Lastname, Firstnames` format:
```suggestion
      author={Javadi-Abhari, Ali and Treinish, Matthew and Krsulich, Kevin and Wood, Christopher J. and Lishman, Jake and Gacon, Julien and Martiel, Simon and Nation, Paul D. and Bishop, Lev S. and Cross, Andrew W. and Johnson, Blake R. and Gambetta, Jay M.},
```

@levbishop: do you remember if it actually matters / in what circumstances?

CITATION.bib
```diff
@@ -1,6 +1,8 @@
-@misc{Qiskit,
-    author = {{Qiskit contributors}},
-    title = {Qiskit: An Open-source Framework for Quantum Computing},
-    year = {2023},
-    doi = {10.5281/zenodo.2573505}
+@misc{javadiabhari2024quantum,
+      title={Quantum computing with Qiskit},
```
jakelishman:
```suggestion
      title={Quantum computing with {Q}iskit},
```
we don't care much about how people's style files do the rest of the title, but Qiskit always wants a capital letter

CITATION.bib
```diff
@@ -1,6 +1,8 @@
...
+      author={Ali Javadi-Abhari and Matthew Treinish and Kevin Krsulich and Christopher J. Wood and Jake Lishman and Julien Gacon and Simon Martiel and Paul D. Nation and Lev S. Bishop and Andrew W. Cross and Blake R. Johnson and Jay M. Gambetta},
+      year={2024},
```
jakelishman, line 4:
```suggestion
      year={2024},
      doi={10.48550/arXiv.2405.08810},
```
arXiv has direct DOIs these days

CITATION.bib
mtreinish: FWIW, I literally just copied this bibtex from what the "export BibTeX citation" button on the arxiv page generated.

mtreinish: For future reference the source code that generated this is here: https://github.com/arXiv/arxiv-browse/blob/97eeef6ab8db5833141c69c055ef08cd05a0b6e7/browse/formatting/cite.py#L11 if we have concrete improvements to the formatting we should open a PR proposing them over there.

jakelishman: (Copied from slack, about the above link)

> I would imagine that they very specifically don’t want to play with the author lists there
>
> the reason for the explicit comma form being preferable is (iirc) because it’s basically impossible to know which names should be considered "first" and which "last"

CITATION.bib
```diff
@@ -1,6 +1,8 @@
-@misc{Qiskit,
-    author = {{Qiskit contributors}},
-    title = {Qiskit: An Open-source Framework for Quantum Computing},
-    year = {2023},
-    doi = {10.5281/zenodo.2573505}
+@misc{javadiabhari2024quantum,
```
1ucian0: Maybe keep the same citekey?
```suggestion
@misc{qiskit2024,
```

1ucian0: Also, I saw `techreport` used for this kind of white papers. Not sure if it makes any difference.

```suggestion
@techreport{javadiabhari2024quantum,
```

jakelishman: should probably be at least `Qiskit2024` or something by default - it's a pretty non-zero chance that "the Qiskit developers" will publish another paper at some point, but this is just the citation key anyway; it doesn't show up anywhere except in a user's TeX source

sbrandhsn: I would maybe even suggest to call it something like `Qiskit_SDK_2024` in case there will be another paper this year and in case we may want to list all 'Qiskit-papers' in that file at some point in the future.

mtreinish: I went with `qiskit2024` for simplicity.

1ucian0 (approved):

PR #12418, opened 2024-05-16
Add convert data

alexanderivrii:
### Summary

When converting from a `DAGCircuit` to a `DAGDependency`, in some cases it would be useful to compute and return additional information related to this conversion, such as the mappings between the input `DAGOpNode`s and the output `DAGDepNode`s. In particular, this would allow analysis passes that translate a `DAGCircuit` to a `DAGDependency`, collect blocks of nodes from the `DAGDependency` (exploiting the commutativity information offered by the latter structure), find the lists of corresponding nodes in the origin `DAGCircuit` and (for example) resynthesize these nodes. A direct application would be to extend the new `StarPreRouting` pass to collect and resynthesize "star-shaped" pieces of the circuit while also exploiting the commutativity between the nodes.

This PR adds a new class `CircuitConversionData` that stores the mappings and extends the conversion functions to `dag_to_dagdependency_with_data` and `_dag_to_dagdependency_v2_with_data` (all of the code pertaining to `DAGDependency_V2` is private for now).

The reason that `CircuitConversionData` is a class rather than a mapping is that we might want to store more information in the future, and we would not need to keep extending the API. And it would also probably be easier to convert this to Rust in the future. 

I am wondering whether I should make the class `CircuitConversionData` also private or also extend other circuit converters such as `circuit_to_dagcircuit`, etc. What do you think?

### DIFF
--- a/qiskit/converters/__init__.py
+++ b/qiskit/converters/__init__.py
@@ -23,23 +23,25 @@ QuantumCircuit -> circuit components
 .. autofunction:: circuit_to_instruction
 .. autofunction:: circuit_to_gate
 
-QuantumCircuit <-> DagCircuit 
+QuantumCircuit <-> DagCircuit
 =============================
 
 .. autofunction:: circuit_to_dag
 .. autofunction:: dag_to_circuit
 
-QuantumCircuit <-> DagDependency 
+QuantumCircuit <-> DagDependency
 ================================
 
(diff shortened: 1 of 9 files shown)


alexanderivrii:
Closing this (with this being really outdated after porting the main data structures to Rust).

PR #12428, opened 2024-05-17
Add merge queue to required tests on github actions

mtreinish:
### Summary

This commit adds the missing config to the github actions workflow for running required tests (currently only arm64 macOS test jobs) to the merge queue. This is necessary to make the job required in the branch protection rules, because the jobs will need to pass as part of the merge queue too.

### Details and comments

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -5,6 +5,7 @@ on:
     branches: [ main, 'stable/*' ]
   pull_request:
     branches: [ main, 'stable/*' ]
+  merge_group:
 
 concurrency:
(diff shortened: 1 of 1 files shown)


jakelishman (approved):
Looks fine to me.  Fwiw, [the docs for `merge_group`](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#merge_group) say that we could also add a `types: [ checks_requested ]` limiter to it, but at the moment, `checks_requested` is the _only_ type anyway.  If it expands in the future, we can decide if we need to be more limiting.

PR #12429, opened 2024-05-17
Fix `QuantumCircuit.depth` with zero-operands and `Expr` nodes

### LINKED ISSUE #12426: GlobalPhaseGate causing `circuit.depth()` error
### Environment

- **Qiskit version**: 1.0.2
- **Python version**: 3.11.9

### What is happening?

Greetings,

Hope all are well. I have noticed that when you apply the GlobalPhaseGate to a circuit, and then try to retrieve the depth of the circuit, it causes an error due to the GlobalPhaseGate instruction not using any qubit indices. I think this should be fixed inside the `.depth()` method as perhaps a special case.

```py
from qiskit.circuit.library import GlobalPhaseGate


qc_qiskit = QuantumCircuit(2, 2)

qc_qiskit.cx(0, 1)

transpiled_qc = transpile(qc_qiskit,
                          basis_gates=['u3', 'cx'],
                          optimization_level=3,
                          seed_transpiler=0)

transpiled_qc.append(GlobalPhaseGate(2.5928519981876477), [], [])

transpiled_qc.global_phase

transpiled_qc.depth()
```

This will output the following error:
```
ValueError                                Traceback (most recent call last)
Cell In[15], [line 19](vscode-notebook-cell:?execution_count=15&line=19)
     [15](vscode-notebook-cell:?execution_count=15&line=15) transpiled_qc.append(GlobalPhaseGate(2.5928519981876477), [], [])
     [17](vscode-notebook-cell:?execution_count=15&line=17) transpiled_qc.global_phase
---> [19](vscode-notebook-cell:?execution_count=15&line=19) transpiled_qc.depth()

File c:\Users\A.C.EA\OneDrive\Documents\GitHub\QICKIT\.venv\Lib\site-packages\qiskit\circuit\quantumcircuit.py:1909, in QuantumCircuit.depth(self, filter_function)
   [1906](file:///C:/Users/A.C.EA/OneDrive/Documents/GitHub/QICKIT/.venv/Lib/site-packages/qiskit/circuit/quantumcircuit.py:1906)             reg_ints.append(idx)
   [1907](file:///C:/Users/A.C.EA/OneDrive/Documents/GitHub/QICKIT/.venv/Lib/site-packages/qiskit/circuit/quantumcircuit.py:1907)             levels.append(op_stack[idx] + 1)
-> [1909](file:///C:/Users/A.C.EA/OneDrive/Documents/GitHub/QICKIT/.venv/Lib/site-packages/qiskit/circuit/quantumcircuit.py:1909) max_level = max(levels)
   [1910](file:///C:/Users/A.C.EA/OneDrive/Documents/GitHub/QICKIT/.venv/Lib/site-packages/qiskit/circuit/quantumcircuit.py:1910) for ind in reg_ints:
   [1911](file:///C:/Users/A.C.EA/OneDrive/Documents/GitHub/QICKIT/.venv/Lib/site-packages/qiskit/circuit/quantumcircuit.py:1911)     op_stack[ind] = max_level

ValueError: max() arg is an empty sequence
```

### What should happen?

Inside the `.depth()` method, the code should simply pass when the instruction is a GlobalPhaseGate. Since it's not, because GlobalPhaseGate has no qubit index as parameter, it causes the `levels` variable to be `[]`, thus raising the error.

### Any suggestions?

I tried this and it worked:
```py
def depth(
        self,
        filter_function: Callable[..., int] = lambda x: not getattr(
            x.operation, "_directive", False
        ),
    ) -> int:
        """Return circuit depth (i.e., length of critical path).

        Args:
            filter_function (callable): A function to filter instructions.
                Should take as input a tuple of (Instruction, list(Qubit), list(Clbit)).
                Instructions for which the function returns False are ignored in the
                computation of the circuit depth.
                By default filters out "directives", such as barrier or snapshot.

        Returns:
            int: Depth of circuit.

        Notes:
            The circuit depth and the DAG depth need not be the
            same.
        """
        # Assign each bit in the circuit a unique integer
        # to index into op_stack.
        bit_indices: dict[Qubit | Clbit, int] = {
            bit: idx for idx, bit in enumerate(self.qubits + self.clbits)
        }

        # If no bits, return 0
        if not bit_indices:
            return 0

        # A list that holds the height of each qubit
        # and classical bit.
        op_stack = [0] * len(bit_indices)

        # Here we are playing a modified version of
        # Tetris where we stack gates, but multi-qubit
        # gates, or measurements have a block for each
        # qubit or cbit that are connected by a virtual
        # line so that they all stacked at the same depth.
        # Conditional gates act on all cbits in the register
        # they are conditioned on.
        # The max stack height is the circuit depth.
        for instruction in self._data:
            levels = []
            reg_ints = []
            for ind, reg in enumerate(instruction.qubits + instruction.clbits):
                # Add to the stacks of the qubits and
                # cbits used in the gate.
                reg_ints.append(bit_indices[reg])
                if filter_function(instruction):
                    levels.append(op_stack[reg_ints[ind]] + 1)
                else:
                    levels.append(op_stack[reg_ints[ind]])
            # Assuming here that there is no conditional
            # snapshots or barriers ever.
            if getattr(instruction.operation, "condition", None):
                # Controls operate over all bits of a classical register
                # or over a single bit
                if isinstance(instruction.operation.condition[0], Clbit):
                    condition_bits = [instruction.operation.condition[0]]
                else:
                    condition_bits = instruction.operation.condition[0]
                for cbit in condition_bits:
                    idx = bit_indices[cbit]
                    if idx not in reg_ints:
                        reg_ints.append(idx)
                        levels.append(op_stack[idx] + 1)

            if getattr(instruction.operation, 'name') == 'global_phase':
                levels = [0]

            max_level = max(levels)
            for ind in reg_ints:
                op_stack[ind] = max_level

        return max(op_stack)
```

jakelishman:
### Summary

This causes `QuantumCircuit.depth` to correctly handle cases where a circuit instruction has zero operands (such as `GlobalPhaseGate`), and to treat classical bits and real-time variables used inside `Expr` conditions as part of the depth calculations.  This is in line with `DAGCircuit`.

This commit still does not add the same `recurse` argument from `DAGCircuit.depth`, because the arguments for not adding it to `QuantumCircuit.depth` at the time still hold; there is no clear meaning to it for general control flow from a user's perspective, and it was only added to the `DAGCircuit` methods because there it is more of a proxy for optimising over all possible inner blocks.

### Details and comments

Fix #12426.

### DIFF
--- a/qiskit/circuit/quantumcircuit.py
+++ b/qiskit/circuit/quantumcircuit.py
@@ -45,7 +45,7 @@ from qiskit.circuit.parameter import Parameter
 from qiskit.circuit.exceptions import CircuitError
 from . import _classical_resource_map
 from ._utils import sort_parameters
-from .controlflow import ControlFlowOp
+from .controlflow import ControlFlowOp, _builder_utils
 from .controlflow.builder import CircuitScopeInterface, ControlFlowBuilderBlock
 from .controlflow.break_loop import BreakLoopOp, BreakLoopPlaceholder
 from .controlflow.continue_loop import ContinueLoopOp, ContinueLoopPlaceholder
@@ -3307,6 +3307,9 @@ class QuantumCircuit:
     ) -> int:
         """Return circuit depth (i.e., length of critical path).
 
+        .. warning::
+            This operation is not well defined if the circuit contains control-flow operations.
+
         Args:
             filter_function: A function to decide which instructions count to increase depth.
                 Should take as a single positional input a :class:`CircuitInstruction`.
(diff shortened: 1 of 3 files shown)


qiskit/circuit/quantumcircuit.py
```diff
@@ -3307,6 +3307,9 @@ def depth(
     ) -> int:
         """Return circuit depth (i.e., length of critical path).
 
+        .. warning::
+            This operation is not well defined if the circuit contains control-flow operations.
```
ElePT, line 3311: would it make sense to add more info in this warning? would it be inaccurate to say, for example:

```suggestion
            This operation is not well defined if the circuit contains control-flow operations, but 
            their variables and clbits are counted in the depth calculation.
            
```

ElePT (commented):
LGTM! I only have one minor comment.

qiskit/circuit/quantumcircuit.py
jakelishman: I was mostly deliberately trying to avoid saying _anything_ about what happens in the return value of this if there's control-flow operations, since the whole result is ill defined.  I didn't want to commit to anything one way or the other.

ElePT: I suspected that was the case, and I think that makes sense too.

ElePT (approved):

PR #12446, opened 2024-05-20
 Remove the duplicated docs for a `BackendV1` classmethod (backport #12443)

arnaucasau:
The `BackendV1` [page](https://docs.quantum.ibm.com/api/qiskit/qiskit.providers.BackendV1#_default_options) has a duplicated entry for the `_default_options()` method as we can see in this screenshot highlight it with two red arrows.

323689526-2d78f5f6-ab3f-402c-a53d-d684078c208f

This PR removes the duplicated autoclass directive for `_default_options()` and the line of documentation explaining that subclasses need to implement it, as it is also explained in line 66. 

https://github.com/Qiskit/qiskit/blob/a83c99b7f8f0bf03dd2513e7410fc1575e30d284/qiskit/providers/backend.py#L66-L69

With this removal, the docs will match the documentation of `BackendV2` where we use the same explanation once:

https://github.com/Qiskit/qiskit/blob/a83c99b7f8f0bf03dd2513e7410fc1575e30d284/qiskit/providers/backend.py#L316-L319

The change will need backport to `stable/1.1`, `stable/1.0`, and `stable/0.46`

### DIFF
--- a/qiskit/providers/backend.py
+++ b/qiskit/providers/backend.py
@@ -88,12 +88,6 @@ class BackendV1(Backend, ABC):
             This next bit is necessary just because autosummary generally won't summarise private
             methods; changing that behaviour would have annoying knock-on effects through all the
             rest of the documentation, so instead we just hard-code the automethod directive.
-
-        In addition to the public abstract methods, subclasses should also implement the following
-        private methods:
-
-        .. automethod:: _default_options
-           :noindex:
         """
         self._configuration = configuration
         self._options = self._default_options()


jakelishman (approved):
Fwiw, you don't need to backport manually.  If you want a PR backporting, we just ask the mergify bot to do it; if it's going the 0.x LTS branch then we do it with a direct command comment rather than the label (which causes it to target the current backport branch), but the bot will still manage things.

arnaucasau:
Oh, good to know, that will save me time for the next one, thanks Jake!

jakelishman:
I'm not sure what permissions you need to do it, but the command is
```
```
You can do more branches as well with (e.g.)
```
@Mergifyio backport stable/0.46 stable/1.4 stable/2.4
```
but since the bot takes care of the `stable/1.1` bit automatically if we label the issue "stable backport potential", we usually only need to manually specify `stable/0.46`.  You can try it on your next issue if you like - we can see if your permissions do it (I think they will).

arnaucasau:
That's awesome!

> You can try it on your next issue if you like - we can see if your permissions do it (I think they will).

Sounds good, thanks again!

jakelishman:
I think you have to put the mergify command in a command all by itself, as in by copy-pasting my first command there into a comment with no other text.  But it also might have got smarter about things since the last time I tried to put the command in with other text.

PR #11565, opened 2024-01-15
Remove execute() function

1ucian0:
### Summary

Removal of `execute` which was deprecated in 0.46 https://github.com/Qiskit/qiskit/pull/11044

PR #12448, opened 2024-05-20
Remove reference to `transpile` from Pulse docs

### LINKED ISSUE #12447: No support for transpiling `Schedule`
### Environment

- **Qiskit version**: 1.1.0
- **Python version**: 3.11
- **Operating system**: macos

### What is happening?

[Documentation](https://docs.quantum.ibm.com/api/qiskit/pulse) says that is possible to run a pulse level program by calling `backend.run(transpile(pulse_prog, backend))`. However, that is not yet supported.

Just run the example in the documentation 

```python
from qiskit import pulse, transpile
 
d0 = pulse.DriveChannel(0)
 
with pulse.build() as pulse_prog:
    pulse.play(pulse.Constant(100, 1.0), d0)

transpile(pulse_prog, backend)
```

### What should happen?

transpile successfully 

### Any suggestions?

_No response_

jakelishman:
### Summary

This reference to `transpile` was mistakenly inserted as part of 1a027ac (gh-11565); `execute` used to ignore the `transpile` step for `Schedule`/`ScheduleBlock` and submit directly to `backend.run`.

I am not clear if there are actually any backends that *accept* pulse jobs anymore, but if there are, this is what the text should have been translated to.

### Details and comments

It's not at all clear to me how accurate this documentation remains at all (if nothing else, doesn't the pulse builder return `ScheduleBlock` now?), but this is at least a fix for the change in #11565.

Fix #12447.

### DIFF
--- a/qiskit/pulse/builder.py
+++ b/qiskit/pulse/builder.py
@@ -74,8 +74,8 @@ a pulse:
 
 The builder initializes a :class:`.pulse.Schedule`, ``pulse_prog``
 and then begins to construct the program within the context. The output pulse
-schedule will survive after the context is exited and can be transpiled and executed like a
-normal Qiskit schedule using ``backend.run(transpile(pulse_prog, backend))``.
+schedule will survive after the context is exited and can be executed like a
+normal Qiskit schedule using ``backend.run(pulse_prog)``.
 
 Pulse programming has a simple imperative style. This leaves the programmer
 to worry about the raw experimental physics of pulse programming and not
--- a/qiskit/pulse/builder.py
+++ b/qiskit/pulse/builder.py
@@ -74,8 +74,8 @@ a pulse:
 
 The builder initializes a :class:`.pulse.Schedule`, ``pulse_prog``
 and then begins to construct the program within the context. The output pulse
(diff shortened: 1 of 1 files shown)


qiskit/pulse/builder.py
```diff
@@ -74,8 +74,8 @@
 
 The builder initializes a :class:`.pulse.Schedule`, ``pulse_prog``
 and then begins to construct the program within the context. The output pulse
-schedule will survive after the context is exited and can be transpiled and executed like a
-normal Qiskit schedule using ``backend.run(transpile(pulse_prog, backend))``.
+schedule will survive after the context is exited and can be executed like a
+normal Qiskit schedule using ``backend.run(pulse_prog)``.
```
wshanks: The providers have not supported executing pulse schedules for a while now. They only support attaching schedules to circuits as calibrations. Since pulse gates are not talked about in this section, that might be too much to reference right here. Maybe just this?

```suggestion
schedule will survive after the context is exited and can be used like a
normal Qiskit schedule.
```

wshanks (approved):

PR #12456, opened 2024-05-23
binary matrices utils in rust

### LINKED ISSUE #12330: Move binary linear matrices utils to rust
Transfer to rust binary matrices utils from:
https://github.com/Qiskit/qiskit/blob/main/qiskit/synthesis/linear/linear_matrix_utils.py.
 
This is needed for many linear functions and clifford synthesis methods: 
#12238 
#12237 
#12228 
#12242 
#12231

ShellyGarion:
### Summary
close #12330

Transfer to rust binary matrices utils from:
https://github.com/Qiskit/qiskit/blob/main/qiskit/synthesis/linear/linear_matrix_utils.py.

joint with @alexanderivrii


sbrandhsn:
I'm excited for this PR! Can you add a graph highlighting the runtime speedup of your Rust code compared to the old Python code? I don't think this graph would need to be comprehensive but it would be nice to get a clearer picture for some relevant procedures. :-)

crates/accelerate/src/linear_matrix.rs
```diff
@@ -0,0 +1,111 @@
...
+                    _row_op(&mut mat, r, i);
+                }
+            }
+        }
+
+        for i in r + 1..m {
+            if mat[(i, new_k)] == 1 {
+                _row_op(&mut mat, r, i);
+            }
+        }
```
mtreinish: You can replace this block with something like:

```suggestion
        // Copy source row to avoid trying multiple borrows at once
        let row0 = mat.row(r).to_owned();
        if full_elim.is_some() && full_elim == Some(true) {
            mat.axis_iter_mut(Axis(0))
                .into_par_iter()
                .enumerate()
                .filter(|(idx, _ax)| *idx >= r)
                .for_each(|(_i, mut row)| {
                    row.zip_mut_with(&row0, |x, &y| *x ^= y);
                });
        }

        mat.axis_iter_mut(Axis(0))
            .into_par_iter()
            .enumerate()
            .filter(|(idx, _ax)| *idx < r + 1 || *idx >= m)
            .for_each(|(_i, mut row)| {
                row.zip_mut_with(&row0, |x, &y| *x ^= y);
            });
```

to do a parallel row replacement. This will require adding `use rayon::prelude::*` to the top. Also instead of unconditionally using `into_par_iter()` we should condition parallelism using `getenv_use_multiple_threads()` like we do in other parallel functions. It's probably also worth benchmarking to see if parallelism actually helps performance in this case.

crates/accelerate/src/linear_matrix.rs
```diff
@@ -0,0 +1,111 @@
...
+    full_elim: Option<bool>,
+) -> Array2<i8> {
+    let (m, mut n) = (mat.nrows(), mat.ncols()); // no. of rows and columns
+    if let Some(ncols_val) = ncols {
+        n = usize::min(n, ncols_val); // no. of active columns
+    }
+
+    let mut r = 0; // current rank
+    let k = 0; // current pivot column
+    let mut new_k = 0;
```
mtreinish: I don't think `new_k` is being used by anything, it's written to in L51 and L55 but nothing ever accesses the value.

mtreinish (commented):
Thanks for doing this, it's a great start. I left a couple inline comments for some suggestions. The larger comment shows how you can leverage rayon for multithreading part of the implementation. But we should benchmark it to make sure that multithreading actually speeds it up in this case. I could see the benefits from it not being huge here. In that suggestion you can just remove the `.into_par_iter()` line and that will make the iterator chain serially execute instead of doing it in parallel.

crates/accelerate/src/linear_matrix.rs
ShellyGarion: `new_k` is used in L55. If I write there `k=j` then I get a lint error (since `k` appears in the range of the for loop).

alexanderivrii:
Shelly and I were experimenting with this code trying to understand how to modify matrix elements in-place (rather than copying the matrix from Python space to Rust space and back), and this is what we currently have.

The wrapper function looks as follows:

```rust
#[pyfunction]
#[pyo3(signature = (mat, ncols=None, full_elim=false))]
fn _gauss_elimination(
    _py: Python,
    mat: &PyArray2<bool>,
    ncols: Option<usize>,
    full_elim: Option<bool>,
) {
    unsafe {
        let view = mat.as_array_mut();
        gauss_elimination(view, ncols, full_elim);
    }
}
```
This seems to work, producing the expected results, yet there are a few points we are not sure about.

1. We are using `&PyArray2` rather than `PyReadwriteArray2` that was suggested elsewhere. How can we be absolutely certain that the memory is modified in-place rather than copied?

2. `cargo build` complains that `&PyArray2` is deprecated:
```
warning: use of deprecated method `pyo3::deprecations::GilRefs::<T>::function_arg`: use `&Bound<'_, T>` instead for this function argument
```
however replacing `&PyArray2<bool>` by something like `mat: &Bound<PyArray2<bool>>` or `mat: &Bound<'_PyArray2<bool>>` does not compile, with the error
```
no method named `as_array_mut` found for reference `&pyo3::Bound<'_, PyArray<bool, Dim<[usize; 2]>>>` in the current scope 
```

3. The code from the python space must be careful to use the right type for the entries, e.g. this works:
```
mat = rng.integers(2, size=(num_qubits, num_qubits)).astype(bool)
_gauss_elimination(mat)
```
but this does not:
```
mat = rng.integers(2, size=(num_qubits, num_qubits))
_gauss_elimination(mat)
```
giving the error
```
TypeError: argument 'mat': type mismatch:
 from=int64, to=bool
```
which is probably to be expected.

crates/accelerate/src/linear_matrix.rs
ShellyGarion: I think that in your suggestion there is no check that `mat[i,new_k] == 1`. @alexanderivrii and I fixed it and will push a commit.

mtreinish:
If you want to make the rust function mutate the input numpy array this isn't too difficult. You need to change the types on things and that might require input normalization in the python side. I did a quick test locally and you would need to change the PR like:

```diff
diff --git a/crates/accelerate/src/linear_matrix.rs b/crates/accelerate/src/linear_matrix.rs
index d708e71bd..197c13c3e 100644
--- a/crates/accelerate/src/linear_matrix.rs
+++ b/crates/accelerate/src/linear_matrix.rs
@@ -10,19 +10,19 @@
 // copyright notice, and modified files need to carry a notice indicating
 // that they have been altered from the originals.
 
-use ndarray::{s, Array2};
-use numpy::{AllowTypeChange, IntoPyArray, PyArray2, PyArrayLike2};
+use ndarray::{s, ArrayViewMut2};
+use numpy::PyReadwriteArray2;
 use pyo3::prelude::*;
 
 // Perform ROW operation on a matrix mat
-fn _row_op(mat: &mut Array2<i8>, ctrl: usize, trgt: usize) {
+fn _row_op(mat: &mut ArrayViewMut2<i8>, ctrl: usize, trgt: usize) {
     let row0 = mat.row(ctrl).to_owned();
     let mut row1 = mat.row_mut(trgt);
     row1.zip_mut_with(&row0, |x, &y| *x ^= y);
 }
 
 // Perform COL operation on a matrix mat
-fn _col_op(mat: &mut Array2<i8>, ctrl: usize, trgt: usize) {
+fn _col_op(mat: &mut ArrayViewMut2<i8>, ctrl: usize, trgt: usize) {
     let col0 = mat.column(ctrl).to_owned();
     let mut col1 = mat.column_mut(trgt);
     col1.zip_mut_with(&col0, |x, &y| *x ^= y);
@@ -31,11 +31,7 @@ fn _col_op(mat: &mut Array2<i8>, ctrl: usize, trgt: usize) {
 // Gauss elimination of a matrix mat with m rows and n columns.
 // If full_elim = True, it allows full elimination of mat[:, 0 : ncols]
 // Returns the matrix mat.
-fn gauss_elimination(
-    mut mat: Array2<i8>,
-    ncols: Option<usize>,
-    full_elim: Option<bool>,
-) -> Array2<i8> {
+fn gauss_elimination(mut mat: ArrayViewMut2<i8>, ncols: Option<usize>, full_elim: Option<bool>) {
     let (m, mut n) = (mat.nrows(), mat.ncols()); // no. of rows and columns
     if let Some(ncols_val) = ncols {
         n = usize::min(n, ncols_val); // no. of active columns
@@ -62,7 +58,7 @@ fn gauss_elimination(
             }
         }
         if !is_non_zero {
-            return mat; // A is in the canonical form
+            return; // A is in the canonical form
         }
 
         if new_r != r {
@@ -87,21 +83,17 @@ fn gauss_elimination(
         }
         r += 1;
     }
-    mat
 }
 
 #[pyfunction]
 #[pyo3(signature = (mat, ncols=None, full_elim=false))]
 fn _gauss_elimination(
-    py: Python,
-    mat: PyArrayLike2<i8, AllowTypeChange>,
+    mut mat: PyReadwriteArray2<i8>,
     ncols: Option<usize>,
     full_elim: Option<bool>,
-) -> PyResult<Py<PyArray2<i8>>> {
-    let view = mat.as_array().to_owned();
-    Ok(gauss_elimination(view, ncols, full_elim)
-        .into_pyarray_bound(py)
-        .unbind())
+) {
+    let view = mat.as_array_mut();
+    gauss_elimination(view, ncols, full_elim);
 }
 
 #[pymodule]
```

The tradeoff here is that you'll probably need to do a `np.asarray(dtype=np.int8)` call in python to ensure the input array is typed with `i8` as it will error if it's not typed correctly.

crates/accelerate/src/linear_matrix.rs
ShellyGarion: done in [678ec0a](https://github.com/Qiskit/qiskit/pull/12456/commits/678ec0aed9b06c0d082875ee28451a8745dc2e0d)

alexanderivrii:
Thanks, @mtreinish, your code using `PyReadwriteArray2` works perfectly! As far as I can tell, the difference between this and what we tried together with Shelly is that you have also removed the `py: Python` argument. I don't really understand when this argument should or should not be present. 

Regarding `np.asarray(dtype=np.int8)`, doesn't this copy the array?

mtreinish:
> Regarding np.asarray(dtype=np.int8), doesn't this copy the array?

It will make a copy if the dtype doesn't match the input array's type: https://numpy.org/doc/stable/reference/generated/numpy.asarray.html. 

The copy is needed to convert the elements to the specified dtype and would be unavoidable in this case. The previous `PyArrayLike2<>` usage was actually doing this for you because you had the `AllowTypeChange` argument: https://docs.rs/numpy/latest/numpy/struct.PyArrayLike.html

crates/accelerate/src/linear_matrix.rs
mtreinish: Ah my bad I missed that in my suggestion and you're right I was looking at this in my local editor after testing: https://github.com/Qiskit/qiskit/pull/12456#discussion_r1625887372

alexanderivrii:
A very quick experiment _on my laptop_:
```
for num_qubits in [50, 100, 250, 500, 1000, 2500]:
    rng = np.random.default_rng(0)
    mats = [rng.integers(2, size=(num_qubits, num_qubits)).astype(bool) for _ in range(10)]

    time_start = time.time()

    for mat in mats:
        _gauss_elimination(mat)

    time_end = time.time()
    print(f"{num_qubits = }, time = {time_end-time_start:.2f}")
```
Old version (python/numpy):
```
num_qubits = 50, time = 0.02
num_qubits = 100, time = 0.05
num_qubits = 250, time = 0.39
num_qubits = 500, time = 1.89
num_qubits = 1000, time = 7.56
num_qubits = 2500, time = 66.32
```
New code (ref. https://github.com/Qiskit/qiskit/pull/12456/commits/0cf7502425b66426877fe845d6c6c6a90bacf531) but without using parallel iterators:
```
num_qubits = 50, time = 0.00
num_qubits = 100, time = 0.01
num_qubits = 250, time = 0.08
num_qubits = 500, time = 0.63
num_qubits = 1000, time = 5.11
num_qubits = 2500, time = 35.89
```
New code (ref. https://github.com/Qiskit/qiskit/pull/12456/commits/0cf7502425b66426877fe845d6c6c6a90bacf531) with using parallel iterators:
```
num_qubits = 50, time = 0.01
num_qubits = 100, time = 0.02
num_qubits = 250, time = 0.15
num_qubits = 500, time = 0.69
num_qubits = 1000, time = 2.99
num_qubits = 2500, time = 34.27
```

alexanderivrii:
The Rust code is compiled in release mode (without it it's significantly slower).

alexanderivrii:
> The copy is needed to convert the elements to the specified dtype and would be unavoidable in this case. The previous PyArrayLike2<> usage was actually doing this for you because you had the AllowTypeChange argument: https://docs.rs/numpy/latest/numpy/struct.PyArrayLike.html

Thank you! In terms of the preferred API, would it be better to be more general on the Rust side (i.e. restore `AllowTypeChange`) or more restrictive on the Rust side (i.e. only allow arrays of type `bool` as per the latest version)?

mtreinish:
I typically prefer doing the type normalization in Python directly. I think in this case that's preferable because you avoid a copy if the input is the correct dtype. While in the previous implementation you always needed to copy to get a mutable array.

crates/accelerate/src/linear_matrix.rs
```diff
@@ -0,0 +1,111 @@
...
+                });
+        }
+
+        mat.axis_iter_mut(Axis(0))
+            .into_par_iter()
+            .enumerate()
+            .filter(|(i, row)| (*i > r) && (*i < m) && row[new_k])
+            .for_each(|(_i, mut row)| {
+                row.zip_mut_with(&row0, |x, &y| *x ^= y);
+            });
```
mtreinish: This can be simplified to a single iterator with something like (double check that I got the filter condition correct:

```suggestion
        mat.axis_iter_mut(Axis(0))
            .into_par_iter()
            .enumerate()
            .filter(|(i, row)| (full_elim == Some(true)  && (*i < r) && row[new_k]) || (*i > r && *i < m))
            .for_each(|(_i, mut row)| {
                row.zip_mut_with(&row0, |x, &y| *x ^= y);
            });
```

You can also move the `full_elim` condition into the `for_each` block and just change the filter to be `i < m` if that's easier.

mtreinish: The other thing I was thinking about is you can probably get a slice view of the rows you care about and iterate over that slice instead of the full matrix. It's might be worth investigating that instead of doing a single big iterator like this. TBH, I'm not sure what is going to be the most efficient here we need to benchmark all the different possible ways to write this and figure out what ends up being the best performing.

crates/accelerate/src/linear_matrix.rs
```diff
@@ -0,0 +1,104 @@
...
+    row1.zip_mut_with(&row0, |x, &y| *x ^= y);
+}
+
+// Perform COL operation on a matrix mat
+fn _col_op(mat: &mut ArrayViewMut2<bool>, ctrl: usize, trgt: usize) {
+    let col0 = mat.column(ctrl).to_owned();
+    let mut col1 = mat.column_mut(trgt);
+    col1.zip_mut_with(&col0, |x, &y| *x ^= y);
+}
+
```
mtreinish: In the current version this is no longer called by anything.

```suggestion
```

crates/accelerate/src/linear_matrix.rs
```diff
@@ -0,0 +1,104 @@
...
+            .enumerate()
+            .filter(|(i, row)| {
+                (full_elim == Some(true) && (*i < r) && row[new_k])
+                    || (*i > r && *i < m && row[new_k])
+            })
+            .for_each(|(_i, mut row)| {
+                row.zip_mut_with(&row0, |x, &y| *x ^= y);
+            });
+
+        r += 1;
```
mtreinish: Besides the open question around whether parallelism actually improves performance here and also the open question about the an iterator vs a for loop too. If we do end up using parallelism we should add a guard here to prevent running in multiple threads at the same time qiskit is using multiple processes. We don't want to spawn a threadpool if Qiskit is running in multiple processes.

You can see the pattern we use for this in other places: https://github.com/Qiskit/qiskit/blob/main/crates/accelerate/src/pauli_exp_val.rs#L186-L192 

The key is to check `getenv_use_multiple_threads()` to get a boolean value based on Qiskit's environment variables whether we're in a multiprocessing context currently or not. If we are then we should not include the `into_par_iter()` otherwise it's safe to do.

crates/accelerate/src/linear_matrix.rs
ShellyGarion: these  functions are not needed for Gaussian elimination, however, they will be needed later for [KMS synthesis](https://docs.quantum.ibm.com/api/qiskit/synthesis#synth_cnot_depth_line_kms)

alexanderivrii:
I have seen that some of our code (e.g. in `sparse_pauli_ops.rs`) packs vectors of booleans into vectors of `u64`s. Do you think this is something worth considering when working with really large matrices?

crates/accelerate/src/linear_matrix.rs
ShellyGarion: Indeed, after benchmarking the code on a linux VM with 8 cores, it seems that doing multiple threading only makes it worse, so I removed it for now.

Original python code:
```
num_qubits = 50, time = 0.02
num_qubits = 100, time = 0.06
num_qubits = 250, time = 0.38
num_qubits = 500, time = 1.68
num_qubits = 1000, time = 8.08
num_qubits = 2500, time = 83.79
```
rust code (single thread):
```
num_qubits = 50, time = 0.00
num_qubits = 100, time = 0.00
num_qubits = 250, time = 0.03
num_qubits = 500, time = 0.20
num_qubits = 1000, time = 2.18
num_qubits = 2500, time = 29.95
```
rust code (multiple threads):
```
num_qubits = 50, time = 0.30
num_qubits = 100, time = 0.53
num_qubits = 250, time = 2.75
num_qubits = 500, time = 8.72
num_qubits = 1000, time = 22.15
num_qubits = 2500, time = 140.68 
```

crates/accelerate/src/linear_matrix.rs
Cryoris: Maybe in KMS synthesis we can also use closures, like you now used below. If it's not needed in this PR I'd also say removing them is better, we can still add them later if we really need them 🙂

crates/accelerate/src/linear_matrix.rs
```diff
@@ -0,0 +1,102 @@
...
+}
+
+// Perform COL operation on a matrix mat
+fn _col_op(mat: &mut ArrayViewMut2<bool>, ctrl: usize, trgt: usize) {
+    let col0 = mat.column(ctrl).to_owned();
+    let mut col1 = mat.column_mut(trgt);
+    col1.zip_mut_with(&col0, |x, &y| *x ^= y);
+}
+
+// Gauss elimination of a matrix mat with m rows and n columns.
```
Cryoris: As a docstring, should this have 3 backslashes? Is a non-full elimination only the lower triangular elimination? If yes it might be nice to mention it explicitly

crates/accelerate/src/linear_matrix.rs
```diff
@@ -0,0 +1,102 @@
...
+// If full_elim = True, it allows full elimination of mat[:, 0 : ncols]
+// Returns the matrix mat.
+fn gauss_elimination(mut mat: ArrayViewMut2<bool>, ncols: Option<usize>, full_elim: Option<bool>) {
```
Cryoris: No comment, just a question: why is the `mut` needed here? Don't we already have a mutable reference because of the `ArrayViewMut` type?

Cryoris (commented):
Some small comments/questions 🙂

Cryoris:
> I have seen that some of our code (e.g. in sparse_pauli_ops.rs) packs vectors of booleans into vectors of u64s. Do you think this is something worth considering when working with really large matrices?

That would limit us to 64 bits only, but we likely want to deal with bigger systems, no?

ShellyGarion:
> > I have seen that some of our code (e.g. in sparse_pauli_ops.rs) packs vectors of booleans into vectors of u64s. Do you think this is something worth considering when working with really large matrices?
> 
> That would limit us to 64 bits only, but we likely want to deal with bigger systems, no?

The idea is to divide `n` qubits into `n/64` words of 64 bits each (pad with 0's if needed)

Cryoris:
Ah! Like a custom bit array. IMO if it's no bottleneck (which it doesn't look like right now) it might be a bit overkill, but good to keep in mind in case it becomes important 🤔

crates/accelerate/src/linear_matrix.rs
ShellyGarion: otherwise I get compilation errors: " cannot borrow as mutable"

crates/accelerate/src/linear_matrix.rs
ShellyGarion: Indeed, it's still a comment. I'll add docstrings when I'm done.

crates/accelerate/src/linear_matrix.rs
ShellyGarion: For now, I'll move the code to rust, as all the code in `linear_matrix_utils.py`, handling KMS linear synthesis will be done in a separate PR.

crates/accelerate/src/linear_matrix.rs
```diff
@@ -119,9 +119,23 @@ fn _gauss_elimination(
...
 }
 
+#[pyfunction]
+#[pyo3(signature = (mat))]
+// Given a boolean matrix A after Gaussian elimination, computes its rank
+// (i.e. simply the number of nonzero rows)"""
+fn _compute_rank_after_gauss_elim(py: Python, mat: PyReadonlyArray2<bool>) -> PyResult<PyObject> {
+    let view = mat.as_array();
+    let rank: usize = view
+        .axis_iter(Axis(0))
```
Cryoris: Just a tiny suggestion if you'd like to use it, I think this should be the same but shorter 🙂 
```suggestion
        .rows()
```

ShellyGarion: thanks, but your suggestion raises the following compilation error:
error[E0599]: `Lanes<'_, bool, Dim<[usize; 1]>>` is not an iterator

crates/accelerate/src/linear_matrix.rs
```diff
@@ -0,0 +1,264 @@
...
+    }
+
+    let mut result: Array2<bool> = Array2::from_shape_fn((n1_rows, n2_cols), |(_i, _j)| false);
+    for i in 0..n1_rows {
+        for j in 0..n2_cols {
+            for k in 0..n2_rows {
+                result[(i, j)] ^= mat1[(i, k)] & mat2[(k, j)];
+            }
+        }
+    }
```
alexanderivrii:
```suggestion
Array2::from_shape_fn((n1_rows, n2_cols), |(i, j)| {
        (0..n2_rows)
            .map(|k| mat1[[i, k]] & mat2[[k, j]])
            .fold(false, |acc, v| acc ^ v)
    })
```

alexanderivrii: This is about 1.5x faster. I have also experimented with calling `ndarray.dot` as follows
```
    let mat1_usize = mat1.map(|v| *v as usize);
    let mat2_usize = mat2.map(|v| *v as usize);
    mat1_usize.dot(&mat2_usize).map(|v| (*v % 2) == 1)
```
but this version had about the same runtime as your original version, probably because of needing to copy the matrices 3 times.

alexanderivrii: We should also avoid `panic!` and return `Result`

crates/accelerate/src/synthesis/linear/mod.rs
```diff
@@ -0,0 +1,192 @@
...
+/// Gauss elimination of a matrix mat with m rows and n columns.
+/// If full_elim = True, it allows full elimination of mat[:, 0 : ncols]
+/// Returns the updated matrix mat.
+/// Args:
+///     mat: a boolean matrix with n rows and m columns
+///     ncols: the number of columns for the gaussian elimination,
+///            if ncols=None, then the elimination is done over all the colums
+///     full_elim: whether to do a full elimination, or partial (upper triangular form)
+/// Returns:
+///     mat: the matrix after the gaussian elimination
```
Cryoris: It seems like the matrix is modified in-place and not returned 🙂 
```suggestion
/// This function modifies the input matrix in-place.
/// Args:
///     mat: a boolean matrix with n rows and m columns
///     ncols: the number of columns for the gaussian elimination,
///            if ncols=None, then the elimination is done over all the colums
///     full_elim: whether to do a full elimination, or partial (upper triangular form)
```

crates/accelerate/src/synthesis/linear/mod.rs
```diff
@@ -0,0 +1,192 @@
...
+/// Returns the matrix mat, and the permutation perm that was done on the rows during the process.
+/// perm[0 : rank] represents the indices of linearly independent rows in the original matrix.
+/// Args:
+///     mat: a boolean matrix with n rows and m columns
+///     ncols: the number of columns for the gaussian elimination,
+///            if ncols=None, then the elimination is done over all the colums
+///     full_elim: whether to do a full elimination, or partial (upper triangular form)
+/// Returns:
+///     mat: the matrix after the gaussian elimination
+///     perm: the permutation perm that was done on the rows during the process
```
Cryoris:
```suggestion
/// Modifies the matrix mat in-place, and returns the permutation perm that was done 
/// on the rows during the process. perm[0 : rank] represents the indices of linearly 
/// independent rows in the original matrix.
/// Args:
///     mat: a boolean matrix with n rows and m columns
///     ncols: the number of columns for the gaussian elimination,
///            if ncols=None, then the elimination is done over all the colums
///     full_elim: whether to do a full elimination, or partial (upper triangular form)
/// Returns:
///     perm: the permutation perm that was done on the rows during the process
```

crates/accelerate/src/synthesis/linear/utils.rs
```diff
@@ -0,0 +1,200 @@
...
+    perm
+}
+
+/// Given a boolean matrix A after Gaussian elimination, computes its rank
+/// (i.e. simply the number of nonzero rows)
+pub fn compute_rank_after_gauss_elim_inner(mat: ArrayView2<bool>) -> usize {
+    let rank: usize = mat
+        .axis_iter(Axis(0))
+        .map(|row| row.fold(false, |out, val| out | *val) as usize)
+        .sum();
```
Cryoris, line 110: Nothing you have to change, but I was thinking about how to "rustify" it even a bit more: instead of summing over the elements in a row, we only need to know whether any single element is `true`, right? Then we could also just use `.find` on the row, like
```rust
    let rank: usize = mat
        .axis_iter(Axis(0))
        .filter(|row| row  // filter on having a non-zero element
            .iter()
            .find(|&val| *val)  // find non-zero element
            .is_some()
        )
        .count();  // count remaining rows
```
and avoid doing a sum. I don't think this is a noticeable speedup for our system sizes, but I spent so much time trying that I still wanted to share it 😉

crates/accelerate/src/synthesis/linear/utils.rs
```diff
@@ -0,0 +1,200 @@
...
+                    break;
+                }
+            }
+            if is_non_zero {
+                break;
+            }
+        }
+        if !is_non_zero {
+            return perm; // A is in the canonical form
+        }
```
Cryoris, line 77: In the eternal quest to turn Qiskit into a 1-line Rust code, here's a suggestion on how to find the first non-zero element with iterators 🙂 As explanation: we take the relevant slice, enumerate over the rows, enumerate over the columns, and the use `find_map` to find the first non-zero element and return the corresponding index. Note that this only works because `find_map` returns the _first_ result that is not None (if it would return any result that is not None we couldn't be sure it's the first one).
```suggestion
    let index = mat
        .slice(s![r.., k..])  // slice with rows r..m and cols k..n
        .columns()
        .into_iter()
        .enumerate()
        .find_map(|(j, col)|  // find first col with nonzero term
            col
            .iter()
            .enumerate()
            .find_map(|(i, &el)|  // find first row with nonzero term
                if el { 
                    Some((i + r, j + k)) // map slice index to global matrix index
                } else {
                    None
                }
            )
        );
        if index.is_none() {
             // no nonzero index found, the matrix is in canonical form
            return perm; 
        }
        let (new_r, new_k) = index.unwrap();
```

crates/accelerate/src/synthesis/linear/utils.rs
```diff
@@ -0,0 +1,200 @@
...
+
+    let invmat = mat1.slice(s![.., n..2 * n]).to_owned();
+
+    if verify {
+        let mat2 = binary_matmul_inner(mat, (&invmat).into())?;
+        let identity_matrix: Array2<bool> = Array2::from_shape_fn((n, n), |(i, j)| i == j);
+        if mat2.ne(&identity_matrix) {
+            return Err("The inverse matrix is not correct.".to_string());
+        }
+    }
```
Cryoris, line 151: I understand that this was here before, but it seems very strange to essentially have a runtime test of our code here... if we provide the function then we should also be sure it is correct and not need to verify the result 😅 I'm fine keeping it because it existed before, but I would prefer removing this.

qiskit/synthesis/clifford/clifford_decompose_layers.py
```diff
@@ -33,9 +33,10 @@
 from qiskit.synthesis.linear_phase import synth_cz_depth_line_mr, synth_cx_cz_depth_line_my
 from qiskit.synthesis.linear.linear_matrix_utils import (
     calc_inverse_matrix,
-    _compute_rank,
-    _gauss_elimination,
-    _gauss_elimination_with_perm,
+    compute_rank,
+    gauss_elimination,
+    gauss_elimination_with_perm,
+    binary_matmul,
```
Cryoris, line 39: Is there a reason to make these utils public? Just thinking that it'll be easier to change if we keep them private 🙂

qiskit/quantum_info/operators/dihedral/random.py
```diff
@@ -53,7 +53,8 @@ def random_cnotdihedral(num_qubits, seed=None):
         random_invertible_binary_matrix,
     )
 
-    linear = random_invertible_binary_matrix(num_qubits, seed=rng)
+    seed = rng.integers(100000, size=1, dtype=np.uint64)[0]
+    linear = random_invertible_binary_matrix(num_qubits, seed=seed).astype(int)
```
mtreinish: This will always result in a copy. You could use `.astype(int, copy=False)` but it still would need to copy. Also I think the dtype is incorrect shouldn't this be int8 because that's what cnot dihedral is using: https://github.com/qiskit/qiskit/blob/main/qiskit/quantum_info/operators/dihedral/dihedral.py#L129.

That being said wouldn't it just make more sense to change all the binary matrices in rust to use `Array2<i8>` to match this? It'd be the same size as `Array2<bool>` but then you could do a no-copy return to cnotdihedral and anywhere else using these functions.

qiskit/synthesis/clifford/clifford_decompose_layers.py
```diff
@@ -267,7 +269,7 @@ def _decompose_graph_state(cliff, validate, cz_synth_func):
     """
 
     num_qubits = cliff.num_qubits
-    rank = _compute_rank(cliff.stab_x)
+    rank = compute_rank((cliff.stab_x).astype(bool))
```
mtreinish: Same comment as above
```suggestion
    rank = compute_rank((cliff.stab_x).astype(bool, copy=False))
```

qiskit/synthesis/linear/linear_depth_lnn.py
```diff
@@ -222,8 +222,8 @@ def _optimize_cx_circ_depth_5n_line(mat):
 
     # According to [1] the synthesis is done on the inverse matrix
     # so the matrix mat is inverted at this step
-    mat_inv = mat.copy()
-    mat_cpy = calc_inverse_matrix(mat_inv)
+    mat_inv = mat.copy().astype(bool)
```
mtreinish: This is doing 2 copies and you only need one. You can do that with:

```suggestion
    mat_inv = np.asarray(mat, copy=True, dtype=bool)
```

qiskit/synthesis/linear/linear_depth_lnn.py
```diff
@@ -222,8 +222,8 @@ def _optimize_cx_circ_depth_5n_line(mat):
...
+    mat_cpy = calc_inverse_matrix(mat_inv).astype(bool)
```
mtreinish: The return from rust is already of dtype bool, so you shouldn't need to cast anything. Do you need a copy here, if so I'd do this rust side instead. You can do that with `to_pyarray()` instead of `into_pyarray()`. Either way you can remove the `astype()` call here:

```suggestion
    mat_cpy = calc_inverse_matrix(mat_inv)
```

qiskit/synthesis/linear_phase/cx_cz_depth_lnn.py
```diff
@@ -245,6 +245,7 @@ def synth_cx_cz_depth_line_my(mat_x: np.ndarray, mat_z: np.ndarray) -> QuantumCi
     n = len(mat_x)
     mat_x = calc_inverse_matrix(mat_x)
 
+    mat_x = mat_x.astype(bool)
```
mtreinish: Same comment as before `calc_inverse_matrix` is returning a dtype bool array. Do you need to copy this here? If so I would do that from rust (especially since this would be a pattern of usage). But either way you can drop this line

```suggestion
````

File: qiskit/synthesis/stabilizer/stabilizer_decompose.py [hunk 21]
```diff
@@ -143,7 +143,7 @@ def _calc_pauli_diff_stabilizer(cliff, cliff_target):
     phase.extend(phase_stab)
     phase = np.array(phase, dtype=int)
 
-    A = cliff.symplectic_matrix.astype(int)
+    A = cliff.symplectic_matrix.astype(bool)
```
mtreinish: This is already of dtype bool in `Clifford` so we should avoid copying here unless we need to:
```suggestion
    A = cliff.symplectic_matrix.astype(bool, copy=False)
```

qiskit/transpiler/passes/synthesis/high_level_synthesis.py
```diff
@@ -779,7 +779,7 @@ def run(self, high_level_object, coupling_map=None, target=None, qubits=None, **
         use_inverted = options.get("use_inverted", False)
         use_transposed = options.get("use_transposed", False)
 
-        mat = high_level_object.linear.astype(int)
+        mat = high_level_object.linear.astype(bool)
```
mtreinish: This is already dtype bool in `LinearFunction` so we should avoid a copy here:

```suggestion
        mat = high_level_object.linear.astype(bool, copy=False)
```

qiskit/transpiler/passes/synthesis/high_level_synthesis.py
```diff
@@ -831,7 +831,7 @@ def run(self, high_level_object, coupling_map=None, target=None, qubits=None, **
...
```
mtreinish:
```suggestion
        mat = high_level_object.linear.astype(bool, copy=False)
```

qiskit/synthesis/clifford/clifford_decompose_layers.py
```diff
@@ -203,24 +204,25 @@ def _create_graph_state(cliff, validate=False):
...
```
mtreinish: You should do this to avoid a copy since the stabilizer slice should be of `dtype=bool` already.
```suggestion
    rank = compute_rank((cliff.stab_x).astype(bool, copy=False))
```

That being said I typically prefer to do:

```suggestion
    rank = compute_rank(np.asarray(cliff.stab_x, dtype=bool))
```

because it defaults to no copy behavior if all the data is aligned.

qiskit/synthesis/clifford/clifford_decompose_layers.py
```diff
@@ -203,24 +204,25 @@ def _create_graph_state(cliff, validate=False):
...
     H1_circ = QuantumCircuit(num_qubits, name="H1")
     cliffh = cliff.copy()
 
     if rank < num_qubits:
         stab = cliff.stab[:, :-1]
-        stab = _gauss_elimination(stab, num_qubits)
+        stab = stab.astype(bool)
```
mtreinish: Same comment as above:

```suggestion
        stab = stab.astype(bool, copy=False)
```

mtreinish (commented):
I only took a quick look at the numpy usage in this PR. I feel like there are a lot of extra copies happening which we can mostly avoid I think. Although I haven't reviewed the rust code in depth I was leaving that for @kevinhartman (as I think he volunteered to review it). As we're accelerating this code extra copies will have a real noticeable overhead so we should be intentional around when we copy.

qiskit/synthesis/linear/linear_matrix_utils.py
```diff
@@ -12,164 +12,16 @@
 
 """Utility functions for handling binary matrices."""
 
-from typing import Optional, Union
-import numpy as np
-from qiskit.exceptions import QiskitError
-
-
-def check_invertible_binary_matrix(mat: np.ndarray):
```
Cryoris, line 20: These functions used to accept also integer matrices, but now strictly require boolean types. Since these were not marked as private I'm not sure if we can change this -- maybe yes because they didn't appear in the API docs as far as I can tell 🤔 alternatively we could keep the Python function and normalize the input before using the Rust function, like
```python
def check_invertible_binary_matrix(mat):
    return _rust_implementation(mat.astype(bool))
```

releasenotes/notes/linear-binary-matrix-utils-rust-c48b5577749c34ab.yaml
```diff
@@ -0,0 +1,8 @@
+---
+features_synthesis:
+  - |
+    Port internal binary matrix utils from Python to rust, including
```
Cryoris: Is it `rust` or `Rust`? 🙂 Also depending on how we remain for the input types of the function, we should mention if only boolean matrices are supported now.

test/python/transpiler/test_high_level_synthesis.py
```diff
@@ -534,22 +534,22 @@ def test_section_size(self):
             hls_config = HLSConfig(linear_function=[("pmh", {"section_size": 1})])
             qct = HighLevelSynthesis(hls_config=hls_config)(qc)
             self.assertEqual(LinearFunction(qct), LinearFunction(qc))
-            self.assertEqual(qct.size(), 22)
-            self.assertEqual(qct.depth(), 20)
+            self.assertEqual(qct.size(), 30)
```
Cryoris, line 537: Do these change due to a different random matrix? Did you check by chance that with the old Python implementation of `random_invertible_binary_matrix` the results are still the same?

Cryoris (commented):
Overall this looks good, and I'm eager to merge it! Below are some questions and comments, my main concern being on whether we handle the functions as API-documented (meaning we keep the signature the same) or whether we can change them as if they were private 🙂

qiskit/quantum_info/operators/dihedral/random.py
alexanderivrii: The original motivation for rust binary matrices to be of type `Array2<bool>` is that python classes `Clifford` and `LinearFunction` internally store a matrix of booleans, and we wanted to avoid copying there. IMHO, the dihedral class is less important then either of the above (at least right now). However, I do agree that the python synthesize size needs to be cleaned up.

ShellyGarion: I agree with @alexanderivrii . As follows form the function name, the aim is only to handle binary matrices (e.g. bool).
In the future, the plan is to port the synthesis code to rust, and so we can handle the other classes later.

qiskit/synthesis/clifford/clifford_decompose_layers.py
ShellyGarion: if I change this line, it makes the tests in `test_cifford_decompose_layers` fail

mtreinish: Looking more closely at the code this is because `gauss_elimination()` is mutating the array in place and this will be a reference to the slice of the input clifford, so making this suggestion is causing the input clifford to be modified. The goal of this `astype()` call then is to make a copy. I think you should make that explicit since the intent isn't super clear with using `astype()` to do the copy. I'd suggest just use `stab.copy()` or `np.asarray(stab, copy=True, dtype=bool)`.

qiskit/synthesis/clifford/clifford_decompose_layers.py
ShellyGarion: these functions are not part of qiskit API. the reason is that we would like to reuse them for various synthesis algorithms and for tests

qiskit/synthesis/linear/linear_matrix_utils.py
ShellyGarion: by their name and also the docstring below, these functions are for Boolean matrices

test/python/transpiler/test_high_level_synthesis.py
ShellyGarion: these tests are based on fixed seed, and so the numbers were changed since the random function is different in Python and Rust

crates/accelerate/src/synthesis/linear/utils.rs
ShellyGarion: There is a parameter "verify" in several parts of the synthesis code, which is mainly used for tests, since the algorithms are quite complicated (the default value is False)

crates/accelerate/src/synthesis/linear/utils.rs
Cryoris: Actually, this ends up being slower than the for-loops -- nevermind then!

qiskit/synthesis/linear/linear_matrix_utils.py
Cryoris: I'm good with the stricter input type, I was just concerned that they might be public -- but if they're not I'm good with the change 👍🏻

crates/accelerate/src/synthesis/linear/utils.rs
Cryoris: We can leave it for this PR 👍🏻 but a common subroutine like matrix inversion seems like something stable enough that we don't need to bake in a verification and instead just have some tests in the test suite (fwiw other packages like NumPy/SciPy also don't have any arguments to check that 🙂).

crates/accelerate/src/synthesis/linear/utils.rs
Cryoris: After some benchmarking it, this is actually slower than what you already had. So nevermind this as well, seems like my "rustification" suggestions are not very good 😂

Cryoris (approved):
LGTM thanks for the effort! 🦀

Issue #12089, opened 2024-03-27
justify keyword arg in circuit draw does not raise on bad input


nonhermitian:
### Environment

- **Qiskit version**:  latest
- **Python version**: 
- **Operating system**: 

### What is happening?

`justify` will not raise on bogus inputs, e.g.

```python
qc.draw('mpl',justify='midldsfasdfse')
```

### How can we reproduce the issue?

Try above

### What should happen?

Bad inputs should raise rather than default implicitly

Waliur003:
I want to solve this... Can you make me an assignee, Please?

jakelishman:
Sure, thanks!

GuillermoAbadLopez:
@jakelishman I see this was not tackled in the end, right? Do you still want to implement it?

If so, could you reassign it to me? Thanks! 🙌

GuillermoAbadLopez:
Now that the [PR #12458](https://github.com/Qiskit/qiskit/pull/12458) is merged 🚀🚀:

Should we wait for the full deprecation cycle (raising error), to close the Issue? Or would you close it already? 
For me do as you consider 👍

----
@1ucian0, @Eric-Arellano @kevinsung

1ucian0:
It can be closed I think. Thanks!

PR #12458, opened 2024-05-25
Add warning for bad `justify` input, in `circuit_drawer`

GuillermoAbadLopez:
### CHECKS:
- [x] I have added the tests to cover my changes.
- [x] All available test passing:
Screenshot 2024-06-17 at 17 41 26

------------------------------

### Summary current state:
PR originally tackling https://github.com/Qiskit/qiskit/issues/12089, to raise a warning for bad `justify` argument inputs.

### The relevant code changes of this PR (+ after first review), are:
- Changed the default value of `justify`, to `"left"` (previously `None`).
- Added the corresponding `DeprecationWarning`, saying we will error for wrong `justify` arguments in the future.
- Added the corresponding tests for the warning.
- Some minor auto-formater reordering of imports (can be undone, if desired).

### Corner cases now working better:

1) If the user manually passes `justify=None` (which ends as `justify="left"`) we raise a warning, which before since itwas the default value, we couldn't... 

2) If the user manually passes any `justify` that passes as True `if justify:`, but which is not a `string`. Since before we would have done `justify.lower()` to it, getting a non-personalized error message: `'X'object has no attribute 'lower'`...

-----
-----
-----

### OUTDATED original Summary (before first review):
(before changing the default value `None` to `"left"`, when only a `warning` was raised [except for `None`])

PR tackling https://github.com/Qiskit/qiskit/issues/12089, where I have added a warning, for when the `justify` argument is not in the three accepted cases `"left"`, `"right"` and `"none"`.


qiskit/visualization/circuit/_utils.py
```diff
@@ -397,8 +401,15 @@ def _get_layered_instructions(
...
-    # default to left
-    justify = justify if justify in ("right", "none") else "left"
+    # If wrong input, default ('left') will be used.
+    if justify not in ("left", "right", "none"):
+        if justify is not None:
+            warn(
+                f"justify must be 'left, right or none', not: '{justify}'. Default (left) will be used.",
+                UserWarning,
+                2,
+            )
```
Eric-Arellano: @1ucian0 @mtreinish can we error here instead of warn? I'm not sure if the deprecation policy allows it: https://github.com/Qiskit/qiskit/blob/main/DEPRECATION.md#changing-behavior

This code path was always broken of using an unrecognized arg for `justify`. Before, we silently switched to the fallback. Can we now error?

GuillermoAbadLopez: Whatever you guys prefer/decide is okey with me! 🙌 


But of course raising an `error` will be less backwards compatible, at least for people with bad `justify`'s around there (for who, until now, we were secretly changing it) 😅.

----

Also notice, that the corner case of manually inputting `justify=None`, will still not be covered in either case, since it's the default value all the way up... and we can't `warn`/`raise error` for it. And so we change it to `"left"` (without a warning), when there exists `justify="none"`, which might be confusing? 🤔 .

To solve this, personally, I would change the default to the real one: `"left"` in cascade up, without having `None`'s, like with [`vertical_compression` argument](https://github.com/Qiskit/qiskit/blob/main/qiskit/visualization/circuit/circuit_visualization.py#L65), but I would like to know your opinions first since that would be another, bigger issue/change.

Eric-Arellano: That's a good point about `justify=None`. I like your suggestion to change the default to be `justify="left"`. That is more obvious code to understand, including generating more clear documentation at https://docs.quantum.ibm.com/api/qiskit/qiskit.circuit.QuantumCircuit#draw. It also allows us to warn/error when explicitly setting `=None`, like you suggest. 

For now, please still keep the explicit fallback to `left` (line 412). But you could remove `if justify is not None:` because we can now always warn if the value is not already one of the 3 valid options.

You'll also want to update all the functions that (transitively) call `_get_layered_instructions` to use the new default of `="left"`.

--

After thinking about it more, I agree with you that it would not yet be safe to error because it would result in someone's program that used to work no longer running. 

However, erroring is still the better behavior in general. So, we can introduce a deprecation so that we can switch to erroring in 2.0.

The best way to deprecate this is to use https://docs.quantum.ibm.com/api/qiskit/utils#deprecate_arg. You'll use that on every _public_ interface where a user can set `justify`, like `QuantumCircuit.draw`. It's useful to put the deprecations closer to where the users set the argument so that you can make the message more clear of where the bad code is coming from. You don't want to set it on all the private helper functions because it would be noisy to have the message in both places; so you would actually remove the `warning` inside `_get_layered_instructions`. 

You'll want to set `deprecation_description` to something like `Setting QuantumCircuit.draw()’s justify argument to a value other than 'left', 'right', or 'none'`. You also need to set `predicate`.

test/python/visualization/test_circuit_drawer.py
```diff
@@ -239,6 +238,15 @@ def test_reverse_bits(self):
         result = visualization.circuit_drawer(circuit, output="text", reverse_bits=True)
         self.assertEqual(result.__str__(), expected)
 
+    def test_justify_warning(self):
```
Eric-Arellano: Thanks for adding a test! If you take the above recommendation to deprecate setting an invalid value, then it'd be great to test here that every one of the public interfaces is warning correctly.

qiskit/visualization/circuit/_utils.py
GuillermoAbadLopez: Great! 
Those changes were exactly what I was thinking is best 👌, I'll apply them when I have a free moment 👍

--

And respect the deprecation, yeah I think that is definetly the way to go, honestly didn't know how you did it, so I didn't want to jump into deprecations, but with your instructions and that wrapper, this will be much easier, thanks!

Edit: Also the thing about putting the deprecation error closer to the user makes a lot of sense, I'll also do so, thanks again!

GuillermoAbadLopez: ### I have a question, regarding the `@deprecate_arg`: 

In this concrete case, we have `Circuit.draw()` and `circuit_drawer()`, where the first, calls the second one, but both are public... 

--- 
So if I add the wrapper in both public methods, I will get the error twice for `Circuit.draw()`, like in 🟨 here:
Screenshot 2024-06-06 at 02 20 01

But if instead, I only leave the wrapper in the common one (`circuit_drawer()`), I won't get two warns, but I won't get a very accurate position of the error, like in 🟨 here:
Screenshot 2024-06-06 at 02 19 41

### I'm not sure what you prefer, so I'll leave the first one until I receive further feedback 👍 
(this first one, also includes more info, since it shows correctly what you inputted)

test/python/visualization/test_circuit_drawer.py
GuillermoAbadLopez: No problem! I'll do that indeed! 🙌

qiskit/visualization/circuit/_utils.py
Eric-Arellano: This is a little convoluted, but you can set the default value to `left` when `QuantumCircuit.draw()` calls `circuit_drawer`. So, you fix the bad arg before the second warning is relevant.

You'll be repeating this logic multiple times to determine if `justify` is an invalid value. So you can define a private helper function like `_is_invalid_justify_arg`.

GuillermoAbadLopez: Ohh, I see, that makes sense, great! 👍

qiskit/circuit/quantumcircuit.py
```diff
@@ -3115,6 +3117,13 @@ def decompose(
         # do not copy operations, this is done in the conversion with circuit_to_dag
         return dag_to_circuit(dag, copy_operations=False)
 
+    @deprecate_arg(
+        "justify",
+        since="1.1.0",
+        deprecation_description="Setting QuantumCircuit.draw()’s justify argument to a value other than "
+        "'left', 'right', or 'none'. Until deprecation, default 'left' will be used instead.",
+        predicate=lambda justify: justify not in ["left", "right", "none"],
```
Eric-Arellano:
```suggestion
        predicate=_is_valid_justify_arg,
```

qiskit/circuit/quantumcircuit.py
```diff
@@ -3260,6 +3269,8 @@ def draw(
         # pylint: disable=cyclic-import
         from qiskit.visualization import circuit_drawer
 
+        justify = justify if _is_valid_justify_arg(justify) else "left"
```
Eric-Arellano:
```suggestion
        # We set this here to avoid the deprecation triggering again
        # from circuit_drawer. This code should be removed once
        # the deprecation is complete.
        justify = justify if _is_valid_justify_arg(justify) else "left"
```

qiskit/visualization/circuit/_utils.py
```diff
@@ -370,8 +374,22 @@ def generate_latex_label(label):
...
 
 
+def _is_valid_justify_arg(justify):
+    """Given a justify argument, return True if it is valid, False otherwise.
+    Args:
+        justify (str): The justify argument to check.
+
+    Returns:
+        bool: True if the justify argument is valid, False otherwise.
+    """
```
Eric-Arellano:
```suggestion
def _is_valid_justify_arg(justify: str) -> bool:
    """Return True if the `justify` arg is valid."""
```

qiskit/visualization/circuit/circuit_visualization.py
```diff
@@ -28,30 +28,40 @@
...
 
 logger = logging.getLogger(__name__)
 
 
+@deprecate_arg(
+    "justify",
+    since="1.1.0",
+    deprecation_description="Setting circuit_drawer()’s justify argument to a value other than "
+    "'left', 'right', or 'none'. Until deprecation, default 'left' will be used instead.",
+    predicate=lambda justify: justify not in ["left", "right", "none"],
```
Eric-Arellano: You'll need to import the function

```suggestion
    predicate=_is_valid_justify_arg,
```

qiskit/circuit/quantumcircuit.py
```diff
@@ -3115,6 +3116,13 @@ def decompose(
         # do not copy operations, this is done in the conversion with circuit_to_dag
         return dag_to_circuit(dag, copy_operations=False)
 
+    @deprecate_arg(
+        "justify",
+        since="1.1.0",
```
Eric-Arellano: Maintainers, do we need to set `pending=True`? I don't know how that works now that we're on Qiskit 1.0.

qiskit/circuit/quantumcircuit.py
```diff
@@ -3115,6 +3116,13 @@ def decompose(
...
+        deprecation_description="Setting QuantumCircuit.draw()’s justify argument to a value other than "
+        "'left', 'right', or 'none'. Until deprecation, default 'left' will be used instead.",
```
Eric-Arellano: You could add this:

> Starting in Qiskit 2.0, invalid arguments will error.

Eric-Arellano (commented):
Thanks! This looks good to me 🚀 I'll be OOO till the 15th, so will defer to others to finish review.

qiskit/circuit/quantumcircuit.py
GuillermoAbadLopez: Its says "Set to True if the deprecation is still pending.", and technically we haven't deprecated it yet, so I guess I should add it yes.

Also, just checked, and the rest of @deprecate_arg with `since="1.1.0"` have also `pending=True`.

qiskit/circuit/quantumcircuit.py
GuillermoAbadLopez: Done!

qiskit/circuit/quantumcircuit.py
GuillermoAbadLopez: Obviously! My bad, adding it :+1:

qiskit/circuit/quantumcircuit.py
GuillermoAbadLopez: Its says `"Set to True if the deprecation is still pending."`, and technically we haven't deprecated it yet, so I guess I should add it yes.

Also, just checked, and the rest of `@deprecate_arg` with `since="1.1.0"` have also `pending=True`, which further supports it, but let's see what do the maintainers say 👌 .

qiskit/circuit/quantumcircuit.py
GuillermoAbadLopez: Actually, I added something on the lines, I guess a commit later, from your review 

I'll merge both now 👌

qiskit/visualization/circuit/_utils.py
GuillermoAbadLopez: Committing it 👍

qiskit/visualization/circuit/_utils.py
GuillermoAbadLopez: Did it!

The only problem I see with this implementation is the repetitive use of the `.lower()` method, but I hope that is acceptable, since the other ways I tried don't look as readable...

---

Also, mention that this solves a previous error, for when the user didn't pass a `string` neither `None`, for which `justify.lower()`, would have raised the standard non-personalized error.

qiskit/circuit/quantumcircuit.py
GuillermoAbadLopez: Actually, the `@deprecate_arg`, already automatically appends this, to the end of the message:
`"is deprecated as of qiskit 1.1.0. It will be removed no earlier than 3 months after the release date."`

So since both, don't fit very well, if its okei with you, I'll try to write something else, that looks better with such automation 👍

qiskit/circuit/quantumcircuit.py
Eric-Arellano: Ah, this gets to my other comment I forgot to say: `since` should be `1.2.0` because 1.1.0 is already released. I'm not convinced we want to set `pending=True` because there won't be a 1.3 release, so we wouldn't have the opportunity to switch off `pending` before the 2.0 release, so we could not safely remove this mechanism in 2.0.

This is our justification for `pending`, but it was written before the Qiskit 1.0 deprecation policy, so it might be stale:

> The alternative path must be in place for one minor version before any warnings are issued. For example, if we want to replace the function foo() with bar(), we must make at least one minor release with both functions before issuing any warnings within foo(). You may issue PendingDeprecationWarnings from the old paths immediately, but this is not necessary and does not affect any timelines for removal.
>
> Reason: we need to give people time to swap over without breaking their code as soon as they upgrade.

Let's see what the others think.

GuillermoAbadLopez:
Currently, tests are not passing due to a `circular import`, most likely due to using `_is_valid_justify_arg`, in all the files...

I had a couple of quick ideas, but it requires moving this function definition, to places, where it doesn't feel right..

-----
Update: Regex check has been removed, for the sake of not complicating this more! 👍

qiskit/circuit/quantumcircuit.py
GuillermoAbadLopez: In the end, I went back to rewriting the lambda function in the `@deprecate_arg`, instead of importing it, due to circular import errors it produces...

And since it's just a temporary one-liner, I hope having duplicated code there is not too bad... If it is, please tell me 👌

qiskit/circuit/quantumcircuit.py
GuillermoAbadLopez: Changed`"1.1.0"` to `"1.2.0"`.

And also removed `pending=True` for the moment, since with it, the warning was not being raised..

----
**Update:** 
If I recall correctly Luciano confirmed this should not be `pending=True` in the call 👍

1ucian0:
I think we should keep `None` as a valid argument.

GuillermoAbadLopez:
# History of the PR (context to new reviewers):

## 0) Original code:
```python
if justify:
    justify = justify.lower()
    

# default to left
justify = justify if justify in ("right", "none") else "left"
```

### Problems:
- Any value different than  `"left"`, `"right"` or `"none"`, would be changed to `"left"`, without raising anything...
- If the user manually passes any `justify` with bool value `True`, but which is not a `string`.  Since we would do `justify.lower()` to it, getting a non-personalized error message: `'X'object has no attribute 'lower'`...

-----
## 1) PR first implementation:
Added a warning, for when the `justify` argument is not in the three accepted cases "left", "right" and "none":
```python

if justify:
    justify = justify.lower()

if justify not in ("left", "right", "none"):
    if justify is not None:
        warn(
            f"justify must be 'left, right or none', not: '{justify}'. Default (left) will be used.",
            UserWarning,
            2,
         )
    justify = "left"
```

### Problems:
- If the user manually inputs `justify=None`, it would be changed to `"left"`, without raising anything...
- If the user manually passes any `justify` with bool value `True`, but which is not a `string`.  Since we would do `justify.lower()` to it, getting a non-personalized error message: `'X'object has no attribute 'lower'`...

-----
## 2) PR implementation after @Eric-Arellano review // [current state]:
Changed the default value of `justify` from `None` to `"left"` in cascade up, calls the `.lower()` method only for strings, and instead of raising a `warn()`, used the argument wrapper for `DeprecationWarning`:
```python
# In `QuantumCircuit` and `drawer()`:
@deprecate_arg(
    "justify",
    since="1.2.0",
    deprecation_description="Setting QuantumCircuit.draw()’s justify argument to a value other than"
    "'left', 'right', or 'none'. Default 'left' will be used. Support for invalid justify arguments",
    predicate=lambda justify: (
        not isinstance(justify, str) or justify.lower() not in ("left", "right", "none")
    ),
)
def draw(...
```
\+
```python
def _is_valid_justify_arg(justify):
    """Return True if the `justify` argument is valid."""
    if isinstance(justify, str):
        justify = justify.lower()

    return justify in ("left", "right", "none")

justify = justify.lower() if _is_valid_justify_arg(justify) else "left"
```
\+
```python
# Change justify default value from None to "left" in cascade up
```

### Problems:
- It adds more complexity to what needs to be removed in the future deprecation...
- It changes more code, making this PR deviate further from the original Issue...

GuillermoAbadLopez:
# Future of the PR (two proposals):
## 3.A) PR proposal from @1ucian0 & @kevinsung:
Go back to point "1) PR first implementation", so `None` is still a valid argument (equals to `"left"`), and to remove the complicated `DeprecationWarning` argument wrapper logic, which in this case instead of helping complicates things:
```python

if isinstance(justify, str):
    justify = justify.lower()

if justify is None:
    justify = "left"

if justify not in ("left", "right", "none"):
    warn(
        f"Setting QuantumCircuit.draw()’s justify argument to a value other than "
        "'left', 'right', 'none' or None (='left'). Default 'left' will be used. Starting "
        " in Qiskit 2.0, invalid arguments will error.",
        DeprecationWarning,
        2,
     )
    justify = "left"
```
\+ 
```python
# Improve documentation, to make clear that None, will get changed to "left"
```
\+
```python
# New release note
```
_(won't apply `.lower()` if its not a string anymore)_

### Problems:
- If the user manually inputs `justify=None`, it would be changed to `"left"`, without raising anything... But this shouldn't be a major problem since, it should be documented.

-----
## 3.B) Mix between previous proposal (3.A) and PR current state:
Remove the complex logic from the `DeprecationWarning` argument wrapper, but not remove the change of the default value of `justify` from `None` to `"left"`:

```python

if isinstance(justify, str):
    justify = justify.lower()

if justify not in ("left", "right", "none"):
    warn(
        f"Setting QuantumCircuit.draw()’s justify argument to a value other than "
        "'left', 'right' or 'none'. Default 'left' will be used. Starting in Qiskit 2.0, "
        "invalid arguments will error.",
        DeprecationWarning,
        2,
     )
    justify = "left"
```
\+ 
```python
# Change justify default value from None to "left" in cascade up
```
\+
```python
# New release note
```
_(won't apply `.lower()` if its not a string anymore)_

### Problems:
- It changes a bit more code, making this PR deviate further from the original Issue...

qiskit/circuit/quantumcircuit.py
GuillermoAbadLopez: **Update, from call with luciano and Kevin:**

They think in this case the wrapper complicates things instead of helping, so we should maybe remove it 👍

----
**New update:**
We might leave it as it is in the end actually, we'll see what Luciano says 🙌

qiskit/circuit/quantumcircuit.py
GuillermoAbadLopez: (repeated comment)

----
**Update, from call with luciano and Kevin:**

They think in this case the wrapper complicates things instead of helping, so we should maybe remove it 👍

qiskit/circuit/quantumcircuit.py
GuillermoAbadLopez: (repeated comment, since this is also involved)

----
**Update, from call with luciano and Kevin:**

They think in this case the wrapper complicates things instead of helping, so we should maybe remove it 👍

kevinsung (dismissed):
On further thought, the current implementation is fine with me.

Eric-Arellano (dismissed):
Thanks for iterating on this! I agree with @kevinsung that this approach looks good. I'm fine with duplicating the logic in the `lambda` because this is temporary code and the logic is acceptable complexity. 

I agree that this is awkward transition code, but it's temporary. The benefit is a much clearer user experience, including better documentation by showing explicitly in the docs that the default is `='left'` rather than it being `None` and them having to read that that really means `left`.

This looks good to me! Thanks again for your work on this and the test.

qiskit/circuit/quantumcircuit.py
GuillermoAbadLopez: Resolving this conversation given your last comment: https://github.com/Qiskit/qiskit/pull/12458#pullrequestreview-2126105662

qiskit/circuit/quantumcircuit.py
GuillermoAbadLopez: **In the end the error message with the `@deprecate_arg`, looks like:**

"Setting QuantumCircuit.draw()’s justify argument to a value other than 'left', 'right', or 'none'. Default 'left' will be used. Support for invalid justify arguments is deprecated as of qiskit 1.2.0. It will be removed no earlier than 3 months after the release date."`

GuillermoAbadLopez:
Just added a release note for the PR as discussed in the call, if you think:
- its not necessary in the end,
- some note is not in the correct section

please tell me, and I'll change it 🙌

GuillermoAbadLopez:
I'm getting this error in the Preliminary tests now, due to the release note:
```
lint: commands[4]> pylint -rn qiskit test tools

------------------------------------
Your code has been rated at 10.00/10

lint: commands[5]> python /home/vsts/work/1/s/tools/verify_headers.py qiskit test tools examples crates
lint: commands[6]> python /home/vsts/work/1/s/tools/find_optional_imports.py
lint: commands[7]> python /home/vsts/work/1/s/tools/find_stray_release_notes.py
lint: commands[8]> reno -q lint
  docs: FAIL code 1 (606.14=setup[410.84]+cmd[24.06,171.24] seconds)
  lint: OK (673.33=setup[175.87]+cmd[39.08,2.25,0.73,53.68,397.27,0.16,0.66,0.08,3.56] seconds)
  evaluation failed :( (1279.88 seconds)

##[error]Bash exited with code '255'.
```
In local everything passes correctly, but I think its not checking new files... 
Screenshot 2024-06-19 at 05 31 30

Also, I have tried various changes I thought would solve it, but nothing worked so far.
If you have encountered this before, any help would be appreciated.

Eric-Arellano:
@GuillermoAbadLopez `reno lint` was a red herring :/ The warning is from Sphinx higher up in CI logs:

> /home/vsts/work/1/s/docs/release_notes.rst:438: WARNING: Inline interpreted text or phrase reference start-string without end-string.

--

I talked more to @1ucian0 and @kevinsung we decided to indeed go back to using `None` as the default argument. While it makes the documentation a little less clear, Qiskit SDK tends to use `None` as the default like this and then set the "true default" at runtime because it gives more flexibility for the future. 

For example, imagine we want to change the default of `left` to a new value like `justified` -- that is a breaking behavior change and would need a deprecation cycle. To detect when people are using the default—rather than explicitly setting `='left'`—we would need to check `if arg is None`. But we can't do that check anymore if we've now set the default directly to be `'left'`; we no longer know if you explicitly set the value or you're relying on the default.

So, please go back to using `None` as the default arg, which results in `left` being used. But, of course, still warn on otherwise invalid arguments.

Pardon the back-and-forth and thank you for persistence on this issue!

GuillermoAbadLopez:
> @GuillermoAbadLopez reno lint was a red herring :/ The warning is from Sphinx higher up in CI logs:
> 
> > /home/vsts/work/1/s/docs/release_notes.rst:438: WARNING: Inline interpreted text or phrase reference start-string without end-string.

That makes a lot of sense, found something, thanks! 👍

GuillermoAbadLopez:
No worries at all, we should do what its best for the stack, in the end!

 I'll change it, but it might take a bit, since I'm going OOO for a couple of days  👍 
 
> I talked more to @1ucian0 and @kevinsung we decided to indeed go back to using None as the default argument. While it makes the documentation a little less clear, Qiskit SDK tends to use None as the default like this and then set the "true default" at runtime because it gives more flexibility for the future.
> ...

GuillermoAbadLopez:
## Last suggestions addressed!

1) Have gone back to **`None` as the default value for `justify`**

2) **Simplified the deprecation logic** to a simple `warn`, similar to the starting one.
(Since even though the `@arg_dreprecation` decorator/wrapper was better for localizing the error, it seemed a bit confusing for the future maintainer that would have finally deprecated this, with single lines around to delete, that he could have missed..., now it's super simple! Also this way allows for a `Regex` check in the test)

3) Some final **changes to the docstrings** (both public and private), and **updated the release note**.

@Eric-Arellano @kevinsung @1ucian0

kevinsung (approved):

Eric-Arellano (approved):
Thank you! @Qiskit/terra-core @1ucian0 can you please merge?

1ucian0 (approved):

PR #12461, opened 2024-05-26
Added parameters property to BindingsArray

### LINKED ISSUE #12436: Add `parameters` property to `BindingsArray`
### What should we add?

Bindings array has no easy way to get an ordered list of parameters it contains. It would be nice to have a property method `BindingsArray.parameters` that returns the ordered list of parameter values correpsonding to the array order returned by `BindingsArray.as_array` when no parameters kwarg is provided. Eg `ba.as_array() == ba.as_array(ba.parameters)`

shraddha-aangiras:
### Summary
Fixes #12436


t-imamichi:
Could you add a unit test of the method? I assigned @chriseclectic to review this PR.

PR #12463, opened 2024-05-27
set a dummy QISKIT_SETTINGS so tests do not interact with local env

1ucian0:
Some of the tests were failing with in my machine and I noticed that it was because my particular setting in `~/.qiskit/settings.conf`. Tests should not depend on the situation of the `setting.conf`.

This PR sets a dummy `QISKIT_SETTINGS` to avoid that.


test/utils/base.py
```diff
@@ -245,9 +245,9 @@ def setUpClass(cls):
 
 
 class FullQiskitTestCase(QiskitTestCase):
-    """Terra-specific further additions for test cases, if ``testtools`` is available.
+    """Qiskit-specific further additions for test cases, if ``testtools`` is available.
 
-    It is not normally safe to derive from this class by name; on import, Terra checks if the
+    It is not normally safe to derive from this class by name; on import, Qiskit checks if the
```
1ucian0: unrelated. Just removing Terra things.

eliarbel (dismissed):
LGTM

ElePT (dismissed):

ElePT (commented):
@1ucian0,  why revert the removal of "Terra"? [3f26e14](https://github.com/Qiskit/qiskit/pull/12463/commits/3f26e141efe093f307827cacae2df0528737962a)

ElePT (approved):
Aha, I see, it's been moved to #12720. LGTM.

jakelishman:
Fwiw, this was already handled by `tox` environment isolation; `QISKIT_SETTINGS` is not passed on by `tox` to the runner.  There's many environment variables that can influence testing, and it's also not a done deal that doing this at `setUp` is good enough - Qiskit's "config" module happens to reload the file on basically every check of it (iirc), but a more sane implementation might have loaded the file just once at `import qiskit` (which happens before this environment patching) and so be unaffected by it.

Tests that can be influenced unduly by options in the settings file should probably be passing the options precisely within the tests to avoid this.

PR #12469, opened 2024-05-28
Fix a bug in isometry.rs

ShellyGarion:
### Summary
After PR #12197 was merged, there is a bug with the following code:

```
num_qubits = 3
num_controls = 4
u = random_unitary(2 ** num_qubits)
ug = UnitaryGate(u)
cug = ug.control(num_controls)
cug 
```

The code is working for `num_qubits = 1` or `2` but not for `3` or `4` or `5`, and raises the following error:

```
/qiskit/circuit/library/generalized_gates/isometry.py:221, in Isometry._disentangle(self, circuit, q, diag, remaining_isometry, column_index, s)
    217 diagonal_mcg = self._append_mcg_up_to_diagonal(
    218     circuit, q, gate, control_labels, target_label
    219 )
    220 # apply the MCG to the remaining isometry
--> 221 v = isometry_rs.apply_multi_controlled_gate(v, control_labels, target_label, gate)
    222 # correct for the implementation "up to diagonal"
    223 diag_mcg_inverse = np.conj(diagonal_mcg).astype(complex, copy=False)

PanicException: assertion failed: j <= 1
```
### Details and comments
The solution is to remove the assertion `assert!(j <= 1)` in `crates/accelerate/src/isometry.rs:215:13`

### DIFF
--- a/releasenotes/notes/fix-isometry-rust-adf0eed09c6611f1.yaml
+++ b/releasenotes/notes/fix-isometry-rust-adf0eed09c6611f1.yaml
@@ -2,5 +2,5 @@
 fixes:
   - |
     Fix a bug in :class:`~.library.Isometry` due to an unnecessary assertion,
-    that led to an error in :method:`~.library.UnitaryGate.control(num_controls)`
+    that led to an error in :meth:`~.library.UnitaryGate.control`
     when :class:`~.library.UnitaryGate` had more that two qubits.
--- a/releasenotes/notes/fix-isometry-rust-adf0eed09c6611f1.yaml
+++ b/releasenotes/notes/fix-isometry-rust-adf0eed09c6611f1.yaml
@@ -2,5 +2,5 @@
 fixes:
   - |
     Fix a bug in :class:`~.library.Isometry` due to an unnecessary assertion,
-    that led to an error in :meth:`~.library.UnitaryGate.control`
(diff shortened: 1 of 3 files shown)


releasenotes/notes/fix-isometry-rust-adf0eed09c6611f1.yaml
```diff
@@ -0,0 +1,6 @@
+---
+fixes:
+  - |
+    Fix a bug in :class:`~.library.Isometry` due to an unnecessary assertion,
+    that led to an error in :meth:`~.library.UnitaryGate.control`
```
mtreinish: This will make the display text `UnitaryGate.control` and link to the same thing which will probably be a bit simpler:

```suggestion
    that led to an error in :meth:`.UnitaryGate.control`
```

mtreinish (commented):
This LGTM thanks for catching and fixing this. TBH, at the time when I added that assert I can't remember what I thought I was guarding against, but do remember explicitly thinking it was needed. But looking at it in retrospect it's clearly not correct for more than 2 target qubits so I'm not sure what I was thinking (I probably had tunnel vision trying to get the tests to pass).

Just one minor nit in the release notes but otherwise this is ready to go.

mtreinish (approved):

PR #12470, opened 2024-05-28
Adds various convenience functions to qiskit

sbrandhsn:
### Summary

This PR adds some convenience functions to Qiskit's `QuantumCircuit` class. The functionality was previously available but required users to dive into the API before determining the wanted quantities. The new 'depth' and 'number of gates' methods were previously available by defining a custom filter function for `depth` and `size` methods of `QuantumCircuit` but I thought it might be worthwhile for new users to have a convenience function that defines such a filter function for them (and saves them e.g. from accidentally counting barriers towards their metrics).

### DIFF
--- a/qiskit/circuit/library/blueprintcircuit.py
+++ b/qiskit/circuit/library/blueprintcircuit.py
@@ -190,10 +190,10 @@ class BlueprintCircuit(QuantumCircuit, ABC):
             self._build()
         return super().count_ops()
 
-    def num_nonlocal_gates(self):
+    def num_nonlocal_gates(self, n=2):
         if not self._is_built:
             self._build()
-        return super().num_nonlocal_gates()
+        return super().num_nonlocal_gates(n)
 
     def num_connected_components(self, unitary_only=False):
         if not self._is_built:
--- a/qiskit/circuit/library/blueprintcircuit.py
+++ b/qiskit/circuit/library/blueprintcircuit.py
@@ -191,6 +191,7 @@ class BlueprintCircuit(QuantumCircuit, ABC):
         return super().count_ops()
 
     def num_nonlocal_gates(self, n=2):
+        """Returns the number of n-qubit gates that contain at least n qubits."""
         if not self._is_built:
             self._build()
         return super().num_nonlocal_gates(n)
--- a/qiskit/circuit/quantumcircuit.py
+++ b/qiskit/circuit/quantumcircuit.py
@@ -3398,10 +3398,10 @@ class QuantumCircuit:
         """
         if not include_directives:
             return self.depth(
-                lambda x: not getattr(x.operation, "_directive", False) and x.qubits >= n
+                lambda x: not getattr(x.operation, "_directive", False) and len(x.qubits) >= n
             )
         else:
-            return self.depth(lambda x: x.qubits >= n)
+            return self.depth(lambda x: len(x.qubits) >= n)
 
     def num_nq_gates(self, n: int = 2, include_directives: bool = False) -> int:
         """Returns the number of gates with exactly n qubits (default: n=2), excluding directives by
@@ -3417,10 +3417,10 @@ class QuantumCircuit:
         """
         if not include_directives:
             return self.size(
-                lambda x: not getattr(x.operation, "_directive", False) and x.qubits == n
+                lambda x: not getattr(x.operation, "_directive", False) and len(x.qubits) == n
             )
         else:
-            return self.size(lambda x: x.qubits == n)
+            return self.size(lambda x: len(x.qubits) == n)
 
     def num_nonidle_qubits(self) -> int:
         """Returns the number of qubits in the quantum circuit that are non-idle,
--- a/qiskit/circuit/quantumcircuit.py
+++ b/qiskit/circuit/quantumcircuit.py
@@ -3394,7 +3394,7 @@ class QuantumCircuit:
             include_directives (bool): Whether to include directives such as barriers
 
         Returns:
-            int: Quantum circuit depth in terms of two-qubit gates
+            int: Quantum circuit depth in terms of two or more qubit gates
         """
         if not include_directives:
             return self.depth(
--- a/qiskit/circuit/quantumcircuit.py
+++ b/qiskit/circuit/quantumcircuit.py
@@ -3394,7 +3394,7 @@ class QuantumCircuit:
             n (int): The minimum number of qubits in an instruction that counts towards the depth.
             include_directives (bool): Whether to include directives such as barriers
         Returns:
-            int: Quantum circuit depth in terms of n-qubit gates (default n=2).
+            int: Quantum circuit depth in terms of gates with at least n qubits (default n=2).
         """
         if not include_directives:
             return self.depth(
--- a/qiskit/circuit/quantumcircuit.py
+++ b/qiskit/circuit/quantumcircuit.py
@@ -3386,6 +3386,90 @@ class QuantumCircuit:
 
         return max(op_stack)
 
+    def depth_2q(self, include_directives: bool = False) -> int:
+        """Returns the depth in terms of two-qubit quantum gates, excluding directives by default.
+
+        Args:
+            include_directives (bool): Whether to include directives such as barriers
+
+        Returns:
+            int: Quantum circuit depth in terms of two-qubit gates
+        """
+        if not include_directives:
+            return self.depth(
+                lambda x: not getattr(x.operation, "_directive", False) and x.qubits == 2
+            )
+        else:
+            return self.depth(lambda x: x.qubits == 2)
+
+    def depth_nq(self, n: int, include_directives: bool = False) -> int:
+        """Returns the depth in terms of quantum gates with at least n-qubits,
+        excluding directives by default.
+
+        Args:
+            n (int): The minimum number of qubits in an instruction that counts towards the depth.
+            include_directives (bool): Whether to include directives such as barriers
+        Returns:
+            int: Quantum circuit depth in terms of n-qubit gates
+        """
+        if not include_directives:
+            return self.depth(
+                lambda x: not getattr(x.operation, "_directive", False) and x.qubits >= n
+            )
+        else:
+            return self.depth(lambda x: x.qubits >= n)
+
+    def num_2q_gates(self, include_directives: bool = False) -> int:
+        """Returns the number of gates with exactly 2-qubits, excluding directives by default.
+
+        Args:
+            include_directives (bool): Whether to include directives such as barriers
+
+        Returns:
+            int: The number of gates in the quantum circuit with exactly n-qubits.
+        """
+        if not include_directives:
+            return self.size(
+                lambda x: not getattr(x.operation, "_directive", False) and x.qubits == 2
+            )
+        else:
+            return self.size(lambda x: x.qubits == 2)
+
+    def num_nq_gates(self, n: int, include_directives: bool = False) -> int:
+        """Returns the number of gates with exactly n-qubits, excluding directives by default.
+        See :meth:`.num_nonlocal_gates` to retrieve the number of gates with at least two-qubits,
+        (excluding directives).
+
+        Args:
+            n (int): The exact number of qubits in an instruction.
+            include_directives (bool): Whether to include directives such as barriers
+
+        Returns:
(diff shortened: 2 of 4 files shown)


releasenotes/notes/adds-convenience-functions-f14907a13fcb7929.yaml
```diff
@@ -0,0 +1,7 @@
+---
+features:
+  - |
+    Added convenience functions to the :class:`.QuantumCircuit` class that return the depth in terms of
```
garrison: The proposed new features are methods, not functions.

qiskit/circuit/quantumcircuit.py
```diff
@@ -3386,6 +3386,90 @@ def depth(
...
+        """
+        if not include_directives:
+            return self.depth(
+                lambda x: not getattr(x.operation, "_directive", False) and x.qubits == 2
+            )
+        else:
+            return self.depth(lambda x: x.qubits == 2)
+
+    def depth_nq(self, n: int, include_directives: bool = False) -> int:
+        """Returns the depth in terms of quantum gates with at least n-qubits,
```
garrison: The key phrase seems to be _at least_.  And for this reason, `depth_nq` with `n=2` is _not_ the same thing as `depth_2q`.  This is potentially confusing.

sbrandhsn:
Thanks for spotting these @garrison - I addressed your comments in the recent commit.

releasenotes/notes/adds-convenience-functions-f14907a13fcb7929.yaml
```diff
@@ -0,0 +1,7 @@
+---
+features:
+  - |
+    Added convenience methods to the :class:`.QuantumCircuit` class that return the depth in terms of
+    gates that have exactly two-qubits (:meth:`.QuantumCircuit.depth_2q`) or at least n-qubits (:meth:`.QuantumCircuit.depth_nq`).
```
garrison:
```suggestion
    gates that have at least two-qubits (:meth:`.QuantumCircuit.depth_2q`) or at least n-qubits (:meth:`.QuantumCircuit.depth_nq`).
```

qiskit/circuit/quantumcircuit.py
```diff
@@ -3386,6 +3386,91 @@ def depth(
...
+            return self.depth(lambda x: x.qubits >= n)
+
+    def num_2q_gates(self, include_directives: bool = False) -> int:
+        """Returns the number of gates with exactly 2-qubits, excluding directives by default.
+
+        Args:
+            include_directives (bool): Whether to include directives such as barriers
+
+        Returns:
+            int: The number of gates in the quantum circuit with exactly n-qubits.
```
ShellyGarion: should be 2 qubits and not n qubits.

qiskit/circuit/quantumcircuit.py
ShellyGarion: why the function `depth_nq` is with `>=n` and `num_nq_gates` is with `==n` ?

qiskit/circuit/quantumcircuit.py
```diff
@@ -3386,6 +3386,91 @@ def depth(
 
         return max(op_stack)
 
+    def depth_2q(self, include_directives: bool = False) -> int:
+        """Returns the depth in terms of quantum gates with at least 2-qubits,
```
ShellyGarion: why the function `depth_2q` is with `>=2` and `num_2q_gates` is with `==2` ?

ShellyGarion (commented):
I think these functions are generally useful, but do you see any specific use-cases? especially for the n-qubit functions.

For example, the function of `depth_2q` has already been used in several tests, including:

```
test/python/synthesis/test_cx_cz_synthesis.py:            
self.assertTrue(depth2q <= 5 * num_qubits)

test/python/synthesis/test_clifford_decompose_layers.py:            
self.assertTrue(depth2q <= 7 * num_qubits + 2)

test/python/synthesis/test_cz_synthesis.py:            
self.assertTrue(depth2q == 2 * num_qubits + 2)

test/python/synthesis/test_stabilizer_synthesis.py:            
self.assertTrue(depth2q == 2 * num_qubits + 2)
```

qiskit/circuit/quantumcircuit.py
sbrandhsn: `num_nq_gates` exactly checks the number of qubits because I thought users might be interested in that quantity in general. Also, `num_nonlocal_gates` determines the number of gates with at least two qubits. I extended the arguments of `num_nonlocal_gates` to include an integer `n` that specifies the minimum number of qubits in a gate to still be counted towards the quantity returned by `num_nonlocal_gates`. The default behavior of `num_nonlocal_gates` stays the same, so I hope our stability policy is not violated...

sbrandhsn:
Thanks for your comments, I addressed them in the most recent commits! :-) 

> I think these functions are generally useful, but do you see any specific use-cases? especially for the n-qubit functions.
> 
> For example, the function of `depth_2q` has already been used in several tests, including:
> 
> ```
> test/python/synthesis/test_cx_cz_synthesis.py:            
> self.assertTrue(depth2q <= 5 * num_qubits)
> 
> test/python/synthesis/test_clifford_decompose_layers.py:            
> self.assertTrue(depth2q <= 7 * num_qubits + 2)
> 
> test/python/synthesis/test_cz_synthesis.py:            
> self.assertTrue(depth2q == 2 * num_qubits + 2)
> 
> test/python/synthesis/test_stabilizer_synthesis.py:            
> self.assertTrue(depth2q == 2 * num_qubits + 2)
> ```

I think these functions will be useful to any user that is interested in building their own circuits or looking into the characteristics of these circuits. Regarding the `n`-qubit versions of the introduced functions: they would be useful to anyone working more with n-qubit gates either because the hardware they are working with is potentially able to realize competitive n-qubit gates or because they are looking at synthesis/transpilation problems that involve n-qubit gates.

On the other hand, the two-qubit versions are going to be searched for first by users that only worked with two-qubit gates yet.

qiskit/circuit/quantumcircuit.py
```diff
@@ -3386,6 +3386,91 @@ def depth(
...
+        excluding directives by default.
+
+        Args:
+            include_directives (bool): Whether to include directives such as barriers
+
+        Returns:
+            int: Quantum circuit depth in terms of two-qubit gates
```
eliarbel:
```suggestion
            int: Quantum circuit depth in terms of two or more qubit gates
```

qiskit/circuit/quantumcircuit.py
eliarbel: Agree the name is a bit confusing an inconsistent with the others. Consider changing to something like `depth_multi_qubit`

qiskit/circuit/quantumcircuit.py
eliarbel: Maybe have just this one with default `n = 2` and drop `depth_2q`

qiskit/circuit/quantumcircuit.py
```diff
@@ -3386,6 +3386,91 @@ def depth(
...
+            int: The number of gates in the quantum circuit with exactly two qubits.
+        """
+        if not include_directives:
+            return self.size(
+                lambda x: not getattr(x.operation, "_directive", False) and x.qubits == 2
+            )
+        else:
+            return self.size(lambda x: x.qubits == 2)
+
+    def num_nq_gates(self, n: int, include_directives: bool = False) -> int:
```
eliarbel: Same suggestion: have `n:int = 2` and drop the `num_2q_gates`

sbrandhsn: Thanks for your suggestion. I removed the `n=2` versions of the methods and updated the release notes accordingly!

jakelishman:
I think there was a meeting where these were discussed that I wasn't in, so apologies if I'm repeating something:

Did we consider the alternative of doing all these by "examples" documentation within the relevant functions or discussion of the circuit data model?  The depth ones in particular feel like they could have been examples, and the others are relatively simple filters on an explicitly public data field - I'm concerned that adding more methods makes it harder to find _everything_ on `QuantumCircuit`, and whenever there's a new method, people have to learn all its intricacies about what it does and doesn't do.  It feels like teaching people to do it themselves might be more scalable on both flexibility and discoverability.

sbrandhsn:
That's a good option and I think we should include the added procedures in the documentation at some point. However, it turns out many users need these functions and defining them repeatedly seems bothersome. There likely are also users that don't want to learn Qiskit but still want to use it in order to get accurate information on a circuit they were cooking up somehow.

To make it worse, the definition of `depth_nq` provided in the documentation at https://github.com/Qiskit/qiskit/blob/bac400d29d5072429786fb4088a24a9f197f167e/qiskit/circuit/quantumcircuit.py#L3333 is inaccurate (it would also include n-qubit barriers in the depth calculation which I think is at least unexpected by users). So, having one source of truth for crucial circuit characteristics seems worthwhile to me.

qiskit/circuit/quantumcircuit.py
sbrandhsn: I was also thinking about whether `depth_nq` should work on gates with exactly `n` qubits or for `>= n` qubits. I decided for the latter because I thought the reason an user might want `depth_nq` to exclude certain gates is that they are comparatively cheap. I think it is safe to assume that if an user regards a gate with `n` qubits to be relatively expensive, they would also consider a gate with `n+i` qubits to be expensive (if there are no other properties to decide on). I also was not able to see a situation where an user might only be interested in the two-qubit gate depth while ignoring all other gates in the presence of e.g. three-qubit gates.

Cryoris:
Generally I agree with Jake that we've been trying to avoid adding methods that don't have an explicit use-case to not clutter the circuit (which already has loads of methods) with more. Since most functionality here is already in place, a compromise could be to extend docstrings or functions of the existing methods.
* instead of `depth_nq`, we could add a concrete example in the docstring of the `depth` method (or even add an argument to do that)
* same for `num_nonlocal_gates` and `num_nq_gates`
* the non-idle wires are already easy to get from `dag.idle_wires`, though granted that users would first have to convert to a `DAGCircuit`

garrison:
> * the non-idle wires are already easy to get from `dag.idle_wires`, though granted that users would first have to convert to a `DAGCircuit`

This is the only one that I think (IMO) may be a good candidate for its own new method.  And I think it'd be more useful as a `nonidle_qubits()` method that returns the `set`, rather than taking `len(set(...))`.

qiskit/circuit/library/blueprintcircuit.py
```diff
@@ -190,10 +190,10 @@ def count_ops(self):
             self._build()
         return super().count_ops()
 
-    def num_nonlocal_gates(self):
+    def num_nonlocal_gates(self, n=2):
```
1ucian0, line 193: `n` might need some documentation. A short one, can be a comment.

sbrandhsn:
I'm fine with having the functionality provided by these convenience procedures as a parameter to the default `depth` and `size` methods - although I think that this may not be a terrible good fit and would actually be better represented by new methods. I will just summarize my arguments and then leave it at that: :-)
-  As soon as a user works with circuits in any way (e.g. construction, transpilation, synthesis, debugging low fidelity), the user will need the information provided by these procedures.
- A user would absolutely expect this information to be available at hand without having to understand the details of the data structures involved or defining their own lambda function.
- Adding these procedures to the docstring can be missed easily, especially by novice users.
- The methods proposed in this PR have been implemented incorrectly in the past (https://github.com/Qiskit/qiskit/blob/bac400d29d5072429786fb4088a24a9f197f167e/qiskit/circuit/quantumcircuit.py#L3333, https://github.com/Qiskit-Extensions/circuit-knitting-toolbox/issues/621 ). Especially novice users will not know to exclude directives in their lambda function when determining `depth_nq`, `nonidle_qubits`, or `num_nq_gates` and will thus get an inaccurate/incorrect return.
- Having these functions not accept a lambda function may yield a more efficient implementation in Rust space (which may get relevant with quantum error correction).
- We already have convenience methods in `QuantumCircuit` such as `num_nonlocal_gates` - this PR only supplements what has been started by that method and provides a potentially novice user with a good picture of their quantum circuit.

jakelishman:
I'll close this given I think the total disposition was against merging more opinionated helper methods.

PR #12475, opened 2024-05-29
add a new parameter num_qubits_distribution to random_circuit

### LINKED ISSUE #12059: Controlling the Insertion of Multi-Qubit Gates in the Generation of Random Circuits
### What should we add?

The [generation of random circuits](https://github.com/Qiskit/qiskit/blob/main/qiskit/circuit/random/utils.py) should be extended by the ability to specify a distribution `num_operand_distribution` that specifies the ratio of  1-qubit, 2-qubit, ..., n-qubit gates in the random circuit. For instance, `random_circuit` with `num_operand_distribution={1: 0.5, 2:0.5}` should yield random circuits with roughly 50% single-qubit quantum gates and 50% two-qubit gates while disallowing n-qubit gates with n>2. 

The parameter `max_operands` should take precedence, i.e. if `max_operands` is set, draw a random distribution `num_operand_distribution`. Raise an error if `sum(num_operand_distribution.values()) != 1.0` or if `max_operands` is set at the same time as `num_operand_distribution` or if any key in `num_operand_distribution` is larger than the largest available n-qubit gate (see `get_standard_gate_name_mapping`). 

Please also add tests to validate your changes and add a release note according to https://github.com/Qiskit/qiskit/blob/main/CONTRIBUTING.md 

Note that this is a subtask of #11718. :-)

KoheiSuda:
### Summary
fixes #12059 
This PR extends the random_circuit function in Qiskit to allow users to specify a distribution num_operand_distribution that defines the ratio of 1-qubit, 2-qubit, ..., n-qubit gates in the generated random circuit. For example, calling random_circuit with num_operand_distribution={1: 0.5, 2: 0.5} will produce random circuits with approximately 50% single-qubit gates and 50% two-qubit gates, while disallowing gates with more than 2 qubits.

Additionally, the following points have been addressed:
> The parameter max_operands should take precedence, i.e. if max_operands is set, draw a random distribution num_operand_distribution. Raise an error if sum(num_operand_distribution.values()) != 1.0 or if max_operands is set at the same time as num_operand_distribution or if any key in num_operand_distribution is larger than the largest available n-qubit gate (see get_standard_gate_name_mapping).

### Details and comments
I worked on this issue as part of [unitaryHACK](https://unitaryhack.dev/), so please assign this issue to me if this PR is accepted.

### DIFF
--- a/qiskit/circuit/random/utils.py
+++ b/qiskit/circuit/random/utils.py
@@ -12,8 +12,8 @@
 
 """Utility functions for generating random circuits."""
 
-import numpy as np
 from math import isclose
+import numpy as np
 
 from qiskit.circuit import ClassicalRegister, QuantumCircuit, CircuitInstruction
 from qiskit.circuit import Reset
--- a/qiskit/circuit/random/utils.py
+++ b/qiskit/circuit/random/utils.py
@@ -13,6 +13,7 @@
 """Utility functions for generating random circuits."""
 
 import numpy as np
+from math import isclose
 
 from qiskit.circuit import ClassicalRegister, QuantumCircuit, CircuitInstruction
 from qiskit.circuit import Reset
@@ -60,7 +61,7 @@ def random_circuit(
         CircuitError: when invalid options given
     """
     if num_operand_distribution is not None:
-        if sum(num_operand_distribution.values()) != 1.0:
+        if not isclose(sum(num_operand_distribution.values()), 1.0, rel_tol=1e-7):
             raise CircuitError("The sum of num_operand_distribution values must be 1.0.")
         if max_operands is not None:
             raise CircuitError(
--- a/qiskit/circuit/random/utils.py
+++ b/qiskit/circuit/random/utils.py
@@ -21,7 +21,14 @@ from qiskit.circuit.exceptions import CircuitError
 
 
 def random_circuit(
-    num_qubits, depth, max_operands=4, measure=False, conditional=False, reset=False, seed=None
+    num_qubits,
+    depth,
+    max_operands=None,
+    measure=False,
+    conditional=False,
+    reset=False,
+    seed=None,
+    num_operand_distribution=None,
 ):
     """Generate random circuit of arbitrary size and form.
 
@@ -44,6 +51,7 @@ def random_circuit(
         conditional (bool): if True, insert middle measurements and conditionals
         reset (bool): if True, insert middle resets
         seed (int): sets random seed (optional)
+        num_operand_distribution (dict): A dictionary of the ratio of n-qubit gates
 
     Returns:
         QuantumCircuit: constructed circuit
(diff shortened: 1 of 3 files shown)


sbrandhsn:
Dear Kohei-san, thanks for showing interest in this issue. I will be on vacation and look into this PR next week! In the meantime make yourself familiar with https://github.com/Qiskit/qiskit/blob/main/CONTRIBUTING.md and the unitaryhack rules (if you have not already). :-)

sbrandhsn (changes requested):
Thanks for this contribution @KoheiSuda! :-) This looks like an elegant solution that needs a couple of modifications before it can be accepted.
- There are formatting issues which causes our CI to fail. Please run `tox -eblack` and `tox -elint-incr` locally to figure out what needs to be changed.
- I think I made a mistake when specifying the precedence of `max_operands`, sorry! Having both set, `max_operands` and `num_operand_distribution` set is fine, please revert the initial default value of `max_operands`. If `num_operand_distribution` is not set but `max_operands` is set, draw a random `num_operand_distribution` and continue with your code or use the initial code.
- Please add a test case with a distribution that disallows single-qubit gates, i.e. with {1:0, …}. I think some logic around how single-qubit gates are used to fill the non-used qubits in layer would fail such a test case with the current code.  
- Please add a test case with some distribution and specify a particularly long `depth` as a parameter. This should make your `num_operand_distribution` match closely what is contained in the returned random circuit.
- The CLAAssistant appears to have an issue with your e-mail address, can you fix that? :-)

KoheiSuda:
@sbrandhsn 
Thank you for your thoughtful advice!

> Please add a test case with a distribution that disallows single-qubit gates, i.e. with {1:0, …}. I think some logic around how single-qubit gates are used to fill the non-used qubits in layer would fail such a test case with the current code.

I have a question. If a distribution disallows single-qubit gates, how should I fill the non-used qubits in layer? Is it okay to use single-qubit gates?

sbrandhsn:
Dear @KoheiSuda, I'm sorry but a PR was submitted by a different user that addresses https://github.com/Qiskit/qiskit/issues/12059 completely. Unfortunately, unitaryHACK explicitly allows multiple developers to work on the same issue at once ( https://unitaryhack.dev/rules/ ) and essentially requires us to select the first valid PR for the bounty. Qiskit still has open unitaryHACK issues available at https://github.com/Qiskit/qiskit/issues?q=is%3Aopen+label%3Aunitaryhack-bounty+is%3Aissue that (partially) attracted interest by other developers. 

Apart from this, I was very happy about your elegant contribution so far and can only encourage you to continue your good work on open source projects that you find interesting! :-)

KoheiSuda:
@sbrandhsn 
OK, Thank you for your reviewing!

jakelishman:
I'll close this as I understand the discussion above to mean the PR is obsoleted.

PR #12459, opened 2024-05-25
Add infrastructure for gates, instruction, and operations in Rust

mtreinish:
### Summary

This commit adds a native representation of Gates, Instruction, and Operations to rust's circuit module. At a high level this works by either wrapping the Python object in a rust wrapper struct that tracks metadata about the operations (name, num_qubits, etc) and then for other details it calls back to Python to get dynamic details like the definition, matrix, etc. For standard library gates like Swap, CX, H, etc this replaces the on-circuit representation with a new rust enum StandardGate. The enum representation is much more efficient and has a minimal memory footprint (just the enum variant and then any parameters or other mutable state stored in the circuit instruction). All the gate properties such as the matrix, definiton, name, etc are statically defined in rust code based on the enum variant (which represents the gate).

The use of an enum to represent standard gates does mean a change in what we store on a CircuitInstruction. To represent a standard gate fully we need to store the mutable properties of the existing Gate class on the circuit instruction as the gate by itself doesn't contain this detail. That means, the parameters, label, unit, duration, and condition are added to the rust side of circuit instrucion. However no Python side access methods are added for these as they're internal only to the Rust code. In Qiskit 2.0 to simplify this storage we'll be able to drop, unit, duration, and condition from the api leaving only label and parameters. But for right now we're tracking all of the fields.

To facilitate working with circuits and gates full from rust the setting the `operation` attribute of a `CircuitInstruction` object now transltates the python object to an internal rust representation. For standard gates this translates it to the enum form described earlier, and for other circuit operations 3 new Rust structs: PyGate, PyInstruction, and PyOperation are used to wrap the underlying Python object in a Rust api. These structs cache some commonly accessed static properties of the operation, such as the name, number of qubits, etc. However for dynamic pieces, such as the definition or matrix, callback to python to get a rust representation for those.

Similarly whenever the `operation` attribute is accessed from Python it converts it back to the normal Python object representation. For standard gates this involves creating a new instance of a Python object based on it's internal rust representation. For the wrapper structs a reference to the wrapped PyObject is returned.

To manage the 4 variants of operation (`StandardGate`, `PyGate`, `PyInstruction`, and `PyOperation`) a new Rust trait `Operation` is created that defines a standard interface for getting the properties of a given circuit operation. This common interface is implemented for the 4 variants as well as the `OperationType` enum which wraps all 4 (and is used as the type for `CircuitInstruction.operation` in the rust code.

[...]

PR #12481, opened 2024-05-30
Make fewer accesses, `instruction.operation`, in exporter.py

jlapeyre:
EDIT: To clarify the motivation. This fixes a bug in the exporter that is exposed by #12459 .

Code in qiskit/qasm3/exporter.py uses the object id to track processing gates. But with upcoming changes in #12459 a new object is created every time an operation is accessed within an instruction. Python allows an object's id to be reused after its reference count goes to zero. As a result, sometimes a stored id refers to two different gates (one of them no longer exists as an object) This will cause errors to be raised.

This PR replaces several repeated accesses with a single access. This is also more efficient since each access actually constructs an object.

In particular when testing https://github.com/Qiskit/qiskit/pull/12459  a code path is found in which a gate is accessed, its id is cached, it is freed, and another gate is accessed; all with no intervening statements. This PR prevents the first gate object from being freed until after the second gate is accessed and analyzed.

However, this PR does not remove all possible recycling of ids during a code section that uses them to track gate processing. This design should be replaced with a new approach.

### DIFF
--- a/qiskit/qasm3/exporter.py
+++ b/qiskit/qasm3/exporter.py
@@ -907,6 +907,7 @@ class QASM3Builder:
 
     def build_switch_statement(self, instruction: CircuitInstruction) -> Iterable[ast.Statement]:
         """Build a :obj:`.SwitchCaseOp` into a :class:`.ast.SwitchStatement`."""
+        operation = instruction.operation
         real_target = self.build_expression(expr.lift(operation.target))
         global_scope = self.global_scope()
         target = self._reserve_variable_name(
@@ -935,7 +936,7 @@ class QASM3Builder:
                     target,
                     (
                         case(values, block)
-                        for values, block in instruction.operation.cases_specifier()
+                        for values, block in operation.cases_specifier()
                     ),
                 ),
             ]
@@ -943,7 +944,7 @@ class QASM3Builder:
         # Handle the stabilised syntax.
         cases = []
         default = None
-        for values, block in instruction.operation.cases_specifier():
+        for values, block in operation.cases_specifier():
             self.push_scope(block, instruction.qubits, instruction.clbits)
             case_body = ast.ProgramBlock(self.build_current_scope())
             self.pop_scope()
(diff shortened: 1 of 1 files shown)


ElePT (approved):
I don't think I'm familiar enough with the code to comment on how to avoid the recycling of ids mentioned in the PR description, but the changes from this PR look straightforward to me.

jakelishman (commented):
Given that the exporter is tracking objects by using the `id` and objects are no longer referentially identical for standard gates (but should be for user-defined gates), I don't think this PR is solving the underlying problem, so there's a pretty high chance that there's paths through, or simple changes we might make in the future that will invalidate it.

If we need the exact id objects to hang around, then we need to be storing at least a ref to the `operation` in the symbol table, so that the lifetimes of the use of the id (in the table) and the object itself are bound together.  That said, given that the symbol tables uses `id` at all, and that's no longer stable for standard gates, we need to change the symbol table to handle the storage of standard gates _without_ use of the `id`.

It might be good to look at how hard it is to do that immediately - if we need to merge this as a stop-gap measure, we can, but there's a clear underlying logic bug in the OQ3 exporter that still needs addressing.

jlapeyre:
> I don't think this PR is solving the underlying problem

Right, this is what I meant to say with the last line of the opening comment.

I was working kind of quickly, so I didn't think yet of a better solution. But maybe it's not too difficult. I agree that merging this bandaid is not anything near ideal.

jakelishman:
#12776 implements the type of fix I was talking about above by rewriting how the symbol table works to remove the assumption that `instruction.operation` has a stable ID for standard-gate objects.

The other components of this might still be a performance upgrade - I haven't checked - so this might still be something we want to merge.

1ucian0:
Should this PR be closed in favor of https://github.com/Qiskit/qiskit/pull/12776 ?

jlapeyre:
@1ucian0,  https://github.com/Qiskit/qiskit/pull/12776#discussion_r1683579441 is a comment on the utility of this PR.

* #12776 heavily edits the same code that I touch here. So trying to merge this would not be clean.
* I doubt the performance benefit would be significant. Given this and the previous item, I don't want to invest time in this at present (if ever).

But I'd prefer to wait til #12776 lands in case @jakelishman wants to weigh in.

In the meantime, I converted this to a draft to backburner it.

1ucian0:
Sounds good! Let's do that: wait for the merge of https://github.com/Qiskit/qiskit/pull/12776 and reevaluate this one.

jakelishman:
I'll close this as stale / not necessary.

PR #12482, opened 2024-05-30
Disable Neko on the extended support branch (stable/0.46)

1ucian0:
The neko workflow only supports the most recent stable branch.


mtreinish (approved):
LGTM, we can also remove the `stable/` entry from the branches list but it's not a big deal this will also stop running the job post merge so that's good

PR #12484, opened 2024-05-31
Updated bit.py Bit._repr when no register is assigned

J-C-Q:
### Summary
bit.py Bit._repr with or without a provided register should be consistent.

### Details and comments
When a Bit is created without providing a register and index, ```Bit._repr``` is not set and ```Bit.__repr__(self)``` returns the default string representation of the object. This seems inconsistent to me. I suggest that in this case, ```Bit._repr``` is set to ```f"{self.__class__.__name__}(no register assigned)"``` (or something similar).

### DIFF
--- a/qiskit/circuit/bit.py
+++ b/qiskit/circuit/bit.py
@@ -36,6 +36,7 @@ class Bit:
             # To sidestep the overridden Bit.__hash__ and use the default hash
             # algorithm (only new-style Bits), call default object hash method.
             self._hash = object.__hash__(self)
+            self._repr = f"{self.__class__.__name__}(no register assigned)"
         else:
             try:
                 index = int(index)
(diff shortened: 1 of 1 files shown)


jakelishman (commented):
I'm sorry this went a while with no response. Thanks for thinking about this contribution to Qiskit.

While the current repr isn't perfect, I think we might be a bit better sticking with it - this one will make all registerless bits appear identically, so we won't be able to tell them apart on print. The current form is also very common for Python objects that can't be recreated from their `repr`, and since the important information a `Bit` carries is _only_ its pointer identity, I'm not sure there's much more we can do in the current data model than the Python builtin.

J-C-Q:
I appreciate the detailed response! You are right, I just wanted prettier printing, but I guess being able to tell the instances apart this way makes sense. Feel free to close this PR. :)

PR #12488, opened 2024-05-31
Use hash of numeric value for bound parameter expressions

### LINKED ISSUE #12487: Bound `ParameterExpression`s do not match the hash of an equal object of a numeric type
### Environment

- **Qiskit version**: 1.1.0
- **Python version**: 3.11
- **Operating system**: Fedora Linux 40

### What is happening?

When all the parameters in a `ParameterExpression` are bound, it becomes equal to a `float`, `int`, or `complex`. However, it still uses its own hashing algorithm, so it does not hash to the same value as the numeric type. The Python data model documents that objects that compare equal should hash to the same value.

```


a = Parameter("a")
b = a.bind({a: 1.2})
b == 1.2  # True
hash(b)  == hash(1.2)  # False
```

### What should happen?

As much as possible, the Python data model should be followed. At some points in the parameter handling code, bound `ParameterExpression`s are eagerly cast to `float`. For example, the `float(p)` here:

https://github.com/Qiskit/qiskit/blob/26886a1b2926a474ea06aa3f9ce9e11e6ce28020/qiskit/transpiler/passes/scheduling/scheduling/base_scheduler.py#L71

I think this is done to avoid a bound `p` failing the lookup if the calibrations dict was keyed with a plain `float`. It would be preferable not to have to deal with type casting in these cases.

### Any suggestions?

_No response_

wshanks:
### Summary

If a `ParameterExpression` has no unbound parameters, the underlying bound value can be hashed instead of the tuple that accounts for the symbolic expression. Doing this allows for the `ParameterExpression` to match the hash for the numeric value it compares equal to.

### Details and comments

Closes #12487

### DIFF
--- a/qiskit/circuit/parameterexpression.py
+++ b/qiskit/circuit/parameterexpression.py
@@ -439,6 +439,9 @@ class ParameterExpression:
             raise TypeError("could not cast expression to int") from exc
 
     def __hash__(self):
+        if not self._parameter_symbols:
+            # For fully bound expressions, fall back to the underlying value
+            return hash(self.numeric())
         return hash((self._parameter_keys, self._symbol_expr))
 
     def __copy__(self):
--- a/releasenotes/notes/parameterexpression-hash-d2593ab1715aa42c.yaml
+++ b/releasenotes/notes/parameterexpression-hash-d2593ab1715aa42c.yaml
@@ -1,0 +1,8 @@
+---
+fixes:
+  - |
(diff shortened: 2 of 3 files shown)


jakelishman (dismissed):
Thanks for finding this, Will. This looks right to me.

jakelishman:
Oh, actually, can we have a bugfix release note?

jakelishman:

jakelishman (approved):
Thanks Will.

PR #12490, opened 2024-06-02
Add alternative draw methods to example code

nihaocami:
### Summary

Just adding the `.draw()` separately from directly printing the circuit.

### Details and comments

The original example here `examples/python/circuit_draw.py:32` says "# Use the internal .draw() to print the circuit" however we don't actually see the `.draw()` function being used. Just adding that additional distinction there to avoid any confusion.


1ucian0:
Not sure if I follow what you mean. Printing a circuit will `draw` it (and print it) using `output="text"`.

```
$ python examples/python/circuit_draw.py
          ┌───┐   ┌─┐
q_1: ─────┤ X ├───┤M├
     ┌───┐└─┬─┘┌─┐└╥┘
q_0: ┤ H ├──■──┤M├─╫─
     └───┘     └╥┘ ║
c: 2/═══════════╩══╩═
                0  1
```

It seems to work as expected to me.

nihaocami:
> Not sure if I follow what you mean. Printing a circuit will `draw` it (and print it) using `output="text"`.
> 
> ```
> $ python examples/python/circuit_draw.py
>           ┌───┐   ┌─┐
> q_1: ─────┤ X ├───┤M├
>      ┌───┐└─┬─┘┌─┐└╥┘
> q_0: ┤ H ├──■──┤M├─╫─
>      └───┘     └╥┘ ║
> c: 2/═══════════╩══╩═
>                 0  1
> ```
> 
> It seems to work as expected to me.

If you look at the diff, the old doc says they're using the `draw()` method but it's not actually shown in code which can be confusing. I am just making a more clear distinction between the two. Functionally they are the same.

examples/python/circuit_draw.py
```diff
@@ -29,5 +29,8 @@ def build_bell_circuit():
 # Create the circuit
 bell_circuit = build_bell_circuit()
 
-# Use the internal .draw() to print the circuit
+# Print the circuit directly to stdout
 print(bell_circuit)
+
+# Or use the internal .draw() to print the circuit
+bell_circuit.draw()
```
Cryoris: Since this is a script it won't print anything (unless you copy it over e.g. to a Jupyter notebook cell). A better solution might just be to change the comment on top of `print(circuit)` to just say something like "Print the circuit, which implicitly calls bell_circuit.draw(output="text")" or something?

1ucian0: Alternately, just avoid talking about internals in the comment:
 
```suggestion
# Print the circuit
print(bell_circuit)
```

1ucian0:
> If you look at the diff, the old doc says they're using the draw() method but it's not actually shown in code which can be confusing. 

the key word is "internal" :)

I honestly see no problem. But if `print(bell_circuit.draw())` helps somebody, sure.

examples/python/circuit_draw.py
nihaocami: yeah that's even better! thx

nihaocami: sweet. done

nihaocami:
Your comment made sense. Closing PR. Thx all!

PR #12494, opened 2024-06-03
Create custom 4x4 and 2x2 linalg operations (matmul, kron, determinant)

### LINKED ISSUE #12120: Investigate using something besides ndarray for matmul

arnaucasau:
Closes #12120

This is an alternative to #12193, where instead of using `faer` for matrix multiplications, we have a new module with some functions implemented by hand to calculate the kronecker product, determinant, and matmul using `ndarray` data types.

In the new module we can find the following functions:

- **kron_matrix_x_identity():** A kron function that performs the product of a given matrix and the identity
- **kron_identity_x_matrix():** A kron function that performs the product of the identity and a given matrix.
- **matrix_multiply_2x2():** A 2x2 matmul completely unrolled
- **matrix_multiply_4x4():** A 4x4 matmul using three for loops which I think are easily vectorizable/unrolled for the compiler
- **determinant_4x4():** A function to calculate the determinant of a 4x4 matrix
- **change_basis():** The same function previously living on `convert_2q_block_matrix.rs` and moved here to be reused in more scripts

Running the transpiler benchmarks we can see how the performance has increased:
```

| Change   | Before [9d03b4bf] <main>   | After [a650f588] <AC/custom-linalg>   |   Ratio | Benchmark (Parameter)                                                                                      |
|----------|----------------------------|---------------------------------------|---------|------------------------------------------------------------------------------------------------------------|
| -        | 126±0.8ms                  | 109±0.7ms                             |    0.86 | transpiler_qualitative.TranspilerQualitativeBench.time_transpile_time_cnt3_5_179(2, 'sabre', 'dense')      |
| -        | 161±1ms                    | 125±0.4ms                             |    0.78 | transpiler_qualitative.TranspilerQualitativeBench.time_transpile_time_cnt3_5_179(3, 'sabre', 'dense')      |
| -        | 158±0.4ms                  | 133±0.3ms                             |    0.84 | transpiler_qualitative.TranspilerQualitativeBench.time_transpile_time_cnt3_5_179(3, 'sabre', 'sabre')      |
| -        | 249±0.4ms                  | 212±0.9ms                             |    0.85 | transpiler_qualitative.TranspilerQualitativeBench.time_transpile_time_cnt3_5_179(3, 'stochastic', 'sabre') |
| -        | 264±0.7ms                  | 222±3ms                               |    0.84 | transpiler_qualitative.TranspilerQualitativeBench.time_transpile_time_cnt3_5_180(2, 'sabre', 'dense')      |
| -        | 302±0.2ms                  | 264±0.5ms                             |    0.87 | transpiler_qualitative.TranspilerQualitativeBench.time_transpile_time_cnt3_5_180(2, 'sabre', 'sabre')      |
| -        | 564±2ms                    | 512±0.8ms                             |    0.91 | transpiler_qualitative.TranspilerQualitativeBench.time_transpile_time_cnt3_5_180(2, 'stochastic', 'sabre') |
| -        | 347±0.7ms                  | 259±0.4ms                             |    0.75 | transpiler_qualitative.TranspilerQualitativeBench.time_transpile_time_cnt3_5_180(3, 'sabre', 'dense')      |
| -        | 391±0.5ms                  | 306±0.5ms                             |    0.78 | transpiler_qualitative.TranspilerQualitativeBench.time_transpile_time_cnt3_5_180(3, 'sabre', 'sabre')      |
| -        | 719±0.9ms                  | 599±0.2ms                             |    0.83 | transpiler_qualitative.TranspilerQualitativeBench.time_transpile_time_cnt3_5_180(3, 'stochastic', 'sabre') |
```

See this comment to see the benchmarks of all the operations and for more information about the other approach.

### DIFF
--- a/crates/accelerate/src/common.rs
+++ b/crates/accelerate/src/common.rs
@@ -1,0 +1,113 @@
+// This code is part of Qiskit.
+//
+// (C) Copyright IBM 2024
+//
+// This code is licensed under the Apache License, Version 2.0. You may
+// obtain a copy of this license in the LICENSE.txt file in the root directory
+// of this source tree or at http://www.apache.org/licenses/LICENSE-2.0.
+//
+// Any modifications or derivative works of this code must retain this
+// copyright notice, and modified files need to carry a notice indicating
+// that they have been altered from the originals.
+
+use num_complex::Complex64;
+use numpy::ndarray::{array, Array2, ArrayView2};
+
+#[inline]
+pub fn kron_matrix_x_identity(lhs: ArrayView2<Complex64>) -> Array2<Complex64> {
+    let zero = Complex64::new(0., 0.);
+    array![
+        [lhs[(0, 0)], zero, lhs[(0, 1)], zero],
+        [zero, lhs[(0, 0)], zero, lhs[(0, 1)]],
+        [lhs[(1, 0)], zero, lhs[(1, 1)], zero],
+        [zero, lhs[(1, 0)], zero, lhs[(1, 1)]],
+    ]
+}
+
+#[inline]
+pub fn kron_identity_x_matrix(rhs: ArrayView2<Complex64>) -> Array2<Complex64> {
+    let zero = Complex64::new(0., 0.);
+    array![
+        [rhs[(0, 0)], rhs[(0, 1)], zero, zero],
+        [rhs[(1, 0)], rhs[(1, 1)], zero, zero],
+        [zero, zero, rhs[(0, 0)], rhs[(0, 1)]],
+        [zero, zero, rhs[(1, 0)], rhs[(1, 1)]],
+    ]
+}
+
+#[inline]
+pub fn matrix_multiply_2x2(
+    a: ArrayView2<Complex64>,
+    b: ArrayView2<Complex64>,
+) -> Array2<Complex64> {
+    let mut result = Array2::uninit((2, 2));
+
+    result[(0, 0)].write(a[(0, 0)] * b[(0, 0)] + a[(0, 1)] * b[(1, 0)]);
+    result[(0, 1)].write(a[(0, 0)] * b[(0, 1)] + a[(0, 1)] * b[(1, 1)]);
+    result[(1, 0)].write(a[(1, 0)] * b[(0, 0)] + a[(1, 1)] * b[(1, 0)]);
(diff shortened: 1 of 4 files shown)


crates/accelerate/src/common.rs
```diff
@@ -0,0 +1,113 @@
...
+            }
+            result[(i, j)].write(sum);
+        }
+    }
+
+    unsafe { result.assume_init() }
+}
+
+#[inline]
+pub fn determinant_4x4(a: ArrayView2<Complex64>) -> Complex64 {
```
jlapeyre, line 74: It was pointed out earlier that this naive expression for the determinant carries a large error in some cases. We confirmed it with some examples. I strongly suspect that the large error will *not* be present for unitary matrices. But I don't have evidence for this.

arnaucasau: Oh, interesting. I didn't know about that. Out of curiosity, do you have the examples you tried? I wasn't able to find anything about it in the search I did or in the [paper](https://www.researchgate.net/publication/275580759_New_Method_to_Compute_the_Determinant_of_a_4x4_Matrix)

Edit: Never mind, I can see the error calculating the determinant for random matrices. Thanks for pointing that out John

PR #12495, opened 2024-06-03
Implement `num_qubits` and `num_clbits` in Rust

jlapeyre:
* `QuantumCircuit.num_qubits` and `num_clbits` are twice as fast after this PR.
* These are used in several places. For example `QuantumCircuit.width()` is three times faster.
* `num_qubits` and `num_clbits` are introduced in circuit_data.rs. These functions are called by the corresponding Python methods. They are also used in circuit_data.rs itself.

### DIFF
--- a/crates/circuit/src/circuit_data.rs
+++ b/crates/circuit/src/circuit_data.rs
@@ -255,6 +255,11 @@ impl CircuitData {
         self.clbits_native.len()
     }
 
+    /// Return the width of the circuit. This is the number of qubits plus the
+    /// number of clbits.
+    ///
+    /// Returns:
+    ///     int: The width of the circuit.
     pub fn width(&self) -> usize {
         self.num_qubits() + self.num_clbits()
     }
--- a/crates/circuit/src/circuit_data.rs
+++ b/crates/circuit/src/circuit_data.rs
@@ -220,6 +220,11 @@ impl CircuitData {
         self.qubits.clone_ref(py)
     }
 
+    /// Return the number of qubits. This is equivalent to the length of the list returned by
+    /// :meth:`.CircuitData.qubits`
+    ///
+    /// Returns:
+    ///     int: The number of qubits.
     #[getter]
     pub fn num_qubits(&self) -> usize {
         self.qubits_native.len()
@@ -240,6 +245,11 @@ impl CircuitData {
         self.clbits.clone_ref(py)
     }
 
+    /// Return the number of clbits. This is equivalent to the length of the list returned by
+    /// :meth:`.CircuitData.clbits`.
+    ///
+    /// Returns:
+    ///     int: The number of clbits.
     #[getter]
     pub fn num_clbits(&self) -> usize {
         self.clbits_native.len()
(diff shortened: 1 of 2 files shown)


crates/circuit/src/circuit_data.rs
```diff
@@ -220,6 +220,11 @@ impl CircuitData {
         self.qubits.clone_ref(py)
     }
 
+    #[getter]
```
mtreinish, line 228: Do you want to throw a docstring on this and the other new methods just for completeness? While we don't publish docs for this as a pyclass (since it's internal only), it sometimes is a nice hint for people using IDEs that show the `__doc__` attribute of a python class.

mtreinish (commented):
This LGTM, in general as the source of truth moves more to the rust side I think this is fine. 

There is an open question about a rust native representation of bits after #12459 but in the meantime asserting the bit pyobject vec in the circuit data is the source of truth seems fine to me (which this PR is doing implicitly), especially once we move to the gate representation being owned by the rust domain.

jlapeyre:
> asserting the bit pyobject vec in the circuit data is the source of truth seems fine to me

It wasn't 100% clear to me that `qubits_native: Vec<PyObject>` is already the fundamental object. But the word "cached"  here
https://github.com/Qiskit/qiskit/blob/797bb285632a4b58c35c9e4cf5be44502b5bd78a/crates/circuit/src/circuit_data.rs#L151-L152

made it seem more likely.

> an open question about a rust native representation of bits

Unfortunately, I expect there will be a lot of churn like this is gradually moving things to Rust.  But this PR is pretty light in terms of changed code, so not too onerous to replace.

crates/circuit/src/circuit_data.rs
jlapeyre: Oh yeah. I forgot to add some of this when I moved it out of draft.

mtreinish:
> It wasn't 100% clear to me that qubits_native: Vec<PyObject> is already the fundamental object. But the word "cached" here

My mistake tou're right it already is, I was misremembering the state of the code. The Python side `QuantumCircuit.qubits` is returning the inner `QuantumCircuit._data.qubits` already. So there is no concern at all with doing this via rust for improved efficiency.

crates/circuit/src/circuit_data.rs
jlapeyre: Done in 2ffac76

crates/circuit/src/circuit_data.rs
```diff
@@ -235,6 +245,20 @@ impl CircuitData {
...
+    /// :meth:`.CircuitData.clbits`.
+    ///
+    /// Returns:
+    ///     int: The number of clbits.
+    #[getter]
+    pub fn num_clbits(&self) -> usize {
+        self.clbits_native.len()
+    }
+
+    pub fn width(&self) -> usize {
```
mtreinish, line 263: This is missing a docstring too.

jlapeyre: Done here 16755cd

mtreinish (approved):

PR #12496, opened 2024-06-03
Add __pos__ for Parameter

eendebakpt:
### Summary

Add `__pos__` for `Parameter` and `ParameterExpression`

### Details and comments

The implementation is simular to the `__neg__`

### DIFF
--- a/qiskit/circuit/parameterexpression.py
+++ b/qiskit/circuit/parameterexpression.py
@@ -328,10 +328,10 @@ class ParameterExpression:
         return self._apply_operation(operator.mul, other)
 
     def __pos__(self):
-        return self._apply_operation(operator.mul, 1.0)
+        return self._apply_operation(operator.mul, 1)
 
     def __neg__(self):
-        return self._apply_operation(operator.mul, -1.0)
+        return self._apply_operation(operator.mul, -1)
 
     def __rmul__(self, other):
         return self._apply_operation(operator.mul, other, reflected=True)
--- a/qiskit/circuit/parameterexpression.py
+++ b/qiskit/circuit/parameterexpression.py
@@ -327,6 +327,9 @@ class ParameterExpression:
     def __mul__(self, other):
         return self._apply_operation(operator.mul, other)
 
+    def _pos__(self):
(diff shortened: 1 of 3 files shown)


qiskit/circuit/parameterexpression.py
```diff
@@ -327,6 +327,9 @@ def __rsub__(self, other):
     def __mul__(self, other):
         return self._apply_operation(operator.mul, other)
 
+    def _pos__(self):
```
jakelishman:
```suggestion
    def __pos__(self):
```

jakelishman (commented):
This seems sensible in spirit to me, thanks.

I guess it would have made more sense if we were able to use `operator.pos` in the `_apply_operation`, but that function seems to be set up to assume binary operations, so never mind on that front.

I'm not wild about introducing a floating-point number with this, though, and the same goes for `__neg__`.  Definitely feels like we might at least want to make them multiplication by _integers_ rather than floats (if we can't just use `operator.pos` and `operator.neg` directly) to avoid introducing imprecisions into the evaluation contexts.

Please could you also add a circuits feature release note about this?

eendebakpt:
> This seems sensible in spirit to me, thanks.
> 
> I guess it would have made more sense if we were able to use `operator.pos` in the `_apply_operation`, but that function seems to be set up to assume binary operations, so never mind on that front.
> 
> I'm not wild about introducing a floating-point number with this, though, and the same goes for `__neg__`. Definitely feels like we might at least want to make them multiplication by _integers_ rather than floats (if we can't just use `operator.pos` and `operator.neg` directly) to avoid introducing imprecisions into the evaluation contexts.
> 
> Please could you also add a circuits feature release note about this?

Multiplication with an int seems nicer indeed. I can change it for this PR, but I am not sure whether there are any implications for `__neg__`. Let me know

jakelishman:
Let's make at least `__pos__` be multiplication by integer here. I think we ought to change `__neg__` too - if the tests don't complain about it (and hopefully nothing in our test suite depends on that questionable behaviour...), but happy to do that in a follow-up if you prefer.

jakelishman (approved):
Ace, thanks!

PR #11614, opened 2024-01-22
[WIP] more general block collection

alexanderivrii:
### Summary

This PR aims to start a discussion on the proper interface for the _extended_ `BlockCollector` class and for the transpiler passes that can work with either `DAGCircuit` or `DAGDependency`.

**Extended `BlockCollector` interface**

The existing `BlockCollector` class allows to greedily collect maximal blocks of nodes matching a given filter function. This serves to collect blocks of linear gates, blocks of Clifford gates, blocks of 1-qubit gates, blocks of 2-qubits gates, and so on. In addition, `BlockCollector` can exploit commutativity between gates by using the `DAGDependency` representation of the circuit.

However, this does not encompass use-cases where the decision of whether to add an eligible node to the current block should depend both on the node _and on the current block_. Furthermore, we may want the collected blocks to carry more information that just the list of nodes contained in the block.

Consider the use-case of collecting and reimplementing "star-shaped" subcircuits (see PR #11387). Deciding whether a given 2-qubit gate can be added to the current "star" depends on whether the star's center qubit has already been chosen and whether this center qubit is part of the node's qubits. 

Or consider the use-case of collecting blocks of nodes all controlled by the same set of qubits. Deciding whether a node can be added to the current block depends on whether this node is also controlled by the same qubits as the other nodes in the block. 

This PR proposes the following extension of the `BlockCollector` interface. There is a new abstract class called `Block`; previous block collection strategies work with the `DefaultBlock` subclass; collecting blocks from a `DAGCircuit`/`DAGDependency` returns a list of these blocks (rather than a list of lists of nodes); yet the change is backward compatible with the help of an additional argument `output_nodes`.

As an example, a `StarBlock` used to collect star-shaped subcircuits contained the list of nodes in the block, the "center" node of the star, and the number of 2-qubit gates in the block.

**Star to linear prerouting pass**

This PR includes a specific example of the star-to-line prerouting pass: the relevant code is copied/adapted from PR #11387. However, an additional effort is taken to make both pieces of collecting star-shaped subcircuits and rewriting star-shaped subcircuits work with either `DAGCircuit` and `DAGDependency`. This is greatly simplified by the new DAG-dependency interface proposed in PR #11705, and for this reason this PR is implemented on top of #11705.

[...]

PR #12498, opened 2024-06-04
Extended block collection 

alexanderivrii:
### Summary

This PR extends the functionality of the `BlockCollector` class, replacing the previous PR #11614. The key new feature is that the check whether a node should be added to the current block may now depend both on the ``filter_fn`` (which specifies _which_ nodes should be collected) _and the block collected so far_. The logic of the `StarPreRouting` pass has been simplified to use the 
`BlockCollector` functionality instead of essentially implementing this in-place. In addition to avoiding code duplication, this also allows to extend the `StarPreRouting` pass with additional options such as using `DAGDependency` dag representation or collecting star-shaped blocks nodes from the back of the circuit (this is not done in this PR since it requires #12418). Note that for `StarPreRouting` the logic of whether a node can be added to the block collected so far depends on the block, i.e. whether it remains star-shaped. 

The new class `Block` defines the abstract API that a block class should satisfy in order to use it for block collection. The main method is ``append_node`` which specifies whether a node can be added to the current block. The method ``size`` returns the size of the block which is useful when filtering blocks based on the ``min_block_size`` argument. The method ``get_nodes`` is primarily needed for backwards compatibility and should return the list of nodes in the block (however it would not be called when using block collection with a specified block class). The method ``reverse`` needs to be implemented to support collecting nodes from the back of the circuit. The method ``split`` needs to be implemented to support splitting collected blocks into smaller sub-blocks.

The new class `DefaultBlock` implements the simple functionality that can be used when the decision of whether to add a node in the current block only depends on ``filter_fn`` and not on the block collected so far. 

The block collection class and functionality are backward compatible, any code that worked before should remain working. (This is why ``collect_all_matching_blocks`` returns a list of blocks where each block is either a subclass of `Block` or a list of nodes). However, the previously public methods ``collect_matching_block`` and ``collect_all_matching_blocks`` in `StarPreRouting` were completely removed (making this specific change not backward-compatible). It was an oversight to make this methods public in the first place, and unfortunately these methods don't make much sense outside of `BlockCollector`.


qiskit/dagcircuit/collect_blocks.py
```diff
@@ -25,6 +27,139 @@
...
+
+        Args:
+            split_blocks (bool): specifies to split the block into sub-blocks
+                over disconnected qubit subsets.
+            split_layers (bool): specifies to split the blocks into layers
+                of non-overlapping instructions (in other words, into depth-1
+                sub-blocks).
+
+        """
+        raise NotImplementedError
```
henryzou50: I like how well-documented this abstract base class is and it is well designed. One potential improvement here is clarifying the return times in order to make the return types consistent. 

For methods like `get_nodes`, `reverse`, and `split`, which could return `None`, it might be more intuitive to always return a list. Here an empty list could signify the same thing as `None`, but it would also make the return type consistent, and thus allow the methods to be used in a wider range of contexts without having to check for `None`.

qiskit/transpiler/passes/routing/star_prerouting.py
```diff
@@ -11,34 +11,31 @@
...
 from qiskit.circuit import Barrier
 from qiskit.dagcircuit import DAGOpNode, DAGDepNode, DAGDependency, DAGCircuit
+from qiskit.dagcircuit.collect_blocks import Block, BlockCollector
 from qiskit.transpiler import Layout
 from qiskit.transpiler.basepasses import TransformationPass
 from qiskit.circuit.library import SwapGate
 
 
-class StarBlock:
+class StarBlock(Block):
```
henryzou50: From an architectural standpoint, I think it makes more sense for `StarBlock` class to be placed in the `collect_blocks.py`. Here are some reasons for moving it:
- Separation of concerns: `collect_blocks,py is focused on defining different types of blocks and the logic for collecting them. Since `StarBlock is a specific type of block, its definition fits naturally within this module.
- Reusability: It might be useful to use `StarBlock` outside of `star_prerouting` and having `StarBlock` in `collect_blocks.py` makes it easier to reuse. For instance, If other passes or functionalities require star-shaped block logic, they can import `StarBlock` from a centralized module.

qiskit/transpiler/passes/routing/star_prerouting.py
```diff
@@ -288,10 +166,26 @@ def determine_star_blocks_processing(
...
+                and getattr(node.op, "condition", None) is None
+                and not isinstance(node.op, Barrier)
+            )
+
+        block_collector = BlockCollector(dag)
+        matching_blocks = block_collector.collect_all_matching_blocks(
+            filter_fn=filter_fn,
+            block_class=StarBlock,
+            output_nodes=False,
+            min_block_size=min_block_size,
```
henryzou50: Since the `split` method for the `Starblock` does not do anything at the moment, and `collect_all_matching_blocks` defaults `split_block=True`, we should add the parameter that `split_blocks=False` to make it clear that we are not splitting at this point.

henryzou50: This also makes me a bit confused on why we did not catch a `DAGCircuitError` in our testing of `star_prerouting`during the section of `collect_all_matching_block`:

```python
        if split_blocks or split_layers:
            tmp_blocks = []
            for block in matching_blocks:
                sub_blocks = block.split(split_blocks=split_blocks, split_layers=split_layers)
                if sub_blocks is None:
                    raise DAGCircuitError(
                        "The block class does not implement the method ``split``."
                    )
                tmp_blocks.extend(sub_blocks)
            matching_blocks = tmp_blocks
 ```

File: qiskit/transpiler/passes/routing/star_prerouting.py [hunk 4]
```diff
@@ -70,10 +67,31 @@ def append_node(self, node):
...
+        return self.nodes
+
+    def split(self, split_blocks, split_layers):
+        """
+        In theory, this method can be used to split a star-shaped
+        block into star-shaped sub-blocks. For now, this does not
+        do anything and simply returns the list consisting of this
+        very block.
+        """
+        return [self]
```
henryzou50: We might want to include in the release notes that the `split` method for `StarBlock` acts as a placeholder and is not fully implemented yet.

qiskit/transpiler/passes/routing/star_prerouting.py
henryzou50: Oh I see, we do not catch the `DAGCircuitError` here as the `split` method for `StarBlock` always returns `[self]`, which is not `None`. Still we should have `split_blocks=False` for this section to avoid confusion and unnecessary compute.

henryzou50 (commented):
Overall, the API changes LGTM, and is well designed. I left a few comments about the structure and details about the `split` method.

alexanderivrii:
Sorry for the late reply. I think at this point it is better to merge #12761 first, and then to think how to best make use of the extended block collection class within `StarPreroute`. So for now I have reverted the changes to the `StarPreroute` pass (and updated the release notes accordingly), see [1a5ef91](https://github.com/Qiskit/qiskit/pull/12498/commits/1a5ef91b27d920b8c582bac2cab10805a81aec19).

qiskit/dagcircuit/collect_blocks.py
alexanderivrii: I think treating an empty list of nodes as `None` might be misleading, and I think it's safer to return `None` in the case that a certain method for a given `Block` type is not supported. I clarified the return types in [1bab966](https://github.com/Qiskit/qiskit/pull/12498/commits/1bab966e6b35794fdcee062e54ec1ddd38066c49).

qiskit/transpiler/passes/routing/star_prerouting.py
alexanderivrii: This and other comments regarding `star_preroute.py` are no longer relevant after in [1a5ef91](https://github.com/Qiskit/qiskit/pull/12498/commits/1a5ef91b27d920b8c582bac2cab10805a81aec19) I've reverted the changes to this file, in order not to interfere with #12761.

I completely agree that it would be best to create a "library" of useful and reusable block types (like `DefaultBlock`). I am not completely convinced that specifically `StarBlock` is going to be useful beyond the `StarPrerouting` pass, so I am not sure we should expose this particular block type. In addition, I have locally tried to improve the quality of `StarPreroute` pass by experimenting with different variants of the `StarBlock` (for example only collecting stars with non-repeating non-center nodes, the intuition being that repeating nodes prevent mapping a star to a line, or only collecting stars consisting of pairwise commuting gates). Unfortunately, I did not find anything that made `StarPrerouting` better on say QAOA circuits, however I do find likely that a good star block for the star prerouting pass would be very specific to this pass and thus should be packaged together with the pass.

qiskit/transpiler/passes/routing/star_prerouting.py
alexanderivrii: Yeah, I agree, the previous handling was sloppy, as you have pointed out.

qiskit/transpiler/passes/routing/star_prerouting.py
alexanderivrii: I agree, though this is no longer relevant at the moment.

alexanderivrii:
Hi @henryzou50, can you please take another look at this PR and see what's missing? Thanks!

alexanderivrii:
I am closing this PR since it's stale and should be reimplemented in rust.

PR #12507, opened 2024-06-04
Implement some standard gates in rust

jlapeyre:
#### This PR ...
*  Implements `RGate` in Rust
* Does some refactoring and rewriting to reduce boilerplate and improve ergonomics. These are applied throughout all crates.
    * `ZERO`, `ONE`, `IM`, `M_IM`, `M_ONE`, etc.
    * `c64!` and method `c64` for constructing `Complex64`. Perhaps we only need the former.
    * `GateArray2Q` and similar type aliases for very common types (`[[Complex64; 4]; 4];`)

This PR builds on:
*  #12459

This PR is blocked by #12459.

Here is a diff with far fewer commits,  taking #12459 as the base:  https://github.com/mtreinish/qiskit-core/compare/gates-in-rust...jlapeyre:qiskit-core:new-some-gates-in-rust

~It really looks like the online diff between this PR branch and #12459 is broken. The diff does not agree with a local diff.~ Seems to have been a GH glitch that has resolved.


crates/circuit/src/operations.rs
```diff
@@ -602,6 +630,7 @@ impl Operation for StandardGate {
                 )
             }),
             Self::UGate => None,
+            Self::RGate => todo!("Add when we have U3"),
```
mtreinish: FWIW, what I did on the other gates that depending on `U3` was to just switch them to use `U` instead for their definitions. The whole `U3`, `U2`, `U1` thing is an artifact of the old IBM basis set and we've just never updated the definitions.

jlapeyre: I implemented some of `define` for `RGate` in terms of `U`. I think there are four cases for input types and I did two of them. But I noticed that these definitions are not used anywhere at the moment; there is even a comment to this effect. So I don't plan to continue.

jlapeyre:
I am closing this in favor of splitting this PR into two parts.

* Part of this PR will be done instead in #12651

PR #12511, opened 2024-06-05
fix issue/12311 with fractional gate in basis

### LINKED ISSUE #12311: 2Q gate synthesis with fractional gates
### Environment

- **Qiskit version**: main
- **Python version**: doesn't matter
- **Operating system**: doesn't matter

### What is happening?

When I build a Qiskit `Target` that supports two qubit basis gates with arbitrary rotation angle, it cannot transpile the circuit if the added basis gate is other than `RZXGate` or `RXXGate`.

```python
from qiskit import quantum_info as qi, transpile, QuantumCircuit

backend = GenericBackendV2(2, basis_gates=["rz", "rx", "rzz"])
qc = QuantumCircuit(2)
qc.unitary(qi.random_unitary(4), [0, 1])
isa = transpile(qc, target=backend.target)
```
this results in an error
```
TypeError: ParameterExpression with unbound parameters (dict_keys([Parameter(ϴ)])) cannot be cast to a float.
```
because it tries to get the unitary matrix of parameterized `RZZGate(Parameter(ϴ))` instruction.

On the other hand, it transpiles circuit when I directly specify the basis gates.
```python
isa = transpile(qc, basis_gates=["rz", "rx", "rzz"])
```

### What should happen?

- Qiskit transpiler should support arbitrary fractional two qubit basis gates, e.g. `RZX, RXX, RYY, RZZ, RXY, ...`. 
- Transpiling with `Target` and giving `basis_gates` should behave identically.

This is related but might be a feature request; even if I give `RZXGate(Parameter(ϴ))` in the target, current synthesis algorithm only picks pi/4 rotation. The synthesis pass should consider another angle if there is possibility of reducing the circuit depth.

### Any suggestions?

No. I'm not familiar with synthesis passes.

ewinston:
### Summary
This fixes an issue where an exception would be raised if transpiling a unitary gate to a backend with an RZZGate in its basis however transpiling with only `basis_gates` specified would transpile fine. 

fixes #12311 

### Details and comments
In this case, where RZZ is the only entangling gate in the basis, decomposition of a 2q unitary gate wouldn't get synthesized in the `UnitarySynthesis` pass but would instead be synthesized in the `HighLevelSynthesis` pass because this would call the instruction's `_define` method which calls `two_qubit_cnot_decompose`. A subsequent `BasisTranslator` pass ultimately maps the cx to the rzz basis.  I'm not sure whether this indirect path to unitary synthesis is intentional.

This pr resolves the specific example posed in the issue.


qiskit/transpiler/passes/synthesis/unitary_synthesis.py
```diff
@@ -782,15 +782,15 @@ def _replace_parameterized_gate(op):
         def is_supercontrolled(gate):
             try:
                 operator = Operator(gate)
-            except QiskitError:
+            except (QiskitError, TypeError):  # TypeError: parametrized gate
```
mtreinish: What is the source of a `TypeError` here? Is it coming from `__array__` ? 

I ask because I'm wondering if this is the correct fix, if it's coming from `__array__`  because the gate is parameterized in the target, I'm wondering if we should instantiate it with a fixed angle instead. Basically update `_replace_parameterized_gate` to handle `RZZ` too, or instead maybe try to generalize it. A parameterized gate in the target represents that any angle value is valid input so we're free to evaluate the matrix at a fixed angle if we want.

ewinston: Yes this seems to be the cause. I assigned it as done in `_replace_parameterized_gate` for now. I don't love the idea of implicitly making that choice here but at least there's precedence. [c45d02d](https://github.com/Qiskit/qiskit/pull/12511/commits/c45d02d0f090f849b23303f5098fe49a6ffb42d7)

test/python/transpiler/test_basis_translator.py
```diff
@@ -1223,3 +1228,48 @@ def __init__(self):
...
+        seed = 9169
+        basis_gates = ["rz", "rx", "rzz"]
+        qc = QuantumCircuit(num_qubits)
+        mat = scipy.stats.unitary_group.rvs(2**num_qubits, random_state=seed)
+        qc.unitary(mat, range(num_qubits))
+        backend = GenericBackendV2(num_qubits, basis_gates=basis_gates)
+        target = backend.target
+        pm = generate_preset_pass_manager(optimization_level=1, target=target, seed_transpiler=134)
+        cqc = pm.run(qc)
+        self.assertEqual(Operator(qc), Operator(cqc))
```
mtreinish: Not that it matters here because there aren't any connectivity constraints so layout and routing won't run. We could use `Operator.from_circuit` here to reverse the permutations caused by routing and layout.

```suggestion
        self.assertEqual(Operator(qc), Operator.from_circuit(cqc))
```

test/python/transpiler/test_basis_translator.py
```diff
@@ -1223,3 +1228,48 @@ def __init__(self):
...
+        qc.unitary(mat, range(num_qubits))
+        target = Target()
+        target.add_instruction(RZGate(self.theta), {(i,): None for i in range(qc.num_qubits)})
+        target.add_instruction(RXGate(self.phi), {(i,): None for i in range(qc.num_qubits)})
+        target.add_instruction(
+            RZZGate(self.lam), {(i, i + 1): None for i in range(qc.num_qubits - 1)}
+        )
+        pm = generate_preset_pass_manager(optimization_level=1, target=target, seed_transpiler=134)
+        cqc = pm.run(qc)
+        self.assertEqual(Operator(qc), Operator(cqc))
```
mtreinish: We definitely should do this here though, because there isn't a guarantee with a coupling map that the transpiler won't permute anything. Although this case is trivial and unlikely to swap it might pick a different layout in the future (if we change the construction of the pass manager at level1).

```suggestion
        self.assertEqual(Operator(qc), Operator.from_circuit(cqc))
```

mtreinish (commented):
This LGTM, I just have a small nit in the tests for future proofing (since I just had to deal with related failures caused by it on a different PR). I'll just apply that and enqueue this for merge.

mtreinish (approved):

PR #12512, opened 2024-06-05
Fix version table in qpy docs

mtreinish:
### Summary

The QPY docs included a format version table that matched up the Qiskit releases to the supported QPY format versions for that release. However, a typo resulted in it being treated as a comment instead of table this commit fixes this and a small copy paste error in one of the versions listed in the table.

### Details and comments


ElePT (approved):
LGTM!

PR #12520, opened 2024-06-06
Bump pylint to 3.2.2

### LINKED ISSUE #12510: AttributeError when running pylint in python 3.12
### Environment

- **Qiskit version**:  1.2.0
- **Python version**:  3.12.1
- **Operating system**:  Windows 11

### What is happening?

When I try to run `tox -elint` after modifying some files locally this is what happens:
- black runs successfully
- ruff runs successfully
- pylint throws error!!

The error that I'm getting is not due to some linting issues in my code but instead its an 
`AttributeError: 'TreeRebuilder' object has no attribute 'visit_typealias'` 
I should mention that I am not getting any error when installing the editable version of `qiskit` after cloning the repo. So the source of this error is not due to some installation problem.

I did some google search and I found this issue on `pylint-dev` [github page](https://github.com/pylint-dev/pylint/issues/8782). Apparently this is an issue that is related to python 3.12 and this issue has been fixed sometime in August 2023. But the pylint version that we use from `requirements-dev.txt` is `pylint==2.16.2` (released in Feb 2023). Can we update the pylint version that `requirements-dev.txt` has (I know this is an involved task) because `qiskit` now supports python 3.12.

Let me know if you want the full error message, I'm not including that full error message here becasue it is very very long!

Run `tox -elint` on the qiskit code in python 3.12.1 virtual environment.

### What should happen?

The command `tox -elint` should not give any AttributeError.

### Any suggestions?

Maybe we will have to upgrade `pylint` to a version that was probably released after August 2023 (this is when this issue was fixed on pylint-dev repo, see the link in the description)

jakelishman:
### Summary

This upgrades pylint to a version compatible with Python 3.8 through 3.12, like Qiskit.  There are a couple of false positives (pylint is incorrect about its `use-yield-from` in the cases it flags), and has included a couple of new stylistic opinions that are not necessary to enforce.

The `possibly-used-before-assignment` lint is quite possibly a good one, but there are far too many instances in the Qiskit codebase right now to fix, whereas our current version of pylint is preventing us from running it with Python 3.12.

### Details and comments

pylint-dev/pylint#9700 will fix the `use-yield-from` false positives we're suppressing here, but there's probably no need to wait for 3.2.3 to release for this PR; we can remove the suppressions later.

- Close #12320 (obsolete)
- Fix #12510

### DIFF
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -17,8 +17,8 @@ black[jupyter]~=24.1
 #
 # These versions are pinned precisely because pylint frequently includes new
 # on-by-default lint failures in new versions, which breaks our CI.
-astroid==2.14.2
-pylint==2.16.2
+astroid==3.2.2
+pylint==3.2.2
 ruff==0.0.267
 
 
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -18,7 +18,7 @@ black[jupyter]~=24.1
 # These versions are pinned precisely because pylint frequently includes new
 # on-by-default lint failures in new versions, which breaks our CI.
 astroid==3.2.2
-pylint==3.2.2
+pylint==3.2.3
 ruff==0.0.267
 
 
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -213,6 +213,7 @@ disable = [
     "docstring-first-line-empty", # relax docstring style
     "import-outside-toplevel", "import-error", # overzealous with our optionals/dynamic packages
     "nested-min-max", # this gives false equivalencies if implemented for the current lint version
+    "consider-using-max-builtin", "consider-using-min-builtin",  # unnecessary stylistic opinion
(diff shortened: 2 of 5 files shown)


requirements-dev.txt
```diff
@@ -17,8 +17,8 @@ black[jupyter]~=24.1
 #
 # These versions are pinned precisely because pylint frequently includes new
 # on-by-default lint failures in new versions, which breaks our CI.
-astroid==2.14.2
-pylint==2.16.2
+astroid==3.2.2
+pylint==3.2.2
```
Pierre-Sassoulas:
```suggestion
astroid==3.2.2
pylint==3.2.3
```

Pierre-Sassoulas (commented):
We released your fix 10mn ago :)

requirements-dev.txt
jakelishman: Oh wow, that was fast thank you!!

mtreinish (approved):

PR #12526, opened 2024-06-06
Avoid exception in `Target.has_calibration` for instruction without properties

### LINKED ISSUE #12525: `Target.has_calibration` errors for instructions without properties
### Environment

- **Qiskit version**: 1.1.0
- **Python version**: 3.12
- **Operating system**: Fedora Linux 40

### What is happening?

If a `Target` has an `Instruction` with `None` for its `InstructionProperties`, `Target.has_calibration` raises an `AttributeError`.

Consider this case from the `Target` docstring with the `has_calibration` call added at the end:

```python
from qiskit.circuit.library import CXGate
from qiskit.transpiler import Target, InstructionProperties

target = Target()
cx_properties = {
    (0, 1): None,
    (1, 0): None,
    (0, 2): None,
    (2, 0): None,
    (0, 3): None,
    (2, 3): None,
    (3, 0): None,
    (3, 2): None
}
target.add_instruction(CXGate(), cx_properties)

target.has_calibration("cx", (0, 1))  # AttributeError!
```

### What should happen?

The last line above should return `False`

### Any suggestions?

I will submit a PR.

wshanks:
### Summary

Avoid exception in `Target.has_calibration` for instruction without properties

### Details and comments

`Target.add_instruction` allows passing `None` in place of an `InstructionProperties` instance. In this case, there will be no `_calibration` attribute, so the `getattr` call properties needs to provide a default value, as done in this change.

Closes #12525

### DIFF
--- a/qiskit/transpiler/target.py
+++ b/qiskit/transpiler/target.py
@@ -892,7 +892,7 @@ class Target(Mapping):
             return False
         if qargs not in self._gate_map[operation_name]:
             return False
-        return getattr(self._gate_map[operation_name][qargs], "_calibration") is not None
+        return getattr(self._gate_map[operation_name][qargs], "_calibration", None) is not None
 
     def get_calibration(
         self,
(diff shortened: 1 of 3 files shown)


mtreinish (approved):
LGTm, thanks for catching and fixing this.

PR #12206, opened 2024-04-18
[WIP] adding "final_permutation" attribute to pass manager's property set 

alexanderivrii:
### Summary

This PR implements Jake's proposal in #12127 (see the issue for details). 

### Details and comments

This is still work-in-progress, but creating a draft PR to try to make the new features deadline. I will try to clean this up tomorrow as well as add a better summary for the issue.

Update: this is trickier than I thought, I am changing this to WIP status.

Issue #12127, opened 2024-04-03
Update existing transpiler passes to support `final_permutation`


jakelishman:
Depending on the results of the conversation here: https://github.com/Qiskit/qiskit/pull/9523#discussion_r1546923018, there are a few transpiler passes that would also need updating to support either `final_layout` (the current form) or `final_permutation` (a proposed alternate form).

```[tasklist]
### Needed in all cases
- [ ] `OptimizeSwapBeforeMeasure`
```

```[tasklist]
### Needed if we replace `final_layout`
- [ ] `ApplyLayout`
- [ ] `EnlargeWithAncilla`
- [ ] `SabreLayout`
- [ ] `BasicSwap`
- [ ] `LookaheadSwap`
- [ ] `SabreSwap`
- [ ] `StochasticSwap`
```

I think it's a current bug in O3 that `OptimizeSwapBeforeMeasure` can modify the virtual mapping at the end of the circuit without updating the final layout in a way that the routers will detect.  Using the explicit permutation form as part of the IR (in both virtual and physical domains) as we were talking about in https://github.com/Qiskit/qiskit/pull/9523#discussion_r1546923018 would be a good way of avoiding that happening.

PR #9523, opened 2023-02-02
Add ElidePermutations transpiler pass

mtreinish:
### Summary

This commit adds a new transpiler pass ElidePermutations which is a logical optimization pass designed to run prior to layout or any other physical embedding steps in the transpiler pipeline. It traverses the DAG in topological order and when a swap gate is encountered it removes that gate and instead permutes the logical qubits for any subsequent gates in the DAG. This will eliminate any swaps or permutation gates in a circuit not caused by routing. 

~~Additionally, this pass is added to the preset pass manager for optimization level 3, we can consider adding it to other levels too if we think it makes sense (there is little overhead, and almost 0 if there are no swaps).~~
_edit_: integrating the pas into the preset pass manager turned out a bit tricky and is delegated to a follow-up PR. in addition, when `ElidePermutations` is present, we do not need `RemoveSwapsBeforeMeasure`.

One thing to note is that this pass does not recurse into control flow blocks at all, it treats them as black box operations. So if any control flow blocks contain swap gates those will not be optimized away. This change was made because mapping the permutation outside and inside any control flow block was larger in scope than what the intent for this pass was. That type of work is better suited for the routing passes which already have to reason about this.

[...]

PR #12534, opened 2024-06-09
[WIP] Adding FinalPermutation attribute to DAGCircuit and QuantumCircuit

alexanderivrii:
### Summary

This PR addresses #12127. It supersedes #12206. See also @jakelishman's comments in #9523.

The current way of reasoning about the implicit permutation applied to DAGCircuit's output qubits throughout the transpiler flow is very complicated and non-intuitive, involving reasoning about "layouts", storing both the "final_layout" and the "virtual_permutation_layout" attributes in the property set, and updating the "final_layout" at the end of the transpiler run.

A simpler way is to simply keep track of this implicit permutation throughout the transpiler flow, with every transpiler pass having to update both the dag circuit and the implicit output permutation accordingly. This PR implements this proposal, and in addition keeps the final permutation as an attribute of DAGCircuit and Quantum circuit (and not of the pass manager).

I would love to have feedback on the current implementation, as well as a discussion on how to move forward and hopefully completely replace the "final_layout", etc. attributes by this new functionality.

### DIFF
--- a/qiskit/circuit/final_permutation.py
+++ b/qiskit/circuit/final_permutation.py
@@ -13,9 +13,14 @@
 ## ToDo: reimplement using PermutationGate ?
 
 class FinalPermutation:
-
+    """
+    The notation is the same as for permutation gate, that is
+    [which qubit gets mapped to location 0, which qubit gets mapped to location 1, ...]
+    in this way circuit with implicit final permutation sigma can be replaced
+    by circuit with PermutationGate([sigma])
+    """
     def __init__(self, permutation=None):
-        # for now: original index -> mapped to index
+
         if permutation is None:
             permutation = []
         self.permutation = permutation
@@ -57,6 +62,8 @@ class FinalPermutation:
     def copy(self):
         print(f"COPY")
         return FinalPermutation(self.permutation.copy())
+
+
 
 def _invert_permutation(perm):
     assert isinstance(perm, list)
--- a/qiskit/circuit/final_permutation.py
+++ b/qiskit/circuit/final_permutation.py
@@ -30,3 +30,42 @@ class FinalPermutation:
     def to_identity(self):
         for from_index in range(len(self.permutation)):
             self.permutation[from_index] = from_index
+
+    def compose(self, other, front=True):
+        """front=True: apply self, then other; front=False: apply other, then self."""
+        if front:
+            self.permutation = _compose_permutations(self.permutation, other)
+        else:
+            self.permutation = _compose_permutations(other, self.permutation)
+
+    def __repr__(self):
+        return str(self.permutation)
+
+    def push_forward(self, forward_map):
+        """"
+        RENAME THIS FUNCTION
+        permutation {a: b} replaced by {sigma(a): sigma(b)}
+        for now both have same size, so can compute as: first apply
+        sigma-inverse, then perm, then sigma
+        """
+        assert isinstance(forward_map, list)
+        forward_map_inverse = _invert_permutation(forward_map)
+        self.permutation = _compose_permutations(forward_map_inverse, self.permutation, forward_map)
+
+
+def _invert_permutation(perm):
+    assert isinstance(perm, list)
+
+    """Finds inverse of a permutation."""
+    inverse_map = {inp: out for out, inp in enumerate(perm)}
+    return [inverse_map[inp] for inp in range(len(perm))]
+
+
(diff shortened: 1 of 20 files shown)


qiskit/circuit/final_permutation.py
```diff
@@ -0,0 +1,96 @@
...
+    def num_qubits(self) -> int:
+        """Returns the length of the permutation (i.e. the total number of qubits)."""
+        return len(self.permutation)
+
+    def is_identity(self) -> bool:
+        """Returns whether the permutation is the identity permutation."""
+        return all(from_index == to_index for from_index, to_index in enumerate(self.permutation))
+
+    def compose_with_permutation(self, permutation, front) -> "FinalPermutation":
+        """Composes FinalPermution with a permutation."""
```
alexanderivrii, line 52: I am looking for a better name for this function.

qiskit/circuit/final_permutation.py
```diff
@@ -0,0 +1,96 @@
...
+        return FinalPermutation(composed_permutation)
+
+    def __repr__(self):
+        return str(self.permutation)
+
+    def copy(self) -> "FinalPermutation":
+        """Creates a copy of the FinalPermutation object."""
+        return FinalPermutation(self.permutation.copy())
+
+    def push_using_mapping(self, forward_map, num_target_qubits=None) -> "FinalPermutation":
```
alexanderivrii, line 69: And for a better name for this function.

qiskit/quantum_info/operators/operator.py
```diff
@@ -391,6 +392,8 @@ def from_circuit(
             final_layout (Layout): If specified this kwarg can be used to represent the
                 output permutation caused by swap insertions during the routing stage
                 of the transpiler.
+            original_qubit_indices (dict): The mapping from qubits to positional indices
+                for the ``layout`` argument.
```
alexanderivrii, line 396: This argument allows creating a callback function that checks the equivalence of the current operator and the original operator after each transpiler pass, something along the following:

```
    def callback_func(pass_, dag, time, property_set, count):
        circuit = dag_to_circuit(dag)
        current_op = Operator.from_circuit(
            circuit,
            layout=property_set['layout'],
            final_layout=property_set['final_layout'],
            original_qubit_indices=property_set['original_qubit_indices']
        )

        current_op_new = Operator._from_circuit_new(
            circuit,
            layout=property_set['layout'],
            original_qubit_indices=property_set['original_qubit_indices']
        )

        original_op = op if circuit.num_qubits == original_num_qubits else extended_op

        print(f"After-pass: equiv =  {original_op.equiv(current_op)}")
        print(f"After-pass-new: equiv =  {original_op.equiv(current_op_new)}")
```
I have found this very useful to catch problems, however this is still not fully correct: the `SetLayout` pass sets the pass manager's "layout" attribute, but it's only the later pass `ApplyLayout` that effectively applies it to the circuit, so in-between these two passes the equivalence is not preserved. Though a simple way would be to have `ApplyLayout` set an extra attribute `layout_is_applied` and use `layout` when checking equivalence only when `layout_is_applied` is set.

qiskit/transpiler/passes/routing/stochastic_swap.py
```diff
@@ -385,6 +386,14 @@ def _mapper(self, circuit_graph, coupling_graph, trials=20):
                 self.property_set["final_layout"], circuit_graph.qubits
             )
 
+        layout_permutation = _inverse_pattern(layout.to_permutation(circuit_graph.qubits))
+
```
alexanderivrii: Here and in other similar code things could be done more directly. I.e. instead of computing the final layout and converting it to a permutation, we could directly compute the permutation.

qiskit/synthesis/permutation/permutation_utils.py
```diff
@@ -71,3 +71,13 @@ def _decompose_cycles(cycles):
         for i in range(m // 2):
             swap_list.append((cycle[i - 1], cycle[m - 2 - i]))
     return swap_list
+
+
+def _compose_permutations(*perms):
```
ShellyGarion, line 76: note that this file was moved to rust in #12327

alexanderivrii:
By now, this is completely outdated. The best path forward would be to close this and to open a new PR, if we still think it's a good idea to add the final permutation to the DAGCircuit model (now in Rust).

PR #12535, opened 2024-06-10
Fix annotation

EarlMilktea:
### Summary

Fix the annotated type of `filters` argument.

### Details and comments

This PR is consisting of two parts:

#### Narrow types

As `filters` is used as the first argument of `filter` function against `list[Backends]`, we can narrow the type of `filters`.

#### Use `| None` explicitly

It is explicitly specified in PEP484 (see [here](https://docs.astral.sh/ruff/rules/implicit-optional/) for information/reference).

### DIFF
--- a/qiskit/providers/providerutils.py
+++ b/qiskit/providers/providerutils.py
@@ -21,7 +21,9 @@ from qiskit.providers.backend import Backend
 logger = logging.getLogger(__name__)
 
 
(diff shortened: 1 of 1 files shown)


1ucian0 (approved):
Thanks!

EarlMilktea:
I received a test failure notification. Am I doing something wrong?

1ucian0:
I think it was a network issue in CI. Nothing bad on your side. Let's just try again.

PR #12543, opened 2024-06-11
Oxidize `synth_permutation_acg` 

### LINKED ISSUE #12226: Port `synth_permutation_acg` to Rust

### LINKED ISSUE #12233: Port `synth_clifford_ag` to Rust

Cryoris:
### Summary

Closes #12226. This could in future be further improved by constructing the circuit in Rust directly.

### Details and comments

A small benchmark for 1000 qubit circuit on random circuits with depth 1000 gives
```
(main) Took 46.50923681259155s for 10000 reps on 1000 qubits
(this PR) Took 16.09170889854431s for 10000 reps on 1000 qubits
```
It seems that 99% of this improvement comes from constructing the circuit in Rust, not from porting doing the permutation logic.

I'm happy about any comments or possible improvements, as this is my first Rust PR 🙂


crates/accelerate/src/permutation.rs
```diff
@@ -111,10 +170,30 @@ fn _get_ordered_swap(py: Python, permutation_in: PyArrayLike1<i64>) -> PyResult<
...
+/// Decompose a SWAP pattern into a series of SWAP gate indices to implement them.
+/// For example, let the pattern be [1, 2, 3, 4, 0, 6, 5], which contains the two cycles
+/// [1, 2, 3, 4, 0] and [6, 5]. These can then be implemented with the SWAPs
+/// [(0, 3), (1, 2), (0, 4), (1, 3)], respectively [(6, 5)].
+/// If ``invert`` is True, reverse the indices before computing the SWAPs.
+#[pyfunction]
+#[pyo3(signature = (pattern, invert_order=true))]
+fn _pattern_to_swaps(
+    py: Python,
+    pattern: PyArrayLike1<i64>,
```
ElePT: Naive question, why did you pick i64 instead of i32? I also wonder if there is something like `PyArrayViewLike1` you could use where you directly work with references instead of manually getting the reference in line 186. I am not 100% sure if that is better but it feels like it could save a few steps.

crates/accelerate/src/permutation.rs
```diff
@@ -77,6 +77,65 @@ fn get_ordered_swap(pattern: &ArrayView1<i64>) -> Vec<(i64, i64)> {
...
+
+    // vector to store the cycles
+    let mut cycles: Vec<Vec<usize>> = Vec::new();
+
+    // cast the input pattern in terms of integers to usize, such that it can be used as index
+    // also invert the bit ordering if ``invert_order`` is true
+    let permutation: Array1<usize> = if *invert_order {
+        invert(pattern) // implies cast to usize
+    } else {
+        // TODO change to using Array1 insted of Vec
```
ElePT: Does this TODO refer to the output type? (returning Array1?), because it seems like pattern is already an Array1.

crates/accelerate/src/permutation.rs
```diff
@@ -77,6 +77,65 @@ fn get_ordered_swap(pattern: &ArrayView1<i64>) -> Vec<(i64, i64)> {
...
+        while !explored[i] {
+            cycle.push(i);
+            explored[i] = true;
+            i = permutation[i];
+        }
+        // cycles must have more than 1 element
+        if cycle.len() > 1 {
+            cycles.push(cycle);
+        }
+    }
```
ElePT: maybe this part could be done using iterators, or at least not cloning the permutation?

crates/accelerate/src/permutation.rs
```diff
@@ -77,6 +77,65 @@ fn get_ordered_swap(pattern: &ArrayView1<i64>) -> Vec<(i64, i64)> {
...
+        }
+
+        // no size check needed, cycles always have at least 2 elements
+        swaps.push((cycle[length - 1], cycle[length - 2]));
+        for i in 1..length / 2 {
+            swaps.push((cycle[i - 1], cycle[length - 2 - i]));
+        }
+    }
+
+    swaps
```
ElePT: this also feels like it could be done with iterators, but I'm not sure if would be too convoluted. I can think about it a bit more.

ElePT (commented):
I tried taking a look as an exercise and added a few comments, I hope they can be helpful.

crates/accelerate/src/permutation.rs
Cryoris: Naive answer: because the other functions ported before also used `i64`. This is also the answer to the type choice, but we should check if using `PyArray...` would be better for all these -- let's sync offline 🙂

crates/accelerate/src/permutation.rs
Cryoris: Whoops, forgot to remove the comment after fixing it. Thanks for noticing!

crates/accelerate/src/permutation.rs
Cryoris: I couldn't come up with an iterator solution since the we're changing the index during the iteration (which is why the compiler complains if `permutation` is not cloned), but I'm happy to discuss and see if we find a way 🙂

crates/accelerate/src/permutation.rs
Cryoris: I was discussing this one with @jakelishman, and since the resulting swaps are not iterated over but returned to Python as list, there's likely not much advantage in turning this into an iterator. However, if we can find a way to make `cycles` an iterator, that might be worthwhile!

crates/accelerate/src/permutation.rs
jpacold: Hi @ElePT and @Cryoris -- I chose `i64` for `_get_ordered_swap` because it's the default integer type for Numpy arrays on Linux. There is some cost for converting from other types. On my machine, if I run the Rust version of `_get_ordered_swap` 10k times I get:

| Input type | Time (seconds) |
|---------------|-------------|
| Numpy array (int32) | 0.04141 |
| Numpy array (int64) | 0.00921 |
| Numpy array (uint32)	| 0.04026 |
| Numpy array (uint64)	| 0.04066 |
| Python list | 0.01595 |

I don't know how much this matters, because building the actual circuit will probably always be the bottleneck.

ShellyGarion (commented):
I think that this PR also needs release notes.

Cryoris:
Let's merge #12605 first to have the right directory structure.

crates/accelerate/src/permutation.rs
```diff
@@ -77,6 +82,86 @@ fn get_ordered_swap(pattern: &ArrayView1<i64>) -> Vec<(i64, i64)> {
     swaps
 }
 
+fn pattern_to_cycles(pattern: &ArrayView1<i64>, invert_order: &bool) -> Vec<Vec<usize>> {
```
ShellyGarion: it may be useful to add some docstring to explain what this function is doing

crates/accelerate/src/permutation.rs
```diff
@@ -77,6 +82,86 @@ fn get_ordered_swap(pattern: &ArrayView1<i64>) -> Vec<(i64, i64)> {
...
+        // cycles must have more than 1 element
+        if cycle.len() > 1 {
+            cycles.push(cycle);
+        }
+    }
+
+    cycles
+}
+
+/// Given a disjoint cycle decomposition, decomposes every cycle into a SWAP
```
ShellyGarion: I think this should be better explained, as it should be a list of SWAPs and not a single SWAP

crates/accelerate/src/permutation.rs
```diff
@@ -111,10 +196,31 @@ fn _get_ordered_swap(py: Python, permutation_in: PyArrayLike1<i64>) -> PyResult<
...
 }
 
+/// Decompose a SWAP pattern into a series of SWAP gate indices to implement them.
+/// For example, let the pattern be [1, 2, 3, 4, 0, 6, 5], which contains the two cycles
+/// [1, 2, 3, 4, 0] and [6, 5]. These can then be implemented with the SWAPs
+/// [(0, 3), (1, 2), (0, 4), (1, 3)], respectively [(6, 5)].
+/// If ``invert`` is True, reverse the indices before computing the SWAPs.
+#[pyfunction]
+#[pyo3(signature = (pattern, invert_order=true))]
+fn _pattern_to_swaps(
```
ShellyGarion: why this function is needed in addition to `_synth_permutation_acg` ? where is it being used?

ShellyGarion (changes requested):
Generally, this code looks good. It only needs to be refactored since it conflicts with the main branch, and some further documentation is needed.

crates/accelerate/src/permutation.rs
alexanderivrii: What about the following approach to avoid cloning the permutation:
```
    for pos in &permutation {
        let mut cycle: Vec<usize> = Vec::new();

        // follow the cycle until we reached an entry we saw before
        let mut i = *pos;
        while !explored[i] {
            cycle.push(i);
            explored[i] = true;
            i = permutation[i];
        }
        // cycles must have more than 1 element
        if cycle.len() > 1 {
            cycles.push(cycle);
        }
    }
```

alexanderivrii: Or this (this compiles but I did not run it):
```
fn pattern_to_cycles(pattern: &ArrayView1<i64>, invert_order: &bool) -> Vec<Vec<usize>> {
    // vector keeping track of which elements in the permutation pattern have been visited
    let mut explored: Vec<bool> = vec![false; pattern.len()];

    // cast the input pattern in terms of integers to usize, such that it can be used as index
    // also invert the bit ordering if ``invert_order`` is true
    let permutation: Array1<usize> = if *invert_order {
        invert(pattern) // implies cast to usize
    } else {
        pattern.mapv(|x| x as usize)
    };

    permutation.iter().map(|pos| {
        let mut cycle: Vec<usize> = Vec::new();

        // follow the cycle until we reached an entry we saw before
        let mut i = *pos;
        while !explored[i] {
            cycle.push(i);
            explored[i] = true;
            i = permutation[i];
        }
        cycle
    }).filter(|c| c.len() > 1).collect()
}
```

alexanderivrii: Lol, I don't think this is very efficient, but you can rewrite the inner while-loop as an iterator too, something along the following lines:
```

fn pattern_to_cycles(pattern: &ArrayView1<i64>, invert_order: &bool) -> Vec<Vec<usize>> {
    // vector keeping track of which elements in the permutation pattern have been visited
    let mut explored: Vec<bool> = vec![false; pattern.len()];

    // cast the input pattern in terms of integers to usize, such that it can be used as index
    // also invert the bit ordering if ``invert_order`` is true
    let permutation: Array1<usize> = if *invert_order {
        invert(pattern) // implies cast to usize
    } else {
        pattern.mapv(|x| x as usize)
    };

    permutation
        .iter()
        .map(|pos| {
            // follow the cycle until we reached an entry we saw before
            std::iter::successors(if !explored[*pos] { Some(*pos) } else { None }, |&i| {
                explored[i] = true;
                if !explored[permutation[i]] {
                    Some(permutation[i])
                } else {
                    None
                }
            })
            .collect::<Vec<usize>>()
        })
        .filter(|c| c.len() > 1)
        .collect()
}
```

Cryoris: As discussed offline, iterators turned out not to speed up the code, so for a simpler code we'll stick to the current version. The iterator-only version is available here if anyone is interested: https://github.com/Cryoris/qiskit-terra/blob/92d91440a9c970410ebe337f8cf425bcb065c999/crates/accelerate/src/synthesis/permutation/utils.rs#L101 🙂

crates/accelerate/src/permutation.rs
Cryoris: Removed in 0743d23 🙂

crates/accelerate/src/permutation.rs
Cryoris: More explanations added in 0743d23

crates/accelerate/src/permutation.rs
Cryoris: Added in 0743d23 🙂

ShellyGarion (dismissed):
LGTM, thanks @Cryoris !

ShellyGarion (approved):

PR #12545, opened 2024-06-11
Fix `ParameterVector` methods: `params` and `index`

### LINKED ISSUE #12541: Strange things about `ParameterVector`
### Environment

- **Qiskit version**: main
- **Python version**: 
- **Operating system**: 

### What is happening?

There are some weird things I have noticed while staring at the `ParameterVector` code.  One design choice is that it stores previous elements even if an array has been down-sized:

https://github.com/Qiskit/qiskit/blob/03d107e1f68f2a47ffdcf5599f906120e4efba63/qiskit/circuit/parametervector.py#L99-L104

Some methods, such as `__iter__` and `__str__` and a few others, correctly account for this:

https://github.com/Qiskit/qiskit/blob/03d107e1f68f2a47ffdcf5599f906120e4efba63/qiskit/circuit/parametervector.py#L87-L88
https://github.com/Qiskit/qiskit/blob/03d107e1f68f2a47ffdcf5599f906120e4efba63/qiskit/circuit/parametervector.py#L93-L94

However, other methods do not account for this, such as `index` and the `params` property:

https://github.com/Qiskit/qiskit/blob/03d107e1f68f2a47ffdcf5599f906120e4efba63/qiskit/circuit/parametervector.py#L69-L76

It appears that it may be intentional for `params` to return everything.  If so, this should be documented.  Also, it might be worth changing `index` to return only elements that currently exist.

```python
pv = ParameterVector("theta", 20)
last = pv[-1]
pv.resize(3)
pv.index(last)  # returns 19, which is greater than len(pv) - 1
pv.params  # is a list of 20 elements, which is greater than len(pv)
```

### What should happen?

Up for discussion

### Any suggestions?

_No response_

garrison:
### Summary

This changes the `params` and `index` methods on `ParameterVector` to _not_ return historical parameters, following
https://github.com/Qiskit/qiskit/issues/12541#issuecomment-2160631785.

Fixes #12541.

### Details and comments

I also simplified a bit of the implementation and renamed `_params` to `_all_params` so it is more unlikely someone might misuse it and create a similar problem in the future.  I would even be open to calling this `_all_historical_params` to be even more clear.

### DIFF
--- a/qiskit/circuit/parametervector.py
+++ b/qiskit/circuit/parametervector.py
@@ -50,14 +50,14 @@ class ParameterVectorElement(Parameter):
 class ParameterVector:
     """ParameterVector class to quickly generate lists of parameters."""
 
-    __slots__ = ("_name", "_params", "_size", "_root_uuid")
+    __slots__ = ("_name", "_all_params", "_size", "_root_uuid")
 
     def __init__(self, name, length=0):
         self._name = name
         self._size = length
         self._root_uuid = uuid4()
         root_uuid_int = self._root_uuid.int
-        self._params = [
+        self._all_params = [
             ParameterVectorElement(self, i, UUID(int=root_uuid_int + i)) for i in range(length)
         ]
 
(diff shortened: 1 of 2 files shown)


garrison:
It looks (from the test failures) like there is code outside this file -- in QPY -- that references `_params`.  IMO this is even better reason to call it `_all_historical_params`, so that a user knows exactly what they are getting if they use this private attribute.

garrison:
> It looks (from the test failures) like there is code outside this file -- in QPY -- that references `_params`.

Update: I realized that qpy uses this private attribute because `ParameterVector` does not implement `__setitem__`.  I updated the qpy file with the new name, `_all_params`.

garrison:
I believe the CI failures are unrelated to this PR.

jakelishman (commented):
Julien and I had spoken about this a little offline, and as I understand things, the purpose of keeping the old parameters around is not so that the _instances_ are the same, but so the UUIDs would be.

At the time the comment and class was written, all the parameters had separately drawn entropy, so couldn't be recreated. Since then, we swapped them to derive their UUID deterministically from a parent (for speed), so now they can. I think we can get rid of the "keep old parameters around" logic entirely now and just truncate the list on resize, since regrowing the vector will recreate objects with the same IDs as before.

jakelishman:
(Also yes, the test failure is just an ephemeral issue)

garrison:
Thanks; I opened a new PR: #12561.

PR #12546, opened 2024-06-11
Use relative import for `_accelerate`

garrison:
### Summary

I don't understand the python import mechanism enough to understand _why_, but this PR fixes some problems for me with importing from Rust code when on Fedora or RHEL.

### Details and comments

If one tries to run `tox -epy` from a dependent of Qiskit on Fedora or RHEL, there will be lots of errors about `qiskit._accelerate`.  For the sake of example, we will use the circuit-knitting-toolbox as an example of a Qiskit dependent, but I've also been able to reproduce this with qiskit-experiments.

Here is a Dockerfile that tries to do this using the current latest commit on Qiskit `main`:

```dockerfile
FROM fedora:40
RUN dnf install -y pip git rust cargo
RUN pip install tox
RUN git clone https://github.com/Qiskit-Extensions/circuit-knitting-toolbox.git
WORKDIR /circuit-knitting-toolbox
RUN echo '[tool.hatch.metadata]' >> pyproject.toml && \
    echo 'allow-direct-references = true' >> pyproject.toml && \
    sed -i 's/qiskit>=1.0.0, <2.0/qiskit @ git+https:\/\/github.com\/Qiskit\/qiskit.git@03d107e1f68f2a47ffdcf5599f906120e4efba63/' pyproject.toml && \
    cat pyproject.toml
RUN tox -epy
```

When one attempts to build this image using `docker build -f Dockerfile.qiskit-main .`, the following error is raised repeatedly:

```
    from qiskit.circuit import (
.tox/py/lib64/python3.12/site-packages/qiskit/__init__.py:62: in <module>
    sys.modules["qiskit._accelerate.circuit"] = qiskit._accelerate.circuit
E   AttributeError: partially initialized module 'qiskit' has no attribute '_accelerate' (most likely due to a circular import)
```

If one instead uses a Dockerfile that points to this fork (same as above except one line),

```dockerfile
FROM fedora:40
RUN dnf install -y pip git rust cargo
RUN pip install tox
RUN git clone https://github.com/Qiskit-Extensions/circuit-knitting-toolbox.git
WORKDIR /circuit-knitting-toolbox
RUN echo '[tool.hatch.metadata]' >> pyproject.toml && \
    echo 'allow-direct-references = true' >> pyproject.toml && \
    sed -i 's/qiskit>=1.0.0, <2.0/qiskit @ git+https:\/\/github.com\/garrison\/qiskit.git@relative-toplevel-imports/' pyproject.toml && \
    cat pyproject.toml
RUN tox -epy
```

then the errors are very different:

```
ImportError while importing test module '/circuit-knitting-toolbox/test/utils/test_transforms.py'.
Hint: make sure your test modules/packages have valid Python names.
Traceback:
/usr/lib64/python3.12/importlib/__init__.py:90: in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
test/utils/test_transforms.py:17: in <module>
    from qiskit.circuit import (
.tox/py/lib64/python3.12/site-packages/qiskit/__init__.py:90: in <module>
    from qiskit.circuit import ClassicalRegister
.tox/py/lib64/python3.12/site-packages/qiskit/circuit/__init__.py:1213: in <module>
    from . import _utils
.tox/py/lib64/python3.12/site-packages/qiskit/circuit/_utils.py:22: in <module>
    from .parametervector import ParameterVectorElement
.tox/py/lib64/python3.12/site-packages/qiskit/circuit/parametervector.py:17: in <module>
    from .parameter import Parameter
.tox/py/lib64/python3.12/site-packages/qiskit/circuit/parameter.py:20: in <module>
    import symengine
.tox/py/lib/python3.12/site-packages/symengine/__init__.py:12: in <module>
    import symengine.lib.symengine_wrapper as wrapper
E   ModuleNotFoundError: No module named 'symengine.lib.symengine_wrapper'
```

which is a different, known, issue (but I am having a hard time finding a link at the moment).  EDIT: I believe it may be related to https://github.com/symengine/symengine-wheels/issues/16.

Thus, I propose that the current PR is merged to get Qiskit one step closer to working properly on Fedora and RHEL.

### DIFF
--- a/qiskit/__init__.py
+++ b/qiskit/__init__.py
@@ -60,9 +60,7 @@ import qiskit._numpy_compat
 # We manually define them on import so people can directly import qiskit._accelerate.* submodules
 # and not have to rely on attribute access.  No action needed for top-level extension packages.
 sys.modules["qiskit._accelerate.circuit"] = _accelerate.circuit
-sys.modules["qiskit._accelerate.convert_2q_block_matrix"] = (
-    _accelerate.convert_2q_block_matrix
-)
+sys.modules["qiskit._accelerate.convert_2q_block_matrix"] = _accelerate.convert_2q_block_matrix
 sys.modules["qiskit._accelerate.dense_layout"] = _accelerate.dense_layout
 sys.modules["qiskit._accelerate.error_map"] = _accelerate.error_map
 sys.modules["qiskit._accelerate.isometry"] = _accelerate.isometry
--- a/qiskit/__init__.py
+++ b/qiskit/__init__.py
@@ -52,37 +52,37 @@ if sys.version_info < (3, 9):
     )
 
 
-import qiskit._accelerate
+from . import _accelerate
 import qiskit._numpy_compat
 
 # Globally define compiled submodules. The normal import mechanism will not find compiled submodules
 # in _accelerate because it relies on file paths, but PyO3 generates only one shared library file.
 # We manually define them on import so people can directly import qiskit._accelerate.* submodules
 # and not have to rely on attribute access.  No action needed for top-level extension packages.
-sys.modules["qiskit._accelerate.circuit"] = qiskit._accelerate.circuit
+sys.modules["qiskit._accelerate.circuit"] = _accelerate.circuit
 sys.modules["qiskit._accelerate.convert_2q_block_matrix"] = (
-    qiskit._accelerate.convert_2q_block_matrix
+    _accelerate.convert_2q_block_matrix
 )
-sys.modules["qiskit._accelerate.dense_layout"] = qiskit._accelerate.dense_layout
-sys.modules["qiskit._accelerate.error_map"] = qiskit._accelerate.error_map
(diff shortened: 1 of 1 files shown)


qiskit/__init__.py
```diff
@@ -52,37 +52,37 @@
     )
 
 
-import qiskit._accelerate
+from . import _accelerate
```
jakelishman, line 55: After this line, we probably don't actually need to change `qiskit._accelerate` into `_accelerate`, because this line still causes `_accelerate` to be a name that's defined on the `qiskit` object.

But also, I don't mind at all.

jakelishman (commented):
I'm fine with this, especially if it's helping you.  Black can probably join together some of the lines now, since they're shorter.

garrison:
> Black can probably join together some of the lines now, since they're shorter.

Sorry, I will fix this.  I fixed it once, but I messed it up again when I when to rebase to the latest `main` and had to resolve a conflict.

jakelishman (approved):

PR #12551, opened 2024-06-11
Run test suite (stestr) during release builds instead of stochastic_swap.py

### LINKED ISSUE #12202: Run test suite during release builds
### What should we add?

Currently our release builds are configured to run this small test script: https://github.com/Qiskit/qiskit/blob/main/examples/python/stochastic_swap.py which executes the `StochasticSwap` transpiler pass. This is an historical artifact from before the use of Rust by qiskit, the stochastic swap pass was our only compiled extension (written in Cython) and the script was exercising solely that to speed up the validation phase of release builds. However in the intervening years more and more of Qiskit is being written in rust (with the 1.1.0 release poised to be >10% written in Rust by lines of code). We should update our wheel build configuration to instead run the unittests to validate that the binaries we publish work on all our supported python versions more broadly than just a single transpiler pass.

This primarily involves updating the `cibuildwheel` configuration: https://github.com/Qiskit/qiskit/blob/9ede1b6ace329b399fbf226b839df8190f1b4115/pyproject.toml#L136-L163 to run our unittests instead of the standalone script. We already run the tests once for PGO, but it would be good to validate that the wheels we publish pass the tests as part of the `cibuildwheel` test phase.
 
There is an open question, whether we should execute the full test suite or just a subset of the tests. I think for our tier 1 platforms (https://docs.quantum.ibm.com/start/install#operating-system-support) we probably should have enough CI time to run the full test suite. But for our tier 2 platforms this might be tricky because we run those builds under QEMU emulation. So for tier 2 platforms we might want a "smoke test" subset of the test suite that executes quicker because everything is very slow under emulation.

ljcamargo:
### Summary

I changed the test-command in [tool.cibuildwheel] to run the test suite stestr instead of running the old stochastic_swap.py script. I should work linux, macosx, and windows because the test command is not overriden

### Details and comments

As it is a safecheck only I added the --abbreviate command to produce a shorter output and didn't activate the QISKIT_TEST_CAPTURE_STREAMS=1

Closes: https://github.com/Qiskit/qiskit/issues/12202 
[Unitary Hack 2024]


pyproject.toml
```diff
@@ -138,12 +138,12 @@ manylinux-x86_64-image = "manylinux2014"
 manylinux-i686-image = "manylinux2014"
 skip = "pp* cp36-* cp37-* *musllinux* *win32 *i686 cp38-macosx_arm64"
 test-skip = "*win32 *linux_i686"
-test-command = "python {project}/examples/python/stochastic_swap.py"
+test-command = "stestr run --abbreviate"
```
mtreinish, line 141: Does the path for this work correctly? `stestr` only works if it's in the repository root with it's config file.

pyproject.toml
```diff
@@ -138,12 +138,12 @@ manylinux-x86_64-image = "manylinux2014"
...
 # We need to use pre-built versions of Numpy and Scipy in the tests; they have a
 # tendency to crash if they're installed from source by `pip install`, and since
 # Numpy 1.22 there are no i686 wheels, so we force pip to use older ones without
 # restricting any dependencies that Numpy and Scipy might have.
-before-test = "pip install --only-binary=numpy,scipy numpy scipy"
+before-test = "pip install --only-binary=numpy,scipy numpy scipy | pip install -r requirements-dev.txt"
```
mtreinish: Is this syntax correct? I would expect this to pipe stdout from the first pip command into stdin of the second. I think you can just do:

```suggestion
before-test = "pip install --only-binary=numpy,scipy numpy scipy -r {project}/requirements-dev.txt"
```

mtreinish (commented):
Thanks for getting this started. I left a couple comments inline, the other thing is as discussed in the original issue we probably will need to run a subset of tests for the tier 2 supported platforms (currently only aarch64 linux) as the CI environment is much slower.

pyproject.toml
ljcamargo: yes @mtreinish totally right, I mistyped | instead of && for running this command sequentially. I was also wondering whether should we install everything on requirements-dev.txt or just stestr in aims of efficiency, but didn't have the time to check if the tests used all of these dev-dependencies.

1ucian0:
Hi @ljcamargo,
While this effort looks like starting in the right direction, it seems that it wont make it [before tomorrow's deadline](https://unitaryhack.dev/project-guide/#%F0%9F%97%93%EF%B8%8F-maintainer-timeline-%F0%9F%97%93%EF%B8%8F). Of course, you are more than welcome to continue working on it and @mtreinish will continue reviewing it. However, it wont be eligible for the unitaryHack bounty.

PR #12561, opened 2024-06-13
Do not retain and expose old elements of `ParameterVector`

### LINKED ISSUE #12541: Strange things about `ParameterVector`

garrison:
### Summary

This fixes #12541 according to https://github.com/Qiskit/qiskit/pull/12545#pullrequestreview-2114202382.

### Details and comments

Note that qpy does _not_ appear to store the `_root_uuid`, so if one deserializes a `ParameterVector` and then removes elements and later adds new ones, the uuids of the regenerated elements will change.

### DIFF
--- a/qiskit/circuit/parametervector.py
+++ b/qiskit/circuit/parametervector.py
@@ -50,11 +50,10 @@ class ParameterVectorElement(Parameter):
 class ParameterVector:
     """ParameterVector class to quickly generate lists of parameters."""
 
-    __slots__ = ("_name", "_params", "_size", "_root_uuid")
+    __slots__ = ("_name", "_params", "_root_uuid")
 
     def __init__(self, name, length=0):
         self._name = name
-        self._size = length
         self._root_uuid = uuid4()
         root_uuid_int = self._root_uuid.int
         self._params = [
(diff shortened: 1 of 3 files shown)


jakelishman (approved):
Nice, thanks!

PR #12563, opened 2024-06-13
Oxidize TwoQubitDecomposeUpToDiagonal

mtreinish:
### Summary

This commit ports the TwoQubitDecomposeUpToDiagonal class from Python to rust. This internal private class is used internally by the quantum shannon decomposition code, and while not performance critical was simple to port. One difference is while the original Python implementation was a class, it acted more like a function in practice. So the new rust version is exposed as a function.

### Details and comments

TODO:

- [x] Generate circuit rust side after #12459 merges


crates/accelerate/src/two_qubit_decompose.rs
```diff
@@ -1015,6 +961,71 @@ impl TwoQubitWeylDecomposition {
...
+                self.calculated_fidelity,
+                self.requested_fidelity,
+            ),
+        )
+            .into_py(py))
+    }
+
+    #[new]
+    #[pyo3(signature=(unitary_matrix, fidelity=DEFAULT_FIDELITY, _specialization=None))]
+    fn new(
```
kevinhartman, line 1022: Is there value in following the pattern of naming this `py_new` and naming the Rust-side constructor `new`?

crates/accelerate/src/two_qubit_decompose.rs
```diff
@@ -1739,6 +1718,152 @@ impl TwoQubitBasisDecomposer {
...
+            TwoQubitWeylDecomposition::new_inner(unitary, Some(DEFAULT_FIDELITY), None)?;
+        let traces = self.traces(&target_decomposed);
+        let best_nbasis = traces
+            .into_iter()
+            .enumerate()
+            .map(|(idx, trace)| (idx, trace.trace_to_fid() * basis_fidelity.powi(idx as i32)))
+            .min_by(|(_idx1, fid1), (_idx2, fid2)| fid2.partial_cmp(fid1).unwrap())
+            .unwrap()
+            .0;
+        let best_nbasis = _num_basis_uses.unwrap_or(best_nbasis as u8);
```
kevinhartman: Can you calculate `best_nbasis` here in the block of an `unwrap_or_else` instead of using `unwrap_or`?

kevinhartman (commented):
Looks good, just a few super minor comments.

crates/accelerate/src/two_qubit_decompose.rs
mtreinish: The pattern that's used everywhere in accelerate for this was rust is `foo_inner` and python is `foo`. I think we should normalize on a single pattern across all the crates, but maybe do that in a separate standalone PR?

crates/accelerate/src/two_qubit_decompose.rs
mtreinish: Yeah I can do that, it is a potential optimization. Although realistically the hot path here is `best_nbasis` that private flag `_num_basis_uses`  I think is only used in tests.

EDIT: Although this code isn't actually from this PR, it's just an artifact of reorganizing the constructors. This was just moved from `new()`

mtreinish: Done in: https://github.com/Qiskit/qiskit/pull/12563/commits/655ce1d677d0f66265b8dd9a1e0631c31f2c26f9

kevinhartman (approved):

Issue #12566, opened 2024-06-13
Add remaining standard gates to rust representation


mtreinish:
### What should we add?

After #12459 has merged there is now a native rust representation of the standard gate library. However, due to the size of #12459 not all gates were added to this rust representation. This issue tracks the migration of the remaining gates so that we have the complete standard gate library in rust. The remaining gates are:

```[tasklist]
### Tasks
- [x] C3XGate https://github.com/Qiskit/qiskit/pull/12659
- [x] C3SXGate https://github.com/Qiskit/qiskit/pull/12659
- [ ] C4XGate https://github.com/Qiskit/qiskit/pull/12659
- [x] DCXGate https://github.com/Qiskit/qiskit/pull/12644
- [x] CHGate https://github.com/Qiskit/qiskit/pull/12639
- [x] CPhaseGate https://github.com/Qiskit/qiskit/pull/12639
- [x] CRXGate https://github.com/Qiskit/qiskit/pull/12648
- [x] CRYGate https://github.com/Qiskit/qiskit/pull/12648
- [x] CRZGate https://github.com/Qiskit/qiskit/pull/12648
- [x] CSGate https://github.com/Qiskit/qiskit/pull/12639
- [x] CSdgGate https://github.com/Qiskit/qiskit/pull/12639
- [x] CSwapGate https://github.com/Qiskit/qiskit/pull/12639
- [x] CSXGate https://github.com/Qiskit/qiskit/pull/12639
- [x] CUGate https://github.com/Qiskit/qiskit/pull/12659
- [x] CU1Gate https://github.com/Qiskit/qiskit/pull/12659
- [x] CU3Gate https://github.com/Qiskit/qiskit/pull/12659
- [x] CCZGate https://github.com/Qiskit/qiskit/pull/12659
- [x] MSGate
- [x] PhaseGate
- [ ] RCCXGate https://github.com/Qiskit/qiskit/pull/12659
- [ ] RC3XGate https://github.com/Qiskit/qiskit/pull/12659
- [x] RGate https://github.com/Qiskit/qiskit/pull/12662 ~https://github.com/Qiskit/qiskit/pull/12507~
- [x] RXXGate https://github.com/Qiskit/qiskit/pull/12672
- [x] RYYGate https://github.com/Qiskit/qiskit/pull/12672
- [x] RZZGate https://github.com/Qiskit/qiskit/pull/12672
- [x] RZXGate https://github.com/Qiskit/qiskit/pull/12672
[...]

PR #12572, opened 2024-06-13
Add rust representation for the u1, u2, and u3 gates

mtreinish:
### Summary

This commit adds the rust representation of the U1Gate, U2Gate, and U3Gate to the `StandardGates` enum in rust.

### Details and comments

Part of #12566

### DIFF
--- a/crates/circuit/src/imports.rs
+++ b/crates/circuit/src/imports.rs
@@ -143,9 +143,9 @@ static STDGATE_IMPORT_PATHS: [[&str; 2]; STANDARD_GATE_SIZE] = [
     ],
     // U1Gate = 26
     ["qiskit.circuit.library.standard_gates.u1", "U1Gate"],
-    // U1Gate = 27
+    // U2Gate = 27
     ["qiskit.circuit.library.standard_gates.u2", "U2Gate"],
-    // U1Gate = 28
+    // U3Gate = 28
     ["qiskit.circuit.library.standard_gates.u3", "U3Gate"],
 ];
 
--- a/crates/circuit/src/imports.rs
+++ b/crates/circuit/src/imports.rs
@@ -141,6 +141,12 @@ static STDGATE_IMPORT_PATHS: [[&str; 2]; STANDARD_GATE_SIZE] = [
         "qiskit.circuit.library.standard_gates.xx_plus_yy",
         "XXPlusYYGate",
     ],
+    // U1Gate = 26
+    ["qiskit.circuit.library.standard_gates.u1", "U1Gate"],
+    // U1Gate = 27
+    ["qiskit.circuit.library.standard_gates.u2", "U2Gate"],
+    // U1Gate = 28
+    ["qiskit.circuit.library.standard_gates.u3", "U3Gate"],
 ];
 
 /// A mapping from the enum variant in crate::operations::StandardGate to the python object for the
--- a/crates/circuit/src/operations.rs
+++ b/crates/circuit/src/operations.rs
@@ -205,14 +205,17 @@ pub enum StandardGate {
     ISwapGate = 23,
     XXMinusYYGate = 24,
     XXPlusYYGate = 25,
+    U1Gate = 26,
+    U2Gate = 27,
+    U3Gate = 28,
 }
 
 static STANDARD_GATE_NUM_QUBITS: [u32; STANDARD_GATE_SIZE] = [
-    1, 1, 1, 2, 2, 2, 3, 1, 1, 1, 2, 2, 1, 0, 1, 1, 1, 1, 1, 1, 1, 1, 1, 2, 2, 2,
+    1, 1, 1, 2, 2, 2, 3, 1, 1, 1, 2, 2, 1, 0, 1, 1, 1, 1, 1, 1, 1, 1, 1, 2, 2, 2, 1, 1, 1,
 ];
 
 static STANDARD_GATE_NUM_PARAMS: [u32; STANDARD_GATE_SIZE] = [
-    0, 0, 0, 0, 0, 0, 0, 1, 1, 1, 0, 0, 0, 1, 0, 0, 1, 3, 0, 0, 0, 0, 0, 0, 2, 2,
+    0, 0, 0, 0, 0, 0, 0, 1, 1, 1, 0, 0, 0, 1, 0, 0, 1, 3, 0, 0, 0, 0, 0, 0, 2, 2, 1, 2, 3,
 ];
 
 static STANDARD_GATE_NAME: [&str; STANDARD_GATE_SIZE] = [
@@ -242,6 +245,9 @@ static STANDARD_GATE_NAME: [&str; STANDARD_GATE_SIZE] = [
     "iswap",
     "xx_minus_yy",
     "xx_plus_yy",
+    "u1",
+    "u2",
+    "u3",
 ];
 
 #[pymethods]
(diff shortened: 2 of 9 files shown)


crates/circuit/src/operations.rs
```diff
@@ -374,6 +382,23 @@ impl Operation for StandardGate {
                 }
                 _ => None,
             },
+            Self::U1Gate => match params[0] {
+                Param::Float(val) => Some(aview2(&gate_matrix::u1_gate(val)).to_owned()),
+                Param::ParameterExpression(_) => None,
```
Cryoris: This looks different than other 1-parameter gates above, which have
```
gate => params  {
  [Param::Float(...)] => matrix,
  _ => None,
}
```
why is the additional match with `ParameterExpression` not supported here, but is with the others?

test/python/circuit/test_rust_equivalence.py
```diff
@@ -79,12 +79,23 @@ def test_definitions(self):
                             )
                         # Rust uses P but python still uses u1
                         elif rs_inst.operation.name == "p":
-                            self.assertEqual(py_inst.operation.name, "u1")
-                            self.assertEqual(rs_inst.operation.params, py_inst.operation.params)
-                            self.assertEqual(
-                                [py_def.find_bit(x).index for x in py_inst.qubits],
-                                [rs_def.find_bit(x).index for x in rs_inst.qubits],
-                            )
+                            if py_inst.operation.name == "u1":
```
Cryoris, line 82: I'm not sure I understand this new addition: there is a gate which in the Python decomposition uses u1 (or u3) whereas the Rust decomposition uses the phase gate? If yes, which gates are these?

Cryoris (commented):
Overall LGTM, just two questions from my side 🙂

crates/circuit/src/operations.rs
mtreinish: The `_ =>` syntax is basically a catch-all match instead of that here I'm just explicitly matching the other two choices. I'll update this to use the catch-all match syntax though, I didn't actually do this with any intention, this PR came from a test branch and I updated it from when #12459 only had 2 variants for the `Param` enum to add the 3rd condition.

test/python/circuit/test_rust_equivalence.py
mtreinish: This is either z or rz (maybe both). This condition existed before this PR, but I updated the logic here because the u1 definition is now also in terms of p gate in rust, but is in terms of u3 in Python.

Cryoris: Ah I see! So 
```
Python: {Z, RZ} -> U1, U1 -> U3
Rust: {Z, RZ, U1} -> Phase
```
-- thanks!

crates/circuit/src/operations.rs
```diff
@@ -374,6 +382,23 @@ impl Operation for StandardGate {
                 }
                 _ => None,
             },
+            Self::U1Gate => match params[0] {
+                Param::Float(val) => Some(aview2(&gate_matrix::u1_gate(val)).to_owned()),
```
ElePT, line 441: I've noticed that `&gate_matrix::u1_gate(val)` is not dereferencing `val`, while the other gate calls do: `&gate_matrix::u2_gate(*phi, *lam)`, `u3_gate(*theta, *phi, *lam)`.  Was this intentional? (aha, it's because it's just a float, never mind).

mtreinish: The difference here is that the match is on `params[0]` instead of `params`. The `val` is already an `f64` and you can't dereference that, while for the other gates the angles have `&f64`. 

I haven't dug into the test failures yet, but I suspect it's more likely I screwed up something in the translation. The qasm3 failure at least is because there are no longer fixed object ids for the gates and I need to update the test to account for that (i.e. use a regex).

mtreinish: If this is a blocker for #12598 feel free to just rebase it on main and move ahead without this. I can reindex the gates in this PR without too much trouble.

crates/circuit/src/imports.rs
```diff
@@ -119,6 +119,12 @@ static STDGATE_IMPORT_PATHS: [[&str; 2]; STANDARD_GATE_SIZE] = [
     ["qiskit.circuit.library.standard_gates.p", "PhaseGate"],
     // UGate = 17
     ["qiskit.circuit.library.standard_gates.u", "UGate"],
+    // U1Gate = 18
+    ["qiskit.circuit.library.standard_gates.u1", "U1Gate"],
+    // U1Gate = 19
+    ["qiskit.circuit.library.standard_gates.u2", "U2Gate"],
+    // U1Gate = 20
```
ElePT:
```suggestion
    // U2Gate = 19
    ["qiskit.circuit.library.standard_gates.u2", "U2Gate"],
    // U3Gate = 20
```

crates/circuit/src/operations.rs
mtreinish: I fixed the test failures in: https://github.com/Qiskit/qiskit/pull/12572/commits/65ff00991a3acd6d3619686a9ced942ebced0662 the optimize 1q gates decomposition failures were because the tests were doing:

`gate.definition.global_phase = val`

and that doesn't do anything to the gate object for a standard gate in rust. TBH, I have no idea what this test is trying to do but to fix it I just used a custom gate class instead of `U1Gate`.

crates/circuit/src/operations.rs
mtreinish: Fixed in: https://github.com/Qiskit/qiskit/pull/12572/commits/65ff00991a3acd6d3619686a9ced942ebced0662

test/python/circuit/test_rust_equivalence.py
```diff
@@ -102,7 +113,7 @@ def test_matrix(self):
                 continue
 
             with self.subTest(name=name):
-                params = [pi] * standard_gate._num_params()
+                params = [0.1] * standard_gate._num_params()
```
ElePT, line 116: Was this change because `math.pi` was triggering some of the test failures? I feel like it's commonly used for assigning parameter values.

crates/circuit/src/operations.rs
ElePT: So `gate.definition.global_phase = val` is another of the "hidden forbidden moves", like `dag_node.op.params[0]`. Do you think we should document it in the reno too? At a first glance people would probably not expect it to fail.

ElePT (approved):
The implementation LGTM, I just had a couple of questions about the cases that triggered test failures.

test/python/circuit/test_rust_equivalence.py
mtreinish: Oh, I thought I changed it back. While debugging I did this because one of the test failures was using 0.1 as the angle value and I wanted to rule that out as the cause of the failure (checking the matrix was also correct with 0.1). I can change it back to pi, it realistically doesn't matter too much though.

crates/circuit/src/operations.rs
mtreinish: I'm not sure it needs one, the definition was never really supposed to be mutable like this as it's an inherent attribute of a gate, changing it doesn't make much sense (for most things it doesn't get used either). This already wouldn't work with singletons as changing the global phase would affect every instance of the gate. I viewed this as test bug, because I have no idea what that test was trying to accomplish. But I can add a release note out of an abundance of caution if you think it's warranted here.

test/python/circuit/test_rust_equivalence.py
Cryoris: I'd even expect 0.1 to be a better value  for testing since for pi the matrices take on some special forms and 0.1 is more "irregular" in that sense 🤔

ElePT: Sounds good, I just wanted to make sure we were not ignoring pi.

crates/circuit/src/operations.rs
ElePT: Hmmm I think it would be nice to have it documented somewhere, but I am ok just keeping an eye on it, I don't think people refer to the release notes anywhere and I see how documenting every edge case can be a problem.

PR #12585, opened 2024-06-14
Add Rust representation of `EquivalenceLibrary`

### LINKED ISSUE #12278: Port `EquivalenceLibrary` to Rust
### What should we add?

As part of our effort to migrate more of Qiskit's code into rust for better runtime performance and efficiency we need to look at migrating more functionality to Rust. As part of this we're looking at moving the implementation of more transpiler passes into Rust in #12208. The `BasisTranslator` is one of the passes included in that epic (individually being tracked in #12246), and to enable the maximum efficiencies in that pass we will need to port the `EquivalenceLibrary` to rust. Right now it's built using a rustworkx graph, and we'll want to move the internals one level down to a petgraph graph and leverage rustworkx-core.

In practice this is probably blocked on #12205, but an initial implementation can leverage python space for dealing with gates and circuits and this can be refactored after #12205 is implemented to use the rust objects directly.

The one thing to keep an eye out for here is that after #12205 we'll still need a mechanism for Python space users to add onto the equivalence library with custom gates. It likely will just need to reuse whatever mechanism we use for `QuantumCircuit` but it's worth highlighting here too.

raynelfss:
Fixes #12278

### Summary
As we continue our efforts to make more of the transpiler and circuit's functionality into rust. After #12459 added an infrastructure for `Gates` and `Operations` in Rust, it will be important for us to think of how to keep track of equivalent implementations of said gates that might not work in different Backends, which is why the `EquivalenceLibrary` exists.

The following commits add a rust representation of the `EquivalenceLibrary` structure.

#### Implementation:
These modifications consist of two parts:
- A rust `PyO3` class named: `EquivalenceLibrary` that implements most of the functionality of the python counterpart while keeping equivalent functionality in rust.
- A python extension of the `PyO3` class which is only used for the `draw()` method.

#### Modifications:
- A couple of modifications were made to some of the defined structures under the `qiskit._accelerate.circuit` crate.

#### Possible issues:
- There is currently no way of creating a `PyDiGraph` natively from rust. However, `rustworkx-core` allows the creation of a `DiGraph` that has the same functionality. If an instance of `PyDiGraph` is needed from `Python`, an alternative method called `graph` will assemble and store a `PyDiGraph` instance equivalent to the one present in the `EquivalenceLibrary` instance. This being only a representation should not modify the internal data of the graph stored in rust and will be discarded every time an equivalency is added or modified.
- This implementation uses `PyO3`'s `subclass` property to allow the rust `PyClass` to be extended from Python. However, this might lead to unexpected behavior if one were to extend the python class, as these need to be extended using the `__new__` method instead of `__init__`. It is highly unlikely that this class would be extended further, but it is important to note this change in behavior.

#### Blocks
- [x] #12913
- [x] #12794 
- [x] #12730
- [x] #12766
- [x] #12697
- [x] #12698

#### Other comments:
- Feel free to leave any reviews or comments.


Cryoris:
Nice! Did you already get to benchmark the basis translator with the oxidized equivalence library? 🙂

raynelfss:
> Nice! Did you already get to benchmark the basis translator with the oxidized equivalence library? 🙂

Not yet @Cryoris, I am having some issues with the BasisTranslator, once I look at those will get to benchmarking it.

crates/circuit/src/equivalence.rs
```diff
@@ -0,0 +1,844 @@
...
+        (self.name.to_string(), self.num_qubits).hash(&mut hasher);
+        hasher.finish()
+    }
+
+    fn __repr__(slf: PyRef<'_, Self>) -> String {
+        slf.to_string()
+    }
+
+    fn __getstate__(slf: PyRef<Self>) -> (String, u32) {
+        (slf.name.to_owned(), slf.num_qubits)
```
jlapeyre: I might do this
```suggestion
        (slf.name.clone(), slf.num_qubits)
```
because, unlike with `to_owned`, I know the output type immediately.
On my machine `clippy` does not care which I use. If we are lucky enough to have a consensus among devs on guidelines for choosing between the two, then that should determine which we use here.

crates/circuit/src/equivalence.rs
```diff
@@ -0,0 +1,844 @@
...
+        self.eq(&other)
+    }
+
+    fn __hash__(&self) -> u64 {
+        let mut hasher = DefaultHasher::new();
+        (self.name.to_string(), self.num_qubits).hash(&mut hasher);
+        hasher.finish()
+    }
+
+    fn __repr__(slf: PyRef<'_, Self>) -> String {
```
jlapeyre: In this block (and maybe elsewhere), I find that I can choose to include or elide the argument `'_` in `PyRef` and `PyRefMut`, and neither rustc nor clippy complain. We should find some reason to pick one and use it uniformly (where the "some reason" applies).
The descriptions of lifetime elision rules that I find in rust documentation are slightly ambiguous and incomplete. It seems clear that if there are no lifetimes (elided or explicit) in the output types, then all input lifetimes may be elided. And in that case, we should probably elide them in the input. Someone might argue that you can choose what you like. I'd say within one code base you should  make one choice. Some might say that's too restrictive. I'd say that in one file or module you should make one choice. If that's too restrictive, I'd still assert that within one `impl` block you should make one choice.

raynelfss:
### About the failing test:
#### ``test.python.transpiler.test_basis_translator.TestBasisExamples.test_cx_bell_to_ecr``
I'm considering changing that one failing test to use the equivalencies selected using the current methods since they seem to produce similarly equivalent results with the only difference being which circuit is selected for it, since it seems to be the only thing holding this redefinition back.

This is what the test does:
```py
from qiskit.circuit import QuantumCircuit, QuantumRegister
from qiskit.converters import dag_to_circuit, circuit_to_dag
from qiskit.transpiler.passes.basis import BasisTranslator
from qiskit.circuit.equivalence_library import (StandardEquivalenceLibrary as std_eqlib)

pi: float = 3.14

"""Verify we can translate a CX bell to ECR,U."""
bell = QuantumCircuit(2)
bell.h(0)
bell.cx(0, 1)

in_dag = circuit_to_dag(bell)
out_dag = BasisTranslator(std_eqlib, ["ecr", "u"]).run(in_dag)

qr = QuantumRegister(2, "q")
expected = QuantumCircuit(2)
expected.u(pi / 2, 0, pi, qr[0])
expected.u(0, 0, -pi / 2, qr[0])
expected.u(pi, 0, 0, qr[0])
expected.u(pi / 2, -pi / 2, pi / 2, qr[1])
expected.ecr(0, 1)
expected_dag = circuit_to_dag(expected)

# Use assert_eq or just print the result
out_dag == expected_dag
Operator.from_circuit(bell) == Operator.from_circuit(dag_to_circuit(out_dag)
```

The problem happens when the `BasisTranslator` picks the equivalence for a CX gate. Where it would usually pick the following equivalence:
```
global phase: 7π/4
     ┌──────────┐┌───────┐┌──────┐
q_0: ┤ Rz(-π/2) ├┤ Ry(π) ├┤0     ├
     ├─────────┬┘└───────┘│  Ecr │
q_1: ┤ Rx(π/2) ├──────────┤1     ├
     └─────────┘          └──────┘
```
It ends up picking the following:
```
global phase: π/4
     ┌─────┐ ┌──────┐┌───┐
q_0: ┤ Sdg ├─┤0     ├┤ X ├
     ├─────┴┐│  Ecr │└───┘
q_1: ┤ √Xdg ├┤1     ├─────
     └──────┘└──────┘    
```

And so the produced graph is different:

Produced:
```
        ┌────────────┐   ┌─────────────┐┌──────┐┌──────────┐
q_0: ───┤ U(π/2,0,π) ├───┤ U(0,0,-π/2) ├┤0     ├┤ U(π,0,π) ├
     ┌──┴────────────┴──┐└─────────────┘│  Ecr │└──────────┘
q_1: ┤ U(-π/2,-π/2,π/2) ├───────────────┤1     ├────────────
```

Expected:
```
       ┌────────────────┐  ┌──────────────┐┌─────────────┐┌──────┐
q_0: ──┤ U(1.57,0,3.14) ├──┤ U(0,0,-1.57) ├┤ U(3.14,0,0) ├┤0     ├
     ┌─┴────────────────┴─┐└──────────────┘└─────────────┘│  Ecr │
q_1: ┤ U(1.57,-1.57,1.57) ├───────────────────────────────┤1     ├
```
I'm curious as to what the @Qiskit/terra-core team thinks about this.

qiskit/transpiler/passes/basis/basis_translator.py
```diff
@@ -577,7 +577,12 @@ def _basis_search(equiv_lib, source_basis, target_basis):
 
     # we add a dummy node and connect it with gates in the target basis.
     # we'll start the search from this dummy node.
-    dummy = graph.add_node(NodeData(key="key", equivs=[("dummy starting node", 0)]))
+    dummy = graph.add_node(
+        NodeData(
+            key=Key("key", 0),
```
jlapeyre: Would this potentially cause a problem if a user uses a gate named "key" on zero qubits? I see that the name and number of qubits are used to hash the key.

crates/accelerate/src/equivalence.rs
```diff
@@ -0,0 +1,786 @@
...
+
+// Custom Structs
+
+#[pyclass(frozen, sequence, module = "qiskit._accelerate.equivalence")]
+#[derive(Debug, Clone, PartialEq, Eq, PartialOrd, Ord, Hash)]
+pub struct Key {
+    #[pyo3(get)]
+    pub name: String,
+    #[pyo3(get)]
+    pub num_qubits: u32,
```
jlapeyre, line 59:
```suggestion
pub(crate) struct Key {
    #[pyo3(get)]
    pub(crate) name: String,
    #[pyo3(get)]
    pub(crate) num_qubits: u32,
```
I think visibility should be low by default.

crates/accelerate/src/equivalence.rs
```diff
@@ -0,0 +1,786 @@
...
+
+    fn __hash__(&self) -> u64 {
+        let mut hasher = DefaultHasher::new();
+        (self.name.to_string(), self.num_qubits).hash(&mut hasher);
+        hasher.finish()
+    }
+
+    fn __repr__(slf: PyRef<Self>) -> String {
+        slf.to_string()
+    }
```
jlapeyre, line 78:
```suggestion
    fn __repr__(&self) -> String {
        self.to_string()
    }
```
I think that making a habit of minimizing entanglement with Python will make separating the Python layer easier in the future. It's true that this is a dunder and would have to be modified.

crates/accelerate/src/equivalence.rs
```diff
@@ -0,0 +1,786 @@
...
+    ///
+    /// Check if a library contains any decompositions for gate.
+    ///
+    /// # Arguments:
+    /// * `operation` OperationType: A Gate instance.
+    ///
+    /// # Returns:
+    /// `bool`: `true` if gate has a known decomposition in the library.
+    ///         `false` otherwise.
+    pub fn has_entry(&self, operation: &PackedOperation) -> bool {
```
jlapeyre, line 655: I think this does not need to be `pub`.  Same thing occurs in other places as well.

jlapeyre (commented):
I have not yet finished. But I have made a few comments.
A couple of things I have looked for
* Reducing visibility. Things that are `pub` can be made `pub(crate)` or even private.
* Reducing "gratuitous" dependence on Python. I noted one or two cases below.

qiskit/circuit/equivalence.py
```diff
@@ -12,190 +12,24 @@
...
-Key = namedtuple("Key", ["name", "num_qubits"])
-Equivalence = namedtuple("Equivalence", ["params", "circuit"])  # Ordered to match Gate.params
-NodeData = namedtuple("NodeData", ["key", "equivs"])
-EdgeData = namedtuple("EdgeData", ["index", "num_gates", "rule", "source"])
+from qiskit.exceptions import InvalidFileError
+from qiskit._accelerate.equivalence import (  # pylint: disable=unused-import
+    BaseEquivalenceLibrary,
+    Key,
+    Equivalence,
+    NodeData,
```
jlapeyre, line 24: Some things, eg. `Key` and `NodeData`, are not used apparently, and the lint for this is suppressed. Why is this?

raynelfss: This is to allow the user to still be able to import these things directly from `qiskit.circuit.equivalence` as they were able to before. If this is not needed anymore I can easily remove.

crates/accelerate/src/equivalence.rs
raynelfss: Oh, this one was made to be Python only on purpose. If a user wants the string representation of this item, it already implements the `Display` trait, so users in Rust can call `to_string()`. If the case happens that `EquivalenceLibrary` will not be used through Python any longer, this function can easily be removed.

qiskit/transpiler/passes/basis/basis_translator.py
raynelfss: I can see how it can potentially cause an issue. Should I maybe use a really big number close to infinity as the number of qubits in this case?

jlapeyre: I was thinking of some sort of generated string rather than "key". But changing the integer is probably fine. We probably want to avoid exceeding the size of the appropriate machine integer.
$2^{31} - 1$ is the maximum value of `i32`.

crates/accelerate/src/equivalence.rs
jlapeyre: I think I was getting at your last point in a very lazy way when I said "true that this is a dunder". I meant modified, removed, whatever.  It's just that there is no advantage to adding the (slight) Python complexity. ... Idk, on second thought because using `PyRef<Self>` prevents it being called from plain `Rust`, it forces plain `Rust` consumers to use the correct interface `to_string`. I'm ok with leaving this as it is.

qiskit/circuit/equivalence.py
jlapeyre: Oh sorry, I did not check who calls these. `Key` and `NodeData` are called from Python in `class BasisTranslator`. Looks like they cannot be made private after all.

crates/accelerate/src/equivalence.rs
```diff
@@ -0,0 +1,786 @@
...
+        self.ne(other)
+    }
+    fn __ge__(&self, other: &Self) -> bool {
+        self.ge(other)
+    }
+    fn __gt__(&self, other: &Self) -> bool {
+        self.gt(other)
+    }
+}
+
```
jlapeyre, line 116: With the following, you can replace the five constructions of `Key` with `Key::from_operation` and `Key::from_gate`
```suggestion
impl Key {
    fn from_operation(operation: &PackedOperation) -> Self {
        let op_ref: OperationRef = operation.view();
        Key {
            name: op_ref.name().to_string(),
            num_qubits: op_ref.num_qubits(),
        }        
    }
    fn from_gate(gate: &GateOper) -> Self {
        Key::from_operation(&gate.operation)
    }
}

```
Or,  you might even omit `from_gate` and just do `let key = Key::from_operation(&gate.operation)`

crates/accelerate/src/equivalence.rs
```diff
@@ -0,0 +1,786 @@
...
+            operation: op_struct.operation,
+            params: op_struct.params,
+        })
+    }
+}
+
+/// Representation of QuantumCircuit which the original circuit object + an
+/// instance of `CircuitData`.
+#[derive(Debug, Clone)]
+pub struct CircuitRep(pub CircuitData);
```
jlapeyre, line 304: Why do you need to wrap `CircuitData` in `CircuitRep`?

jlapeyre:
This looks great for the most part. I left a few questions and suggestions.

crates/accelerate/src/equivalence.rs
raynelfss: This is here just for seamless conversion to Python, since right now the `BasisTranslator` needs to receive circuits, although I will be ultimately changing this since my implementation of `BasisTranslator` is made to use `CircuitData` by default. I will add a `TODO` comment describing why this is necessary at the moment.

crates/accelerate/src/equivalence.rs
raynelfss: Hmm... Why is that the case here? I understand that some methods are not currently being used, however this is part of the `EquivalenceLibrary`'s original functionality and it was a public function by definition.

crates/accelerate/src/equivalence.rs
raynelfss: Key in this case should stay as a public struct because the function `EquivalenceLibrary::keys()` returns an iterator of `Key` instances, I can make the method public only to the crate as well but it is a little inconvenient since `EquivalenceLibrary::node_index(key)` is also a public method that in this case accepts `Key` as an argument. I could use `map` to turn them into tuple instances of `(&str, u32)` in both cases. However, this will require a `Clone` in the case of `EquivalenceLibrary::node_index(key)` since the `Key` struct stores a `String` instanced which would require for me to call `to_owned` on the `&str` instance.

jlapeyre (approved):
Looks good.

PR #12588, opened 2024-06-17
Oxidize `synth_cnot_count_full_pmh`

### LINKED ISSUE #12106: The synthesis function ``synth_cnot_count_full_pmh`` returns invalid circuits for certain ``section_size`` values

### LINKED ISSUE #12227: Port `synth_cnot_count_full_pmh` to Rust

Cryoris:
### Summary

Closes #12227. 

In a benchmark to synthesize 100 qubit linear networks 1000 times, the runtime is ~44 times faster:
```
(main) Took 172.75505709648132s
(this PR) Took 3.920625925064087s
```
I would assume the main speedup is due to a more efficient circuit construction.

### Details and comments

* Does not address #12106, which is currently tackled in a different PR. 
* There certainly is some way to use iterators more by iterating over sections, but I couldn't yet wrap my head around it (and it might not give a significant speedup), suggestions are welcome 🙂

### DIFF
--- a/crates/accelerate/src/synthesis/linear/pmh.rs
+++ b/crates/accelerate/src/synthesis/linear/pmh.rs
@@ -1,0 +1,195 @@
+// This code is part of Qiskit.
+//
+// (C) Copyright IBM 2024
+//
+// This code is licensed under the Apache License, Version 2.0. You may
+// obtain a copy of this license in the LICENSE.txt file in the root directory
+// of this source tree or at http://www.apache.org/licenses/LICENSE-2.0.
+//
+// Any modifications or derivative works of this code must retain this
+// copyright notice, and modified files need to carry a notice indicating
+// that they have been altered from the originals.
+
+use hashbrown::HashMap;
+use ndarray::{s, Array1, Array2, ArrayViewMut2, Axis};
+use numpy::PyReadonlyArray2;
+use smallvec::smallvec;
+use std::cmp;
+
+use qiskit_circuit::circuit_data::CircuitData;
+use qiskit_circuit::operations::{Param, StandardGate};
+use qiskit_circuit::Qubit;
+
+use pyo3::prelude::*;
+
+use super::utils::_add_row_or_col;
+
+/// This helper function allows transposed access to a matrix.
+fn _index(transpose: &bool, i: &usize, j: &usize) -> (usize, usize) {
+    if *transpose {
+        (*j, *i)
+    } else {
+        (*i, *j)
+    }
+}
+
+fn _ceil_fraction(numerator: &usize, denominator: &usize) -> usize {
+    let mut fraction = numerator / denominator;
+    if numerator % denominator > 0 {
+        fraction += 1;
+    }
+    fraction
+}
+
+/// This function is a helper function of the algorithm for optimal synthesis
+/// of linear reversible circuits (the Patel–Markov–Hayes algorithm). It works
+/// like gaussian elimination, except that it works a lot faster, and requires
+/// fewer steps (and therefore fewer CNOTs). It takes the matrix and
+/// splits it into sections of size section_size. Then it eliminates all non-zero
+/// sub-rows within each section, which are the same as a non-zero sub-row
+/// above. Once this has been done, it continues with normal gaussian elimination.
+/// The benefit is that with small section sizes, most of the sub-rows will
+/// be cleared in the first step, resulting in a factor ``section_size`` fewer row row operations
+/// during Gaussian elimination.
+///
+/// The algorithm is described in detail in the following paper
+/// "Optimal synthesis of linear reversible circuits."
+/// Patel, Ketan N., Igor L. Markov, and John P. Hayes.
+/// Quantum Information & Computation 8.3 (2008): 282-294.
+///
+/// Note:
+/// This implementation tweaks the Patel, Markov, and Hayes algorithm by adding
+/// a "back reduce" which adds rows below the pivot row with a high degree of
+/// overlap back to it. The intuition is to avoid a high-weight pivot row
+/// increasing the weight of lower rows.
+///
+/// Args:
+///     matrix: square matrix, describing a linear quantum circuit
+///     section_size: the section size the matrix columns are divided into
+///
+/// Returns:
+///     A vector of CX locations (control, target) that need to be applied.
+fn lower_cnot_synth(
+    mut matrix: ArrayViewMut2<bool>,
+    section_size: &usize,
+    transpose: &bool,
+) -> Vec<(usize, usize)> {
+    // The vector of CNOTs to be applied. Called ``circuit`` here for consistency with the paper.
+    let mut circuit: Vec<(usize, usize)> = Vec::new();
+    let cutoff = 1;
+
+    // to apply to the transposed matrix, we can just set axis = 1
+    let row_axis = if *transpose { Axis(1) } else { Axis(0) };
+
+    // get number of columns (same as rows) and the number of sections
+    let n = matrix.raw_dim()[0];
+    let num_sections = _ceil_fraction(&n, section_size);
+
+    // iterate over the columns
+    for section in 1..num_sections + 1 {
+        // store sub section row patterns here, which we saw already
+        let mut patterns: HashMap<Array1<bool>, usize> = HashMap::new();
+        let section_slice = s![(section - 1) * section_size..cmp::min(section * section_size, n)];
+
+        // iterate over the rows (note we only iterate from the diagonal downwards)
+        for row_idx in (section - 1) * section_size..n {
+            // we need to keep track of the rows we saw already, called ``pattern`` here
+            let pattern: Array1<bool> = matrix
+                .index_axis(row_axis, row_idx)
+                .slice(section_slice)
+                .to_owned();
+
+            // skip if the row is empty (i.e. all elements are false)
+            if pattern.iter().any(|&el| el) {
+                if patterns.contains_key(&pattern) {
+                    // store CX location
+                    circuit.push((patterns[&pattern], row_idx));
+                    // remove the row
+                    _add_row_or_col(matrix.view_mut(), transpose, patterns[&pattern], row_idx);
+                } else {
+                    // if we have not seen this pattern yet, keep track of it
+                    patterns.insert(pattern, row_idx);
+                }
+            }
+        }
+
+        // gaussian eliminate the remainder of the section
+        for col_idx in (section - 1) * section_size..cmp::min(section * section_size, n) {
+            let mut diag_el = matrix[[col_idx, col_idx]];
+
+            for r in col_idx + 1..n {
+                if matrix[_index(transpose, &r, &col_idx)] {
+                    if !diag_el {
+                        _add_row_or_col(matrix.view_mut(), transpose, r, col_idx);
+                        circuit.push((r, col_idx));
+                        diag_el = true
+                    }
+                    _add_row_or_col(matrix.view_mut(), transpose, col_idx, r);
+                    circuit.push((col_idx, r));
+                }
+
+                // back-reduce to the pivot row: this one-line-magic checks if the logical AND
+                // between the two target rows has more ``true`` elements is larger than the cutoff
+                if matrix
+                    .index_axis(row_axis, col_idx)
+                    .iter()
+                    .zip(matrix.index_axis(row_axis, r).iter())
+                    .map(|(&i, &j)| i & j)
+                    .filter(|&x| x)
+                    .count()
+                    > cutoff
+                {
+                    _add_row_or_col(matrix.view_mut(), transpose, r, col_idx);
+                    circuit.push((r, col_idx));
+                }
+            }
+        }
+    }
+    circuit
+}
+
+/// Synthesize a linear function, given by a boolean square matrix, into a circuit.
+/// This function uses the Patel-Markov-Hayes algorithm, described in arXiv:quant-ph/0302002,
+/// using section-wise elimination of the rows.
+#[pyfunction]
+#[pyo3(signature = (matrix, section_size=None))]
+pub fn synth_cnot_count_full_pmh(
+    py: Python,
+    matrix: PyReadonlyArray2<bool>,
+    section_size: Option<i64>,
+) -> PyResult<CircuitData> {
+    let arrayview = matrix.as_array();
+    let mut mat: Array2<bool> = arrayview.to_owned();
+    let num_qubits = mat.nrows(); // is a quadratic matrix
+
+    // If given, use the user-specified input size. If None, we default to
+    // alpha * log2(num_qubits), as suggested in the paper, where the coefficient alpha
+    // is calibrated to minimize the upper bound on the number of row operations.
+    // In addition, we at least set a block size of 2, which, in practice, minimizes the bound
+    // until ~100 qubits.
+    let alpha = 0.56;
+    let blocksize = match section_size {
+        Some(section_size) => section_size as usize,
+        None => std::cmp::max(2, (alpha * (num_qubits as f64).log2()).floor() as usize),
+    };
+
+    // compute the synthesis for the lower triangular part of the matrix, and then
+    // apply it on the transposed part for the full synthesis
+    let lower_cnots = lower_cnot_synth(mat.view_mut(), &blocksize, &false);
+    let upper_cnots = lower_cnot_synth(mat.view_mut(), &blocksize, &true);
+
+    // iterator over the gates
+    let instructions = upper_cnots
+        .iter()
+        .map(|(i, j)| (*j, *i))
+        .chain(lower_cnots.into_iter().rev())
+        .map(|(ctrl, target)| {
+            (
+                StandardGate::CXGate,
+                smallvec![],
+                smallvec![Qubit(ctrl as u32), Qubit(target as u32)],
+            )
+        });
+
+    CircuitData::from_standard_gates(py, num_qubits as u32, instructions, Param::Float(0.0))
+}
--- a/crates/accelerate/src/synthesis/linear/pmh.rs
+++ b/crates/accelerate/src/synthesis/linear/pmh.rs
@@ -25,15 +25,15 @@ use pyo3::prelude::*;
 use super::utils::_add_row_or_col;
 
 /// This helper function allows transposed access to a matrix.
-fn _index(transpose: &bool, i: &usize, j: &usize) -> (usize, usize) {
-    if *transpose {
-        (*j, *i)
+fn _index(transpose: bool, i: usize, j: usize) -> (usize, usize) {
+    if transpose {
+        (j, i)
     } else {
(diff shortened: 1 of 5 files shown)


crates/accelerate/src/synthesis/cnot_synthesis.rs
```diff
@@ -0,0 +1,163 @@
...
+/// This function uses the Patel-Markov-Hayes algorithm, described in arXiv:quant-ph/0302002,
+/// using section-wise elimination of the rows.
+#[pyfunction]
+#[pyo3(signature = (matrix, section_size=2))]
+pub fn synth_cnot_count_full_pmh(
+    py: Python,
+    matrix: PyReadonlyArray2<bool>,
+    section_size: i64,
+) -> PyResult<CircuitData> {
+    let mut mat: Array2<bool> = matrix.as_array().to_owned();
```
ShellyGarion: perhaps you can use: 
`let mut mat: ArrayViewMut2<bool> = = matrix.as_array_mut()` and then you don't need to use `.to_owned()`

crates/accelerate/src/lib.rs
```diff
@@ -29,6 +29,7 @@ pub mod sabre;
 pub mod sampled_exp_val;
 pub mod sparse_pauli_op;
 pub mod stochastic_swap;
+pub mod synthesis;
```
ShellyGarion: we should discuss the lib structure. in my PR #12456 I called the module `linear_matrix`, here it's `synthesis` and in #12601 it's `clifford`. we probably need some hierarchy here.

crates/accelerate/src/synthesis/cnot_synthesis.rs
```diff
@@ -0,0 +1,163 @@
...
+            (
+                StandardGate::CXGate,
+                smallvec![],
+                smallvec![Qubit(ctrl as u32), Qubit(target as u32)],
+            )
+        });
+
+    Ok(
+        CircuitData::from_standard_gates(py, num_qubits as u32, instructions, Param::Float(0.0))
+            .expect("Something went sideways in Qiskit's Python realm!"),
```
ShellyGarion: is it a standard Qiskit error?

crates/accelerate/src/synthesis/cnot_synthesis.rs
```diff
@@ -0,0 +1,163 @@
...
+                    patterns.insert(pattern, row_idx);
+                }
+            }
+        }
+
+        // gaussian eliminate the remainder of the section
+        for col_idx in (section - 1) * section_size..section * section_size {
+            let mut diag_el = matrix[[col_idx, col_idx]];
+
+            for r in col_idx + 1..n {
```
ShellyGarion: I think this can be done using iterators like in the usual gaussian elimination in #12456

crates/accelerate/src/synthesis/cnot_synthesis.rs
```diff
@@ -0,0 +1,163 @@
...
+
+use qiskit_circuit::circuit_data::CircuitData;
+use qiskit_circuit::operations::{Param, StandardGate};
+use qiskit_circuit::Qubit;
+
+use pyo3::prelude::*;
+
+/// Mutate a matrix inplace by adding the value of the ``ctrl`` row to the
+/// ``target`` row. If ``add_columns`` is true, add columns instead of rows.
+fn _add(mat: &mut Array2<bool>, add_columns: &bool, ctrl: usize, trgt: usize) {
```
ShellyGarion: is it the same function as `_row_op` and `_col_op` in #12456 ?
do we want to keep just one of these suggestions?

ShellyGarion (commented):
Generally, it looks very well. Please see some comments and questions. 
Also, release notes are needed.

crates/accelerate/src/synthesis/cnot_synthesis.rs
```diff
@@ -0,0 +1,163 @@
...
+/// using section-wise elimination of the rows.
+#[pyfunction]
+#[pyo3(signature = (matrix, section_size=2))]
+pub fn synth_cnot_count_full_pmh(
+    py: Python,
+    matrix: PyReadonlyArray2<bool>,
+    section_size: i64,
+) -> PyResult<CircuitData> {
+    let mut mat: Array2<bool> = matrix.as_array().to_owned();
+    // let mut mat = matrix.to_owned();
```
ShellyGarion: this comment should be removed (also the one 3 lines later)

crates/accelerate/src/synthesis/cnot_synthesis.rs
Cryoris: Yes it's the same indeed! I don't mind in which PR we add it, maybe we could just keep which ever merges first and then update the other one? Or we can wait with this until #12456 merges 🙂

crates/accelerate/src/synthesis/cnot_synthesis.rs
Cryoris: Yeah, from what I understand this must be handled every time we use the `from_standard_gates`, see e.g. https://github.com/mtreinish/qiskit-core/commit/4edf3fbd3d6f763053d67cc6dadec83f5174228f#diff-61c072175672ef132e5e682a51401abd08288938d864b2090f4dc446e772837eR855-R868 🙂 (Also the compiler is complaining otherwise)

crates/accelerate/src/lib.rs
Cryoris: Yes, let's sync offline 👍🏻

crates/accelerate/src/synthesis/cnot_synthesis.rs
Cryoris: The `to_owned` is currently necessary since the Python matrix remains unchanged -- so we copy it here to be able to modify it in Rust space 🙂

crates/accelerate/src/synthesis/cnot_synthesis.rs
ShellyGarion: in this case we should discuss the preferred API (let's talk offline) - 
do we prefer one function or two (for row and col)?
the name "_add" is too general. maybe "_add_row_or_col" ?

crates/accelerate/src/synthesis/cnot_synthesis.rs
Cryoris: I could get it more or less to work without the "backreduce" in the end of the iteration (L147 and following). But the backreduce accesses the pivot row repeatedly and Rust is complaining about multiple mutable borrows in the iterator... If you have any more ideas I'm happy to try more 🙂

crates/accelerate/src/synthesis/cnot_synthesis.rs
```diff
@@ -0,0 +1,188 @@
...
+
+use qiskit_circuit::circuit_data::CircuitData;
+use qiskit_circuit::operations::{Param, StandardGate};
+use qiskit_circuit::Qubit;
+
+use pyo3::prelude::*;
+
+/// Mutate a matrix inplace by adding the value of the ``ctrl`` row to the
+/// ``target`` row. If ``add_cols`` is true, add columns instead of rows.
+fn _add_row_or_col(mat: &mut Array2<bool>, add_cols: &bool, ctrl: usize, trgt: usize) {
```
ShellyGarion: I think it is better to have:
`fn _add_row_or_col(mut mat: ArrayViewMut2<bool>, add_cols: &bool, ctrl: usize, trgt: usize) {`
I added this function to linear matrix utils in #12456

Cryoris: Done in 5684f62 🙂

crates/accelerate/src/synthesis/cnot_synthesis.rs
```diff
@@ -0,0 +1,188 @@
...
+
+use qiskit_circuit::circuit_data::CircuitData;
+use qiskit_circuit::operations::{Param, StandardGate};
+use qiskit_circuit::Qubit;
+
+use pyo3::prelude::*;
+
+/// Mutate a matrix inplace by adding the value of the ``ctrl`` row to the
+/// ``target`` row. If ``add_cols`` is true, add columns instead of rows.
+fn _add_row_or_col(mut mat: ArrayViewMut2<bool>, add_cols: &bool, ctrl: usize, trgt: usize) {
```
ShellyGarion: this function is now in synthesis/linear/utils.rs

crates/accelerate/src/synthesis/cnot_synthesis.rs
```diff
@@ -0,0 +1,188 @@
...
+        (s![ctrl, ..], s![trgt, ..])
+    };
+    let (row0, mut row1) = mat.multi_slice_mut(info);
+
+    // add them inplace
+    row1.zip_mut_with(&row0, |x, &y| *x ^= y);
+}
+
+/// This helper function allows transposed access to a matrix.
+fn _index(transpose: &bool, i: &usize, j: &usize) -> (usize, usize) {
```
ShellyGarion: if this is a helper function for linear matrices, I would suggest to add it to synthesis/linear/utils.rs

crates/accelerate/src/synthesis/cnot_synthesis.rs
```diff
@@ -0,0 +1,188 @@
...
+#[pyo3(signature = (matrix, section_size=2))]
+pub fn synth_cnot_count_full_pmh(
+    py: Python,
+    matrix: PyReadonlyArray2<bool>,
+    section_size: i64,
+) -> PyResult<CircuitData> {
+    let arrayview = matrix.as_array();
+    let mut mat: Array2<bool> = arrayview.to_owned();
+    // let mat: ArrayViewMut2<bool> = matrix.as_array().to_owned().view_mut();
+    // let mut mat: ArrayViewMut2<bool> = matrix.as_array().to_owned().view_mut();
```
ShellyGarion: these 2 commented lines should be removed

crates/accelerate/src/synthesis/cnot_synthesis.rs
```diff
@@ -0,0 +1,188 @@
...
+    } else {
+        (*i, *j)
+    }
+}
+
+/// Synthesize a linear function, given by a boolean square matrix, into a circuit.
+/// This function uses the Patel-Markov-Hayes algorithm, described in arXiv:quant-ph/0302002,
+/// using section-wise elimination of the rows.
+#[pyfunction]
+#[pyo3(signature = (matrix, section_size=2))]
```
ShellyGarion: I would suggest to set the default value of section_size to `None` and if it's not given then use `log(n)` instead, as it produces a lower number of CX gates, after the bugs mentioned in #12166 will be fixed (see the discussion in #12166).

ShellyGarion (changes requested):
This code looks good, but it should be refactored as #12456 was merged.

My main comment is that perhaps we should consider incorporating the suggestion in 
https://github.com/Qiskit/qiskit/pull/12166#issuecomment-2198448523 to handle invalid values `section_size` as well as tests to verify the correctness of the valid circuit (see https://github.com/Qiskit/qiskit/pull/12166#issuecomment-2088040391).
Then, perhaps it would be better to have the default `section_size` as `log(n)` since it produces a circuit with smaller CX count.

qiskit/synthesis/linear/cnot_synth.py
```diff
@@ -17,33 +17,34 @@
...
+            the state of the input circuit.
         section_size: The size of each section in the Patel–Markov–Hayes algorithm [1].
 
     Returns:
-        QuantumCircuit: a CX-only circuit implementing the linear transformation.
+        A CX-only circuit implementing the linear transformation.
 
     Raises:
-        QiskitError: when variable ``state`` isn't of type ``numpy.ndarray``
+        QiskitError: when variable ``state`` isn't of type ``numpy.ndarray``.
```
ShellyGarion: do we still raise this error?

crates/accelerate/src/synthesis/cnot_synthesis.rs
ShellyGarion: Actually I'm not sure that it's better, see https://github.com/Qiskit/qiskit/pull/12166#issuecomment-2198521174

crates/accelerate/src/synthesis/cnot_synthesis.rs
Cryoris: I don't think this is used anywhere else, so it might be cleaner to keep it private in this file -- if we find that we need it elsewhere I agree we should move it 🙂

qiskit/synthesis/linear/cnot_synth.py
Cryoris: Yeah we still do, I didn't touch that check (just added proper sentence punctuation 😉) but looking at it I would prefer removing it as the type is given in the type hint so we assume it is correctly given.

releasenotes/notes/oxidize-pmh-ec74e4002510eaad.yaml
```diff
@@ -0,0 +1,6 @@
+---
+upgrade_synthesis:
```
alexanderivrii: We should mark this as `features_synthesis` not as `upgrade_synthesis`; I have just pushed a tiny PR #12715 that fixes it in other places.

crates/accelerate/src/synthesis/cnot_synthesis.rs
Cryoris: Done as discussed offline 🙂

qiskit/synthesis/linear/cnot_synth.py
Cryoris: Now we don't anymore, I removed that check.

qiskit/synthesis/linear/cnot_synth.py
```diff
@@ -17,126 +17,53 @@
...
            Quantum Information & Computation 8.3 (2008): 282-294.
            `arXiv:quant-ph/0302002 [quant-ph] <https://arxiv.org/abs/quant-ph/0302002>`_
     """
-    if not isinstance(state, (list, np.ndarray)):
-        raise QiskitError(
-            f"state should be of type list or numpy.ndarray, but was of the type {type(state)}"
+    normalized = np.asarray(state).astype(bool)
+    if section_size is not None and normalized.shape[1] < section_size:
+        raise ValueError(
+            f"The section_size ({section_size}) is cannot be larger than the number of columns "
```
ShellyGarion: either "cannot be larger" or "is larger" but not both

releasenotes/notes/oxidize-pmh-ec74e4002510eaad.yaml
```diff
@@ -0,0 +1,6 @@
+---
+features_synthesis:
```
ShellyGarion, line 2: I think that it's worth to add that it also fixes a bug for certain values of section_size #12106,
and also that the default value of section_size has been changed to allow a more efficient synthesis for larger number of qubits

test/python/synthesis/test_linear_synthesis.py
```diff
@@ -139,6 +140,29 @@ def test_synth_lnn_kms(self, num_qubits):
...
+        qc = QuantumCircuit(n)
+        for i in range(n - 1):
+            qc.cx(i, i + 1)
+        expected = Operator(qc)
+
+        mat = LinearFunction(qc).linear
+        for section_size in range(1, n + 1):
+            synthesized = synth_cnot_count_full_pmh(mat, section_size=section_size)
+            with self.subTest(section_size=section_size):
+                self.assertEqual(expected, Operator(synthesized))
```
ShellyGarion: instead of using `Operator`, you can use: `mat_synth = LinearFunction(qc).linear` and check that it's identical to the original `mat`. The matrix is `nxn` and not `(2^n)x(2^n)`, so you can also check larger values of `n`.

test/python/synthesis/test_linear_synthesis.py
```diff
@@ -139,6 +140,29 @@ def test_synth_lnn_kms(self, num_qubits):
                 dist = abs(q0 - q1)
                 self.assertEqual(dist, 1)
 
+    def test_pmh_section_sizes(self):
```
ShellyGarion: I would be happy if we can add some pseudo-random tests with various values of `section_size` as well as `None`, see e.g  this test:
https://github.com/Qiskit/qiskit/blob/2c3ddf1e6f81c497297b1d3531f20ea62b4cb620/test/python/synthesis/test_linear_synthesis.py#L118

ShellyGarion (commented):
Thanks @Cryoris - this PR looks great, I have some small comments.

crates/accelerate/src/synthesis/linear/pmh.rs
```diff
@@ -0,0 +1,195 @@
...
+use qiskit_circuit::operations::{Param, StandardGate};
+use qiskit_circuit::Qubit;
+
+use pyo3::prelude::*;
+
+use super::utils::_add_row_or_col;
+
+/// This helper function allows transposed access to a matrix.
+fn _index(transpose: &bool, i: &usize, j: &usize) -> (usize, usize) {
+    if *transpose {
```
alexanderivrii: Is there a reason to use `&bool`, `&usize` instead of `bool`, `usize`? I don't think we get any memory saving by passing the arguments "by reference" rather than directly.

crates/accelerate/src/synthesis/linear/pmh.rs
```diff
@@ -0,0 +1,195 @@
...
+        (*j, *i)
+    } else {
+        (*i, *j)
+    }
+}
+
+fn _ceil_fraction(numerator: &usize, denominator: &usize) -> usize {
+    let mut fraction = numerator / denominator;
```
alexanderivrii: same question here about `&usize` vs `usize`

alexanderivrii (commented):
The code looks amazing! I just had one tiny question about function arguments.

test/python/synthesis/test_linear_synthesis.py
Cryoris: I just checked the smallest value for which we saw failures 🙂 I'll increase a bit.

crates/accelerate/src/synthesis/linear/pmh.rs
Cryoris: Nope, just initial Rust hiccups 😉 I'll change to use by value (same for your comment below).

releasenotes/notes/oxidize-pmh-ec74e4002510eaad.yaml
Cryoris: Added 🙂

test/python/synthesis/test_linear_synthesis.py
Cryoris: Done!

ShellyGarion (approved):
LGTM! Thanks @Cryoris !

PR #12593, opened 2024-06-17
Accept `Option<&str>` instead of `&Option<String>`, etc

jlapeyre:
Accept a wider range of input types, and more idiomatic input types in a few functions. This affects code added in the gates-in-rust PR.

In a few places, this removes unnecessary object copies. 

DONE: ~The same change could be made in several other places. I can add these in another commit to this PR.~


crates/circuit/src/circuit_instruction.rs
```diff
@@ -163,7 +163,7 @@ impl CircuitInstruction {
         qubits: Option<&Bound<PyAny>>,
         clbits: Option<&Bound<PyAny>>,
         params: SmallVec<[Param; 3]>,
-        label: Option<String>,
+        label: Option<&str>,
```
jakelishman: I think this particular one is mistaken; it's correct that the `label` is owned in this constructor.

crates/circuit/src/circuit_instruction.rs
```diff
@@ -363,7 +363,7 @@ impl CircuitInstruction {
         qubits: Option<&Bound<PyAny>>,
         clbits: Option<&Bound<PyAny>>,
         params: Option<SmallVec<[Param; 3]>>,
-        label: Option<String>,
+        label: Option<&str>,
```
jakelishman: This is similarly correct that it's actually able to transfer the ownership.

jakelishman (commented):
This is a good idea in general, thanks.  I have a branch locally that does much of this as part of a different performance change, but I can update that.

jakelishman:
#12594 is my PR that'll clash, btw - I can update that, though.

jlapeyre:
That was faster than I expected. I fixed some clippy complaints and rebased in order to force push because surely no one has looked at this PR yet. I'll push it anyway. I'll compare to your PR; maybe this one is not necessary.

jakelishman:
I'm fine to merge this (or some form of it) before mine - the similar changes in mine are coincidental.

crates/circuit/src/circuit_instruction.rs
jlapeyre: My intention was to move the creation of the `String` into the constructor rather than the caller. Is the problem that this method `py_new` is called from Python and has to build a `String` rather than `&str`?  Then a ref is passed and another `String` is created with `label.map(|x| x.to_string())` which is of course wasteful ?

crates/circuit/src/circuit_instruction.rs
jlapeyre: Hmmm. I guess if  we want to transfer ownership, then there is no advantage to building a `String` in the constructor. Instead require that the caller constructs a `String`. And move it. Otherwise you construct the `String` twice, once in the caller and once in the callee.

jlapeyre:
> I'm fine to merge this (or some form of it) before mine - the similar changes in mine are coincidental.

Ok. Then I should 1) correct the observations you made above and 2) make the same changes where possible in the signatures.

crates/circuit/src/circuit_instruction.rs
jakelishman: What I'm thinking of is that if the `label` is `Some`, then we're always going to need to get an owned object.  There's no situation in which we pass `Some` to the `label` and we _won't_ need to clone it to get an owned copy, so there's nothing gained by allowing it to accept a ref, and there's something potentially lost if the caller had to create the label specifically (say from a `format!`) but can't give over the ownership.

I don't know 100% what PyO3 does when it's got an incoming `Bound<'py, PyString>` (i.e. if it makes a `String` or directly a `&str` - I suspect the latter), but accepting an owned `String` allows all possible callers to use the most efficient form.

jakelishman (dismissed):
Sorry this didn't merge sooner - I'd forgotten it hadn't before that other PR it clashed with did.

jlapeyre:
TY. NP. I removed one more obsolete thing today. If we had waited and kept rebasing every so often, it would have disappeared entirely!

jakelishman (approved):

PR #12597, opened 2024-06-18
Add remaining tests for `ParameterVector` and tweak its `repr`

garrison:
### Summary

This PR adds a few tests to bring `parametervector.py` to 100% test coverage.

### Details and comments

I also tweaked `ParameterVector`'s `__repr__` so that it puts quotes around the `name`.  Before this, one would get

```
>>> repr(ParameterVector("x", 2))
'ParameterVector(name=x, length=2)'
```

### DIFF
--- a/qiskit/circuit/parametervector.py
+++ b/qiskit/circuit/parametervector.py
@@ -87,7 +87,7 @@ class ParameterVector:
         return f"{self.name}, {[str(item) for item in self.params]}"
 
     def __repr__(self):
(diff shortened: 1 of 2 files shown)


jakelishman (approved):
I've personally never been a huge fan of making assertions about the output of `repr`, since imo it's kind of an implementation-detail subject-to-change type of thing, but I may well be in the minority on that.

At any rate, fixing the repr to have the quote marks is definitely a good thing, thanks.

PR #12598, opened 2024-06-18
Add rust representation for SGates, TGates, and iSwap gate

ElePT:
### Summary
This PR adds some of the remaining gates from #12566. The changes are related to #12572, which adds u1, u2 and u3, and #12507, which adds the RGate.

I am tempted to modify the order of the gate table to something a bit nicer but with so many open parallel PRs that modify it I think it would be a source of bugs, and it might make more sense to leave it for later.


mtreinish (commented):
The other thing you probably want to update here is adding the fast path for rust gates to the circuit methods. For example:

https://github.com/Qiskit/qiskit/blob/6d6dce327264201d7158baf6af2152bbe387fc57/qiskit/circuit/quantumcircuit.py#L4481-L4492

crates/circuit/src/operations.rs
```diff
@@ -445,9 +483,56 @@ impl Operation for StandardGate {
...
+                            (Self::CXGate, smallvec![], q0_2.clone()),
+                            (Self::TGate, smallvec![], q2.clone()),
+                            (Self::CXGate, smallvec![], q1_2),
+                            (Self::TdgGate, smallvec![], q2.clone()),
+                            (Self::CXGate, smallvec![], q0_2),
+                            (Self::TGate, smallvec![], q1.clone()),
+                            (Self::TGate, smallvec![], q2.clone()),
+                            (Self::HGate, smallvec![], q2),
+                            (Self::CXGate, smallvec![], q0_1.clone()),
+                            (Self::TGate, smallvec![], smallvec![Qubit(0)]),
```
mtreinish, line 527: No variable for this one? :)

mtreinish (approved):
LGTM, thanks for tackling these.

PR #12605, opened 2024-06-19
Follow up on #12327: circuit construction in Rust

Cryoris:
### Summary

#12327 ported the utility functions for `synth_permutation_basic` to Rust, but not the circuit construction which, at the time, wasn't possible. Now that we can construct circuits in Rust directly, this PR follows up by doing exactly that (and also adds a reno).


crates/accelerate/src/permutation.rs
```diff
@@ -111,10 +116,31 @@ fn _get_ordered_swap(py: Python, permutation_in: PyArrayLike1<i64>) -> PyResult<
...
+        get_ordered_swap(&view).iter().map(|(i, j)| {
+            (
+                StandardGate::SwapGate,
+                smallvec![],
+                smallvec![Qubit(*i as u32), Qubit(*j as u32)],
+            )
+        }),
+        Param::Float(0.0),
+    )
+    .expect("Something went wrong in Qiskit's Python realm"))
```
mtreinish: I'm going to enjoy when we have a user complaining about an uncatchable `PanicException` complaining about the python realm when calling `Qubit()` fails unexpectedly. :p

crates/accelerate/src/permutation.rs
```diff
@@ -111,10 +116,31 @@ fn _get_ordered_swap(py: Python, permutation_in: PyArrayLike1<i64>) -> PyResult<
...
 
+#[pyfunction]
+#[pyo3(signature = (pattern))]
+fn _synth_permutation_basic(py: Python, pattern: PyArrayLike1<i64>) -> PyResult<CircuitData> {
+    let view = pattern.as_array();
+    let num_qubits = view.len();
+    Ok(CircuitData::from_standard_gates(
+        py,
+        num_qubits as u32,
+        get_ordered_swap(&view).iter().map(|(i, j)| {
```
mtreinish: Is there a reason to leave this as a `pyfunction`? It seems more natural to have this take an `ArrayView1` or a slice and leave it as solely a private rust function.

crates/accelerate/src/permutation.rs
Cryoris: Yup that'll inevitably happen at some point 😅 is there a better way of handling it right now? I was mimicking how it's done in other place so far 😬

crates/accelerate/src/permutation.rs
Cryoris: It's currently also used in the construction of the `PermutationGate` (and circuit, they are doubled atm). But I can change that code to instead use `synth_permutation_basic` which would be a bit better as it is actually part of the API as opposed to `_get_ordered_swap`.

Cryoris: Done in b0fcf2a

crates/accelerate/src/permutation.rs
```diff
@@ -95,26 +109,30 @@ fn _inverse_pattern(py: Python, pattern: PyArrayLike1<i64>) -> PyResult<PyObject
...
+            (
+                StandardGate::SwapGate,
+                smallvec![],
+                smallvec![Qubit(*i as u32), Qubit(*j as u32)],
+            )
+        }),
+        Param::Float(0.0),
+    )
+    .expect("Something went wrong in Qiskit's Python realm"))
```
mtreinish: Based on our offline discussion I think:

```suggestion
    CircuitData::from_standard_gates(
        py,
        num_qubits as u32,
        get_ordered_swap(&view).iter().map(|(i, j)| {
            (
                StandardGate::SwapGate,
                smallvec![],
                smallvec![Qubit(*i as u32), Qubit(*j as u32)],
            )
        }),
        Param::Float(0.0),
    )
```
should work and answer: https://github.com/Qiskit/qiskit/pull/12605/files#r1647033844

mtreinish: Of course this will fail rustfmt and I also didn't test it locally, but it's the basic idea.

Cryoris: Thanks! Done in b860eb0 🙂

Cryoris:
I've also added the module structure in bb03417 as discussed with @ShellyGarion, so after this merged we can follow it in the other open PRs 🙂

alexanderivrii (approved):
LGTM! Special thanks for organizing the synthesis directory structure.

ShellyGarion (approved):

PR #12608, opened 2024-06-20
Fix v2 pulse drawer 

nkanazawa1989:
### Summary

v2 pulse drawer crashes when the draw function is called with a V2 backend that doesn't have configuration and defaults attributes. Since qiskit-ibm-runtime 0.24 was released and fake V1 backends were deprecated, we should to support visualization with fake V2 backends that doen't have configuration and defaults attributes.

### DIFF
--- a/qiskit/visualization/pulse_v2/device_info.py
+++ b/qiskit/visualization/pulse_v2/device_info.py
@@ -40,7 +40,7 @@ from typing import Dict, List, Union, Optional
 
 from qiskit import pulse
 from qiskit.providers import BackendConfigurationError
-from qiskit.providers.backend import Backend
+from qiskit.providers.backend import Backend, BackendV2
 
 
 class DrawerBackendInfo(ABC):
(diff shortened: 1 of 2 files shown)


qiskit/visualization/pulse_v2/device_info.py
```diff
@@ -106,40 +106,65 @@ def create_from_backend(cls, backend: Backend):
...
+            # load qubit channel mapping
+            for qind in range(configuration.n_qubits):
+                qubit_channel_map[qind].append(configuration.drive(qubit=qind))
+                qubit_channel_map[qind].append(configuration.measure(qubit=qind))
+                for tind in range(configuration.n_qubits):
+                    try:
+                        qubit_channel_map[qind].extend(configuration.control(qubits=(qind, tind)))
+                    except BackendConfigurationError:
+                        pass
+        elif isinstance(backend, BackendV2):
```
ElePT, line 147: Lint is complaining about this being an elif instead of an else. I think it is reasonable to assume that the default is `BackendV2` and not do the instance check

ElePT (commented):
Thanks for the fix @nkanazawa1989, if you'd like we can try to include it in the 1.1.1 release today, I want to do it in a few hours, so all we need is for lint to pass (added a suggestion below). Else, we can always just add it to 1.2.

releasenotes/notes/fix-v2-pulse-drawer-d05e4e392766909f.yaml
```diff
@@ -0,0 +1,7 @@
+---
+fixes:
+  - |
+    Fixed a bug in :func:`qiskit.visualization.pulse_v2.interface.draw` that doesn't
```
ElePT, line 4:
```suggestion
    Fixed a bug in :func:`qiskit.visualization.pulse_v2.interface.draw` that didn't 
```

releasenotes/notes/fix-v2-pulse-drawer-d05e4e392766909f.yaml
```diff
@@ -0,0 +1,7 @@
+---
+fixes:
+  - |
+    Fixed a bug in :func:`qiskit.visualization.pulse_v2.interface.draw` that doesn't 
+    draw pulse schedule when the draw function is called with a :class:`.BackendV2` argument.
```
ElePT: We could use the past here to show that the bug is fixed :)

```suggestion
    Fixed a bug in :func:`qiskit.visualization.pulse_v2.interface.draw` that didn't 
    draw pulse schedules when the draw function was called with a :class:`.BackendV2` argument.
```

qiskit/visualization/pulse_v2/device_info.py
jakelishman: The other thing we sometimes do is to be more defensive and do the check, in order to raise an explicit `RuntimeError("backend object not yet supported")` for a clearer message - we don't have a `BackendV3`, but it's not inconceivable that we will expand in the future and have this same bug again.

qiskit/visualization/pulse_v2/device_info.py
```diff
@@ -106,40 +106,65 @@ def create_from_backend(cls, backend: Backend):
...
-                    qubit_channel_map[qind].extend(configuration.control(qubits=(qind, tind)))
-                except BackendConfigurationError:
+                    qubit_channel_map[qind].append(backend.measure_channel(qind))
+                except Exception:  # pylint: disable=broad-except
                     pass
+                for tind in range(backend.num_qubits):
+                    try:
+                        qubit_channel_map[qind].extend(backend.control_channel(qubits=(qind, tind)))
+                    except Exception:  # pylint: disable=broad-except
+                        pass
```
ElePT, line 167:
```suggestion
        else:
                raise RuntimeError("Backend object not yet supported")
```

ElePT, line 167:
```suggestion
                        pass
        else:
            raise RuntimeError("Backend object not yet supported")        
```

ElePT (dismissed):
LGTM, I pushed Jake's suggestion to be able to include the PR in 1.1.1.

ElePT:
https://github.com/Mergifyio backport stable/0.46 stable/1.1

ElePT (approved):

Issue #12414, opened 2024-05-15
precision argument of BackendEstimatorV2.run should account for observable norm


kevinsung:
### What should we add?

The [current implementation](https://github.com/Qiskit/qiskit/blob/c6c45a1de9140d3051f34a90a8486e689f0a8749/qiskit/primitives/backend_estimator_v2.py#L165) sets `shots = 1 / precision**2`. This is sufficient for a single Pauli operator with norm 1, but it is insufficient for a Hamiltonian with more terms and greater norm. I think it should be `A / precision**2` where `A` is the sum of the absolute value squared of the coefficients of the Hamiltonian, `sum(np.abs(sparse_pauli_op.coeffs)**2)`.

ihincks:
I think that this issue should be filed against particular EstimatorV2 implementations, rather than the base class. This is because the base class can't really enforce how any child chooses to deal with precision. Note that "precision" should always be interpreted as "give me your best guess attempt to satisfy this precision".

For example, the StatevectorSimulator has no notion of shots whatsoever, because it just grabs observable estimates straight from the state vector, and sets the stds to 0 because they're perfect estimates.

The other thing to point out is that `shots = A / precision**2` is not always the right thing to do even if the simulation is shot based, because effects like drift might make it so that shot noise is not the only source of noise affecting precision.

kevinsung:
I agree that `shots = A / precision**2` where A scales with the Hamiltonian norm is not perfect, but I think it's a better representation of a "best guess attempt" than setting A = 1, so I think it makes sense to update the default implementation to use it.

ihincks:
But what's the default implementation? If `qiskit_ibm_runtime.EstimatorV2`, then we can move this issue over there, or to a server-side implementation. If `BackendEstimatorV2` or something in Aer, we can also move or specify.

kevinsung:
Oh I see what you mean. I've changed the title of this issue because it actually refers to BackendEstimatorV2 specifically.

kevinsung:
I opened https://github.com/Qiskit/qiskit-ibm-runtime/issues/1684 for the implementation in Runtime. It appears that Runtime uses their own implementation of BackendEstimatorV2.

PR #12618, opened 2024-06-20
Improving shots estimate to give errors within user-specified precision.

jsaroni:
### Summary
Fixes #issue-12414 (https://github.com/Qiskit/qiskit/issues/12414). When the user specifies the precision in EstimatorV2.run, the actual standard deviation from results is not bounded by or close to the user-specified precision for an observable that is a general linear combination of Pauli strings. This pull request gives a more accurate approximation to the number of shots needed for the desired precision for such observables. The tests to verify this are based on https://docs.quantum.ibm.com/api/qiskit-ibm-runtime/qiskit_ibm_runtime.EstimatorV2 and https://github.com/Qiskit/qiskit-ibm-runtime/issues/1684. 

### Details and comments
The shots estimation accounts for the sum of the absolute value squared of the coefficients of the observable as proposed in https://github.com/Qiskit/qiskit/issues/12414 to give a standard deviation that is bounded by and closer to the user-specified precision for any observable. In backend_estimator_v2.py, the _run function is modified to include the sum of the absolute value squared of the coefficients of the observable in the shots estimation.

The following plot shows that the ratio of precision from results to the user-specified precision changes with varying observable term coefficients using the current implementation. In the proposed implementation, this ratio is fixed and close to 1. 
average_precision_over_runs
The plot below shows that in the current implementation, having more than one occurrence of a term in an observable leads to precision results that are different from the user-specified precision due to a fixed number of shots for observables with different numbers of terms and error propagation. In the proposed implementation, the precision from the results remains close to the user-specified precision.
ZZ_linear_combinations


qiskit/primitives/backend_estimator_v2.py
```diff
@@ -160,9 +160,15 @@ def _validate_pubs(self, pubs: list[EstimatorPub]):
 
     def _run(self, pubs: list[EstimatorPub]) -> PrimitiveResult[PubResult]:
         pub_dict = defaultdict(list)
+        for pub in pubs:
+            data = self._preprocess_pub(pub)
+        bc_obs = data.observables
+        coeff_sq_sum = 0
+        for pauli, coeff in bc_obs[0].items():
+            coeff_sq_sum += coeff**2
```
ihincks: This has an easy 1-line replacement. You're also not using `pauli`, so it shouldn't be part of the iteration.

```suggestion
        coeff_sq_sum = sum(coeff**2 for coeff in bc_obs[0].values())
```

qiskit/primitives/backend_estimator_v2.py
```diff
@@ -160,9 +160,15 @@ def _validate_pubs(self, pubs: list[EstimatorPub]):
 
     def _run(self, pubs: list[EstimatorPub]) -> PrimitiveResult[PubResult]:
         pub_dict = defaultdict(list)
+        for pub in pubs:
+            data = self._preprocess_pub(pub)
+        bc_obs = data.observables
```
ihincks: You're preprocessing all of the pubs, but then just looking at the last one.

qiskit/primitives/backend_estimator_v2.py
```diff
@@ -160,9 +160,15 @@ def _validate_pubs(self, pubs: list[EstimatorPub]):
...
+        coeff_sq_sum = 0
+        for pauli, coeff in bc_obs[0].items():
+            coeff_sq_sum += coeff**2
         # consolidate pubs with the same number of shots
         for i, pub in enumerate(pubs):
-            shots = int(math.ceil(1.0 / pub.precision**2))
+            shots = int(math.ceil(coeff_sq_sum / pub.precision**2))
```
ihincks, line 168: Why not make coeff_sq_sum depend on the pub? With what you've written, it only depends on the last pub.

Moreover, it only depends on the first element of the last pub.

qiskit/primitives/backend_estimator_v2.py
```diff
@@ -160,9 +160,15 @@ def _validate_pubs(self, pubs: list[EstimatorPub]):
 
     def _run(self, pubs: list[EstimatorPub]) -> PrimitiveResult[PubResult]:
         pub_dict = defaultdict(list)
+        for pub in pubs:
+            data = self._preprocess_pub(pub)
```
ihincks: Please look through the `_preprocess_pub` function. It doesn't do anything relevant to your changes, so it seems like a waste to call it at all.

ihincks (commented):
Thanks for your submission.

Part of this PR should be evidence that it indeed helps the situation. This could be a plot commented on the PR, exploring different regimes. It should also include tests in the test folder.

qiskit/primitives/backend_estimator_v2.py
jsaroni: Thanks. I included test plots and made a commit to include a test function in qiskit/test/python/primitives/test_backend_estimator_v2.py. The tests are passed by the proposed implementation but not the current. Also used the 1-line replacement and removed pauli.

qiskit/primitives/backend_estimator_v2.py
jsaroni: Right. I fixed this and checked that the code works when there is more than one pub in test_backend_estimator_v2.py test_error.

qiskit/primitives/backend_estimator_v2.py
jsaroni: Sounds good. I guess it was indirectly dependent it. 

From the EstimatorV2 tutorial example bc_obs is of the form [{'II': 1.0, 'IZ': 2.0, 'XI': 3.0}] so bc_obs[0] gets all the dictionary terms {'II': 1.0, 'IZ': 2.0, 'XI': 3.0}.

qiskit/primitives/backend_estimator_v2.py
jsaroni: I had to use _preprocess_pub to get the right data type for using .values().

qiskit/primitives/backend_estimator_v2.py
```diff
@@ -162,7 +162,10 @@ def _run(self, pubs: list[EstimatorPub]) -> PrimitiveResult[PubResult]:
         pub_dict = defaultdict(list)
         # consolidate pubs with the same number of shots
         for i, pub in enumerate(pubs):
-            shots = int(math.ceil(1.0 / pub.precision**2))
+            data = self._preprocess_pub(pub)
+            bc_obs = data.observables
```
ihincks: Why not?
```suggestion
            bc_obs = pub.observables
```
It looks like you're doing a binding and then throwing it away. Why can't you just get the observables directly?

qiskit/primitives/backend_estimator_v2.py
```diff
@@ -162,7 +162,10 @@ def _run(self, pubs: list[EstimatorPub]) -> PrimitiveResult[PubResult]:
...
+            coeff_sq_sum = sum(coeff**2 for coeff in bc_obs[0].values())
```
ihincks: Why are we looking only at the first observable? Don't we want to look at the worst case or something? Eg, consider the case

```
observables = [{"XX": 1}, {"YY": 100}]
```

Since we only look at the first observable, we'll be way off for the second one.

ihincks: Eg you could try
```python
        for i, pub in enumerate(pubs):
            coeff_sq_sum = 0
            for obs in pub.observables.ravel():
                coeff_sq_sum = max(coeff_sq_sum, sum(coeff**2 for coeff in obs.values()))
            shots = int(math.ceil(coeff_sq_sum / pub.precision**2))
            pub_dict[shots].append(i)
```

test/python/primitives/test_backend_estimator_v2.py
```diff
@@ -388,6 +388,24 @@ def test_diff_precision(self, backend, abelian_grouping):
...
+        hamiltonian1 = self.hamiltonian[0].apply_layout(psi1.layout)
+        theta1 = self.theta[0]
+        job = estimator.run([(psi1, hamiltonian1, [theta1])], precision=self._precision)
+        result = job.result()
+        np.testing.assert_allclose(result[0].data.stds, [self._precision], rtol=self._rtol)
+        # The result of the second run is the same
+        job = estimator.run([(psi1, hamiltonian1, [theta1]), (psi1, hamiltonian1, [theta1])], precision=self._precision)
+        result = job.result()
+        np.testing.assert_allclose(result[0].data.stds, [self._precision], rtol=self._rtol)
+        np.testing.assert_allclose(result[1].data.stds, [self._precision], rtol=self._rtol)
```
ihincks: This should also test observable arrays that aren't 1D.

Also, since your change is specifically about the role of precision in the context of multi-term observables, you should define your own observables here, and not trust that hamiltonian1 is in the format that you like; someone might change it later and not read this test.

test/python/primitives/test_backend_estimator_v2.py
```diff
@@ -388,6 +388,24 @@ def test_diff_precision(self, backend, abelian_grouping):
         np.testing.assert_allclose(result[0].data.evs, [1.901141473854881], rtol=self._rtol)
         np.testing.assert_allclose(result[1].data.evs, [1.901141473854881], rtol=self._rtol)
 
+    @combine(backend=BACKENDS, abelian_grouping=[True, False])
+    def test_error(self, backend, abelian_grouping):
+        "Test for error"
```
ihincks: This name is a bit too generic because `error` can mean many things. Consider renaming to something that includes the word "precision". The string describing the test is also a bit too generic.

ihincks (commented):
Thanks @jsaroni! I've left a few comments.

qiskit/primitives/backend_estimator_v2.py
jsaroni: I tried this and it worked, thanks.

test/python/primitives/test_backend_estimator_v2.py
jsaroni: Changed test function name to test_precision_error.

test/python/primitives/test_backend_estimator_v2.py
jsaroni: Ok.

I defined my own observables including multi-term observables.

jsaroni:
Closed in favor of #12711

PR #12619, opened 2024-06-20
Invalidate `parameters` cache on circuit copy

### LINKED ISSUE #12617: Parameter cache is preserved after copy_empty_like
### Environment

- **Qiskit version**: main
- **Python version**: 
- **Operating system**: 

### What is happening?

Parameter appears in empty circuit after copy

```python
p1 = Parameter("P1")
qc = QuantumCircuit(1)
qc.rz(p1, 0)
qc.parameters  # dummy call to fill cache

cpy = qc.copy_empty_like()
cpy.parameters
```
I see
```
ParameterView([Parameter(P1)])
```
which should not exist.

### What should happen?

Parameter should be empty

### Any suggestions?

Nullify `_parameters` after copy

jakelishman:
### Summary

Previously, the caching of the parameter view could persist between copies of the circuit because it was part of the `copy.copy`.

### Details and comments

Fix #12617.

I verified that this is present in 1.1.0 as well.

### DIFF
--- a/qiskit/circuit/quantumcircuit.py
+++ b/qiskit/circuit/quantumcircuit.py
@@ -3674,6 +3674,8 @@ class QuantumCircuit:
         cpy._data = CircuitData(
             self._data.qubits, self._data.clbits, global_phase=self._data.global_phase
         )
+        # Invalidate parameters caching.
+        cpy._parameters = None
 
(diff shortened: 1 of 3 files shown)


Cryoris (approved):

jakelishman:
@mergifyio backport stable/0.46 stable/1.1

PR #12623, opened 2024-06-20
Add test case to validate the rust->Python gate conversion

mtreinish:
### Summary

This commit adds a test to the test_rust_equivalence module to assert that the Python gate objects returned from the Rust CircuitData is the correct type.

### Details and comments


ElePT (commented):
I think that the test is not behaving as intended, shouldn't it fail without fixing the [wrong import paths](https://github.com/Qiskit/qiskit/blob/b6c61661272c2e242963c416e64a9a2e050ea25d/crates/circuit/src/imports.rs#L127) we currently have for the T gates? (I did check it out locally and there it fails as expected). Maybe something related to the test config in CI?

mtreinish:
There is a path where those imports don't get executed: https://github.com/Qiskit/qiskit/blob/main/crates/circuit/src/circuit_instruction.rs#L743-L785 If a `TGate` python object is added to **any** circuit directly then we get the mapping automatically from that for all future uses of `StandardGate::TGate` from the class attribute. Those import paths are just a fallback that gets executed if we don't go through python on the construction side and we need to generate the py gate from something that was created directly from rust space. I think in the case of `TGate` and `TdgGate` we are unconditionally adding them to circuits here: https://github.com/Qiskit/qiskit/blob/main/qiskit/circuit/library/standard_gates/equivalence_library.py#L136-L145 when we do `import qiskit`.

ElePT (approved):
Ok, so we might not be able to catch the fallback path on CI but the test is still good to have. I guess the best we can do in this case is keep this in mind and run the test locally when relevant.

ElePT:
I removed the PR from the queue because, if I'm not mistaken, it would fail if #12646 got merged. So maybe we can wait until all gates have been added to merge the test.

mtreinish:
> I removed the PR from the queue because, if I'm not mistaken, it would fail if https://github.com/Qiskit/qiskit/pull/12646 got merged. So maybe we can wait until all gates have been added to merge the test.

I think it actually would be fine, because the test only validates if the python side class attribute `_standard_gate` is set. So for the placeholder gates we've not done that and it'll be a skip. But lets wait for #12646 to merge first and then test my hypothesis, rather than potentially blocking #12646.

ElePT (approved):
Looks to be working!

PR #12630, opened 2024-06-21
Fix `CircuitInstruction` legacy iterable typing

### LINKED ISSUE #12626: CircuitInstruction qarg type changed without communication
### Environment

- **Qiskit version**: main
- **Python version**: 
- **Operating system**: 

### What is happening?

CircuitInstruction qarg type is suddenly changed.

image

image

This can cause problem in unittest if one is testing instruction qubits without typecast.

See above screenshot

### What should happen?

Data type must be preserved, or change must be mentioned in the release note

### Any suggestions?

_No response_

jakelishman:
### Summary

The legacy 3-tuple format of `CircuitInstruction` still exposes the object in the old tuple-like way of `(Operation, list[Qubit], list[Clbit])`, rather than the new-style attribute access using tuples for the qargs and cargs.  This was inadvertantly changed when it moved to Rust.

### Details and comments

Fix #12626.

No changelog because it's not in a release version of Qiskit.

### DIFF
--- a/crates/circuit/src/circuit_instruction.rs
+++ b/crates/circuit/src/circuit_instruction.rs
@@ -542,7 +542,11 @@ impl CircuitInstruction {
 
         Ok(PyTuple::new_bound(
             py,
-            [op, self.qubits.to_object(py), self.clbits.to_object(py)],
+            [
+                op,
+                self.qubits.bind(py).to_list().into(),
+                self.clbits.bind(py).to_list().into(),
+            ],
         ))
     }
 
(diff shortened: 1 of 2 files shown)


ElePT (approved):
The fix looks good. Do you see this behavior being deprecated in the near future, or would it cause too much trouble? (seeing that it is not ideal for runtime but needed for backwards compat.)

jakelishman:
I opened #12631 to track deprecation - I'm just throwing together a PR to do it now too.

PR #6266, opened 2021-04-20
fix error when `dag_drawer` removes a temporary file that is still open

peendebak:
### Summary

When drawing a DAG with `dag_drawer` an exception can occur when the method tries to remove the temporary file while the temporary file is still open in the PIL Image. The PR fixes the problem by making a copy of the `Image` and then closing the original `Image` before removing the file. Fixes #6256

### Details and comments

For opening and closing a context manager is used as provided by `PIL`.

PR #12635, opened 2024-06-21
Fix pm visualization winerror

### LINKED ISSUE #12634: WinError 32 during using qiskit.visualization.pass_manager_visualization
### Environment

- **Qiskit version**: 1.1.1
- **Python version**: 3.12.3
- **Operating system**: Windows 11

### What is happening?

When you try to draw the Graphviz diagram with `pass_manager_visualization`, You get a `PermissionError: [WinError 32] The process cannot access the file because it is being used by another process` during running `os.remove(tmppath)`.

```
from qiskit_ibm_runtime.fake_provider import FakeKyiv

pm = generate_preset_pass_manager(optimization_level=2, backend=FakeKyiv())
pm.draw()
```

### What should happen?

Draw the Graphviz diagram

### Any suggestions?

Kind of deep copying the image would work.

```
with Image.open(tmppath) as image_:
    image = image_.copy()
os.remove(tmppath)
```

HuangJunye:
### Summary

Fix #12634, similar to #6266

### DIFF
--- a/qiskit/visualization/pass_manager_visualization.py
+++ b/qiskit/visualization/pass_manager_visualization.py
@@ -314,7 +314,7 @@ def make_output(graph, raw, filename):
 
         with Image.open(tmppath) as test_image:
             image = test_image.copy()
-
+        os.remove(tmppath)
         if filename:
             image.save(filename, "PNG")
         return image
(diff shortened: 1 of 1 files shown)


ElePT (commented):
Thanks for the fix @HuangJunye! could you please add a quick reno for the fix?

PR #12639, opened 2024-06-21
Add Rust representation for CHGate, CPhaseGate, CSGate, CSdgGate, CSXGate, CSwapGate

ElePT:
### Summary
This PR adds Rust representations for:
- `CHGate`
- `CPhaseGate`
- `CSGate`
- `CSdgGate`
- `CSXGate`
- `CSwapGate`

As part of #12566.

### Details and comments


qiskit/circuit/quantumcircuit.py
```diff
@@ -4998,7 +5008,7 @@ def s(self, qubit: QubitSpecifier) -> InstructionSet:
         Returns:
             A handle to the instructions created.
         """
-        return self._append_standard_gate(StandardGate.SGate, [], [qubit], cargs=None)
+        return self._append_standard_gate(StandardGate.SGate, [], qargs=[qubit])
```
Cryoris, line 5003: Do you want to apply this change consistently in all calls to `_standard_gate`? Above there are some that still explicitly pass `None` for `cargs`.

test/python/circuit/test_rust_equivalence.py
```diff
@@ -110,7 +111,14 @@ def test_definitions(self):
                                     [py_def.find_bit(x).index for x in py_inst.qubits],
                                     [rs_def.find_bit(x).index for x in rs_inst.qubits],
                                 )
-
+                        # Rust uses cp but python still uses cu1 in some cases
+                        elif rs_inst.operation.name == "cp":
```
Cryoris, line 115: This also applies to the exception above: would it be safer to explicitly point out the gates that behave differently?
```suggestion
                        # Rust uses cp but python still uses cu1 in some cases
                        elif name == "csx":  
                           if rs_inst.operation.name == "cp":
                               # code as before...
```
That would also mean that in the above `elif ... == "p"` we can more clearly apply special logic to the gates that behave differently in Py/Rust and don't need to add exceptions for gates that behave normally (like `cp cs csdg`).

Cryoris (commented):
Just two small comments, otherwise LGTM 🙂

qiskit/circuit/quantumcircuit.py
ElePT: I might have overlooked the newly merged ones. Given that there's another PR lined up after this one (#12659), I could make sure everything is consistent there.

test/python/circuit/test_rust_equivalence.py
ElePT: Yes, these exceptions are getting confusing. But again, I'd suggest doing the test clean-up in https://github.com/Qiskit/qiskit/pull/12659 because it introduces new cases anyways.

Cryoris (approved):
LGTM, I'm fine moving the suggestions to the follow up #12659 👍🏻

PR #8093, opened 2022-05-20
Change data type of `QuantumCircuit.data` elements

jakelishman:
### Summary

These two commits change the data type of the elements of `QuantumCircuit.data` from an ad-hoc 3-tuple `(Instruction, List[Qubit], List[Clbit])` to an encapsulating namedtuple-like class called `CircuitInstruction` with three attributes:

- `operation`: currently an `Instruction`, but will presumably be `Operation` soon.
- `qubits`: `Tuple[Qubit, ...]`
- `clbits`: `Tuple[Clbit, ...]`

The first commit, b554a7140, is the minimal set of changes needed to convert the data type, including backwards-compatibility shims.  I suggest reviewers look at this first.  The test suite passes (without modification) after this commit, which is a test of backwards compatibility.

The second commit, fb0fb32ed, changes every single usage of this tuple in Terra and its test suite over to using the new attribute-access pattern.  This is almost entirely mechanical, but there is a slight backwards-compatibility break here.  Previously, the data types of `qargs` and `cargs` in `DAGOpNode` and `DAGDepNode` were not specified, but in practice were lists most (but not all) of the time.  After this commit, they are now tuples in 100% of Terra code.  This should have little-to-no effect on downstream code, because _everything_ would have broken if anybody actually attempted to mutate one of these lists.  The only likely possibilities are if people are trying to add the values to a list (Terra _very_ occasionally tried this), or if they are comparing for equality with an exact list.

### Details and comments

See #7624 for a lot of the justification for why we want to change from an ad-hoc tuple to an extensible class as the backing scalar type.  See also #7020 for a previous attempt at this.  #7020 did not make the second commit in this PR, which is why this PR does not show the same performance regressions as that one.

I ran the test suites from Experiments and Nature against this, with no failures, suggesting the backwards compatibility shims in `CircuitInstruction` work pretty well, and the changes to `DAGOpNode` don't have an effect on downstream code.

The reason to use tuples not lists in the new instruction is to reduce memory usage on copy (such as from circuit to DAG); tuples can be bound directly, whereas lists need to be reallocated.  They are also directly usable as dictionary keys (which we sometimes do), and we have no need to mutate them.

The new `CircuitInstruction` class is not immutable itself, although I don't propose we use that in any current usages.  I anticipate we may want to use mutability when more items are in it (such as simple binding of parameters), especially if the collection underlying `QuantumCircuit` changes from a `list` to a DAG at any point.  This may not be necessary, though.

[...]

PR #12640, opened 2024-06-21
Deprecate tuple-like access to `CircuitInstruction`

### LINKED ISSUE #12631: Deprecate legacy iterable access to `CircuitInstruction`
Ever since the scalar type of `QuantumCircuit.data` changed from a 3-tuple of `(Operation, list[Qubit], list[Clbit])` to the record-like `CircuitInstruction` in #8093, we've maintained the iterable interfaces as a legacy format that makes it appear tuple-like for the purposes of `for inst, qargs, cargs in qc.data` and such.

The new-style format has been around two years now and using the legacy format involves memory and runtime penalties (to allocate new lists rather than re-using the same interned tuples).  We should deprecate the legacy interface pending removal in Qiskit 2.0 to push any remaining users onto the modern form.

jakelishman:
### Summary

This has been the legacy path since `CircuitInstruction` was added in gh-8093.  It's more performant to use the attribute-access patterns, and with more of the internals moving to Rust and potentially needing more use of additional class methods and attributes, we need to start shifting people away from the old form.

### Details and comments

Fix #12631.

The sister PR to Aer that fixes its remaining uses of this is Qiskit/qiskit-aer#2179.

### DIFF
--- a/crates/circuit/src/circuit_instruction.rs
+++ b/crates/circuit/src/circuit_instruction.rs
@@ -14,7 +14,7 @@
 use std::cell::RefCell;
 
 use pyo3::basic::CompareOp;
-use pyo3::exceptions::PyValueError;
+use pyo3::exceptions::{PyDeprecationWarning, PyValueError};
 use pyo3::prelude::*;
 use pyo3::types::{IntoPyDict, PyList, PyTuple, PyType};
 use pyo3::{intern, IntoPy, PyObject, PyResult};
@@ -22,7 +22,7 @@ use smallvec::{smallvec, SmallVec};
 
 use crate::imports::{
     get_std_gate_class, populate_std_gate_map, GATE, INSTRUCTION, OPERATION,
-    SINGLETON_CONTROLLED_GATE, SINGLETON_GATE,
+    SINGLETON_CONTROLLED_GATE, SINGLETON_GATE, WARNINGS_WARN,
 };
 use crate::interner::Index;
 use crate::operations::{OperationType, Param, PyGate, PyInstruction, PyOperation, StandardGate};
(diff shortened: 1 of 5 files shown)


crates/circuit/src/circuit_instruction.rs
```diff
@@ -939,3 +944,29 @@ pub(crate) fn convert_py_to_operation_type(
...
+                concat!(
+                    "Treating CircuitInstruction as an iterable is deprecated legacy behavior",
+                    " since Qiskit 1.2, and will be removed in Qiskit 2.0.",
+                    " Instead, use the `operation`, `qubits` and `clbits` named attributes."
+                )
+            ),
+            py.get_type_bound::<PyDeprecationWarning>(),
+            // Stack level.  Compared to Python-space calls to `warn`, this is unusually low
+            // beacuse all our internal call structure is now Rust-space and invisible to Python.
+            1,
```
mtreinish, line 969: This isn't great, but yeah there isn't really an alternative because the call stack isn't really accessible here.

mtreinish (approved):
LGTM, this is straightforward enough and it's good to have a deprecation warning pattern for Python from rust now.

crates/circuit/src/circuit_instruction.rs
jakelishman: tbh, I think it might actually work out better from Rust space than it otherwise would, because this is only called by magic methods, which have weird callstacks already, since some of the "logical" parts of it are CPython internals already.

Setting `stacklevel=1` here causes everything I could think of to get blamed on the actual triggering line in Python space.

1ucian0:
I think, randomized tests is still using this deprecated behavior https://github.com/Qiskit/qiskit/actions/runs/9656096284/job/26633022278

PR #12644, opened 2024-06-24
Add Rust representation for DCXGate

eliarbel:
### Summary
As the name suggests. 

### Details and comments
Addresses #12566


ElePT:
Hi @eliarbel, I have opened a PR with placeholders for all missing standard gates in rust so that we don't have clashing IDs when working on gates in parallel. Do you think it would be helpful? https://github.com/Qiskit/qiskit/pull/12646

eliarbel:
> Hi @eliarbel, I have opened a PR with placeholders for all missing standard gates in rust so that we don't have clashing IDs when working on gates in parallel. Do you think it would be helpful? #12646

Yeap, I've used it. Thanks @ElePT !

mtreinish (commented):
Can you update the `QuantumCircuit.dcx()` method to directly append the rust gate too? You can look at cx or h for an example of how to do this.

eliarbel:
> Can you update the `QuantumCircuit.dcx()` method to directly append the rust gate too? You can look at cx or h for an example of how to do this.

Good catch! done

crates/circuit/src/gate_matrix.rs
```diff
@@ -324,3 +331,5 @@ pub fn xx_plus_yy_gate(theta: f64, beta: f64) -> [[Complex64; 4]; 4] {
         [c64(0., 0.), c64(0., 0.), c64(0., 0.), c64(1., 0.)],
     ]
 }
+
+
```
ElePT:
```suggestion
```

crates/circuit/src/operations.rs
```diff
@@ -523,7 +523,10 @@ impl Operation for StandardGate {
             Self::CSwapGate => todo!(),
             Self::CUGate | Self::CU1Gate | Self::CU3Gate => todo!(),
             Self::C3XGate | Self::C3SXGate | Self::C4XGate => todo!(),
-            Self::DCXGate => todo!(),
+            Self::DCXGate => match params {
+                [] => Some(aview2(&gate_matrix::DCX_GATE).to_owned()),
+                _ => None,
+            }
```
ElePT:
```suggestion
            },
```

ElePT (commented):
Almost there! I left a couple of suggestions to fix the formatting complaints.

ElePT (approved):

PR #12646, opened 2024-06-24
Add placeholders for all mising standard gates in Rust

ElePT:
### Summary
This PR proposes placeholders for all missing gates in https://github.com/Qiskit/qiskit/issues/12566 with a dummy number of qubits and number of parameters (34) that should be replaced with the actual number when implemented. The idea of this proposal is to allow for a better parallelization of the work that avoids conflicts. Note that I don't particularly like this order, and I think we should reorder the lists once the implementations are completed, but I think this order is better than having to resolve conflicts every time we merge a new PR. I have also added annotations in the static vectors to increase readability.

### DIFF
--- a/crates/circuit/src/operations.rs
+++ b/crates/circuit/src/operations.rs
@@ -205,43 +205,109 @@ pub enum StandardGate {
     ISwapGate = 23,
     XXMinusYYGate = 24,
     XXPlusYYGate = 25,
+    U1Gate = 26,
+    U2Gate = 27,
+    U3Gate = 28,
+    CRXGate = 29,
+    CRYGate = 30,
+    CRZGate = 31,
+    RGate = 32,
+    CHGate = 33,
+    CPhaseGate = 34,
+    CSGate = 35,
+    CSdgGate = 36,
+    CSXGate = 37,
+    CSwapGate = 38,
+    CUGate = 39,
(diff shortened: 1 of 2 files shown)


crates/circuit/src/operations.rs
```diff
@@ -205,43 +205,109 @@ pub enum StandardGate {
...
+    "u1",           // 26
+    "u2",           // 27
+    "u3",           // 28
+    "crx",          // 29
+    "cry",          // 30
+    "crz",          // 31
+    "r",            // 32
+    "ch",           // 33
+    "cphase",       // 34
+    "cs",           // 35
```
ElePT:
```suggestion
    "cp",           // 34
    "cs",           // 35
```

crates/circuit/src/operations.rs
```diff
@@ -878,6 +953,24 @@ impl Operation for StandardGate {
                     .expect("Unexpected Qiskit python bug"),
                 )
             }),
+            Self::U1Gate | Self::U2Gate | Self::U3Gate => todo!(),
```
ElePT:
```suggestion
```

mtreinish (approved):
LGTM, this should reduce merge conflicts a bit, and also make it easier to resolve them when they do occur.

PR #12648, opened 2024-06-24
Port CRX/Y/Z gates to Rust

Cryoris:
### Summary

Port the controlled Pauli rotations to Rust. Checks off parts of #12566.


mtreinish (commented):
Can you update the `QuantumCircuit` methods to directly append the rust gate too? You can look at cx or h for an example of how to do this.

qiskit/circuit/quantumcircuit.py
```diff
@@ -4776,6 +4776,12 @@ def crx(
         """
         from .library.standard_gates.rx import CRXGate
 
+        # if the control state is |1> use the fast Rust version of the gate
+        if ctrl_state is None or ctrl_state in ["1", 1]:
```
ElePT, line 4778: Hmm I get that you added the `["1", 1]` cases because it's the state set by default when `ctrl_state` is None, but I wonder if we would be overseeing some edge case by adding this?? (If there is indeed no difference, I think we should do it in all controlled gate methods).

Cryoris: If `ctrl_state` is `None` it is directly set to 1 in the constructor of any controlled gate, e.g.
```
>>> CRXGate(theta=0.5, ctrl_state=None).ctrl_state 
1
```
so I think there should be no functional difference 🙂 

But I agree we should generally cover all `|1>` state definitions we can use the Rust implementation more generally.

Cryoris:
Picking up the `ctrl_state` edge case @ElePT mentioned above, there is a potential sharp bit we're introducing with the change to Rust (this happens independently of the `None` vs `1 "1"` discussion): changing the `ctrl_state` of a mutable gate, like CRX, once the gate is in the circuit does not change the matrix anymore, but did before. 

For example
```python

qc = QuantumCircuit(2)
qc.crx(2, 0, 1)
print(Operator(qc))  # matrix with ctrl_state 1

# not very legal but not explicitly illegal?
qc.data[0].operation.ctrl_state = 0
# w/o Rust gates this prints the correct matrix with ctrl_state 0,
# with Rust gates it still shows the matrix for ctrl_state 1
print(Operator(qc)) 
```
This may be something we need to fix globally.

ElePT:
> Picking up the `ctrl_state` edge case @ElePT mentioned above, there is a potential sharp bit we're introducing with the change to Rust (this happens independently of the `None` vs `1 "1"` discussion): changing the `ctrl_state` of a mutable gate, like CRX, once the gate is in the circuit does not change the matrix anymore, but did before.
> 
> For example
> 
> ```python
> from qiskit.quantum_info import Operator
> from qiskit.circuit import QuantumCircuit
> 
> qc = QuantumCircuit(2)
> qc.crx(2, 0, 1)
> print(Operator(qc))  # matrix with ctrl_state 1
> 
> # not very legal but not explicitly illegal?
> qc.data[0].operation.ctrl_state = 0
> # w/o Rust gates this prints the correct matrix with ctrl_state 0,
> # with Rust gates it still shows the matrix for ctrl_state 1
> print(Operator(qc)) 
> ```
> 
> This may be something we need to fix globally.

This sounds like a friend of:

```
gate.definition.global_phase = 3.14159
dag_node.op.params[0] = 3.14159
```

I don't think we'll be able to mutate any gate element post-Rust migration using that syntax ([reno](https://github.com/Qiskit/qiskit/blob/2fab2007e3d7384e9e55a10340952c4974ece03c/releasenotes/notes/circuit-gates-rust-5c6ab6c58f7fd2c9.yaml#L4))

mtreinish:
The other issue with `ctrl_state` that I realized we're missing right now is what I pointed out here: https://github.com/Qiskit/qiskit/pull/12659#discussion_r1654466050 the handling in `CircuitInstruction`'s rust code for when you do `CircuitInstruction(CRXGate(pi, ctrl_state=0), ...)` doesn't understand that `CRXGate` with `ctrl_state=0` is not the standard gate definition. We will need to update the logic in that function to check the value of `ctrl_state` for a non-singleton controlled gate.

ElePT:
> The other issue with `ctrl_state` that I realized we're missing right now is what I pointed out here: [#12659 (comment)](https://github.com/Qiskit/qiskit/pull/12659#discussion_r1654466050) the handling in `CircuitInstruction`'s rust code for when you do `CircuitInstruction(CRXGate(pi, ctrl_state=0), ...)` doesn't understand that `CRXGate` with `ctrl_state=0` is not the standard gate definition. We will need to update the logic in that function to check the value of `ctrl_state` for a non-singleton controlled gate.

Yes, I am trying to fix it as part of #12659.

ElePT (approved):
Thanks a lot, I added a path to handle custom control states in [9b47121](https://github.com/Qiskit/qiskit/pull/12659/commits/9b47121e1afb51e8f3bdd609edbb67939fd0286d), and I will also update the other circuit construction methods to add the `ctrl_state in ["1", 1]` case once this PR is merged.

PR #10792, opened 2023-09-07
Deprecate `bind_parameters` in favor of `assign_parameters`

ElePT:
### Summary
Fixes #7057 (unless there is any strong argument to keep both methods in Qiskit 1.0).
Fix #8958

### Details and comments
`bind_parameters` was only used in a couple of places internally. I went through these to make sure that we would not need any additional checks for the assigned values (as `bind_parameters` would fail for `ParameterExpression`s):

- `Sampler, Estimator, BackendSampler, BackendEstimator`
- `_expand_parameters` in `qiskit/compiler/assembler.py`
- `DefaultUnitarySynthesis` transpiler pass
- `UnrollForLoops` transpiler pass

I don't think these checks would be needed, but let me know if you disagree. 

I was also unsure about modifying the `qiskit.algorithms` related code, as it is going to be removed anyway, but honestly the search and replace was pretty much 0 effort compared to catching the deprecation warnings or creating exceptions in the test case. I have in any case done this in a separate commit in case we think it should be reverted.

PR #12649, opened 2024-06-24
Deprecate Qobj and assemble

1ucian0:
### Summary

Deprecating `assemble` and `Qobj`.

## Detailed aspects

The main approach was to rename `assemble`  (as public API) to `_assemble`. Ditto for `assemble_circuits`.

Because Aer is using some of the deprecated functionality, the warnings are catched by few `warnings.catch_warnings()`.

There are several follow ups, because GenericBackendV2 still uses deprecated functionality for pulse.


qiskit/providers/models/pulsedefaults.py
```diff
@@ -282,10 +283,12 @@ def from_dict(cls, data):
...
-                if isinstance(value, list):
-                    in_data[key] = list(map(schema[key].from_dict, value))
-                else:
-                    in_data[key] = schema[key].from_dict(value)
+                with warnings.catch_warnings():
+                    warnings.simplefilter("ignore", category=DeprecationWarning)
+                    if isinstance(value, list):
+                        in_data[key] = list(map(schema[key].from_dict, value))
+                    else:
+                        in_data[key] = schema[key].from_dict(value)
```
ElePT: Once `PulseQobjDef` gets removed, do we have plan for the `PulseDefaults` class? For context, `GenericBackendV2` builds an instance of `PulseDefaults`, which is the reason why it's also using the now deprecated `PulseQobjInstruction`. If the plan is to deprecate `PulseDefaults`, does the warning need to be released as the same time as the qobj deprecation?

qiskit/assembler/assemble_circuits.py
```diff
@@ -118,7 +124,7 @@ def _assemble_circuit(
 
     instructions = []
     for op_context in circuit.data:
-        instruction = op_context.operation.assemble()
+        instruction = op_context.operation._assemble()
```
1ucian0, line 129: By calling the private method, the functionality is supported until the removal of `_assemble_circuit`

qiskit/primitives/backend_estimator.py
```diff
@@ -67,14 +68,17 @@ def _run_circuits(
...
-        jobs = [
-            backend.run(circuits[pos : pos + max_circuits], **run_options)
-            for pos in range(0, len(circuits), max_circuits)
-        ]
-        result = [x.result() for x in jobs]
-    else:
-        result = [backend.run(circuits, **run_options).result()]
+    with warnings.catch_warnings():
+        # TODO remove this catch once Aer stops using QobjDictField
+        warnings.simplefilter("ignore", category=DeprecationWarning)
```
ElePT: In these cases where the warning is coming from a module outside of qiskit (or you need some more granularity because you might be suppressing unwanted warnings), it might make sense to filter warnings by module, something like:

```suggestion
        warnings.filterwarnings("ignore", category=DeprecationWarning, module=r"qiskit_aer(\.[a-zA-Z0-9_]+)*")
```

it's something I have tried to do in past deprecations, for example: [here](https://github.com/Qiskit/qiskit/pull/10792/files#diff-7bcb16fad17dea49a12edcf727cdf3c32ef85fec8c97f5cb3b14bdca155611bfR117).

For an even better control, you can also specify a concrete message to filter with, like it's done in the [QiskitTestCase](https://github.com/Qiskit/qiskit/blob/419f40e9e72b4981a6536d08d6b0fa91c13021d5/test/utils/base.py#L234C1-L239C10) class.

ElePT (commented):
I have taken a first look and I left a few suggestions. From my experience with large deprecations, it's always hard to get every single detail right at the first try, but the rule of thumb I think is making sure that the deprecation warnings are raised (which is done in the assertions), and making sure we suppress the least amount of warnings possible, and when there's no other option (I know it's sometimes hard), that the warnings are properly characterized.

qiskit/primitives/backend_estimator.py
1ucian0: Great idea. done in c2f2fdb3de97a6f5a44c8a6ba73442e75f99b106

qiskit/providers/models/pulsedefaults.py
1ucian0: In this context, the problem is `PulseLibraryItem`. I added it in a comment.

doichanj:
I made PR to remove qobj from Aer https://github.com/Qiskit/qiskit-aer/pull/2187
Aer has been deprecated qobj and standalone since 0.14 released 3 month ago, so we can remove them

qiskit/assembler/assemble_circuits.py
```diff
@@ -402,7 +393,54 @@ def assemble_circuits(
...
-    return QasmQobj(
-        qobj_id=qobj_id, config=qobj_config, experiments=experiments, header=qobj_header
-    )
+
+@deprecate_func(
+    since="1.2",
+    removal_timeline="in the 2.0 release",
+    additional_msg="The function assemble_circuits is being deprecated "
+    "as they are not necessary for BackendV2. If user still need Qobj, that probably "
+    "means that they are using a backend based on the deprecated BackendV1 class.",
```
ElePT: I think it's not necessary to mention the function that is being deprecated in the additional message, the decorator already introduces the name in the message, see sample output:

```
DeprecationWarning: The function ``qiskit.compiler.assembler.assemble()`` is deprecated as of qiskit 1.2. It will be removed in the 2.0 release. 
The function assemble is being deprecated as they are not necessary for BackendV2. 
If user still need Qobj, that probably means that they are using a backend based on the deprecated BackendV1 class.
```

So I suggest creating a generic additional message that you can directly copy-paste in all deprecations:

```suggestion
    additional_msg="The `Qobj` class and related functionality are part of the deprecated `BackendV1` "
    "workflow,  and no longer necessary for `BackendV2`. If a user workflow requires `Qobj` it likely relies "
    "on deprecated functionality and should be updated to use `BackendV2`.",
```

qiskit/compiler/assembler.py
```diff
@@ -191,12 +265,16 @@ def assemble(
...
         _log_assembly_time(start_time, end_time)
-        return assemble_circuits(
-            circuits=bound_experiments,
-            qobj_id=qobj_id,
-            qobj_header=qobj_header,
-            run_config=run_config,
-        )
+        with warnings.catch_warnings():
+            # The Qobj is deprecated
+            # warnings.filterwarnings("ignore", category=DeprecationWarning, module="qiskit")
```
ElePT: Any specific reason why the `module` field is not being used here? In any case we should remove the comment:

```suggestion
```

qiskit/compiler/assembler.py
```diff
@@ -191,12 +265,16 @@ def assemble(
...
+            warnings.filterwarnings("ignore", category=DeprecationWarning)
```
ElePT: Maybe: ?
```suggestion
            warnings.filterwarnings("ignore", category=DeprecationWarning, module="qiskit")
```

qiskit/qobj/common.py
```diff
@@ -21,6 +23,13 @@ class QobjDictField(SimpleNamespace):
     previously constructed using marshmallow.
     """
 
+    @deprecate_func(
+        since="1.2",
+        removal_timeline="in the 2.0 release",
+        additional_msg="The full Qobj module is being deprecated, including QobjDictField's subclasses, "
+        "as they are not necessary for BackendV2. If user still need Qobj, that probably "
+        "means that they are using a backend based on the deprecated BackendV1 class.",
+    )
```
ElePT: I think my message suggestion from the beginning applies in this group of classes too.

releasenotes/notes/deprecate_assemble-67486b4d0a8d4f96.yaml
```diff
@@ -0,0 +1,10 @@
+---
+deprecations_providers:
+  - |
+    The structures related to Qobj are now deprecated. Interacting with remote backends should be done via QPY or
+    OpenQASM instead.
```
ElePT: I would appreciate a full list of the classes being deprecated for future reference/being able to search for a specific deprecation in the reno, I can help if it adds to much overhead.

```suggestion
    The `Qobj` structure and related classes are now deprecated, they were introduced as part of the `BackendV1` workflow and are no longer necessary for interacting with  :class:`.BackendV2` backends. Remote backend interaction should be done via `QPY` or `OpenQASM` instead.
```

releasenotes/notes/deprecate_assemble-67486b4d0a8d4f96.yaml
```diff
@@ -0,0 +1,10 @@
...
+deprecations_transpiler:
+  - |
+    The function `assemble` is now deprecated as the Qobj structure is not needed any more. The function was used,
+    among other things, to create a Qobj. It was also used for binding parameter and that functionality is now
+    fully covered by :meth:`~.QuantumCircuit.assign_parameters`.
```
ElePT:
```suggestion
    The `assemble` function is now deprecated and will be removed in the 2.0 release. The function was primarily used to create a `Qobj`, which is no longer necessary in :class:`.BackendV2`-based workflows. It was also used for binding parameters, a functionality fully covered by :meth:`~.QuantumCircuit.assign_parameters`.
```

ElePT (commented):
Thanks for this titanic work @1ucian0, I have gone through the main code and release notes and left a few suggestions (mostly wording). I have also superficially browsed the tests and in principle they look good, but I will have to take a second look.

qiskit/compiler/assembler.py
1ucian0: hmm.. it is actually not needed. Removed in ac54cd8b364cfb5b9a2171897c40077044a2b508

qiskit/assembler/assemble_circuits.py
1ucian0: much better. Thanks! I added this `additional_msg` in all the decorators.

qiskit/providers/models/pulsedefaults.py
```diff
@@ -271,7 +272,7 @@ def from_dict(cls, data):
             PulseDefaults: The PulseDefaults from the input dictionary.
         """
         schema = {
-            "pulse_library": PulseLibraryItem,
+            "pulse_library": PulseLibraryItem,  # The class PulseLibraryItem` is deprecated
```
ElePT:
```suggestion
            "pulse_library": PulseLibraryItem,  # The class PulseLibraryItem is deprecated
```

qiskit/providers/models/pulsedefaults.py
```diff
@@ -282,10 +283,13 @@ def from_dict(cls, data):
         in_data = {}
         for key, value in data.items():
             if key in schema:
-                if isinstance(value, list):
-                    in_data[key] = list(map(schema[key].from_dict, value))
-                else:
-                    in_data[key] = schema[key].from_dict(value)
+                with warnings.catch_warnings():
+                    # The class PulseLibraryItem` is deprecated
```
ElePT:
```suggestion
                    # The class PulseLibraryItem is deprecated
```

test/benchmarks/assembler.py
```diff
@@ -46,4 +46,6 @@ def setup(self, n_qubits, depth, number_of_circuits):
         self.qobj = assemble(self.circuits)
 
     def time_disassemble_circuit(self, _, __, ___):
+        # TODO: QObj is getting deprecated. Remove once that happens
+        # https://github.com/Qiskit/qiskit/pull/12649
         disassemble(self.qobj)
```
ElePT, line 51: benchmarks don't error on deprecation warnings, right? just making sure that not catching the warning here won't break anything.

test/python/circuit/test_parameters.py
```diff
@@ -1018,6 +1021,40 @@ def test_binding_parameterized_circuits_built_in_multiproc(self):
             )
         )
 
+    def test_binding_parameterized_circuits_built_in_multiproc(self):
+        """Verify subcircuits built in a subprocess can still be bound.
+        REMOVE this test once assemble is REMOVED"""
```
ElePT: I think this was a copy-pasting error. This is the new test that won't be removed.
```suggestion
        """Verify subcircuits built in a subprocess can still be bound."""
```

test/python/transpiler/test_sabre_swap.py
```diff
@@ -273,15 +275,21 @@ def test_no_infinite_loop(self, method):
 
         # Assert that the same keys are produced by a simulation - this is a test that the inserted
         # swaps route the qubits correctly.
-        if not optionals.HAS_AER:
-            return
+        with warnings.catch_warnings():
+            # TODO remove Aer stops using Provider Qiskit class
+            warnings.filterwarnings("ignore", category=DeprecationWarning, module="qiskit")
+            if not optionals.HAS_AER:
+                return
```
ElePT: This seems to be caught already by the base class filter, right?
```suggestion
        if not optionals.HAS_AER:
            return
```

test/python/transpiler/test_sabre_swap.py
```diff
@@ -273,15 +275,21 @@ def test_no_infinite_loop(self, method):
...
-        with self.assertWarns(DeprecationWarning):
+        with warnings.catch_warnings():
+            # TODO remove this catch once Aer stops using QobjDictField and Provider ABC
+            # https://github.com/Qiskit/qiskit-aer/pull/2184
+            warnings.filterwarnings("ignore", category=DeprecationWarning, module="qiskit")
             sim = Aer.get_backend("aer_simulator")
-        in_results = sim.run(qc, shots=4096).result().get_counts()
-        out_results = sim.run(routed, shots=4096).result().get_counts()
+            in_results = sim.run(qc, shots=4096).result().get_counts()
+            out_results = sim.run(routed, shots=4096).result().get_counts()
```
ElePT: This one should also be caught by the base class filter, although I am not sure what deprecation the original assertWarns was trying to assert.

test/python/transpiler/test_star_prerouting.py
```diff
@@ -258,8 +260,11 @@ def test_100q_grid_full_path(self, opt_level):
         )
         pm.init += StarPreRouting()
         result = pm.run(qc)
-        counts_before = AerSimulator().run(qc).result().get_counts()
-        counts_after = AerSimulator().run(result).result().get_counts()
+        with warnings.catch_warnings():
+            # TODO remove this catch once Aer stops using qiskit.qobj.common.QobjDictField`
+            warnings.filterwarnings("ignore", category=DeprecationWarning, module="qiskit")
+            counts_before = AerSimulator().run(qc).result().get_counts()
+            counts_after = AerSimulator().run(result).result().get_counts()
```
ElePT: also caught by base filter?

```suggestion
        counts_before = AerSimulator().run(qc).result().get_counts()
        counts_after = AerSimulator().run(result).result().get_counts()
```

test/python/transpiler/test_star_prerouting.py
```diff
@@ -298,8 +303,11 @@ def test_100q_grid_full_path_no_barrier(self, opt_level):
...
```
ElePT:
```suggestion
        counts_before = AerSimulator().run(qc).result().get_counts()
        counts_after = AerSimulator().run(result).result().get_counts()
```

test/python/providers/test_fake_backends.py
```diff
@@ -104,15 +105,18 @@ def setUpClass(cls):
...
+            warnings.filterwarnings("ignore", category=DeprecationWarning, module="qiskit")
+            if not optionals.HAS_AER and backend.num_qubits > 20:
+                self.skipTest(f"Unable to run fake_backend {backend.name} without qiskit-aer")
+            job = backend.run(
+                transpile(
+                    self.circuit, backend, seed_transpiler=42, optimization_level=optimization_level
+                ),
+                seed_simulator=42,
+            )
+            result = job.result()
```
ElePT: This catch should no longer be necessary with the base class filter, right?

test/python/providers/test_fake_backends.py
```diff
@@ -128,21 +132,25 @@ def test_circuit_on_fake_backend(self, backend, optimization_level):
...
+        with warnings.catch_warnings():
+            # TODO remove this catch once Aer stops using QobjDictField
+            warnings.filterwarnings("ignore", category=DeprecationWarning, module="qiskit")
+            job = backend.run(
+                transpile(
+                    self.circuit, backend, seed_transpiler=42, optimization_level=optimization_level
+                ),
+                seed_simulator=42,
+            )
+            result = job.result()
```
ElePT: see above

test/python/providers/fake_provider/test_generic_backend_v2.py
```diff
@@ -136,7 +137,11 @@ def test_run(self):
 
         backend = GenericBackendV2(num_qubits=5, basis_gates=["cx", "id", "rz", "sx", "x"])
         tqc = transpile(qc, backend=backend, optimization_level=3, seed_transpiler=42)
-        result = backend.run(tqc, seed_simulator=42, shots=1000).result()
+        with warnings.catch_warnings():
+            # TODO remove this catch once Aer stops using Provider ABC
+            # https://github.com/Qiskit/qiskit-aer/pull/2184
+            warnings.filterwarnings("ignore", category=DeprecationWarning, module="qiskit")
+            result = backend.run(tqc, seed_simulator=42, shots=1000).result()
```
ElePT:
```suggestion
        result = backend.run(tqc, seed_simulator=42, shots=1000).result()
```

test/python/providers/fake_provider/test_fake_backends.py
```diff
@@ -45,7 +46,10 @@ def test_fake_backends_get_kwargs(self):
         qc.measure_all()
 
         trans_qc = transpile(qc, backend)
-        raw_counts = backend.run(trans_qc, shots=1000).result().get_counts()
+        with warnings.catch_warnings():
+            # TODO remove this catch once Aer stops using QobjDictField
+            warnings.filterwarnings("ignore", category=DeprecationWarning, module="qiskit")
+            raw_counts = backend.run(trans_qc, shots=1000).result().get_counts()
```
ElePT:
```suggestion
        raw_counts = backend.run(trans_qc, shots=1000).result().get_counts()
```

test/python/providers/fake_provider/test_fake_backends.py
```diff
@@ -57,6 +61,10 @@ def test_fake_backend_v2_noise_model_always_present(self):
         qc = QuantumCircuit(1)
         qc.x(0)
         qc.measure_all()
-        res = backend.run(qc, shots=1000).result().get_counts()
+        with warnings.catch_warnings():
+            with warnings.catch_warnings():
+                # TODO remove this catch once Aer stops using QobjDictField
+                warnings.filterwarnings("ignore", category=DeprecationWarning, module="qiskit")
+                res = backend.run(qc, shots=1000).result().get_counts()
```
ElePT: (I think it's no longer necessary and there was a strange nesting when catching warnings)

```suggestion
        res = backend.run(qc, shots=1000).result().get_counts()
```

ElePT (commented):
Thanks again @1ucian0, I think the PR is almost there, just a few more final touches. I hope I am not wrong but I think that all the individual "Aer filter warnings" in the tests are no longer necessary because you added the global warning filter in [2a085b5](https://github.com/Qiskit/qiskit/pull/12649/commits/2a085b5356f693184c5b7ff8cfa5aaeedba4bf25). I tried to point all of the instances out and give suggestions when possible, because the less filters we have in place, the easier it is to clean up afterwards.

test/python/providers/fake_provider/test_fake_backends.py
1ucian0: good catch!

test/python/providers/test_fake_backends.py
1ucian0: done in bd216a17e8e0f3f20b6b5862c0b7c4f20d45ff56

test/python/providers/test_fake_backends.py
1ucian0: done c1c3b32caf4d040b756e3f05b08d743cefb76d5a

test/benchmarks/assembler.py
1ucian0: I think so, they do not subclass the regular TestCase class that raise them to errors.

1ucian0:
You are right, they are not necessary. I think I removed them all. However, CI was flooding so I have to silent them in globally. Hopefully, they can be return to `default` once Aer releases.

ElePT (commented):
I noticed there are 2 tests that didn't get updated (aka removed the Aer warning filter):

- test/python/transpiler/test_sabre_swap.py
- test/python/transpiler/test_star_prerouting.py

If there are no specific reasons to leave those filters in, once they are removed, I am happy to approve :)

1ucian0:
> If there are no specific reasons to leave those filters in, once they are removed, I am happy to approve :)

Nope.. Just overlooked them. Removed in b69bdf96b5663ddb6bb8ffb367398701df515cef

ElePT (approved):
Great, LGTM!! 🚀

PR #12550, opened 2024-06-11
[DAGCircuit Oxidation] Port `DAGCircuit` to Rust

kevinhartman:
### Summary
Ports `DAGCircuit` itself to Rust. The primary benefit to doing this comes from migrating from Rustworkx's `PyDiGraph` PyO3 to a native Rust `petgraph` graph type, which can be accessed directly by Rust-based transpiler passes in the future (much faster). We should also pick up memory savings since our DAG's graph no longer stores `DAGNode` instances, and instead stores a lighter enum `NodeType`, which simply contains a `PackedInstruction`.

#### Status
Currently, the foundational stuff is in place, namely adding and removing bits/wires, registers, etc. along with conversion between `DAGNode` used by the Python API and the Rust struct `NodeType`, used for graph storage.

**Remaining work includes**
- [x] Port remaining `DAGCircuit` methods (I have stubbed any methods not yet implemented using `todo()!`).
- [x] Upgrade Rustworkx version (or point at `main`) to pick up any new `rustworkx-core` functionality needed to support the above.
- [x] Delete `dagcircuit.py` and wire-up the `DAGCircuit` Rust `pyclass` in its place (easy).
- [x] Rebase onto #12459 once it merges, which will involve potentially a lot of changes to use the native `Operation` it introduces.
- [x] Get existing Python testing to pass.
- [x] Clean up struct naming, fix warnings.
- [ ] Performance profiling and tuning.
- [ ] Memory profiling and tuning.

**Open questions**
- [x] Should we port `BitLocations` to Rust and make its management part of the Rust `BitData` struct?

### Details and comments
Internally, the new Rust-based `DAGCircuit` uses a `petgraph::StableGraph` with node weights of type `NodeType` and edge weights of type `Wire`. The `NodeType` enum contains variants for `QubitIn`, `QubitOut`, `ClbitIn`, `ClbitOut`, and `Operation`, which should save us from all of the `isinstance` checking previously needed when working with `DAGNode` Python instances. The `Wire` enum contains variants `Qubit` and `Clbit`.

API access from Python-space is (still) done in terms of `DAGNode` instances to remain API compatible with the Python implementation. However, internally, we convert to and deal in terms of `NodeType`. When the user requests a particular node via lookup or iteration, we inflate an ephemeral `DAGNode` based on the internal `NodeType` and give them that. This is very similar to what we've done in #10827 when porting `CircuitData` to Rust.

#### For the future
[...]

PR #12650, opened 2024-06-24
Use rust gates for Optimize1QGatesDecomposition

mtreinish:
### Summary

This commit moves to using rust gates for the Optimize1QGatesDecomposition transpiler pass. It takes in a sequence of runs (which are a list of DAGOpNodes) from the python side of the transpiler pass which are generated from DAGCircuit.collect_1q_runs() (which in the future should be moved to rust after #12550 merges). The rust portion of the pass now iterates over each run, performs the matrix multiplication to compute the unitary of the run, then synthesizes that unitary, computes the estimated error of the circuit synthesis and returns a tuple of the circuit sequence in terms of rust StandardGate enums. The python portion of the code then takes those sequences and does inplace substitution of each run with the sequence returned from rust.

Once #12550 merges we should be able to move the input collect_1q_runs() call and perform the output node substitions in rust making the full pass execute in the rust domain without any python interaction.

Additionally, the OneQubitEulerDecomposer class is updated to use rust for circuit generation instead of doing this python side. The internal changes done to use rust gates in the transpiler pass meant we were half way to this already by emitting rust StandardGates instead of python gate objects. The dag handling is still done in Python however until #12550 merges.

~This also includes an implementation of the r gate, I temporarily added this to unblock this effort as it was the only gate missing needed to complete this. We can rebase this if a standalone implementation of the gate merges before this.~ Rebased on top of: https://github.com/Qiskit/qiskit/pull/12662

### Details and comments

TODO:

- [x] Investigate multithreading across runs. This might not make sense until #12550 merges so we can do a parallel bridge iterator over runs and avoid having to collect runs generated in python


qiskit/transpiler/passes/optimization/optimize_1q_decomposition.py
```diff
@@ -198,32 +205,47 @@ def run(self, dag):
...
-                    node._node_id = dag._multi_graph.add_node(node)
-                    dag._increment_op(op)
-                    dag._multi_graph.insert_node_on_in_edges(node._node_id, run[0]._node_id)
-                dag.global_phase += best_circuit_sequence.global_phase
-                # Delete the other nodes in the run
-                for current_node in run:
-                    dag.remove_op_node(current_node)
+            runs.append(run)
+            qubits.append(qubit)
+            bases.append(_possible_decomposers(basis))
```
mtreinish: I need to add back the caching mechanism here so we don't need to recompute this for every qubit.

crates/accelerate/src/euler_one_qubit_decomposer.rs
```diff
@@ -883,6 +999,115 @@ pub fn params_zxz(unitary: PyReadonlyArray2<Complex64>) -> [f64; 4] {
...
+    runs: Vec<Vec<PyRef<DAGOpNode>>>,
+    qubits: Vec<usize>,
+    bases: Vec<Vec<PyBackedStr>>,
+    simplify: bool,
+    error_map: Option<&OneQubitGateErrorMap>,
+    atol: Option<f64>,
+) -> Vec<OptimizeDecompositionReturn> {
+    // The array collection is done serially because we need to use a PyRef to get a reference
+    // input to the DAGOpNode (as a side effect of being a subclass in rust space). This prohibits
+    // using paralleism because PyRef is neither sync or send.
```
mtreinish:
```suggestion
```

mtreinish:
I took a quick attempt at implementing this with multithreading locally and I could get it to work but it required doing extra collection of all the data from the collect runs into a second vec to strip the `PyRef` from each node. This added more overhead than parallelism saved in this case. I think it should be possible to investigate doing this in parallel after #12550 merges because then we can generate the runs from rust directly. For multithreading we'll likely still need to collect the runs into a vec (to avoid mutation while iterating) but we can measure if that overhead is higher than the speedup from doing the synthesis in parallel and investigate in a followup.

mtreinish:
I threw together a quick benchmark to test this:

```python
import statistics
import time

from qiskit.transpiler.passes import Optimize1qGatesDecomposition

seed = 42
n_qubits = 20
depth = 102400

qc = random_circuit(n_qubits, depth, measure=True, seed=seed)
dag = circuit_to_dag(qc)
basis_gates = ['sx', 'x', 'rz', 'cx', 'id']
opt_pass = Optimize1qGatesDecomposition(basis=basis_gates)
times = []
for i in range(10):
    start = time.perf_counter()
    opt_pass.run(dag)
    stop = time.perf_counter()
    times.append(stop - start)
print(f"Mean runtime: {statistics.geometric_mean(times)}")
```

and it's showing a pretty noticeable speedup over Qiskit 1.1. With this PR I got:

```
3.931521232593536 sec
```

and with Qiskit 1.1 I got:

```
6.184648169533495 sec
```

Looking at a profile with this script the majority of the time with this PR is spent in collect 1q runs and dag interactions in python.

mtreinish:
To compare this against the current state of main I ran asv for the pass specific benchmarks:

```

| Change   | Before [1ed5951a] <optimize-1q-more-rust~3^2>   | After [130969a3] <optimize-1q-more-rust>   |   Ratio | Benchmark (Parameter)                                                                                         |
|----------|-------------------------------------------------|--------------------------------------------|---------|---------------------------------------------------------------------------------------------------------------|
| -        | 134±1ms                                         | 46.9±0.4ms                                 |    0.35 | passes.MultipleBasisPassBenchmarks.time_optimize_1q_decompose(20, 1024, ['rz', 'x', 'sx', 'cx', 'id'])        |
| -        | 99.3±1ms                                        | 34.0±0.2ms                                 |    0.34 | passes.MultipleBasisPassBenchmarks.time_optimize_1q_decompose(14, 1024, ['rz', 'x', 'sx', 'cx', 'id'])        |
| -        | 116±0.7ms                                       | 39.0±0.5ms                                 |    0.34 | passes.MultipleBasisPassBenchmarks.time_optimize_1q_decompose(20, 1024, ['u', 'cx', 'id'])                    |
| -        | 85.6±0.3ms                                      | 27.9±0.1ms                                 |    0.33 | passes.MultipleBasisPassBenchmarks.time_optimize_1q_decompose(14, 1024, ['u', 'cx', 'id'])                    |
| -        | 123±1ms                                         | 41.1±0.5ms                                 |    0.33 | passes.MultipleBasisPassBenchmarks.time_optimize_1q_decompose(20, 1024, ['rx', 'ry', 'rz', 'r', 'rxx', 'id']) |
| -        | 92.0±0.5ms                                      | 29.6±0.2ms                                 |    0.32 | passes.MultipleBasisPassBenchmarks.time_optimize_1q_decompose(14, 1024, ['rx', 'ry', 'rz', 'r', 'rxx', 'id']) |
| -        | 47.4±0.5ms                                      | 12.6±0.04ms                                |    0.27 | passes.MultipleBasisPassBenchmarks.time_optimize_1q_decompose(5, 1024, ['rz', 'x', 'sx', 'cx', 'id'])         |
| -        | 43.6±0.3ms                                      | 10.6±0.05ms                                |    0.24 | passes.MultipleBasisPassBenchmarks.time_optimize_1q_decompose(5, 1024, ['rx', 'ry', 'rz', 'r', 'rxx', 'id'])  |
| -        | 40.7±0.4ms                                      | 9.31±0.06ms                                |    0.23 | passes.MultipleBasisPassBenchmarks.time_optimize_1q_decompose(5, 1024, ['u', 'cx', 'id'])                     |

| Change   | Before [1ed5951a] <optimize-1q-more-rust~3^2>   | After [130969a3] <optimize-1q-more-rust>   |   Ratio | Benchmark (Parameter)                                                                                           |
|----------|-------------------------------------------------|--------------------------------------------|---------|-----------------------------------------------------------------------------------------------------------------|
|          | 2.65±0.01s                                      | 2.66±0.02s                                 |    1    | passes.PassBenchmarks.time_optimize_swap_before_measure(20, 1024)                                               |
|          | 231±1ms                                         | 230±1ms                                    |    1    | passes.PassBenchmarks.time_optimize_swap_before_measure(5, 1024)                                                |
|          | 737±10ms                                        | 719±6ms                                    |    0.98 | passes.MultipleBasisPassBenchmarks.time_optimize_1q_commutation(14, 1024, ['rx', 'ry', 'rz', 'r', 'rxx', 'id']) |
|          | 1.53±0.01s                                      | 1.51±0s                                    |    0.98 | passes.PassBenchmarks.time_optimize_swap_before_measure(14, 1024)                                               |
|          | 1.14±0.01s                                      | 1.11±0s                                    |    0.97 | passes.MultipleBasisPassBenchmarks.time_optimize_1q_commutation(20, 1024, ['rx', 'ry', 'rz', 'r', 'rxx', 'id']) |
|          | 256±2ms                                         | 247±1ms                                    |    0.97 | passes.MultipleBasisPassBenchmarks.time_optimize_1q_commutation(5, 1024, ['rx', 'ry', 'rz', 'r', 'rxx', 'id'])  |
|          | 822±10ms                                        | 778±3ms                                    |    0.95 | passes.MultipleBasisPassBenchmarks.time_optimize_1q_commutation(14, 1024, ['rz', 'x', 'sx', 'cx', 'id'])        |
|          | 746±7ms                                         | 706±2ms                                    |    0.95 | passes.MultipleBasisPassBenchmarks.time_optimize_1q_commutation(14, 1024, ['u', 'cx', 'id'])                    |
|          | 1.26±0.01s                                      | 1.20±0s                                    |    0.95 | passes.MultipleBasisPassBenchmarks.time_optimize_1q_commutation(20, 1024, ['rz', 'x', 'sx', 'cx', 'id'])        |
|          | 1.16±0.02s                                      | 1.10±0s                                    |    0.95 | passes.MultipleBasisPassBenchmarks.time_optimize_1q_commutation(20, 1024, ['u', 'cx', 'id'])                    |
|          | 298±2ms                                         | 277±1ms                                    |    0.93 | passes.MultipleBasisPassBenchmarks.time_optimize_1q_commutation(5, 1024, ['rz', 'x', 'sx', 'cx', 'id'])         |
|          | 260±2ms                                         | 239±1ms                                    |    0.92 | passes.MultipleBasisPassBenchmarks.time_optimize_1q_commutation(5, 1024, ['u', 'cx', 'id'])                     |

PERFORMANCE INCREASED.
```

crates/accelerate/src/euler_one_qubit_decomposer.rs
```diff
@@ -672,6 +715,28 @@ fn compare_error_fn(
...
+    match error_map {
+        Some(err_map) => {
+            let num_gates = gates.len();
+            let gate_fidelities: f64 = gates
+                .iter()
+                .map(|x| 1. - err_map.error_map[qubit].get(x.0.name()).unwrap_or(&0.))
+                .product();
+            (1. - gate_fidelities, num_gates)
+        }
+        None => (gates.len() as f64, gates.len()),
```
ElePT, line 701: why is `gates.len()` returned twice here? I would expect this output to be something like `None => (None, gates.len()),`

crates/accelerate/src/euler_one_qubit_decomposer.rs
```diff
@@ -755,6 +829,48 @@ pub fn unitary_to_gate_sequence_inner(
...
+    unitary: PyReadonlyArray2<Complex64>,
+    target_basis_list: Vec<PyBackedStr>,
+    qubit: usize,
+    error_map: Option<&OneQubitGateErrorMap>,
+    simplify: bool,
+    atol: Option<f64>,
+) -> PyResult<Option<CircuitData>> {
+    let mut target_basis_vec: Vec<EulerBasis> = Vec::with_capacity(target_basis_list.len());
+    for basis in target_basis_list {
+        let basis_enum = EulerBasis::__new__(basis.deref())?;
```
ElePT: Is there any difference between `basis.deref()` and `*basis`?

ElePT (commented):
I tried taking a look at the implementation, I didn't see anything that stood out as wrong but I'm sure I wasn't able to catch every detail.

crates/accelerate/src/euler_one_qubit_decomposer.rs
mtreinish: This is for compatibility with the python side scoring. The scoring function in python are designed to do sorting with a tuple of `(error_rate, len)` and in the case without error rates we just use `(len, len)`. I just retained this from what the python side was doing before.

crates/accelerate/src/euler_one_qubit_decomposer.rs
mtreinish: Using `*basis` would end up in a copy I think but this does it all by reference. Although I'm not sure it would work because basis is a `PyBackedStr` so I don't think we can use `*basis`. `PyBackedStr.deref()`  returns a `&str` which is what we need in this case.

mtreinish:
We should merge this after #12662 merges since this depends on having an RGate implementation in rust. I did it inline here but it would be good to review that in isolation, I'll rebase this after that merges.

crates/accelerate/src/euler_one_qubit_decomposer.rs
```diff
@@ -633,7 +677,7 @@ fn compare_error_fn(
             let fidelity_product: f64 = circuit
                 .gates
                 .iter()
-                .map(|x| 1. - err_map.get(&x.0).unwrap_or(&0.))
+                .map(|x| 1. - err_map.get(x.0.name()).unwrap_or(&0.))
```
jlapeyre: Very minor, but would save looking upward a few lines for context. I tried with rustfmt; no change in linewrap!
```suggestion
                .map(|gate| 1. - err_map.get(gate.0.name()).unwrap_or(&0.))
```

jlapeyre: Have we considered using a non-string key? (Or some kind of [optimized string](https://swatinem.de/blog/optimized-strings/))

crates/accelerate/src/euler_one_qubit_decomposer.rs
```diff
@@ -642,6 +686,28 @@ fn compare_error_fn(
...
+    gates: &[(StandardGate, SmallVec<[f64; 3]>)],
+    error_map: Option<&OneQubitGateErrorMap>,
+    qubit: usize,
+) -> (f64, usize) {
+    match error_map {
+        Some(err_map) => {
+            let num_gates = gates.len();
+            let gate_fidelities: f64 = gates
+                .iter()
+                .map(|x| 1. - err_map.error_map[qubit].get(x.0.name()).unwrap_or(&0.))
```
jlapeyre:
```suggestion
                .map(|gate| 1. - err_map.error_map[qubit].get(gate.0.name()).unwrap_or(&0.))
```
Also no fmt change.

crates/accelerate/src/euler_one_qubit_decomposer.rs
```diff
@@ -651,7 +717,7 @@ fn compute_error(
             let num_gates = gates.len();
             let gate_fidelities: f64 = gates
                 .iter()
-                .map(|x| 1. - err_map.error_map[qubit].get(&x.0).unwrap_or(&0.))
+                .map(|x| compute_error_term(x.0.as_str(), err_map, qubit))
```
jlapeyre:
```suggestion
                .map(|gate| compute_error_term(gate.0.as_str(), err_map, qubit))
```
No change with `rustfmt`

crates/accelerate/src/euler_one_qubit_decomposer.rs
mtreinish: Not really, these strings are lookup keys from the `Target`, the error map here is just a rust view of the error data in the target that we construct in Python to provide the data to this pass (when we have #12292 merged we can drop this structure and query it directly). We need something that interoperates with a Python string which is realistically just `String` or `&str`.

jlapeyre: I assumed that, at least for now, we need regular strings for Python interop.

crates/accelerate/src/euler_one_qubit_decomposer.rs
```diff
@@ -725,6 +800,48 @@ pub fn unitary_to_gate_sequence_inner(
...
+) -> PyResult<Option<CircuitData>> {
+    let mut target_basis_vec: Vec<EulerBasis> = Vec::with_capacity(target_basis_list.len());
+    for basis in target_basis_list {
+        let basis_enum = EulerBasis::__new__(basis.deref())?;
+        target_basis_vec.push(basis_enum)
+    }
+    let unitary_mat = unitary.as_array();
+    let circuit_sequence = unitary_to_gate_sequence_inner(
+        unitary_mat,
+        &target_basis_vec,
```
jlapeyre: EDIT: This should have the same behavior.
```suggestion
    let target_basis_vec: PyResult<Vec<EulerBasis>> = target_basis_list
        .iter()
        .map(|basis| EulerBasis::__new__(basis))
        .collect();
    Ok(unitary_to_gate_sequence_inner(
        unitary.as_array(),
        &target_basis_vec?,
```

crates/accelerate/src/euler_one_qubit_decomposer.rs
```diff
@@ -853,6 +970,112 @@ pub fn params_zxz(unitary: PyReadonlyArray2<Complex64>) -> [f64; 4] {
...
+                (1. - error, raw_run.len())
+            } else {
+                (error, raw_run.len())
+            };
+            let target_basis_list = &bases[index];
+            let mut target_basis_vec: Vec<EulerBasis> = Vec::with_capacity(target_basis_list.len());
+            for basis in target_basis_list {
+                let basis_enum = EulerBasis::__new__(basis.deref()).unwrap();
+                target_basis_vec.push(basis_enum)
+            }
```
jlapeyre:
```suggestion
            let target_basis_vec: Vec<EulerBasis> = bases[index]
                .iter()
                .map(|basis| EulerBasis::__new__(basis).unwrap())
                .collect();
```

crates/accelerate/src/euler_one_qubit_decomposer.rs
jlapeyre: [Here is a diff](https://github.com/mtreinish/qiskit-core/compare/optimize-1q-more-rust...jlapeyre:qiskit-core:optimize-1q-more-rust-gjl-change) showing  everywhere the pattern occurs

crates/accelerate/src/euler_one_qubit_decomposer.rs
jlapeyre: I might be mistaken, but the docs suggest that `PyBackedStr` is meant to be used with deref coercion. You can use `__new__(basis)`.  [For example.](https://github.com/mtreinish/qiskit-core/compare/optimize-1q-more-rust...jlapeyre:qiskit-core:optimize-1q-more-rust-gjl-change)

jlapeyre:
In the code in https://github.com/Qiskit/qiskit/pull/12650#issuecomment-2189462949 inserting nonsense for `basis_gates`, eg `basis_gates = ['NONSENSE', 'x', 'rz', 'cx', 'id']` throws no error. The code runs to completion silently (I made the sizes and reps small so it runs faster)

Of course, the same thing happens on the `main` branch. Maybe there is a good reason.

crates/accelerate/src/euler_one_qubit_decomposer.rs
mtreinish: I applied this to the branch: https://github.com/Qiskit/qiskit/pull/12650/commits/68af440ca1ad08d3a0551c7e00013eddb29b2434

crates/accelerate/src/euler_one_qubit_decomposer.rs
mtreinish: I applied this suggestion to the branch in: https://github.com/Qiskit/qiskit/pull/12650/commits/68af440ca1ad08d3a0551c7e00013eddb29b2434

mtreinish:
> In the code in https://github.com/Qiskit/qiskit/pull/12650#issuecomment-2189462949 inserting nonsense for basis_gates, eg basis_gates = ['NONSENSE', 'x', 'rz', 'cx', 'id'] throws no error. The code runs to completion silently (I made the sizes and reps small so it runs faster)
>
> Of course, the same thing happens on the main branch. Maybe there is a good reason.

That's expected behavior. The target euler basis is selected as a subset of the provided basis gate names. So it looks for which gate names it knows about and figures out which are valid decomposition bases to use. So in this context `cx` is treated the same way as `NONSENSE`. If we were using a target we'd have to associate `NONSENSE` with an actual python gate definition. But  this pass would still work the same way because the decomposition is only defined for a fixed set of target gates. You can see the table here: https://docs.quantum.ibm.com/api/qiskit/qiskit.synthesis.OneQubitEulerDecomposer

crates/accelerate/src/euler_one_qubit_decomposer.rs
```diff
@@ -853,6 +965,110 @@ pub fn params_zxz(unitary: PyReadonlyArray2<Complex64>) -> [f64; 4] {
...
+            )
+            .map(|out_seq| {
+                let new_error = compute_error_one_qubit_sequence(&out_seq, qubit, error_map);
+                (old_error, new_error, out_seq)
+            })
+        })
+        .collect()
+}
+
+fn matmul_1q(operator: &mut [[Complex64; 2]; 2], other: Array2<Complex64>) {
```
jlapeyre, line 1034: EDIT: No not faster.
I tried this, thinking it might be faster:
```rust
fn matmul_1q(operator: &mut [[Complex64; 2]; 2], other: Array2<Complex64>) {
    let (a00, a01, a10, a11) = (operator[0][0], operator[0][1], operator[1][0], operator[1][1]);
    operator[0][0] =  other[[0, 0]] * a00 + other[[0, 1]] * a10;
    operator[0][1] =  other[[0, 0]] * a01 + other[[0, 1]] * a11;
    operator[1][0] =  other[[1, 0]] * a00 + other[[1, 1]] * a10;
    operator[1][1] =  other[[1, 0]] * a01 + other[[1, 1]] * a11;
}
```
Perhaps this avoids an allocation.
But performance of this code https://github.com/Qiskit/qiskit/pull/12650#issuecomment-2189462949 did not appear to change.

crates/accelerate/src/euler_one_qubit_decomposer.rs
```diff
@@ -853,6 +965,110 @@ pub fn params_zxz(unitary: PyReadonlyArray2<Complex64>) -> [f64; 4] {
...
+                .iter()
+                .map(|node| {
+                    if let Some(err_map) = error_map {
+                        error *=
+                            compute_error_term(node.instruction.operation.name(), err_map, qubit)
+                    }
+                    node.instruction
+                        .operation
+                        .matrix(&node.instruction.params)
+                        .unwrap()
```
jlapeyre: There may be a better message, but something like:
```suggestion
                        .expect("No matrix defined for operation")
```
My understanding is `expect` will only panic here if there is a programming error, not a user error; and that this is a legitimate use of `expect`. Not great error handling, but a step up from `unwrap`.

If we follow the rust doc style guidelines, the message should be something like `"Operation must have a matrix definition"`. IDK which is preferrable.

We could also do `unwrap_or_else(|| panic!("{}", format!("No matrix defined for operation {}", node.instruction.operation.name()` (untested)

mtreinish: Sure, I can add an `expect()` call here. It shouldn't be possible to reach this code and not have a matrix defined since the input nodes to this function have passed through this filter function: https://github.com/Qiskit/qiskit/pull/12650/files#diff-61c072175672ef132e5e682a51401abd08288938d864b2090f4dc446e772837eR1048-R1069 (which checks that `matrix()` `is_some()`)

crates/accelerate/src/euler_one_qubit_decomposer.rs
mtreinish: I'm not sure where this would save an allocation this is just destructuring the array reference to individual variables. I would expect the compiler to generate the same exact code for this as the code in this function already.

The one allocation point I'm not happy about but I couldn't really figure out how to avoid was around the `Array2` for `other` which we're getting from `Operation::matrix()`. I needed to do that from a typing perspective in the `Operation` trait, because using an array view wouldn't work since we may get the array view from Python or it might point to a static array so I couldn't see a way to declare a lifetime parameter for the trait method that would work (or at least I think that's why it returns an owned array, it's been >1 month since I wrote that class so I'm not 100%).

jlapeyre: Another thing. We can test if all gates are standard, and if so do some optimizations. Like use fixed sized `Array`s throughout the multiplication. Doesn't need to be in this PR of course.

mtreinish: Yeah, let's look at that kind of thing in a followup. Right now the runtime overhead is mostly dominated by the rust python boundary and I think we'll have another round or two of optimizations to perform on this pass after #12550 merges. We also should have more flexibility when everything is handled in rust.

jlapeyre: > just destructuring the array reference to individual variables. 

I see that now. In fact, if there were an allocation, there is no way the caller can know where it is.

jlapeyre: > avoid was around the `Array2` for other which we're getting from `Operation::matrix()`

This is the optimization I was talking about above. I didn't think it through all the way. But if you could scan or use a flag to know that all matrices are slices of fixed size, then multiplication would be much faster. They are all 2x2 of course, but `Array2` doesnt carry the info in the type.
But i wouldnt want to slow this PR down for this optimization.

crates/accelerate/src/euler_one_qubit_decomposer.rs
jlapeyre: Right. It's meant to signal somehow that this is unreachable unless something is broken, ie a programming error creeps in. People debate over `expect`, `unreachable!`, etc. I like something better than `unwrap` anyway.  You could ask, is the programmer not handling an error that might arise with bad input? In part this is meant to signal to the reader that this will only fail (we believe) if someone breaks code elsewhere.

We could even write "you have discovered a bug in qiskit", etc. But I think something minimally visually distracting is good.

jlapeyre:
Another note for later. This function
https://github.com/Qiskit/qiskit/blob/3fb175390d5fa007150fb76a23b914059010788a/crates/accelerate/src/euler_one_qubit_decomposer.rs#L386-L386
has return type `PyResult<OneQubitGateSequence>`. But as far as I can tell it is only called in one place, and that is in the same file. The returned value is `unwrap`ped with no check. In fact the return value is always `ok`. This is so the function can be called from Python. (maybe in the future, since I don't see this called anywhere) Reading and understanding the `unwrap` takes a bit of time.

I'd prefer that the Python-side function (if needed) wraps a function that returns `OneQubitGateSequence` rather than having this special requirement of the Python interface intrude in code that is only used in this function.

This is a quick fix. But probably outside the scope of this PR.

crates/accelerate/src/euler_one_qubit_decomposer.rs
```diff
@@ -853,6 +965,110 @@ pub fn params_zxz(unitary: PyReadonlyArray2<Complex64>) -> [f64; 4] {
...
+                    .matrix(&node.instruction.params)
+                    .is_some()
+                && (node.instruction.extra_attrs.is_none()
+                    || node
+                        .instruction
+                        .extra_attrs
+                        .as_ref()
+                        .unwrap()
+                        .condition
+                        .is_none())
```
jlapeyre:
```suggestion
                && match &node.instruction.extra_attrs {
                    None => true,
                    Some(attrs) => attrs.condition.is_none(),
                }
```

jlapeyre:
I left one more suggested simplification edit.

Other than that, this looks good to go.

jlapeyre (approved):

PR #12651, opened 2024-06-25
Add some abbreviations for gates and linear algebra

jlapeyre:
Add constant abbreviations for some values and types.

This PR introduces some abbreviations for repetitive Rust code.  Motivations are reducing clutter, improving readability, and perhaps modest support for rapid development.

#### Construction of complex numbers

Complex numbers are common in scientific computing and ubiquitous in quantum computation. They are supported in Rust via libraries, one in particular, with few ergonomic features beyond those provided for any `struct`. We do have `Complex64` for `Complex<f64>` so that numbers are constructed with
```rust
Complex64::new(a, b)
```

In other scientific languages and platforms, developers typically appreciate ergonomic features[^1] reflecting special status of complex numbers. Improving ergonomic for complex numbers in Qiskit is a worthwhile task.

In this PR, I chose to continue using the definition of `const fn 64` that was introduced in #12459 https://github.com/Qiskit/qiskit/blob/1ed5951a98b594808525c8428e06178c160cfcbb/crates/circuit/src/gate_matrix.rs#L19-L22.  I have moved the definition to a more neutral location and used it uniformly in all crates.

I chose this in part to try to make this appealing to Qiskit devs. However there are choices.
1. The current `c64` (referenced above). Can be used in `static` and `const` constructs. Requires `f64` arguments.
2.  [`c64`](https://docs.rs/num-complex/latest/num_complex/fn.c64.html) introduced recently in `num-complex`. This creates a new `Complex<f64>` from arguments that can convert `Into<f64>`. For example `c64(1.0, 2.0)` and `c64(1, 2)`. `static` and `const` not supported, which is a major disadvantage for Qiskit.
3. I prefer defining `c64` slightly more generally so that `c64(1, 2.0)` is allowed (The version in item 2. does not support this.). This is a minor tweak to the implementation. However, `static` and `const` are still not supported.
4. A macro `c64!` is the most capable and flexible. This is functionally nearly the same as item 3. with the major advantage that `static` and `const` *are* supported.

One might argue that having `c64(1, 2)` (or `c64!(1, 2)`) return a `Complex64` goes against the spirit of Rust. But I find decimal points that exist only to satisfy language semantics irksome and distracting. For rapid development, allowing more flexibility within constraints of semantics is useful.  It would be better that `c64(1, 2)` construct a complex number with integer types in combination with facilities mix complex types. But Rust in general is too explicit for this.

However, examining this PR, you can see that the number of superfluous decimal points is actually rather small at least so far. So retaining choice `1` is not a big burden.

#### Constants for common complex and `f64` values

These are `const` bindings for very common complex values, such as `C_ONE`, `C_ZERO` etc. I think just `ONE` and `ZERO` are just as good, a bit less explicit, but neater and more readable (@sbrandhsn). These make source more readable and less cluttered, and likely less error prone. A couple of common `f64` values that are defined repeatedly locally have been moved to a common location.  (@ShellyGarion @alexanderivrii)

#### Type definitions for arrays representing gates

For example `GateArray1Q = [[Complex64; 2]; 2];`.  Advantages are reducing clutter, signalling intent, etc.

One could object to this and the previous item. These are less explicit and one might mistakenly assume that an identifier is bound to a different value. But these are generic objections to binding an identifier to a value. The statements of the objections are always true. Yet we find these refactorings sometimes worthwhile.

#### Relation to other PRs

* This PR is replaces part of  #12507

The other part of #12507, which implements a gate, will be included in a separate PR.

[^1]: In Python the suffix `j` is recognized by the parser, so `2j` represents the complex value `(0.0, 2.0)`.
    
    In Julia, `Complex{T<:Real}` is a parametric type and the constant `im` is bound to the value `Complex{Bool}(false, true)`. The parser reads a numeric literal followed by an identifier as a call to the function `*`. Methods are defined for arithmetic with mixed types, so that `2.0im` dispatches to `*(x::Real, z::Complex{Bool}) = Complex(x * real(z), x * imag(z))`.

    Technical computing is not a core domain of C. However concessions to ergonomics exist. In C99 a complex number is typically constructed like this: `1.0 + 2.0*I`. In C11, `CMPLX(1.0, 2.0)`.

    Fortran has special support for complex numbers, since it's main application domain is scientific/technical computing. For example `z%re = 3.0` assigns the real part of `z`.

### DIFF
--- a/crates/accelerate/src/convert_2q_block_matrix.rs
+++ b/crates/accelerate/src/convert_2q_block_matrix.rs
@@ -20,10 +20,7 @@ use numpy::ndarray::{aview2, Array2, ArrayView2};
 use numpy::{IntoPyArray, PyArray2, PyReadonlyArray2};
 use smallvec::SmallVec;
 
-static ONE_QUBIT_IDENTITY: [[Complex64; 2]; 2] = [
-    [Complex64::new(1., 0.), Complex64::new(0., 0.)],
-    [Complex64::new(0., 0.), Complex64::new(1., 0.)],
-];
+use qiskit_circuit::gate_matrix::ONE_QUBIT_IDENTITY;
 
 /// Return the matrix Operator resulting from a block of Instructions.
 #[pyfunction]
--- a/crates/accelerate/src/euler_one_qubit_decomposer.rs
+++ b/crates/accelerate/src/euler_one_qubit_decomposer.rs
@@ -31,6 +31,7 @@ use ndarray::prelude::*;
 use numpy::PyReadonlyArray2;
 use pyo3::pybacked::PyBackedStr;
 
+use qiskit_circuit::util::c64;
 use qiskit_circuit::SliceOrInt;
 
 pub const ANGLE_ZERO_EPSILON: f64 = 1e-12;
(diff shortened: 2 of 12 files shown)


jlapeyre:
In an out of band discussion we agreed to reduce use of the constants `PI2` and `PI4`. In particular, the commit above that makes them `pub const` has been overwritten to remove these in b6b315e .

The commit message for b6b315e is written as if the `pub const` definitions never existed. So the message can be preserved as is in a squash commit. But the message for the commit introducing `pub const PI2` etc. should be removed when squashing.

jakelishman (approved):
This looks fine to me and inline with what we discussed.  If you want to squash-rebase / change your commit messages or anything, feel free to and I'll re-approve.  If you don't mind, just tag it for merge.

PR #12653, opened 2024-06-25
GenericBackendV2 should fail when the backend cannot allocate the basis gate because its size

1ucian0:
### Summary

GenericBackendV2 should not accept backends that are too small for certain gates. The following examples should be rejected:

`GenericBackendV2(num_qubits=1, basis_gates=["cx", "id"])`

`GenericBackendV2(num_qubits=2, basis_gates=["ccx", "id"])`

@ElePT found this!

### DIFF
--- a/releasenotes/notes/fixes_GenericBackendV2-668e40596e1f070d.yaml
+++ b/releasenotes/notes/fixes_GenericBackendV2-668e40596e1f070d.yaml
@@ -1,4 +1,4 @@
 ---
 fixes:
   - |
-    The constructor :class:`.GenericBackendV2` was allowing to create malformed backends because accepted backend sizes that cannot allocate basis gates. That is, a backend with a single qubit should not accept a basis with two-qubit gates.
+    The constructor :class:`.GenericBackendV2` was allowing to create malformed backends because it accepted basis gates that couldn't be allocated in the backend size . That is, a backend with a single qubit should not accept a basis with two-qubit gates.
--- a/releasenotes/notes/fixes_GenericBackendV2-668e40596e1f070d.yaml
+++ b/releasenotes/notes/fixes_GenericBackendV2-668e40596e1f070d.yaml
@@ -1,0 +1,4 @@
+---
+fixes:
+  - |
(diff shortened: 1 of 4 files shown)


releasenotes/notes/fixes_GenericBackendV2-668e40596e1f070d.yaml
```diff
@@ -0,0 +1,4 @@
+---
+fixes:
+  - |
+    The constructor :class:`.GenericBackendV2` was allowing to create malformed backends because accepted backend sizes that cannot allocate basis gates. That is, a backend with a single qubit should not accept a basis with two-qubit gates.
```
ElePT:
```suggestion
  - |
    The constructor :class:`.GenericBackendV2` was allowing to create malformed backends because it accepted basis gates that couldn't be allocated in the backend size . That is, a backend with a single qubit should not accept a basis with two-qubit gates.
```

ElePT (dismissed):
thanks @1ucian0 LGTM, I left a small reno suggestion (am I allowed to approve?)

ElePT (dismissed):

releasenotes/notes/fixes_GenericBackendV2-668e40596e1f070d.yaml
1ucian0: nice one, much better.

1ucian0:
> am I allowed to approve?

It seems you are.

ElePT:
I think we need to add an internal path for building a target for 1 qubit, because `GenericBackendV2` contains 2q basis gates by default (that's why the test fails currently).

1ucian0:
> I think we need to add an internal path for building a target for 1 qubit, because `GenericBackendV2` contains 2q basis gates by default (that's why the test fails currently).

I think I prefer to keep the default simple (without different values depending on the number of qubits) and be explicit on the `basis _gate` (see https://github.com/Qiskit/qiskit/pull/12653/commits/43a94f1d79f72c482a38820ddef63df85836a0d0). What do you think?

ElePT:
I think that can work too. If we see this is causing problems we can always correct the 1q path.

ElePT (approved):

1ucian0:

PR #12628, opened 2024-06-21
Enable the new efficient MCX decompose

t-imamichi:
### Summary

This PR enables #11993 as `QuantumCircuit.mcx`.
We can reduce the number of CNOT gates as follows.

```python
from qiskit import QuantumCircuit

def bench_mcx(n: int):
    qc = QuantumCircuit(n)
    qc.mcx(list(range(n - 1)), n - 1)
    #print(qc.decompose(reps=1))
    print(n, qc.decompose(reps=4).count_ops()["cx"])

for i in range(6, 10):
    bench_mcx(i)
```

main
```
6 92
7 188
8 380
9 764
```

this PR
```
6 84
7 140
8 220
9 324
```

PR #12659, opened 2024-06-25
Add Rust representation for most controlled gates

ElePT:
### Summary
This PR adds Rust implementations for most controlled gates left in https://github.com/Qiskit/qiskit/issues/12566.

- `CUGate`
- `CU1Gate`
- `CU3Gate`
- `C3XGate`
- `C3SXGate`
- `CCZGate`
- `RCCXGate`
- `RC3XGate`

This is done through a newly implemented `make_n_controlled_gate` macro (which is currently limited to adding controls to a single-qubit gate).

### Details and comments

I attempted to implement C4XGate but got some issues with the implementation of the [FixedIntializer](https://docs.rs/ndarray/latest/ndarray/trait.FixedInitializer.html) traits for arrays of more than 16 elements. There are workarounds, but given its low priority the implementation has been left for a follow-up.


test/python/circuit/test_rust_equivalence.py
```diff
@@ -19,10 +19,15 @@
 import numpy as np
 
 from qiskit.circuit import QuantumCircuit
+from qiskit.circuit.library.standard_gates import *
```
ElePT:
```suggestion
from qiskit.circuit.library.standard_gates import C3XGate
```

test/python/visualization/test_circuit_text_drawer.py
```diff
@@ -5208,7 +5208,7 @@ def test_open_controlled_u1(self):
         expected = "\n".join(
             [
                 "                                                           ",
-                "qr_0: |0>─o─────────o─────────o─────────o─────────■────────",
+                "qr_0: |0>─■─────────o─────────o─────────o─────────■────────",
```
ElePT: I am not sure why this test changed, but I don't think it's wrong like this.

mtreinish: I think you'll need to update the logic here: https://github.com/Qiskit/qiskit/blob/main/crates/circuit/src/circuit_instruction.rs#L834-L851 or in the circuit method (it depends how the gate is being appended to the circuit).

The issue is a controlled gate with a non default ctrl_state is not a standard gate. The standard gate definition is not taken into account and we should treat a custom `ctrl_state` as a `PyGate`. It looks like the logic I had for this only covers when the input is a singleton which ignores standard controlled gates that take a parameter.

ElePT: Updated logic and added test in [9b47121](https://github.com/Qiskit/qiskit/pull/12659/commits/9b47121e1afb51e8f3bdd609edbb67939fd0286d)

crates/circuit/src/circuit_instruction.rs
```diff
@@ -878,20 +878,37 @@ pub fn convert_py_to_operation_type(
...
+        let is_default_ctrl_state: bool = match py_op.getattr(py, intern!(py, "ctrl_state")) {
+            Ok(c_state) => match c_state.extract::<Option<i32>>(py) {
+                Ok(c_state_int) => match c_state_int {
+                    Some(c_int) => c_int == 1,
+                    None => true,
+                },
+                Err(_) => false,
+            },
+            Err(_) => false,
+        };
```
mtreinish: I feel like we should only do this check if we know we're a non-singleton controlled gate. If we're a singleton the `mutable` flag is sufficient (since a custom ctrl state is always mutable in python). I just think we should try to avoid the extra getattr overhead except in cases where we know we need to check. Maybe we could just make this a closure and call it in L911.

crates/circuit/src/operations.rs
```diff
@@ -305,13 +303,13 @@ static STANDARD_GATE_NAME: [&str; STANDARD_GATE_SIZE] = [
     "cu",           // 39
     "cu1",          // 40
     "cu3",          // 41
-    "c3x",          // 42
+    "c3x",          // 42 ("mcx")
```
mtreinish: I feel like this mismatch will cause a problem eventually. Like it's not an issue right now because we never use these gates out of the box anywhere. But we use these names as the canonical identifier in a lot of places (which is why the overloaded `mcx` name is a real problem). For example, when we port the qasm exporters to rust they'll just call `op.name()` internally and then this will return `c3x` instead of `mcx` which will cause issues.

crates/circuit/src/operations.rs
```diff
@@ -1406,6 +1875,21 @@ fn add_param(param: &Param, summand: f64, py: Python) -> Param {
     }
 }
 
+fn sum_params(param1: Param, param2: Param, py: Python) -> Param {
```
mtreinish: Maybe `radd` instead of `sum`? I was a bit confused by this name at first because I knew we added `add` already for `PGate`.

crates/circuit/src/operations.rs
ElePT: I agree. But it seems like the solution would be not using the overloaded `mcx` name in Python, wouldn't it?

mtreinish: It would, and ideally that's how I would expect us to fix this. But when we discussed potentially doing that as part of #12628 I thought the concern was around backwards compatibility and the consensus was we shouldn't change the name until 2.0?

crates/circuit/src/circuit_instruction.rs
ElePT: Ok, so the status of things after [da3e598](https://github.com/Qiskit/qiskit/pull/12659/commits/da3e598b91c216cbddad3413990a90b3151196a9) (and [b70799a](https://github.com/Qiskit/qiskit/pull/12659/commits/b70799a4e35a722e06cc7bde18816a89acb8a725) which I needed to fix the logic that I screwed up in [05fe760](https://github.com/Qiskit/qiskit/pull/12659/commits/05fe7608072dcb146189a0ca59c13864cafb272a)) is that we check for the following conditions to skip the rust definition:

a. mutable and (singleton or controlled singleton)
or
b. controlled and non-default-ctrl-state (I understand there's no need to add the mutable condition here because it will always be mutable?)

crates/circuit/src/operations.rs
ElePT: Absolutely, this was created before `add` and I didn't update the name. I renamed it to `radd_param` (singular for consistency) in [da3e598](https://github.com/Qiskit/qiskit/pull/12659/commits/da3e598b91c216cbddad3413990a90b3151196a9).

crates/circuit/src/operations.rs
ElePT: That's what I vaguely remembered. If we want to rename the gates for 2.0, should we then introduce deprecation warnings in 1.2? I am not sure what the protocol is with name changes, we can't add the new path in advance, right?

ElePT: hah, this just became a problem after the recent commits: https://github.com/Qiskit/qiskit/actions/runs/9764823435/job/26954059268?pr=12659#step:8:16894

mtreinish: The deprecation path for something like this isn't clear as it's a straight change in behavior. There isn't a good way to signal this to users unless we emit a `FutureWarning` every time a `C3XGate` is created. This might be the case we just document this in the 2.0 release notes and say it wasn't easy to raise a deprecation warning for.

crates/circuit/src/circuit_instruction.rs
mtreinish: Yeah if you do things in that order it should work I think.

crates/circuit/src/operations.rs
ElePT: Ok, I didn't see a way around the recent failures other than renaming "c3x" to "mcx" in rust space ([f05c9a1](https://github.com/Qiskit/qiskit/pull/12659/commits/f05c9a1b320ad61e79ecc711e66208de083bc11a)). I don't love it, but I guess we can then change both names at the same time?

crates/circuit/src/circuit_instruction.rs
```diff
@@ -878,24 +878,49 @@ pub fn convert_py_to_operation_type(
...
+        // These are the only cases where controlled gates can be standard.
+        let is_default_ctrl_state = || match py_op.getattr(py, intern!(py, "ctrl_state")) {
+            Ok(c_state) => match c_state.extract::<Option<i32>>(py) {
+                Ok(c_state_int) => match c_state_int {
+                    Some(c_int) => {
+                        let qubits: u32 = py_op
+                            .getattr(py, intern!(py, "num_qubits"))
+                            .unwrap()
+                            .extract(py)
+                            .unwrap();
```
mtreinish: I think we can preserve the errors here with:

```suggestion
                            .getattr(py, intern!(py, "num_qubits"))?
                            .extract(py)?;
```

Then in L918 we can just do `!is_default_ctrl_state()?)` to return that back to the users.

crates/circuit/src/operations.rs
```diff
@@ -311,13 +309,13 @@ static STANDARD_GATE_NAME: [&str; STANDARD_GATE_SIZE] = [
     "cu",           // 39
     "cu1",          // 40
     "cu3",          // 41
-    "c3x",          // 42
+    "mcx",          // 42 ("mcx")
     "c3sx",         // 43
-    "c4x",          // 44
+    "c4x",          // 44 ("mcx")
```
mtreinish: Doesn't this need to be `mcx` too?

crates/circuit/src/operations.rs
```diff
@@ -311,13 +309,13 @@ static STANDARD_GATE_NAME: [&str; STANDARD_GATE_SIZE] = [
...
     "dcx",          // 45
     "ccz",          // 46
     "rccx",         // 47
-    "rc3x",         // 48
+    "rcccx",        // 48 ("rc3x")
```
mtreinish, line 318: Same here shouldn't this match the python side name and be `rc3x`?

test/python/circuit/test_rust_equivalence.py
```diff
@@ -175,3 +190,46 @@ def test_num_params(self):
...
+                "11": default_op,
+                None: default_op,
+                0: non_default_op_0,
+                1: non_default_op_1,
+                "01": non_default_op_1,
+                "10": non_default_op_2,
+            }
+            for state, op in state_out_map.items():
+                circuit = QuantumCircuit(3)
+                circuit.append(z_gate.control(2, ctrl_state=state), [0, 1, 2])
```
mtreinish: Does this actually create a `CCZGate` object? Or does this create a custom `ControlledGate` instance? To test this effectively we'll need to ensure it's actually a `CCZGate` instance because we want both the `._standard_gate` attribute set and also a custom `ctrl_state` .

mtreinish (commented):
This is looking good, thanks for doing all of this. I just have a couple of small inline comments.

crates/circuit/src/operations.rs
ElePT: One would think so (that's how I originally called it), but the python side name is `rcccx` :sweat_smile: https://github.com/Qiskit/qiskit/blob/419f40e9e72b4981a6536d08d6b0fa91c13021d5/qiskit/circuit/library/standard_gates/x.py#L874

crates/circuit/src/operations.rs
ElePT: It would if we were actually adding the gate, but it's not implemented at the moment, so it doesn't make a difference. We could change it for consistency.

crates/circuit/src/circuit_instruction.rs
ElePT: This was my original idea, but adding the `?` triggers this annoying compilation errror:

```
error[E0277]: the `?` operator can only be used in a closure that returns `Result` or `Option` (or another type that implements `FromResidual`)
   --> crates/circuit/src/circuit_instruction.rs:902:68
    |
897 |         let is_default_ctrl_state = || match py_op.getattr(py, intern!(py, "ctrl_state")) {
    |                                     -- this function should return `Result` or `Option` to accept `?`
...
902 |                             .getattr(py, intern!(py, "num_qubits"))?
    |                                                                    ^ cannot use the `?` operator in a closure that returns `bool`
    |
    = help: the trait `FromResidual<Result<Infallible, pyo3::PyErr>>` is not implemented for `bool`
```

Aha, I can just change the typehint, one moment.

crates/circuit/src/operations.rs
ElePT: Ok, changed in [8c922aa](https://github.com/Qiskit/qiskit/pull/12659/commits/8c922aa5031b6bdf9c0782869156eba8ffe210f9).

test/python/circuit/test_rust_equivalence.py
ElePT: Addressed in [6e7b076](https://github.com/Qiskit/qiskit/pull/12659/commits/6e7b0768e29370aa13c297f9ea9ee15d2f791c42)

crates/circuit/src/circuit_instruction.rs
ElePT: Ok, either I'm missing something or it's not so easy to preserve the error. Changing the return type doesn't help: 
```
the trait `FromResidual<Result<Infallible, pyo3::PyErr>>` is not implemented for `Option<bool>`
```

crates/circuit/src/operations.rs
jlapeyre: The change has already been made but: Also, it seems customary in PLs to use "add" for two items, and "sum" for a list of items.

crates/circuit/src/circuit_instruction.rs
jlapeyre: No you're not missing something obvious. This is a common problem. You can find buzz about it online. I finally found a clean solution for a PR yesterday. I'll look it up and see if it applies here.

I don't think the stuff I found online was directly relevant to the solution I found. In fact I think a lot of the discussion was not helpful. You can search for variants of : "rust question mark operator return  closure"

jlapeyre: EDIT: Committed in 2a98908
Here is a PR to this branch https://github.com/ElePT/qiskit/pull/34 that returns `PyResult` rather than panic on the two `unwrap`s.  This is slightly different from yesterday's issue that I mentioned above, so I don't reference the solution here.

ElePT: Thanks a lot @jlapeyre, it would have taken me a long time to find the fix. I will keep it in mind for future PRs.

crates/circuit/src/circuit_instruction.rs
```diff
@@ -878,24 +878,49 @@ pub fn convert_py_to_operation_type(
...
     // is the control state, so for `SingletonControlledGates` we'll still need
     // this check.
     if standard.is_some() {
         let mutable: bool = py_op.getattr(py, intern!(py, "mutable"))?.extract(py)?;
-        if mutable
+        // The default ctrl_states are the all 1 state and None.
+        // These are the only cases where controlled gates can be standard.
+        let is_default_ctrl_state = || -> PyResult<bool> {
+            match py_op.getattr(py, intern!(py, "ctrl_state")) {
+                Ok(c_state) => match c_state.extract::<Option<i32>>(py) {
```
mtreinish, line 899: One thing we can do here is since we're not returning the error to python is use `downcast_bound()` instead of `extract()` (see: https://github.com/Qiskit/qiskit/commit/2c3ddf1e6f81c497297b1d3531f20ea62b4cb620 where I did this too recently) which saves us the python error generation.

crates/circuit/src/operations.rs
mtreinish: Ah ok, I was just thrown by the comment here. Because the others had the Python name. But now it's all consistent with the desired name in the comment.

test/python/circuit/test_rust_equivalence.py
```diff
@@ -148,6 +161,9 @@ def test_name(self):
             if standard_gate is None:
                 # gate is not in rust yet
                 continue
+            if gate_class.name == "mcx":
+                # ambiguos gate name
```
mtreinish:
```suggestion
                # ambiguous gate name
```

test/python/circuit/test_rust_equivalence.py
```diff
@@ -12,17 +12,21 @@
...
 
 from qiskit.circuit import QuantumCircuit, CircuitInstruction
+from qiskit.circuit.library.standard_gates import C3XGate, CU1Gate, CZGate, CCZGate
 from qiskit.circuit.library.standard_gates import get_standard_gate_name_mapping
+from qiskit.quantum_info import Operator
 
 SKIP_LIST = {"rx", "ry", "ecr"}
-CUSTOM_MAPPING = {"x", "rz"}
+CUSTOM_NAME_MAPPING = {"mcx": C3XGate()}
+MATRIX_SKIP_LIST = {"c3sx"}
```
mtreinish, line 29: I thought I saw the rust definition of this matrix using the macro in the code?

mtreinish (commented):
I think this looks good to go now. I just a had a few very small comments inline. The only other question I have is about `c4x` if we're not going to implement it in rust should we remove it from the list?

test/python/circuit/test_rust_equivalence.py
ElePT: Yes, the test is being skipped because of the python definition, not the rust one.
`qiskit.circuit.exceptions.CircuitError: "to_matrix not defined for this <class '_SingletonC3SXGate'>"`

crates/circuit/src/circuit_instruction.rs
ElePT: I think it's not possible to downcast to a rust-native type. The compiler at least doesn't like it:

```
899  |                 Ok(c_state) => match c_state.downcast_bound::<Option<i32>>() {
     |                                                               ^^^^^^^^^^^ the trait `PyTypeInfo` is not implemented for `Option<i32>`
```

test/python/circuit/test_rust_equivalence.py
mtreinish: Hmm, that seems like a bug in the python side then. But lets fix that in a separate PR

ElePT: I will try to find a fix tomorrow in a separate PR.

mtreinish (approved):
This LGTM now, thanks for doing this. The only open question is around whether we should remove `c4x` from the list. But we can decide that in a follow up, or add a comment. It's more effort to remove it as part of this since it already had a placeholder.

test/python/circuit/test_rust_equivalence.py
ElePT: #12742

PR #12629, opened 2024-06-21
Deprecate BackendV1 and `qiskit.providers.models`

1ucian0:
### Summary

Time to deprecate BackendV1 in favor of `Target` (for hardware description) and primitives (for execution). We should still keep BackendV2 as a supported alternative.

This requieres (i.e this branch is on top of these PRs, only makes sense once they are merged. Therefore `on hold`):

- [x] https://github.com/Qiskit/qiskit/pull/12649
- [x] https://github.com/Qiskit/qiskit/pull/12660
- [x] ~https://github.com/Qiskit/qiskit/pull/12788 (for after merging this PR)~

once this one is merged, the rest of the `_assemble` code needs to be cleaned. Hopefully, it can be done after the release of this PR.

PR #12660, opened 2024-06-25
Deprecate fake backends based on BackendV1

1ucian0:
~Branches from `1ucian0:deprecate/qobj_assemble/1` https://github.com/Qiskit/qiskit/pull/12649, so blocked on that~

### Summary

Some of the introduced TODOs are waiting for fixes:

- [x] https://github.com/Qiskit/qiskit/issues/12760
- [x] https://github.com/Qiskit/qiskit/issues/12759
- [x] https://github.com/Qiskit/qiskit/issues/12832


1ucian0:
https://github.com/Qiskit/qiskit/pull/12649 merged. Removing `on hold`.

test/python/compiler/test_compiler.py
```diff
@@ -198,6 +197,7 @@ def test_parallel_compile(self):
         qc.measure(qr[5], cr[0])
         qlist = [qc for k in range(10)]
         with self.assertWarns(DeprecationWarning):
+            backend = Fake20QV1()
```
ElePT, line 200: I liked your comment from the previous file :)

```suggestion
            # Ignorable, because assemble and BackendV1 are deprecated
            backend = Fake20QV1()
```

test/python/compiler/test_sequencer.py
```diff
@@ -27,7 +27,8 @@ class TestSequence(QiskitTestCase):
 
     def setUp(self):
         super().setUp()
-        self.backend = Fake127QPulseV1()
+        with self.assertWarns(DeprecationWarning):
+            self.backend = Fake127QPulseV1()
```
ElePT, line 33: What would be the migration path for this test once BackendV1 is removed? is this one of the pending TODOs? (maybe it could be marked more clearly).

test/python/compiler/test_transpiler.py
```diff
@@ -1511,7 +1511,8 @@ def test_scheduling_timing_constraints(self):
         """Test that scheduling-related loose transpile constraints
         work with both BackendV1 and BackendV2."""
 
-        backend_v1 = Fake27QPulseV1()
+        with self.assertWarns(DeprecationWarning):
+            backend_v1 = Fake27QPulseV1()
         backend_v2 = BackendV2Converter(backend_v1)
```
ElePT: This can be left for a follow-up, but eventually this test should be using `GenericBackendV2` as `backend_v2`. It doesn't make much more sense to do it now because it's easier if  the properties of`backend_v1` and `backend_v2` are identical.

test/python/compiler/test_transpiler.py
```diff
@@ -2068,14 +2072,81 @@ def test_transpile_annotated_ops(self, opt_level):
...
         """Test transpilation of circuits with annotated operations given a backend."""
         qc = QuantumCircuit(3)
         qc.append(AnnotatedOperation(SGate(), InverseModifier()), [0])
         qc.append(AnnotatedOperation(XGate(), ControlModifier(1)), [1, 2])
         qc.append(AnnotatedOperation(HGate(), PowerModifier(3)), [2])
-        backend = Fake20QV1()
+
+        backend = GenericBackendV2(
+            num_qubits=20,
+            coupling_map=[
```
ElePT: Can be left for a follow-up, but I don't think this backend needs to have 20 qubits (could be 3 or 4 with an equivalent cmap for those qubits).

test/python/circuit/test_scheduled_circuit.py
```diff
@@ -50,21 +51,24 @@ def setUp(self):
...
+        qc.delay(0.1, 0, unit="ms")  # 450450[dt]
         qc.delay(100, 0, unit="ns")  # 450[dt]
-        qc.h(0)  # 160[dt]
-        qc.h(1)  # 160[dt]
-        sc = transpile(qc, self.backend_with_dt, scheduling_method="alap", layout_method="trivial")
-        self.assertEqual(sc.duration, 450546)
+        qc.h(0)  # 195[dt]
+        qc.h(1)  # 210[dt]
+
+        backend = GenericBackendV2(27, calibrate_instructions=True, seed=42)
```
ElePT: Can be left for a follow-up, but I don't think this backend needs to have 27 qubits, this is a bit of a quirk that came from the lack of V1 backends with the correct pulse support. But a 2 qubit backend should do the job.

ElePT: I am not 100% sure it will not trigger changes, but in case you want to try:

```suggestion
        backend = GenericBackendV2(2, calibrate_instructions=True, seed=42)
```

test/python/compiler/test_transpiler.py
```diff
@@ -2383,9 +2454,12 @@ def test_qpy_roundtrip_standalone_var_target(self, optimization_level):
     @data(0, 1, 2, 3)
     def test_qasm3_output(self, optimization_level):
         """Test that the output of a transpiled circuit can be dumped into OpenQASM 3."""
+        with self.assertWarns(DeprecationWarning):
+            backend = Fake20QV1()
```
ElePT, line 2450: Is this test supposed to work with BackendV2? If so, we could maybe duplicate it to have a V1 and V2 version like it's done in other tests.

test/python/primitives/test_primitive.py
```diff
@@ -142,7 +142,10 @@ def test_with_scheduling(n):
                 qc = QuantumCircuit(1)
                 qc.x(0)
                 qc.add_calibration("x", qubits=(0,), schedule=custom_gate)
-                return transpile(qc, Fake20QV1(), scheduling_method="alap")
+                backend = GenericBackendV2(
+                    num_qubits=20, basis_gates=["id", "u1", "u2", "u3", "cx"], seed=42
+                )
```
ElePT: This one probably can work with a 2-q backend (avoiding having only 1 qubit not to run into edge cases with the 2q basis gates).
```suggestion
                backend = GenericBackendV2(
                    num_qubits=2, basis_gates=["id", "u1", "u2", "u3", "cx"], seed=42
                )
```

test/python/providers/test_backendconfiguration.py
```diff
@@ -26,7 +26,8 @@ class TestBackendConfiguration(QiskitTestCase):
 
     def setUp(self):
         super().setUp()
-        backend = FakeOpenPulse2Q()
+        with self.assertWarns(DeprecationWarning):
+            backend = FakeOpenPulse2Q()
```
ElePT, line 33: Just confirming, there is no migration path here because BackendConfiguration will get deprecated with BackendV1? then you could add the todo message:

```
    # TODO the full file can be removed once BackendV1 is removed, since it is the
    #  only one with backend.configuration()
```

test/python/providers/test_fake_backends.py
```diff
@@ -256,8 +260,8 @@ def test_converter_delay_circuit(self):
 
     def test_converter_with_missing_gate_property(self):
         """Test converting to V2 model with irregular backend data."""
-        backend = FakeOpenPulse2Q()
-
+        with self.assertWarns(DeprecationWarning):
+            backend = FakeOpenPulse2Q()
```
ElePT, line 264: I just realized that once V1 is gone, we cannot really test the BackendConverter anymore. I guess it will be deprecated too?

test/python/providers/test_pulse_defaults.py
```diff
@@ -26,8 +26,13 @@ class TestPulseDefaults(QiskitTestCase):
 
     def setUp(self):
         super().setUp()
-        self.defs = FakeOpenPulse2Q().defaults()
-        self.inst_map = self.defs.instruction_schedule_map
+        with self.assertWarns(DeprecationWarning):
```
ElePT, line 29: I am guessing the plan is to remove this test with BackendV1. Should this be noted in a comment?

test/python/result/test_mitigators.py
```diff
@@ -119,8 +119,8 @@ def counts_data(circuit, assignment_matrices, shots=1024):
     def test_mitigation_improvement(self):
         """Test whether readout mitigation led to more accurate results"""
         shots = 1024
-        assignment_matrices = self.assignment_matrices()
-        num_qubits = len(assignment_matrices)
+        with self.assertWarns(DeprecationWarning):
+            assignment_matrices = self.assignment_matrices()
```
ElePT: What triggers a warning in this test?

test/python/transpiler/test_calibrationbuilder.py
```diff
@@ -266,7 +266,8 @@ def build_reverse(
     @data(-np.pi / 4, 0.1, np.pi / 4, np.pi / 2, np.pi)
     def test_rzx_calibration_cr_pulse_stretch(self, theta: float):
         """Test that cross resonance pulse durations are computed correctly."""
-        backend = Fake27QPulseV1()
+        with self.assertWarns(DeprecationWarning):
+            backend = Fake27QPulseV1()
         inst_map = backend.defaults().instruction_schedule_map
```
ElePT, line 271: This can be left for a follow-up, but if I'm not mistaken these tests should be migrated to use `GenericBackendV2`

test/python/transpiler/test_passmanager_run.py
```diff
@@ -83,12 +83,63 @@ def test_default_pass_manager_single(self):
         circuit.cx(qr[1], qr[2])
         circuit.cx(qr[2], qr[3])
 
-        coupling_map = Fake20QV1().configuration().coupling_map
+        coupling_map = [
```
ElePT: can be left for a follow-up but I think these tests can run with a smaller number of qubits (the 20 was kind of arbitrary).

test/python/visualization/test_gate_map.py
```diff
@@ -139,7 +258,295 @@ def test_plot_error_map_backend_v2(self):
     @unittest.skipUnless(optionals.HAS_SEABORN, "Seaborn not installed")
     def test_plot_error_map_over_100_qubit(self):
         """Test plotting error map with large fake backend."""
-        backend = Fake127QPulseV1()
+        coupling_map = [
```
ElePT: Can be left for a follow-up but we could maybe take advantage of some of the pre-stored coupling maps in `test` not to have to hard-code it here. Visual tests are annoying to change so I am ok with this as an intermediate step.

ElePT (commented):
Thanks a lot @1ucian0. The PR looks pretty good. I left some comments but they are mostly remarks and reminders for follow-ups.

test/python/compiler/test_compiler.py
1ucian0: multi branch PRing is hard :P

1ucian0: Removing all these comments in f34620cebab81a486fb4b6646a228b81c7a233a1 . They are obvious, even for future @1ucian0

test/python/compiler/test_transpiler.py
1ucian0: oh! you created them already. They live in `test/python/legacy_cmaps.py`. ~I will move to them in this PR~ finished in bd2e363ea5694261a2a42e60e791eeeb8634fcf2

test/python/compiler/test_sequencer.py
1ucian0: I added a note in 469189c7a77f59016954836791a41e1c4f564462. 
I dont think there is a migration path. They will be removed with pulse. Is that right @mtreinish ?

test/python/compiler/test_transpiler.py
1ucian0: I move it to `GenericBackendV2` in 6013044499e0a8949d0aa2a44d9f34f8f38c1468 and it seems to work.

test/python/circuit/test_scheduled_circuit.py
1ucian0: oh! good point. I just blindly put a similar backend. Moved to 2 qubits in 5eb561e94b07e53119e1c5e75bce69236f387afd

test/python/compiler/test_transpiler.py
1ucian0: done in 11a82d43a9fd62f6777b00e45b3edf2cfb889197

test/python/providers/test_backendconfiguration.py
1ucian0: correct. Comment added in e005b48671e171d81ec5a2940c8183d757a35356

test/python/circuit/test_scheduled_circuit.py
1ucian0: yeap.. 2 works

test/python/primitives/test_primitive.py
1ucian0: 2 qubits work

test/python/providers/test_pulse_defaults.py
1ucian0: I need to instantiate a V1 because V2 does not have `defaults()`

test/python/transpiler/test_calibrationbuilder.py
1ucian0: They use `defaults()` tho

test/python/transpiler/test_passmanager_run.py
1ucian0: The test seems to try to be a challenge for layout and routing, so I prefer to keep it in a bigger backend. However, I used `ALMADEN_CMAP` in 982e274d80572cef4c207eafb6dda7a49250c1b6

test/python/visualization/test_gate_map.py
1ucian0: I added `KYOTO_CMAP` (127 qubits) in 36633d17c5db1e8718ac069e35fe4c3fe22b8125

test/python/providers/test_fake_backends.py
1ucian0: Hmm... not sure... I assumed that it will be removed with pulse. 

Is there a possibility of a gate in BackendV2 with calibration but no "property"?

```python
        # The backend includes pulse calibration definition for U2, but its property is gone.
        # Note that u2 is a basis gate of this device.
        # Since gate property is not provided, the gate broadcasts to all qubits as ideal instruction.
        del backend._properties._gates["u2"]
```

Honestly, I think I'm missing what this test is testing. @nkanazawa1989 ? @mtreinish ?

test/python/providers/test_pulse_defaults.py
1ucian0: will be gone with pulse, I assume

test/python/result/test_mitigators.py
1ucian0: good catch! I added a comment in 889173bc154b5fe79594694a2fadd8da59986753 and submitted https://github.com/Qiskit/qiskit/issues/12832

test/python/compiler/test_transpiler.py
ElePT: nice 👍

test/python/transpiler/test_calibrationbuilder.py
ElePT: The `instruction_schedule_map` can be retrieved from the target in the V2 model as `backend.target.instruction_schedule_map()`, no need to go through `defaults`

test/python/providers/test_fake_backends.py
ElePT: My comment might have been confusing, it's not necessarily about this test. I was thinking of the general context of the `BackendV1` deprecation (#12629) and whether we should also deprecate `BackendConverter` with it. I find it tricky because `BackendConverter` is part of the migration path, so it's strange to have it raise a deprecation warning when we tell people to use it, but it cannot be tested once we get rid of `BackendV1`, so in that sense I think it should be deprecated. We can maybe raise a custom warning to explain this situation with it.

ElePT (commented):
I think the changes look very good. My only concern is clearing up what parts will stay and what parts will go with the pulse deprecation. Mostly because the `instruction_schedule_map` is currently part of the target model, so I am not sure how these would be decoupled, or whether it will be kept as a standalone element. Similarly, I am not sure whether the scheduling passes will all be deprecated or whether these will be kept somehow. (Now that I think of it, the concerns are mostly for #12629 so I am fine approving this PR and moving the discussion there, where the context might be more fitting).

ElePT (approved):
As mentioned above, my current questions can be moved to the `BackendV1` deprecation PR directly, so approving this one.

1ucian0:
For reference:

https://github.com/Qiskit/qiskit/pull/12660#discussion_r1694781920 : 

> The `instruction_schedule_map` can be retrieved from the target in the V2 model as
> `backend.target.instruction_schedule_map()`, no need to go through `defaults`

Trying to move that test to backendV2 ended up in https://github.com/Qiskit/qiskit/issues/12834 . I added a comment in 0c2b6487ad7ea45300f297321bf04843ba2b9dab included in https://github.com/Qiskit/qiskit/pull/12629

https://github.com/Qiskit/qiskit/pull/12660#discussion_r1694786739 : follow up in [#12629](https://github.com/Qiskit/qiskit/pull/12629#issuecomment-2255558384)

PR #12661, opened 2024-06-25
Simplify QuantumCircuit._from_circuit_data bit handling

mtreinish:
### Summary

This commit simplifies the logic around bit handling in the `QuantumCircuit._from_circuit_data()` constructor. Previously it was calling `add_bits()` for each bit in the `CircuitData` object to update the output circuit's accounting for each qubit. But this was needlessly heavy as the `CircuitData` is already the source of truth for the bits in a circuit and we just need to update the indices dictionary. The `add_bits()` method attempts to add the bits to the `CircuitData` too but this is wasted overhead because the `CircuitData` already has the bits as that's where the came from. This changes the constructor to just directly set the bit indices as needed and return the circuit.

### Details and comments

### DIFF
--- a/qiskit/circuit/quantumcircuit.py
+++ b/qiskit/circuit/quantumcircuit.py
@@ -1162,10 +1162,8 @@ class QuantumCircuit:
         """A private constructor from rust space circuit data."""
         out = QuantumCircuit()
         out._data = data
-        for index, bit in enumerate(data.qubits):
-            out._qubit_indices[bit] = BitLocations(index, [])
-        for index, bit in enumerate(data.clbits):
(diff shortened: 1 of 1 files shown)


jakelishman (dismissed):
Was this noticeable in performance scanning?  If so, perhaps dictionary comprehensions will be marginally faster than the loop too.

Otherwise, no problem and feel free to tag for merge.

mtreinish:
It showed up in py-spy profiling I was doing of permutation synthesis. It wasn't a major component but I was surprised to see it there so I pushed this up to fix it. A dict comprehension is a good idea and will make it a bit faster, I'll tweak this shortly.

jakelishman (approved):

PR #12662, opened 2024-06-26
Implement RGate in Rust

jlapeyre:
This completes all tasks of implementing `RGate` in Rust rather than Python.

This replaces 
* #12507

#12507 has been modified and split into the present PR and #12651

### DIFF
--- a/crates/circuit/src/operations.rs
+++ b/crates/circuit/src/operations.rs
@@ -968,7 +968,7 @@ impl Operation for StandardGate {
                     Param::Obj(_) => unreachable!(),
                 };
                 let (phi_expr1, phi_expr2) = match &params[1] {
-                    Param::Float(phi) => (Param::Float(*phi - 1.0), Param::Float(-*phi + 1.0)),
+                    Param::Float(phi) => (Param::Float(*phi - PI2), Param::Float(-*phi + PI2)),
                     Param::ParameterExpression(phi) => {
                         let phiexpr1 = phi
                             .call_method1(py, intern!(py, "__add__"), ((-PI / 2.0),))
--- a/crates/circuit/src/operations.rs
+++ b/crates/circuit/src/operations.rs
@@ -967,7 +967,7 @@ impl Operation for StandardGate {
                     }
                     Param::Obj(_) => unreachable!(),
                 };
-                let (phi_expr1, phi_expr2) = match &params[0] {
+                let (phi_expr1, phi_expr2) = match &params[1] {
                     Param::Float(phi) => (Param::Float(*phi - 1.0), Param::Float(-*phi + 1.0)),
                     Param::ParameterExpression(phi) => {
                         let phiexpr1 = phi
--- a/crates/circuit/src/operations.rs
+++ b/crates/circuit/src/operations.rs
@@ -960,13 +960,7 @@ impl Operation for StandardGate {
             }),
             Self::CRXGate | Self::CRYGate | Self::CRZGate => todo!(),
             Self::RGate => Python::with_gil(|py| -> Option<CircuitData> {
-                let theta_expr = match &params[0] {
-                    Param::Float(theta) => Param::Float(*theta),
-                    Param::ParameterExpression(theta) => {
-                        Param::ParameterExpression(theta.clone_ref(py))
-                    }
-                    Param::Obj(_) => unreachable!(),
-                };
+                let theta_expr = clone_param(&params[0], py);
                 let (phi_expr1, phi_expr2) = match &params[1] {
                     Param::Float(phi) => (Param::Float(*phi - PI2), Param::Float(-*phi + PI2)),
                     Param::ParameterExpression(phi) => {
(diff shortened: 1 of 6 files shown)


test/python/circuit/test_rust_equivalence.py
```diff
@@ -21,7 +21,7 @@
 from qiskit.circuit import QuantumCircuit
 from qiskit.circuit.library.standard_gates import get_standard_gate_name_mapping
 
-SKIP_LIST = {"rx", "ry", "ecr"}
+SKIP_LIST = {"rx", "ry", "ecr", "r"}
```
ElePT: Why add the `r` gate to the SKIP_LIST? All others haven't been implemented yet, but `r` has an implementation.

crates/circuit/src/operations.rs
```diff
@@ -954,7 +959,41 @@ impl Operation for StandardGate {
...
+                Some(
+                    CircuitData::from_standard_gates(
+                        py,
+                        1,
+                        [(Self::UGate, defparams, smallvec![Qubit(0)])],
+                        FLOAT_ZERO,
+                    )
+                    .expect("Unexpected Qiskit python bug"),
+                )
+            }),
```
ElePT: I think this could be more readable if we used helper functions like [`multiply_param`](https://github.com/Qiskit/qiskit/blob/6974b4500f6c716b407b599e5cec80afbb757516/crates/circuit/src/operations.rs#L983), and maybe add one for adding a scalar to a parameter in the same fashion. But I am fine merging the implementation as-is and beautifying it later.

ElePT (commented):
Thanks for the PR! it looks pretty good, I just left a couple of comments.

crates/circuit/src/operations.rs
```diff
@@ -954,7 +959,41 @@ impl Operation for StandardGate {
...
-            Self::RGate => todo!(),
+            Self::RGate => Python::with_gil(|py| -> Option<CircuitData> {
+                let theta_expr = match &params[0] {
+                    Param::Float(theta) => Param::Float(*theta),
+                    Param::ParameterExpression(theta) => {
+                        Param::ParameterExpression(theta.clone_ref(py))
+                    }
+                    Param::Obj(_) => unreachable!(),
+                };
+                let (phi_expr1, phi_expr2) = match &params[0] {
```
ElePT: Isn't phi the second parameter:

```suggestion
                let (phi_expr1, phi_expr2) = match &params[1] {
```

crates/circuit/src/operations.rs
jlapeyre: My first iteration did not fit with the helper function, nor one for `add`. But it's changed a lot since I last tried. I'll revisit it; readability and encapsulation is important.

test/python/circuit/test_rust_equivalence.py
jlapeyre: Good question. I added and removed several times while debugging. I'll recheck.

jlapeyre: Fixed in [`7eef0f3` (#12662)](https://github.com/Qiskit/qiskit/pull/12662/commits/7eef0f3d8292f12ebbdcb04b1fb3834e5141ea1d). It failed because the decomposition had an error.

crates/circuit/src/operations.rs
jlapeyre: I did something similar to your suggest in [`49c8dcd` (#12662)](https://github.com/Qiskit/qiskit/pull/12662/commits/49c8dcd69f8c31dd0dd48be4425e9af6b3357dd4). This factors out cloning a `Param`. This is only used once now, but can be used whenever a `Param` needs to be cloned. I find the code much more readable at the call point.

Regarding factoring out `add_param`: There are pros and cons to doing this. Disadvantages: The branch with `-*phi + PI / 2.0` is more readable if it is not nested calls. In fact, this made it slightly easier for me to spot the error `-*phi + 1.0`.

Another one is that it requires two match statements rather than one, so `-*phi + PI / 2.0`  won't be optimized. I'm not sure, but I doubt this will matter in the future for performance.

But, I'm on the fence. Currently, the second match arm would be much more readable with your suggestion. I can implement and use `add_param` quickly if you prefer. I'm inclined to do it just to look at it.

jlapeyre: Implemented your suggestion in [`07fec37` (#12662)](https://github.com/Qiskit/qiskit/pull/12662/commits/07fec379319090b0b065b2589b63ddc75b7430c4). I'm not 100% sure there isn't a non-negligible efficiency hit. But I think it's easier to read and will be easier to build on.

ElePT: I see your point, I think we could merge the code as-is and once all Rust standard gates are in place, maybe run the benchmarks with/without `add_param` to see if there is indeed a significant performance hit.

ElePT (approved):
Thanks a lot @jlapeyre! I see that there might be a trade-off between readability and some performance here, but given that the RGate is used in #12650, I would merge it now to unblock downstream PRs and leave any further optimizations for a follow-up, as I mentioned in my other comment.

PR #10325, opened 2023-06-23
DAGCircuit: Add get_causal_cone method

raynelfss:
### Summary
The causal node of a qubit is the set of all the qubits that affect the resulting output of the aforementioned qubit. This PR aims to add the `get_causal_node` method to `DAGCircuit` to obtain the causal node of a qubit in a circuit.

A couple of helper methods have also been added:
- `get_qubit_input_node`: to get the input node of a certain qubit in the circuit.
- `get_qubit_output_node`: to get the output node of a certain qubit in the circuit.

### Details and comments
- The implementation of these methods may not be the most optimal one, so it is open to suggestions from the community.

PR #12668, opened 2024-06-26
improving `quantum_causal_cone` method in python

alexanderivrii:
### Summary

When trying to port the method `DAGCircuit.quantum_causal_cone` (introduced in #10325) to Rust, I have noticed that the existing algorithm examines each node multiple times (when reached from different paths), making the algorithm worst-case exponential. 

For a specific example, consider the circuit
```
q0_0: ──■─────────■─────────■─────────■─────────■─────────■───────
      ┌─┴─┐     ┌─┴─┐     ┌─┴─┐     ┌─┴─┐     ┌─┴─┐     ┌─┴─┐     
q0_1: ┤ X ├──■──┤ X ├──■──┤ X ├──■──┤ X ├──■──┤ X ├──■──┤ X ├──■──
      └───┘┌─┴─┐└───┘┌─┴─┐└───┘┌─┴─┐└───┘┌─┴─┐└───┘┌─┴─┐└───┘┌─┴─┐
q0_2: ──■──┤ X ├──■──┤ X ├──■──┤ X ├──■──┤ X ├──■──┤ X ├──■──┤ X ├      ...
      ┌─┴─┐└───┘┌─┴─┐└───┘┌─┴─┐└───┘┌─┴─┐└───┘┌─┴─┐└───┘┌─┴─┐└───┘
q0_3: ┤ X ├─────┤ X ├─────┤ X ├─────┤ X ├─────┤ X ├─────┤ X ├─────
      └───┘     └───┘     └───┘     └───┘     └───┘     └───┘     
q0_4: ────────────────────────────────────────────────────────────
```

In this PR we hash the set of processed **non-directive** nodes. 

Let me also explain the point about not hashing "directive" nodes such as barriers (or measures). A barrier may involve all the qubits in the circuit, however not all of these should be considered to be a part of the causal cone. So when processing a barrier node the original algorithm only adds the quantum predecessors of that barrier node that intersect the current set of qubits in the quantum cone. However, this does potentially require to examine each barrier multiple times. (_Aside: I am still not fully certain that this logic is 100% correct, however I was not able to find an example where this does not do what's intended_). See also #10325.

This PR does **not** change this logic, however for clarity the two cases (directives and not-directives) are treated separately (and there are some other minor renaming and other minor code simplifications). It also adds more Python tests that explore the corner-cases.

It would be nice to have this PR merged before adding the Rust version of this to the new DAGCircuit, as it can serve as a reference for that implementation and also adds more python tests.

### DIFF
--- a/releasenotes/notes/improve-quantum-causal-cone-f63eaaa9ab658811.yaml
+++ b/releasenotes/notes/improve-quantum-causal-cone-f63eaaa9ab658811.yaml
@@ -1,0 +1,5 @@
+---
+upgrade_circuits:
+  - |
+    Improved the method :meth:`.DAGCircuit.quantum_causal_cone` to avoid examining the same
+    non-directive node multiple times when reached from different paths.
--- a/releasenotes/notes/improve-quantum-causal-cone-f63eaaa9ab658811.yaml
+++ b/releasenotes/notes/improve-quantum-causal-cone-f63eaaa9ab658811.yaml
@@ -1,5 +1,5 @@
 ---
-upgrade_circuits:
+features_circuits:
   - |
-    Improved the method :meth:`.DAGCircuit.quantum_causal_cone` to avoid examining the same
-    non-directive node multiple times when reached from different paths.
+    Improved performance of the method :meth:`.DAGCircuit.quantum_causal_cone` by not examining
+    the same non-directive node multiple times when reached from different paths.
--- a/qiskit/dagcircuit/dagcircuit.py
+++ b/qiskit/dagcircuit/dagcircuit.py
@@ -2264,36 +2264,44 @@ class DAGCircuit:
         output_node = self.output_map.get(qubit, None)
         if not output_node:
             raise DAGCircuitError(f"Qubit {qubit} is not part of this circuit.")
-        # Add the qubit to the causal cone.
-        qubits_to_check = {qubit}
-        # Add predecessors of output node to the queue.
-        queue = deque(self.predecessors(output_node))
 
-        # While queue isn't empty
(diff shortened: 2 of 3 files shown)


raynelfss (dismissed):
This LGTM, thank you for improving this function!

releasenotes/notes/improve-quantum-causal-cone-f63eaaa9ab658811.yaml
```diff
@@ -0,0 +1,5 @@
+---
+upgrade_circuits:
+  - |
+    Improved the method :meth:`.DAGCircuit.quantum_causal_cone` to avoid examining the same
+    non-directive node multiple times when reached from different paths.
```
mtreinish: This isn't really an upgrade note, an upgrade note is reserved for something that we want to document might require user input when upgrading from the previous release to the new one. Examples would be a breaking api change, minimum dependency version bump, etc. I would expect this to be more of a feature note (to indicate improved performance) or a bug fix release note if we want to document this in the release.

alexanderivrii: Thanks, this is not the first time I am making this mistake, but I will eventually remember what "upgrade" stands for.

alexanderivrii: done in [23877fe](https://github.com/Qiskit/qiskit/pull/12668/commits/23877fe9fcb07837b6bba3e085597ad1c881c79e)

raynelfss (approved):
Thank you for fixing the release note, this should be good to merge.

PR #12669, opened 2024-06-26
Encapsulate Python sequence-like indexers

jakelishman:
### Summary

This encapsulates a lot of the common logic around Python sequence-like indexers (`SliceOrInt`) into iterators that handle adapting negative indices and slices in `usize` for containers of a given size.

These indexers now all implement `ExactSizeIterator` and `DoubleEndedIterator`, so they can be used with all `Iterator` methods, and can be used (with `Iterator::map` and friends) as inputs to `PyList::new_bound`, which makes code simpler at all points of use.

The special-cased uses of this kind of thing from `CircuitData` are replaced with the new forms.  This had no measurable impact on performance on my machine, and removes a lot noise from error-handling and highly specialised functions.

### Details and comments

This PR was not meant to be this big - I only started writing something up because I needed similar logic to things we already had in 3+ different places for `SparseObservable`, and then I let it get a bit more out of hand.

This introduces `thiserror` as a crate for generating the boilerplate for implementing `Error` for return values.  I barely used it in this PR, but I split this off from a follow-up defining `SparseObservable`, which uses it more heavily.

### DIFF
--- a/crates/circuit/src/circuit_data.rs
+++ b/crates/circuit/src/circuit_data.rs
@@ -22,11 +22,12 @@ use crate::imports::{BUILTIN_LIST, QUBIT};
 use crate::interner::{IndexedInterner, Interner, InternerKey};
 use crate::operations::{Operation, OperationType, Param, StandardGate};
 use crate::parameter_table::{ParamEntry, ParamTable, GLOBAL_PHASE_INDEX};
-use crate::{Clbit, Qubit, SliceOrInt};
+use crate::slice::{PySequenceIndex, SequenceIndex};
+use crate::{Clbit, Qubit};
 
 use pyo3::exceptions::{PyIndexError, PyValueError};
 use pyo3::prelude::*;
-use pyo3::types::{PyList, PySet, PySlice, PyTuple, PyType};
+use pyo3::types::{PyList, PySet, PyTuple, PyType};
 use pyo3::{intern, PyTraverseError, PyVisit};
 
 use hashbrown::{HashMap, HashSet};
@@ -321,7 +322,7 @@ impl CircuitData {
     }
 
     pub fn append_inner(&mut self, py: Python, value: PyRef<CircuitInstruction>) -> PyResult<bool> {
-        let packed = self.pack(py, value)?;
+        let packed = self.pack(value)?;
         let new_index = self.data.len();
         self.data.push(packed);
         self.update_param_table(py, new_index, None)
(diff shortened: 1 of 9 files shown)


crates/circuit/src/slice.rs
```diff
@@ -0,0 +1,375 @@
...
+}
+
+impl<'py> PySequenceIndex<'py> {
+    /// Specialize this index to a collection of the given `len`, returning a Rust-native type.
+    pub fn with_len(&self, len: usize) -> Result<SequenceIndex, PySequenceIndexError> {
+        match self {
+            PySequenceIndex::Int(index) => {
+                let index = if *index >= 0 {
+                    let index = *index as usize;
+                    if index >= len {
```
Cryoris, line 51: Could this also be written in a single line as follows? 🙂 
```suggestion
                    if *index as usize >= len {
```

crates/circuit/src/slice.rs
```diff
@@ -0,0 +1,375 @@
...
+                    index
+                } else {
+                    len.checked_sub(index.unsigned_abs())
+                        .ok_or(PySequenceIndexError::OutOfRange)?
+                };
+                Ok(SequenceIndex::Int(index))
+            }
+            PySequenceIndex::Slice(slice) => {
+                let indices = slice
+                    .indices(len as ::std::os::raw::c_long)
```
Cryoris, line 63: Why do we need this type instead of e.g. using `as isize` (which `PySliceMethods.indices` seems to take)?

crates/circuit/src/slice.rs
```diff
@@ -0,0 +1,375 @@
...
+    /// The number of indices this refers to.
+    pub fn len(&self) -> usize {
+        match self {
+            Self::Int(_) => 1,
+            Self::PosRange { start, stop, step } => {
+                let gap = stop.saturating_sub(*start);
+                gap / *step + (gap % *step != 0) as usize
+            }
+            Self::NegRange { start, stop, step } => 'arm: {
+                let Some(start) = start else { break 'arm 0 };
```
Cryoris, line 134: Now I'm relieved we don't share an office with you breaking limbs 😉 but on a more serious note, I'm a bit confused about this, couldn't I have a Python slice with negative start and stop and then this would end up with two `None` values?

Cryoris (commented):
Nice to see so much code repetition to go away! I mainly have a question about negative Python slices (like `vec[-5:-1]`) 🙂

crates/circuit/src/slice.rs
jakelishman: I think you're looking at PyO3 0.22 docs - in 0.21 it takes a `c_long` (see PyO3/pyo3#3761).

crates/circuit/src/slice.rs
jakelishman: Haha yeah, I meant the "match arm" but I didn't even think about that!

Yes, you can have two `None`, and in fact I think that's the _only_ time `indices.start` can be `None`.  Either if it weren't, the length would still always be zero because the start point isn't a valid index.

crates/circuit/src/slice.rs
jakelishman: I don't remember exactly why I split it off like this, but if you want to change it, line 54 would also need changing to the same thing.  If you want to remake the suggestion with both I'll commit it.

crates/circuit/src/circuit_data.rs
```diff
@@ -744,184 +745,130 @@ impl CircuitData {
...
+            slf.update_param_table(py, index, None)?;
+            Ok(())
+        }
 
-                for (i, v) in slice.iter().zip(values.iter()) {
-                    self.__setitem__(py, SliceOrInt::Int(*i), v)?;
+        let py = value.py();
+        match index.with_len(self.data.len())? {
+            SequenceIndex::Int(index) => set_single(self, index, value),
+            indices @ SequenceIndex::PosRange {
```
mtreinish, line 797: I think this is the first time I've seen the match binding syntax used in practice.

crates/circuit/src/circuit_data.rs
```diff
@@ -1175,56 +1122,22 @@ impl CircuitData {
...
+        if !indices.is_empty() {
+            self.reindex_parameter_table(py)?;
         }
-        Ok(index as usize)
+        Ok(())
     }
 
-    fn pack(&mut self, py: Python, inst: PyRef<CircuitInstruction>) -> PyResult<PackedInstruction> {
+    fn pack(&mut self, inst: PyRef<CircuitInstruction>) -> PyResult<PackedInstruction> {
+        let py = inst.py();
```
mtreinish, line 1140: This feels a bit unnecessary but it's not a big deal I guess.

crates/circuit/src/slice.rs
```diff
@@ -0,0 +1,375 @@
...
+impl<'py> FromPyObject<'py> for PySequenceIndex<'py> {
+    fn extract_bound(ob: &Bound<'py, PyAny>) -> PyResult<Self> {
+        // `slice` can't be subclassed in Python, so it's safe (and faster) to check for it exactly.
+        // The `downcast_exact` check is just a pointer comparison, so while `slice` is the less
+        // common input, doing that first has little-to-no impact on the speed of the `isize` path,
+        // while the reverse makes `slice` inputs significantly slower.
+        if let Ok(slice) = ob.downcast_exact::<PySlice>() {
+            return Ok(Self::Slice(slice.clone()));
+        }
+        Ok(Self::Int(ob.extract()?))
```
mtreinish, line 40: Heh, this makes me think of: https://github.com/Qiskit/rustworkx/pull/1096 where you opted for just flipping the order. This is starting to make me think maybe we want to contribute this to PyO3 so there is a common impl of this for everyone.

mtreinish (approved):
This lgtm, this is a great abstraction to add and makes working with slices much less error prone. All the logic seems super through and accounted for edge cases I didn't even realize were there. As a standalone PR I think this is good to merge. 

But I'm wondering about rustworkx and probably any other pyo3 users out there. I'm thinking we should contribute this to PyO3 or have it somewhere more publicly accessible because I think it'll be useful for any custom sequence types people are creating with PyO3. It's providing a nice API to work with slices which are super error prone with the current pyo3 interface because they're super low level and the Python C API doesn't document things super well.

crates/circuit/src/circuit_data.rs
jakelishman: I think I used it in the OQ2 lexer as well - I don't use it often, but it's occasionally useful.

crates/circuit/src/circuit_data.rs
jakelishman: Yeah, I don't think I intended to make a change here - it might be a remnant of something else I was up to.

crates/circuit/src/slice.rs
jakelishman: You're just watching me learning more of the low-level internals of everything live haha.

In Rustworkx I didn't know enough other than to leave the `FromPyObject` derivation in place, in which case flipping the order just makes things faster.  But for the sake of completeness here, since I'd already wasted enough time that a little longer made no difference, I actually checked whether `slice` is subclassable (turns out it isn't), in which case the derived `FromPyObject` is needlessly slow; you can't subclass it, so the `type is slice` check is sufficient and is just a pointer comparison.

jakelishman:
Yeah, I considered talking to PyO3 about contributing it upstream, though I'm thinking maybe we play with it a shade longer within Qiskit first to see if the abstraction actually bears weight.

I'm _ok_ with the interface I've come up with here, but I think it could still be better.  If nothing else, there's tons of specialisations of `Range` that could be passed through the various iterator methods.  I've mostly just added the bits we need for Qiskit immediately (though tbh `DoubleEndedIterator` for `Descending` was probably not 100% required) to avoid further bloating a PR that was already more involved than I was expecting.

PR #12670, opened 2024-06-26
Fixing and documenting how the `Estimator` calculates stds

SamFerracin:
### Summary
This PR slightly modifies the formula used by the `Estimator` to calculate `std`s, and adds details on how the `std`s are computed.

### Details and comments
Addresses issue [1751](https://github.com/Qiskit/qiskit-ibm-runtime/issues/1751).


qiskit/primitives/backend_estimator_v2.py
```diff
@@ -254,8 +265,8 @@ def _postprocess_pub(
             for pauli, coeff in bc_obs[index].items():
                 expval, variance = expval_map[param_index, pauli]
                 evs[index] += expval * coeff
-                variances[index] += variance * coeff**2
-        stds = np.sqrt(variances / shots)
+                variances[index] += coeff * variance**0.5
+        stds = variances / np.sqrt(shots)
```
t-imamichi: Why do you suggest this change? It does not make sense to me to store `variance ** 0.5` to `variances`.
What is the issue of the current implementation?

SamFerracin: @t-imamichi this addresses [issue 1751](https://github.com/Qiskit/qiskit-ibm-runtime/issues/1751)

After considering a few different approaches, I opted for:
- Slightly modifying the way the stds are calculated, so that our stds equal the square root of the parameter epsilon in Eq. 5 of [this paper](https://quantum-journal.org/papers/q-2021-01-20-385/pdf/) and have the units of standard deviation.
- Improving the documentation, making sure to explain what exactly we are reporting as std.

qiskit/primitives/backend_estimator_v2.py
```diff
@@ -87,7 +87,18 @@ class BackendEstimatorV2(BaseEstimatorV2):
     precludes doing any provider- or backend-specific optimizations.
 
     This class does not perform any measurement or gate mitigation, and, presently, is only
-    compatible with Pauli-based observables.
+    compatible with Pauli-based observables. More formally, given an observable of the type
+    :math:`O=\sum_{i=1}^Na_iP_i`, where :math:`a_i` is a complex number and :math:`P_i` is a
+    Pauli operator, the estimator calculates the expectation :math:`E(P_i)` of each :math:`P_i`
+    and finally calculates the expectation value of :math:`O` as
+    :math:`E(O)==\sum_{i=1}^Na_iE(P_i)`. The reported ``std`` is calculated as
```
t-imamichi: `==` does not look good. Could you use `=`, `:=`, or `\equiv`, etc. as you wish?   
image

qiskit/primitives/backend_estimator_v2.py
```diff
@@ -87,7 +87,18 @@ class BackendEstimatorV2(BaseEstimatorV2):
...
+
+    .. math::
+
+        \frac{\sum_{i=1}^{n}a_i\sqrt{\textrm{Var}\big(P_i\big)}}{\sqrt{N}}\:,
```
t-imamichi: I guess you need `abs` as you did at line 268. Is it correct?
```suggestion
        \frac{\sum_{i=1}^{n}|a_i|\sqrt{\textrm{Var}\big(P_i\big)}}{\sqrt{N}}\:,
```

qiskit/primitives/backend_estimator_v2.py
t-imamichi: Thank you for the details. I understand that the current implementation of stds is too small.
It might be nice to rename array `variances` because it stores square root of variances.
But since this method is short, I'm fine to leave the name as is.

t-imamichi:
Thank you for fixing the issue. Could you write a release note of bugfix?
Could you also update the title of this PR because it fixes stds of BackendEstimatorV2 in addition to the documentation?

qiskit/primitives/backend_estimator_v2.py
SamFerracin: Oh, thank you for spotting this! It should just be `=`

qiskit/primitives/backend_estimator_v2.py
```diff
@@ -72,7 +72,7 @@ class _PreprocessedData:
 
 
 class BackendEstimatorV2(BaseEstimatorV2):
-    """Evaluates expectation values for provided quantum circuit and observable combinations
+    r"""Evaluates expectation values for provided quantum circuit and observable combinations
```
ihincks:
```suggestion
    r"""Evaluates expectation values for provided quantum circuit and observable combinations.
```

qiskit/primitives/backend_estimator_v2.py
```diff
@@ -87,7 +87,18 @@ class BackendEstimatorV2(BaseEstimatorV2):
     precludes doing any provider- or backend-specific optimizations.
 
     This class does not perform any measurement or gate mitigation, and, presently, is only
-    compatible with Pauli-based observables.
+    compatible with Pauli-based observables. More formally, given an observable of the type
+    :math:`O=\sum_{i=1}^Na_iP_i`, where :math:`a_i` is a complex number and :math:`P_i` is a
+    Pauli operator, the estimator calculates the expectation :math:`E(P_i)` of each :math:`P_i`
```
ihincks: mathbb or rn for the E's?

qiskit/primitives/backend_estimator_v2.py
```diff
@@ -87,7 +87,18 @@ class BackendEstimatorV2(BaseEstimatorV2):
...
+    and finally calculates the expectation value of :math:`O` as
+    :math:`E(O)==\sum_{i=1}^Na_iE(P_i)`. The reported ``std`` is calculated as
+
+    .. math::
+
+        \frac{\sum_{i=1}^{n}|a_i|\sqrt{\textrm{Var}\big(P_i\big)}}{\sqrt{N}}\:,
```
ihincks, line 99: what is `\:` ? I've never used that
```suggestion
        \frac{\sum_{i=1}^{n}|a_i|\sqrt{\textrm{Var}\big(P_i\big)}}{\sqrt{N}},
```

ihincks: what's the link to that paper which suggests this? it's not intuitive to me.

SamFerracin: Eq. 5 of https://arxiv.org/abs/1908.06942. I added a reference to it in the docstring

SamFerracin: `\:` leaves a blank space. I normally use it when I have a comma or a break at the end of a full-line equation

releasenotes/notes/backend-estimator-v2-variance-905c953415ad0e29.yaml
```diff
@@ -0,0 +1,5 @@
+---
+fixes:
+  - |
+    Changes the way in which the `BackendEstimatorV2` class calculates the `std` to ensure that
```
t-imamichi: FYI: Reno uses sphinx format not markdown.

```suggestion
    Changes the way in which the :class:`.BackendEstimatorV2` class calculates the ``std`` to ensure that
```

jsaroni:
In improving the backend_estimator_v2.py documentation, (unrelated to stds, sorry), there is this comment under `_preprocess_pub`

`# calculate expectation values for each pair of parameter value set and pauli `

but the lines following the comment don't actually calculate the expectation values.

```
        # calculate expectation values for each pair of parameter value set and pauli
        param_obs_map = defaultdict(set)
        for index in np.ndindex(*bc_param_ind.shape):
            param_index = bc_param_ind[index]
            param_obs_map[param_index].update(bc_obs[index])
```

t-imamichi:
Good catch, @jsaroni. We will update the comment when we update the method.

releasenotes/notes/backend-estimator-v2-variance-905c953415ad0e29.yaml
SamFerracin: Thank you for this!

t-imamichi (approved):
LGTM

PR #12671, opened 2024-06-26
Add base representation of `SparseObservable`

### LINKED ISSUE #12282: Sparse operators on extended alphabets
### What should we add?

The current primary operator class in Qiskit, `SparsePauliOp`, is limited to only Pauli observables, and is actually not sparse, but rather holds data in the form of two dense NumPy arrays.  This design choice comes with several limitations.  First, the restriction to Pauli operators means that one cannot use the current class for computing expectation values of projector operators, like those used in QML kernel methods, requiring a Pauli decomposition that scales exponentially.  Second, the choice of internal representation results in a memory footprint and operator construction time that is larger than other quantum SDKs.

Instead, Qiskit should move to a real sparse format that allows for operations on extended alphabets that include projection operators, the most important being `'0'` and `'1'`.  The limited alphabet can be stored efficiently where the indices of only non-identity terms should be kept in the data structure.

jakelishman:
### Summary

﻿﻿This adds the base representation of `SparseObservable`, including the simple constructors from Python space and the ability to view the data buffers.

This commit does not include the mathematical manipulations of the operators, nor some of the helper methods that will be used to manipulate the operators in the context of primitives execution. These will follow in subsequent patches.

The design and implementation notes of `SparseObservable` are described in a Qiskit RFC that preceeded this patch series[^1], and it's best to consult that document for full details on the operator considerations.

[^1]: https://github.com/Qiskit/RFCs/blob/7a74b08793475b7b0142d3a3f7142cabcfd33ab8/0021-sparse-observable.md

### Details and comments

This is the beginnings of the implementation of Qiskit/RFCs#74, and subject to change based on decisions made there.

A current simple example of what works with the PR:
```python
import numpy as np
from qiskit.quantum_info import SparseObservable

# Using the copy-in constructor to make the Z_0 + Z_1 + Z_2 + ... operator
# by manually providing the data buffers.
num_qubits = 100
terms = np.full((num_qubits,), SparseObservable.BitTerm["+"], dtype=np.uint8)
indices = np.arange(num_qubits, dtype=np.uint32)
coeffs = np.ones((num_qubits,), dtype=complex)
boundaries = np.arange(num_qubits + 1, dtype=np.uintp)
# This copies out of the buffers, because we need the base allocations to be
# owned by Rust space.
so = SparseObservable.from_raw_parts(num_qubits, coeffs, terms, indices, boundaries)

print(so.bit_terms)
```

This PR depends on #12669.

Close #12282.  This PR only adds the representation of such operators, but that's sufficient to immediately fulfil the feature as an MVP.  Qiskit/RFCs#75 tracks the rest of the implementation of the RFC.


jakelishman:
No longer "on hold" because its dependency merged, but I do still need to write a lot more of the initialisation methods for this.

jakelishman:
This base PR for `SparseObservable` is now ready for review.  I'll follow it (at some point) with another adding mathematical operators to the observable, and one (or more) adding helper methods for use by and for the primitives.

jakelishman:
The only missing lines in the coverage report are ones that are unreachable, either clearly from the code, or because they're error paths to handle things like a `SparsePauliOp` that violates its internal data structures during the extraction of the data to Rust space.

(Also there's a line about `copy=False` in an `__array__` method, but that actually _is_ tested, it's just the coverage run uses `numpy<2` due to some transitive dependency, which suppresses that test.)

crates/accelerate/src/sparse_observable.rs
```diff
@@ -0,0 +1,1677 @@
...
+        BitTerm::Right,
+        BitTerm::Left,
+        BitTerm::Z,
+        BitTerm::Zero,
+        BitTerm::One,
+    ]
+    .into_iter()
+    .flat_map(|term| {
+        let mut out = vec![(term.py_name(), term as u8)];
+        if term.py_name() != term.py_label() {
```
Cryoris, line 190: I'm not sure I fully understand this: This will attach another tuple `(label, term as u8)` for the projectors, such that the Python enum will allow construction from both the label and the name i.e. `IntEnum.PLUS` and `IntEnum.+` (<--- hang on this can't be right??), right?

crates/accelerate/src/sparse_observable.rs
```diff
@@ -0,0 +1,1677 @@
...
+                        py_enum
+                            .getattr(term.py_name())
+                            .expect("the created `BitTerm` enum should have matching attribute names to the terms")
+                            .unbind()
+                    })
+            })
+        });
+        terms[self as usize]
+            .as_ref()
+            .expect("the lookup table initialiser populated a 'Some' in all valid locations")
```
Cryoris: Optional, of course 🇬🇧 😛 
```suggestion
            .expect("the lookup table initializer populated a 'Some' in all valid locations")
```

crates/accelerate/src/sparse_observable.rs
```diff
@@ -0,0 +1,1677 @@
...
+        ),
+    )?;
+    Ok(obj.downcast_into::<PyType>()?.unbind())
+}
+
+// Return the relevant value from the Python-space sister enumeration.  These are Python-space
+// singletons and subclasses of Python `int`.  We only use this for interaction with "high level"
+// Python space; the efficient Numpy-like array paths use `u8` directly so Numpy can act on it
+// efficiently.
+impl IntoPy<Py<PyAny>> for BitTerm {
```
Cryoris, line 216: Just for my understanding: Could this also have been solved by adding a 
```
#[pyclass]
pub enum PyBitTerm {
    // all enum options here ...
}
```
which is exposed to Python and then `impl IntoPy<...> for BitTerm` would internally use the `PyBitTerm`? The current creation looks like black magic to me but I'll try to understand it better 😄

crates/accelerate/src/sparse_observable.rs
```diff
@@ -0,0 +1,1677 @@
...
+///
+/// The length parameter :math:`t` is the number of terms in the sum, and the parameter :math:`s` is
+/// the total number of non-identity single-qubit terms.
+///
+/// As illustrative examples:
+///
+/// * in the case of a zero operator, :attr:`boundaries` is length 1 (a single 0) and all other
+///   vectors are empty.
+/// * in the case of a fully simplified identity operator, :attr:`boundaries` is ``[0, 0]``,
+///   :attr:`coeffs` has a single entry, and :attr:`bit_terms` and :attr:`indices` are empty.
```
Cryoris, line 436: A simple (but non-trivial 😉 ) example might be nice. This is partially for my own understanding; `s` includes all single-qubit Paulis, right? (I.e. `ZZ` would have `s=2`?)
```suggestion
///   :attr:`coeffs` has a single entry, and :attr:`bit_terms` and :attr:`indices` are empty.
/// * a ``ZZ - IX`` observable would have :math:`t=2` coefficients (namely ``1`` and ``-1``)
///   and :math:`s=3` single-qubit Pauli terms (namely ``Z, Z, X``)
```

crates/accelerate/src/sparse_observable.rs
```diff
@@ -0,0 +1,1677 @@
...
+///
+///     The numeric structure of these is that they are all four-bit values of which the low two
+///     bits are the (phase-less) symplectic representation of the Pauli operator related to the
+///     object, where the low bit denotes a contribution by :math:`Z` and the second lowest a
+///     contribution by :math:`X`, while the upper two bits are ``00`` for a Pauli operator, ``01``
+///     for the negative-eigenstate projector, and ``10`` for teh positive-eigenstate projector.
+///
+///     .. autoattribute:: qiskit.quantum_info::SparseObservable.BitTerm.X
+///
+///         The Pauli :math:`X` operator.
```
Cryoris: It's sort of obvious but with the other labels mentioned below it might avoid some confusion to add it here (confusion as in: "doesn't Pauli X have a label too?")
```suggestion
///         The Pauli :math:`X` operator, labelled ``"X"``.
```

crates/accelerate/src/sparse_observable.rs
```diff
@@ -0,0 +1,1677 @@
...
+///     :ref:`sparse-observable-alphabet`.
+///
+///     This class is attached to :class:`.SparseObservable`.  Access it as
+///     :class:`.SparseObservable.BitTerm`.
+///
+///     The numeric structure of these is that they are all four-bit values of which the low two
+///     bits are the (phase-less) symplectic representation of the Pauli operator related to the
+///     object, where the low bit denotes a contribution by :math:`Z` and the second lowest a
+///     contribution by :math:`X`, while the upper two bits are ``00`` for a Pauli operator, ``01``
+///     for the negative-eigenstate projector, and ``10`` for teh positive-eigenstate projector.
```
Cryoris:
```suggestion
///     for the negative-eigenstate projector, and ``10`` for the positive-eigenstate projector.
```

crates/accelerate/src/sparse_observable.rs
```diff
@@ -0,0 +1,1677 @@
...
+///     else in this source file. The use of `autoattribute` is because it pulls in the numeric
+///     value.
+///
+/// .. py:class:: SparseObservable.BitTerm
+///
+///     An :class:`~enum.IntEnum` that provides named access to the numerical values used to
+///     represent each of the single-qubit alphabet terms enumerated in
+///     :ref:`sparse-observable-alphabet`.
+///
+///     This class is attached to :class:`.SparseObservable`.  Access it as
```
Cryoris, line 464: Users are not expected to handle `BitTerm`s directly, since they can just use the labels, right? Otherwise the access via `SparseObservable` seems like it could be a bit much typing 🙂

crates/accelerate/src/sparse_observable.rs
```diff
@@ -0,0 +1,1677 @@
...
+/// :class:`~numpy.ndarray` containing a copy of the relevant data with the natural ``dtype`` of the
+/// field; this lets you easily do mathematics on the results, like bitwise operations on
+/// :attr:`bit_terms`.  You can assign to indices or slices of each of the attributes, but beware
+/// that you must uphold :ref:`the data coherence rules <sparse-observable-arrays>` while doing
+/// this.  For example::
+///
+///     >>> obs = SparseObservable.from_list([("XZY", 1.5j), ("+1r", -0.5)])
+///     >>> assert isinstance(obs.coeffs[:], np.ndarray)
+///     >>> # Reduce all single-qubit terms to the relevant Pauli operator, if they are a projector.
+///     >>> obs.bit_terms[:] = obs.bit_terms[:] & 0b00_11
```
Cryoris, line 537: That's really neat 😁

crates/accelerate/src/sparse_observable.rs
```diff
@@ -0,0 +1,1677 @@
...
+///     >>> assert obs == SparseObservable.from_list([("XZY", 1.5j), ("XZY", -0.5)])
+///
+///
+/// Construction
+/// ============
+///
+/// :class:`SparseObservable` defines several constructors.  The default constructor will attempt to
```
Cryoris, line 544: Above this always had a leading `.` but here (and below in some instances) it doesn't -- I _guess_ it's not necessary since this should be injected to where the `SparseObservable` is imported but I'm not sure at all. So I'm just highlighting this in case it needs changing 🙂 
```suggestion
/// :class:`.SparseObservable` defines several constructors.  The default constructor will attempt to
```

crates/accelerate/src/sparse_observable.rs
```diff
@@ -0,0 +1,1677 @@
...
+/// .. py:function:: SparseObservable.__new__(data, /, num_qubits=None)
+///
+///     The default constructor of :class:`SparseObservable`.
+///
+///     This delegates to one of :ref:`the explicit conversion-constructor methods
+///     <sparse-observable-convert-constructors>`, based on the type of the ``data`` argument.  If
+///     ``num_qubits`` is supplied and constructor implied by the type of ``data`` does not accept a
+///     number, the given integer must match the input.
+///
+///     :param data: The data type of the input.  This can be another :class:`SparseObservable`, in
```
Cryoris, line 581: Is that in the `:param ...:` format due to being written in the Rust file, or could this also be our typical `Args: ...`?

crates/accelerate/src/sparse_observable.rs
```diff
@@ -0,0 +1,1677 @@
...
+                bit_terms: &self.bit_terms[start..end],
+                indices: &self.indices[start..end],
+            }
+        })
+    }
+
+    /// Add the term implied by a dense string label onto this observable.
+    pub fn add_dense_label<L: AsRef<[u8]>>(
+        &mut self,
+        label: L,
```
Cryoris, line 718: Is that to allow both `&[u8]` and `&[BitTerm]` as inputs? Or why is this needed instead of e.g. `label: &[u8]`? 🙂

crates/accelerate/src/sparse_observable.rs
```diff
@@ -0,0 +1,1677 @@
...
+        coeff: Complex64,
+    ) -> Result<(), LabelError> {
+        let label = label.as_ref();
+        if label.len() != self.num_qubits() as usize {
+            return Err(LabelError::WrongLengthDense {
+                num_qubits: self.num_qubits(),
+                label: label.len(),
+            });
+        }
+        // The only valid characeters in the alphabet are ASCII, so if we see something other than
```
Cryoris:
```suggestion
        // The only valid characters in the alphabet are ASCII, so if we see something other than
```

crates/accelerate/src/sparse_observable.rs
```diff
@@ -0,0 +1,1677 @@
...
+        self.boundaries.push(self.bit_terms.len());
+        Ok(())
+    }
+}
+
+#[pymethods]
+impl SparseObservable {
+    #[pyo3(signature = (data, /, num_qubits=None))]
+    #[new]
+    fn py_new(data: Bound<PyAny>, num_qubits: Option<u32>) -> PyResult<Self> {
```
Cryoris, line 756: Just a question: does `py_new` then get called as `__new__` or is this literally a method that will be called `SparseObservable.py_new`?

crates/accelerate/src/sparse_observable.rs
```diff
@@ -0,0 +1,1677 @@
...
+                    "explicitly given 'num_qubits' ({}) does not match label ({})",
+                    num_qubits,
+                    label.len(),
+                )));
+            }
+            return Self::py_from_label(&label).map_err(PyErr::from);
+        }
+        if let Ok(observable) = data.downcast::<Self>() {
+            check_num_qubits(&data)?;
+            return Ok(observable.borrow().clone());
```
Cryoris, line 792: The check in here is just a sanity check in case the `SparseObservable` has some weird setting, is it? I mean like
```python
obs = SparseObservable("111")  # this is always a valid SparseObservable
obs.num_qubits = 1  # but we can mutilate it
new_obs = SparseObservable(obs)
```
Otherwise I would've thought that the `SparseObservable` is always in a valid format and we could omit that check 😄

crates/accelerate/src/sparse_observable.rs
```diff
@@ -0,0 +1,1677 @@
...
+        }
+        if let Ok(vec) = data.extract() {
+            return Self::py_from_list(vec, num_qubits);
+        }
+        if let Ok(vec) = data.extract() {
```
Cryoris, line 800: I'm sure this works, but I don't understand it: why would the second if branch ever be taken here? 🤔

crates/accelerate/src/sparse_observable.rs
```diff
@@ -0,0 +1,1677 @@
...
+    #[getter]
+    #[inline]
+    pub fn num_qubits(&self) -> u32 {
+        self.num_qubits
+    }
+
+    /// The number of terms in the sum this operator is tracking.
+    #[getter]
+    #[inline]
+    pub fn num_ops(&self) -> usize {
```
Cryoris: I'm pondering whether we could find a slightly better name, since we don't use abbreviations that much (but `num_operators` is also long to type...). How about `terms`? That would allow a natural distinction between `bit_terms` and `terms` 🙂 
```suggestion
    pub fn num_terms(&self) -> usize {
```

crates/accelerate/src/sparse_observable.rs
```diff
@@ -0,0 +1,1677 @@
...
+        self.coeffs.len()
+    }
+
+    /// The coefficients of each abstract term in in the sum.  This has as many elements as terms in
+    /// the sum.
+    #[getter]
+    fn get_coeffs(slf_: Py<Self>) -> ArrayView {
```
Cryoris, line 834: Would this be called Python-side as `observable.get_coeffs` or `observable.get_coeffs()` (asking due to the `#[getter]` qualifier). If it were the first, should this just be `coeffs`?

crates/accelerate/src/sparse_observable.rs
```diff
@@ -0,0 +1,1677 @@
...
+        BIT_TERM_PY_ENUM
+            .get_or_try_init(py, || make_py_bit_term(py))
+            .map(|obj| obj.clone_ref(py))
+    }
+
+    /// Get the zero operator over the given number of qubits.
+    ///
+    /// Examples:
+    ///
+    ///     Get the zero operator for 100 qubits::
```
Cryoris, line 903: When I first read this I expected this to return `|000....><000...|` on `n` qubits 😅 but I also don't really have a better naming suggestions... maybe `empty`? But I don't really like that either...

crates/accelerate/src/sparse_observable.rs
```diff
@@ -0,0 +1,1677 @@
...
+            num_qubits,
+            coeffs: vec![Complex64::new(1.0, 0.0)],
+            bit_terms: vec![],
+            indices: vec![],
+            boundaries: vec![0, 0],
+        }
+    }
+
+    fn __repr__(&self) -> String {
+        let num_terms = format!(
```
Cryoris, line 940: This variable naming is one argument in favor of renaming `num_ops() -> num_terms()` 😜

crates/accelerate/src/sparse_observable.rs
```diff
@@ -0,0 +1,1677 @@
...
+            "{} term{}",
+            self.num_ops(),
+            if self.num_ops() == 1 { "" } else { "s" }
```
Cryoris: Kudos to using the right plural form of "terms" 👍🏻

crates/accelerate/src/sparse_observable.rs
```diff
@@ -0,0 +1,1677 @@
...
+    }
+
+    /// Construct an observable from a list of dense labels and coefficients.
+    ///
+    /// This is analagous to :meth:`.SparsePauliOp.from_list`, except it uses
+    /// :ref:`the extended alphabet <sparse-observable-alphabet>` of :class:`.SparseObservable`.  In
+    /// this dense form, you must supply all identities explicitly in each label.
+    ///
+    /// The label must be a sequence of the alphabet ``'IXYZ+-rl01'``.  The label is interpreted
+    /// analagously to a bitstring.  In other words, the right-most letter is associated with qubit
```
Cryoris:
```suggestion
    /// analogously to a bitstring.  In other words, the right-most letter is associated with qubit
```

crates/accelerate/src/sparse_observable.rs
```diff
@@ -0,0 +1,1677 @@
...
+    ///     Construct an observable from a list of labels of the same length::
+    ///
+    ///         >>> SparseObservable.from_list([
+    ///         ...     ("III++", 1.0),
+    ///         ...     ("II--I", 1.0j),
+    ///         ...     ("I++II", -0.5),
+    ///         ...     ("--III", -0.25j),
+    ///         ... ])
+    ///         <SparseObservable with 4 terms on 5 qubits:
+    ///             (1+0j)(+_1 +_0) + (0+1j)(-_2 -_1) + (-0.5+0j)(+_3 +_2) + (-0-0.25j)(-_4 -_3)>
```
Cryoris, line 1080: I like that the order of operators in the print reflects the qubit indexing 👍🏻  (i.e. that it is `+_1 +_0` instead of `+_0 +_1` -- ok the difference would be more visible if the bit terms were different haha)

crates/accelerate/src/sparse_observable.rs
```diff
@@ -0,0 +1,1677 @@
...
+    fn py_from_sparse_list(
+        iter: Vec<(String, Vec<u32>, Complex64)>,
+        num_qubits: u32,
+    ) -> Result<Self, LabelError> {
+        let coeffs = iter.iter().map(|(_, _, coeff)| *coeff).collect();
+        let mut boundaries = Vec::with_capacity(iter.len() + 1);
+        boundaries.push(0);
+        let mut indices = Vec::new();
+        let mut bit_terms = Vec::new();
+        let mut sorted = btree_map::BTreeMap::new();
```
Cryoris, line 1189: This is to ensure the indices are sorted in ascending order, right? It would be nice to add a comment for our future selves 🙂

crates/accelerate/src/sparse_observable.rs
```diff
@@ -0,0 +1,1677 @@
...
+            for (index, term) in sorted.iter() {
+                let Some(term) = term else {
+                    continue;
+                };
+                indices.push(*index);
+                bit_terms.push(*term);
+            }
+            boundaries.push(bit_terms.len());
+        }
+        Ok(Self {
```
Cryoris, line 1219: Does this call `SparseObservable::new` or does it directly set the values? If it calls `new`, should we call `new_unchecked` instead since we know the data is valid?

Cryoris (commented):
I'm not quite through with the first look, but there's some bits I don't 100% understand (like how `bytemuck` works) but overall this looks really nice and the the comments help a ton to understand it 🙂 I have mainly questions below for now and will have a look at the rest (and play with it) in a bit!

crates/accelerate/src/sparse_observable.rs
jakelishman: Yeah, technically it makes `IntEnum.+` possible as well, but `Enum` subclasses know about this possibility - you can use `[]` syntax to get those, like `BitTerm["+"]`.  You can also do `getattr(BitTerm, "+")` and it'll work as well - Python class attributes really are just strings, and just because you can't spell it as legal syntax in the attribute getter doesn't mean it's not possible to access.

But the "canonical" form of them (and the only one I properly documented) is the all-caps names forms.

crates/accelerate/src/sparse_observable.rs
jakelishman: The reason to have a bit of a split is that PyO3's `pyclass` derivations for Rust `enum`s don't work as pleasantly in Python as proper Python enumerations.  For example, making the Python-exposed class an `IntEnum` means that all its elements are subclasses of `int` properly, so you can put them in a Numpy array of `u8`, which means you can use the proper named access if you like, rather than having to deal with the magic numbers.

crates/accelerate/src/sparse_observable.rs
jakelishman: The intention is that you can use them when you're constructing raw Numpy arrays, or things like that, if you like.  We can teach in documentation that you can do `terms = SparseObservable.BitTerm` or something to shorten it, if we like?  My problem is that the name `BitTerm` is really generic, and there's a lot of different things in `quantum_info`, and I wanted it to be tied to `SparseObservable` in some form, just like `qiskit.circuit.classical` has `expr.Unary.Op` and `expr.Binary.Op` and whatnot.

crates/accelerate/src/sparse_observable.rs
jakelishman: Haha thanks - this is one of the direct motivating examples for the particular data format.

crates/accelerate/src/sparse_observable.rs
jakelishman: Within this docstring, which is the documentation of `SparseObservable`, the lookup `` :class:`SparseObservable` `` will always succeed with or without the leading dot - Sphinx will find it in the first/second place it looks, whereas the `.` affects where Sphinx starts looking after that.

It's similar to how the methods don't need the leading `.` - because they're in the `SparseObservable` namespace, they'll always be automatically found.

crates/accelerate/src/sparse_observable.rs
jakelishman: I don't think I actually tested it the other way round, but the `Args:` thing isn't a Sphinx built-in, it's a pre-processing step done by `sphinx.ext.napoleon`, which (I _think_) only acts as part of the processing of a function docstring.  Since this particular function isn't being pulled from a standalone docstring, but it's documented _very_ literally in Sphinx, I think I had to use the Sphinx-native syntax to do it.

Fwiw, this form is what Napoleon transforms the `Args:` stuff into.

crates/accelerate/src/sparse_observable.rs
jakelishman: Most specifically, it's to allow us to pass string literals in Rust space directly to the function.  So you can call `obs.add_dense_label("XYZ+rl")` and it'll do what you expect.

crates/accelerate/src/sparse_observable.rs
jakelishman: The default constructor in Python is super special.  For a static type object (almost all defined by an extension module, like this one), the default constructor gets placed as a function pointer into a slot in the type object's underlying C struct that's called `tp_new`.  The lookup rules in Python cause attribute-access to `__new__` to wrap that function-pointer slot into a callable Python method.

The `#[new]` attribute in PyO3 triggers that behaviour from C space.  It doesn't get attached as a proper Python method at all, so the `py_new` name is entirely arbitrary - that name doesn't get exposed to Python space.  That's also why you can't put the docstring on this method and have it work - it's not a proper Python method at all - though I think PyO3 are interested in modifying their docs-generation logic to concatenate the default-constructor docstring onto the end of the class docstring.

crates/accelerate/src/sparse_observable.rs
jakelishman: `num_qubits` doesn't have a defined setter from Python space, so you can't mutate it.

This check is because you can pass lots of things to the default constructor that have a very inherent number of qubits in them (like a `SparsePauliOp` or a `Pauli`).  We'd create incoherent data if we allowed the `num_qubits` to be modified away from the thing we're converting, so the test here just validates that the number matches, if it was explicitly given.

crates/accelerate/src/sparse_observable.rs
jakelishman: There's type-inference magic going on here.  In the first branch, the extraction type is `Vec<(String, Complex64)>`, and in the second branch, it's `Vec<(String, Vec<u32>, Complex64)>`.  The first branch will fail if the Python-space import is something like `[("XY", (2, 1), 1.0j)]`, but the second branch will succeed.

The empty list `[]` can succeed either extraction, but it'd have the same effect in both branches, so it doesn't matter that it's ambiguous (and there's a test that it works).

crates/accelerate/src/sparse_observable.rs
jakelishman: I'm happy to switch over to `num_terms` if you think that's better, and not distracting with the single-bit terms.  I personally _think_ about them as terms, I think I just got a bit weird about the naming.

crates/accelerate/src/sparse_observable.rs
jakelishman: You can think of the PyO3 `#[getter]` tag as being give-or-take the same thing as the Python-space `@property` decorator - it causes the function following it to be used for base attribute access, and doesn't define a setter.  In addition, though, PyO3 also automatically strips any `get_` prefix from the front of a getter - that's useful in Rust if you have both a getter _and_ a setter, because in Rust you can't name two methods the same thing.

I used the `get_` prefix here to avoid a naming conflict in Rust with the general `coeffs` field, and with a view that it feels quite likely that at some point in the future we'll make a `pub fn coeffs(&self) -> &[Complex64]` view method.

crates/accelerate/src/sparse_observable.rs
jakelishman: I'm fairly in favour of keeping the name as it is, but I'm happy to improve the documentation explaining the concept of the zero operator if it's not clear.  To me, `zeros` matches the Numpy terminology, and `empty` would mean an uninitialised operator.

(Also, personally "zero operator" to me means exactly what this function does - the object you named is what I'd call an "all-zeros projector" or something.)

crates/accelerate/src/sparse_observable.rs
jakelishman: Haha sure - I think I said above, personally I call this `terms` in my head.

crates/accelerate/src/sparse_observable.rs
jakelishman: Ha, thanks - that was completely deliberate.  I thought it's hard enough to understand what's going on with the sparsity as things are, so I wanted to be really crystal clear.

I documented somewhere that `from_list` is equivalent to `from_sparse_list` if all the coeffs are `reversed(range(num_qubits))` as well, so these two are the same:
```python
SparseObservable.from_list([("IXIZ", 1j), ("+-rl", 0.5)])
SparseObservable.from_sparse_list([("IXIZ", (3, 2, 1, 0), 1j), ("+-rl", (3, 2, 1, 0), 0.5)])
```
to try and drive the point home.

crates/accelerate/src/sparse_observable.rs
jakelishman: Yeah, a core purpose of a `BTreeMap` is that its iteration methods all return their results in sorted order.  It modifies the asymptotic complexities of a few actions away from a "pure" hashmap - it's kind of a pragmatic hybrid of a binary tree, a hashmap and a list.

crates/accelerate/src/sparse_observable.rs
jakelishman: This constructs the struct directly.  It's effectively the same as `new_unchecked` in this case, but `new_unchecked` was me being slightly forwards-looking to a time when `SparseObservable` is accessed and created from within Rust.  Since the fields of the struct aren't public (deliberately), anywhere outside this file wouldn't be allowed to use the `SparseObservable { .. }` constructor form, but I still wanted the zero-copy, zero-validation form to be available if the user promised to be safe about it.

crates/accelerate/src/sparse_observable.rs
Cryoris: Yeah that is actually a nice way of writing the sparse labels, maybe we should update the `SparsePauliOp` docstring to add this example 🤔

crates/accelerate/src/sparse_observable.rs
Cryoris: I see... yeah using an alias like that in the docs would be good enough I think, that's a nice solution to keep things precise but short 🙂

crates/accelerate/src/sparse_observable.rs
Cryoris: Ohh I see, right, so PyO3 `pyclass`es can pretty much only be standalone classes but not derive from a Python type like `enum.IntEnum`?

crates/accelerate/src/sparse_observable.rs
Cryoris: Ok, I see 👍🏻 and the support of both `"+"` and `"PLUS"` is just a convenience, or is it necessary? (Not asking to change it, just to understand 🙂 )

crates/accelerate/src/sparse_observable.rs
Cryoris: Wait what, a `&str` is a valid `&[u8]`? I mean yes it makes sense, but it sometimes surprises me what the compiler allows and what isn't 😄

crates/accelerate/src/sparse_observable.rs
Cryoris: Ok gotcha, that's indeed a bit special. Is `__new__` the only method that has this special behavior as it creates the object or are there other methods that have a special flow?

crates/accelerate/src/sparse_observable.rs
jakelishman: Just a convenience - I thought it might be useful to be able to do stuff like `[BitTerm[x] for x in "XrlYX+-01"]` potentially.

crates/accelerate/src/sparse_observable.rs
Cryoris: Ah ok because the type is inferred from how `vec` is used in the if-body, i.e. because `Self::py_from_list` expects `Vec<(String, Complex64)>`? That's kind of neat, but confusing to read without an IDE completion 😛

crates/accelerate/src/sparse_observable.rs
jakelishman: There are mechanisms to make them subclass each other, but I'm not sure about types defined from Python space (like `IntEnum`). Either way, they're really really fiddly to work with if you do it.

crates/accelerate/src/sparse_observable.rs
jakelishman: `String` is actually just a wrapper around a `Vec<u8>`. The references aren't exactly the same thing as each other, but every `&str` can be losslessly converted into a `&[u8]`. It's the `AsRef` trait implemention doing the work here.

crates/accelerate/src/sparse_observable.rs
jakelishman: Many of the magic methods have similar trickery about them, but PyO3 hides it behind the scenes for you. `__new__` has the added complication of being a static method, so needs an annotation anyway, I guess?

I don't know exactly why `new` gets the special tag and the others don't.

crates/accelerate/src/sparse_observable.rs
jakelishman: tbh, the `SparsePauliOp` documentation needs a whole overhaul - it's really unclear for how core the class is to the V1 and V2 primitives.  I was trying to get ahead of the game with this operator.  We also should overhaul the `quantum_info` module page in a similar vein, and much of the classes.  Abby and the docs team have got a bunch of open issues about that in #13236.

crates/accelerate/src/sparse_observable.rs
jakelishman: Yeah, exactly.

I don't actually use inline IDE type hints (I have it set so I can query it on demand, but I don't like them being in the way by default), so I was fine with it like this.  If you think it's too confusing, I can add a comment?

crates/accelerate/src/sparse_observable.rs
jakelishman: More comment added in 9dd0187213.

crates/accelerate/src/sparse_observable.rs
jakelishman: Done in 34a5490b9f (along with all other instances of "analagous", because apparently I just don't know how to spell that word).

crates/accelerate/src/sparse_observable.rs
jakelishman: Done in 4ca948eaa2.

crates/accelerate/src/sparse_observable.rs
jakelishman: Done in 4ca948eaa2.

crates/accelerate/src/sparse_observable.rs
jakelishman: Done in 34a5490b9f.

crates/accelerate/src/sparse_observable.rs
jakelishman: Done in 9dd0187213.

crates/accelerate/src/sparse_observable.rs
jakelishman: Done in 34a5490b9f.

crates/accelerate/src/sparse_observable.rs
jakelishman: Done in 9dd0187213.

crates/accelerate/src/sparse_observable.rs
jakelishman: Done in 9dd0187213.

crates/accelerate/src/sparse_observable.rs
jakelishman: Done in 34a5490b9f.

crates/accelerate/src/sparse_observable.rs
Cryoris: Ahh ok that's cool, I didn't know about the use of `AsRef` 😄

crates/accelerate/src/sparse_observable.rs
Cryoris: Yeah I think that would be helpful, if you don't mind! But I'm fine keeping it as is, I think from context it's clear what's happening I was just not aware type inference is that strong 😄

crates/accelerate/src/sparse_observable.rs
Cryoris: Oh right, that makes sense -- thanks for the info!

crates/accelerate/src/sparse_observable.rs
Cryoris: Yeah I guess strictly speaking what the all-zeros projector would rather be `zeros` and not `zero` 😄 To be extra clear I think it wouldn't hurt to be explicit in the docstring, maybe something like "Get the zero operator (i.e, without any terms)." ?

jakelishman: Done in 94ab2d6925.

crates/accelerate/src/sparse_observable.rs
jakelishman: Done in 94ab2d6925.

Cryoris (approved):
Thanks for the detailed comments and responses, this looks really nice (and I learned quite a bunch in reviewing 😄)!

PR #12675, opened 2024-06-28
Fix clippy warnings on latest stable rust

mtreinish:
### Summary

In Rust 1.79.0 several new clippy rules were added and/or enabled by default. This was causing some new issues to be flagged when building qiskit with the this release of Rust. This commit fixes these issues flagged by clippy.

### Details and comments

### DIFF
--- a/crates/circuit/src/circuit_instruction.rs
+++ b/crates/circuit/src/circuit_instruction.rs
@@ -825,10 +825,7 @@ pub(crate) fn convert_py_to_operation_type(
     };
     let op_type: Bound<PyType> = raw_op_type.into_bound(py);
     let mut standard: Option<StandardGate> = match op_type.getattr(attr) {
-        Ok(stdgate) => match stdgate.extract().ok() {
-            Some(gate) => gate,
-            None => None,
-        },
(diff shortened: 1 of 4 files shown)


crates/circuit/src/circuit_instruction.rs
```diff
@@ -825,10 +825,7 @@ pub(crate) fn convert_py_to_operation_type(
     };
     let op_type: Bound<PyType> = raw_op_type.into_bound(py);
     let mut standard: Option<StandardGate> = match op_type.getattr(attr) {
-        Ok(stdgate) => match stdgate.extract().ok() {
-            Some(gate) => gate,
-            None => None,
-        },
+        Ok(stdgate) => stdgate.extract().ok().unwrap_or_default(),
```
jakelishman, line 828: This looks like it's `stdgate.extract().ok().flatten()` to me, but this works fine.

jakelishman (approved):

crates/circuit/src/circuit_instruction.rs
mtreinish: Clippy didn't tell me to do that though :)

PR #12676, opened 2024-06-28
Suppress nonsense `DeprecationWarning` caused by `unittest`

jakelishman:
### Summary

`unittest.TestCase.assertWarns` in context-manager form has an awkward habit of querying the `__warningregistry__` attribute on every module in existence.  This interacts poorly with a Numpy 2 deprecation warning trigger for code that's attempting to import functions from modules that became private in Numpy 2, if a warning has previously been triggered out of `numpy.linalg._linalg`.

This simply suppresses that particular warning from the test suite.

### DIFF
--- a/constraints.txt
+++ b/constraints.txt
@@ -7,6 +7,10 @@ scipy<1.11; python_version<'3.12'
 # See https://github.com/Qiskit/qiskit/issues/12655 for current details.
 scipy==1.13.1; python_version=='3.12'
 
+# Rustworkx 0.15.0 contains a bug that breaks graphviz-related tests.
+# See https://github.com/Qiskit/rustworkx/pull/1229 for the fix.
+rustworkx==0.14.2
+
 # z3-solver from 4.12.3 onwards upped the minimum macOS API version for its
 # wheels to 11.7. The Azure VM images contain pre-built CPythons, of which at
 # least CPython 3.8 was compiled for an older macOS, so does not match a
--- a/test/utils/base.py
+++ b/test/utils/base.py
@@ -214,8 +214,7 @@ class QiskitTestCase(BaseQiskitTestCase):
         warnings.filterwarnings(
             "ignore",
             category=DeprecationWarning,
-            message="__warningregistry__",
-            module=r"numpy(\.\w+)*",
+            message=r".*numpy\.(\w+\.)*__warningregistry__",
         )
 
         # We only use pandas transitively through seaborn, so it's their responsibility to mark if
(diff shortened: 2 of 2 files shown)


constraints.txt
```diff
@@ -7,6 +7,10 @@ scipy<1.11; python_version<'3.12'
 # See https://github.com/Qiskit/qiskit/issues/12655 for current details.
 scipy==1.13.1; python_version=='3.12'
 
+# Rustworkx 0.15.0 contains a bug that breaks graphviz-related tests.
```
Cryoris, line 10: ~~Is this then unrelated to the NumPy deprecation warning?~~ Edit: Nvm just saw the offline discussion 🙂

Eric-Arellano (dismissed):
Consider creating tracking issues so that we don't forget to follow up on these things

constraints.txt
jakelishman: Yes, it's just that Rustworkx 0.14 wasn't compatible with Numpy 2 and Rustworkx 0.15 is, so the release caused us to start pulling in Numpy 2 in CI.  We didn't see the failures before because not only do you need Numpy 2, you also need to have the test suite generate a warning out of `np.linalg._linalg`, which we don't do on most platforms, but there are some about numerical-precision things on macOS ARM specifically, and we weren't running the Numpy 2 tests on macOS ARM previously because it wasn't available to us at the time.

test/utils/base.py
```diff
@@ -204,6 +204,19 @@ def setUpClass(cls):
         warnings.filterwarnings("error", category=DeprecationWarning)
         warnings.filterwarnings("error", category=QiskitWarning)
 
+        # Numpy 2 made a few new modules private, and have warnings that trigger if you try to
+        # access attributes that _would_ have existed.  Unfortunately, Python's `warnings` module
+        # adds a field called `__warningregistry__` to any module that triggers a warning, and
+        # `unittest.TestCase.assertWarns` then queries said fields on all existing modules.  On
+        # macOS ARM, we see some (we think harmless) warnings come out of `numpy.linalg._linalg` (a
+        # now-private module) during transpilation, which means that subsequent `assertWarns` calls
+        # can spuriously trick Numpy into sending out a nonsense `DeprecationWarning`.
```
ElePT, line 213:
```suggestion
        # can spuriously trick Numpy into sending out a nonsense `DeprecationWarning`.
        # Tracking issue: https://github.com/Qiskit/qiskit/issues/12679
```

ElePT (dismissed):
I created a tracking issue and added it to the comment! Let me know if it works.

test/utils/base.py
jakelishman: If you think this is valuable feel free to commit this. I mentioned in the comment over there: this is an entirely spurious warning and there's no real path to removal of this skip; it's simply an awkward interaction between two libraries not under our control, neither of which are likely to update any time soon, and we're not doing anything wrong.

The one thing that _might_ change the system Qiskit-side is that the only reason we see this in the first place is that macOS ARM transpiler tests trigger a couple of (seemingly harmless) `np.linalg._linalg` runtime warnings for mathematical precision things, and that's a necessary component of the `DeprecationWarning` triggering.  If we somehow got rid of those (which we don't have any path to - we don't really know what triggers them or why they happen), the warning would disappear here, though it'd be fragile to some other warning appearing later.

ElePT (approved):

Issue #9031, opened 2022-10-28
Update plot_gate_map (and related visualization functions) to leverage rustworkx.visualization.graphviz_draw for unknown coupling maps


mtreinish:
### What should we add?

Right now the `plot_gate_map()` visualization function (and similar functions like `plot_circuit_layout()` use rustworkx's `spring_layout()` function when there a backend is passed in and there isn't a hardcoded layout available and passes that to a mpl visualization. For moderate numbers of qubits this works reasonably well, but for large numbers of qubits the layout doesn't scale positions appropriately and the output is typically cluttered and not a useful visualization. Graphviz is particularly well suited for doing this type of graph visualization (as it is a specialized tool for doing graph visualization). So instead of trying to rebuild what we get from graphviz we should just call out to graphviz if it's available and leverage the rustwork function for graph visualization using graphivz to generate the visualization. The only potential issue is potentially `plot_error_map` which has histograms and color bars built using matplotlib and integrating the two different visualization tools might prove difficult.

Here is some example code I used to leverage `graphviz_draw()` to build a view like `plot_circuit_layout()`:

```python
from rustworkx.visualization import graphviz_draw

graph = target.build_coupling_map().graph
for node in graph.node_indices():
    graph[node] = node
    
for edge, triple in graph.edge_index_map().items():
    graph.update_edge_by_index(edge, (triple[0], triple[1]))

physical_bits = layout.get_physical_bits()
qubit_indices = {bit: index for index, bit in enumerate(linear_circuit.qubits)}
    
def color_node(node):
    if node in physical_bits:
        out_dict = {
            "label": str(qubit_indices[physical_bits[node]]),
            "color": "red",
            "fillcolor": "red",
            "style": "filled",
            "shape": "circle",
        }
    else:
        out_dict = {
            "label": "",
            "color": "blue",
            "fillcolor": "blue",
            "style": "filled",
            "shape": "circle",
        }
    return out_dict

def color_edge(edge):
    if all(x in physical_bits for x in edge):
        out_dict = {
            "color": "red",
            "fillcolor": "red",
        }
    else:
        out_dict = {
            "color": "blue",
            "fillcolor": "blue",
        }
    return out_dict
        
graphviz_draw(graph, method="neato", node_attr_fn=color_node, edge_attr_fn=color_edge)
```
which in this case output something like
[...]

PR #12680, opened 2024-06-28
Added coordinate map for Heron devices

quantumjim:
### Summary

Added a coordinate map, so that `plot_gate_map` can use the same layout the the IBM Quantum Platform, and not create one randomly.

### Details and comments

Image below shows the output for `from qiskit.visualization import plot_gate_map
plot_gate_map(backend, figsize=(10,10))`.
image


mtreinish:
I'm not a huge fan of continuing this pattern, we moved to using graphviz for graph visualization (see: https://github.com/Qiskit/qiskit/issues/9031) so we didn't have to continue hard coding magic layouts in the visualization functions moving forward. It's not a really good solution because it assumes that all backends prefixed with `ibm_` or `fake_` and n qubits want to use the same layout. Prior to #9031 being implemented it was not great but better than using the algorithmic layout from `rx.spring_layout()` (which is garbage for more than a handful of nodes) but we're in a better state now. If the graphviz layout is deficient in some way it might be a better approach if we can come up with a better algorithmic layout for this.

quantumjim:
I agree that `qubit_coordinates_map` is not the best way to do things and should disappear in a future release. But while it is in the code, shouldn't it reflect public devices?

PR #12692, opened 2024-06-28
Enable avoiding Python operation creation in transpiler

mtreinish:
### Summary

Since #12459 accessing `node.op` in the transpiler eagerly creates a Python object on access. This is because we now are no longer storing a Python object internally and we need to rebuild the object to return the python object as expected by the api. This is causing a significant performance regression because of the extra overhead. The longer term goal is to move as much of the performance critical passes to operate in rust which will eliminate this overhead. But in the meantime we can mitigate the performance overhead by changing the Python access patterns to avoid the operation object creation. This commit adds some new getter methods to DAGOpNode to give access to the inner rust data so that we can avoid the extra overhead. As a proof of concept this updates the unitary synthesis pass in isolation. Doing this fixes the regression caused by #12459 for that pass. We can continue this migration for everything else in follow up PRs. This commit is mostly to establish the pattern and add the python space access methods.

### Details and comments

### DIFF
--- a/crates/circuit/src/dag_node.rs
+++ b/crates/circuit/src/dag_node.rs
@@ -15,9 +15,11 @@ use crate::circuit_instruction::{
     ExtraInstructionAttributes,
 };
 use crate::operations::Operation;
+use numpy::IntoPyArray;
 use pyo3::prelude::*;
 use pyo3::types::{PyDict, PyList, PySequence, PyString, PyTuple};
-use pyo3::{intern, PyObject, PyResult};
+use pyo3::{intern, IntoPy, PyObject, PyResult};
+use smallvec::smallvec;
 
 /// Parent class for DAGOpNode, DAGInNode, and DAGOutNode.
 #[pyclass(module = "qiskit._accelerate.circuit", subclass)]
@@ -70,12 +72,19 @@ pub struct DAGOpNode {
 
 #[pymethods]
 impl DAGOpNode {
+    #[allow(clippy::too_many_arguments)]
     #[new]
+    #[pyo3(signature = (op, qargs=None, cargs=None, params=smallvec![], label=None, duration=None, unit=None, condition=None, dag=None))]
     fn new(
         py: Python,
-        op: PyObject,
+        op: crate::circuit_instruction::OperationInput,
         qargs: Option<&Bound<PySequence>>,
         cargs: Option<&Bound<PySequence>>,
+        params: smallvec::SmallVec<[crate::operations::Param; 3]>,
+        label: Option<String>,
+        duration: Option<PyObject>,
+        unit: Option<String>,
+        condition: Option<PyObject>,
         dag: Option<&Bound<PyAny>>,
     ) -> PyResult<(Self, DAGNode)> {
         let qargs =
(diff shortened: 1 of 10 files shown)


qiskit/transpiler/passes/synthesis/unitary_synthesis.py
```diff
@@ -32,16 +32,16 @@
...
 from qiskit.transpiler.passes.optimization.optimize_1q_decomposition import _possible_decomposers
 from qiskit.synthesis.two_qubit.xx_decompose import XXDecomposer, XXEmbodiments
 from qiskit.synthesis.two_qubit.two_qubit_decompose import (
     TwoQubitBasisDecomposer,
     TwoQubitWeylDecomposition,
-    GATE_NAME_MAP,
 )
 from qiskit.quantum_info import Operator
+from qiskit.circuit.controlflow import CONTROL_FLOW_OP_NAMES
 from qiskit.circuit import ControlFlowOp, Gate, Parameter
```
mtreinish:
```suggestion
from qiskit.circuit import Gate, Parameter
```

crates/circuit/src/dag_node.rs
```diff
@@ -110,34 +119,16 @@ impl DAGOpNode {
...
-                condition: res.condition,
-            }))
-        } else {
-            None
-        };
+        let mut instruction = CircuitInstruction::py_new(
+            py, op, None, None, params, label, duration, unit, condition,
+        )?;
+        instruction.qubits = qargs.into();
+        instruction.clbits = cargs.into();
```
ElePT, line 127: I might be missing something here, but why aren't  the qubits and clbits directly given to the `CircuitInstruction` constructor?

ElePT (commented):
I think that it's nice to have an intermediate solution until the passes are migrated to rust, but I wonder how do we want to approach the follow-ups. Do we want to try migrating as many python passes to not access `node.op` first, independently of when we plan to have a rust implementation for them? Or does it make sense to wait and see which passes we don't manage to migrate to rust in the near term and only update those?

crates/circuit/src/dag_node.rs
mtreinish: It's because of a type mismatch. The constructor for a circuit instruction takes in an `impl IntoIterator` while we have a `PyTuple` here which isn't doesn't implement the `IntoItertator` trait. I thought about collecting the tuples into a `Vec` so we could pass them to `new()`, but at the end of the day the `CircuitInstruction` has a `PyTuple` in it, so I just elected to do this to avoid needing any conversions or tweaks to the interfaces.

mtreinish:
> I think that it's nice to have an intermediate solution until the passes are migrated to rust, but I wonder how do we want to approach the follow-ups. Do we want to try migrating as many python passes to not access `node.op` first, independently of when we plan to have a rust implementation for them? Or does it make sense to wait and see which passes we don't manage to migrate to rust in the near term and only update those?

Yeah I was thinking of trying to do this first. At this point I'm a bit skeptical we'll get any passes ported before the 1.2 PR proposal freeze deadline in ~2 weeks given the current state of the dagcircuit PR. So I pushed this up to give us a way to mitigate the regression caused by #12459 in the short term for 1.2. But there are probably some passes that we can work with just `DAGOpNode`/`CircuitInstruction` from rust like what I did in #12650 and we should try to do that instead of updating the Python code in those cases when we encounter them.

ElePT (approved):
Sounds good. Approving!

Issue #12688, opened 2024-06-28
Add post-selection method to `BitArray`


ihincks:
### What should we add?

We should add a method that helps to filter shots by post-selection. The user should be able to make requests such as "I would like all of the shots where bits 15 and 18 are in the state `01`".

There is a complication to do with the shaped nature of BitArray: each index of the shape might result in a different number of post selections. For example, if the original BitArray has 100 shots, the example above might result in 57 post-selections on the first index, 17 on the second, and so forth. To get around this, I propose that the post-selection method always flattens the data. Alternate proposals welcome. For example, if a data BitArray has shape (2,3,4) and num_shots=20, then the post-selection specification would first flatten into a BitArray of shape () and num_shots=20*2*3*4, and apply the post selection there.

Proposed signature:
```python
def postselect(self, selection: BitArray, idxs: Iterable[int]) -> BitArray:
    """Post-select this bit array based on sliced equality with a given bitstring.

    .. note::
        If this bit array contains any shape axes, it is first flattened into a long list of shots before 
        applying post-selection. This is done because :class:`~BitArray` cannot handle ragged
        numbers of shots across axes.

    Args:
        selection: A bit array containing only one bitstring, specifying what to post-select on.
        idxs: A list of length matching ``selection``, specifying which bits of ``data`` it corresponds to.

    Returns:
        A new bit array with ``shape=(), num_bits=data.num_bits, num_shots<=data.num_shots``.

    Raises:
        ValueError: If ``selection`` has more bits than :attr:`num_bits``.
        ValueError: If the lengths of ``selection`` and ``idxs`` do not match.
        ValueError: If selection contains more than one bitstring.
    """

**Comment by ihincks** [comment]
cc @aeddins-ibm

**Comment by aeddins-ibm** [comment]
I wonder if we can make this easier for the user than the proposed signature above. Having `selection` be a `BitArray` means a user will need to understand how the `BitArray` class works, and also what convention `Sampler` uses when it returns a result. (Is it right that `self` is almost always going to be a `Sampler` result?)

Can we simplify the signature to more closely match this?
```
"I would like all of the shots where bits 15 and 18 are in the state 01"
```

The values of `idxs` can be the classical bit index in the sampled circuit, since users are already familiar with that. This matches the precedent of `BitArray.slice_bits()`, line 418-419:
[...]

PR #12693, opened 2024-06-28
Bitarray postselect

aeddins-ibm:
### Summary
Add method `BitArray.postselect()`, per issue #12688 .

[edit July 22, 2024: deleting outdated, distracting details from this description. See linked issue and the method docstring for more information.]


test/python/primitives/containers/test_bit_array.py
```diff
@@ -695,3 +695,214 @@ def test_expectation_values(self):
...
+                [45, 32],
+                [40, 1],
+                [18, 48],
+                [42, 15],
+                [25, 5],
+            ],
+            dtype=int,
+        )
+
+        np.random.seed(0)
```
t-imamichi: Thank you for implementing this new feature.
Could you use `Generator` (`np.random.default_rng`) instead of `RandomState` (`np.random`)?

> [Generator](https://numpy.org/doc/stable/reference/random/generator.html#numpy.random.Generator) and its associated infrastructure was introduced in NumPy version 1.17.0. There is still a lot of code that uses the older [RandomState](https://numpy.org/doc/stable/reference/random/legacy.html#numpy.random.RandomState) and the functions in [numpy.random](https://numpy.org/doc/stable/reference/random/index.html#module-numpy.random). While there are no plans to remove them at this time, we do recommend transitioning to [Generator](https://numpy.org/doc/stable/reference/random/generator.html#numpy.random.Generator) as you can.

https://numpy.org/doc/stable/reference/random/index.html#module-numpy.random

```suggestion
        rng = np.random.default_rng(0)
```

test/python/primitives/containers/test_bit_array.py
```diff
@@ -695,3 +695,214 @@ def test_expectation_values(self):
                 _ = ba.expectation_values("Z")
             with self.assertRaisesRegex(ValueError, "is not diagonal"):
                 _ = ba.expectation_values("X" * ba.num_bits)
+
+    def test_postselection(self):
```
ihincks, line 699: This test seems to be somewhat indirect for a unit test. It seems to boil down to making sure that `postselect` works for 0-d bit arrays, in which case I wonder if we couldn't just do something like

```python
# manually construct some bitstrings with num_bits > 16, maybe 21 or something.
arr = BitArray.from_bitstrings(
    [
        "01001"....<maybe 4 or 5 bitstrings>
    ]
)

# now test a handful of cases with post-select, manually.
```

I'd be most concerned about testing:
 - various indexes that straddle or don't straddle byte alignments
 - indices that are out of order
 - at least 1 multi-dim case

qiskit/primitives/containers/bit_array.py
```diff
@@ -464,6 +464,38 @@ def slice_shots(self, indices: int | Sequence[int]) -> "BitArray":
...
+        """Post-select this bit array based on sliced equality with a given bitstring.
+
+        .. note::
+            If this bit array contains any shape axes, it is first flattened into a long list of shots
+            before applying post-selection. This is done because :class:`~BitArray` cannot handle
+            ragged numbers of shots across axes.
+
+        Args:
+            indices: A list of the indices of the cbits on which to postselect.
+                This matches the indexing used by BitArray.slice_bits().
```
ihincks:
```suggestion
                This matches the indexing used by :meth:`~slice_bits`.
```

qiskit/primitives/containers/bit_array.py
```diff
@@ -464,6 +464,38 @@ def slice_shots(self, indices: int | Sequence[int]) -> "BitArray":
...
+                If this bit array was produced by a sampler, then an index ``i`` corresponds to the
+                :class:`~.ClassicalRegister` location ``creg[i]``.
+
+            selection: A list of bools of length matching ``indices``, with `indices[i]` corresponding
+              to `selection[i]`. Shots will be discarded unless all cbits specified by `indices` have
+              the values given by `selection`.
```
ihincks: nit: in ReST, single ticks are italics, double ticks are inline code.
```suggestion
            selection: A list of bools of length matching ``indices``, with ``indices[i]`` corresponding
              to ``selection[i]``. Shots will be discarded unless all cbits specified by ``indices`` have
              the values given by ``selection``.
```

ihincks (commented):
Thanks for opening the PR!

This change will also need a reno.

test/python/primitives/containers/test_bit_array.py
aeddins-ibm: Added more-focused test cases, postselecting on a multi-dimensional BitArray.

test/python/primitives/containers/test_bit_array.py
aeddins-ibm: Ended up replacing the randomized test with smaller ones without randomness.

aeddins-ibm:
The error in the latest automated run of **Tests / macOS-arm64-tests-Python-3.10 (pull_request)** seems to be [a known bug](https://github.com/Qiskit/qiskit/pull/12628#issuecomment-2186575922), and presumably (?) unrelated to `BitArray`. It ran successfully before the last lint commit. Do I need to run the test again somehow?

test/python/primitives/containers/test_bit_array.py
```diff
@@ -695,3 +695,69 @@ def test_expectation_values(self):
...
+            for name, indices, selection in test_cases:
+                with self.subTest(dataname + "_" + name):
+                    answer = bool_array[np.all(bool_array[:, indices] == selection, axis=-1)]
+                    postselected_bools = np.unpackbits(
+                        bit_array.postselect(indices, selection).array[:, ::-1],
+                        count=num_bits,
+                        axis=-1,
+                        bitorder="little",
+                    ).astype(bool)
+                    self.assertTrue((postselected_bools == answer).all())
```
t-imamichi: I would suggest `np.testing.assert_equal` because it is more informative when it fails.
```suggestion
                    np.testing.assert_equal(postselected_bools, answer)
```

test/python/primitives/containers/test_bit_array.py
```diff
@@ -695,3 +695,69 @@ def test_expectation_values(self):
...
+
+            num_bits = bool_array.shape[-1]
+            bool_array = bool_array.reshape(-1, num_bits)
+
+            test_cases = [
+                ("basic", [0, 1], [0, 0]),
+                ("multibyte", [0, 9], [0, 1]),
+                ("repeated", [5, 5, 5], [0, 0, 0]),
+                ("contradict", [5, 5, 5], [0, 0, 1]),
+                ("unsorted", [5, 0, 9, 3], [1, 0, 1, 0]),
```
t-imamichi: `selection` part of test cases is `list[int]` though the signature of `selection` of `postselect` is `Sequence[bool]`.
```python
    def postselect(self, indices: Sequence[int], selection: Sequence[bool]) -> BitArray:
```
Could you update the test cases such as `("basic", [0, 1], [False, False]),` or the signature of `postselect` such as `def postselect(self, ..., selection: Sequence[bool | int]`)?

test/python/primitives/containers/test_bit_array.py
ihincks: thanks

aeddins-ibm:
At the cost of moderate code complexity, there's a moderate speedup if we implement this method using bitwise operations.

Blue is the old method, orange is the new method, and green is the new method if the requested input is guaranteed not to contain any contradictions (e.g. the user promises not to postselect that bit-indices [5, 5] have values [0, 1]).

image

I think it's worthwhile so I'll commit the faster version here.

I first attempted to use bitwise operations to speed up `slice_bits()`, but I could not get a numpy implementation of `slice_bits()` that was competitive with just unpacking/repacking. I found postselection to be more compatible with numpy bitwise ops.

aeddins-ibm:
(deleted previous comment due to issues with those plots)

Will push a faster version of `postselect` that uses bitwise operations. The code is a little more complicated but I think it's worthwhile.

I didn't check the timings through all of parameter space, but here are a few sweeps using `timeit`. AFAICT the new version is always faster. Disclaimer that I haven't used `timeit` much before but believe this is correct.

The blue trace is the old method. Orange trace is new method. Green trace is new method if the user promises that the postselection condition does not contain any contradictions (e.g. select on bits [5, 5] being [0, 1]).

This plot shows times to postselect a random BitArray of 5000 shots and varying numbers of bits, where we post-select on half of them:
image

Like above, but we post-select on 10 bits in each case:
image

500 shots, post-select on 1 bit:
image

500 shots, post-select on half of the bits:
image

test/python/primitives/containers/test_bit_array.py
aeddins-ibm: Done, thanks.

test/python/primitives/containers/test_bit_array.py
aeddins-ibm: Yes, thank you! I think this actually fixed a major bug with how the tests were evaluating.

qiskit/primitives/containers/bit_array.py
```diff
@@ -464,6 +464,100 @@ def slice_shots(self, indices: int | Sequence[int]) -> "BitArray":
...
+                Negative indices are allowed.
+
+            selection: A list of bools of length matching ``indices``, with ``indices[i]`` corresponding
+              to ``selection[i]``. Shots will be discarded unless all cbits specified by ``indices`` have
+              the values given by ``selection``.
+
+            may_have_contradiction: If the args ``indices`` and ``selection`` are known not to contain
+              contradictions, one can set this to False for a moderate speed-up, depending on the problem
+              size. Here "contradiction" means requiring that a single bit has two different values, e.g.
+              ``indices=[5,5]`` and ``selection=[0,1]``. If a contradiction is present, then setting
```
t-imamichi:
```suggestion
              ``indices=[5,5]`` and ``selection=[False,True]``. If a contradiction is present, then setting
```

qiskit/primitives/containers/bit_array.py
```diff
@@ -464,6 +464,100 @@ def slice_shots(self, indices: int | Sequence[int]) -> "BitArray":
         arr = arr[..., indices, :]
         return BitArray(arr, self.num_bits)
 
+    def postselect(
+        self,
+        indices: Sequence[int] | int,
+        selection: Sequence[bool] | bool,
+        may_have_contradiction: bool = True,
```
t-imamichi: I would suggest requiring name `may_have_contradiction` always, i.e., disallow `ba.postselect([1], [True], False)` and it should be `ba.postselect([1], [True], may_have_contradiction=False)`. It is clear what option users turn off.
```suggestion
        *,
        may_have_contradiction: bool = True,
```

test/python/primitives/containers/test_bit_array.py
```diff
@@ -695,3 +695,84 @@ def test_expectation_values(self):
...
+                        if isinstance(indices, int):
+                            indices = (indices,)
+                        if isinstance(selection, bool):
+                            selection = (selection,)
+                        answer = bool_array[np.all(bool_array[:, indices] == selection, axis=-1)]
+                        if name == "contradict":
+                            if check_for_contradiction:
+                                np.testing.assert_equal(postselected_bools, answer)
+                        else:
+                            self.assertTrue(len(answer) > 0)  # avoiding trivial test case
```
t-imamichi:
```suggestion
                            self.assertGreater(len(answer), 0)  # avoiding trivial test case
```

qiskit/primitives/containers/bit_array.py
```diff
@@ -464,6 +464,100 @@ def slice_shots(self, indices: int | Sequence[int]) -> "BitArray":
...
+            return flattened
+
+        # Allow for negative bit indices:
+        is_negative = indices < 0
+        indices[is_negative] = np.mod(indices[is_negative], self.num_bits, dtype=int)
+
+        # Check for contradictory conditions:
+        if may_have_contradiction:
+            if np.intersect1d(indices[selection], indices[np.logical_not(selection)]).size > 0:
+                return BitArray(np.empty((0, num_bytes), dtype="uint8"), num_bits=self.num_bits)
```
t-imamichi: Is there any reason to specify dtype by string? I usually do it by `dtype=np.uint8` because it is easier for IDE to autocomplete and verify it.

qiskit/primitives/containers/bit_array.py
```diff
@@ -464,6 +464,100 @@ def slice_shots(self, indices: int | Sequence[int]) -> "BitArray":
...
+
+        num_indices = len(indices)
+
+        if len(selection) != num_indices:
+            raise ValueError("Lengths of indices and selection do not match.")
+
+        num_bytes = self._array.shape[-1]
+        indices = np.asarray(indices)
+
+        if num_indices > 0 and np.max(indices) >= self.num_bits:
```
t-imamichi: Could you add a check of the min side, i.e., `np.min(indices) < 0` → ValueError?

qiskit/primitives/containers/bit_array.py
aeddins-ibm: Thanks, that's good to know! I'll change the strings to `np.uint8`.

qiskit/primitives/containers/bit_array.py
```diff
@@ -464,6 +464,100 @@ def slice_shots(self, indices: int | Sequence[int]) -> "BitArray":
...
+            indices: A list of the indices of the cbits on which to postselect.
+                If this bit array was produced by a sampler, then an index ``i`` corresponds to the
+                :class:`~.ClassicalRegister` location ``creg[i]`` (as in :meth:`~slice_bits`).
+                Negative indices are allowed.
+
+            selection: A list of bools of length matching ``indices``, with ``indices[i]`` corresponding
+              to ``selection[i]``. Shots will be discarded unless all cbits specified by ``indices`` have
+              the values given by ``selection``.
+
+            may_have_contradiction: If the args ``indices`` and ``selection`` are known not to contain
```
ihincks: I just noticed this addition. I'm wondering what the use case is for duplicating index values? It seems to me that the check

```len(set(indices)) == len(indices)``

is so cheap compared to the rest of the work done by this function that it's not worth skimping on, so we should just always do it.

Comment (aeddins-ibm) [reply qiskit/primitives/containers/bit_array.py]: I just wasn't sure if the loss in generality could be a problem. A repeated index seems like a mathematically valid input with well-defined results, so not supporting it could surprise users. It also more closely matches how other arrays behave, like in this example selecting all rows where index 0 has value 1:
```
a = np.array([
    [0,1,2],
    [1,2,3]])
indices = [0,0,0,0,0]
selection = [1,1,1,1,1]
a[np.all(a[:,indices] == selection, axis=-1)]
```
```
array([[1, 2, 3]])
```

For a use case, maybe this ends up as a subroutine buried in some complex codebase, where a list of postselection conditions is generated. Maybe the validities of two different data bits both hinge on the same ancilla bit being 0, leading to that ancilla bit's index being repeated in that list of conditions. It would be best if the user could just pass the list of conditions to `postselect`, without having to first manually detect and handle the special cases of duplicate or contradictory conditions.

Comment (aeddins-ibm) [reply qiskit/primitives/containers/bit_array.py]: (Though I am definitely on the fence about just removing the `may_have_contradiction` argument, essentially fixing it to always be True, since it's complicated and so far the benefit seems marginal.)

File: qiskit/primitives/containers/bit_array.py (hunk 11 above)
Comment (aeddins-ibm) [reply qiskit/primitives/containers/bit_array.py]: In a recent commit I added support for negative indices -- i.e., index -1 will get the last bit (most-significant bit). Sorry for the confusion.

This is a break from the current behavior of `slice_bits`, but I felt it would be nice to have, and in the spirit of this class behaving like a numpy array. If there are reasons not to support negative indices I'm open to reverting the change.

Comment (aeddins-ibm) [reply qiskit/primitives/containers/bit_array.py]: https://github.com/Qiskit/qiskit/blob/17ac5ecaf7498f0b9e8e200ca4adae76105421df/qiskit/primitives/containers/bit_array.py#L481-L484

https://github.com/Qiskit/qiskit/blob/17ac5ecaf7498f0b9e8e200ca4adae76105421df/qiskit/primitives/containers/bit_array.py#L529-L531

File: qiskit/primitives/containers/bit_array.py (hunk 12 above)
Comment (ihincks) [reply qiskit/primitives/containers/bit_array.py]: Thanks. If it helps, I think you should fall off the fence, with a nudge toward "fix it at True". Like you say, since it's well defined to just return an empty array whenever the conditions are impossible to satisfy.

File: qiskit/primitives/containers/bit_array.py (hunk 7 above)
Comment (aeddins-ibm) [reply qiskit/primitives/containers/bit_array.py]: I have now simplified this optional argument and its description in the docstring. I changed it from `may_have_contradiction` to `assume_unique`, which I hope is easier for a user to interpret. This also avoids this formatting issue of int vs bool.

Comment (aeddins-ibm) [reply qiskit/primitives/containers/bit_array.py]: Actually per Ian's suggestion, I just removed this optional argument altogether, as the benefits did not seem worth the complexity.

File: qiskit/primitives/containers/bit_array.py (hunk 12 above)
Comment (aeddins-ibm) [reply qiskit/primitives/containers/bit_array.py]: I just pushed a (second) commit that removes this argument.

File: qiskit/primitives/containers/bit_array.py [hunk 13]
```diff
@@ -464,6 +464,91 @@ def slice_shots(self, indices: int | Sequence[int]) -> "BitArray":
...
+
+        flattened = self.reshape((), self.size * self.num_shots)
+
+        # If no conditions, keep all data, but flatten as promised:
+        if num_indices == 0:
+            return flattened
+
+        # Allow for negative bit indices:
+        is_negative = indices < 0
+        indices[is_negative] = np.mod(indices[is_negative], self.num_bits, dtype=int)
```
t-imamichi: I'm fine to allow negative indices. But I suggest raising IndexError if indices are too small instead of applying mod because numpy ndarray behaves as follows. Moreover, this PR does not apply mod to large indices.

```python
import numpy as np
a = np.array([0, 1, 2])
print(a[-1])  # 2
print(a[-2])  # 1
print(a[-3])  # 0
print(a[-4])  # IndexError: index -4 is out of bounds for axis 0 with size 3
```

qiskit/primitives/containers/bit_array.py
```diff
@@ -464,6 +464,91 @@ def slice_shots(self, indices: int | Sequence[int]) -> "BitArray":
...
+        num_indices = len(indices)
+
+        if len(selection) != num_indices:
+            raise ValueError("Lengths of indices and selection do not match.")
+
+        num_bytes = self._array.shape[-1]
+        indices = np.asarray(indices)
+
+        if num_indices > 0 and np.max(indices) >= self.num_bits:
+            raise ValueError(
```
t-imamichi: IndexError might be better perhaps.

qiskit/primitives/containers/bit_array.py
t-imamichi: Thank you for the details. I agree to allow negative indices. But I think it's consistent to raise an error for too small indices because it raises an error for too large indices at line 512.

qiskit/primitives/containers/bit_array.py
aeddins-ibm: Good point, thanks! Just made this change.

qiskit/primitives/containers/bit_array.py
aeddins-ibm: Agreed. Just updated to use IndexError.

Seems out of scope for this PR but currently slice_bits() raises a ValueError, if that matters:
https://github.com/Qiskit/qiskit/blob/3e02934f113c33a95bdb0e45c3e45a3777574794/qiskit/primitives/containers/bit_array.py#L432-L435

qiskit/primitives/containers/bit_array.py
aeddins-ibm: > I suggest raising IndexError if indices are too small instead of applying mod

To clarify, after checking indices are in-range, the code still applies mod because it simplifies the check for contradictory postselection conditions. Handles the (weird) case where the same bit is indexed once with a negative index and once with a positive index, where one has a 0 and the other a 1 in `selection`.

aeddins-ibm: ^ Just added a test for this weird case.

qiskit/primitives/containers/bit_array.py
t-imamichi: Thanks. slice_bits should be updated.

aeddins-ibm: I don't know what's normally done, should that be a separate PR since it's technically off-topic, or included in this PR since it's a tiny change? (and if the latter, should it be added as a bug-fix release note?)

I went ahead and committed the change to this PR for the sake of time.

aeddins-ibm: Made the same change in `slice_shots` (replaced ValueError with IndexError).

There are other places in `bit_array.py` where I'm not sure whether it should be ValueError or IndexError. OK to leave those for now?

t-imamichi: It would be nice to make as separate PR because we may have discussions on the change.
I also found that there are other methods involved in `indices`. So, the PR could be a bit large.
Could you revert the change of `slice_bits`? It would be nice to focus on postselection in this PR.
I will make a separate PR to fix both `slice_bits` and `slice_shots`.

t-imamichi:
LGTM. I made a separate PR #12755 to update slice_bits, slice_shots, etc. I will review this PR again after reverting some changes of slice_bits.

qiskit/primitives/containers/bit_array.py
aeddins-ibm: That sounds good, I have reverted the changes to `slice_bits` and `slice_shots` here.

qiskit/primitives/containers/bit_array.py
```diff
@@ -464,6 +464,96 @@ def slice_shots(self, indices: int | Sequence[int]) -> "BitArray":
...
+            selection: A list of bools of length matching ``indices``, with ``indices[i]`` corresponding
+              to ``selection[i]``. Shots will be discarded unless all cbits specified by ``indices`` have
+              the values given by ``selection``.
+
+        Returns:
+            A new bit array with ``shape=(), num_bits=data.num_bits, num_shots<=data.num_shots``.
+
+        Raises:
+            IndexError: If ``max(indices)`` is greater than or equal to :attr:`num_bits``.
+            IndexError: If ``min(indices)`` is less than negative :attr:`num_bits``.
```
t-imamichi:
```suggestion
            IndexError: If ``max(indices)`` is greater than or equal to :attr:`num_bits`.
            IndexError: If ``min(indices)`` is less than negative :attr:`num_bits`.
```
image

qiskit/primitives/containers/bit_array.py
```diff
@@ -464,6 +464,96 @@ def slice_shots(self, indices: int | Sequence[int]) -> "BitArray":
         arr = arr[..., indices, :]
         return BitArray(arr, self.num_bits)
 
+    def postselect(
+        self,
+        indices: Sequence[int] | int,
+        selection: Sequence[bool] | bool,
```
t-imamichi: I played around `postselct` method and felt that `selection` as `Sequence[int] | int` might be more intuitive because bitstrings are displayed with 0 or 1.

```python
ba = BitArray.from_counts({'111': 2, '000': 3})
print(ba.postselect(1, True).get_counts())  # the current signature expects this
# {'110': 2}
print(ba.postselect(1, 1).get_counts())    # this might be more intuitive
# {'110': 2}
```

I checked a similar method `QuantumCiruict.if_test` and found that it takes `int` as condition.
https://github.com/Qiskit/qiskit/blob/99032fc042b5f4f57956f451578c3476a672ddb9/qiskit/circuit/quantumcircuit.py#L6243-L6245

What do you think of it from your use case? Is bool more useful than int? @aeddins-ibm?

releasenotes/notes/bitarray-postselect-659b8f7801ccaa60.yaml
```diff
@@ -0,0 +1,4 @@
+---
+features_primitives:
+  - |
+    Added a new method :meth:`.BitArray.postselect` that returns all shots containing specified bit values.
```
t-imamichi, line 4: It would be nice to have a simple example in reno because the unit test looks a bit complicated.

E.g.,
```python

ba = BitArray.from_counts({'111': 2, '000': 3})
print(ba.postselect(1, True).get_counts())
# {'111': 2}
```
It depends on the discussion of `selection` type.
https://github.com/Qiskit/qiskit/pull/12693/files#r1675377387

qiskit/primitives/containers/bit_array.py
aeddins-ibm: Yes, that suggested change makes sense to me.

I did like how the typehint `bool` helped distinguish the roles of the two arguments. But I agree that users should input 0 and 1.

`Sequence[int] | int` sounds good. I'll confirm that then update the typehint/docstring/tests to match.

If we specify `int` instead of `bool`, maybe we should raise an error if any values are not 0 or 1.

releasenotes/notes/bitarray-postselect-659b8f7801ccaa60.yaml
aeddins-ibm: Added, with small adjustments.

(Looking at other release notes, I assumed the format to include a code block is to end the preceding line with `::` and then indent the code block).

qiskit/primitives/containers/bit_array.py
aeddins-ibm: Ended up going with `Sequence[bool | int] | bool | int`, and just stating in the docstring that `selection` should be a list of binary values, and that these values will be cast to type `bool`. I hope this is sufficiently clear that a user should not normally enter a value of `2`, and if they do then it will be cast to `True` as usual.

aeddins-ibm:
Before merging and committing to this function signature, should we consider if there are more general postselection conditions we could easily support?

For example, a plausible use case is someone wants all shots where, for all even k, `cbits[k] != cbits[k+1]`.

--
Update: I think this is probably not necessary, at least not urgent. For N values of k, the above example case could be resolved by calling `postselect` 2N times, which isn't great but is much better than the exponential blowup that I was initially imagining. (I was incorrectly concerned one would have to call postselect 2^N times, once for each of the possible 2N-qubit bitstrings. But one only has to call `postselect` 2N times, with `selection=[0, 1]` and `selection=[1,0]` for each value of k.)

releasenotes/notes/bitarray-postselect-659b8f7801ccaa60.yaml
t-imamichi: Thanks. reno looks good.

qiskit/primitives/containers/bit_array.py
t-imamichi: Thank you for updating the code. Looks good about this change.

t-imamichi:
If you have any use case of  `cbits[k] != cbits[k+1]` in mind, we have a little time left to discuss the extension (before 1.2 release).
If it's not urgent, I agree to merge this PR. This PR is LGTM already.

aeddins-ibm:
I think it's OK to merge as-is. The use-case I imagined above can at least be done in a linear number of calls to this function, which seems tolerable for now.

There are other hypothetical use cases that would require an exponential number of calls AFAICT. Like if we wanted only those shots where there are an odd number of 1's (odd parity) in a set of N bits, for large N. Then I think one would need to call `postselect` an exponential number of times to pick out each of the (2^N)/2 odd-parity bitstrings individually. This is not a fundamental limitation -- `postselect` could keep all shots where `bitwise_xor.reduce` of the selected bits gives 1. But I don't see a natural way to generalize the `postselect` interface, and likewise don't see an obvious place to stop generalizing the logic (if we add `xor`, do we also need to add `xnor`? And `any`? And...).

So I think it's better to leave `postselect` doing just the one thing it already does.

Someday, if we want to add support for e.g. selecting shots with a certain parity on many bits, we might instead add a separate method `BitArray.parity(indices)` that returns the parity of the specified bits for each shot, and then the user can use that to create a boolean mask to keep shots with the desired parity. (There may be overlap with code to compute expectation values). This would require modifying `slice_shots` to accept a boolean mask, which at a glance looks straightforward.

Frankly, we could also use this approach instead of having a `postselect` method at all, except that currently `postselect` should be faster than `slice_bits` due to the bit unpacking/repacking. I was not able to make `slice_bits` faster via numpy bitwise operations, but maybe a Rust version of `slice_bits` would be faster, at which point `postselect` could optionally be deprecated (or not) in favor of doing all types of postselection by creating a boolean mask along the shots axis.

Anyway, that's all to say I agree this PR is good to merge. Thanks for your patient reviewing!

t-imamichi (dismissed):
Thank you for your contributions. LGTM.

Post-selection with parity sounds an interesting use case. I agree to think of a separate method to realize it.

aeddins-ibm:
In case it's any use as a future reference, here is one approach to compute parity of specified bits.

This snippet uses `bitmask` as defined in the `postselect` code. It first uses `bitwise_xor` to collapse together all bytes in the bits axis, then broadcasts the remaining 8 bits to a byte each to take their xor.
```
shot_mask = np.bitwise_xor.reduce(flattened._array & bitmask, axis=-1, dtype=np.uint8, keepdims=True)
shot_mask = np.logical_xor.reduce(shot_mask & (np.uint8(1) << np.arange(8, dtype=np.uint8)), axis=-1, dtype=bool)
```

test/python/primitives/containers/test_bit_array.py
```diff
@@ -738,7 +738,7 @@ def test_postselection(self):
                 ("basic", [0, 1], [0, 0]),
                 ("multibyte", [0, 9], [0, 1]),
                 ("repeated", [5, 5, 5], [0, 0, 0]),
-                ("contradict", [5, 5, 5], [0, 0, 0]),
+                ("contradict", [5, 5, 5], [1, 0, 0]),
```
t-imamichi: I'm curious why the test passes even if test case was wrong. Should we update the test to catch the wrong case?

test/python/primitives/containers/test_bit_array.py
```diff
@@ -695,3 +695,80 @@ def test_expectation_values(self):
...
+                        axis=-1,
+                        bitorder="little",
+                    ).astype(bool)
+                    if isinstance(indices, int):
+                        indices = (indices,)
+                    if isinstance(selection, bool):
+                        selection = (selection,)
+                    answer = bool_array[np.all(bool_array[:, indices] == selection, axis=-1)]
+                    if name not in ["contradict", "negcontradict"]:
+                        self.assertGreater(len(answer), 0)  # avoiding trivial test case
```
t-imamichi: How about the following test?
```suggestion
                    if name in ["contradict", "negcontradict"]:
                        self.assertEqual(len(answer), 0)
                    else:
                        self.assertGreater(len(answer), 0)                    
```

If the test case was wrong `("contradict", [5, 5, 5], [0, 0, 0])`, it can catch it as follows.

```
FAIL: test_postselection (test.python.primitives.containers.test_bit_array.BitArrayTestCase.test_postselection)
test.python.primitives.containers.test_bit_array.BitArrayTestCase.test_postselection
----------------------------------------------------------------------
testtools.testresult.real._StringException: Traceback (most recent call last):
  File "/Users/ima/tasks/4_2024/qiskit/terra/test/python/primitives/containers/test_bit_array.py", line 763, in test_postselection
    self.assertEqual(len(answer), 0)  # avoiding trivial test case
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/homebrew/Cellar/python@3.12/3.12.4/Frameworks/Python.framework/Versions/3.12/lib/python3.12/unittest/case.py", line 885, in assertEqual
    assertion_func(first, second, msg=msg)
  File "/opt/homebrew/Cellar/python@3.12/3.12.4/Frameworks/Python.framework/Versions/3.12/lib/python3.12/unittest/case.py", line 878, in _baseAssertEqual
    raise self.failureException(msg)
AssertionError: 1 != 0
```

test/python/primitives/containers/test_bit_array.py
t-imamichi: I made a suggestion to catch the wrong test case. Could you take a look at it?
https://github.com/Qiskit/qiskit/pull/12693#discussion_r1688011473

aeddins-ibm: Thank you, I applied the suggested change.

The reason the test passed, as you figured out, is that the typo was in the problem definition. `BitArray.postselect()` solved that incorrect problem correctly, and verified it against the answer obtained for the same problem without using `BitArray.postselect()`. But the problem itself did not match the label. In a way, the new condition now also checks that the problem matches the string label.

t-imamichi: Thank you

t-imamichi (approved):
LGTM

t-imamichi:
Could you review this again, @ihincks? I would like to merge it if there is no problem.

t-imamichi:
Since there is no comment from @ihincks, I will merge this PR once CI gets recovered.

ElePT:
Neko won't block the merge queue so I think we can already add the PR to the queue :)

PR #12695, opened 2024-06-30
Oxidize random clifford

### LINKED ISSUE #12696: Port `random_clifford` to Rust
### What should we add?

In the theme of moving code to Rust, we should port the `random_clifford` function to Rust space.

alexanderivrii:
### Summary

Closes #12696.

This PR implements the function `random_clifford` in Rust. The Rust function is called `random_clifford_tableau`. 

This is a close port of the python function, with the exception that (1) optimizations for low-rank matrices were not yet included, (2) some of the numpy optimizations were replaced by the simpler approach presented in the appendix of the original paper.

UPDATE: 

By far the main bottleneck of the process is the (already existing) Rust function `binary_matmul_inner` for multiplying binary matrices. With the older implementation of this function, the Rust code happens to be _slower_ than the Python code, likely because `numpy` does some clever optimizations.

However, parallelizing `binary_matmul_inner` makes things better; finally, the Rust code is about 3x faster than Python code.

Here are the detailed results for generating 50 random Cliffords for different numbers of qubits:

```
PYTHON CODE: num_qubits = 5: time elapsed: 0.04
RUST CODE: num_qubits = 5: time elapsed: 0.00
PAR PYTHON CODE: num_qubits = 5: time elapsed: 0.03
=============================================================
PYTHON CODE: num_qubits = 10: time elapsed: 0.06
RUST CODE: num_qubits = 10: time elapsed: 0.00
PAR RUST CODE: num_qubits = 10: time elapsed: 0.00
=============================================================
PYTHON CODE: num_qubits = 50: time elapsed: 0.30
RUST CODE: num_qubits = 50: time elapsed: 0.11
PAR RUST CODE: num_qubits = 50: time elapsed: 0.05
=============================================================
PYTHON CODE: num_qubits = 100: time elapsed: 1.23
RUST CODE: num_qubits = 100: time elapsed: 0.88
PAR RUST CODE: num_qubits = 100: time elapsed: 0.27
=============================================================
PYTHON CODE: num_qubits = 150: time elapsed: 3.02
RUST CODE: num_qubits = 150: time elapsed: 2.86
PAR RUST CODE: num_qubits = 150: time elapsed: 0.74
=============================================================
PYTHON CODE: num_qubits = 200: time elapsed: 6.46
RUST CODE: num_qubits = 200: time elapsed: 6.92
PAR RUST CODE: num_qubits = 200: time elapsed: 1.78
=============================================================
PYTHON CODE: num_qubits = 250: time elapsed: 12.03
RUST CODE: num_qubits = 250: time elapsed: 14.30
PAR RUST CODE: num_qubits = 250: time elapsed: 3.54
=============================================================
PYTHON CODE: num_qubits = 300: time elapsed: 20.58
RUST CODE: num_qubits = 300: time elapsed: 26.63
PAR RUST CODE: num_qubits = 300: time elapsed: 6.37
=============================================================
PYTHON CODE: num_qubits = 350: time elapsed: 34.84
RUST CODE: num_qubits = 350: time elapsed: 43.99
PAR RUST CODE: num_qubits = 350: time elapsed: 10.45
=============================================================
PYTHON CODE: num_qubits = 400: time elapsed: 58.67
RUST CODE: num_qubits = 400: time elapsed: 67.97
PAR RUST CODE: num_qubits = 400: time elapsed: 15.98
=============================================================
PYTHON CODE: num_qubits = 450: time elapsed: 73.17
RUST CODE: num_qubits = 450: time elapsed: 109.03
PAR RUST CODE: num_qubits = 450: time elapsed: 26.85
```


ShellyGarion (commented):
LGTM. the code looks ok but is there a way to check the randomness or at least that assuming the same random choice, we get the same output as in the python code?

ShellyGarion:
How much this code improves when you port it from python to rust?

crates/accelerate/src/synthesis/clifford/mod.rs
```diff
@@ -43,6 +44,28 @@ fn synth_clifford_greedy(py: Python, clifford: PyReadonlyArray2<bool>) -> PyResu
     CircuitData::from_standard_gates(py, num_qubits as u32, clifford_gates, Param::Float(0.0))
 }
 
+#[pyfunction]
+#[pyo3(signature = (num_qubits, seed=None))]
```
Cryoris: More a question than a comment: Don't we typically put the docstring before the signature? Just to be consistent with the rest 🙂

crates/accelerate/src/synthesis/linear/utils.rs
```diff
@@ -10,9 +10,15 @@
...
-use ndarray::{concatenate, s, Array2, ArrayView2, ArrayViewMut2, Axis};
+use ndarray::{azip, concatenate, s, Array2, ArrayView1, ArrayView2, ArrayViewMut2, Axis, Zip};
 use rand::{Rng, SeedableRng};
 use rand_pcg::Pcg64Mcg;
+use rayon::iter::{IndexedParallelIterator, ParallelIterator};
+use rayon::prelude::IntoParallelIterator;
+
+use crate::getenv_use_multiple_threads;
+
+const PARALLEL_THRESHOLD: usize = 10;
```
Cryoris, line 23: Would it be helpful to add a comment on this number? I assume it's a heuristic from when the parallel execution is actually faster? 🙂

crates/accelerate/src/synthesis/clifford/random_clifford.rs
```diff
@@ -0,0 +1,163 @@
...
+        let k = if index < m { index } else { 2 * m - index - 1 };
+        perm[i] = inds[k];
+        inds.remove(k);
+    }
+    (had, perm)
+}
+
+/// Add symmetric random boolean value to off diagonal entries.
+fn fill_tril(mut mat: ArrayViewMut2<bool>, rng: &mut Pcg64Mcg, symmetric: bool) {
+    // todo: implement & benchmark optimizations for low-rank matrices
```
Cryoris: What kind of optimizations are possible for low-rank matrices?

crates/accelerate/src/synthesis/clifford/random_clifford.rs
```diff
@@ -0,0 +1,163 @@
...
+
+    for i in 0..n {
+        let m = n - i;
+        let eps: f64 = 4f64.powi(-(m as i32));
+        let r: f64 = rng.gen();
+        let index: usize = -((r + (1f64 - r) * eps).log2().ceil() as isize) as usize;
+        had[i] = index < m;
+        let k = if index < m { index } else { 2 * m - index - 1 };
+        perm[i] = inds[k];
+        inds.remove(k);
```
Cryoris, line 37: Can this ever run into the problem that it's trying to access an entry that has been removed? Maybe it's obvious that it cannot but I don't quite see it directly 🤔

crates/accelerate/src/synthesis/clifford/random_clifford.rs
```diff
@@ -0,0 +1,163 @@
...
+    binary_matmul_inner, calc_inverse_matrix_inner, replace_row_inner, swap_rows_inner,
+};
+use ndarray::{concatenate, s, Array1, Array2, ArrayView2, ArrayViewMut2, Axis};
+use rand::{Rng, SeedableRng};
+use rand_pcg::Pcg64Mcg;
+
+/// Sample from the quantum Mallows distribution.
+fn sample_qmallows(n: usize, rng: &mut Pcg64Mcg) -> (Array1<bool>, Array1<usize>) {
+    // Hadamard layer
+    let mut had = Array1::from_shape_fn(n, |_| false);
```
Cryoris: Optional change, but would `from_elem(n, false)` be slightly cleaner? (I only know about this nice constructor since you used it below 🙂)

crates/accelerate/src/synthesis/clifford/random_clifford.rs
```diff
@@ -0,0 +1,163 @@
...
+
+    let mut delta1: Array2<bool> = Array2::from_shape_fn((num_qubits, num_qubits), |(i, j)| i == j);
+    fill_tril(delta1.view_mut(), &mut rng, false);
+
+    let mut delta2: Array2<bool> = Array2::from_shape_fn((num_qubits, num_qubits), |(i, j)| i == j);
+    fill_tril(delta2.view_mut(), &mut rng, false);
+
+    // Compute stabilizer table
+    let zero = Array2::from_elem((num_qubits, num_qubits), false);
+    let prod1 = binary_matmul_inner(gamma1.view(), delta1.view()).unwrap();
```
Cryoris, line 95: If you're looking for more improvements, we could have a custom multiplication that takes into account that the upper triangular part of delta is always false (or by adding a flag to the `binary_matmul_inner`) 🤔

crates/accelerate/src/synthesis/clifford/random_clifford.rs
```diff
@@ -0,0 +1,163 @@
...
+    for i in 0..num_qubits {
+        gamma1[[i, i]] = rng.gen();
+    }
+    fill_tril(gamma1.view_mut(), &mut rng, true);
+
+    let mut gamma2: Array2<bool> = Array2::from_elem((num_qubits, num_qubits), false);
+    for i in 0..num_qubits {
+        gamma2[[i, i]] = rng.gen();
+    }
+    fill_tril(gamma2.view_mut(), &mut rng, true);
```
Cryoris, line 85: Another idea for improvement would be to have `fill_tril` do the random diagonal, and avoid constructing the matrix just to rewrite almost all entries again 🙂 E.g.
```
fn get_tril(rng: ... , symmetric: bool, random_diag: bool) -> Array2<bool>
```
where `random_diag=false` sets the diagonal to `true` (for the `delta` matrices) and `random_diag=true` creates the random diagonal for the `gamma` matrices

crates/accelerate/src/synthesis/linear/utils.rs
```diff
@@ -30,11 +36,29 @@ pub fn binary_matmul_inner(
         ));
     }
 
-    Ok(Array2::from_shape_fn((n1_rows, n2_cols), |(i, j)| {
-        (0..n2_rows)
-            .map(|k| mat1[[i, k]] & mat2[[k, j]])
-            .fold(false, |acc, v| acc ^ v)
-    }))
+    let run_in_parallel = getenv_use_multiple_threads();
```
Cryoris, line 41: I'm not very familiar with this part yet -- is this `true` per default for users or is this something we have to set manually in a config?

Cryoris (commented):
Nice that it's faster in the parallel version 👍🏻 👍🏻 Just curious: did you run a profiling to see which parts of the Rust code are the slowest? 🙂

crates/accelerate/src/synthesis/clifford/mod.rs
alexanderivrii: Indeed :). Fixed in [a020dc1](https://github.com/Qiskit/qiskit/pull/12695/commits/a020dc1191e3abdd979f95b510fc14f0aa5c9778).

crates/accelerate/src/synthesis/linear/utils.rs
alexanderivrii: This code for parallel execution mimics the code in `pauli_exp_val.rs`, though there the threshold is chosen to be `19`. Indeed, I have chosen `10` based on some local experiments.

crates/accelerate/src/synthesis/clifford/random_clifford.rs
alexanderivrii: The original python code (which I forgot to delete in the version of the PR that you have reviewed, and finally deleted in [a020dc1](https://github.com/Qiskit/qiskit/pull/12695/commits/a020dc1191e3abdd979f95b510fc14f0aa5c9778)) computed the inverses of triangular matrices of size less than or equal to 5 explicitly. I have indeed planned (and forgot) to try out this potential optimization. But I did experiment with it right now, generating 100,000 random circuits for 1, 2, 3, 4, and 5 qubits, and did not see any noticeable difference in runtime. So I have ended up simply removing the "todo" comment. 

For posterity, here is the Rust function:
```
/// Invert a lower-triangular matrix with unit diagonal.
fn inverse_tril(mat: ArrayView2<bool>) -> Array2<bool> {
    let num_rows = mat.nrows();
    match num_rows {
        1 => {
            mat.to_owned()
        },
        2 => {
            mat.to_owned()
        },
        3 => {
            let mut inv = mat.to_owned();
            inv[[2, 0]] = mat[[2, 0]] ^ (mat[[1, 0]] & mat[[2, 1]]);
            inv
        },
        4 => {
            let mut inv = mat.to_owned();
            inv[[2, 0]] = mat[[2, 0]] ^ (mat[[1, 0]] & mat[[2, 1]]);
            inv[[3, 1]] = mat[[3, 1]] ^ (mat[[2, 1]] & mat[[3, 2]]);
            inv[[3, 0]] = mat[[3, 0]] ^ (mat[[3, 2]] & mat[[2, 0]]) ^ (mat[[1, 0]] & inv[[3, 1]]);
            inv
        },
        5 => {
            let mut inv = mat.to_owned();
            inv[[2, 0]] = mat[[2, 0]] ^ (mat[[1, 0]] & mat[[2, 1]]);
            inv[[3, 1]] = mat[[3, 1]] ^ (mat[[2, 1]] & mat[[3, 2]]);
            inv[[3, 0]] = mat[[3, 0]] ^ (mat[[3, 2]] & mat[[2, 0]]) ^ (mat[[1, 0]] & inv[[3, 1]]);
            inv[[4, 2]] = (mat[[4, 2]] ^ (mat[[3, 2]] & mat[[4, 3]])) & true;
            inv[[4, 1]] = mat[[4, 1]] ^ (mat[[4, 3]] & mat[[3, 1]]) ^ (mat[[2, 1]] & inv[[4, 2]]);
            inv[[4, 0]] = mat[[4, 0]] ^ (mat[[1, 0]] & inv[[4, 1]]) ^ (mat[[2, 0]] & inv[[4, 2]]) ^ (mat[[3, 0]] & mat[[4, 3]]);
            inv
        },
        _ => {
            calc_inverse_matrix_inner(mat, false).unwrap()
        },
    }
}
```

crates/accelerate/src/synthesis/clifford/random_clifford.rs
alexanderivrii: Indeed, this was not clear to me either, but this is both a direct port of the now removed python function, and the pseudocode in the referenced paper.

crates/accelerate/src/synthesis/clifford/random_clifford.rs
alexanderivrii: Done in [a020dc1](https://github.com/Qiskit/qiskit/pull/12695/commits/a020dc1191e3abdd979f95b510fc14f0aa5c9778).

crates/accelerate/src/synthesis/clifford/random_clifford.rs
alexanderivrii: Hmm, indeed we probably could exploit this somehow. If you think it's worth it, we can think about this and treat this in a potential follow-up.

crates/accelerate/src/synthesis/clifford/random_clifford.rs
alexanderivrii: Technically speaking only half of the entries (when `symmetric` is False). :) I will use the same excuse that I tried to keep the code as close as possible to the original. In any case I don't think this makes much runtime difference (compared to the matrix multiplication between `table1` and `table`).

crates/accelerate/src/synthesis/linear/utils.rs
alexanderivrii: As mentioned above, this code is adapted from `pauli_exp_val.rs`. The code for `getenv_use_multiple_threads` is in `accelerate/src/lib.rs` and checks whether Qiskit is already running in parallel. If not (which I believe is the usual scenario), we will use multiple threading to multiply matrices, but otherwise we will not.

alexanderivrii:
Thanks for the review! To answer your question, I was manually instrumenting my code to print the times to execute various blocks until I saw that the single most expensive line is the binary multiplication of `table1` by `table`.

crates/accelerate/src/synthesis/clifford/random_clifford.rs
```diff
@@ -0,0 +1,162 @@
...
+        had[i] = index < m;
+        let k = if index < m { index } else { 2 * m - index - 1 };
+        perm[i] = inds[k];
+        inds.remove(k);
+    }
+    (had, perm)
+}
+
+/// Add symmetric random boolean value to off diagonal entries.
+fn fill_tril(mut mat: ArrayViewMut2<bool>, rng: &mut Pcg64Mcg, symmetric: bool) {
```
ShellyGarion, line 43: perhaps this function should be in linear utils? it's useful to generate a random symmetric boolean matrix, which can represent a CZ circuit. In this case, it should be called something like "random_boolean matrix".

qiskit/quantum_info/operators/symplectic/random.py
```diff
@@ -111,155 +112,6 @@ def random_clifford(num_qubits: int, seed: int | np.random.Generator | None = No
...
-                mat[1, 2] = mat[2, 1]
-        if dim > 3:
-            mat[3, 0] = rng.integers(2, dtype=np.int8)
-            mat[3, 1] = rng.integers(2, dtype=np.int8)
-            mat[3, 2] = rng.integers(2, dtype=np.int8)
-            if symmetric:
-                mat[0, 3] = mat[3, 0]
-                mat[1, 3] = mat[3, 1]
-                mat[2, 3] = mat[3, 2]
-        return
```
ShellyGarion, line 213: this code above was not used in rust? why?

ShellyGarion (commented):
LGTM, thanks @alexanderivrii

qiskit/quantum_info/operators/symplectic/random.py
ShellyGarion: I've just noticed your explanation here: https://github.com/Qiskit/qiskit/pull/12695#discussion_r1731413331

crates/accelerate/src/synthesis/clifford/random_clifford.rs
Cryoris: It was just a suggestion since you mentioned that the matrix multiplications are slow, but I'm fine not adding this at this point, as I'm not sure `random_clifford` is a bottleneck -- so up to you 🙂

crates/accelerate/src/synthesis/linear/utils.rs
Cryoris: Could you add a comment explaining this, so we know it in the future?

alexanderivrii: Sure, added an explicit comment in [7f443e0](https://github.com/Qiskit/qiskit/pull/12695/commits/7f443e02088f4f452ac433972f12520de3ce0967)

crates/accelerate/src/synthesis/clifford/random_clifford.rs
alexanderivrii: I don't think we should expose the two very special "random linear" functions required for this PR: one that generates random symmetric binary matrices with 0s on the diagonal, and another that generates upper-triangular random symmetric binary matrices with 1s on the diagonal.

crates/accelerate/src/synthesis/clifford/random_clifford.rs
alexanderivrii: I like the suggestion, but would rather not go down this rabbit hole for this PR. As they say, "premature optimization is the root of all evil" 🙂.

crates/accelerate/src/synthesis/clifford/random_clifford.rs
ShellyGarion: ok, but I still think that the name "fill_tril" could be changed to something more meaningful (it also doesn't appear in the paper)

alexanderivrii: But... this is the name from our Python code :). Ok, will try to think of better names for this and other functions.

ShellyGarion (commented):
LGTM. Thanks @alexanderivrii .
If @Cryoris has no further comments, then we can merge it.

Cryoris (approved):
LGTM thanks for all the work! If we ever need to revisit I think we have some ideas in this PR to even optimize the code a bit more 🙂

PR #12698, opened 2024-07-01
Expose operation conversion methods to other crates.

raynelfss:
### Summary
After #12459 merged, a couple of conversion methods were introduced to be able to represent operations in Rust. These commits aim to expose these methods so that they can be used on other crates within `qiskit._accelerate`. One example of its use is for #12585.

### Details and comments
- Expose `operation_type_to_py`, `operation_type_and_data_to_py`, `convert_py_to_operation_type`, methods to other rust crates.
- Expose `OperationTypeConstruct` struct to other crates.


jlapeyre:
Can  you add some motivation? Is there something in the works that intends to use these methods?

raynelfss:
@jlapeyre, Here's one use case:

https://github.com/raynelfss/qiskit/blob/77dbec896677a67df3ab350448c537b23df973f9/crates/circuit/src/equivalence.rs#L237-L252

This is the way we represent gates within the EquivalenceLibrary as we still need to check the parameters that are used within each `Operation`.

jlapeyre (approved):

PR #12453, opened 2024-05-22
Add `DenseLayout` trial to `SabreLayout`

mtreinish:
### Summary

Building on the work done in #10829, #10721, and #12104 this commit adds
a new trial to all runs of `SabreLayout` that runs the dense layout
pass. In general the sabre layout algorithm starts from a random layout
and then runs a routing algorithm to permute that layout virtually where
swaps would be inserted to select a layout that would result in fewer
swaps. As was discussed in #10721 and #10829 that random starting point
is often not ideal especially for larger targets where the distance
between qubits can be quite far. Especially when the circuit qubit count
is low relative to the target's qubit count this can result it poor
layouts as the distance between the qubits is too large. In qiskit we
have an existing pass, `DenseLayout`, which tries to find the most
densely connected n qubit subgraph of a connectivity graph. This
algorithm necessarily will select a starting layout where the qubits are
near each other and for those large backends where the random starting
layout doesn't work well this can improve the output quality.

As the overhead of `DenseLayout` is quite low and the core algorithm is
written in rust already this commit adds a default trial that uses
DenseLayout as a starting point on top of the random trials (and any
input starting points). For example if the user specifies to run
SabreLayout with 20 layout trials this will run 20 random trials and
one trial with `DenseLayout` as the starting point. This is all done
directly in the sabre layout rust code for efficiency. The other
difference between the standalone `DenseLayout` pass is that in the
standalone pass a sparse matrix is built and a reverse Cuthill-McKee
permutation is run on the densest subgraph qubits to pick the final
layout. This permutation is skipped because in the case of Sabre's
use of dense layout we're relying on the sabre algorithm to perform
the permutation.

### Details and comments

Depends on: #12104

This PR is based on top of #12104 and will be manually rebased after that merges. You can look at: https://github.com/Qiskit/qiskit/commit/06e208f71b776400e57aef4fd2b04df1bf00f790 for the details of just this PR.

PR #12701, opened 2024-07-01
Avoid Python op creation in commutative cancellation

mtreinish:
### Summary

This commit updates the commutative cancellation and commutation
analysis transpiler pass. It builds off of #12692 to adjust access
patterns in the python transpiler path to avoid eagerly creating a
Python space operation object. The goal of this PR is to mitigate the
performance regression on these passes introduced by the extra
conversion cost of #12459.

### Details and comments

~This is based on top of #12692 and will need to be rebased after #12692 merges. To see the diff of just this PR you can look at the last commit: https://github.com/Qiskit/qiskit/commit/5beaad493aadb0de356b95add5590d7783eed166~ Rebased on main now that #12692 has merged


mtreinish:
With this PR we're getting closer to the performance of 1.1.0, most importantly the QFT benchmarks are no longer timing out with this PR: 

```

| Change   |   Before [9092ee78] <1.1.1^0> |   After [94abf30f]  |   Ratio | Benchmark (Parameter)                                       |
|----------|-------------------------------|---------------------|---------|-------------------------------------------------------------|
| -        |                          2582 |                1954 |    0.76 | utility_scale.UtilityScaleBenchmarks.track_qft_depth('cx')  |
| -        |                          2582 |                1954 |    0.76 | utility_scale.UtilityScaleBenchmarks.track_qft_depth('cz')  |
| -        |                          2582 |                1954 |    0.76 | utility_scale.UtilityScaleBenchmarks.track_qft_depth('ecr') |

| Change   | Before [9092ee78] <1.1.1^0>   | After [94abf30f]    | Ratio   | Benchmark (Parameter)                                                     |
|----------|-------------------------------|---------------------|---------|---------------------------------------------------------------------------|
|          | 23.2±0.03s                    | 26.4±0.04s          | ~1.14   | utility_scale.UtilityScaleBenchmarks.time_qft('ecr')                      |
|          | 23.2±0.04s                    | 25.4±0.03s          | 1.10    | utility_scale.UtilityScaleBenchmarks.time_qft('cz')                       |
|          | 20.6±0.06s                    | 21.0±0.03s          | 1.02    | utility_scale.UtilityScaleBenchmarks.time_qft('cx')                       |
|          | 444                           | 435                 | 0.98    | utility_scale.UtilityScaleBenchmarks.track_square_heisenberg_depth('cx')  |
|          | 444                           | 435                 | 0.98    | utility_scale.UtilityScaleBenchmarks.track_square_heisenberg_depth('cz')  |
|          | 444                           | 435                 | 0.98    | utility_scale.UtilityScaleBenchmarks.track_square_heisenberg_depth('ecr') |
|          | 1607                          | 1483                | 0.92    | utility_scale.UtilityScaleBenchmarks.track_qaoa_depth('cx')               |
|          | 1622                          | 1488                | 0.92    | utility_scale.UtilityScaleBenchmarks.track_qaoa_depth('cz')               |
|          | 1622                          | 1488                | 0.92    | utility_scale.UtilityScaleBenchmarks.track_qaoa_depth('ecr')              |

Benchmarks that have got worse:

| Change   | Before [9092ee78] <1.1.1^0>   | After [94abf30f]    |   Ratio | Benchmark (Parameter)                                                         |
|----------|-------------------------------|---------------------|---------|-------------------------------------------------------------------------------|
| +        | 1.77±0.01s                    | 3.97±0.01s          |    2.24 | utility_scale.UtilityScaleBenchmarks.time_square_heisenberg('ecr')            |
| +        | 1.27±0s                       | 2.80±0.01s          |    2.21 | utility_scale.UtilityScaleBenchmarks.time_square_heisenberg('cx')             |
| +        | 1.93±0.01s                    | 4.07±0.01s          |    2.11 | utility_scale.UtilityScaleBenchmarks.time_square_heisenberg('cz')             |
| +        | 97.3±0.9ms                    | 176±4ms             |    1.81 | utility_scale.UtilityScaleBenchmarks.time_parse_qft_n100('cz')                |
| +        | 97.8±0.7ms                    | 176±3ms             |    1.8  | utility_scale.UtilityScaleBenchmarks.time_parse_qft_n100('cx')                |
| +        | 31.9±0.4ms                    | 57.2±0.9ms          |    1.79 | utility_scale.UtilityScaleBenchmarks.time_parse_square_heisenberg_n100('cz')  |
| +        | 2.45±0.02s                    | 4.38±0.02s          |    1.79 | utility_scale.UtilityScaleBenchmarks.time_qaoa('ecr')                         |
| +        | 97.9±0.4ms                    | 174±2ms             |    1.78 | utility_scale.UtilityScaleBenchmarks.time_parse_qft_n100('ecr')               |
| +        | 9.02±0.06ms                   | 15.9±0.3ms          |    1.76 | utility_scale.UtilityScaleBenchmarks.time_parse_qaoa_n100('cz')               |
| +        | 9.02±0.06ms                   | 15.9±0.2ms          |    1.76 | utility_scale.UtilityScaleBenchmarks.time_parse_qaoa_n100('ecr')              |
| +        | 32.2±0.4ms                    | 56.6±0.9ms          |    1.76 | utility_scale.UtilityScaleBenchmarks.time_parse_square_heisenberg_n100('cx')  |
| +        | 32.2±0.1ms                    | 56.7±1ms            |    1.76 | utility_scale.UtilityScaleBenchmarks.time_parse_square_heisenberg_n100('ecr') |
| +        | 9.18±0.2ms                    | 16.0±0.3ms          |    1.74 | utility_scale.UtilityScaleBenchmarks.time_parse_qaoa_n100('cx')               |
| +        | 1.12±0.01s                    | 1.91±0.01s          |    1.71 | utility_scale.UtilityScaleBenchmarks.time_qaoa('cx')                          |
| +        | 2.96±0.05s                    | 4.60±0.01s          |    1.56 | utility_scale.UtilityScaleBenchmarks.time_qaoa('cz')                          |

PERFORMANCE DECREASED.
```

(the sha1 is different because I had it rebased on main locally for testing and the quality changes are unrelated and likely caused by #12453 which merged after 1.1.0)

The next passes causing bottlenecks around gate object creation after this PR are:
1. `BasisTranslator`
2. `ConsolidateBlocks`
3. `Optimize1qGatesDecomposition` (which is covered by #12650) 
4. `Collect2qBlocks`

crates/circuit/src/circuit_instruction.rs
```diff
@@ -432,6 +432,41 @@ impl CircuitInstruction {
...
+        if copy == Some(false) {
+            return Err(PyValueError::new_err(
+                "A copy is needed to return an array from this object.",
+            ));
+        }
+        let res =
+            match self.to_matrix(py) {
+                Some(res) => res,
+                None => return Err(PyTypeError::new_err(
+                    "ParameterExpression with unbound parameters cannot be represented as an array",
```
sbrandhsn: Is this the only way `self.to_matrix(py)` can fail - can we forward the error message produced by python here?

sbrandhsn (commented):
This generally looks good to me, thanks! :-) I had one question on handling error messages but apart from this I'd be happy to approve this PR!

crates/circuit/src/circuit_instruction.rs
mtreinish: Yeah, the `matrix()` method defined in the `Operation` trait only returns an `Option<Array2<Complex64>>` . The goal of the trait was to define pure rust based semantics for querying operations. So for standard gates it's either a matrix if all the parameters are floats or `None` if any parameter is a `ParameterExpression`. For the `PyGate` rust space wrapper it converting any potential error (or the lack of a matrix) to `None`: https://github.com/Qiskit/qiskit/blob/main/crates/circuit/src/operations.rs#L1549-L1565.

We could add an alternative path that tried to preserve the error message from Python, but the point of the `Operation` trait was to abstract some of this away for a rust context. Realistically though I modeled this error message after what happens in the common case:

```
>>> from qiskit.circuit import Parameter
>>> from qiskit.circuit.library import RZGate
>>> gate = RZGate(Parameter('a'))
>>> gate.__array__()
  File "<stdin>", line 1, in <module>
  File "/tmp/test_11/lib/python3.12/site-packages/qiskit/circuit/library/standard_gates/rz.py", line 139, in __array__
    ilam2 = 0.5j * float(self.params[0])
                   ^^^^^^^^^^^^^^^^^^^^^
  File "/tmp/test_11/lib/python3.12/site-packages/qiskit/circuit/parameterexpression.py", line 415, in __float__
    raise TypeError(
TypeError: ParameterExpression with unbound parameters (dict_keys([Parameter(a)])) cannot be cast to a float.
```

There is a loss of information for opaque gates that don't define a matrix where if a user did
```python
from qiskit.circuit import Gate, CircuitInstruction
custom_gate = Gate(name='foo', num_qubits=2, params=[])
custom_gate.__array__()
custom_instruction = CircuitInstruction(custom_gate)
custom_instruction.__array__()
```
The `custom_instruction.__array__()` path would return this parameter expression error vs an `AttributeError` for the custom_gate.__array__()` call.

That all being said this path doesn't really get called in this PR, as custom gates will work in the python domain. We only use this `to_matrix()` and `__array__` if it's a standard gate: https://github.com/Qiskit/qiskit/pull/12701/files/fa774b3e3d493d9b04cb98166e40673301abae10#diff-f652dda983f1a9918c329ea373b9e570cc2f037ab22c59a968424deea2610832R70
but I guess it's a potential future hazard for ourselves and maybe we want to avoid doing this. The other alternative I guess would be to type check in `commute_matmul()` and call `DAGOpNode.matrix` to build the `Operator` instead of passing the `DAGOpNode` directly to `Operator`. I went back and forth on this when I wrote the PR. Which would you prefer?

sbrandhsn: I think the current version is fine, thanks for explaining! :-)

sbrandhsn (dismissed):
LGMT, thanks!

crates/circuit/src/circuit_instruction.rs
mtreinish: Actually after writing that all out, I've convinced myself that adding `to_matrix()` and `__array__` is the incorrect path forward here. It is too big an interface to add haphazardly and it's easy enough to workaround in the commutation checker code directly. I'll refactor this as part of the rebase I need to do to fix merge conflicts anyway.

mtreinish: Removed in: https://github.com/Qiskit/qiskit/pull/12701/commits/5f94bf07ee0d38a40e02f0befec048a204c44737

sbrandhsn (approved):

PR #12704, opened 2024-07-01
Use rust gates for ConsolidateBlocks

mtreinish:
### Summary

This commit moves to use rust gates for the ConsolidateBlocks transpiler
pass. Instead of generating the unitary matrices for the gates in a 2q
block Python side and passing that list to a rust function this commit
switches to passing a list of DAGOpNodes to the rust and then generating
the matrices inside the rust function directly. This is similar to what
was done in #12650 for Optimize1qGatesDecomposition. Besides being faster
to get the matrix for standard gates, it also reduces the eager
construction of Python gate objects which was a significant source of
overhead after #12459. To that end this builds on the thread of work in
the two PRs #12692 and #12701 which changed the access patterns for
other passes to minimize eager gate object construction.

### Details and comments

TODO:

- [x] Fix last test failures
- [x] Benchmark and profile
- [x] Rebase after #12692 and #12701 merge

This PR is built on top of #12692 and #12701 and will need to rebased after both merge. To view the contents of just this PR you can look at the last commit on the branch (I'll be force pushing updates to that last commit): https://github.com/Qiskit/qiskit/pull/12704/commits/e449357e45acc3b3c25aad1ef1088706ff0a51a0


crates/accelerate/src/convert_2q_block_matrix.rs
```diff
@@ -20,32 +22,85 @@ use numpy::ndarray::{aview2, Array2, ArrayView2};
...
     py: Python,
-    op_list: Vec<(PyReadonlyArray2<Complex64>, SmallVec<[u8; 2]>)>,
+    op_list: Vec<PyRef<DAGOpNode>>,
+    block_index_map_dict: &Bound<PyDict>,
 ) -> PyResult<Py<PyArray2<Complex64>>> {
+    // Build a BitData in block_index_map_dict order. block_index_map_dict is a dict of bits to
+    // indices mapping the order of the qargs in the block. There should only be 2 entries since
+    // there are only 2 qargs here (e.g. `{Qubit(): 0, Qubit(): 1}`) so we need to ensure that
+    // we added the qubits to bit data in the correct index order.
+    let mut index_map: Vec<PyObject> = (0..block_index_map_dict.len()).map(|_| py.None()).collect();
```
jlapeyre, line 68: This is a bit easier to read
```suggestion
    let mut index_map = vec![py.None(); block_index_map_dict.len()];
```
But it's probably not exactly the same. `vec!` calls `clone`. And it looks like this increments a ref count for a `PyAny`. The test suite passes with this edit. Still not sure the change is a good idea. `py.None()` eventually does this `ffi::Py_None().assume_borrowed(py).downcast_unchecked()`

mtreinish: I wrote it this way specifically to avoid the clone, since it's a pyobject we'd need the GIL to do that and `clone()` will get it internally by calling `with_gil()`, but this has extra overhead.

jlapeyre: Yeah, I saw the thing about "faster if you happen to be already holding the GIL", but I had no idea how much faster. If the overhead is at all significant, I agree the change is not worth it.

crates/accelerate/src/convert_2q_block_matrix.rs
```diff
@@ -20,32 +22,85 @@ use numpy::ndarray::{aview2, Array2, ArrayView2};
...
     let identity = aview2(&ONE_QUBIT_IDENTITY);
-    let input_matrix = op_list[0].0.as_array();
-    let mut matrix: Array2<Complex64> = match op_list[0].1.as_slice() {
+    let first_node = &op_list[0];
+    let input_matrix = get_matrix_from_inst(py, &first_node.instruction)?;
+    let mut matrix: Array2<Complex64> = match bit_map
+        .map_bits(first_node.instruction.qubits.bind(py).iter())?
+        .map(|x| x as u8)
+        .collect::<SmallVec<[u8; 2]>>()
+        .as_slice()
```
jlapeyre: Why not this?
```rust
        .map_bits(first_node.instruction.qubits.bind(py).iter())?
        .collect::<Vec<_>>()
        .as_slice()
```
Storage should not an issue here, and `u8` at best won't be more performant. Maybe the `2` here in `SmallVec<[u8; 2]` allows the compiler to optimize?

crates/accelerate/src/convert_2q_block_matrix.rs
```diff
@@ -20,32 +22,85 @@ use numpy::ndarray::{aview2, Array2, ArrayView2};
...
-        [1, 0] => change_basis(input_matrix),
+        [0, 1] => input_matrix,
+        [1, 0] => change_basis(input_matrix.view()),
         [] => Array2::eye(4),
         _ => unreachable!(),
     };
-    for (op_matrix, q_list) in op_list.into_iter().skip(1) {
-        let op_matrix = op_matrix.as_array();
+    for node in op_list.into_iter().skip(1) {
+        let op_matrix = get_matrix_from_inst(py, &node.instruction)?;
```
jlapeyre, line 93: Line 91 would convey more information as `[..] => unreachable!()`

crates/accelerate/src/convert_2q_block_matrix.rs
```diff
@@ -71,8 +126,42 @@ pub fn change_basis(matrix: ArrayView2<Complex64>) -> Array2<Complex64> {
     trans_matrix
 }
 
+#[pyfunction]
```
jlapeyre, line 128: Line 50 could be `[..] => None`

jlapeyre:
This PR does not touch `map_bits` itself. But you might want to consider 
 [this change](https://github.com/mtreinish/qiskit-core/compare/py-access-rs-data-consolidate-blocks...jlapeyre:qiskit-core:py-access-rs-data-consolidate-blocks-map-bits). (https://github.com/jlapeyre/qiskit-core/commit/71e92a6aab82a3df7cdb5783dcbc3936841196e2) This simplifies how `map_bits` is implemented and how it is called.

mtreinish:
> This PR does not touch `map_bits` itself. But you might want to consider [this change](https://github.com/mtreinish/qiskit-core/compare/py-access-rs-data-consolidate-blocks...jlapeyre:qiskit-core:py-access-rs-data-consolidate-blocks-map-bits). ([jlapeyre@71e92a6](https://github.com/jlapeyre/qiskit-core/commit/71e92a6aab82a3df7cdb5783dcbc3936841196e2)) This simplifies how `map_bits` is implemented and how it is called.

I would want to check with @kevinhartman and @jakelishman before making this change since there is a fair amount in flux around using `BitData` between #12550 and #12730. I was trying to avoid chaning the rust interface for dealing with bit data too much in this PR.

jakelishman:
imo if we were to touch any meaningful parts of `BitData` (beyond just making it public), then we'd do it in a separate PR - this PR doesn't logically need to change the interfaces, and doing it separately would let us track performance changes.

I haven't currently touched `BitData`, though I imagine it probably will change a bit in coming PRs, so best to just leave it un-churned right now.

jlapeyre:
I was on the fence about it. But `map_bits` was called four times in `circuit_data`, and two calls were added here in `convert_2q_block_matrix`. There are no other calls.

I think a separate PR is good. Especially because, yeah, there is a possible change in performance.

 In fact, I would add a method on `PackedInstruction` (if that's the right struct) because every call to `map_bits` starts with an instruction (or higher).

crates/accelerate/src/convert_2q_block_matrix.rs
jlapeyre: These two suggestions, doing `[..] => None`, change lines not yet touched by this PR. Whether to do it here depends on whether in general you favor cleaning these things up in a PR or doing them in a separate PR. An argument against is that this is not strictly part of the point of the PR. An argument for is that a PR to do these things would not be a high priority, so it might never be done. So I favor making the change. OTOH, the arguments for leaving these as is are not unreasonable.

crates/accelerate/src/convert_2q_block_matrix.rs
mtreinish: I was thinking it was for typing reasons, but I realized I was confusing this with the 2q decomposer and the euler 1q decomposer which are using `SmallVec<u8>` as the output type. I agree in this case just collecting to a vec of u32 is fine. I can change it

jakelishman:
I have a half-done PR that's reworking some of the `Interner` interfaces to avoid a couple of unnecessary allocations.  I think I'm likely to touch `BitData` as part of that.

crates/accelerate/src/convert_2q_block_matrix.rs
jlapeyre: That's what I guessed. It was not a straight copy and paste, but something like that, from a situation where it *is* important that the type be `u8`.

jlapeyre:
```suggestion
        .map_bits(first_node.instruction.qubits.bind(py).iter())?
        .collect::<Vec<_>>()
        .as_slice()
```

jlapeyre (approved):

PR #12705, opened 2024-07-01
Avoid Python op creation in BasisTranslator

mtreinish:
### Summary

This commit updates the BasisTranslator transpiler pass. It builds off
of #12692 and #12701 to adjust access patterns in the python transpiler
path to avoid eagerly creating a Python space operation object. The goal
of this PR is to mitigate the performance regression introduced by the
extra conversion cost of #12459 on the BasisTranslator.

### Details and comments

TODO:

- [x] Fix 2 failing test cases with control flow
- [x] Benchmark and profile
- [x] Rebase after #12692 and #12701 merge

This PR is built on top of https://github.com/Qiskit/qiskit/pull/12692 and https://github.com/Qiskit/qiskit/pull/12701 and will need to rebased after both merge. To view the contents of just this PR you can look at the last commit on the branch (I'll be force pushing updates to that last commit): https://github.com/Qiskit/qiskit/commit/d95ef618157d11580ff4a8b31f34562477dde706


jakelishman:
Note: this will conflict with #12730.

mtreinish:
I'm marking this as ready to review. From profiling there is still significant overhead in `substitute_node_with_dag()` that gets called a lot by the basis translator. While I could work on mitigating that I think between #12730 and then eventually #12550 will just take care of that. So if we need to do something for 1.2 there we can handle it in a follow up.

jakelishman (approved):
I'm fine to merge this now, and I'll fix up the merge conflict in #12730.  I think that #12730 will already fix very nearly all the remaining overhead in `substitute_node_with_dag` (all the stuff about "additional wires") around `node.op` creation, and the last change (an `isinstance(node.op, ControlFlowOp)`) is trivial.

PR #12711, opened 2024-07-02
Improving shots estimate to give errors within user-specified precision.

jsaroni:
### Summary 
New PR to replace #12618 because that one was from an incorrect branch.

Fixes #issue-12414 (https://github.com/Qiskit/qiskit/issues/12414). When the user specifies the precision in EstimatorV2.run, the actual standard deviation from results is not bounded by or close to the user-specified precision for an observable that is a general linear combination of Pauli strings. This pull request gives a more accurate approximation to the number of shots needed for the desired precision for such observables. The tests to verify this are based on https://docs.quantum.ibm.com/api/qiskit-ibm-runtime/qiskit_ibm_runtime.EstimatorV2 and https://github.com/Qiskit/qiskit-ibm-runtime/issues/1684.

### Details and comments
The shots estimation accounts for the sum of the absolute value squared of the coefficients of the observable as proposed in https://github.com/Qiskit/qiskit/issues/12414 to give a standard deviation that is bounded by and closer to the user-specified precision for any observable. In backend_estimator_v2.py, the _run function is modified to include the sum of the absolute value squared of the coefficients of the observable in the shots estimation.

The following plot shows that the ratio of precision from results to the user-specified precision changes with varying observable term coefficients using the current implementation. In the proposed implementation, this ratio is fixed and close to 1.

image

The plot below shows that in the current implementation, having more than one occurrence of a term in an observable leads to precision results that are different from the user-specified precision due to a fixed number of shots for observables with different numbers of terms and error propagation. In the proposed implementation, the precision from the results remains close to the user-specified precision.

image


qiskit/primitives/backend_estimator_v2.py
```diff
@@ -161,8 +161,11 @@ def _validate_pubs(self, pubs: list[EstimatorPub]):
     def _run(self, pubs: list[EstimatorPub]) -> PrimitiveResult[PubResult]:
         pub_dict = defaultdict(list)
         # consolidate pubs with the same number of shots
+        coeff_sq_sum = 0
```
SamFerracin: I am a little bit confused by the math here..

The variance of a linear combination `aX + bY + ...` of observables `X, Y, ...` is given by:

> `var(aX + bY + ...) = a**2 Var(X) + b**2 Var(Y) + ...`, 


where `Var(O) = 1/N_O` is the variance of the observable `O` and `N_O` is the number of shots for the observable `O`.
In our case, we want `var(aX + bY + ...)` to be equal to `epsilon**2`, and we want to use the same number of shots `N` for every observable `O`.  Therefore, we have: 

> `epsilon**2 = a**2/N + b**2/N + ...`, 


which gives:

> `N = (a**2 + b**2 + ...)/epsilon**2`

So here is how I would code this up:
```
for i, pub in enumerate(pubs):
    coeff_sq_sum = 0
    for obs in pub.observables.ravel():
        coeff_sq_sum += sum(coeff**2 for coeff in obs.values())
    shots = int(math.ceil(coeff_sq_sum / pub.precision**2))
```

Am I missing something?

SamFerracin: FYI, there's a related issue that may shed some light
https://github.com/Qiskit/qiskit-ibm-runtime/issues/1684

jsaroni: Right. This is how it should be as written by @ihincks in https://github.com/Qiskit/qiskit/pull/12618. I have updated it. Given that we are estimating the precision without knowing the target state, setting `Var(O)=1/N_O` bounds the shot number.

ihincks: > Am I missing something?

I think you might be: each element of `pub.observables.ravel()` is itself a multi-term observable. The point of the max was to find the worst offender.

SamFerracin: I see. So every `obs` is not an observable, but a list of (potentially more than one) observables.
@jsaroni could you add a comment about this? It will help us remember in the future

jsaroni: Done.

SamFerracin (dismissed):

test/python/primitives/test_backend_estimator_v2.py
```diff
@@ -388,6 +388,29 @@ def test_diff_precision(self, backend, abelian_grouping):
         np.testing.assert_allclose(result[0].data.evs, [1.901141473854881], rtol=self._rtol)
         np.testing.assert_allclose(result[1].data.evs, [1.901141473854881], rtol=self._rtol)
 
+    @combine(backend=BACKENDS, abelian_grouping=[True, False])
+    def test_precision_error(self, backend, abelian_grouping):
+        "Test for error"
```
t-imamichi: Could you write a docstring `"""..."""` instead of a string `"..."`?
"Test for error" looks too generic. Could you write a little more details?

t-imamichi:
Could you apply black following https://github.com/Qiskit/qiskit/blob/main/CONTRIBUTING.md#style-and-lint? CI fails due to a style error.

test/python/primitives/test_backend_estimator_v2.py
```diff
@@ -388,6 +388,29 @@ def test_diff_precision(self, backend, abelian_grouping):
         np.testing.assert_allclose(result[0].data.evs, [1.901141473854881], rtol=self._rtol)
         np.testing.assert_allclose(result[1].data.evs, [1.901141473854881], rtol=self._rtol)
 
+    @combine(backend=BACKENDS, abelian_grouping=[True, False])
+    def test_precision_error(self, backend, abelian_grouping):
```
t-imamichi: The test function named "test_precision_error", but there is no error checking in the test.
`test_something_error` usually checks a particular error is actually raised or not by `self.assertRaises`.
The current test looks checking `stds` with the target precision. If it's correct, I would suggest a function name `test_precision_and_stds`, for example.

test/python/primitives/test_backend_estimator_v2.py
jsaroni: How does """Test that errors are within user-specified precision""" sound?

test/python/primitives/test_backend_estimator_v2.py
jsaroni: Ok, changed it to test_precision_and_stds.

jsaroni:
> Could you apply black following https://github.com/Qiskit/qiskit/blob/main/CONTRIBUTING.md#style-and-lint? CI fails due to a style error.

I used the black formatting. I checked that lint does not return an error locally. I still get the CI error

```
1 file would be reformatted, 1277 files would be left unchanged. 
lint: exit 1 (38.82 seconds) /home/vsts/work/1/s> black --check qiskit test tools examples setup.py pid=13021 
docs: OK (598.62=setup[407.74]+cmd[21.00,169.88] seconds) 
lint: FAIL code 1 (209.40=setup[170.58]+cmd[38.82] seconds) 
evaluation failed :( (808.81 seconds) 
##[error]Bash exited with code '255'. 
```

t-imamichi:
No problem. I applied black to this branch. Hopefully CI passes now.

t-imamichi:
Could you write a release note following the guide? I think this PR should be "features_primitives" section.
https://github.com/Qiskit/qiskit/blob/main/CONTRIBUTING.md#release-notes

jsaroni:
> Could you write a release note following the guide? I think this PR should be "features_primitives" section. https://github.com/Qiskit/qiskit/blob/main/CONTRIBUTING.md#release-notes

Thanks. I have added a release note for review https://github.com/jsaroni/qiskit/blob/estimator_primitive/releasenotes/notes/backend_estimator_stds-aa6893ab647567f1.yaml

releasenotes/notes/backend_estimator_stds-aa6893ab647567f1.yaml
```diff
@@ -0,0 +1,29 @@
...
+      service = QiskitRuntimeService()
+      backend = service.least_busy(operational=True, simulator=False)
+      
+      psi = RealAmplitudes(num_qubits=2, reps=2)
+      hamiltonian = SparsePauliOp.from_list([("II", 1), ("IZ", 2), ("XI", 3)])
+      theta = [0, 1, 1, 2, 3, 5]
+      
+      pm = generate_preset_pass_manager(backend=backend, optimization_level=1)
+      isa_psi = pm.run(psi)
+      isa_observables = hamiltonian.apply_layout(isa_psi.layout)
```
t-imamichi, line 22:
```suggestion
    An estimation of the shots has been updated in :meth:`.BackendEstimatorV2.run` to account for the error from multi-term observables.
```

releasenotes/notes/backend_estimator_stds-aa6893ab647567f1.yaml
```diff
@@ -0,0 +1,29 @@
...
+      
+      estimator = Estimator(backend=backend)
+      
+      # calculate [ <psi(theta)|hamiltonian|psi(theta)> ]
+      job = estimator.run([(isa_psi, isa_observables, [theta])], precision=0.01)
+      pub_result = job.result()[0]
+      print(f"Expectation values: {pub_result.data.evs}")
```
t-imamichi: Thank you for writing a reno. You don't need to write an example because this PR does not change API.
```suggestion
    An estimation of the shots has been updated in :meth:`.BackendEstimatorV2.run` to account for the error from multi-term observables.
```

qiskit/primitives/backend_estimator_v2.py
```diff
@@ -162,7 +162,11 @@ def _run(self, pubs: list[EstimatorPub]) -> PrimitiveResult[PubResult]:
         pub_dict = defaultdict(list)
         # consolidate pubs with the same number of shots
         for i, pub in enumerate(pubs):
-            shots = int(math.ceil(1.0 / pub.precision**2))
+            coeff_sq_sum = 0
+            for obs in pub.observables.ravel():
+                # each element of pub.observables.ravel() may be a multi-term observable
+                coeff_sq_sum = max(coeff_sq_sum, sum(coeff**2 for coeff in obs.values()))
+            shots = int(math.ceil(coeff_sq_sum / pub.precision**2))
```
t-imamichi: If all coefficients are 0, shots is set 0 and it results in an error. You need to set at least 1 shot.
Could you add the following case to the unit test?
```suggestion
            shots = int(math.ceil(coeff_sq_sum / pub.precision**2))
            shots = max(1, shots)
```

The following script causes an error only for this PR.
```python
from qiskit.primitives import BackendEstimatorV2

backend = AerSimulator()
estimator = BackendEstimatorV2(backend=backend)

qc = QuantumCircuit(1)
qc.h(0)
op = SparsePauliOp.from_list([("X", 0), ("Y", 0)])

result = estimator.run([(qc, op)]).result()
print(result[0].data.evs)
```

this PR
```
qiskit.exceptions.QiskitError: 'No counts for experiment "0"'
```

main branch
```
0.0
```

jsaroni: @t-imamichi I see, thanks. Could I use one line?
`shots = max(1, int(math.ceil(coeff_sq_sum / pub.precision**2)))`

t-imamichi: Sure

t-imamichi:
Could you write a test of zero coefficient based on this script?
https://github.com/Qiskit/qiskit/pull/12711#discussion_r1673645697

jsaroni:
> Could you write a test of zero coefficient based on this script? [#12711 (comment)](https://github.com/Qiskit/qiskit/pull/12711#discussion_r1673645697)

Ok. I added a test in `test_precision_and_stds` for when all observable term coefficients are zero. In this case both stds and evs are zero regardless of the set precision.

t-imamichi:
Thank you for adding the test. I'm running CI...

t-imamichi:
The recent changes breaks the style. You need to apply black again.

test/python/primitives/test_backend_estimator_v2.py
```diff
@@ -388,6 +388,78 @@ def test_diff_precision(self, backend, abelian_grouping):
         np.testing.assert_allclose(result[0].data.evs, [1.901141473854881], rtol=self._rtol)
         np.testing.assert_allclose(result[1].data.evs, [1.901141473854881], rtol=self._rtol)
 
+    @combine(backend=BACKENDS, abelian_grouping=[True, False])
+    def test_precision_and_stds(self, backend, abelian_grouping):
```
ihincks: Optional

```suggestion
    @combine(backend=BACKENDS, abelian_grouping=[True, False], hamiltonian=...)
    def test_precision_and_stds(self, backend, abelian_grouping, hamiltonian):
```

test/python/primitives/test_backend_estimator_v2.py
```diff
@@ -388,6 +388,78 @@ def test_diff_precision(self, backend, abelian_grouping):
...
+                ],
+                precision=self._precision,
+            )
+            result = job.result()
+            np.testing.assert_array_less(
+                result[0].data.stds, [self._precision] * len(hamiltonian[i])
+            )
+            np.testing.assert_array_less(
+                result[1].data.stds, [self._precision] * len(hamiltonian[i])
+            )
```
ihincks:
```suggestion
        for hamiltonian in hamiltonians:
            job = estimator.run(
                [(self.psi[0], hamiltonian, [self.theta[0]])], precision=self._precision
            )
            result = job.result()
            np.testing.assert_array_less(
                result[0].data.stds, [self._precision] * len(hamiltonian)
            )
            # The result of the second run is the same
            job = estimator.run(
                [
                    (self.psi[0], hamiltonian, [self.theta[0]]),
                    (self.psi[0], hamiltonian, [self.theta[0]]),
                ],
                precision=self._precision,
            )
            result = job.result()
            np.testing.assert_array_less(
                result[0].data.stds, [self._precision] * len(hamiltonian)
            )
            np.testing.assert_array_less(
                result[1].data.stds, [self._precision] * len(hamiltonian)
            )
```

test/python/primitives/test_backend_estimator_v2.py
```diff
@@ -388,6 +388,78 @@ def test_diff_precision(self, backend, abelian_grouping):
...
+            SparsePauliOp.from_list([("ZZ", 1), ("ZZ", 1)]),
+            [
+                SparsePauliOp.from_list([("II", 1), ("IZ", 2), ("XI", 3)]),
+                SparsePauliOp.from_list([("ZZ", 1)]),
+                SparsePauliOp.from_list([("ZZ", 1), ("ZZ", 1)]),
+            ],
+        ]
+        for i in range(len(hamiltonian)):
+            job = estimator.run(
+                [(self.psi[0], hamiltonian[i], [self.theta[0]])], precision=self._precision
```
ihincks: Why is the dimension being increased here? Not a big deal, but don't add it unless there's a reason.
```suggestion
                [(self.psi[0], hamiltonian[i], self.theta[0]]), precision=self._precision
```

test/python/primitives/test_backend_estimator_v2.py
```diff
@@ -388,6 +388,78 @@ def test_diff_precision(self, backend, abelian_grouping):
...
+            )
+            result = job.result()
+            np.testing.assert_array_less(
+                result[0].data.stds, [self._precision] * len(hamiltonian[i])
```
ihincks: Two issues here:
 1. assert_array_less already does broadcasting, so no need to replicate the precision ourselves
 2. len(hamilotian) for hamiltonian[0] [1] and [2] does not correspond to the size of `.stds`

```suggestion
                result[0].data.stds, self._precision
```

test/python/primitives/test_backend_estimator_v2.py
ihincks:
```suggestion
```

test/python/primitives/test_backend_estimator_v2.py
```diff
@@ -388,6 +388,78 @@ def test_diff_precision(self, backend, abelian_grouping):
         np.testing.assert_allclose(result[0].data.evs, [1.901141473854881], rtol=self._rtol)
         np.testing.assert_allclose(result[1].data.evs, [1.901141473854881], rtol=self._rtol)
 
+    @combine(backend=BACKENDS, abelian_grouping=[True, False])
```
ihincks: What is the value of looping over backends and abelian_grouping? Does it actually increase test coverage meaningfully?

test/python/primitives/test_backend_estimator_v2.py
```diff
@@ -388,6 +388,78 @@ def test_diff_precision(self, backend, abelian_grouping):
...
+            result = job.result()
+            np.testing.assert_array_less(
+                result[0].data.stds, [self._precision] * len(hamiltonian[i])
+            )
+            np.testing.assert_array_less(
+                result[1].data.stds, [self._precision] * len(hamiltonian[i])
+            )
+
+    @combine(backend=BACKENDS, abelian_grouping=[True, False])
+    def test0_precision_and_stds(self, backend, abelian_grouping):
```
ihincks:
```suggestion
    def test_precision_with_zero_coefficients(self, backend, abelian_grouping):
```

test/python/primitives/test_backend_estimator_v2.py
```diff
@@ -388,6 +388,78 @@ def test_diff_precision(self, backend, abelian_grouping):
...
+        hamiltonian = [
+            SparsePauliOp.from_list([("II", 0), ("IZ", 0), ("XI", 0)]),
+            SparsePauliOp.from_list([("ZZ", 0)]),
+            SparsePauliOp.from_list([("ZZ", 0), ("ZZ", 0)]),
+            [
+                SparsePauliOp.from_list([("II", 0), ("IZ", 0), ("XI", 0)]),
+                SparsePauliOp.from_list([("ZZ", 0)]),
+                SparsePauliOp.from_list([("ZZ", 0), ("ZZ", 0)]),
+            ],
+        ]
```
ihincks: Avoid testing more than necessary
```suggestion
        hamiltonian = [
            SparsePauliOp.from_list([("II", 0), ("IZ", 0), ("XI", 0)]),
            SparsePauliOp.from_list([("ZZ", 0)]),
            [
                SparsePauliOp.from_list([("II", 0), ("IZ", 0), ("XI", 0)]),
                SparsePauliOp.from_list([("ZZ", 1)]),
                SparsePauliOp.from_list([("ZZ", 0), ("ZZ", 0)]),
            ],
        ]
```

test/python/primitives/test_backend_estimator_v2.py
```diff
@@ -388,6 +388,78 @@ def test_diff_precision(self, backend, abelian_grouping):
...
+        estimator = BackendEstimatorV2(backend=backend, options=self._options)
+        estimator.options.abelian_grouping = abelian_grouping
+        hamiltonian = [
+            SparsePauliOp.from_list([("II", 0), ("IZ", 0), ("XI", 0)]),
+            SparsePauliOp.from_list([("ZZ", 0)]),
+            SparsePauliOp.from_list([("ZZ", 0), ("ZZ", 0)]),
+            [
+                SparsePauliOp.from_list([("II", 0), ("IZ", 0), ("XI", 0)]),
+                SparsePauliOp.from_list([("ZZ", 0)]),
+                SparsePauliOp.from_list([("ZZ", 0), ("ZZ", 0)]),
```
ihincks: Could we get more test coverage by making one of these non-zero?

ihincks (dismissed):

test/python/primitives/test_backend_estimator_v2.py
jsaroni: Ok.

test/python/primitives/test_backend_estimator_v2.py
jsaroni: Added a zero and non-zero coefficient test case.

test/python/primitives/test_backend_estimator_v2.py
jsaroni: I removed the looping abelian_grouping and selected just one backend.

t-imamichi:
Could you apply black and check lint too?

jsaroni:
> Could you apply black and check lint too?

Applied it.

t-imamichi:
Could you replied to Ian's questions and suggestions?

test/python/primitives/test_backend_estimator_v2.py
jsaroni: Resolved.

jsaroni:
> Could you replied to Ian's questions and suggestions?

Replied. Resolved the posted comments.

t-imamichi:
Could you resolve the conflict?

jsaroni:
> Could you resolve the conflict?

It shows this section in red. I'm not sure why.
sshot

t-imamichi:
That's the conflict part you need to resolve. Could you take a look at the documentation? https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/addressing-merge-conflicts/resolving-a-merge-conflict-on-github

t-imamichi (approved):
LGTM

PR #12712, opened 2024-07-02
virtual padding for invalid section_size in synth_cnot_count_full_pmh

### LINKED ISSUE #12106: The synthesis function ``synth_cnot_count_full_pmh`` returns invalid circuits for certain ``section_size`` values

ghost:
### Summary
Suggest using "virtual padding" or matrices filled with zeros to address empty and invalid circuits when the PMH algorithm cannot handle a `section_size` larger than the number of qubits.
#12106 
### Details 

The issue in the code stems from how the algorithm handles the `section_size` parameter. The algorithm is designed to divide the input matrix into sections and process these sections iteratively. However, when the `section_size` is larger than the matrix size, there are no sections to iterate over, causing the algorithm to fail.

we can try to solve this by virtual padding, If the `section_size` exceeds the matrix size (n), we can virtually pad the matrix with additional columns of zeros to make it divisible by the `section_size`. This padding is "virtual" in the sense that it's not explicitly added to the matrix data structure, but rather accounted for in the algorithm's logic.

The changes:

- Virtual Padding: If the `section_size` is larger than the number of qubits (`num_qubits`), the matrix is padded with zeros to the right using `np.hstack`.

- **Modified Loop:** The loop in `_lwr_cnot_synth` now iterates up to `state.shape[1]` (the new padded size) to cover all sections.

- **Handling Zero Sub-rows:** The code now checks if a sub-row is all zeros `(np.sum(sub_row_patt) == 0)` and skips it if so.

- **Removing Padding:** When creating the circuit, we only add CNOT gates if both the control and target qubits are within the original `num_qubits`, effectively discarding operations on the padded columns.

- **Modified Loop Range:** The range of the inner loop in `_lwr_cnot_synth` is modified to `range((sec - 1) * section_size, min(sec * section_size, num_qubits))`. This ensures that the loop only iterates over the original columns, even if the `section_size` is larger than the matrix size.

### DIFF
--- a/qiskit/synthesis/linear/cnot_synth.py
+++ b/qiskit/synthesis/linear/cnot_synth.py
@@ -5,7 +5,6 @@
 # This code is licensed under the Apache License, Version 2.0. You may
 # obtain a copy of this license in the LICENSE.txt file in the root directory
 # of this source tree or at http://www.apache.org/licenses/LICENSE-2.0.
-#
 # Any modifications or derivative works of this code must retain this
 # copyright notice, and modified files need to carry a notice indicating
 # that they have been altered from the originals.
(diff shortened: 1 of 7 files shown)


ghost:
The circuits correctness in old vs. new code:

```
from qiskit.circuit.library import LinearFunction

for nq in range(2, 10):
    bad_section_sizes = set()

    for seed in range(100, 200):
        mat = random_invertible_binary_matrix(nq, seed=seed)

        expected_qc = QuantumCircuit(nq)
        expected_qc.append(LinearFunction(mat), range(nq))
        expected_op = Operator(expected_qc)

        for section_size in range(1, nq+1):
            qc = synth_cnot_count_full_pmh(mat, section_size=section_size)
            op = Operator(qc)
            ok = op.equiv(expected_op)
            # print(f"section size is {section_size}: {ok}")
            if not ok:
                bad_section_sizes.add(section_size)

    print(f"For {nq = }: {bad_section_sizes = }")
```

without virtual padding:

```
For nq = 2: bad_section_sizes = set()
For nq = 3: bad_section_sizes = set()
For nq = 4: bad_section_sizes = set()
For nq = 5: bad_section_sizes = {3}
For nq = 6: bad_section_sizes = {4}
For nq = 7: bad_section_sizes = {4, 5}
For nq = 8: bad_section_sizes = {3, 5, 6}
For nq = 9: bad_section_sizes = {5, 6, 7}
```

with virtual padding:

```
For nq = 2: bad_section_sizes = set()
For nq = 3: bad_section_sizes = set()
For nq = 4: bad_section_sizes = set()
For nq = 5: bad_section_sizes = set()
For nq = 6: bad_section_sizes = set()
For nq = 7: bad_section_sizes = set()
For nq = 8: bad_section_sizes = set()
For nq = 9: bad_section_sizes = set()
```

qiskit/synthesis/linear/cnot_synth.py
```diff
@@ -26,102 +25,65 @@
 def synth_cnot_count_full_pmh(
     state: list[list[bool]] | np.ndarray[bool], section_size: int = 2
 ) -> QuantumCircuit:
-    """
```
ShellyGarion: please keep this docstring and don't remove it (you can slightly update it if needed).

qiskit/synthesis/linear/cnot_synth.py
```diff
@@ -26,102 +25,65 @@
...
         i.reverse()
-    # Convert the list into a circuit of C-NOT gates
-    circ = QuantumCircuit(state.shape[0])
-    for i in circuit_u + circuit_l:
-        circ.cx(i[0], i[1])
-    return circ
-
 
-def _lwr_cnot_synth(state, section_size):
-    """
```
ShellyGarion: please keep this docstring and don't remove it (you can slightly update it if needed).

ShellyGarion (changes requested):
@Abdalla01001 - thanks for contributing to Qiskit and fixing the bug, but this PR needs some more work to fit into Qiskit standards.
I advise you to look at [Qiskit contributing guidelines](https://github.com/Qiskit/qiskit/blob/main/CONTRIBUTING.md) first.

1. Note that you should sign the CLA.

2. See the section on "Style and Lint".

3. See the section on "Release notes".

4. The functions API docs should not be removed. You can add a comment on virtual padding, but keep the existing format of the docstrings, since it appears [here](https://docs.quantum.ibm.com/api/qiskit/synthesis#synth_cnot_count_full_pmh).

5. Please also add a test to this [test file](https://github.com/Qiskit/qiskit/blob/main/test/python/synthesis/test_linear_synthesis.py) to check the correctness of the function with different values of `section_size` (which can be higher or smaller than the number of qubits).
See for example this test:
https://github.com/Qiskit/qiskit/blob/bfc69a499c035bc3b8af31b37ba0469d7e89e67a/test/python/synthesis/test_linear_synthesis.py#L118

qiskit/synthesis/linear/cnot_synth.py
```diff
@@ -26,102 +25,65 @@
...
-            f"state should be of type list or numpy.ndarray, but was of the type {type(state)}"
+            "state should be of type list or numpy.ndarray, "
+            "but was of the type {}".format(type(state))
         )
     state = np.array(state)
+    num_qubits = state.shape[0]
+
+    # Virtual Padding
+    if section_size > num_qubits:
+        padding_size = section_size - (num_qubits % section_size)
```
alexanderivrii: could this be simplified to `padding_size = section_size - num_qubits`?

ShellyGarion:
Following on [this comment](https://github.com/Qiskit/qiskit/pull/12166#issuecomment-2198521174) on the default value of `section_size` (which is currently 2).

@Cryoris noticed that in the paper they say one can choose `section_size = alpha * log2(n)` with `alpha < 1` but the PRs only tested `alpha = 1` maybe that's why the tests didn't perform any better? In the [paper](https://arxiv.org/abs/quant-ph/0302002) they do an experiment with `alpha=1/2` (but say it may be suboptimal). Perhaps you can do some further tests with several values of `alpha` ?

See also [this comment](https://github.com/Qiskit/qiskit/pull/12166#issuecomment-2205363365).

qiskit/synthesis/linear/cnot_synth.py
Abdalla01001: Done

qiskit/synthesis/linear/cnot_synth.py
Abdalla01001: Done

qiskit/synthesis/linear/cnot_synth.py
Abdalla01001: @alexanderivrii  the goal of padding is to extend the matrix to a size divisible by the `section_size`.  Simply subtracting the number of qubits from the section size does not guarantee divisibility.

ghost: @alexanderivrii the goal of padding is to extend the matrix to a size divisible by the `section_size`. Just subtracting the number of qubits from the section size does not guarantee divisibility.

qiskit/synthesis/linear/cnot_synth.py
```diff
@@ -53,25 +58,38 @@ def synth_cnot_count_full_pmh(
...
     if not isinstance(state, (list, np.ndarray)):
         raise QiskitError(
-            f"state should be of type list or numpy.ndarray, but was of the type {type(state)}"
+            "state should be of type list or numpy.ndarray, "
+            "but was of the type {}".format(type(state))
         )
     state = np.array(state)
+    num_qubits = state.shape[0]
+
+    # Virtual Padding
```
Cryoris: After discussing with @alexanderivrii and @ShellyGarion we came to the conclusion that the section size being larger than the number of qubits is an invalid input. It would be better to raise an error if `section_size > num_qubits`, as originally done in #12166. Could you do that and remove the virtual padding? 

We'll go forward with your PR then, as it still fixes the bad `section_size` values due to your fixes on L131 and L148.

ghost: That's right, I did it

ShellyGarion:
Since the general plan is to port this code to Rust and we have a time pressure to make it to Qiskit 1.2 release, I'll close this PR is favor of https://github.com/Qiskit/qiskit/pull/12588.
Both @Abdalla01001 and @Tarun-Kumar07 were added as co-authors of PR https://github.com/Qiskit/qiskit/pull/12588. 
We very much appreciate your suggestions and efforts in fixing this bug!

PR #12713, opened 2024-07-03
Fixed the _skip_target argument in `generate_preset_pass_manager` 

Siddharthgolecha:
the _skip_target argument was set in config but was not exactly used

### Summary
The `generate_preset_pass_manager` configured Pass Manager configurations from backend but `_skip_target` parameter was not correctly configured. 

### Details and comments
While building the Pass Manager configurations from backend while generating the preset pass manager, the `_skip_target` parameter was not correctly configured. It was added in the `pm_options` but in `from_backend` function argument `_skip_target` was directly used instead of `_skip_target` from the `pm_options`. So, I passed the argument in the function and kept the original option in the `pm_options` just in case it is being used in future pass managers

### DIFF
--- a/qiskit/transpiler/preset_passmanagers/__init__.py
+++ b/qiskit/transpiler/preset_passmanagers/__init__.py
@@ -389,7 +389,7 @@ def generate_preset_pass_manager(
 
     if backend is not None:
         pm_options["_skip_target"] = _skip_target
-        pm_config = PassManagerConfig.from_backend(backend, **pm_options)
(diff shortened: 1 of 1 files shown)


ElePT:
Hi @Siddharthgolecha, you can see from the current failed tests that `_skip_target` was used, as it's one of the keyword arguments specified through `**pm_options`. This change causes the following error:
```
TypeError: from_backend() got multiple values for keyword argument '_skip_target'
```
I will close this PR as there is no need for the change, thanks for your interest.

PR #12714, opened 2024-07-03
Port `synth_clifford_bm` to Rust.

### LINKED ISSUE #12234: Port `synth_clifford_bm` to Rust

alexanderivrii:
### Summary

Closes #12234. 

This ports the function `synth_clifford_bm` to Rust. Recall that this function optimally synthesizes ``Clifford`` operators on 1, 2 or 3 qubits with respect to the number of CX-gates.

Here are the performance comparisons on my laptop between Python and Rust versions for 1000 random `n`-qubit Cliffords, with `n=1, 2, 3`.

1000 random 1-qubit cliffords:
Python version: 0.05
Rust version: 0.02
Improvement: 2.5x

1000 random 2-qubit cliffords:
Python version: 14.68
Rust version: 0.03
Improvement: 400x.

1000 random 3-qubit cliffords:
Python version: 30.95
Rust version: 0.38
Improvement: 80x

### Details and comments

The circuits produced by the Python and Rust versions should be identical (up to register naming).

This is an almost line-by-line port of the Python file. The only minor difference is that the function `reduce_cost` returns the list of inverted gates, so that the gates in the final circuit only needs to be reversed and not both reversed and inverted. 

There are a few places where I think the Rust code could be improved, but I did not find a good solution. I will highlight this as separate comments and ask for feedback/suggestions.


crates/accelerate/src/synthesis/clifford/bm_synthesis.rs
```diff
@@ -0,0 +1,377 @@
...
+/// also inverted).
+fn reduce_cost(
+    cliff: &Clifford,
+    cost: usize,
+    gates: &mut CliffordGatesVec,
+) -> Result<(Clifford, usize), String> {
+    // All choices for a 2-qubit block
+    for qubit0 in 0..cliff.num_qubits {
+        for qubit1 in qubit0 + 1..cliff.num_qubits {
+            for (n0, n1) in iproduct!(0..3, 0..3) {
```
alexanderivrii, line 151: **Feedback wanted!** I very much like the `iproduct!` macro as it allows to write concisely nested `for` loops. Would it be possible to put `for` loops for `qubit0` and `qubit1` inside the same macro? Note that `qubit1` iterates from `qubit0+1`.

crates/accelerate/src/synthesis/clifford/bm_synthesis.rs
```diff
@@ -0,0 +1,377 @@
...
+        );
+        let mut mask = arr1(&[false, false, false, false, false, false]);
+        mask[q2] = true;
+        mask[q2 + 3] = true;
+        let xs = clifford.tableau.slice(s![q1, 0..6]).to_owned();
+        let zs = clifford.tableau.slice(s![q1 + 3, 0..6]).to_owned();
+        let ys = xs.clone() ^ zs.clone();
+        let is_loc_x = (xs.clone() & mask.clone()) == xs;
+        let is_loc_z = (zs.clone() & mask.clone()) == zs;
+        let is_loc_y = (ys.clone() & mask.clone()) == ys;
```
alexanderivrii: **Feedback wanted!**. This is almost a 1-1 rewrite of the Python code. I had to use `to_owned` and `clone` to make this compile. This is probably not a huge deal since all the arrays have only 6 elements, but I would like to get rid of this extra copying if possible. I also do not understand why `let ys = xs ^ zx;` does not compile (it seems that it needs to move both `xs` or `ys`). Of course, checking that `x & mask == x` is equivalent to checking that `x` has no nonzero-elements on non-masked qubits, leading to an alternative implementation, but the current form is probably a bit clearer.

ShellyGarion (commented):
LGTM. The functionality looks ok to me. thanks!

crates/accelerate/src/synthesis/clifford/bm_synthesis.rs
raynelfss: I think you can get away with passing most of these by reference, even your views of `xs` and `zs`. Try this:

```suggestion
        let xs = clifford.tableau.slice(s![q1, 0..6]);
        let zs = clifford.tableau.slice(s![q1 + 3, 0..6]);
        let ys = &xs ^ &zs;
        let is_loc_x = (&xs & &mask) == xs;
        let is_loc_z = (&zs & &mask) == zs;
        let is_loc_y = (&ys & &mask) == ys;
```

alexanderivrii: Thank you, @raynelfss! This solves the problem. Done in [a89ed9e](https://github.com/Qiskit/qiskit/pull/12714/commits/a89ed9e3343c380d55d1030089bd212962fb169b).

crates/accelerate/src/synthesis/clifford/bm_synthesis.rs
```diff
@@ -0,0 +1,377 @@
...
+    } else {
+        0
+    }
+}
+
+/// Return the number of CX gates required for Clifford decomposition
+/// for a 3-qubit Clifford.
+fn cx_cost3(clifford: &Clifford) -> usize {
+    // create information transfer matrices r1, r2
+    let mut r1 = arr2(&[[0, 0, 0], [0, 0, 0], [0, 0, 0]]);
```
Cryoris, line 67: You could also use the designated `Array::zero` function here if you want to type less 0s, as
```suggestion
    let mut r1 = Array2::<u32>::zeros((3, 3));
```
🙂

crates/accelerate/src/synthesis/clifford/bm_synthesis.rs
Cryoris: With `iproduct` probably not as qubit0 and qubit1 are not independent... unless you'd want to do something like `iproduct!(0..n, 0..n, 0..3, 0..3).filter(|(q0, q1, n0, n1)| q1 > q0)` -- but it seems to me being explicit like you have it is benevolent for people reading the code in the future 😉

Cryoris (approved):
LGTM as well!

ShellyGarion (approved):

PR #12715, opened 2024-07-03
fixing synthesis release notes

alexanderivrii:
### Summary

Just a few days ago @mtreinish has reminded me (yet again) about the difference between an upgrade note and a feature note:

> An upgrade note is reserved for something that we want to document might require user input when upgrading from the previous release to the new one. Examples would be a breaking api change, minimum dependency version bump, etc.

and of course we forgot about this in some of the recently merged PRs (probably because we keep copying and adapting the notes from previously merged PRs). 

This tiny PRs classifies the release notes correctly. And attention @Cryoris with #12588.


Cryoris (approved):
Noted 😉

PR #12716, opened 2024-07-03
Add ``PySequenceIndex::convert_index`` to convert Python-like indices to unsigned

Cryoris:
### Summary

This allows to convert a Python-like signed index into an unsigned index suitable in Rust notation.

### Details and comments

For example
```rust
use qiskit_circuit::slice::PySequenceIndex;

let vec = // some vector...
let py_index = -2;
let index = PySequenceIndex::convert_idx(py_index, vec.len()).unwrap();
assert_eq!(index, vec.len() - 2);
```
This is particularly useful in some places where we port Python code to Rust, and Python index conventions are used.

### DIFF
--- a/crates/circuit/src/slice.rs
+++ b/crates/circuit/src/slice.rs
@@ -46,17 +46,8 @@ impl<'py> PySequenceIndex<'py> {
     pub fn with_len(&self, len: usize) -> Result<SequenceIndex, PySequenceIndexError> {
         match self {
             PySequenceIndex::Int(index) => {
-                let index = if *index >= 0 {
-                    let index = *index as usize;
-                    if index >= len {
-                        return Err(PySequenceIndexError::OutOfRange);
-                    }
-                    index
-                } else {
-                    len.checked_sub(index.unsigned_abs())
-                        .ok_or(PySequenceIndexError::OutOfRange)?
-                };
-                Ok(SequenceIndex::Int(index))
+                let wrapped_index = PySequenceIndex::convert_idx(*index, len)?;
+                Ok(SequenceIndex::Int(wrapped_index))
             }
             PySequenceIndex::Slice(slice) => {
                 let indices = slice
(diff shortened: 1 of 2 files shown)


crates/circuit/src/slice.rs
```diff
@@ -80,6 +80,15 @@ impl<'py> PySequenceIndex<'py> {
...
     }
+
+    /// Given an integer (which may be negative) get a valid unsigned index for a sequence.
+    pub fn convert_idx(index: isize, length: usize) -> Result<usize, PySequenceIndexError> {
+        let wrapped_index = match PySequenceIndex::Int(index).with_len(length)? {
+            SequenceIndex::Int(result) => result,
+            _ => unreachable!(), // we will always match on Int(..)
+        };
+        Ok(wrapped_index)
+    }
```
jakelishman: It might be marginally cleaner to pull out this logic from `with_len`, and have `with_len` call this instead.

crates/circuit/src/slice.rs
```diff
@@ -372,4 +382,37 @@ mod test {
...
+
+    /// Test SequenceIndex::from_int correctly handles positive and negative indices
+    #[test]
+    fn convert_py_idx() {
+        let cases: IntoIter<(isize, usize, usize), 3> = [
+            (2, 5, 2), // (index, sequence length, expected result)
+            (-2, 5, 3),
+            (0, 2, 0),
+        ]
+        .into_iter();
```
jakelishman: You don't need to manually call `into_iter` - `for` loops do it for you.  I think you could drop the type annotation and the `into_iter()` call, and it'd all work the way you want.

If Rust struggles with determining the integer types, you can make the first tuple super explicit as `(2isize, 5usize, 2usize)` and then it should be happy.

jakelishman (commented):
Thanks for doing this - looks like a helpful idea!

crates/circuit/src/slice.rs
Cryoris: Ah yes, then we can avoid the `unreachable!` branch, good idea! Done in [f9c114c](https://github.com/Qiskit/qiskit/pull/12716/commits/f9c114cfaefbeccdfc3d9fda5af4a2c4b1849afe)

crates/circuit/src/slice.rs
Cryoris: Great, thanks for the hint! Done in [f9c114c](https://github.com/Qiskit/qiskit/pull/12716/commits/f9c114cfaefbeccdfc3d9fda5af4a2c4b1849afe).

jakelishman (approved):
Looks good, thanks for doing this!

PR #12719, opened 2024-07-03
Optimize DAGCircuit.collect_1q_runs() filter function

mtreinish:
### Summary

In the recently merged #12650 a new rust function was added for the filter function of the collect_1q_runs() method's internal filter function. As part of this we need to do exclude any non-DAGOpNode dag nodes passed to the filter function. This was previously implemented using the pyo3 extract() method so that if the pyobject can be coerced into a DAGOpNode for later use we continue but if it errors we know the input object is not a DAGOpNode and return false. However, as we don't need to return the error to python we should be using downcast instead of extract (as per the pyo3 performance guide [1]) to avoid the error conversion cost. This was an oversight in #12650 and I only used extract() out of habit.

[1] https://pyo3.rs/v0.22.0/performance#extract-versus-downcast


kevinhartman (approved):
LGTM!

PR #11370, opened 2023-12-04
Optimize standard gate library equality

mtreinish:
### Summary

This commit updates the standard library gate classes' to define an __eq__ methods to optimize the runtime performance. Their parent class `Instruction`'s __eq__ is optimized for generality to compare equality between any gate instruction object. But this generality has a runtime cost which is unnecessary for the standard library gates which it equality is much simpler. This commit adds the faster path equality check for the standard gate library classes which will improve the performance of anything doing a lot gate equality.

### Details and comments

TODO:

- [x] Add UGate and other missing gates

PR #12721, opened 2024-07-04
Remove some of the entries in allow_DeprecationWarning

1ucian0:
### Summary

Tests have a list of `filterwarnings` that have been accumulating with time. Most of them make little sense nowadays. Let's clean them up.

### DIFF
--- a/qiskit/circuit/instruction.py
+++ b/qiskit/circuit/instruction.py
@@ -188,11 +188,12 @@ class Instruction(Operation):
             return False
 
         for self_param, other_param in zip_longest(self.params, other.params):
-            try:
+            if isinstance(self_param, numpy.ndarray):
+                if numpy.array_equal(self_param, other_param):
+                    continue
+            else:
                 if self_param == other_param:
                     continue
-            except ValueError:
-                pass
 
             try:
                 self_asarray = numpy.asarray(self_param)
--- a/test/utils/base.py
+++ b/test/utils/base.py
@@ -244,12 +244,8 @@ class QiskitTestCase(BaseQiskitTestCase):
         #  Ideally, use
         # `allow_DeprecationWarning_message` instead, because it can be much more specific.
         allow_DeprecationWarning_modules = [
-            "test.python.pulse.test_builder",
-            "test.python.pulse.test_block",
-            "qiskit.pulse.schedule",
-            "qiskit.pulse.instructions.instruction",
-            "qiskit.pulse.instructions.play",
-            "qiskit.quantum_info.operators.symplectic.pauli",
+            # Example - Because the deprecation of X #<issue/PR>. Can be removed once X is removed.
+            # "qiskit.quantum_info.operators.symplectic.pauli",
         ]
         for mod in allow_DeprecationWarning_modules:
             warnings.filterwarnings("default", category=DeprecationWarning, module=mod)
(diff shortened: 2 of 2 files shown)


test/utils/base.py
```diff
@@ -238,30 +238,24 @@ def setUpClass(cls):
...
             "qiskit.pulse.instructions.instruction",
             "qiskit.pulse.instructions.play",
-            "qiskit.pulse.library.parametric_pulses",
             "qiskit.quantum_info.operators.symplectic.pauli",
         ]
         for mod in allow_DeprecationWarning_modules:
             warnings.filterwarnings("default", category=DeprecationWarning, module=mod)
+
         allow_DeprecationWarning_message = [
-            r"elementwise comparison failed.*",
```
1ucian0, line 259: The fix in `qiskit/circuit/instruction.py` makes this entry unnecessary

qiskit/circuit/instruction.py
```diff
@@ -188,11 +188,12 @@ def __eq__(self, other):
...
         for self_param, other_param in zip_longest(self.params, other.params):
-            try:
+            if isinstance(self_param, numpy.ndarray):
+                if numpy.array_equal(self_param, other_param):
+                    continue
+            else:
                 if self_param == other_param:
                     continue
-            except ValueError:
-                pass
```
jakelishman, line 195: This _might_ have had an effect on timings, but since we gave standard gates their own `__eq__` overrides in #11370 it shouldn't make any difference at all (and when we move to Rust properly, it really won't).

test/utils/base.py
```diff
@@ -232,36 +232,26 @@ def setUpClass(cls):
         # Safe to remove once https://github.com/Qiskit/qiskit-aer/pull/2179 is in a release version
         # of Aer.
         warnings.filterwarnings(
-            "default",
+            "ignore",
```
jakelishman: I very very gently prefer `default` here personally, because it is a true warning (as opposed to the `__warningregistry__` one above, which is triggering falsely), and showing it means it's still at least _somewhat_ visible.

I don't feel strongly, though.

test/utils/base.py
```diff
@@ -232,36 +232,26 @@ def setUpClass(cls):
...
-            "qiskit.pulse.schedule",
-            "qiskit.pulse.instructions.instruction",
-            "qiskit.pulse.instructions.play",
-            "qiskit.pulse.library.parametric_pulses",
-            "qiskit.quantum_info.operators.symplectic.pauli",
+            # Example - Because the deprecation of X #<issue/PR>. Can be removed once X is removed.
+            # "qiskit.quantum_info.operators.symplectic.pauli",
         ]
         for mod in allow_DeprecationWarning_modules:
             warnings.filterwarnings("default", category=DeprecationWarning, module=mod)
```
jakelishman: We might consider removing this entirely; there's very little reason to ever blanket-allow warnings from an entire module, so imo we shouldn't make it easy to do like this.  The examples above with explicit and very tight warning filters are better.

jakelishman (commented):
Thanks for looking at this Luciano - it's great to have this debt cleared up.

test/utils/base.py
1ucian0: reverted in cb1d6ef170ae85cd1399410d34f0bacb2f753a7e

test/utils/base.py
1ucian0: done in cb1d6ef170ae85cd1399410d34f0bacb2f753a7e

qiskit/circuit/instruction.py
1ucian0: I was unsure about the performance implications of this. If the answer is none-in-the-future, I would prefer to avoid the `==` for ndarrays here.

jakelishman (approved):
Thanks for this tidying - long overdue!

PR #12722, opened 2024-07-04
Randomized test errors are failing because Aer uses deprecated functionality

1ucian0:
The randomized tests are [failing](https://github.com/Qiskit/qiskit/issues/2645#issuecomment-2201834078) because Aer uses deprecated functionality.

Maybe randomized test should not fail on `DeprecationWarning`s?


ElePT (approved):
Let's at least merge this fix now. I guess there is still some value to randomized tests failing on deprecation warnings, if only to make us aware that they are being raised (for example, by Aer).

PR #12723, opened 2024-07-04
Revert "set a dummy QISKIT_SETTINGS so tests do not interact with local env"

1ucian0:
Reverts Qiskit/qiskit#12463, following https://github.com/Qiskit/qiskit/pull/12463#issuecomment-2208626799

Tox should take care of env isolation.


ElePT (approved):
Given that I approved the PR originally, I guess it's on me to revert it too :P (and start using tox).

PR #12727, opened 2024-07-05
'Peephole' optimization - or: collecting and optimizing two-qubit blocks - before routing

### LINKED ISSUE #12562: Apply Peephole Optimization Before Routing
### What should we add?

We should investigate to apply peephole optimization before routing.

sbrandhsn:
### Summary

Fixes #12562 

### Details and comments

This PR collects and consolidates two-qubit gates into blocks before routing. Each block is then subsequently analysed to determine whether the two-qubit unitary representing the block equals the identity or is a product of single-qubit gates. In the former case, the gates in the block are removed and in the latter case the block is replaced by single-qubit gates correspondingly.

Points up for debate:
* Do we want to run the new pass `Split2QUnitaries` as part of `ConsolidatedBlocks`?
* Do the passes for gate direction checks and fixes always play nicely with having two-qubit unitaries before routing? There were no test failures but I'm still wondering.
* Is the current path of checking for the two-qubit unitaries for being a product of single-qubit unitaries the fastest or should we rely on Weyl-decomposition? If the former is true, we should probably port the corresponding functions to rust as they will touch a lot of two-qubit unitaries. :-)


mtreinish:
> Is the current path of checking for the two-qubit unitaries for being a product of single-qubit unitaries the fastest or should we rely on Weyl-decomposition? If the former is true, we should probably port the corresponding functions to rust as they will touch a lot of two-qubit unitaries. :-)

I took care of this in: https://github.com/Qiskit/qiskit/pull/12739 since it was very simple. From a complexity PoV I would expect the `local_invariants()` function to be much faster since it's far less complex the the `TwoQubitWeylDecomposition` class.

qiskit/transpiler/passes/optimization/split_2q_unitaries.py
```diff
@@ -0,0 +1,49 @@
...
+
+    def run(self, dag: DAGCircuit):
+        """Run the Split2QUnitaries pass on `dag`."""
+        for node in dag.topological_op_nodes():
+            # skip operations without two-qubits and for which we can not determine a potential 1q split
+            if (
+                len(node.cargs) > 0
+                or len(node.qargs) != 2
+                or not (hasattr(node.op, "to_matrix") and hasattr(node.op, "__array__"))
+                or (hasattr(node.op, "is_parameterized") and node.op.is_parameterized())
```
mtreinish:
```suggestion
                or node.matrix is None
                or node.is_parameterized())
```

qiskit/transpiler/passes/optimization/split_2q_unitaries.py
```diff
@@ -0,0 +1,49 @@
...
+            ):
+                continue
+
+            # check if the node can be represented by single-qubit gates
+            nmat = node.op.to_matrix()
```
mtreinish:
```suggestion
            nmat = node.matrix
```

test/python/compiler/test_transpiler.py
```diff
@@ -109,6 +111,16 @@ def _define(self):
         self._definition = QuantumCircuit(2)
         self._definition.cx(0, 1)
 
+    def to_matrix(self) -> np.ndarray:
```
mtreinish: Why is this needed? Should this skip the gate if the matrix is empty?

qiskit/transpiler/passes/optimization/split_2q_unitaries.py
```diff
@@ -0,0 +1,49 @@
...
+
+class Split2QUnitaries(TransformationPass):
+    """Splits each two-qubit gate in the `dag` into two single-qubit gates, if possible without error."""
+
+    def run(self, dag: DAGCircuit):
+        """Run the Split2QUnitaries pass on `dag`."""
+        for node in dag.topological_op_nodes():
+            # skip operations without two-qubits and for which we can not determine a potential 1q split
+            if (
+                len(node.cargs) > 0
```
mtreinish, line 46: Do you think it makes sense to have a unitary gate name check here with:

```suggestion
                node.name != "unitary"
                or len(node.cargs) > 0
```

Or are you thinking there is a case where we have a another 2q gate type we can decompose into a product of 1q gates?

qiskit/transpiler/passes/optimization/split_2q_unitaries.py
```diff
@@ -0,0 +1,49 @@
...
+            nmat = node.op.to_matrix()
+            if np.all(two_qubit_local_invariants(nmat) == [1, 0, 3]):
+                ul, ur, phase = decompose_two_qubit_product_gate(nmat)
+                dag_node = DAGCircuit()
+                dag_node.add_qubits(node.qargs)
+
+                dag_node.apply_operation_back(UnitaryGate(ur), qargs=(node.qargs[0],))
+                dag_node.apply_operation_back(UnitaryGate(ul), qargs=(node.qargs[1],))
+                dag_node.global_phase += phase
+                dag.substitute_node_with_dag(node, dag_node)
```
mtreinish: While we can use substitute node with dag here, it'll probably be a bit more performant to do something like:

https://github.com/Qiskit/qiskit/blob/main/qiskit/transpiler/passes/optimization/optimize_1q_decomposition.py#L244-L248

Where we just insert the decomposition gate before the 2q gate and then just remove the node. This should be a bit faster because it avoids all the extra complexity in the dag substitution code.

qiskit/transpiler/preset_passmanagers/builtin_plugins.py
```diff
@@ -154,6 +159,49 @@ def pass_manager(self, pass_manager_config, optimization_level=None) -> PassMana
...
+                    if (
+                        op is None
+                        or not isinstance(op, Instruction)
+                        or op.name in _discrete_skipped_ops
+                    ):
+                        continue
+
+                    if len(op.params) > 0:
+                        return True
+                return False
```
mtreinish: We should check with @wshanks to see if this will work for the experiments RB case we hit before with #11354. I'm mostly wondering if they're using rz and allowing the angles to be multiples of `pi/2`. It was long enough ago I don't remember the specific details of what the input was.

releasenotes/notes/peephole-before-routing-c3d184b740bb7a8b.yaml
```diff
@@ -0,0 +1,16 @@
...
+    the passes :class:`.Collect2qBlocks`, :class:`.ConsolidateBlocks` and :class:`.Split2QUnitaries` are
+    added to the `init` stage of the preset pass managers with optimization level 2 and optimization level 3
+    if the pass managers target a :class:`.Target` that has a discrete basis gate set, i.e. all basis gates
+    have are not parameterized. The modification of the `init` stage should allow for a more efficient routing
+    for quantum circuits that (a) contain two-qubit unitaries/gates that are actually a product of single-qubit gates
+    or (b) contain multiple two-qubit gates in a continuous block of two-qubit gates. In the former case,
+    the routing of the two-qubit gate can simply be skipped as no real interaction between a pair of qubits
+    occurs. In the latter case, the lookahead space of routing algorithms is not 'polluted' by superfluous
+    two-qubit gates, i.e. for routing it is sufficient to only consider one single two-qubit gate per
+    continuous block of two-qubit gates.
```
mtreinish: I would split this into two release notes (with separate bullet points under features). One documenting the new pass, the other documenting the modifications to the preset pass manager.

releasenotes/notes/peephole-before-routing-c3d184b740bb7a8b.yaml
```diff
@@ -0,0 +1,16 @@
+---
+features:
```
mtreinish:
```suggestion
features_transpiler:
```

qiskit/transpiler/passes/optimization/split_2q_unitaries.py
```diff
@@ -0,0 +1,49 @@
...
+                len(node.cargs) > 0
+                or len(node.qargs) != 2
+                or not (hasattr(node.op, "to_matrix") and hasattr(node.op, "__array__"))
+                or (hasattr(node.op, "is_parameterized") and node.op.is_parameterized())
+            ):
+                continue
+
+            # check if the node can be represented by single-qubit gates
+            nmat = node.op.to_matrix()
+            if np.all(two_qubit_local_invariants(nmat) == [1, 0, 3]):
```
mtreinish: In this case it might be faster to do this manually with:

```suggestion
            local_invairants = two_qubit_local_invariants(nmat)
            if local_invariants[0] == 1 and local_invariants[1] == 0 and local_invariants == 3:
```

since there are only 3 floating point values I would expect the numpy overhead from `np.all()` to actually be higher than just doing the comparisons directly. Also do you need any tolerance on the comparisons here?

qiskit/transpiler/passes/optimization/split_2q_unitaries.py
```diff
@@ -0,0 +1,49 @@
...
+                ul, ur, phase = decompose_two_qubit_product_gate(nmat)
+                dag_node = DAGCircuit()
+                dag_node.add_qubits(node.qargs)
+
+                dag_node.apply_operation_back(UnitaryGate(ur), qargs=(node.qargs[0],))
+                dag_node.apply_operation_back(UnitaryGate(ul), qargs=(node.qargs[1],))
```
mtreinish:
```suggestion
                dag_node.apply_operation_back(UnitaryGate(ur, check_input=False), qargs=(node.qargs[0],))
                dag_node.apply_operation_back(UnitaryGate(ul, check_input=False), qargs=(node.qargs[1],))
```

mtreinish (commented):
Overall this looks great to me, I really like this idea of collecting 2q unitary blocks before routing and deferring synthesis until the translation phase. It avoids the potential risk of early synthesis making routing worse and will potentially make routing faster because we have reduced the total gate count.

I left some inline comments, on the new pass itself most of the suggestions are adjusting the access patterns to get attributes from the `DAGOpNode` directly instead of using the python op. Although some of these access patterns will shift after #12730 merges so we should keep an eye on potential conflicts there. The other thing is after #12739 merges and `decompose_two_qubit_product_gate()` is ported to rust we actually can move the majority of this pass to rust, where we pass a list of `DAGOpNodes` to rust do the filtering and resynthesis in rust, and return a list of 1q unitary pairs. This is basically what's done in `Optimize1qGatesDecomposition` now and I expect that it will be faster than doing it all python side. But that can wait until we have the necessary features in rust.

qiskit/transpiler/preset_passmanagers/builtin_plugins.py
wshanks: We do have to use rz with multiples of `pi/2` for RB since the only singles in most IBM targets are id, sx, x, and rz. I don't see a reference to RB in #11354. What is the concern?

mtreinish: We rolled back #11354 originally because it was causing issues with experiments. Since running synthesis before layout required us to not specify a target basis gate set and it was basically converting all 1q sequences of gates into `U` gates. So we quickly reverted it: https://github.com/Qiskit/qiskit/pull/11360 (and skipped an issue) to unblock experiments job against main. That's what this function and the if statement calling it are trying to guard against.

But now that I wrote it out, I'm not sure we actually need to guard against this with the way this implementation is being done. Since we're not actually synthesizing the unitaries until after layout and just collecting the blocks before layout, the synthesis should still target the correct basis and it probably would work as well as it did prior to this PR. So we probably can drop this conditional logic altogether because of the synthesis doesn't work with a discrete basis here then it wouldn't have worked before.

wshanks: Okay, so it sounds like we are okay, but I could try running the experiments tests with this PR if you want.

I looked up the details of the previous issue. It was a niche case. In the experiments tests, there is a test that sets the basis to `["h", "s", "x"]` and that test was failing because the conversion to parameterized gates couldn't be mapped back to the discrete basis (it seems like it should still be possible in principle?). In practice we usually run against IBM backends with rz.

At the time I noted that the following code would trigger the error:

```python


circ = QuantumCircuit(1)
circ.s(0)
circ.y(0)

transpile(circ, basis_gates=["h", "s", "x"])
```

qiskit/transpiler/passes/optimization/split_2q_unitaries.py
sbrandhsn: I would suggest to no limit this pass to unitary gates, as this is more general and appears to work well. We can process conditional rotation gates with a zero gate here or look into any two-qubit gate that was not picked up by `ConsolidateBlocks` for any reason when not limiting to unitary gates only.

qiskit/transpiler/passes/optimization/split_2q_unitaries.py
sbrandhsn: We do not appear to require any tolerance on the comparison here. `DAGOpNode(op=Instruction(name='cp', num_qubits=2, num_clbits=0, params=[7.490140565847857e-07])` would also yield exactly `[1, 0, 3]`

qiskit/transpiler/passes/optimization/split_2q_unitaries.py
sbrandhsn: Sounds good, thanks. I will modify this.

qiskit/transpiler/preset_passmanagers/builtin_plugins.py
sbrandhsn: This pass would not run with `basis_gates=["h", "s", "x"]`, so we are probably good to go? :-) Thanks @wshanks !

test/python/compiler/test_transpiler.py
sbrandhsn: Looking into this - a couple of tests were failing...

sbrandhsn (commented):
Sounds good, thanks for the review! I will change the release notes, modify the dag substitution and revisit the failing test cases caused by missing `to_matrix`

qiskit/transpiler/preset_passmanagers/builtin_plugins.py
mtreinish: Well I think the question is do we need to have this filter at all anymore? I'm thinking we might be ok just skipping this conditional logic but instead always run these 3 passes during init at level 2 and 3. It would be a potential issue if we were going to synthesize anything pre-layout we might avoid all the issues we had in the earlier attempt.

sbrandhsn: It may be worthwhile to leave it in for any user that wants to play with Clifford circuits, right?

test/python/transpiler/test_preset_passmanagers.py
```diff
@@ -126,14 +127,14 @@ def test_layout_3239(self, level=3):
         qc.measure(range(4), range(4))
         result = transpile(
             qc,
-            basis_gates=["u1", "u2", "u3", "cx"],
+            basis_gates=["h", "z", "cx"],
```
mtreinish: I have the same question here as on the other tests, we shouldn't need to change the input side here. Especially since this is recreating a documented issue we don't want to regress it. Is there a reason this change was needed?

test/python/transpiler/test_preset_passmanagers.py
```diff
@@ -1455,6 +1456,16 @@ def _define(self):
...
 
+            def to_matrix(self) -> np.ndarray:
+                return np.asarray(
+                    [
+                        [1.0 + 0.0j, 0.0 + 0.0j, 0.0 + 0.0j, 0.0 + 0.0j],
+                        [0.0 + 0.0j, 0.0 + 0.0j, 0.0 + 0.0j, 1.0 + 0.0j],
+                        [0.0 + 0.0j, 0.0 + 0.0j, 1.0 + 0.0j, 0.0 + 0.0j],
+                        [0.0 + 0.0j, 1.0 + 0.0j, 0.0 + 0.0j, 0.0 + 0.0j],
+                    ]
+                )
```
mtreinish: Same question here, we shouldn't need a custom matrix definition here afaict.

releasenotes/notes/peephole-before-routing-c3d184b740bb7a8b.yaml
```diff
@@ -0,0 +1,18 @@
...
+    the passes :class:`.Collect2qBlocks`, :class:`.ConsolidateBlocks` and :class:`.Split2QUnitaries` are
+    added to the `init` stage of the preset pass managers with optimization level 2 and optimization level 3
+    if the pass managers target a :class:`.Target` that has a discrete basis gate set, i.e. all basis gates
+    have are not parameterized. The modification of the `init` stage should allow for a more efficient routing
+    for quantum circuits that (a) contain two-qubit unitaries/gates that are actually a product of single-qubit gates
+    or (b) contain multiple two-qubit gates in a continuous block of two-qubit gates. In the former case,
+    the routing of the two-qubit gate can simply be skipped as no real interaction between a pair of qubits
+    occurs. In the latter case, the lookahead space of routing algorithms is not 'polluted' by superfluous
+    two-qubit gates, i.e. for routing it is sufficient to only consider one single two-qubit gate per
+    continuous block of two-qubit gates.
```
mtreinish:
```suggestion
    The passes :class:`.Collect2qBlocks`, :class:`.ConsolidateBlocks` and :class:`.Split2QUnitaries` have been
    added to the ``init`` stage of the preset pass managers with optimization level 2 and optimization level 3.
    The modification of the `init` stage should allow for a more efficient routing for quantum circuits that either:
    
      * contain two-qubit unitaries/gates that are actually a product of single-qubit gates
      * contain multiple two-qubit gates in a continuous block of two-qubit gates.
      
    In the former case, the routing of the two-qubit gate can simply be skipped as no real interaction
    between a pair of qubits occurs. In the latter case, the lookahead space of routing algorithms is not
    'polluted' by superfluous two-qubit gates, i.e. for routing it is sufficient to only consider one single
    two-qubit gate per continuous block of two-qubit gates. These passes are not run if the pass
    managers target a :class:`.Target` that has a discrete basis gate set, i.e. all basis gates have are not
    parameterized.
```

qiskit/transpiler/preset_passmanagers/builtin_plugins.py
mtreinish: Let's leave it for now then. But, my thinking was it wouldn't make a difference because we'll collect these gates post-routing as part of the peephole optimization we run during the optimization stage so filtering here doesn't make much of a difference.

mtreinish (commented):
Overall I think this looks about ready. I just left a couple of nits inline about the tests, and the release notes. The other thing I think would be good to see before approving and merging this are some benchmark results to see what the impact on runtime and output quality are more broadly. We can use some of the asv benchmarks to compare this against main for some transpile cases.

qiskit/synthesis/two_qubit/two_qubit_decompose.py
```diff
@@ -113,7 +113,7 @@ def decompose_two_qubit_product_gate(special_unitary_matrix: np.ndarray):
 
     temp = np.kron(L, R)
     deviation = abs(abs(temp.conj().T.dot(special_unitary_matrix).trace()) - 4)
-    if deviation > 1.0e-13:
+    if deviation > 1.0e-8:
```
levbishop: :-( why? I worked so hard to make these functions maintain precision

qiskit/synthesis/two_qubit/local_invariance.py
```diff
@@ -65,7 +65,7 @@ def two_qubit_local_invariants(U: np.ndarray) -> np.ndarray:
     # to better equate to the Weyl chamber coordinates (c0,c1,c2)
     # and explore the parameter space.
     # Also do a FP trick -0.0 + 0.0 = 0.0
-    return np.round([G1.real, G1.imag, G2.real], 12) + 0.0
+    return np.round([G1.real, G1.imag, G2.real], 9) + 0.0
```
levbishop:
```suggestion
    return [G1.real, G1.imag, G2.real]
```
Or just drop this wrapper completely, since it's pretty unnecessary without the rounding. 

I guess you loosened this rounding to catch more approximately factorized gates but I think that's better done with a tolerance at the point of comparison rather than losing precision for all users of this function.

I know this tighter tolerance was pre existing behavior but it makes much more sense to go the other way and remove rounding completely here - i don't understand why we want to throw away information.

qiskit/transpiler/passes/optimization/split_2q_unitaries.py
```diff
@@ -0,0 +1,65 @@
...
+                or len(node.qargs) != 2
+                or node.matrix is None
+                or node.is_parameterized()
+            ):
+                continue
+
+            # check if the node can be represented by single-qubit gates
+            nmat = node.matrix
+            local_invariants = two_qubit_local_invariants(nmat)
+            if local_invariants[0] == 1 and local_invariants[1] == 0 and local_invariants[2] == 3:
```
levbishop: If needed, I think much better to use a tolerance for the comparison here in place of rounding the Makhlin params for everyone?

levbishop (dismissed):

levbishop:
In general this is something I'm in favor of and was meaning to do for a long time. The specific implementation I don't like throwing away precision for all users of functions, just to support this pass having a tolerance. I think a better approach would be to use TwoQubitWeylDecomposition which has a proper fidelity parameter that can give a controlled infidelity consistently applied in both this early splitting and also in the final synthesis. Also I'm not sure that the decompose two qubit product gate with such a loose tolerance is giving the closest product approximation to the original unitary (the resulting 1q gates might not even be themselves unitary) whereas the TwoQubitWeylDecomposition do give the the closest available product of 1q unitary.

Another advantage of using TwoQubitWeylDecomposition is that it would also spot the case where a unitary is single qubit + SWAP which can be split with a wire permutation at this pre routing stage. 

Understood  that TwoQubitWeylDecomposition is quite a bit slower than just getting the invariants, but it's already going to get called on each unitary later in the synthesis stage. (In principle you could refactor these two calls into a single call by inserting TwoQubitWeylDecomposition objects into the Dag rather than unitary objects, but that's a big change including  to the synthesis passes.)

qiskit/transpiler/passes/optimization/split_2q_unitaries.py
levbishop: Could check for SWAP like gates here too and factorize as product of 1q plus a wire permutation

sbrandhsn: Sounds nice, is this just a check on the values of `local_invariants`? What are the corresponding values for swap gates?

sbrandhsn:
We can use `TwoQubitWeylDecomposition` instead - what do you think @mtreinish? As this pass will mostly run on two-qubit unitaries that can not be decomposed, I thought minimizing runtime overhead is crucial. We can also revisit later to see how we can insert the information necessary for `TwoQubitWeylDecomposition` into the circuit. I didn't intent to merge the code with these tolerances without consulting you - sorry for causing confusion.

The previous tolerances would fail on 
```
qc = QuantumCircuit(2)
qc.cp(pi*2**-(22), 0, 1)
``` 
with `qiskit.exceptions.QiskitError: 'decompose_two_qubit_product_gate: decomposition failed: deviation too large: 2.1049828546892968e-13'`

I see two ways of handling this:
- Setting the tolerance in `decompose_two_qubit_product_gate` to the same tolerance as in `two_qubit_local_invariants` (1-e13 would be fine).
- Checking for the identity in `Split2QUnitaries` with a corresponding tolerance before trying to decompose into single-qubit gates.

I like the idea of also detecting SWAP gates even though I would rather expect this behaviour in `ElidePermutations` and I would also expect the number of 'hidden' SWAP gates to be even lower than the number of decomposable two-qubit unitaries - what is your expectation?

mtreinish:
> We can use `TwoQubitWeylDecomposition` instead - what do you think @mtreinish? As this pass will mostly run on two-qubit unitaries that can not be decomposed, I thought minimizing runtime overhead is crucial. We can also revisit later to see how we can insert the information necessary for `TwoQubitWeylDecomposition` into the circuit. I didn't intent to merge the code with these tolerances without consulting you - sorry for causing confusion.

I'm fine using the `TwoQubitWeylDecomposition` especially if it lets us do a more robust analysis and gives us extra opportunities for optimizations then using that makes sense. While it will be a bit slower we already are calling it as part of the `TwoQubitBasisDecomposer` over all the 2q unitary blocks and it's quite fast so I doubt it'll be a measurable time difference. For example: https://github.com/Qiskit/qiskit/pull/12809#issuecomment-2247437326 was me timing the time to build a circuit using it (which we won't do here) and with that it was taking on average ~ 3e-5 seconds per unitary on main right now. So I'd say it's in fast enough territory if we want to use it.

If there is a performance issue in the future we can revisit it then, because we'll probably move the entire pass triplet into a single rust pass in the near future anyway.

test/python/transpiler/test_preset_passmanagers.py
```diff
@@ -279,7 +280,7 @@ def counting_callback_func(pass_, dag, time, property_set, count):
             callback=counting_callback_func,
             translation_method="synthesis",
         )
-        self.assertEqual(gates_in_basis_true_count + 1, collect_2q_blocks_count)
+        self.assertEqual(gates_in_basis_true_count + 2, collect_2q_blocks_count)
```
mtreinish, line 283: If anyone else is looking at this, this change is needed because with this PR applied we're calling `Collect2qBlocks` an extra time during the `init` stage now (it took me a sec to realize why this changed).

qiskit/transpiler/preset_passmanagers/builtin_plugins.py
```diff
@@ -160,6 +165,49 @@ def pass_manager(self, pass_manager_config, optimization_level=None) -> PassMana
...
+            do_consolidate_blocks_init = target is None and basis is None
+            do_consolidate_blocks_init |= target is not None and _is_one_op_non_discrete(
+                target.operations
+            )
+            do_consolidate_blocks_init |= basis is not None and _is_one_op_non_discrete(basis)
+
+            if do_consolidate_blocks_init:
+                init.append(Collect2qBlocks())
+                init.append(ConsolidateBlocks())
+                init.append(Split2QUnitaries())
```
mtreinish: Shouldn't we set fidelity based on the `approximation_degree` here?

mtreinish (commented):
Overall I think this is mostly ready to go now. I just have one small inline question about how we handle the fidelity argument in the preset pass manager.

levbishop:
The code looks fine, module plumbing in the approximation. I haven't tried checking the compile speed and/or quality of output. Would intuitively expect modest improvement in both most of the time, but there are probably pathological cases where it is worse. Looks like you are pushing the swap case for a follow up

qiskit/transpiler/preset_passmanagers/builtin_plugins.py
jakelishman: The logic here seems intended to avoid doing an unsafe code path if we know the basis to be unsafe (i.e. one of the basis gates is continuous in a way that we understand).

I think we might want to flip it a little bit, especially with this merging this close to the rc; rather than assuming we will do the pass, I think we should assume that we _won't_, and only run it if we can prove the basis is made entirely of standard gates we understand, of which at least one is a continuously parametric 1q.

I'm thinking of the case where a downstream user has a custom gate that they've added to the basis translator's equivalences and they want to transpile against.  We have downstream packages like that (e.g. the circuit-knitting toolbox and some of our researchers), but either way it's part of our interface.

mtreinish: I don't think there really is anything different being introduced by this and I viewed the if statement as unnecessary, see our earlier discussion thread here: https://github.com/Qiskit/qiskit/pull/12727#discussion_r1672182189 . Basically we either run `ConsolidateBlocks` here or we run it after routing/basis translation. From the preserving custom gates or non-standard gate workflows perspective nothing should be different by doing this. That being said I was fine leaving this in as extra insurance and given our proximity to the 1.2.0rc1 tag restricting it even further makes sense. We can loosen these requirements post 1.2.0.

jakelishman: Ah yeah, it's a fair point that we do it after routing again anyway.  Yeah, I guess my concern here applies equally to our pre-existing situation - basically, my concern is that we should collect up gate if we can't be sure we'll be able to synthesise ourselves back out of them.

I'm ok leaving it in as-is, given we already run `ConsolidateBlocks` at these levels.

qiskit/transpiler/preset_passmanagers/builtin_plugins.py
```diff
@@ -207,7 +214,7 @@ def _is_one_op_non_discrete(ops):
             if do_consolidate_blocks_init:
                 init.append(Collect2qBlocks())
                 init.append(ConsolidateBlocks())
-                init.append(Split2QUnitaries())
+                init.append(Split2QUnitaries(pass_manager_config.approximation_degree))
```
mtreinish: This needs to handle the case of `None` correctly. This would be something like:

```suggestion
                # If approximation degree is None that indicates a request to approximate up to the
                # error rates in the target. However, in the init stage we don't yet know the target
                # qubits being used to figure out the fidelity so just use the default fidelity parameter
                # in this case.
                if pass_manager_config.approximation_degree is not None:
                    init.append(Split2QUnitaries(pass_manager_config.approximation_degree))
                else:
                    init.append(Split2QUnitaries())
```

test/python/compiler/test_transpiler.py
```diff
@@ -872,6 +874,42 @@ def test_do_not_run_gatedirection_with_symmetric_cm(self):
...
+        qc.cx(0, 1)
+        qc.cx(1, 2)
+
+        orig_pass = Split2QUnitaries()
+        with patch.object(Split2QUnitaries, "run", wraps=orig_pass.run) as mock_pass:
+            basis = ["t", "sx", "cx"]
+            backend = GenericBackendV2(3, basis_gates=basis)
+            transpile(qc, backend=backend)
+            transpile(qc, basis_gates=basis)
+            transpile(qc, target=backend.target)
```
mtreinish, line 891: It's probably good to be explicit about the optimization level here, just to ensure we're using 2:

```suggestion
            transpile(qc, backend=backend, optimization_level=2)
            transpile(qc, basis_gates=basis, optimization_level=2)
            transpile(qc, target=backend.target, optimization_level=2)
```

test/python/transpiler/test_split_2q_unitaries.py
```diff
@@ -0,0 +1,190 @@
...
+
+    def test_2_1q(self):
+        """Test that a Kronechker product of two X gates is correctly processed."""
+        x_mat = np.array([[0, 1], [1, 0]])
+        multi_x = np.kron(x_mat, x_mat)
+        qr = QuantumRegister(2, "qr")
+        backend = BasicSimulator()
+        qc = QuantumCircuit(qr)
+        qc.unitary(multi_x, qr)
+        qct = transpile(qc, backend)
```
mtreinish:
```suggestion
        qct = transpile(qc, backend, optimization_level=2)
```

test/python/transpiler/test_split_2q_unitaries.py
```diff
@@ -0,0 +1,190 @@
...
+from qiskit.circuit.library import UnitaryGate
+from qiskit.quantum_info import Operator
+from qiskit.transpiler import PassManager
+from qiskit.quantum_info.operators.predicates import matrix_equal
+from qiskit.transpiler.passes import Collect2qBlocks, ConsolidateBlocks
+from qiskit.providers.basic_provider import BasicSimulator
+from qiskit.transpiler.passes.optimization.split_2q_unitaries import Split2QUnitaries
+
+
+class TestSplit2QUnitaries(QiskitTestCase):
```
mtreinish, line 31: It might be good to have some tests for the skip conditions in the pass, like a circuit with only 1q gates, or a parameterized 2q gate, etc. We get that coverage from the full path transpiler tests because this is in the preset pass manager, but having explicit tests will help us catch regressions in the future. But we can save this for a follow up.

qiskit/transpiler/passes/optimization/split_2q_unitaries.py
```diff
@@ -0,0 +1,76 @@
...
+
+from qiskit.transpiler.basepasses import TransformationPass
+from qiskit.circuit.quantumcircuitdata import CircuitInstruction
+from qiskit.dagcircuit.dagcircuit import DAGCircuit, DAGOpNode
+from qiskit.circuit.library.generalized_gates import UnitaryGate
+from qiskit.synthesis.two_qubit.two_qubit_decompose import TwoQubitWeylDecomposition
+
+
+class Split2QUnitaries(TransformationPass):
+    """Splits each two-qubit gate in the `dag` into two single-qubit gates, if possible without error."""
```
mtreinish: I feel like expanding the docs here to be a bit more explicit about what is going on in the pass would be good. But we can do this in a follow up for 1.2.0 final. If I were to read this I would be kind of confused by this docstring because I wouldn't understand how to split a 2q gate into two 1q gates without error.

```suggestion
    """Attempt to splits two-qubit gates in a :class:`.DAGCircuit` into two single-qubit gates
    
    This pass will analyze all the two qubit gates in the circuit and analyze the gate's unitary
    matrix to determine if the gate is actually a product of 2 single qubit gates. In these
    cases the 2q gate can be simplified into two single qubit gates and this pass will
    perform this optimization and will replace the two qubit gate with two single qubit
    :class:`.UnitaryGate`\s.
```

mtreinish (commented):
I think this is basically ready. I just left some small inline comments on the tests and docs but we can do these in follow ups pretty easily as long as we do it before the 1.2.0 final release.

qiskit/transpiler/preset_passmanagers/builtin_plugins.py
sbrandhsn: I made the checks a bit stricter, I think this makes sense for a new pass that will essentially run on all input circuits, we can relax in a follow-up.

sbrandhsn:
> The code looks fine, module plumbing in the approximation. I haven't tried checking the compile speed and/or quality of output. Would intuitively expect modest improvement in both most of the time, but there are probably pathological cases where it is worse. Looks like you are pushing the swap case for a follow up

Thanks, I decided to not include the swap gate case here because we will need to build a new dag for every input circuit. I assume most input circuits won't have swap-like circuits, so we would end up wasting that runtime while not yielding an improvement in circuit size for most cases. Also, when a user inputs a prerouted quantum circuit he would be very surprised to find that his routing solution has been replaced by something qiskit came up with which may or may not be better. I still think this is an interesting approach and should be reconsidered in the future.

qiskit/transpiler/preset_passmanagers/builtin_plugins.py
sbrandhsn: I wasn't aware it can be `None`, the default seems to be setting `approximate_degree=1`, thanks!

mtreinish (approved):
LGTM, thanks for the quick updates.

PR #10591, opened 2023-08-08
Keep resets at beginning of circuits

nonhermitian:
### Summary
Fixes #6943
Fixes #5921
Fix #5127

### Details and comments

Disables the removal of `reset` instructions if performed at the beginning of a circuit.  Previous behavior assumed that the input state of a circuit is zero, as is the case when executed on most (perhaps all) quantum hardware to date.  However, the circuit object is abstract and exists without this context.  Moreover, if one wanted to explicitly set the initial state to be the zeros state by adding resets, then the preset passes would remove this, invalidating the intention of the user.  

This PR disables the pass that does this in the preset passmanagers.  The tests that looked for removed resets are turned into tests that verify the resets are still there.

Issue #5127, opened 2020-09-24
Reset optimization in the transpiler


zachschoenfeld33:
### What is the expected enhancement?

Modification of the behavior of the `reset` instruction in the transpiler.

There is a standard transpiler pass `RemoveResetInZeroState`, which understandably deletes a `reset` when the qubit is already in the ground state. According to documentation, this should be applied for `optimization_level >= 1`.

However, it does not always act as expected. Consider
```
qc = QuantumCircuit(1, 1)
qc.x(0)
qc.x(0)
qc.reset(0)

print(qc)
print(transpile(qc, optimization_level=0))
print(transpile(qc, optimization_level=1))
print(transpile(qc, optimization_level=2))
print(transpile(qc, optimization_level=3))
```

This gives the output
```
     ┌───┐┌───┐     
q_0: ┤ X ├┤ X ├─|0>─
     └───┘└───┘     
c: 1/═══════════════
                    
     ┌───┐┌───┐     
q_0: ┤ X ├┤ X ├─|0>─
     └───┘└───┘     
c: 1/═══════════════
                    
     ┌───┐┌───┐     
q_0: ┤ X ├┤ X ├─|0>─
     └───┘└───┘     
c: 1/═══════════════
                    
          
q_0: ─|0>─
          
c: 1/═════
          
     
q_0: 
     
c: 1/
```
We would expect that level 1 removes the two `x` gates in line w/ the `RemoveResetInZeroState`, but it does not.

Furthermore, when using a backend, interchanging of `reset` instructions for `swap` gates can be unpredictable.

Consider the following circuits which use a single `reset` on 3 different qubits. This is run on the public 5q device `ibmq_ourense`
```
IBMQ.load_account()
provider = IBMQ.get_provider(hub='ibm-q', group='open', project='main')
backend = provider.get_backend('ibmq_ourense')

def multiq_custom_circ(n_resets):
    """Multi qubit circuit on qubits 0, 1, 2 w/ n_resets reset instructions used."""
    qc = QuantumCircuit(3,3)
    qc.h(0)
    qc.h(1)
    qc.h(2)
    for _ in range(n_resets):
        qc.reset(0)
    qc.cx(1,0)
    for _ in range(n_resets):
        qc.reset(2)
    qc.cx(0,2)
    for _ in range(n_resets):
        qc.reset(1)
    qc.x(1)
    qc.measure([0,1,2], [0,1,2])
    return qc

qc = multiq_custom_circ(1)
[...]

PR #12729, opened 2024-07-07
Adding ``is_zero_initialized`` to transpile and extending HighLevelSynthesis with ancilla qubits

alexanderivrii:
### Updated Summary

This PR adds a new argument ``is_zero_initialized`` to ``transpile(...)``, as well as to ``PassManagerConfig`` and ``generate_pass_manager``. If set to ``True`` (the default) the qubits are assumed to be initially in state ``<0|``. For some additional context, see #10591, #5127 and other linked issues. After an offline discussion, we have concluded that having an explicit argument in ``transpile`` achieves the best of both worlds.

In addition, this PR passes this flag to ``HighLevelSynthesis`` transpiler pass. If set to ``True``, the pass can initially label all qubits as "clean auxiliary qubits". The pass then keeps track of clean and dirty auxiliary qubits throughout the run, and passes this information to plugins via kwargs ``_num_clean_ancillas`` and ``_num_dirty_ancillas``. Additionally, the pass has been extended to handle the case that the circuit synthesizing a high-level-object is defined over these additional qubits.

I have removed the MCX synthesis methods and MCX synthesis plugins from this PR. We will handle a more organized restructuring of the multi-controlled synthesis algorithms and plugins in a follow-up.

### Original Summary (outdated)

For now this is a proof-of-concept PR based on internal Qiskit discussions. Everything is up to debate.

The idea is to extend the `HighLevelSynthesis` transpiler pass with the ability to use "auxiliary qubits" when synthesizing high-level operations using plugins. In this initial implementation, the allowed synthesis methods should support "dirty" auxiliary qubits, that is the synthesis methods are required to bring each auxiliary qubit into its original state. This allows to use any qubit that is not a part of the object's definition as an auxiliary qubit. It should be also possible to support "clean" auxiliary qubits (or a combination of clean and dirty auxiliary qubits). 

This is illustrated for two MCX synthesis functions: the default synthesis function (based on the reduction to MCPhase gate) that does not require any auxiliary qubits, and the "recursive synthesis" function (extracted from the definition of the `MCXRecursive` gate) that requires a single auxiliary qubit for MCX gates with more than 4 control qubits. The key difference between the `MCXRecursiveGate` and the new functionality is that for `MCXRecursiveGate` the number and the auxiliary qubits used are baked into the gate's definition, while with the new flow the gate is an `MCXGate` only defined on its control and target qubits, and it's the actual synthesis that decides how many auxiliary qubits are available.

This PR does not deprecate or remove `MCXRecursiveGate` (and other custom MCX variants). This can be done in some follow-up PR if we agree that this is the right strategy. Also note that since the `MCXRecursiveGate` has a name "mcx_recursive" (or something like that) the two new "mcx" plugins do not apply, and the old synthesis behavior based on the gate's definition is used. 

More comments appear as notes within the code. 

### Details and comments

Here is a concrete example:

```
qc = QuantumCircuit(10)
# This is a bona fide MCX gate, not its recursive version
qc.mcx(control_qubits=[0, 1, 2, 3, 4, 5, 6], target_qubit=[7])

# default synthesis plugin (from this PR) that uses the current default synthesis algorithm
qct1 = transpile(qc, basis_gates=['cx', 'u'])
print(qct1.count_ops())

# mcx synthesis plugin (from this PR) that uses the recursive synthesis algorithm with 1 extra ancilla
hls_config = HLSConfig(mcx=["recursive"])
qct2 = transpile(qc, basis_gates=['cx', 'u'], hls_config=hls_config)
print(qct2.count_ops())
```
The output is:
```
OrderedDict([('u', 248), ('cx', 220)])
OrderedDict([('u', 159), ('cx', 144)])
```
---

Feedback is welcome!


alexanderivrii:
Heads up: the number of 2-qubit gates for the synthesized circuit on the benchpress `multi_control_circuit` benchmark (that contains multiple MCX gates of different sizes):
* up to 10 qubits: tket - 937, qiskit - 1269, qiskit* - 985
* up to 15 qubit: tket - 3617, qiskit - 5849, qiskit* - 3825

Here `qiskit*` corresponds to running transpile with this PR and with the following config:
```
hls_config = HLSConfig(mcx=["recursive", "default"], plugin_selection="all", plugin_evaluation_fn=size2q)
transpile(qiskit_qc, basis_gates=["cx", "u"], hls_config=hls_config)
```

Now, let's add an extra free qubit to the circuit (i.e. the circuit is now defined over `n+1` qubits with multiple MCX gates up to `n` qubits). The number of gates for tket and for qiskit do not change, but qiskit* uses the extra qubit.
* up to 10 qubits: tket - 937, qiskit - 1269, qiskit* - 765
* up to 15 qubit: tket - 3617, qiskit - 5849, qiskit* - 3197

qiskit/transpiler/passes/synthesis/high_level_synthesis.py
```diff
@@ -410,43 +414,122 @@ def run(self, dag: DAGCircuit) -> DAGCircuit:
             TranspilerError: when the transpiler is unable to synthesize the given DAG
             (for instance, when the specified synthesis method is not available).
         """
+        return self._run_inner(dag, self.is_zero_initialized)
 
-        # copy dag_op_nodes because we are modifying the DAG below
-        dag_op_nodes = dag.op_nodes()
+    def _run_inner(self, dag: DAGCircuit, is_zero_initialized=False) -> DAGCircuit:
```
Cryoris:
```suggestion
    def _run_inner(self, dag: DAGCircuit, is_zero_initialized: bool = False) -> DAGCircuit:
```

qiskit/transpiler/passes/synthesis/high_level_synthesis.py
```diff
@@ -410,43 +414,122 @@ def run(self, dag: DAGCircuit) -> DAGCircuit:
...
+
+                else:
+                    if isinstance(decomposition, QuantumCircuit):
+                        decomposition = circuit_to_dag(decomposition, copy_operations=False)
+
+                    if isinstance(decomposition, DAGCircuit):
+                        # Compute the qubits on which to apply the decomposition
+                        # (including auxiliary qubits).
+                        num_ancillas_used = decomposition.num_qubits() - len(node.qargs)
+                        if not use_ancillas or num_ancillas_used == 0:
```
Cryoris: If `use_ancillas` is `False`, then `num_ancillas_used` must be 0 because we didn't give the synthesis methods any ancilla qubits to use, so this is redundant (unless something really goes wrong in the synthesis, in which case it breaks later either way)
```suggestion
                        if num_ancillas_used == 0:
```

qiskit/transpiler/passes/synthesis/high_level_synthesis.py
```diff
@@ -410,43 +414,122 @@ def run(self, dag: DAGCircuit) -> DAGCircuit:
...
+                            extended_qargs = node.qargs
+                        elif num_ancillas_used <= len(clean_ancillas_available):
```
Cryoris: It might be good to state this behavior somewhere: that clean ancillas, if available, are preferably used over dirty ones.

qiskit/transpiler/passes/synthesis/high_level_synthesis.py
```diff
@@ -410,43 +414,122 @@ def run(self, dag: DAGCircuit) -> DAGCircuit:
...
+                        # CCX-gate). There should be no case when this Operation would have more qubits
+                        # than the original operation.
+                        if decomposition.num_qubits != len(node.qargs):
+                            raise TranspilerError(
+                                "HighLevelSynthesis reached an unexpected result."
+                            )
+                        new_dag.apply_operation_back(decomposition, node.qargs, node.cargs)
+
+            # update the list of clean ancilla qubits
+            clean_ancillas = [q for q in clean_ancillas if q not in node.qargs]
```
Cryoris: We should add re-using clean ancillas in the future. One option would be to let the synthesis method report what it does with ancilla qubits, e.g. how many of which kind it requires, and how they are returned.

releasenotes/notes/hls-with-ancillas-d6792b41dfcf4aac.yaml
```diff
@@ -0,0 +1,18 @@
...
+    A new argument ``is_zero_initialized`` has been added to :func:`qiskit.compiler.transpile`,
+    :func:`.generate_preset_pass_manager`, and to :class:`~.PassManagerConfig`.
+    If set to ``True``, the qubits are assumed to be initially in the state :math:`|0\rangle`,
+    potentially allowing additional optimization opportunities for individual transpiler passes.
+  - |
+    The constructor for :class:`.HighLevelSynthesis` transpiler pass now accepts an
+    additional argument ``is_zero_initialized``. If set to ``True``, the pass assumes that the
+    qubits are initially in the state :math:`|0\rangle`. In addition, the pass keeps track of
+    clean and dirty auxiliary qubits throughout the run, and passes this information to plugins
+    via kwargs ``_num_clean_ancillas`` and ``_num_dirty_ancillas``.
```
Cryoris: Is there a reason to have these with leading underscores? It seems that no other arguments are with leading underscores and the number of ancillas seem like normal arguments to me 🙂

releasenotes/notes/hls-with-ancillas-d6792b41dfcf4aac.yaml
```diff
@@ -0,0 +1,18 @@
...
+upgrade_transpiler:
+  - |
+    The :func:`qiskit.compiler.transpile` now assumes that the qubits are initially in the state
```
Cryoris:
```suggestion
    The :func:`~qiskit.compiler.transpile` now assumes that the qubits are initially in the state
```

test/python/transpiler/test_high_level_synthesis.py
```diff
@@ -481,7 +481,7 @@ def test_target_gets_passed_to_plugins(self):
                 [
                     HighLevelSynthesis(
                         hls_config=hls_config,
-                        target=GenericBackendV2(num_qubits=5, basis_gates=["u", "cx"]).target,
+                        target=GenericBackendV2(num_qubits=5, basis_gates=["u", "cx", "id"]).target,
```
Cryoris, line 511: Just curious: why is this change needed?

Cryoris (commented):
tl;dr: I would suggest we only add `is_zero_initialized` (and add `RemoveResetInZeroState` in this case) and only add the HLS logic once we have plugins that can use it.

-----

Overall this looks like a solid start for handling ancillas. But I have one question upfront: if our main goal is to introduce the new argument `is_zero_initialized` for 1.2, wouldn't it make more sense to defer the HLS changes to when we actually add the new plugins that can use ancillas? That way we can also add tests and ensure the HLS logic works as expected, and we don't add "dead code" that is in the release but never used.

Another important point for is being consistent with the assumption that the initial state is $|0\rangle$ throughout the compiling process. That would mean running the `RemoveResetInZeroState` pass if `is_zero_initialized is True` -- otherwise it seems a bit unexpected that we state the qubits are already reset but don't remove initial resets.

I can open a PR with these changes to your branch, and if you agree, I'd be happy to move forward with the reduced changes for 1.2.

test/python/transpiler/test_high_level_synthesis.py
alexanderivrii: The `id` gate neither has a definition nor is present in the equivalence library, so HLS complains about it. This manifested only right now since HLS now recursively processes the circuit obtained by synthesizing a high-level-object (allowing a synthesis method to replace a high-level-object by other high-level-objects, somewhat similar to unrolling nested definitions) + one of the mock plugins replaces a custom gate by `id` gates.

qiskit/transpiler/passes/synthesis/high_level_synthesis.py
alexanderivrii: I agree! We can probably use circuit's `metadata` for this (and any additional information that the synthesis method needs to return).

releasenotes/notes/hls-with-ancillas-d6792b41dfcf4aac.yaml
alexanderivrii: I was hesitating about this and changed back and forth several times myself. The original reason is to differentiate between user-specified plugin options and options set by HighLevelSynthesis pass itself. Right now I agree that the version without the underscores is better.

ElePT:
Given that this PR is in ongoing discussion, I think it will likely not make it for 1.2, and be for 1.3.

alexanderivrii:
Superseded by #12911.

alexanderivrii:
This can be closed as the better PR #12911 has been merged!

PR #12734, opened 2024-07-08
Extending control flow op names

sbrandhsn:
### Summary

I think `CONTROL_FLOW_OP_NAMES` should contain all control flow operations currently available to qiskit by default. :-)

### DIFF
--- a/qiskit/circuit/controlflow/__init__.py
+++ b/qiskit/circuit/controlflow/__init__.py
@@ -24,4 +24,5 @@ from .for_loop import ForLoopOp
 from .switch_case import SwitchCaseOp, CASE_DEFAULT
 
 
-CONTROL_FLOW_OP_NAMES = frozenset(("for_loop", "while_loop", "if_else", "switch_case"))
+CONTROL_FLOW_OP_NAMES = frozenset(("for_loop", "while_loop", "if_else", "switch_case", "continue_loop",
+                                   "break_loop"))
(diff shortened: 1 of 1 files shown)


qiskit/circuit/controlflow/__init__.py
```diff
@@ -24,4 +24,6 @@
 from .switch_case import SwitchCaseOp, CASE_DEFAULT
 
 
-CONTROL_FLOW_OP_NAMES = frozenset(("for_loop", "while_loop", "if_else", "switch_case"))
+CONTROL_FLOW_OP_NAMES = frozenset(
+    ("for_loop", "while_loop", "if_else", "switch_case", "continue_loop", "break_loop")
+)
```
jakelishman, line 29: Fwiw, `ContinueLoopOp` and `BreakLoopOp` aren't subclasses of `ControlFlowOp`, which is what this was for checking, and they can't exist other than as the child of one of the others.  `ControlFlowOp` is about defining a small interface for instructions that contain other `QuantumCircuit`s as basic blocks.  These do impact the flow of control, but from the `QuantumCircuit` tree representation, that's often not actually particularly important (though I guess we wouldn't know - nothing supports them).

This probably doesn't hurt to add, but if anything's using this as a check and _not_ doing `isinstance(op, ControlFlowOp)` in addition, then it might cause breakage.

PR #12739, opened 2024-07-08
Oxidize two qubit local invariance functions

mtreinish:
### Summary

This commit migrates the two functions in the private module `qiskit.synthesis.two_qubit.local_invariance` to primarily be implemented in rust. Since the two_qubit_local_invariants() function is being used in #12727 premptively porting these functions to rust will both potentially speed up the new transpiler pass in that PR and also facilitate a future porting of that pass to rust. The two python space functions continue to exist as a small wrapper to do input type checking/conversion and rounding of the result (since the python API for rounding is simpler). There is no release note included for these functions as they are internal utilities in Qiskit and not exposed as a public interface.

### Details and comments


crates/accelerate/src/two_qubit_decompose.rs
```diff
@@ -1883,9 +1883,76 @@ impl TwoQubitBasisDecomposer {
...
+    // Get determinate since +- one is allowed.
+    let det_bell_basis = bell_basis_unitary
+        .view()
+        .into_faer_complex()
+        .determinant()
+        .to_num_complex();
+    let m = bell_basis_unitary.t().dot(&bell_basis_unitary);
+    let mut m_tr2 = m.diag().sum();
+    m_tr2 *= m_tr2;
+    // Table II of Ref. 1 or Eq. 28 of Ref. 2.
```
jakelishman, line 1961: This comment doesn't make sense in the absence of the docstring.

crates/accelerate/src/two_qubit_decompose.rs
```diff
@@ -1883,9 +1883,76 @@ impl TwoQubitBasisDecomposer {
...
+        C_ZERO,
+        c64(0., -FRAC_1_SQRT_2),
+    ],
+];
+
+#[pyfunction]
+pub fn two_qubit_local_invariants(unitary: PyReadonlyArray2<Complex64>) -> [f64; 3] {
+    let mat = unitary.as_array();
+    // Transform to bell basis
+    let bell_basis_unitary = transpose_conjugate(aview2(&MAGIC)).dot(&mat.dot(&aview2(&MAGIC)));
```
jakelishman: There's possibly a speedup and accuracy improvement to be had by storing the adjoint statically, and doing the asymmetric non-normalised trick we do in the other Weyl code.

crates/accelerate/src/two_qubit_decompose.rs
```diff
@@ -1883,9 +1883,76 @@ impl TwoQubitBasisDecomposer {
...
+    // Here we split the real and imag pieces of G1 into two so as
+    // to better equate to the Weyl chamber coordinates (c0,c1,c2)
+    // and explore the parameter space.
+    // Also do a FP trick -0.0 + 0.0 = 0.0
+    [g1.re + 0., g1.im + 0., g2.re + 0.]
+}
+
+#[pyfunction]
+pub fn local_equivalence(weyl: PyReadonlyArray1<f64>) -> PyResult<[f64; 3]> {
+    let weyl = weyl.as_slice()?;
```
jakelishman: This can fail if the input array isn't contiguous, I think.

mtreinish: Yeah, that's why it returns a `PyResult`. As it's internal I didn't think it mattered, although I debated moving to just passing in 3 floats in a `[f64; 3]` directly because there are only 3 values. I can make that change if you think it's clearer.

crates/accelerate/src/two_qubit_decompose.rs
mtreinish: I'll add the comment about the original paper to a docstring on the rust function too

crates/accelerate/src/two_qubit_decompose.rs
jakelishman: Sorry, I guess I didn't make the thought clear: this returns a Python exception, except it's needless because we don't _need_ this to be contiguous.  Even if we keep passing an array, we can still do what we need with strided access, right?

mtreinish: Ah ok, yeah I can use an `ArrayView1` here instead too. Using arrays here just felt like overkill here for something that is just 3 floats

crates/accelerate/src/two_qubit_decompose.rs
mtreinish: Done in: https://github.com/Qiskit/qiskit/pull/12739/commits/cf6c5a4f254a49b813e15432cb28f555d6922c85

crates/accelerate/src/two_qubit_decompose.rs
mtreinish: Added the static in: https://github.com/Qiskit/qiskit/pull/12739/commits/c4e369aad539e80b2264d9cabcf06fafcbd771fd

crates/accelerate/src/two_qubit_decompose.rs
mtreinish: Used `as_array()` here to keep everything as an arrayview1: https://github.com/Qiskit/qiskit/pull/12739/commits/21279873e74df243776b390de8bc548594576607

crates/accelerate/src/two_qubit_decompose.rs
```diff
@@ -1883,9 +1883,110 @@ impl TwoQubitBasisDecomposer {
...
+pub fn local_equivalence(weyl: PyReadonlyArray1<f64>) -> PyResult<[f64; 3]> {
+    let weyl = weyl.as_array();
+    let weyl_2_cos_squared_product: f64 = weyl.iter().map(|x| (x * 2.).cos().powi(2)).product();
+    let weyl_2_sin_squared_product: f64 = weyl.iter().map(|x| (x * 2.).sin().powi(2)).product();
+    let g0_equiv = weyl_2_cos_squared_product - weyl_2_sin_squared_product;
+    let g1_equiv = weyl.iter().map(|x| (x * 4.).sin()).product::<f64>() / 4.;
+    let g2_equiv = 4. * weyl_2_cos_squared_product
+        - 4. * weyl_2_sin_squared_product
+        - weyl.iter().map(|x| (4. * x).cos()).product::<f64>();
+    Ok([g0_equiv + 0., g1_equiv + 0., g2_equiv + 0.])
```
jlapeyre, line 1982: This compiles with no clippy complaints for me.
```suggestion
    Ok([g0_equiv, g1_equiv, g2_equiv])
```
~Was this a holdover from Python that coerced the values to floats?~ No, see  below

mtreinish: It's actually a trick to convert a floating point negative zero: `-0.` to a `0.` without any branching. This was done in the python code so I figured retain the behavior of only returning `0.`

jlapeyre (approved):
LGTM

PR #10296, opened 2023-06-15
Use singleton matrices for unparametrised standard gates

jakelishman:
### Summary

This makes the array form of standard gates with zero parameters singleton class attributes that reject modification. The class-level `__array__` methods are updated to return exactly the same instance, except in very unusual circumstances, which means that `Gate.to_matrix()` and `numpy.asarray()` calls on the objects will return the same instance. This avoids a decent amount of construction time, and avoids several Python-space list allocations and array allocations.

The dtypes of the static arrays are all standardised to be complex128. Gate matrices are in general unitary, `Gate.to_matrix()` already enforces a cast to `complex128`.  For gates that allowed their dtypes to be inferred, there were several cases where native ints and floats would be used, meaning that `Gate.to_matrix()` would also involve an extra matrix allocation to hold the cast, which just wasted time.

For standard controlled gates, we store both the closed- and open-controlled matrices singly controlled gates.  For gates with more than one control, we only store the "all ones" controlled case, as a memory/speed trade-off; open controls are much less common than closed controls.

For the most part this won't have an effect on peak memory usage, since all the allocated matrices in standard Qiskit usage would be freed by the garbage collector almost immediately.  This will, however, reduce construction costs and garbage-collector pressure, since fewer allocations+frees will occur, and no calculations will need to be done.

### Details and comments

I ran some bits of the benchmark suite, not the full thing.  One particularly notable speedup was in `CommutationAnalysis`, where I'm seeing ~15% improvements in the benchmarks:
```
       before           after         ratio
     [81964e64]       [339f4dca]
     <singleton-matrices~1>       <singleton-matrices>
-      1.86±0.02s       1.65±0.01s     0.89  passes.PassBenchmarks.time_commutation_analysis(20, 1024)
-        563±10ms         488±20ms     0.87  passes.PassBenchmarks.time_commutation_analysis(5, 1024)
-       1.33±0.1s       1.12±0.01s     0.84  passes.PassBenchmarks.time_commutation_analysis(14, 1024)
```

I'd expect to see some improvements bits of unitary synthesis and the 1q and 2q conversions into unitaries too for real workloads, though I didn't immediately see them in my quick runs - the benchmark suite might not use too many of the gates that see the speedup, though, or I might have missed those benchmarks.

PR #12742, opened 2024-07-09
Fix `C3SXGate` `to_matrix` method

ElePT:
### Summary
This PR adds the `with_controlled_gate_array` decorator missing in `C3SXGate` that enables the `to_matrix` method. This bug was found while porting `C3SXGate` to Rust (https://github.com/Qiskit/qiskit/pull/12659).

### Details and comments
I have also noticed that the path of `C3SXGate` is incorrect. It's currently in `qiskit/circuit/library/standard_gates/x.py` while it should be in `qiskit/circuit/library/standard_gates/sx.py`. I did not address this issue in the current PR because I believe the move to the correct location is a breaking change and should probably be left for 2.0 (although most users won't import gates directly from the file).

### DIFF
--- a/qiskit/circuit/library/standard_gates/x.py
+++ b/qiskit/circuit/library/standard_gates/x.py
@@ -22,6 +22,7 @@ from qiskit.circuit._utils import _ctrl_state_to_int, with_gate_array, with_cont
 from qiskit._accelerate.circuit import StandardGate
 
 _X_ARRAY = [[0, 1], [1, 0]]
+_SX_ARRAY = [[0.5 + 0.5j, 0.5 - 0.5j], [0.5 - 0.5j, 0.5 + 0.5j]]
 
 
 @with_gate_array(_X_ARRAY)
@@ -571,6 +572,7 @@ class RCCXGate(SingletonGate):
         return isinstance(other, RCCXGate)
 
 
+@with_controlled_gate_array(_SX_ARRAY, num_ctrl_qubits=3, cached_states=(7,))
 class C3SXGate(SingletonControlledGate):
     """The 3-qubit controlled sqrt-X gate.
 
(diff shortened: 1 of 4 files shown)


qiskit/circuit/library/standard_gates/x.py
```diff
@@ -591,7 +593,7 @@ def __init__(
         unit="dt",
         _base_label=None,
     ):
-        """Create a new 3-qubit controlled sqrt-X gate.
+        """Create a new 3-qubit contƒCrolled sqrt-X gate.
```
jakelishman: wtf

releasenotes/notes/fix-c3sx-gate-matrix-050cf9f9ac3b2b82.yaml
```diff
@@ -0,0 +1,5 @@
+---
+fixes:
+  - |
+    Fixed a missing decorator in :class:`.C3SXGate` that made it fail if `gate.to_matrix()` was called.
```
jakelishman: You can link `` :meth:`.Gate.to_matrix` `` to get the proper cross-ref.

jakelishman (commented):
Thanks for the catch.  Looking (very briefly) at #10296, it seems like the array was always missing, even before the decorator.

releasenotes/notes/fix-c3sx-gate-matrix-050cf9f9ac3b2b82.yaml
ElePT:
```suggestion
    Fixed a missing decorator in :class:`.C3SXGate` that made it fail if :meth:`.Gate.to_matrix` was called.
```

qiskit/circuit/library/standard_gates/x.py
ElePT: when I switch keyboards option and command are reversed and stuff happens XD

jakelishman (approved):

jakelishman:

PR #12746, opened 2024-07-09
Port `synth_permutation_depth_lnn_kms` to Rust

### LINKED ISSUE #12224: Port `synth_permutation_depth_lnn_kms` to Rust

Procatv:
### Summary
Porting synth_permutation_depth_lnn_kms to Rust

### Details and comments
Fixes #12224


crates/accelerate/src/synthesis/permutation/mod.rs
```diff
@@ -82,11 +82,38 @@ fn _synth_permutation_acg(py: Python, pattern: PyArrayLike1<i64>) -> PyResult<Ci
...
+/// architecture using the Kutin, Moulton, Smithline method.
+#[pyfunction]
+#[pyo3(signature = (pattern))]
+pub fn _synth_permutation_depth_lnn_kms(py: Python, pattern: PyArrayLike1<i64>) -> PyResult<CircuitData> {
+    let inverted = utils::invert(&pattern.as_array());
+    let view = inverted.view();
+    let num_qubits = view.len();
+
+    let swap_layer: Vec<(usize, usize)> = utils::_create_swap_layer(&view);
+
```
alexanderivrii: The original Python code called this function multiple times with alternating odd/even layers:
```
for i in range(num_qubits):
    _create_swap_layer(qc, cur_pattern, i % 2)
```
And here it seems that you are only calling the function once.

crates/accelerate/src/synthesis/permutation/utils.rs
```diff
@@ -144,3 +144,20 @@ pub fn decompose_cycles(cycles: &Vec<Vec<usize>>) -> Vec<(usize, usize)> {
 
     swaps
 }
+
+// create swaps
+//Need to fix the below implementation
+pub fn _create_swap_layer(pattern: &ArrayView1<usize>) -> Vec<(usize, usize)> {
+    // let mut permutation: Vec<usize> = pattern.iter().map(|&x| x as usize).collect();
```
alexanderivrii: We need to call this function both for constructing even layers (pairing qubits starting from `0`) and odd layers (pairing qubits starting with `1`), so you probably need to pass the `starting_point` (either `0` or `1`) as the argument.

releasenotes/notes/synth_permutation_depth_lnn_kms-c444f3a363f3a903.yaml
```diff
@@ -0,0 +1,4 @@
+---
+upgrade_synthesis:
+  - |
```
alexanderivrii: This should be labeled as a feature note, please see https://github.com/Qiskit/qiskit/pull/12715.

test/python/synthesis/test_permutation_synthesis.py
```diff
@@ -24,6 +24,7 @@
     synth_permutation_depth_lnn_kms,
     synth_permutation_basic,
     synth_permutation_reverse_lnn_kms,
+    synth_permutation_depth_lnn_kms,
```
alexanderivrii: I think `synth_permutation_depth_lnn_kms` now appears twice in this import list.

releasenotes/notes/synth_permutation_depth_lnn_kms-c444f3a363f3a903.yaml
```diff
@@ -0,0 +1,4 @@
...
+    Port :func:`.synth_permutation_depth_lnn_kms`, used to synthesize qubit permutations, to Rust.
```
alexanderivrii: maybe expand this a bit to: "used to synthesize permutations _for linear connectivity_, to Rust."

alexanderivrii (changes requested):
Many thanks for doing this! The code already looks great, but I believe you have somewhat over-simplified the construction (and this is why some the correctness tests fail).

crates/accelerate/src/synthesis/permutation/utils.rs
```diff
@@ -144,3 +144,18 @@ pub fn decompose_cycles(cycles: &Vec<Vec<usize>>) -> Vec<(usize, usize)> {
...
+    let mut pattern_vec = pattern.to_vec();
+    let num_qubits = pattern_vec.len();
+    let mut gates = Vec::new();
+
+    for j in (starting_point..num_qubits).step_by(2) {
+        if pattern_vec[j] > pattern_vec[(j + 1) % num_qubits] {
+            gates.push((j, (j + 1) % num_qubits));
+            pattern_vec.swap(j, (j + 1) % num_qubits);
+        }
+    }
```
raynelfss: For this I believe that the higher bound should be `num_qubits-1` instead of `num_qubits` as per the original function. You told me you're doing the modulo `num_qubits` so it doesn't go out of bounds, but I think this should pass.
```suggestion
    for j in (starting_point..num_qubits - 1).step_by(2) {
        if pattern_vec[j] > pattern_vec[(j + 1)] {
            gates.push((j, (j + 1)));
            pattern_vec.swap(j, (j + 1));
        }
    }
```

raynelfss: You might want to use `&ArrayViewMut1<_>` here, the changes you're making to this pattern aren't being applied here.

Once you do so you shouldn't need to convert your `ArrayView` into a `vec` and the changes you're making to it should be applied in place. The way this is being done now does not save any of the swaps you're performing on `pattern`.

raynelfss (changes requested):
This is a good start, thank you for working on this! I just have a comment about the `create_swap_layer` function here.

crates/accelerate/src/synthesis/permutation/mod.rs
```diff
@@ -82,11 +82,44 @@ fn _synth_permutation_acg(py: Python, pattern: PyArrayLike1<i64>) -> PyResult<Ci
     )
 }
 
+/// Synthesize a permutation circuit for a linear nearest-neighbor
+///     architecture using the Kutin, Moulton, Smithline method.
+#[pyfunction]
+#[pyo3(signature = (pattern))]
+pub fn _synth_permutation_depth_lnn_kms(
+    py: Python,
+    pattern: PyArrayLike1<i64>,
```
alexanderivrii, line 91: In most of our synthesis code, we are using `PyReadonly1` instead of `PyArrayLike1`, however the current form is consistent with other methods in `synthesis/permutation/mod.rs`. So I am suggesting to leave this as is, but possibly refactor this file in a follow-up.

crates/accelerate/src/synthesis/permutation/utils.rs
```diff
@@ -144,3 +145,22 @@ pub fn decompose_cycles(cycles: &Vec<Vec<usize>>) -> Vec<(usize, usize)> {
 
     swaps
 }
+
+/// Implements a single swap layer, consisting of conditional swaps between each
+/// neighboring couple. The starting_point is the first qubit to use (either 0 or 1
+///     for even or odd layers respectively). Mutates the permutation pattern ``pattern``.
```
alexanderivrii: Do you need the extra spaces on line 151?

crates/accelerate/src/synthesis/permutation/utils.rs
```diff
@@ -144,3 +145,22 @@ pub fn decompose_cycles(cycles: &Vec<Vec<usize>>) -> Vec<(usize, usize)> {
...
+    pattern: &mut ArrayViewMut1<usize>,
+    starting_point: usize,
+) -> Vec<(usize, usize)> {
+    let num_qubits = pattern.len();
+    let mut gates = Vec::new();
+
+    for j in (starting_point..num_qubits - 1).step_by(2) {
+        if pattern[j] > pattern[j + 1] {
+            gates.push((j, (j + 1)));
+            pattern.swap(j, j + 1);
```
alexanderivrii: it's probably slightly clearer to remove the extra parenthesis around `j+1`, i.e. write `gates.push((j, j+1));`.

crates/accelerate/src/synthesis/permutation/mod.rs
```diff
@@ -82,11 +82,44 @@ fn _synth_permutation_acg(py: Python, pattern: PyArrayLike1<i64>) -> PyResult<Ci
     )
 }
 
+/// Synthesize a permutation circuit for a linear nearest-neighbor
+///     architecture using the Kutin, Moulton, Smithline method.
+#[pyfunction]
```
alexanderivrii: You probably do not need extra spaces on line 86.

alexanderivrii (commented):
This looks good to me, I only had some really-really minor nitpicks. This PR was almost ready for the past week, but for some reason was not tagged for 1.2, but I think we should include it.

crates/accelerate/src/synthesis/permutation/utils.rs
Procatv: Without the parenthesis I get an error that it says that 'expected 1 argument and found 2'

alexanderivrii (approved):
LGTM! Thank you!

alexanderivrii:
This is good to go on my part. The system does not allow me to merge it before @raynelfss approves it as well (since there are requested changes, which I believe have all been addressed).

ShellyGarion (approved):

raynelfss (approved):
LGTM! Thank you for working on this!

PR #12750, opened 2024-07-10
Add clifford gates to collect_cliffords

ShellyGarion:
### Summary

This PR adds some Clifford gates to the `CollectClifford()` transpiler pass.

### Details and comments

This is an immediate bugfix which does not handle the following problems:

- When adding a gate to the Clifford class, it should be added in other places of the code, 
and there is no single list of gates that the Clifford class supports.
https://github.com/Qiskit/qiskit/blob/6b57b7832feaf5334b0b8ec671943c0b7a81ec50/qiskit/transpiler/passes/optimization/collect_cliffords.py#L27
and
https://github.com/Qiskit/qiskit/blob/fa3d6df04f57d6a5bcc658a7f1a6e5a99984b949/qiskit/circuit/random/utils.py#L290

- The current Clifford class can handle certain `u` gates (for certain angles), and this does not appear in the docs, neither in the code above:
https://github.com/Qiskit/qiskit/blob/fa3d6df04f57d6a5bcc658a7f1a6e5a99984b949/qiskit/quantum_info/operators/symplectic/clifford_circuits.py#L100


alexanderivrii:
Thanks @ShellyGarion, this looks good! If we want a more general approach that supports whatever is supported by `clifford_circuits`, what do you think about changing the `filter_fn` to the following:

```
def _is_clifford_gate(node):
    """Specifies whether a node holds a clifford gate."""
    if getattr(node.op, "condition", None) is not None:
        return False
    is_clifford_gate = True
    try:
        Clifford(node.op)
    except QiskitError:
        is_clifford_gate = False
    return is_clifford_gate
```
or
```
def _is_clifford_gate(node):
    """Specifies whether a node holds a clifford gate."""
    if getattr(node.op, "condition", None) is not None:
        return False
    if node.op.name in clifford_gate_names:
        return True
    # test other gates
    is_clifford_gate = True
    try:
        Clifford(node.op)
    except QiskitError:
        is_clifford_gate = False
    return is_clifford_gate
```
My only concern with these suggestions is whether they are not prohibitively slow for large clifford circuits.

UPDATE: unfortunately the suggestions do make the pass slower: a quick experiment on 10 random circuits (created by `circuits = [random_circuit(20, 1000, max_operands=4, seed=seed) for seed in range(10)]`) gives
* creating circuits: 6.13 seconds
* name-based method: 15.97 seconds
* Clifford(gate)-based method: 82.83 seconds

alexanderivrii:
another suggestion: can we use `_BASIS_1Q` and `_BASIS_2Q` defined at the end of `clifford_circuits.py`?

ShellyGarion:
@alexanderivrii - thanks for your suggestion. I also fixed `random_clifford_circuit` in a similar way.
I think it's a good solution, although it still does not handle certain `u` gates (and other 1 and 2-qubit gates) which are clifford for certain angles.

qiskit/transpiler/passes/optimization/collect_cliffords.py
```diff
@@ -69,21 +70,9 @@ def __init__(
...
-    "cx",
-    "cy",
-    "cz",
-    "swap",
-    "clifford",
-    "linear_function",
-    "pauli",
-]
+clifford_gate_names = (
+    list(_BASIS_1Q.keys()) + list(_BASIS_2Q.keys()) + ["clifford", "linear_function", "pauli"]
```
ShellyGarion: should we also add "permutation" here?

alexanderivrii: I think it's a good idea. I have checked and we can indeed construct a Clifford from a circuit containing `PermutationGate` objects, so it makes sense for the collection pass to handle these as well

alexanderivrii (approved):
LGTM, thanks for doing this!

Issue #12135, opened 2024-04-03
Controlled UGate with a parameterized angle fails with 4 or more controls


nepomechie:
### Environment

- **Qiskit version**:  1.0.0
- **Python version**: 3.11.7
- **Operating system**:  macOS 14.4.1 

### What is happening?

A controlled UGate with a parametrized angle k and with 4 (or more) controls throws an error.

Such controlled parametrized gates are useful for constructing certain custom trial states for vqe.

### How can we reproduce the issue?

```
from qiskit import *
from qiskit import QuantumRegister
from qiskit import QuantumCircuit
from qiskit.circuit.library import UGate
from qiskit.circuit import Parameter

k=Parameter("k")
q=QuantumRegister(5)
qc = QuantumCircuit(q)
ccU = UGate(k,0,0).control(4)
qc.append(ccU,[0,1,2,3,4])```

```
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
Cell In[3], line 10
      8 q=QuantumRegister(5)
      9 qc = QuantumCircuit(q)
---> 10 ccU = UGate(k,0,0).control(4)
     11 qc.append(ccU,[0,1,2,3,4])

File ~/Library/Python/3.12/lib/python/site-packages/qiskit/circuit/library/standard_gates/u.py:130, in UGate.control(self, num_ctrl_qubits, label, ctrl_state, annotated)
    128     gate.base_gate.label = self.label
    129 else:
--> 130     gate = super().control(
    131         num_ctrl_qubits=num_ctrl_qubits,
    132         label=label,
    133         ctrl_state=ctrl_state,
    134         annotated=annotated,
    135     )
    136 return gate

File ~/Library/Python/3.12/lib/python/site-packages/qiskit/circuit/gate.py:121, in Gate.control(self, num_ctrl_qubits, label, ctrl_state, annotated)
    117 if not annotated:
    118     # pylint: disable=cyclic-import
    119     from .add_control import add_control
--> 121     return add_control(self, num_ctrl_qubits, label, ctrl_state)
    123 else:
    124     return AnnotatedOperation(
    125         self, ControlModifier(num_ctrl_qubits=num_ctrl_qubits, ctrl_state=ctrl_state)
    126     )

File ~/Library/Python/3.12/lib/python/site-packages/qiskit/circuit/add_control.py:61, in add_control(operation, num_ctrl_qubits, label, ctrl_state)
     58 if isinstance(operation, UnitaryGate):
[...]

Issue #10697, opened 2023-08-23
parameters in RXXGate


Muzhou-Ma:
### Environment

python version 3.10.11

### What is happening?

I'm trying to give **`RXXGate`** a **parameter** `phi` when I encounter this error:
**_ParameterExpression with unbound parameters ({Parameter(a)}) cannot be cast to a float._**

Here is my codes:
```
phi = Parameter('a')
qc = QuantumCircuit(4)
a = RXXGate(phi).control(2)
qc = qc.compose(a)
qc.draw()
```

However, it's interesting that everything works fine if the `control `method isn't used:
```
phi_1 = Parameter('a')
qc = QuantumCircuit(4)
a = RXXGate(theta = phi_1)
qc = qc.compose(a, [2, 3])
qc.draw()
```
which can give the result:
                
```
q_0: ───────────
                
q_1: ───────────
     ┌─────────┐
q_2: ┤0        ├
     │  Rxx(a) │
q_3: ┤1        ├
     └─────────┘
```

I'm wondering what I've done wrong and how to fix it?

**Besides**, is it possible for me to make this parameter **trainable** for optimization tasks like using gradient descend?

### How can we reproduce the issue?

Perhaps making the Parameter class more flexible can help.

### What should happen?

I suppose the gate should take this parameter and give it a random(or zero) initialization.

Raghav-Bell:
You can stop the error even without assigning any value in `control()`
![image](https://github.com/Qiskit/qiskit/assets/83136390/7d9989f3-7600-48af-98b1-16e8f511ee86)

Raghav-Bell:
This error is raised from the following method.
https://github.com/Qiskit/qiskit/blob/57137ff7f2cdd6d7671aa0e6889258f28a70e5a1/qiskit/circuit/parameterexpression.py#L474-L482

Check PR #10244 .

jakelishman:
I strongly suspect that this is the same root as #7326, because the internal definition of the multi-controlled $R_{xx}$ gate will use `CU`, which has issues in Qiskit at the moment.  There's a potential fix in #9118 that I've just bumped up the priority list, because it keeps getting forgotten (by me - sorry!).

Muzhou-Ma:
Thanks for pointing out, hope to see the update soon!

jakelishman:
[...]

PR #12752, opened 2024-07-10
Add `AnnotatedOperation.params` and fix some control issues

### LINKED ISSUE #10311: `Gate.control(num_ctrl_qubits)` is not working when `num_ctrl_qubits` > 1
### Environment

- **Qiskit Terra version**: 0.24.1
- **Python version**: 3.11.3

### What is happening?

The call `gate.control(num_ctrl_qubits)` is raising an unexpected error when `gate` is one of the 2-qubit interaction `RXXGate`, `RYYGate`, `RZZGate`, `RZXGate` and `num_ctrl_qubits` is greater than 1.

```python
from qiskit.circuit.library import RXXGate

p = Parameter('p')
gate = RXXGate(p).control(2)
```

### What should happen?

The code above should work and return a multi-controlled parametric `RXXGate`.

Cryoris:
### Summary

Adds support for `AnnotatedOperation.params` and, leveraging that, fixes some issues when controlling standard gates that are parameterized. Closes #10311, #10697 and #12135.

Since this PR also changes `AnnotatedOperation` it's probably too large for a backport?

### Details and comments

* `AnnotatedOperation.params` is a proxy to to `base_op.params`, if it exists. If the `base_op` has no `params`, returns an empty list (as discussed with the team offline).
* The `annotated` argument in `Gate.control` now is `None` per default, which allows to set it to `True` for gates that cannot be eagerly synthesized. A test for each gate that failed previously is added.
* The `S/Sdg` gates didn't return `CS/CSdg` gates for a single control, that's also been fixed.


releasenotes/notes/annotated-params-116288d5628f7ee8.yaml
```diff
@@ -0,0 +1,23 @@
...
+    Add support for :meth:`.AnnotatedOperation.params`, which enables
+    circuit-level parameter handling (such as binding parameters) for
+    annotated operations.
+fixes:
+  - |
+    Fix a series of issues when controlling parameterized standard gates.
+    The controlled version of some gates (e.g. :class:`.RXXGate` or
+    :class:`.RYGate` for more than 1 control) cannot be synthesized if
+    they contain unbound parameters. Previously, this raised an error, but
+    now we create an :class:`.AnnotatedOperation` as placeholder. This
```
alexanderivrii: Maybe say something like: "Previously, this raised an error when such gates were added to the circuit, but ..."

qiskit/circuit/annotated_operation.py
```diff
@@ -219,6 +220,29 @@ def power(self, exponent: float, annotated: bool = False):
...
+        """The params of the underlying base operation."""
+        return getattr(self.base_op, "params", [])
+
+    @params.setter
+    def params(self, value: list[ParameterValueType]):
+        if hasattr(self.base_op, "params"):
+            self.base_op.params = value
+        else:
+            raise NotImplementedError(
+                f"Cannot set attribute ``params`` on the base operation {self.base_op}."
```
alexanderivrii: I am wondering if there might be a better error message type than `NotImplementedError`.

qiskit/circuit/gate.py
```diff
@@ -118,16 +117,16 @@ def control(
                 operation.
             ctrl_state: the control state in decimal or as a bitstring
                 (e.g. ``'111'``). If ``None``, use ``2**num_ctrl_qubits-1``.
-            annotated: indicates whether the controlled gate can be implemented
-                as an annotated gate.
+            annotated: indicates whether the controlled gate is implemented
+                as an annotated gate. If ``None``, this is handled as ``False``.
```
alexanderivrii: I have not looked through all of the PR yet, but I am wondering if there is ever a difference between `annotated=False` and `annotated=None`.

qiskit/circuit/library/standard_gates/multi_control_rotation_gates.py
```diff
@@ -383,6 +389,9 @@ def mcrz(
         else:
             self.append(CRZGate(lam), control_qubits + [target_qubit])
     else:
+        if isinstance(lam, ParameterExpression):
+            raise QiskitError(f"Cannot synthesize MCRZ with unbound parameter: {lam}.")
+
```
alexanderivrii, line 394: The improved error messages look very helpful!

qiskit/circuit/library/standard_gates/rxx.py
```diff
@@ -111,6 +114,39 @@ def _define(self):
 
         self.definition = qc
 
+    def control(
+        self,
+        num_ctrl_qubits: int = 1,
+        label: str | None = None,
+        ctrl_state: str | int | None = None,
+        annotated: bool | None = None,
+    ):
```
alexanderivrii, line 123: Would it have been possible to set the default value `annotated=True` without introducing `None`?

alexanderivrii (commented):
This looks really great!

One thing I am confused about is the new ternary semantics of `annotated`: with this PR it can be either `False`, `True` or `None`. Would it have been possible to achieve the same result without introducing the `None` value but by changing the default in the newly added control methods to `annotated=True`?

Previously the semantics of `annotated` was as follows: `False` - means do not create an annotated gate, this was required for backward compatibility. `True` - means introduce an annotated gate unless there is a friendlier gate available, e.g. `x.control(1, annotated=True)` would produce a `CX`-gate, not an annotated operation.

alexanderivrii:
Could you please also a test for reading/setting params for an annotated operation of an annotated operation of a gate?

Cryoris:
> Would it have been possible to achieve the same result without introducing the None value but by changing the default in the newly added control methods to annotated=True?

That would've been my preferred solution as well, but I thought the default argument cannot change in between classes derived from the same object. If that's allowed, then it would be the nicer solution 🙂

Edit: but the default value also changes with respect to the number of controls and whether the gate parameters are unbound... so we cannot really set it to `True` or `False` in the signature 🤔

qiskit/circuit/gate.py
Cryoris: Yes there is: for all parameterized gates that cannot be synthesized until the parameters are bound. Some examples are: 
- RX or U with more than 1 control
- RXX with any number of controls

alexanderivrii:
Suppose we keep the default `annotated=False` so as not change any previous behavior, but we modify the semantics of `gate.control(n, annotated)` to the following:
* if `annotated` is `False`, then we do not create an annotated operation, unless it depends on parameters
* if `annotated` is `True`, then we choose the version that is more convenient [as before]

In this way we do not change anything that used to work before, and obtain annotated operations for controlled parametric gates. Personally I prefer this slightly over the ternary semantics, but I am happy to change my opinion.

Cryoris:
The advantage of what you're describing is that we don't have to change the signature, which would be nice as it reduces the amount of required change. But this has two downsides:
1. We're implicitly changing default values (e.g. changing from `False` to `True` in some cases)
2. We're overriding user inputs (e.g. if a user explicitly sets `False` but we override it to `True`)

In the `None` scenario, I think it is easier to understand what default values are chosen, plus users can still force us to use whatever value they want. 

> if annotated is True, then we choose the version that is more convenient [as before]

I'm not sure I fully understand: if I do `RYGate(theta).control(annotated=True)` it still uses an `AnnotatedOperation` even though a `CRYGate` exists, which we could use.

qiskit/circuit/annotated_operation.py
Cryoris: Maybe `AttributeError`?

alexanderivrii:
Thanks @Cryoris. I am not against merging this PR in the current form, and as it fixes an important set of problems we should definitely include it in 1.2. However, let me try to describe again the two options which I think we have.

First option (as I am proposing). We only have two values ``annotated=False`` and ``annotated=True``. I am thinking of these as user-specified _hints_ that the user passes to a gate's ``control``/``inverse``/``power`` methods, and Qiskit is allowed to override these values. For instance, it can override ``False`` to ``True`` for controlled parametric gates. And it can override ``True`` to ``False`` if Qiskit has a native representation of this gate that it can use. Personally, I think that `RYGate(theta).control(annotated=True)` (with 1 control) should be the `CRYGate`: since we have it in Qiskit, we should probably use it (in particular, maybe we have a great default definition for this gate). In this case, the default value is always ``False``.

Second option (as you are proposing). We have three values ``annotated=False``, ``annotated=True``, and ``annotated=None`` with the default value being ``None``. I am interpreting these as follows. The value ``None`` means let Qiskit choose the best possible representation. The value of ``False`` should then represent "do not create an ``AnnotatedOperation`` and fail horribly when writing ``RYGate(theta).control(annotated=False)``". The value of ``True`` means "always introduce an annotated operation, with in particular ``XGate().control(annotated=True)`` producing an ``AnnotatedOperation`` instead of ``CXGate()``."

After writing this down so explicitly, I actually start liking the second choice much better :). Especially if the default value is ``None`` in which case Qiskit decides how the gate should be implemented. And if a user wants to shoot oneself in the foot by providing non-default values, let it be. However, it also gives the user (and transpiler passes) much better control on what the gate really is. 

So, provided that we fully agree to go with the second option, the only thing I would like to ask you is to make the semantics of ``annotated`` clear both in the documentation and in the release notes. And also please check all the other places where ``annotated`` is used to make sure it adheres with this semantics.

Cryoris:
Option 2 it is then 🙂 I added more detailed explanations about the `None` argument and resolved the other comments 👍🏻

qiskit/circuit/annotated_operation.py
ElePT: AttributeError sounds about right:

```suggestion
            raise AttributeError(
                f"Cannot set attribute ``params`` on the base operation {self.base_op}."
```

qiskit/circuit/annotated_operation.py
```diff
@@ -219,6 +220,29 @@ def power(self, exponent: float, annotated: bool = False):
...
+            )
+
+    def validate_parameter(self, parameter: ParameterValueType) -> ParameterValueType:
+        """Validate a parameter for the underlying base operation."""
+        if hasattr(self.base_op, "validate_parameter"):
+            return self.base_op.validate_parameter(parameter)
+
+        raise NotImplementedError(
```
ElePT:
```suggestion
        raise AttributeError(
```

qiskit/circuit/annotated_operation.py
```diff
@@ -219,6 +220,29 @@ def power(self, exponent: float, annotated: bool = False):
...
+    @params.setter
+    def params(self, value: list[ParameterValueType]):
+        if hasattr(self.base_op, "params"):
+            self.base_op.params = value
+        else:
+            raise NotImplementedError(
+                f"Cannot set attribute ``params`` on the base operation {self.base_op}."
+            )
+
+    def validate_parameter(self, parameter: ParameterValueType) -> ParameterValueType:
```
ElePT, line 237: Small question, would it be more coherent if we called it `validate_param`? I am a bit on the fence, I tend to prefer full words but I find it strange that the other methods say `params` and this one says `parameter`

qiskit/circuit/library/standard_gates/rx.py
```diff
@@ -90,15 +92,23 @@ def control(
             ctrl_state: control state expressed as integer,
                 string (e.g.``'110'``), or ``None``. If ``None``, use all 1s.
             annotated: indicates whether the controlled gate can be implemented
-                as an annotated gate.
+                as an annotated gate. If ``None``, this is set to ``False``, unless the
+                gate contains free parameters and has more than one control qubit,
+                in which case it cannot yet be synthesized.
```
ElePT: Is this trying to say that `None` will be treated as `True` if the gate contains free parameters and has more than one control? (I believe this is the case). Or that users are required to set it explicitly to `True`? If it's the first one, I would say:

```suggestion
                as an annotated gate. If ``None``, this is set to ``False`` except for 
                the case where the gate contains free parameters and has more than 
                one control qubit, where it is set to ``True``.
```

If it's the second one, I would say:

```suggestion
                as an annotated gate. If ``None``, this is set to ``False``. If the
                gate contains free parameters and has more than one control qubit,
                it cannot yet be synthesized unless this input is set to ``True``.
```

qiskit/circuit/library/standard_gates/rx.py
```diff
@@ -90,15 +92,23 @@ def control(
             ctrl_state: control state expressed as integer,
                 string (e.g.``'110'``), or ``None``. If ``None``, use all 1s.
             annotated: indicates whether the controlled gate can be implemented
```
ElePT: a bit out of scope, but does this indicate whether it can be implemented or whether it should be implemented? It looks to me like the second.

qiskit/circuit/library/standard_gates/rxx.py
```diff
@@ -111,6 +114,39 @@ def _define(self):
...
+
+        Args:
+            num_ctrl_qubits: number of control qubits.
+            label: An optional label for the gate [Default: ``None``]
+            ctrl_state: control state expressed as integer,
+                string (e.g.``'110'``), or ``None``. If ``None``, use all 1s.
+            annotated: indicates whether the controlled gate can be implemented
+                as an annotated gate. If ``None``, this is set to ``True`` if
+                the gate contains free parameters, in which case it cannot
+                yet be synthesized.
```
ElePT: This formulation already works a bit better than the one where I gave suggestions, they probably should be unified (I don't care about the specific message, it's just that I think it should be explicit about when it's set to true or false).

releasenotes/notes/annotated-params-116288d5628f7ee8.yaml
```diff
@@ -0,0 +1,23 @@
+---
+features_circuits:
+  - |
+    Add support for :meth:`.AnnotatedOperation.params`, which enables
```
ElePT: should the reno include the other new method too?

```suggestion
    Added support for :meth:`.AnnotatedOperation.params` and :meth:`.AnnotatedOperation.validate_param`, which enable
```

releasenotes/notes/annotated-params-116288d5628f7ee8.yaml
```diff
@@ -0,0 +1,23 @@
...
+    circuit-level parameter handling (such as binding parameters) for
+    annotated operations.
+fixes:
+  - |
+    Fix a series of issues when controlling parameterized standard gates.
```
ElePT: This is optional but I think it makes the reno nicer :)

```suggestion
    Fixed a series of issues when controlling parameterized standard gates.
```

ElePT (commented):
I think the PR looks very good, just left a few minor wording suggestions.

qiskit/circuit/annotated_operation.py
Cryoris: This is coming from `Instruction.validate_parameter`, to have `AnnotatedOperation` work as drop-in-replacement 🙂

qiskit/circuit/library/standard_gates/rx.py
Cryoris: I also think the second is correct... I can change it if we're already touching it 👍🏻

Cryoris: Done in 7a03c4f

qiskit/circuit/library/standard_gates/rxx.py
Cryoris: Updated to use this one consistently in 7a03c4f 👍🏻

qiskit/circuit/library/standard_gates/rx.py
Cryoris: I used the RXX message as you suggested below 🙂

qiskit/circuit/annotated_operation.py
Cryoris: Done in 148b5c7

releasenotes/notes/annotated-params-116288d5628f7ee8.yaml
```diff
@@ -20,4 +21,10 @@ fixes:
...
     :class:`.CSGate`, resp. :class:`.CSdgGate`, if they are controlled on
     a single control qubit.
-
+upgrade_circuits:
+  - |
+    The ``annotated`` argument of the :meth:`.Gate.control` method is now 
+    ``None``, which allows Qiskit to choose whether to annotate a controlled operation.
+    Generally, the concrete implementation (``annotated=False``) is returned, if it
+    is available, otherwise we set ``annotated=True``. This allows, for example, to 
+    defer the synthesis of controlled, parameterized gates.
```
ElePT:
```suggestion
    The ``annotated`` argument of the :meth:`.Gate.control` method is now 
    ``None``, which allows Qiskit to choose whether to annotate a controlled operation.
    If the concrete implementation (``annotated=False``) is available, it will be returned by
    default. Otherwise, the annotated implementation will be returned (``annotated=True``). 
    This allows, for example, to defer the synthesis of controlled, parameterized gates.
```

qiskit/circuit/gate.py
```diff
@@ -118,7 +118,11 @@ def control(
             ctrl_state: the control state in decimal or as a bitstring
                 (e.g. ``'111'``). If ``None``, use ``2**num_ctrl_qubits-1``.
             annotated: indicates whether the controlled gate is implemented
-                as an annotated gate. If ``None``, this is handled as ``False``.
+                as an annotated gate. If ``None``, this is set to ``False``,
+                if the controlled gate can directly be constructed, and otherwise
+                set to ``True``. This allows defering the construction process in case the
+                synthesis of the controlled gate requires more information (e.g.
+                values of unbound parameters).
```
ElePT:
```suggestion
                as an annotated gate. If ``None``, this is set to ``False``
                if the controlled gate can directly be constructed, and otherwise
                set to ``True``. This allows defering the construction process in case the
                synthesis of the controlled gate requires more information (e.g.
                values of unbound parameters).
```

ElePT (dismissed):
LGTM! I just added two comma-related minor suggestions. After they are applied I think the PR is ready to be merged.

ElePT (approved):

PR #12755, opened 2024-07-11
Make `BitArray.{slice_bits,slice_shots,__getitem__}` raise `IndexError` when indices are not valid

t-imamichi:
### Summary

This PR updates `BitArray.{slice_bits,slice_shots,__getitem__}` to raise `IndexError` if indices are not valid.

Especially, `__getitem__` raise IndexError with appropriate message as follows (`ba[0, 0, 0]` case).

```python

ba = BitArray.from_counts({"0": 2, "1": 3})
ba = ba.reshape(1, 1)

try:
    ba[1]
except Exception as ex:
    print(type(ex), ex)
try:
    ba[0, 0, 0]
except Exception as ex:
    print(type(ex), ex)
try:
    ba[0, 0, 0, 0]
except Exception as ex:
    print(type(ex), ex)
```

main branch
```
<class 'IndexError'> index 1 is out of bounds for axis 0 with size 1
<class 'ValueError'> The input array must have at least two axes.
<class 'ValueError'> BitArrays cannot be sliced along the bits axis, see slice_bits() instead.
```

this PR
```
<class 'IndexError'> index 1 is out of bounds for axis 0 with size 1
<class 'IndexError'> BitArray cannot be sliced along the shots axis, use slice_shots() instead.
<class 'IndexError'> BitArray cannot be sliced along the bits axis, use slice_bits() instead.
```

### DIFF
--- a/qiskit/primitives/containers/bit_array.py
+++ b/qiskit/primitives/containers/bit_array.py
@@ -130,10 +130,15 @@ class BitArray(ShapedMixin):
         return f"BitArray({desc})"
 
     def __getitem__(self, indices):
-        if isinstance(indices, tuple) and len(indices) >= self.ndim + 2:
-            raise ValueError(
-                "BitArrays cannot be sliced along the bits axis, see slice_bits() instead."
-            )
+        if isinstance(indices, tuple):
+            if len(indices) == self.ndim + 1:
+                raise IndexError(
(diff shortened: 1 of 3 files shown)


ElePT (approved):
This looks like a straightforward fix that improves the BitArray user experience. I think it should be backported to 1.1 and 1.2.

ElePT:

Issue #12156, opened 2024-04-08
Update `DAGCircuit` to use `rustworkx` 0.15 methods


jakelishman:
### What should we add?

There's a couple of things added to `rustworkx` that we might want to look into using for `DAGCircuit` and various transpiler passes, once `rustworkx` releases:

- `remove_node_retain_edges_by_key` in place of `remove_node_retain_edges`: this fixes the computational complexity of node removal for our wire structure for $\mathcal O(n^2)$ to $\mathcal O(n)$ for node degree $n$.  That's a _huge_ win for things like final barriers on very wide circuits. (See also #11677.)
- the topological sorters now have keyword arguments `reverse` and `initial`.  Places that use a reverse topological sort can now just manually specify that, and places that only care about starting the topological sort from a certain set of nodes that dominate all the areas of interest can use `initial`.  For places that will only need to look a couple of steps into the topological order, using `TopologicalSorter` with `initial` can avoid a scan over the entire DAG (though beware that `TopologicalSorter` defaults to `check_cycles=True` which is a waste of time for us).

There may well be more, this issue is just to remind us to check before the release.

PR #12756, opened 2024-07-11
Use rustworkx 0.15.0 features in DAGCircuit.remove_op_node

### LINKED ISSUE #11677: Fix overhead of `DAGCircuit.remove_op_node`
### What should we add?

Right now the `DAGCircuit,remove_op_node` function has quadratic overhead as a function of the number of bits/dag edges are on an operation. This is do to how the underlying `rustworkx.PyDiGraph.remove_node_retain_edges` method is used by `remove_op_node`. Normally this doesn't matter because in most cases this is only ever called on operations with 1 or 2 qubits. However, since the introduction of #10323 this is now getting called on a barrier of n qubits for the the width of the target backend. This can become a large performance bottleneck, for example when targeting a 60k qubit backend calling `remove_op_node` took 47 min while it was processing the list of qubits.

To fix this optimally we'll need to likely add a new interface to rustworkx to handle this more efficiently.

mtreinish:
### Summary

This commit updates the minimum rustworkx version to 0.15.0 to pull in the new PyDiGraph.remove_node_retain_edges_by_id() method introduced in that release. This new function is used for the DAGCircuit.remove_op_node() method instead of the
PyDiGraph.remove_node_retain_edges() function. This new method has much better scaling characteristics and should improve the performance characteristics of removing very wide operations from a DAGCircuit.

Fixes #11677
Part of #12156

### Details and comments

### DIFF
--- a/qiskit/dagcircuit/dagcircuit.py
+++ b/qiskit/dagcircuit/dagcircuit.py
@@ -1986,9 +1986,7 @@ class DAGCircuit:
                 "node type was wrongly provided."
             )
 
-        self._multi_graph.remove_node_retain_edges(
-            node._node_id, use_outgoing=False, condition=lambda edge1, edge2: edge1 == edge2
-        )
+        self._multi_graph.remove_node_retain_edges_by_id(node._node_id)
         self._decrement_op(node.name)
 
     def remove_ancestors_of(self, node):
--- a/releasenotes/notes/update-rustworkx-min-version-4f07aacfebccae80.yaml
+++ b/releasenotes/notes/update-rustworkx-min-version-4f07aacfebccae80.yaml
@@ -1,0 +1,7 @@
+---
+upgrade_misc:
+  - |
+    The minimum version of rustworkx required to run this release has been
+    increased from 0.14.0 to 0.15.0. This is required because Qiskit is now
+    using new functionality added in the rustworkx 0.15.0 release which
+    improves performance.
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,4 +1,4 @@
-rustworkx>=0.14.0
+rustworkx>=0.15.0
 numpy>=1.17,<3
 scipy>=1.5
 sympy>=1.3


qiskit/dagcircuit/dagcircuit.py
```diff
@@ -1986,9 +1986,7 @@ def remove_op_node(self, node):
                 "node type was wrongly provided."
             )
 
-        self._multi_graph.remove_node_retain_edges(
-            node._node_id, use_outgoing=False, condition=lambda edge1, edge2: edge1 == edge2
-        )
+        self._multi_graph.remove_node_retain_edges_by_id(node._node_id)
```
jakelishman, line 1989: I think `by_id` is slightly risky here while bits still have an equality that isn't entirely referential.

I'm fairly sure it's possible to end up with edges that are constructed from different `bit` instances, since the objects that go into the wire payload are passed in by the user to `apply_operation_back`. It sounds unlikely, but I'm fairly sure it is possible, and I'd hate to debug that.

Maybe use `by_key` just until we've closed that loophole?

mtreinish: I was thinking about that, but I wasn't too worried because I was thinking the only object identity we care about is between incoming and outgoing edge pairs around a node. We don't have to worry about using the same object globally for a dag as long as the local edge weights use the same object. I was thinking that we'll always use the same object when doing that because we always add the edges in pairs.

But if you think it's safe to use `by_key` I'll update this.

jakelishman: I wouldn't put it past either us or a user pass doing `canonical_register = QuantumRegister("q", dag.num_qubits)` and using that in `apply_operation_back` on an existing circuit. That's still technically allowed behaviour, so the key is probably safer for now. The move to Rust space or removal of the register/index bit properties will let us swap to the preferential id one.

mtreinish: But that wouldn't matter for this as `apply_operation_back()` uses rustworkx's `insert_node_on_in_edges_multiple()` which pulls the edge weights from the existing edge on the graph and just uses the same object for the two new edges it adds when inserting a node in between an existing edge. That's what I was saying, the only way I could see us using a different object id is if we manipulated the graph directly to do so, otherwise we do everything together.

But as I said before I'll update this to use by_key instead if you're worried about the implications here it's not that much slower.

jakelishman: Ahh yeah, I forgot that the weight given isn't used verbatim on the edges in the split, just used to make a key function. In that case I'm fine with id.

jakelishman (approved):

PR #12761, opened 2024-07-11
 Port `star_preroute` to rust

### LINKED ISSUE #12337: Port `StarPrerouting` to Rust
This PR can be started now and partially completed before #11721 and #12205 as it only does structural analysis and inserts swaps. We can leverage the existing mechanisms used for `SabreLayout` and `SabreSwap` in rust currently to do this. To full mark this issue as complete we will need #11721 and #12205 so that the pass can operate solely in rust, but as a first step this isn't blocked.

henryzou50:
### Summary

Closes [#12337 ](https://github.com/Qiskit/qiskit/issues/12337)

This PR ports the core logic of `star_preroute` from Python to Rust. The changes involve creating a new Rust module for the star prerouting algorithm and updating the corresponding Python code to integrate with this new Rust functionality. 

These changes only involve the first part of porting `star_prerouting` to rust, as it mainly handles porting the logic of transforming the DAG with the StarBlocks to rust, but we will still need to decide how we should port collecting the StarBlocks to rust. This is a bit more tricky as there are still ongoing discussion on how we can extend the block collection, such as in [PR #12498 ](https://github.com/Qiskit/qiskit/pull/12498). 

### Details and comments
- New Rust file: Added `star_preroute.rs` to handle the core logic of the function `star_preroute` from the python side. This file defines the type aliases for node and block representations, which matches the current block representation of the `StarBlock` (except that the center is just a bool, as we only need to know if there is a center), and the node representation matches how the nodes used in `SabreDAG`. The `star_preroute` function processes the star blocks witihin the `SabreDAG`  to finds the linear routing equivalent and then returns the result as a `SabreResult`.
- Node representation: A key part of this implementation is how it takes advantage of `SabreResult` and `SabreDAG`, so the node representation is a tuple of the node id, list of qubit indices, a set of classical bit indices, and a directive flag. However, once we update the regular DAG to rust, this part may change significantly, but the main logic will remain unchanged.
- Updates in the SABRE rust module: To use `sabre_dag.rs` and `swap_map.rs` in `star_prerouting`, I change them to be public in `crates/accelerate/src/sabre/mod.rs`. Not sure if it makes more sense to do it this way or to move `star_prerouting` to `crates/accelerate/src/sabre/` since it mimics the methods used in Sabre to change the dag.
- Python side updates: Imported the necessary modules and only modified the function `star_preroute` so that now the function performs the heavy lifting of transforming the DAG within the Rust space, leveraging `_build_sabre_dag` and `_apply_sabre_result`.
- Issues: I was not sure how correctly handle nodes with control flow from the rust side. I know that `route.rs` handles this with `route_control_flow_block` to populate the `node_block_results` for `SabreResult`, but I was not sure how to best take advantage of this function for `star_prerouting`. Currently, the `node_block_results` for `star_prerouting` essentially always empty and just there to have`SabreResult`. There also seems to be no unit tests for `star_prerouting` that includes control flow.


crates/accelerate/src/star_prerouting.rs
```diff
@@ -0,0 +1,215 @@
...
+/// - List of involved qubit indices (Vec<VirtualQubit>)
+/// - Set of involved classical bit indices (HashSet<usize>)
+/// - Directive flag (bool)
+type Nodes = (usize, Vec<VirtualQubit>, HashSet<usize>, bool);
+
+/// Type alias for a block representation.
+/// Each block is represented by a tuple containing:
+/// - A boolean indicating the presence of a center (bool)
+/// - A list of nodes (Vec<Nodes>)
+type Block = (bool, Vec<Nodes>);
```
sbrandhsn, line 25: I'm not sure whether we wanted to keep `type Block = (NodeId, Vec<Nodes>)` - the memory savings are marginal and we may need to do follow-up analysis on the center qubit at some point in order to decide whether we would actually like to run star-prerouting (SPR). We can also add this later when/if it becomes necessary. :-)

crates/accelerate/src/star_prerouting.rs
```diff
@@ -0,0 +1,215 @@
...
+
+    // Structures for SabreResult
+    let mut out_map: HashMap<usize, Vec<[PhysicalQubit; 2]>> =
+        HashMap::with_capacity(dag.dag.node_count());
+    let mut gate_order: Vec<usize> = Vec::with_capacity(dag.dag.node_count());
+    let node_block_results: HashMap<usize, Vec<BlockResult>> = HashMap::new();
+
+    // Process each node in the given processing order
+    for node in processing_order.iter() {
+        if let Some(block_id) = find_block_id(&blocks, node) {
```
sbrandhsn: This is n² time complexity, we can introduce a hash map from node to block to make this n + n at the cost of memory (if it ever became a problem).

crates/accelerate/src/star_prerouting.rs
```diff
@@ -0,0 +1,215 @@
...
+}
+
+/// Processes a star block, applying operations and handling swaps.
+///
+/// Args:
+///
+/// * `qubit_mapping` - A mutable reference to the qubit mapping vector.
+/// * `block` - A tuple containing a boolean indicating the presence of a center and a vector of nodes representing the star block.
+/// * `last_2q_gate` - The last two-qubit gate in the processing order.
+/// * `is_first_star` - A mutable reference to a boolean indicating if this is the first star block being processed.
```
sbrandhsn, line 130: Does `is_first_star` need to be mutable? I expected that the processing order would guarantee that we do not process the same block twice.

qiskit/transpiler/passes/routing/star_prerouting.py
```diff
@@ -305,113 +309,84 @@ def star_preroute(self, dag, blocks, processing_order):
...
-            last_2q_gate = last_2q_gate[0]
-        else:
-            last_2q_gate = None
 
+        # Determine the processing order of the nodes in the DAG for the Rust representation
         int_digits = floor(log10(len(processing_order))) + 1
         processing_order_index_map = {
-            node: f"a{str(index).zfill(int(int_digits))}"
-            for index, node in enumerate(processing_order)
+            node: f"a{index:0{int_digits}}" for index, node in enumerate(processing_order)
```
sbrandhsn, line 329: Neat, I did not know about this syntax! :-)

sbrandhsn (changes requested):
This mostly looks great to me, thanks! :-) I left a couple of in-line comments that should partially be addressed in my opinion. Apart from this:
1. Do you think porting SPR to Rust requires any extra testing?
2. I think a good additional test would be to route our standard QFT definition and check whether the resulting circuit requires n-1 swap gates where n is the number of CNOT gates and check whether the QFT circuit has been linearized (you can run https://github.com/Qiskit/qiskit/blob/1191fcbc7277216ce03d30f1bf0d2767ea681cd9/qiskit/transpiler/passes/layout/vf2_utils.py#L29 and check whether #edges -1 == #vertices)?
3. Can you run a quick benchmark on QFT to see how the runtime improves? I'm thinking about just routing the QFT circuit and measuring the time of doing that (no ASV or so required).

WRT circuits with classical feedback, I think it is fine to just have it as is right now. Ideally, we would check whether we wanted to run SPR at all and a `ControlFlowOp` may very likely result in blocking a SPR run.

crates/accelerate/src/star_prerouting.rs
henryzou50: Yes, I agree. At the time, I was uncertain as well. Currently, the logic in `star_preroute` only requires us to know if the block is the center or not, so I used `type Block = (bool, Vec<Nodes>)`. When we determine the specifics of the follow-up analysis on the center qubit, we can include more information. For now, I think this approach is sufficient.

crates/accelerate/src/star_prerouting.rs
henryzou50: Nice catch and thank you! I will make that change.

crates/accelerate/src/star_prerouting.rs
henryzou50: You are correct that the processing order would guarantee that we do not process the same block twice, but I think it is good to have `is_first_star` be mutable here since the `is_first_star flag` is used to indicate whether the current block being processed is the first star block encountered. It needs to be mutable because it gets updated within the `process_block` function to ensure that subsequent blocks are not treated as the first star block.

henryzou50:
> This mostly looks great to me, thanks! :-) I left a couple of in-line comments that should partially be addressed in my opinion. Apart from this:
> 
>     1. Do you think porting SPR to Rust requires any extra testing?
> 
>     2. I think a good additional test would be to route our standard QFT definition and check whether the resulting circuit requires n-1 swap gates where n is the number of CNOT gates and check whether the QFT circuit has been linearized (you can run https://github.com/Qiskit/qiskit/blob/1191fcbc7277216ce03d30f1bf0d2767ea681cd9/qiskit/transpiler/passes/layout/vf2_utils.py#L29
>         and check whether #edges -1 == #vertices)?
> 
>     3. Can you run a quick benchmark on QFT to see how the runtime improves? I'm thinking about just routing the QFT circuit and measuring the time of doing that (no ASV or so required).
> 
> 
> WRT circuits with classical feedback, I think it is fine to just have it as is right now. Ideally, we would check whether we wanted to run SPR at all and a `ControlFlowOp` may very likely result in blocking a SPR run.

1. I don't think porting `StarPrerouting` to rust requires any extra testing at the moment, especially since this PR only ports the logic of the function `star_preroute`. I think the tests in `test_star_prerouting.py` are cover the correctness of `star_preroute. 

2. That is a great test! I just added it, but I have some clarifying questions: 
- Shouldn't the resulting circuit requires **n-2** swap gates where n is the number of **cp** gates 
- For instance, in the 6-qubit QFT after applying `StarPreRouting` we get

```                                                                          ░ ┌───┐                                                  ░ ┌───┐                                     ░ ┌───┐                         ░ ┌───┐          ░       ░ 
q_4 -> 0 ──────■──────────────────────────────────────────────────────────░─┤ H ├─■────────X───────────────────────────────────────░─┤ H ├─■────────X──────────────────────────░─┤ H ├─■────────X──────────────░─┤ H ├─■────────░───────░─
         ┌───┐ │P(π/2)                                                    ░ └───┘ │P(π/2)  │                                       ░ └───┘ │P(π/2)  │                          ░ └───┘ │P(π/2)  │              ░ └───┘ │P(π/2)  ░ ┌───┐ ░ 
q_5 -> 1 ┤ H ├─■────────■────────X────────────────────────────────────────░───────■────────X──■────────X───────────────────────────░───────■────────X──■────────X──────────────░───────■────────X──■────────X──░───────■────────░─┤ H ├─░─
         └───┘          │P(π/4)  │                                        ░                   │P(π/4)  │                           ░                   │P(π/4)  │              ░                   │P(π/4)  │  ░                ░ └───┘ ░ 
q_3 -> 2 ───────────────■────────X──■────────X────────────────────────────░───────────────────■────────X──■────────X───────────────░───────────────────■────────X──■────────X──░───────────────────■────────X──░────────────────░───────░─
                                    │P(π/8)  │                            ░                               │P(π/8)  │               ░                               │P(π/8)  │  ░                               ░                ░       ░ 
q_2 -> 3 ───────────────────────────■────────X──■─────────X───────────────░───────────────────────────────■────────X──■─────────X──░───────────────────────────────■────────X──░───────────────────────────────░────────────────░───────░─
                                                │P(π/16)  │               ░                                           │P(π/16)  │  ░                                           ░                               ░                ░       ░ 
q_1 -> 4 ───────────────────────────────────────■─────────X──■─────────X──░───────────────────────────────────────────■─────────X──░───────────────────────────────────────────░───────────────────────────────░────────────────░───────░─
                                                             │P(π/32)  │  ░                                                        ░                                           ░                               ░                ░       ░ 
q_0 -> 5 ────────────────────────────────────────────────────■─────────X──░────────────────────────────────────────────────────────░───────────────────────────────────────────░───────────────────────────────░────────────────░───────░─
                                                                          ░                                                        ░                                           ░                               ░                ░       ░ 
                                                                          
``` 
where only the first and last cp gates do not have a corresponding swap gate. 

3. In a benchmark with a 100 qubit QFT: 
``` 
(main)     Took: 29.963886976242065s
(this PR) Took: 23.62939691543579s              
```                                          
Which shows that this PR is about 21% faster.

henryzou50:
@sbrandhsn Thank you for your comments and I finished adding the changes!

While doing these changes, I also made another change of changing how the `swap_map` of the `SabreResult` is handled in this case. I changed the `node_id` in `apply_swap` of `star_preroute.rs` to use the current node id instead of the next node id as it is more efficient, but just does not match how we normally do it in `SabreSwap`. This makes it so that we apply the gate first then the swap, which also fixes an error we had before where we never placed a swap gate at the end of the processing a block. This only affected tests where we had multiple blocks to process. To make sure we apply the results correctly from `SabreResult`, I added a flag to `_apply_sabre_result` to treat the case of `StarPreRouting` differently so that it applies the swap after applying the node.

sbrandhsn:
Looks good, thanks for addressing these. Yeah it is `n-2` I forgot about not needing a swap gate for the first 2q-gate. :-) Do you see a scenario where a user may want to have control over `apply_swap_first`? I currently do not see one, so I would suggest to remove that flag and keep whatever default behavior you deem to be better. :-) This behavior is not covered by our stability policy, so I don't see an issue in changing it. :-)

Thanks for your work. I will accept the PR as soon as `apply_swap_first` is handled. :-)

henryzou50:
> Looks good, thanks for addressing these. Yeah it is `n-2` I forgot about not needing a swap gate for the first 2q-gate. :-) Do you see a scenario where a user may want to have control over `apply_swap_first`? I currently do not see one, so I would suggest to remove that flag and keep whatever default behavior you deem to be better. :-) This behavior is not covered by our stability policy, so I don't see an issue in changing it. :-)
> 
> Thanks for your work. I will accept the PR as soon as `apply_swap_first` is handled. :-)

Thanks! No, I do not see a scenario where a user may want to have control over `apply_swap_first`, as this should just mainly be used for SABRE. I removed the flag and reimplemented the default behavior :)

sbrandhsn:
LGTM! Can you resolve the merge conflict? :-) I would be happy to approve then! Thanks!

henryzou50:
> LGTM! Can you resolve the merge conflict? :-) I would be happy to approve then! Thanks!

Done! Thanks :)

sbrandhsn (approved):
LGTM, thanks!

PR #12762, opened 2024-07-11
Move `generate_preset_pass_manager` to a standalone file

ElePT:
### Summary
`generate_preset_pass_manager` deserves a proper home. This PR moves this function from the `__init__` file in `transpiler/preset_passmanagers` to a standalone file in the same directory, `generate_preset_pass_manager.py`. 
Note that this move doesn't affect the public interface or documented import path, but it makes it easier to find the source code when traversing the repo.

### Details and comments
The PR now adds a global import path for `generate_preset_pass_manager` and contributes to #11897.


mtreinish (commented):
The other thing we should do as part of this is expose this from `qiskit/__init__.py` since it's "commonly" used.

releasenotes/notes/add-generate-preset-pm-global-import-efb12f185f3f738b.yaml
```diff
@@ -0,0 +1,11 @@
+---
+upgrade_transpiler:
```
ElePT: I think this is probably an upgrade? It does increase usability. I am also not sure if "global import" is the best term, I also thought of "top level import".

mtreinish: It should probably a feature note since we're adding something new. An upgrade note would normally be reserved if we were changing something that might require user input when upgrading to the new version. This is just allowing a simpler import but the old path still works fine and is a supported part of our public api.

```suggestion
features_transpiler:
```

releasenotes/notes/add-generate-preset-pm-global-import-efb12f185f3f738b.yaml
```diff
@@ -0,0 +1,11 @@
...
+  - |
+    Added a new global import for :func:`.generate_preset_pass_manager`, so that it can be imported as::
```
mtreinish:
```suggestion
    Added a new import path option for :func:`.generate_preset_pass_manager`, so that it can now be imported as::
```

mtreinish (commented):
LGTM, thanks for the quick update and adding the top level import too. Just a few small nits in the release note, but otherwise this is ready to merge.

mtreinish (dismissed):

jakelishman:
Might we be amenable to making this name available from `qiskit.transpiler` as well?

ElePT:
> Might we be amenable to making this name available from `qiskit.transpiler` as well?

@jakelishman Makes sense to me, added path in [ebb48ac](https://github.com/Qiskit/qiskit/pull/12762/commits/ebb48acd2beeb6ec84068b26a2d79e05d27e0138).

jakelishman (dismissed):
Nice, thanks!

jakelishman (dismissed):

ElePT:
While dealing with the cyclic imports I realized that, in #12185, I hadn't migrated two input handling functions (the ones for `layout` and `approximation_degree`) to `generate_preset_pass_manager` to keep a consistent behavior between transpile and generate ppm. Fixed in [2bac8f1](https://github.com/Qiskit/qiskit/pull/12762/commits/2bac8f19f7f53f8982234b1c9fc18eab1ae4de96).

Now that I think of it, maybe I should open an independent PR with these fixes, it might be a bit clearer.

Edit: changes extracted to #12799.

releasenotes/notes/add-generate-preset-pm-global-import-efb12f185f3f738b.yaml
```diff
@@ -0,0 +1,14 @@
...
+  - |
+    Added a new import path option for :func:`.generate_preset_pass_manager`, so that it can now be imported as::
+    
+      from qiskit import generate_preset_pass_manager
+    
+    instead of having to type the full path::
+      
+      from qiskit.transpiler.preset_passmanagers import generate_preset_pass_manager
+    
+    Starting in Qiskit 1.2, the function is also importable from the :mod:`qiskit.transpiler` module as::
```
mtreinish:
```suggestion
    The function is also importable from the :mod:`qiskit.transpiler` module as::
```

mtreinish (dismissed):
LGTM, thanks for getting to the bottom of those hairy import cycles

mtreinish (approved):

PR #12730, opened 2024-07-07
Rebalance `CircuitInstruction` and `PackedInstruction`

jakelishman:
### Summary

This is a large overhaul of how circuit instructions are both stored in Rust (`PackedInstruction`) and how they are presented to Python (`CircuitInstruction`).  In summary:

* The old `OperationType` enum is now collapsed into a manually managed `PackedOperation`.  This is logically equivalent, but stores a `PyGate`/`PyInstruction`/`PyOperation` indirectly through a boxed pointer, and stores a `StandardGate` inline.  As we expect the vast majority of gates to be standard, this hugely reduces the memory usage.  The enumeration is manually compressed to a single pointer, hiding the discriminant in the low, alignment-required bytes of the pointer.

* `PackedOperation::view()` unpacks the operation into a proper reference-like enumeration `OperationRef<'a>`, which implements `Operation` (though there is also a `try_standard_gate` method to get the gate without unpacking the whole enumeration).

* Both `PackedInstruction` and `CircuitInstruction` use this `PackedOperation` as the operation storage.

* `PackedInstruction` is now completely the Rust-space format for data, and `CircuitInstruction` is purely for communication with Python.

On my machine, this commit brings the utility-scale benchmarks to within 10% of the runtime of 1.1.0 (and some to parity), despite all the additional overhead.

Changes to accepting and building Python objects
------------------------------------------------

* A `PackedInstruction` is created by copy constructor from a `CircuitInstruction` by `CircuitData::pack`.  There is no `pack_owned` (really, there never was - the previous method didn't take ownership) because there's never owned `CircuitInstruction`s coming in; they're Python-space interop, so we never own them (unless we clone them) other than when we're unpacking them.

* `PackedInstruction` is currently just created manually when not coming from a `CircuitInstruction`.  It's not hard, and makes it easier to re-use known intern indices than to waste time re-interning them. There is no need to go via `CircuitInstruction`.

* `CircuitInstruction` now has two separated Python-space constructors: the old one, which is the default and takes `(operation, qubits, clbits)` (and extracts the information), and a new fast-path `from_standard` which asks only for the standard gate, qubits and params, avoiding operator construction.

* To accept a Python-space operation, extract a Python object to `OperationFromPython`.  This extracts the components that are separate in Rust space, but joined in Python space (the operation, params and extra attributes).  This replaces `OperationInput` and `OperationTypeConstruct`, being more efficient at the extraction, including providing the data in the formats needed for `PackedInstruction` or `CircuitInstruction`.

[...]

PR #12766, opened 2024-07-11
Make `CircuitData` inner data accessible in Rust

raynelfss:
### Summary
The following commits aim to add a way to iterate through all the instructions contained in `CircuitData` to make these instructions accessible through Rust.

### Details and comments
- Implement an `iter()` method that iterates through the instructions in `CircuitData` in Rust.
- Make `PackedInstruction` public:
   This class is being made public and reorganized by #12730.

### Use cases:
In #12585 there's a temporary struct called `CircuitRep` which aims to represent the circuit. This struct has a `data()` method that retrieves the data from the circuit once and stores it as a `Vec<CircuitInstruction>`.
https://github.com/raynelfss/qiskit/blob/bf9aedd72b019600b223e71f46e0107a295457b3/crates/circuit/src/equivalence.rs#L311-L319
```rust
    pub fn data(&mut self, py: Python) -> PyResult<&[CircuitInstruction]> {
        if self.data.is_none() {
            let data = self.object.bind(py).getattr("data")?.extract()?;
            self.data = Some(data);
            return Ok(self.data.as_ref().unwrap());
        }
        return Ok(self.data.as_ref().unwrap());
    }
```

The method is used once to create a set of instructions based on their name and the number of qubits they affect. https://github.com/raynelfss/qiskit/blob/bf9aedd72b019600b223e71f46e0107a295457b3/crates/circuit/src/equivalence.rs#L448-L452

```rust
        let sources: HashSet<Key> =
            HashSet::from_iter(equivalent_circuit.data(py)?.iter().map(|inst| Key {
                name: inst.operation.name().to_string(),
                num_qubits: inst.operation.num_qubits(),
            }));
```
By extracting from `CircuitData` and avoiding an extra conversion to a different DataType, it will be easier for us to access those instruction attributes from the Rust side of things without needing to invoke Python.

### DIFF
--- a/crates/circuit/src/circuit_data.rs
+++ b/crates/circuit/src/circuit_data.rs
@@ -1167,4 +1167,9 @@ impl CircuitData {
             py_op: RefCell::new(inst.py_op.clone()),
         })
     }
+
+    /// Rust native method, returns an iterator over all the instructions present in the circuit.
+    pub fn iter(&self) -> impl Iterator<Item = &PackedInstruction> {
+        self.data.iter()
+    }
 }
--- a/crates/circuit/src/circuit_data.rs
+++ b/crates/circuit/src/circuit_data.rs
@@ -1168,7 +1168,7 @@ impl CircuitData {
         })
     }
 
-    /// Rust native method, returns an iterator over all the instructions present in the circuit.
+    /// Returns an iterator over all the instructions present in the circuit.
     pub fn iter(&self) -> impl Iterator<Item = &PackedInstruction> {
         self.data.iter()
     }
--- a/crates/circuit/src/circuit_instruction.rs
+++ b/crates/circuit/src/circuit_instruction.rs
@@ -44,7 +44,7 @@ pub struct ExtraInstructionAttributes {
 
 /// Private type used to store instructions with interned arg lists.
 #[derive(Clone, Debug)]
-pub(crate) struct PackedInstruction {
+pub struct PackedInstruction {
     /// The Python-side operation instance.
     pub op: OperationType,
     /// The index under which the interner has stored `qubits`.


crates/circuit/src/circuit_data.rs
```diff
@@ -1167,4 +1167,9 @@ impl CircuitData {
             py_op: RefCell::new(inst.py_op.clone()),
         })
     }
+
+    /// Rust native method, returns an iterator over all the instructions present in the circuit.
```
jakelishman: We won't need to say "Rust native method" - _all_ the methods we call from Rust space are Rust native, even the ones that PyO3 provides a wrapper for so that Python can call them.  Rust doesn't pay any overhead to call `pymethod`s.

jakelishman (commented):
This seems generally very sensible to me.  Will `CircuitData::iter` be enough - will you not need to access the actual qubits / clbits at some point?  If so, we'll not get far without having access to the interners as well.

raynelfss:
> will you not need to access the actual qubits / clbits at some point?  If so, we'll not get far without having access to the interners as well.

For the time being I only need access to the instructions name and the number of qubits affected by it. 

Although it will be needed to know what qargs and/or cargs is the instruction affecting, I'd rather wait for #12730 to merge and do a follow up to this to use the newer architecture your rebalancing will introduce.

Cryoris:
I just came across this in a case where we need to repeat the circuit data on other qubits -- in that case we'd also need access to the qubits/clbits 🤔

Cryoris:
I've also come across this but would need access to the qubits and clbits. Couldn't we just already include that info in this PR and extend iter to something like 
```rust
    pub fn iter(&self) -> impl Iterator<Item = (&PackedInstruction, &Vec<Qubit>, &Vec<Clbit>)> {
        self.data.iter().map(|inst| {
            (
                inst,
                self.qargs_interner.intern(inst.qubits_id).value,
                self.cargs_interner.intern(inst.clbits_id).value,
            )
        })
    }
```
We could argue that iterating over circuit data should not just give the instructions but all the information (including qubits/clbits) 🙂

jakelishman:
Julien: what you're getting at here is kind of the "middle type" between `PackedInstruction` and `CircuitInstruction` that you might have heard referred to in meetings / on PRs with Ray, me or Kevin, etc.

Can you give a bit more context at what exactly you're doing that wants it?

Some points: in general, we want to avoid doing work that we don't always need to do.  Not all uses of the iterator actually need to unwrap the qubit/clbit slices, which means we don't need to look them up at all, and don't need to copy around the references. Having access to the intern index gives the most performant access to the qubits/clbits in situations like copy, etc - if you have the intern key and access to the interner, you can just ask for it when you need it (my comment above).  That's not at all to say that there _aren't_ uses for a middle, fully unwrapped type as a convenience (why I'm asking), but we most likely don't want the non-critical-loop convenience method to be the only thing available.

jakelishman:
Just as a very rough example, consider lots of block-collection passes for peephole quantum optimisation.  They don't care _what_ the clbits are other than if they're 0 or not; for those, we don't need to waste time and memory unpacking the clbit from the intern index on each iteration through.

jakelishman (approved):
Thanks Ray.

From offline discussion: Julien's going to take a look at reimplementing `QuantumCircuit.compose` with an inner `CircuitData::compose`, which will immediately alleviate some of the things he's talking about, but is also an entry point to us figuring out more of what we might need with the "middle type".

The concerns are still real, but I don't think we need to address them immediately as part of this PR, since we don't know exactly what path we'll take.

PR #12767, opened 2024-07-12
Fixed ZZFeatureMap not accepting a list of entanglement

### LINKED ISSUE #12432: ZZFeatureMap does not accept layer lists for entanglement kwarg
### Environment

- **Qiskit version**: 1.0.2
- **Python version**: 
- **Operating system**: 

### What is happening?

```python

N = 3
layer1 = [(0,1)]
layer2 = [(1,2)]

qc = ZZFeatureMap(N, reps=1, entanglement=[layer1, layer2])
qc.draw()
```
`CircuitError: 'Number of items in qubits parameter (2) does not match number of qubits in the circuit (1).'`

run above

### What should happen?

It should work like `TwoLocal` does.

### Any suggestions?

_No response_

shravanpatel30:
### Summary
This pull request fixes #12432 

When a List[List(int)] or List[List[List[int]]] or List[List[List[List[int]]]] is passed to the entanglement in `ZZFeatureMap`, it returns an error. This should work like `TwoLocal` does. 

### Details and comments

Because both `ZFeatureMap` and `ZZFeatureMap` inherits from the `PauliFeatureMap` the problem was with how `PauliFeatureMap` handles the entanglement lists.  To generate proper entanglement layers `PauliFeatureMap` relies on the `get_entangler_map()` (the method which takes 3 arguments) method from `NLocal` which handles invalid inputs and generates properly formatted entanglement layers. But unlike `TwoLocal` where we specify rotation blocks and the entanglement blocks as separate arguments, for `PauliFeatureMap` the paulis are specified as a single list. Due to this, we need some additional logic to handle how entanglement layers are generated from the user specified entanglement list. 

To fix this I have added a helper method `_selective_entangler_map()` and also added `get_entangler_map()` method to `PauliFeatureMap`. The newly specified `get_entangler_map()` method will override the `get_entangler_map()` method from the `NLocal` class for List[List(int)] or List[List[List[int]]] or List[List[List[List[int]]]] but if any other input is passed to entanglement (like a str `full`) then we fall back on the method from `NLocal` to handle those cases.

With my fix all the below examples work as expected:

- If entanglement is a List[List(int)]:
```
N = 5
layer1 = (0,1)
layer2 = (1,0)

qc = PauliFeatureMap(N, reps=2, paulis=['Z', 'ZZ'], entanglement=[layer1, layer3], insert_barriers=True)
qc.decompose().draw('mpl')
```
List List int

-  If entanglement is a List[List[List[int]]] (This means that entanglement will change according to reps):
```
N = 5
layer1 = [(0,1,2)]
layer2 = [(3,2,1)]

qc = PauliFeatureMap(N, reps=3, paulis=['Z', 'ZZZ'], entanglement=[layer1, layer2], insert_barriers=True)
qc.decompose().draw('mpl', fold=-1)
```
List List List int

- If entanglement is List[List[List[List[int]]]] (This means that entanglement will change according to reps and paulis):
```
N = 5
layers = [[[(0,1), (2,3)], [(1,2), (3,4)]], [[(1,0), (3,2)] , [(2,1), (4,3)]]]

qc = PauliFeatureMap(N, reps=3, paulis=['Z', 'ZZ', 'XX'], entanglement=layers, insert_barriers=True)
qc.decompose().draw('mpl')
```
List List List List int

@nonhermitian please review this and let me know if I should make any further changes.

### DIFF
--- a/qiskit/circuit/library/data_preparation/pauli_feature_map.py
+++ b/qiskit/circuit/library/data_preparation/pauli_feature_map.py
@@ -347,12 +347,27 @@ class PauliFeatureMap(NLocal):
                             raise ValueError(f"Invalid value of entanglement:{e3}")
                         e2[ind] = tuple(map(int, e3))
 
-            chosen_entanglement = self.entanglement[i % num_i][j % num_j]
+            # Unlike Twolocal where we can specify entanglement blocks and rotation
+            # blocks separately, for PauliFeatureMap, all the paulis are specified
+            # as a single list (like ['Z', 'ZZ', 'YY']) and so if we just use
+            # self.entanglement[i % num_i][j % num_j] as the entanglement we will be
+            # choosing incorrect entanglement from the specified entanglement. So,
+            # here we subtract the number of single-qubit paulis from the j % num_j
+            # to pick correct entanglement from the specified List[List[List[List[int]]]]
+            count_single_qubit_paulis = 0
+            for pauli in self.paulis:
+                if len(pauli) == 1:
+                    count_single_qubit_paulis += 1
+
+            chosen_entanglement = self.entanglement[i % num_i][
+                (j % num_j) - count_single_qubit_paulis
+            ]
             return self._selective_entangler_map(num_block_qubits, chosen_entanglement)
 
         else:
-            # if the entanglement is not List[List[int]] or List[List[List[int]]]
-            # then we fall back on the original `get_entangler_map()` method from NLocal
+            # if the entanglement is not List[List[int]] or List[List[List[int]]] or
+            # List[List[List[List[int]]]] then we fall back on the original
+            # `get_entangler_map()` method from NLocal
             return super().get_entangler_map(rep_num, block_num, num_block_qubits)
 
 
--- a/qiskit/circuit/library/data_preparation/pauli_feature_map.py
+++ b/qiskit/circuit/library/data_preparation/pauli_feature_map.py
@@ -115,7 +115,12 @@ class PauliFeatureMap(NLocal):
         self,
         feature_dimension: Optional[int] = None,
         reps: int = 2,
-        entanglement: Union[str, Dict[int, List[Tuple[int]]], Dict[int, List[List[int]]]] = "full",
+        entanglement: Union[
+            str,
+            Dict[int, List[Tuple[int]]],
+            Dict[int, List[List[int]]],
+            Callable[[int], Union[str, Dict[int, List[Tuple[int]]], Dict[int, List[List[int]]]]],
+        ] = "full",
         alpha: float = 2.0,
         paulis: Optional[List[str]] = None,
         data_map_func: Optional[Callable[[np.ndarray], float]] = None,
@@ -132,7 +137,9 @@ class PauliFeatureMap(NLocal):
                 ``'linear'``, ``'reverse_linear'``, ``'circular'`` or ``'sca'``) or can be a
                 dictionary where the keys represent the number of qubits and the values are list
                 of integer-pairs specifying the indices of qubits that are entangled with one
-                another. For example: ``{1: [(0,), (2,)], 2: [(0,1), (2,0)]}``
+                another, for example: ``{1: [(0,), (2,)], 2: [(0,1), (2,0)]}`` or can be a
+                ``Callable[[int], Union[str | Dict[...]]]`` to return an entanglement specific for
+                a repetition
             alpha: The Pauli rotation factor, multiplicative to the pauli rotations
             paulis: A list of strings for to-be-used paulis. If None are provided, ``['Z', 'ZZ']``
                 will be used.
@@ -158,7 +165,6 @@ class PauliFeatureMap(NLocal):
         self._data_map_func = data_map_func or self_product
         self._paulis = paulis or ["Z", "ZZ"]
         self._alpha = alpha
-        self.entanglement = entanglement
 
     def _parameter_generator(
         self, rep: int, block: int, indices: List[int]
@@ -288,29 +294,34 @@ class PauliFeatureMap(NLocal):
         self, rep_num: int, block_num: int, num_block_qubits: int
     ) -> Sequence[Sequence[int]]:
 
+        # if entanglement is a Callable[[int], Union[str | Dict[...]]]
+        if callable(self._entanglement):
+            entanglement = self._entanglement(rep_num)
+        else:
+            entanglement = self._entanglement
+
         # entanglement is Dict[int, List[List[int]]]
-        if isinstance(self.entanglement, dict):
+        if isinstance(entanglement, dict):
             if all(
                 isinstance(e2, (int, np.int32, np.int64))
-                for key in self.entanglement.keys()
-                for en in self.entanglement[key]
+                for key in entanglement.keys()
+                for en in entanglement[key]
                 for e2 in en
             ):
-                for qb, ent in self.entanglement.items():
-                    for ind, en in enumerate(ent):
+                for qb, ent in entanglement.items():
+                    for en in ent:
                         if len(en) != qb:
                             raise ValueError(
                                 f"For num_qubits = {qb}, entanglement must be a "
                                 f"tuple of length {qb}. You specified {en}."
                             )
-                        self.entanglement[qb][ind] = tuple(map(int, en))
 
             # Check if the entanglement is specified for all the pauli blocks being used
             for pauli in self.paulis:
-                if len(pauli) not in self.entanglement.keys():
+                if len(pauli) not in entanglement.keys():
                     raise ValueError(f"No entanglement specified for {pauli} pauli.")
 
-            return self.entanglement[num_block_qubits]
+            return entanglement[num_block_qubits]
 
         else:
             # if the entanglement is not Dict[int, List[List[int]]] or
--- a/qiskit/circuit/library/data_preparation/pauli_feature_map.py
+++ b/qiskit/circuit/library/data_preparation/pauli_feature_map.py
@@ -128,8 +128,11 @@ class PauliFeatureMap(NLocal):
         Args:
             feature_dimension: Number of qubits in the circuit.
             reps: The number of repeated circuits.
-            entanglement: Specifies the entanglement structure. Refer to
-                :class:`~qiskit.circuit.library.NLocal` for detail.
+            entanglement: Specifies the entanglement structure. Can be a string (``'full'``,
+                ``'linear'``, ``'reverse_linear'``, ``'circular'`` or ``'sca'``) or can be a
+                dictionary where the keys represent the number of qubits and the values are list
+                of integer-pairs specifying the indices of qubits that are entangled with one
+                another. For example: ``{1: [(0,), (2,)], 2: [(0,1), (2,0)]}``
             alpha: The Pauli rotation factor, multiplicative to the pauli rotations
             paulis: A list of strings for to-be-used paulis. If None are provided, ``['Z', 'ZZ']``
                 will be used.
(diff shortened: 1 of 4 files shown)


Cryoris:
Thanks for the contribution @shravanpatel30! This is indeed a missing feature or wrong documentation. However, I'm not sure this is the right approach to solving it, as this selection might get very messy if you have Pauli connections with differing numbers of qubits. 

Is there a use-case for which you need specific entanglements? If yes, another approach could be to support a dictionary with `{num_qubits: entanglement}` pairs, e.g.
```python
entanglement = {
    1: [(0,), (2,)],
    2: [(0, 1), (1, 2)],
    3: [(0, 1, 2)]
}
zz_feature_map = ZZFeatureMap(.., paulis=["z", "zz", "zzz"], entanglement=entanglement)
```

shravanpatel30:
Hi @Cryoris, the data encoding using a feature map is an active area of research and so being able to control the entanglement in a feature map is really important. This gives more flexibility in creating custom feature maps. 

Currently, classes like `TwoLocal`, `NLocal`, and others that handle encoding allow entanglement to be specified using various formats like List[List[int]], List[List[List[int]]], or List[List[List[List[int]]]]. If we switch to using a dictionary for `ZZFeatureMap` and `PauliFeatureMap`, could that introduce inconsistencies or issues, given that these classes inherit from `NLocal`? However, if you believe that using a dictionary wouldn’t cause any issues, I would be happy to implement this approach.

Also, I might be misunderstanding your concern, but I’m unclear on what you meant by:
> as this selection might get very messy if you have Pauli connections with differing numbers of qubits.

The current implementation works even when the entanglement is `layer1 = [(0,), (2,), (1,2), (0,1,2)]`:
```
from qiskit.circuit.library import PauliFeatureMap, ZZFeatureMap

N = 3
layer1 = [(0,), (2,), (1,2), (0,1,2)]

qc = PauliFeatureMap(N, reps=2, paulis=['Z', 'ZZ', 'ZZZ'], entanglement=[layer1], insert_barriers=True)
qc.decompose().draw('mpl')
```
image

Cryoris:
> If we switch to using a dictionary for ZZFeatureMap and PauliFeatureMap, could that introduce inconsistencies or issues, given that these classes inherit from NLocal? However, if you believe that using a dictionary wouldn’t cause any issues, I would be happy to implement this approach.

It's good to think about this -- in this case I would say it is fine to have the subclass have a specific behavior, since it does have a special structure and the `list[list[list...` does not apply here.

> Also, I might be misunderstanding your concern, but I’m unclear on what you meant by:
>> as this selection might get very messy if you have Pauli connections with differing numbers of qubits.

> The current implementation works even when the entanglement is layer1 = [(0,), (2,), (1,2), (0,1,2)]

I mean that the entanglement list can (1) get very large if we have all layers in one list, and (2) it can get difficult to read as one could even accidentially mix the layers, such as `[(0,1), (1,2), (2,3), (1,2,3), (3,4), (0,1,2), (0,),...]`.

shravanpatel30:
Hi @Cryoris,

I have made some changes to the `PauliFeatureMap` so that it now supports a dictionary `{num_qubits: entanglement}`.

For `entanglement = {
    1: [(0,), (2,)],
    2: [(0, 1), (1, 2)],
    3: [(0, 1, 2)]
}`:
image

And if the  `entanglement = {
    1: [(0,), (2,)],
    3: [(0, 1, 2)]
}`, then `full` entanglement (which is default for `PauliFeatureMap`) is used for the 2-qubit pauli `ZZ`:
image

If you are fine with these changes then I'll go ahead and modify the typing hints and docstrings of `PauliFeatureMap`, `ZFeatureMap` and `ZZFeatureMap`. I'll also add few tests and releasenotes for these changes.

Cryoris:
Thanks for the update! Could you update the code such that there's no default `full` entanglement when the number of qubits is not specified? I think it would be better if the code is as explicit as possible, instead of adding entanglement that is not specified. Users can always add the full entanglement if it's needed 🙂 

And yes to tests and releasenotes 👍🏻  The section `features_circuits` should be the right section.

shravanpatel30:
Hi @Cryoris, I have added tests and releasenotes. I have also updated the typing hints and docstrings to reflect new changes. Let me know if you want me to make any further changes.

shravanpatel30:
Sorry about the failing test.

I replaced all the tab characters with spaces as suggested by the linting error, but it's still failing. When I check the compiled release notes on my PC, my recently added release note appears correctly, so I'm not sure where the problem lies.

releasenotes/notes/paulifeaturemap-takes-dictionary-as-entanglement-02037cb2d46e1c41.yaml
```diff
@@ -0,0 +1,21 @@
...
+                    1: [(0,), (2,)],
+                    2: [(0, 1), (1, 2)],
+                    3: [(0, 1, 2)],
+            }
+            params=[0, 1, 3.14]
+            qc = PauliFeatureMap(3, reps=2, paulis=['Z', 'ZZ', 'ZZZ'], entanglement=entanglement, insert_barriers=True)
+            qc.decompose().draw('mpl')
+
+
+
```
ElePT:
```suggestion
---
features_circuits:
  - |
    The entanglement argument in :class:`.PauliFeatureMap` and :class:`.ZZFeatureMap` will no longer support ``List[int]`` or ``List[List[int]]``. Instead,
    the entanglement structure in :class:`.PauliFeatureMap` and :class:`.ZZFeatureMap`  can now be specified as a dictionary where the keys represent  
    the number of qubits, and the values are lists of integer tuples that define which qubits are entangled with one another. This
    allows for more flexibility in constructing feature maps tailored to specific quantum algorithms. 
    Example usage::

      from qiskit.circuit.library import PauliFeatureMap
      entanglement = {
        1: [(0,), (2,)],
        2: [(0, 1), (1, 2)],
        3: [(0, 1, 2)],
      }
      params=[0, 1, 3.14]
      qc = PauliFeatureMap(3, reps=2, paulis=['Z', 'ZZ', 'ZZZ'], entanglement=entanglement, insert_barriers=True)
      qc.decompose().draw('mpl')
```

ElePT (commented):
Hi @shravanpatel30, I have left a suggestion that might fix the issue in the reno, I think the cause of the lint complaint is related to the indent in the code example. I have also added a few formatting suggestions for the referenced class names. You can directly click on "commit suggestion" to try it out.

shravanpatel30:
Hi @ElePT, thanks for the suggestion. Both code and the release notes now pass all the checks. Let me know if I need to make any more changes.

qiskit/circuit/library/data_preparation/pauli_feature_map.py
```diff
@@ -116,7 +115,7 @@ def __init__(
         self,
         feature_dimension: Optional[int] = None,
         reps: int = 2,
-        entanglement: Union[str, List[List[int]], Callable[[int], List[int]]] = "full",
+        entanglement: Union[str, Dict[int, List[Tuple[int]]], Dict[int, List[List[int]]]] = "full",
```
Cryoris: This should still accept a callable, which can return an entanglement specific for a repetition. So the signature should be: `Union[str, Dict[...], Callable[[int], Union[str | Dict[...]]`.

qiskit/circuit/library/data_preparation/pauli_feature_map.py
```diff
@@ -156,6 +158,7 @@ def __init__(
         self._data_map_func = data_map_func or self_product
         self._paulis = paulis or ["Z", "ZZ"]
         self._alpha = alpha
+        self.entanglement = entanglement
```
Cryoris: Is this necessary? This attribute should already be accessible via `self._entanglement`.

qiskit/circuit/library/data_preparation/pauli_feature_map.py
```diff
@@ -281,6 +284,40 @@ def cx_chain(circuit, inverse=False):
...
+                for e2 in en
+            ):
+                for qb, ent in self.entanglement.items():
+                    for ind, en in enumerate(ent):
+                        if len(en) != qb:
+                            raise ValueError(
+                                f"For num_qubits = {qb}, entanglement must be a "
+                                f"tuple of length {qb}. You specified {en}."
+                            )
+                        self.entanglement[qb][ind] = tuple(map(int, en))
```
Cryoris: I'm not sure we need this; users should already provide integers (as per the type hint) and I think the code works with both tuples and lists.

Cryoris (commented):
Looks almost good, thanks for the adjustments!

shravanpatel30:
@Cryoris, I have updated the typing hints and added some code to handle entanglement specified as a callable. I have also incorporated other changes in the code according to your comments.

qiskit/circuit/library/data_preparation/pauli_feature_map.py
```diff
@@ -116,7 +115,12 @@ def __init__(
         self,
         feature_dimension: Optional[int] = None,
         reps: int = 2,
-        entanglement: Union[str, List[List[int]], Callable[[int], List[int]]] = "full",
+        entanglement: Union[
+            str,
+            Dict[int, List[Tuple[int]]],
+            Dict[int, List[List[int]]],
+            Callable[[int], Union[str, Dict[int, List[Tuple[int]]], Dict[int, List[List[int]]]]],
```
Cryoris: We only need one of `List[List]` or `List[Tuple]`, they are equivalent 🙂  (same for the ZZ feature map)
```suggestion
            Dict[int, List[Tuple[int]]],
            Callable[[int], Union[str, Dict[int, List[Tuple[int]]]]],
```

releasenotes/notes/paulifeaturemap-takes-dictionary-as-entanglement-02037cb2d46e1c41.yaml
```diff
@@ -0,0 +1,20 @@
...
+      from qiskit.circuit.library import PauliFeatureMap
+      entanglement = {
+        1: [(0,), (2,)],
+        2: [(0, 1), (1, 2)],
+        3: [(0, 1, 2)],
+      }
+      qc = PauliFeatureMap(3, reps=2, paulis=['Z', 'ZZ', 'ZZZ'], entanglement=entanglement, insert_barriers=True)
+      qc.decompose().draw('mpl')
+
```
Cryoris: It was a bug that we could specify the entanglement as lists to begin with -- that didn't make sense 🙂 I reformulated the reno a bit to reflect this:
```suggestion
fixes:
  - |
    Fixed that the entanglement in :class:`.PauliFeatureMap` and :class:`.ZZFeatureMap`
    could be given as ``List[int]`` or ``List[List[int]]``, which was incompatible with the fact
    that entanglement blocks of different sizes are used. Instead, the entanglement can be 
    given as dictionary with ``{block_size: entanglement}`` pairs.
features_circuits:
  - |
    :class:`.PauliFeatureMap` and :class:`.ZZFeatureMap` now support specifying the 
    entanglement as a dictionary where the keys represent the number of qubits, and 
    the values are lists of integer tuples that define which qubits are entangled with one another. This
    allows for more flexibility in constructing feature maps tailored to specific quantum algorithms. 
    Example usage::

      from qiskit.circuit.library import PauliFeatureMap
      entanglement = {
        1: [(0,), (2,)],
        2: [(0, 1), (1, 2)],
        3: [(0, 1, 2)],
      }
      qc = PauliFeatureMap(3, reps=2, paulis=['Z', 'ZZ', 'ZZZ'], entanglement=entanglement, insert_barriers=True)
      qc.decompose().draw('mpl')

```

shravanpatel30:
@Cryoris, I have made all the requested changes.

Cryoris (approved):
Thanks for the fix & new feature!

PR #12769, opened 2024-07-12
Setting noise_info to false in GenericBackendV2 creates issues later in the transpiler pipeline

sbrandhsn:
### Summary

Fixes bugs introduced in #12747

### Details and comments

I added some tests to check whether transpiling with noise-less and/or pulse-channel-less backends still works and noticed test failures. I fixed these bugs and added the corresponding test cases. I don't think this warrants a new release note but let me know if you disagree. :-)

### DIFF
--- a/qiskit/providers/fake_provider/generic_backend_v2.py
+++ b/qiskit/providers/fake_provider/generic_backend_v2.py
@@ -388,11 +388,6 @@ class GenericBackendV2(BackendV2):
                 defaults = self._generate_calibration_defaults()
                 calibration_inst_map = defaults.instruction_schedule_map
 
-        is_fully_connected = (
-            self._coupling_map.graph.num_edges()
-            == self._coupling_map.graph.num_nodes() * (self._coupling_map.graph.num_nodes() - 1)
-        )
-
         # Iterate over gates, generate noise params from defaults,
         # and add instructions, noise and calibrations to target.
         for name in self._basis_gates:
@@ -411,12 +406,9 @@ class GenericBackendV2(BackendV2):
                 noise_params = self._get_noise_defaults(name, gate.num_qubits)
                 self._add_noisy_instruction_to_target(gate, noise_params, calibration_inst_map)
             else:
-                if is_fully_connected:
-                    self._target.add_instruction(gate, properties=None, name=name)
-                else:
-                    qarg_set = self._coupling_map if gate.num_qubits > 1 else range(self.num_qubits)
-                    props = {(qarg,) if isinstance(qarg, int) else qarg: None for qarg in qarg_set}
-                    self._target.add_instruction(gate, properties=props, name=name)
+               qarg_set = self._coupling_map if gate.num_qubits > 1 else range(self.num_qubits)
+               props = {(qarg,) if isinstance(qarg, int) else qarg: None for qarg in qarg_set}
+               self._target.add_instruction(gate, properties=props, name=name)
 
         if self._control_flow:
             self._target.add_instruction(IfElseOp, name="if_else")
--- a/qiskit/providers/fake_provider/generic_backend_v2.py
+++ b/qiskit/providers/fake_provider/generic_backend_v2.py
@@ -406,9 +406,9 @@ class GenericBackendV2(BackendV2):
                 noise_params = self._get_noise_defaults(name, gate.num_qubits)
                 self._add_noisy_instruction_to_target(gate, noise_params, calibration_inst_map)
             else:
-               qarg_set = self._coupling_map if gate.num_qubits > 1 else range(self.num_qubits)
-               props = {(qarg,) if isinstance(qarg, int) else qarg: None for qarg in qarg_set}
-               self._target.add_instruction(gate, properties=props, name=name)
+                qarg_set = self._coupling_map if gate.num_qubits > 1 else range(self.num_qubits)
+                props = {(qarg,) if isinstance(qarg, int) else qarg: None for qarg in qarg_set}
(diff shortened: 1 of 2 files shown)


ElePT (commented):
Thanks @sbrandhsn for looking into this. I am guessing that the root of the issue was related to the instructions being global (i.e. `properties=None`), but what was exactly failing, was it the pm construction or specific passes? Because as far as I understand it global instructions should be allowed too in the pipeline (so maybe the bug is down the line, not here).

sbrandhsn:
Sure, they might be another issue down the line. However, if the user specifies a backend with a coupling map but without noise info, I guess we should not return a target with global instructions. :-) This is the current behavior of the code I pushed last week.

ElePT:
Both options should be valid (target with global instructions and with local instructions but `properties=None`) as long as they are properly documented, I guess I was asking whether this was just unexpected behavior for you or whether you saw any failure caused by global instructions that motivated this PR. That being said, I am not against the "unexpected behavior" argument.

qiskit/providers/fake_provider/generic_backend_v2.py
```diff
@@ -406,7 +411,12 @@ def _build_generic_target(self):
                 noise_params = self._get_noise_defaults(name, gate.num_qubits)
                 self._add_noisy_instruction_to_target(gate, noise_params, calibration_inst_map)
             else:
-                self._target.add_instruction(gate, properties=None, name=name)
+                if is_fully_connected:
+                    self._target.add_instruction(gate, properties=None, name=name)
+                else:
+                    qarg_set = self._coupling_map if gate.num_qubits > 1 else range(self.num_qubits)
+                    props = {(qarg,) if isinstance(qarg, int) else qarg: None for qarg in qarg_set}
+                    self._target.add_instruction(gate, properties=props, name=name)
```
ElePT: hmmm what makes the fully connected cmap special in this case? Is this purely an efficiency matter?

sbrandhsn: Yep, it is much more efficient to set `properties=None' here in that case, otherwise we create a dictionary with entries for each edge of the fully connected coupling map that all point to None...

sbrandhsn:
> Both options should be valid (target with global instructions and with local instructions but `properties=None`) as long as they are properly documented, I guess I was asking whether this was just unexpected behavior for you or whether you saw any failure caused by global instructions that motivated this PR. That being said, I am not against the "unexpected behavior" argument.

Sorry for not being clear in my first response. :-/ I was getting an intermittent error that I can not replicate easily right now. I think the gist was that `build_coupling_map` was called by something in the transpiler which returned `None` because `self.qargs` is `None`. Then some code expected `None` to be a valid coupling map and called `size` on it. Interestingly this did not occur for all instance of `CouplingMap.from_heavy_hex` that I was running. Maybe something in main fixes that behaviour right now.

qiskit/providers/fake_provider/generic_backend_v2.py
ElePT: I see, but I am not sure I like this approach. This would make the behavior differ "arbitrarily" depending on the connectivity of the coupling map used, right? Some times the instructions would be global, some times they would be local with no properties... I think that the behavior should be consistent independently of the connectivity (plus, if having properties=None does introduce issues in the pipeline, then the issues aren't being fixed for the case of fully-connected cmaps, right?).

sbrandhsn: I see - the issue here is that the memory overhead is quite large when representing an all-to-all connected by local instructions without properties. It can be up to ~2 GB for large backends whereas a backend with global instructions and no properties needs a couple of MB. Granted, one should likely not create a large all-to-all connected backend but if a user does, he will be surprised by that memory overhead. 

However, I agree that a discrete change based on that fully-connected flag is similarly confusing - a backend missing only one edge to be all-to-all connected would exhibit a different memory behaviour than a fully-connected backend. Should I remove the optimizations based on `is_fully_connected`? I think the pre-PR behaviour of noiseless `BackendV2` is definitely incorrect...

ElePT: I think that it's safer to remove the optimizations for `is_fully_connected`. It's the default behavior of the class and I can see how users would get confused if specifying a custom cmap suddenly affects the gate properties.

qiskit/providers/fake_provider/generic_backend_v2.py
```diff
@@ -388,6 +388,11 @@ def _build_generic_target(self):
                 defaults = self._generate_calibration_defaults()
                 calibration_inst_map = defaults.instruction_schedule_map
 
+        is_fully_connected = (
+            self._coupling_map.graph.num_edges()
+            == self._coupling_map.graph.num_nodes() * (self._coupling_map.graph.num_nodes() - 1)
+        )
+
```
ElePT: If we agree on the not separating fully connected cmaps path, I'd suggest:
```suggestion
```

qiskit/providers/fake_provider/generic_backend_v2.py
ElePT:
```suggestion
               qarg_set = self._coupling_map if gate.num_qubits > 1 else range(self.num_qubits)
               props = {(qarg,) if isinstance(qarg, int) else qarg: None for qarg in qarg_set}
               self._target.add_instruction(gate, properties=props, name=name)
```

ElePT (commented):
@sbrandhsn I have added some suggestions not to separate out the behavior of the fully connected cmap. Let me know if you agree with the approach.

ElePT (approved):
LGTM

PR #12770, opened 2024-07-12
Adjust graphviz options for better layout with plot_gate_map()

mtreinish:
### Summary

This commit fixes an issue with the visualizations of some backends/coupling maps. The default neato settings worked well in most cases but for some graphs it was not generating a good layout. This tweaks the settings a bit so that layouts work with more classes of graphs.

### Details and comments


1ucian0:
`ibm_fez`
image

mtreinish:
Based on that image I'm wondering if we want to scale up the fontsize a bit?

1ucian0:
`ibm_quebec`

image

mtreinish:
That second example looks like we should condition:

https://github.com/Qiskit/qiskit/pull/12770/files#diff-ff70fb816cde3f453d870891f47ede8245855a9e156228201a53fd613ad00659R1096-R1100

on `if not qubit_coordinates` and only use these graph attributes if we're doing an algorithmic layout.

1ucian0:
In the last change, I changed to prism and I got the following results:

`ibm_fez`
image

`ibm_cusconazca`
image

`ibm_torino`
image

`ibm_kyiv`
image

If you are happy when them, I will update the image references and merge.

mtreinish:
> If you are happy when them, I will update the image references and merge.

Those look fine to me, the `ibm_torino` one is maybe a bit less pretty than `ibm_fez` and `ibm_cusconazca` but it looks fine. So let's just go with this.

ElePT:
It looks like the font size still needs some tweaking:

1_qubit_gate_map
5_qubit_gate_map
16_qubit_gate_map

1ucian0:
Here, another round of changes:
image
image
image
image
image
image
image
image
image

1ucian0:
image
image
image
image
image
image
image
image
image

Cryoris (approved):
Thanks for the final tweaks @1ucian0 !

PR #12140, opened 2024-04-04
Handle huge integers in OpenQASM 2 expression evaluator

jakelishman:
### Summary

This modifies the expression evaluator to directly parse the backing string data of an integer token in a floating-point context, which lets us handle numbers that would otherwise overflow a `usize`.  It's possible for this to be totally valid, if, for example, the integer is a multiple of some very large power of two that doesn't overflow a double-precision float.

We already needed to immediately cast the integer to a float, so this is just a more accurate way of doing the evaluation, and doesn't affect when we use integers in other contexts.

### Details and comments

Fix #11561

PR #12774, opened 2024-07-15
Fix parsing of huge OpenQASM 2 conditionals

### LINKED ISSUE #12773: Integer overflow when creating circuits from QASM files with large ints
### Environment

- **Qiskit version**:  1.1.1
- **Python version**: 
- **Operating system**: 

### What is happening?

As reported https://github.com/Qiskit/benchpress/issues/17, qasm imports lead to int overflows when qasm files contain large ints.  E.g. this example will fail

```
OPENQASM 2.0;
include "qelib1.inc";
qreg q0[301];
creg c0[301];
h q0[300];
measure q0[300] -> c0[300];
if(c0==2037035976334486086268445688409378161051468393665936250636140449354381299763336706183397376) h q0[0];
```

with 

```
thread '<unnamed>' panicked at crates/qasm2/src/lex.rs:279:42:
called `Result::unwrap()` on an `Err` value: ParseIntError { kind: PosOverflow }
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

```
PanicException: called `Result::unwrap()` on an `Err` value: ParseIntError { kind: PosOverflow }
```

Run the above QASM file with `QuantumCircuit.from_qasm_file`

### What should happen?

It should be imported

### Any suggestions?

The int is larger than int64, which I assume is the root cause.

jakelishman:
### Summary

We fixed handling of giant integers in gate expression positions gh-12140, and this commit fixes the handling in conditionals. Unfortunately, this means pulling in big-int handling properly; the integers simply _are_ bigints, and we're not immediately converting them into something else.

The need to support this may influence how the Rust-space data models of `QuantumCircuit` evolve.

### Details and comments

Fix #12773


crates/qasm2/Cargo.toml
```diff
@@ -10,6 +10,7 @@ name = "qiskit_qasm2"
 doctest = false
 
 [dependencies]
+num-bigint = "0.4"
```
mtreinish: Should we move this to the workspace, because we use big int in accelerate too.

mtreinish (commented):
This LGTM it's a pretty straightforward change, and you pre-empted my question about using `u128` as a short term bandaid with your unit tests. :) Just one inline comment about the cargo config, but that's small.

crates/qasm2/Cargo.toml
jakelishman: Sure, done in 27e1689c6a.

mtreinish (approved):

jakelishman:

PR #12775, opened 2024-07-16
Fix pass-manager drawing tests for pydot 3.0

jakelishman:
### Summary

Two changes in pydot 3 changed the reference output:

1. previously, attributes were sorted before being output.  Pydot 3 stores them in declaration order.  Here, we sort our attributes to maintain the closest behaviour between the two versions.

2. Text fields (like `label`) that contain special characters now have their values enclosed in quote marks.  This is a difference between the two versions, and we update the reference files to the new version since that's what we'll be using in CI.

Qiskit is still compatible with pydot 2 in general usage, it's just our tests that are a little more tied to the current version.


jakelishman:

mtreinish (dismissed):
This looks fine to me assuming it passes CI, it's just an argument reordering and escaping change in the expected output. I guess the open question is did you want to throw a test version constraint on > 3.0.0 and leave the optional requirement at whatever the current minimum is? It doesn't matter too much so feel free to just enqueue this for merging if you're fine leaving the minimum for testing unchanged.

jakelishman:
That sounds like a good idea - done in 7158eef.

mtreinish (approved):

PR #12776, opened 2024-07-16
Rewrite OpenQASM 3 exporter symbol table

### LINKED ISSUE #7136: Some internal polishing is needed in the QASM3 exporter
This is a https://github.com/Qiskit/qiskit-terra/pull/6565 follow up

This particular issue tracks:
- [ ] The `GlobalNamespace._data` is horrendous and needs to be split. More [here](https://github.com/Qiskit/qiskit-terra/pull/6565#discussion_r669145097).  
- [ ] Merge `GlobalNamespace` and `Exporter`.
- [ ] Better Namespace validation and renaming, see [here](https://github.com/Qiskit/qiskit-terra/pull/6565#discussion_r714732791) and [here](https://github.com/Qiskit/qiskit-terra/pull/6565#discussion_r714733921).

### LINKED ISSUE #7335: QASM 2 and 3 exporters do not handle parameterised gates correctly
### Environment

- **Qiskit Terra version**: 0.19.dev @ 2c402dd5c
- **Python version**: any
- **Operating system**: any

### What is happening?

The QASM3 exporter (and the QASM2 exporter) does not handle arbitrary parameterised gates correctly.  It emits a gate definition which includes parameters in the signature, but hardcodes them in the body to the first instance of the gate it found.  This is because there's currently no way in Terra to get from a gate with its parameters assigned back to the fully parameterised version.

```python
In [1]: import qiskit.qasm3
   ...: qc = qiskit.QuantumCircuit(2, 2)
   ...: qc.rzx(0.2, 0, 1)
   ...: qc.rzx(0.3, 0, 1)
   ...:
   ...: print(qiskit.qasm3.dumps(qc))
OPENQASM 3;
include "stdgates.inc";
gate rzx(_gate_p_0) _gate_q_0, _gate_q_1 {
  h _gate_q_1;
  cx _gate_q_0, _gate_q_1;
  rz(0.2) _gate_q_1;
  cx _gate_q_0, _gate_q_1;
  h _gate_q_1;
}
bit[2] c;
qubit[2] _all_qubits;
let q = _all_qubits[0:1];
rzx(0.2) q[0], q[1];
rzx(0.3) q[0], q[1];
```

The two `rzx` calls have different parameters, but the parameter is hardcoded to 0.2 in the actual definition.

### What should happen?

A full fix of this issue will have the `rz` line in the `rzx` gate definition use the gate parameter.

### Any suggestions?

As a temporary measure before the 0.19 release, we can hack the exporter to always output a new declaration for each instance of every gate that is not in the include files.  This is pretty awful, and the output QASM won't correctly represent the structure of the program with re-used gates, but it will at least produce the correct behaviour.

### LINKED ISSUE #9428: QASM 3 exporter makes multiple definitions for same gate
### Environment

- **Qiskit Terra version**: lastest
- **Python version**: 
- **Operating system**: 

### What is happening?

```python

qc = QuantumCircuit(2)
qc.rzz(1,0,1)
qc.rzz(1,0,1)
qc.rzz(1,0,1)
```
gives a unique QASM3 gate definition for each gate above even though they are identical.

```text
OPENQASM 3;
include "stdgates.inc";
gate rzz_4592196368(_gate_p_0) _gate_q_0, _gate_q_1 {
  cx _gate_q_0, _gate_q_1;
  rz(1) _gate_q_1;
  cx _gate_q_0, _gate_q_1;
}
gate rzz_4592194880(_gate_p_0) _gate_q_0, _gate_q_1 {
  cx _gate_q_0, _gate_q_1;
  rz(1) _gate_q_1;
  cx _gate_q_0, _gate_q_1;
}
gate rzz_4592190992(_gate_p_0) _gate_q_0, _gate_q_1 {
  cx _gate_q_0, _gate_q_1;
  rz(1) _gate_q_1;
  cx _gate_q_0, _gate_q_1;
}
qubit[2] _all_qubits;
let q = _all_qubits[0:1];
rzz_4592196368(1) q[0], q[1];
rzz_4592194880(1) q[0], q[1];
rzz_4592190992(1) q[0], q[1];
```

Run above

### What should happen?

Ideally, every unique gate would have a single definition.

### Any suggestions?

_No response_

### LINKED ISSUE #11560: Basis gate set with RXX gate converted with QASM3 and run with Aer simulator results in 'unknown instruction: rxx_139974866104288'
### Environment

- **Qiskit Terra version**: 0.45.1
- **Python version**: 3.10.1

### What is happening?

Creating a circuit that results in an RXX gate after being transpiled with a basis set of rx, ry, and rxx and then converting with QASM3 and running on an Aer simulator results in the following error:

```
  Cell In[79], line 18
    counts = simulator.run(qc, shots=100).result().get_counts(qc)
  File /opt/conda/lib/python3.10/site-packages/qiskit_aer/jobs/utils.py:42 in _wrapper
    return func(self, *args, **kwargs)
  File /opt/conda/lib/python3.10/site-packages/qiskit_aer/jobs/aerjob.py:114 in result
    return self._future.result(timeout=timeout)
  File /opt/conda/lib/python3.10/concurrent/futures/_base.py:458 in result
    return self.__get_result()
  File /opt/conda/lib/python3.10/concurrent/futures/_base.py:403 in __get_result
    raise self._exception
  File /opt/conda/lib/python3.10/concurrent/futures/thread.py:58 in run
    result = self.fn(*self.args, **self.kwargs)
  File /opt/conda/lib/python3.10/site-packages/qiskit_aer/backends/aerbackend.py:435 in _execute_circuits_job
    aer_circuits, idx_maps = assemble_circuits(circuits)
  File /opt/conda/lib/python3.10/site-packages/qiskit_aer/backends/aer_compiler.py:675 in assemble_circuits
    aer_circuits, idx_maps = zip(*[assemble_circuit(circuit) for circuit in circuits])
  File /opt/conda/lib/python3.10/site-packages/qiskit_aer/backends/aer_compiler.py:675 in <listcomp>
    aer_circuits, idx_maps = zip(*[assemble_circuit(circuit) for circuit in circuits])
  File /opt/conda/lib/python3.10/site-packages/qiskit_aer/backends/aer_compiler.py:530 in assemble_circuit
    num_of_aer_ops += _assemble_op(
  File /opt/conda/lib/python3.10/site-packages/qiskit_aer/backends/aer_compiler.py:646 in _assemble_op
    raise AerError(f"unknown instruction: {name}")
AerError: 'unknown instruction: rxx_139974866104288'
```

Having the basis gate set, the QASM3 conversion, and the aer simulator are all required to reproduce the bug in this manner.

Run the following:

```python
import qiskit
from qiskit import QuantumCircuit, ClassicalRegister, QuantumRegister
from qiskit.circuit.library.standard_gates import *
qr = QuantumRegister(2, name='qr')
cr = ClassicalRegister(2, name='cr')
qc = QuantumCircuit(qr, cr, name='qc')
qc.append(RXXGate(3.1067098743999053), qargs=[qr[0], qr[1]], cargs=[])
qc.measure(qr, cr)

from qiskit import transpile
qc = transpile(qc, basis_gates=['rx', 'ry', 'rxx'])

from qiskit import qasm3
qc = qasm3.loads(qasm3.dumps(qc))

simulator = AerSimulator()
counts = simulator.run(qc, shots=100).result().get_counts(qc)
```

### What should happen?

The circuit should  be able to run without error.

### Any suggestions?

_No response_

### LINKED ISSUE #11761: Qiskit’s OpenQASM 3 exporter: not correctly escaping the names of custom gates it’s defining
### Environment

- **Qiskit version**: 
- **Python version**: 
- **Operating system**: 

### What is happening?

I submit a job to the IBM Platform, a max cut using QAOA, for very simple circuits (less than 10 nodes, few edges). This will produce a parameterized circuit.
The OpenQASM 3 string created cannot be drawn - there's something wrong with the string, likely not correctly escaping the names of custom gates it's defining.

Jake Lishman has requested I submit this bug report.

Here's an example of a string it produces:

"""
OPENQASM 3;
include "stdgates.inc";
input float[64] _β_0_;
input float[64] _γ_0_;
gate exp(it IZ)(_γ_0_) _gate_q_0, _gate_q_1 {
  rz(5.0*_γ_0_) _gate_q_0;
}
gate rzz_140434630696240(_γ_0_) _gate_q_0, _gate_q_1 {
  cx _gate_q_0, _gate_q_1;
  rz(1.0*_γ_0_) _gate_q_1;
  cx _gate_q_0, _gate_q_1;
}
gate exp(it ZZ)(_γ_0_) _gate_q_0, _gate_q_1 {
  rzz_140434630696240(1.0*_γ_0_) _gate_q_0, _gate_q_1;
}
gate PauliEvolution(_γ_0_) _gate_q_0, _gate_q_1 {
  exp(it IZ)(_γ_0_) _gate_q_0, _gate_q_1;
  exp(it ZZ)(_γ_0_) _gate_q_0, _gate_q_1;
}
bit[2] meas;
qubit[2] q;
h q[0];
h q[1];
PauliEvolution(_γ_0_) q[0], q[1];
PauliEvolution(_β_0_) q[0], q[1];
barrier q[0], q[1];
meas[0] = measure q[0];
meas[1] = measure q[1];
"""

This code will not run for you as it depends on some custom functions and config files.

However, the problem should be reproducible from the concept - we are creating a really straightforward QAOA / maxcut problem.

```
from docplex.mp.model import Model
from qiskit_optimization.algorithms import MinimumEigenOptimizer
from qiskit_optimization.translators import from_docplex_mp
from qiskit_ibm_runtime import QiskitRuntimeService, Session, Sampler, Options
from qiskit_algorithms import QAOA
from qiskit_algorithms.optimizers import COBYLA, SPSA
from qiskit_aer.noise import NoiseModel

from quantum_computing_for_energy_sector.settings import load_config
from quantum_computing_for_energy_sector.utils import root_path, read_args, show_graph
from quantum_computing_for_energy_sector.logging_config import logger

token = load_config(root_path(".venv", "IBMQ-API-token.yml"))
args = read_args()

device_config = load_config(args.device_conf_file)
logger.info(f"Device config: {device_config}")

service = QiskitRuntimeService(
    channel = "ibm_quantum",
    token = token["IBM_Quantum_API_key"]
)

if device_config["device"] == "Simulator":
    backend = service.backend(device_config["simulator"])

backend = service.least_busy(operational=True, simulator=False)
graph_config = load_config(args.graph_conf_file)

number_of_nodes = len(graph_config["nodes"])
edges = [(edge['first_node'], edge['second_node'], edge['weight']) for edge in graph_config["edges"]]

model = Model()

x = model.binary_var_list(number_of_nodes)
model.maximize(model.sum(w * x[i] * (1 - x[j]) + w * (1 - x[i]) * x[j] for i, j, w in edges))

model.add(x[0] == 1)

problem = from_docplex_mp(model)

optimizer = COBYLA(maxiter = device_config["maxiter"])

session = Session(backend = backend)
options = Options()

sampler = Sampler(session = session, options = options)

qaoa = QAOA(sampler=sampler, optimizer=optimizer)
algorithm = MinimumEigenOptimizer(qaoa)
result = algorithm.solve(problem)

```

### What should happen?

The code above will run - that is not the problem.

The problem is the OpenQASM 3 string that qiskit generates on the IBM Quantum Platform - this string has a bug in it. A circuit cannot be drawn from that string.

### Any suggestions?

_No response_

### LINKED ISSUE #11954: qasm3 returns wrong results with pauli evolution gate
When trying to turn this circuit into a qasm3 : 

```
from qiskit.circuit import QuantumCircuit, ParameterVector
from qiskit.qasm3 import dumps, loads, dump, load

t = ParameterVector("t", 3)
X, Y, Z = SparsePauliOp("X"), SparsePauliOp("Y"), SparsePauliOp("Z")
op1, op2, op3 = (X^X), (Y^Y), (Z^Z)
evo1 = PauliEvolutionGate(op1, time=t[0])
evo2 = PauliEvolutionGate(op2, time=t[1])
evo3 = PauliEvolutionGate(op3, time=t[2])
 
dec0 = QuantumCircuit(2)
dec0.append(evo1, range(2))
dec0.append(evo2, range(2))
dec0.append(evo3, range(2))
dec0.count_ops() #returns OrderedDict([('PauliEvolution', 3)])
```

after dumping into a qasm3, we would expect the qasm to have 3 different gates for the XX, YY, ZZ but it only looks into the first Pauli evolution gate and repeats it 3 times : 

```
q0 = dumps(dec0)
print(q0)
#returns 
OPENQASM 3.0;
include "stdgates.inc";
input float[64] _t_0_;
input float[64] _t_1_;
input float[64] _t_2_;
gate rxx_4948547280(_t_0_) _gate_q_0, _gate_q_1 {
  h _gate_q_0;
  h _gate_q_1;
  cx _gate_q_0, _gate_q_1;
  rz(2.0*_t_0_) _gate_q_1;
  cx _gate_q_0, _gate_q_1;
  h _gate_q_1;
  h _gate_q_0;
}
gate PauliEvolution(_t_0_) _gate_q_0, _gate_q_1 {
  rxx_4948547280(2.0*_t_0_) _gate_q_0, _gate_q_1;
}
qubit[2] q;
PauliEvolution(_t_0_) q[0], q[1];
PauliEvolution(_t_1_) q[0], q[1];
PauliEvolution(_t_2_) q[0], q[1];
```

jakelishman:
### Summary

This rewrites the symbol handling of the OpenQASM 3 exporter to decouple object identities from the necessary object identifiers.  As part of this, we use the same trick of standard-gate reparametrisation to produce gate-definition sources for Qiskit built-ins, which fixes many cases of bad parametrisation of gates like `rzx`.

This kind of rewrite was necessary to fix the now-bad assumption within the OQ3 exporter that "gate identity" is always static within a circuit. Since standard gate `Gate` instances are now only generated on demand, there is no guarantee of stability of them.  The fix to the definition source for these makes them independent of object identity. User-defined gates can still use the identity, as these are still guaranteed static.

This commit fixes almost all of the "bad parametrisation" tests in the test suite.  There are several other changes in the test suite necessary:

* since the uniqueness of the identifier is now independent of how the lookup of a Qiskit object works, there is no need to include the highly non-deterministic `id` in the generated symbols for user gates. Several tests changed to use the new, simple count-based unique names.

* the escaping and uniqueness rules now apply uniformly to all gate definitions, fixing several bad test cases that previously were testing invalid OpenQASM 3.

* the escaping rules changed slightly for naming collisions with keywords, making them slightly more consistent with how other renaming rules worked.

### Details and comments

This unblocks #12730 by fixing the broken assumption about object identity for standard gates.

* Fix #7136 (not to say it's perfect now, but this addresses all the problems tracked by that issue).
* Fix #7335
* Fix #9428
* Fix #11954
* Fix #11560
* Fix #11761

* Supersede #12481
    This fixes the underlying bug that PR 12481 was attempting to address, though reducing the use of `instruction.operation` as that PR does could still be a benefit to performance.

### DIFF
--- a/qiskit/qasm3/exporter.py
+++ b/qiskit/qasm3/exporter.py
@@ -439,8 +439,8 @@ class SymbolTable:
         else:
             canonical = _gate_canonical_form(gate)
             self.gates[name] = GateInfo(canonical, None)
-            if (standard_gate := getattr(canonical, "_standard_gate")) is not None:
-                self.standard_gate_idents[standard_gate] = ident
+            if canonical._standard_gate is not None:
+                self.standard_gate_idents[canonical._standard_gate] = ident
             else:
                 self.user_gate_idents[id(canonical)] = ident
         return ident
@@ -463,8 +463,8 @@ class SymbolTable:
         # Add the gate object with a magic lookup keep to the objects dictionary so we can retrieve
         # it later.  Standard gates are not guaranteed to have stable IDs (they're preferentially
         # not even created in Python space), but user gates are.
-        if (standard_gate := getattr(source, "_standard_gate")) is not None:
-            self.standard_gate_idents[standard_gate] = ident
+        if source._standard_gate is not None:
+            self.standard_gate_idents[source._standard_gate] = ident
         else:
             self.user_gate_idents[id(source)] = ident
         return ident
@@ -477,8 +477,8 @@ class SymbolTable:
             our_defn.canonical is None or our_defn.canonical == canonical
         ):
             return ast.Identifier(gate.name)
-        if (standard_gate := getattr(canonical, "_standard_gate")) is not None:
-            if (our_ident := self.standard_gate_idents.get(standard_gate)) is None:
+        if canonical._standard_gate is not None:
+            if (our_ident := self.standard_gate_idents.get(canonical._standard_gate)) is None:
                 return None
             return our_ident if self.gates[our_ident.string].canonical == canonical else None
         # No need to check equality if we're looking up by `id`; we must have the same object.
--- a/qiskit/qasm3/exporter.py
+++ b/qiskit/qasm3/exporter.py
@@ -246,12 +246,17 @@ _BUILTIN_GATES = {
 
 
 @dataclasses.dataclass
-class GateDefinition:
+class GateInfo:
     """Symbol-table information on a gate."""
 
-    source: Gate | None
-    """The definition source of the gate.  This can be ``None`` if the gate was an overridden "basis
-    gate" for this export."""
+    canonical: Gate | None
+    """The canonical object for the gate.  This is a Qiskit object that is not necessarily equal to
+    any usage, but is the canonical form in terms of its parameter usage, such as a standard-library
+    gate being defined in terms of the `_FIXED_PARAMETERS` objects.  A call-site gate whose
+    canonical form equals this can use the corresponding symbol as the callee.
+
+    This can be ``None`` if the gate was an overridden "basis gate" for this export, so no canonical
+    form is known."""
     node: ast.QuantumGateDefinition | None
     """An AST node containing the gate definition.  This can be ``None`` if the gate came from an
     included file, or is an overridden "basis gate" of the export."""
@@ -261,7 +266,7 @@ class SymbolTable:
     """Track Qiskit objects and the OQ3 identifiers used to refer to them."""
 
     def __init__(self):
-        self.gates: collections.OrderedDict[str, GateDefinition | None] = {}
+        self.gates: collections.OrderedDict[str, GateInfo | None] = {}
         """Mapping of the symbol name to the "definition source" of the gate, which provides its
         signature and decomposition.  The definition source can be `None` if the user set the gate
         as a custom "basis gate".
@@ -430,14 +435,14 @@ class SymbolTable:
         name = self.escaped_declarable_name(name, allow_rename=False, unique=False)
         ident = ast.Identifier(name)
         if gate is None:
-            self.gates[name] = GateDefinition(None, None)
+            self.gates[name] = GateInfo(None, None)
         else:
-            source = _gate_definition_source(gate)
-            self.gates[name] = GateDefinition(source, None)
-            if (standard_gate := getattr(source, "_standard_gate")) is not None:
+            canonical = _gate_canonical_form(gate)
+            self.gates[name] = GateInfo(canonical, None)
+            if (standard_gate := getattr(canonical, "_standard_gate")) is not None:
                 self.standard_gate_idents[standard_gate] = ident
             else:
-                self.user_gate_idents[id(source)] = ident
+                self.user_gate_idents[id(canonical)] = ident
         return ident
 
     def register_gate(
@@ -452,7 +457,7 @@ class SymbolTable:
         full AST definition."""
         name = self.escaped_declarable_name(name, allow_rename=True, unique=False)
         ident = ast.Identifier(name)
-        self.gates[name] = GateDefinition(
+        self.gates[name] = GateInfo(
             source, ast.QuantumGateDefinition(ident, tuple(params), tuple(qubits), body)
         )
         # Add the gate object with a magic lookup keep to the objects dictionary so we can retrieve
@@ -466,27 +471,28 @@ class SymbolTable:
 
     def get_gate(self, gate: Gate) -> ast.Identifier | None:
         """Lookup the identifier for a given `Gate`, if it exists."""
-        source = _gate_definition_source(gate)
-        # `our_defn.source` means a basis gate that we should assume is always valid.
+        canonical = _gate_canonical_form(gate)
+        # `our_defn.canonical is None` means a basis gate that we should assume is always valid.
         if (our_defn := self.gates.get(gate.name)) is not None and (
-            our_defn.source is None or our_defn.source == source
+            our_defn.canonical is None or our_defn.canonical == canonical
         ):
             return ast.Identifier(gate.name)
-        if (standard_gate := getattr(source, "_standard_gate")) is not None:
+        if (standard_gate := getattr(canonical, "_standard_gate")) is not None:
             if (our_ident := self.standard_gate_idents.get(standard_gate)) is None:
                 return None
-            return our_ident if self.gates[our_ident.string].source == source else None
+            return our_ident if self.gates[our_ident.string].canonical == canonical else None
         # No need to check equality if we're looking up by `id`; we must have the same object.
-        return self.user_gate_idents.get(id(source))
+        return self.user_gate_idents.get(id(canonical))
 
 
-def _gate_definition_source(gate: Gate) -> Gate:
-    """Get the "definition source" of a gate.
+def _gate_canonical_form(gate: Gate) -> Gate:
+    """Get the canonical form of a gate.
 
     This is the gate object that should be used to provide the OpenQASM 3 definition of a gate (but
     not the call site; that's the input object).  This lets us return a re-parametrised gate in
     terms of general parameters, in cases where we can be sure that that is valid.  This is
-    currently only  Qiskit standard gates.
+    currently only Qiskit standard gates.  This lets multiple call-site gates match the same symbol,
+    in the case of parametric gates.
 
     The definition source provides the number of qubits, the parameter signature and the body of the
     `gate` statement.  It does not provide the name of the symbol being defined."""
@@ -693,8 +699,8 @@ class QASM3Builder:
             return self.symbols.register_gate(gate.name, gate, (), (control, target), body)
         if gate.definition is None:
             raise QASM3ExporterError(f"failed to export gate '{gate.name}' that has no definition")
-        source = _gate_definition_source(gate)
-        with self.new_context(source.definition):
+        canonical = _gate_canonical_form(gate)
+        with self.new_context(canonical.definition):
             defn = self.scope.circuit
             # If `defn.num_parameters == 0` but `gate.params` is non-empty, we are likely in the
             # case where the gate's circuit definition is fully bound (so we can't detect its inputs
@@ -734,7 +740,7 @@ class QASM3Builder:
         # We register the gate only after building its body so that any gates we needed for that in
         # turn are registered in the correct order.  Gates can't be recursive in OQ3, so there's no
         # problem with delaying this.
-        return self.symbols.register_gate(source.name, source, params, qubits, body)
+        return self.symbols.register_gate(canonical.name, canonical, params, qubits, body)
 
     def assert_global_scope(self):
         """Raise an error if we are not in the global scope, as a defensive measure."""
--- a/qiskit/qasm3/exporter.py
+++ b/qiskit/qasm3/exporter.py
@@ -120,7 +120,8 @@ _RESERVED_KEYWORDS = frozenset(
 # This probably isn't precisely the same as the OQ3 spec, but we'd need an extra dependency to fully
 # handle all Unicode character classes, and this should be close enough for users who aren't
 # actively _trying_ to break us (fingers crossed).
-_VALID_IDENTIFIER = re.compile(r"(\$[\d]+|[\w][\w\d]*)", flags=re.U)
+_VALID_DECLARABLE_IDENTIFIER = re.compile(r"([\w][\w\d]*)", flags=re.U)
+_VALID_HARDWARE_QUBIT = re.compile(r"\$[\d]+", flags=re.U)
 _BAD_IDENTIFIER_CHARACTERS = re.compile(r"[^\w\d]", flags=re.U)
 
 
@@ -332,14 +333,15 @@ class SymbolTable:
         name_allowed = (
             (lambda name: not self.symbol_defined(name)) if unique else self.can_shadow_symbol
         )
+        valid_identifier = _VALID_DECLARABLE_IDENTIFIER
         if allow_rename:
-            if not _VALID_IDENTIFIER.fullmatch(name):
+            if not valid_identifier.fullmatch(name):
                 name = "_" + _BAD_IDENTIFIER_CHARACTERS.sub("_", name)
             base = name
             while not name_allowed(name):
                 name = f"{base}_{next(self._counter)}"
             return name
-        if not _VALID_IDENTIFIER.fullmatch(name):
+        if not valid_identifier.fullmatch(name):
             raise QASM3ExporterError(f"cannot use '{name}' as a name; it is not a valid identifier")
         if name in _RESERVED_KEYWORDS:
             raise QASM3ExporterError(f"cannot use the keyword '{name}' as a variable name")
@@ -364,7 +366,8 @@ class SymbolTable:
         variable: object,
         *,
         allow_rename: bool,
-        global_: bool = False,
+        force_global: bool = False,
+        allow_hardware_qubit: bool = False,
     ) -> ast.Identifier:
         """Register a variable in the symbol table for the given scope, returning the name that
         should be used to refer to the variable.  The same name will be returned by subsequent calls
@@ -377,14 +380,23 @@ class SymbolTable:
                 it.
             allow_rename: whether to allow the name to be mutated to escape it and/or make it safe
                 to define (avoiding keywords, subject to shadowing rules, etc).
-            global_: force this declaration to be in the global scope.
+            force_global: force this declaration to be in the global scope.
+            allow_hardware_qubit: whether to allow hardware qubits to pass through as identifiers.
+                Hardware qubits are a dollar sign followed by a non-negative integer, and cannot be
+                declared, so are not suitable identifiers for most objects.
         """
-        scope_index = 0 if global_ else -1
-        # We still need to do this escaping and shadow checking if `global_`, because we don't want
-        # a previous variable declared in the currently active scope to shadow the global.  This
-        # kind of logic would be cleaner if we made the naming choices later, after AST generation
-        # (e.g. by using only indices as the identifiers until we're ready to output the program).
-        name = self.escaped_declarable_name(name, allow_rename=allow_rename, unique=global_)
+        scope_index = 0 if force_global else -1
+        # We still need to do this escaping and shadow checking if `force_global`, because we don't
+        # want a previous variable declared in the currently active scope to shadow the global.
+        # This logic would be cleaner if we made the naming choices later, after AST generation
+        # (e.g. by using only indices as the identifiers until we're outputting the program).
+        if allow_hardware_qubit and _VALID_HARDWARE_QUBIT.fullmatch(name):
+            if self.symbol_defined(name):  # pragma: no cover
+                raise QASM3ExporterError(f"internal error: cannot redeclare hardware qubit {name}")
+        else:
+            name = self.escaped_declarable_name(
+                name, allow_rename=allow_rename, unique=force_global
+            )
         identifier = ast.Identifier(name)
         self.variables[scope_index][name] = variable
         if variable is not None:
@@ -415,7 +427,7 @@ class SymbolTable:
         "basis gate" that assumes that all calling signatures are valid and that all gates of this
         name are exactly compatible, which is somewhat dangerous."""
         # Validate the name is usable.
-        name = self.escaped_declarable_name(name, allow_rename=False)
+        name = self.escaped_declarable_name(name, allow_rename=False, unique=False)
         ident = ast.Identifier(name)
         if gate is None:
             self.gates[name] = GateDefinition(None, None)
@@ -438,7 +450,7 @@ class SymbolTable:
     ) -> ast.Identifier:
         """Register the given gate in the symbol table, using the given components to build up the
         full AST definition."""
-        name = self.escaped_declarable_name(name, allow_rename=True)
+        name = self.escaped_declarable_name(name, allow_rename=True, unique=False)
         ident = ast.Identifier(name)
         self.gates[name] = GateDefinition(
             source, ast.QuantumGateDefinition(ident, tuple(params), tuple(qubits), body)
@@ -829,7 +841,9 @@ class QASM3Builder:
             # We're referring to physical qubits.  These can't be declared in OQ3, but we need to
             # track the bit -> expression mapping in our symbol table.
             for i, bit in enumerate(circuit.qubits):
-                self.symbols.register_variable(f"${i}", bit, allow_rename=False)
+                self.symbols.register_variable(
+                    f"${i}", bit, allow_rename=False, allow_hardware_qubit=True
+                )
             return []
         if any(len(circuit.find_bit(q).registers) > 1 for q in circuit.qubits):
             # There are overlapping registers, so we need to use aliases to emit the structure.
@@ -986,7 +1000,7 @@ class QASM3Builder:
         """Build a :obj:`.SwitchCaseOp` into a :class:`.ast.SwitchStatement`."""
         real_target = self.build_expression(expr.lift(instruction.operation.target))
         target = self.symbols.register_variable(
-            "switch_dummy", None, allow_rename=True, global_=True
+            "switch_dummy", None, allow_rename=True, force_global=True
         )
         self._global_classical_forward_declarations.append(
             ast.ClassicalDeclaration(ast.IntType(), target, None)
--- a/qiskit/qasm3/exporter.py
+++ b/qiskit/qasm3/exporter.py
@@ -682,7 +682,7 @@ class QASM3Builder:
         This recurses through gate-definition statements."""
         if issubclass(gate.base_class, library.CXGate) and gate.ctrl_state == 1:
             # CX gets super duper special treatment because it's the base of Qiskit's definition
-            # tree, but isn't an OQ3 built-in (it was in OQ2).  We use `isinstance` because we
+            # tree, but isn't an OQ3 built-in (it was in OQ2).  We use `issubclass` because we
             # haven't fully fixed what the name/class distinction is (there's a test from the
             # original OQ3 exporter that tries a naming collision with 'cx').
             control, target = ast.Identifier("c"), ast.Identifier("t")
--- a/qiskit/qasm3/exporter.py
+++ b/qiskit/qasm3/exporter.py
@@ -12,20 +12,26 @@
 
 """QASM3 Exporter"""
 
+from __future__ import annotations
+
 import collections
-import re
+import contextlib
+import dataclasses
 import io
 import itertools
+import math
 import numbers
-from os.path import dirname, join, abspath
+import re
 from typing import Iterable, List, Sequence, Union
 
+from qiskit._accelerate.circuit import StandardGate
 from qiskit.circuit import (
+    library,
     Barrier,
     CircuitInstruction,
     Clbit,
+    ControlledGate,
     Gate,
-    Instruction,
     Measure,
     Parameter,
     ParameterExpression,
@@ -47,7 +53,6 @@ from qiskit.circuit.controlflow import (
     ContinueLoopOp,
     CASE_DEFAULT,
 )
-from qiskit.circuit.library import standard_gates
 from qiskit.circuit.register import Register
 from qiskit.circuit.tools import pi_check
 
@@ -115,13 +120,8 @@ _RESERVED_KEYWORDS = frozenset(
 # This probably isn't precisely the same as the OQ3 spec, but we'd need an extra dependency to fully
 # handle all Unicode character classes, and this should be close enough for users who aren't
 # actively _trying_ to break us (fingers crossed).
-_VALID_IDENTIFIER = re.compile(r"[\w][\w\d]*", flags=re.U)
-
-
-def _escape_invalid_identifier(name: str) -> str:
-    if name in _RESERVED_KEYWORDS or not _VALID_IDENTIFIER.fullmatch(name):
-        name = "_" + re.sub(r"[^\w\d]", "_", name)
-    return name
+_VALID_IDENTIFIER = re.compile(r"(\$[\d]+|[\w][\w\d]*)", flags=re.U)
+_BAD_IDENTIFIER_CHARACTERS = re.compile(r"[^\w\d]", flags=re.U)
 
 
 class Exporter:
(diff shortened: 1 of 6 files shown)


jakelishman:
Marking as high priority because this blocks #12730.

jakelishman:
The failing test is fixed by #12775.  I'll push a fix to the lint failure once that's merged.

qiskit/qasm3/exporter.py
```diff
@@ -115,13 +120,8 @@
...
 # handle all Unicode character classes, and this should be close enough for users who aren't
 # actively _trying_ to break us (fingers crossed).
-_VALID_IDENTIFIER = re.compile(r"[\w][\w\d]*", flags=re.U)
-
-
-def _escape_invalid_identifier(name: str) -> str:
-    if name in _RESERVED_KEYWORDS or not _VALID_IDENTIFIER.fullmatch(name):
-        name = "_" + re.sub(r"[^\w\d]", "_", name)
-    return name
+_VALID_IDENTIFIER = re.compile(r"(\$[\d]+|[\w][\w\d]*)", flags=re.U)
```
jlapeyre: EDIT: This is fixed by https://github.com/jakelishman/qiskit-terra/pull/25

This allows things like `register_gate_without_definition("$1", None)`. ~~Is this intended?~~

For example
```python
mygate = Gate("$1", 1, [])
qc = QuantumCircuit(1)
qc.append(mygate, [0])
qcdef = QuantumCircuit(1)
qcdef.x(0)
mygate.definition = qcdef
```
writes an OQ3 gate named `$1`  when exported, whereas `Gate("&1"...` is properly escaped.

jlapeyre: I might factor out the two uses of this regular expression with
```python
def is_valid_identifier(string):
    return _VALID_IDENTIFIER.fullmatch(string)
```
But it's ok the way it is, too.

A comment: Because of Perl habits, I would have used `^` and `$` to anchor the pattern instead of `fullmatch`. But I think using `fullmatch`  (regardless of whether the matching is refactored)  makes it easier to understand the code, both because the reader doesn't have to look back to the regular expression to know if the entire string must match, and becuase they need less knowledge of regular expressions.

qiskit/qasm3/exporter.py
```diff
@@ -829,10 +900,13 @@ def build_current_scope(self) -> List[ast.Statement]:
...
-            ast.ClassicalDeclaration(_build_ast_type(var.type), self._register_variable(var, scope))
-            for var in scope.circuit.iter_declared_vars()
+            ast.ClassicalDeclaration(
+                _build_ast_type(var.type),
+                self.symbols.register_variable(var.name, var, allow_rename=True),
+            )
+            for var in self.scope.circuit.iter_declared_vars()
         ]
-        for instruction in scope.circuit.data:
+        for instruction in self.scope.circuit.data:
```
jlapeyre, line 929: As mentioned in the PR description, a possible remaining benefit
of the superseded #12481 is fewer attribute lookups: `instruction.operation`.

However, the code block that's touched in #12481 is removed in the present PR, #12776.
I guess the time to make attribute lookups is negligible compared to other tasks, such
as string building (even though this is done by copying short strings to a stream).

A remaining benefit in my view is readability. But that's debatable. You might consider factoring out
the attribute lookup here. It's fine as it is, as well.

qiskit/qasm3/exporter.py
```diff
@@ -196,139 +196,308 @@ def dump(self, circuit, stream):
...
-        "u3": standard_gates.U3Gate,
-    }
-    include_paths = [abspath(join(dirname(__file__), "..", "qasm", "libs"))]
+# Just needs to have enough parameters to support the largest standard (non-controlled) gate in our
+# standard library.  We have to use the same `Parameter` instances each time so the equality
+# comparisons will work.
+_FIXED_PARAMETERS = (Parameter("p0"), Parameter("p1"), Parameter("p2"), Parameter("p3"))
+
+# Mapping of symbols defined by `stdgates.inc` to their gate definition source.
+_KNOWN_INCLUDES = {
```
jlapeyre, line 206: I might do something like this. Just a suggestion:
```python
from qiskit.circuit import(Gate, ControlledGate, Parameter, library)

_FIXED_PARAMETERS = (Parameter("p0"), Parameter("p1"), Parameter("p2"), Parameter("p3"))

def _gate_definition_source(gate: Gate) -> Gate:
    if gate._standard_gate and not isinstance(gate, ControlledGate):
        return gate.base_class(*_FIXED_PARAMETERS[: len(gate.params)])
    elif gate._standard_gate:
        return gate.base_class(*_FIXED_PARAMETERS[: len(gate.params)], ctrl_state=gate.ctrl_state)
    return gate

def _std_known_includes():
    std_gate_names = ['p', 'x', 'y', 'z', 'h', 's', 'sdg', 't', 'tdg',
                      'sx', 'rx', 'ry', 'rz', 'cx', 'cy', 'cz', 'cp',
                      'crx', 'cry', 'crz', 'ch', 'swap', 'ccx', 'cswap',
                      'cu', 'id', 'u1', 'u2', 'u3']
    alias_map = {'cx': 'CX', 'p': 'phase', 'cp': 'cphase'}

    qiskit_map = library.get_standard_gate_name_mapping()
    source_gates = {name: _gate_definition_source(qiskit_map[name]) for name in std_gate_names}
    for name, alias in alias_map.items():
        source_gates[alias] = _gate_definition_source(source_gates[name])
    return source_gates

_KNOWN_INCLUDES = {"stdgates.inc": _std_known_includes()}

```

qiskit/qasm3/exporter.py
```diff
@@ -196,139 +196,308 @@ def dump(self, circuit, stream):
...
+class GateDefinition:
+    """Symbol-table information on a gate."""
+
+    source: Gate | None
+    """The definition source of the gate.  This can be ``None`` if the gate was an overridden "basis
+    gate" for this export."""
+    node: ast.QuantumGateDefinition | None
+    """An AST node containing the gate definition.  This can be ``None`` if the gate came from an
+    included file, or is an overridden "basis gate" of the export."""
+
```
jlapeyre, line 263: If I understand correctly, `GateDefinition` links the qiskit "definition" with the OQ3 definition. And the qiskit definition is a `Gate` object instance with canonical parameters (fixed parameters shared by all gates).
The word "definition" is used here and repeatedly throughout the file. It's often not immediatley clear whether we are referring to the qiskit or the OQ3 defnition.  I wonder if we can find terms that make it easier to disitinguish.

For example, here we have "definition source of the gate" and "gate definition".  The word "source" is only used in this file for qiskit representations. So that must mean the second "gate definition" refers to OQ3. And of course `Gate` is a qiskit class representing gates. And `ast` must mean qasm. 

Maybe "qiskit definition source" and "OpenQASM 3 gate defintion". And also maybe state explicilty that this is the data that associates two representations in the two systems.

qiskit/qasm3/exporter.py
```diff
@@ -196,139 +196,308 @@ def dump(self, circuit, stream):
...
+                return out
+        raise KeyError(f"'{variable}' is not defined in the current context")
+
+    def register_gate_without_definition(self, name: str, gate: Gate | None) -> ast.Identifier:
+        """Register a gate that does not require an OQ3 definition.
+
+        If the ``gate`` is given, it will be used to validate that a call to it is compatible (such
+        as a known gate from an included file).  If it is not given, it is treated as a user-defined
+        "basis gate" that assumes that all calling signatures are valid and that all gates of this
+        name are exactly compatible, which is somewhat dangerous."""
```
jlapeyre, line 433: "somewhat dangerous"  haha

qiskit/qasm3/exporter.py
```diff
@@ -196,139 +196,308 @@ def dump(self, circuit, stream):
...
+        # Validate the name is usable.
+        name = self.escaped_declarable_name(name, allow_rename=False)
+        ident = ast.Identifier(name)
+        if gate is None:
+            self.gates[name] = GateDefinition(None, None)
+        else:
+            source = _gate_definition_source(gate)
+            self.gates[name] = GateDefinition(source, None)
+            if (standard_gate := getattr(source, "_standard_gate")) is not None:
```
jlapeyre: Why is it `getattr` here, but `._standard_gate` in `_gate_definition_source`?

jlapeyre (commented):
This looks great so far. I made a few comments.

I still have a couple of questions on the design. When I finish reading exporter.py tomorrow they may evaporate.
EDIT: The questions dissipated.

qiskit/qasm3/exporter.py
```diff
@@ -502,172 +648,98 @@ def build_program(self):
...
+            yield ast.Include(filename)
+
+    def define_gate(self, gate: Gate) -> ast.Identifier:
+        """Define a gate in the symbol table, including building the gate-definition statement for
+        it.
+
+        This recurses through gate-definition statements."""
+        if issubclass(gate.base_class, library.CXGate) and gate.ctrl_state == 1:
+            # CX gets super duper special treatment because it's the base of Qiskit's definition
+            # tree, but isn't an OQ3 built-in (it was in OQ2).  We use `isinstance` because we
```
jlapeyre, line 667: We use `issubclass` ?

jlapeyre: We use `issubclass` ?

qiskit/qasm3/exporter.py
```diff
@@ -196,139 +196,308 @@ def dump(self, circuit, stream):
...
+        self.user_gate_idents: dict[int, ast.Identifier] = {}
+        """Mapping of `id`s of user gates to the identifier we use for it."""
+
+        self.variables: list[dict[str, object]] = [{}]
+        """Stack of mappings of variable names to the Qiskit object that represents them.
+
+        The zeroth index corresponds to the global scope, the highest index to the current scope."""
+        self.objects: list[dict[object, ast.Identifier]] = [{}]
+        """Stack of mappings of Qiskit objects to the identifier (or subscripted identifier) that
+        refers to them.  This is similar to the inverse mapping of ``variables``.
```
jlapeyre, line 288: (See also the previous comment on descriptions of mappings)
I'm looking for a way to describe a map without too much ambiguity and subject/verb disagreement. Maybe something like:
```suggestion
        """Mapping from standard gate enumeration values to the identifiers we represent them as."""
        self.user_gate_idents: dict[int, ast.Identifier] = {}
        """Mapping from `id`s of user gates to the identifiers we use for them."""

        self.variables: list[dict[str, object]] = [{}]
        """Stack of mappings from variable names to the Qiskit objects that represent them.

        The zeroth index corresponds to the global scope, the highest index to the current scope."""
        self.objects: list[dict[object, ast.Identifier]] = [{}]
        """Stack of mappings from Qiskit objects to the identifiers (or subscripted identifiers) that
        refer to them.  This is similar to the inverse mapping of ``variables``.
```

qiskit/qasm3/exporter.py
```diff
@@ -196,139 +196,308 @@ def dump(self, circuit, stream):
...
-        "u2": standard_gates.U2Gate,
-        "u3": standard_gates.U3Gate,
-    }
-    include_paths = [abspath(join(dirname(__file__), "..", "qasm", "libs"))]
+# Just needs to have enough parameters to support the largest standard (non-controlled) gate in our
+# standard library.  We have to use the same `Parameter` instances each time so the equality
+# comparisons will work.
+_FIXED_PARAMETERS = (Parameter("p0"), Parameter("p1"), Parameter("p2"), Parameter("p3"))
+
+# Mapping of symbols defined by `stdgates.inc` to their gate definition source.
```
jlapeyre, line 205: I think I actually read this smoothly without big hitches for disambiguation. But it still might be more clear like this:
```suggestion
# Mapping from symbols defined by `stdgates.inc` to their gate definition sources.
```

qiskit/qasm3/exporter.py
```diff
@@ -196,139 +196,308 @@ def dump(self, circuit, stream):
...
+    """An AST node containing the gate definition.  This can be ``None`` if the gate came from an
+    included file, or is an overridden "basis gate" of the export."""
+
+
+class SymbolTable:
+    """Track Qiskit objects and the OQ3 identifiers used to refer to them."""
+
+    def __init__(self):
+        self.gates: collections.OrderedDict[str, GateDefinition | None] = {}
+        """Mapping of the symbol name to the "definition source" of the gate, which provides its
```
jlapeyre, line 270: Elsewhere, I make suggested changes to the descriptions of mappings. But line 264 is a good  choice.

Two good ways to describe the maps are

"Mapping of the red item to the blue item"

"Mapping from the red items to the blue items"

qiskit/qasm3/exporter.py
```diff
@@ -196,139 +196,308 @@ def dump(self, circuit, stream):
...
+
+@dataclasses.dataclass
+class BuildScope:
+    """The structure used in the builder to store the contexts and re-mappings of bits from the
+    top-level scope where the bits were actually defined."""
+
+    circuit: QuantumCircuit
+    """The circuit block that we're currently working on exporting."""
+    bit_map: dict[Bit, Bit]
+    """Mapping of bit objects in ``circuit`` to the bit objects in the global-scope program
```
jlapeyre, line 517: This is readable. I don't think the English-language description needs to be mathematically precise, as long as a reader easily and unconsciously gets a precise understanding.
I'd suggest making the pattern of descriptions uniform, or at least intentionally choosing the pattern based on readability.

In most cases, this is a very minor point. But it also would take only a couple of minutes to edit the 10 or so cases.
It's somewhat more important to make a good choice with more complicated objects, such  as "stack of mappings".

qiskit/qasm3/exporter.py
```diff
@@ -196,139 +196,308 @@ def dump(self, circuit, stream):
...
+        """Enter a new variable scope."""
+        self.variables.append({})
+        self.objects.append({})
+
+    def pop_scope(self):
+        """Exit the current scope, returning to a previous scope."""
+        self.objects.pop()
+        self.variables.pop()
+
+    def new_context(self) -> SymbolTable:
```
jlapeyre, line 305: You might consider `def new_symbol_table(self) -> SymbolTable:`. This is a method in the class `SymbolTable` and the doc strings elsewhere in the class refer to `SymbolTable`. I suppose developer docs will never be extracted and formatted. But if they were, the method name and description here would be puzzling.

I find the contexts, symbol tables, and scopes a bit confusing. Maybe it's unavoidable. But a brief overview might be helpful.

qiskit/qasm3/exporter.py
```diff
@@ -196,139 +196,308 @@ def dump(self, circuit, stream):
...
+    # (and control state, if appropriate).
+    if gate._standard_gate and not isinstance(gate, ControlledGate):
+        return gate.base_class(*_FIXED_PARAMETERS[: len(gate.params)])
+    elif gate._standard_gate:
+        return gate.base_class(*_FIXED_PARAMETERS[: len(gate.params)], ctrl_state=gate.ctrl_state)
+    return gate
+
+
+@dataclasses.dataclass
+class BuildScope:
```
jlapeyre, line 510: There are two kinds of scope, right? Is it: The scope in `SymbolTable` is an OQ3 scope. And `BuildScope` is a scope relevant only to the process of exporting... or maybe a qiskit-side scope. In the doc string, we have
"top-level scope where the bits were actually defined". Does "scope" here refer to qiskit? Maybe using terms "qiskit-scope" and "qasm-scope", or whatever the correct categories are, would be helpful.

qiskit/qasm3/exporter.py
```diff
@@ -196,139 +196,308 @@ def dump(self, circuit, stream):
...
+    node: ast.QuantumGateDefinition | None
+    """An AST node containing the gate definition.  This can be ``None`` if the gate came from an
+    included file, or is an overridden "basis gate" of the export."""
+
+
+class SymbolTable:
+    """Track Qiskit objects and the OQ3 identifiers used to refer to them."""
+
+    def __init__(self):
+        self.gates: collections.OrderedDict[str, GateDefinition | None] = {}
```
jlapeyre: Did you consider keeping everything in the same stack of symbol tables? I suppose you'd have attach a type or other label to each object, in order to do this. But I don't see a lot of types otherwise. To me, keeping the gates separate like you have done looks cleaner and more manageable. Maybe it's the only reasonable way.

qiskit/qasm3/exporter.py
jlapeyre: In fact, I'm looking at the long list of lines like ` CustomGate(library.XGate, "x", 0, 1)` in `__init__.py` and the list here, and the list in `library.get_standard_gate_name_mapping`. It seems like it would not be hard to encapsulate this data in one place with simple API.

The snippet above is a nod in this direction. It reduces the sources of truth a bit.

qiskit/qasm3/exporter.py
```diff
@@ -196,139 +196,308 @@ def dump(self, circuit, stream):
...
+            return ast.Identifier(gate.name)
+        if (standard_gate := getattr(source, "_standard_gate")) is not None:
+            if (our_ident := self.standard_gate_idents.get(standard_gate)) is None:
+                return None
+            return our_ident if self.gates[our_ident.string].source == source else None
+        # No need to check equality if we're looking up by `id`; we must have the same object.
+        return self.user_gate_idents.get(id(source))
+
+
+def _gate_definition_source(gate: Gate) -> Gate:
```
jlapeyre: I've been wondering why you don't store something like `(GateClass, num_params)`, instead of constructing a gate with fixed, otherwise useless, params. It looks like there are several reasons. One is that we don't have a convenient interface for obtaining information about a gate other than calling methods on instances of the gate. So, you'd have to construct a gate anyway to get it. Another is you don't try to do anything clever with user gates; they are stored as instances, and are their own source, and you hope for the best. So if you treated standard gates differently, your definition of source would have a more heterogeneous format. Although this is a sort of vague view of the problem, trying to store the definition source differently seems to cause problems, and not to really solve any.

What you have here is pretty straightforward, given the constraints of current qiskit data models.

qiskit/qasm3/exporter.py
```diff
@@ -364,108 +531,86 @@ def __init__(
...
-            circuit.data, opaques=[], gates=[]
-        )
-        opaque_definitions = [
-            self.build_opaque_definition(instruction) for instruction in opaques_to_declare
-        ]
-        gate_definitions = [
-            self.build_gate_definition(instruction) for instruction in gates_to_declare
-        ]
+        # The order we build parts of the AST has an effect on which names will get escaped to avoid
+        # collisions.  The current ideas are:
```
jlapeyre, line 609: This is pretty clear and useful.

test/python/qasm3/test_export.py
```diff
@@ -494,7 +542,7 @@ def test_unbound_circuit(self):
         )
         self.assertEqual(Exporter().dumps(qc), expected_qasm)
 
-    def test_unknown_parameterized_gate_called_multiple_times(self):
+    def test_standard_parameterized_gate_called_multiple_times(self):
```
jlapeyre, line 545: EDIT: I pushed a commit add a test here https://github.com/jakelishman/qiskit-terra/pull/24

I don't see a test for the case that first instance of the gate is **not** generic.
```python
        qc.rzx(0.5, 0, 1)
        qc.rzx(x, 0, 1)
        qc.rzx(y, 0, 1)
```
With this PR this is exported correctly!

jlapeyre (commented):
@jakelishman  This is the last part of my review. 

Above, I mentioned a design question. I wondered why a choice was made. I think I see why and mention this in a comment.

This looks like a great improvement. I was not able to go over all of it.

One of the more important comments below is an apparently missing test for a bug fix. (The test still assumes the bug!)

jlapeyre:
TLDR; https://github.com/jakelishman/qiskit-terra/pull/25 and https://github.com/jakelishman/qiskit-terra/pull/24 fix a couple of oversights/bugs in this PR.

qiskit/qasm3/exporter.py
jakelishman: My plan was to leave this for your PR anyway - I wasn't trying to override that.

qiskit/qasm3/exporter.py
jakelishman: The mapping between Qiskit's gate names and the name that they're given in OQ2 and OQ3 isn't entirely identical (Qiskit's OQ2 `qelib.inc` extension calls `c3sx` something different, iirc), and there's the aliasing changes, so I wouldn't necessarily say that there's a duplication in sources of truth here; the "linking" between Qiskit's source of truth and OpenQASM's is itself a new set of facts to be encoded.

Personally I wrote it the way I did because I find the (relatively short) table easier to read and understand than generation code, and there's less to debug in terms of logic.  The more code I write, I find that I'm raising my internal barrier to wanting to write "code that writes code" and so on - I used to do it a lot more, then it was often more effort to maintain than a slightly longer but more explicit list.

Obviously there's a limit to that - I'm of course not suggesting that tablegen has no purpose.

qiskit/qasm3/exporter.py
jakelishman: We don't subclass `CXGate` within Qiskit, but I think think this is what would be expected if we did.  It also avoids pylint complaining about unidiomatic type checks (`gate.base_class is library.CXGate` would be a complaint, I think).

qiskit/qasm3/exporter.py
jakelishman: What are you thinking isn't precise about this statement?

qiskit/qasm3/exporter.py
jakelishman: I think the docstring of this method fairly clearly defines what a "new context" is?  imo, `new_symbol_table` erases part of the description of the method - if I read it, I'd think "what's the difference from `SymbolTable()`?".

qiskit/qasm3/exporter.py
jakelishman: The "scope" in `SymbolTable` and `BuildScope` refer to the same concept - I was trying to unify those names.  OpenQASM 3 has more than one type of scope (gates introduce different scoping rules to non-gate blocks, e.g.), which is also represented here by the idea of a "context", but a context still includes a current scope.

qiskit/qasm3/exporter.py
jakelishman: Gates have different rules to variables with regards to definitions, shadowing, etc, so it's a bit clearer (imo) to track them separately - for example, there's no concept of scopes for gate identifiers, since they've got to be global definitions and can't be shadowed ever.

We also need to track different information for `Gate` objects compared to other symbols, so it's convenient to put them in a separate table.

qiskit/qasm3/exporter.py
jakelishman: Hopefully this is a little clearer with the renaming of this method to `_gate_canonical_form`.  It's clearer for typing and usage if everything's always just a `Gate` (or `None`) - no need to do type reflection on lookups, just use the same comparisons as always.

test/python/qasm3/test_export.py
jakelishman: Thanks - cherry-picked as 2cbf15ff4d0ae35e845e9db9a66070597b14b679.

qiskit/qasm3/exporter.py
jakelishman: I'm not entirely sure.  Everything's required to be `Gate` (or `None`, but that's already handled) throughout these paths, so the attribute always exists.  Removed unnecessary `getattr` in b51386fac2.

qiskit/qasm3/exporter.py
jakelishman: Good idea, thanks.  Much of what we're doing is object canonicalisation for equality comparisons, so I've renamed a bunch of things around that idea in 771676c412.

qiskit/qasm3/exporter.py
jakelishman: I cherry-picked jakelishman/qiskit-terra#25 (with minor modifications) as 598fae3b5a2df29a9aeba4d8bac6bdc3f7656d89.

imo the ^$ anchoring restricts the regex beyond what its name (`_VALID_IDENTIFIER`) actually suggests - it works well when the regex is defined inline, but I agree that `fullmatch` is clearer when it's encapsulated in a variable.

qiskit/qasm3/exporter.py
jakelishman: I've just "mapping of" consistently, and I'm not sure I agree there's any distinction in the preposition implied on plurality - both read fine to me.

jakelishman:
I've addressed all of the above actionable comments.  I left the preposition/plurality in "mapping of X(s) to Y(s)" as-is, since I don't think there's any meaningful difference, and we already use both "mapping of" and "mapping from" with plurals and singulars throughout Qiskit.

qiskit/qasm3/exporter.py
jakelishman: Leaving as is, per https://github.com/Qiskit/qiskit/pull/12776#issuecomment-2242631327.

qiskit/qasm3/exporter.py
jakelishman: Leaving as is, per https://github.com/Qiskit/qiskit/pull/12776#issuecomment-2242631327.

qiskit/qasm3/exporter.py
jakelishman: Leaving as is, per https://github.com/Qiskit/qiskit/pull/12776#issuecomment-2242631327.

qiskit/qasm3/exporter.py
jlapeyre: Yes, thanks for that. We can think about that after this PR then.

qiskit/qasm3/exporter.py
jlapeyre: Sure, those are good arguments. Simple, readable, self-contained code is nice to find.

jakelishman:
The last couple of remaining comments are about potentially improving internal developer documentation in the exporter.  That could well be valuable for the future, but let's do it in a follow-up if so, so we can get this merged and unblock #12730 in the interim.

qiskit/qasm3/exporter.py
jlapeyre: My comment was not clear!  I meant that the doc string says "We use `isinstance`", the code uses `issubclass`

jakelishman: Oh I see it now - thanks!  Fixed in 984c4220c1.

jlapeyre:
If you've reviewed the language around mappings and are satisfied, then so am I.

jlapeyre (approved):
Looks great

test/python/qasm3/test_export.py
```diff
@@ -429,11 +422,14 @@ def test_custom_gate_with_params_bound_main_call(self):
 
         circuit.assign_parameters({parameter0: pi, parameter1: pi / 2}, inplace=True)
 
+        # NOTE: this isn't exactly what we want; note that the parameters in the signature are not
+        # actually used.  It would be fine to change the output of the exporter to make `custom` non
+        # parametric in this case.
```
ElePT, line 427: (just to make sure I am following) This isn't what we want because we want the parameters in the signature to be used in the definition instead of `pi`/`pi/4`? Where would the parameter value be stored then?

I can see that at least the issue from #7335 (the value being copied over) isn't there anymore.

ElePT (approved):
LGTM too!

test/python/qasm3/test_export.py
jakelishman: Ideally, the output of this test should be
```qasm
OPENQASM 3.0;
include "stdgates.inc";
qubit[2] q;
gate custom(p0, p1) q0, q1 {
  rz(p0) q0;
  rz(p1 / 2) q1;
}
custom(pi, pi/2) q[0], q[1];
```
since that's how the `custom` gate is originally defined in Qiskit.  We can't actually do that, though, because the prior call to `assign_parameters` erases the original parametric definition, so we can't find it again.  The test as written doesn't use `p0` or `p1`.

With this PR, this "bad" behaviour is only limited to custom user gates, though, so it's not such a huge deal anymore, as it's rather less likely to come up in practice.

PR #12780, opened 2024-07-17
Add config option to leverage all cores for sabre

mtreinish:
### Summary

By default when running sabre in parallel we use a fixed number of threads (depending on optimization level). This was a tradeoff made for having deterministic results across multiple systems with a fixed seed set. However when running qiskit on systems with a lot of CPUs available we're leaving potential performance on the table by not using all the available cores. This new flag lets users opt-in to running sabre with n trials for n CPUs to potentially get better output results from the transpiler, with minimal to no runtime overhead, at the cost of the results not necessarily being reproducible when run on a different computer.

### Details and comments

### DIFF
--- a/qiskit/transpiler/preset_passmanagers/builtin_plugins.py
+++ b/qiskit/transpiler/preset_passmanagers/builtin_plugins.py
@@ -12,6 +12,8 @@
 
 """Built-in transpiler stage plugins for preset pass managers."""
 
+import os
+
 from qiskit.transpiler.passmanager import PassManager
 from qiskit.transpiler.exceptions import TranspilerError
 from qiskit.transpiler.passes import BasicSwap
@@ -63,6 +65,10 @@ from qiskit.circuit.library.standard_gates import (
     SXGate,
     SXdgGate,
 )
+from qiskit.utils.parallel import CPU_COUNT
+from qiskit import user_config
+
+CONFIG = user_config.get_config()
 
 
 class DefaultInitPassManager(PassManagerStagePlugin):
@@ -397,11 +403,16 @@ class SabreSwapPassManager(PassManagerStagePlugin):
             pass_manager_config.initial_layout,
         )
         if optimization_level == 0:
+            if CONFIG.get("sabre_all_threads", None) or os.getenv("QISKIT_SABRE_ALL_THREADS"):
+                trial_count = CPU_COUNT
+            else:
+                trial_count = 5
+
             routing_pass = SabreSwap(
                 coupling_map_routing,
                 heuristic="basic",
                 seed=seed_transpiler,
-                trials=5,
+                trials=trial_count,
             )
             return common.generate_routing_passmanager(
                 routing_pass,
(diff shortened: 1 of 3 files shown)


releasenotes/notes/add-sabre-all-threads-option-ad4ff7a4d045cb2b.yaml
```diff
@@ -0,0 +1,26 @@
...
+    corresponding environment variable ``QISKIT_SABRE_ALL_THREADS``. When this
+    flag is set the preset pass managers will run the :class:`.SabreLayout`
+    and :class:`.SabreSwap` transpiler passes using all the available
+    CPUs on the local system. Using this option is a tradeoff between
+    determinism of output between different computers and potentially better
+    output with fewer :class:`.SwapGate`\s.
+
+    These transpiler passes run multiple random trials in parallel and pick
+    the result which results in the fewest :class:`.SwapGate`\s. As a rule of
+    thumb if you run more trials the provides the algorithm more opportunities
```
ElePT:
```suggestion
    the output which results in the fewest :class:`.SwapGate`\s. As a rule of
    thumb, if you run more trials, this provides the algorithm more opportunities
```

releasenotes/notes/add-sabre-all-threads-option-ad4ff7a4d045cb2b.yaml
```diff
@@ -0,0 +1,26 @@
...
+    to find a better result. By default the preset pass managers used a fixed
```
ElePT: (still being used)
```suggestion
    to find a better result. By default, the preset pass managers use a fixed
```

ElePT (commented):
LGTM, not that I'll ever have the chance to see the effects of the option in my laptop 😝. I just left a few suggestions on the reno.

qiskit/transpiler/preset_passmanagers/builtin_plugins.py
```diff
@@ -397,11 +403,16 @@ def pass_manager(self, pass_manager_config, optimization_level=None) -> PassMana
             pass_manager_config.initial_layout,
         )
         if optimization_level == 0:
+            if CONFIG.get("sabre_all_threads", None) or os.getenv("QISKIT_SABRE_ALL_THREADS"):
+                trial_count = CPU_COUNT
+            else:
+                trial_count = 5
+
```
eliarbel: It would be nice to replace this repeating block with a small util function like this:

```python
def get_trial_count(default_trials=5):
    if CONFIG.get("sabre_all_threads", None) or os.getenv("QISKIT_SABRE_ALL_THREADS"):
        return CPU_COUNT
    return default_trials
```

And use it here and in the other places below

mtreinish: This is a good idea, I'll refactor this. I was also thinking about this logic overnight and was wondering if maybe it would make more sense to `max(default_trials, CPU_COUNT)` if the option is set instead of always `CPU_COUNT` . Then we don't need to caveat about not setting the flag if you have less than 20 CPUs on your local system.

mtreinish: Done in: https://github.com/Qiskit/qiskit/pull/12780/commits/21ff70fc5f7c09521a3ca90a117030d2b7b7349e

eliarbel (approved):
LGTM

ElePT (approved):
LGTM too

PR #12785, opened 2024-07-18
Deprecate accidentally public internal helper methods on QuantumCircuit

### LINKED ISSUE #12417: Deprecate accidentally public internal helper methods on QuantumCircuit
### What should we add?

The `QuantumCircuit` class has 5 methods which shouldn't really be exposed publically as they're only used as internal helpers as part of other functionality. However they have accidentally been documented as part of the public API in Qiskit and we need to support them for the remainder of 1.x. We should however mark these public entrypoints as deprecated ASAP and work towards making them internal only in 2.0. These methods are:

- `.cast()`
- `.cbit_argument_conversion()`
- `.cls_instances()`
- `.cls_prefix()`
- `.qbit_argument_conversion()`

ElePT:
### Summary
Fixes  #12417. I added the "private version" of the methods so that we can then remove the public one in 2.0. Let me know if you had a different idea for the deprecation.


qiskit/circuit/quantumcircuit.py
```diff
@@ -2267,18 +2291,52 @@ def __getitem__(self, item):
...
     def qbit_argument_conversion(self, qubit_representation: QubitSpecifier) -> list[Qubit]:
         """
         Converts several qubit representations (such as indexes, range, etc.)
         into a list of qubits.
 
+        Args:
+            qubit_representation (Object): representation to expand
+
+        Returns:
+            List(Qubit): the resolved instances of the qubits.
```
Cryoris: Since the typehints are inline in the signature, I don't think they are needed in the docstring anymore 🙂 (Same below for the `clbit_argument_conversion`)
```suggestion
            qubit_representation: Representation to expand.

        Returns:
            The resolved instances of the qubits.
```

qiskit/circuit/quantumcircuit.py
```diff
@@ -1400,24 +1401,47 @@ def _increment_instances(cls):
...
     @classmethod
+    def _cls_instances(cls) -> int:
+        """Return the current number of instances of this class,
+        useful for auto naming."""
+        return cls.instances
+
+    @classmethod
+    @deprecate_func(
+        since=1.2,
+        removal_timeline="in qiskit 2.0",
```
Cryoris: Should we uppercase this? 🙂 
```suggestion
        removal_timeline="in Qiskit 2.0",
```

Cryoris (commented):
Two tiny comments otherwise LGTM 🙂

qiskit/circuit/quantumcircuit.py
ElePT: I wrote it in lowercase because the beginning of the message writes `qiskit` in lowercase. But I also don't love it. Anyways, I modified the timeline message in [a7b7361](https://github.com/Qiskit/qiskit/pull/12785/commits/a7b73616bc3004b31218fa5d48e23c81845f89cb) to say 
"in the 2.0 release" to be consistent with other deprecations (qobj and assemble).

Cryoris: Oh I didn't notice... that seems odd. Maybe we should change that as well 🤔 (not here ofc)

qiskit/circuit/quantumcircuit.py
```diff
@@ -2267,38 +2291,91 @@ def __getitem__(self, item):
...
+        """
+        return self._cbit_argument_conversion(clbit_representation)
+
+    def _cbit_argument_conversion(self, clbit_representation: ClbitSpecifier) -> list[Clbit]:
+        """
+        Converts several classical bit representations (such as indexes, range, etc.)
+        into a list of classical bits.
+
+        Args:
+            clbit_representation : Representation to expand.
```
Cryoris:
```suggestion
            clbit_representation: Representation to expand.
```

Cryoris (approved):
LGTM, thanks for the updates! 🙂

PR #12794, opened 2024-07-19
Move `QuantumCircuit.assign_parameters` to Rust

jakelishman:
### Summary

This is (as far as I could tell), the last really major performance regression in our asv suite compared to 1.1.0, so with this commit, we should be at _not worse_ for important utility-scale benchmarks.

This largely rewrites `ParamTable` (renamed back to `ParameterTable` because I kept getting confused with `Param`) to have more Rust-friendly interfaces available, so that `assign_parameters` can then use them.

This represents a 2-3x speedup in `assign_parameters` performance over 1.1.0, when binding simple `Parameter` instances.  Approximately 75% of the time is now spent in Python-space `Parameter.assign` and `ParameterExpression.numeric` calls; almost all of this could be removed were we to move `Parameter` and `ParameterExpression` to have their data exposed directly to Rust space.  The percentage of time spent in Python space only increases if the expressions to be bound are actual `ParameterExpression`s and not just `Parameter`.

Most changes in the test suite are because of the use of internal-only methods that changed with the new `ParameterTable`.  The only discrepancy is a bug in `test_pauli_feature_map`, which was trying to assign using a set.

### Details and comments

Built on #12730, so will need rebasing over it.

~I think this first commit might accidentally have introduced a small (10%) regression to parametric-circuit construction time over its parent.   That's a mistake if so - I should be able to fix that later.~ *edit*: on retiming, I couldn't reproduce a problem - if anything, this commit is a minor improvement.

Timings for parametric circuit benchmarks compared to 1.1.0 (the different SHA1 is because I hadn't written the commit message when I took the benchmark):

```

| Change   | Before [7d29dc1b] <1.1.0^0>   | After [f391e6d4]    | Ratio   | Benchmark (Parameter)                                                                                           |
|----------|-------------------------------|---------------------|---------|-----------------------------------------------------------------------------------------------------------------|
| -        | 1.50±0.05ms                   | 589±40μs            | 0.39    | circuit_construction.ParameterizedCircuitBindBench.time_bind_params(20, 128, 128)                               |
| -        | 1.19±0.02ms                   | 427±10μs            | 0.36    | circuit_construction.ParameterizedCircuitBindBench.time_bind_params(20, 128, 8)                                 |
| -        | 1.16±0.04s                    | 358±4ms             | 0.31    | circuit_construction.ParameterizedCircuitBindBench.time_bind_params(20, 131072, 128)                            |
| -        | 1.52±0.02s                    | 634±8ms             | 0.42    | circuit_construction.ParameterizedCircuitBindBench.time_bind_params(20, 131072, 131072)                         |
| -        | 1.07±0.02s                    | 350±5ms             | 0.33    | circuit_construction.ParameterizedCircuitBindBench.time_bind_params(20, 131072, 2048)                           |
| -        | 1.13±0.01s                    | 410±3ms             | 0.36    | circuit_construction.ParameterizedCircuitBindBench.time_bind_params(20, 131072, 32768)                          |
| -        | 1.12±0.01s                    | 370±2ms             | 0.33    | circuit_construction.ParameterizedCircuitBindBench.time_bind_params(20, 131072, 8)                              |
| -        | 1.06±0.01s                    | 362±2ms             | 0.34    | circuit_construction.ParameterizedCircuitBindBench.time_bind_params(20, 131072, 8192)                           |
| -        | 15.6±0.2ms                    | 5.37±0.2ms          | 0.34    | circuit_construction.ParameterizedCircuitBindBench.time_bind_params(20, 2048, 128)                              |
| -        | 22.4±1ms                      | 7.90±0.2ms          | 0.35    | circuit_construction.ParameterizedCircuitBindBench.time_bind_params(20, 2048, 2048)                             |
| -        | 15.2±0.6ms                    | 5.26±0.3ms          | 0.35    | circuit_construction.ParameterizedCircuitBindBench.time_bind_params(20, 2048, 8)                                |
| -        | 270±6ms                       | 85.5±2ms            | 0.32    | circuit_construction.ParameterizedCircuitBindBench.time_bind_params(20, 32768, 128)                             |
| -        | 264±10ms                      | 86.4±1ms            | 0.33    | circuit_construction.ParameterizedCircuitBindBench.time_bind_params(20, 32768, 2048)                            |
| -        | 370±10ms                      | 147±3ms             | 0.40    | circuit_construction.ParameterizedCircuitBindBench.time_bind_params(20, 32768, 32768)                           |
| -        | 271±10ms                      | 90.9±2ms            | 0.34    | circuit_construction.ParameterizedCircuitBindBench.time_bind_params(20, 32768, 8)                               |
| -        | 274±3ms                       | 98.1±0.8ms          | 0.36    | circuit_construction.ParameterizedCircuitBindBench.time_bind_params(20, 32768, 8192)                            |
| -        | 347±5μs                       | 144±8μs             | 0.42    | circuit_construction.ParameterizedCircuitBindBench.time_bind_params(20, 8, 8)                                   |
| -        | 63.5±2ms                      | 21.6±1ms            | 0.34    | circuit_construction.ParameterizedCircuitBindBench.time_bind_params(20, 8192, 128)                              |
| -        | 67.1±2ms                      | 23.8±0.6ms          | 0.35    | circuit_construction.ParameterizedCircuitBindBench.time_bind_params(20, 8192, 2048)                             |
| -        | 61.6±2ms                      | 21.4±0.6ms          | 0.35    | circuit_construction.ParameterizedCircuitBindBench.time_bind_params(20, 8192, 8)                                |
| -        | 87.3±2ms                      | 34.3±0.6ms          | 0.39    | circuit_construction.ParameterizedCircuitBindBench.time_bind_params(20, 8192, 8192)                             |
| -        | 3.84±0.02ms                   | 2.02±0.1ms          | 0.53    | circuit_construction.ParameterizedCircuitConstructionBench.time_build_parameterized_circuit(20, 128, 128)       |
| -        | 3.53±0.4ms                    | 1.38±0.1ms          | 0.39    | circuit_construction.ParameterizedCircuitConstructionBench.time_build_parameterized_circuit(20, 128, 8)         |
| -        | 2.57±0.01s                    | 995±10ms            | 0.39    | circuit_construction.ParameterizedCircuitConstructionBench.time_build_parameterized_circuit(20, 131072, 128)    |
| -        | 3.52±0.06s                    | 1.85±0.02s          | 0.53    | circuit_construction.ParameterizedCircuitConstructionBench.time_build_parameterized_circuit(20, 131072, 131072) |
| -        | 2.64±0.06s                    | 1.03±0.02s          | 0.39    | circuit_construction.ParameterizedCircuitConstructionBench.time_build_parameterized_circuit(20, 131072, 2048)   |
| -        | 2.84±0.02s                    | 1.23±0.01s          | 0.43    | circuit_construction.ParameterizedCircuitConstructionBench.time_build_parameterized_circuit(20, 131072, 32768)  |
| -        | 2.62±0.05s                    | 998±10ms            | 0.38    | circuit_construction.ParameterizedCircuitConstructionBench.time_build_parameterized_circuit(20, 131072, 8)      |
| -        | 2.63±0.03s                    | 1.09±0.02s          | 0.41    | circuit_construction.ParameterizedCircuitConstructionBench.time_build_parameterized_circuit(20, 131072, 8192)   |
| -        | 45.1±1ms                      | 16.5±1ms            | 0.37    | circuit_construction.ParameterizedCircuitConstructionBench.time_build_parameterized_circuit(20, 2048, 128)      |
| -        | 50.9±0.7ms                    | 29.4±0.7ms          | 0.58    | circuit_construction.ParameterizedCircuitConstructionBench.time_build_parameterized_circuit(20, 2048, 2048)     |
| -        | 38.7±0.3ms                    | 15.4±0.6ms          | 0.40    | circuit_construction.ParameterizedCircuitConstructionBench.time_build_parameterized_circuit(20, 2048, 8)        |
| -        | 640±20ms                      | 247±2ms             | 0.39    | circuit_construction.ParameterizedCircuitConstructionBench.time_build_parameterized_circuit(20, 32768, 128)     |
| -        | 650±7ms                       | 273±2ms             | 0.42    | circuit_construction.ParameterizedCircuitConstructionBench.time_build_parameterized_circuit(20, 32768, 2048)    |
| -        | 867±20ms                      | 468±9ms             | 0.54    | circuit_construction.ParameterizedCircuitConstructionBench.time_build_parameterized_circuit(20, 32768, 32768)   |
| -        | 672±30ms                      | 248±2ms             | 0.37    | circuit_construction.ParameterizedCircuitConstructionBench.time_build_parameterized_circuit(20, 32768, 8)       |
| -        | 703±10ms                      | 313±6ms             | 0.45    | circuit_construction.ParameterizedCircuitConstructionBench.time_build_parameterized_circuit(20, 32768, 8192)    |
| -        | 873±10μs                      | 501±40μs            | 0.57    | circuit_construction.ParameterizedCircuitConstructionBench.time_build_parameterized_circuit(20, 8, 8)           |
| -        | 159±5ms                       | 66.4±1ms            | 0.42    | circuit_construction.ParameterizedCircuitConstructionBench.time_build_parameterized_circuit(20, 8192, 128)      |
| -        | 174±2ms                       | 77.1±0.5ms          | 0.44    | circuit_construction.ParameterizedCircuitConstructionBench.time_build_parameterized_circuit(20, 8192, 2048)     |
| -        | 156±4ms                       | 64.1±1ms            | 0.41    | circuit_construction.ParameterizedCircuitConstructionBench.time_build_parameterized_circuit(20, 8192, 8)        |
| -        | 210±2ms                       | 113±1ms             | 0.54    | circuit_construction.ParameterizedCircuitConstructionBench.time_build_parameterized_circuit(20, 8192, 8192)     |
```


crates/circuit/src/circuit_data.rs
```diff
@@ -160,172 +164,73 @@ impl CircuitData {
                 #[cfg(feature = "cache_pygates")]
                 py_op: RefCell::new(None),
             });
+            res.track_instruction_parameters(py, res.data.len() - 1)?;
```
mtreinish, line 167: This is technically an oversight because I guess it's possible for someone in rust space to create a parameter object and use that here. But, in practice I didn't think anyone would. Do you think it's worth the overhead of doing this for the rare case?

mtreinish:
I've tagged this as high priority because there are some bugs in parameter binding that @woodsp-ibm identified on main that this branch fixes. So we shouldn't tag 1.2.0rc1 without this included.

jakelishman:
Oh, what bugs?

*edit*: the point being: let's add unit tests.

woodsp-ibm:
I linked the algorithms PR I have which was to address issues arising when testing against Qsikit main that showed the bug being referred to. In that PR one was due to use of some internals that are no longer there, the other, where I had added a .data copy as a workaround was the bug being referred to. I had some standalone code as a sample if its of help - I showed it to Matthew when discussing the issue so he has it too.

mtreinish:
The example Steve shared was:
```python
from qiskit.circuit import QuantumCircuit, Parameter, ClassicalRegister, QuantumRegister

from qiskit.transpiler.passes import TranslateParameterizedGates

SUPPORTED_GATES = [
    "rx",
    "ry",
    "rz",
    "rzx",
    "rzz",
    "ryy",
    "rxx",
    "cx",
    "cy",
    "cz",
    "ccx",
    "swap",
    "iswap",
    "h",
    "t",
    "s",
    "sdg",
    "x",
    "y",
    "z",
]

a = Parameter("a")
qc = QuantumCircuit(1)
qc.h(0)
qc.p(a, 0)
qc.h(0)

print(qc)
print(qc.parameters)

translator = TranslateParameterizedGates(SUPPORTED_GATES)
qc = translator(qc)

print(qc)
print(qc.parameters)


qc1 = qc.copy()
qr_aux = QuantumRegister(1, "qr_aux")
cr_aux = ClassicalRegister(1, "cr_aux")
qc1.add_register(qr_aux)
qc1.add_register(cr_aux)
qc1.h(qr_aux)
qc1.data.insert(0, qc1.data.pop()) # This line seems to mess things up

print(qc1)
print(qc1.parameters)

cct = qc1.assign_parameters({a: 0.44})

print(cct)
print(cct.parameters)
```

on main it looked like the final circuit printed was not binding `a` on the correct index in `CircuitData` in the rust code. So `Parameter('a')`  was still in the `params` field on the first `PackedInstruction`/`CircuitInstruction` but the parameter table didn't have it anymore and said it was bound. I assume it was something around the `pop()` or `insert()` where there was an edge case in the tracking/reindexing of the parameter table on main. But this fixes it as part of your larger refactoring. We probably should add a test that does something like this to make sure we don't regress.

jakelishman:
Yeah, I'm pretty sure I know exactly where that bug had come in, and I remember moving a line of code that I imagine is what fixed it. In the insert code, I think the new data is written in _before_ we called untrack on it, so it untracked the wrong thing.

crates/circuit/src/parameter_table.rs
```diff
@@ -10,164 +10,395 @@
...
+impl From<ParameterTableError> for PyErr {
+    fn from(value: ParameterTableError) -> PyErr {
+        CircuitError::new_err(value.to_string())
+    }
+}
+
+/// A single use of a symbolic parameter.
+#[derive(Clone, Copy, Hash, PartialEq, Eq, Debug)]
+pub enum ParameterUse {
+    Index { instruction: usize, parameter: u32 },
```
Cryoris, line 43: Is there a reason for specifically choosing `u32` and `usize` here?

crates/circuit/src/parameter_table.rs
```diff
@@ -10,164 +10,395 @@
...
+        &mut self,
+        uuid: ParameterUuid,
+    ) -> Result<HashSet<ParameterUse>, ParameterTableError> {
+        let info = self
+            .by_uuid
+            .remove(&uuid)
+            .ok_or(ParameterTableError::ParameterNotTracked(uuid))?;
+        self.by_name
+            .remove(&info.name)
+            .expect("each parameter should be tracked by both UUID and name");
```
Cryoris, line 326: Very much optional, but it might be nice to add "but the name was not found" or something to give a better hint? But I don't know if this could ever be triggered anyways

crates/circuit/src/imports.rs
```diff
@@ -45,6 +45,8 @@ impl ImportOnceCell {
                 .unwrap()
                 .getattr(self.object)
                 .unwrap()
+                .downcast_into()
+                .unwrap()
```
jakelishman: Oh, this is totally unnecessary - it was accidentally left over from some playing I did with `ImportOnceCell` that I then (mostly) reverted.

```suggestion
```
```suggestion
                .downcast_into()
                .unwrap()
```

jakelishman:
I added a test that catches Steve's case in 1cd8d08.  It turned out to be a bit more complex than I'd originally thought - it's a particular interaction with the parametric global-phase tracking in addition to the pop/insert retracking required, and somewhere along the line, that was causing the mess-up.

crates/circuit/src/parameter_table.rs
```diff
@@ -10,164 +10,395 @@
...
+        self.vectors.clear();
+        self.order.clear();
+        self.py_parameters = None;
     }
 
-    fn set(&mut self, uuid: u128, name: String, param: PyObject, refs: ParamEntry) {
-        self.names.insert(name, uuid);
-        self.table.insert(uuid, refs);
-        self.uuid_map.insert(uuid, param);
+    /// Expose the trackec data for a given parameter as directly as possible to Python space.
```
Cryoris:
```suggestion
    /// Expose the tracker data for a given parameter as directly as possible to Python space.
```

crates/circuit/src/circuit_data.rs
```diff
@@ -430,6 +335,32 @@ impl CircuitData {
         self.clbits.len()
     }
 
+    /// Return the number of unbound compile-time symbolic parameters tracked by the circuit.
+    pub fn num_params(&self) -> usize {
```
Cryoris: This is the number of `Parameter` objects, right? Should we name this consistently with the current `QuantumCircuit.num_parameters`?
```suggestion
    pub fn num_parameters(&self) -> usize {
```

crates/circuit/src/circuit_data.rs
```diff
@@ -823,16 +744,40 @@ impl CircuitData {
...
+        let py = value.py();
         let new_index = self.data.len();
         let packed = self.pack(py, &value.borrow())?;
         self.data.push(packed);
-        self.update_param_table(py, new_index, params)
+        self.track_instruction_parameters(py, new_index)
+    }
+
+    /// Backup entry point for appending an instruction from Python space, in the unusual case that
+    /// one of the instruction parameters contains a cyclical reference to the circuit itself.
```
Cryoris, line 757: To make sure I understand: you mean something like this with cyclical?
```
circuit = # some circuit with free parameters
circuit.append(circuit.to_gate(), circuit.qubits)
```

crates/circuit/src/circuit_data.rs
```diff
@@ -1105,4 +1024,197 @@ impl CircuitData {
...
+                                    // Historically, `assign_parameters` called `validate_parameter`
+                                    // only when a `ParameterExpression` became fully bound.  Some
+                                    // "generalised" (or user) gates fail without this, though
+                                    // arguably, that's them indicating they shouldn't be allowed to
+                                    // be parametric.
+                                    //
+                                    // Our `bind_expr` coercion means that a non-parametric
+                                    // `ParameterExperssion` after binding would have been coerced
+                                    // to a numeric quantity already, so the match here is
+                                    // definitely paramterised.
```
Cryoris:
```suggestion
                                    // definitely parameterised.
```

Cryoris (commented):
One comment about function naming and otherwise only minuscule comments and questions 🙂

crates/circuit/src/parameter_table.rs
jakelishman: `instruction` indexes into `CircuitData::data`, which has no bounds on it.  `parameter` indexes into an instruction's `params` field, but the `Operation` trait we have in Rust space requires `num_params` to be a `u32`, so a `u32` is sufficient to index that.

On a 64-bit system, this wouldn't actually cause any different in size of the `Index` variant, because `usize` is 8 bytes and `u32` is 4, so alignment of the struct would require that it gets padded up to 16 bytes.  _However_, this is defined inline in an `enum`, so we also need to store the enum discriminant in that data as well, which is an extra bit of data.  By ensuring that there's at least a padding byte in all variants, we can enable a Rust compiler optimisation that stores the variant in what would otherwise be padding, so `ParameterUse` remains 16 bytes, rather than naively getting expanded to 24.

On 32-bit systems, `ParameterUse` is 12 bytes and we don't get the padding trick, but we don't care much about that.

crates/circuit/src/parameter_table.rs
jakelishman: This isn't a user-facing warning; it's triggering an internal panic if one of the core invariants of the `ParameterTable` is violated.  It _shouldn't_ be possible for it to occur at all, which is why we're using `Option::expect`, rather than mapping it into a Python-space error.

If this triggers, it'll be because a developer (maybe me) made a mistake, and the backtrace will point immediately back to this line of code.

crates/circuit/src/circuit_data.rs
jakelishman: This is actually most copied from the original source - I just rearranged the functions a bit.  Matt originally wrote it, and I think the original problem was somehow we had a test that had a control-flow operation that put the same circuit to be one of its blocks?? I don't entirely remember.

crates/circuit/src/circuit_data.rs
mtreinish: :us:
```suggestion
                                    // definitely parameterized.
```

crates/circuit/src/circuit_data.rs
mtreinish: Yeah, I don't remember exactly where it was happening, but there was a case with a control flow op was causing an error in the runtime borrow checking because the same `CircuitData` object was being passed in here.

crates/circuit/src/parameter_table.rs
Cryoris: Nice, thanks for explanation 😄

crates/circuit/src/circuit_data.rs
jakelishman: Yeah, yours is better and in keeping with the rest of the terminology in Rust space.  I'll update it everywhere.

crates/circuit/src/circuit_data.rs
jakelishman: I think as long as we're accepting general `Param` instances, I think it's a mistake not to.  At the moment, the way we use it makes it very unlikely for the `Param`s to be anything other than floats, but I don't think it'd be too hard to have someone call this on a circuit they got from Python space after applying some filter / join of the iterables or something.

The `ParameterTable` isn't a public field (and imo absolutely shouldn't be), so unless the typing of this function is changed so that it only accepts `f64` as the params, I think we need to do this.

The overhead shouldn't be much, since we can determine that it's a no-op without any Python-space calls.

crates/circuit/src/circuit_data.rs
jakelishman: Done in 8c554ba7e0.

crates/circuit/src/parameter_table.rs
jakelishman: Done in 7e9aa1c212.

crates/circuit/src/circuit_data.rs
jakelishman: Done in 7e9aa1c212.

jakelishman:
Ok, comments should be addressed, and the `AnnotatedOperation` problems should be fixed by 83da2f4662.

Fwiw they've been pre-existing in some form or another ever since `AnnotatedOperation` was first introduced.

crates/circuit/src/circuit_data.rs
```diff
@@ -1105,4 +1024,213 @@ impl CircuitData {
...
+                                                    .into_py_dict_bound(py),
+                                            ),
+                                        )?,
+                                    )?
+                                }
+                            };
+                            op.getattr(params_attr)?.set_item(parameter, new_param)?;
+                            let mut new_op = op.extract::<OperationFromPython>()?;
+                            previous.op = new_op.operation;
+                            previous.params_mut().swap_with_slice(&mut new_op.params);
```
mtreinish, line 1175: I didn't know about this method, I like it. It's uses are a bit niche, but when you need it like here it is handy.

crates/circuit/src/parameter_table.rs
```diff
@@ -10,164 +10,395 @@
...
-        slf.into()
+/// This implementation of `FromPyObject` is for the UUID itself, which is what the `ParameterUuid`
+/// struct actually encapsulates.
+impl<'py> FromPyObject<'py> for ParameterUuid {
+    fn extract_bound(ob: &Bound<'py, PyAny>) -> PyResult<Self> {
+        if ob.is_exact_instance(UUID.get_bound(ob.py())) {
+            ob.getattr(intern!(ob.py(), "int"))?.extract().map(Self)
+        } else {
+            Err(PyTypeError::new_err("not a UUID"))
+        }
```
mtreinish, line 93: We might want to make a note for the future that pyo3 has an open PR adding uuid conversion support for:

https://docs.rs/uuid/latest/uuid/

that might be something we want to leverage in the future here.

crates/circuit/src/parameter_table.rs
```diff
@@ -10,164 +10,395 @@
...
+        // If `order` is already populated, it's sorted; it's the responsibility of the methods of
+        // this struct that mutate it to invalidate the cache.
+        if !self.order.is_empty() {
+            return;
         }
-        self.table.insert(uuid, entry);
-        self.names.insert(name, uuid);
-        self.uuid_map.insert(uuid, parameter);
+        self.order.reserve(self.by_uuid.len());
+        self.order.extend(self.by_uuid.keys());
```
mtreinish, line 257: `extend()` doesn't do the allocation all at once for us?

crates/circuit/src/parameter_table.rs
```diff
@@ -10,164 +10,395 @@
...
+        self.order.sort_unstable_by_key(|uuid| {
```
mtreinish, line 258: I wonder if there is a threshold where rayon's `par_sort_unstable_by_key()` would be useful here. We can look at that in the future.

qiskit/circuit/quantumcircuit.py
```diff
@@ -4319,108 +4296,28 @@ def assign_parameters(  # pylint: disable=missing-raises-doc
...
-        # Normalize the inputs into simple abstract interfaces, so we've dispatched the "iteration"
-        # logic in one place at the start of the function.  This lets us do things like calculate
-        # and cache expensive properties for (e.g.) the sequence format only if they're used; for
-        # many large, close-to-hardware circuits, we won't need the extra handling for
-        # `global_phase` or recursive definition binding.
-        #
-        # During normalisation, be sure to reference 'parameters' and related things from 'self' not
-        # 'target' so we can take advantage of any caching we might be doing.
-        if isinstance(parameters, dict):
+        if isinstance(parameters, collections.abc.Mapping):
```
mtreinish, line 4308: Are there cases people are passing dict likes to `assign_parameters`? Also I assume you're not worried about the extra runtime overhead of the registration hooks of the collections types here.

mtreinish (approved):
This lgtm, I had a few questions/musings inline but nothing worth blocking over. Thanks for doing this, it's a really nice improvement both in the organization of the rust space data structures for dealing with parameters but also the runtime improvements.

crates/circuit/src/circuit_data.rs
jakelishman: Yeah, I learned about when writing this as well.  It _felt_ like a method that would exist in Rust, so I had a quick look.

crates/circuit/src/parameter_table.rs
jakelishman: I think it's possible that this is left over from some point when the internal data structures were different, and I was getting the keys from something that didn't necessarily have an accurate `size_hint`.  I don't entirely remember.

crates/circuit/src/parameter_table.rs
jakelishman: Quite possibly yes, unless it interferes with the fastest paths for data that's already sorted.  The cases we need this to have the highest performance correspond to IBM backends' fast parametric updates, where all the parameters are likely to be in the circuit in order to begin with.

qiskit/circuit/quantumcircuit.py
jakelishman: No, and in fact it would have been impossible before because of the type hint.  I'd loosened this because a) that matches the documentation and b) I think I was passing the `ParameterBindsDict` custom object recursively into subcalls at some point (though I think I changed the logic subsequently).

PR #12799, opened 2024-07-22
Fix oversights in `generate_preset_pass_manager`

ElePT:
### Summary
This PR fixes a couple of issues that should have been addressed in #12185. Found while working on #12762 and changes built on top of it (must be rebased).

### DIFF
--- a/qiskit/compiler/transpiler.py
+++ b/qiskit/compiler/transpiler.py
@@ -20,7 +20,6 @@ import warnings
 
 from qiskit import user_config
 from qiskit.circuit.quantumcircuit import QuantumCircuit
-from qiskit.circuit.quantumregister import Qubit
 from qiskit.dagcircuit import DAGCircuit
 from qiskit.providers.backend import Backend
 from qiskit.providers.backend_compat import BackendV2Converter
@@ -339,10 +338,7 @@ def transpile(  # pylint: disable=too-many-return-statements
         if translation_method is None and hasattr(backend, "get_translation_stage_plugin"):
             translation_method = backend.get_translation_stage_plugin()
 
-    initial_layout = _parse_initial_layout(initial_layout)
-    approximation_degree = _parse_approximation_degree(approximation_degree)
     output_name = _parse_output_name(output_name, circuits)
-
     coupling_map = _parse_coupling_map(coupling_map)
     _check_circuits_coupling_map(circuits, coupling_map, backend)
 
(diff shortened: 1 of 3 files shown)


releasenotes/notes/add-generate-preset-pm-global-import-efb12f185f3f738b.yaml
```diff
@@ -0,0 +1,20 @@
...
+    
+    Starting in Qiskit 1.2, the function is also importable from the :mod:`qiskit.transpiler` module as::
+    
+      from qiskit.transpiler import generate_preset_pass_manager
+
+fixes:
+  - |
+    Fixed input-handling inconsistencies between :func:`.transpile` and :func:`.generate_preset_pass_manager` that
+    would lead to different transpilation outputs for the same inputs, or :func:`.generate_preset_pass_manager`
+    failing for certain input combinations accepted by :func:`.transpile`.
```
ElePT:
```suggestion
```

qiskit/compiler/transpiler.py
```diff
@@ -339,10 +338,7 @@ def callback_func(**kwargs):
         if translation_method is None and hasattr(backend, "get_translation_stage_plugin"):
             translation_method = backend.get_translation_stage_plugin()
 
-    initial_layout = _parse_initial_layout(initial_layout)
-    approximation_degree = _parse_approximation_degree(approximation_degree)
```
ElePT, line 343: Before this PR, setting a custom `initial_layout` or `approximation_degree` could lead to different results on `generate_preset_pass_manager` vs `transpile`, while the rest of inputs would be consistent.

releasenotes/notes/use-target-in-generate-preset-pm-5215e00d22d0205c.yaml
```diff
@@ -12,3 +12,9 @@ upgrade_transpiler:
     If a `backend` input of type :class:`.BackendV1` is provided, it will be 
     converted to :class:`.BackendV2` to expose its :class:`.Target`. This change does 
     not require any user action.
+
+fixes:
+  - |
+    A series of input-handling inconsistencies between :func:`.transpile` and :func:`.generate_preset_pass_manager` 
+    have been fixed. These inconsistencies would lead to different transpilation outputs for the same inputs, 
+    or :func:`.generate_preset_pass_manager` failing for certain input combinations accepted by :func:`.transpile`.
```
ElePT, line 20: Most of the inconsistencies were fixed in the previous PR but we didn't have any reno to refer to. Given that there have been issues about these inconsistencies, I think it would be good to mention that they will no longer happen (but we can adjust the wording).

mtreinish (approved):
LGTM, this is straightforward enough and inline with the other changes made.

Issue #12765, opened 2024-07-11
`BitArray.from_samples/from_counts` fails with `num_bits=None` for str outcomes and 0 outcome.


chriseclectic:
### Environment

- **Qiskit version**: 1.1
- **Python version**: 3.11
- **Operating system**: MacOS

### What is happening?

Calling `BitArray.from_counts` or `BitArray.from_samples` without specifying the number of bits is supposed to infer the number of bits from the input data, however this currently only seems to work for integer keys > 0.

### How can we reproduce the issue?

`BitArray.from_samples([s])` or `BitArray.from_counts({s: shots})`

errors with
```
File ~/mambaforge/envs/qiskit1/lib/python3.11/site-packages/qiskit/primitives/containers/bit_array.py:305, in BitArray.from_samples(samples, num_bits)
    303 data = b"".join(val.to_bytes(num_bytes, "big") for val in ints)
    304 array = np.frombuffer(data, dtype=np.uint8, count=len(data))
--> 305 return BitArray(array.reshape(-1, num_bytes), num_bits)

ValueError: cannot reshape array of size 0 into shape (0)
```
for `s = 0` or a string, eg `s = "000"`.

Calling `BitArray.from_counts` ultimately ends up calling `BitArray.from_samples` and returns the same error.

### What should happen?

The number of bits should be inferred correctly for bitstring dicts, or for trivial case of `{0: shots}` (which should default to 1 bit)

aeddins-ibm:
If I understand correctly, there are two issues: 
- enforcing the convention that 1 bit is needed to represent `0` (seems straightforward)
- inferring `num_bits` from the bitstring lengths (less clear)

Currently there are unit tests requiring that `from_samples` runs even when given bitstrings with different numbers of bits:

https://github.com/Qiskit/qiskit/blob/ec1f42cd785805264b86445ac92bf2a22f10bf92/test/python/primitives/containers/test_bit_array.py#L241-L247

I think this means that to infer `num_bits` from the bitstring key lengths, we have to check each key length individually and take the maximum, or else drop support for this case of inconsistent bitstring lengths (so we can just check the length of the first key).

Quick test on laptop suggests for 10^8 bitstrings with up to 1000 bits, `max(map(len, samples))` takes ~15-20 seconds.

Since checking all the key lengths seems slow, I would rather remove support for inputs containing bitstrings of inconsistent lengths. Is it safe to remove, or is there an actual use case for this / are we obligated to continue supporting it?

[...]

PR #12800, opened 2024-07-22
Fix `BitArray.from_counts`/`from_samples` to not fail for input with only `0` outcome and `num_bits=None`

aeddins-ibm:
### Summary
Partial fix for #12765.

That issue points out two problems with `BitArray.from_counts` / `BitArray.from_samples`. The more severe problem is that these methods fail with an error if the input (e.g. a counts dictionary) contains only the `0` outcome, and `num_bits` is left at the default value of `None`. 

The methods try to infer `num_bits` as the number of bits needed to represent the largest outcome (e.g. bitstring) contained in the input data. If the largest outcome is `0` when converted to an integer (e.g. the bitstring `"0"`), the number of bits inferred is `0`, which can be surprising, and also leads to an error.

This small PR interprets `0` as represented by 1 bit instead of by 0 bits, the suggested correct behavior in #12765.

The less severe problem, not fixed here, is that currently any information about `num_bits` present in the input bitstring lengths gets ignored. That problem seemed less trivial to me. I put some thoughts [here](https://github.com/Qiskit/qiskit/issues/12765#issuecomment-2240676887).

### DIFF
--- a/qiskit/primitives/containers/bit_array.py
+++ b/qiskit/primitives/containers/bit_array.py
@@ -230,7 +230,7 @@ class BitArray(ShapedMixin):
         Args:
             counts: One or more counts-like mappings with the same number of shots.
             num_bits: The desired number of bits per shot. If unset, the biggest value found sets
-                this value.
+                this value, with a minimum of one bit.
 
         Returns:
             A new bit array with shape ``()`` for single input counts, or ``(N,)`` for an iterable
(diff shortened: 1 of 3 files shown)


ElePT:
Hi @aeddins-ibm, could you add a short release note documenting the fix?

ElePT (approved):
This fix looks straightforward, I think we should include it in 1.2 and backport it to 1.1

ElePT:
https://github.com/Mergifyio backport stable/1.1 stable/1.2

t-imamichi (approved):
LGTM

Issue #12552, opened 2024-06-12
deprecate StochasticSwap 


1ucian0:
`StochasticSwap` is currently use in optmization level 0 and superseded by SabreSwap.
 - [x] remove from level 0 in 1.2 https://github.com/Qiskit/qiskit/pull/12806
 - [ ] deprecate in 1.2/1.3
 - [ ] remove it in 2.0
 - [ ]  close dangling issues, like https://github.com/Qiskit/qiskit/issues/4094

weechinghwa:
Hi, I am interested in taking up this issue.

1ucian0:
This is more of an epic than an issue (tagging as such). Would you like to help with any of the items? I can take `remove from level 0 in 1.2` because it is time sensitive.

weechinghwa:
I am interested in `deprecate 1.2/1.3`, but this is rather new to me. 
Could you provide some keywords or guides to help me understand it thoroughly from A to Z.

1ucian0:
> Could you provide some keywords or guides to help me understand it thoroughly from A to Z.

I think a good starting point is https://github.com/Qiskit/qiskit/blob/main/CONTRIBUTING.md#adding-deprecation-warnings

Feel free to create a draft PR soon, so we can discuss it more concretely.

PR #12806, opened 2024-07-23
Change the default `routing_method` to `sabre`

1ucian0:
In order to get ready [for deprecating `StochasticSwap`](https://github.com/Qiskit/qiskit/issues/12552) the only default pipeline that uses is level 0. Changing it to `sabre`.

### DIFF
--- a/releasenotes/notes/sabre_level0-1524f01965257f3f.yaml
+++ b/releasenotes/notes/sabre_level0-1524f01965257f3f.yaml
@@ -1,5 +1,13 @@
 ---
-features_transpiler:
+upgrade_transpiler:
   - |
-    The default `routing_method` for optimization level 0 was changed from ``stochastic`` to ``sabre``.
-    The pass :class:`.SabreSwap` fully supersedes :class:`.StochasticSwap` and the latter will be deprecated in a future release.
+    The default routing pass used by optimization level 0 for :func:`.generate_preset_pass_manager`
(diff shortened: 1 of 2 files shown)


releasenotes/notes/sabre_level0-1524f01965257f3f.yaml
```diff
@@ -0,0 +1,5 @@
+---
+features_transpiler:
+  - |
+    The default `routing_method` for optimization level 0 was changed from ``stochastic`` to ``sabre``.
+    The pass :class:`.SabreSwap` fully supersedes :class:`.StochasticSwap` and the latter will be deprecated in a future release.
```
mtreinish:
```suggestion
upgrade_transpiler:
  - |
    The default routing pass used by optimization level 0 for :func:`.generate_preset_pass_manager`
    and :func:`.transpile` has been changed from :class:`.StochasticSwap` to :class:`.SabreSwap`.
    The :class:`.SabreSwap` pass performs exactly the same function but performs better in both
    runtime and output quality (in number of swap gates and depth) compared to
    :class:`.StochasticSwap`. For ``optimization_level=0`` this shouldn't matter because it's not
    expected to run routing for the typical use case of level 0.
       
    If you were relying on the previous default routing algorithm for any reason you can use the
    ``routing_method`` argument for :func:`.transpile` and :func:`.generate_preset_pass_manager`
    to ``"stochastic"`` to use the :class:`.StochasticSwap` pass.
```

mtreinish (approved):

PR #12740, opened 2024-07-08
Use Rust gates for 2q unitary synthesis

mtreinish:
### Summary

This commit builds off of what #12650 did for the 1q decomposer and moves to using rust gates for the 2q decomposer too. This means that the circuit sequence generation is using rust's StandardGate representation directly instead of relying on mapping strings. For places where circuits are generated (calling `TwoQubitWeylDecomposition.circuit()` or or `TwoQubitBasisDecomposer.__call__` without the `use_dag` flag) the entire circuit is generated in Rust and returned to Python.

### Details and comments

PR #12809, opened 2024-07-24
Directly construct CircuitData in TwoQubitWeylDecomposition

mtreinish:
### Summary

In the recently merged #12740 a path was added for constructing the `CircuitData` in rust when synthesizing to a `QuantumCircuit` directly. However, in that PR this was done through a layer of indirection by first collecting the gates into an intermediate `Vec` and then passing an iterator of that into
`CircuitData::from_standard_gates()`. However this resulted in an unecessary allocation for two `Vec`s and it would have been better to just directly construct the `CircuitData` object directly. However, to accomplish this we needed more rust space methods for creating and manipulating a `CircuitData` object as it's mostly being constructed from Python space or using `CircuitData::from_standard_gates()` with an iterator so far. This commit makes those additions and then updates the `TwoQubitWeylDecomposition` code to directly construct the `CircuitData` object instead of using an intermediate `Vec`.

### Details and comments


crates/circuit/src/circuit_data.rs
```diff
@@ -164,6 +150,58 @@ impl CircuitData {
...
+    /// Append a standard gate to this CircuitData
+    pub fn push_standard_gate(
+        &mut self,
+        operation: StandardGate,
+        params: SmallVec<[Param; 3]>,
+        qargs: SmallVec<[Qubit; 2]>,
+    ) -> PyResult<()> {
+        let no_clbit_index = (&mut self.cargs_interner)
+            .intern(InternerKey::Value(Vec::new()))?
+            .index;
```
mtreinish, line 206: I'm not super happy with this, it feels like unnecessary overhead on every call. It's why I didn't update `from_standard_gates()` to call this new method because it does this outside the loop. It feels like maybe we should have an implicit empty index that's a constant in the interner so we can avoid this.

crates/circuit/src/circuit_data.rs
```diff
@@ -96,7 +96,7 @@ pub struct CircuitData {
     clbits: BitData<Clbit>,
     param_table: ParamTable,
     #[pyo3(get)]
-    global_phase: Param,
+    pub global_phase: Param,
```
jakelishman: At least from #12794 onwards (and probably now anyway), this needs to be a private attribute - it's not safe to directly set it in general, because you always have to remember to untrack any `Parameter` uses within the global phase and track any new ones (not there are any here).

In #12794 there's a setter `CircuitData::set_global_phase` for it.

crates/circuit/src/circuit_data.rs
```diff
@@ -164,6 +150,58 @@ impl CircuitData {
         Ok(res)
     }
 
+    /// Build an empty CircuitData object with an initially allocated instruction capacity
+    pub fn with_capacity(
+        py: Python,
+        num_qubits: u32,
+        instruction_capacity: usize,
+        global_phase: Param,
+    ) -> PyResult<Self> {
```
jakelishman: If we're going to have a `with_capacity` setter (which we should), we probably ought to be general and have it accept a number of clbits as well.

mtreinish:
I've run a quick little benchmark script:
```python
import statistics
import time

from qiskit.synthesis.two_qubit import TwoQubitWeylDecomposition
from qiskit.circuit.library import CXGate

unitaries = [random_unitary(4) for _ in range(10000)]
runtimes = []
for mat in unitaries:
    start = time.perf_counter()
    TwoQubitWeylDecomposition(mat).circuit()
    stop = time.perf_counter()
    runtimes.append(stop - start)
print(f"Mean runtime per unitary: {statistics.geometric_mean(runtimes)} sec.")
print(f"Total runtime for 10k unitary matrices: {sum(runtimes)} sec.")
```

to compare the performance with and without this PR. There is a consistent small speedup I'm seeing with this PR applied. With this PR:
```
Mean runtime per unitary: 2.8342069525664723e-05 sec.
Total runtime for 10k unitary matrices: 0.3291525247041136 sec.
```

on current main:

```
Mean runtime per unitary: 3.0709930013551664e-05 sec.
Total runtime for 10k unitary matrices: 0.3557815661188215 sec.
```

There is some fluctuation in the numbers obviously especially with fully random unitaries, but I haven't seen a case where the PR is slower (or the same speed) than main. Not that `TwoQubitWeylDecomposition.circuit()` is performance critical at all, but I wanted to put some numbers behind why we would want to do this.

crates/circuit/src/circuit_data.rs
```diff
@@ -164,6 +150,58 @@ impl CircuitData {
...
+            qubits,
+            clbits: no_clbit_index,
+            params,
+            extra_attrs: None,
+            #[cfg(feature = "cache_pygates")]
+            py_op: RefCell::new(None),
+        });
+        Ok(())
+    }
+
```
jakelishman: I don't think we need to take `params` and `qargs` as particular owned types here, and especially not these two owned types. Taking `&[Param]` and `&[Qubit]` would let us use any type that makes sense in the calling functions, including direct on-stack slices like `&[Qubit(0)]` rather than `smallvec![Qubit(0)]`, etc (though the smallvec would still be fine).

That's definitely for `qargs`.  For `params`, there's imo an argument for accepting `Option<Box<SmallVec<[Param; 3]>>>` directly, since that then allows the caller to potentially avoid an allocation and move.

crates/circuit/src/circuit_data.rs
mtreinish: Lets wait for #12794 then, this PR isn't critical

crates/circuit/src/circuit_data.rs
mtreinish: Done in: https://github.com/Qiskit/qiskit/pull/12809/commits/6e6cb4f68ea3f03a2fc26f34e0336e6aa3b2c569

crates/circuit/src/circuit_data.rs
mtreinish: Done in: https://github.com/Qiskit/qiskit/pull/12809/commits/d5c96937d55687a57baeabc52d23fe12c0631add

crates/circuit/src/circuit_data.rs
mtreinish: Done in: https://github.com/Qiskit/qiskit/pull/12809/commits/cfe878d0a7f452b3a6a8ea76e053107945f72444

crates/circuit/src/circuit_data.rs
jakelishman: I have a change to the interners I want to make once the `DAGCircuit` PR merges that will avoid the need for the `Vec` here in favour of the static empty slice (which is at least a small saving), but I think the idea of the special "intern key to the 'zero' object" is a good one too. I can see it coming up a lot, and I think it's doable without losing any particular generality in the interners - we can make it like
```rust
pub struct Interner<T> { ... };
pub struct InternKey<T> {
   index: u32,
   _phantom: PhantomData<T>,
};
impl<T: Default> InternKey<T> {
    pub fn empty() -> Self {
        Self {
            val: 0,
            _phantom: PhantomData,
        }
    }
}
```
or something like that, and arrange that the 0 index of `Interner<T: Default>` always corresponds to `<T as Default>::default()`.

jakelishman (approved):
This seems like a good idea to me, thanks for the updates.

PR #12812, opened 2024-07-24
Leverage rustworkx node and edge count hints in transpiler

mtreinish:
### Summary

This commit updates the DAGCircuit construction in the transpiler to use the newly introduced node_count_hint and edge_count_hint kwargs on the rustworkx graph constructors. In situations where we know the size of the dag or at least the lower bound we can reduce the number of underlying memory allocations by preallocating that size when we create the inner graph object. This commit adds private arguments to the DAGCircuit constructor which are used to provide these size hints to rustworkx.

### Details and comments


qiskit/transpiler/passes/layout/disjoint_utils.py
```diff
@@ -111,7 +111,10 @@ def split_barriers(dag: DAGCircuit):
             barrier_uuid = f"{node.op.label}_uuid={uuid.uuid4()}"
         else:
             barrier_uuid = f"_none_uuid={uuid.uuid4()}"
-        split_dag = DAGCircuit()
+        split_dag = DAGCircuit(
+            _node_count_hint=3 * num_qubits,
+            _edge_count_hint=2 * num_qubits,
```
ElePT: Usual question for me, how did you reach a node count (hint) of `3*num_qubits` here?

ElePT (commented):
The use of count hints looks straightforward to me. I would have to look into the passes better to fully get where some of the counts are coming from, but I guess that there isn't any big issue if the tests pass.

qiskit/transpiler/passes/layout/disjoint_utils.py
mtreinish: The dagcircuit being built here is taking a n qubit barrier and splitting it into n 1 qubit barriers so we can compute the connected components of the circuit. So if it's a 3q barrier it would looks something like:

dag_barrier

So there are 3 nodes for every qubit and 2 edges.

qiskit/dagcircuit/dagcircuit.py
```diff
@@ -73,7 +73,7 @@ class DAGCircuit:
 
     # pylint: disable=invalid-name
 
-    def __init__(self):
+    def __init__(self, _node_count_hint=0, _edge_count_hint=0):
```
jakelishman: Is there any harm in making this public?  I don't really like having "private" arguments unless we really need them, and there's no real safety concerns with these.

mtreinish: Not really, I was mostly worried about making it public because I pushed this up after the feature freeze and that would make it a feature. But I think this makes a straightforward exception to the process since it's so simple.

jakelishman: Yeah, I'm fine with allowing it, and I'd rather just make the things public than have awkward "private" arguments show up in the documentation.

mtreinish: Made public in: https://github.com/Qiskit/qiskit/pull/12812/commits/eb91a48c2e3478577cf0f3494f23ee1397fe9384

qiskit/dagcircuit/dagcircuit.py
```diff
@@ -686,7 +699,10 @@ def copy_empty_like(self, *, vars_mode: _VarsMode = "alike"):
...
             DAGCircuit: An empty copy of self.
         """
-        target_dag = DAGCircuit()
+        target_dag = DAGCircuit(
+            node_count_hint=self._multi_graph.num_nodes(),
+            edge_count_hint=self._multi_graph.num_edges(),
+        )
         target_dag.name = self.name
         target_dag._global_phase = self._global_phase
         target_dag.duration = self.duration
```
jakelishman, line 708: Should `copy_empty_like` eagerly allocate to the full size? Feels like that could be overallocating a bit.  Maybe we want to control it with a single flag?

qiskit/transpiler/passes/layout/disjoint_utils.py
```diff
@@ -111,7 +111,10 @@ def split_barriers(dag: DAGCircuit):
             barrier_uuid = f"{node.op.label}_uuid={uuid.uuid4()}"
         else:
             barrier_uuid = f"_none_uuid={uuid.uuid4()}"
-        split_dag = DAGCircuit()
+        split_dag = DAGCircuit(
+            node_count_hint=3 * num_qubits,
+            edge_count_hint=2 * num_qubits,
+        )
```
jakelishman, line 117: Is this just a guess, or is there a reason to it?  Maybe not worth bothering if it's a guess.

qiskit/transpiler/passes/layout/sabre_layout.py
```diff
@@ -303,7 +303,18 @@ def run(self, dag):
...
+        for component in components:
+            component_dag = component.dag
+            size = len(component_dag._multi_graph) - 2 * len(component_dag._wires)
+            num_nodes += size
+            num_edges += component_dag._multi_graph.num_edges()
+
+        mapped_dag = DAGCircuit(
+            node_count_hint=num_nodes,
+            edge_count_hint=num_edges,
+        )
```
jakelishman, line 317: I'm suspicious that the Python-space time to calculate this stuff is quite possibly similar to the Rust-space time to just grow the vectors.

jakelishman (commented):
In general this seems like a good idea.  There's a few places where the amount of calculation / private-method accessing we're doing is sufficiently high that it might not be worthwhile in Python space, but I don't feel strongly.

The hints in `copy_empty_like` feel a bit more awkward - they make sense in in the context of `copy` or if the DAG will be refilled, but for things like `layers`, the overallocation feels like it could get quite out-of-hand.

qiskit/transpiler/passes/layout/disjoint_utils.py
mtreinish: See my answer to Elena's earlier question about this: https://github.com/Qiskit/qiskit/pull/12812#discussion_r1691320571

jakelishman: Whoops, sorry.

qiskit/transpiler/passes/layout/sabre_layout.py
mtreinish: Yeah, I was worried about this too. In practice I think the list is only ever a single list entry here though. The code handles disjoint connectivity, but iirc it was unsound to do the routing split up so we bail return an unrouted circuit before this gets called if the number of components is >1.

jakelishman: I'm fine to just leave it anyway - it's probably just going to be a wash at worst, rather than hurting.

mtreinish:
I'm going to close this as it's no longer relevant since we've moved the `DAGCircuit` to rust. There might still be places we want to do this from python but the implementation will have to be added as part of the rust implementation of the dag now.

Issue #12792, opened 2024-07-19
Light Cone Transpiler Pass


MarcDrudis:
### What should we add?

I think it would be a good idea to add a Transpiler pass that computes the light cone of an observable. For those who are not familiar, a light cone is the set of gates in a circuit that affect the measurement of a subset of the qubits. For example, if you have a shallow EfficientSU2 ansatz with pairwise entanglement on 127 qubits and you are trying to measure an expectation value on a single qubit you can typically reduce the circuit to a 20 qubit circuit (just making up the number, in practice it depends). I use it a lot in my research and I know of other researchers that do it as well. We all typically have a naive implementation of it, but it would be a powerful tool to have one that handles edge cases directly implemented into Qiskit.

I can take care of this issue as I already have code that does it. Before tacking this problem, however, it would be wise to first implement the proposal in #12790 to allow for computation of light cones also in parameterized circuits.

spiccinelli:
I would be interested in participating and happy to help 😄

Cryoris:
This sounds like a nice addition. Though the main motivation might come from classically simulating the circuit, this could  be useful when compiling circuits for hardware as we could simplify the compilation problem. Reducing the circuit size could also potentially allow for better routing, but also requires some care (e.g. regarding the user-provided initial layout). 

If we'd select the "active" qubits by the presence of the measurement we might need to take special care of mid-circuit measurements. For example, we might only care about the final measurements and could allow removing mid-circuit measurements if they are part of a control-flow that doesn't affect the target measurements. But we can discuss these details at a later stage.

alexanderivrii:
I think Andrew Eddins (@aeddins-ibm) already has a pass that does exactly this.

aeddins-ibm:
Hi @alexanderivrii, thanks for pointing that out. @Cryoris reached out to me separately to coordinate, and I believe we are on the same page that my rough-draft code may be used as a starting point if that turns out to be helpful.

Cryoris:
Closed by #12814.

Issue #12790, opened 2024-07-19
Add parameterized gates to the `CommutationChecker`


MarcDrudis:
### What should we add?

I realized that currently the `CommutationChecker` class can not handle Pauli Rotations if they are parameterized. In many cases our circuits are paramterized and the commutation relationships for this rotations are just the same ones as the ones for their generators. I propose to add this functionality into Qiskit. There are 2 ways to go about it. One would be to make the `standard_gate_commuation` dictionary bigger or to map the rotation gates to their generators.

I already have already implemented it for myself because I need it for my research so I would be happy to implement it for Qiskit.

jakelishman:
I believe your #13302 solved this, but feel free to re-open if there was more to go.

aeddins-ibm:
Are `Pauli` gates currently supported by the commutation checker? (they can have e.g. `'ZZ'` as a parameter)

On 1.3.0rc2, I'm getting `QiskitError: 'Unable to hash a non-float instruction parameter.'`, but I'm also doing some hacky stuff so the problem could be unique to my end.

PR #12814, opened 2024-07-25
Light Cone Transpiler Pass

MarcDrudis:
### Summary
When measuring or sampling a subset of the qubits of our circuit there will be some gates in our circuit that will not affect the outcome of the measurement. One can thus compute the causal light cone of the circuit and measurement, i.e. the set of gates that does affect the measurement. This transpiler pass provides an easy tool to reduce the number of gates in our circuit provided a Pauli observable to be measured or  a subset of measurements in the Z basis.
This PR adresses  #12792 

### Details and comments
This code takes inspiration from an implementation done by @aeddins-ibm.
Before finishing the unit-tests for this PR it would be interesting to solve #12790 .


qiskit/transpiler/passes/optimization/light_cone.py
```diff
@@ -0,0 +1,141 @@
...
+            of measurements in the circuit.
+            If a Pauli operator is specified, the lightcone will correspond to
+            the reduced circuit with the
+            same expectation value for the observable.
+        """
+        super().__init__()
+        self.observable = observable
+
+    @staticmethod
+    def _find_measurement_qubits(dag: DAGCircuit):
```
Cryoris: We can do this more efficiently if we only consider final measurements, see how it's done in the `RemoveFinalMeasurements` pass 🙂

qiskit/transpiler/passes/optimization/light_cone.py
```diff
@@ -0,0 +1,141 @@
...
+
+class LightCone(TransformationPass):
+    """Remove the gates that do not affect the outcome of a measurement on a circuit.
+
+    Pass for computing the light-cone of an observable or measurement. The Pass can
+    handle either a Pauli operator one would like to measure or a measurement on a set
+    of qubits.
+    """
+
+    def __init__(self, observable: Pauli | None = None):
```
Cryoris: This also takes a list of Paulis, right?

qiskit/transpiler/passes/optimization/light_cone.py
```diff
@@ -0,0 +1,141 @@
...
+        """
+        LightCone initializer.
+
+        Args:
+            observable: If None the lightcone will be computed for the set
+            of measurements in the circuit.
+            If a Pauli operator is specified, the lightcone will correspond to
+            the reduced circuit with the
+            same expectation value for the observable.
```
Cryoris:
```suggestion
        """
        Args:
            observable: If None the lightcone will be computed for the set
                of measurements in the circuit. If a Pauli operator is specified, 
                the lightcone will correspond to the reduced circuit with the
                same expectation value for the observable.
```

qiskit/transpiler/passes/optimization/light_cone.py
```diff
@@ -0,0 +1,141 @@
...
+            light_cone = qubits_measured
+            light_cone_ops = [(PauliGate("Z"), [qubit_index]) for qubit_index in qubits_measured]
+        else:
+            pauli_string = str(self.observable)
+            light_cone = [dag.qubits[i] for i, p in enumerate(pauli_string) if p != "I"]
+            light_cone_ops = [(PauliGate(pauli_string.replace("I", "")), light_cone)]
+
+        return set(light_cone), light_cone_ops
+
+    def run(self, dag: DAGCircuit):
```
Cryoris: We should not modify the input `dag`. It would probably be most efficient to just rebuild the dag, we can use `apply_operation_front` (since we iterate in reverse) for operations inside the light cone.

qiskit/transpiler/passes/optimization/light_cone.py
```diff
@@ -0,0 +1,141 @@
...
+        return dag
+
+
+if __name__ == "__main__":
+    """This main will be only for testing purposes and will be removed
+    before merging the PR."""
+    from qiskit.circuit.library import RealAmplitudes
+    from qiskit.converters import circuit_to_dag
+
+    qc = RealAmplitudes(10, entanglement="pairwise").decompose()
```
Cryoris: That's a good start for tests, you can just create a new file in `test/python/transpiler` that contains tests for this. Beyond testing that it works, we should also ensure failures are correctly handled, such as no measurements and no `observable`, an `observable` with mismatching size to the dag, checking mid-circuit measurements, ...

Cryoris (commented):
Looks like a good start! Here's some initial comments 🙂

qiskit/transpiler/passes/optimization/light_cone.py
MarcDrudis: For now only one Pauli. We could easily adapt it to a list of Paulis if we think that would make more sense.

qiskit/transpiler/passes/optimization/light_cone.py
MarcDrudis: I thought about it. `RemoveFinalMeasurements` has a function `_calc_final_ops` that would be ideal to reuse. Can I change the class `RemoveFinalMeasurements` to have that function outside of the class so that I can call it from this other class or would it be better to copy-paste the function into `LightCone`?

qiskit/transpiler/passes/optimization/light_cone.py
MarcDrudis: Okay

qiskit/transpiler/passes/optimization/light_cone.py
MarcDrudis: Great!

qiskit/transpiler/passes/optimization/light_cone.py
Cryoris: We could just make the method public, it seems generally useful. In this case we should give it a more precise name, e.g. something like `final_measurement_nodes`, and add a short test for it.

qiskit/transpiler/passes/optimization/light_cone.py
Cryoris: I'm asking because in your code the default path for `observable=None` seems to use a list of Paulis already 🙂

nonhermitian: A list of Pauli terms would, in general, lead to a collection of different circuits as each Pauli term may have its own independent causal structure.  A one circuit in, many circuits out, is not currently supported in the transpiler interface.  But perhaps should be as it is useful for twirling as well

Cryoris: Oh I meant that in this case the gates would have to commute with each of the Paulis in the list (how the measurements are treated as list of single-Z Paulis in the code right now). That would e.g. allow to decompose a larger observable into smaller (e.g. `ZZZ...` into `Z0, Z1, ...`) -- this would give a larger light cone since now things might not commute which would in the bigger observable, but it would make the commutation check faster (which I think would use a matrix multiplication check for non-standard gates).

But that's probably not required for the beginning and we could just issue a warning if the commutation check becomes problematic 🙂

qiskit/transpiler/passes/optimization/light_cone.py
```diff
@@ -0,0 +1,141 @@
...
+    ) -> tuple[set[int], tuple[Instruction, list[int]]]:
+        """Returns the initial lightcone.
+        If obervable is None, the lightcone is a different Z operator in each one of the
+        qubits with a measurement. If a Pauli observable is defined, that will be
+        the first gate in the circuit.
+        """
+        if self.observable is None:
+            qubits_measured = self._find_measurement_qubits(dag)
+            light_cone = qubits_measured
+            light_cone_ops = [(PauliGate("Z"), [qubit_index]) for qubit_index in qubits_measured]
```
Cryoris: It's likely better to use `ZGate` here so we can access the pre-computed commutations of the commutation checker. I don't think these are defined for the PauliGate.
```suggestion
            light_cone_ops = [(ZGate(), [qubit_index]) for qubit_index in qubits_measured]
```

qiskit/transpiler/passes/optimization/light_cone.py
```diff
@@ -0,0 +1,141 @@
...
+        print(light_cone, light_cone_ops)
+        # iterate over the nodes in reverse topological order
+        for node in reversed(list(dag.topological_op_nodes())):
+            if not light_cone.intersection(node.qargs):
+                dag.remove_op_node(node)
+                continue
+
+            commutes_bool = True
+            for op in light_cone_ops:
+                commute_bool = commutator.commute(op[0], op[1], [], node.op, node.qargs, [])
```
Cryoris: We should handle the `max_num_qubits` arg, which currently defaults to 3. E.g. if a user would give a `PauliGate("ZZZZ")` I think the code would just forego the commutation check and return `False` for any commutation.

So we should probably set `max_num_qubits=max(len(op[1]), len(node.qargs))` and add a warning if that number becomes too big.

qiskit/transpiler/passes/optimization/light_cone.py
```diff
@@ -0,0 +1,141 @@
...
+        qubits with a measurement. If a Pauli observable is defined, that will be
+        the first gate in the circuit.
+        """
+        if self.observable is None:
+            qubits_measured = self._find_measurement_qubits(dag)
+            light_cone = qubits_measured
+            light_cone_ops = [(PauliGate("Z"), [qubit_index]) for qubit_index in qubits_measured]
+        else:
+            pauli_string = str(self.observable)
+            light_cone = [dag.qubits[i] for i, p in enumerate(pauli_string) if p != "I"]
```
aeddins-ibm: The string representation of a Pauli in Qiskit puts qubit-0 at the right end of the string*. So I believe this list comprehension incorrectly reverses the observable and must be fixed.

Example of the problem here. Observable has `Z` on qubit 1 (bottom qubit in drawing), which does NOT commute with the CX gate, but the indexing problem puts `Z` on the wrong qubit so the transpiler pass concludes that it DOES commute with the CX gate:
image
(I suppressed a debug print statement in the code). Giving the reversed observable `Pauli('IZ')` yields a circuit that does still include the CNOT (but it should be an empty circuit).

Since this kind of bug is extremely common when working with `str` representations of Pauli strings, in general I **strongly** suggest not working with the `str` representation of Paulis unless/until you absolutely have to. A safer approach is to index the `Pauli` object directly: 
```
pauli = Pauli('XYZ')
bit_idx = 2
pauli[bit_idx] ## this returns Pauli('X'), which is correct
str(pauli)[bit_idx] ## this returns 'Z', which in this context would be incorrect
```
Indexing the Pauli directly interfaces with the numpy arrays backing the Pauli object, so in addition to being safer it's also often faster than working with the string representation.

\* - Often this is called being "little-endian", though I'm told that technically that term is reserved for byte-ordering, not bit-ordering, so "MSb0" or "MSbF" might be more appropriate, not sure.

test/python/circuit/test_commutation_checker.py
Cryoris, line 1: These changes are probably by accident (and seem to cause the tests to fail)

test/python/transpiler/test_light_cone.py
```diff
@@ -0,0 +1,294 @@
...
+        pm = PassManager([LightCone(Pauli("Z" + "I" * (num_qubits - 1)))])
+        new_circuit = pm.run(qc)
+
+        self.assertEqual(new_circuit.num_parameters, 11)
+
+        qr = QuantumRegister(num_qubits)
+        expected = QuantumCircuit(qr)
+        theta = ParameterVector("θ", num_qubits**2)
+
+        expected.ry(theta[0], qr[0])
```
Cryoris: Would it be simpler (and less error prone to type) to choose more specific, smaller (i.e. shallower/less ops) circuits to test all edge cases? 🙂

test/python/transpiler/test_light_cone.py
```diff
@@ -0,0 +1,294 @@
...
+        expected.cx(qr[2], qr[1])
+        expected.cx(qr[3], qr[4])
+        expected.cx(qr[1], qr[0])
+        expected.cx(qr[2], qr[3])
+        expected.cx(qr[2], qr[1])
+        expected.cx(qr[2], qr[1])
+
+        new_circuit = pm.run(qc)
+
+        self.assertEqual(Operator(expected), Operator(new_circuit))
```
Cryoris: If you're already building the expected circuit, you could check the equality more efficient by just comparing the circuits
```suggestion
        self.assertEqual(expected, new_circuit)
```
You might have to remove the use of an explicit `QuantumRegister` and just construct the circuits as `QuantumCircuit(num_qubits + 1)`. (Also side remark, it's a bit weird to have `num_qubits` not actually be the number of qubits in the circuit, but you can keep it if you prefer this 🙂 )

qiskit/transpiler/passes/optimization/light_cone.py
```diff
@@ -0,0 +1,143 @@
...
+        return set(light_cone), light_cone_ops
+
+    def run(self, dag: DAGCircuit):
+        """Run the LightCone pass on `dag`.
+
+        Args:
+            dag (DAGCircuit): the DAG to be optimized.
+
+        Returns:
+            DAGCircuit: the optimized DAG.
```
Cryoris: You can skip the typehints here since if they are present in the signature 🙂 I'd also suggest to specify what "optimize" means here.
```suggestion
            dag: The DAG to reduce.

        Returns:
            The DAG reduced to the light cone of the observable.
```

qiskit/transpiler/passes/optimization/light_cone.py
```diff
@@ -0,0 +1,143 @@
...
+        """
+        # Get the initial light cone and operations
+        light_cone, light_cone_ops = self._get_initial_lightcone(dag)
+
+        #  Initialize a new, empty DAG
+        new_dag = DAGCircuit()
```
Cryoris: You should be able to use `new_dag = dag.copy_empty_like()` here instead of creating a new one manually

MarcDrudis: Okay

qiskit/transpiler/passes/optimization/light_cone.py
MarcDrudis: Yes. Good catch. I always mess this up and iterating over the Pauli object is a good way not to. I implemented it like that but had to revert back to `str` as we wanted to make it compatible with `SparsePauliOp.to_sparse_list()` (and also `SparseObservable`).

qiskit/transpiler/passes/optimization/light_cone.py
MarcDrudis: Yes. This is actually a very good point. I am working on fixing it right now , but it creates a bug for some reason.

MarcDrudis:
I created a test that creates the lightcone of a large Pauli string and it is not passing. This is linked to #13828. How should we proceed? Do we wait until the issue is fixed for merging this PR, or should we just throw a warning for now if the Pauli string gets too big?

Cryoris:
You could comment out the test case saying that it requires #13828 to be fixed, but I wouldn't add a custom warning, since that's extra cleanup work once the bug has been fixed. The expected behavior is for this to work 🙂

qiskit/transpiler/passes/optimization/light_cone.py
```diff
@@ -0,0 +1,136 @@
...
+    either an observable one would like to measure or a measurement on a set of qubits.
+    """
+
+    def __init__(self, bit_terms: str | None = None, indices: list[int] | None = None) -> None:
+        """
+        Args:
+            bit_terms: If `None` the light-cone will be computed for the set of measurements
+            in the circuit. If a string is specified, the light-cone will correspond to the
+            reduced circuit with the same expectation value for the observable.
+            indices: list of non-trivial indices corresponding to the observable in `bit_terms`.
```
Cryoris:
```suggestion
            bit_terms: If ``None`` the light-cone will be computed for the set of measurements
                in the circuit. If a string is specified, the light-cone will correspond to the
                reduced circuit with the same expectation value for the observable.
            indices: list of non-trivial indices corresponding to the observable in ``bit_terms``.
```

qiskit/transpiler/passes/optimization/light_cone.py
```diff
@@ -0,0 +1,136 @@
...
+        self, dag: DAGCircuit
+    ) -> tuple[set[Qubit], list[tuple[Gate, list[Qubit]]]]:
+        """Returns the initial light-cone.
+        If obervable is `None`, the light-cone is the set of measured qubits.
+        If a `bit_terms` is provided, the qubits corresponding to the
+        non-trivial Paulis define the light-cone.
+        """
+        lightcone_qubits = self._find_measurement_qubits(dag)
+        if self.indices and len(dag.qubits) < max(self.indices) + 1:
+            raise ValueError("`indices` contains values outside the qubit rage.")
```
Cryoris: This could be a bit optimized: you're checking later if `self.bit_terms` is not None, in which case you _know_ that `self.indices` is also provided. So you can skip the `if self.indices` check here and move the index check inside the `if self.bit_terms ...` part.

qiskit/transpiler/passes/optimization/light_cone.py
```diff
@@ -0,0 +1,136 @@
...
+                # Check commutation with all previous operations
+                commutes_bool = True
+                for op in lightcone_operations:
+                    max_num_qubits = max(len(op[1]), len(node.qargs))
+                    if max_num_qubits > 10:
+                        raise Warning(
+                            "LightCone pass is checking commutation of"
+                            f"operators of size {max_num_qubits}."
+                            "This operation can be slow."
+                        )
```
Cryoris: This will exit the program, what you should use here is `warnings.warn` with a `RuntimeWarning` as category

test/python/transpiler/test_light_cone.py
```diff
@@ -0,0 +1,386 @@
...
+from qiskit.transpiler.passes.optimization.light_cone import LightCone
+from qiskit.transpiler.passmanager import PassManager
+
+
+@ddt.ddt
+class TestLightConePass(QiskitTestCase):
+    """Test the LightCone pass."""
+
+    def __init__(self, methodName: str = "runTest") -> None:
+        super().__init__(methodName)
```
Cryoris: This looks very strange, why do you need this?

test/python/transpiler/test_light_cone.py
```diff
@@ -0,0 +1,386 @@
...
+
+    @ddt.data("X", "Z")
+    def test_nonparameterized_noncommuting(self, pauli_label):
+        """Test for a non-commuting, asymmetric, weight-one Pauli."""
+        bit_terms = pauli_label
+        light_cone = LightCone(bit_terms=bit_terms, indices=[1])
+        pm = PassManager([light_cone])
+
+        q0 = QuantumRegister(2, "q0")
+        qc = QuantumCircuit(q0)
```
Cryoris: You can ofc use registers, but there's not really a reason in these tests. You can also just give the number of qubits.
```suggestion
        qc = QuantumCircuit(2)
```

releasenotes/notes/add-light-cone-pass-6c56085734512e98.yaml
```diff
@@ -0,0 +1,53 @@
...
+        .. plot::
+
+          from qiskit.transpiler.passes.optimization.light_cone import LightCone
+          from qiskit.transpiler.passmanager import PassManager
+          from qiskit.circuit import QuantumCircuit
+
+          qc = QuantumCircuit(3,1)
+          qc.h(0)
+          qc.h(1)
+          qc.h(2)
```
Cryoris: this could be written more concise if you want
```suggestion
          qc.h(range(3))
```

releasenotes/notes/add-light-cone-pass-6c56085734512e98.yaml
```diff
@@ -0,0 +1,53 @@
...
+          qc.cx(0,1)
+          qc.cx(2,1)
+          qc.h(0)
+          qc.h(1)
+          qc.h(2)
+          qc.measure(0,0)
```
Cryoris: Why is this measurement repeated?

Cryoris (commented):
Overall this LGTM, I left some comments below 🙂

releasenotes/notes/add-light-cone-pass-6c56085734512e98.yaml
MarcDrudis: Just a typo.

test/python/transpiler/test_light_cone.py
MarcDrudis: You are right. Now it is fixed.

test/python/transpiler/test_light_cone.py
MarcDrudis: We don't. I have removed it.

qiskit/transpiler/passes/optimization/light_cone.py
MarcDrudis: This leads to a bug. I have created a unitest and opened an issue #13828 . Once this issue is solved, the test can be enabled and everything else should work well.

qiskit/transpiler/passes/optimization/light_cone.py
```diff
@@ -0,0 +1,137 @@
...
+import warnings
+from qiskit.circuit import Gate, Qubit
+from qiskit.circuit.commutation_checker import CommutationChecker
+from qiskit.circuit.commutation_library import standard_gates_commutations
+from qiskit.circuit.library import PauliGate, ZGate
+from qiskit.dagcircuit import DAGCircuit
+from qiskit.transpiler.basepasses import TransformationPass
+from qiskit.transpiler.passes.utils.remove_final_measurements import calc_final_ops
+
+commutator = CommutationChecker(standard_gates_commutations)
```
Cryoris: Could you use the `SessionCommutationChecker` here (same import path as `CommutationChecker`)? That way we can save the initialization and re-use already cached relations

qiskit/transpiler/passes/optimization/light_cone.py
```diff
@@ -0,0 +1,137 @@
...
+        qubits_measured = set()
+        for node in final_nodes:
+            qubits_measured |= set(node.qargs)
+        return qubits_measured
+
+    def _get_initial_lightcone(
+        self, dag: DAGCircuit
+    ) -> tuple[set[Qubit], list[tuple[Gate, list[Qubit]]]]:
+        """Returns the initial light-cone.
+        If obervable is `None`, the light-cone is the set of measured qubits.
```
Cryoris:
```suggestion
        If observable is `None`, the light-cone is the set of measured qubits.
```

qiskit/transpiler/passes/optimization/light_cone.py
```diff
@@ -0,0 +1,137 @@
...
+        If a `bit_terms` is provided, the qubits corresponding to the
+        non-trivial Paulis define the light-cone.
+        """
+        lightcone_qubits = self._find_measurement_qubits(dag)
+        if self.bit_terms is None:
+            lightcone_operations = [(ZGate(), [qubit_index]) for qubit_index in lightcone_qubits]
+        else:
+            # Having both measurements and an observable is not allowed
+            if len(dag.qubits) < max(self.indices) + 1:
+                raise ValueError("`indices` contains values outside the qubit rage.")
```
Cryoris: Though I like raging qubits, this seems to miss an "n" 😉 
```suggestion
                raise ValueError("`indices` contains values outside the qubit range.")
```

Cryoris (commented):
Some tiny comments, then I think this is good to go!

qiskit/transpiler/passes/optimization/light_cone.py
MarcDrudis: Yes

Cryoris (approved):
LGTM thanks for all the work @MarcDrudis and @spiccinelli!

PR #12816, opened 2024-07-25
Commutation Checker with rotation gates

MarcDrudis:
### Summary

Changing the logic of commutation checker so that it can handle pauli rotation gates. We simply check if the gate is parameterized and a pauli rotation gate and if so we use the commutation relationship of its underlying PauliGate.

### Details and comments

The comments in the unittest already hinted at this feature being implemented in the future. I changed the relevant unit tests and added some for multi qubit Pauli rotations.


qiskit/circuit/commutation_checker.py
```diff
@@ -103,6 +119,20 @@ def commute(
         Returns:
             bool: whether two operations commute.
         """
+        # The rotation gates commute like their respective generators.
+        if (
+            op1.name in _rotation_gates
+            and getattr(op1, "is_parameterized", False)
+            and op1.is_parameterized()
```
Cryoris: I thought we could reduce rotation gates to Paulis independent of whether they have free parameters or not. If they have concrete parameters, the current code computes the matrix representation afaik, but if we'd check the underlying Pauli instead we could leveraged the cached table which should be faster 🙂

qiskit/circuit/commutation_checker.py
```diff
@@ -20,17 +20,31 @@
...
+    "rzx": PauliGate("XZ"),
+}
+
 
 @lru_cache(maxsize=None)
 def _identity_op(num_qubits):
     """Cached identity matrix"""
     return Operator(
-        np.eye(2**num_qubits), input_dims=(2,) * num_qubits, output_dims=(2,) * num_qubits
+        np.eye(2**num_qubits),
```
Cryoris: It seems your formatter is not set up the same way as Qiskit, maybe make sure to run black again with the local settings or update your lint to allow line lengths of 105 (see our `pyproject.toml`)

qiskit/circuit/commutation_checker.py
```diff
@@ -20,17 +20,31 @@
...
 from qiskit.circuit.controlflow import CONTROL_FLOW_OP_NAMES
+from qiskit.circuit.library import XGate, YGate, ZGate
+from qiskit.circuit.library.generalized_gates.pauli import PauliGate
 from qiskit.quantum_info.operators import Operator
 
 _skipped_op_names = {"measure", "reset", "delay", "initialize"}
 _no_cache_op_names = {"annotated"}
 
+_rotation_gates = {
+    "rx": XGate(),
```
Cryoris: Could we add controlled Pauli rotations and the phase gate (-> RZ) as well?

Cryoris (commented):
This is a very nice addition, thanks for tackling it! I left some comments below and I think we should add some more tests. Maybe something as simple as checking that `R<Pauli>(theta)` and `<Pauli>` commute?

qiskit/circuit/commutation_checker.py
MarcDrudis: I agree with that. We would need to be careful that we are not missing some cases in which the general rotation does not commute but the paramterized one does. I am thinking mainly of theta%2*np.pi is close to 0.

qiskit/circuit/commutation_checker.py
MarcDrudis: Yes, this should be trivial. I can look into all the paramterized gates. As long as they span from a single gate rotation it is as easy as making this dictionary bigger.

Cryoris:
Now that #12870 is almost ready, it would probably be best to directly implement these changes in Rust, once it is merged 🙂

Cryoris:
Outdated since the port of commutation checker to rust

PR #12575, opened 2024-06-13
Deprecate PrimitivesV1 reference implementations and their utils

LeanderCS:
Fixes #12497
### Summary
Deprecate

- BaseEstimatorV1
- BaseSamplerV1
- Estimator
- Sampler
- BackendEstimator
- BackendSampler
- init_circuit function
- init_observable function
- final_measurement_mapping function

Update
- BasePrimitive docstrings

PR #12818, opened 2024-07-25
Follow up on non-versioned primitives deprecation

1ucian0:
follow up on #12575

## `BackendSampler`

Deprecating `BackendSampler` is not as easy as use `BackendSamplerV2`, because the former is `BaseSamplerV1`.

We either move the tests that use `BackendSampler` to `BackendSamplerV2` (which is not super direct).

So, this PR creates a `BackendSamplerV1` as an alternative to `BackendSampler`.

## `BackendEstimator`

ditto `BackendSampler`

## utils

Because the deprecation is only on the aliases and not on PrimitivesV1, the utilities are still needed.

## replacement recommendation

While the recommendation is to move to primitives V2, the drop-in replacement for the alias is primitives V1. For some users, that might be the quickest solution. I think both alternatives should be mentioned.


1ucian0:
I lived in a confused universe... this PR makes no sense.

PR #12825, opened 2024-07-26
Context-aware dynamical decoupling

Cryoris:
### Summary

Implement context-aware dynamical decoupling (DD) from [arXiv:2403.06852](https://arxiv.org/abs/2403.06852).

### Details and comments

When selecting decoupling sequences for an idle qubit, context-aware DD takes into account what actions are occuring on neighboring qubits. It ensures that neighboring sequences are mutually orthogonal (to avoid collisions, like in staggered DD) but also whether the qubit is a CX or ECR target/control spectator. This is nicely summarized in a figure of the paper, where the different wire colors denote different decoupling sequences:

image

This particularly leads to improvement if a lot of qubits are idle at the same time or if they are next to ECR/CX gates. One important category of circuits in this class are the layers in twirled circuits. This pass also works best if the ECR/CX gate times are the same for all connections, which is almost always the case on IBM devices.

Some simple experiments show that this pass performs better than the current standard in Qiskit, `PadDynamicaDecoupling` with an `X-X` sequence. Here's some basic benchmarks, see the paper for more: 

image

This code is based off @kdk and @ajavadia 🙂


qiskit/transpiler/passes/scheduling/padding/context_aware_dynamical_decoupling.py
```diff
@@ -0,0 +1,785 @@
...
+        if len(closed_delay.active_qubits) > 0:
+            raise RuntimeError("Failed remove active qubits on closed delays.")
+
+        if validate:
+            closed_delay.validate()
+
+    return closed_delay_blocks
+
+
+def _split_adjacent_block(
```
eliarbel: Any particular reason why this and all the functions below are not instance methods of the `ContextAwareDynamicalDecoupling` class?

Cryoris: Simply because they don't depend on `self` and I wanted to keep the class lean, but looking at it again, the functions depend on the class context (e.g. `self.property_set`) so putting them inside the class would make sense 🙂 I'll update!

Cryoris: Done in c6700ebb7.

miamico:
Anything holding this up from being merged? Happy to lend a hand if needed to speed up the process @Cryoris @eliarbel

qiskit/transpiler/passes/__init__.py
```diff
@@ -286,6 +286,7 @@
 from .scheduling import ASAPSchedule
 from .scheduling import DynamicalDecoupling
 from .scheduling import AlignMeasures
+from .scheduling import ContextAwareDynamicalDecoupling
```
eliarbel, line 261: Need to add the class also to the Scheduling section above in the documentation part so it'll be included in the docs.

test/python/transpiler/test_context_aware_dd.py
```diff
@@ -0,0 +1,734 @@
...
+        elif name == "x":
+            return (1e-10, 1.1e-10, 9e-5, 6e-3)
+        elif name == "sx":
+            return (5e-11, 5.5e-11, 9e-5, 6e-3)
+        return super()._get_noise_defaults(name, num_qubits)
+
+
+@ddt
+class TestContextAwareDD(QiskitTestCase):
+    """Test context-aware dynamical decoupling."""
```
eliarbel: For convenience I would put the link to the paper also here

test/python/transpiler/test_context_aware_dd.py
```diff
@@ -0,0 +1,734 @@
...
+    def paper_setting(self):
+        """Get the circuit from Fig. 5 of the context-aware DD paper."""
+        circuit = QuantumCircuit(6)
+        circuit.barrier()
+        circuit.cx(1, 0)
+        circuit.cx(3, 4)
+        circuit.barrier()
+        circuit.cx(1, 2)
+        circuit.cx(4, 5)
+        circuit.barrier()
```
eliarbel: The penultimate layer of Fig. 5 is missing here, namely cx(1,0) and cx(4,5). Is this intentional?

qiskit/transpiler/passes/scheduling/padding/context_aware_dynamical_decoupling.py
```diff
@@ -0,0 +1,756 @@
...
+from qiskit.transpiler.coupling import CouplingMap
+from qiskit.transpiler.exceptions import TranspilerError
+
+from .pad_delay import PadDelay
+
+logger = logging.getLogger(__name__)
+
+
+class ContextAwareDynamicalDecoupling(TransformationPass):
+    """Context-aware dynamical decoupling.
```
eliarbel: Since the summary in the Scheduling TOC in the docs is taken from here, I would make this a bit more informative otherwise it's just repeating the class name.

qiskit/transpiler/passes/scheduling/padding/context_aware_dynamical_decoupling.py
```diff
@@ -0,0 +1,756 @@
...
+
+    This pass implements a context-aware dynamical decoupling (DD) [1], which ensures that
+
+        (1) simultaneously occuring DD sequences on device-adjacent qubits are mutually orthogonal, and
```
eliarbel:
```suggestion
        (1) simultaneously occurring DD sequences on device-adjacent qubits are mutually orthogonal, and
```

qiskit/transpiler/passes/scheduling/padding/context_aware_dynamical_decoupling.py
```diff
@@ -0,0 +1,756 @@
...
+        (2) DD sequences on spectator qubits of ECR and CX gates are orthogonal to the echo
+            pulses on the neighboring target/control qubits.
+
+    The mutually orthogonal DD sequences are currently Walsh-Hadamard sequences, consisting of only
+    X gates. In some cases it might therefore be benefitial to use :class:`.PadDynamicalDecoupling`
```
eliarbel:
```suggestion
    X gates. In some cases it might therefore be beneficial to use :class:`.PadDynamicalDecoupling`
```

qiskit/transpiler/passes/scheduling/padding/context_aware_dynamical_decoupling.py
```diff
@@ -0,0 +1,756 @@
...
+                )
+                op.replacement.compose(dd_circuit, inplace=True, copy=False)
+                if len(op.start_times) == 0:
+                    op.start_times += start_times
+                else:
+                    start_times = [op.start_times[-1] + time for time in start_times]
+
+                op.start_times += start_times
+
+        # 5) Replace each delay operation with it's individual DD sequence
```
eliarbel:
```suggestion
        # 5) Replace each delay operation with its individual DD sequence
```

qiskit/transpiler/passes/scheduling/padding/context_aware_dynamical_decoupling.py
```diff
@@ -0,0 +1,756 @@
...
+        qubit_map = dict(enumerate(dag.qubits))
+        for delay in all_delays:
+            # replace it with it's stored replacement
```
eliarbel:
```suggestion
            # replace it with its stored replacement
```

qiskit/transpiler/passes/scheduling/padding/context_aware_dynamical_decoupling.py
```diff
@@ -0,0 +1,756 @@
...
+        elif order == 2:
+            spacing = [1 / 4, 1 / 4, 1 / 4, 1 / 4, 0]
+        elif order == 3:
+            spacing = [1 / 8, 1 / 4, 1 / 4, 1 / 4, 1 / 8]
+        elif order == 4:
+            spacing = [1 / 8, 1 / 4, 1 / 8, 1 / 8, 1 / 4, 1 / 8, 0]
+        elif order == 5:
+            spacing = [1 / 8, 1 / 8, 1 / 8, 1 / 4, 1 / 8, 1 / 8, 1 / 8]
+        else:
+            max_order = ContextAwareDynamicalDecoupling.MAX_ORDER
```
eliarbel: What scenarios would lead to asking for orders higher than 5? Worth explaining why 5 is the max order.

qiskit/transpiler/passes/scheduling/padding/context_aware_dynamical_decoupling.py
```diff
@@ -0,0 +1,756 @@
...
+
+        sorted_events = sorted(eligible_delays, key=DelayEvent.sort_key)
+        logger.debug("Sorted delay events: %s", sorted_events)
+
+        return sorted_events
+
+    def _get_wire_coloring(self, dag: DAGCircuit, merged_delay: MultiDelayOp) -> dict[int, int]:
+        """Find a wire coloring for a multi-delay operation.
+
+        This function returns a dictionay that includes the coloring (as int) for the indices in the
```
eliarbel:
```suggestion
        This function returns a dictionary that includes the coloring (as int) for the indices in the
```

qiskit/transpiler/passes/scheduling/padding/context_aware_dynamical_decoupling.py
```diff
@@ -0,0 +1,756 @@
...
+                )
+            )
+            return open_delay_blocks[-1]
+
+        def _update_delay_block(open_delay_block, delay_event):
+            """Add another delay event to an existing block to either extend or close it."""
+            open_delay_block.events.append(delay_event)
+
+            if delay_event.type == EventType.BEGIN:  # TODO: this is never called
+                open_delay_block.active_qubits += set(delay_event.op_node.qargs)
```
eliarbel: This is branch is indeed not taken as `_update_delay_block` is only called with EventType.END. I would clean this then and put `assert delay_event.type != EventType.BEGIN`

qiskit/transpiler/passes/scheduling/padding/context_aware_dynamical_decoupling.py
```diff
@@ -0,0 +1,756 @@
...
+                    for open_delay_qubit in open_delay.active_qubits
+                )
+            ]
+
+            if delay_event.type == EventType.BEGIN:
+                # If crossing a begin edge, check if there are any open delays that are adjacent.
+                # If so, add current event to that group.
+
+                if len(adjacent_open_delay_blocks) == 0:
+                    # Make a new delay
```
eliarbel:
```suggestion
                    # Make a new delay block
```

qiskit/transpiler/passes/scheduling/padding/context_aware_dynamical_decoupling.py
```diff
@@ -0,0 +1,756 @@
...
+                    _open_delay_block(delay_event)
+                else:
+                    # Make a new block and combine that with adjacent open blocks
+                    new_block = _open_delay_block(delay_event)
+                    _combine_delay_blocks(adjacent_open_delay_blocks + [new_block])
+
+            if delay_event.type == EventType.END:
```
eliarbel: Just nitpicking, but I would make this an `else:` instead.

qiskit/transpiler/passes/scheduling/padding/context_aware_dynamical_decoupling.py
```diff
@@ -0,0 +1,756 @@
...
+                # If crossing a end edge, remove this qubit from the actively delaying qubits"
+                if len(adjacent_open_delay_blocks) != 1:
+                    raise RuntimeError("Closing edge w/o an open delay?")
```
eliarbel: There are several `RuntimeError` raised in the code but it's not clear if those are meant to be for the user or of the developer. This one in particular seems more like a developer check to me, so I would transform it to an `assert` statement. I would only raise exceptions for cases the users should act upon, e.g. in cases of invalid inputs (e.g. DAG/Target).

qiskit/transpiler/passes/scheduling/padding/context_aware_dynamical_decoupling.py
```diff
@@ -0,0 +1,756 @@
...
+        logger.debug("len(closed delays): %s", len(closed_delay_blocks))
+
+        # validate the results, there should be no open delays and all active qubits
+        # should be accounted for
+        if len(open_delay_blocks) > 0:
+            raise RuntimeError("Failed to close all open delays.")
+
+        for closed_delay in closed_delay_blocks:
+            if len(closed_delay.active_qubits) > 0:
+                raise RuntimeError("Failed remove active qubits on closed delays.")
```
eliarbel: Also here and a few lines above - are these developer checks or user errors?

qiskit/transpiler/passes/scheduling/padding/context_aware_dynamical_decoupling.py
```diff
@@ -0,0 +1,756 @@
...
+
+            if validate:
```
eliarbel: There are several validation checks across the code. I suggest enabling and using them in `test_context_aware_dd.py`

qiskit/transpiler/passes/scheduling/padding/context_aware_dynamical_decoupling.py
```diff
@@ -0,0 +1,756 @@
...
+                if open_group in next_groups:
+                    # extend the delay operation and remove a breakpoint
+                    delay.end = window[1]
+                    for op in delay.ops:
+                        op.breakpoints.remove(window[0])
+
+                    # TODO this is a sanity check
+                    involved_ops = {op_map[(index, window)] for index in open_group}
+                    if len(involved_ops.difference(delay.ops)) > 0:
+                        raise RuntimeError("I cannot think")
```
eliarbel: You are thinking too classically ;-)
Make it an assertion or part of a validation block

qiskit/transpiler/passes/scheduling/padding/context_aware_dynamical_decoupling.py
```diff
@@ -0,0 +1,756 @@
...
+class DelayEvent:
+    """Represent a single-qubit delay event, which is either begin or end of a delay instruction."""
+
+    type: EventType
+    time: int  # Time in this circuit, in dt
+    op_node: DAGOpNode  # The node for the circuit delay
+
+    @staticmethod
+    def sort_key(event: DelayEvent) -> tuple(int, int):
+        """Sort events, first by time then by type ('end' events come for 'begin' events)."""
```
eliarbel:
```suggestion
        """Sort events, first by time then by type ('end' events come before 'begin' events)."""
```

qiskit/transpiler/passes/scheduling/padding/context_aware_dynamical_decoupling.py
```diff
@@ -0,0 +1,756 @@
...
+@dataclass
+class AdjacentDelayBlock:
+    """Group of circuit delays which are collectively adjacent in time and on device.
+
+    For example, here the 3 delay operations on q0, q1 and q2 form an adjacent delay block.
+
+        q0: -██████---------  |  qubits q0,q1,q2 have adjacent delay
+        q1: ------███████---  |  operations, since the delay operations
+        q2: --█████████-----  |  all overlap
+        q3: -----------████-  -> this delay starts when delay on q2 end, so they have no overlap
```
eliarbel:
```suggestion
        q3: -----------████-  -> this delay starts when delay on q2 ends, so they have no overlap
```

eliarbel (changes requested):
Thanks Julien, overall seems well written and documented. I didn't read the paper so the review is not about the actual algorithmic part of the DD method itself. I suggest getting input from Kevin or Ali if you want another pair of eyes on this aspect. 
I've left line comments, most of which are minor but still require addressing. 
In terms of performence - do you have a sense about the overhead of the block collection and manipulation (e.g. `_collect_adjacenet_delay_blocks` )?

qiskit/transpiler/passes/scheduling/padding/context_aware_dynamical_decoupling.py
mtreinish: I agree with Eli here about the intent of these exception not being clear or user actionable, but I don't think a runtime assertion is a good choice. 

Using `assert` is not really desirable either because the raised `AssertionError` is also not really actionable and typically ends up being even more cryptic, especially since there is little control over the exception raised. I'd argue either change the exceptions raised to a qiskit exception type and change the error message to indicate how a user can address the failure mode or just remove the check completely. If it's just an internal sanity check for debugging we should remove it and avoid the runtime overhead completely. In theory this is what you can do with asserts, because `python -O` will strip the asserts but in practice I've very rarely seen anyone use that flag.

jakelishman: Fwiw, we have used `RuntimeError` for other places in Qiskit to represent an internal logic error, but the idea is that it should be entirely impossible for a user to trigger them (unless they're doing something they're not allowed to), or they're a flag to implementers that they've done something wrong.

That said, I believe that the majority of places it's used in Qiskit already are places where it has no runtime cost to do, because it's the "default" case of a lookup table or an `if/else` chain, or something like that, and we need to do _something_ other than just fall off the end of the chain silently.  I think quite a few of the uses here don't fall into those buckets, and should be replaced with proper errors - some of them are user actionable, and some are internal logic validation that has non-trivial overhead to check.

qiskit/transpiler/passes/scheduling/padding/context_aware_dynamical_decoupling.py
```diff
@@ -0,0 +1,756 @@
...
+        skip_dd_threshold: float = 1.0,
+        pulse_alignment: int | None = None,
+        coloring_strategy: rx.ColoringStrategy = rx.ColoringStrategy.Saturation,
+    ) -> None:
+        """
+        Args:
+            target: The :class:`.Target` of the device to run the circuit.
+            min_joinable_duration: Do not join delays with others below this duration. This
+                can be useful, e.g. if a big delay block would be interrupted and split into
+                smaller blocks due to a very short, adjacent delay.
```
ElePT: I would find it very helpful if the documentation mentioned something that could be linked easier to "minimal delay time" and what units are used, something like (I still find it a bit clunky):

```suggestion
            min_joinable_duration: Minimal duration to join delays (in dt units). Do not join delays with
                others below this duration . This can be useful, e.g. if a big delay block would be interrupted
                and split into smaller blocks due to a very short, adjacent delay.
```

qiskit/transpiler/passes/scheduling/padding/context_aware_dynamical_decoupling.py
```diff
@@ -0,0 +1,756 @@
...
+            skip_reset_qubits: Skip initial delays and delays after a reset.
+            skip_dd_threshold: Skip dynamical decoupling on an idle qubit, if the duration of
```
ElePT, line 118: In what units is `skip_dd_threshold`? I have noticed that `min_joinable_duration` is an int but this one is a float, so maybe it's not dt?

qiskit/transpiler/passes/scheduling/padding/context_aware_dynamical_decoupling.py
```diff
@@ -0,0 +1,756 @@
...
+                the decoupling sequence exceeds this fraction of the idle window. Per default,
+                the sequence is applied if it fits into the window.
+            pulse_alignment: The hardware constraints for gate timing allocation.
+                This is usually provided om ``backend.configuration().timing_constraints``.
+                If provided, the delay length, i.e. ``spacing``, is implicitly adjusted to
+                satisfy this constraint. If ``None``, this is extracted from the ``target``.
```
ElePT: Same comment about units

ElePT (commented):
Thanks @Cryoris, as you know I have ran this pass and noticed a few painpoints I wanted to point out. In addition to the documentation suggestions, I have noticed that `min_joinable_duration` is a key parameter to avoid 1q gate explosion in some circuits, I wonder if you had considered setting it no a non-0 default value?

ElePT (commented):
Another comment, I believe there is a bug in the instruction duration assignation, because the timeline drawer fails after running this pass complaining about missing gate durations:

```
  File "/Users/ept/qiskit_workspace/qiskit/debug_dd.py", line 84, in <module>
    timeline_drawer(bound_scheduled_circuit, filename="out.png")
  File "/Users/ept/qiskit_workspace/qiskit/qiskit/utils/deprecation.py", line 183, in wrapper
    return func(*args, **kwargs)
  File "/Users/ept/qiskit_workspace/qiskit/qiskit/utils/deprecation.py", line 183, in wrapper
    return func(*args, **kwargs)
  File "/Users/ept/qiskit_workspace/qiskit/qiskit/visualization/timeline/interface.py", line 379, in draw
    canvas.load_program(program=program)
  File "/Users/ept/qiskit_workspace/qiskit/qiskit/visualization/timeline/core.py", line 193, in load_program
    for datum in gen(gate_source, formatter=self.formatter):
  File "/Users/ept/qiskit_workspace/qiskit/qiskit/visualization/timeline/generators.py", line 158, in gen_sched_gate
    if gate.duration > 0:
TypeError: '>' not supported between instances of 'NoneType' and 'int'
```

test/python/transpiler/test_context_aware_dd.py
Cryoris: Good catch! Though this is actually legacy code and no longer used in the testing 🙂 I'll remove the whole method.

qiskit/transpiler/passes/scheduling/padding/context_aware_dynamical_decoupling.py
Cryoris: Whoops I didn't know this spelling is incorrect -- thanks!

qiskit/transpiler/passes/scheduling/padding/context_aware_dynamical_decoupling.py
Cryoris: Because I got to lazy to write the code that generalizes this and thought there's not frequently a device which has a qubit with more than 5 connections 😉 I'll add the code that does this generally

qiskit/transpiler/passes/scheduling/padding/context_aware_dynamical_decoupling.py
Cryoris: done in 13802fc36

qiskit/transpiler/passes/scheduling/padding/context_aware_dynamical_decoupling.py
Cryoris: fixed in 13802fc36

test/python/transpiler/test_context_aware_dd.py
Cryoris: added in 13802fc36

qiskit/transpiler/passes/__init__.py
Cryoris: added in 13802fc36

test/python/transpiler/test_context_aware_dd.py
Cryoris: removed in 13802fc36

qiskit/transpiler/passes/scheduling/padding/context_aware_dynamical_decoupling.py
Cryoris: fixed in 13802fc36

qiskit/transpiler/passes/scheduling/padding/context_aware_dynamical_decoupling.py
Cryoris: fixed in 13802fc36

qiskit/transpiler/passes/scheduling/padding/context_aware_dynamical_decoupling.py
Cryoris: fixed in 13802fc36

qiskit/transpiler/passes/scheduling/padding/context_aware_dynamical_decoupling.py
Cryoris: fixed in 13802fc36

qiskit/transpiler/passes/scheduling/padding/context_aware_dynamical_decoupling.py
Cryoris: Added in 13802fc36. This can now compute Walsh Hadamard sequences up to arbitrary orders.

qiskit/transpiler/passes/scheduling/padding/context_aware_dynamical_decoupling.py
Cryoris: fixed in 13802fc36

qiskit/transpiler/passes/scheduling/padding/context_aware_dynamical_decoupling.py
Cryoris: Removed entirely in 13802fc36

qiskit/transpiler/passes/scheduling/padding/context_aware_dynamical_decoupling.py
Cryoris: fixed in 13802fc36

qiskit/transpiler/passes/scheduling/padding/context_aware_dynamical_decoupling.py
Cryoris: done in 13802fc36

qiskit/transpiler/passes/scheduling/padding/context_aware_dynamical_decoupling.py
Cryoris: I opted for the third variant: adding optional `logger.debug` statements, if the respective logging level is enabled. This code has been tested on dummy circuits and a variety of smaller hardware experiments, but I'd prefer keeping an easy way of printing debug info as long as the code is relatively new. Let me know if this is fine, otherwise I can also remove these!

qiskit/transpiler/passes/scheduling/padding/context_aware_dynamical_decoupling.py
Cryoris: They are dev checks -- see comment above 🙂

qiskit/transpiler/passes/scheduling/padding/context_aware_dynamical_decoupling.py
eliarbel: Yes, a logging approach is good

eliarbel (commented):
Thanks @Cryoris for applying all the changes. This looks good and I have no further comments. There are still some comments from @ElePT.

eliarbel (dismissed):
Fine from my side. Please address the remaining comments from @ElePT

qiskit/transpiler/passes/scheduling/padding/context_aware_dynamical_decoupling.py
Cryoris: It's in the fraction of the idle window 🙂 
> Skip dynamical decoupling on an idle qubit, if the duration of
                the decoupling sequence exceeds this fraction of the idle window.
                
I'll add a clarifying comment, maybe something like
```
For example, to skip DD if the DD sequence would take up
more than 95% of the idle time, set this to 0.95. A value of 
1.0 means that DD is always used if it fits into the sequence.
```

Cryoris:
https://github.com/Qiskit/qiskit/pull/12825#pullrequestreview-2404327735 is a very good catch -- the newly added gates that implement the DD sequences didn't have a `duration` attached, since they weren't present when the scheduling pass was run. I'll add a test to check for this 🙂

Cryoris:
After discussion with @ElePT I also updated the default behavior of `min_joinable_duration`: On realistic devices, the 2q gate lengths vary a little -- yet might not want to insert DD sequences for two concurrent 2q gates, just because they naturally differ slightly. The new default is therefore that we only add DD sequences for idle times beyond `2 * (max_2q_duration - min_2q_duration)`. This needs some benchmarking on real devices before merging.

qiskit/transpiler/passes/scheduling/padding/context_aware_dynamical_decoupling.py
Cryoris: Done in [6c6f6d7](https://github.com/Qiskit/qiskit/pull/12825/commits/6c6f6d7ff633c1246a3ac08dce07fb922011f8a2)

qiskit/transpiler/passes/scheduling/padding/context_aware_dynamical_decoupling.py
Cryoris: Done in [6c6f6d7](https://github.com/Qiskit/qiskit/pull/12825/commits/6c6f6d7ff633c1246a3ac08dce07fb922011f8a2)

qiskit/transpiler/passes/scheduling/padding/context_aware_dynamical_decoupling.py
Cryoris: Done in [6c6f6d7](https://github.com/Qiskit/qiskit/pull/12825/commits/6c6f6d7ff633c1246a3ac08dce07fb922011f8a2)

mtreinish:
> [#12825 (review)](https://github.com/Qiskit/qiskit/pull/12825#pullrequestreview-2404327735) is a very good catch -- the newly added gates that implement the DD sequences didn't have a `duration` attached, since they weren't present when the scheduling pass was run. I'll add a test to check for this 🙂

Just so you know I'm working on removing per instruction durations in #13506 so we probably don't need to change this now. It would have been important for 1.3, but in case 2.0 we just making extra work for both of us.

Cryoris:
I removed any setting and usage of `Instruction.duration` in 21c4ea0, so hopefully you don't have to update the context-aware DD in #13506 🙂 That means the timeline drawer will fail now, but I assume this will be fixed in your PR 🙂

eliarbel:
> After discussion with @ElePT I also updated the default behavior of `min_joinable_duration`: On realistic devices, the 2q gate lengths vary a little -- yet might not want to insert DD sequences for two concurrent 2q gates, just because they naturally differ slightly. The new default is therefore that we only add DD sequences for idle times beyond `2 * (max_2q_duration - min_2q_duration)`. This needs some benchmarking on real devices before merging.

Hi @Cryoris are you planning to benchmark the new `min_joinable_duration` default on real devices before the 2.1 release?

Cryoris:
Yep, it's good from my side now 🙂 The number of X gates is rather large though, per design of the algorithm, I think that was just surprising to see. 

E.g. for an N-local circuit with linear entanglement on `n` qubits (not that someone would run this), the number of X gates goes with `O(n^2)` instead of `O(n)` for "normal" DD. But the results of CA-DD look promising on the eagle devices and the overhead can be controlled with the `min_joinable_duration` parameter.

eliarbel:
Overall this looks in good shape to me both from code and testing perspectives. I haven't reviewed the algorithm in depth against the paper, but I ran several examples and the results look OK. Still, I'd like to look into some of the algorithmic aspects a bit more. In the meantime, I have a couple questions and possible suggestions, both not major: 

The rationale behind the default value for `min_joinable_duration` is not so clear to me. Also, it doesn't feel very intuitive as a knob for controlling the amount of DD inserted into the circuit. Is there some underlying intuition behind the choice, something that could be added to the docstring to help clarify this a bit?

'cz' is included in the calculation of the 2q gate length variance in `_gate_length_variance`. However, the logic for handling spectator qubits does not account for `CZGate`s. Was this intentional, perhaps due to the differences in crosstalk levels between Heron and Eagle devices? If so, this introduces a technology assumption in the implementation, isn't it? I'm wondering if it would make sense to provide a way for users to opt in or out of spectator qubit analysis for specific gate types?

Cryoris:
Re: `min_joinable_duration` -- this shouldn't set the amount of DD inserted, but rather control how adjacent idle blocks are joined. It's a knob that lets you trade-off between potentially many short sequences or longer blocks. The default value is picked to ensure that a layer of CX/ECR will not have a lot of short DD sequences just due to the difference in duration. But this is not discussed in the paper afaik and I'm happy to change this if you think there's a more reasonable choice 🙂 

Re: CZ -- The paper only investigates CX/ECR. CZ would probably require a different spectator qubit handling, but this is not explained in the paper and hence not supported in the code. We can add an argument that allows to turn off the spectator-specific logic for CX/ECR.

eliarbel:
> Re: min_joinable_duration -- this shouldn't set the amount of DD inserted, but rather control how adjacent idle blocks are joined. It's a knob that lets you trade-off between potentially many short sequences or longer blocks. The default value is picked to ensure that a layer of CX/ECR will not have a lot of short DD sequences just due to the difference in duration. But this is not discussed in the paper afaik and I'm happy to change this if you think there's a more reasonable choice 🙂

The reason I treated this is as a knob to control the amount of DD is because of this check: 
```python
if (
    node.op.name == "delay"
    and not is_after_reset(node)
    and self._duration(node, qubit_map) > self._min_joinable_duration
   )
```
in `_get_sorted_delays`.  It skips durations shorter than `min_joinable_duration` when collecting eligible delays, effectively becoming a filter of idle blocks based on their lengths. This is because shorter delays are not just omitted from adjacency blocks and have their own DD sequences, they are skipped altogether. If we want a truly `min_joinable_duration` behavior in the "joinable" sense, we'll probably want to handle the parameter in `_collect_adjacent_delay_blocks`.  As for the value default value itself, I see the reasoning for layers so I think it's OK. But now I think that a name that better reflects how the logic is implemented currently is `min_duration` 😄

Cryoris:
Fair point, let's rename it 👍🏻

qiskit/transpiler/passes/scheduling/padding/context_aware_dynamical_decoupling.py
```diff
@@ -0,0 +1,875 @@
...
+    .. note::
+
+        If this pass is run within a pass manager (as in the example below), it will
+        automatically run :class:`.PadDelay` to allocate the delays. If instead it is run as
+        standalone (not inside a :class:`.PassManager`), the delays must already be inserted.
+
+
+    Example::
+
+        from qiskit.circuit.library import QFTGate
```
eliarbel, line 69:
```suggestion
        from qiskit.circuit import QuantumCircuit
        from qiskit.circuit.library import QFTGate
```

qiskit/transpiler/passes/scheduling/padding/context_aware_dynamical_decoupling.py
```diff
@@ -0,0 +1,875 @@
...
+class AdjacentDelayBlock:
+    """Group of circuit delays which are collectively adjacent in time and on device.
+
+    For example, here the 3 delay operations on q0, q1 and q2 form an adjacent delay block.
+
+        q0: -██████---------  |  qubits q0,q1,q2 have adjacent delay
+        q1: ------███████---  |  operations, since the delay operations
+        q2: --█████████-----  |  all overlap
+        q3: -----------████-  -> this delay starts when delay on q2 ends, so they have no overlap
+        q4: ----████--------  -> this clearly has no overlap with someting else
```
eliarbel:
```suggestion
        q4: ----████--------  -> this clearly has no overlap with something else
```

eliarbel (changes requested):
Thanks @Cryoris for the patience with the final review. I only have 2 minor docs comments, otherwise this looks very good!

eliarbel (approved):
Great work - good to go! 🚀

PR #12826, opened 2024-07-26
Avoid operator creation in transpiler

jakelishman:
### Summary

This removes very nearly all of the use of `DAGOpNode.op` in the default transpiler paths.  The sole exception is in `InverseCancellation`, which currently would involve some quite awkward gymnastics for little near-term benefit. The pass should move fully to Rust soon, making it not worth the effort.

Most of the tricks here involve using the knowledge that most operations will involve only Rust-space standard gates, and that these cannot be control-flow operations.

### Details and comments

This always includes a reno for a couple of the new methods added to `DAGOpNode` and `CircuitInstruction`.

I measured this as being ~25% faster for some toy workloads at transpilation levels 0 and 1, probably mostly driven by the new methods letting us avoid a lot of `isinstance` checks in most cases.


qiskit/transpiler/passes/layout/vf2_utils.py
```diff
@@ -37,7 +37,7 @@ class MultiQEncountered(Exception):
 
     def _visit(dag, weight, wire_map):
         for node in dag.op_nodes(include_directives=False):
-            if isinstance(node.op, ControlFlowOp):
+            if node.is_control_flow():
                 if isinstance(node.op, ForLoopOp):
```
mtreinish, line 41: You could avoid the op usage here too if you did:
```suggestion
                if node.name == "for_loop":
```

mtreinish (dismissed):
Overall this LGTM, I'm happy to wrap up this thread of workarounds before we start migrating passes to rust (which will be doing essentially this but without the language barrier). Just one inline comment but something that's applicable to a bunch of places, basically we can use the name of a control flow op instead of an `isinstance` check to defer the op creation until we actually use it. Although as those are always `PyInstruction` it doesn't really matter from a performance perspective as we always have a cached pyobject at that point. So I'm fine merging this as-is if you don't think it's important. Feel free to just enqueue this for merge in that case.

mtreinish (approved):

PR #12829, opened 2024-07-26
[WIP] Try to speed up Target python-side mapping access.

raynelfss:
### Summary
Exploratory follow-up PR to #12292, attempts to speedup `gate_map` access in python by caching each access instance to not convert more than once.

### Known issues
- The way `InstructionProperties` is stored in the Target (using `BaseInstructionProperties`) on rust does not contain the `calibration` attribute. Which causes errors.

### Details and comments
WIP.


raynelfss:
The efforts for this PR will be superseded by the results of #13308, in which we will hopefully address this.

PR #12839, opened 2024-07-29
Build `ibm-platform-services` with known-good setuptools

jakelishman:
### Summary

`ibm-platform-services<=0.55.1` only supply an `sdist` on PyPI, and the `setup.py` of that sdist imports the deprecated
`setuptools.commands.test`, which is removed in `setuptools==72.0.0`. This makes the package unbuildable with the latest `setuptools`, and it's hard to thread through a constraint on build dependencies right into the depths of recursive pip/tox/whatever commands.

This commit forcibly builds the bad dependency with an older version of `setuptools` and adds the wheel to the pip cache, so subsequent installation commands will be able to install from a pre-built wheel regardless of the `setuptools` version.

### Details and comments

See IBM/python-sdk-core#202 and IBM/platform-services-python-sdk#268 for the tracking issues on the particular repos.

### DIFF
--- a/.azure/lint_docs_qpy-linux.yml
+++ b/.azure/lint_docs_qpy-linux.yml
@@ -28,7 +28,7 @@ jobs:
           .build-deps/bin/python -m pip install 'setuptools<72.0' wheel
           .build-deps/bin/python -m pip wheel --no-build-isolation ibm-platform-services
           rm -rf .build-deps
-        displayName: Prebuild old-setuptools dependencies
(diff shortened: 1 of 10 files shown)


jakelishman:
With the release of `setuptools==72.1.0`, the immediate problems here are resolved.  Our only problematic dependency, `ibm-platform-services` (via `qiskit-ibm-runtime`) released version 0.55.2, which fixes the underlying issue, so this is no longer necessary.

PR #12842, opened 2024-07-29
Fix `StatevectorSampler` to raise an error if a circuit with  c_if is passed

### LINKED ISSUE #12281: midcircuit measurement issue with `StatevectorSampler`
### Environment

- **Qiskit version**: 1.01
- **Python version**: 3.11
- **Operating system**: macOS

### What is happening?

`StatevectorSampler` is not supposed to support midcircuit measurements. However, in its current implementation, there are instances where instead of throwing an error, it produces the wrong result.

Example:
```python
from qiskit.primitives import StatevectorSampler
from qiskit.providers.basic_provider import BasicSimulator

simulator = BasicSimulator()
sampler = StatevectorSampler()

qc = QuantumCircuit(2,2)
qc.h(1)
qc.measure(1,1)
qc.x(0).c_if(1,1)
qc.measure(0,0)
display(qc.draw('text',cregbundle=False))

sim_counts = simulator.run(qc, shots=1024).result().get_counts()
samp_counts = sampler.run([qc], shots=1024).result()[0].data.c.get_counts()
print(f'BasicSimulator Counts:     {sim_counts}')
print(f'StatevectorSampler Counts: {samp_counts}')
```

Output:
```
     ┌───┐┌─┐        
q_1: ┤ H ├┤M├────────
     └───┘└╥┘┌───┐┌─┐
q_0: ──────╫─┤ X ├┤M├
           ║ └─╥─┘└╥┘
c_1: ══════╩═══■═══╬═
                   ║ 
c_0: ══════════════╩═
                     
BasicSimulator Counts:     {'00': 519, '11': 505}
StatevectorSampler Counts: {'01': 503, '11': 521}
```

Notice that the expected output is ~ 50% `00` - 50% `11`, but sampler returns ~ 50% `01` - 50% `11`.

### What should happen?

`StatevectorSampler` should throw an error because the measurement result in `c_1` is being used to control an `XGate`, but instead it is ignoring the control and always applying the gate independent on the value on `c_1`, producing incorrect results.

### Any suggestions?

Haven't spent much time looking at the code, but it seems that the issue stems from `StatevectorSampler` only checking if there are repeated measurement blocks on the same qubit. In this example, there's only one measurement per qubit, so both get removed as if they were "final measurements". The circuit then executes without error, but obviously produces incorrect results because it is treating `c_1` as always being equal to `0`.

Not sure what the right solution would be at this point, but probably checking for dynamic-circuit conditionals?

t-imamichi:
### Summary

`StatevectorSampler` cannot handle both control flow and c_if because `Statevector` cannot.
But, it ignores `c_if` by mistake. This PR makes `StatevectorSampler` raise an error if a circuit with c_if is passed.
A unit test of `StatevectorSampler` used `c_if` accidentally. So, this PR also fixes the case without passing c_if to `StatevectorSampler`.

Fixed #12281

### DIFF
--- a/test/python/primitives/test_statevector_sampler.py
+++ b/test/python/primitives/test_statevector_sampler.py
@@ -592,16 +592,12 @@ class TestStatevectorSampler(QiskitTestCase):
         c2 = ClassicalRegister(1, "c2")
 
         qc = QuantumCircuit(q, c1, c2)
-        qc.ry(np.pi / 4, 2)
-        qc.cx(2, 1)
-        qc.cx(0, 1)
-        qc.h(0)
-        qc.measure(0, c1)
-        qc.measure(1, c2)
         qc.z(2).c_if(c1, 1)
         qc.x(2).c_if(c2, 1)
         qc2 = QuantumCircuit(5, 5)
         qc2.compose(qc, [0, 2, 3], [2, 4], inplace=True)
+        # Note: qc2 has aliased cregs, c0 -> c[2] and c1 -> c[4].
+        # copy_empty_like copies the aliased cregs of qc2 to qc3.
         qc3 = QuantumCircuit.copy_empty_like(qc2)
         qc3.ry(np.pi / 4, 2)
         qc3.cx(2, 1)
@@ -609,6 +605,7 @@ class TestStatevectorSampler(QiskitTestCase):
         qc3.h(0)
         qc3.measure(0, 2)
         qc3.measure(1, 4)
+        self.assertEqual(len(qc3.cregs), 3)
         cregs = [creg.name for creg in qc3.cregs]
         target = {
             cregs[0]: {0: 4255, 4: 4297, 16: 720, 20: 726},
--- a/test/python/primitives/test_statevector_sampler.py
+++ b/test/python/primitives/test_statevector_sampler.py
@@ -280,6 +280,11 @@ class TestStatevectorSampler(QiskitTestCase):
         qc3 = QuantumCircuit(1, 1)
         with qc3.for_loop(range(5)):
             qc3.h(0)
+        qc4 = QuantumCircuit(2, 2)
+        qc4.h(0)
+        qc4.measure(1, 1)
+        qc4.x(0).c_if(1, 1)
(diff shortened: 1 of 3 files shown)


test/python/primitives/test_statevector_sampler.py
```diff
@@ -594,15 +602,22 @@ def test_circuit_with_aliased_cregs(self):
...
         qc2 = QuantumCircuit(5, 5)
         qc2.compose(qc, [0, 2, 3], [2, 4], inplace=True)
-        cregs = [creg.name for creg in qc2.cregs]
+        qc3 = QuantumCircuit.copy_empty_like(qc2)
+        qc3.ry(np.pi / 4, 2)
+        qc3.cx(2, 1)
+        qc3.cx(0, 1)
+        qc3.h(0)
+        qc3.measure(0, 2)
+        qc3.measure(1, 4)
```
ElePT: it looks like qc2 is not used, so maybe `qc3` can replace `qc2` directly?

ElePT (commented):
The fix looks reasonable to me.

test/python/primitives/test_statevector_sampler.py
t-imamichi: I used copy_empty_like to copy the aliased registers.
I updated the circuit and comments to clarify it.

t-imamichi: If there is more straightforward way to make aliased registers, I would like to use it. This case does not look so intuitive.

ElePT: got it, I am not so familiar with aliased registers either, but the comment already makes it much clearer 👍

ElePT (approved):
LGTM, I think we can try to get the fix into 1.2, and potentially backport to 1.1? what do you think @t-imamichi

ihincks (approved):
Thanks :)

ElePT:
[@Mergifyio](https://github.com/Mergifyio) backport stable/1.1

t-imamichi:
Thanks. It makes sense to backport to 1.1

PR #12848, opened 2024-07-29
Fix dag visualization with Var wires

mtreinish:
### Summary

This commit fixes the dag visualization for DAGs with classical variables. The Var type was not handled in the attribute callback functions for nodes and edges. This was causing the visualizer to fail if the dag contained these types. This fixes this by adding explicit handling for the Var types and using the name attribute of the Var object.

### Details and comments


jakelishman (commented):
This generally looks fine, thanks for the catch.  Is there a simple regression test we can throw in for it?  Also: bugfix release note.

mtreinish:
I'm skeptical of a test here, mostly because it's a visualization test and the only thing I can do besides asserting it doesn't fail is compare the output visualization. I'm not a big fan of image diff tests, and in this case will be very tricky for me to write because I have a newer version of graphviz installed locally and none of the existing graphviz based visualization tests pass locally.

jakelishman:
For visualisation, I'm basically fine just asserting that it doesn't fail for the reasons you give. It would be good to make sure that the path is at least _run_ in the tests.

jakelishman (approved):
Thanks!

jakelishman:
@Mergifyio backport stable/1.1 stable/1.2

PR #12850, opened 2024-07-30
Update transpiler pipeline to (only) use target internally

### LINKED ISSUE #9256: Migrate to `Target` as only backend input for transpiler passes

ElePT:
### Summary

This PR is a step towards the target-only transpiler pipeline. It manages to almost exclusively use the target input to communicate transpilation constraints. To do this, I designed a (very simple) "FakeTarget" (name tbd) designed for internal use that augments the Target data model supporting the use-case of coupling map provided with no basis gates, and modified the preset pass manager creation mechanisms to contemplate this new use-case.

The creation of `FakeTarget` is not documented as part the public API, as it's designed for internal use.

### Details and comments
This PR addresses the following points from #9256:

- [x] Update preset pass manager construction to only pass a Target to passes. Stage plugins initialization will need to fill out all the fields in the PassManagerConfig from the Target for backwards compatibility. We can stop doing this as we deprecate non-target constraint fields in PassManagerConfig.

- [x]  Deprecate anything we previously could represent that is lost in conversion to Target (or alternatively augment Target's data model to support these edge cases). 

**Note:** There are 2 use cases where the Target is still skipped in the pass manager construction. Both are deprecated and will be removed in 2.0:

1. Whenever a custom basis gate is provided in `basis_gates` ->  this use-case is now deprecated, the recommended alternative in these cases is to construct a `Target` user-side, which can be done through the `Target.from_configuration` method (there's a `custom_name_mapping` argument that allows to use custom gate definitions). Until the deprecation period is over, we still allow custom basis gates by skipping the target.

2. When `instruction_durations` are provided without a target or backend. The argument is deprecated, but we must support the use-case until the deprecation period is over.


ElePT:
As of [9078df2](https://github.com/Qiskit/qiskit/pull/12850/commits/9078df2010d0e0f251b269fcc04bd319cce12947), there are 347 failing tests that must be classified into one of these 2 use cases depending on the different input combinations:

1. User path we can support and must be updated to use target
2. User path we can't/ don't want to support and must be deprecated

More updates will follow. 

```
======
Totals
======
Ran: 13302 tests in 180.7444 sec.
 - Passed: 12876
 - Skipped: 79
 - Expected Fail: 0
 - Unexpected Success: 0
 - Failed: 347
Sum of execute time for each test: 2002.1354 sec.
```

ElePT:
Status after [df78b1c](https://github.com/Qiskit/qiskit/pull/12850/commits/df78b1c459cca906f9ac387e1051253bd2d6b64d):

```
======
Totals
======
Ran: 13562 tests in 234.3214 sec.
 - Passed: 13146
 - Skipped: 90
 - Expected Fail: 0
 - Unexpected Success: 0
 - Failed: 326
Sum of execute time for each test: 2640.2146 sec.
```

ElePT:
Status after [8eea4cf](https://github.com/Qiskit/qiskit/pull/12850/commits/8eea4cfb49bd84a4bd2b20c7d344af64960611c8)

```
======
Totals
======
Ran: 16317 tests in 2518.9866 sec.
 - Passed: 16236
 - Skipped: 66
 - Expected Fail: 0
 - Unexpected Success: 0
 - Failed: 15
Sum of execute time for each test: 27668.6588 sec.
```

qiskit/transpiler/preset_passmanagers/generate_preset_pass_manager.py
```diff
@@ -460,7 +470,15 @@ def _parse_basis_gates(basis_gates, backend, inst_map, skip_target):
             return None, name_mapping, skip_target
 
         for inst in instructions:
-            if inst not in standard_gates or inst not in default_gates:
+            if inst not in standard_gates and inst not in default_gates:
```
ElePT, line 331: After this change, I noticed that 2-5 flaky unit tests started to panic when running unitary synthesis. All tests were passing before changing this `or` to an `and`, but I believe this was an oversight that was causing the target to be skipped more than it should have. Now the target is not skipped, which means the Rust variant of unitary synthesis is run most of the time, and I am afraid this might have uncovered a bug in the unitary synthesis Rust code. I will take a closer look next week.

qiskit/transpiler/target.py
```diff
@@ -1303,3 +1324,29 @@ def target_to_backend_properties(target: Target):
             return BackendProperties.from_dict(properties_dict)
     else:
         return None
+
+
+class FakeTarget(Target):
```
eliarbel: Maybe `_PartialTarget` is a better name? Also I would add to its docstring that this is an internal-only class and should not be used directly by the user. Though it's not in the API doc but still it wouldn't hurt to highlight.

qiskit/transpiler/target.py
```diff
@@ -1303,3 +1324,29 @@ def target_to_backend_properties(target: Target):
...
+    """
+    Pseudo-target class for internal use in the transpilation pipeline.
+    Differently to the :class:`.Target` class, this preudo-target is initialized
+    with a `coupling_map` argument that allows to store connectivity constraints
+    without basis gates.
+    """
+
```
eliarbel, line 858: When checking a `FakeTarget` instance in a boolean context, e.g. like:
```python
target = Target.from_configuration(basis_gates=None, coupling_map=CouplingMap([[0,1]]))
if target:
    print("I'm alive")
```

The evaluation fails.


This would be problematic for all the passes that can work with partial target and do a check like `self.target`. Overriding `__len__` can mitigate this. 
Also in the `level_x..` pass managers (e.g. `level_1_pass_managers`) functions, there are checks like this: 
```python
if (coupling_map and not coupling_map.is_symmetric) or (
        target is not None and target.get_non_global_operation_names(strict_direction=True)
    ):
```
`target is not None` will also always evaluate to `False` with `FakeTarget`.

qiskit/transpiler/target.py
```diff
@@ -1303,3 +1324,29 @@ def target_to_backend_properties(target: Target):
...
+    def __init__(self, coupling_map=None, **kwargs):
+        super().__init__(**kwargs)
+        self._coupling_map = coupling_map
+
+    def build_coupling_map(self, two_q_gate=None, filter_idle_qubits=False):
+        return self._coupling_map
+
+    def instruction_supported(self, *args, **kwargs):
```
eliarbel, line 872: `Target.instruction_supported` is implemented in Rust and used in `gates_in_bases.rs` and `gate_direction.rs`. If these get a `FakeTarget` from Python how would this overriden method is used? 

Additionally, and even more important, returning `True` here by default is not a good solution for the `GateDirection` pass. We need to think about a way to solve this.

qiskit/transpiler/target.py
```diff
@@ -1066,59 +1067,79 @@ def from_configuration(
...
-            name_mapping.update(custom_name_mapping)
-
+        if basis_gates is None:
+            # The Target class requires basis_gates.
+            # If there are no basis_gates, we can't build a proper Target instance,
+            # but we can generate a "pseudo" target that holds connectivity constraints
+            # and is compatible with our transpilation pipeline
+            if num_qubits is None and coupling_map is not None:
+                num_qubits = len(coupling_map.graph)
+            target = FakeTarget(
```
eliarbel: Nitpicking alert: 
I'm not a big fan in general of the pattern that a class' constructor instantiates an object which is not of the class itself, even if it's a subclass. And here even more so since now we are basically allowing a user to create the internal `FakeTarget` via a user-facing API  when passing `basis_gates=None` to `Target.from_configuration`. Do we really want to support this use-case? I am not sure, probably not until at least `Target` fully and officially supports partial initialization. 

One way to alleviate this could be to have `generate_preset_pass_manager` the only factory of `FakeTarget` via say `FakeTarget.from_configuration`. This may require refactoring `Target.from_configuration` a bit, namely making the code which adds instructions to the newly instantiated target a class function of `Target` such that `FakeTarget` can also use it it's own `from_configuration` method. Though if only coupling map is supplied none of the code is actually used so maybe this type of refactoring is not even needed if we only want to support coupling-based only initialization of `FakeTarget`.

eliarbel (changes requested):
It seems to me that this PR is probably premature for 1.3. For one, consider the following difference compared to `main`:

```python
qc = QuantumCircuit(2,2)
qc.cx(0,1)

pm = generate_preset_pass_manager(1, coupling_map=CouplingMap([[1,0]]))
tqc = pm.run(qc)
print(tqc)
```

The `main` branch gives this: 
```
         ┌───┐┌───┐┌───┐
q_0 -> 0 ┤ H ├┤ X ├┤ H ├
         ├───┤└─┬─┘├───┤
q_1 -> 1 ┤ H ├──■──┤ H ├
         └───┘     └───┘
    c: 2/═══════════════
```

But on `transpile-only-target-pipeline` branch we get this: 

```
q_0 -> 0 ──■──
         ┌─┴─┐
q_1 -> 1 ┤ X ├
         └───┘
    c: 2/═════ 
```

(GateDirection does not do what it's supposed to do here. See my line comments for more details).

The reason this is not caught by our unit testing is that we don't really have a way to directly compare transpilation results between versions AFAIK. Transpilation tests which do use pre-defined circuits as groundtruth (e.g. like in `test_gate_direction.py`) could have catch this, but I guess none really use `generate_preset_pass_manager` at all or
maybe use it but without just a coupling-map. I do see the `FakeTarget` instantiation branch in `target.py` reported as covered in the coverage report but probably not with those tests that compare output circuits and can flag a difference? 

Anyway, once we have the above concern addressed, we should have explicit tests for the loose constraints and `FakeTarget` execution path to make sure all passes are included as before and we get the expected output circuits.

ElePT:
Hi @eliarbel! Thanks for taking such a detailed look into the PR. I think you are raising some very good points. I definitely agree with avoiding the user exposure to the class, and I will work on an alternative not to generate the `FakeTarget` from `Target.from_configuration`. I also agree with the `target is None` issue, as well as the documentation, that should be easy to fix.

As for the implementation in Rust and use in `GateDirection` (and other passes), I wanted to share a few points:

1. This PR modifies the pass manager generation logic to skip passes that require basis gates, under the assumption that no basis gates means no synthesis/gate direction, etc.  The current status in this branch is that, for the example you mentioned, the `GateDirection` pass is not run at all. Using a callback, this is how the pipeline looks like for the following branches (not saying this is correct, I am willing to revise it):

  `transpile-only-target-pipeline`:
   Passes: ['ContainsInstruction', 'UnitarySynthesis', 'HighLevelSynthesis', 'Unroll3qOrMore', 'InverseCancellation', 'SetLayout', 'TrivialLayout', 'CheckMap', 'FullAncillaAllocation', 'EnlargeWithAncilla', 'ApplyLayout', 'CheckMap', 'FilterOpNodes', 'HighLevelSynthesis', 'Depth', 'FixedPoint', 'Size', 'FixedPoint', 'InverseCancellation', 'GatesInBasis', 'HighLevelSynthesis', 'Depth', 'FixedPoint', 'Size', 'FixedPoint', 'ContainsInstruction']

  ```
  Out:
  q_0 -> 0 ──■──
           ┌─┴─┐
  q_1 -> 1 ┤ X ├
           └───┘
      c: 2/═════
  ```
  `main`:
    Passes: ['ContainsInstruction', 'UnitarySynthesis', 'HighLevelSynthesis', 'Unroll3qOrMore', 'InverseCancellation', 'SetLayout', 'TrivialLayout', 'CheckMap', 'FullAncillaAllocation', 'EnlargeWithAncilla', 'ApplyLayout', 'CheckMap', **'FilterOpNodes', 'UnitarySynthesis', 'HighLevelSynthesis', 'BasisTranslator', 'CheckGateDirection', 'GateDirection', 'Depth', 'FixedPoint', 'Size', 'FixedPoint', 'Optimize1qGatesDecomposition', 'InverseCancellation', 'GatesInBasis', 'Depth', 'FixedPoint', 'Size', 'FixedPoint', 'ContainsInstruction'**]

  ```
  Out:
           ┌───┐┌───┐┌───┐
  q_0 -> 0 ┤ H ├┤ X ├┤ H ├
           ├───┤└─┬─┘├───┤
  q_1 -> 1 ┤ H ├──■──┤ H ├
           └───┘     └───┘
      c: 2/═══════════════
  ```

2. So far, I thought I would be able to deal with this logic purely in Python space because, well, the coupling map doesn't exist in Rust, and that is all this `FakeTarget` contains (I acknowledge that the way it's constructed, being a subclass, this is not very well conveyed). The `instruction_supported` method, as you mentioned, is a bit of an exception, because there are truly no instructions in the fake target, so we have to come up with a convention. Our options here are to set it to `True` or `False` by default. And I believe that we can add a path in Rust space to support this use case where for an empty gate map, we return `true` or `false`. The issue that you pointed out is that some passes expect a `true`, while others expect a `false`.  In the case of `GateDirection` with no basis gates, I am not sure what is the expected behavior. If we provide a coupling map exclusively, should we assume that all instructions in the circuit should follow the direction of the coupling map? 

3. As for the addition of new unit tests for loose-constraints vs. target, I agree with further testing this case you brought up, and you are right in pointing out that we don't have enough sanity checks in the preset pm generation pipeline. However, this is also partially due to the large number of possible combinations that can be generated. I have a notable number of PRs dealing with target vs loose constraints and while things are a bit more standardized now, there has been historically very little understanding of the differences in behavior that come from different input combinations. In a nutshell, this could be a huge can of worms. I guess we now need to decide if we consider it important enough to block this PR until we have thorough tests everywhere, or whether we could find a middle ground where we add some key tests and follow up gradually with improvements in the testing base.

ElePT:
After an offline discussion, we have concluded that:

1. We know how to fix the particular example raised by @eliarbel (the use of `GateDirection` and with the new fake target), by extracting the coupling map in python space and not providing the fake target to the Rust implementation.

2. However, this is just an example of the potential disruption this change could cause in user workflows. And while we don't make any promises on the specific transpiler output (in terms of what gates are used), we might be breaking user assumptions that could be perceived as a change in our interface. For this reason we have decided to leave this PR for 2.0 instead of merging it in a minor release.

The plan is to fix the individual issues raised in the review and expand the test base with some significant tests that could give us more assurance/ a better understanding of the differences in output this PR could cause.

The deprecation warnings and a few fixes are elegible for 1.3 so they have been singled out in #13394.

qiskit/transpiler/target.py
ElePT: `target is not None` does evaluate to `True` because it's an explicit check for `None`, so at least we can confirm that the pass manager generation logic was working as expected (else it would have been a bigger issue). However, it's true that the output of `__len__` is `0`.  I am wondering if it might make more sense not making `FakeTarget` a subclass of `Target` because it inherits all of these method that make little sense for the object we have.

qiskit/transpiler/target.py
ElePT: The class was refactored in [e79f06c](https://github.com/Qiskit/qiskit/pull/12850/commits/e79f06c34cb8af3acef9ce90da6ee160a5ba1b2e), I agree that this wasn't the best pattern. Now there's a clear path for `Target` and `_FakeTarget` creation and the logic is left to the pass manager generation (no user-facing API changes).

qiskit/transpiler/target.py
eliarbel: You're right, `if target is None` is checking the reference itself. But `if target:` is the boolean context where `__len__` is used. There are a couple places in the codebase in which this pattern in used so it's still a potential issue. 

Regarding whether or not to inherit from `Target` - Ideally the best way going forward might be to support partial initialization of `Target` itself. So, for example, in the case of calling `Target.from_configuration` with only a coupling map, internally `Target` will end up containing dummy instructions that will make constructing a coupling map and other functionality like `instruction_supported` work as expected. This will probably require changing the `Target` internals. Or maybe `FakeTarget` can use `Target.add_instruction` using an internal only instruction which might require changes also on the Rust side. 
@raynelfss 

Still I think we'll want to have a subclass like `FakeTarget` or some other special factory anyway as a way to ensure partial target initialization is only supported for internal code callers.

qiskit/transpiler/preset_passmanagers/generate_preset_pass_manager.py
ElePT: The panic was due to a small bug in unitary synthesis fixed here: https://github.com/Qiskit/qiskit/pull/13651

qiskit/transpiler/target.py
```diff
@@ -979,3 +980,49 @@ def target_to_backend_properties(target: Target):
             return BackendProperties.from_dict(properties_dict)
     else:
         return None
+
+
+class _FakeTarget(Target):
+    """
+    Pseudo-target class for INTERNAL use in the transpilation pipeline.
+    It's essentially an empty :class:`.Target` instance with a a `coupling_map`
```
eliarbel:
```suggestion
    It's essentially an empty :class:`.Target` instance with a `coupling_map`
```

eliarbel (changes requested):
So first pass over the renewed PR surfaced this: 
1. We need to remove the deprecated the `instruction_durations` and `timing_constraints` args from `generate_preset_pass_manager`. This is PR is a good place (really the last place :-) ) to do that.
2. Can we stop propagating `coupling_map` and  `basis_gates` to `PassManagerConfig` already in this PR? 

The first point is a blocker for this PR, the second one is a strong nice-to-have but I wouldn't consider as a blocker.

test/python/transpiler/test_target.py
```diff
@@ -1747,3 +1748,20 @@ def test_over_two_qubits_with_coupling(self):
...
+        target = _FakeTarget(coupling_map=cmap)
+        self.assertEqual(target.num_qubits, 0)
+        self.assertEqual(target.build_coupling_map(), cmap)
+
+    def test_fake_from_configuration(self):
+        cmap = CouplingMap([[0, 1]])
+        target = _FakeTarget.from_configuration(coupling_map=cmap)
+        self.assertNotEqual(target, None)
+        self.assertEqual(target.num_qubits, 2)
+        self.assertEqual(target.build_coupling_map(), cmap)
```
eliarbel, line 1767: Since `_FakeTarget` is an internal-only path, we need to make sure it doesn't drift over time towards duplicating `Target` too much. So I would add a check like this:

``` python
     # Make sure _FakeTarget cannot be instantiated if there is enough info for a real Target
     with self.assertRaises(TypeError):
         target = _FakeTarget.from_configuration(basis_gates=["cx"], coupling_map=CouplingMap([[0, 1]]))
```

qiskit/transpiler/preset_passmanagers/common.py
```diff
@@ -85,9 +85,9 @@ class _InvalidControlFlowForBackend:
     # Explicitly stateful closure to allow pickling.
 
     def __init__(self, basis_gates=(), target=None):
-        if target is not None:
+        if target is not None and len(target.operation_names) > 0:
             self.unsupported = [op for op in CONTROL_FLOW_OP_NAMES if op not in target]
-        elif basis_gates is not None:
+        elif basis_gates is not None and len(basis_gates) > 0:
```
eliarbel, line 90: The `len(basis_gates) > 0` check changes how `self.unsupported` is initialized in case `basis_gates` was not passed. Was that intentional?

qiskit/transpiler/preset_passmanagers/builtin_plugins.py
```diff
@@ -546,6 +550,10 @@ def _opt_control(property_set):
                     ContractIdleWiresInControlFlow(),
                 ]
 
+            # Basic steps for optimization level 2:
+            # 1. RemoveIdentityEquivalent
+            # 2. Optimize1qGatesDecomposition
+            # 3. CommutativeCancelation
```
eliarbel:
```suggestion
            # 3. CommutativeCancellation
```

qiskit/transpiler/preset_passmanagers/builtin_plugins.py
```diff
@@ -558,8 +566,14 @@ def _opt_control(property_set):
                     CommutativeCancellation(target=pass_manager_config.target),
                     ContractIdleWiresInControlFlow(),
                 ]
+
+            # Basic steps for optimization level 3:
+            # 1. ConsolidateBlocks
+            # 2. UnitarySynthesis
+            # 3. RemoveIdentityEquivalent
+            # 4. Optimize1qGatesDecomposition
+            # 5. CommutativeCancelation
```
eliarbel:
```suggestion
            # 5. CommutativeCancellation
```

qiskit/transpiler/passes/utils/gate_direction.py
```diff
@@ -81,6 +81,12 @@ def run(self, dag):
             TranspilerError: If the circuit cannot be mapped just by flipping the
                 cx nodes.
         """
+        # Only use "fix_gate_direction_target" if a target exists and target.operation_names
+        # is not empty, else use "fix_gate_direction_coupling".
         if self.target is None:
             return fix_gate_direction_coupling(dag, set(self.coupling_map.get_edges()))
+        elif len(self.target.operation_names) == 0:
```
eliarbel, line 88:
```suggestion
        elif len(self.target.operation_names) == 0:
            # A  _FakeTarget path, no basis gates, just use the coupling map
```

qiskit/transpiler/passes/synthesis/unitary_synthesis.py
```diff
@@ -138,9 +138,11 @@ def __init__(
         self._pulse_optimize = pulse_optimize
         self._natural_direction = natural_direction
         self._plugin_config = plugin_config
-        self._target = target
+        # Bypass target if it doesn't contain any basis gates, as this
```
eliarbel:
```suggestion
        # Bypass target if it doesn't contain any basis gates (i.e it's _FakeTarget), as this
```

qiskit/transpiler/passes/optimization/optimize_1q_decomposition.py
```diff
@@ -81,17 +81,19 @@ def __init__(self, basis=None, target=None):
         """
         super().__init__()
 
-        if basis:
+        if basis and len(basis) > 0:
```
eliarbel, line 83: Why do we need the `len` check if basis is not None?

qiskit/transpiler/passes/optimization/optimize_1q_decomposition.py
```diff
@@ -81,17 +81,19 @@ def __init__(self, basis=None, target=None):
...
             self._basis_gates = set(basis)
         else:
             self._basis_gates = None
-        self._target = target
+        # Bypass target if it doesn't contain any basis gates, as this
```
eliarbel:
```suggestion
        # Bypass target if it doesn't contain any basis gates (i.e. it's a _FakeTarget), as this
```

qiskit/transpiler/passes/optimization/consolidate_blocks.py
```diff
@@ -69,7 +69,9 @@ def __init__(
         """
         super().__init__()
         self.basis_gates = None
-        self.target = target
+        # Bypass target if it doesn't contain any basis gates, as this
```
eliarbel:
```suggestion
        # Bypass target if it doesn't contain any basis gates (i.e. it's a _FakeTarget), as this
```

qiskit/transpiler/passes/basis/basis_translator.py
```diff
@@ -98,15 +98,16 @@ def __init__(self, equivalence_library, target_basis, target=None, min_qubits=0)
             min_qubits (int): The minimum number of qubits for operations in the input
                 dag to translate.
         """
-
         super().__init__()
         self._equiv_lib = equivalence_library
         self._target_basis = target_basis
-        self._target = target
+        # Bypass target if it doesn't contain any basis gates, as this
```
eliarbel:
```suggestion
        # Bypass target if it doesn't contain any basis gates (i.e. it's a _FakeTarget), as this
```

qiskit/transpiler/target.py
```diff
@@ -979,3 +980,49 @@ def target_to_backend_properties(target: Target):
...
+    @classmethod
+    def from_configuration(
+        cls,
+        *args,
+        num_qubits: int | None = None,
+        coupling_map: CouplingMap | list | None = None,
+        **kwargs,
+    ) -> _FakeTarget:
+
+        if num_qubits is None and coupling_map is not None:
```
eliarbel, line 891: This makes sense to me. However, `Target.from_configuration` does not update `self.num_qubits` if it's None. Are we OK with this divergence?

qiskit/transpiler/passes/utils/check_gate_direction.py
```diff
@@ -45,8 +45,15 @@ def run(self, dag):
...
-            if self.target
-            else check_gate_direction_coupling(dag, set(self.coupling_map.get_edges()))
-        )
+        # Only use "check_gate_direction_target" if a target exists and target.operation_names
+        # is not empty, else use "check_gate_direction_coupling".
+        if self.target is None:
+            self.property_set["is_direction_mapped"] = check_gate_direction_coupling(
+                dag, set(self.coupling_map.get_edges())
+            )
+        elif len(self.target.operation_names) == 0:
```
eliarbel, line 54:
```suggestion
        elif len(self.target.operation_names) == 0:
            # A  _FakeTarget path, no basis gates, just use the coupling map
```

eliarbel (changes requested):
I went over the passes, thanks @ElePT for addressing the comments from my initial review!
I think it's a good start for the Target only flow that will facilitate cleaning the pass manager flows now, passing only target.

I've left suggestions and some questions in line.

ElePT:
Thanks Eli! I still need to make sure tests pass (pretty sure there are still a few failures here and there after merging from main), and then I will make sure to apply your comments and update the reno with the input parameter removal from `generate_preset_pass_manager`.

qiskit/transpiler/passes/optimization/optimize_1q_decomposition.py
ElePT: During the preset pm generation, basis_gates is extracted from the target. In the cases where the target has no basis gates, it returns an empty list `[]`, instead of `None`. As far as I understand it there isn't any difference between basis gates `None` and `[]` in terms of how the pass is expected to work, so I added the extra check to catch both cases.

qiskit/transpiler/preset_passmanagers/common.py
ElePT: it was intentional, for the reason explained in the comment about `len(basis_gates)`. There is an additional conversion inside the preset pm generation where, if a user sets `basis_gates = None `, they can only be communicated through the target as `basis_gates = []`. As far as I know, we didn't really associate any special meaning to `basis_gates = []` vs `basis_gates=None`, differently to, for example, `num_qubits` where `0` is different from `None`. So I went ahead and merged both paths into one.

test/python/transpiler/test_target.py
ElePT: Added

qiskit/transpiler/target.py
ElePT: This was added because the layout/routing passes check against `target.num_qubits` quite a lot, and after all, this fake target object is created for almost exclusive use of these passes. Given that it is not intended for external use, I personally am ok with the divergence, but I understand the doubt.

test/python/transpiler/test_target.py
eliarbel: Thanks!

qiskit/transpiler/passes/optimization/optimize_1q_decomposition.py
eliarbel: OK, thanks.

qiskit/transpiler/preset_passmanagers/generate_preset_pass_manager.py
```diff
@@ -88,8 +70,7 @@ def generate_preset_pass_manager(
 
     The target constraints for the pass manager construction can be specified through a :class:`.Target`
     instance, a :class:`.BackendV2` instance, or via loose constraints
-    (``basis_gates``, ``coupling_map``, ``instruction_durations``,
-    ``dt`` or ``timing_constraints``).
+    (``basis_gates``, ``coupling_map``, or ``dt``).
```
eliarbel, line 73:
```suggestion
    (``basis_gates``, ``coupling_map`` or ``dt``).
```

qiskit/transpiler/preset_passmanagers/generate_preset_pass_manager.py
```diff
@@ -123,14 +102,12 @@ def generate_preset_pass_manager(
 
         backend (Backend): An optional backend object which can be used as the
             source of the default values for the ``basis_gates``,
-            ``coupling_map``, ``instruction_durations``,
-            ``timing_constraints``, and ``target``. If any of those other arguments
+            ``coupling_map``, and ``target``. If any of those other arguments
```
eliarbel, line 105:
```suggestion
            ``coupling_map`` and ``target``. If any of those other arguments
```

qiskit/transpiler/preset_passmanagers/generate_preset_pass_manager.py
```diff
@@ -258,61 +198,72 @@ def generate_preset_pass_manager(
...
+
+    # Check if coupling map has been provided (either standalone or through backend)
+    # with user-defined basis_gates, and whether these have 3q or more.
+    if coupling_map is not None and basis_gates is not None:
+        for gate in OVER_3Q_GATES:
+            if gate in basis_gates:
+                raise ValueError(
+                    f"Gates with 3 or more qubits ({gate}) in `basis_gates` or `backend` are "
+                    "incompatible with a custom `coupling_map`. To include 3-qubit or larger "
+                    " gates in the transpilation basis, use a custom `target` instance instead."
```
eliarbel, line 229:
```suggestion
                    " gates in the transpilation basis, use a custom `Target` instance instead."
```

qiskit/transpiler/preset_passmanagers/generate_preset_pass_manager.py
```diff
@@ -258,61 +198,72 @@ def generate_preset_pass_manager(
...
-    # (known edge case).
-    basis_gates, name_mapping, _skip_target = _parse_basis_gates(basis_gates, backend, _skip_target)
+    instruction_durations = _parse_instruction_durations(backend, dt)
+    timing_constraints = _parse_timing_constraints(backend)
     coupling_map = _parse_coupling_map(coupling_map, backend)
+    basis_gates, name_mapping = _parse_basis_gates(basis_gates, backend)
+
+    # Check if coupling map has been provided (either standalone or through backend)
+    # with user-defined basis_gates, and whether these have 3q or more.
+    if coupling_map is not None and basis_gates is not None:
```
eliarbel, line 223: That was a good catch! 👏

releasenotes/notes/remove-deprecated-transpile-args-d6e212a659eec7cb.yaml
```diff
@@ -0,0 +1,50 @@
...
+upgrade_transpiler:
+  - |
+    The following :func:`.transpile` and :func:`.generate_preset_pass_manager` input arguments,
+    deprecated since Qiskit 1.3 , have been removed from the API:
+
+      * `instruction_durations`
+
+      * `timing_constraints`
+    
+    In addition to this, the specification of custom basis gates through the ``basis`` gate argument of
```
eliarbel, line 11:
```suggestion
    In addition to this, the specification of custom basis gates through the ``basis_gates`` argument of
```

eliarbel (approved):
This looks ready for me. Thanks @ElePT for the quick updates especially with the removals and 3Q gates and their related tests.

I've left a few tiny line comments for the docs; let's not block this PR over them but rather apply those in the final docs + reno updates before the final 2.0 release.

ElePT:
Sounds good! Thanks for your review. I can also open a follow-up PR for the docs changes during the rc period.

PR #12858, opened 2024-07-30
Add 100q QV circuit to PGO scripts

mtreinish:
### Summary

To get better coverage for PGO, including more examples of two qubit synthesis, this commit adds a new PGO script to the benchmarks for running a 100q quantum volume circuit. This provides a different dense circuit than the other 100q circuits we were running. At the time the pgo scripts were added this was excluded because it was too slow at the time. However for the upcoming 1.2 release the transpile time has sped up a bit so this isn't such a concern anymore as the local transpile time is now ~2min which seems within the threshold for pgo time.

### Details and comments


ElePT (approved):
LGTM

PR #12860, opened 2024-07-31
Make `ImportOneCell::new` method public.

raynelfss:
### Summary
The following commits aim to make this custom `ImportOneCell` structure more publicly available by exposing the `::new()` method public.

### Details and comments
As we move continue to connect more parts of Python to Rust, there are parts of our code that still live in python and we will need to access these modules/classes through the `PyO3` import API. For these tasks we have the `GILOneCell` that allows us to access python once and keep a reference whatever we call/do. 

To make the import process more accessible, the `ImportOneCell` struct was created with the purpose of creating imports more easily. Since it's already been implemented once, it would be beneficial for us to expose it more to other crates within `qiskit._accelerate`.

### DIFF
--- a/crates/circuit/src/imports.rs
+++ b/crates/circuit/src/imports.rs
@@ -28,7 +28,7 @@ pub struct ImportOnceCell {
 }
 
 impl ImportOnceCell {
-    const fn new(module: &'static str, object: &'static str) -> Self {
+    pub const fn new(module: &'static str, object: &'static str) -> Self {
         Self {
             module,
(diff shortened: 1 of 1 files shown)


jakelishman (approved):
I'd roughly intended that all Python-space imports should only live in the one file of `crates/circuit` so that that Rust-core interaction with the Python library should be limited, but in retrospect, that probably didn't make a huge amount of sense if much of the transpiler is going to live elsewhere (which it will).

PR #12862, opened 2024-07-31
Expose seed in `Estimator` and `StatevectorEstimator` 

### LINKED ISSUE #12519: Incorrect expectation estimation when Reset is used in circuit
### Environment

- **Qiskit version**:  1.1.0
- **Python version**:  3.12.3
- **Operating system**:  Fedora 40 Workstation

### What is happening?

In a circuit (aimed at simulation of a hamiltonian with disipation) I am getting incorrect estimation values of observables. The system I am simulating has two identical (non-interacting) qubit oscilators. For dissipation a non-unitary operation is implemented with an ancilla qubit that is reset at the start of a trotterisation step and measured at the end of the step to get the dissipation factor.

See [GitLab repo](https://gitlab.com/SillieWous/qiskit-bug-report/-/blob/main/bug.ipynb) for an example.

### What should happen?

Since these are two identical systems without interaction they ought to measure the exact same state trajectory. However there are frequently values that do not agree between the two systems.

The image that is in the example (`Out[8]`) illustrates this pretty obviously.

Moreover the way the observables are defined they should always sum to 1 (see `Out[6]` in the example notebook). But at the indices where the image shows odd behaviour this is not the case. Looking at the values of the observables it seems a value of an earlier measurement is used instead of the correct one.

### Any suggestions?

_No response_

t-imamichi:
### Summary

The Estimator implementations based on `Statevector`, i.e., `Estimator` and `StatevectorEstimator` may return expectation values in a stochastic way if input circuits include reset gates for a some subsystems.
This is because `Statevector` behaves in a stochastic way in such a situation and we need to set a random seed to make it reproducible. Here are the details.

https://github.com/Qiskit/qiskit/blob/b75c2e093ebd4417906edfb159e557f3585b7c4e/qiskit/quantum_info/states/statevector.py#L611-L616

This PR introduces a private utility to function to set a random number generator to make the output reprodicible.

Fixes #12519 (result with this PR https://github.com/Qiskit/qiskit/issues/12519#issuecomment-2262272093)

### Details and comments

Based on https://github.com/Qiskit/qiskit/pull/12862#issuecomment-2260809268

```python
import numpy as np
from qiskit.primitives import Estimator, StatevectorEstimator

qc = QuantumCircuit(2)
qc.h(0)
qc.cx(0, 1)
qc.reset(0)

op = SparsePauliOp("ZI")

n = 500

for seed in [123, None]:
    print("seed", seed)
    est1 = Estimator(options={"seed": seed})
    est2 = StatevectorEstimator(seed=seed)
    for _ in range(3):
        result = est1.run([qc for _ in range(n)], n * [op]).result()
        values = result.values
        print("Estimator\t\t", np.mean(values))

        result = est2.run([(qc, n * [op])]).result()
        values = result[0].data.evs
        print("StatevectorEstimator\t", np.mean(values))
    print()
```

main (the values are close to zero, but random)
```
seed 123
Estimator		 0.012
StatevectorEstimator	 0.016
Estimator		 0.008
StatevectorEstimator	 0.04
Estimator		 -0.008
StatevectorEstimator	 -0.008

seed None
Estimator		 0.012
StatevectorEstimator	 0.024
Estimator		 0.028
StatevectorEstimator	 -0.04
Estimator		 0.048
StatevectorEstimator	 -0.02
```

this PR (the values are close to zero and reproducible thanks to `seed=123`)
```
seed 123
Estimator		 0.04
StatevectorEstimator	 0.04
Estimator		 0.04
StatevectorEstimator	 0.04
Estimator		 0.04
StatevectorEstimator	 0.04

seed None
Estimator		 0.068
StatevectorEstimator	 0.012
Estimator		 -0.02
StatevectorEstimator	 0.06
Estimator		 0.04
StatevectorEstimator	 0.032
```

### DIFF
--- a/qiskit/primitives/estimator.py
+++ b/qiskit/primitives/estimator.py
@@ -55,7 +55,7 @@ class Estimator(BaseEstimator[PrimitiveJob[EstimatorResult]]):
         The result of this class is exact if the circuit contains only unitary operations.
         On the other hand, the result could be stochastic if the circuit contains a non-unitary
         operation such as a reset for a some subsystems.
-        The stochastic result can be reproducible by setting ``seed``, e.g.,
+        The stochastic result can be made reproducible by setting ``seed``, e.g.,
         ``Estimator(options={"seed":123})``.
     """
 
--- a/qiskit/primitives/estimator.py
+++ b/qiskit/primitives/estimator.py
@@ -42,8 +42,9 @@ class Estimator(BaseEstimatorV1[PrimitiveJob[EstimatorResult]]):
     :Run Options:
 
         - **shots** (None or int) --
-          The number of shots. If None, it calculates the exact expectation
-          values. Otherwise, it samples from normal distributions with standard errors as standard
+          The number of shots. If None, it calculates the expectation values
+          with full state vector simulation.
+          Otherwise, it samples from normal distributions with standard errors as standard
           deviations using normal distribution approximation.
 
         - **seed** (np.random.Generator or int) --
@@ -111,7 +112,7 @@ class Estimator(BaseEstimatorV1[PrimitiveJob[EstimatorResult]]):
                     f"The number of qubits of a circuit ({circ.num_qubits}) does not match "
                     f"the number of qubits of a observable ({obs.num_qubits})."
                 )
-            final_state = _statevector_from_circuit(circ, seed)
+            final_state = _statevector_from_circuit(circ, rng)
             expectation_value = final_state.expectation_value(obs)
             if shots is None:
                 expectation_values.append(expectation_value)
--- a/qiskit/primitives/estimator.py
+++ b/qiskit/primitives/estimator.py
@@ -22,7 +22,6 @@ import numpy as np
 
 from qiskit.circuit import QuantumCircuit
 from qiskit.exceptions import QiskitError
-from qiskit.quantum_info import Statevector
 from qiskit.quantum_info.operators.base_operator import BaseOperator
 from qiskit.utils.deprecation import deprecate_func
 
@@ -31,7 +30,7 @@ from .primitive_job import PrimitiveJob
 from .utils import (
     _circuit_key,
     _observable_key,
-    bound_circuit_to_instruction,
+    _statevector_from_circuit,
     init_observable,
 )
 
@@ -112,7 +111,7 @@ class Estimator(BaseEstimatorV1[PrimitiveJob[EstimatorResult]]):
                     f"The number of qubits of a circuit ({circ.num_qubits}) does not match "
                     f"the number of qubits of a observable ({obs.num_qubits})."
                 )
-            final_state = Statevector(bound_circuit_to_instruction(circ))
+            final_state = _statevector_from_circuit(circ, seed)
             expectation_value = final_state.expectation_value(obs)
             if shots is None:
                 expectation_values.append(expectation_value)
--- a/qiskit/primitives/estimator.py
+++ b/qiskit/primitives/estimator.py
@@ -50,6 +50,13 @@ class Estimator(BaseEstimator[PrimitiveJob[EstimatorResult]]):
         - **seed** (np.random.Generator or int) --
           Set a fixed seed or generator for the normal distribution. If shots is None,
           this option is ignored.
+
+    .. note::
+        The result of this class is exact if the circuit contains only unitary operations.
+        On the other hand, the result could be stochastic if the circuit contains a non-unitary
+        operation such as a reset for a some subsystems.
+        The stochastic result can be reproducible by setting ``seed``, e.g.,
+        ``Estimator(options={"seed":123})``.
     """
 
     @deprecate_func(
--- a/qiskit/primitives/statevector_estimator.py
+++ b/qiskit/primitives/statevector_estimator.py
@@ -19,13 +19,13 @@ from collections.abc import Iterable
 
 import numpy as np
 
-from qiskit.quantum_info import SparsePauliOp, Statevector
+from qiskit.quantum_info import SparsePauliOp
 
 from .base import BaseEstimatorV2
 from .containers import DataBin, EstimatorPubLike, PrimitiveResult, PubResult
 from .containers.estimator_pub import EstimatorPub
 from .primitive_job import PrimitiveJob
-from .utils import bound_circuit_to_instruction
+from .utils import _statevector_from_circuit
 
 
 class StatevectorEstimator(BaseEstimatorV2):
@@ -151,7 +151,7 @@ class StatevectorEstimator(BaseEstimatorV2):
         for index in np.ndindex(*bc_circuits.shape):
             bound_circuit = bc_circuits[index]
             observable = bc_obs[index]
-            final_state = Statevector(bound_circuit_to_instruction(bound_circuit))
+            final_state = _statevector_from_circuit(bound_circuit, self._seed)
             paulis, coeffs = zip(*observable.items())
             obs = SparsePauliOp(paulis, coeffs)  # TODO: support non Pauli operators
             expectation_value = np.real_if_close(final_state.expectation_value(obs))
--- a/qiskit/primitives/statevector_estimator.py
+++ b/qiskit/primitives/statevector_estimator.py
@@ -45,7 +45,7 @@ class StatevectorEstimator(BaseEstimatorV2):
         The result of this class is exact if the circuit contains only unitary operations.
         On the other hand, the result could be stochastic if the circuit contains a non-unitary
         operation such as a reset for a some subsystems.
-        The stochastic result can be reproducible by setting ``seed``, e.g.,
+        The stochastic result can be made reproducible by setting ``seed``, e.g.,
         ``StatevectorEstimator(seed=123)``.
 
     .. plot::
--- a/qiskit/primitives/statevector_estimator.py
+++ b/qiskit/primitives/statevector_estimator.py
@@ -151,7 +151,7 @@ class StatevectorEstimator(BaseEstimatorV2):
         for index in np.ndindex(*bc_circuits.shape):
             bound_circuit = bc_circuits[index]
             observable = bc_obs[index]
-            final_state = _statevector_from_circuit(bound_circuit, self._seed)
+            final_state = _statevector_from_circuit(bound_circuit, rng)
             paulis, coeffs = zip(*observable.items())
             obs = SparsePauliOp(paulis, coeffs)  # TODO: support non Pauli operators
             expectation_value = np.real_if_close(final_state.expectation_value(obs))
--- a/qiskit/primitives/statevector_estimator.py
+++ b/qiskit/primitives/statevector_estimator.py
@@ -41,6 +41,13 @@ class StatevectorEstimator(BaseEstimatorV2):
     called an estimator primitive unified bloc (PUB), produces its own array-based result. The
     :meth:`~.EstimatorV2.run` method can be given a sequence of pubs to run in one call.
 
+    .. note::
+        The result of this class is exact if the circuit contains only unitary operations.
+        On the other hand, the result could be stochastic if the circuit contains a non-unitary
+        operation such as a reset for a some subsystems.
+        The stochastic result can be reproducible by setting ``seed``, e.g.,
+        ``StatevectorEstimator(seed=123)``.
+
     .. plot::
         :include-source:
 
--- a/releasenotes/notes/fix-estimator-reset-9e7539776df4cac4.yaml
+++ b/releasenotes/notes/fix-estimator-reset-9e7539776df4cac4.yaml
@@ -1,0 +1,16 @@
+---
+fixes:
+  - |
+    Fixed a bug of :class:`.Estimator` and :class:`.StatevectorEstimator` that
+    they may behave in a stochastic way if input circuits include resets for a some subsystems.
+    They behave in a deterministic way if a random seed is given. For example::
+
+      from qiskit.primitives import StatevectorEstimator
+
+      estimator = StatevectorEstimator(seed=123)
+
+    or::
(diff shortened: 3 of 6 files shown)


Cryoris:
tl;dr: It seems good to expose the `seed`, but I think the user has the wrong expectation for what this class does.

---------

While I would agree that we can expose the `seed`, I don't think it fixes the expectation the user describes in the issue. By setting the seed we're effectively fixing the state of the mixed state and averaging over multiple executions won't give the correct result. For example, creating a Bell state and then resetting qubit 1, we would expect qubit 2 to be in a mixture of $|0\rangle$ and $|1\rangle$, but if we set the seed we fix this outcome. E.g.
```python
import numpy as np
from qiskit.primitives import Estimator

qc = QuantumCircuit(2)
qc.h(0)
qc.cx(0, 1)
qc.reset(0)

op = SparsePauliOp("ZI")

n = 100
est = Estimator(options={"seed": 1})  
values = est.run([qc for _ in range(n)], n * [op]).result().values
print(np.mean(values))  # should be ~0, but with the seed it is either +1 or -1
```
If the user wants correct behavior they should probably average over multiple runs or use another primitive?

SillieWous:
(Commenting as the user who filed the issue)
@Cryoris From [Estimator docs](https://docs.quantum.ibm.com/api/qiskit/qiskit.primitives.Estimator)

> shots (None or int) – The number of shots. If None, it calculates the exact expectation values

Since default is `None`, and I do not set it, I do not see how 'exact' can result in stochastic behaviour.

t-imamichi:
~~@Cryoris I see. Using `seed` to make `Statevector` could be confusing for your example. It might be nice to provide a separate option to make `Statevector` deterministic.~~

Updated: the next commit seems to work fine.

t-imamichi:
I updated this PR to make the result of this case https://github.com/Qiskit/qiskit/pull/12862#issuecomment-2260809268  stochastic but reproducible by fixing a random seed.

output of this PR
```
0.06
```

It also works #12519 case too if we set `seed`. 
I noticed that the case #12519 requires reproducible results.
If results are reproducible, the stochastic results due to reset works fine for the case. 
Here is a plot https://github.com/Qiskit/qiskit/issues/12519#issuecomment-2262272093

I updated the document of `Estimator`.  As https://github.com/Qiskit/qiskit/pull/12862#issuecomment-2261132537 mentioned, "exact expectation values" is not an accurate description with resets of subsystems. I copied a phrase from `StatevectorEstimator`.

> If None, it calculates the exact expectation values

t-imamichi:
updated description https://github.com/Qiskit/qiskit/pull/12862#issue-2439612798

Cryoris:
> Since default is None, and I do not set it, I do not see how 'exact' can result in stochastic behaviour.

tl;dr: The statevector class cannot represent classical mixtures (states that require a density matrix representation) and randomly picks one of the valid statevectors.

----

The statevector simulation is only that -- simulation of statevectors by means of unitary evolution 🙂 
Adding non-unitary operations such as resets results in mixed states, which cannot generally be represented by a statevector but would need a density matrix representation. In your case, for example, we can write the reset operation $\mathcal{R}$ using Kraus operators as

$$\mathcal{R}: \rho \mapsto K_1\rho K_1^\dagger + K_2\rho K_2^\dagger$$

with 

```math
K_1 = \begin{pmatrix} 1 & 0 \\ 0 & 0 \end{pmatrix},~ K_2 = \begin{pmatrix} 0 & 0 \\ 1 & 0 \end{pmatrix}.
```

If we start from a Bell state $|\psi\rangle = (|00\rangle + |11\rangle)/\sqrt{2}$ and apply a reset on the second qubit, we perform the operation

```math
|\psi\rangle\langle\psi| \mapsto \rho = (\mathbb 1 \otimes \mathcal{R})(|\psi\rangle\langle\psi|) = \frac{1}{2} \begin{pmatrix} 1 & 0 & 0 & 0 \\ 0 & 0 & 0 & 0 \\ 0 & 0 & 1 & 0 \\ 0 & 0 & 0 & 0 \end{pmatrix}
```

which is mixed ($Tr(\rho^2) = 1/2$) and can be understood as a equiprobable classical mixture of the two pure states $|00\rangle$ and $|01\rangle$. Since the `Statevector` class cannot represent that, it picks either of the two possible states, which makes the "exact" statevector calculation stochastic.

So to get an exact solution to your circuits that include resets, you might want to use a density matrix simulator (Aer should have one). Alternatively you can run a large number of statevector simulations, and average over the results, which equals a manual average over the classical mixture.

SillieWous:
Thank you for taking the time to explain it. I will be reading up on the Kraus operators.

Let me start by saying that I am and was not questioning your expertise. Quite the opposite, you definitely understand the code and physics much better than me. But due to that expertise your expectations are based on the implementation, whereas user expectations come from the documentation.

Therefore when the documentation states 'exact'  it is not strange that the user thinks it is deterministic. Apparently that is not possible due to the way it is implemented (which probably make a lot of sense), but in my opinion it is not fair to assume the user knows implementation details.

As someone with more of a background in EE and software, I see the reset gate as just a (non-unitary) matrix multiplication. My knowledge reaches as far as that non-unitary operations are forbidden in a real circuit, but running locally is not a real circuit. So when I read 'exact', that makes me think no such limitations apply (running locally).

PS
I'm sorry if I come across as rude or something, that is not my intention. I just get triggered slightly at things that come even remotely close to 'blaming the user'.

t-imamichi:
Thank you for the detailed explanation, @Cryoris. Yes, since the statevector class cannot represent the classical mixture, I suggest rewording "exact" in the documentation in this PR.

As for the issue #12519, the script expects the same statevector for all observables and this PR helps do it. Of course, they still need to conduct the experiments multiple times with different seeds to deal with the stochastic statevector behavior due to reset for subsystems.

releasenotes/notes/fix-estimator-reset-9e7539776df4cac4.yaml
```diff
@@ -0,0 +1,16 @@
+---
+fixes:
```
Cryoris: I'm not sure this is a fix, since the behavior before was valid. Maybe rather a feature since it now exposes setting the seed?

test/python/primitives/test_estimator.py
```diff
@@ -355,6 +355,27 @@ def get_op(i):
...
+        qc.reset(0)
+        op = SparsePauliOp("ZI")
+
+        seed = 12
+        n = 250
+        with self.assertWarns(DeprecationWarning):
+            estimator = Estimator(options={"seed": seed})
+            result = estimator.run([qc for _ in range(n)], [op] * n).result()
+        # expectation values should be stochastic due to reset for subsystems
+        np.testing.assert_allclose(result.values.mean(), 0, atol=1e-1)
```
Cryoris: If this runs very fast, could we increase `n` (maybe 1000)? If I did the math right this has still a ~1% chance of failing, which could be quite a lot in the long run. If it is not very fast I think we can also skip this test, as the behavior of `reset` should be tested in the `Statevector` class.

Cryoris:
@SillieWous no worries, happy to discuss 🙂 

@t-imamichi maybe we could add a note to the `StatevectorEstimator` docstring that explaining that it is exact if the circuit contains only unitary operations?

test/python/primitives/test_estimator.py
t-imamichi: n = 1000 takes around 0.3 sec on mac M1 (while n = 250 takes around 0.09 sec).
I will increase the number to n = 1000.

releasenotes/notes/fix-estimator-reset-9e7539776df4cac4.yaml
t-imamichi: You are right. It should be a new feature.

test/python/primitives/test_estimator.py
Cryoris: We discussed with the team and the failure likelihood (now ~0.1%) is still too high as it would likely fail the CI once a day. Could you move that test to `test/randomized`? Then CI will not be blocked on the failures but we can still get a notice 🙂

qiskit/primitives/estimator.py
```diff
@@ -43,13 +42,21 @@ class Estimator(BaseEstimator[PrimitiveJob[EstimatorResult]]):
...
         - **seed** (np.random.Generator or int) --
           Set a fixed seed or generator for the normal distribution. If shots is None,
           this option is ignored.
+
+    .. note::
+        The result of this class is exact if the circuit contains only unitary operations.
+        On the other hand, the result could be stochastic if the circuit contains a non-unitary
+        operation such as a reset for a some subsystems.
+        The stochastic result can be reproducible by setting ``seed``, e.g.,
+        ``Estimator(options={"seed":123})``.
```
Cryoris:
```suggestion
        The stochastic result can be made reproducible by setting ``seed``, e.g.,
        ``Estimator(options={"seed":123})``.
```

qiskit/primitives/statevector_estimator.py
```diff
@@ -41,6 +41,13 @@ class StatevectorEstimator(BaseEstimatorV2):
     called an estimator primitive unified bloc (PUB), produces its own array-based result. The
     :meth:`~.EstimatorV2.run` method can be given a sequence of pubs to run in one call.
 
+    .. note::
+        The result of this class is exact if the circuit contains only unitary operations.
+        On the other hand, the result could be stochastic if the circuit contains a non-unitary
+        operation such as a reset for a some subsystems.
+        The stochastic result can be reproducible by setting ``seed``, e.g.,
```
Cryoris:
```suggestion
        The stochastic result can be made reproducible by setting ``seed``, e.g.,
```

test/python/primitives/test_estimator.py
t-imamichi: Because I fixed `seed`, it does not fail as far as there is no breaking change in `np.random` or `Statevector.reset`.

I printed `result.values.mean()` several times and the values are consistent.

```
(dev-qiskit) ima@mbp3 ~/t/4/q/terra (primitives/estimator-reset)> python -m unittest test/python/primitives/test_estimator.py test/python/primitives/test_statevector_estimator.py
.....0.006
0.032
........................0.006
0.03223976655569642
......
----------------------------------------------------------------------
Ran 35 tests in 1.942s

OK
(dev-qiskit) ima@mbp3 ~/t/4/q/terra (primitives/estimator-reset)> python -m unittest test/python/primitives/test_estimator.py test/python/primitives/test_statevector_estimator.py
.....0.006
0.032
........................0.006
0.03223976655569642
......
----------------------------------------------------------------------
Ran 35 tests in 1.838s

OK
(dev-qiskit) ima@mbp3 ~/t/4/q/terra (primitives/estimator-reset)> python -m unittest test/python/primitives/test_estimator.py test/python/primitives/test_statevector_estimator.py
.....0.006
0.032
........................0.006
0.03223976655569642
......
----------------------------------------------------------------------
Ran 35 tests in 1.834s

OK
```

diff to display mean values.
```diff
diff --git a/test/python/primitives/test_estimator.py b/test/python/primitives/test_estimator.py
index 2c251af65..f17b0a7d7 100644
--- a/test/python/primitives/test_estimator.py
+++ b/test/python/primitives/test_estimator.py
@@ -370,7 +370,8 @@ class TestEstimator(QiskitTestCase):
                 estimator = Estimator(options={"seed": seed})
                 result = estimator.run([qc for _ in range(n)], [op] * n).result()
             # expectation values should be stochastic due to reset for subsystems
-            np.testing.assert_allclose(result.values.mean(), 0, atol=1e-1)
+            np.testing.assert_allclose(z:=result.values.mean(), 0, atol=1e-1)
+            print(z)

             with self.assertWarns(DeprecationWarning):
                 result2 = estimator.run([qc for _ in range(n)], [op] * n).result()
@@ -383,7 +384,8 @@ class TestEstimator(QiskitTestCase):
                 estimator = Estimator(options={"seed": seed})
                 result = estimator.run([qc for _ in range(n)], [op] * n, shots=shots).result()
             # expectation values should be stochastic due to reset for subsystems
-            np.testing.assert_allclose(result.values.mean(), 0, atol=1e-1)
+            np.testing.assert_allclose(z:=result.values.mean(), 0, atol=1e-1)
+            print(z)

             with self.assertWarns(DeprecationWarning):
                 result2 = estimator.run([qc for _ in range(n)], [op] * n, shots=shots).result()
diff --git a/test/python/primitives/test_statevector_estimator.py b/test/python/primitives/test_statevector_estimator.py
index 0cd76550f..2913e8c53 100644
--- a/test/python/primitives/test_statevector_estimator.py
+++ b/test/python/primitives/test_statevector_estimator.py
@@ -321,7 +321,8 @@ class TestStatevectorEstimator(QiskitTestCase):
         with self.subTest("precision=0"):
             result = estimator.run([(qc, [op] * n)]).result()
             # expectation values should be stochastic due to reset for subsystems
-            np.testing.assert_allclose(result[0].data.evs.mean(), 0, atol=1e-1)
+            np.testing.assert_allclose(z:=result[0].data.evs.mean(), 0, atol=1e-1)
+            print(z.item())

             result2 = estimator.run([(qc, [op] * n)]).result()
             # expectation values should be reproducible due to seed
@@ -331,7 +332,8 @@ class TestStatevectorEstimator(QiskitTestCase):
             precision = 0.01
             result = estimator.run([(qc, [op] * n)], precision=precision).result()
             # expectation values should be stochastic due to reset for subsystems
-            np.testing.assert_allclose(result[0].data.evs.mean(), 0, atol=1e-1)
+            np.testing.assert_allclose(z:=result[0].data.evs.mean(), 0, atol=1e-1)
+            print(z.item())

             result2 = estimator.run([(qc, [op] * n)], precision=precision).result()
             # expectation values should be reproducible due to seed
```

Cryoris (approved):

PR #12869, opened 2024-07-31
Fix ``AnnotatedOperation`` being modified during parameter assignment

Cryoris:
### Summary

Copying an annotated operation currently does not copy the underlying `base_op`. This leads to problems down the line, as we have multiple annotated operations referencing the same `base_op`, for example in parameter assignment. This PR copies the `base_op` upon copying the annotated operation to fix this problem.

### Details and comments

No reno (yet), as this bug was probably mostly noticable in this release (1.2) which adds parameter handling to `AnnotatedOperation`.

Here's a minimal breaking example, which is fixed by this PR.
```python
from qiskit.circuit import Parameter, QuantumCircuit
from qiskit.circuit.library import RYGate

theta = Parameter("theta")
ry = RYGate(theta)
qc = QuantumCircuit(5)
qc.append(ry.control(4, annotated=True), qc.qubits)
qc.assign_parameters([1.23], inplace=True)
print(qc)  # prints the circuit, which should be unmodified, with a bound parameter

qc.assign_parameters([1.23])  # fails as the parameter is already bound
```

### DIFF
--- a/qiskit/circuit/annotated_operation.py
+++ b/qiskit/circuit/annotated_operation.py
@@ -133,7 +133,7 @@ class AnnotatedOperation(Operation):
 
     def copy(self) -> "AnnotatedOperation":
         """Return a copy of the :class:`~.AnnotatedOperation`."""
-        return AnnotatedOperation(base_op=self.base_op, modifiers=self.modifiers.copy())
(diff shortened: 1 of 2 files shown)


ElePT (approved):
LGTM

PR #12871, opened 2024-07-31
Revert base class change introduced in Primitives V1 deprecation

ElePT:
### Summary
https://github.com/Qiskit/qiskit/pull/12575 deprecated the non-versioned aliases for the V1 interfaces of the primitives as well as the V1 implementations. The base class of the deprecated implementations was changed from the non-versioned to the versioned class, and looking back I think this change should be reverted.

This change breaks instance checks on implementations based on the primitives, and prevents the deprecation warning from being raised properly. Given that the implementations are deprecated and will be removed as the same time as the base class aliases, we didn't really gain much from changing to an un-deprecated base class, and I think it is safer to keep the base classes unchanged and preserve the instance checks.

### DIFF
--- a/qiskit/primitives/backend_estimator.py
+++ b/qiskit/primitives/backend_estimator.py
@@ -36,7 +36,7 @@ from qiskit.transpiler.passes import (
 )
 from qiskit.utils.deprecation import deprecate_func
 
-from .base import BaseEstimatorV1, EstimatorResult
+from .base import BaseEstimator, EstimatorResult
 from .primitive_job import PrimitiveJob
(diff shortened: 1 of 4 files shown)


mtreinish (approved):

1ucian0 (commented):
Does the reno from https://github.com/Qiskit/qiskit/pull/12575 need to be changed?

ElePT:
@1ucian0 This was a "side effect" form the original PR, it doesn't affect the deprecation or the reno (double checked the latest version of the reno from #12835, not affected) 👍

1ucian0 (approved):

PR #12882, opened 2024-08-01
Bump macos version in wheels gha workflow

mtreinish:
### Summary

During the 1.2.0rc1 release process we had a failure to build the x86_64 macOS jobs. This was because the job was still trying to use the macos-11 runner which is no longer supported. [1] This commit fixes this by bumping the version in the wheel job to macos-12.

### Details and comments

[1] https://github.blog/changelog/2024-05-20-actions-upcoming-changes-to-github-hosted-macos-runners/

### DIFF
--- a/.github/workflows/wheels.yml
+++ b/.github/workflows/wheels.yml
@@ -12,7 +12,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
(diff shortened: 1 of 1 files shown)


ElePT (approved):
LGTM

PR #12884, opened 2024-08-01
Fix setter so that `SparsePauliOp.paulis.phase` stays zero

### LINKED ISSUE #12883: several `SparsePauliOp` methods ignore `self.paulis.phase`
### Environment

- **Qiskit version**: 1.1.1
- **Python version**: 3.11.6
- **Operating system**: macos

### What is happening?

If I understand it right, `SparsePauliOp.simplify()` does not correctly take into account the `phase` attribute of its underlying `PauliList`.

The same also seems true for several `SparsePauliOp` methods under the header "LinearOp Methods", like `conjugate`, and `adjoint`, which update `self.coeffs` but AFAIK do not look at `self.paulis.phase`.

(I haven't looked exhaustively for other methods that do not handle `paulis.phase`.)

For example, define a `SparsePauliOp`, and then set its `paulis` attribute to a new value with nonzero `phase`:
```
spo = qi.SparsePauliOp(['X', 'X'])
spo.paulis = ['X','-X']
print(spo)
```
```
SparsePauliOp(['X', '-X'],
              coeffs=[1.+0.j, 1.+0.j])
```
So far so good. But now call simplify:
```
spo.simplify()
```
```
SparsePauliOp(['X'],
              coeffs=[2.+0.j])
```
AFAIK this is incorrect. 

### What should happen?

The example above should instead match this result where we construct the SPO differently:
```
spo_2 = qi.SparsePauliOp(['X','-X'])
spo_2.simplify()
```
```
SparsePauliOp(['I'],
              coeffs=[0.+0.j])
```

### Any suggestions?

Either we need to fix all the methods to consider `paulis.phase`, or (probably better) we need to more strictly require that the `paulis` attribute never has any phase, e.g. updating the `paulis()` setter to move the phase into the `coeffs` (don't know what else would need to change).

Another example:
```
spo_1 = qi.SparsePauliOp(['X'])
spo_1.paulis = qi.PauliList(['-1jX'])
spo_1
```
```
SparsePauliOp(['-iX'],
              coeffs=[1.+0.j])
```
Now try `conjugate()`:
```
spo_1.conjugate()
```
```
SparsePauliOp(['-iX'],
              coeffs=[1.-0.j])
```

aeddins-ibm:
### Summary
Goal is to fix #12883.

Currently several `SPO` methods assume that `SPO.paulis.phase` is all zero. However nonzero phase is allowed by the `paulis()` setter, leading to incorrect results when those other methods are called. 

### Details and comments
In this small PR we update the `paulis()` setter to absorb the phase into the SPO `coeffs`.


aeddins-ibm:
Trying to decide between a few approaches.

Approaches I'm leaning away from:
1. Copy `input_paulis` before assigning to `self.paulis`. This has some overhead, and is a change from the current setter which does not copy. And it does not leave a public way to set `self.paulis` without copying.
2. _Current state of this PR_: Don't copy the input, and also don't mutate the input. Make `self.paulis.x` (`z`) the same object as `input_paulis.x` (`z`) going forward. But `self.paulis.phase` will not be the same object as `input_paulis.phase`. This could lead to weird, subtle bugs as some later operations will update both `self.paulis` and `input_paulis`, while others update only one of them.

Proposed new approach is to mutate the input and raise a warning:
```
if np.any(input_paulis.phase):
    warnings.warn(warning_message, UserWarning) 
    self.coeffs *= (-1j) ** input_paulis.phase
    input_paulis.phase = 0
self.paulis = input_paulis
```
I think that `SparsePauliOp(input_paulis, copy=False)` basically does this mutation currently, so there is some precedent for it. But it seems more surprising here, where we mutate what's on the RHS of an `=`. So I think it's worth including the warning if `input_paulis.phase` is nonzero.

Alternatively we could raise an _error_ when `input_paulis.phase` is nonzero, but I think it's better to support the ability to assign a phased `PauliList`, as long as the user is aware of what's happening. An error would also be more disruptive for existing code.

I don't have experience making this kind of call in a large codebase, so am very open to suggestions, but will try the warning for now.

qiskit/quantum_info/operators/symplectic/sparse_pauli_op.py
```diff
@@ -256,6 +257,14 @@ def paulis(self, value):
             raise ValueError(
                 f"incorrect number of operators: expected {len(self.paulis)}, got {len(value)}"
             )
+        if np.any(value.phase):
```
Cryoris: Since the initializer also silently maps the phase to the coefficients, I think we don't need a warning here, as this is the standard behavior. Rather, the setter was inconsistent with the initializer 🙂

test/python/quantum_info/operators/symplectic/test_sparse_pauli_op.py
```diff
@@ -1097,6 +1097,21 @@ def test_paulis_setter_rejects_bad_inputs(self):
         with self.assertRaisesRegex(ValueError, "incorrect number of operators"):
             op.paulis = PauliList([Pauli("XY"), Pauli("ZX"), Pauli("YZ")])
 
+    def test_paulis_setter_absorbs_phase(self):
```
Cryoris, line 1100: Could you also add a test to check the `simplify()` behavior you described in your original issue?

Cryoris (commented):
Nice fix! Just two small comments below.

qiskit/quantum_info/operators/symplectic/sparse_pauli_op.py
aeddins-ibm: Got it, thanks for the feedback on this.

I'm still a little concerned users may get burned if they try to use the input `PauliList` for other stuff after the assignment, like
```
spo_storing_interim_result.paulis = paulis_with_phase
final_result = some_function_that_depends_critically_on_phase(paulis_with_phase)
```
but agree that 1) this PR is consistent with the initializer, 2) the above example is a somewhat unusual code structure, 3) as Jake mentioned on the issue page, this PR is kind of a band-aid since really `SparsePauliOp` should not have a phase attribute at all.

I'll go ahead and remove the warning per your suggestion.

test/python/quantum_info/operators/symplectic/test_sparse_pauli_op.py
aeddins-ibm: Added.

qiskit/quantum_info/operators/symplectic/sparse_pauli_op.py
jakelishman: Andrew: I'd suggest that that code example is a problem outside `SparsePauliOp` - after the line `spo_storing_interim_result.paulis = paulis_with_phase`, it's not valid to make any assumptions about the content of `paulis_with_phase`, because the `SparsePauliOp` owns the object.  I think in general, users can hopefully either re-order the functions so that they still haven't given away ownership of the object while they're using it, or they explicitly copy it if they're still using it.

Thanks for thinking about this, though - it's good to have performance and/or mutability concerns in mind when dealing with Python.

Cryoris (approved):
Thanks for the fix! 🙂

PR #12889, opened 2024-08-02
Avoid exporting incorrect `PyInit_*` symbols

jakelishman:
### Summary

Using the `#[pymodule]` derive macro in PyO3 0.21 always causes a `PyInit_*` symbol with a matching name to be exported in the output `cdylib`.  This is required for the top-level module, in order for Python to import it---it needs to know which symbol in a shared library file it should call---but submodules must be manually initialised, so do not need it.  Including it is typically harmless (and something we've been doing for a long time), but it is technically against the coding rules for CPython extensions[^1].

Recent versions of `abi3audit` (0.0.11+) have tightened their symbol checkers to better match the CPython guidelines, which causes our wheels to be rejected by their audits.  This is, in theory, not a break of abi3 because CPython could never introduce an API-elvel `PyInit_*` function themselves without causing problems, so there ought to be no problems for our users, even with future Python versions. That said, we still want to pass the audit, because the coding guidelines are useful.

This commit is not the cleanest way of doing things.  PyO3 0.22 includes a `#[pymodule(submodule)]` option on the attribute macro, which lets us use all the standard code generation while suppressing the unnecessary `PyInit_*` symbol.  When we are ready to move to PyO3 0.22, we probably want to revert this commit to switch to that form.

[^1]: https://docs.python.org/3/c-api/intro.html

### Details and comments

We found this during the deployment of 1.2.0rc1, since that's the most recent PyPI release since `abi3audit` released version 0.0.11.  This commit creates a wheel that passes the audit at its current version 0.0.14.

### DIFF
--- a/crates/accelerate/src/convert_2q_block_matrix.rs
+++ b/crates/accelerate/src/convert_2q_block_matrix.rs
@@ -145,7 +145,6 @@ pub fn collect_2q_blocks_filter(node: &Bound<PyAny>) -> Option<bool> {
     }
 }
 
-#[pymodule]
 pub fn convert_2q_block_matrix(m: &Bound<PyModule>) -> PyResult<()> {
     m.add_wrapped(wrap_pyfunction!(blocks_to_matrix))?;
     m.add_wrapped(wrap_pyfunction!(collect_2q_blocks_filter))?;
--- a/crates/accelerate/src/dense_layout.rs
+++ b/crates/accelerate/src/dense_layout.rs
@@ -244,7 +244,6 @@ pub fn best_subset_inner(
     [rows, cols, best_map]
 }
 
-#[pymodule]
 pub fn dense_layout(m: &Bound<PyModule>) -> PyResult<()> {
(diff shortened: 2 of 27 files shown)


mtreinish:
We'll probably need to do this on `stable/1.1` too so we can push out 1.1.2, but I think things between main/1.2 and 1.1 have diverged enough it's not worth attempting a backport and we should just do it manually there.

jakelishman:
Yeah, I can open another PR to do that there.  It should be a smaller job.

mtreinish (approved):
The code LGTM, I haven't personally validated the symbols being exported in the shared library but I trust that you tested this locally and it matches the code change so I'm good with merging this now.

jakelishman:
Fwiw, what I did to test was to pull this commit into a Linux VM (needs to be Linux, since `abi3audit` only does this strictest check on ELF libraries), run `pip wheel .` on the library, then use `abi3audit` on the result Qiskit wheel.  I verified that it fails the audit without the commit and passes with it.

PR #12890, opened 2024-08-02
Support commutations of parameterized gates

Cryoris:
### Summary

Add support for basic parameterized gates in the commutation checker (props to @MarcDrudis). This is currently only done for rotation gates that have a single generator (e.g. done for `RXX` but not for `XXPlusYY`). This has the power to improve compiler quality, but also speeds up the `CommutationChecker` in cases where we have to check if gates are parameterized a lot. 

### Details and comments

This works by replacing a parameterized gate by it's generator. For example, `RZ(theta)` commutes with a gate exactly if `Z` commutes with that gate. The special cases of `theta mod 2pi = 0` are also handled. For compiling a 100-qubit QFT circuit (which has loads of `CPhase` gates), the runtime on my laptop was reduced as:
```
(main) Took 2.1887816429138183 +- 0.3050131081682874s
(this PR) Took 1.8230558633804321 +- 0.19875805494141444s
``` 
which is more than I expected given the small substitution.


qiskit/dagcircuit/dagdependency.py
```diff
@@ -461,14 +461,27 @@ def _update_edges(self):
...
                 prev_node = self.get_node(prev_node_id)
 
-                if not self.comm_checker.commute(
-                    prev_node.op,
-                    prev_node.qargs,
-                    prev_node.cargs,
-                    max_node.op,
-                    max_node.qargs,
-                    max_node.cargs,
+                # TODO For some reason, DAGDependency needs gates *not* to commute if they are
```
Cryoris: Before this change, DAGDependency used to look like (and still does with this hacky flag)
image

but since P and RZ do commute, it would actually look like this
image

However, correctly taking into account these commutations breaks the template matching -- it seems like this is something the template matching should be able to handle, but for now I added this flag.

Cryoris: Resolved -- thanks @alexanderivrii !

Cryoris:
Outdated since the port of commutation checker to rust

PR #12898, opened 2024-08-03
Remove init peephole optimization discrete basis check

mtreinish:
### Summary

In #12727 a check was added to the default init stage's construction to avoid running 2q gate consolidation in the presence of targets with only discrete gates. However the way the target was being used in this check was incorrect. The name for an instruction in the target should be used as its identifier and then if we need the object representation that should query the target for that object based on the name. However the check was doing this backwards getting a list of operation objects and then using the name contained in the object. This will cause issues for instructions that use custom names such as when there are tuned variants or a custom gate instance with a unique name.

While there is some question over whether we need this check as we will run the consolidate 2q blocks pass as part of the optimization stage which will have the same effect, this opts to just fix the target usage for it to minimize the diff. Also while not the explicit goal of this check it is protecting against some bugs in the consolidate blocks pass that occur when custom gates are used. So for the short term this check is retained, but in the future when these bugs in consolidate blocks are fixed we can revisit whether we want to remove the conditional logic.

### Details and comments

### DIFF
--- a/qiskit/transpiler/preset_passmanagers/builtin_plugins.py
+++ b/qiskit/transpiler/preset_passmanagers/builtin_plugins.py
@@ -14,7 +14,6 @@
 
 import os
 
-from qiskit.circuit import Instruction
 from qiskit.transpiler.passes.optimization.split_2q_unitaries import Split2QUnitaries
 from qiskit.transpiler.passmanager import PassManager
 from qiskit.transpiler.exceptions import TranspilerError
@@ -47,7 +46,6 @@ from qiskit.transpiler.passes.optimization import (
 from qiskit.transpiler.passes import Depth, Size, FixedPoint, MinimumPoint
 from qiskit.transpiler.passes.utils.gates_basis import GatesInBasis
 from qiskit.transpiler.passes.synthesis.unitary_synthesis import UnitarySynthesis
-from qiskit.transpiler.target import Target
 from qiskit.passmanager.flow_controllers import ConditionalController, DoWhileController
 from qiskit.transpiler.timing_constraints import TimingConstraints
 from qiskit.transpiler.passes.layout.vf2_layout import VF2LayoutStopReason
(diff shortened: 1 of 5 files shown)


qiskit/transpiler/preset_passmanagers/builtin_plugins.py
```diff
@@ -178,28 +179,35 @@ def pass_manager(self, pass_manager_config, optimization_level=None) -> PassMana
...
             stdgates = get_standard_gate_name_mapping()
 
-            def _is_one_op_non_discrete(ops):
+            def _is_one_op_non_discrete(target):
                 """Checks if one operation in `ops` is not discrete, i.e. is parameterizable
                 Args:
                     ops (List(Operation)): list of operations to check
                 Returns
                     True if at least one operation in `ops` is not discrete, False otherwise
                 """
```
jakelishman: This documentation comment isn't accurate now (though I think it wasn't 100% perfect before either).

jakelishman (commented):
I'm getting more confused by the logic of this check.  The name of it (`is_one_op_nondiscrete`) implies `any`-like short-circuiting logic, except there are points that we short-circuit to return `False` _and_ points that we short-circuit to return `True`?

I'm not sure what's up - maybe the function is just named incorrectly - but something still feels wrong about its logic here.

mtreinish:
Fundamentally the check is not solely what its name indicated before or now; it really was "are_any_gates_not_discrete_and_are_all_operations_known" or something like that. Realistically, the piece that's actually needed here is actually the known check; it's protecting us against issues running `ConsolidateBlocks` with custom operations that it can't reason about. I need to open another bug about `ConsolidateBlocks` crashing with custom opaque gates so we can track fixing that too..

jakelishman:
Given the problems we've had here, I'm leaning a bit more towards making the check only the bit that's actually required to match the previous behaviour, since the more superfluous conditions we're adding onto the top, the harder it is to read, and as seen here and before, the easier it is for us to have mistakes in it.

I am still roughly interested in us having a better story for discrete-basis translation in the future, but that's very much a future thing (probably we need a performant `TemplateMatching` before it becomes entirely feasible for optimisations).

mtreinish:
Ok, I pivoted the PR to just remove the check and fix the underlying bug in `ConsolidateBlocks` it was protecting against.

jakelishman (approved):
Thanks - this does seem like the better way to go overall.

Issue #12863, opened 2024-07-31
Organize the synthesis library of multi-controlled gates 


ShellyGarion:
### What should we add?

Currently the synthesis methods of the multi-controlled gates are part of their definition. 
We would like to have a single `MCXGate()` class, and an HLS plugin that checks the number of clean or dirty ancillas, and chooses the optimal synthesis method (see https://github.com/Qiskit/qiskit/pull/12729).

The synthesis library will therefore include a new section **Multi-controlled gates synthesis** which will include the following functions:
- [x] `synth_mcx_no_ancillas` - the current definition of the `MCXGate`
- [x] `synth_mcx_one_clean_ancilla` - the current definition of the `MCXRecursiveGate`
- [x] `synth_mcx_n_dirty_ancillas` - the current definition of the `MCVChainGate`
- [x] `synth_mcx_graysynth_no_ancillas` - the current definition of the `MCXGrayCode` (and the internal `MCU1Gate`). Note that the number of CX gates grows exponentially in the number of qubits.
- [x] `synth_mcp_no_ancillas` - the current definition if the `MCPhaseGate`
- `synth_mcp_graysynth_no_ancillas` - the current definition of the `MCXGrayCode` (and the internal `MCU1Gate`). Only for parametrized gates. Note that the number of CX gates grows exponentially in the number of qubits.
- [x] `synth_mcmt_no_ancillas` - the current definition of the `MCMTGate`
- [x] `synth_mcmt_n_clean_ancillas` - the current definition of the `MCMTVChainGate`
- [x] `synth_mcrx_no_ancillas` - the current synthesis of the [mcrx](https://docs.quantum.ibm.com/api/qiskit/qiskit.circuit.QuantumCircuit#mcrx) method 
- [x] `synth_mcry_no_ancillas` - the current synthesis of the [mcrx](https://docs.quantum.ibm.com/api/qiskit/qiskit.circuit.QuantumCircuit#mcry) method 
- [x] `synth_mcrz_no_ancillas` - the current synthesis of the [mcrz](https://docs.quantum.ibm.com/api/qiskit/qiskit.circuit.QuantumCircuit#mcrx) method 

The following classes will be deprecated in Qiskit 2.0:
- [`MCMTVChain`](https://docs.quantum.ibm.com/api/qiskit/qiskit.circuit.library.MCMTVChain)
[...]

PR #12904, opened 2024-08-05
Move mcx synthesis methods with ancillas to the synthesis library

ShellyGarion:
### Summary
See details in issue #12863.
In this PR we transfer the mcx synthesis methods with clean and dirty ancillas to the synthesis library.

### Details and comments

Added 3 functions in the `qiskit/synthesis/multi-controlled` library:

- `synth_mcx_1_clean_b95`
- `synth_n_dirty_i15`
- `synth_n_clean_m15`

### DIFF
--- a/qiskit/synthesis/multi_controlled/mcx_with_ancillas_synth.py
+++ b/qiskit/synthesis/multi_controlled/mcx_with_ancillas_synth.py
@@ -107,7 +107,7 @@ def synth_mcx_n_dirty_ancillas_ickhc(
     return qc
 
 
-def synth_mcx_n_clean_ancillas(num_qubits: int, num_ctrl_qubits: int = None):
+def synth_mcx_n_clean_ancillas(num_ctrl_qubits: int):
     """Synthesis of an MCX gate with n controls and n-2 clean ancillary qubits,
     producing a circuit with at most 6*n-6 CX gates"""
 
@@ -115,6 +115,7 @@ def synth_mcx_n_clean_ancillas(num_qubits: int, num_ctrl_qubits: int = None):
     from qiskit.circuit.quantumregister import QuantumRegister
     from qiskit.circuit.quantumcircuit import QuantumCircuit
 
+    num_qubits = 2 * num_ctrl_qubits - 1
     q = QuantumRegister(num_qubits, name="q")
     qc = QuantumCircuit(q, name="mcx_vchain")
     q_controls = q[:num_ctrl_qubits]
@@ -140,7 +141,7 @@ def synth_mcx_n_clean_ancillas(num_qubits: int, num_ctrl_qubits: int = None):
     return qc
 
 
-def synth_mcx_one_clean_ancilla_bbcdmssw(num_qubits, num_ctrl: int):
+def synth_mcx_one_clean_ancilla_bbcdmssw(num_ctrl_qubits: int):
     """Implement an MCX gate with n controls using one clean ancilla qubit,
     producing a circuit with at most 16*n-8 CX gates."""
 
@@ -149,16 +150,21 @@ def synth_mcx_one_clean_ancilla_bbcdmssw(num_qubits, num_ctrl: int):
     from qiskit.circuit.quantumcircuit import QuantumCircuit
     from qiskit.circuit.library.standard_gates.x import C3XGate, C4XGate
 
-    q = QuantumRegister(num_qubits, name="q")
-    qc = QuantumCircuit(q, name="mcx_recursive")
-
-    if num_ctrl == 3:
+    if num_ctrl_qubits == 3:
+        q = QuantumRegister(4, name="q")
+        qc = QuantumCircuit(q, name="mcx_recursive")
         qc._append(C3XGate(), q[:], [])
         return qc
 
-    elif num_ctrl == 4:
+    elif num_ctrl_qubits == 4:
+        q = QuantumRegister(5, name="q")
+        qc = QuantumCircuit(q, name="mcx_recursive")
         qc._append(C4XGate(), q[:], [])
         return qc
+
+    num_qubits = num_ctrl_qubits + 2
+    q = QuantumRegister(num_qubits, name="q")
+    qc = QuantumCircuit(q, name="mcx_recursive")
 
     num_ctrl_qubits = len(q) - 1
     q_ancilla = q[-1]
--- a/qiskit/synthesis/multi_controlled/mcx_with_ancillas_synth.py
+++ b/qiskit/synthesis/multi_controlled/mcx_with_ancillas_synth.py
@@ -10,15 +10,15 @@
 # copyright notice, and modified files need to carry a notice indicating
 # that they have been altered from the originals.
 
+"""Module containing multi-controlled circuits synthesis with ancillary qubits."""
+
 from qiskit.circuit.quantumregister import QuantumRegister
 from qiskit.circuit.quantumcircuit import QuantumCircuit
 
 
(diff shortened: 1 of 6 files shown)


Cryoris (commented):
Nice, thanks for the effort! Should we already pending-deprecate the gate classes and refer to the synthesis functions? Once we have the MCX HLS by Sasha we can extend the deprecation message to state that users can also use the HLS plugin.

ShellyGarion:
Note that in commit https://github.com/Qiskit/qiskit/pull/12904/commits/7b68e5745541cb732d961f9781ab69214760162a I had to revert the new synthesized circuit names since QASM based tests are failing.
Should I update the names and fix the tests? or is it considered breaking API?

qiskit/synthesis/multi_controlled/mcx_with_ancillas_synth.py
```diff
@@ -0,0 +1,203 @@
...
+"""Module containing multi-controlled circuits synthesis with ancillary qubits."""
+
+from math import ceil
+
+
+def synth_mcx_n_dirty_i15(
+    num_ctrl_qubits: int,
+    relative_phase: bool = False,
+    action_only: bool = False,
+):
```
alexanderivrii, line 25: Could you please describe the arguments to this function in the docstring (e.g., what are `relative_phase` and `action_only`)?

test/python/circuit/test_controlled_gate.py
```diff
@@ -507,89 +507,37 @@ def test_multi_controlled_u1_matrix(self, num_controls):
             with self.subTest(msg=f"control state = {ctrl_state}"):
                 self.assertTrue(matrix_equal(simulated, expected))
 
-    @data(1, 2, 3, 4)
-    def test_multi_control_toffoli_matrix_clean_ancillas(self, num_controls):
-        """Test the multi-control Toffoli gate with clean ancillas.
```
alexanderivrii, line 512: Do I understand correctly that this and other tests were removed because the new `test_multi_control_toffoli_matrix_advanced_num_ancillas` test handles all of the different cases?

alexanderivrii (commented):
Thanks, this looks great! The new function names, such as `synth_mcx_n_dirty_i15`, `synth_mcx_1_clean_b95`, etc. are reasonable. One small question: once we integrate these into plugins, would it make sense for the plugin names to be `n_dirty_i15`, `1_clean_b95`, etc. as in ``hls = HLSConfig(mcx=['1_clean_b95`])``?

test/python/circuit/test_controlled_gate.py
ShellyGarion: yes

qiskit/synthesis/multi_controlled/mcx_with_ancillas_synth.py
```diff
@@ -0,0 +1,245 @@
...
+
+    if num_ctrl_qubits == 1:
+        qc.cx(q_controls, q_target)
+        return qc
+    elif num_ctrl_qubits == 2:
+        qc.ccx(q_controls[0], q_controls[1], q_target)
+        return qc
+    elif not relative_phase and num_ctrl_qubits == 3:
+        # pylint: disable=cyclic-import
+        from qiskit.circuit.library.standard_gates.x import C3XGate
```
alexanderivrii: Would it be possible to import this at the top-level, and instead add `disable=cyclic-import` to `x.py`?

qiskit/synthesis/multi_controlled/mcx_with_ancillas_synth.py
```diff
@@ -0,0 +1,245 @@
...
+
+def synth_mcx_n_dirty_i15(
+    num_ctrl_qubits: int,
+    relative_phase: bool = False,
+    action_only: bool = False,
+):
+    """
+    Synthesize a multi-controlled X gate with :math:`k` controls using :math:`k - 2`
+    dirty ancillary qubits producing a circuit with at most :math:`8 * k - 6` CX gates,
+    by Iten et. al. [1].
```
alexanderivrii: Can you explicitly state how many qubits will be in the synthesized circuit, i.e. something like "... producing a circuit with 2 * k - 1 qubits at at most ... gates".

qiskit/synthesis/multi_controlled/mcx_with_ancillas_synth.py
```diff
@@ -0,0 +1,245 @@
...
+            break
+
+    return qc
+
+
+def synth_mcx_n_clean_m15(num_ctrl_qubits: int):
+    """
+    Synthesize a multi-controlled X gate with :math:`k` controls using :math:`k - 2`
+    clean ancillary qubits producing a circuit with at most :math:`6 * k - 6` CX gates,
+    by Maslov [1].
```
alexanderivrii: And also say here how many qubits are in the synthesized circuit.

qiskit/synthesis/multi_controlled/mcx_with_ancillas_synth.py
```diff
@@ -0,0 +1,245 @@
...
+    qc.rccx(q_controls[0], q_controls[1], q_ancillas[i])
+
+    return qc
+
+
+def synth_mcx_1_clean_b95(num_ctrl_qubits: int):
+    """
+    Synthesize a multi-controlled X gate with :math:`k` controls using a single
+    clean ancillary qubit producing a circuit with at most :math:`16 * k - 8` CX gates,
+    by Barenco et al. [1].
```
alexanderivrii: and here

releasenotes/notes/add-synth-mcx-with-ancillas-6a92078d6b0e1de4.yaml
```diff
@@ -0,0 +1,17 @@
...
+  - |
+    Add a synthesis function :func:`.synth_mcx_n_dirty_i15` that
+    synthesizes a multi-controlled X gate with :math:`k` controls using :math:`k - 2`
+    dirty ancillary qubits producing a circuit with at most :math:`8 * k - 6` CX gates,
+    by Iten et. al.
+  - |
+    Add a synthesis function :func:`.synth_mcx_n_clean_m15` that
+    synthesizes a multi-controlled X gate with :math:`k` controls using :math:`k - 2`
+    clean ancillary qubits producing a circuit with at most :math:`6 * k - 6` CX gates,
+    by Maslov.
```
alexanderivrii: Do you think it's worth providing the full reference to the paper, the link to arxiv, or at least the name, i.e. instead of saying "by Maslov" say something like "as per the paper 'On the advantages of using relative phase Toffolis with an application to multiple control Toffoli optimization' by Maslov"?

alexanderivrii (commented):
Thanks @ShellyGarion! This looks very good. I had a few minor comments about explicitly adding the information about the total numbers of qubits to the new synthesis function docstrings and to the release notes , moving imports to the top-level, and providing more explicit links to the papers in the release notes. After this is addressed, I would be happy to merge.

qiskit/synthesis/multi_controlled/mcx_with_ancillas_synth.py
ShellyGarion: no, this will not work, since there is a real cyclic import here

releasenotes/notes/add-synth-mcx-with-ancillas-6a92078d6b0e1de4.yaml
ShellyGarion: I'll add a reference to the arxiv.

alexanderivrii (approved):
LGTM, thanks for making the changes!

PR #12906, opened 2024-08-05
Internally importing `qiskit.providers.backend.Backend` should not raise `DeprecationWarning`

### LINKED ISSUE #12902: Importing BackendV2 results in a deprecation warning of `qiskit.providers.models`
### Environment

- **Qiskit version**: main, 1.2.0rc1
- **Python version**: 3.12.4
- **Operating system**: macOS 14.6

### What is happening?

Even if I imported only BackendV2, qiskit raises a deprecation warning of `qiskit.providers.models`, which is related to BackendV1, I guess.

```python
# test.py
from qiskit.providers.backend import BackendV2
```

```
$ python -W error::DeprecationWarning test.py
  File "/Users/ima/tasks/4_2024/qiskit/terra/tmp/test.py", line 1, in <module>
    from qiskit.providers.backend import BackendV2
  File "/Users/ima/envs/qiskit312/lib/python3.12/site-packages/qiskit/__init__.py", line 91, in <module>
    from qiskit.circuit import ClassicalRegister
  File "/Users/ima/envs/qiskit312/lib/python3.12/site-packages/qiskit/circuit/__init__.py", line 1237, in <module>
    from . import library
  File "/Users/ima/envs/qiskit312/lib/python3.12/site-packages/qiskit/circuit/library/__init__.py", line 495, in <module>
    from .generalized_gates import (
  File "/Users/ima/envs/qiskit312/lib/python3.12/site-packages/qiskit/circuit/library/generalized_gates/__init__.py", line 22, in <module>
    from .linear_function import LinearFunction
  File "/Users/ima/envs/qiskit312/lib/python3.12/site-packages/qiskit/circuit/library/generalized_gates/linear_function.py", line 22, in <module>
    from qiskit.quantum_info import Clifford
  File "/Users/ima/envs/qiskit312/lib/python3.12/site-packages/qiskit/quantum_info/__init__.py", line 116, in <module>
    from .analysis import hellinger_distance, hellinger_fidelity, Z2Symmetries
  File "/Users/ima/envs/qiskit312/lib/python3.12/site-packages/qiskit/quantum_info/analysis/__init__.py", line 17, in <module>
    from .z2_symmetries import Z2Symmetries
  File "/Users/ima/envs/qiskit312/lib/python3.12/site-packages/qiskit/quantum_info/analysis/z2_symmetries.py", line 26, in <module>
    from ..operators import Pauli, SparsePauliOp
  File "/Users/ima/envs/qiskit312/lib/python3.12/site-packages/qiskit/quantum_info/operators/__init__.py", line 18, in <module>
    from .measures import average_gate_fidelity, diamond_norm, gate_error, process_fidelity
  File "/Users/ima/envs/qiskit312/lib/python3.12/site-packages/qiskit/quantum_info/operators/measures.py", line 27, in <module>
    from qiskit.quantum_info.states.densitymatrix import DensityMatrix
  File "/Users/ima/envs/qiskit312/lib/python3.12/site-packages/qiskit/quantum_info/states/__init__.py", line 16, in <module>
    from .statevector import Statevector
  File "/Users/ima/envs/qiskit312/lib/python3.12/site-packages/qiskit/quantum_info/states/statevector.py", line 28, in <module>
    from qiskit.quantum_info.states.quantum_state import QuantumState
  File "/Users/ima/envs/qiskit312/lib/python3.12/site-packages/qiskit/quantum_info/states/quantum_state.py", line 27, in <module>
    from qiskit.result.counts import Counts
  File "/Users/ima/envs/qiskit312/lib/python3.12/site-packages/qiskit/result/__init__.py", line 62, in <module>
    from .result import Result
  File "/Users/ima/envs/qiskit312/lib/python3.12/site-packages/qiskit/result/result.py", line 19, in <module>
    from qiskit.pulse.schedule import Schedule
  File "/Users/ima/envs/qiskit312/lib/python3.12/site-packages/qiskit/pulse/__init__.py", line 65, in <module>
    from qiskit.pulse.builder import (
  File "/Users/ima/envs/qiskit312/lib/python3.12/site-packages/qiskit/pulse/builder.py", line 427, in <module>
    from qiskit.pulse import (
  File "/Users/ima/envs/qiskit312/lib/python3.12/site-packages/qiskit/pulse/macros.py", line 22, in <module>
    from qiskit.providers.backend import BackendV2
  File "/Users/ima/envs/qiskit312/lib/python3.12/site-packages/qiskit/providers/__init__.py", line 786, in <module>
    from qiskit.providers.backend import Backend
  File "/Users/ima/envs/qiskit312/lib/python3.12/site-packages/qiskit/providers/backend.py", line 24, in <module>
    from qiskit.providers.models.backendstatus import BackendStatus
  File "/Users/ima/envs/qiskit312/lib/python3.12/site-packages/qiskit/providers/models/__init__.py", line 56, in <module>
    warnings.warn(
DeprecationWarning: qiskit.providers.models is deprecated since Qiskit 1.2 and will be removed in Qiskit 2.0.With the removal of Qobj, there is no need for these schema-conformant objects. If you still needto use them, it could be because you are using a BackendV1, which is also deprecated in favorof BackendV2
```

### What should happen?

It might be nice perhaps to suppress the deprecation warning related to BackendV1 in this case.

### Any suggestions?

_No response_

1ucian0:
Fixes #12902

### DIFF
--- a/qiskit/providers/fake_provider/generic_backend_v2.py
+++ b/qiskit/providers/fake_provider/generic_backend_v2.py
@@ -74,17 +74,25 @@ _QUBIT_PROPERTIES = {
     "frequency": (5e9, 5.5e9),
 }
 
-# The number of samples determines the pulse durations of the corresponding
-# instructions. This default defines pulses with durations in multiples of
-# 16 dt for consistency with the pulse granularity of real IBM devices, but
-# keeps the number smaller than what would be realistic for
-# manageability. If needed, more realistic durations could be added in the
-# future (order of 160dt for 1q gates, 1760dt for 2q gates and measure).
-_PULSE_LIBRARY = [
-    PulseLibraryItem(name="pulse_1", samples=np.linspace(0, 1.0, 16, dtype=np.complex128)),  # 16dt
-    PulseLibraryItem(name="pulse_2", samples=np.linspace(0, 1.0, 32, dtype=np.complex128)),  # 32dt
-    PulseLibraryItem(name="pulse_3", samples=np.linspace(0, 1.0, 64, dtype=np.complex128)),  # 64dt
-]
+
+def _pulse_library():
+    # The number of samples determines the pulse durations of the corresponding
+    # instructions. This default defines pulses with durations in multiples of
+    # 16 dt for consistency with the pulse granularity of real IBM devices, but
+    # keeps the number smaller than what would be realistic for
+    # manageability. If needed, more realistic durations could be added in the
+    # future (order of 160dt for 1q gates, 1760dt for 2q gates and measure).
+    return [
+        PulseLibraryItem(
+            name="pulse_1", samples=np.linspace(0, 1.0, 16, dtype=np.complex128)
+        ),  # 16dt
+        PulseLibraryItem(
+            name="pulse_2", samples=np.linspace(0, 1.0, 32, dtype=np.complex128)
+        ),  # 32dt
+        PulseLibraryItem(
+            name="pulse_3", samples=np.linspace(0, 1.0, 64, dtype=np.complex128)
+        ),  # 64dt
+    ]
 
 
 class GenericBackendV2(BackendV2):
@@ -262,7 +270,7 @@ class GenericBackendV2(BackendV2):
         acting on qargs.
         """
 
-        pulse_library = _PULSE_LIBRARY
+        pulse_library = _pulse_library()
         # Note that the calibration pulses are different for
         # 1q gates vs 2q gates vs measurement instructions.
         if inst == "measure":
@@ -352,7 +360,7 @@ class GenericBackendV2(BackendV2):
             qubit_freq_est=qubit_freq_est,
             meas_freq_est=meas_freq_est,
             buffer=0,
-            pulse_library=_PULSE_LIBRARY,
+            pulse_library=_pulse_library(),
             cmd_def=cmd_def,
         )
 
--- a/qiskit/providers/fake_provider/generic_backend_v2.py
+++ b/qiskit/providers/fake_provider/generic_backend_v2.py
@@ -16,6 +16,7 @@ from __future__ import annotations
 import warnings
 
 from collections.abc import Iterable
+from typing import List, Dict, Any, Union
 import numpy as np
 
 from qiskit import pulse
@@ -35,12 +36,12 @@ from qiskit.transpiler import CouplingMap, Target, InstructionProperties, QubitP
 from qiskit.providers import Options
 from qiskit.providers.basic_provider import BasicSimulator
 from qiskit.providers.backend import BackendV2
-from qiskit.providers.models import (
-    PulseDefaults,
-    Command,
-)
-from qiskit.qobj import PulseQobjInstruction, PulseLibraryItem
 from qiskit.utils import optionals as _optionals
+from qiskit.providers.models.pulsedefaults import Command
+from qiskit.qobj.converters.pulse_instruction import QobjToInstructionConverter
+from qiskit.pulse.calibration_entries import PulseQobjDef
+from qiskit.providers.models.pulsedefaults import MeasurementKernel, Discriminator
+from qiskit.qobj.pulse_qobj import QobjMeasurementOption
 
 # Noise default values/ranges for duration and error of supported
 # instructions. There are two possible formats:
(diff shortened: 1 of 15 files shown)


qiskit/providers/__init__.py
```diff
@@ -779,11 +779,20 @@ def status(self):
...
-from qiskit.providers.backend import Backend
+
+with warnings.catch_warnings():
+    warnings.filterwarnings(
+        "ignore",
+        category=DeprecationWarning,
+        message=r"qiskit\.providers\.models.+",
+        module="qiskit",
+    )
+    from qiskit.providers.backend import Backend
```
jakelishman: The fact that we're doing this means that the warning will not be firing correctly.  If `import qiskit.providers` causes an import of `qiskit.providers.models` internally, then user code subsequently calling `import qiskit.providers.models` will not re-execute the module (since it'll already be in `sys.modules`), and therefore they'll not see the warning.

A correct fix to this should prevent the bad import from happening at all, unless the user specifically does the bad thing.

1ucian0:
@jakelishman I took a different approach, but also not sure if it is the right one. Have a look and let me know.

1ucian0:
Extended the PR because `import qiskit.providers.fake_provider` had the same issue.

qiskit/providers/backend.py
```diff
@@ -172,13 +172,24 @@ def status(self):
...
+            )
+            from qiskit.providers.models.backendstatus import BackendStatus
+
+            return BackendStatus(
+                backend_name=self.name(),
+                backend_version="1",
+                operational=True,
+                pending_jobs=0,
+                status_msg="",
+            )
```
jakelishman: I personally wouldn't put this warning catch in - it's expected that deprecated code will call other deprecated code, and the `stacklevel` of the `BackendStatus` warning should be set such that `qiskit` is "blamed" for the warning, which will mean it's suppressed for users by default (which is good - we want that).  It would show up if all warnings are set to display, but that's expected behaviour.

qiskit/providers/fake_provider/generic_backend_v2.py
```diff
@@ -74,17 +71,34 @@
...
+        PulseLibraryItem(
+            name="pulse_2", samples=np.linspace(0, 1.0, 32, dtype=np.complex128)
+        ),  # 32dt
+        PulseLibraryItem(
+            name="pulse_3", samples=np.linspace(0, 1.0, 64, dtype=np.complex128)
+        ),  # 64dt
+    ]
+
+if typing.TYPE_CHECKING:
+    from qiskit.providers.models import PulseDefaults
```
jakelishman: `TYPE_CHECKING` guards are conventionally put up at the top with the rest of the imports.

jakelishman (commented):
It's good that you've been very specific with the warning filters, but in general, I don't think we should be using `warnings.catch_warnings` like this in library code.  Seems like there's two cases we're currently using it:

1. when we're calling deprecated functionality from a class that is itself deprecated
2. when an eager import involves calling deprecated code

Imo, case 1 doesn't need suppression at all.  The functionality is deprecated already; it's entirely expected that deprecated functions call other deprecated functions, and correctly configured stack levels in the warnings should mean that the deprecated code gets "blamed", not the user / downstream library.  Using the filter makes the code slower to run, harder to read, and runs more risk of inadvertently suppressing a true warning, for (imo) no benefit - users won't see it, and downstream tests that _need_ to test deprecated code paths should understand how to configure their test suite correctly to say "we expect to be blamed for _this_ warning, and we understand that internal deprecation warnings will be issued from `qiskit.providers.model`" - it's a standard part of testing.

In case 2, in general, importing Qiskit should not _use_ deprecated code at all - deprecated code should only run when specifically triggered by a user.  If it's still necessary for us to run, it indicates we're not ready to be deprecating it.  In rare cases where it remains necessary to exist in imports that don't directly touch deprecated code, we can use lazy attribute getters like module-level `__getattr__` to only call the deprecated functionality if/when it's actually used by something.

Context-managers like `catch_warnings` are also inherently non-thread-safe, which means that any downstream libraries that use internal threading (like several of the primitives implementations) can cause breakages to the warning filters, such as permanent application of one of these catch filters, if we have them in library code.

jakelishman (approved):
I'm not _fully_ convinced that it's a great idea for us to be switching our imports to the inner modules - I haven't had chance to check in full detail - but at this point in the release cycle, I'm happy to just do it and go ahead like this.  The fix to the deprecation for `GenericBackendV2` looks like the most important bit - we're now doing what we'd expect others to do, and defining "our own" calibration things, if that's something we're going to expose to users.

We ought to take a note to remove those fake calibrations from `GenericBackendV2` in the future as well, if it's something we're generally dropping support for; `GenericBackendV2` is supposed to the lowest common denominator backend support, so when pulse things are no longer considered part of that, we'll want to drop them there.

Issue #12807, opened 2024-07-23
Migrate CI to github actions


mtreinish:
### What should we add?

We've been using azure pipelines as our primary CI provider for ~5 years now. We have continued to use it because it provided an isolated pool of resources without a share quota with any other repositories. But we're at a point now where the shared quota isn't an issue and running with github actions would likely improve our total commit throughput. We should look at migrating all of our CI job definitions from azure into github actions.

PR #12910, opened 2024-08-06
Raise minimum python version to 3.9

mtreinish:
### Summary

Qiskit 1.2.0 was the final minor version release of qiskit with Python 3.8 support. As Python 3.8 is going EoL before the next Qiskit minor release 1.3.0 we can drop support for running with 3.8 on the main branch now. This commit makes that change and updates everything using python 3.8 currently to use our new minimum instead.

### Details and comments


raynelfss (commented):
This all looks good to me.

I'm thinking that the CI tests might have to do with numpy deprecating `np.product` after the release of [Numpy 1.25](https://numpy.org/devdocs/release/1.25.0-notes.html), this an old release but it also happens to be the point at which Numpy stopped supported Python 3.8. I don't know how deep this issue goes but hopefully this helps shine a light.

mtreinish:
Thanks, yeah it's catching a real issue we'll need to fix by updating the test setup. Since the qpy backwards compatibility tests are installing historical releases of qiskit that predate numpy 1.25 those earlier releases are not compatible with the newer numpy releases. We'll have to update: https://github.com/Qiskit/qiskit/blob/main/test/qpy_compat/process_version.sh#L48-L50 to set a back version for compatibility.

mtreinish:
This should be good to go now (assuming I updated the correct reference images).

.github/workflows/tests.yml
```diff
@@ -18,7 +18,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        # Normally we test min and max version but we can't run python 3.8 or
+        # Normally we test min and max version but we can't run python
         # 3.9 on arm64 until actions/setup-python#808 is resolved
         python-version: ["3.10", "3.12"]
```
raynelfss, line 23: This issue appears to be closed now actions/setup-python#808, are we planning on making any updates?

raynelfss (dismissed):
Thank you for this, glad to see it works well now. Just wondering whether we should remove support for 3.8 in asv too. I think we only need to change this line:
https://github.com/Qiskit/qiskit/blob/86a1b496ac44698422c0428b9d76820148c2a707/asv.conf.json#L20

Other than that, it LGTM!

.github/workflows/tests.yml
mtreinish: We can fix this in a follow-up. There isn't a super big motivation to change it as using 3.10 works fine, but if we could get 3.9 running instead of 3.10 that would at least make everything consistent. But I think we can also look at doing it as part of #12807 too

mtreinish:
> Just wondering whether we should remove support for 3.8 in asv too.

Yes I think we should, I updated this in: https://github.com/Qiskit/qiskit/pull/12910/commits/9d19eae3f717a6c3bb9f87c8f533f16fc1d60b95

raynelfss (approved):
LGTM Thank you for the quick updates 🚀

PR #12911, opened 2024-08-06
Auxiliary qubit tracking in ``HighLevelSynthesis``

Cryoris:
### Summary

Enable tracking the state of auxiliary qubits in high-level synthesis. This allows synthesis plugins to leverage additional qubits in a decomposition, leading to a potential reduction of circuit depth and number of operations (which is why I tagged it for the performance demo 🙂). This PR unblocks work on the circuit library, in particular refactoring of the MCX and MCMT gates.

### Details and comments

This first implementation follows a greedy approach, where each operation is given access to all currently available auxiliary qubits. It does so by using the newly added `QubitTracker`, which keeps track of which qubits in the circuit/dag are "clean" (in state $|0\rangle$) or "dirty" (an unknown state). Generally, if qubits start out as clean, and as soon as it is involved in an operation, it changes to dirty (special rules apply for operations like resets, barriers or delays).

Another important change is that synthesized blocks must be synthesized again. This recursion is necessary, since unrolling or decomposing an operation, could yield other operations that must be synthesized. This slows down the code compared to the current state, but only by another iteration over the DAG, so that should be a reasonably small overhead. 

Follow-ups could involve:
* speedup the code by having the synthesis plugins return DAG circuits (will be easier once we could construct a DAG from Rust's `CircuitData` or an iterator over circuit instructions)
* avoid redundant recursion if plugins can report the basis gate set of their decomposition
* do the whole thing in Rust 🙂 

<details closed>
<summary>The asv utility benchmark finds that this does not significantly reduce performance.</summary>
<br>

```

| Change   | Before [1fdd527f] <main>   | After [9726d19b] <hls+aux~1>   | Ratio   | Benchmark (Parameter)                                                         |
|----------|----------------------------|--------------------------------|---------|-------------------------------------------------------------------------------|
|          | 0                          | 0                              | n/a     | utility_scale.UtilityScaleBenchmarks.track_bvlike_depth('cx')                 |
|          | 0                          | 0                              | n/a     | utility_scale.UtilityScaleBenchmarks.track_bvlike_depth('cz')                 |
|          | 0                          | 0                              | n/a     | utility_scale.UtilityScaleBenchmarks.track_bvlike_depth('ecr')                |
|          | 701±6ms                    | 756±6ms                        | 1.08    | utility_scale.UtilityScaleBenchmarks.time_square_heisenberg('cx')             |
|          | 877±9ms                    | 939±10ms                       | 1.07    | utility_scale.UtilityScaleBenchmarks.time_square_heisenberg('ecr')            |
|          | 3.40±0.02s                 | 3.61±0.08s                     | 1.06    | utility_scale.UtilityScaleBenchmarks.time_qv('ecr')                           |
|          | 3.33±0.01s                 | 3.51±0.08s                     | 1.05    | utility_scale.UtilityScaleBenchmarks.time_qv('cx')                            |
|          | 5.46±0s                    | 5.74±0.06s                     | 1.05    | utility_scale.UtilityScaleBenchmarks.time_qv('cz')                            |
|          | 1.02±0.02s                 | 1.07±0.01s                     | 1.05    | utility_scale.UtilityScaleBenchmarks.time_square_heisenberg('cz')             |
|          | 223±5ms                    | 231±9ms                        | 1.04    | utility_scale.UtilityScaleBenchmarks.time_bv_100('cx')                        |
|          | 1.38±0.01s                 | 1.43±0.01s                     | 1.04    | utility_scale.UtilityScaleBenchmarks.time_qaoa('ecr')                         |
|          | 248±2ms                    | 256±3ms                        | 1.03    | utility_scale.UtilityScaleBenchmarks.time_bv_100('cz')                        |
|          | 15.3±0.01s                 | 15.7±0.1s                      | 1.03    | utility_scale.UtilityScaleBenchmarks.time_qft('ecr')                          |
|          | 238±1ms                    | 243±2ms                        | 1.02    | utility_scale.UtilityScaleBenchmarks.time_bv_100('ecr')                       |
|          | 6.81±0.03ms                | 6.93±0.2ms                     | 1.02    | utility_scale.UtilityScaleBenchmarks.time_parse_qaoa_n100('cx')               |
|          | 917±5ms                    | 931±6ms                        | 1.02    | utility_scale.UtilityScaleBenchmarks.time_qaoa('cx')                          |
|          | 73.4±0.4ms                 | 74.0±0.2ms                     | 1.01    | utility_scale.UtilityScaleBenchmarks.time_bvlike('cx')                        |
|          | 73.5±0.2ms                 | 74.1±0.6ms                     | 1.01    | utility_scale.UtilityScaleBenchmarks.time_bvlike('cz')                        |
|          | 73.3±0.2ms                 | 73.7±0.3ms                     | 1.01    | utility_scale.UtilityScaleBenchmarks.time_bvlike('ecr')                       |
|          | 6.78±0.05ms                | 6.86±0.08ms                    | 1.01    | utility_scale.UtilityScaleBenchmarks.time_parse_qaoa_n100('cz')               |
|          | 6.86±0.08ms                | 6.90±0.05ms                    | 1.01    | utility_scale.UtilityScaleBenchmarks.time_parse_qaoa_n100('ecr')              |
|          | 73.0±0.4ms                 | 73.5±0.3ms                     | 1.01    | utility_scale.UtilityScaleBenchmarks.time_parse_qft_n100('cx')                |
|          | 23.3±0.2ms                 | 23.5±0.2ms                     | 1.01    | utility_scale.UtilityScaleBenchmarks.time_parse_square_heisenberg_n100('cx')  |
|          | 23.3±0.09ms                | 23.7±0.3ms                     | 1.01    | utility_scale.UtilityScaleBenchmarks.time_parse_square_heisenberg_n100('cz')  |
|          | 1.78±0.01s                 | 1.80±0.01s                     | 1.01    | utility_scale.UtilityScaleBenchmarks.time_qaoa('cz')                          |
|          | 14.7±0.03s                 | 14.8±0.01s                     | 1.01    | utility_scale.UtilityScaleBenchmarks.time_qft('cx')                           |
|          | 15.8±0.01s                 | 16.1±0.03s                     | 1.01    | utility_scale.UtilityScaleBenchmarks.time_qft('cz')                           |
|          | 3.84±0.01s                 | 3.84±0.01s                     | 1.00    | utility_scale.UtilityScaleBenchmarks.time_circSU2('cz')                       |
|          | 395                        | 395                            | 1.00    | utility_scale.UtilityScaleBenchmarks.track_bv_100_depth('cx')                 |
|          | 397                        | 397                            | 1.00    | utility_scale.UtilityScaleBenchmarks.track_bv_100_depth('cz')                 |
|          | 397                        | 397                            | 1.00    | utility_scale.UtilityScaleBenchmarks.track_bv_100_depth('ecr')                |
|          | 300                        | 300                            | 1.00    | utility_scale.UtilityScaleBenchmarks.track_circSU2_depth('cx')                |
|          | 300                        | 300                            | 1.00    | utility_scale.UtilityScaleBenchmarks.track_circSU2_depth('cz')                |
|          | 300                        | 300                            | 1.00    | utility_scale.UtilityScaleBenchmarks.track_circSU2_depth('ecr')               |
|          | 1483                       | 1483                           | 1.00    | utility_scale.UtilityScaleBenchmarks.track_qaoa_depth('cx')                   |
|          | 1488                       | 1488                           | 1.00    | utility_scale.UtilityScaleBenchmarks.track_qaoa_depth('cz')                   |
|          | 1488                       | 1488                           | 1.00    | utility_scale.UtilityScaleBenchmarks.track_qaoa_depth('ecr')                  |
|          | 1954                       | 1954                           | 1.00    | utility_scale.UtilityScaleBenchmarks.track_qft_depth('cx')                    |
|          | 1954                       | 1954                           | 1.00    | utility_scale.UtilityScaleBenchmarks.track_qft_depth('cz')                    |
|          | 1954                       | 1954                           | 1.00    | utility_scale.UtilityScaleBenchmarks.track_qft_depth('ecr')                   |
|          | 2538                       | 2538                           | 1.00    | utility_scale.UtilityScaleBenchmarks.track_qv_depth('cx')                     |
|          | 2538                       | 2538                           | 1.00    | utility_scale.UtilityScaleBenchmarks.track_qv_depth('cz')                     |
|          | 2538                       | 2538                           | 1.00    | utility_scale.UtilityScaleBenchmarks.track_qv_depth('ecr')                    |
|          | 435                        | 435                            | 1.00    | utility_scale.UtilityScaleBenchmarks.track_square_heisenberg_depth('cx')      |
|          | 435                        | 435                            | 1.00    | utility_scale.UtilityScaleBenchmarks.track_square_heisenberg_depth('cz')      |
|          | 435                        | 435                            | 1.00    | utility_scale.UtilityScaleBenchmarks.track_square_heisenberg_depth('ecr')     |
|          | 3.85±0.06s                 | 3.80±0.05s                     | 0.99    | utility_scale.UtilityScaleBenchmarks.time_circSU2('ecr')                      |
|          | 73.1±0.9ms                 | 72.7±0.8ms                     | 0.99    | utility_scale.UtilityScaleBenchmarks.time_parse_qft_n100('cz')                |
|          | 73.0±0.5ms                 | 72.4±0.6ms                     | 0.99    | utility_scale.UtilityScaleBenchmarks.time_parse_qft_n100('ecr')               |
|          | 23.4±0.3ms                 | 23.2±0.1ms                     | 0.99    | utility_scale.UtilityScaleBenchmarks.time_parse_square_heisenberg_n100('ecr') |
|          | 3.78±0.1s                  | 3.67±0.02s                     | 0.97    | utility_scale.UtilityScaleBenchmarks.time_circSU2('cx')                       |

```
<\details>


qiskit/synthesis/clifford/clifford_decompose_bm.py
```diff
@@ -20,6 +20,7 @@
 from qiskit._accelerate.synthesis.clifford import (
     synth_clifford_bm as synth_clifford_bm_inner,
 )
+from qiskit.transpiler.passes import Decompose
```
henryzou50: I noticed that `from qiskit.transpiler.passes import Decompose` is an unused import in this file. Removing it should resolve the lint issues you're encountering. Could you please remove this import?

```suggestion
```

qiskit/transpiler/passes/synthesis/high_level_synthesis.py
henryzou50, line 1: I noticed several comments in this file that seem to be outdated or left over from previous implementations. I will mark them.

qiskit/transpiler/passes/synthesis/high_level_synthesis.py
```diff
@@ -418,151 +449,227 @@ def run(self, dag: DAGCircuit) -> DAGCircuit:
...
+            ):
+                pass
+
+            # next check control flow
+            elif node.is_control_flow():
+                node.op = control_flow.map_blocks(
+                    partial(self._run, tracker=tracker.copy()),
+                    node.op,
+                    # self.run,
+                    # node.op,
```
henryzou50: Appears to be outdated

qiskit/transpiler/passes/synthesis/high_level_synthesis.py
```diff
@@ -418,151 +449,227 @@ def run(self, dag: DAGCircuit) -> DAGCircuit:
...
+                    for sub_node in op.op_nodes():
+                        out.apply_operation_back(
+                            sub_node.op, tuple(qubit_map[qarg] for qarg in sub_node.qargs)
+                        )
+                    out.global_phase += op.global_phase
+                    # handle different types of ops
+                    # out.compose(op, qargs, inplace=True)
+                else:
+                    raise RuntimeError(f"Unexpected synthesized type: {type(op)}")
+                # out.apply_operation_back(op, qubits, cargs=())
```
henryzou50: Appears to be outdated

qiskit/transpiler/passes/synthesis/high_level_synthesis.py
```diff
@@ -418,151 +449,227 @@ def run(self, dag: DAGCircuit) -> DAGCircuit:
...
+                    qubit_map = {
+                        qubit: index_to_qubit[index] for index, qubit in zip(qubits, op.qubits)
+                    }
+                    for sub_node in op.op_nodes():
+                        out.apply_operation_back(
+                            sub_node.op, tuple(qubit_map[qarg] for qarg in sub_node.qargs)
+                        )
+                    out.global_phase += op.global_phase
+                    # handle different types of ops
+                    # out.compose(op, qargs, inplace=True)
```
henryzou50: Appears to be outdated

qiskit/transpiler/passes/synthesis/high_level_synthesis.py
```diff
@@ -418,151 +449,227 @@ def run(self, dag: DAGCircuit) -> DAGCircuit:
...
+        #       and then apply the modifiers.
+        #  (2) High-level objects: try running the battery of high-level synthesis plugins
+        #  (3) Unrolling custom definitions: try defining the operation if it is not yet
+        #       in the set of supported instructions
+        # If any of the above were triggered, we will recurse and go again through these steps
+        # until no further change occurred. If there was no change, we just return ``None``.
+        synthesized = None
+
+        # try synthesizing via AnnotatedOperation
+        # if isinstance(operation, AnnotatedOperation):
```
henryzou50: Appears to be outdated

qiskit/transpiler/passes/synthesis/high_level_synthesis.py
```diff
@@ -418,151 +449,227 @@ def run(self, dag: DAGCircuit) -> DAGCircuit:
...
+            # used in the modifiers.
+            num_ctrl = sum(
+                mod.num_ctrl_qubits for mod in modifiers if isinstance(mod, ControlModifier)
+            )
+            baseop_qubits = qubits[num_ctrl:]  # reminder: control qubits are the first ones
+            baseop_tracker = tracker.copy()
+            # dict(zip(baseop_qubits, range(operation.base_op.num_qubits)))
+            # )
+            baseop_tracker.drop(qubits[:num_ctrl])  # no access to control qubits
+            # baseop_tracker.used(qubits[:num_ctrl])  # no access to control qubits
```
henryzou50: Appears to be outdated

qiskit/transpiler/passes/synthesis/high_level_synthesis.py
```diff
@@ -418,151 +449,227 @@ def run(self, dag: DAGCircuit) -> DAGCircuit:
...
+        # if isinstance(operation, AnnotatedOperation):
+        if len(modifiers := getattr(operation, "modifiers", [])) > 0:
+            # The base operation must be synthesized without using potential control qubits
+            # used in the modifiers.
+            num_ctrl = sum(
+                mod.num_ctrl_qubits for mod in modifiers if isinstance(mod, ControlModifier)
+            )
+            baseop_qubits = qubits[num_ctrl:]  # reminder: control qubits are the first ones
+            baseop_tracker = tracker.copy()
+            # dict(zip(baseop_qubits, range(operation.base_op.num_qubits)))
```
henryzou50: Appears to be outdated

qiskit/transpiler/passes/synthesis/high_level_synthesis.py
```diff
@@ -418,151 +449,227 @@ def run(self, dag: DAGCircuit) -> DAGCircuit:
...
+        # synthesize via HLS
+        elif len(hls_methods := self._methods_to_try(operation.name)) > 0:
+            # TODO just pass qubits if use_qubit_indices is removed
+            indices = qubits if self._use_qubit_indices else None
+            synthesized = self._synthesize_op_using_plugins(
+                hls_methods,
+                operation,
+                indices,
+                tracker.num_clean(qubits),
+                tracker.num_dirty(qubits),
```
henryzou50: The TODO seems to be completed and the code does correctly address the conditional check. The TODO here can be removed.

qiskit/transpiler/passes/synthesis/high_level_synthesis.py
```diff
@@ -418,151 +449,227 @@ def run(self, dag: DAGCircuit) -> DAGCircuit:
...
+            if isinstance(synthesized, Operation):
+                re_synthesized, qubits = self._synthesize_operation(
+                    synthesized, qubits, tracker.copy()
+                )
+                if re_synthesized is not None:
+                    synthesized = re_synthesized
+                used_qubits = qubits
+
+                # if synthesized.num_qubits != len(qubits):
+                # raise RuntimeError("currently cannot annotate HLS stuff")
```
henryzou50: Appears to be outdated

qiskit/transpiler/passes/synthesis/high_level_synthesis.py
```diff
@@ -418,151 +449,227 @@ def run(self, dag: DAGCircuit) -> DAGCircuit:
...
+
+            elif isinstance(synthesized, QuantumCircuit):
+                aux_qubits = tracker.borrow(synthesized.num_qubits - len(qubits), qubits)
+                used_qubits = qubits + tuple(aux_qubits)
+                as_dag = circuit_to_dag(synthesized, copy_operations=False)
+                # map used qubits to subcircuit
+                new_qubits = [as_dag.find_bit(q).index for q in as_dag.qubits]
+                qubit_map = dict(zip(used_qubits, new_qubits))
+                # qubit_map = {}
```
henryzou50: Appears to be outdated

qiskit/transpiler/passes/synthesis/high_level_synthesis.py
```diff
@@ -646,79 +768,92 @@ def _synthesize_op_using_plugins(
                     best_decomposition = decomposition
                     best_score = current_score
 
+        # if best_decomposition is None:
+        #     raise TranspilerError(
+        #         f"HLS failed to synthesize {op.name} given the provided methods {hls_methods}."
+        #     )
+
```
henryzou50: Appears to be outdated

qiskit/compiler/transpiler.py
```diff
@@ -66,6 +66,7 @@ def transpile(  # pylint: disable=too-many-return-statements
     optimization_method: Optional[str] = None,
     ignore_backend_supplied_default_methods: bool = False,
     num_processes: Optional[int] = None,
+    qubits_initially_zero: bool = True,
```
henryzou50, line 69: I remember we talked about the variable naming earlier today. This works well, but if you want to try out a shorter name, I think `qubits_zeroed` works here :)

qiskit/transpiler/passes/synthesis/qubit_tracker.py
```diff
@@ -0,0 +1,123 @@
...
+
+    def borrow(self, num_qubits: int, active_qubits: Iterable[int] | None = None) -> list[int]:
+        """Get ``num_qubits`` qubits, excluding ``active_qubits``."""
+        if active_qubits is None:
+            active_qubits = []
+
+        if num_qubits > (available := len(self.qubits) - len(active_qubits)):
+            raise RuntimeError(f"Cannot borrow {num_qubits} qubits, only {available} available.")
+
+        return [qubit for qubit in self.qubits if qubit not in active_qubits][:num_qubits]
```
henryzou50: We can optimized the `borrow` method by converting `active_qubits` to a set for efficient membership testing. This reduces the complexity of the `not in` checks within the list comprehension and simplifies the code. The updated method ensures faster execution, especially when `active_qubits` is large. 
```suggestion
        active_qubits = set(active_qubits or [])
        available_qubits = [qubit for qubit in self.qubits if qubit not in active_qubits]

        if num_qubits > len(available_qubits):
            raise RuntimeError(
                f"Cannot borrow {num_qubits} qubits, only {len(available_qubits)} available."
            )

        return available_qubits[:num_qubits]
```
The approach simplifies the check for available qubits, making it more readable and potentially more efficient.

qiskit/transpiler/passes/synthesis/qubit_tracker.py
```diff
@@ -0,0 +1,123 @@
...
+        if check:
+            if len(untracked := set(qubits).difference(self.qubits)) > 0:
+                raise ValueError(
+                    "Setting state of untracked qubits:", untracked, "\nTracker state:", self
+                )
+
+        for qubit in qubits:
+            self.clean.discard(qubit)
+
+        self.dirty.update(qubits)
```
henryzou50: Nice, this works well, but I think we can optimized this a bit further by converting `qubits` to a set once, which improves the efficiency of set operations. Using set operations (`-=` and `|=`) directly on `self.clean` and `self.dirty` is more efficient than looping through the qubits and performing individual operations.

```suggestion
        qubits = set(qubits)

        if check:
            untracked = qubits - set(self.qubits)
            if untracked:
                raise ValueError(
                    "Setting state of untracked qubits:", untracked, "\nTracker state:", self
                )

        self.clean -= qubits
        self.dirty |= qubits
```

qiskit/transpiler/passes/synthesis/qubit_tracker.py
```diff
@@ -0,0 +1,123 @@
...
+        """Set the state of ``qubits`` to 0 (i.e. True)."""
+        if check:
+            if len(untracked := set(qubits).difference(self.qubits)) > 0:
+                raise ValueError(
+                    "Setting state of untracked qubits:", untracked, "\nTracker state:", self
+                )
+
+        self.clean.update(qubits)
+        for qubit in qubits:
+            self.dirty.discard(qubit)
```
henryzou50: Same idea as above, if we want to change to that. 
```suggestion
        qubits = set(qubits)

        if check:
            untracked = qubits - set(self.qubits)
            if untracked:
                raise ValueError(
                    "Setting state of untracked qubits:", untracked, "\nTracker state:", self
                )

        self.clean |= qubits
        self.dirty -= qubits
```

qiskit/transpiler/passes/synthesis/qubit_tracker.py
```diff
@@ -0,0 +1,123 @@
...
+        if check:
+            if len(untracked := set(qubits).difference(self.qubits)) > 0:
+                raise ValueError(
+                    "Dropping state of untracked qubits:", untracked, "\nTracker state:", self
+                )
+
+        self.qubits = tuple(qubit for qubit in self.qubits if qubit not in qubits)
+        for qubit in qubits:
+            self.clean.discard(qubit)
+            self.dirty.discard(qubit)
```
henryzou50: Same idea as above
```suggestion
        qubits = set(qubits)

        if check:
            untracked = qubits - set(self.qubits)
            if untracked:
                raise ValueError(
                    "Dropping state of untracked qubits:", untracked, "\nTracker state:", self
                )

        self.qubits = tuple(qubit for qubit in self.qubits if qubit not in qubits)
        self.clean -= qubits
        self.dirty -= qubits
```

henryzou50 (changes requested):
Overall, LGTM Great job on the implementation and the thorough testing.

However, I noticed that it fails a lint test. This should be fixed by removing the unused import in `clifford_decompose_bm.py`.

Additionally, there were several sections in `high_level_synthesis.py` that seemed to have outdated comments. It would be helpful to update these to maintain clarity.

I also left some suggestions for small improvements to `qubit_tracker.py` that could enhance its efficiency. These changes involve optimizing set operations and membership testing, which should help improve the performance of the qubit tracking.

Once these minor issues are addressed, it should be good to merge!

qiskit/transpiler/passes/synthesis/high_level_synthesis.py
Cryoris: What I was trying to say is that the line
```
indices = qubits if self._use_qubit_indices else None
```
can be removed once `use_qubit_indices` is deprecated, so more of a note for us in the future 🙂 I'll add a few more words to make it clearer

qiskit/transpiler/passes/synthesis/qubit_tracker.py
Cryoris: Ah great, I was looking for something like `set.update` for removal but didn't know that we can just use `-=`. Thanks!

qiskit/transpiler/passes/synthesis/qubit_tracker.py
```diff
@@ -0,0 +1,123 @@
...
+"""A qubit state tracker for synthesizing operations with auxiliary qubits."""
+
+from __future__ import annotations
+from collections.abc import Iterable
+from dataclasses import dataclass
+
+
+@dataclass
+class QubitTracker:
+    """Track qubits per index as and their state.
```
alexanderivrii:
```suggestion
    """Track qubits by index and their state.
```

qiskit/transpiler/passes/synthesis/qubit_tracker.py
```diff
@@ -0,0 +1,123 @@
...
+
+    The states are distinguished into clean (meaning in state :math:`|0\rangle`) or dirty (an
+    unkown state).
```
alexanderivrii:
```suggestion
    unknown state).
```

qiskit/synthesis/clifford/clifford_decompose_bm.py
Cryoris: whoops, probably a remnant of rebasing -- thanks!

qiskit/transpiler/passes/synthesis/high_level_synthesis.py
Cryoris: Sorry, I should've cleaned that up before opening the PR -- done in  c87cf0b!

qiskit/compiler/transpiler.py
Cryoris: nice idea, or maybe `qubits_reset` if we want to avoid the "zero" 🤔

qiskit/transpiler/passes/synthesis/qubit_tracker.py
```diff
@@ -0,0 +1,123 @@
...
+    def num_clean(self, active_qubits: Iterable[int] | None = None):
+        """Return the number of clean qubits, not considering the active qubits."""
+        # this could be cached if getting the set length becomes a performance bottleneck
+        return len(self.clean.difference(active_qubits or set()))
+
+    def num_dirty(self, active_qubits: Iterable[int] | None = None):
+        """Return the number of dirty qubits, not considering the active qubits."""
+        return len(self.dirty.difference(active_qubits or set()))
+
+    def borrow(self, num_qubits: int, active_qubits: Iterable[int] | None = None) -> list[int]:
```
alexanderivrii, line 44: The `borrow` method does not prioritize returning clean qubits over dirty ones? Could this be a problem for synthesis plugins that require clean ancillas?

qiskit/transpiler/passes/synthesis/qubit_tracker.py
```diff
@@ -0,0 +1,123 @@
...
+
+"""A qubit state tracker for synthesizing operations with auxiliary qubits."""
+
+from __future__ import annotations
+from collections.abc import Iterable
+from dataclasses import dataclass
+
+
+@dataclass
+class QubitTracker:
```
alexanderivrii, line 21: I really like this new class that very clearly exposes the functionality of how we track clean and dirty ancilla qubits. A couple of questions:
* Do we really need `self.qubits`, i.e. wouldn't it always be the same as the union of `self.clean` and `self.dirty`? Do we need this for the `borrow` method to be deterministic? Or for something else as well?
* We should probably specify the numbers of required clean and dirty qubits for the `borrow` method, i.e. have the method be something like `borrow(self, num_clean_qubits, num_dirty_qubits, active_qubits)`?
* When would a method `drop` be used? 
* It might be nice to add some example on using various methods in this class.

qiskit/transpiler/passes/synthesis/high_level_synthesis.py
```diff
@@ -345,19 +350,41 @@ class HighLevelSynthesis(TransformationPass):
...
+        as dirty qubits.
+
     """
 
+    @deprecate_arg(
+        "use_qubit_indices",
+        since="1.3",
+        additional_msg="The qubit indices will then always be used.",
+        pending=True,
+    )
```
alexanderivrii: I am still trying to figure out whether we need `use_qubit_indices` or not. The motivation was to provide the context whether HLS runs before or after layout/routing. In the default transpiler flow, it runs before, i.e. on the logical circuit. Note that HLS has access to `target`/`coupling_map` but (as no layout/routing has been done) it does not know the mapping between logical and physical qubits. In addition, most of the available synthesis plugins do not take the coupling map into account, e.g., all clifford synthesis methods synthesizes cliffords implicitly assuming the all-to-all connectivity. However, in some cases we might want to run HLS after layout/routing and we have a few internal and external plugins that ensure adherence to the coupling map (e.g., the token swapper plugin, or the external sat-based synthesis plugins for cliffods). 

Hmm... I think the only problematic usecase would be if we wanted to run say token swapper on the logical circuit, in which case it would wrongly assume that it has to adhere to the target connectivity and introduce more logic (that would later be routed potentially introducing even more swaps).

qiskit/transpiler/passes/synthesis/high_level_synthesis.py
```diff
@@ -418,151 +449,227 @@ def run(self, dag: DAGCircuit) -> DAGCircuit:
...
+            else:
+                # this returns the synthesized operation and the qubits it acts on -- note that this
+                # may be different than the original qubits, since we may use auxiliary qubits
+                synthesized, used_qubits = self._synthesize_operation(node.op, qubits, tracker)
+
+            # if the synthesis changed the operation (i.e. it is not None), store the result
+            # and mark the operation qubits as used
+            if synthesized is not None:
+                synthesized_nodes[node] = (synthesized, used_qubits)
+                tracker.used(qubits)  # assumes that auxiliary are returned in the same state
```
alexanderivrii, line 492: Should this be `tracker.used(used_qubits)`? Update: no, this is probably correct, `used_qubits` is the union of the original nodes's qubits + additional ancilla qubits employed by synthesis. If some of these ancillas were in `|0>`, the synthesis had to return them back to `|0>`, and they are still clean.

qiskit/transpiler/passes/synthesis/high_level_synthesis.py
```diff
@@ -418,151 +449,227 @@ def run(self, dag: DAGCircuit) -> DAGCircuit:
...
-                    self._equiv_lib is not None
-                    and equivalence.Key(name=node.name, num_qubits=node.num_qubits)
-                    in self._equiv_lib._key_to_node_index
-                )
+    def _synthesize_operation(
+        self,
+        operation: Operation,
+        qubits: tuple[int],
+        tracker: QubitTracker,
+    ) -> QuantumCircuit | Operation | DAGCircuit | None:
```
alexanderivrii: This now returns a tuple, right? We should change the output typehint.

alexanderivrii (commented):
I am half-way looking at the refactoring and I really like it! I have left a couple of interim questions and suggestions.

qiskit/transpiler/passes/synthesis/high_level_synthesis.py
```diff
@@ -418,151 +449,215 @@ def run(self, dag: DAGCircuit) -> DAGCircuit:
...
+                tracker.num_dirty(qubits),
+            )
 
-    def _recursively_handle_op(
-        self, op: Operation, qubits: Optional[List] = None
-    ) -> Tuple[Union[QuantumCircuit, DAGCircuit, Operation], bool]:
-        """Recursively synthesizes a single operation.
+        # try unrolling custom definitions
+        elif not self._top_level_only:
+            synthesized = self._unroll_custom_definition(operation, qubits)
```
alexanderivrii: I am not sure about the `else` here. Suppose that we have an object for which we both have synthesis plugins and `definition`. Previously we would try to run the synthesis plugins first, and, in the case the synthesis plugins are not successful to synthesize the object, we would resort to the `definition`; with this change the `definition` is not used.

qiskit/transpiler/passes/synthesis/high_level_synthesis.py
alexanderivrii: I am also wondering if there is any text from the older docstring of `_recursively_handle_op` that we might want to keep/adapt, in particular explaining in which circumstances the output can be a `QuantumCircuit`, a `DAGCircuit`, an `Operation` or `None`.

qiskit/transpiler/passes/synthesis/high_level_synthesis.py
```diff
@@ -418,151 +449,215 @@ def run(self, dag: DAGCircuit) -> DAGCircuit:
...
+        #  (2) High-level objects: try running the battery of high-level synthesis plugins
+        #  (3) Unrolling custom definitions: try defining the operation if it is not yet
+        #       in the set of supported instructions
+        # If any of the above were triggered, we will recurse and go again through these steps
+        # until no further change occurred. If there was no change, we just return ``None``.
+        synthesized = None
+
+        # Try synthesizing via AnnotatedOperation. This is faster than an isinstance check
+        # but a bit less safe since someone could create operations with a ``modifiers`` attribute.
+        if len(modifiers := getattr(operation, "modifiers", [])) > 0:
```
alexanderivrii, line 565: I really like how this passes clean/dirty ancilla qubits when synthesizing annotated operations! I also like how the code becomes a bit simpler, though I am wondering if we are not missing any edgecases. Previously, when we have an annotated operation we would first synthesize the base operation (which itself could be an annotated operation, but this is not the complication I am thinking about), and then we would handle modifiers one-by-one (which we also do now).

To describe what I am worried about, let's say that synthesizing the base operation produces the quantum circuit `qc`, and we need to apply first the `power` and then the `control` modifiers to this `qc`. Previously we would call `qc = qc.power(...)` to raise the quantum circuit to the given power, **apply the pass recursively to unroll obtained quantum circuit**, call `qc.control(...)` to add the control, and again apply the pass recursively to unroll the new circuit. Right now, it seems all intermediate recursive calls (like the one in boldface), and only apply a single recursive call at the very end. If this works as intended, the new way is definitely better, yet the `add_control` logic is quite tricky and assumes that the circuit only consists of gates of specific types; can we run into problems by removing the intermediate calls?

qiskit/transpiler/passes/synthesis/high_level_synthesis.py
```diff
@@ -418,151 +449,215 @@ def run(self, dag: DAGCircuit) -> DAGCircuit:
...
+            # The base operation must be synthesized without using potential control qubits
+            # used in the modifiers.
+            num_ctrl = sum(
+                mod.num_ctrl_qubits for mod in modifiers if isinstance(mod, ControlModifier)
             )
-        )
+            baseop_qubits = qubits[num_ctrl:]  # reminder: control qubits are the first ones
+            baseop_tracker = tracker.copy()
+            baseop_tracker.drop(qubits[:num_ctrl])  # no access to control qubits
```
alexanderivrii: This is a very minor point. Would it make sense to combine these two lines into one, having the `QubitTracker` initializer receive `qubits_to_drop` (empty set by default)?

alexanderivrii (commented):
Thanks @Cryoris for taking over and greatly improving the PR. I have left a few additional questions related to potential edgecases which I am not sure might or might not happen.

qiskit/compiler/transpiler.py
henryzou50: Oh yes, I like that suggestion. It is clear and avoids any confusion with the term "zero"  :)

henryzou50 (commented):
Thank you for changes in the first round of reviews. I've looked over the updates, and everything appears to be in order. Great work!

I noticed that Sasha has also left some insightful comments. Once those are addressed, I believe we'll be in a good position to merge the changes

qiskit/transpiler/passes/synthesis/high_level_synthesis.py
```diff
@@ -418,151 +449,215 @@ def run(self, dag: DAGCircuit) -> DAGCircuit:
...
+                    op = circuit_to_dag(op, copy_operations=False)
+
+                if isinstance(op, DAGCircuit):
+                    qubit_map = {
+                        qubit: index_to_qubit[index] for index, qubit in zip(qubits, op.qubits)
+                    }
+                    for sub_node in op.op_nodes():
+                        out.apply_operation_back(
+                            sub_node.op, tuple(qubit_map[qarg] for qarg in sub_node.qargs)
+                        )
```
alexanderivrii: Should we also worry about `cargs` here?

qiskit/transpiler/passes/synthesis/high_level_synthesis.py
alexanderivrii: Update: I have run multiple experiments with this code and it seems to function correctly! So most probably there is nothing to do here.

qiskit/transpiler/passes/synthesis/qubit_tracker.py
Cryoris: That's a very good point, that would be a problem, let's change that! In future we might need to update the tracker such that we can request a certain number of qubits of certain type.

qiskit/transpiler/passes/synthesis/qubit_tracker.py
Cryoris: * Yes, we need it to be deterministic, but this is a choice we have. We could also change `borrow` to sort the qubits to make the output deterministic.
* We could add that, but currently it would remain unused since HLS plugins have no way of telling how many ancillas of which _kind_ they are using -- only how many. I'm slightly leaning towards adding this feature once we can use it, but I'm open to changing my mind 🙂 
* `drop` is used for annotated controlled operations. There, we do not want to include the control qubits when synthesizing the base op, hence we do `tracker.drop(control_qubits)` 🙂 
* Sure, I can add some 👍🏻

qiskit/transpiler/passes/synthesis/high_level_synthesis.py
Cryoris: Summarizing the offline discussion: let's keep it and if `use_qubit_indices=False`, then we assume we cannot query the target whether a gate is supported. Should we rename this to `physical_qubits` as well to make it clearer that the circuit has been mapped to physical qubits already?

qiskit/transpiler/passes/synthesis/high_level_synthesis.py
Cryoris: It should be correct as is, because of the explanation you gave 🙂

qiskit/transpiler/passes/synthesis/high_level_synthesis.py
Cryoris: I see, I had assumed that the plugin will always be successful if there is one 🙂 I can change it accordingly. We should also add a test, it seems this scenario doesn't come up yet 😄

qiskit/transpiler/passes/synthesis/high_level_synthesis.py
Cryoris: Good point, I think this might work as is because `control` will unroll the circuit, right?

qiskit/transpiler/passes/synthesis/high_level_synthesis.py
Cryoris: Yeah we can do that 👍🏻

qiskit/transpiler/passes/synthesis/high_level_synthesis.py
Cryoris: Yes we probably should, e.g. when unrolling custom definitions that have clbits -- but I'm wondering if that's already working unexpectedly, otherwise shouldn't there be a test failing? 🤔

Cryoris: Edit: it didn't work, there was just no test for this 🙂

qiskit/compiler/transpiler.py
alexanderivrii: FWIW, personally I like the current name `qubits_initially_zero` a bit more than `qubits_reset` or `qubits_zeroed` since it's more clear what it means.

alexanderivrii (commented):
Thanks for addressing all of my review comments and adding tests for different cornercases (plugin mechanism returning `None`, having `clbits`).

The only remaining minor nitpick is that `_synthesize_operation` should return a `tuple` (consisting of the synthesized object and used qubits), can you please update the signature of this function accordingly?

alexanderivrii (approved):
Thanks Julien! I am now fully happy with this PR. @henryzou50, please take another look and see if it can be merged.

henryzou50 (approved):
LGTM! Great work and thanks for the changes :)

PR #7625, opened 2022-02-05
Add support for multi-qubit-gates to CSPLayout

Silrem:
### Summary
Fixes #7155. Added support for multi-qubit-gates to CSPLayout. CSPLayout now assumes for every 3+-qubit-gate we need a full connection between all involved qubits.

### Details and comments
I've implemented the version suggested in the issue, assuming all qubits involved in the gate, need a full connection. As this is my first issue and I'm not too familiar with the library yet, I'm wondering about a few things:

* Should I add a test case for this fix?
* Do we need to take into account direction? Should we require a bidirectional full connection?
* This approach does improve CSVLayout, but is the assumption actually fair? For Toffoli it holds, but can we say every 3+-qubit-gate does require a full connection? 

- [x] Possibly add testcase
- [x] Add release notes 
- [ ] Check how/if we need to consider direction
- [ ] Check assumption

PR #12912, opened 2024-08-06
Fix CSPLayout to ensure qubits in multi qubit gates are connected

### LINKED ISSUE #7155: CSP ignores 3-or-more-qubit gates
### Information

- **Qiskit Terra version**: main

### What is the current behavior?

Take the following circuit:

```python
qc = QuantumCircuit(3)
qc.cx(0,1)
qc.ccx(2,0,1)
qc.draw()
```
```
q_0: ──■────■──
     ┌─┴─┐┌─┴─┐
q_1: ┤ X ├┤ X ├
     └───┘└─┬─┘
q_2: ───────■──
```

`CSPLayout` should not be able to find a solution with a layout `0- > 1 -> 2`. However:
```python
from qiskit.transpiler.passes import CSPLayout

pass_ = CSPLayout(CouplingMap([(0, 1), (1, 2)]))
result = pass_(qc)
pass_.property_set["CSPLayout_stop_reason"]
```
```
'solution found'
```

### Suggested solutions

There are two options. If `CSPLayout` encounters a 3-or-more-qubit gate, either assumes that this needs complete connection among the involved qubits, or just fails with reason `3-or-more-qubit gate found`.

ChilliPenguin:
### Summary
Adds more qubit connection requirements for CSPLayout to ensure 3+ qubit gates work correctly.
This fixes #7155 and is essentially just adding onto PR #7625 with slight modification and unit tests added.

### Details and comments
This doesn't address the issue of directionality (specifically how should a connection between two control qubits be enforced if directionality matters)


1ucian0:
Hi @ChilliPenguin, thanks for taking care of it. Can you @Silrem as a co-author? as you are based on their PR. I think enforcing the 3q-gate bidirectional when you dont know how the box is implemented makes sense. If it is a controlled-2q or a 2c-1q gate, maybe you can relax it?

1ucian0:
This PR is stalling and, I think, https://github.com/Qiskit/qiskit/pull/15921 goes in the right direction.

PR #12913, opened 2024-08-06
Add rust friendly `assign parameters` methods

raynelfss:
### Summary
Follow up on #12794.
These commits expose some rust-friendly `assign_parameters` methods in `CircuitData` allowing us to perform parameter assignments without having call to python directly.

### Details and comments
The following commits add the methods `assign_paramters_from_slice` and `assign_parameters_from_mapping` which enable the assignment of parameters in rust in a friendly way instead of using hard-Python structures (`Py<_>` or `Bound<'_, _>`. Here are their use cases:
- `assign_paramters_from_slice`: Use whenever a collection of `Param` needs to be assigned. The amount of parameters sent needs to be of the same size as the the number of parameters in the `CircuitData` instance.
- `assign_parameters_from_mapping`: Use whenever of mapping between `ParameterUuid` instances from the `CircuitData` and `Param` instances coming from `Gates` or `Operations`. This can be used for partial assignment, and the number of parameters can be smaller or equal to the amount of parameters in the `CircuitData` instance.

### Known issues
- ~~The mapping of `Params` : `Params` may not be optimal as we wouldn't know if the types of parameters we're binding are valid, `Param::Float` does not posses a `uuid` and therefore is not a valid key for the mapping.~~ Will do a mapping between `ParameterUuid` and `Param` to assume that all keys are valid instances with valid `uuid`s.
- ~~The mapping of `assign_parameters_from_mapping` does not accept instances of `Map<Param, Param>` due to param not implementing the hash property in Rust.~~ Now it accepts any mapping of `ParameterUuid` : `Param` as long as it can be turned into an iterator of type `(ParameterUuid, Param)`.

### Blockers
- [x] #12926

### Other comments
- Feel free to point out any issues in the comments and I will quickly address.


crates/circuit/src/circuit_data.rs
```diff
@@ -949,45 +949,42 @@ impl CircuitData {
...
             let mut old_table = std::mem::take(&mut self.param_table);
+            let owned_iter: Vec<Param> = array.iter().map(|value| Param::Float(*value)).collect();
             self.assign_parameters_inner(
                 sequence.py(),
-                array
+                owned_iter
                     .iter()
                     .zip(old_table.drain_ordered())
-                    .map(|(value, (param_ob, uses))| (param_ob, Param::Float(*value), uses)),
+                    .map(|(value, (obj, uses))| (obj, value, uses)),
```
jakelishman: It's not clear to me that an entire reallocation of the entire array here is suitable in order to accept `&'a Param` in the inner methods.  It suggests to me that there's ownership problems going on, or that the new interface isn't entirely correct - we didn't need to allocate before, but we are now, so there's a regression.

I suspect that there's a way to make this work without allocation even in the current signature of the inner method, but if not, two points:
1. `Param` is very cheap to clone if the GIL is held; _mostly_ it's just going to be copying floats, which are `Copy`, and the rest of it is going to be ref-count updates in Python space.
2. We might be able to change the signature of the inner assigners to accept the iterator item as `AsRef<Param>`, and define `AsRef<Param> for Param` (there's no reflexive blanket impl built into the language only as a technical limitation).  That would let the inner method handle both `Param` and `&Param` iterators monomorphically.

crates/circuit/src/circuit_data.rs
```diff
@@ -949,45 +949,42 @@ impl CircuitData {
...
         let py = mapping.py();
         let mut items = Vec::new();
+        let mut objs = Vec::new();
         for item in mapping.call_method0("items")?.iter()? {
             let (param_ob, value) = item?.extract::<(Py<PyAny>, AssignParam)>()?;
-            let uuid = ParameterUuid::from_parameter(param_ob.bind(py))?;
-            items.push((param_ob, value.0, self.param_table.pop(uuid)?));
+            items.push(value);
+            objs.push(param_ob); // We need to separate the objects to avoid cloning.
         }
```
jakelishman: Can you explain why separating into multiple collections avoids cloning?  It feels like it shouldn't matter, or at a minimum all we'd be cloning is Python-space pointers anyway, which are very cheap to clone.

crates/circuit/src/circuit_data.rs
```diff
@@ -1135,6 +1132,44 @@ impl CircuitData {
...
+    {
+        let mut items = Vec::new();
+        for (param_uuid, value) in iter {
+            // Assume all the Parameters are already in the circuit
+            let param_obj = self.get_parameter_by_uuid(param_uuid);
+            items.push((
+                param_obj.unwrap().clone_ref(py),
+                value,
+                self.param_table.pop(param_uuid)?,
+            ));
```
jakelishman: Since we're already returning a `PyResult` from this function, I don't think there's any pain from having it correctly handle the error condition of a bad `ParameterUuid` here.

crates/circuit/src/circuit_data.rs
```diff
@@ -1135,6 +1132,44 @@ impl CircuitData {
...
+            slice
+                .iter()
+                .zip(old_table.drain_ordered())
+                .map(|(value, (param_ob, uses))| (param_ob, value, uses)),
+        )
+    }
+
+    /// Assigns parameters to circuit data based on a mapping of `ParameterUuid` : `Param`.
+    /// This mapping assumes that the provided `ParameterUuid` keys are instances
+    /// of `ParameterExpression`.
```
jakelishman, line 1146: It's conventional in Rust documentation to put the assumptions that cause panics in a section called `# Panics` to really make them stand out - idiomatic Rust is _far_ stricter about errors than Python, where the conventions are more like "just raise an exception".  Panics are serious business.

(But also see the other comment - imo it'd be cleaner just to handle the error properly, since it won't cost us anything.)

jakelishman (commented):
Thanks for this Ray.  My biggest concern here is in the first review comment - the new interface gives the Python-space forms additional allocations that they didn't need before.  I think we ought to either rework the Python forms (which I suspect might be possible) or at least rework the inner interface a little such that those allocations aren't necessary; they logically aren't, since they weren't there before.

crates/circuit/src/circuit_data.rs
raynelfss: It was an incorrect way of passing `Param` as a reference. However, I will remove this for a better structure.

crates/circuit/src/circuit_data.rs
raynelfss: **1.**  Would it be possible for me to clone `Param` safely by doing something like this:
```rs
...
match param {
    Param::Float(float) => Param::Float(*float),
    Param::ParameterExpression(exp) => Param::ParameterExpression(exp.clone_ref(py)),
    Param::Object(obj) => Param::Object(obj.clone_ref(py)),
}
...
```
If so, maybe this should be a method added to `Param`, if it is not as expensive to simply call `.clone()` considering that this will most likely copy floats usually.

**2.** So if we're able to implement the `AsRef<Param>` trait, this would allow us to receive both referenced and owned `Param instances as long as we specify it? Hmm... I'll give it a try then.

crates/circuit/src/circuit_data.rs
raynelfss: Sounds good!

crates/circuit/src/circuit_data.rs
jakelishman: 1. yes it's fine to add that, and for PyO3 style, we'd probably call that method `clone_ref`
2. well, the method would be generic and the Rust compiler actually ends up compiling two versions of it (one for the reference form, one for the owned form), but in terms of _writing_ the code, it appears like we just wrote one version

crates/circuit/src/operations.rs
```diff
@@ -125,6 +125,21 @@ impl Param {
...
+            Param::Float(float) => Param::Float(*float),
+            Param::Obj(obj) => Param::Obj(obj.clone_ref(py)),
+        }
+    }
+}
+
+impl AsRef<Param> for Param {
+    fn as_ref(&self) -> &Param {
+        self
+    }
```
jakelishman, line 145: Might be worth just putting in a comment here about why this exists potentially with a link to the Rust docs bit about why the blanket impl doesn't: https://doc.rust-lang.org/std/convert/trait.AsRef.html#reflexivity

jakelishman (commented):
Thanks Ray, this looks much cleaner to me now, and it's good to see no new allocations added in the existing paths.  One tiny comment about a comment is all, but code-wise this is good to merge.

jakelishman (approved):
Ace, thanks!

PR #12915, opened 2024-08-06
Deprecate more utils and parameters that depend on `qiskit.providers.models`

1ucian0:
As part of #12906 I found more places where object from the deprecated `qiskit.providers.models` are used, so they can be deprecated too.


ElePT:
Hi @Luciano!! This PR contains some useful deprecations, but also overlaps with some of the work in #https://github.com/Qiskit/qiskit/pull/13719 (mainly the `backend_properties` parts). I believe it shouldn't be too much work to get it up to speed and remove the overlapping deprecations, but let me know if you need help!

ElePT (approved):
LGTM

PR #12917, opened 2024-08-07
Introduce a gate name interner to avoid duplicating `String` in the Target

raynelfss:
### Summary
The current implementation of `Target` in Rust, performs a lot of cloning when adding an instruction to the Target. These commits allow it to keep one copy of these `String` instances and have copy-able (`usize`) values represent them in the other structures of the target

### Details and comments
The following commits add an interner which enable us to avoid multiple cloning of attributes with repeated use in multiple mapping structures within the target; In this case, with the gate names.

These changes are made along the following other changes:
- Remove `variable_operation_class`, use enum nature to discriminate.
- Hide `gate_name_map` from Python, use `operation_from_name`.
- Other small tweaks and fixes.

### Known issues:
Nothing so far, make sure to bring up during review.


crates/accelerate/src/target_transpiler/mod.rs
```diff
@@ -170,13 +233,13 @@ pub(crate) struct Target {
...
     pub concurrent_measurements: Option<Vec<Vec<PhysicalQubit>>>,
     gate_map: GateMap,
-    #[pyo3(get)]
-    _gate_name_map: IndexMap<String, TargetOperation, RandomState>,
-    global_operations: IndexMap<u32, HashSet<String>, RandomState>,
-    variable_class_operations: IndexSet<String, RandomState>,
-    qarg_gate_map: NullableIndexMap<Qargs, Option<HashSet<String>>>,
-    non_global_strict_basis: Option<Vec<String>>,
-    non_global_basis: Option<Vec<String>>,
+    gate_name_map: IndexMap<usize, TargetOperation, RandomState>,
```
mtreinish: One small question here, is there a reason we the full width of a `usize`? It is natural because that's the index type for `IndexSet`, but I'm pretty confident we don't the full 64 bit (or 32 bit on those platform) address space. Realistically I feel like we could get away with a u8 because it seems unlikely for a target to support more than >255 gates at once. If we wanted to be a bit conservative we could use a u16, there is almost no possibility of a target having over 65k gates names it.

mtreinish (commented):
I'm glad to see this, I meant to open an issue from: https://github.com/Qiskit/qiskit/pull/12292#discussion_r1676473673 but it's good you beat me to that. I haven't had a chance to review the code in depth yet, but I had a high level question inline off the top.

crates/accelerate/src/target_transpiler/mod.rs
jakelishman: I can imagine a world with >255 operation names, if we had something like multiple fractional 2q gates (heterogeneous arch) exposed both continuously and with various precisely calibrated points on the parametrisation.

mtreinish: 255 is more than it sounds like. I would be surprised if anyone spent the time to calibrate all those different variants even on a far future system. But I also agree with you that it is within the realm of plausibility, however `u16::MAX` is definitely sufficiently large though so I think using a u16 newtype here for the target string index would make a lot of sense.

raynelfss: This makes sense. Although the default indexing unit for `IndexSet` is mainly `usize`, I see how it could help to use a smaller datatype to store these indices in.

jakelishman:
Ray: I don't know your exact plans for the type here, but I think you should be able to create an `Interner<str>` now from the existing `Interner`, and it'll fulfil (almost) everything you need.

jakelishman:
Then all the magic `u16`s will become `Interned<str>` - that's a `u32` internally, but it's a) possible to make `Interner` and `Interned` more generic so they use `u16` instead and b) probably not going to matter for the `Target` anyway (personally I'd probably just use `u32`).

raynelfss:
Hey @jakelishman yes I'm currently playing with it. However, there are certain traits that I will need to implement beforehand. Will get back to you on this.

jakelishman:
The concept of the existing `Interner` is still based around index-set lookups, just like your new one is, so I'd be surprised if there's _too_ much work needed to make the existing one do everything you need.

raynelfss:
The efforts for this PR will be superseded by the results of #13308, in which we will hopefully address this.

raynelfss:
The efforts for this PR will be superseded by the results of #13308, in which we will hopefully address this.

PR #12918, opened 2024-08-07
Allow immutable borrow to access `QuantumCircuit.parameters`

jakelishman:
### Summary

`QuantumCircuit.parameters` is logically a read-only operation on `QuantumCircuit`.  For efficiency in multiple calls to `assign_parameters`, we actually cache the sort order of the internal `ParameterTable` on access.  This is purely a caching effect, and should not leak out to users.

The previous implementation took a Rust-space mutable borrow out in order to (potentially) mutate the cache.  This caused problems if multiple Python threads attempted to call `assign_parameters` simultaneously; it was possible for one thread to give up the GIL during its initial call to `CircuitData::copy` (so an immutable borrow was still live), allowing another thread to continue on to the getter `CircuitData::get_parameters`, which required a mutable borrow, which failed due to the paused thread in `copy`.

This moves the cache into a `RefCell`, allowing the parameter getters to take an immutable borrow as the receiver.  We now write the cache out only if we *can* take the mutable borrow out necessary.  This can mean that other threads will have to repeat the work of re-sorting the parameters, because their borrows were blocking the saving of the cache, but this will not cause failures.

The methods on `ParameterTable` that invalidate the cache all require a mutable borrow on the table itself.  This makes it impossible for an immutable borrow to exist simultaneously on the cache, so these methods should always succeed to acquire the cache lock to invalidate it.

### Details and comments

This won't be the last time we have problems with Python threading letting people see Rust reject conditions that allow data races. This particular case isn't the user's fault, though - `QuantumCircuit.parameters` _shouldn't_ need a mutable borrow, and should be thread-safe to be combined with other thread-safe methods on `QuantumCircuit`.  The issue was an internal caching detail, and that's something Qiskit should help.

An example script that produced errors before:
```python
from concurrent.futures import ThreadPoolExecutor

qc = QuantumCircuit(1, 1)
for _ in [None]*100_000:
    qc.measure(0, 0)

workers = 2

with ThreadPoolExecutor(max_workers=workers) as executor:
    fs = [executor.submit(qc.assign_parameters, []) for _ in [None]*workers]
    for future in fs:
        future.result()
```
The thread workers in this example aren't doing anything wrong; `assign_parameters(inplace=False)` should be safe to run in multiple threads.  Before the Rust-space work, Python would silently ignore the potential data race, and the GIL would have prevented anything catastrophic happening (though multiple threads might have separately calculated the sort order).

In that example, the `QuantumCircuit.copy` call within `QuantumCircuit.assign_parameters` takes an immutable borrow in Rust space, but the Rust space method can become interrupted part way through, while copying the `Measure` instructions (since this requires yielding control to the Python interpreter to run `Measure.copy`).  This allows another thread to progress to `QuantumCircuit.parameters`, which needed a mutable borrow, which could not be taken out due to the other paused thread.

### DIFF
--- a/crates/circuit/src/parameter_table.rs
+++ b/crates/circuit/src/parameter_table.rs
@@ -216,7 +216,7 @@ impl ParameterTable {
                     None
                 };
                 self.by_name.insert(name.clone(), uuid);
-                self.order_cache.borrow_mut().invalidate();
+                self.order_cache.get_mut().invalidate();
                 let mut uses = HashSet::new();
                 if let Some(usage) = usage {
                     uses.insert_unique_unchecked(usage);
@@ -335,7 +335,7 @@ impl ParameterTable {
                     vec_entry.remove_entry();
                 }
             }
-            self.order_cache.borrow_mut().invalidate();
+            self.order_cache.get_mut().invalidate();
             entry.remove_entry();
         }
         Ok(())
@@ -361,7 +361,7 @@ impl ParameterTable {
                     (vector_info.refcount > 0).then_some(vector_info)
                 });
         }
-        self.order_cache.borrow_mut().invalidate();
+        self.order_cache.get_mut().invalidate();
         Ok(info.uses)
     }
 
@@ -371,7 +371,7 @@ impl ParameterTable {
     pub fn drain_ordered(
         &mut self,
     ) -> impl ExactSizeIterator<Item = (Py<PyAny>, HashSet<ParameterUse>)> {
-        let mut cache = self.order_cache.borrow_mut();
+        let cache = self.order_cache.get_mut();
         cache.py_parameters = None;
         let order = if cache.uuids.is_empty() {
             self.sorted_order()
@@ -393,7 +393,7 @@ impl ParameterTable {
         self.by_uuid.clear();
         self.by_name.clear();
         self.vectors.clear();
-        self.order_cache.borrow_mut().invalidate();
+        self.order_cache.get_mut().invalidate();
     }
 
     /// Expose the tracked data for a given parameter as directly as possible to Python space.
--- a/crates/circuit/src/parameter_table.rs
+++ b/crates/circuit/src/parameter_table.rs
@@ -10,7 +10,7 @@
 // copyright notice, and modified files need to carry a notice indicating
 // that they have been altered from the originals.
 
-use std::cell::RefCell;
+use std::cell::OnceCell;
 
 use hashbrown::hash_map::Entry;
 use hashbrown::{HashMap, HashSet};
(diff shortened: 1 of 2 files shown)


crates/circuit/src/parameter_table.rs
```diff
@@ -332,26 +327,28 @@ impl ParameterTable {
...
 
     /// Clear this table, yielding the Python parameter objects and their uses in sorted order.
+    ///
+    /// The clearing effect is eager and not dependent on the iteration.
     pub fn drain_ordered(
-        &'_ mut self,
-    ) -> impl Iterator<Item = (Py<PyAny>, HashSet<ParameterUse>)> + '_ {
-        self.ensure_sorted();
+        &mut self,
+    ) -> impl ExactSizeIterator<Item = (Py<PyAny>, HashSet<ParameterUse>)> {
```
kevinhartman, line 339: Is there any reason to use an abstract return type here, or can we just return `ParameterTableDrain`?

In the case of `Vec::drain`, a public `Drain` struct gets returned explicitly. That seems like a useful pattern for the general case since we can then implement additional traits for the iterator if needed.

kevinhartman (commented):
This looks good to me.

I'm much happier now with the `OnceCell` being used instead of `RefCell`, which had felt like it required a little bit too much manual coordination when managing the cache.

As you mentioned in the PR description, this won't be the last time we encounter borrowing issues between threads. In this case, our caching was the reason we needed a mutable borrow. But, in cases where we actually have these semantics (where it _is_ the user's fault, in that they're doing something that requires multiple threads to borrow the same thing concurrently), are we planning to replicate the behavior of the old Python code (doing extra work but providing a correct result), or to produce a runtime failure?

crates/circuit/src/parameter_table.rs
jakelishman: Mostly I'd done it because I didn't want `ParameterTableDrain` to exist at all, so I kept the type private. The original version of this function had the iterator type unnameable, because I could borrow everything out of the struct without needing a manual implementation.  The Rust stdlib exports all its iterator type objects possibly in part because anonymous `impl Trait` in return position only turned up in 1.26, so there was no alternative before that.

I can make it a public type if you think it's better that way, but I'm not sure I agree with the justification "can implementation additional traits [...] if needed" in this case - if we need them, _then_ we can name the type and make it public, but at the moment, we've no need.

jakelishman:
> I'm much happier now with the OnceCell being used instead of RefCell, which had felt like it required a little bit too much manual coordination when managing the cache.

Oddly, I'm unhappy with this PR because I think this actually requires _more_ manual co-ordination when managing the cache, because now you have to take a lot of care (that the compiler won't protect you from) not to yield control to the Python interpreter during the initialiser, which makes `get_or_init` unsafe for us in most cases.  Hopefully a swap to `GILOnceCell` or something similar is possible for us in the future, though, which would remove my concern.

In the meantime, I'm happy though because it uses less memory, the point-of-use semantics are clearer, and it requires less runtime checking haha.

crates/circuit/src/parameter_table.rs
kevinhartman: I suppose it's fine as is, given that we can swap in an explicit struct if needed.

(Though I'd be curious if Rust would have used abstract return types for this kind of thing in the standard lib if they'd be around from the beginning.)

kevinhartman:
> you have to take a lot of care (that the compiler won't protect you from) not to yield control to the Python interpreter during the initialiser, which makes get_or_init unsafe for us in most cases.

Can we add a code comment that warns devs of this?

With `RefCell`, the coordination I was mostly referring to was what I'd remembered from the original version of this PR, where there were uses of `try_borrow_mut` and uses of `borrow(_mut)`, which I see now wasn't the case in the final `RefCell`-based approach.

jakelishman:
Oh, I got confused with the two PRs sorry - in this particular PR, the creation of the cache object doesn't require yielding to the Python interpreter.  The operations in #12942 do, though, and they include big explanatory code comments about the problem.

crates/circuit/src/parameter_table.rs
jakelishman: Yeah, I don't know if they would have either.  The anonymous types don't make for the prettiest reading of documentation, tbf, though equally, all the structs named `Iter` just mean you have to go to a mostly empty separate docs page to find which of the `Iterator` extension traits it implements.

kevinhartman (approved):
LGTM, thanks @jakelishman!

PR #12919, opened 2024-08-08
Fix warnings with latest clippy

jakelishman:
### Summary

These aren't failing CI, which uses `clippy` at MSRV, they're just triggered by more recent versions.


mtreinish (approved):

mtreinish:
I removed the stable backport tag here because the target rust code does not exist in `stable/1.2` it was merged on main for 1.3.

jakelishman:
Oh yeah thanks, I forgot about that.

PR #12925, opened 2024-08-08
Reorder Pauli terms before Trotterization

altaris:
### Summary

Reordering the terms of a Pauli operator can reduce the depth of the evolution circuit if using the Lie-Trotter or Suzuki-Trotter methods. This PR adds the option to do just that.

### Details and comments

This PR adds 
* the method `synthesis.evolution.product_formula.reorder_paulis` that reorders Pauli terms based on a greedy graph coloring heuristic;
* the `reorder` option in the `ProductFormula`, `LieTrotter` and `SuzukiTrotter` constructors to allow for reordering before synthesis.


qiskit/synthesis/evolution/product_formula.py
```diff
@@ -177,6 +186,64 @@ def evolve_pauli(
...
+    indexed_nodes = list(enumerate(graph.nodes()))
+    for (idx1, term1), (idx2, term2) in combinations(indexed_nodes, 2):
+        # Add an edge between term1 and term2 if they touch the same qubit
+        for a, b in zip(term1[0], term2[0]):
+            if not (a == "I" or b == "I"):
+                graph.add_edge(idx1, idx2, None)
+                break
+    # TODO: The graph is likely not very big, so an exact coloring could be
+    # computed in a reasonable time.
+    coloring = rx.graph_greedy_color(graph)
```
alexanderivrii: A quick comment: the Rustworkx's `graph_greedy_color` function can take an optional argument `strategy` that specifies the heuristic coloring method used, please see https://www.rustworkx.org/dev/apiref/rustworkx.ColoringStrategy.html for detail. You might want to experiment with the different strategies available to see which performs best for this application (or run all three one after another and choose the result with the smallest number of colors).

qiskit/synthesis/evolution/product_formula.py
```diff
@@ -177,6 +186,64 @@ def evolve_pauli(
...
+    graph.add_nodes_from(sum((o.to_list() for o in operators), []))
+    indexed_nodes = list(enumerate(graph.nodes()))
+    for (idx1, term1), (idx2, term2) in combinations(indexed_nodes, 2):
+        # Add an edge between term1 and term2 if they touch the same qubit
+        for a, b in zip(term1[0], term2[0]):
+            if not (a == "I" or b == "I"):
+                graph.add_edge(idx1, idx2, None)
+                break
+    # TODO: The graph is likely not very big, so an exact coloring could be
+    # computed in a reasonable time.
```
alexanderivrii: Out of curiosity, how many nodes are typically there and how much is it important to get the optimum number of colors? (For instance, there are standard ways to translate a graph coloring problem into SAT which could then be solved with say Z3).

qiskit/synthesis/evolution/lie_trotter.py
```diff
@@ -78,6 +78,7 @@ def __init__(
             | None
         ) = None,
         wrap: bool = False,
+        reorder: bool = False,
```
Cryoris: We can discuss this, but I think this should be `True` per default as it can significantly optimize the circuit depth 🙂 It will change the output of the current circuits, but as this is a synthesis routine which only promises to output a circuit implementing the Trotter formula, changes like this are fine I think.

Cryoris:
@altaris are you still pursuing this or do would you like us to help? 🙂

altaris:
I thought so too, but I didn't want to change the behavior of anything else
when I wrote this. I can flip it to default to `True`.

2024年9月2日(月) 17:24 Julien Gacon ***@***.***>:

altaris:
Sure! I'd love to see it through 💪

qiskit/synthesis/evolution/lie_trotter.py
sbrandhsn: I disagree. We should only consider performing Pauli term reordering by default for a given Hamiltonian, if we
- understand the potential classical runtime implications
- (strongly) suspect we can gain reductions in circuit depth by inspecting the Hamiltonian
- care about circuit depth at all as specified by some transpilation option
- leveraged Rust for getting the most efficient implementation
The underlying combinatorial problem is NP-complete and while heuristics yield a great runtime improvement they can still be quite slow compared to the rest of the transpilation steps, especially on larger Hamiltonians. Otherweise, I love the PR, thanks @altaris !

I'm mentioning @ajavadia in case he wants to chime in. :-)

Cryoris: Runtime cost is a good point 👍🏻 it would be interesting to see the that for large scale Hamiltonians, as in most cases they have an underlying structure. Maybe we could check some from the Hamlib? 🙂

qiskit/synthesis/evolution/product_formula.py
```diff
@@ -177,6 +186,79 @@ def evolve_pauli(
...
+) -> SparsePauliOp | list[SparsePauliOp]:
+    r"""
+    Creates an equivalent operator by reordering terms in order to yield a
+    shallower circuit after evolution synthesis. The original operator remains
+    unchanged.
+
+    This method works in three steps. First, a graph is constructed, where the
+    nodes are the terms of the operator and where two nodes are connected if
+    their term acts on the same qubit (for example, the terms $IXX$ and $IYI$
+    would be connected, but not $IXX$ and $YII$). Then, the graph is colored.
```
Cryoris: This is not API documented, so this is just as info: to render math in our docs we have to use 
```
:math:`XII`
```
instead of `$XII$` 🙂

qiskit/synthesis/evolution/product_formula.py
```diff
@@ -177,6 +186,79 @@ def evolve_pauli(
...
+        len(operators) > 1
+        # operators has a single SparsePauliOp, which has > 1 terms
+        or (len(operators) == 1 and len(operators[0]) > 1)
+    ):
+        return operators
+    if not isinstance(operators, list):
+        operators = [operators]
+
+    terms = sum((o.to_list() for o in operators), [])
+    terms = sorted(terms, key=_term_sort_key)
```
Cryoris: Why do we need to sort the operators? To enable a deterministic outcome?

qiskit/synthesis/evolution/product_formula.py
```diff
@@ -177,6 +186,79 @@ def evolve_pauli(
...
+    # Do nothing in trivial cases
+    if isinstance(operators, SparsePauliOp) and len(operators) <= 1:
+        return operators
+    if isinstance(operators, list) and not (
+        # operators is a list of > 1 SparsePauliOp
+        len(operators) > 1
+        # operators has a single SparsePauliOp, which has > 1 terms
+        or (len(operators) == 1 and len(operators[0]) > 1)
+    ):
+        return operators
```
Cryoris: Pauli operators that are given in a list all commute, e.g. we would give
```
[SparsePauliOp(["ZZI", "IZZ", "ZII"]), SparsePauliOp(["YII", "YYY"])]
```
This information should be preserved if provided by the users 🙂 So I think we should simply do a re-ordering per element in the list, instead of merging all operators and discarding this initial information. If users would like to merge all operators, this could simply be done by adding the operators before passing them into the evolution.

qiskit/synthesis/evolution/product_formula.py
```diff
@@ -177,6 +186,79 @@ def evolve_pauli(
...
+
+    terms = sum((o.to_list() for o in operators), [])
+    terms = sorted(terms, key=_term_sort_key)
+    graph = rx.PyGraph()
+    graph.add_nodes_from(terms)
+    indexed_nodes = list(enumerate(graph.nodes()))
+    for (idx1, term1), (idx2, term2) in combinations(indexed_nodes, 2):
+        # Add an edge between term1 and term2 if they touch the same qubit
+        for a, b in zip(term1[0], term2[0]):
+            if not (a == "I" or b == "I"):
```
Cryoris:
```suggestion
            if a != "I" and b != "I":
```
should be slightly faster since the expression can terminate already if the first part of the `and` statement is `False` 🙂

qiskit/synthesis/evolution/product_formula.py
Cryoris: I agree, we should try this or expose an argument allowing to set the coloring strategy.

Cryoris (commented):
Overall the PR looks very nice! I left some comments below, in particular about the handling of lists of operators. Could you also add a release note describing the new argument? 🙂

qiskit/synthesis/evolution/product_formula.py
```diff
@@ -177,6 +186,79 @@ def evolve_pauli(
...
+                graph.add_edge(idx1, idx2, None)
+                break
```
mrossinek: I would think there is a better way to do this based on the `PauliList` boolean array saved inside the `SparsePauliOp` instances. A simple way is the following:
```python
x_or_z = op.paulis.x | op.paulis.z
overlaps = x_or_z @ x_or_z.T
```
This yields a matrix indicating on entry `(i, j)` whether the Pauli term at index `i` has overlap with that at index `j`.

This computes the full matrix but you are really only interested in the upper triangular indices. So there may be an even more efficient way to do this...

qiskit/synthesis/evolution/product_formula.py
mrossinek: I agree that preserving this information is needed. I suppose you could simply call the algorithm below on each entry of this list separately and then return a new list of operators.

qiskit/synthesis/evolution/product_formula.py
mrossinek: Actually, this might also help with the algorithm becoming deterministic without requiring the terms to be sorted first :thinking:

mrossinek (commented):
Just adding my 2 cents here

alexanderivrii:
A quick question. The overall consensus is that "reorder" should be "False" by default? Or that it should be "True" if reordering is sufficiently fast?

qiskit/synthesis/evolution/lie_trotter.py
ShellyGarion: Is there a benchmark such as Benchpress HamLib https://github.com/Qiskit/benchpress/tree/main that indicates the CX cost/depth improvement and the time consumption ?

qiskit/synthesis/evolution/product_formula.py
Cryoris: This is now obsolete as we deal with sparse pauli labels 🙂

qiskit/synthesis/evolution/lie_trotter.py
Cryoris: Conclusion: we're not enabling this per default for now.

ShellyGarion (commented):
LGTM. thanks for the nice contribution to Qiskit. Some release notes are needed.

qiskit/synthesis/evolution/qdrift.py
```diff
@@ -125,4 +132,8 @@ def expand(self, evolution: PauliEvolutionGate) -> list[tuple[str, tuple[int], f
         sampled_paulis = [
             (pauli[0], pauli[1], np.real(np.sign(pauli[2])) * rescaled_time) for pauli in sampled
         ]
+
+        if self.preserve_order:
```
Cryoris: Only re-order if the order is _not_ preserved 🙂 
```suggestion
        if not self.preserve_order:
```

qiskit/circuit/library/pauli_evolution.py
```diff
@@ -143,7 +149,14 @@ def time(self, time: ParameterValueType) -> None:
 
     def _define(self):
         """Unroll, where the default synthesis is matrix based."""
-        self.definition = self.synthesis.synthesize(self)
+        if self.synthesis is None:
+            from qiskit.synthesis.evolution import LieTrotter
+
+            synthesis = LieTrotter()
+        else:
+            synthesis = self.synthesis
```
alexanderivrii: Do we need this, i.e. shouldn't `self.synthesis` always return a non-`None` synthesis method?

qiskit/synthesis/evolution/lie_trotter.py
```diff
@@ -79,8 +80,19 @@ def __init__(
                 built.
             wrap: Whether to wrap the atomic evolutions into custom gate objects. This only takes
                 effect when ``atomic_evolution is None``.
+            preserve_order: Whether to allow preserve_ordering the terms of the operator to
```
alexanderivrii: "preserve_order" allows "preserve_ordering" 😄 you might want to rephrase this slightly

qiskit/synthesis/evolution/product_formula.py
```diff
@@ -84,11 +90,15 @@ def __init__(
             wrap: Whether to wrap the atomic evolutions into custom gate objects. Note that setting
                 this to ``True`` is slower than ``False``. This only takes effect when
                 ``atomic_evolution is None``.
+            preserve_order: Whether to allow preserve_ordering the terms of the operator to
+                potentially yield a shallower evolution circuit. Not relevant
```
alexanderivrii: and here too

qiskit/synthesis/evolution/product_formula.py
```diff
@@ -239,3 +250,74 @@ def real_or_fail(value, tol=100):
         return np.real(value)
 
     raise ValueError(f"Encountered complex value {value}, but expected real.")
+
+
+def reorder_paulis(
+    paulis: SparsePauliLabel, strategy: rx.ColoringStrategy = rx.ColoringStrategy.Saturation
```
alexanderivrii: Out of curiosity, why did you choose this specific coloring strategy? 

One other thing you can do is to first call `rx.graph_bipartite_edge_color`, and if the graph happens to be bipartite (which is the case for many structured graphs of interest, like heavy-hexes), then we will automatically get a coloring with the minimum number of colors. And if the graph is not bipartite, we could use the greedy coloring function.

UPDATE: sorry, my comment above makes no sense whatsoever, in this algorithm you need a _node coloring_ algorithm, while the function I mentioned is for _edge coloring_.

qiskit/synthesis/evolution/product_formula.py
```diff
@@ -239,3 +250,74 @@ def real_or_fail(value, tol=100):
...
+) -> SparsePauliOp | list[SparsePauliOp]:
+    r"""
+    Creates an equivalent operator by preserve_ordering terms in order to yield a
```
alexanderivrii: At this point I am starting to doubt myself, and think that `preserve_ordering` is a valid English word.

qiskit/synthesis/evolution/product_formula.py
```diff
@@ -239,3 +250,74 @@ def real_or_fail(value, tol=100):
...
+    shallower circuit after evolution synthesis. The original operator remains
+    unchanged.
+
+    This method works in three steps. First, a graph is constructed, where the
+    nodes are the terms of the operator and where two nodes are connected if
+    their term acts on the same qubit (for example, the terms :math:`IXX` and
```
alexanderivrii: "their term" -> "their terms"

qiskit/synthesis/evolution/product_formula.py
```diff
@@ -239,3 +250,74 @@ def real_or_fail(value, tol=100):
...
+
+    # rx.graph_greedy_color is supposed to be deterministic
+    coloring = rx.graph_greedy_color(graph, strategy=strategy)
+    terms_by_color = defaultdict(list)
+
+    for term_idx, color in sorted(coloring.items()):
+        term = graph.nodes()[term_idx]
+        terms_by_color[color].append(term)
+
+    terms = sum(terms_by_color.values(), [])
```
alexanderivrii: This is probably obvious, but how is `sum` used here exactly? And why at the end `terms` is of the form  `SparsePauliOp | list[SparsePauliOp]`?

qiskit/synthesis/evolution/suzuki_trotter.py
```diff
@@ -104,6 +105,9 @@ def __init__(
                 built.
             wrap: Whether to wrap the atomic evolutions into custom gate objects. This only takes
                 effect when ``atomic_evolution is None``.
+            preserve_order: Whether to allow reordering the terms of the operator to
+                potentially yield a shallower evolution circuit. Not relevant
+                when synthesizing operator with a single term. Defaults to ``True``.
```
alexanderivrii: Now I like the docstring 😄 

Actually, the way this is worded, it feels the other way around, i.e. `preserve_order` restricts the order to the original order.

alexanderivrii (commented):
Thanks @Cryoris  and the original contributor @altaris! I think allowing Pauli reordering is a useful option for the `PauliEvolutionGate` synthesis algorithms.

I have left a few minor comments and questions inline.

One more question: do you think that for Rustiq-based synthesis algorithm (and the plugin already accepts "preserve_order"), it might be beneficial to rearrange Paulis before passing the pauli network to Rustiq?

qiskit/synthesis/evolution/lie_trotter.py
Cryoris: Hehe, this happened due to `%s/reorder/preserve_order` 😅

Cryoris:
> One more question: do you think that for Rustiq-based synthesis algorithm (and the plugin already accepts "preserve_order"), it might be beneficial to rearrange Paulis before passing the pauli network to Rustiq?

That is already possible to do with this PR, since the re-ordering happens before `expand`. So we can expose this in the Rustiq plugin, just like in the default plugin. So we would re-order with the graph coloring and then allow Rustiq to also re-order? Or are they redundant together?

qiskit/synthesis/evolution/product_formula.py
Cryoris: It misuses `sum` to concatenate lists -- it's better to use `itertools.chain` though, I'll update

alexanderivrii:
> That is already possible to do with this PR, since the re-ordering happens before expand. So we can expose this in the Rustiq plugin, just like in the default plugin. So we would re-order with the graph coloring and then allow Rustiq to also re-order? Or are they redundant together?

I think it makes sense to expose this to Rustiq plugin as well. And I think it makes sense to let Rustiq do the additional reordering.

qiskit/synthesis/evolution/suzuki_trotter.py
Cryoris: Fixed in 56e619c

qiskit/synthesis/evolution/product_formula.py
Cryoris: Fixed in 56e619c

qiskit/synthesis/evolution/product_formula.py
Cryoris: Fixed in 56e619c

qiskit/synthesis/evolution/product_formula.py
Cryoris: without underscore I think it is, but it's wrong here 😂

qiskit/synthesis/evolution/product_formula.py
Cryoris: Fixed in 56e619c

qiskit/synthesis/evolution/lie_trotter.py
Cryoris: Fixed in 56e619c

qiskit/circuit/library/pauli_evolution.py
Cryoris: Correct, fixed in 56e619c

Cryoris:
56e619c includes this option in the Rustiq plugin too 👍🏻

qiskit/synthesis/evolution/lie_trotter.py
```diff
@@ -60,6 +60,7 @@ def __init__(
             | None
         ) = None,
         wrap: bool = False,
+        preserve_order: bool = False,
```
mtreinish: I thought the consensus here was to not reorder terms by default because of the potential performance overhead. Shouldn't the default be to `True`?

qiskit/synthesis/evolution/product_formula.py
```diff
@@ -63,6 +69,7 @@ def __init__(
             | None
         ) = None,
         wrap: bool = False,
+        preserve_order: bool = False,
```
mtreinish: Same question here

qiskit/synthesis/evolution/qdrift.py
```diff
@@ -68,6 +68,7 @@ def __init__(
         ) = None,
         seed: int | None = None,
         wrap: bool = False,
+        preserve_order: bool = True,
```
mtreinish, line 71: This one defaults to `True` that's just inconsistent, but I think the expected value.

qiskit/transpiler/passes/synthesis/hls_plugins.py
```diff
@@ -1528,6 +1535,9 @@ def run(self, high_level_object, coupling_map=None, target=None, qubits=None, **
             )
             return None
 
+        if "preserve_order" in options and isinstance(algo, ProductFormula):
+            algo.preserve_order = options["preserve_order"]
+
```
alexanderivrii: Thanks for adding this. A very small nitpick: we don't need another call to `isinstance(also, ProductFormula)` here, since we've just checked this on the line above.

mtreinish:
I think given the open questions and the proximity to 1.3.0 (ie this is the last PR being tracked) it's probably better to defer this to 2.0.0 rather than force it in. This feature is a nice to have, but I don't think it's absolutely necessary for 1.3.0

raynelfss:
Pushing to 2.0.0 after further discussion...

qiskit/synthesis/evolution/product_formula.py
mtreinish:
```suggestion
        preserve_order: bool = True,
```

alexanderivrii (approved):
Thanks @Cryoris. This PR very nicely concludes the work that was done to improve the support for Pauli evolutions: representing them as gates, improving expansion methods, adding the plugin interface, new synthesis algorithms and new options, porting relevant code to Rust.

qiskit/synthesis/evolution/lie_trotter.py
```diff
@@ -79,8 +80,19 @@ def __init__(
                 built.
             wrap: Whether to wrap the atomic evolutions into custom gate objects. This only takes
                 effect when ``atomic_evolution is None``.
+            preserve_order: If ``False``, allows reordering the terms of the operator to
+                potentially yield a shallower evolution circuit. Not relevant
+                when synthesizing operator with a single term.
```
mtreinish, line 85: I think from a docs perspective it would be good to explain the tradeoffs here, especially around runtime performance.

qiskit/synthesis/evolution/product_formula.py
```diff
@@ -239,3 +251,61 @@ def real_or_fail(value, tol=100):
...
+        return (term[1], term[0])
+
+    # Do nothing in trivial cases
+    if len(paulis) <= 1:
+        return paulis
+
+    terms = sorted(paulis, key=_term_sort_key)
+    graph = rx.PyGraph()
+    graph.add_nodes_from(terms)
+    indexed_nodes = list(enumerate(graph.nodes()))
```
mtreinish, line 296: I think you don't actually need to cast this is as a list, it should all work fine without it and save a copy.

mtreinish (commented):
I'm fine with merging this as is. I just have 2 small nits inline, but they can be fixed in a followup without issue and aren't worth blocking this over.

PR #12926, opened 2024-08-08
Add `get_parameter_by_uuid` to `CircuitData`

raynelfss:
### Summary
The following commits add a method `get_parameter_by_uuid` to obtain the `Parameter` object from `CircuitData` by `uuid`.

### Details and comments
The method retrieves a `Parameter` object from the `ParameterTable` based on its `uuid` without exposing any extra data. This could be useful when performing mapping based parameter assignment (as one of the methods from #12913 attempts to do), from rust-space, without having to necessarily pass the a `Parameter` object  to do so.

### Known issues:
Nothing so far but, if anything is found, leave a comment or review. Thank you!

### DIFF
--- a/crates/circuit/src/circuit_data.rs
+++ b/crates/circuit/src/circuit_data.rs
@@ -361,12 +361,6 @@ impl CircuitData {
             .map(|ob| ob.clone_ref(py))
     }
 
-    pub fn get_parameter_by_uuid(&self, py: Python, uuid: ParameterUuid) -> Option<Py<PyAny>> {
-        self.param_table
-            .py_parameter_by_uuid(uuid)
-            .map(|ob| ob.clone_ref(py))
-    }
-
     /// Return the width of the circuit. This is the number of qubits plus the
     /// number of clbits.
     ///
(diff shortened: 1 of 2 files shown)


crates/circuit/src/circuit_data.rs
```diff
@@ -361,6 +361,12 @@ impl CircuitData {
             .map(|ob| ob.clone_ref(py))
     }
 
+    pub fn get_parameter_by_uuid(&self, py: Python, uuid: ParameterUuid) -> Option<Py<PyAny>> {
+        self.param_table
+            .py_parameter_by_uuid(uuid)
+            .map(|ob| ob.clone_ref(py))
+    }
+
```
jakelishman: Does this need to be a `pymethod`?

Do you think there might any reason to call this without needing the GIL?  If so, we _could_ return the `Option<&Py<PyAny>>` directly and let the caller handle cloning the reference if they need it. I can't think of a huge amount of use for that, though.

jakelishman (commented):
Thanks Ray, this looks very sensible to me.

crates/circuit/src/circuit_data.rs
raynelfss: I don't see a need for it to be a `pymethod` mainly other than keeping it consistent with the counterpart method that allows for you to look for a parameter by name. Now, I could turn it into a rust-native method and then have a `pymethod` call directly to it, unless we won't use this at all.

jakelishman (approved):
Cool, thanks!

PR #12927, opened 2024-08-08
Prepare 1.1.2

kevinhartman:
### Summary
Prepares the release of minor version 1.1.2.

### Details and comments
There's a fix in here for version parsing of `matplotlib` (their current version is `matplotlib 3.9.1.post1` which we fail to parse) which we also need on `main` (parsing logic is borrowed from our qpy module). I will open a separate PR on that for main, and we can choose to either backport it to this branch or just keep this PR as is.


releasenotes/notes/fix-bitarray-slice-bits-shots-c9cb7e5d907722f5.yaml
```diff
@@ -3,9 +3,9 @@ upgrade_primitives:
...
     :meth:`.BitArray.slice_bits` and :meth:`.BitArray.slice_shots`
     will now raise ``IndexError`` when indices are out of bounds.
-    They used to raise ``ValueError`` in the case.
+    They used to raise ``ValueError`` in this case.
   - |
     :meth:`.BitArray.__getitem__` will now raise ``IndexError``
     when indices are out of bounds or the number of dimensions
     of indices does not match that of BitArray.
-    They used to raise ``ValueError`` in the case.
+    They used to raise ``ValueError`` in this case.
```
mtreinish: I hadn't seen this release note before today, but this is an API breaking change. For a bugfix release we should revert this. I see in the original PR #12755 the argument was made this is a bugfix to be consistent with other things and normal index based access. I think given that we can potentially make the case for this in 1.2.0, but for a bugfix release we shouldn't do something like this. I think before we can tag this 1.1.2 release we should revert #12845

mtreinish (commented):
Besides the one release note I flagged in my previous comment I think this LGTM. Once the proposed revert in https://github.com/Qiskit/qiskit/pull/12929 merges I think we can merge this and go ahead and tag the release.

mtreinish (approved):
LGTM, thanks for doing this.

PR #12928, opened 2024-08-08
Fix version parsing issue in bloch.py.

kevinhartman:
### Summary
Adds more robust version parsing in `bloch.py`, which we use to check the version of `matplotlib`. The latest version of `matplotlib` is `matplotlib 3.9.1.post1` and causes the old code to fail.

### Details and comments
This version regex stuff is taken from our `qpy` module.

### DIFF
--- a/qiskit/visualization/bloch.py
+++ b/qiskit/visualization/bloch.py
@@ -50,6 +50,7 @@ __all__ = ["Bloch"]
 
 import math
 import os
+import re
 import numpy as np
 import matplotlib
 import matplotlib.pyplot as plt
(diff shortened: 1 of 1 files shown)


qiskit/visualization/bloch.py
```diff
@@ -60,6 +61,47 @@
...
+            (?P<dev_l>dev)
+            [-_\.]?
+            (?P<dev_n>[0-9]+)?
+        )?
+    )
+    (?:\+(?P<local>[a-z0-9]+(?:[-_\.][a-z0-9]+)*))?       # local version
+"""
+    + "$"
+)
+VERSION_PATTERN_REGEX = re.compile(VERSION_PATTERN, re.VERBOSE | re.IGNORECASE)
```
mtreinish, line 102: Can we just import this from qpy it seems a bit weird to duplicate this a 4th time in the code: https://github.com/search?q=repo%3AQiskit%2Fqiskit%20VERSION_PATTERN&type=code

but at the end of the day I guess it doesn't matter too much.

kevinhartman: Personally, I would rather just duplicate it unless we move it to a common location. I think importing from `qpy` would be weirder in terms of module dependencies. It really shouldn't change anyway since it's not our code.

mtreinish: I think we should look at consolidating these down to one shared copy somewhere, but we can save that for a separate PR. I'm fine to add another copy of this here in the medium term to get this fix into 1.2.0.

mtreinish (approved):

PR #12935, opened 2024-08-09
Use ahash for IndexMap and IndexSet hasher

mtreinish:
### Summary

The ahash hasher offers better performace as a drop in replacement for the stdlib hash algorithm. This is a large reason we use the hashbrown crate for our HashMap type in Qiskit (along with rayon support). The indexmap crate doesn't use ahash by default however, so we typically switch to it anywhere we use IndexMap or IndexSet to match the lookup performance of hashbrown in places where we need to preserve insertion order. However, there were a couple of spots where we missed this in the rust code, this commit corrects these oversights.

### Details and comments


jakelishman (approved):
The super intermediate use in `qasm3` feels like it's probably not too worth bothering with, but it's not bad to be uniform.

PR #12943, opened 2024-08-12
Allow `CircuitData` construction from `PackedOperation`s

Cryoris:
### Summary

As the title says 🙂 This allows to construct more general circuits than using just `CircuitData::from_standard_gates`, including e.g. barriers.

### Details and comments

This is essentially a copy of `CircuitData::from_standard_gates`. Qubits and Clbits are now `Vec`, because
* `SmallVec` cannot be used as we cannot predict the maximum length
* arrays (`&[Qubit]`) turned out to be a bit restrictive as the length must be known at compile-time (but maybe there's some trick to use this in the signature but pass vectors or something?)

I didn't yet add a test since it's essentially a copy of the standard gates constructor, and this will be used a bunch by the circuit library refactor. But we could add some `cfg` test that maybe checks this gives the same circuit data as using the standard gates construct -- though not sure how difficult that comparison is 🙂 

Example usage:
```rust
let n = 50;  // number of qubits
let h_layer = (0..n).map(|i| (
    StandardGate::HGate.into(),  // PackedOperation
    smallvec![],  //  params
    vec![Qubit(i)],  // qubit
    vec![], // clbit
);
let barrier = (  // barrier on all qubits
    PyInstruction { /* define barrier by calling Python ... */ },
    smallvec![],
    (0..n).map(|i| Qubit(i)).collect(),  // all qubits
    vec![], 
);

let data = CircuitData::from_packed_operations(h_layer.chain(barrier));
```

### DIFF
--- a/crates/circuit/src/circuit_data.rs
+++ b/crates/circuit/src/circuit_data.rs
@@ -105,6 +105,71 @@ pub struct CircuitData {
 
 impl CircuitData {
     /// An alternate constructor to build a new `CircuitData` from an iterator
+    /// of packed operations. This can be used to build a circuit from a sequence
+    /// of `PackedOperation` without needing to involve Python.
+    ///
+    /// This can be connected with the Python space
+    /// QuantumCircuit.from_circuit_data() constructor to build a full
+    /// QuantumCircuit from Rust.
+    ///
+    /// # Arguments
+    ///
+    /// * py: A GIL handle this is needed to instantiate Qubits in Python space
+    /// * num_qubits: The number of qubits in the circuit. These will be created
+    ///     in Python as loose bits without a register.
+    /// * num_clbits: The number of classical bits in the circuit. These will be created
+    ///     in Python as loose bits without a register.
+    /// * instructions: An iterator of the (packed operation, params, qubits, clbits) to
+    ///     add to the circuit
+    /// * global_phase: The global phase to use for the circuit
+    pub fn from_packed_operations<I>(
+        py: Python,
+        num_qubits: u32,
+        num_clbits: u32,
+        instructions: I,
+        global_phase: Param,
+    ) -> PyResult<Self>
+    where
+        I: IntoIterator<
+            Item = (
+                PackedOperation,
+                SmallVec<[Param; 3]>,
+                Vec<Qubit>,
+                Vec<Clbit>,
+            ),
+        >,
+    {
+        let instruction_iter = instructions.into_iter();
+        let mut res = Self::with_capacity(
+            py,
+            num_qubits,
+            num_clbits,
+            instruction_iter.size_hint().0,
+            global_phase,
+        )?;
+        for (operation, params, qargs, cargs) in instruction_iter {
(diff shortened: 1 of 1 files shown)


crates/circuit/src/circuit_data.rs
```diff
@@ -104,6 +104,71 @@ pub struct CircuitData {
...
+    /// # Arguments
+    ///
+    /// * py: A GIL handle this is needed to instantiate Qubits in Python space
+    /// * num_qubits: The number of qubits in the circuit. These will be created
+    ///     in Python as loose bits without a register.
+    /// * num_clbits: The number of classical bits in the circuit. These will be created
+    ///     in Python as loose bits without a register.
+    /// * instructions: An iterator of the (packed operation, params, qubits, clbits) to
+    ///     add to the circuit
+    /// * global_phase: The global phase to use for the circuit
```
jakelishman, line 124: It's less common in Rust land to be this explicit about every individual argument - things like `global_phase` are pretty self explanatory, and if the comment is just explaining the type, it might not be offering too much.

We don't have any team policy on this yet, so no need to change anything, I'm just comparing to the Rust ecosystem at large.

crates/circuit/src/circuit_data.rs
```diff
@@ -104,6 +104,71 @@ pub struct CircuitData {
...
+            instruction_iter.size_hint().0,
+            global_phase,
+        )?;
+        for (operation, params, qargs, cargs) in instruction_iter {
+            let qubits = (&mut res.qargs_interner)
+                .intern(InternerKey::Value(qargs.to_vec()))?
+                .index;
+            let clbits = (&mut res.cargs_interner)
+                .intern(InternerKey::Value(cargs.to_vec()))?
+                .index;
```
jakelishman: Is the `to_vec` necessary here?  I think `qargs` and `cargs` should already be owned, and already be the correct types.

jakelishman (commented):
This generally seems sensible to me, and useful in the short term.

I think at some point in the indeterminate future we'll want to re-examine our internal Rust interfaces to work out exactly what we want and the best interface to expose to ourselves, but let's not get bogged down in bikeshedding right now when we're just getting started.

crates/circuit/src/circuit_data.rs
Cryoris: whoops, my oversight 👍🏻  Fixed in d632427

crates/circuit/src/circuit_data.rs
Cryoris: Yep, good point, I followed what existed for `from_standard_gates` here 🙂

jakelishman (approved):
Let's go for this for now.  We'll need to improve the testing story of things like this in the future, but given you're about to start using it in another PR, we'll get the testing from that form.

ElePT:
[@Mergifyio](https://github.com/Mergifyio) backport stable/1.2

ElePT:
@Mergifyio backport stable/1.2

PR #12950, opened 2024-08-13
Move the circuit library's entanglement logic to Rust

Cryoris:
### Summary

Move the entanglement logic used in the circuit library's `NLocal` circuits & friends to Rust. This PR is also the basis to move frequently used circuits, such as `EfficientSU2`, `TwoLocal` or `ZZFeatureMap` to Rust.

This also fixes a bug in the `circular` entanglement which missed some entanglement blocks for entangling gates with 3+ qubits.

### Details and comments

`NLocal` will directly use this function, but there's also a series of tests to check the different errors and possible input types. 

The attentive reader might notice that handling elaborate constructs such as `list[list[list[int]]]` are not handled, which does not affect the current code, but will change what we accept once the circuit library is moved to Rust. That's an intentional change as these types were used to allow different entanglements for different circuit layers (controlled by `offset`), but this was redundantly covered by passing a callable. Moving forward, the suggestion is to only allow a callable in these cases to have only 1 method of doing 1 thing 🙂

### DIFF
--- a/crates/accelerate/src/circuit_library/entanglement.rs
+++ b/crates/accelerate/src/circuit_library/entanglement.rs
@@ -1,0 +1,203 @@
+// This code is part of Qiskit.
+//
+// (C) Copyright IBM 2024
+//
+// This code is licensed under the Apache License, Version 2.0. You may
+// obtain a copy of this license in the LICENSE.txt file in the root directory
+// of this source tree or at http://www.apache.org/licenses/LICENSE-2.0.
+//
+// Any modifications or derivative works of this code must retain this
+// copyright notice, and modified files need to carry a notice indicating
+// that they have been altered from the originals.
+
+use crate::QiskitError;
+use pyo3::{
+    types::{PyAnyMethods, PyInt, PyList, PyListMethods, PyString, PyTuple},
+    Bound, PyAny, PyResult,
+};
+use qiskit_circuit::slice::PySequenceIndex;
+
+fn _combinations(n: u32, repetitions: u32) -> Vec<Vec<u32>> {
+    if repetitions == 1 {
+        (0..n).map(|index| vec![index]).collect()
+    } else {
+        let mut result = Vec::new();
+        for indices in _combinations(n, repetitions - 1) {
+            let last_element = indices[indices.len() - 1];
+            for index in last_element + 1..n {
+                let mut extended_indices = indices.clone();
+                extended_indices.push(index);
+                result.push(extended_indices);
+            }
+        }
+        result
+    }
+}
+
+pub fn full(num_qubits: u32, block_size: u32) -> impl Iterator<Item = Vec<u32>> {
+    // this should be equivalent to itertools.combinations(list(range(n)), m)
+    _combinations(num_qubits, block_size).into_iter()
+}
+
+/// Return the qubit indices for linear entanglement.
+/// For a block_size of ``m``, this is defined as [(0..m-1), (1..m), (2..m+1), ..., (n-m..n-1)]
+pub fn linear(num_qubits: u32, block_size: u32) -> impl DoubleEndedIterator<Item = Vec<u32>> {
+    (0..num_qubits - block_size + 1)
+        .map(move |start_index| (start_index..start_index + block_size).collect())
+}
+
+/// Return the qubit indices for a reversed linear entanglement.
+pub fn reverse_linear(num_qubits: u32, block_size: u32) -> impl Iterator<Item = Vec<u32>> {
+    linear(num_qubits, block_size).rev()
+}
+
+/// Return the qubit indices for circular entanglement. This is defined as tuples of length ``m``
+/// starting at each possible index ``(0..n)``. Historically, Qiskit starts with index ``n-m+1``.
+/// This is probably easiest understood for a concerete example of 4 qubits and block size 3:
+/// [(2,3,0), (3,0,1), (0,1,2), (1,2,3)]
+pub fn circular(num_qubits: u32, block_size: u32) -> Box<dyn Iterator<Item = Vec<u32>>> {
+    if block_size == 1 || num_qubits == block_size {
+        Box::new(linear(num_qubits, block_size))
+    } else {
+        let historic_offset = num_qubits - block_size + 1;
+        Box::new((0..num_qubits).map(move |start_index| {
+            (0..block_size)
+                .map(|i| (historic_offset + start_index + i) % num_qubits)
+                .collect()
+        }))
+    }
+}
+
+pub fn pairwise(num_qubits: u32) -> impl Iterator<Item = Vec<u32>> {
+    // for Python-folks (like me): pairwise is equal to linear[::2] + linear[1::2]
+    linear(num_qubits, 2)
+        .step_by(2)
+        .chain(linear(num_qubits, 2).skip(1).step_by(2))
+}
+
+pub fn shift_circular_alternating(
+    num_qubits: u32,
+    block_size: u32,
+    offset: usize,
+) -> Box<dyn Iterator<Item = Vec<u32>>> {
+    // index at which we split the circular iterator -- remember that circular entanglement is a
+    // list of length ``num_qubits``, which is what we use in the ``convert_idx`` function here
+    let split = PySequenceIndex::convert_idx(
+        -((offset % num_qubits as usize) as isize),
+        num_qubits as usize,
+    )
+    .expect("Something went wrong converting the offset to a negative index.");
+    let shifted = circular(num_qubits, block_size)
+        .skip(split)
+        .chain(circular(num_qubits, block_size).take(split));
+    if offset % 2 == 0 {
+        Box::new(shifted)
+    } else {
+        // if the offset is odd, reverse the indices inside the qubit block (e.g. turn CX
+        // gates upside down)
+        Box::new(shifted.map(|indices| indices.into_iter().rev().collect()))
+    }
+}
+
+/// Get an entangler map for an arbitrary number of qubits.
+///
+/// Args:
+///     num_qubits: The number of qubits of the circuit.
+///     block_size: The number of qubits of the entangling block.
+///     entanglement: The entanglement strategy as string.
+///     offset: The block offset, can be used if the entanglements differ per block,
+///         for example used in the "sca" mode.
+///
+/// Returns:
+///     The entangler map using mode ``entanglement`` to scatter a block of ``block_size``
+///     qubits on ``num_qubits`` qubits.
+pub fn get_entanglement_from_str(
+    num_qubits: u32,
+    block_size: u32,
+    entanglement: &str,
+    offset: usize,
+) -> PyResult<Box<dyn Iterator<Item = Vec<u32>>>> {
+    if block_size > num_qubits {
+        return Err(QiskitError::new_err(format!(
+            "block_size ({}) cannot be larger than number of qubits ({})",
+            block_size, num_qubits
+        )));
+    }
+
+    if entanglement == "pairwise" && block_size > 2 {
+        return Err(QiskitError::new_err(format!(
+            "block_size ({}) can be at most 2 for pairwise entanglement",
+            block_size
+        )));
+    }
+
+    // if block size is 2 and pairwise, this is just linear: [0, 1, 2, ...]
+    match (entanglement, block_size) {
+        ("full", _) => Ok(Box::new(full(num_qubits, block_size))),
+        ("linear", _) => Ok(Box::new(linear(num_qubits, block_size))),
+        ("reverse_linear", _) => Ok(Box::new(reverse_linear(num_qubits, block_size))),
+        ("sca", _) => Ok(shift_circular_alternating(num_qubits, block_size, offset)),
+        ("circular", _) => Ok(circular(num_qubits, block_size)),
+        ("pairwise", 1) => Ok(Box::new(linear(num_qubits, 1))),
+        ("pairwise", 2) => Ok(Box::new(pairwise(num_qubits))),
+        _ => Err(QiskitError::new_err(format!(
+            "Unsupported entanglement: {}",
+            entanglement
+        ))),
+    }
+}
+
+/// Get an entangler map for an arbitrary number of qubits.
+///
+/// Args:
+///     num_qubits: The number of qubits of the circuit.
+///     block_size: The number of qubits of the entangling block.
+///     entanglement: The entanglement strategy.
+///     offset: The block offset, can be used if the entanglements differ per block,
+///         for example used in the "sca" mode.
+///
+/// Returns:
+///     The entangler map using mode ``entanglement`` to scatter a block of ``block_size``
+///     qubits on ``num_qubits`` qubits.
+pub fn get_entanglement<'a>(
+    num_qubits: u32,
+    block_size: u32,
+    entanglement: &'a Bound<PyAny>,
+    offset: usize,
+) -> PyResult<Box<dyn Iterator<Item = PyResult<Vec<u32>>> + 'a>> {
+    // unwrap the callable, if it is one
+    let entanglement = if entanglement.is_callable() {
+        entanglement.call1((offset,))?
+    } else {
+        entanglement.to_owned()
+    };
+
+    if let Ok(strategy) = entanglement.downcast::<PyString>() {
+        let as_str = strategy.to_string();
+        return Ok(Box::new(
+            get_entanglement_from_str(num_qubits, block_size, as_str.as_str(), offset)?
+                .map(|connections| Ok(connections)),
+        ));
+    } else if let Ok(list) = entanglement.downcast::<PyList>() {
+        let entanglement_iter = list.iter().map(move |el| {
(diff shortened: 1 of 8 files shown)


crates/accelerate/src/circuit_library/entanglement.rs
```diff
@@ -0,0 +1,202 @@
...
+/// Offset 2 -> [(0,1,2), (1,2,3), (2,3,0), (3,0,1)]
+/// ...
+pub fn shift_circular_alternating(
+    num_qubits: u32,
+    block_size: u32,
+    offset: usize,
+) -> Box<dyn Iterator<Item = Vec<u32>>> {
+    // index at which we split the circular iterator -- remember that circular entanglement is a
+    // list of length ``num_qubits``, which is what we use in the ``convert_idx`` function here
+    let split = PySequenceIndex::convert_idx(
```
jlapeyre: I think it's better to maintain a cleaner separation between the implementation of Qiskit data and functions in Rust on the one hand, and on the Python wrappers on the other. Separating them in the future will be a very large undertaking even if we try to prepare for it.

I'd even go so far as to put the two concerns in separate files, so that the core implementation does not import pyo3 things.

I realize that `convert_idx` exists before this PR, but I still don't favor kicking the can on this.

crates/accelerate/src/circuit_library/entanglement.rs
```diff
@@ -0,0 +1,202 @@
...
+    (0..num_qubits - block_size + 1)
+        .map(move |start_index| (start_index..start_index + block_size).collect())
+}
+
+/// Get a reversed linear entanglement. This is like linear entanglement but in reversed order:
+/// [(n-m..n-1), ..., (1..m), (0..m-1)]
+/// This is particularly interesting, as CX+"full" equals CX+"reverse_linear", even though the
+/// latter uses a quadratically less number of gates.
+pub fn reverse_linear(num_qubits: u32, block_size: u32) -> impl Iterator<Item = Vec<u32>> {
+    linear(num_qubits, block_size).rev()
```
jlapeyre, line 40: We could avoid the bizarre `DoubleEndedIterator` with the following
```rust
    (0..num_qubits - block_size + 1)
        .rev()
        .map(move |start_index| (start_index..start_index + block_size).collect())
```
Until someone is familiar with the limitations of rust ranges, they might look around for code here that pulls items from both ends. (I did). On the other hand, my suggestion is not obviously better. It's more food for thought, that you might prefer to use.
Since more robust and convenient ranges are useful in technical computing, we may want to shop around for a replacement. But I favor putting that off till this first major RIIR is done.

Cryoris: Yeah I can update this -- I just thought the `DoubleEndedIterator` would be cleaner, but the code duplication is tiny so I'm happy to change it 🙂

crates/accelerate/src/circuit_library/entanglement.rs
Cryoris: The ``convert_idx`` function doesn't actually use any `pyo3` or Python logic, it's really just a util for Python-like access. I can also avoid using it but if we have the function already, then I'm not sure we should duplicate the logic 🙂

jlapeyre: > The `convert_idx function` doesn't actually use any `pyo3`

That's part of the point I'm trying to make. When I first started to the comment, I mistakenly believed that  `convert_idx` was introduced in this PR. This makes dealing with this a bit less straightforward. `convert_idx` uses no `pyo3`. The two calls to `convert_idx` are in functions that otherwise do not use pyo3. Part of what we are trying to do is redesign qiskit in a way that is independent of Python. Allowing pyo3 to intrude where it is not necessary makes this more difficult.

crates/accelerate/src/circuit_library/entanglement.rs
jlapeyre: `DoubleEndedIterator` might be cleaner. It does avoid code duplication. I was trying to come up with something obviously better, but didn't really get there. 

I don't have a strong preference, so I'd say whatever you prefer.

Cryoris: In that case I'd slightly prefer the `DoubleEndedIterator`, as it doesn't require the duplication 🙂

crates/accelerate/src/circuit_library/entanglement.rs
Cryoris: Ok, this seems part of a more general discussion -- to move forward I removed the usage of `convert_idx` from this PR 🙂

crates/accelerate/src/circuit_library/entanglement.rs
```diff
@@ -0,0 +1,197 @@
...
+/// [(0..m-1), (1..m), (2..m+1), ..., (n-m..n-1)]
+pub fn linear(num_qubits: u32, block_size: u32) -> impl DoubleEndedIterator<Item = Vec<u32>> {
+    (0..num_qubits - block_size + 1)
+        .map(move |start_index| (start_index..start_index + block_size).collect())
+}
+
+/// Get a reversed linear entanglement. This is like linear entanglement but in reversed order:
+/// [(n-m..n-1), ..., (1..m), (0..m-1)]
+/// This is particularly interesting, as CX+"full" equals CX+"reverse_linear", even though the
+/// latter uses a quadratically less number of gates.
```
ShellyGarion: maybe say explicitly that CX+"full" is n(n-1)/2 while CX+"reverese_linear" is only n-1 ?

crates/accelerate/src/circuit_library/entanglement.rs
```diff
@@ -0,0 +1,197 @@
...
+pub fn reverse_linear(num_qubits: u32, block_size: u32) -> impl Iterator<Item = Vec<u32>> {
+    linear(num_qubits, block_size).rev()
+}
+
+/// Return the qubit indices for circular entanglement. This is defined as tuples of length ``m``
+/// starting at each possible index ``(0..n)``. Historically, Qiskit starts with index ``n-m+1``.
```
ShellyGarion, line 44: is there a reason or reference for choosing this offset?

crates/accelerate/src/circuit_library/mod.rs
```diff
@@ -0,0 +1,69 @@
...
+
+/// Get the entanglement for given number of qubits and block size.
+///
+/// Args:
+///     num_qubits: The number of qubits to entangle.
+///     block_size: The entanglement block size (e.g. 2 for CX or 3 for CCX).
+///     entanglement: The entanglement strategy. This can be one of:
+///
+///         * string: Available options are ``"linear"``, ``"reverse_linear"``, ``"circular"``,
+///             ``"pairwise"`` or ``"sca"``.
```
ShellyGarion: also add "full"

releasenotes/notes/fix-circular-entanglement-5aadd5adf75c0c13.yaml
```diff
@@ -0,0 +1,16 @@
+fixes:
```
ShellyGarion, line 1: is there a performance improvement by porting this code into rust? if so - we can also indicate it in the release notes.

ShellyGarion (commented):
Thanks @Cryoris - I only have some minor comments.
BTW, does porting this code to rust improves its performance ?

crates/accelerate/src/circuit_library/entanglement.rs
alexanderivrii: I am also confused about using `DoubleEndedIterator` here: looking at the following examples
```
let numbers = vec![1, 2, 3, 4, 5, 6];

let mut iter = numbers.iter();

assert_eq!(Some(&1), iter.next());
assert_eq!(Some(&6), iter.next_back());
assert_eq!(Some(&5), iter.next_back());
assert_eq!(Some(&2), iter.next());
assert_eq!(Some(&3), iter.next());
assert_eq!(Some(&4), iter.next());
assert_eq!(None, iter.next());
assert_eq!(None, iter.next_back());
```
in
https://doc.rust-lang.org/std/iter/trait.DoubleEndedIterator.html,
it seems that the primary usage of this iterator is the ability to pull items from both ends. But we don't use this ability anywhere in the code, right? Oh, you mean that for `linear(num_qubits, block_size).rev()` to work, we need the ability to pull the items from the back?

crates/accelerate/src/circuit_library/entanglement.rs
```diff
@@ -0,0 +1,197 @@
...
+use pyo3::{
+    types::{PyAnyMethods, PyInt, PyList, PyListMethods, PyString, PyTuple},
+    Bound, PyAny, PyResult,
+};
+
+use crate::QiskitError;
+
+/// Get all-to-all entanglement. For 4 qubits and block size 2 we have:
+/// [(0, 1), (0, 2), (0, 3), (1, 2), (1, 3), (2, 3)]
+pub fn full(num_qubits: u32, block_size: u32) -> impl Iterator<Item = Vec<u32>> {
```
alexanderivrii: I am trying to understand when a function can return `impl Iterator<Item = Vec<u32>>` vs `Box<dyn Iterator<Item = Vec<u32>>>`. From what I see in the documentation, we can do the former when the concrete iterator type is known at compile time, and we should do the latter when we want to return multiple iterator types. But why does `full` fall into the first category, while `shift_circular_alternating` into the second one?

In any case, would it be better if all the functions returned the same type, be it `Box<dyn Iterator<Item = Vec<u32>>>` or maybe even `Box<dyn DoubleEndedIterator<Item = Vec<u32>>>`?

releasenotes/notes/fix-circular-entanglement-5aadd5adf75c0c13.yaml
```diff
@@ -0,0 +1,16 @@
...
+    code previously used::
+
+      [(2, 3, 0), (0, 1, 2), (1, 2, 3)]
+
+    but now is correctly adding the ``(3, 0, 1)`` connections, that is::
+
+      [(2, 3, 0), (3, 0, 1), (0, 1, 2), (1, 2, 3)]
+
+    As such, the ``"circular"`` and ``"sca"`` entanglements use ``num_qubits``
+    entangling blocks per layer.
```
alexanderivrii, line 16: Great catch!

test/python/circuit/library/test_nlocal.py
```diff
@@ -928,5 +932,127 @@ def test_full_vs_reverse_linear(self, num_qubits):
...
+        with self.assertRaises(QiskitError):
+            _ = fast_entangler_map(num_qubits=4, block_size=3, entanglement="pairwise", offset=0)
+
+    def test_invalid_blocksize(self):
+        """Test the block size being too large."""
+        with self.assertRaises(QiskitError):
+            _ = fast_entangler_map(num_qubits=2, block_size=3, entanglement="linear", offset=0)
+
+    def test_invalid_entanglement_str(self):
+        """Go drunk, you're home"""
```
alexanderivrii: Hmmm....

crates/accelerate/src/circuit_library/entanglement.rs
Cryoris: > Oh, you mean that for linear(num_qubits, block_size).rev() to work, we need the ability to pull the items from the back?

Exactly 🙂 here, `rev` can only be called on types that implement the `DoubleEndedIterator` trait

crates/accelerate/src/circuit_library/entanglement.rs
Cryoris: I don't remember the details, though it might have been connected to the shifted-circular-alternating below... I added this as comment to explicitly state this behavior, which was uncommented before 🙂

crates/accelerate/src/circuit_library/entanglement.rs
Cryoris: As discussed in a comment above, we could also avoid using `DoubleEndedIterator`, but we would still need the `Box<dyn Iterator>` since the functions return different iterator types (they use `Chain` `Map` `Slice` in different combinations, so would still need to be coerced to the same dynamic type). But yeah, we could make all functions return `Box<dyn Iterator<Item=Vec<u32>>` -- I kept the signatures as the minimal type we need, but I'm also happy to change 🙂

releasenotes/notes/fix-circular-entanglement-5aadd5adf75c0c13.yaml
Cryoris: There's no performance benefit it only constructing the entanglement, this is a prerequisite so we can fully construct the circuit library circuits in Rust, without needing to go in Python to construct the entanglement in between 🙂 There's an immediate follow-up on porting `PauliFeatureMap` to Rust (which is significantly faster), but putting all together would've been quite a big PR, so I split the features 🙂

test/python/circuit/library/test_nlocal.py
Cryoris: Made serious in 7d8c8a4 🙂

crates/accelerate/src/circuit_library/mod.rs
Cryoris: done in 7d8c8a4

crates/accelerate/src/circuit_library/entanglement.rs
Cryoris: done in 7d8c8a4

crates/accelerate/src/circuit_library/entanglement.rs
alexanderivrii: Thanks! I think the current way is fine. Oh, one more question: do the functions like `full`, `linear` need to be public (`pub`)?

crates/accelerate/src/circuit_library/entanglement.rs
```diff
@@ -0,0 +1,197 @@
...
+                    connections, block_size
+                )));
+            }
+            Ok(connections)
+        });
+        return Ok(Box::new(entanglement_iter));
+    }
+    Err(QiskitError::new_err(
+        "Entanglement must be a string or list of qubit indices.",
+    ))
```
alexanderivrii: To me this looks good, however I would appreciate if someone with more knowledge of Rust/Pyo3 would take a look as well (maybe @raynelfss)? In particular, is there a more ergonomic way to convert python lists of integers into a rust vector (avoiding some of the downcasting/extracting)?

crates/accelerate/src/circuit_library/entanglement.rs
```diff
@@ -0,0 +1,197 @@
...
+            block_size, num_qubits
+        )));
+    }
+
+    if entanglement == "pairwise" && block_size > 2 {
+        return Err(QiskitError::new_err(format!(
+            "block_size ({}) can be at most 2 for pairwise entanglement",
+            block_size
+        )));
+    }
```
alexanderivrii: Could we add this to the `match` code below (by including `("pairwise", _)` after handling the other "pairwise" cases)?

crates/accelerate/src/circuit_library/entanglement.rs
Cryoris: They currently are since the `get_entangler_map` with the Python logic is defined in `mod.rs` -- we can move that into the `entanglement.rs` file and then they can be private 🙂

crates/accelerate/src/circuit_library/entanglement.rs
Cryoris: If there's a way I'd be glad to know it -- the `NLocal` refactoring (on my local branch, not here) currently has quadruple nested lists which need casting to `PyList` 😅

crates/accelerate/src/circuit_library/entanglement.rs
Cryoris: Good idea 👍🏻

crates/accelerate/src/circuit_library/entanglement.rs
```diff
@@ -0,0 +1,197 @@
...
+
+    // if block size is 2 and pairwise, this is just linear: [0, 1, 2, ...]
+    match (entanglement, block_size) {
+        ("full", _) => Ok(Box::new(full(num_qubits, block_size))),
+        ("linear", _) => Ok(Box::new(linear(num_qubits, block_size))),
+        ("reverse_linear", _) => Ok(Box::new(reverse_linear(num_qubits, block_size))),
+        ("sca", _) => Ok(shift_circular_alternating(num_qubits, block_size, offset)),
+        ("circular", _) => Ok(circular(num_qubits, block_size)),
+        ("pairwise", 1) => Ok(Box::new(linear(num_qubits, 1))),
+        ("pairwise", 2) => Ok(Box::new(pairwise(num_qubits))),
```
alexanderivrii, line 134: Oh, I see, you have simplified the original Python logic, or rather moved it to the individual functions. I think the code became much more clear now. However, do we have enough tests to cover all the corner-case scenarios: such as `block_size` being 1, `block_size` being `num_qubits`, etc.?

Cryoris: I can add some more 🙂

alexanderivrii (approved):
LGTM, thanks for this development!

PR #12951, opened 2024-08-13
Use hashbrown::HashSet instead of ahash::HashSet

mtreinish:
### Summary

In the `target_transpiler/mod.rs` module we were using ahash::HashSet for a hash set implementation, but the rest of Qiskit has standardized on using hashbrown for the `HashSet` and `HashMap` types. Hashbrown uses ahash for it's hashing algorithm but it also provides other advantages. To ensure that hash sets are compatible across the library we should be using the same library for everything. To support this goal, this commit also adds a clippy rule that raises a warning if the std library hashmap or hashset is used, or the versions from ahash. This means with our current dependency set the only allowed hashset types are `hashbrown::HashMap`/`HashSet` and `indexmap::IndexMap`/`IndexSet` (for where we need to maintain insertion order). Ideally we'd have a rule that forces the use of ahash with `IndexMap` and `IndexSet` (see #12935) but I don't think clippy exposes an option to enable something like that.

### Details and comments

### DIFF
--- a/.clippy.toml
+++ b/.clippy.toml
@@ -1,0 +1,6 @@
+disallowed-types = [
+    "std::collections::HashSet",
+    "std::collections::HashMap",
+    "ahash::HashSet",
+    "ahash::HashMap",
+]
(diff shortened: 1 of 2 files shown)


raynelfss (approved):
This is sensible to me, there's still a lot of confusion as to which `Mapping` or `Collection` structures to use. This should allow us to be more consistent about allowing certain things.

PR #12952, opened 2024-08-14
Fix error message upon misalignment in `PadDynamicalDecoupling`

Cryoris:
### Summary

If a delay is misaligned and following a `DAGInNode` (or preceding a `DAGOutNode`), the error message in `PadDynamicalDecoupling` previously failed, as it was trying to access non-existing attributes. This is because, unlike `DAGOpNode`, the in and out nodes do not have a `name` or `qargs` attribute.

### DIFF
--- a/releasenotes/notes/fix-dd-misalignment-msg-76fe16e5eb4ae670.yaml
+++ b/releasenotes/notes/fix-dd-misalignment-msg-76fe16e5eb4ae670.yaml
@@ -1,0 +1,7 @@
+---
+fixes:
+  - |
+    Fixed a bug in :class:`.PadDynamicalDecoupling`, which previously
+    did not correctly display the error message that a delay is not
+    pulse-aligned, if the previous or following node was a input/output
+    node. Now, the error message is correctly displayed.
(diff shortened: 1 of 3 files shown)


releasenotes/notes/fix-dd-misalignment-msg-76fe16e5eb4ae670.yaml
```diff
@@ -0,0 +1,7 @@
+---
+fixes:
+  - |
+    Fixed a bug in :class:`.PadDynamicalDecoupling`, which previously
+    did not correctly display the error message that a delay is not
+    pulse-aligned, if the previous or following node was a input/output
```
jakelishman: "an input/output"

jakelishman (commented):
Just for my understanding: is the expectation that the (correct) error message should only show if there's been a bug in the prior `PadDelay` pass, or are we expecting users to feed in manually scheduled circuits?

Cryoris:
As I understand it, this error message should only come up if users put manual delays into the circuit and either didn't run `ConstrainedReschedule` or put barriers, such that the reschedule doesn't merge delays into multiples of the alignment.

jakelishman (approved):

PR #12957, opened 2024-08-14
Remove stale `variable_class_operations` set from `Target`

### LINKED ISSUE #12953: Transpile fails when multiple circuits with UnitaryGate or DiagonalGate
### Environment

- **Qiskit version1.3: 
- **Python version**: 
- **Operating system**: 

### What is happening?

Transpile fails when the list of input circuits contains `UnitaryGate` or `DiagonalGate` with error as following:

```
pyo3_runtime.PanicException: internal error: entered unreachable code: 'num_qubits' property is reserved for normal operations only.
```
And single circuit input with `UnitaryGate` or `DiagonalGate` 

see https://github.com/Qiskit/qiskit-aer/actions/runs/10381689813/job/28743530145

```
from qiskit.quantum_info.random import random_unitary

backend = AerSimulator(method="statevector", device="CPU")

qc1 = QuantumCircuit(2)
qc1.unitary(random_unitary(4, seed=123), [0, 1])
# using followings instead of unitary is OK
#qc1.h(0)
#qc1.cx(0, 1)
qc1.measure_all()

qc2 = QuantumCircuit(2)
qc2.unitary(random_unitary(4, seed=123), [0, 1])
# using followings instead of unitary is OK
#qc2.h(1)
#qc2.cx(1, 0)
qc2.measure_all()

circuits = [qc1, qc2]
# circuits = [qc1]   this is also OK
circuits = transpile(circuits, backend)

result = backend.run(circuits).result()
print(result)
```

### What should happen?

No error in transpile

### Any suggestions?

_No response_

raynelfss:
**NOTE** This is a refactored PR since the cause of the errors was found to be a different reason.

I will leave the old description attached, the following description is only valid for commits after 0caa7d5e2375310b7a468b20a51bd66c7dd610cc

### Summary
Fixes #12953 

The following commits remove a stale attribute of the rust target called `variable_class_operations` which was used to keep track of all the variadics within the target but was not kept during the serialization step.

### Details and comments:

When performing serialization, we were forgetting to include `variable_class_operations` set of names in the state mapping. Since the nature of `TargetOperation` is to work as an enum of either `Instruction` instances or class aliases that would represent `Variadic` instructions. The usage of that structure was redundant, so it was removed. 

The following changes were made:
- `num_qubits` returns an instance of `u32`, callers will need to make sure they're dealing with a `NormalOperation`.
- `params` behaves more similarly, returning a slice of `Param` instances. Will panic if called on a `Variadic` operation.

A test case was added to check for something similar to what was mentioned by @doichanj in #12953.

### Old description
```
The following commit ensures that when the `num_qubits` or `params` are extracted from a variadic, we always return `None`.

### Details and comments
- The `num_qubits` property for `TargetOperation` returns an `Option<u32>` instead of panicking in the case that the instruction is a `Variadic`.
- The `params` property for `TargetOperation` returns an `Option<[&Param]>]` instance instead of panicking in the case of a `Variadic`.
```


ElePT (commented):
Thanks @raynelfss! I think it would be good to also have a unit test for the fix, but looking at the issue it looks like you were not able to reproduce it locally, did you think of an alternative we could use?

raynelfss:
Thank you for checking this @ElePT, I couldn't replicate Jun's specific issue. But I do have an idea of how to replicate it. I will add unit tests soon. Thank you!

raynelfss:
Actually the test-case for this has proven a bit harder to compose. This problem appeared to happen on previous versions of our code and only if Qiskit is ran with parallelization during some stage of `instruction_supported`. I will continue to look into this, the bug has been fixed but it would be great to include a test case for this as well.

test/python/transpiler/test_target.py
```diff
@@ -1166,6 +1166,24 @@ def test_instruction_supported_no_args(self):
     def test_instruction_supported_no_operation(self):
         self.assertFalse(self.ibm_target.instruction_supported(qargs=(0,), parameters=[math.pi]))
 
+    def test_target_serialization_preserve_variadic(self):
+        """Checks that variadics are still seen as variadic after serialization"""
+        from pickle import loads, dumps
```
mtreinish: You can move this import to the top of the module, we only need to do a runtime import if there is a potential cycle, but with pickle in the std lib that's not possible.

test/python/transpiler/test_target.py
```diff
@@ -1166,6 +1166,24 @@ def test_instruction_supported_no_args(self):
...
+
+        target = Target("test", 1)
+        # Add variadic example gate with no properties.
+        target.add_instruction(XGate, None, "x_var")
```
mtreinish: It might be better to add a gate that doesn't have a fixed number of qubits, you can use `UnitaryGate` since it can have a variable number of qubits based on the size of the unitary matrix.

mtreinish (commented):
This LGTM it makes sense to remove the duplicated tracking indexmap if it's not needed. Just a couple of small inline comments about the tests.

mtreinish (approved):
LGTM, thanks for the quick update.

Issue #12208, opened 2024-04-18
Move preset pass manager transpiler passes to rust


mtreinish:
### What should we add?

As part of our ongoing effort to migrate more of the rust code for better runtime performance and efficiency we are looking to migrate all the transpiler passes in the preset pass managers to be written in rust. We will still primarily use the python interface for passes as this is a critical part of the transpiler, but the default passes we use should execute solely in rust space and the python portion should just control the dispatch of the rust function.

This is blocked on #11721 and in many cases on #12205 too

The current list of passes involved for this are tracked here:

### Tasks
- [ ] https://github.com/Qiskit/qiskit/issues/12211
- [ ] https://github.com/Qiskit/qiskit/issues/12248
- [ ] https://github.com/Qiskit/qiskit/issues/12249
- [ ] https://github.com/Qiskit/qiskit/issues/12250
- [ ] https://github.com/Qiskit/qiskit/issues/12251
- [ ] https://github.com/Qiskit/qiskit/issues/12253
- [ ] https://github.com/Qiskit/qiskit/issues/12252
- [ ] https://github.com/Qiskit/qiskit/issues/12254
- [ ] https://github.com/Qiskit/qiskit/issues/12255
- [ ] https://github.com/Qiskit/qiskit/issues/12336
- [ ] https://github.com/Qiskit/qiskit/issues/12337
- [ ] https://github.com/Qiskit/qiskit/issues/12256
- [ ] https://github.com/Qiskit/qiskit/issues/12257
- [ ] https://github.com/Qiskit/qiskit/issues/12258
- [ ] https://github.com/Qiskit/qiskit/issues/12259
- [ ] https://github.com/Qiskit/qiskit/issues/12260
- [ ] https://github.com/Qiskit/qiskit/issues/12261
- [ ] https://github.com/Qiskit/qiskit/issues/12262
- [ ] https://github.com/Qiskit/qiskit/issues/12263
- [ ] https://github.com/Qiskit/qiskit/issues/12264
- [ ] https://github.com/Qiskit/qiskit/issues/12265
- [ ] https://github.com/Qiskit/qiskit/issues/12266
- [ ] https://github.com/Qiskit/qiskit/issues/12267
- [ ] https://github.com/Qiskit/qiskit/issues/12268
- [ ] https://github.com/Qiskit/qiskit/issues/12269
- [ ] https://github.com/Qiskit/qiskit/issues/12270
- [ ] https://github.com/Qiskit/qiskit/issues/12271
[...]

PR #12959, opened 2024-08-14
Fully port Optimize1qGatesDecomposition to Rust

mtreinish:
### Summary

This commit builds off of #12550 and the other data model in Rust
infrastructure and migrates the Optimize1qGatesDecomposition pass to
leaves rust until it has finished modifying the DAGCircuit. There is
model that are still in Python, mainly calibrations and parameter
expressions (for global phase). But otherwise the entirety of the pass
operates in rust now.
    
This is just a first pass at the migration here, it moves the pass to be
a single for loop in rust. The next steps here are to look at operating
optimizations being done by the pass so we should be able to the
throughput of the pass by leveraging multithreading to handle each run
in parallel. This commit does not attempt this though, because of the
Python dependency and also the data structures around gates and the
dag aren't really setup for multithreading yet and there likely will
need to be some work to support that (this pass is a good candidate to
work through the bugs on that).

### Details and comments

Part of #12208

This PR is based on top of #12550 and as such github shows the entire contents of #12550 in addition to the contents of this PR. To see the contents of this PR you can look at HEAD on this branch, or just look at:
https://github.com/Qiskit/qiskit/pull/12959/commits/0c8b668f1d16783ed864b6dcbf3366692fce82b3

TODO:

- [x] Rebase after #12550 merges
- [x] Benchmark and tune the implementation
- [x] Investigate if there is any dead code in the `euler_one_qubit_decomposer` module


mtreinish:
I ran some asv benchmarks on this and it's yielding very good looking performance improvements:
```

| Change   | Before [76eb568c] <use-dag-1q-decompose~9^2>   | After [7eaa9544] <use-dag-1q-decompose>   |   Ratio | Benchmark (Parameter)                                                                                         |
|----------|------------------------------------------------|-------------------------------------------|---------|---------------------------------------------------------------------------------------------------------------|
| -        | 5.46±0.02ms                                    | 1.23±0ms                                  |    0.23 | passes.MultipleBasisPassBenchmarks.time_optimize_1q_decompose(5, 1024, ['rx', 'ry', 'rz', 'r', 'rxx', 'id'])  |
| -        | 13.9±0.06ms                                    | 2.96±0.04ms                               |    0.21 | passes.MultipleBasisPassBenchmarks.time_optimize_1q_decompose(14, 1024, ['rx', 'ry', 'rz', 'r', 'rxx', 'id']) |
| -        | 25.4±0.2ms                                     | 4.02±0.02ms                               |    0.16 | passes.MultipleBasisPassBenchmarks.time_optimize_1q_decompose(20, 1024, ['rx', 'ry', 'rz', 'r', 'rxx', 'id']) |
| -        | 4.11±0.01ms                                    | 595±20μs                                  |    0.14 | passes.MultipleBasisPassBenchmarks.time_optimize_1q_decompose(5, 1024, ['u', 'cx', 'id'])                     |
| -        | 16.4±0.03ms                                    | 2.05±0.01ms                               |    0.13 | passes.MultipleBasisPassBenchmarks.time_optimize_1q_decompose(20, 1024, ['u', 'cx', 'id'])                    |
| -        | 5.85±0.01ms                                    | 782±10μs                                  |    0.13 | passes.MultipleBasisPassBenchmarks.time_optimize_1q_decompose(5, 1024, ['rz', 'x', 'sx', 'cx', 'id'])         |
| -        | 15.6±0.06ms                                    | 1.95±0.02ms                               |    0.12 | passes.MultipleBasisPassBenchmarks.time_optimize_1q_decompose(14, 1024, ['rz', 'x', 'sx', 'cx', 'id'])        |
| -        | 11.9±0.04ms                                    | 1.47±0.02ms                               |    0.12 | passes.MultipleBasisPassBenchmarks.time_optimize_1q_decompose(14, 1024, ['u', 'cx', 'id'])                    |
| -        | 30.4±0.3ms                                     | 2.65±0.04ms                               |    0.09 | passes.MultipleBasisPassBenchmarks.time_optimize_1q_decompose(20, 1024, ['rz', 'x', 'sx', 'cx', 'id'])        |

PERFORMANCE INCREASED.
```

mtreinish:
Now that #12550 has merged I've rebased this and it's ready for review.

mtreinish:
Now that this is ready for merging I ran two benchmark runs for this pass against main:

```

| Change   | Before [d3040a0b] <main>            | After [2c0aad55] <use-dag-1q-decompose>   |   Ratio | Benchmark (Parameter)                                                                                         |
|----------|-------------------------------------|-------------------------------------------|---------|---------------------------------------------------------------------------------------------------------------|
| -        | 5.50±0.03ms                         | 1.24±0ms                                  |    0.23 | passes.MultipleBasisPassBenchmarks.time_optimize_1q_decompose(5, 1024, ['rx', 'ry', 'rz', 'r', 'rxx', 'id'])  |
| -        | 14.0±0.08ms                         | 2.95±0.02ms                               |    0.21 | passes.MultipleBasisPassBenchmarks.time_optimize_1q_decompose(14, 1024, ['rx', 'ry', 'rz', 'r', 'rxx', 'id']) |
| -        | 24.3±0.2ms                          | 4.11±0.02ms                               |    0.17 | passes.MultipleBasisPassBenchmarks.time_optimize_1q_decompose(20, 1024, ['rx', 'ry', 'rz', 'r', 'rxx', 'id']) |
| -        | 4.12±0.03ms                         | 572±1μs                                   |    0.14 | passes.MultipleBasisPassBenchmarks.time_optimize_1q_decompose(5, 1024, ['u', 'cx', 'id'])                     |
| -        | 15.9±0.06ms                         | 2.01±0.01ms                               |    0.13 | passes.MultipleBasisPassBenchmarks.time_optimize_1q_decompose(20, 1024, ['u', 'cx', 'id'])                    |
| -        | 5.99±0.02ms                         | 768±4μs                                   |    0.13 | passes.MultipleBasisPassBenchmarks.time_optimize_1q_decompose(5, 1024, ['rz', 'x', 'sx', 'cx', 'id'])         |
| -        | 15.7±0.2ms                          | 1.92±0.03ms                               |    0.12 | passes.MultipleBasisPassBenchmarks.time_optimize_1q_decompose(14, 1024, ['rz', 'x', 'sx', 'cx', 'id'])        |
| -        | 11.8±0.05ms                         | 1.44±0.01ms                               |    0.12 | passes.MultipleBasisPassBenchmarks.time_optimize_1q_decompose(14, 1024, ['u', 'cx', 'id'])                    |
| -        | 28.1±0.2ms                          | 2.63±0.01ms                               |    0.09 | passes.MultipleBasisPassBenchmarks.time_optimize_1q_decompose(20, 1024, ['rz', 'x', 'sx', 'cx', 'id'])        |

PERFORMANCE INCREASED.
```

and against 1.2.0 (to discount any potential performance change from https://github.com/Qiskit/qiskit/pull/12550):
```

| Change   | Before [7e2e835e] <1.2.0^0>   | After [2c0aad55] <use-dag-1q-decompose>   |   Ratio | Benchmark (Parameter)                                                                                         |
|----------|-------------------------------|-------------------------------------------|---------|---------------------------------------------------------------------------------------------------------------|
| -        | 5.53±0.02ms                   | 1.23±0.01ms                               |    0.22 | passes.MultipleBasisPassBenchmarks.time_optimize_1q_decompose(5, 1024, ['rx', 'ry', 'rz', 'r', 'rxx', 'id'])  |
| -        | 14.0±0.1ms                    | 2.98±0.03ms                               |    0.21 | passes.MultipleBasisPassBenchmarks.time_optimize_1q_decompose(14, 1024, ['rx', 'ry', 'rz', 'r', 'rxx', 'id']) |
| -        | 25.7±0.1ms                    | 4.08±0.03ms                               |    0.16 | passes.MultipleBasisPassBenchmarks.time_optimize_1q_decompose(20, 1024, ['rx', 'ry', 'rz', 'r', 'rxx', 'id']) |
| -        | 4.14±0.02ms                   | 584±10μs                                  |    0.14 | passes.MultipleBasisPassBenchmarks.time_optimize_1q_decompose(5, 1024, ['u', 'cx', 'id'])                     |
| -        | 5.93±0.02ms                   | 778±10μs                                  |    0.13 | passes.MultipleBasisPassBenchmarks.time_optimize_1q_decompose(5, 1024, ['rz', 'x', 'sx', 'cx', 'id'])         |
| -        | 15.8±0.09ms                   | 1.95±0.03ms                               |    0.12 | passes.MultipleBasisPassBenchmarks.time_optimize_1q_decompose(14, 1024, ['rz', 'x', 'sx', 'cx', 'id'])        |
| -        | 12.1±0.09ms                   | 1.49±0.02ms                               |    0.12 | passes.MultipleBasisPassBenchmarks.time_optimize_1q_decompose(14, 1024, ['u', 'cx', 'id'])                    |
| -        | 16.5±0.1ms                    | 2.06±0.04ms                               |    0.12 | passes.MultipleBasisPassBenchmarks.time_optimize_1q_decompose(20, 1024, ['u', 'cx', 'id'])                    |
| -        | 30.5±0.2ms                    | 2.61±0.05ms                               |    0.09 | passes.MultipleBasisPassBenchmarks.time_optimize_1q_decompose(20, 1024, ['rz', 'x', 'sx', 'cx', 'id'])        |

PERFORMANCE INCREASED.
```

crates/accelerate/src/nlayout.rs
```diff
@@ -72,6 +73,25 @@ impl PhysicalQubit {
...
+
+/// This is only safe in the context of a physical circuit during transpilation
+/// after the qubit indices of the circuit/dag circuit refer to the physical
+/// qubits (once we've run a layout pass and applied it).
+impl From<&Qubit> for PhysicalQubit {
+    fn from(s: &Qubit) -> PhysicalQubit {
+        PhysicalQubit::new(s.0)
+    }
+}
+
```
jakelishman: These are pretty huge caveats on a really core trait implementation.  I think leaving it as requiring `PhysicalQubit::new(q.0)` is probably better, or adding `Qubit::as_physical` or something, so you always have to be explicit about what you're doing.  `.into()`'s a bit magical to have any sort of caveat for me, personally.

ShellyGarion (commented):
I think that `euler_one_qubit_decomposer.rs` should be moved to `crates/accelerate/src/synthesis/one_qubit` to match with the structure of the current synthesis library.
Similarly, it would be good if `two_qubit_decompose.rs` would be moved to `crates/accelerate/src/synthesis/two_qubits`

crates/accelerate/src/nlayout.rs
mtreinish: That's a fair point, I'll add a dedicated method to `Qubit` for it. I started this by doing `PhysicalQubit::new(q.0)` and thought it was a bit tedious to do it that way.

mtreinish:
> I think that euler_one_qubit_decomposer.rs should be moved to crates/accelerate/src/synthesis/one_qubit to match with the structure of the current synthesis library.
Similarly, it would be good if two_qubit_decompose.rs would be moved to crates/accelerate/src/synthesis/two_qubits

We can reorganize these modules in a follow up PR. When I created these modules there wasn't any other synthesis in rust so there wasn't anywhere else to put them. I think it'll probably be best to do this at the same time we create a transpiler crate so the `Optimize1qGatesDecomposition` function that I modify here is separate from the base synthesis code.

crates/accelerate/src/nlayout.rs
mtreinish: Actually I took a quick look at doing this and it will require some reorganization of the code I think. The problem is `PhysicalQubit` is defined in `accelerate::nlayout` and `Qubit` is defined in `qiskit_circuit`. I don't want to add a bidirectional dependency between `circuit` and `accelerate`. I think in a separate PR we'll want to probably move `nlayout` into circuit (as `PhysicalQubit` and `VirtualQubit`'s methods have a dependency on `NLayout`) and define all the `Qubit`  types together. Then we can do `Qubit::as_physical` as part of that reorganization.

mtreinish: I removed the trait impl for this here: https://github.com/Qiskit/qiskit/pull/12959/commits/9b43ea7c02fdba8d8c4ea6a4c70c5961f4faf61a

crates/accelerate/src/euler_one_qubit_decomposer.rs
```diff
@@ -571,21 +577,116 @@ pub fn generate_circuit(
...
+    U = 2,
+    PSX = 3,
+    U1X = 4,
+    RR = 5,
+    ZYZ = 6,
+    ZXZ = 7,
+    XZX = 8,
+    XYX = 9,
+    ZSXX = 10,
+    ZSX = 11,
```
ElePT, line 689: What are these indices used for? I'm sure it's there but I haven't been able to find it in the code

crates/accelerate/src/euler_one_qubit_decomposer.rs
```diff
@@ -965,70 +1045,198 @@ pub fn params_zxz(unitary: PyReadonlyArray2<Complex64>) -> [f64; 4] {
...
+                        break;
+                    }
+                }
+            }
+        } else {
+            outside_basis = false;
+        }
+        if outside_basis
+            || new_error < old_error
+            || new_error.0.abs() < 1e-9 && old_error.0.abs() >= 1e-9
```
ElePT, line 1230: Just because we were discussing tolerances earlier, why did you decide to use 1e-9 here?

crates/accelerate/src/nlayout.rs
```diff
@@ -24,7 +24,7 @@ use hashbrown::HashMap;
 macro_rules! qubit_newtype {
     ($id: ident) => {
         #[derive(Debug, Clone, Copy, PartialEq, Eq, PartialOrd, Ord, Hash)]
-        pub struct $id(u32);
+        pub struct $id(pub u32);
```
ElePT, line 27: This is to unlock what's done in line 1090, right? `PhysicalQubit::new(dag.get_qargs(inst.qubits)[0].0)` (not actionable, I just want to understand it because I will need to access info similarly).

crates/accelerate/src/target_transpiler/mod.rs
```diff
@@ -940,6 +940,17 @@ impl Target {
         });
     }
 
+    /// Get the error rate of a given instruction in the target
+    pub fn get_error(&self, name: &str, qargs: &[PhysicalQubit]) -> Option<f64> {
+        self.gate_map.get(name).and_then(|gate_props| {
```
ElePT, line 945: I hadn't seen (or noticed) the `and_then()` before. Looks handy.

releasenotes/notes/optimize-1q-gates-decomposition-ce111961b6782ee0.yaml
```diff
@@ -0,0 +1,13 @@
+---
+features_transpiler:
+  - |
+    Added a new method :meth:`.DAGCircuit.control_flow_ops` which provides a fast
+    path to get all the :class:`.DAGOpNode` in a :class:`.DAGCircuit` that
+    contain a :class:`.ControlFlowOp`. This was possible before using the
+    :meth:`.DAGCircuit.op_nodes` method and passing the ``ControlFlowOp`` class
+    as a filter, but this new function will perform the same operation but
+    perform it faster.
```
ElePT: small suggestion to avoid the double "but":

```suggestion
    :meth:`.DAGCircuit.op_nodes` method and passing the :class:`.ControlFlowOp` class
    as a filter, but this new function will perform the operation faster.
```

ElePT (dismissed):
Thanks a lot @mtreinish!! Apart from the cool optimizations, the PR also includes a lot of handy dag circuit utilities, and I think we would all benefit from having it merged soon (I definitely will). I left a few questions (just for my info) and a tiny non-blocking reno suggestion, but LGMT.

crates/accelerate/src/euler_one_qubit_decomposer.rs
mtreinish: It's to force an specific integer for each variant, to be explicit so that the variant matches it's position in each of the static arrays. It's implicit with the ordering, but it's best to be explicit to avoid potential issues since we're doing all the look ups in the statics.

crates/accelerate/src/nlayout.rs
mtreinish: Yeah it's so I can access the inner u32 with `.0`.

ElePT (approved):

crates/accelerate/src/euler_one_qubit_decomposer.rs
mtreinish: It's because the python code this is based on was using it: https://github.com/Qiskit/qiskit/blob/main/qiskit/transpiler/passes/optimization/optimize_1q_decomposition.py#L199 the default tolerance for `math.is_close()` is 1e-9 https://docs.python.org/3/library/math.html#math.isclose

PR #12961, opened 2024-08-15
Adding MCX synthesis plugins

alexanderivrii:
### Summary

This PR builds on top of #12904 and #12911 and adds additional MCX synthesis methods and MCX synthesis plugins, some support clean and dirty auxiliary qubits.

The default plugin for MCX gates runs other plugins maximally trying to exploit the sets of clean and dirty ancilla qubits available: it first tries the 'clean-n' method, then the 'dirty-n' method, then the 'clean-1' method, and finally the reduction to mcphase.

The `MCXRecursive` gate and other custom variants are marked for pending deprecation (and will be removed some time in the future).

### Details and comments

The `high_level_synthesis.py` file was split into two: one containing the transpiler pass, and one containing the plugins.

There were a few additional caveats:
- The `MCXRecursive` (and other custom MCX variants) inherit from `MCXGate` but are not called `"mcx"`
- The `C3XGate` and `C4XGate` are called `"mcx"` but do not inherit from `MCXGate`.

The first point will be cleaned up by deprecating custom MCX variants. The second point might be cleaned up by renaming the names for `C3XGate` and `C4XGate`, however might introduce some breaking changes and should be done for 2.0.


alexanderivrii:
One other thing: I would like to replace `_define` for `C3XGate` and `C4XGate` in `x.py` by `synthesis/multi-controlled/synth_c3x` and `synth_c4x`, but this leads to circular imports (these are singleton gates so the imports seem to be happening on the top-level). Help appreciated.

alexanderivrii:
A quick experiment on benchpress multi_control_circuit benchmark (see #12729 for some prior numbers):

Number of 2-qubit gates:
* 10 qubits: tket: 937, before this PR: 1269, with this PR: 809
* 15 qubits: tket: 3617, before this PR: 5849, with this PR: 2293

The improvement comes from better synthesis algorithms that use clean/dirty ancilla qubits

qiskit/synthesis/__init__.py
```diff
@@ -180,8 +180,12 @@
     two_qubit_cnot_decompose,
     TwoQubitWeylDecomposition,
 )
-from .multi_controlled.mcx_with_ancillas_synth import (
+from .multi_controlled import (
     synth_mcx_n_dirty_i15,
     synth_mcx_n_clean_m15,
     synth_mcx_1_clean_b95,
+    synth_mcx_mcphase,
```
ShellyGarion: could you add the synth_mcx functions to the API docs? (you need to add them a few lines above on this file too).

qiskit/synthesis/multi_controlled/mcx_synthesis.py
```diff
@@ -230,3 +233,113 @@ def synth_mcx_1_clean_b95(num_ctrl_qubits: int):
     )
 
     return qc
+
+
+def synth_mcx_gray_code(num_ctrl_qubits: int) -> QuantumCircuit:
```
ShellyGarion, line 238: Do we have a reference to this method? if so, we could change the name to show it.
Otherwise, I'm fine with the name gray_synth.

qiskit/synthesis/multi_controlled/mcx_synthesis.py
```diff
@@ -230,3 +233,113 @@ def synth_mcx_1_clean_b95(num_ctrl_qubits: int):
...
+    r"""
+    Synthesize a multi-controlled X gate with :math:`k` controls using the Gray code.
+
+    Produces a quantum circuit with :math:`num_ctrl_qubits + 1` qubits.
```
ShellyGarion: Please add a note that this synthesis method is not advised to use since it's not scalable as the number of CX gates grows exponentially in the number of qubits

qiskit/synthesis/multi_controlled/mcx_synthesis.py
```diff
@@ -230,3 +233,113 @@ def synth_mcx_1_clean_b95(num_ctrl_qubits: int):
...
+    num_qubits = num_ctrl_qubits + 1
+    q = QuantumRegister(num_qubits, name="q")
+    qc = QuantumCircuit(q, name="mcx_gray")
+    qc._append(HGate(), [q[-1]], [])
+    qc._append(MCU1Gate(np.pi, num_ctrl_qubits=num_ctrl_qubits), q[:], [])
+    qc._append(HGate(), [q[-1]], [])
+    return qc
+
+
+def synth_mcx_mcphase(num_ctrl_qubits: int) -> QuantumCircuit:
```
ShellyGarion: Any better suggestions for this function name? the mcp synth is based on iteratively using mcrz synth.
I would suggest `synth_mcx_basic` ?

qiskit/synthesis/multi_controlled/mcx_synthesis.py
```diff
@@ -230,3 +233,113 @@ def synth_mcx_1_clean_b95(num_ctrl_qubits: int):
...
+    r"""
+    Synthesize a multi-controlled X gate with :math:`k` controls based on
+    the implementation for MCPhaseGate.
+
+    Produces a quantum circuit with :math:`num_ctrl_qubits + 1` qubits.
```
ShellyGarion: please add a comment that the number of CX gates grows quadratically in the number of qubits

qiskit/circuit/library/standard_gates/x.py
```diff
@@ -1305,15 +1304,9 @@ def inverse(self, annotated: bool = False):
...
-        from .u1 import MCU1Gate
-        from .h import HGate
+        from qiskit.synthesis.multi_controlled import synth_mcx_gray_code
 
-        q = QuantumRegister(self.num_qubits, name="q")
-        qc = QuantumCircuit(q, name=self.name)
-        qc._append(HGate(), [q[-1]], [])
-        qc._append(MCU1Gate(numpy.pi, num_ctrl_qubits=self.num_ctrl_qubits), q[:], [])
-        qc._append(HGate(), [q[-1]], [])
+        qc = synth_mcx_gray_code(self.num_ctrl_qubits)
```
ShellyGarion, line 1322: I think we would also like to deprecate (pending) MCXGrayCode too ?

qiskit/synthesis/__init__.py
```diff
@@ -180,8 +180,12 @@
...
+    synth_mcx_gray_code,
+    synth_c3x,
+    synth_c4x,
```
ShellyGarion, line 194: I don't think we would like to add synt_c3x and synt_c4x to the docs (since we don't want to copy the circuit library into the synthesis library).
in fact, why can't we use the C3XGate and C4XGate directly?

qiskit/synthesis/multi_controlled/mcx_synthesis.py
```diff
@@ -230,3 +233,113 @@ def synth_mcx_1_clean_b95(num_ctrl_qubits: int):
...
+    qc = QuantumCircuit(q)
+    q_controls = list(range(num_ctrl_qubits))
+    q_target = num_ctrl_qubits
+    qc.h(q_target)
+    qc.mcp(np.pi, q_controls, q_target)
+    qc.h(q_target)
+    return qc
+
+
+def synth_c3x() -> QuantumCircuit:
```
ShellyGarion, line 300: why should we copy the synthesis code here and not use C3XGate and C4XGate directly?

qiskit/synthesis/multi_controlled/mcx_synthesis.py
```diff
@@ -122,8 +131,8 @@ def synth_mcx_n_dirty_i15(
     return qc
 
 
-def synth_mcx_n_clean_m15(num_ctrl_qubits: int):
-    """
+def synth_mcx_n_clean_m15(num_ctrl_qubits: int) -> QuantumCircuit:
```
ShellyGarion, line 134: please note that in this code we also provide the circuit names (this is relevant to all the functions).
these are currently "mcx_vchain" or "mcx_recursive".
I would suggest to change them to "mcx" and update the QASM tests accordingly (if it's not considered breaking API)

ShellyGarion (commented):
Thanks @alexanderivrii! I've made some preliminary comments (mainly on the functions names and API)

qiskit/synthesis/__init__.py
alexanderivrii: This is done in [4135344](https://github.com/Qiskit/qiskit/pull/12961/commits/41353440aae98f7f94407a58b9e2652ee244c474).

qiskit/synthesis/multi_controlled/mcx_synthesis.py
Cryoris: Doesn't `mcrz` use the `_mcsu2` work? Maybe we can use that paper as reference?

qiskit/synthesis/multi_controlled/mcx_synthesis.py
Cryoris: I think the problem is that `C3XGate` and `C4XGate` are called `"mcx"` which would lead to an infinite recursion in the HLS plugin. Also I'm not sure what the benefit of having these gates is, it might be best to deprecate them as well in favor of just using `MCXGate` -- what do you think?

If we want to keep them, an alternative solution to the infinite-recursion could be to instance-check on `C3XGate/C4XGate` and return `None`, which will then trigger the unroller.

qiskit/synthesis/multi_controlled/mcx_synthesis.py
alexanderivrii: I was not able to find the reference (and it does not appear in the `MCXGrayCode` gate's docstring).

qiskit/synthesis/multi_controlled/mcx_synthesis.py
alexanderivrii: Good point, adding a sentence on that in the docstring.

qiskit/synthesis/multi_controlled/mcx_synthesis.py
Cryoris: We should be careful with the name `"mcx"`, for example, I could imagine that calling the `AerSimulator` with a decomposition with ancillas and the name `"mcx"` will cause wrong results, as the simulator could think that we have more controls.

Wouldn't it be safer to just call them individually (i.e. `mcx_m15`)? If we reach this function, then the circuit is getting synthesized, so I think it's fine for the circuit to change and for operations to change names.

qiskit/synthesis/multi_controlled/mcx_synthesis.py
alexanderivrii: I was looking for a better name for this too, so ideas are welcome. I am not particularly fond of `synth_mcx_basic` though, as the word "basic" is too generic. By the way, do we have a reference for this?

qiskit/synthesis/multi_controlled/mcx_synthesis.py
alexanderivrii: adding

qiskit/circuit/library/standard_gates/x.py
alexanderivrii: I did this (in the original commit).

qiskit/synthesis/multi_controlled/mcx_synthesis.py
alexanderivrii: I actually did this in one of the previous commits and I believe it broke some of the QASM tests: for instance `synth_mcx_n_clean_m15` is used in the definition of `MCXVChain`. We will be able to clean it up after the custom MCX gates are deprecated. Though please note that the name "mcx" is not ideal either, since we might want to eventually enforce that an object called "mcx" is an "MCXGate" (the same way as an object called "h'" is a Hadamard); this is already not the case (and requires additional `isinstance` checks in the plugins), but let's not make the situation worse.

qiskit/synthesis/__init__.py
alexanderivrii: I believe that the actual synthesis methods should be in the synthesis library and not in the circuit library (as part of gates' `definition`s).

Some other things to consider:
* Do we want to reuse these functions from other places? I believe we might.
* What if at some point we will find better decompositions for 3- and 4- controlled X-gates? I think changing the synthesis methods is easier than going inside the gate's definition.
* Do we want to remove C3X and C4X gates altogether? I don't think we have plans for that, but this point periodically comes up in offline discussions.

qiskit/synthesis/multi_controlled/mcx_synthesis.py
alexanderivrii: We discussed this offline with @Cryoris and thought that calling `synth_c3x()` and `synth_c4x()` instead of using `C3XGate` and `C4XGate` directly is a slightly better approach. A more minor point is that this decreases the level of "nestedness" and should be slightly faster. A more important problem (as I have already complained in the PR summary) is that C3XGate and C4XGate have name `"mcx"`, so they are caught by `HighLevelSynthesis` plugins for MCX-gates (which I think they should). (Note: fortunately CX and CCX gates are _not_ called `"mcx"`). In order to avoid recursion, we need to ensure that a synthesis plugin for MCX gates simplifies the logic, i.e. it may replace a 10-controlled MCX-gate by some smaller MCX-gates, but it should not replace a `C3XGate` by a `C3XGate`. This is exactly what was happening before. We could have changed this both on the plugin level or the synthesis function level, but thought that the second option is slightly cleaner. I can also do it the other way around, if you think that's better.

qiskit/circuit/library/standard_gates/x.py
```diff
@@ -1280,6 +1268,17 @@ def __new__(
             return gate
         return super().__new__(cls)
 
+    @deprecate_func(
+        additional_msg=(
+            "It is recommended to use :class:`.MCXGate` and let :class:`.HighLevelSynthesis` choose "
```
Cryoris: The release-notes-reader in me is happy about the Sphinx links, but this message will only be displayed on the command line, so no need for annotations like `:class:` or `:func:` 🙂 In fact, personally I would even just write plain `MCXGate` (without the `'` bits) to make it easier to read, but this is up to you

qiskit/circuit/library/standard_gates/x.py
```diff
@@ -1280,6 +1268,17 @@ def __new__(
...
+            "the best synthesis method depending on the number of ancilla qubits available. "
+            "If this specific synthesis method is required, one can specify it using the plugin "
+            "interface. Alternatively, one can use :func:`.synth_mcx_gray_code` to construct "
```
Cryoris: Do you want to point out the name of the plugin option? Otherwise we could add an example in the release notes.

qiskit/circuit/quantumcircuit.py
```diff
@@ -3212,10 +3212,17 @@ def decompose(
         from qiskit.converters.dag_to_circuit import dag_to_circuit
 
         dag = circuit_to_dag(self, copy_operations=True)
-        dag = HighLevelSynthesis().run(dag)
+
+        if gates_to_decompose is None:
```
Cryoris, line 3216: Personally I would expect that an `MCXGate` or `Clifford` will be decomposed when I call `.decompose`, no?

qiskit/synthesis/multi_controlled/mcx_synthesis.py
```diff
@@ -10,20 +10,28 @@
...
 
 
 def synth_mcx_n_dirty_i15(
     num_ctrl_qubits: int,
     relative_phase: bool = False,
     action_only: bool = False,
-):
-    """
+) -> QuantumCircuit:
+    r"""
```
Cryoris, line 34: Good catch!

qiskit/synthesis/multi_controlled/mcx_synthesis.py
```diff
@@ -59,7 +67,8 @@ def synth_mcx_n_dirty_i15(
         qc.ccx(q_controls[0], q_controls[1], q_target)
         return qc
     elif not relative_phase and num_ctrl_qubits == 3:
-        qc._append(C3XGate(), [*q_controls, q_target], [])
+        circuit = synth_c3x()
+        qc.append(circuit, [*q_controls, q_target])
```
Cryoris: I'm wondering whether we should skip the extra wrapping here, which would be more performant in unrolling and save us circuit copies in the construction.
```suggestion
        qc.compose(synth_c3x(), [*q_controls, q_target], inplace=True, copy=False)
```

qiskit/synthesis/multi_controlled/mcx_synthesis.py
```diff
@@ -230,3 +233,116 @@ def synth_mcx_1_clean_b95(num_ctrl_qubits: int):
...
+        num_ctrl_qubits: The number of control qubits.
+
+    Returns:
+        The synthesized quantum circuit.
+    """
+    num_qubits = num_ctrl_qubits + 1
+    q = QuantumRegister(num_qubits, name="q")
+    qc = QuantumCircuit(q, name="mcx_gray")
+    qc._append(HGate(), [q[-1]], [])
+    qc._append(MCU1Gate(np.pi, num_ctrl_qubits=num_ctrl_qubits), q[:], [])
```
Cryoris, line 256: It would be more straightforward to just call `_gray_code_chain` with `gate=CXGate()` from `../library/u3.py` (and less confusing IMO since we have very different algorithms for `MCPhase` and `MCU1` even they should be literally the same) -- should we just update this while we're at it?

qiskit/synthesis/multi_controlled/mcx_synthesis.py
```diff
@@ -230,3 +233,116 @@ def synth_mcx_1_clean_b95(num_ctrl_qubits: int):
...
+
+def synth_c4x() -> QuantumCircuit:
+    """Efficient synthesis of 4-controlled X-gate."""
+
+    q = QuantumRegister(5, name="q")
+    qc = QuantumCircuit(q, name="mcx")
+
+    rules = [
+        (HGate(), [q[4]], []),
+        (CU1Gate(np.pi / 2), [q[3], q[4]], []),
```
Cryoris, line 344: Could we update all `U1` occurences to `Phase` gates or does that break the tests? I think we should move away from these where we can 🙂

qiskit/transpiler/passes/synthesis/hls_plugins.py
```diff
@@ -0,0 +1,924 @@
...
+
+    References:
+        1. Iten et. al., *Quantum Circuits for Isometries*, Phys. Rev. A 93, 032318 (2016),
+           `arXiv:1501.06911 <http://arxiv.org/abs/1501.06911>`_
+    """
+
+    def run(self, high_level_object, coupling_map=None, target=None, qubits=None, **options):
+        """Run synthesis for the given MCX gate."""
+
+        if high_level_object.name != "mcx":
```
Cryoris: This shouldn't be necessary since the HLS should only give us things called `mcx` right? The other plugins also don't have this additional safety 🙂

qiskit/transpiler/passes/synthesis/hls_plugins.py
```diff
@@ -0,0 +1,924 @@
...
+            return None
+
+        # Iteratively run other synthesis methods available
+
+        if (
+            decomposition := MCXSynthesisNCleanM15().run(
+                high_level_object, coupling_map, target, qubits, **options
+            )
+        ) is not None:
+            return decomposition
```
Cryoris, line 910: This works and we can probably do it this way for the beginning, but in future we'll surely need a method to query how many auxiliary qubits a plugin requires. We could already add it here and have a code structure that directly selects the correct plugin based on the number of available auxiliary qubits. What do you think?

Cryoris (commented):
This already looks great, thanks for the effort! I left some comments below 🙂

qiskit/synthesis/multi_controlled/mcx_synthesis.py
ShellyGarion: here is a reference for the mcrz (mcsu2) paper, but this code is a variant that does not appear in the paper.

[1] R. Vale et al. Decomposition of Multi-controlled Special Unitary Single-Qubit Gates
            arXiv:2302.06377 (2023) <https://arxiv.org/abs/2302.06377>

alexanderivrii: Thanks for the reference; adding.

qiskit/circuit/library/standard_gates/x.py
alexanderivrii: I like the "release-notes-reader in you" comment :). This is now fixed.

qiskit/circuit/library/standard_gates/x.py
alexanderivrii: Good point, I have done this.

qiskit/circuit/quantumcircuit.py
alexanderivrii: This is a bit of a corner-case. When a circuit has an `MCXGate` or a `Clifford`, the `decompose` function does decompose the object into lower-level gates. 

The question is what we should do when `decompose` is called with a non-`None` list of gates to decompose,  in which case it is expected that only the gates in this list will get decomposed but not any other gates (and in fact we have a test to check that an MCX gate will not get decomposed in this case). However, HighLevelSynthesis now does decompose MCXGates. So the change is that in this case we do not run `HighLevelSynthesis` prior to running `Decompose`. This is not a perfect fix, but at least avoids the failing testcases.

A better fix, as `HighLevelSynthesis` mostly subsumes the `Decompose` pass anyway is to add more options to `HighLevelSynthesis` dictating which gates to decompose and specifying the depth of the recursion.

qiskit/synthesis/multi_controlled/mcx_synthesis.py
alexanderivrii: I actually tried this, but there is a case when the circuit has more qubits than the decomposition provided by `synth_c3x()`. In this case, `append` works correctly, but `compose` (wrongly) does not do anything. But I probably need to specify the qubit mapping correctly.

qiskit/synthesis/multi_controlled/mcx_synthesis.py
alexanderivrii: Right now we have just moved the code without doing any real changes, for simplicity let's keep it this way for this PR. But I completely agree that we need to rethink the hierarchy of all the different multi-controlled synthesis methods available.

qiskit/synthesis/multi_controlled/mcx_synthesis.py
alexanderivrii: Again, I like the suggestion but propose to do this in a follow-up (since it may probably break some of the qasm tests).

qiskit/transpiler/passes/synthesis/hls_plugins.py
alexanderivrii: This is a good point and I can remove these checks as in the normal flow only the `HighLevelSynthesis` transpiler pass would call the plugins and the interface guarantees that the names would be correct. 

However, I am slightly worried about the case when a user might decide to call synthesis plugins directly (as they are documented in the API docs) on MCX gates, and in particular on something like a `CCXGate` (which in Qiskit is not an `MCXGate` but this decisions seems to have been rather arbitrary). So I wanted the plugins to complain in this case. Of course, the same problem exists if a user would called a QFT-synthesis plugin on a Clifford object, however intuitively it's harder to confuse `QFT` with `Clifford` than `CCX` with `MCX`. What do you think?

qiskit/transpiler/passes/synthesis/hls_plugins.py
alexanderivrii: As discussed, I have also marked the `MCXGate`'s function `get_num_ancilla_qubits` for pending deprecation. 

IMHO, in general a plugin should not know how many ancilla qubits it would require before it runs: depending on the high-level object itself, on options, on luck, or on random seed, it may return different decompositions using different numbers of ancilla qubits. (But note that a plugin does receive how many ancilla qubits are available, and cannot use more than this number of qubits).

qiskit/synthesis/multi_controlled/mcx_synthesis.py
ShellyGarion: I still think that we should find a better name than " synth_mcx_mcphase"

qiskit/circuit/quantumcircuit.py
Cryoris: I see! It seems indeed that there's some duplicate functionality... let's discuss this in more detail, since HLS already does a lot of things I'm not sure we should add more 🤔

qiskit/synthesis/multi_controlled/mcx_synthesis.py
Cryoris: Yeah `compose` should work 🤔 but not doing anything is strange, that would mean that either the circuit was empty or maybe `inplace=True` was missing?

qiskit/transpiler/passes/synthesis/hls_plugins.py
Cryoris: I agree for plugins that can use different strategies, but the specific MCX plugins e.g. `MCXSynthesisNCleanM15` have exactly one synthesis method that they use, so it would be safe to add it there, no? Also, there's no way to infer the number of ancilla qubits otherwise (without actually running the decomposition), if we deprecate it on `MCXGate`.

qiskit/synthesis/multi_controlled/mcx_synthesis.py
alexanderivrii: After an offline discussion with @ShellyGarion and @Cryoris, we have settled on `synth_v24`. I am not really fond of this or other names, but this is the best we could think of and consistent with the naming of other methods.

qiskit/synthesis/multi_controlled/mcx_synthesis.py
alexanderivrii: Ah, `inplace=True` solved it.

qiskit/transpiler/passes/synthesis/hls_plugins.py
alexanderivrii: Removed (after an offline discussion).

qiskit/transpiler/passes/synthesis/hls_plugins.py
alexanderivrii: We discussed this offline, and decided to leave it as is implemented right now. We will need to rethink this in the future once we start looking at analysis passes that "optimally" divide the ancilla qubits between different high-level-objects.

alexanderivrii:
Thanks @Cryoris for the review. I have addressed of all your review comments (I think), so it's ready for review once again. The only remaining question (for the rest of the team) is what you think about the current plugin names ``"n_clean_m15"``, ``"v24"``, ``"1_clean_b95"`` -- these do look a bit ugly, but the best naming convention we could come up so far is to base the names off what they do (e.g., use ``n`` clean ancillas) and the name of the first author + the year of the paper describing the technique (e.g., Vale et al's 2024 paper).

qiskit/synthesis/multi_controlled/mcx_synthesis.py
Cryoris: Yeah, we should definitely introduce a human-readable alias for these methods, which users can call in the HLS 🙂

qiskit/circuit/library/standard_gates/x.py
```diff
@@ -1330,6 +1336,17 @@ class MCXRecursive(MCXGate):
         2. Iten et al., 2015. https://arxiv.org/abs/1501.06911
     """
 
+    @deprecate_func(
+        additional_msg=(
+            "It is recommended to use MCXGate and let HighLevelSynthesis choose "
+            "the best synthesis method depending on the number of ancilla qubits available. "
+            "If this specific synthesis method is required, one can specify it using the "
+            "high-level-synthesis plugin `mcx_1_clean` for MCX gates, or, alternatively, "
+            "one can use synch_mcx_1_clean to construct the gate directly."
```
Cryoris:
```suggestion
            "one can use synth_mcx_1_clean to construct the gate directly."
```

qiskit/circuit/library/standard_gates/x.py
```diff
@@ -1330,6 +1336,17 @@ class MCXRecursive(MCXGate):
         2. Iten et al., 2015. https://arxiv.org/abs/1501.06911
     """
 
+    @deprecate_func(
+        additional_msg=(
+            "It is recommended to use MCXGate and let HighLevelSynthesis choose "
+            "the best synthesis method depending on the number of ancilla qubits available. "
+            "If this specific synthesis method is required, one can specify it using the "
+            "high-level-synthesis plugin `mcx_1_clean` for MCX gates, or, alternatively, "
```
Cryoris: Are these names still correct after the renaming? In the tests you're using 
```suggestion
            "high-level-synthesis plugin '1_clean_b95' for MCX gates, or, alternatively, "
```

releasenotes/notes/add-mcx-plugins-85e5b248692a36db.yaml
```diff
@@ -0,0 +1,38 @@
...
+
+    * :class:`.MCXSynthesisNCleanM15`, based on :func:`.synth_mcx_n_clean_m15`.
+    * :class:`.MCXSynthesisNDirtyI15`, based on :func:`.synth_mcx_n_dirty_i15`.
+    * :class:`.MCXSynthesis1CleanB95`, based on :func:`.synth_mcx_1_clean_b95`.
+    * :class:`.MCXSynthesisV24`, based on :func:`.synth_mcx_v24`.
+    * :class:`.MCXSynthesisGrayCode`, based on :func:`.synth_mcx_gray_code`.
+
+    As well:
+
+    * :class:`.MCXSynthesisGrayDefault`, choosing the most efficient synthesis
```
Cryoris: Should this be the following?
```suggestion
    * :class:`.MCXSynthesisDefault`, choosing the most efficient synthesis
```

releasenotes/notes/add-mcx-plugins-85e5b248692a36db.yaml
```diff
@@ -0,0 +1,38 @@
...
+      method based on the number of clean and dirty ancilla qubits available.
+
+deprecations:
```
Cryoris: We don't mention pending deprecation afaik, I think this can be removed 🙂

releasenotes/notes/add-mcx-plugins-85e5b248692a36db.yaml
```diff
@@ -0,0 +1,38 @@
...
+  - |
+    Added synthesis functions :func:`.synth_mcx_gray_code` and :func:`.synth_mcx_v24`
+    that synthesize multi-controlled X gates. These functions do not require additional
+    ancilla qubits.
+  - |
+    Added synthesis functions :func:`.synth_c3x` and  :func:`.synth_c4x` that
+    synthesize 3-controlled and 4-controlled X-gates respectively.
+features_transpiler:
+  - |
+    Added multiple high-level-synthesis plugins for synthesizing an :class:`.MCXGate`:
```
Cryoris, line 12: Would you like to add an example code snippet on how this can be used with a special synthesis plugin? That might be cool to see for users 🙂

alexanderivrii: done in [5a8ba64](https://github.com/Qiskit/qiskit/pull/12961/commits/5a8ba642a7024566ef8d78d06a2dd49526d66662)

releasenotes/notes/add-mcx-plugins-85e5b248692a36db.yaml
alexanderivrii: done in [5a8ba64](https://github.com/Qiskit/qiskit/pull/12961/commits/5a8ba642a7024566ef8d78d06a2dd49526d66662)

ShellyGarion (commented):
LGTM. Thanks @alexanderivrii !
I only have one question, I generated the API docs, and could not find the hls_plugins. 
Do you know where should they be located now?

alexanderivrii:
@ShellyGarion, thanks for noticing! The docs now correctly appear in: html_docs/html/apidoc/synthesis.html

ShellyGarion:
Another small comment - the last table in `transpiler_synthesis_plugins.html` shows the following, so all the references are [1].
So either remove it, or specify first author and a year.
image

alexanderivrii:
@ShellyGarion , thanks for noticing that too. Fixed.

Cryoris (approved):
LGTM thanks for all the effort! I'll wait with putting it the merge queue, if @ShellyGarion still has some comments 🙂

PR #12963, opened 2024-08-15
Make PrimitiveJob serializable

### LINKED ISSUE #12787: Make `DataBin` pickle-able
### What should we add?

Currently when using pickle with `PrimitiveResult` or `PubResult` the `DataBin` class will throw a `NotImplementedError` from `__setattr__()` when you try to `pickle.load()`.

The change needed is trivial, so I was not sure if this omission was originally made on purpose... 

However, I think being able to save results locally instead of relying on referencing them by job id would be a benefit.  

There is some value in not requiring network connectivity and making them easily shareable/archive-able.

The necessary change to [databin.py: 96](https://github.com/Qiskit/qiskit/blob/65f709d21612bd38d5f8ae0c8d700892cbec9289/qiskit/primitives/containers/data_bin.py#L96) is :
```
    def __setattr__(self, name, value):
        super().__setattr__(name, value)
```

kt474:
### Summary

In `qiskit-ibm-runtime` we want to add job methods to our `QiskitRuntimeLocalService` (Issue https://github.com/Qiskit/qiskit-ibm-runtime/issues/1607). In order to save `PrimitiveJob` objects locally we need a way to serialize them. It's currently not serializable due to the use of `ThreadPoolExecutor`:

https://github.com/Qiskit/qiskit/blob/35d0954578d5312a63fb702f29bbd3d47ff25cc5/qiskit/primitives/primitive_job.py#L47-L48

Credit to @t-imamichi 
### Details and comments

Also fixes #12787


qiskit/primitives/containers/data_bin.py
```diff
@@ -93,9 +93,6 @@ def __init__(self, *, shape: ShapeInput = (), **data):
     def __len__(self):
         return len(self._data)
 
-    def __setattr__(self, *_):
-        raise NotImplementedError
-
```
kt474, line 108: Is it ok to remove this? This `__setattr__` is preventing the serialization of the job result

t-imamichi: Could you answer the question, @ihincks?

qiskit/primitives/primitive_job.py
```diff
@@ -27,14 +27,16 @@ class PrimitiveJob(BasePrimitiveJob[ResultT, JobStatus]):
     Primitive job class for the reference implementations of Primitives.
     """
 
-    def __init__(self, function, *args, **kwargs):
+    def __init__(self, *args, function=None, **kwargs):
```
t-imamichi: This does not work as expected because `function` is always included in `args` and `self._function` will be always `None`.

t-imamichi: Perhaps you want something like the following?
```suggestion
    def __init__(self, function=None, *args, **kwargs):
```

qiskit/primitives/containers/data_bin.py
t-imamichi: #12787 asked the same question.

qiskit/primitives/primitive_job.py
kt474: I had that originally but got a pylint error, I think we'll have to ignore `keyword-arg-before-vararg`

qiskit/primitives/containers/data_bin.py
ihincks: Please go ahead and remove it. Apologies, this completely fell off my radar.

t-imamichi:
It looks good overall. Could you add a reno to describe that both `DataBin` and `PrimitiveJob` become serializable?

t-imamichi (dismissed):
LGTM. Could you ask someone else to review it too just in case?

qiskit/primitives/primitive_job.py
```diff
@@ -27,14 +27,16 @@ class PrimitiveJob(BasePrimitiveJob[ResultT, JobStatus]):
     Primitive job class for the reference implementations of Primitives.
     """
 
-    def __init__(self, function, *args, **kwargs):
+    def __init__(self, function=None, *args, **kwargs):  # pylint: disable=keyword-arg-before-vararg
```
ihincks: What is the reason for setting function to None as default? I expect, for example, that `executor.submit` will fail with an error if the default is kept.

kt474: You're right, we don't actually need to default to `None` here, we only need to set `self._function = None` in `_prepare_dump` to allow serializing/deserializing. Otherwise, we'd run into `TypeError: cannot pickle '_thread.RLock' object`

Tested against my PR in qiskit-ibm-runtime and fixed in [a6d1314](https://github.com/Qiskit/qiskit/pull/12963/commits/a6d1314247cc4004e9a91777577822e998b40962)

qiskit/primitives/primitive_job.py
```diff
@@ -46,19 +48,30 @@ def _submit(self):
         self._future = executor.submit(self._function, *self._args, **self._kwargs)
         executor.shutdown(wait=False)
 
+    def _prepare_dump(self):
```
t-imamichi: I recall that this method is used when we serialize a job. But, this method is not used any other methods of this class. So, it might be good to add a comment just in case though it is a private method.

kt474: Added in https://github.com/Qiskit/qiskit/pull/12963/commits/fac959a4f7deaf6375d23f8293c92fe2119ecdd6

qiskit/primitives/primitive_job.py
```diff
@@ -46,19 +48,30 @@ def _submit(self):
         self._future = executor.submit(self._function, *self._args, **self._kwargs)
         executor.shutdown(wait=False)
 
+    def _prepare_dump(self):  # This method allows PrimitiveJob to be serialized
```
t-imamichi:
```suggestion
    def _prepare_dump(self):
    """This method allows PrimitiveJob to be serialized"""
```

releasenotes/notes/serialize-primitive-job-aa97d0bf8221ea99.yaml
```diff
@@ -0,0 +1,5 @@
+---
+features_primitives:
+  - |
+    To make :class:`.PrimitiveJob` serializable, `qiskit.primitives.containers.DataBin` has been
```
t-imamichi:
```suggestion
    To make :class:`.PrimitiveJob` serializable, ``qiskit.primitives.containers.DataBin`` has been 
```

t-imamichi: After #13564 is merged, you can use ":class:\`.DataBin\`"

t-imamichi (dismissed):
LGTM. Could you review it, @ihincks?

ihincks (dismissed):

qiskit/primitives/primitive_job.py
```diff
@@ -46,19 +48,31 @@ def _submit(self):
         self._future = executor.submit(self._function, *self._args, **self._kwargs)
         executor.shutdown(wait=False)
 
+    def _prepare_dump(self):
+        """This method allows PrimitiveJob to be serialized"""
+        _ = self.result()
+        _ = self.status()
+        self._future = None
+
```
raynelfss: When is this function called during serialization? I can't see anywhere where the method is called in the PR itself?

raynelfss (commented):
The removal of the unimplemented `__setattr__` method makes sense to me, but I have a question about the `_prepare_dump` method.

qiskit/primitives/primitive_job.py
kt474: It would be called by the user if they want to serialize `PrimitiveJob` - example [here](https://github.com/Qiskit/qiskit-ibm-runtime/pull/1854/files#diff-80a3263989cc232c76980d9d9cae1a5e9e3216af4d76f8224b6cbd7c261adc07R311)

raynelfss: Would it be better to make this happen in one of the serialization methods that python has built-in? I believe this could be executed on one of the methods, avoiding that function call. You could include the lines here and override one of the reserved methods for serialization: `__reduce__`, `__getstate__`, `__getnewargs__`, depending on your needs.

Note: This might require you to also implement a `__setstate__` method.

t-imamichi: Thank you for your suggestion. I think the following should work (to remove non-pickable object `_future`) by replacing `_prepare_dump` with the following.

```python
    def __getstate__(self):
        _ = self.result()
        _ = self.status()
        state = self.__dict__.copy()
        state["_future"] = None
        return state

    def __setstate__(self, state):
        self.__dict__.update(state)
        self._future = None
```

t-imamichi: I made a patch to this PR https://github.com/kt474/qiskit-terra/pull/1.

raynelfss (dismissed):
This looks much better! Thank you for these additions @kt474, @t-imamichi 🚀

raynelfss (dismissed):

releasenotes/notes/serialize-primitive-job-aa97d0bf8221ea99.yaml
```diff
@@ -0,0 +1,5 @@
+---
+features_primitives:
+  - |
+    To make :class:`.PrimitiveJob` serializable, ``qiskit.primitives.containers.DataBin`` has been
```
t-imamichi: I suggest the following because `DataBin` is public now.
https://docs.quantum.ibm.com/api/qiskit/qiskit.primitives.DataBin
```suggestion
    To make :class:`.PrimitiveJob` serializable, :class:`.DataBin` has been 
```

test/python/primitives/test_primitive_job.py
```diff
@@ -0,0 +1,51 @@
...
+        job2 = pickle.loads(obj)
+        self.assertIsInstance(job2, PrimitiveJob)
+        self.assertEqual(job.job_id(), job2.job_id())
+        self.assertEqual(job.status(), job2.status())
+        self.assertEqual(job.metadata, job2.metadata)
+        result = job.result()
+        result2 = job2.result()
+        self.assertEqual(result.metadata, result2.metadata)
+        self.assertEqual(len(result), len(result2))
+        for _, sampler_pub in enumerate(result):
```
t-imamichi: Sorry. This is my fault. This line should have been simple.
```suggestion
        for sampler_pub in result:
```

t-imamichi (approved):

raynelfss (approved):

PR #12968, opened 2024-08-16
Operator equality with (clean) ancillas

alexanderivrii:
### Summary

This PR adds a new method ``equal_with_ancillas`` to the ``Operator`` class that checks whether the two operators are equal in the subspace where the ancilla qubits are zero.

While this is not immediately required for #12961 and #12911, it allows to check correctness of various synthesis methods and transformations in the presence of clean ancilla qubits. (The dirty ancilla qubits should be interpreted as regular qubits and don't matter).

### Details and comments

Many thanks to @ShellyGarion for the idea how to implement this in a simple way.

Here is an example (which also appears in the release notes):
```
qc1 = QuantumCircuit(4)
qc1.x(0)
qc1.x(2)
qc1.cx(1, 0)
qc1.cx(1, 2)
qc1.cx(1, 3)
op1 = Operator(qc1)

qc2 = QuantumCircuit(4)
qc2.x(0)
qc2.x(2)
op2 = Operator(qc2)

assert op1 != op2
assert op1.equal_with_ancillas(op2, [1])
```

The first assertion checks that the two circuits are not equal, and they are not -- due to the CX gates from qubit 1 to other qubits. However, the two circuits are equal if qubit 1 is initially zero, then the CX gates do not matter.


alexanderivrii:
As @Cryoris suggested offline, it is better to make this function private for now. So I have prefixed it with the underscore and removed the lovingly-created release notes.

qiskit/quantum_info/operators/operator.py
```diff
@@ -615,6 +615,62 @@ def _multiply(self, other):
...
+    ) -> bool:
+        r"""Test if two Operators are equal on the subspace where ancilla qubits
+        are :math:`|0\rangle`.
+
+        Args:
+            other (Operator): an operator object.
+            ancilla_qubits: a list of clean ancilla qubits.
+            rtol (float): relative tolerance value for comparison.
+            atol (float): absolute tolerance value for comparison.
+            ignore_phase (bool): ignore complex-phase difference between matrices.
```
eliarbel: Change the order to match the actual args order

qiskit/quantum_info/operators/operator.py
```diff
@@ -615,6 +615,62 @@ def _multiply(self, other):
...
+
+        Returns:
+            bool: True if operators are equal up to clean ancilla qubits.
```
eliarbel:
```suggestion
            bool: True iff operators are equal up to clean ancilla qubits.
```

qiskit/quantum_info/operators/operator.py
```diff
@@ -615,6 +615,62 @@ def _multiply(self, other):
         ret._data = other * self._data
         return ret
 
+    def _equal_with_ancillas(
+        self,
+        other: Operator,
+        ancilla_qubits: list,
```
eliarbel: Maybe List[int] instead of list?

qiskit/quantum_info/operators/operator.py
```diff
@@ -615,6 +615,62 @@ def _multiply(self, other):
...
+            other (Operator): an operator object.
+            ancilla_qubits: a list of clean ancilla qubits.
+            rtol (float): relative tolerance value for comparison.
+            atol (float): absolute tolerance value for comparison.
+            ignore_phase (bool): ignore complex-phase difference between matrices.
+
+        Returns:
+            bool: True if operators are equal up to clean ancilla qubits.
+        """
+        if self.dim != other.dim:
```
eliarbel: Add type checking on `other` similarly to how it's done in `equiv()`

qiskit/quantum_info/operators/operator.py
```diff
@@ -615,6 +615,62 @@ def _multiply(self, other):
...
+        pattern = [0] * num_qubits
+        pos = 0
+        for q in range(num_qubits):
+            if q not in ancilla_qubits:
+                pattern[pos] = q
+                pos += 1
+        for q in range(num_qubits):
+            if q in ancilla_qubits:
+                pattern[pos] = q
+                pos += 1
```
eliarbel: Might be a nitpicking, but I disliked the loop duplication up to the if condition ... 
So to alleviate this and hopefully also improve readability, consider something like this: 

```python
        pattern = []
        ancillas = []
        for q in range(num_qubits):
            if q in ancilla_qubits:
                ancillas.append(q)
            else:
                pattern.append(q)
        pattern += ancillas
```

eliarbel (changes requested):

alexanderivrii:
Thanks @Cryoris and @eliarbel for the reviews! 

I haven't fully understood Julien's comment the first time: as the new function is not yet used anywhere in the main Qiskit code, for now it's better to make it a private function and not have it in `operator.py`. This is now done in [4de623a](https://github.com/Qiskit/qiskit/pull/12968/commits/4de623a146e16e451e1afadd6f4ef12e27d78402). I have created a special file `operator_utils.py` for this function since no other place seemed suitable: the `utils` directory contains some high-level utilities (commutator, anti-commutator), the `predicates.py` is used by the `Operator` class while the new function uses the `Operator` class, etc. 

Eli, I have addressed your review comments in [4de623a](https://github.com/Qiskit/qiskit/pull/12968/commits/4de623a146e16e451e1afadd6f4ef12e27d78402).

qiskit/quantum_info/operators/operator.py
alexanderivrii: Good point, done in [4de623a](https://github.com/Qiskit/qiskit/pull/12968/commits/4de623a146e16e451e1afadd6f4ef12e27d78402).

qiskit/quantum_info/operators/operator.py
alexanderivrii: Done in [4de623a](https://github.com/Qiskit/qiskit/pull/12968/commits/4de623a146e16e451e1afadd6f4ef12e27d78402).

qiskit/quantum_info/operators/operator.py
alexanderivrii: Done in [4de623a](https://github.com/Qiskit/qiskit/pull/12968/commits/4de623a146e16e451e1afadd6f4ef12e27d78402) (well, `list[int]`, which I believe is now a bit more preferred).

qiskit/quantum_info/operators/operator.py
alexanderivrii: The `equiv` code does more than just type-checking: if converts `other` to an `Operator` when it's not already, so for example one can check the equivalence of an `Operator` and a `QuantumCircuit`. I don't think this fully makes sense for this (currently private) function.

qiskit/quantum_info/operators/operator.py
alexanderivrii: That's a great suggestion, done in [4de623a](https://github.com/Qiskit/qiskit/pull/12968/commits/4de623a146e16e451e1afadd6f4ef12e27d78402).

eliarbel (approved):
LGTM

I approve for merge, though I'm not 100% happy with adding a new file ('operator_utils.py') just for this, knowing that there is a chance that '_equiv_with_ancillas' would become public and even part of Operator. Since there are already _ methods in Operator, roughly in the same vein like `_multiply` and `_add` I would prefer to keep '_equiv_with_ancillas' in Operator and spare an additional new file in the module.

Cryoris:
`_multiply` and `_add` are used in overloading the `__rmult__` and `__add__` operations on the public interface, but the ancilla equality wouldn't be connected to a user-facing API 🙂

Cryoris (approved):

PR #12973, opened 2024-08-16
Add `DepthHeuristic` to SABRE routing heuristics

henryzou50:
### Summary

This pull request introduces a new heuristic, `DepthHeuristic`, to the `SabreSwap`. The `DepthHeuristic` is designed to minimize circuit depth by penalizing swaps that increase the circuit's depth during the routing process. This enhancement provides users with additional flexibility to prioritize depth optimization, complementing the existing heuristics in SABRE.

**Note:** This pull request is currently a draft. I am still working on refining the depth calculation to ensure it's "truly correct," and additional enhancements will be added in future commits.

#### Key Changes:

1. **New Heuristic (`DepthHeuristic`) in `heuristic.rs`:**
   - Added the `DepthHeuristic` struct, similar to `BasicHeuristic` and `DecayHeuristic`.
   - The `DepthHeuristic` includes `weight` and `scale` parameters, allowing users to adjust its impact on routing decisions.
   - Unlike `LookaheadHeuristic` and `DecayHeuristic`, this heuristic does not require additional parameters like `size` or `reset`. It is solely focused on depth-based penalties.

2. **Depth Tracking in `route.rs`:**
   - Introduced `qubit_depths` in `RoutingState` to track the depth of each qubit as routing progresses. It is only updated if the pass is ran with the depth heuristic. 
   - Implemented a `circuit_depth` method to calculate the depth of the circuit as we progress through the routing, and considers the additional swaps as 3 cnots. 
   - Integrated the `DepthHeuristic` into the `choose_best_swap` method to penalize swaps that increase circuit depth.
   - Updated the `update_route` method to correctly adjust `qubit_depths` when applying swaps and gates.

3. **Updates to `sabre_swap.py` and `test_sabre_swap.py`:**
   - Integrated `DepthHeuristic` into the SABRE transpiler, making it available for use alongside other heuristics.
   - Added tests for `DepthHeuristic` to ensure it functions correctly within the SABRE routing framework.

#### Future Considerations:

- **Depth Calculation Refinement:**
Currently, the depth calculations are based on whether a swap adds depth to the circuit but do not consider the impact of immediate gates that can be applied after the swap. This refinement is critical because prioritizing swaps that enable multiple immediate gates without significant depth increase can improve routing efficiency. I have these changes in a separate branch and will be added once the corresponding tests are validated and passing.

- **Upcoming `CriticalHeuristic`:** 
I am also working on a new `CriticalHeuristic` in another branch and will make a seperate PR for that once that passes all the tests. This heuristic will focus on the critical path of the DAG, prioritizing gates that are part of this path.

- **Dynamic Scaling of Depth Impact:**
We may want to think about having future iterations may explore dynamically scaling the depth penalty based on other heuristics or the size of the circuit, providing more nuanced control over depth optimization during routing.

- **Automated Depth Validation:**
While manual tests were conducted to validate `circuit_depth`, an automated test would be beneficial. Specifically, this test should compare `circuit_depth` in `RoutingState` with the depth of a decomposed circuit in Python. At the moment, I could not think of an efficient way that matches qiskit's testing scheme to compare the result of a function in rust vs in python. The test should also address scenarios involving initial DAG circuits with swap gates, which are currently not decomposed into three swaps in the Rust implementation.


crates/accelerate/src/sabre/route.rs
```diff
@@ -383,13 +418,33 @@ impl<'a, 'b> RoutingState<'a, 'b> {
             }
         }
 
+        if let Some(DepthHeuristic { .. }) = self.heuristic.depth {
+            let curr_depth = self.circuit_depth();
+
```
alexanderivrii: This is a really cool development! I have two minor questions:

* Would it be possible to update the current maximum depth as gates are routed; possibly this would be a bit faster than computing the maximum value each time?

* The current computation prioritizes SWAPs that do not increase the current maximum depth to those that do. In the case that multiple swaps-without-depth-increase are available, would it make sense to further prioritize SWAPs that connect a pair of qubits with lowest depths?

henryzou50: Thank you for the feedback—I'm glad you found the development interesting!

1. Yes, updating the current maximum depth as gates are routed is definitely possible and a great idea. It would be faster than computing the maximum value each time. I'll implement this improvement—thanks for the suggestion!

2. It makes sense to prioritize further SWAPs that connect a pair of qubits with the lowest depths, and often multiple swaps without depth increase are available. My current method addresses this is to calculating the "true depth" after inserting a SWAP. This involves determining the immediate gates that can be executed following the swap and then computing the new depth based on those gates. I’ve implemented this in my Python version, where I executed the experiments, and am currently working on integrating it into the Rust side.

crates/accelerate/src/sabre/heuristic.rs
```diff
@@ -165,6 +165,45 @@ impl DecayHeuristic {
...
 }
 
+/// Define the characteristics of the "depth" heuristic. This is used to penalize swaps that increase the circuit depth.
+#[pyclass]
+#[pyo3(module = "qiskit._accelerate.sabre", frozen)]
+#[derive(Clone, Copy, PartialEq)]
+pub struct DepthHeuristic {
+    /// The relative weighting of this heuristic to others.  Typically you should just set this to
+    /// 1.0 and define everything else in terms of this.
+    pub weight: f64,
```
jakelishman: I think this comment is copied from `BasicHeuristic` - this weight probably _shouldn't_ be 1, right?

crates/accelerate/src/sabre/heuristic.rs
```diff
@@ -258,6 +301,15 @@ impl Heuristic {
...
     }
 
+    /// Set the multiplier increment and reset interval of the depth heuristic.  The reset interval
+    /// must be non-zero.
+    pub fn with_depth(&self, weight: f64, scale: SetScaling) -> Self {
+        Self {
+            depth: Some(DepthHeuristic { weight, scale }),
+            ..self.clone()
+        }
+    }
```
jakelishman: The comment here looks off too.

crates/accelerate/src/sabre/layer.rs
```diff
@@ -26,9 +26,10 @@ use crate::nlayout::PhysicalQubit;
 /// makes it more efficient to do everything in terms of physical qubits, so the conversion between
 /// physical and virtual qubits via the layout happens once per inserted swap and on layer
 /// extension, not for every swap trialled.
+#[derive(Clone)]
 pub struct FrontLayer {
     /// Map of the (index to the) node to the qubits it acts on.
-    nodes: IndexMap<NodeIndex, [PhysicalQubit; 2], ::ahash::RandomState>,
+    pub nodes: IndexMap<NodeIndex, [PhysicalQubit; 2], ::ahash::RandomState>,
```
jakelishman: I don't think this needs to become `pub`, and it's probably safer that it doesn't, since you can't mutate this correctly unless you're also mutating `qubits`, and it's a bit tricksy to do that correctly.

crates/accelerate/src/sabre/route.rs
```diff
@@ -90,6 +93,13 @@ impl<'a, 'b> RoutingState<'a, 'b> {
         self.front_layer.apply_swap(swap);
         self.extended_set.apply_swap(swap);
         self.layout.swap_physical(swap[0], swap[1]);
+        if self.heuristic.depth.is_some() {
+            // Add 3 since swaps can be decomposed into 3 CNOTs
+            let new_depth =
+                self.qubit_depths[swap[0].index()].max(self.qubit_depths[swap[1].index()]) + 3.0;
+            self.qubit_depths[swap[0].index()] = new_depth;
+            self.qubit_depths[swap[1].index()] = new_depth;
+        }
```
jakelishman: This is definitely making assumptions about how the swap will actually be synthesised (what if it's an iswap-based backend?) and what the cost of the 2q gate we're routing is (what if it's something like an Rxx that we'll decompose later?).  I'm fine leaving this assumption in for the time being, since we don't currently pass any other synthesis-related information to Sabre, but it'd be good to update the comment to spell out more about the assumptions going into it.

crates/accelerate/src/sabre/route.rs
```diff
@@ -103,6 +113,14 @@ impl<'a, 'b> RoutingState<'a, 'b> {
         })
     }
 
+    fn circuit_depth(&self) -> f64 {
```
jakelishman, line 124: Can you add a docstring comment to this explaining what the input assumptions are, and what precisely the output number means?  From the name alone, it's not necessarily clear that the number will vary over the course of the routing.

crates/accelerate/src/sabre/route.rs
```diff
@@ -171,6 +189,18 @@ impl<'a, 'b> RoutingState<'a, 'b> {
...
+            // Update the qubit depths for the gates in the dag
+            if self.heuristic.depth.is_some() {
+                if let [a, b] = node.qubits[..] {
+                    let qubit_a = a.to_phys(&self.layout).index();
+                    let qubit_b = b.to_phys(&self.layout).index();
+                    let new_depth =
+                        self.qubit_depths[qubit_a].max(self.qubit_depths[qubit_b]) + 1.0;
+                    self.qubit_depths[qubit_a] = new_depth;
+                    self.qubit_depths[qubit_b] = new_depth;
+                }
```
jakelishman, line 209: This applies to directive nodes as well, like a 2q barrier.  Barrier definitely synchronises the depths of the qubits that follow it, but we perhaps shouldn't have the depths _increase_ because of a directive node (so the `+ 1.0` might not be right?).

crates/accelerate/src/sabre/route.rs
```diff
@@ -383,13 +413,144 @@ impl<'a, 'b> RoutingState<'a, 'b> {
...
+                SetScaling::Size => {
+                    if self.front_layer.is_empty() {
+                        0.0
+                    } else {
+                        weight / (self.front_layer.len() as f64)
+                    }
+                }
+            };
+            let orig_depth = self.circuit_depth();
+
```
jakelishman: This `if let` block ends up being very long compared to all the rest of the function.  As a minor stylistic point, perhaps we could factor much of the body of it out into a separate method?

crates/accelerate/src/sabre/route.rs
```diff
@@ -383,13 +413,144 @@ impl<'a, 'b> RoutingState<'a, 'b> {
...
+            // Calculate the change in circuit depth for each swap
+            for (swap, score) in self.swap_scores.iter_mut() {
+                // Update the qubit depths after directly applying the swap trial, treating swaps as 3 CNOTs
+                let mut trial_qubit_depths = self.qubit_depths.to_vec();
+                let new_depth = trial_qubit_depths[swap[0].index()]
+                    .max(trial_qubit_depths[swap[1].index()])
+                    + 3.0;
```
jakelishman: Since the magic number 3 has turned up again here (commented on above too), I think it would be good to centralise the magic somewhere. Maybe it's just in the `Heuristic` itself in a way that we don't expose to Python space to change, since we don't know how we'll want to modify how we pass in synthesis-related information in the future, but I'm very suspicious of a magic-number assumption appearing in more than one place - way easy to get out of sync / mistaken in the future.

crates/accelerate/src/sabre/route.rs
```diff
@@ -383,13 +413,144 @@ impl<'a, 'b> RoutingState<'a, 'b> {
...
+                        weight / (self.front_layer.len() as f64)
+                    }
+                }
+            };
+            let orig_depth = self.circuit_depth();
+
+            // Calculate the change in circuit depth for each swap
+            for (swap, score) in self.swap_scores.iter_mut() {
+                // Update the qubit depths after directly applying the swap trial, treating swaps as 3 CNOTs
+                let mut trial_qubit_depths = self.qubit_depths.to_vec();
```
jakelishman, line 464: I'm totally fine leaving this for the first implementation of the PR, but the $\mathcal O(\text{num qubits})$ clone that occurs on every swap, even though the scope of the modifications should be relatively modest, feels like a place to investigate performance improvements.

crates/accelerate/src/sabre/route.rs
```diff
@@ -548,6 +710,7 @@ pub fn swap_map_trial(
 
     while !state.front_layer.is_empty() {
         let mut current_swaps: Vec<[PhysicalQubit; 2]> = Vec::new();
+        let qubit_depths_backup = state.qubit_depths.to_vec();
```
jakelishman: This is an all-qubits clone for each routed gate, which isn't ideal - it makes us pay a cost for the release valve, even if we don't take it, and when we added the release valve, an important goal of it was to not increase the cost of the happy path.

That said, it's not the end of the world if we need to do it, since at least it's not for each swap.  Some potential strategies for mitigating a bit of the cost:
* only calculate the backup if there's a depth component in the heuristic
* allocate the space for the backup only once at the start of the function, and re-use it rather than reallocating it each time.

My main concern here is to not affect the runtime of Sabre for heuristics that _don't_ have a depth component - we definitely don't need to try and optimise the depth performance right now, but it is important not to have it regress the non-depth cases.

crates/accelerate/src/sabre/route.rs
```diff
@@ -575,6 +738,8 @@ pub fn swap_map_trial(
             // If we exceeded the max number of heuristic-chosen swaps without making progress,
             // unwind to the last progress point and greedily swap to bring a ndoe together.
             // Efficiency doesn't matter much; this path never gets taken unless we're unlucky.
+            // Reset the qubit depths to the backup
+            state.qubit_depths.copy_from_slice(&qubit_depths_backup);
```
jakelishman: This also might want to be protected in a `if self.heuristic.depth.is_some()` block.

crates/accelerate/src/sabre/route.rs
```diff
@@ -383,13 +413,144 @@ impl<'a, 'b> RoutingState<'a, 'b> {
...
+                // Calculate the final maximum depth
+                let final_depth = *trial_qubit_depths
+                    .iter()
+                    .max_by(|a, b| a.partial_cmp(b).unwrap())
+                    .unwrap();
+                let depth_increase = final_depth - orig_depth;
+                // Divide the increase in depth by 3 for how 1 swap is equivalent to 3 CNOTs
+                let depth_score = depth_increase / 3.0;
+
+                *score += weight * (self.front_layer.score(*swap, dist) + depth_score);
```
jakelishman: `self.front_layer.score(*swap, dist)` is already a part of the `BasicHeuristic`.  Putting here seems like it's just adding extra weight to that part of the calculation? I feel like we only need to add on `weight * depth_score` here, but maybe I'm mistaken?

releasenotes/notes/sabre-heuristic-depth-92a6f1f4922478dc.yaml
```diff
@@ -0,0 +1,21 @@
+---
+prelude: >
+    A new heuristic, `DepthHeuristic`, has been added to the `SabreSwap` pass
+    in Qiskit. This heuristic allows users to prioritize swaps that do not
+    increase the overall circuit depth, potentially leading to a more optimized
+    circuit with a much lower depth.
```
jakelishman: We don't put `prelude` components in single-PR release notes, since determining the prelude is the responsibility of the commit that finalises the release.

releasenotes/notes/sabre-heuristic-depth-92a6f1f4922478dc.yaml
```diff
@@ -0,0 +1,21 @@
...
+    like "basic," "lookahead," or "decay," but with a focus on minimizing the circuit 
+    depth. By penalizing swaps that would increase the circuit depth, this feature 
+    helps in routing quantum circuits more efficiently with a potentially lower depth.
+
+    To use this feature, you can add `DepthHeuristic` to the set of heuristics used 
+    in `SabreSwap`, just as you would with "basic," "lookahead," or "decay." For example:
+
+    This enhancement gives users the flexibility to choose the heuristic best suited for 
+    their circuit, helping to improve performance by reducing the overall circuit depth 
+    when necessary.
```
jakelishman: Couple of points:
* things like `DepthHeuristic` and `SabreSwap` should be cross-references if there's something in the public documentation to cross-reference too (e.g. `` :class:`.SabreSwap` ``), and we shouldn't use the name at all if it's not public.
* the example alluded to by the second paragraph is missing
* things like `"basic"`, `"lookahead"` etc are referring to Python literal strings - we can't put punctuation inside those strings without it invalidating them, so it should be

  > This heuristic works similarly to the existing heuristics like "basic", "lookahead" or "decay", [...]
  
  or something like that.  (Basically, the British way of writing quotes haha - for us, you can't move punctuation into quotes like that.)

test/python/transpiler/test_sabre_swap.py
jakelishman, line 1: I know we've not got a huge amount of explicit tests of the other heuristics, but it really would be good to have at least a couple of tests that the output is valid, such as producing an equivalent circuit that only inserts swaps on qubits with a hardware link.

One way to do that might be to route a circuit from the trivial initial layout, then
* use `ElidePermutations` to ensure that the output circuit is equivalent to the input - I haven't worked that through exactly, but `ElidePermutations` should effectively be the "inverse" operation of routing
* use `CheckMap` to validate that each operation (including swaps) is only on valid hardware links.

qiskit/transpiler/passes/routing/sabre_swap.py
```diff
@@ -225,6 +225,12 @@ def run(self, dag):
                 .with_lookahead(0.5, 20, SetScaling.Size)
                 .with_decay(0.001, 5)
             )
+        elif self.heuristic == "depth":
+            heuristic = (
+                Heuristic(attempt_limit=10 * num_dag_qubits)
+                .with_basic(1.0, SetScaling.Size)
+                .with_depth(0.5, SetScaling.Size)
+            )
```
jakelishman: If we're exposing this publicly, we need to list the `depth` option in the docstring, and we probably need to make a similar modification to `SabreLayout`.

jakelishman (commented):
Thanks for this Henry, it's definitely an interesting research direction for the heuristic.  Do you have any case studies where you can show that this new heuristic produces significantly improved output over `lookahead`?  That's something that might be interesting to demonstrate in the release note, and potentially as a test case if we can prove that the improvement is really _inherent_ to the depth heuristic compared to the lookahead one (and not just an RNG thing, etc).

crates/accelerate/src/sabre/route.rs
jakelishman: Also, this code is very heavily duplicated from the main loop of the core Sabre loop.  I'm fine leaving it like this for its first introduction, but we should definitely consider what refactors we might make to avoid the duplication, since it's relatively complex code with a lot of different methods inlined into one place.

crates/accelerate/src/sabre/heuristic.rs
henryzou50: Yes, you are correct, the weight should not be 1. I've adjusted it so that the weight is now less than 1, relative to the 'basic' heuristic. Specifically, if the basic heuristic has a weight of 1, the depth heuristic should default to 1/3. This reflects that a score of 3 in the depth heuristic corresponds to a depth increase of 3, which is equivalent to 1 swap (or 3 CNOT gates), corresponding to a distance of 1.

crates/accelerate/src/sabre/heuristic.rs
henryzou50: Fixed

crates/accelerate/src/sabre/layer.rs
henryzou50: Yes, you are right and that makes sense, this does not need to become `pub` and I adjusted that.

crates/accelerate/src/sabre/route.rs
henryzou50: Good point! I agree it’s fine to leave it as is for now, as you mentioned, with an update planned for later. To keep things consistent, though, I’ve added `swap_depth` to the `DepthHeuristic`, which represents the estimated depth increase per added swap during routing.

crates/accelerate/src/sabre/route.rs
henryzou50: Added and thanks!

crates/accelerate/src/sabre/route.rs
henryzou50: I agree, I'll leave it as is for now, but we should definitely plan some refactoring to avoid duplication in the future.

crates/accelerate/src/sabre/route.rs
henryzou50: Good point, and fixed for now! Resolved by the `swap_depth` in the `DepthHeuristic` above.

crates/accelerate/src/sabre/route.rs
henryzou50: I agree with leaving it as is for now, but I share the concern about the performance impact of the `DepthHeuristic`, especially in terms of runtime compared to other heuristics at a larger scale. This clone on every swap is indeed one of the primary contributors to the performance overhead.

crates/accelerate/src/sabre/route.rs
henryzou50: Nice catch and done!

crates/accelerate/src/sabre/route.rs
henryzou50: Fixed!

crates/accelerate/src/sabre/route.rs
henryzou50: You are right, we just need the `weight * depth_score` here, and I also adjusted it so that we don't divide by 3 here, and that part is just involved in the `weight`.

releasenotes/notes/sabre-heuristic-depth-92a6f1f4922478dc.yaml
henryzou50: Thanks and fixed!

releasenotes/notes/sabre-heuristic-depth-92a6f1f4922478dc.yaml
henryzou50: Great points, and I just applied them!

test/python/transpiler/test_sabre_swap.py
henryzou50: Yes, those are important tests to have, and I just added them on the latest commit with the added tests: 
- `test_do_not_reorder_measurements`
- `test_elide_permutations_equivalence`

qiskit/transpiler/passes/routing/sabre_swap.py
henryzou50: Yes, you're right, I’ve added the depth option to the docstring and corrected a typo there.

`SabreLayout` doesn’t mention the heuristic in its docstring, so I don’t think a similar modification is needed for it at this point. This does remind me, though, that in the future, we might want to add an option for the heuristic in `SabreLayout`. Currently, it defaults to 'decay', which could be limiting when we want to run `SabreLayout` without `SabreSwap`.

crates/accelerate/src/sabre/route.rs
henryzou50: Ohh I see, that makes sense. Since directive nodes, like barriers, don’t correspond to a quantum operation in the way a gate does, they wouldn’t impact the circuit depth. I’m not sure of the best way to check if a node is a directive node or not. Do you have any suggestions on how we might identify these nodes effectively?

henryzou50:
Thank you for the review! I’ve made the changes to address your comments.

The most consistent case studies where this new heuristic shows significant improvement over lookahead are circuits with sequentially dependent interaction chains, like QFT and GHZ. I’ve added a test case to demonstrate this improvement, though I’m not entirely sure how to conclusively prove the improvement is inherent to the depth heuristic rather than due to random variability.

From my testing, for large circuit sizes and coupling maps with higher connectivity than a ring, the depth heuristic consistently yields better depth reductions by leveraging parallel swaps, although it does lead to an increase in the number of gates added. To ensure thoroughness, the test I added runs multiple trials and compares the median results. However, in practice, a single trial often shows this advantage clearly.

crates/accelerate/src/sabre/route.rs
mtreinish: There is a `directive` field in the `DAGNode` struct used by `SabreDAG`: https://github.com/Qiskit/qiskit/blob/main/crates/accelerate/src/sabre/sabre_dag.rs#L26 that should get set if the python object is a directive. If that's true you can ignore the node.

qiskit/transpiler/passes/routing/sabre_swap.py
mtreinish: Hmm, we probably should expose the swap heuristic as an option on `SabreLayout` too. But we can do this in a follow up PR since it's a new API change and probably out of scope for this PR.

jakelishman:
I'll close this as stale, but we might want to revisit in the future

PR #12975, opened 2024-08-18
Implement a capacity-allocated constructor for `DAGCircuit` in Rust.

### LINKED ISSUE #13002: Create a ``DAGCircuit::with_capacity`` method in Rust
### What should we add?

Tracked by #13001

This method should allow us to create a pre-allocated `DAGCircuit` avoiding most of the overhead of sequentially adding instructions to the `DAGCircuit` and re-allocating memory on every addition.

raynelfss:
### Summary
Resolves #13002
After #12550 merges, the following commits aim to provide a`with_capacity` alternative initializer for the `DAGCircuit` in Rust.

### Details and comments
**Will be rebased after #12550 merges**

These commits add the `DAGCircuit::with_capacity` constructor to the rust representation of `DAGCircuit`. This would allow us to allocate an approximation of the necessary space to successfully create a `DAGCircuit` with certain properties to avoid re-allocating space in memory.

This constructor takes the following arguments:
- `num_qubits` for the number of qubits in the specific circuit.
- `num_clbits` for the number of clbits in the circuit.
- `num_ops` for the number of operations in the circuit (Optional).
- `num_vars` for the number of variables in the circuit (Optional).

### Questions
- What is the best way of estimating the possible number of edges required for `num_ops`?
   - An operation (or a block) can have as many input/output qargs/cargs as possible
   - I used 3 (2 qubits + 1 classical or 3 qubits) as a baseline. But this is just speculation, if a better approach exists, let me in a comment.
 - Best way of estimating the names of operations in a circuit?
    - In a `DAGCircuit`, there can be 100 `op_nodes` of just the same operation or 100 different operations occurring just once.
    - I initialized it as `num_ops` but if a different opinion exists, let me know in a comment.

### Blockers
- [x] #12550


mtreinish:
https://github.com/Qiskit/qiskit/pull/12812 might be useful for you on this. This was me adding basically the same interface to the python space constructor for `DAGCircuit`. We get more flexibility with it in Rust including for things like the internal caches so I probably will rebase that PR on top of this after this merges (and add a python interface for the alternative constructor).

crates/circuit/src/dag_circuit.rs
```diff
@@ -1556,7 +1556,14 @@ def _format(operand):
     ///     DAGCircuit: An empty copy of self.
     #[pyo3(signature = (*, vars_mode="alike"))]
     fn copy_empty_like(&self, py: Python, vars_mode: &str) -> PyResult<Self> {
-        let mut target_dag = DAGCircuit::new(py)?;
+        let mut target_dag = DAGCircuit::with_capacity(
+            py,
+            self.num_qubits(),
+            self.num_clbits(),
+            Some(self.num_vars()),
+            None,
```
mtreinish, line 1564: Shouldn't this be:

```suggestion
            None,
            Some(self.num_vars()),
```

Otherwise we potentially will be overallocating if there are vars but no ops. Although it might make more sense to flip the arguments in the `with_capacity` signature as doing `qubits, clbits, vars, ops` makes more sense intuitively to me.

mtreinish (commented):
Overall this LGTM it's a straightforward addition to the rust api for `DAGCircuit`. Just one inline comment where I think one argument is in the wrong position.

crates/circuit/src/dag_circuit.rs
raynelfss: I'll re-arrange the function signature for this as it makes more sense, thank you 🚀

mtreinish (approved):
LGTM, thanks for the quick update

PR #12178, opened 2024-04-14
Replace initialization method by Isometry in StatePreparation

ShellyGarion:
### Summary

Close #12081
Replace initialization method by `Isometry` in `StatePreparation`

PR #12976, opened 2024-08-18
Fix Initialize.gates_to_uncompute method

### LINKED ISSUE #12969: Data preparation initializer.py is calling a method on StatePreparation that no longer exists
### Environment

- **Qiskit version**: 1.2
- **Python version**: [3.8, 3.12]
- **Operating system**: Linux, Windows, Mac

### What is happening?

Given 1.2 just released I quickly looked around the application modules just as a quick check all was good (algorithms runs tests against qiskit main so as to get some heads up, but the apps just run/test off stable release) - Qiskit Finance nightly CI tests failed in several of them due to this:
```
      File "/home/runner/work/qiskit-finance/qiskit-finance/qiskit_finance/circuit/library/probability_distributions/normal.py", line 218, in __init__
    circuit = initialize.gates_to_uncompute().inverse()

      File "/opt/hostedtoolcache/Python/3.10.14/x64/lib/python3.10/site-packages/qiskit/circuit/library/data_preparation/initializer.py", line 91, in gates_to_uncompute
    return self._stateprep._gates_to_uncompute()

    AttributeError: 'StatePreparation' object has no attribute '_gates_to_uncompute'
```    

https://github.com/Qiskit/qiskit/blob/713ab38cccb7e55a40a87372a2590a5a4430fa8f/qiskit/circuit/library/data_preparation/initializer.py#L91

Create an initializer Initialize instance and call its method `gates_to_uncompute`

### What should happen?

The `gates_to_uncompute` public method should work, as it used to, even if internal implementation is altered. 

The change, where the method that initializer was calling got removed, seems to have come in #12178 

### Any suggestions?

_No response_

ShellyGarion:
### Summary
close #12969 
Fix some code that has been removed in https://github.com/Qiskit/qiskit/pull/12178, so that the `Initialize.gates_to_uncompute` method will not fail

### Details and comments
`Initialize.gates_to_uncompute` is based on `StatePreparation._gates_to_uncompute()` that now contains the inverse of `Isomtery` in order to make it more efficient


qiskit/circuit/library/data_preparation/state_preparation.py
```diff
@@ -243,6 +243,20 @@ def validate_parameter(self, parameter):
     def _return_repeat(self, exponent: float) -> "Gate":
         return Gate(name=f"{self.name}*{exponent}", num_qubits=self.num_qubits, params=[])
 
+    ### helper functions for initializer.gates_to_uncompute() method ###
```
Cryoris: What do you think about just implementing this on `Initialize` directly? It's not used in this class and we might accidentally remove it in future again, so it would be safer to just have it in the API-documented method.

ShellyGarion: done in [b6f208a](https://github.com/Qiskit/qiskit/pull/12976/commits/b6f208acc968106349e4f5ef8398f343633db83a)

ShellyGarion: I also added a test that will hopefully prevent us from breaking the API in the future

ShellyGarion:
The CI failures should be fixed after https://github.com/Qiskit/qiskit/pull/12979 will be merged

qiskit/circuit/library/data_preparation/initializer.py
```diff
@@ -86,9 +87,14 @@ def gates_to_uncompute(self) -> QuantumCircuit:
...
 
         Returns:
-            Circuit to take ``self.params`` vector to :math:`|{00\\ldots0}\\rangle`
+            QuantumCircuit: circuit to take ``self.params`` vector to :math:`|{00\\ldots0}\\rangle`
         """
-        return self._stateprep._gates_to_uncompute()
+        q = QuantumRegister(self.num_qubits)
+        circuit = QuantumCircuit(q, name="disentangler")
+
+        isom = Isometry(self.params, 0, 0)
```
Cryoris, line 93: If I recall correctly (maybe we need to double check), `gates_to_uncompute` previously returned a circuit with only `Gate` objects -- now it would return a circuit with an `Instruction`. To fix this we could maybe do 
```python
circuit.compose(isom.definition.inverse(), copy=False, inplace=True)
```
or, the more performant version, just
```python
return isom._gates_to_uncompute()
```
?

Cryoris (commented):
This is very related to the comment below: `StatePreparation` is a `Gate` but currently contains an `Instruction` (namely an `Isometry`). We should update this such that a gate contains only other gates, e.g. by doing something like
```python
def _define_synthesis_isom(self):
        # ... previous code ...
        # we know Isometry produces only Gate objects in this case,
        # so we can directly compose it's definition
        isom = Isometry(self._params_arg, 0, 0)
        initialize_circuit.compose(isom.definition, inplace=True, copy=False)
```

qiskit/circuit/library/data_preparation/initializer.py
ShellyGarion: fixed in [4ae61c7](https://github.com/Qiskit/qiskit/pull/12976/commits/4ae61c78d2d44935076cc0ff84af3cbe731b912b)

Cryoris:
Let's wait for @woodsp-ibm to confirm this is good, but this LGTM -- thanks for the updates! 🙂 

Edit: I just saw https://github.com/Qiskit/qiskit/issues/12969#issuecomment-2296811479, so it's good to go 👍🏻

Cryoris (approved):

woodsp-ibm:
@Cryoris I had checked it yesterday and it was all good https://github.com/Qiskit/qiskit/issues/12969#issuecomment-2296811479 but I just rechecked it now and used both initializer and state_preparation files from here and it passes the finance tests so its all good.

woodsp-ibm (approved):
Passes the Finance tests now - thx!

Cryoris:
Thanks for double checking @woodsp-ibm! (and thanks for remembering the merge queue @alexanderivrii 😉)

PR #11463, opened 2023-12-28
Adding QFT gate to natively reason about Quantum  Fourier Transforms

alexanderivrii:
### Summary

A new "high-level" gate ``QFTGate`` allows to natively add QFTs on a quantum circuit and, in particular, to defer their synthesis to the transpiler. The gates will be synthesized by the ``HighLevelSynthesis`` transpiler pass, using one of several available plugins.

A new function ``synth_qft_full(num_qubits, do_swaps, approximation_degree, insert_barriers, inverse, name)`` allows to synthesize a QFT circuit with the given number of qubits. For now the implementation is taken directly from the ``_build`` method of the existing ``QFT`` (blueprint) circuit class. The name of the method reflects that the synthesized circuit requires full connectivity. Note that another QFT synthesis method has been recently added in #11236, with the synthesized circuit following only linear (aka nearest-neighbor) connectivity.

The two methods above are wrapped into HighLevelSynthesis plugins ``QFTSynthesisFull`` and ``QFTSynthesisLine`` respectively. The default synthesis is set to ``QFTSynthesisFull``, thus fully coinciding with the circuits produced using the existing ``QFT`` circuit class.

In a follow-up PR I am going to replace the construction of QFT circuits by QFT gates whenever possible, such as for example when constructing QFT-based adders and multipliers.

### Details and Comments

[...]

PR #12979, opened 2024-08-19
adding copy argument to QFTGate.__array__

alexanderivrii:
### Summary

Fixes numpy support for QFTGate: the QFT gate PR was developed in parallel with #11999 and did not get the required upgrade.

### DIFF
--- a/qiskit/circuit/library/basis_change/qft.py
+++ b/qiskit/circuit/library/basis_change/qft.py
@@ -315,8 +315,10 @@ class QFTGate(Gate):
         """
         super().__init__(name="qft", num_qubits=num_qubits, params=[])
 
-    def __array__(self, dtype=complex):
+    def __array__(self, dtype=complex, copy=None):
         """Return a numpy array for the QFTGate."""
+        if copy is False:
+            raise ValueError("unable to avoid copy while creating an array as requested")
         n = self.num_qubits
         nums = np.arange(2**n)
         outer = np.outer(nums, nums)


qiskit/circuit/library/basis_change/qft.py
```diff
@@ -315,8 +315,10 @@ def __init__(
         """
         super().__init__(name="qft", num_qubits=num_qubits, params=[])
 
-    def __array__(self, dtype=complex):
+    def __array__(self, dtype=complex, copy=None):
         """Return a numpy array for the QFTGate."""
+        if copy is False:
+            raise ValueError("unable to avoid copy while creating an array as requested")
```
ElePT, line 321: Tiny tiny suggestion:

```suggestion
            raise ValueError("Unable to avoid copy while creating an array as requested")
```

ElePT (approved):
Thanks a lot @alexanderivrii! I added a tiny suggestion but the PR can also be merged as-is, given that this bug is becoming annoying in the merge queue.

ElePT:
The QFT gate was added in 1.2, right? (https://github.com/Qiskit/qiskit/pull/11463) In that case we should also backport the fix.

qiskit/circuit/library/basis_change/qft.py
alexanderivrii: Thanks @ElePT! I agree that the capitalized "U" looks better, but all the other 30+ identical message use lower-case "u".

ElePT: hah, tunnel vision. Let's keep it consistent then.

PR #12982, opened 2024-08-19
Example code for Store instruction

ebaeumer:
### Summary
Created an example code for the Store instruction

image


qiskit/circuit/__init__.py
```diff
@@ -579,6 +579,16 @@
 :class:`Qubit` nor :class:`Clbit` operands, but has an explicit :attr:`~Store.lvalue` and
 :attr:`~Store.rvalue`.
 
+For example, to determine the parity of a bitstring cr and store it in another register creg,
+the :class:`Store` instruction can be used in the following way::
```
1ucian0:
```suggestion
For example, to determine the parity of a bitstring ``cr`` and store it in another register ``creg``,
the :class:`Store` instruction can be used in the following way::
```

1ucian0 (approved):
Thanks! LGTM.  Welcome to the contributor community Elisa!

PR #12985, opened 2024-08-19
Add code example in `Operator` docstring

jschuhmac:
### Summary
Adds code example in the docstring of the `Operator` class.

### Details and comments
Adds the following code example which applies an $X$ operator to the zero state:
code_example


1ucian0 (approved):
thanks!

PR #12986, opened 2024-08-19
Fix `UnitaryGate.repeat()` method

### LINKED ISSUE #11990: Invalid param type <class 'numpy.ndarray'> for gate unitary*1
### Environment

- **Qiskit version**: 1.0.2
- **Python version**: 3.11.8
- **Operating system**: Ubuntu 22.04.3 LTS

### What is happening?

Running the below code triggered a qiskit.circuit.exceptions.CircuitError: "Invalid param type <class 'numpy.ndarray'> for gate unitary*1." I found this bug happens for some other Gates as well. Below is a minimal reproduction example using CUGate.

```
from qiskit import QuantumCircuit, ClassicalRegister, QuantumRegister
from qiskit.circuit.library.standard_gates import CUGate

qr = QuantumRegister(2, name='qr')
cr = ClassicalRegister(2, name='cr')
qc = QuantumCircuit(qr, cr, name='qc')

qc.append(CUGate(0, 0, 0, 0).power(1).repeat(1), qargs=qr)
```

### What should happen?

```
  File "/mnt/tests/reproduction/reproduce_invalid_param_type.py", line 8, in <module>
    qc.append(CUGate(0, 0, 0, 0).power(1).repeat(1), qargs=qr)
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/root/anaconda3/envs/qiskit1.0.2/lib/python3.11/site-packages/qiskit/circuit/instruction.py", line 591, in repeat
    instruction = self._return_repeat(n)
                  ^^^^^^^^^^^^^^^^^^^^^^
  File "/root/anaconda3/envs/qiskit1.0.2/lib/python3.11/site-packages/qiskit/circuit/gate.py", line 87, in _return_repeat
    return Gate(name=f"{self.name}*{exponent}", num_qubits=self.num_qubits, params=self.params)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/root/anaconda3/envs/qiskit1.0.2/lib/python3.11/site-packages/qiskit/circuit/gate.py", line 46, in __init__
    super().__init__(name, num_qubits, 0, params, label=label, duration=duration, unit=unit)
  File "/root/anaconda3/envs/qiskit1.0.2/lib/python3.11/site-packages/qiskit/circuit/instruction.py", line 108, in __init__
    self.params = params  # must be at last (other properties may be required for validation)
    ^^^^^^^^^^^
  File "/root/anaconda3/envs/qiskit1.0.2/lib/python3.11/site-packages/qiskit/circuit/instruction.py", line 287, in params
    self._params.append(self.validate_parameter(single_param))
                        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/root/anaconda3/envs/qiskit1.0.2/lib/python3.11/site-packages/qiskit/circuit/gate.py", line 242, in validate_parameter
    raise CircuitError(f"Invalid param type {type(parameter)} for gate {self.name}.")
qiskit.circuit.exceptions.CircuitError: "Invalid param type <class 'numpy.ndarray'> for gate unitary*1."
```

### Any suggestions?

_No response_

timmintam:
### Summary
Closes #11990 : error when calling the `UnitaryGate.repeat()` method (inherited from the `Gate` class).

### Details and comments
The method `repeat()` internally calls the private method `_return_repeat()` which is inherited from the `Gate` class.
It creates a new `Gate` object with arguments `params=self.params`, which is expected to be of type `ParameterExpression` or `float`. However the parameters are of type `numpy.ndarray` for a `UnitaryGate`. 
A simple fix is to omit these parameters (`params=[]`) as they are not directly used by the returned `Gate`. Actually these parameters are still stored in the `repeat()` method and used only there.


releasenotes/notes/fix_11990-8551c7250207fc76.yaml
```diff
@@ -0,0 +1,6 @@
+---
+fixes:
+  - |
+    Fixes an error when calling the method :meth:`.UnitaryGate.repeat()`.
```
ElePT: The :meth: directive adds the parenthesis when the note is rendered s no need to add it explicitly

```suggestion
    Fixes an error when calling the method :meth:`.UnitaryGate.repeat`. 
```

ElePT (commented):
Thanks a lot @timmintam! The PR looks pretty good, I just added a small suggestion

Cryoris:
This would also happen for a set of other gates that have special `params` attributes, right? E.g. `Isometry` (also array), `UCGate` (list of arrays), `PermutationGate` (matrix of ints)... If it does, it would be nice to fix this generally 🙂 

These would fail because the custom `validate_parameter` functions are not being used, since `_return_repeat` constructs a generic `Gate`. Maybe we could pass that function along?

timmintam:
> This would also happen for a set of other gates that have special `params` attributes, right? E.g. `Isometry` (also array), `UCGate` (list of arrays), `PermutationGate` (matrix of ints)... If it does, it would be nice to fix this generally 🙂
> 
> These would fail because the custom `validate_parameter` functions are not being used, since `_return_repeat` constructs a generic `Gate`. Maybe we could pass that function along?

You are right @Cryoris, this would fail for other gates as well actually. Do you mean something along these lines this in the `Gate` class ?

```python
def _return_repeat(self, exponent: float) -> "Gate":
   gate = Gate(name=f"{self.name}*{exponent}", num_qubits=self.num_qubits, params=[])
   gate.validate_parameter = self.validate_parameter
   gate.params = self.params
   return gate
```
It seems to do the trick.

Cryoris:
Yeah exactly! 🙂 An alternative solution would be to extend the `Gate.__init__` to accept a custom `validate_parameter`, but in Python we can just override this function and it should be fine.

Could you also add tests for these other gate types that fail?

test/python/circuit/test_isometry.py
```diff
@@ -133,6 +133,24 @@ def test_isometry_inverse(self, iso):
...
+        np.eye(2, 2),
+        random_unitary(2, seed=297102).data,
+        np.eye(4, 4),
+        random_unitary(4, seed=123642).data,
+        random_unitary(8, seed=568288).data,
+    )
+    def test_isometry_repeat(self, iso):
+        """Tests for the repeat of isometries from n to n qubits"""
+        if len(iso.shape) == 1:
+            iso = iso.reshape((len(iso), 1))
```
ElePT: The unitaries you are defining in the data all have shape != 1, don't they? in that case we can skip these lines:

```suggestion
```

ElePT (commented):
Thanks a lot [timmintam](https://github.com/timmintam), just to err on the side of caution, could you also add a test for the `PermutationGate` case in `test/python/circuit/library/test_permutation.py`?

ElePT (dismissed):
Nice! The changes look good to me, but let's wait to see if @Cryoris has any final comment before merging.

releasenotes/notes/fix_11990-8551c7250207fc76.yaml
```diff
@@ -0,0 +1,6 @@
+---
+fixes:
+  - |
+    Fixes an error when calling the method :meth:`.UnitaryGate.repeat`. 
+    Refer to `#11990 <https://github.com/Qiskit/qiskit/issues/11990>` for more
```
Cryoris: I think this needs a trailing underscore (but @ElePT might know this better than me 🙂)
```suggestion
    Refer to `#11990 <https://github.com/Qiskit/qiskit/issues/11990>`__ for more
```

Cryoris (commented):
Nice, thanks for the update! Sorry I wasn't explicit about all gates that would need this kind of testing, this should be the complete list:
* `Initialize` (though this is an `Instruction` so this might need the same fix on the `Instruction` class)
* `StatePreparation`
* `DiagonalGate`
* `LinearFunction` (not called gate but is a gate 🙂 )
* `PauliGate`
* `UCPauliRotGate`
* `HamiltonianGate`
Could you add a small test for these as well? They can all also be on 1 or 2 qubits max like what you already have 🙂

Cryoris (approved):
LGTM thanks for the updates 🙂

ElePT (approved):
LGTM to me too!

PR #12870, opened 2024-07-31
Oxidize commutation checker

sbrandhsn:
### Summary

fixes #12702 

### Details and comments
This PR implements the commutation checker with a commutation cache and library as well as a helper function for composing unitary matrices in Rust. I left a couple of things for future work:
- Composing unitary matrices in Rust is currently not as general as the python `Operator.compose` but can be generalized fairly easily when `Operator` is implemented in Rust, too. 
- The function signature of some of the commutation functions is pretty large - this is because qubits and clbits in `DAGOpNode` are a `PyTuple` and I wanted to get rid of Python objects as early as possible. :-) The signatures can be shortened when qubits and clbits are also in Rust space.
Two things are changed lower in the Qiskit stack in `circuit_instruction.rs` and `operations.rs`:
- The methods `is_conditioned` returns whether an instruction is conditioned on classical bits and has been added to `circuit_instruction.rs` 
- ~The `to_matrix` method of `PyGate` and `PyOperation` has been extended to include a call to python-side `Operator(self)` in case the python-side `to_matrix` call fails. This is a change in behavior and I can pull up that code into the `commutation_checker.rs` but I wanted to hear everyone's opinion here. I think that if we have all the necessary information to return a matrix, we should attempt to do that even if it comes at an increase in runtime.~ - removed after further discussion
I'm fairly unfamiliar with Rust so if some of the behavior can be describe more succinctly please let me know. :-)

### Speedups
We can see a speedup of up to 5x depending on the type of input circuit that should be evaluated for commutation relations. I suspect that we do not see a 10x speedup, as some may have expected, because: the heavy-lifting (matrix multiplications) is done by high-performance numpy code, the incorporated commutation caching quickly turns most commutation resolutions into a constant-time lookup the larger the circuit becomes, and the commutation library already contains most of the gates that we have in rust.

Random circuits with parameterized gates only benefit the most from the rust implementation:
rotations_circuits

Followed by random circuits with any kind of gate:
random_circuits

The speedup becomes larger for Clifford circuits on larger circuits as gate checks outweigh the contant-time lookup of clifford gates in the commutation library by far
clifford_circuits

PR #12995, opened 2024-08-20
Oxidize Commutation Analysis

sbrandhsn:
### Details and comments
Part of #12208, oxidises `CommutationAnalysis`. This is based on PR #12959 (and thus #12550) and PR #12870 and shows promising runtime improvements in preliminary tests. When #12550 is merged, the interface to `CommutationChecker` needs a little thought, then this should be ready to go, in principle.


crates/accelerate/src/commutation_analysis.rs
```diff
@@ -0,0 +1,179 @@
...
+use ahash::AHasher;
+use hashbrown::HashMap;
+use indexmap::IndexSet;
+use pyo3::prelude::*;
+
+use pyo3::types::{PyDict, PyList};
+use qiskit_circuit::dag_circuit::{DAGCircuit, NodeType, Wire};
+use rustworkx_core::petgraph::stable_graph::NodeIndex;
+
+type AIndexSet<T> = IndexSet<T, BuildHasherDefault<AHasher>>;
```
mtreinish: This can be expressed a bit more succinctly with:

```suggestion
type AIndexSet<T> = IndexSet<T, ::ahash::RandomState>;
```

That being said I'm not sure I'm a fan of using a type alias for this instead of just using `IndexSet<NodeIndex, RandomState>` (assuming a use `ahash::RandomState` is added) where we need to define a new indexset.

crates/accelerate/src/commutation_analysis.rs
```diff
@@ -0,0 +1,179 @@
...
+                                    None => &empty_params,
+                                };
+                                let params2 = match packed_inst1.params.as_ref() {
+                                    Some(params) => params,
+                                    None => &empty_params,
+                                };
+                                let qargs1 = dag.qargs_interner.get(packed_inst0.qubits);
+                                let qargs2 = dag.qargs_interner.get(packed_inst1.qubits);
+                                let cargs1 = dag.cargs_interner.get(packed_inst0.clbits);
+                                let cargs2 = dag.cargs_interner.get(packed_inst1.clbits);
```
mtreinish:
```suggestion
                                let qargs1 = dag.get_qargs(packed_inst0.qubits);
                                let qargs2 = dag.get_qargs((packed_inst1.qubits);
                                let cargs1 = dag.get_cargs(packed_inst0.clbits);
                                let cargs2 = dag.get_cargs(packed_inst1.clbits);
```

crates/circuit/src/dag_circuit.rs
```diff
@@ -231,9 +231,9 @@ pub struct DAGCircuit {
     cregs: Py<PyDict>,
 
     /// The cache used to intern instruction qargs.
-    qargs_interner: Interner<[Qubit]>,
+    pub qargs_interner: Interner<[Qubit]>,
```
mtreinish: There are `get_qargs` and `get_cargs` methods on `main` now so you don't need to make this public:

```suggestion
    qargs_interner: Interner<[Qubit]>,
```

crates/circuit/src/dag_circuit.rs
```diff
@@ -231,9 +231,9 @@ pub struct DAGCircuit {
...
     /// The cache used to intern instruction cargs.
-    cargs_interner: Interner<[Clbit]>,
+    pub cargs_interner: Interner<[Clbit]>,
```
```suggestion
    cargs_interner: Interner<[Clbit]>,
```

crates/accelerate/src/commutation_analysis.rs
```diff
@@ -0,0 +1,179 @@
...
+                                let op1 = packed_inst0.op.view();
+                                let op2 = packed_inst1.op.view();
+                                let params1 = match packed_inst0.params.as_ref() {
+                                    Some(params) => params,
+                                    None => &empty_params,
+                                };
+                                let params2 = match packed_inst1.params.as_ref() {
+                                    Some(params) => params,
+                                    None => &empty_params,
+                                };
```
mtreinish: You can just use `params_view()` for this:

```suggestion
                                let op1 = packed_inst0.op.view();
                                let op2 = packed_inst1.op.view();
                                let params1 = packed_inst0.params_view();
                                let params2 = packed_inst1.params_view();
```

mtreinish (commented):
This is looking good thanks for doing this. I left some small mostly mechanical comments inline. I still want to review the pass logic to double check it all matches up exactly. Especially as there seem to be a fair amount of `panic!()` calls which I want to double check and make sure we're not masking potential user actionable error conditions (or indicate a logic issue).

crates/accelerate/src/commutation_analysis.rs
Cryoris: ahh perfect -- thanks!

crates/accelerate/src/commutation_analysis.rs
```diff
@@ -0,0 +1,197 @@
...
+    dag: &mut DAGCircuit,
+    commutation_checker: &mut CommutationChecker,
+) -> PyResult<(
+    HashMap<Wire, CommutingNodes>,
+    HashMap<(NodeIndex, Wire), usize>,
+)> {
+    let mut commutation_set: HashMap<Wire, CommutingNodes> = HashMap::new();
+    let mut node_indices: HashMap<(NodeIndex, Wire), usize> = HashMap::new();
+
+    let max_num_qubits = 3;
```
mtreinish: Not that it matters, but I'd have done this with:

```rust
const MAX_NUM_QUBITS: u32 = 3;
```

outside the function definition because it is never changed. But realistically the compiler will likely end up with the same generated code in both cases so it doesn't matter.

crates/accelerate/src/commutation_analysis.rs
```diff
@@ -0,0 +1,197 @@
...
+                    commutation_entry.push(AIndexSet::from_iter([current_gate_idx]))
+                }
+            }
+
+            node_indices.insert(
+                (current_gate_idx, wire.clone()),
+                commutation_entry.len() - 1,
+            );
+        }
+    }
```
mtreinish: This is a probably a good candidate for parallelization. We can naively make this a parallel iterator for the outer loop like:

```rust
0..dag.num_qubits().into_par_iter().map(|qubit| Wire::Qubit(Qubit(qubit as u32))).for_each(|wire| {
```

We'll probably have to play some games around synchronization, although we can probably change it into an iterator collector. But we should do this in a follow up because it might require some deeper changes (as the compiler points them out).

crates/accelerate/src/commutation_analysis.rs
```diff
@@ -0,0 +1,197 @@
...
+            // if the current gate index is not in the set, check whether it commutes with
+            // the previous nodes -- if yes, add it to the commutation set
+            if !last.contains(&current_gate_idx) {
+                let mut all_commute = true;
+
+                for prev_gate_idx in last.iter() {
+                    // if the node is an input/output node, they do not commute, so we only
+                    // continue if the nodes are operation nodes
+                    if let (NodeType::Operation(packed_inst0), NodeType::Operation(packed_inst1)) =
+                        (&dag.dag[current_gate_idx], &dag.dag[*prev_gate_idx])
```
mtreinish, line 83: Can `current_gate_idx` ever point to a non-op node? Not that it really matters because we have need to get the inner packed instruction, I was just wondering from a logic PoV.

crates/accelerate/src/commutation_analysis.rs
```diff
@@ -0,0 +1,197 @@
...
+use pyo3::prelude::*;
+
+use pyo3::types::{PyDict, PyList};
+use qiskit_circuit::dag_circuit::{DAGCircuit, NodeType, Wire};
+use rustworkx_core::petgraph::stable_graph::NodeIndex;
+
+// custom types: IndexSet allows to iterate over the elements in insertion
+// order, but uses the std hasher, which is slower than ahash
+type AIndexSet<T> = IndexSet<T, ::ahash::RandomState>;
+type CommutingNodes = Vec<AIndexSet<NodeIndex>>;
```
mtreinish: I'm wondering about the choice of `IndexSet` here is this just for deduplication? The original python code doesn't seem to use a set for anything.

crates/accelerate/src/commutation_analysis.rs
```diff
@@ -0,0 +1,197 @@
...
+                    )
+                }),
+            ),
+        )?;
+    }
+
+    // Then we add the {(node, wire): index} dictionary
+    for ((node_index, wire), index) in node_indices {
+        // we could cache the py_wires to avoid this match and the python object creation,
+        // but this didn't make a noticable difference in runtime
```
mtreinish: Realistically the `BitData` you're accessing here is the cache. It's basically just doing a vec lookup and returning the python object.

That being said I'm wondering can wire here be anything besides a `Qubit`? Not that it really matters because this code is correct and will handle the other cases as expected.

mtreinish (commented):
Overall this LGTM, after the most recent update the code looks good to me. I had a few questions inline nothing really blocking. But at least for the data structure I'd like to know the rationale behind using a `HashMap<Wire, Vec<IndexSet<NodeIndex>>` as the type when in python it was all lists. I can believe using a `Vec<IndexSet>` as the weight here is the better choice but I'd like to know the rationale before approving.

crates/accelerate/src/commutation_analysis.rs
Cryoris: Still consistency is nice -- I'll move it outside 👍🏻

crates/accelerate/src/commutation_analysis.rs
Cryoris: Yeah good point -- I'll have a look in parallel 🙂 I was trying to propagate the Result coming from `commute_inner` but if we collect then we can still do that 👍🏻

crates/accelerate/src/commutation_analysis.rs
Cryoris: yep it even has to, since the commutation sets also contain indices for the input and output nodes (we iterate over `dag.nodes_on_wire(..., only_ops: false)`)

crates/accelerate/src/commutation_analysis.rs
Cryoris: No you're right, it cannot, since we explicitly only iterate over `0..dag.num_qubits().map(|i| Wire::Qubit(i))`. We can change the code to expect a Qubit, that might be a bit cleaner to the readers

crates/accelerate/src/commutation_analysis.rs
Cryoris: @sbrandhsn motivation was to have a fast lookup (we check if the current gate index is already in the commutation set). I can check what the performance is with a normal vector 🙂

sbrandhsn: I was hoping for a runtime speedup coming from a faster execution of line 79 with a set instead of a list :-) I have not benchmarked this though. It might not be a sane optimization as we probably would only get a runtime speedup when we have large sets of commuting gates. It depends on the input circuit but we might assume that large commuting set do not occur in practice. Besides, I thought if the variable name says that the variable is a set, it should be a set instead of a list. :-D I did not consider deduplication.

crates/accelerate/src/commutation_analysis.rs
sbrandhsn: I was wondering what happens if a dag has non-continuous qubits, i.e. gaps in the set of qubits such as qubits 0...3 and 6..10 with no qubits at index 4, 5 - can this happen and was this supported in the Python implementation?

mtreinish: The indices can't ever be non-continuous in Rust or Python. They're literally the index in the `list`/`Vec` for `dag.qubits`. So if you remove a qubit everything has a shift penalty.  The rust code just makes this a bit more explicit than it ever was in Python because there was a layer of indirection around it in Python with the bit objects.

mtreinish:
I ran ASV comparing this against main to see what the performance looks like and it's really good:
```

| Change   | Before [dff9e812]    | After [70735707] <oxidize-commutative-analysis>   |   Ratio | Benchmark (Parameter)                                             |
|----------|----------------------|---------------------------------------------------|---------|-------------------------------------------------------------------|
| -        | 241±2ms              | 217±0.6ms                                         |    0.9  | utility_scale.UtilityScaleBenchmarks.time_bv_100('cx')            |
| -        | 250±0.8ms            | 224±2ms                                           |    0.9  | utility_scale.UtilityScaleBenchmarks.time_bv_100('cz')            |
| -        | 4.08±0.02s           | 3.64±0.01s                                        |    0.89 | utility_scale.UtilityScaleBenchmarks.time_qv('ecr')               |
| -        | 722±6ms              | 646±3ms                                           |    0.89 | utility_scale.UtilityScaleBenchmarks.time_square_heisenberg('cz') |
| -        | 1.38±0.01s           | 1.20±0.01s                                        |    0.87 | utility_scale.UtilityScaleBenchmarks.time_qaoa('ecr')             |
| -        | 1.75±0.01s           | 1.46±0.01s                                        |    0.83 | utility_scale.UtilityScaleBenchmarks.time_qaoa('cz')              |
| -        | 6.89±0.02s           | 3.61±0.03s                                        |    0.52 | utility_scale.UtilityScaleBenchmarks.time_qft('cz')               |
| -        | 6.42±0.02s           | 3.18±0.02s                                        |    0.5  | utility_scale.UtilityScaleBenchmarks.time_qft('ecr')              |
| -        | 5.53±0.01s           | 2.46±0.02s                                        |    0.44 | utility_scale.UtilityScaleBenchmarks.time_qft('cx')               |
| -        | 84.8±1ms             | 33.1±0.3ms                                        |    0.39 | passes.PassBenchmarks.time_commutation_analysis(14, 1024)         |
| -        | 133±0.8ms            | 48.7±0.2ms                                        |    0.37 | passes.PassBenchmarks.time_commutation_analysis(20, 1024)         |
| -        | 25.9±0.8ms           | 9.55±0.1ms                                        |    0.37 | passes.PassBenchmarks.time_commutation_analysis(5, 1024)          |
| -        | 36.6±0.2ms           | 13.3±0.2ms                                        |    0.36 | utility_scale.UtilityScaleBenchmarks.time_bvlike('cz')            |
| -        | 36.6±0.2ms           | 13.0±0.07ms                                       |    0.35 | utility_scale.UtilityScaleBenchmarks.time_bvlike('cx')            |
| -        | 37.1±0.1ms           | 13.1±0.07ms                                       |    0.35 | utility_scale.UtilityScaleBenchmarks.time_bvlike('ecr')           |

| Change   | Before [dff9e812]    | After [70735707] <oxidize-commutative-analysis>   | Ratio   | Benchmark (Parameter)                                                                                           |
|----------|----------------------|---------------------------------------------------|---------|-----------------------------------------------------------------------------------------------------------------|
|          | 5.40±0.03s           | 4.81±0.03s                                        | ~0.89   | utility_scale.UtilityScaleBenchmarks.time_qv('cz')                                                              |
|          | 0                    | 0                                                 | n/a     | utility_scale.UtilityScaleBenchmarks.track_bvlike_depth('cx')                                                   |
|          | 0                    | 0                                                 | n/a     | utility_scale.UtilityScaleBenchmarks.track_bvlike_depth('cz')                                                   |
|          | 0                    | 0                                                 | n/a     | utility_scale.UtilityScaleBenchmarks.track_bvlike_depth('ecr')                                                  |
|          | 9.83±0.07ms          | 10.3±0.08ms                                       | 1.05    | utility_scale.UtilityScaleBenchmarks.time_parse_qaoa_n100('ecr')                                                |
|          | 34.4±0.1ms           | 36.0±0.1ms                                        | 1.05    | utility_scale.UtilityScaleBenchmarks.time_parse_square_heisenberg_n100('cx')                                    |
|          | 34.3±0.2ms           | 36.0±0.1ms                                        | 1.05    | utility_scale.UtilityScaleBenchmarks.time_parse_square_heisenberg_n100('cz')                                    |
|          | 171±0.8ms            | 177±0.7ms                                         | 1.04    | passes.MultipleBasisPassBenchmarks.time_optimize_1q_commutation(5, 1024, ['rx', 'ry', 'rz', 'r', 'rxx', 'id'])  |
|          | 184±1ms              | 191±0.3ms                                         | 1.04    | passes.MultipleBasisPassBenchmarks.time_optimize_1q_commutation(5, 1024, ['rz', 'x', 'sx', 'cx', 'id'])         |
|          | 166±0.4ms            | 173±0.7ms                                         | 1.04    | passes.MultipleBasisPassBenchmarks.time_optimize_1q_commutation(5, 1024, ['u', 'cx', 'id'])                     |
|          | 9.87±0.05ms          | 10.3±0.03ms                                       | 1.04    | utility_scale.UtilityScaleBenchmarks.time_parse_qaoa_n100('cx')                                                 |
|          | 9.92±0.05ms          | 10.3±0.09ms                                       | 1.04    | utility_scale.UtilityScaleBenchmarks.time_parse_qaoa_n100('cz')                                                 |
|          | 106±0.3ms            | 110±0.5ms                                         | 1.04    | utility_scale.UtilityScaleBenchmarks.time_parse_qft_n100('cx')                                                  |
|          | 107±0.6ms            | 111±0.5ms                                         | 1.04    | utility_scale.UtilityScaleBenchmarks.time_parse_qft_n100('cz')                                                  |
|          | 106±0.8ms            | 110±0.6ms                                         | 1.04    | utility_scale.UtilityScaleBenchmarks.time_parse_qft_n100('ecr')                                                 |
|          | 34.5±0.5ms           | 35.8±0.2ms                                        | 1.04    | utility_scale.UtilityScaleBenchmarks.time_parse_square_heisenberg_n100('ecr')                                   |
|          | 554±2ms              | 570±2ms                                           | 1.03    | passes.MultipleBasisPassBenchmarks.time_optimize_1q_commutation(14, 1024, ['rx', 'ry', 'rz', 'r', 'rxx', 'id']) |
|          | 582±2ms              | 599±2ms                                           | 1.03    | passes.MultipleBasisPassBenchmarks.time_optimize_1q_commutation(14, 1024, ['rz', 'x', 'sx', 'cx', 'id'])        |
|          | 549±2ms              | 563±4ms                                           | 1.02    | passes.MultipleBasisPassBenchmarks.time_optimize_1q_commutation(14, 1024, ['u', 'cx', 'id'])                    |
|          | 908±4ms              | 925±2ms                                           | 1.02    | passes.MultipleBasisPassBenchmarks.time_optimize_1q_commutation(20, 1024, ['rx', 'ry', 'rz', 'r', 'rxx', 'id']) |
|          | 947±2ms              | 966±3ms                                           | 1.02    | passes.MultipleBasisPassBenchmarks.time_optimize_1q_commutation(20, 1024, ['rz', 'x', 'sx', 'cx', 'id'])        |
|          | 896±6ms              | 914±3ms                                           | 1.02    | passes.MultipleBasisPassBenchmarks.time_optimize_1q_commutation(20, 1024, ['u', 'cx', 'id'])                    |
|          | 3.70±0.1s            | 3.73±0.01s                                        | 1.01    | utility_scale.UtilityScaleBenchmarks.time_circSU2('cx')                                                         |
|          | 395                  | 395                                               | 1.00    | utility_scale.UtilityScaleBenchmarks.track_bv_100_depth('cx')                                                   |
|          | 397                  | 397                                               | 1.00    | utility_scale.UtilityScaleBenchmarks.track_bv_100_depth('cz')                                                   |
|          | 397                  | 397                                               | 1.00    | utility_scale.UtilityScaleBenchmarks.track_bv_100_depth('ecr')                                                  |
|          | 300                  | 300                                               | 1.00    | utility_scale.UtilityScaleBenchmarks.track_circSU2_depth('cx')                                                  |
|          | 300                  | 300                                               | 1.00    | utility_scale.UtilityScaleBenchmarks.track_circSU2_depth('cz')                                                  |
|          | 300                  | 300                                               | 1.00    | utility_scale.UtilityScaleBenchmarks.track_circSU2_depth('ecr')                                                 |
|          | 1607                 | 1607                                              | 1.00    | utility_scale.UtilityScaleBenchmarks.track_qaoa_depth('cx')                                                     |
|          | 1622                 | 1622                                              | 1.00    | utility_scale.UtilityScaleBenchmarks.track_qaoa_depth('cz')                                                     |
|          | 1622                 | 1622                                              | 1.00    | utility_scale.UtilityScaleBenchmarks.track_qaoa_depth('ecr')                                                    |
|          | 1954                 | 1954                                              | 1.00    | utility_scale.UtilityScaleBenchmarks.track_qft_depth('cx')                                                      |
|          | 1954                 | 1954                                              | 1.00    | utility_scale.UtilityScaleBenchmarks.track_qft_depth('cz')                                                      |
|          | 1954                 | 1954                                              | 1.00    | utility_scale.UtilityScaleBenchmarks.track_qft_depth('ecr')                                                     |
|          | 2709                 | 2709                                              | 1.00    | utility_scale.UtilityScaleBenchmarks.track_qv_depth('cx')                                                       |
|          | 2709                 | 2709                                              | 1.00    | utility_scale.UtilityScaleBenchmarks.track_qv_depth('cz')                                                       |
|          | 2709                 | 2709                                              | 1.00    | utility_scale.UtilityScaleBenchmarks.track_qv_depth('ecr')                                                      |
|          | 462                  | 462                                               | 1.00    | utility_scale.UtilityScaleBenchmarks.track_square_heisenberg_depth('cx')                                        |
|          | 462                  | 462                                               | 1.00    | utility_scale.UtilityScaleBenchmarks.track_square_heisenberg_depth('cz')                                        |
|          | 462                  | 462                                               | 1.00    | utility_scale.UtilityScaleBenchmarks.track_square_heisenberg_depth('ecr')                                       |
|          | 3.80±0.06s           | 3.73±0.1s                                         | 0.98    | utility_scale.UtilityScaleBenchmarks.time_circSU2('cz')                                                         |
|          | 3.91±0.02s           | 3.85±0.2s                                         | 0.98    | utility_scale.UtilityScaleBenchmarks.time_circSU2('ecr')                                                        |
|          | 2.87±0.02s           | 2.74±0.02s                                        | 0.95    | utility_scale.UtilityScaleBenchmarks.time_qv('cx')                                                              |
|          | 450±6ms              | 423±5ms                                           | 0.94    | utility_scale.UtilityScaleBenchmarks.time_square_heisenberg('cx')                                               |
|          | 613±3ms              | 572±2ms                                           | 0.93    | utility_scale.UtilityScaleBenchmarks.time_square_heisenberg('ecr')                                              |
|          | 241±0.5ms            | 222±1ms                                           | 0.92    | utility_scale.UtilityScaleBenchmarks.time_bv_100('ecr')                                                         |
|          | 810±6ms              | 743±5ms                                           | 0.92    | utility_scale.UtilityScaleBenchmarks.time_qaoa('cx')                                                            |

PERFORMANCE INCREASED.
```

qiskit/transpiler/passes/optimization/commutation_analysis.py
```diff
@@ -33,6 +32,13 @@ def __init__(self, *, _commutation_checker=None):
         # do not care about commutations of all gates, but just a subset
         if _commutation_checker is None:
             _commutation_checker = scc
+
+        if isinstance(_commutation_checker, PyCommutationChecker):
+            # If an instance of the to-be deprecated Python implementation of the CommutationChecker
+            # is passed, defer to the internally held Rust implementation instead. This is needed
+            # as the ``analyze_commutations`` function expects the Rust implementation.
+            _commutation_checker = _commutation_checker.cc
+
```
mtreinish: We can remove this condition, we didn't deprecate the Python side and just unified the implementations. I don't think we're ever be getting the inner rust object directly anymore.

mtreinish (commented):
This LGTM, one super small thing in the Python code that is out of date with what got merged in for the commutation checker. It's not a bug but we just have an extra condition and comment that aren't needed anymore.

qiskit/transpiler/passes/optimization/commutation_analysis.py
mtreinish: Fixed in: https://github.com/Qiskit/qiskit/pull/12995/commits/68f4f4b51748e1e581e11d8e636aceb8df59fe2b

mtreinish (approved):
I fixed that last little nit and now this should be ready to merge. Thanks @sbrandhsn and @Cryoris for doing this!

crates/accelerate/src/commutation_analysis.rs
mtreinish: Switched to a `Vec` in: https://github.com/Qiskit/qiskit/pull/12995/commits/70735707d1a1490502d757528399f6b461fbdbda

PR #12983, opened 2024-08-19
Deprecates `StochasticSwap` and suggests the use of `SabreSwap`

danielbultrini:
### Summary
I added depreciation warnings for issue #12552 and modified unit tests to capture the new errors. Documentation with examples has also been added.

### Details and comments

I have added a deprecation warning to the ``StochasticSwap`` class and a description of how to change things. All relevant tests were made to capture the warning, but some tests which check all different types of these passes (i.e. Sabre, look ahead) where split so that the stochastic method may easily be removed in future. In one class of test (test_mappers.py) StochasticSwap had to be removed, as the method used to run the tests made it prohibitively hard to easily check for warnings in the stochastic case exclusively.

PR #12999, opened 2024-08-20
Tiny follow up to #12983

Cryoris:
### Summary

Tiny follow up on #12983 fixing a typos & updating releasenotes.

### Details and comments

Since `MUMBAI_CMAP` is only internally available, I suggested a change to a full example in the reno 🙂

### DIFF
--- a/releasenotes/notes/deprecate-StochasticSwap-451f46b273602b7b.yaml
+++ b/releasenotes/notes/deprecate-StochasticSwap-451f46b273602b7b.yaml
@@ -36,7 +36,7 @@ deprecations_transpiler:
           target=backend.target
       )
 
-    While for a pass manager change::
+    While for a pass manager, the change would be::
 
       passmanager = PassManager(StochasticSwap(coupling, 20, 13))
       new_qc = passmanager.run(qc)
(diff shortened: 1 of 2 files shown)


releasenotes/notes/deprecate-StochasticSwap-451f46b273602b7b.yaml
```diff
@@ -1,45 +1,49 @@
...
-      qr = QuantumRegister(4, "q")
-      qc = QuantumCircuit(qr)
+          qc,
+          routing_method="sabre",
+          layout_method="sabre",
+          seed_transpiler=12342,
+          target=backend.target
+      )
+
+    While for a pass manager change::
```
ElePT:
```suggestion
    While for a pass manager, the change would be::
```

ElePT (commented):
Thanks a lot! I didn't realize the sphinx syntax in the deprecation message, I guess my brain is trained to ignore it. Now that we are at it, I had wanted to suggest this tiny change in the original PR but didn't not to saturate it further with comments.

ElePT (approved):
LGTM

Issue #13001, opened 2024-08-20
Port `circuit_to_dag` to Rust


raynelfss:
### What should we add?

With #12550 getting merged soon, we should be thinking of moving the ``circuit_to_dag`` converter into Rust. This issue will keep track of the progress on this effort and describe the designated roadmap.

```[tasklist]
### Tasks
- [ ] https://github.com/Qiskit/qiskit/pull/12550
- [ ] https://github.com/Qiskit/qiskit/pull/13006
- [ ] #13002
- [ ] #13003
- [ ] #13004
- [ ] ``circuit_to_dag`` method exposed to Python while using ``DAGCIrcuit::from_circuit_data`` to perform the conversion.
```

PR #13006, opened 2024-08-20
Expose `CircuitData` interners and registers to the `qiskit-circuit` crate

raynelfss:
### Summary
Tracked by #13001 
This commit adds small changes to the visibility of the internal data of the `CircuitData` object, which would allow for an easier conversion into a `DAGCircuit` once #13001 is completed.

### Details and comments:
- Make `qargs_interner`, `cargs_interner`, `qubits` and `clbits` attributes visible to the rest of the `qiskit-circuit` crate. This could allow for easier transfer of data when converting from `CircuitData` to `DAGCircuuit`.
- Make `CircuitData::iter()` method return an exact-size iterator. Which would allow us to determine the size of the circuit without needing to collect the iterator.

### Suggestions
Leave any suggestions in a comment here.

### DIFF
--- a/crates/circuit/src/circuit_data.rs
+++ b/crates/circuit/src/circuit_data.rs
@@ -1142,7 +1142,7 @@ impl CircuitData {
         &self.qargs_interner
     }
 
-    /// Returns an immutable view of the Interner used for Qargs
+    /// Returns an immutable view of the Interner used for Cargs
     pub(crate) fn view_cargs_interner(&self) -> &IndexedInterner<Vec<Clbit>> {
         &self.cargs_interner
     }
--- a/crates/circuit/src/circuit_data.rs
+++ b/crates/circuit/src/circuit_data.rs
@@ -91,13 +91,13 @@ pub struct CircuitData {
     /// The packed instruction listing.
     data: Vec<PackedInstruction>,
     /// The cache used to intern instruction bits.
-    qargs_interner: IndexedInterner<Vec<Qubit>>,
+    pub(crate) qargs_interner: IndexedInterner<Vec<Qubit>>,
     /// The cache used to intern instruction bits.
-    cargs_interner: IndexedInterner<Vec<Clbit>>,
+    pub(crate) cargs_interner: IndexedInterner<Vec<Clbit>>,
     /// Qubits registered in the circuit.
-    qubits: BitData<Qubit>,
+    pub(crate) qubits: BitData<Qubit>,
     /// Clbits registered in the circuit.
-    clbits: BitData<Clbit>,
+    pub(crate) clbits: BitData<Clbit>,
     param_table: ParameterTable,
     #[pyo3(get)]
     global_phase: Param,
@@ -1133,7 +1133,7 @@ impl CircuitData {
     }
 
     /// Returns an iterator over all the instructions present in the circuit.
-    pub fn iter(&self) -> impl Iterator<Item = &PackedInstruction> {
+    pub fn iter(&self) -> impl ExactSizeIterator<Item = &PackedInstruction> {
         self.data.iter()
     }
 
--- a/crates/circuit/src/circuit_data.rs
+++ b/crates/circuit/src/circuit_data.rs
@@ -16,7 +16,7 @@ use std::cell::OnceCell;
 use crate::bit_data::BitData;
 use crate::circuit_instruction::{CircuitInstruction, OperationFromPython};
 use crate::imports::{ANNOTATED_OPERATION, CLBIT, QUANTUM_CIRCUIT, QUBIT};
-use crate::interner::{IndexedInterner, Interner};
+use crate::interner::{Index, IndexedInterner, Interner};
 use crate::operations::{Operation, OperationRef, Param, StandardGate};
 use crate::packed_instruction::{PackedInstruction, PackedOperation};
 use crate::parameter_table::{ParameterTable, ParameterTableError, ParameterUse, ParameterUuid};
@@ -91,13 +91,13 @@ pub struct CircuitData {
     /// The packed instruction listing.
     data: Vec<PackedInstruction>,
     /// The cache used to intern instruction bits.
-    pub(crate) qargs_interner: IndexedInterner<Vec<Qubit>>,
+    qargs_interner: IndexedInterner<Vec<Qubit>>,
     /// The cache used to intern instruction bits.
-    pub(crate) cargs_interner: IndexedInterner<Vec<Clbit>>,
+    cargs_interner: IndexedInterner<Vec<Clbit>>,
     /// Qubits registered in the circuit.
-    pub(crate) qubits: BitData<Qubit>,
+    qubits: BitData<Qubit>,
     /// Clbits registered in the circuit.
-    pub(crate) clbits: BitData<Clbit>,
+    clbits: BitData<Clbit>,
     param_table: ParameterTable,
     #[pyo3(get)]
     global_phase: Param,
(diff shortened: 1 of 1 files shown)


kevinhartman (commented):
I think we do need these changes, but it might make sense to make all of the fields in `CircuitData` public to the crate, since it feels somewhat arbitrary to restrict some but not all in this case.

The `ExactSizeIterator` change is nice.

raynelfss:
@kevinhartman I agree with making most of the fields in `CircuitData` publicly visible except for two of them.
- The `data` field: Although it would be useful, exposing it would allow it to be mutable if the `CircuitData` instance happens to be owned or mutably borrowed. I'd rather just expose an iterator that returns an immutable reference (like we do with `iter()`) and/or expose a function that returns an immutable view of the `data` field.
- The `ParameterTable`: this attribute is fully private and should not be exposed anywhere else as far as I know based on @jakelishman's assessment.

Thus, that's why I think having both interners and the registers (`BitData`) be the only attributes to be made public to the crate is the correct approach for now.

mtreinish:
What about `global_phase`?

kevinhartman:
> exposing it would allow it to be mutable if the CircuitData instance happens to be owned or mutably borrowed.

That's true, but what I'm getting at is having any of the fields public (e.g. the DAG's bits) makes it possible for things within `qiskit-circuit` to modify ` DAGCircuit` in ways that make it invalid. For example, with the current state of this PR, we could remove the DAG's bits even if it contains operations that use them. It seems odd to force the implementor of `circuit_to_dag` (and the rest of this crate) to use some parts of the proper / public `DAGCircuit` (e.g. `DAGCircuit::iter`) but also expose a select few of its internal fields. My thinking is that because what we're exposing here already gives `qiskit-circuit` the power to create bad data, we shouldn't try to pretend that the visibility has been designed to protect anything.

kevinhartman:
I totally closed that by accident!! I'm so sorry @raynelfss >.<

raynelfss:
Well in that case do you suggest to have functions that return immutable references instead, or functions that can unpack things. I.e. `unpack_qarg()` or `get_qubits()`, these could return immutable references and allow for cloning but never in-place modifications.

jakelishman:
Since I was mentioned: my feeling is probably mostly with Kevin here, on the side of don't make things `pub` if you can break data coherence by mutating them (including assigning a new value to the field).  That includes `global_phase`: you can break data coherence if you assign to it without suitably updating the parameter table.

Having immutable-view getters and safe setters is great.  There'll definitely be places where we need to go further than that, and that'll involve thinking about what APIs we should expose that'll be both useful and efficient.

For `CircuitData` in particular, I think it's probably a good idea to not let the `pub` access get out of hand, even if in other places we've fallen more on the side of permissiveness for now - `CircuitData` is way more core, and the data-coherence requirements _much_ harder to get right.

kevinhartman:
> Well in that case do you suggest to have functions that return immutable references instead

I would also vote for this option. Since `PackedInstruction` already exposes interner keys, it feels reasonable to me for `CircuitData` to expose immutable getters for the interners and bits etc. ^^

jakelishman:
Fwiw, I do think many places will need to be able to query the interners in one way or another for performance, so either special methods, or making sure that there's a way to get at an immutable reference to an interner will be important, I think.

raynelfss:
Alright then, I will make those changes promptly later in the day. Thank you for your input, team 🚀

crates/circuit/src/circuit_data.rs
```diff
@@ -1151,6 +1151,47 @@ impl CircuitData {
         self.data.iter()
     }
 
+    /// Returns an immutable view of the Interner used for Qargs
+    pub(crate) fn view_qargs_interner(&self) -> &IndexedInterner<Vec<Qubit>> {
+        &self.qargs_interner
+    }
+
+    /// Returns an immutable view of the Interner used for Qargs
```
eliarbel:
```suggestion
    /// Returns an immutable view of the Interner used for Cargs
```

crates/circuit/src/circuit_data.rs
```diff
@@ -1151,6 +1151,47 @@ impl CircuitData {
         self.data.iter()
     }
 
+    /// Returns an immutable view of the Interner used for Qargs
+    pub(crate) fn view_qargs_interner(&self) -> &IndexedInterner<Vec<Qubit>> {
```
kevinhartman:
```suggestion
    pub fn qargs_interner(&self) -> &Interner<Vec<Qubit>> {
```

I think all of these methods should have public visibility, since accessing them seems fair game for transpiler passes etc. that would live in other crates.

We should also aim to follow Rust's guidelines on getter method naming:

https://rust-lang.github.io/api-guidelines/naming.html#getter-names-follow-rust-convention-c-getter

In this case, it's better to just name the method `qargs_interner`.

The `view_` prefix I believe made sense in the case of `PackedOperation` since it makes it clear that there's some sort of work happening under the covers to produce the result. I do not believe its choice had anything to do with the mutability of the returned reference—the return type already provides this information to the caller.

kevinhartman (commented):
Just a comment on how these getters are named.

crates/circuit/src/circuit_data.rs
jakelishman: Fwiw, `PackedInstruction::params_view` is called that instead of just `params` because the raw `params` field is also a public `Option<Box<...>>`, and `params_view` is a helper to convert that mess of a type into a `&[Param]`.

crates/circuit/src/circuit_data.rs
```diff
@@ -1151,6 +1151,41 @@ impl CircuitData {
...
+    pub fn cargs_interner(&self) -> &IndexedInterner<Vec<Clbit>> {
+        &self.cargs_interner
+    }
+
+    /// Returns an immutable view of the Global Phase `Param` of the circuit
+    pub fn global_phase(&self) -> &Param {
+        &self.global_phase
+    }
+
+    /// Returns an immutable view of the Qubit register of the circuit
```
eliarbel:
```suggestion
    /// Returns an immutable view of the Qubits registered in the circuit
```

crates/circuit/src/circuit_data.rs
```diff
@@ -1151,6 +1151,41 @@ impl CircuitData {
...
+    pub fn qubits(&self) -> &BitData<Qubit> {
+        &self.qubits
+    }
+
+    /// Returns an immutable view of the Classical register of the circuit
```
eliarbel:
```suggestion
    /// Returns an immutable view of the Classical bits registered in the circuit
```

eliarbel (dismissed):
LGTM, just 2 minor updates to doc comments please

kevinhartman (commented):
Looks good, one other thing we should do here is make the `mod interner` public in `lib.rs` so other crates can actually import the `Interner` trait, which is needed to do anything with the interner.

crates/circuit/src/circuit_data.rs
```diff
@@ -1151,6 +1151,41 @@ impl CircuitData {
...
+
+    /// Unpacks from InternerIndex to `[Qubit]`
+    pub fn get_qargs(&self, index: Index) -> &[Qubit] {
+        self.qargs_interner().intern(index).value
+    }
+
+    /// Unpacks from InternerIndex to `[Clbit]`
+    pub fn get_cargs(&self, index: Index) -> &[Clbit] {
+        self.cargs_interner().intern(index).value
+    }
```
jakelishman: We don't necessarily need these two with the interners exposed - the caller can just do it themselves and have access to whatever new methods, etc, get added in the future (I have something in the works).

mtreinish: FWIW, I added similar methods in #12959 see: https://github.com/Qiskit/qiskit/pull/12959/files#diff-9d10dd27b7540092c29eb23e0fe83a90b8661fb1f452178b60f0e8fb4fd54d60R6165-R6168 I think it's nice to have a helper method for this since it'll be a common operation we'll need to do in transpiler passes and it's less typing.

jakelishman: If you guys think they're worthwhile, I'm not particularly opposed to the shorthand helpers.

kevinhartman (approved):
LGTM. Thanks @raynelfss!

PR #13013, opened 2024-08-21
Fully port InverseCancellation to Rust

### LINKED ISSUE #12271: Port `InverseCancellation` to Rust

mtreinish:
### Summary

This commit builds off of #12959 and the other data model in Rust
infrastructure and migrates the InverseCancellation pass to
leaves Rust until it has finished modifying the DAGCircuit. There is
model that are still in Python, mainly for handling parameter
expressions. But otherwise the entirety of the pass
operates in rust now.

This is just a first pass at the migration here, it moves the pass to
use loops in rust. The next steps here are to look at operating
optimizations being done for different inverse gates/pairs so we should
be able to the throughput of the pass by leveraging multithreading to
handle each inverse option in parallel. This commit does not attempt
this though, because of the Python dependency and also the data
structures around gates and the dag aren't really setup for
multithreading yet and there likely will need to be some work to
support that. 

### Details and comments

~This PR is based on top of https://github.com/Qiskit/qiskit/pull/12959 and as such github shows the entire contents of https://github.com/Qiskit/qiskit/pull/12959 in addition to the contents of this PR. To see the contents of this PR you can look at HEAD on this branch, or just look at:
https://github.com/Qiskit/qiskit/pull/13013/commits/00c7e10c23bb15ac2623c0132a90f11004354125~ rebased

Fixes #12271
Part of #12208

TODO:

- [x] Rebase after #12959 merges
- [ ] Benchmark and tune the implementation


crates/accelerate/src/inverse_cancellation.rs
```diff
@@ -0,0 +1,193 @@
...
+    inverse_gates: Vec<[OperationFromPython; 2]>,
+    self_inverse_gates: Vec<OperationFromPython>,
+    inverse_gate_names: HashSet<String>,
+    self_inverse_gate_names: HashSet<String>,
+) -> PyResult<()> {
+    let op_counts = if !self_inverse_gate_names.is_empty() || !inverse_gate_names.is_empty() {
+        dag.count_ops(py, true)?
+    } else {
+        IndexMap::default()
+    };
```
kevinhartman: Can you use a `OnceCell<PyResult<_>>` here so that the `count_ops` only gets initialized 0 to 1 times instead of checking the conditionals ahead of time?

crates/accelerate/src/inverse_cancellation.rs
```diff
@@ -0,0 +1,193 @@
...
+        return Ok(());
+    }
+    for gate in self_inverse_gates {
+        let gate_count = op_counts.get(gate.operation.name()).unwrap_or(&0);
+        if *gate_count <= 1 {
+            continue;
+        }
+        let mut collect_set: HashSet<String> = HashSet::with_capacity(1);
+        collect_set.insert(gate.operation.name().to_string());
+        let gate_runs: Vec<Vec<NodeIndex>> = dag.collect_runs(collect_set).unwrap().collect();
```
kevinhartman, line 62: It looks like `DAGCircuit::collect_runs` returns an `Option` only because it directly returns the result of the Rustworkx core function, which uses `None` to represent a cycle in the graph (...which is questionable lol). 

Perhaps it is beyond the scope of this PR, but it might be nice if you can change `DAGCircuit::collect_runs` to unwrap that option internally. If it fails, it should be on `DAGCircuit` for getting itself into an invalid state.

crates/accelerate/src/inverse_cancellation.rs
```diff
@@ -0,0 +1,193 @@
...
+            let mut partitions: Vec<Vec<NodeIndex>> = Vec::new();
+            let mut chunk: Vec<NodeIndex> = Vec::new();
+            let max_index = gate_cancel_run.len() - 1;
+            for (i, cancel_gate) in gate_cancel_run.iter().enumerate() {
+                let node = &dag.dag[*cancel_gate];
+                if let NodeType::Operation(inst) = node {
+                    if gate_eq(py, inst, &gate)? {
+                        chunk.push(*cancel_gate);
+                    } else {
+                        let is_empty: bool = chunk.is_empty();
```
kevinhartman: IMO having the temporary variable here hampers readability a tad.

crates/accelerate/src/inverse_cancellation.rs
```diff
@@ -0,0 +1,193 @@
...
+                        partitions.push(std::mem::take(&mut chunk));
+                    } else {
+                        let next_qargs = if let NodeType::Operation(next_inst) =
+                            &dag.dag[gate_cancel_run[i + 1]]
+                        {
+                            next_inst.qubits
+                        } else {
+                            panic!("Not an op node")
+                        };
+                        if inst.qubits != next_qargs {
```
kevinhartman, line 88: Cool! I think this is the first example I've seen of the intern IDs saving us some time :)

crates/accelerate/src/inverse_cancellation.rs
```diff
@@ -0,0 +1,193 @@
...
+    inverse_gate_names: HashSet<String>,
+    inverse_gates: Vec<[OperationFromPython; 2]>,
+) -> PyResult<()> {
+    if !inverse_gate_names
+        .iter()
+        .any(|name| op_counts.contains_key(name))
+    {
+        return Ok(());
+    }
+    for pair in inverse_gates {
```
kevinhartman: I suppose this is a faithful port first an foremost, but it might be nice to see the `inverse_gates` get unpacked to `gate_0` and `gate_1`, here.

crates/circuit/src/dag_circuit.rs
```diff
@@ -6277,6 +6241,51 @@ impl DAGCircuit {
...
+                dag: &DAGCircuit,
+                counts: &mut IndexMap<String, usize, RandomState>,
+            ) -> PyResult<()> {
+                for (key, value) in dag.op_names.iter() {
+                    counts
+                        .entry(key.clone())
+                        .and_modify(|count| *count += value)
+                        .or_insert(*value);
+                }
+                let circuit_to_dag = imports::CIRCUIT_TO_DAG.get_bound(py);
```
kevinhartman, line 6356: Maybe it's worth adding a `count_ops` method to `CircuitData`, and calling that directly instead of converting to a `DAGCircuit` first.

I'm not saying that we should necessarily bother maintaining op counts in `CircuitData` like we do in a `DAGCircuit`, but computing on the fly should still be faster than recursively converting everything to a `DAGCircuit`.

Not necessary for this PR, but I wouldn't be opposed to doing it here either.

kevinhartman (commented):
Looks good so far!

Let me know what you think about the comments (they're mostly minor and not blocking).

crates/circuit/src/dag_circuit.rs
mtreinish: It's funny you mention this, that was added last week: https://github.com/Qiskit/qiskit/commit/5fc1635b6c447e910dbf9f390eb1e6ec789e70a4

but the implementation there is governed by the api of the python space method and not how I would implement it for performance in the transpiler.

crates/accelerate/src/inverse_cancellation.rs
mtreinish: Yeah, I'm fine with making this change to `collect_runs()` but I'd say lets save this for a follow-up PR so we can do it in isolation.

crates/accelerate/src/inverse_cancellation.rs
mtreinish: Removed in: https://github.com/Qiskit/qiskit/pull/13013/commits/efeeda5534e7bf778547674d77cdb819409e1c83

crates/accelerate/src/inverse_cancellation.rs
mtreinish: Done in: https://github.com/Qiskit/qiskit/pull/13013/commits/00153dd8e0e9c1b6aeb3135d044c9d08d937bbc5

crates/accelerate/src/inverse_cancellation.rs
mtreinish: Probably, but I'm not sure it would be necessary. I reworked the logic here to avoid the `default()` call in https://github.com/Qiskit/qiskit/pull/13013/commits/7dee70219033ca759be1b603680d7722765230bb and I think it's a lot more clear now.

kevinhartman: Oh lol yeah, that is much better 😆

kevinhartman (dismissed):
LGTM.

Thanks for the changes! We can fix the result type of `DAGCircuit::collect_runs` in a follow-up PR.

kevinhartman (approved):

PR #13014, opened 2024-08-22
Fix Rust-space `TwoQubitBasisDecomposer` for non-standard KAK gate

### LINKED ISSUE #12998: Rust-space `TwoQubitBasisDecomposer` fails if KAK gate is not Rust-space standard
### Environment

- **Qiskit version**: 1.2.0
- **Python version**: 3.10
- **Operating system**: macOS

### What is happening?

Since moving to Rust space, the `TwoQubitBasisDecomposer` has encoded an assumption that the given KAK gate is a Rust-space `StandardGate`.  This is true in most uses within Qiskit, but not necessarily user-defined targets.

```python
from qiskit.circuit import library as std
from qiskit.synthesis import TwoQubitBasisDecomposer

# note that `CXGate(ctrl_state=0)` is not handled as a "standard" gate.
TwoQubitBasisDecomposer(std.CXGate(ctrl_state=0))(std.CXGate())
```
```text
thread '<unnamed>' panicked at /Users/runner/work/qiskit/qiskit/crates/circuit/src/packed_instruction.rs:168:14:
the caller is responsible for knowing the correct type
stack backtrace:

[ ... snip ... ]
  10:        0x10f2b719a - core::option::expect_failed::ha3671a0dd3229d4c
  11:        0x10f1570ba - qiskit_accelerate::two_qubit_decompose::TwoQubitBasisDecomposer::__pymethod_to_circuit__::h78b6cbefd6ac2cc6
  12:        0x10f0c6488 - pyo3::impl_::trampoline::trampoline::hb8b7fc9b34400178
  13:        0x10f137668 - qiskit_accelerate::two_qubit_decompose::<impl pyo3::impl_::pyclass::PyMethods<qiskit_accelerate::two_qubit_decompose::TwoQubitBasisDecomposer> for pyo3::impl_::pyclass::PyClassImplCollector<qiskit_accelerate::two_qubit_decompose::TwoQubitBasisDecomposer>>::py_methods::ITEMS::trampoline::h3e5e5843f2694fc9
[ ... snip ...]
PanicException                            Traceback (most recent call last)
Cell In[1], line 4
      1 from qiskit.circuit import library as std
      2 from qiskit.synthesis import TwoQubitBasisDecomposer
----> 4 TwoQubitBasisDecomposer(std.CXGate(ctrl_state=0))(std.CXGate())

File ~/code/qiskit/venv310-other/lib/python3.10/site-packages/qiskit/synthesis/two_qubit/two_qubit_decompose.py:667, in TwoQubitBasisDecomposer.__call__(self, unitary, basis_fidelity, approximate, use_dag, _num_basis_uses)
    665 else:
    666     if getattr(self.gate, "_standard_gate", None):
--> 667         circ_data = self._inner_decomposer.to_circuit(
    668             np.asarray(unitary, dtype=complex),
    669             self.gate,
    670             basis_fidelity,
    671             approximate,
    672             _num_basis_uses=_num_basis_uses,
    673         )
    674         return QuantumCircuit._from_circuit_data(circ_data)
    675     else:

PanicException: the caller is responsible for knowing the correct type
```

### What should happen?

We should be able to output the circuit in that case.

### Any suggestions?

The decomposition is actually internally successful, the problem is just the assumption that after `let kak_gate = kak_gate.extract::<OperationFromPython>()?;`, then `kak_gate.operation` internally stores a `StandardGate`.

We should actually check whether that's the case, and if not, avoid using the `CircuitData::from_standard_gates` fast-path constructor an instead use a general-purpose one.

ElePT:
### Summary
Proposal to fix #12998. I can probably extend the unit tests further.


test/python/synthesis/test_synthesis.py
```diff
@@ -1270,6 +1270,15 @@ def test_use_dag(self, euler_bases, kak_gates, seed):
...
             self.assertTrue(decomposition_basis.issubset(requested_basis))
 
+    def test_non_std_gate(self):
+        """Test that the TwoQubitBasisDecomposer class can be correctly instantiated with a
+        non-standard gate.
+
+        Reproduce from: https://github.com/Qiskit/qiskit/issues/12998
+        """
+        # note that `CXGate(ctrl_state=0)` is not handled as a "standard" gate.
+        TwoQubitBasisDecomposer(CXGate(ctrl_state=0))(CXGate())
```
mtreinish: The bug was only present when building the circuit. Can you update this to pass it a unitary matrix to synthesize and maybe assert the output circuit's operator is equivalent to the input and your pygate is correctly used.

mtreinish (commented):
Thanks for fixing this, the fix looks good, just one comment on the tests.

test/python/synthesis/test_synthesis.py
mtreinish: Oh I misread this it is creating a circuit from cx's matrix. It'd still be good to add the assertions though.

ElePT: That's totally on me, I just copied the reproducing example without trying to make it readable. It should look better now :)

test/python/synthesis/test_synthesis.py
```diff
@@ -1270,6 +1270,18 @@ def test_use_dag(self, euler_bases, kak_gates, seed):
...
+        """Test that the TwoQubitBasisDecomposer class can be correctly instantiated with a
+        non-standard KAK gate.
+
+        Reproduce from: https://github.com/Qiskit/qiskit/issues/12998
+        """
+        # note that `CXGate(ctrl_state=0)` is not handled as a "standard" gate.
+        decomposer = TwoQubitBasisDecomposer(CXGate(ctrl_state=0))
+        unitary = CXGate()
+        decomposed_unitary = decomposer(unitary)
+        self.assertEqual(Operator(unitary), Operator(decomposed_unitary))
```
mtreinish: Just to make this a bit more complete so we're testing the decomposition too not just that it worked. I might change this to be:

```suggestion
        unitary = SwapGate().to_matrix()
        decomposed_unitary = decomposer(unitary)
        self.assertEqual(Operator(unitary), Operator(decomposed_unitary))
        self.assertNotIn("swap", decomposed_unitary.count_ops())
        self.assertNotIn("cx", decomposed_unitary.count_ops())
        self.assertEqual(3, decomposed_unitary.count_ops()["cx_o0"])
```

I think this is important because I just checked and for open controlled cx in 1.1.0 this was broken too because the decomposer was treating any CXGate object the same ignoring the ctrl state.

mtreinish (approved):

PR #13018, opened 2024-08-23
Filter aer warnings in deprecated `FakeBackend` V1 class

ElePT:
### Summary
The recent qiskit-aer 0.15 release has introduced deprecation warnings for `BackendV1` inputs in `NoiseModel.from_backend`. We were using this method in the deprecated `FakeBackend` class (which is, by definition, V1) and this is triggering test failures. I suggest filtering the warnings in `FakeBackend` until the class is removed in 2.0.

### DIFF
--- a/qiskit/providers/fake_provider/fake_backend.py
+++ b/qiskit/providers/fake_provider/fake_backend.py
@@ -65,7 +65,13 @@ class FakeBackend(BackendV1):
 
             self.sim = AerSimulator()
             if self.properties():
-                noise_model = NoiseModel.from_backend(self)
+                with warnings.catch_warnings():
+                    warnings.filterwarnings(
+                        "ignore",
+                        category=DeprecationWarning,
+                        message=r".*from_backend using V1 based backend is deprecated as of Aer 0.15*",
+                    )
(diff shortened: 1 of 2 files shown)


jakelishman (commented):
Would it be appropriate to use a test-case warning suppression that targets this category and message, with a further limit on `module="qiskit.providers.fake_provider.fake_backend"`?

Since we've had our first reports of threading failures (albeit unrelated to this), I'm getting more nervous about including things like `warnings.catch_warnings` - if more than one thread enters the `fake_backend` path at the same time, or if another thread enters another `warnings.catch_warnings`, etc, the warnings filters can get out-of-sync and result in one of these filters getting applied after all the catches should have left.

ElePT:
Changed :) This actually how I initially approached it, but then I thought it could be more practical for the removal to just have it in `FakeBackend` directly (I remembered the cleanup we had to do on the `QiskitTestCase` class not so long ago because we would forget to get rid of the filters past their "expiration date"). It should be fine though, now that we are aware of it.

jakelishman (approved):
Ah cool, thanks!  At least if we _do_ forget to remove it, the `module` filter is scoped so tightly that it should be fairly obvious in the future.

PR #13019, opened 2024-08-23
Add equivalence-library rules between `rzz` and `cp`

jakelishman:
### Summary

These gates are locally equivalence (as are all the Ising-interaction gates), and this simple additional rule lets things like QFT, which are defined by Qiskit's default constructor in terms of `cp`, get converted into `rzz` or `rzx`.

### Details and comments

I also have a complete rewrite of the equivalence library that is complete and in theory optimal (in 2q gate count) for translations at jakelishman/qiskit-terra@6654059193, but the current `BasisTranslator` algorithm does not work as well with the structuring of the rules.  I have a new algorithm for translation in mind that I'm calling the `BasisConstructor`, which will solve that problem, and also make basis translation for overcomplete basis sets and heterogeneous architectures more accurate, reliable and performant (hopefully!), but I haven't had chance to write it up yet.

### DIFF
--- a/releasenotes/notes/cphase-rzz-equivalence-e8afc37b71a74366.yaml
+++ b/releasenotes/notes/cphase-rzz-equivalence-e8afc37b71a74366.yaml
@@ -1,6 +1,17 @@
 ---
 features_circuits:
   - |
-    New equivalence rules in the standard equivalence library mean that the transpiler can now
-    directly convert between two-qubit continuous Pauli rotations, rather than always decomposing
-    into a discrete two-CX-based solution.
+    The standard equivalence library (:data:`.SessionEquivalenceLibrary`) now has rules that can
+    directly convert between Qiskit's standard-library 2q continuous Ising-type interactions (e.g.
+    :class:`.CPhaseGate`, :class:`.RZZGate`, :class:`.RZXGate`, and so on) using local equivalence
+    relations.  Previously, several of these conversions would go via a 2-CX form, which resulted
+    in less efficient circuit generation.
+
+    .. note::
+
+      In general, the :class:`.BasisTranslator` is not guaranteed to find the "best" equivalence
+      relation for a given :class:`.Target`, but will always find an equivalence if one exists.  We
+      rely on more expensive resynthesis and gate-optimization passes in the transpiler to improve
+      the output.  These passes are currently not as effective for basis sets with a continuously
+      parametrized two-qubit interaction as they are for discrete super-controlled two-qubit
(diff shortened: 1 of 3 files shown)


test/python/circuit/test_controlled_gate.py
```diff
@@ -1438,7 +1438,6 @@ def test_control_zero_operand_gate(self, num_ctrl_qubits):
     @data(
         RXGate,
         RYGate,
-        RZGate,
```
ShellyGarion, line 1441: why did you need to remove this test?

jakelishman: The test is asserting that a failure occurs, and for most of the gates in the list, they shouldn't really fail - we've got enough equivalence and translation rules that it should be possible to translate the gate into one we can represent better. The `add_control` logic is just not super smart about gate translation.

As a side effect of this PR, the RZ path now happens to convert things to `CPhaseGate` rather than `CRZGate` (or something like that), so the control works further than it did before.

ShellyGarion (dismissed):
LGTM. My only comment is that some equivalences related to rotation gates do not have an image, but this is not directly related to this PR, and can be added in another PR.

releasenotes/notes/cphase-rzz-equivalence-e8afc37b71a74366.yaml
```diff
@@ -0,0 +1,6 @@
+---
+features_circuits:
+  - |
+    New equivalence rules in the standard equivalence library mean that the transpiler can now
+    directly convert between two-qubit continuous Pauli rotations, rather than always decomposing
```
Cryoris: Did it always decompose into CX? There are some existing 2q Pauli equivalences that should allow to convert between some two-qubit rotations, no? 🙂 
```suggestion
    directly convert between two-qubit continuous Pauli rotations, rather than sometimes decomposing
```

jakelishman: I'll tweak this note - I wrote it under the same false assumption about how the basis translator works that's alluded to in the PR comment. We need a new translation algorithm to make _any_ choice of translation reliable - the current system is good at finding an arbitrary valid translation, but not at finding _the best_ translation (for any definition of "best").

jakelishman: I've written a new version of the release note in bbe529b604.

Cryoris (approved):

PR #13025, opened 2024-08-23
Fully port Split2QUnitaries to rust

mtreinish:
### Summary

This commit builds off of #13013 and the other data model in Rust
infrastructure and migrates the InverseCancellation pass to
leaves Rust until it has finished modifying the DAGCircuit. There is
model that are still in Python, mainly for creating `UnitaryGate`
instances and `ParameterExpression` for global phase. But otherwise
the entirety of the pass operates in rust now.
    
This is just a first pass at the migration here, it moves the pass to
use loops in rust. The next steps here are to look at operating
optimizations being done for different gates so we should be able to
increase the throughput of the pass by leveraging multithreading to
handle each gate in parallel. This commit does not attempt
this though, because of the Python dependency and also the data
structures around gates and the dag aren't really setup for
multithreading yet and there likely will need to be some work to
support that.
    
### Details and comments

~This PR is based on top of https://github.com/Qiskit/qiskit/pull/13013 and as such github shows the entire contents of https://github.com/Qiskit/qiskit/pull/13013 in addition to the contents of this PR. To see the contents of this PR you can look at HEAD on this branch, or just look at: 
https://github.com/Qiskit/qiskit/pull/13025/commits/0689b736d9f5d49427f07fb6950e5416aedf861d~ Rebased

Part of #12208


crates/accelerate/src/split_2q_unitaries.rs
```diff
@@ -0,0 +1,75 @@
...
+#[pyfunction]
+pub fn split_2q_unitaries(
+    py: Python,
+    dag: &mut DAGCircuit,
+    requested_fidelity: f64,
+) -> PyResult<()> {
+    let nodes: Vec<NodeIndex> = dag.topological_op_nodes()?.collect();
+    for node in nodes {
+        if let NodeType::Operation(inst) = &dag.dag[node] {
+            let qubits = dag.get_qargs(inst.qubits).to_vec();
```
sbrandhsn, line 32: Do we need a `to_vec` here or would it be faster to omit it here?

crates/accelerate/src/split_2q_unitaries.rs
```diff
@@ -0,0 +1,75 @@
...
+
+use crate::two_qubit_decompose::{Specialization, TwoQubitWeylDecomposition};
+
+#[pyfunction]
+pub fn split_2q_unitaries(
+    py: Python,
+    dag: &mut DAGCircuit,
+    requested_fidelity: f64,
+) -> PyResult<()> {
+    let nodes: Vec<NodeIndex> = dag.topological_op_nodes()?.collect();
```
sbrandhsn: This pass does not need to traverse the nodes in topological order - could we skip the linear overhead of constructing that order and instead use `op_nodes`?

crates/circuit/src/dag_circuit.rs
```diff
@@ -6315,6 +6315,60 @@ impl DAGCircuit {
         }
     }
 
+    /// Insert an op given by callback on each individual qubit into a node
+
+    #[allow(unused_variables)]
+    pub fn replace_on_incoming_qubits<F>(
```
sbrandhsn:
```suggestion
    /// Replace a node with individual operations (given by callback function `F`) on each qubit of that node
    #[allow(unused_variables)]
    pub fn replace_node_with_1q_ops<F>(
```
I found the comment and function name a bit confusing and hope this suggestion improves things better. :-)

crates/circuit/src/dag_circuit.rs
```diff
@@ -6315,6 +6315,60 @@ impl DAGCircuit {
...
+        &mut self,
+        py: Python, // Unused if cache_pygates isn't enabled
+        node: NodeIndex,
+        mut insert: F,
```
sbrandhsn: Does `F` need to be mutable?

crates/circuit/src/dag_circuit.rs
```diff
@@ -6315,6 +6315,60 @@ impl DAGCircuit {
...
+                params: (!new_op.params.is_empty()).then(|| Box::new(new_op.params)),
+                extra_attrs: new_op.extra_attrs,
+                #[cfg(feature = "cache_pygates")]
+                py_op: py_op,
+            };
+            let new_index = self.dag.add_node(NodeType::Operation(inst));
+            let parent_index = edge_source;
+            self.dag
+                .add_edge(parent_index, new_index, edge_weight.clone());
+            self.dag.add_edge(new_index, node, edge_weight);
```
sbrandhsn: Is there any merit of connecting `new_op` to the successor of `node` here, instead of `node` directly?

sbrandhsn (commented):
I left a couple of minor comments, apart from these, this LGTM. Do you have some benchmark data? :-)

crates/accelerate/src/split_2q_unitaries.rs
mtreinish: The `to_vec()` here is needed to get an owned object here. `dag.get_qargs()` returns a slice that refers back to qubits owned by `dag`'s qubits list. If I didn't have this the compiler would complain that have I have an immutable reference to `dag` via `qubits` when I try to mutate the dag later in `replace_on_incoming_qubits`. I only added this because the compiler errored when I originally kept it as a slice. This is the same reason I had to collect the node indices into a `Vec` instead of directly iterating over the nodes.

crates/circuit/src/dag_circuit.rs
mtreinish: It does because I used `FnMut` for the the type here. It doesn't have to be mutable if we accepted an `Fn` for `F` instead. I just wasn't sure which kind of callables we'd allow longer term and opted to be a bit more conservative in the typing, and having `FnMut` means the callable is allowed to mutate it's environment. I can change it to `Fn` in this case because the only callback we have so far doesn't need to be `FnMut`.

crates/circuit/src/dag_circuit.rs
mtreinish: The only reason I did it was because this way was because I only had to iterate over the incoming edges and could reuse the logic in the `remove_op_node` but I can inline the logic here and save a few edge operations.

crates/circuit/src/dag_circuit.rs
sbrandhsn: In this case, `FnMut` is probably better as it allows the user more flexibility, it especially allows users to easily track the changes performed by `replace_on_incoming_qubits`.

crates/circuit/src/dag_circuit.rs
sbrandhsn: If it likely won't have a large impact on the runtime, we can leave it as is and reuse the logic. :-)

crates/circuit/src/dag_circuit.rs
mtreinish: Changed in: https://github.com/Qiskit/qiskit/pull/13025/commits/2ff94f5308bfc0b61391ce45b4fd8d0dcdbfde26

crates/circuit/src/dag_circuit.rs
mtreinish: Done in: https://github.com/Qiskit/qiskit/pull/13025/commits/913ec20f546c10233a7aaa9decd1707bf8bfc600

crates/circuit/src/dag_circuit.rs
mtreinish: Done in: https://github.com/Qiskit/qiskit/pull/13025/commits/b2cdb54035363549d6c395c3026b5f7725b3a894

crates/accelerate/src/split_2q_unitaries.rs
mtreinish: Fixed in: https://github.com/Qiskit/qiskit/pull/13025/commits/e924c8af3c9770ad7defec34fe1b6f40c245b321

sbrandhsn (approved):
LGTM, thanks!

mtreinish:
For benchmarks I didn't do anything too extensive. But I ran this script:
```python
import statistics
import time

from qiskit.transpiler.passes import ConsolidateBlocks, Split2QUnitaries

split_pass = Split2QUnitaries()

times = []
for _ in range(100):
    qc = random_circuit(100, 1000)
    blocked = ConsolidateBlocks()(qc)
    start = time.perf_counter()
    split_pass(blocked)
    stop = time.perf_counter()
    runtime = stop - start
    print(runtime)
    times.append(runtime)

mean_runtime = statistics.geometric_mean(times)
print(f"Mean runtime over 100x 100x1000 random circuits: {mean_runtime} sec.")
```

with both `main` and this PR. With `main` it returned:

```
Mean runtime over 100x 100x1000 random circuits: 0.22658202673859693 sec.
```

With this PR it yielded:

```
Mean runtime over 100x 100x1000 random circuits: 0.15633993219475464 sec.
```
It's not as much of as speedup as I was hoping for, but still pretty good. Also, this isn't a perfect comparison because it does include the circuit -> dag overhead also I have no idea how many substitutions are actually getting made by this.

sbrandhsn:
> I have no idea how many substitutions are actually getting made by this.

Probably close to zero, which is the expected case for a reasonable circuit. :-) I think a lot of users would be willing to pay a lot of classical compute if they can save one two-qubit gate without introducing further errors.

mtreinish:
Lol, it's actually zero because I forgot to unroll 3q or more and also the collect blocks pass, so there are no collected unitaries in the circuit. I'll fix it and rerun the script.

PR #13030, opened 2024-08-24
Fully port CheckMap to Rust

### LINKED ISSUE #12251: Port `CheckMap` to Rust
This is blocked on #12052 and #11721

mtreinish:
### Summary

This commit migrates the entirety of the CheckMap analysis pass to Rust.
The pass operates solely in the rust domain and returns an
`Option<(String, [u32; 2])>` to Python which is used to set the two
property set fields appropriately. All the analysis of the dag is done
in Rust. There is still Python interaction required though because
control flow operations are only defined in Python. However the
interaction is minimal and only to get the circuits for control flow
blocks and converting them into DAGs (at least until #13001 is complete).

### Details and comments

~This commit is based on top of #12959 and will need to be rebased after
that merges. In the meantime you can see the contents of this PR at:
https://github.com/Qiskit/qiskit/commit/41fe4539bad0b7c313db54a562c3d699a4cd14b6~ Rebased

Closes #12251
Part of #12208


mtreinish:
I was inspired to quickly do this transpiler pass because asv showed a runtime regression in this pass after #12550 merged. Comparing this PR against main (with that regression) shows a huge speedup:
```

| Change   | Before [d3040a0b] <main>            | After [41fe4539]    |   Ratio | Benchmark (Parameter)                                        |
|----------|-------------------------------------|---------------------|---------|--------------------------------------------------------------|
| -        | 1.81±0.01ms                         | 32.7±0.1μs          |    0.02 | mapping_passes.PassBenchmarks.time_check_map(5, 1024)        |
| -        | 4.41±0.2ms                          | 101±5μs             |    0.02 | mapping_passes.RoutedPassBenchmarks.time_check_map(5, 1024)  |
| -        | 5.38±0.08ms                         | 33.6±0.7μs          |    0.01 | mapping_passes.PassBenchmarks.time_check_map(14, 1024)       |
| -        | 20.9±0.2ms                          | 289±10μs            |    0.01 | mapping_passes.RoutedPassBenchmarks.time_check_map(14, 1024) |
| -        | 38.6±0.6ms                          | 459±8μs             |    0.01 | mapping_passes.RoutedPassBenchmarks.time_check_map(20, 1024) |
| -        | 8.21±0.3ms                          | 32.9±0.05μs         |    0    | mapping_passes.PassBenchmarks.time_check_map(20, 1024)       |

PERFORMANCE INCREASED.
```

(I'm not sure I've ever seen asv report a ratio of `0` before)

mtreinish:
Comparing against 1.2.0 without the regression caused by #12550:
```

| Change   | Before [7e2e835e] <1.2.0^0>   | After [41fe4539]    |   Ratio | Benchmark (Parameter)                                        |
|----------|-------------------------------|---------------------|---------|--------------------------------------------------------------|
| -        | 341±0.6μs                     | 32.7±0.1μs          |    0.1  | mapping_passes.PassBenchmarks.time_check_map(5, 1024)        |
| -        | 866±5μs                       | 33.0±0.3μs          |    0.04 | mapping_passes.PassBenchmarks.time_check_map(14, 1024)       |
| -        | 2.33±0.01ms                   | 99.3±4μs            |    0.04 | mapping_passes.RoutedPassBenchmarks.time_check_map(5, 1024)  |
| -        | 1.24±0.01ms                   | 33.1±0.2μs          |    0.03 | mapping_passes.PassBenchmarks.time_check_map(20, 1024)       |
| -        | 12.0±0.1ms                    | 281±5μs             |    0.02 | mapping_passes.RoutedPassBenchmarks.time_check_map(14, 1024) |
| -        | 20.2±0.4ms                    | 454±7μs             |    0.02 | mapping_passes.RoutedPassBenchmarks.time_check_map(20, 1024) |

PERFORMANCE INCREASED.
```

crates/accelerate/src/check_map.rs
```diff
@@ -0,0 +1,97 @@
...
+            let qubits = dag.get_qargs(inst.qubits);
+            if inst.op.control_flow() {
+                if let OperationRef::Instruction(py_inst) = inst.op.view() {
+                    let raw_blocks = py_inst.instruction.getattr(py, "blocks")?;
+                    let circuit_to_dag = CIRCUIT_TO_DAG.get_bound(py);
+                    for raw_block in raw_blocks.bind(py).iter().unwrap() {
+                        let block_obj = raw_block?;
+                        let block = block_obj
+                            .getattr(intern!(py, "_data"))?
+                            .extract::<CircuitData>()?;
```
raynelfss: I'm not sure if this is less of a cost but what if we immutably borrowed this instead of extracting. I figured, if the logic is correct, it would cost less than taking ownership since we only want to know the number of qubits in this circuit.
```suggestion
                        let block = block_obj
                            .getattr(intern!(py, "_data"))?
                            .downcast::<CircuitData>()?
                            .borrow();
```

crates/accelerate/src/check_map.rs
```diff
@@ -0,0 +1,97 @@
...
+            None => edge_set.contains(&[qubits[0].into(), qubits[1].into()]),
+        }
+    };
+    for node in dag.op_nodes(false) {
+        if let NodeType::Operation(inst) = &dag.dag[node] {
+            let qubits = dag.get_qargs(inst.qubits);
+            if inst.op.control_flow() {
+                if let OperationRef::Instruction(py_inst) = inst.op.view() {
+                    let raw_blocks = py_inst.instruction.getattr(py, "blocks")?;
+                    let circuit_to_dag = CIRCUIT_TO_DAG.get_bound(py);
```
raynelfss, line 48: Since `circuit_to_dag` is coming to Rust soon, would you want to wait until #13036 merges? If it's not too costly then that's fine as well.

raynelfss (commented):
This looks great, thank you for working on this. I just have a couple of in-line comments that I'm a bit iffy on. Other than that, this is about ready to merge. :rocket: (Once you fix your merge conflicts 🐱)

crates/accelerate/src/check_map.rs
mtreinish: If nothing else the error handling will be a bit faster with downcast.

crates/accelerate/src/check_map.rs
mtreinish: I don't think it's particularly worth waiting this is minimal overhead and only comes up with control flow circuits. I think we can easily just update this to use #13036 after that merges.

I think there is actually a path here where we can avoid doing the circuit to dag conversion altogether, like what @kevinhartman did in #13034

crates/accelerate/src/check_map.rs
```diff
@@ -0,0 +1,98 @@
...
+        }
+    }
+    Ok(None)
+}
+
+#[pyfunction]
+pub fn check_map(
+    py: Python,
+    dag: &DAGCircuit,
+    edge_set: HashSet<[u32; 2]>,
```
mtreinish, line 90: Besides dropping `circuit_to_dag` the other potential performance optimization here is building the hash set during init as a pyclass so we avoid converting from a pyset at call time. We can probably do this in a follow up because realistically this probably won't be noticeable except for the cases when there are a lot of edges in the connectivity graph and minimal gates in the circuit (also `CheckMap` is only typically called once during a transpile, for determining whether to run `SabreSwap` or not).

crates/accelerate/src/check_map.rs
```diff
@@ -0,0 +1,98 @@
...
+                let mapped_bits = [
+                    wire_map[qubits[0].0 as usize],
+                    wire_map[qubits[1].0 as usize],
+                ];
+                edge_set.contains(&[mapped_bits[0].into(), mapped_bits[1].into()])
+            }
+            None => edge_set.contains(&[qubits[0].into(), qubits[1].into()]),
+        }
+    };
+    for node in dag.op_nodes(false) {
```
kevinhartman, line 42: Not for this PR, but I wonder if eventually we should add methods like `DAGCircuit::op_node_references` and have that return an iterator of some node reference type which holds the `NodeId` and a `&PackedInstruction` so that users don't need to unwrap. Sort of like what `petgraph` has for their `IntoNodeReferences` trait.

raynelfss: I like the sound of that. Currently a lot of the rust `DAGCircuit` API is limited, and everyone seems to be adding stuff in their transpiler passes PRs to save time. We could discuss this further in its own issue if you're interested in working on it :)

raynelfss (approved):
LGTM! I still think we could wait for #13036 to merge but it seems like you have plans to drop having to perform that conversion, I can add it myself afterwards if anything.

PR #13032, opened 2024-08-25
Add method to add instructions to a DAGCircuit from an iterator of PackedInstruction

### LINKED ISSUE #13003: Create a `DAGCircuit::add_from_iter()` method in rust
### What should we add?

Tracked by #13001

Create a method `DAGCircuit::add_from_iter` to create add a chain of `PackedInstruction` instances to a `DAGCircuit` preventing the duplicate insertion of `DAGOutputNode` instances.

raynelfss:
Solves #13003 

### Summary
Tracked by #13001 and preceeded by #12975
The following commits aim to add a `DAGCircuit::extend()` to add a chain of `DAGOpNodes` to the DAGCircuit based on a sequence of `PackedInstruction`.

### Details and comments
These commits add the following two methods:
- `DAGCircuit::extend()` to add a sequence of `PackedInstruction` continously avoiding having to re-add each bit/var's output node each time an instruction is added. The addition of the output node is only performed once all instructions have been added correctly to the DAGCircuit.

### Known issues:
- At the moment, I'm still thinking of ways of handling vars correctly as we need to keep track of each bit/var's output node and vars are currently represented as `PyObject` which are not hashable.
    - ~~One possible solution would be to use a `PyDict` however the structures that are being stored within this mapping are rust-native (`NodeIndex`, `Wire`)~~.
    - ~~Another solution would be to somehow internalize these structures, which I'm still thinking on how to do.~~
- Feel free to comment with any ideas.

### Blockers
- [x] #12975 (will rebase once merged)
- [x] #12550


ElePT:
I believe that the PR doesn't have any open blocker. I will remove the on hold label. Let me know if I missed something.

raynelfss:
> I believe that the PR doesn't have any open blocker. I will remove the on hold label. Let me know if I missed something.

I was about to do that :) Thank you

crates/circuit/src/dag_circuit.rs
```diff
@@ -6347,6 +6345,139 @@ impl DAGCircuit {
...
+                        .map(|edge| (edge.id(), edge.source()))
+                        .unwrap();
+                    self.dag.remove_edge(edge_id);
+                    predecessor_node
+                };
+
+                vars_last_nodes.set_item(var, new_node.index())?;
+                if var_last_node == new_node {
+                    // TODO: Fix instances of duplicate nodes for Vars
+                    continue;
```
ElePT: If I am not mistaken it looks like you were able to store the var keys as PyDicts, right? Is there something else to fix or was the TODO addressed?

crates/circuit/src/dag_circuit.rs
```diff
@@ -6347,6 +6345,139 @@ impl DAGCircuit {
             Err(DAGCircuitError::new_err("Specified node is not an op node"))
         }
     }
+
+    /// Adds valid instances of [PackedInstruction] to the back of the Circuit.
+    pub fn extend<I>(&mut self, py: Python, iter: I) -> PyResult<Vec<NodeIndex>>
```
ElePT:
```suggestion
    /// Extends the DAG with valid instances of [PackedInstruction]
    pub fn extend<I>(&mut self, py: Python, iter: I) -> PyResult<Vec<NodeIndex>>
```

ElePT (commented):
Thanks @raynelfss! This PR is quite useful for transpiler passes that build new instructions (like unitary synthesis). I have been using these changes in my PR and everything has worked as expected so far, so you can have that as a "unit test" data point. I might be overseeing something but I would approve this PR after addressing a few minor (and quite superficial) comments.

crates/circuit/src/dag_circuit.rs
raynelfss: You are correct, I must have forgotten to remove it.

ElePT (dismissed):
LGTM

ElePT (commented):
Actually, I think I have found an issue when trying to add instructions that have no clbits. When I initialize a packed instruction with `clbits: target_dag.cargs_interner.get_default()`,  I then get a `the caller is responsible for only using interner keys from the correct interner` error. I will try to come up with a minimal example that reproduces the problem.

ElePT:
The issue was fixed by #13092, it was unrelated to the changes in this PR :)

ElePT (dismissed):

kevinhartman:
I can give this a quick look as well if it's helpful (I see it is not in the merge queue yet)

crates/circuit/src/dag_circuit.rs
```diff
@@ -6347,6 +6345,135 @@ impl DAGCircuit {
...
+    {
+        // Create HashSets to keep track of each bit/var's last node
+        let mut qubit_last_nodes: HashMap<Qubit, NodeIndex> = HashMap::default();
+        let mut clbit_last_nodes: HashMap<Clbit, NodeIndex> = HashMap::default();
+        // TODO: Refactor once Vars are in rust
+        // Dict [ Var: (int, VarWeight)]
+        let vars_last_nodes: Bound<PyDict> = PyDict::new_bound(py);
+
+        // Store new nodes to return
+        let mut new_nodes = vec![];
```
kevinhartman: As an optimization, you can use `iter.size_hint()` here to reserve an appropriately sized vector if whatever was passed for `iter` happens to provide `Some(upper_bound)` in the tuple returned by `size_hint`.

crates/circuit/src/dag_circuit.rs
```diff
@@ -6347,6 +6345,135 @@ impl DAGCircuit {
...
+            let new_node = self.dag.add_node(NodeType::Operation(instr));
+            new_nodes.push(new_node);
+
+            // Check all the qubits in this instruction.
+            for qubit in self.qargs_interner.get(qubits_id) {
+                // Retrieve each qubit's last node
+                let qubit_last_node = if let Some(node) = qubit_last_nodes.remove(qubit) {
+                    node
+                } else {
+                    let output_node = self.qubit_io_map[qubit.0 as usize][1];
```
kevinhartman, line 6398: Can you add a comment here to explain that what this code is doing is removing the output node edge the first time a qubit is used in `iter`, and that we re-add them after processing all instructions?

It also may be cleaner to use the `entry` API up on line 6393 to get a mutable reference to the value instead of removing `qubit` from `qubit_last_nodes` only to add it back with a different value, but I think a comment would at least be helpful.

kevinhartman (commented):
This is looking good, thanks for adding this @raynelfss!

The one larger point I think we should start to be weary of is introducing new methods that aren't covered by any of our existing testing. We've operated in a bit of a free-for-all the last few months while porting things to Rust, under the assumption that our existing Python test code would provide sufficient coverage. 

IMO, this new method is sufficiently complex and should be unit tested. I'm not sure if we've decided on an approach to exercise the Rust-only API of our `pyclass`s. Perhaps @mtreinish has opinions / suggestions?

mtreinish:
> IMO, this new method is sufficiently complex and should be unit tested. I'm not sure if we've decided on an approach to exercise the Rust-only API of our pyclasss. Perhaps @mtreinish has opinions / suggestions?

In general having rust unit tests is possible and we have them already in place where it is possible/makes sense for PRs like this with new rust specific functionality. See: https://github.com/Qiskit/qiskit/commit/86c63eb3d719a129dbcb92c56e9a67954030b803 for a recent example (which is already setup to run in CI: https://dev.azure.com/qiskit-ci/qiskit/_build/results?buildId=60452&view=logs&j=b9878877-bc64-52fc-fc03-577dd0aa7cb4&t=2cc4852e-f347-585e-188c-b93a0bf20b0f&l=363 ). But what we will struggle with here is that this new method takes a `py` token and we don't have a mechanism to interact with python via rust unit tests.

I would agree we should have this exercised somehow in an ideal world. But I think we probably are ok to let this slide if we can't come up with a solution for unit testing with a python dependency. Given that there are several PRs actively based on top of this and those are exercising the code and tested via Python AFAICT.

crates/circuit/src/dag_circuit.rs
raynelfss: In this case, since we don't exactly know the type of iterator the received collection will turn into, we can't obtain a size hint until we consume the collection into the iterator. This should be fine as long as we use the iterator right after.

kevinhartman (approved):
LGTM, thanks for the changes!

PR #13033, opened 2024-08-25
Move interners from trait to generic structs

jakelishman:
### Summary

This rewrites the interner mechanisms be defined in terms of two generic structs (`Interner` and `Interned`) that formalise the idea that the interned values are owned versions of a given reference type, and move the logically separate intern-related actions (get a key from the value, and get the value from the key) into separate functions (`get` and `insert`, respectively).

This has a few advantages:

1. we now have both `insert` and `insert_owned` methods, which was awkward to add within the trait-based structure.  This allows a more efficient path when the owned variant has already necessarily been constructed.

2. additionally, the standard `insert` path now takes only a reference type.  For large circuits, most intern lookups will, in general, find a pre-existing key, so in situations where the interned value is sufficiently small that it can be within a static allocation (which is the case for almost all `qargs` and `cargs`), it's more efficient not to construct the owned type on the heap.

3. the type of the values retrieved from an interner are no longer indirected through the owned type that's stored.  For example, where `IndexedInterner<Vec<Qubit>>` gave out `&Vec<Qubit>`s as its lookups, `Interner<[Qubit]>` returns the more standard `&[Qubit]`, which is only singly indirect rather than double.

The following replacements are made:

1. The `IndexedInterner` struct from before is now just called `Interner` (and internally, it uses an `IndexSet` rather than manually tracking the indices).  Its generic type is related to the references it returns, rather than the owned value it stores, so `IndexedInterner<Vec<Qubit>>` becomes `Interner<[Qubit]>`.

2. The `Interner` trait is now gone.  Everything is just accessed as concrete methods on the `Interner` struct.

3. `<&IndexedInterner as Interner>::intern` (lookup of the value from an interner key) is now called `Interner::get`.

4. `<&mut IndexedInterner as Interner>::intern` (conversion of an owned value to an interner key) is now called `Interner::insert_owned`.

5. A new method, `Interner::insert`, can now be used when one need not have an owned allocation of the storage type; the correct value will be allocated if required (which is expected to be less frequent).

6. The intern key is no longer called `interner::Index`, but `Interned<T>`, where the generic parameter `T` matches the generic of the `Interner<T>` that gave out the key.

### Details and comments

This is sufficiently generic now that it can probably be used directly for the purposes of #12917, though I didn't make it generic over the key width.  It probably wouldn't be too hard to do that too, if we want it very much.

I'm intending to follow this PR with one that makes the key corresponding to the "empty" state constructible without any hash lookup, since we end up needing that quite a lot (think the cargs key when pushing a standard gate). (*edit*: now #13035.)

### DIFF
--- a/crates/circuit/src/dag_circuit.rs
+++ b/crates/circuit/src/dag_circuit.rs
@@ -231,9 +231,9 @@ pub struct DAGCircuit {
     cregs: Py<PyDict>,
 
     /// The cache used to intern instruction qargs.
-    qargs_cache: Interner<[Qubit]>,
+    qargs_interner: Interner<[Qubit]>,
     /// The cache used to intern instruction cargs.
-    cargs_cache: Interner<[Clbit]>,
+    cargs_interner: Interner<[Clbit]>,
     /// Qubits registered in the circuit.
     pub qubits: BitData<Qubit>,
     /// Clbits registered in the circuit.
@@ -415,8 +415,8 @@ impl DAGCircuit {
             dag: StableDiGraph::default(),
             qregs: PyDict::new_bound(py).unbind(),
             cregs: PyDict::new_bound(py).unbind(),
-            qargs_cache: Interner::new(),
-            cargs_cache: Interner::new(),
+            qargs_interner: Interner::new(),
+            cargs_interner: Interner::new(),
             qubits: BitData::new(py, "qubits".to_string()),
             clbits: BitData::new(py, "clbits".to_string()),
             global_phase: Param::Float(0.),
@@ -1210,10 +1210,10 @@ def _format(operand):
         for node_weight in self.dag.node_weights_mut() {
             match node_weight {
                 NodeType::Operation(op) => {
-                    let cargs = self.cargs_cache.get(op.clbits);
+                    let cargs = self.cargs_interner.get(op.clbits);
                     let carg_bits = old_clbits.map_indices(cargs).map(|b| b.bind(py).clone());
                     op.clbits = self
-                        .cargs_cache
+                        .cargs_interner
                         .insert_owned(self.clbits.map_bits(carg_bits)?.collect());
                 }
                 NodeType::ClbitIn(c) | NodeType::ClbitOut(c) => {
@@ -1418,10 +1418,10 @@ def _format(operand):
         for node_weight in self.dag.node_weights_mut() {
             match node_weight {
                 NodeType::Operation(op) => {
-                    let qargs = self.qargs_cache.get(op.qubits);
+                    let qargs = self.qargs_interner.get(op.qubits);
                     let qarg_bits = old_qubits.map_indices(qargs).map(|b| b.bind(py).clone());
                     op.qubits = self
-                        .qargs_cache
+                        .qargs_interner
                         .insert_owned(self.qubits.map_bits(qarg_bits)?.collect());
                 }
                 NodeType::QubitIn(q) | NodeType::QubitOut(q) => {
@@ -1562,8 +1562,8 @@ def _format(operand):
         target_dag.duration = self.duration.as_ref().map(|d| d.clone_ref(py));
         target_dag.unit.clone_from(&self.unit);
         target_dag.metadata = self.metadata.as_ref().map(|m| m.clone_ref(py));
-        target_dag.qargs_cache = self.qargs_cache.clone();
-        target_dag.cargs_cache = self.cargs_cache.clone();
+        target_dag.qargs_interner = self.qargs_interner.clone();
+        target_dag.cargs_interner = self.cargs_interner.clone();
 
         for bit in self.qubits.bits() {
             target_dag.add_qubit_unchecked(py, bit.bind(py))?;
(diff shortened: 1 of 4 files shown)


crates/circuit/src/dag_circuit.rs
```diff
@@ -231,9 +231,9 @@ pub struct DAGCircuit {
     cregs: Py<PyDict>,
 
     /// The cache used to intern instruction qargs.
-    qargs_cache: IndexedInterner<Vec<Qubit>>,
+    qargs_cache: Interner<[Qubit]>,
```
kevinhartman: Perhaps this would be a good time for us to also rename `{qargs,cargs}_cache` to `{qargs,cargs}_interner` to match the naming in `CircuitData`?

(I meant to do that originally, but it got away from me before we merged `DAGCircuit`.)

crates/circuit/src/dag_circuit.rs
```diff
@@ -1677,14 +1673,12 @@ def _format(operand):
...
-                &mut self.qargs_cache,
-                self.qubits.map_bits(qargs.iter().flatten())?.collect(),
-            )?;
-            let clbits_id = Interner::intern(
-                &mut self.cargs_cache,
-                self.clbits.map_bits(cargs.iter().flatten())?.collect(),
-            )?;
+            let qubits_id = self
+                .qargs_cache
+                .insert_owned(self.qubits.map_bits(qargs.iter().flatten())?.collect());
```
kevinhartman, line 1678: So in this case, we can use `collect` here which will collect into a native array? This seems very cool, but I think I'm not understanding how this works. Can you explain?

crates/circuit/src/interner.rs
```diff
@@ -10,79 +10,149 @@
...
+/// than a true reference, at the cost that it is explicitly not lifetime bound to the interner it
+/// came from; it is up to the user to ensure that they never attempt to query an interner with a
+/// key from a different interner.
+#[derive(Debug, Eq, PartialEq)]
+pub struct Interned<T: ?Sized> {
+    index: u32,
+    // `Interned` is like a non-lifetime-bound reference to data stored in the interner.  Storing
+    // this adds a small amount more type safety to the interner keys when there's several interners
+    // in play close to each other.
+    _type: PhantomData<*const T>,
```
kevinhartman, line 33: Can you explain (mostly just to help me understand) why you're using `*const T` as the type parameter instead of just `T`? It looks like `PhantomData`'s type parameter just has to be `?Sized`.

kevinhartman (commented):
I love this.

Very cool Rust going on here that gives the interner a much nicer interface. Super curious to hear the answers to some of the questions I've left in line.

Thanks for doing this, @jakelishman!

crates/circuit/src/dag_circuit.rs
jakelishman: This might be more straightforward than you had in mind - `<[T] as ToOwned>::Owned` is `Vec<T>`, so this is basically identical to what it was before.  It just looks shorter because it's easier to access the `insert` method, and the `Vec<T>` is a bit more obfuscated (unfortunately - I actually wanted it to be more obvious, but the only way I could think to do that was to introduce a second type parameter that had no real effect).

crates/circuit/src/interner.rs
jakelishman: Vibes mostly lol.  Logically, `Interned` doesn't _own_ the slice it refers to (e.g. when the `Interned` goes out of scope, the slice doesn't get dropped), it's just a way of retrieving the slice.  That makes it like a reference, except it doesn't have a statically checked lifetime, so `PhantomData<&'a T>` doesn't make sense.

From the perspective of the Rust compiler, there's no difference between me having written `_type: PhantomData<T>` and `_type: PhantomData<*const T>`, I just wrote the latter because that's closer to the actual erased type that `Interned` stores (and in a classical interner, the `Interned` type is often _literally_ a pointer).

jakelishman: (strictly, I think `Send` would be auto-impl'ed if I'd used `PhantomData<T>` instead of `PhantomData<*const T>`, so maybe there's _technically_ a small difference)

crates/circuit/src/dag_circuit.rs
kevinhartman: Oh wow, I would not have guessed that. Thanks for clarifying!

I agree with your choice to avoid a redundant type parameter.

crates/circuit/src/interner.rs
kevinhartman: Hmm, alright, that's fine with me. I don't think we'll need to disturb this once it's in place anyhow, so it doesn't seem too important to explain the nuance with a comment etc.

kevinhartman:
Looks good, I'll approve once conflicts are resolved.

crates/circuit/src/interner.rs
jakelishman: My intention had been that this bit of the comment:

> `Interned` is like a non-lifetime-bound reference to data stored in the interner.

was explaining the type of the `PhantomData`.  I'm fine to expand that code comment to be more in depth - I was already trying to answer this question in it anyway.

crates/circuit/src/dag_circuit.rs
jakelishman: Done in fb71771c33.

crates/circuit/src/interner.rs
jakelishman: Done in 2d56e17543.

kevinhartman (approved):

PR #13034, opened 2024-08-25
Fully port `GatesInBasis` to Rust.

### LINKED ISSUE #12275: Port `GatesInBasis` to Rust
Blocked on #11721 and #12052

kevinhartman:
### Summary
Ports `GatesInBasis` to Rust. The implementation avoids the prior conversion to `DAGCircuit` for each control flow block, which should help with performance.

### Details and comments
Additional changes:
- Adds a `PyInstruction::blocks()` for passes which need to get the blocks of a control flow operation as native `CircuitData`.
- Adds `NodeType::unwrap_operation()` for callers who know that a DAG node is an operation.
- Makes `Target` interface public for use in other crates (needed to accept a `&Target` in a `pyfunction`).
- Fixes some mutability / visibility concerns in `DAGCircuit`, naming making `dag`, `qubits`, and `clbits` private again.

Includes some changes from #13056, #13013 and ~#13006~ which ought to merge first.

Resolves #12275

### DIFF
--- a/crates/accelerate/src/gates_in_basis.rs
+++ b/crates/accelerate/src/gates_in_basis.rs
@@ -23,11 +23,7 @@ use qiskit_circuit::packed_instruction::PackedInstruction;
 use qiskit_circuit::Qubit;
 
 #[pyfunction]
-pub fn any_gate_missing_from_target(
-    _py: Python,
-    dag: &DAGCircuit,
-    target: &Target,
-) -> PyResult<bool> {
+fn any_gate_missing_from_target(_py: Python, dag: &DAGCircuit, target: &Target) -> PyResult<bool> {
     #[inline]
     fn is_universal(gate: &PackedInstruction) -> bool {
         matches!(gate.op.name(), "barrier" | "store")
@@ -101,7 +97,7 @@ pub fn any_gate_missing_from_target(
 }
 
 #[pyfunction]
-pub fn any_gate_missing_from_basis(
+fn any_gate_missing_from_basis(
     py: Python,
     dag: &DAGCircuit,
     basis: HashSet<String>,
--- a/crates/accelerate/src/gates_in_basis.rs
+++ b/crates/accelerate/src/gates_in_basis.rs
@@ -23,7 +23,7 @@ use qiskit_circuit::packed_instruction::PackedInstruction;
 use qiskit_circuit::Qubit;
 
 #[pyfunction]
-fn any_gate_missing_from_target(_py: Python, dag: &DAGCircuit, target: &Target) -> PyResult<bool> {
+fn any_gate_missing_from_target(dag: &DAGCircuit, target: &Target) -> PyResult<bool> {
     #[inline]
     fn is_universal(gate: &PackedInstruction) -> bool {
         matches!(gate.op.name(), "barrier" | "store")
--- a/crates/accelerate/src/gates_in_basis.rs
+++ b/crates/accelerate/src/gates_in_basis.rs
@@ -1,0 +1,121 @@
+// This code is part of Qiskit.
+//
+// (C) Copyright IBM 2024
+//
+// This code is licensed under the Apache License, Version 2.0. You may
+// obtain a copy of this license in the LICENSE.txt file in the root directory
+// of this source tree or at http://www.apache.org/licenses/LICENSE-2.0.
+//
+// Any modifications or derivative works of this code must retain this
+// copyright notice, and modified files need to carry a notice indicating
+// that they have been altered from the originals.
+
+use hashbrown::{HashMap, HashSet};
+use pyo3::prelude::*;
+use qiskit_circuit::circuit_data::CircuitData;
+use smallvec::SmallVec;
+
+use crate::nlayout::PhysicalQubit;
+use crate::target_transpiler::Target;
+use qiskit_circuit::dag_circuit::DAGCircuit;
+use qiskit_circuit::operations::Operation;
+use qiskit_circuit::packed_instruction::PackedInstruction;
+use qiskit_circuit::Qubit;
+
+#[pyfunction]
+pub fn any_gate_missing_from_target(
+    _py: Python,
+    dag: &DAGCircuit,
+    target: &Target,
+) -> PyResult<bool> {
+    #[inline]
+    fn is_universal(gate: &PackedInstruction) -> bool {
+        matches!(gate.op.name(), "barrier" | "store")
+    }
+
+    fn visit_gate(
+        target: &Target,
+        gate: &PackedInstruction,
+        qargs: &[Qubit],
+        wire_map: &HashMap<Qubit, PhysicalQubit>,
+    ) -> PyResult<bool> {
+        let qargs_mapped = SmallVec::from_iter(qargs.iter().map(|q| wire_map[q]));
+        if !target.instruction_supported(gate.op.name(), Some(&qargs_mapped)) {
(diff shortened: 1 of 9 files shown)


crates/accelerate/src/gates_in_basis.rs
```diff
@@ -0,0 +1,125 @@
...
+// that they have been altered from the originals.
+
+use hashbrown::{HashMap, HashSet};
+use pyo3::prelude::*;
+use qiskit_circuit::circuit_data::CircuitData;
+use smallvec::SmallVec;
+
+use crate::nlayout::PhysicalQubit;
+use crate::target_transpiler::Target;
+use qiskit_circuit::dag_circuit::DAGCircuit;
```
ElePT, line 20: This is an almost embarrassing suggestion, feel free to ignore it:

```suggestion
use smallvec::SmallVec;

use crate::nlayout::PhysicalQubit;
use crate::target_transpiler::Target;
use qiskit_circuit::circuit_data::CircuitData;
use qiskit_circuit::dag_circuit::DAGCircuit;
```

crates/accelerate/src/gates_in_basis.rs
```diff
@@ -0,0 +1,125 @@
...
+    fn visit_circuit(
+        target: &Target,
+        circuit: &CircuitData,
+        wire_map: &HashMap<Qubit, PhysicalQubit>,
+    ) -> PyResult<bool> {
+        for gate in circuit.iter() {
+            if is_universal(gate) {
+                continue;
+            }
+            let qargs = circuit.qargs_interner().intern(gate.qubits);
```
ElePT: This seems to be outdated syntax for the interner (changed in #13033)

crates/circuit/src/dag_circuit.rs
```diff
@@ -78,6 +78,15 @@ pub enum NodeType {
...
 }
 
+impl NodeType {
+    pub fn unwrap_operation(&self) -> &PackedInstruction {
+        match self {
+            NodeType::Operation(instr) => instr,
+            _ => panic!("Node is not an operation!"),
+        }
+    }
+}
```
ElePT: I may just go ahead and borrow this in my UnitarySynthesis WIP PR now because it's so much nicer than having to unwrap nodes manually (this will be merged earlier anyways).

ElePT (commented):
Thanks @kevinhartman for this PR. I know it's currently blocked by #13056 but I wanted to take a look now to get more familiar with the latest contributions to our Rust crates. From my (limited) point of view, the changes looks good, I really like the additions you made beyond the `GatesInBasis` pass. I did get confused by some changes but then realized that the branch is a bit out of date so I can take a second look once it's updated.

crates/accelerate/src/gates_in_basis.rs
kevinhartman: It looks nicer like that, thanks! I wish `CLion` were better about organizing the things it imports automatically.

crates/accelerate/src/gates_in_basis.rs
```diff
@@ -0,0 +1,121 @@
...
+use crate::nlayout::PhysicalQubit;
+use crate::target_transpiler::Target;
+use qiskit_circuit::dag_circuit::DAGCircuit;
+use qiskit_circuit::operations::Operation;
+use qiskit_circuit::packed_instruction::PackedInstruction;
+use qiskit_circuit::Qubit;
+
+#[pyfunction]
+pub fn any_gate_missing_from_target(
+    _py: Python,
```
raynelfss: This py token seems to no longer be needed here.
```suggestion
```

crates/accelerate/src/gates_in_basis.rs
```diff
@@ -0,0 +1,121 @@
...
+
+    // In the outer DAG, virtual and physical bits are the same thing.
+    let wire_map: HashMap<Qubit, PhysicalQubit> =
+        HashMap::from_iter((0..dag.num_qubits()).map(|i| {
+            (
+                Qubit(i.try_into().unwrap()),
+                PhysicalQubit::new(i.try_into().unwrap()),
+            )
+        }));
+
```
raynelfss, line 84: Is this mapping here mainly to reduce tha cost of re-building `Qubits` or `PhysicalQubit` instances from each other? Not that it needs any correction I just wanted to understand if this was the case.

crates/circuit/src/dag_circuit.rs
```diff
@@ -81,6 +81,15 @@ pub enum NodeType {
     Operation(PackedInstruction),
 }
 
+impl NodeType {
+    pub fn unwrap_operation(&self) -> &PackedInstruction {
+        match self {
+            NodeType::Operation(instr) => instr,
+            _ => panic!("Node is not an operation!"),
```
raynelfss, line 92: Can we add a docstring to this to document why this method exists and why it panics? I know it's a little redundant, but it would be convenient so a developer can know just from the docstring alone.

crates/accelerate/src/gates_in_basis.rs
```diff
@@ -0,0 +1,121 @@
...
+use qiskit_circuit::dag_circuit::DAGCircuit;
+use qiskit_circuit::operations::Operation;
+use qiskit_circuit::packed_instruction::PackedInstruction;
+use qiskit_circuit::Qubit;
+
+#[pyfunction]
+pub fn any_gate_missing_from_target(
+    _py: Python,
+    dag: &DAGCircuit,
+    target: &Target,
```
raynelfss: Are we planning on using this function anywhere outside of `accelerate`? I don't know if we want to expose the `Target` outside of the crate, mostly because of `NullableIndexMap` as it deosn't have access to things such as the `Entry` api. It was already documented to be an internal struct but, if you think it works well enough, we could keep it public.

raynelfss (commented):
This looks about ready to merge. I just have a couple comments about `Target`'s modified visibility and some extra nitpicks.

crates/accelerate/src/gates_in_basis.rs
kevinhartman: The purpose is actually to map qubits within control flow blocks from the indexing scheme used within the current block back to the indexing of the root circuit.

For example, if the root circuit has 5 qubits and has an `if-else` operation that uses qubits 3 and 4, then the inner circuits corresponding to the `if` and `else` blocks of that operation will have 2 qubits, indexed 0 and 1. To get from 0 and 1 each inner block's indexing scheme to 3 and 4 in the root circuit's indexing scheme, we use this map.

crates/accelerate/src/gates_in_basis.rs
kevinhartman: I actually mistakenly was thinking that these methods needed to be public to be exposed with PyO3, but luckily it doesn't require that. I made these private and reverted the visibility changes to the target stuff in cd7c7642f9ef152a38655cbf6d909840fa8e2703.

crates/circuit/src/dag_circuit.rs
kevinhartman: Done in 6245eeb344e491211fbf29d51b0c8defe60df0f6

crates/accelerate/src/gates_in_basis.rs
kevinhartman: Removed in e62a90765d892b65aeb26139ea9612202bc8925f.

raynelfss (approved):
LGTM!

PR #13036, opened 2024-08-26
Port `circuit_to_dag` to Rust

### LINKED ISSUE #13001: Port `circuit_to_dag` to Rust

### LINKED ISSUE #13004: Create a `DAGCircuit::from_circuit_data` to Rust
### What should we add?

Tracked by #13001 

We need a ``DAGCircuit::from_circuit_data`` which would work as our rust-native `circuit_to_dag` but will actually be `CircuitData` to `DAGCircuit`. It could use the method from #13003 but be more specific with the criteria as `CircuitData` contains more information than an `Iterator` does.

raynelfss:
Is tracked by and resolves #13001 and #13004

### Summary
These commits bring `circuit_to_dag` into rust, leveraging all of the api's introduced by #12550. This also resulted in the addition of a rust-native method. `DAGCircuit::from_circuit_data` to convert from `CircuitData` to `DAGCircuit` from Rust alone.

### Details and comments
The following commits aim to add the circuit converter `circuit_to_dag` to rust while leveraging the rust-side logic for `DAGCircuit`. This is originally used to turn an instance of `QuantumCircuit` into `DAGCircuit`. A rust method to convert `CircuitData` to `DAGCircuit` was also added.

#### Inital changes
- Add new method `DAGCircuit::from_quantum_circuit` which uses the data from a `QuantumCircuit` instance to build a dag_circuit.
   - This method uses the API's introduced by #13006 which exposes the `Interner`s and `BitData` instances for qubits and clbits from the `CircuitData` struct for public view.
   - The generated `DAGCircuit` instances has an estimated capacity that should be enough for most `DAGCircuit` instances with the number of operations provided. Uses #12975.
   - After adding up the inner data, each instruction from the `CircuitData` instance is added onto the `DAGCircuit` by using `.add_from_iter()` (introduced by #13032).
- A new method `DAGCircuit::from_circuit_data` uses the previously describe method to create a `DAGCircuit` based only on the `CircuitData` from a circuit.
   - This method aims to focus on the core structure of the circuit and not any of the other attributes that are only used through python (such as Registers, Variables, Calibrations, etc.)
   - Should be used with caution as `CircuitData != QuantumCircuit` and results may differ.
- Expose the method through a `Python` interface with `circuit_to_dag` which goes by the alias of `core_circuit_to_dag` and is called by the original method.
- Add an arbitrary structure `QuantumCircuitData` that extract attributes from the python `QuantumCircuit` instance and makes it easier to access in rust.
   - This structure is for attribute extraction only and is not clonable/copyable.
- Expose a new module `converters` which should store all of the rust-side converters whenever they get brought into rust.

### Tasks
- [x] Fix the testing suite.
- [x] Clean up API
- [x] Benchmark

### Blockers
- [x] #13032 
- [x] #12975 
- [x] #13006 

### Questions and suggestions
Feel free to ask any questions and give any feedback below, thank you! 🚀

### DIFF
--- a/crates/circuit/src/dag_circuit.rs
+++ b/crates/circuit/src/dag_circuit.rs
@@ -6460,19 +6460,17 @@ impl DAGCircuit {
         I: IntoIterator<Item = PackedInstruction>,
     {
         // Create HashSets to keep track of each bit/var's last node
-        let mut qubit_last_nodes: HashMap<Qubit, NodeIndex> = HashMap::default();
-        let mut clbit_last_nodes: HashMap<Clbit, NodeIndex> = HashMap::default();
-        // TODO: Refactor once Vars are in rust
-        // Dict [ Var: (int, VarWeight)]
-        let vars_last_nodes: Bound<PyDict> = PyDict::new_bound(py);
+        let mut qubit_last_nodes: HashMap<Qubit, (NodeIndex, Wire)> = HashMap::default();
+        let mut clbit_last_nodes: HashMap<Clbit, (NodeIndex, Wire)> = HashMap::default();
+        // TODO: Keep track of vars
 
-        // Consume into iterator to obtain size hint
-        let iter = iter.into_iter();
         // Store new nodes to return
-        let mut new_nodes = Vec::with_capacity(iter.size_hint().1.unwrap_or_default());
+        let mut new_nodes = vec![];
         for instr in iter {
             let op_name = instr.op.name();
-            let (all_cbits, vars): (Vec<Clbit>, Option<Vec<PyObject>>) = {
+            // TODO: Use _vars
+            let (all_cbits, _vars): (Vec<Clbit>, Option<Vec<PyObject>>) = {
+                // Check if the clbits are already included
                 if self.may_have_additional_wires(py, &instr) {
                     let mut clbits: HashSet<Clbit> =
                         HashSet::from_iter(self.cargs_interner.get(instr.clbits).iter().copied());
@@ -6497,98 +6495,78 @@ impl DAGCircuit {
             let new_node = self.dag.add_node(NodeType::Operation(instr));
             new_nodes.push(new_node);
 
+            // For each qubit and cl_bit retrieve the last_nodes
+            let mut nodes_to_connect: HashSet<(NodeIndex, Wire)> = HashSet::default();
             // Check all the qubits in this instruction.
             for qubit in self.qargs_interner.get(qubits_id) {
                 // Retrieve each qubit's last node
-                let qubit_last_node = *qubit_last_nodes.entry(*qubit).or_insert_with(|| {
-                    // If the qubit is not in the last nodes collection, the edge between the output node and its predecessor.
-                    // Then, store the predecessor's NodeIndex in the last nodes collection.
+                let qubit_last_node = if let Some((node, wire)) = qubit_last_nodes.remove(qubit) {
+                    (node, wire)
+                } else {
                     let output_node = self.qubit_io_map[qubit.0 as usize][1];
                     let (edge_id, predecessor_node) = self
                         .dag
                         .edges_directed(output_node, Incoming)
                         .next()
-                        .map(|edge| (edge.id(), edge.source()))
+                        .map(|edge| (edge.id(), (edge.source(), edge.weight().clone())))
                         .unwrap();
                     self.dag.remove_edge(edge_id);
                     predecessor_node
-                });
+                };
                 qubit_last_nodes
                     .entry(*qubit)
-                    .and_modify(|val| *val = new_node);
-                self.dag
-                    .add_edge(qubit_last_node, new_node, Wire::Qubit(*qubit));
+                    .and_modify(|val| *val = (new_node, qubit_last_node.1.clone()));
+                nodes_to_connect.insert(qubit_last_node);
             }
 
             // Check all the clbits in this instruction.
             for clbit in all_cbits {
-                let clbit_last_node = *clbit_last_nodes.entry(clbit).or_insert_with(|| {
-                    // If the qubit is not in the last nodes collection, the edge between the output node and its predecessor.
-                    // Then, store the predecessor's NodeIndex in the last nodes collection.
+                let clbit_last_node = if let Some((node, wire)) = clbit_last_nodes.remove(&clbit) {
+                    (node, wire)
+                } else {
                     let output_node = self.clbit_io_map[clbit.0 as usize][1];
                     let (edge_id, predecessor_node) = self
                         .dag
                         .edges_directed(output_node, Incoming)
                         .next()
-                        .map(|edge| (edge.id(), edge.source()))
-                        .unwrap();
-                    self.dag.remove_edge(edge_id);
-                    predecessor_node
-                });
-                clbit_last_nodes
-                    .entry(clbit)
-                    .and_modify(|val| *val = new_node);
-                self.dag
-                    .add_edge(clbit_last_node, new_node, Wire::Clbit(clbit));
-            }
-
-            // If available, check all the vars in this instruction
-            for var in vars.iter().flatten() {
-                let var_last_node = if let Some(result) = vars_last_nodes.get_item(var)? {
-                    let node: usize = result.extract()?;
-                    vars_last_nodes.del_item(var)?;
-                    NodeIndex::new(node)
-                } else {
-                    // If the var is not in the last nodes collection, the edge between the output node and its predecessor.
-                    // Then, store the predecessor's NodeIndex in the last nodes collection.
-                    let output_node = self.var_output_map.get(py, var).unwrap();
-                    let (edge_id, predecessor_node) = self
-                        .dag
-                        .edges_directed(output_node, Incoming)
-                        .next()
-                        .map(|edge| (edge.id(), edge.source()))
+                        .map(|edge| (edge.id(), (edge.source(), edge.weight().clone())))
                         .unwrap();
                     self.dag.remove_edge(edge_id);
                     predecessor_node
                 };
+                clbit_last_nodes
+                    .entry(clbit)
+                    .and_modify(|val| *val = (new_node, clbit_last_node.1.clone()));
+                nodes_to_connect.insert(clbit_last_node);
+            }
 
-                vars_last_nodes.set_item(var, new_node.index())?;
-                self.dag
-                    .add_edge(var_last_node, new_node, Wire::Var(var.clone_ref(py)));
+            // TODO: Check all the vars in this instruction.
+
+            for (node, wire) in nodes_to_connect {
+                self.dag.add_edge(node, new_node, wire);
             }
         }
 
-        // Add the output_nodes back to qargs
-        for (qubit, node) in qubit_last_nodes {
+        // Add the output_nodes back
+        for (qubit, (node, wire)) in qubit_last_nodes {
             let output_node = self.qubit_io_map[qubit.0 as usize][1];
-            self.dag.add_edge(node, output_node, Wire::Qubit(qubit));
+            self.dag.add_edge(node, output_node, wire);
         }
 
-        // Add the output_nodes back to cargs
-        for (clbit, node) in clbit_last_nodes {
+        for (clbit, (node, wire)) in clbit_last_nodes {
             let output_node = self.clbit_io_map[clbit.0 as usize][1];
-            self.dag.add_edge(node, output_node, Wire::Clbit(clbit));
-        }
-
-        // Add the output_nodes back to vars
-        for item in vars_last_nodes.items() {
-            let (var, node): (PyObject, usize) = item.extract()?;
-            let output_node = self.var_output_map.get(py, &var).unwrap();
-            self.dag
-                .add_edge(NodeIndex::new(node), output_node, Wire::Var(var));
+            self.dag.add_edge(node, output_node, wire);
         }
 
         Ok(new_nodes)
+    }
+
+    /// Creates an instance of DAGCircuit from an iterator over `PackedInstruction`.
+    fn from_iter<I>(_py: Python, _iter: I) -> PyResult<Self>
+    where
+        I: IntoIterator<Item = PackedInstruction>,
+    {
+        todo!()
     }
 }
 
--- a/crates/circuit/src/dag_circuit.rs
+++ b/crates/circuit/src/dag_circuit.rs
@@ -6539,8 +6539,7 @@ impl DAGCircuit {
                 });
                 clbit_last_nodes
                     .entry(clbit)
-                    .and_modify(|val| *val = new_node)
-                    .or_insert(new_node);
+                    .and_modify(|val| *val = new_node);
                 self.dag
                     .add_edge(clbit_last_node, new_node, Wire::Clbit(clbit));
             }
@@ -6641,42 +6640,64 @@ impl DAGCircuit {
         new_dag.metadata = qc.metadata.map(|meta| meta.unbind());
 
         // Add the qubits depending on order.
-        let mut qubit_map: HashMap<Qubit, Bound<PyAny>> = HashMap::new();
-        if let Some(qubit_ordering) = qubit_order {
-            for qubit in qubit_ordering {
-                if new_dag.qubits.find(&qubit).is_some() {
-                    return Err(DAGCircuitError::new_err(format!(
-                        "duplicate qubits {}",
-                        &qubit
-                    )));
-                }
-                qubit_map.insert(new_dag.add_qubit_unchecked(py, &qubit)?, qubit);
-            }
+        let qubit_map: Option<Vec<Qubit>> = if let Some(qubit_ordering) = qubit_order {
+            let mut ordered_vec = Vec::from_iter((0..num_qubits as u32).map(Qubit));
+            qubit_ordering
+                .into_iter()
+                .try_for_each(|qubit| -> PyResult<()> {
+                    if new_dag.qubits.find(&qubit).is_some() {
+                        return Err(DAGCircuitError::new_err(format!(
+                            "duplicate qubits {}",
+                            &qubit
+                        )));
+                    }
+                    let qubit_index = qc_data.qubits().find(&qubit).unwrap();
+                    ordered_vec[qubit_index.0 as usize] =
+                        new_dag.add_qubit_unchecked(py, &qubit)?;
+                    Ok(())
+                })?;
+            Some(ordered_vec)
         } else {
-            for qubit in qc_data.qubits().bits() {
-                let bound = qubit.clone_ref(py).into_bound(py);
-                qubit_map.insert(new_dag.add_qubit_unchecked(py, qubit.bind(py))?, bound);
-            }
-        }
+            qc_data
+                .qubits()
+                .bits()
+                .iter()
+                .try_for_each(|qubit| -> PyResult<_> {
+                    new_dag.add_qubit_unchecked(py, qubit.bind(py))?;
+                    Ok(())
+                })?;
+            None
+        };
 
         // Add the clbits depending on order.
-        let mut clbit_map: HashMap<Clbit, Bound<PyAny>> = HashMap::new();
-        if let Some(clbit_ordering) = clbit_order {
-            for clbit in clbit_ordering {
-                if new_dag.clbits.find(&clbit).is_some() {
-                    return Err(DAGCircuitError::new_err(format!(
-                        "duplicate clbits {}",
-                        &clbit
-                    )));
-                }
-                clbit_map.insert(new_dag.add_clbit_unchecked(py, &clbit)?, clbit);
-            }
+        let clbit_map: Option<Vec<Clbit>> = if let Some(clbit_ordering) = clbit_order {
+            let mut ordered_vec = Vec::from_iter((0..num_clbits as u32).map(Clbit));
+            clbit_ordering
+                .into_iter()
+                .try_for_each(|clbit| -> PyResult<()> {
+                    if new_dag.clbits.find(&clbit).is_some() {
+                        return Err(DAGCircuitError::new_err(format!(
+                            "duplicate clbits {}",
+                            &clbit
+                        )));
+                    };
+                    let clbit_index = qc_data.clbits().find(&clbit).unwrap();
+                    ordered_vec[clbit_index.0 as usize] =
+                        new_dag.add_clbit_unchecked(py, &clbit)?;
+                    Ok(())
+                })?;
+            Some(ordered_vec)
         } else {
-            for clbit in qc_data.clbits().bits() {
-                let bound = clbit.clone_ref(py).into_bound(py);
-                clbit_map.insert(new_dag.add_clbit_unchecked(py, clbit.bind(py))?, bound);
-            }
-        }
+            qc_data
+                .clbits()
+                .bits()
+                .iter()
+                .try_for_each(|clbit| -> PyResult<()> {
+                    new_dag.add_clbit_unchecked(py, clbit.bind(py))?;
+                    Ok(())
+                })?;
+            None
+        };
 
         // Add all of the new vars.
         for var in &qc.declared_vars {
@@ -6704,35 +6725,36 @@ impl DAGCircuit {
             }
         }
 
-        // Collect the re-mapped indices for qubits
-        let remapped_qubits: HashMap<Qubit, Qubit> = qubit_map
-            .into_iter()
-            .map(|(index, bit)| (qc_data.qubits().find(&bit).unwrap(), index))
-            .collect();
-        // Collect the re-mapped indices for clbits
-        let remapped_clbits: HashMap<Clbit, Clbit> = clbit_map
-            .into_iter()
-            .map(|(index, bit)| (qc_data.clbits().find(&bit).unwrap(), index))
-            .collect();
-
         // Pre-process and re-intern all indices again.
         let instructions: Vec<PackedInstruction> = qc_data
             .iter()
             .map(|instr| -> PyResult<PackedInstruction> {
                 // Re-map the qubits
-                let qargs: Vec<Qubit> = qc_data
-                    .get_qargs(instr.qubits)
-                    .iter()
-                    .map(|bit| remapped_qubits[bit])
-                    .collect();
-                let new_qubits = new_dag.qargs_interner.insert_owned(qargs);
+                let new_qargs = if let Some(qubit_mapping) = &qubit_map {
+                    let qargs = qc_data
+                        .get_qargs(instr.qubits)
+                        .iter()
+                        .map(|bit| qubit_mapping[bit.0 as usize])
+                        .collect();
+                    new_dag.qargs_interner.insert_owned(qargs)
+                } else {
+                    new_dag
+                        .qargs_interner
+                        .insert(qc_data.get_qargs(instr.qubits))
+                };
                 // Remap the clbits
-                let cargs: Vec<Clbit> = qc_data
-                    .get_cargs(instr.clbits)
-                    .iter()
-                    .map(|bit| remapped_clbits[bit])
-                    .collect();
-                let new_clbits = new_dag.cargs_interner.insert_owned(cargs);
+                let new_cargs = if let Some(clbit_mapping) = &clbit_map {
+                    let qargs = qc_data
+                        .get_cargs(instr.clbits)
+                        .iter()
+                        .map(|bit| clbit_mapping[bit.0 as usize])
+                        .collect();
+                    new_dag.cargs_interner.insert_owned(qargs)
+                } else {
+                    new_dag
+                        .cargs_interner
+                        .insert(qc_data.get_cargs(instr.clbits))
+                };
                 // Copy the operations
 
                 Ok(PackedInstruction {
@@ -6741,8 +6763,8 @@ impl DAGCircuit {
                     } else {
                         instr.op.clone()
                     },
-                    qubits: new_qubits,
-                    clbits: new_clbits,
+                    qubits: new_qargs,
+                    clbits: new_cargs,
                     params: instr.params.clone(),
                     extra_attrs: instr.extra_attrs.clone(),
                     #[cfg(feature = "cache_pygates")]
--- a/crates/circuit/src/dag_circuit.rs
+++ b/crates/circuit/src/dag_circuit.rs
@@ -6644,11 +6644,6 @@ impl DAGCircuit {
 
         // Pre-process the instructions
         let qubit_set: Vec<Qubit> = if let Some(qubit_ordering) = &qubit_order {
-            if qubit_ordering.len() != num_qubits {
-                return Err(PyValueError::new_err(
-                    "'qubit_order' does not contain exactly the same qubits as the circuit",
-                ));
-            };
             let mut qubits = vec![];
             for qubit in qubit_ordering {
                 let bound = qubit.bind(py);
@@ -6669,11 +6664,6 @@ impl DAGCircuit {
             (0..num_qubits as u32).map(Qubit).collect()
         };
         let clbit_set: Vec<Clbit> = if let Some(clbit_ordering) = &clbit_order {
-            if clbit_ordering.len() != num_clbits {
-                return Err(PyValueError::new_err(
-                    "'clbit_order' does not contain exactly the same clbits as the circuit",
-                ));
-            };
             let mut clbits = vec![];
             for clbit in clbit_ordering {
                 let bound = clbit.bind(py);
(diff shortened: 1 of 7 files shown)


raynelfss:
There are still some bugs to fix but so far benchmarks look promising for this:
```bash
All benchmarks:

| Change   | Before [cc87318f] <main>   | After [24df33df] <rusty-circ-to-dag>   | Ratio   | Benchmark (Parameter)                                        |
|----------|----------------------------|----------------------------------------|---------|--------------------------------------------------------------|
|          | n/a                        | n/a                                    | n/a     | converters.ConverterBenchmarks.time_circuit_to_dag(14, 8192) |
|          | n/a                        | n/a                                    | n/a     | converters.ConverterBenchmarks.time_circuit_to_dag(20, 2048) |
|          | n/a                        | n/a                                    | n/a     | converters.ConverterBenchmarks.time_circuit_to_dag(20, 8192) |
|          | n/a                        | n/a                                    | n/a     | converters.ConverterBenchmarks.time_circuit_to_dag(32, 2048) |
|          | n/a                        | n/a                                    | n/a     | converters.ConverterBenchmarks.time_circuit_to_dag(32, 8192) |
|          | n/a                        | n/a                                    | n/a     | converters.ConverterBenchmarks.time_circuit_to_dag(53, 2048) |
|          | n/a                        | n/a                                    | n/a     | converters.ConverterBenchmarks.time_circuit_to_dag(53, 8192) |
| -        | 17.3±0.2μs                 | 7.56±0.09μs                            | 0.44    | converters.ConverterBenchmarks.time_circuit_to_dag(1, 8)     |
| -        | 29.3±0.4μs                 | 11.1±0.1μs                             | 0.38    | converters.ConverterBenchmarks.time_circuit_to_dag(2, 8)     |
| -        | 11.9±0.1ms                 | 3.67±0.1ms                             | 0.31    | converters.ConverterBenchmarks.time_circuit_to_dag(53, 128)  |
| -        | 283±2μs                    | 84.9±0.4μs                             | 0.30    | converters.ConverterBenchmarks.time_circuit_to_dag(20, 8)    |
| -        | 191±0.8μs                  | 55.1±0.5μs                             | 0.29    | converters.ConverterBenchmarks.time_circuit_to_dag(14, 8)    |
| -        | 62.3±0.2μs                 | 18.0±0.1μs                             | 0.29    | converters.ConverterBenchmarks.time_circuit_to_dag(5, 8)     |
| -        | 882±10μs                   | 259±3μs                                | 0.29    | converters.ConverterBenchmarks.time_circuit_to_dag(53, 8)    |
| -        | 106±0.8μs                  | 30.4±0.4μs                             | 0.29    | converters.ConverterBenchmarks.time_circuit_to_dag(8, 8)     |
| -        | 434±8μs                    | 120±2μs                                | 0.28    | converters.ConverterBenchmarks.time_circuit_to_dag(32, 8)    |
| -        | 169±3μs                    | 44.2±0.4μs                             | 0.26    | converters.ConverterBenchmarks.time_circuit_to_dag(1, 128)   |
| -        | 34.5±0.3ms                 | 8.74±0.2ms                             | 0.25    | converters.ConverterBenchmarks.time_circuit_to_dag(14, 2048) |
| -        | 6.00±0.03ms                | 1.50±0.03ms                            | 0.25    | converters.ConverterBenchmarks.time_circuit_to_dag(32, 128)  |
| -        | 9.80±0.2ms                 | 2.27±0.2ms                             | 0.23    | converters.ConverterBenchmarks.time_circuit_to_dag(1, 8192)  |
| -        | 2.20±0ms                   | 502±6μs                                | 0.23    | converters.ConverterBenchmarks.time_circuit_to_dag(14, 128)  |
| -        | 3.44±0.05ms                | 805±4μs                                | 0.23    | converters.ConverterBenchmarks.time_circuit_to_dag(20, 128)  |
| -        | 76.6±0.8ms                 | 17.3±0.3ms                             | 0.23    | converters.ConverterBenchmarks.time_circuit_to_dag(8, 8192)  |
| -        | 2.44±0.01ms                | 531±4μs                                | 0.22    | converters.ConverterBenchmarks.time_circuit_to_dag(1, 2048)  |
| -        | 18.5±0.1ms                 | 4.05±0.1ms                             | 0.22    | converters.ConverterBenchmarks.time_circuit_to_dag(8, 2048)  |
| -        | 351±1μs                    | 72.6±1μs                               | 0.21    | converters.ConverterBenchmarks.time_circuit_to_dag(2, 128)   |
| -        | 47.6±0.9ms                 | 9.96±0.4ms                             | 0.21    | converters.ConverterBenchmarks.time_circuit_to_dag(5, 8192)  |
| -        | 1.24±0.01ms                | 260±5μs                                | 0.21    | converters.ConverterBenchmarks.time_circuit_to_dag(8, 128)   |
| -        | 20.6±0.09ms                | 4.10±0.2ms                             | 0.20    | converters.ConverterBenchmarks.time_circuit_to_dag(2, 8192)  |
| -        | 755±2μs                    | 147±0.08μs                             | 0.20    | converters.ConverterBenchmarks.time_circuit_to_dag(5, 128)   |
| -        | 5.14±0.06ms                | 880±4μs                                | 0.17    | converters.ConverterBenchmarks.time_circuit_to_dag(2, 2048)  |
| -        | 11.6±0.09ms                | 2.01±0.05ms                            | 0.17    | converters.ConverterBenchmarks.time_circuit_to_dag(5, 2048)  |

PERFORMANCE INCREASED.
```

raynelfss:
After fixing all of the failing tests, there was a bit of a performance regression (probably from cloning all the instructions, which I don't think is needed anymore if we do things through rust). But it's still faster than the original.
```bash
All benchmarks:

| Change   | Before [cc2edc9f] <main>   | After [7b765db2] <rusty-circ-to-dag>   | Ratio   | Benchmark (Parameter)                                        |
|----------|----------------------------|----------------------------------------|---------|--------------------------------------------------------------|
|          | n/a                        | n/a                                    | n/a     | converters.ConverterBenchmarks.time_circuit_to_dag(14, 8192) |
|          | n/a                        | n/a                                    | n/a     | converters.ConverterBenchmarks.time_circuit_to_dag(20, 2048) |
|          | n/a                        | n/a                                    | n/a     | converters.ConverterBenchmarks.time_circuit_to_dag(20, 8192) |
|          | n/a                        | n/a                                    | n/a     | converters.ConverterBenchmarks.time_circuit_to_dag(32, 2048) |
|          | n/a                        | n/a                                    | n/a     | converters.ConverterBenchmarks.time_circuit_to_dag(32, 8192) |
|          | n/a                        | n/a                                    | n/a     | converters.ConverterBenchmarks.time_circuit_to_dag(53, 2048) |
|          | n/a                        | n/a                                    | n/a     | converters.ConverterBenchmarks.time_circuit_to_dag(53, 8192) |
| -        | 17.2±0.1μs                 | 12.0±0.3μs                             | 0.70    | converters.ConverterBenchmarks.time_circuit_to_dag(1, 8)     |
| -        | 61.4±0.3μs                 | 37.9±0.5μs                             | 0.62    | converters.ConverterBenchmarks.time_circuit_to_dag(5, 8)     |
| -        | 190±1μs                    | 115±0.7μs                              | 0.61    | converters.ConverterBenchmarks.time_circuit_to_dag(14, 8)    |
| -        | 29.2±0.3μs                 | 17.9±0.2μs                             | 0.61    | converters.ConverterBenchmarks.time_circuit_to_dag(2, 8)     |
| -        | 277±0.3μs                  | 168±1μs                                | 0.61    | converters.ConverterBenchmarks.time_circuit_to_dag(20, 8)    |
| -        | 107±4μs                    | 65.0±1μs                               | 0.61    | converters.ConverterBenchmarks.time_circuit_to_dag(8, 8)     |
| -        | 425±2μs                    | 243±8μs                                | 0.57    | converters.ConverterBenchmarks.time_circuit_to_dag(32, 8)    |
| -        | 20.6±0.1ms                 | 11.3±0.2ms                             | 0.55    | converters.ConverterBenchmarks.time_circuit_to_dag(2, 8192)  |
| -        | 869±6μs                    | 459±1μs                                | 0.53    | converters.ConverterBenchmarks.time_circuit_to_dag(53, 8)    |
| -        | 34.4±2ms                   | 17.9±0.5ms                             | 0.52    | converters.ConverterBenchmarks.time_circuit_to_dag(14, 2048) |
| -        | 48.5±0.8ms                 | 24.7±0.5ms                             | 0.51    | converters.ConverterBenchmarks.time_circuit_to_dag(5, 8192)  |
| -        | 3.35±0.06ms                | 1.67±0.04ms                            | 0.50    | converters.ConverterBenchmarks.time_circuit_to_dag(20, 128)  |
| -        | 11.8±0.04ms                | 5.86±0.1ms                             | 0.50    | converters.ConverterBenchmarks.time_circuit_to_dag(53, 128)  |
| -        | 78.7±0.9ms                 | 39.5±0.6ms                             | 0.50    | converters.ConverterBenchmarks.time_circuit_to_dag(8, 8192)  |
| -        | 363±8μs                    | 178±0.9μs                              | 0.49    | converters.ConverterBenchmarks.time_circuit_to_dag(2, 128)   |
| -        | 5.84±0.06ms                | 2.84±0.05ms                            | 0.49    | converters.ConverterBenchmarks.time_circuit_to_dag(32, 128)  |
| -        | 2.23±0.1ms                 | 1.06±0.02ms                            | 0.48    | converters.ConverterBenchmarks.time_circuit_to_dag(14, 128)  |
| -        | 11.9±0.3ms                 | 5.66±0.2ms                             | 0.48    | converters.ConverterBenchmarks.time_circuit_to_dag(5, 2048)  |
| -        | 18.7±0.4ms                 | 8.94±0.3ms                             | 0.48    | converters.ConverterBenchmarks.time_circuit_to_dag(8, 2048)  |
| -        | 10.5±0.2ms                 | 4.90±0.3ms                             | 0.47    | converters.ConverterBenchmarks.time_circuit_to_dag(1, 8192)  |
| -        | 5.14±0.06ms                | 2.34±0.07ms                            | 0.46    | converters.ConverterBenchmarks.time_circuit_to_dag(2, 2048)  |
| -        | 755±2μs                    | 349±2μs                                | 0.46    | converters.ConverterBenchmarks.time_circuit_to_dag(5, 128)   |
| -        | 1.23±0.02ms                | 569±3μs                                | 0.46    | converters.ConverterBenchmarks.time_circuit_to_dag(8, 128)   |
| -        | 168±2μs                    | 76.0±4μs                               | 0.45    | converters.ConverterBenchmarks.time_circuit_to_dag(1, 128)   |
| -        | 2.52±0.05ms                | 976±9μs                                | 0.39    | converters.ConverterBenchmarks.time_circuit_to_dag(1, 2048)  |

PERFORMANCE INCREASED.
```

crates/circuit/src/dag_circuit.rs
```diff
@@ -6347,6 +6346,323 @@ impl DAGCircuit {
...
+            let (var, node): (PyObject, usize) = item.extract()?;
+            let output_node = self.var_output_map.get(py, &var).unwrap();
+            self.dag
+                .add_edge(NodeIndex::new(node), output_node, Wire::Var(var));
+        }
+
+        Ok(new_nodes)
+    }
+
+    #[allow(clippy::too_many_arguments)]
```
mtreinish: Is there a reason to not just take `QuantumCircuitData` as the input here? That would let you ignore this clippy and you don't have to pull out each field individually and then each argument could be set by name in the struct.

raynelfss: I was thinking that it would add some versatility to just accept the arguments, but adding it to the strctu seems to make more sense.

raynelfss:
After @mtreinish suggestions, the copying of instructions one by one is more effective.

```bash
All benchmarks:

| Change   | Before [3d4bab2b] <main>   | After [7fe0a365] <rusty-circ-to-dag>   | Ratio   | Benchmark (Parameter)                                        |
|----------|----------------------------|----------------------------------------|---------|--------------------------------------------------------------|
|          | n/a                        | n/a                                    | n/a     | converters.ConverterBenchmarks.time_circuit_to_dag(14, 8192) |
|          | n/a                        | n/a                                    | n/a     | converters.ConverterBenchmarks.time_circuit_to_dag(20, 2048) |
|          | n/a                        | n/a                                    | n/a     | converters.ConverterBenchmarks.time_circuit_to_dag(20, 8192) |
|          | n/a                        | n/a                                    | n/a     | converters.ConverterBenchmarks.time_circuit_to_dag(32, 2048) |
|          | n/a                        | n/a                                    | n/a     | converters.ConverterBenchmarks.time_circuit_to_dag(32, 8192) |
|          | n/a                        | n/a                                    | n/a     | converters.ConverterBenchmarks.time_circuit_to_dag(53, 2048) |
|          | n/a                        | n/a                                    | n/a     | converters.ConverterBenchmarks.time_circuit_to_dag(53, 8192) |
| -        | 15.3±0.05μs                | 7.85±0.2μs                             | 0.51    | converters.ConverterBenchmarks.time_circuit_to_dag(1, 8)     |
| -        | 26.1±0.1μs                 | 11.7±0.03μs                            | 0.45    | converters.ConverterBenchmarks.time_circuit_to_dag(2, 8)     |
| -        | 147±2μs                    | 61.3±0.3μs                             | 0.42    | converters.ConverterBenchmarks.time_circuit_to_dag(14, 8)    |
| -        | 222±5μs                    | 93.4±0.4μs                             | 0.42    | converters.ConverterBenchmarks.time_circuit_to_dag(20, 8)    |
| -        | 81.7±0.8μs                 | 33.2±0.2μs                             | 0.41    | converters.ConverterBenchmarks.time_circuit_to_dag(8, 8)     |
| -        | 50.0±0.1μs                 | 20.2±0.07μs                            | 0.40    | converters.ConverterBenchmarks.time_circuit_to_dag(5, 8)     |
| -        | 346±1μs                    | 130±0.4μs                              | 0.38    | converters.ConverterBenchmarks.time_circuit_to_dag(32, 8)    |
| -        | 703±4μs                    | 271±0.7μs                              | 0.38    | converters.ConverterBenchmarks.time_circuit_to_dag(53, 8)    |
| -        | 9.38±0.08ms                | 3.23±0.02ms                            | 0.34    | converters.ConverterBenchmarks.time_circuit_to_dag(53, 128)  |
| -        | 4.38±0.02ms                | 1.38±0.01ms                            | 0.32    | converters.ConverterBenchmarks.time_circuit_to_dag(32, 128)  |
| -        | 2.46±0.02ms                | 719±3μs                                | 0.29    | converters.ConverterBenchmarks.time_circuit_to_dag(20, 128)  |
| -        | 1.57±0.01ms                | 436±4μs                                | 0.28    | converters.ConverterBenchmarks.time_circuit_to_dag(14, 128)  |
| -        | 145±4μs                    | 38.7±0.3μs                             | 0.27    | converters.ConverterBenchmarks.time_circuit_to_dag(1, 128)   |
| -        | 24.4±0.1ms                 | 6.38±0.2ms                             | 0.26    | converters.ConverterBenchmarks.time_circuit_to_dag(14, 2048) |
| -        | 240±1μs                    | 63.5±0.4μs                             | 0.26    | converters.ConverterBenchmarks.time_circuit_to_dag(2, 128)   |
| -        | 867±1μs                    | 226±1μs                                | 0.26    | converters.ConverterBenchmarks.time_circuit_to_dag(8, 128)   |
| -        | 531±0.6μs                  | 129±0.3μs                              | 0.24    | converters.ConverterBenchmarks.time_circuit_to_dag(5, 128)   |
| -        | 52.1±0.3ms                 | 11.9±0.2ms                             | 0.23    | converters.ConverterBenchmarks.time_circuit_to_dag(8, 8192)  |
| -        | 2.13±0.04ms                | 446±4μs                                | 0.21    | converters.ConverterBenchmarks.time_circuit_to_dag(1, 2048)  |
| -        | 8.46±0.2ms                 | 1.76±0.04ms                            | 0.21    | converters.ConverterBenchmarks.time_circuit_to_dag(1, 8192)  |
| -        | 13.1±0.08ms                | 2.78±0.01ms                            | 0.21    | converters.ConverterBenchmarks.time_circuit_to_dag(8, 2048)  |
| -        | 3.54±0.06ms                | 718±2μs                                | 0.20    | converters.ConverterBenchmarks.time_circuit_to_dag(2, 2048)  |
| -        | 13.8±0.07ms                | 2.71±0.02ms                            | 0.20    | converters.ConverterBenchmarks.time_circuit_to_dag(2, 8192)  |
| -        | 7.86±0.06ms                | 1.55±0.01ms                            | 0.20    | converters.ConverterBenchmarks.time_circuit_to_dag(5, 2048)  |
| -        | 31.6±0.2ms                 | 6.44±0.03ms                            | 0.20    | converters.ConverterBenchmarks.time_circuit_to_dag(5, 8192)  |

PERFORMANCE INCREASED.
```

crates/circuit/src/dag_circuit.rs
```diff
@@ -6448,7 +6450,8 @@ impl DAGCircuit {
                 });
                 clbit_last_nodes
                     .entry(clbit)
-                    .and_modify(|val| *val = new_node);
+                    .and_modify(|val| *val = new_node)
+                    .or_insert(new_node);
```
mtreinish: Was this a bug in `extend()` that you didn't catch until now?

crates/circuit/src/dag_circuit.rs
```diff
@@ -6501,6 +6509,184 @@ impl DAGCircuit {
...
+        qc: QuantumCircuitData,
+        copy_op: bool,
+        qubit_order: Option<Vec<Bound<PyAny>>>,
+        clbit_order: Option<Vec<Bound<PyAny>>>,
+    ) -> PyResult<DAGCircuit> {
+        // Extract necessary attributes
+        let qc_data = qc.data;
+        let num_qubits = qc_data.num_qubits();
+        let num_clbits = qc_data.num_clbits();
+        let num_ops = qc_data.__len__();
```
mtreinish, line 6613: Heh, I feel like we should have a native `len()` method to `CircuitData`. Not for this PR of course, it's just a bit weird to see this in rust code.

crates/circuit/src/dag_circuit.rs
```diff
@@ -6501,6 +6509,184 @@ impl DAGCircuit {
...
+                    )));
+                }
+                qubit_map.insert(new_dag.add_qubit_unchecked(py, &qubit)?, qubit);
+            }
+        } else {
+            for qubit in qc_data.qubits().bits() {
+                let bound = qubit.clone_ref(py).into_bound(py);
+                qubit_map.insert(new_dag.add_qubit_unchecked(py, qubit.bind(py))?, bound);
+            }
+        }
```
mtreinish: I'm pretty sure you could do this as an `iter().map().collect()` instead. Something like:
```suggestion
        let qubit_map: HashMap<Qubit, Bound<PyAny>> = if let Some(qubit_ordering) = qubit_order {
            qubit_ordering.iter().map(|qubit| {
                if new_dag.qubits.find(&qubit).is_some() {
                    Err(DAGCircuitError::new_err(format!(
                        "duplicate qubits {}",
                        &qubit
                    )))
                } else {
                    Ok((new_dag.add_qubit_unchecked(py, &qubit)?, qubit))
                }
            }
        } else {
            qc_data.qubits().bits().map(|qubit| {
                let bound = qubit.clone_ref(py).into_bound(py);
                Ok((new_dag.add_qubit_unchecked(py, qubit.bind(py))?, bound))
            })
        }?;
```

(completely untested I wrote this in the webui). If this works (I'm not sure it will) then using this both `qubit_map` and `clbit_map` would probably be slightly more efficient, but likely not measurably.

crates/circuit/src/dag_circuit.rs
```diff
@@ -6501,6 +6509,184 @@ impl DAGCircuit {
...
+        // Collect the re-mapped indices for qubits
+        let remapped_qubits: HashMap<Qubit, Qubit> = qubit_map
+            .into_iter()
+            .map(|(index, bit)| (qc_data.qubits().find(&bit).unwrap(), index))
+            .collect();
+        // Collect the re-mapped indices for clbits
+        let remapped_clbits: HashMap<Clbit, Clbit> = clbit_map
+            .into_iter()
+            .map(|(index, bit)| (qc_data.clbits().find(&bit).unwrap(), index))
+            .collect();
```
mtreinish: I'm a bit concerned about this and the hashmap above in the default no order arguments being specified. We're spending a fair amount of effort building mappings of basically `{0: 0, 1:1, ..., n: n}` and we should just short circuit that path without any allocations as this becomes unnecessary overhead in those cases. Also is there a reason to not use a `Vec` here for these two? The `Qubit` and `Clbit` are indices into a vec, so they always need to be `0..n`?

crates/circuit/src/lib.rs
```diff
@@ -80,6 +81,17 @@ impl From<Clbit> for BitType {
...
+#[inline(always)]
+#[doc(hidden)]
+fn add_submodule<F>(m: &Bound<PyModule>, constructor: F, name: &str) -> PyResult<()>
+where
+    F: FnOnce(&Bound<PyModule>) -> PyResult<()>,
+{
+    let new_mod = PyModule::new_bound(m.py(), name)?;
+    constructor(&new_mod)?;
+    m.add_submodule(&new_mod)
+}
```
mtreinish: Do we really need a submodule for this? I'm totally fine if you wanted to add this as `qiskit._accelerate.circuit.circuit_to_dag` and eventually as `dag_to_circuit()` too. I don't think we're going to have any other converter functions in rust in the near future. But if we did we can address this then.

mtreinish (commented):
Overall this looks great, I'm very eager to have this as besides being a speedup for conversion it gives us a native method for working with control flow block in the transpiler with minimal python overhead. My main concern is about the extra overhead with building the qubit mapping stuff when we don't have a mapping. I get from a code cleanliness perspective it's a lot easier to have an identity map. But I'm worried about the performance overhead to handle the mapping when there isn't any both in building the maps and then also adding hash lookups on every qubit..

crates/circuit/src/dag_circuit.rs
mtreinish: Oh also the same comment about maybe building a vec here instead of hashmap. When I see `Qubit` as a key my first thought is always "Can we use a vec here instead?"

crates/circuit/src/dag_circuit.rs
raynelfss: This one is not necessarily a bug since nodes should be there already. I believe the `.or_insert()` could be removed with no effect. I had it for debugging. There is an error in `extend()` though that I believe I put comments in for.

crates/circuit/src/dag_circuit.rs
raynelfss: I will try to build a `Vec` instead. The one issue I kept finding is that some circuits would have indices that aren't necessarily in order (Specifically with certain `ControlFlowOp` blocks). I will test it out and update it.

crates/circuit/src/lib.rs
raynelfss: Yeah I was just planning ahead here since both `CircuitData` as a representation of `QuantumCircuit` and `DAGCircuit` both live in rust. We could have them be part of the same module, just to keep things a bit sorted. Although, I do understand how awkward it is to type:
```py
from qiskit._accelerate.circuit.converters import circuit_to_dag
```
It would be cleaner to have both converters be part of a converters module but, again, if it is a problem I can easily remove it.

raynelfss:
```bash
All benchmarks:

| Change   | Before [3d4bab2b] <main>   | After [94f8ab42] <rusty-circ-to-dag>   | Ratio   | Benchmark (Parameter)                                        |
|----------|----------------------------|----------------------------------------|---------|--------------------------------------------------------------|
|          | n/a                        | n/a                                    | n/a     | converters.ConverterBenchmarks.time_circuit_to_dag(14, 8192) |
|          | n/a                        | n/a                                    | n/a     | converters.ConverterBenchmarks.time_circuit_to_dag(20, 2048) |
|          | n/a                        | n/a                                    | n/a     | converters.ConverterBenchmarks.time_circuit_to_dag(20, 8192) |
|          | n/a                        | n/a                                    | n/a     | converters.ConverterBenchmarks.time_circuit_to_dag(32, 2048) |
|          | n/a                        | n/a                                    | n/a     | converters.ConverterBenchmarks.time_circuit_to_dag(32, 8192) |
|          | n/a                        | n/a                                    | n/a     | converters.ConverterBenchmarks.time_circuit_to_dag(53, 2048) |
|          | n/a                        | n/a                                    | n/a     | converters.ConverterBenchmarks.time_circuit_to_dag(53, 8192) |
| -        | 15.0±0.4μs                 | 7.02±0.02μs                            | 0.47    | converters.ConverterBenchmarks.time_circuit_to_dag(1, 8)     |
| -        | 25.6±0.2μs                 | 10.9±0.09μs                            | 0.43    | converters.ConverterBenchmarks.time_circuit_to_dag(2, 8)     |
| -        | 217±0.6μs                  | 88.0±0.2μs                             | 0.41    | converters.ConverterBenchmarks.time_circuit_to_dag(20, 8)    |
| -        | 82.3±2μs                   | 33.0±2μs                               | 0.40    | converters.ConverterBenchmarks.time_circuit_to_dag(8, 8)     |
| -        | 145±0.7μs                  | 56.4±0.4μs                             | 0.39    | converters.ConverterBenchmarks.time_circuit_to_dag(14, 8)    |
| -        | 49.5±0.4μs                 | 18.8±0.3μs                             | 0.38    | converters.ConverterBenchmarks.time_circuit_to_dag(5, 8)     |
| -        | 342±10μs                   | 123±1μs                                | 0.36    | converters.ConverterBenchmarks.time_circuit_to_dag(32, 8)    |
| -        | 711±1μs                    | 259±5μs                                | 0.36    | converters.ConverterBenchmarks.time_circuit_to_dag(53, 8)    |
| -        | 9.27±0.07ms                | 2.95±0.04ms                            | 0.32    | converters.ConverterBenchmarks.time_circuit_to_dag(53, 128)  |
| -        | 4.39±0.03ms                | 1.34±0.02ms                            | 0.31    | converters.ConverterBenchmarks.time_circuit_to_dag(32, 128)  |
| -        | 2.43±0.01ms                | 689±3μs                                | 0.28    | converters.ConverterBenchmarks.time_circuit_to_dag(20, 128)  |
| -        | 1.58±0.02ms                | 417±1μs                                | 0.26    | converters.ConverterBenchmarks.time_circuit_to_dag(14, 128)  |
| -        | 145±1μs                    | 34.8±0.2μs                             | 0.24    | converters.ConverterBenchmarks.time_circuit_to_dag(1, 128)   |
| -        | 24.1±0.1ms                 | 5.68±0.03ms                            | 0.24    | converters.ConverterBenchmarks.time_circuit_to_dag(14, 2048) |
| -        | 236±2μs                    | 57.1±0.3μs                             | 0.24    | converters.ConverterBenchmarks.time_circuit_to_dag(2, 128)   |
| -        | 883±6μs                    | 216±3μs                                | 0.24    | converters.ConverterBenchmarks.time_circuit_to_dag(8, 128)   |
| -        | 529±3μs                    | 119±0.5μs                              | 0.22    | converters.ConverterBenchmarks.time_circuit_to_dag(5, 128)   |
| -        | 51.6±0.4ms                 | 10.9±0.2ms                             | 0.21    | converters.ConverterBenchmarks.time_circuit_to_dag(8, 8192)  |
| -        | 8.17±0.05ms                | 1.61±0.03ms                            | 0.20    | converters.ConverterBenchmarks.time_circuit_to_dag(1, 8192)  |
| -        | 13.3±0.2ms                 | 2.66±0.04ms                            | 0.20    | converters.ConverterBenchmarks.time_circuit_to_dag(8, 2048)  |
| -        | 2.06±0.01ms                | 395±6μs                                | 0.19    | converters.ConverterBenchmarks.time_circuit_to_dag(1, 2048)  |
| -        | 31.6±0.3ms                 | 5.98±0.07ms                            | 0.19    | converters.ConverterBenchmarks.time_circuit_to_dag(5, 8192)  |
| -        | 3.51±0.03ms                | 648±4μs                                | 0.18    | converters.ConverterBenchmarks.time_circuit_to_dag(2, 2048)  |
| -        | 13.8±0.2ms                 | 2.42±0.02ms                            | 0.18    | converters.ConverterBenchmarks.time_circuit_to_dag(2, 8192)  |
| -        | 7.84±0.08ms                | 1.41±0.01ms                            | 0.18    | converters.ConverterBenchmarks.time_circuit_to_dag(5, 2048)  |

PERFORMANCE INCREASED.
```
After @mtreinish 's suggestions there was a slight performance increase.

crates/circuit/src/lib.rs
raynelfss: I ended up adding it through `py_ext` and the new path is exposed as `qiskit._accelerate.converters` to avoid adding all the extra code.

crates/circuit/src/dag_circuit.rs
raynelfss: I removed this second re-mapping, and made everything happen in the first one, so that the outcome is optional if no ordering is provided.

crates/circuit/src/dag_circuit.rs
```diff
@@ -6590,6 +6597,207 @@ impl DAGCircuit {
...
+                    params: instr.params.clone(),
+                    extra_attrs: instr.extra_attrs.clone(),
+                    #[cfg(feature = "cache_pygates")]
+                    py_op: OnceCell::new(),
+                })
+            })
+            .collect::<PyResult<Vec<_>>>()?;
+
+        // Finally add all the instructions back
+        new_dag.extend(py, instructions)?;
```
mtreinish, line 6777: Is the only reason you're collecting into a vec here because the `deepcopy()` call is fallible. I was kind of excited to have this all work with iterators in rust so we avoided a big allocation like this. I don't think we want to mask the error in python if a deepcopy fails, that is important for users to get the python exception. While it ends up being a big block of duplicated code, do you think it would make sense to move the `copy_op` conditional outside the iterator so you do something like:

```rust
if copy_op {
   let instructions = iterator.... .collect::<PyResult<Vec<_>>>()
} else {
   new_dag.extend(py, iterator)
}
```

where iterator is basically this entire chain just with or without the pyresult depending on whether we call `clone()` or `deepcopy`.

mtreinish (commented):
Overall this looks great, thanks for the fast update. Avoiding the extra overhead of an identity mapping definitely is worth it and the code looks much simpler now that everything is in a single collection. I just had one inline question/suggestion that I think could improve performance a bit more at the cost of some code duplication, but I'm curious what you think.

crates/circuit/src/dag_circuit.rs
raynelfss: The other reason we're re-collecting is to re-intern the qargs. But I also believe this could be avoided. After all the only requirement here is that whatever we send to `extend()` can turn into an iterator of PackedInstruction. I think we can avoid the allocation by not collecting the `Vec` but we would still have to re-intern things since we're not copying the interners over.

Edit: This earlier suggestion may not work due to the iterator mutably borrowing the dag while the method already holds right to mutate. :/

mtreinish: I think we can do this without re-interning at all in the case when we're not reordering any qubits or clbits. Since the interners should be the same for the dag and circuit.

But, lets save this as a potential followup then. The gains from doing something like this would be be measurable but even with a collection into a Vec this has a good improvement in speed and the logic to figure out how to do this is probably will probably be pretty verbose to handle all the edge cases around whether we are reordering any bits or not, or we need to return a python exception from deepcopy or not. I'll open an issue after this merges so we can track it as a potential future optimization.

crates/circuit/src/dag_circuit.rs
```diff
@@ -6590,6 +6597,207 @@ impl DAGCircuit {
...
+        let circ = QuantumCircuitData {
+            data: circuit_data,
+            name: None,
+            calibrations: None,
+            metadata: None,
+            qregs: None,
+            cregs: None,
+            input_vars: Vec::with_capacity(0),
+            captured_vars: Vec::with_capacity(0),
+            declared_vars: Vec::with_capacity(0),
```
mtreinish: Is there a reason you didn't just use `new()`?
```suggestion
            input_vars: Vec::new(),
            captured_vars: Vec::new(),
            declared_vars: Vec::new(),
```

mtreinish (dismissed):
One small nit/question inline but this LGTM thanks for doing this. If you don't want to update the nit feel free to enqueue this for merge.

mtreinish (approved):

PR #13040, opened 2024-08-26
Fully port `Depth` and `Size` passes to Rust

### LINKED ISSUE #12272: Port `Depth` to Rust

### LINKED ISSUE #12273: Port `Size` to Rust

kevinhartman:
### Summary
This has a bit of a click-bait title since they're technically already ported (the calls to `DAGCircuit::depth` and `DAGCircuit::size` made by the Python passes are already in Rust). But, this PR avoids an `isinstance` call that was being done for every node in the DAG on calls to these methods when the `recurse` parameter was specified as `true`, which should resolve the performance issues we were seeing with these passes.

I've also done this for `DAGCircuit::count_ops`.

### Details

Resolves #12272, #12273

### DIFF
--- a/crates/circuit/src/dag_circuit.rs
+++ b/crates/circuit/src/dag_circuit.rs
@@ -4568,16 +4568,19 @@ def _format(operand):
                         .or_insert(*value);
                 }
                 let circuit_to_dag = imports::CIRCUIT_TO_DAG.get_bound(py);
-                for node in dag.py_op_nodes(
-                    py,
-                    Some(imports::CONTROL_FLOW_OP.get_bound(py).downcast()?),
-                    true,
-                )? {
-                    let raw_blocks = node.getattr(py, "op")?.getattr(py, "blocks")?;
-                    let blocks: &Bound<PyTuple> = raw_blocks.downcast_bound::<PyTuple>(py)?;
-                    for block in blocks.iter() {
-                        let inner_dag: &DAGCircuit =
-                            &circuit_to_dag.call1((block.clone(),))?.extract()?;
+                for node in dag.dag.node_weights() {
+                    let NodeType::Operation(node) = node else {
+                        continue;
+                    };
+                    if !node.op.control_flow() {
+                        continue;
+                    }
+                    let OperationRef::Instruction(inst) = node.op.view() else {
+                        panic!("control flow op must be an instruction")
+                    };
+                    let blocks = inst.instruction.bind(py).getattr("blocks")?;
+                    for block in blocks.iter()? {
+                        let inner_dag: &DAGCircuit = &circuit_to_dag.call1((block?,))?.extract()?;
                         inner(py, inner_dag, counts)?;
                     }
                 }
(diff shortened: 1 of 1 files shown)


crates/circuit/src/dag_circuit.rs
```diff
@@ -2239,34 +2240,33 @@ def _format(operand):
...
+                } else {
+                    1
+                };
+                if weight == 0 {
+                    node_lookup.insert(node_index, 0);
+                } else {
+                    let raw_blocks = inst_bound.getattr("blocks")?;
+                    let blocks = raw_blocks.downcast::<PyTuple>()?;
+                    let mut block_weights: Vec<usize> = Vec::with_capacity(blocks.len());
+                    for block in blocks.iter() {
```
mtreinish: Do we need the PyTuple downcast here? Should we be able to just call `len()` and `iter()` on a `Bound<PyAny>`. I think the only difference is their fallible in `PyAnyMethods` but not in `PyTupleMethods`.

crates/circuit/src/dag_circuit.rs
```diff
@@ -2239,34 +2240,33 @@ def _format(operand):
         Ok(if recurse {
```
mtreinish: I wonder if we want an `&& CONTROL_FLOW_OP_NAMES.iter().any(|x| self.op_names.contains_key(&x.to_string()))` on this condition so we don't bother with the for loop if we know there are no control flow instructions. The same question applies to `size()` as well.

mtreinish (commented):
This LGTM, thanks for fixing this. Just a couple small inline comments.

crates/circuit/src/dag_circuit.rs
kevinhartman: Good idea. Done in 80ff154cd81771dc546e1e0efe18e6aa6e221c36.

crates/circuit/src/dag_circuit.rs
kevinhartman: That's a good idea, and allowed me to make the code a bit cleaner across both methods. I added a helper as well, which is private to `DAGCircuit` for now.

Done in af244c2f44c2268a7ca92c9419f7a7a80db0c06e.

crates/circuit/src/dag_circuit.rs
```diff
@@ -2153,53 +2153,46 @@ def _format(operand):
...
-                    "Size with control flow is ambiguous.",
-                    " You may use `recurse=True` to get a result",
-                    " but see this method's documentation for the meaning of this."
-                )));
-            }
+        if !self.has_control_flow() {
             return Ok(length);
         }
+        if !recurse {
+            return Err(DAGCircuitError::new_err(concat!(
```
mtreinish, line 2160: Oh, I forgot about `concat!` there are a couple of long error messages in this module we should probably break up using it.

mtreinish (approved):
LGTM, thanks for doing this.

PR #13042, opened 2024-08-26
Oxidize CheckGateDirection

### LINKED ISSUE #12256: Port `CheckGateDirection` to Rust

eliarbel:
### Summary
This PR ports the `CheckGateDirection` analysis pass to Rust. 

### Details and comments
This PR includes:

* Until we have a Rust version of `CouplingMap`, the Python code converts the given coupling map to a set of  edge pairs and pass it as `PySet` to the Rust code. 
* The pass relies on the `DAGCircuit.qubits` registry for mapping instruction qubit args to qubit indices when checking gate direction against either a coupling map edge set or a target.
* An unrelated change in `dag_circuit.rs`: making a pure Rust-friendly version of `two_qubits_op` 

Close #12256


eliarbel:
Some benchmark data, using circuits generated like this:

```python
tqc = generate_preset_pass_manager(1, coupling_map=CouplingMap.from_heavy_hex(7))
    .run(qc)
```
and this:
```python
tqc = generate_preset_pass_manager(1, GenericBackendV2(qubits))
    .run(qc)
```

where: 
```python
 qubits = 100
 qc = random_circuit(qubits, qubits, seed=12234)
 ```
 
(note that `DAGCircuit.count_ops()` gives different stats for the coupling-based vs. target-bases circuits but this is irrelevant for this check)

And results are measured like this:
```python
%%timeit
check_direction_pass.run(dag)
```

|Implementation|CouplingMap|Target|
|-------------|----------|------|
|Python|1.48 s ± 46.5 ms|641 ms ± 11.4 ms|
|Rust|130 ms ± 7.02 ms|131 ms ± 5.23|

So ~5-11X improvement in this check by moving to Rust. Note that CouplingMap is (still) in Python while Target is in Rust, which may explain the run-time differences in Python and the speed-up differences between the target vs coupling map cases.

ElePT (commented):
It might be helpful for benchmarking to generate a target with the same coupling map as the "coupling_map" case, as `GenericBackendV2` generates a target with all-to-all connectivity by default. You could do this by running:

`tqc = generate_preset_pass_manager(1, GenericBackendV2(qubits, coupling_map=CouplingMap.from_heavy_hex(7))).run(qc)`

eliarbel:
Thanks @ElePT . There are 2 paths for this pass: one is based on coupling map and one on target. So I think it doesn't really matter for this benchmarking to have the same coupling constraints, as long as we use the same constraints in each path separately both in Python and Rust . 

Anyway, FWIW, here is another comparison, this time using the same coupling constraints, basis gates and traspiler seed in all 4 runs (python/rust X coupling/target). `DAGCircuit.count_ops()` is now the same for all: `{'rz': 27177, 'sx': 6193, 'cx': 161650, 'x': 70}`: 

|Implementation|CouplingMap|Target|
|-------------|----------|------|
|Python|3.2 s ± 537 ms|3.21 s ± 104 ms|
|Rust|263 ms ± 26.2|213 ms ± 4.2 ms|


I see that I need to resolve conflicts in `dag_circuit.rs` now that Jake's PR #13033 is merged; Will do soon

crates/circuit/src/dag_circuit.rs
```diff
@@ -5788,6 +5781,19 @@ impl DAGCircuit {
         }
     }
 
+    /// Return an iterator of 2 qubit operations. Ignore directives like snapshot and barrier.
+    pub fn two_qubit_ops<'a>(&'a self) -> Box<dyn Iterator<Item = NodeIndex> + 'a> {
+        Box::new(self.op_nodes(false).filter(|index| {
```
mtreinish: Does this need the `Box` wrapping it? I'd have assumed you could have done something like:

```rust
pub fn two_qubit_ops(&self) -> impl Iterator<Item = NodeIndex> + '_ {
```

for the signature and avoid the box usage here.

crates/circuit/src/dag_circuit.rs
```diff
@@ -6157,6 +6163,11 @@ impl DAGCircuit {
         }
         Ok(())
     }
+
+    /// Get the instruction qubits interned at the given index
+    pub fn get_inst_qubits(&self, index: Interned<[Qubit]>) -> &[Qubit] {
+        self.qargs_interner.get(index)
+    }
```
mtreinish: There is a `get_qargs` (and `get_cargs` for classical bits) method already you can use for this.

crates/accelerate/src/gate_direction.rs
```diff
@@ -0,0 +1,159 @@
...
+}
+
+// The main routine for checking gate directionality
+fn check_gate_direction<T>(py: Python, dag: &DAGCircuit, gate_complies: &T) -> PyResult<bool>
+where
+    T: Fn(&DAGCircuit, &PackedInstruction, &[Qubit]) -> bool,
+{
+    for node in dag.op_nodes(false) {
+        let NodeType::Operation(packed_inst) = &dag.dag[node] else {
+            return Err(DAGCircuitError::new_err("PackedInstruction is expected"));
```
mtreinish: This gets returned to python and a python user doesn't know what `PackedInstruction` is. I'd say a `panic!()` here is actually probably fine because it's a rust domain bug if we get a non-operation here as `op_nodes` should be filtering it for us. But that being said I'd say if you did want a Python error return I'd change this to be:

```suggestion
            return Err(DAGCircuitError::new_err("An operation node was expected"));
```

crates/accelerate/src/gate_direction.rs
```diff
@@ -0,0 +1,159 @@
...
+// Required for checking control flow instruction which are represented in blocks (circuits)
+// and converted to DAGCircuit with possibly different qargs than the original one.
+fn map_qubit(py: Python, orig_dag: &DAGCircuit, curr_dag: &DAGCircuit, qubit: Qubit) -> Qubit {
+    let qubit = curr_dag
+        .qubits
+        .get(qubit)
+        .expect("Qubit in curr_dag")
+        .bind(py);
+    orig_dag.qubits.find(qubit).expect("Qubit in orig_dag")
+}
```
mtreinish: I feel like we can do this mapping without needing Python since we should be mapping the indices right? I'm not sure if it's exactly the same, but for `CheckMap` which is a very similar pass I did this with a `Vec<Qubit>`: https://github.com/Qiskit/qiskit/pull/13030/files#diff-6faf9eee12b142b944a69438ee1e1d11aeae4ab022c8a8e6551a5215a3762c6fR56-R64

crates/accelerate/src/gate_direction.rs
```diff
@@ -0,0 +1,159 @@
...
+///     coupling_edges: set of edge pairs representing a directed coupling map, against which gate directionality is checked
+///
+/// Returns:
+///     true iff all two-qubit gates comply with the coupling constraints
+#[pyfunction]
+#[pyo3(name = "check_gate_direction_coupling")]
+fn py_check_with_coupling_map(
+    py: Python,
+    dag: &DAGCircuit,
+    coupling_edges: &Bound<PySet>,
```
mtreinish: Is using a `PySet` here the best choice? I would have expected using a rust space `HashSet` would be a better choice here. We have a one time conversion overhead, but I'd have expected that to be minimal especially compared to the performance improvement from a rust lookup.

crates/accelerate/src/gate_direction.rs
```diff
@@ -0,0 +1,159 @@
...
+///     dag: the DAGCircuit to analyze
+///
+///     target: the Target against which gate directionality compliance is checked
+///
+/// Returns:
+///     true iff all two-qubit gates comply with the target's coupling constraints
+#[pyfunction]
+#[pyo3(name = "check_gate_direction_target")]
+fn py_check_with_target(py: Python, dag: &DAGCircuit, target: &Bound<Target>) -> PyResult<bool> {
+    let target = target.borrow();
```
mtreinish: Could you just do?:

```suggestion
fn py_check_with_target(py: Python, dag: &DAGCircuit, target: &Target) -> PyResult<bool> {
```

crates/accelerate/src/gate_direction.rs
```diff
@@ -0,0 +1,159 @@
...
+            let mut qargs = Qargs::new();
+
+            qargs.push(PhysicalQubit::new(
+                map_qubit(py, dag, curr_dag, op_args[0]).0,
+            ));
+            qargs.push(PhysicalQubit::new(
+                map_qubit(py, dag, curr_dag, op_args[1]).0,
+            ));
+
+            target.instruction_supported(inst.op.name(), Some(&qargs))
```
mtreinish: I'm not a fan that `instruction_supported` requires a `SmallVec` reference, I feel like it could work with a slice input. But we can save that for a follow up. But given these constraints I think you can do or something like this to avoid the new/push since `Qargs` is just a type alias:

```suggestion
            let qargs = smallvec![
                PhysicalQubit::new(
                    map_qubit(py, dag, curr_dag, op_args[0]).0,
                ),
                PhysicalQubit::new(
                    map_qubit(py, dag, curr_dag, op_args[1]).0,
                ),
            ];
            target.instruction_supported(inst.op.name(), Some(&qargs))
```

crates/accelerate/src/gate_direction.rs
```diff
@@ -0,0 +1,159 @@
...
+where
+    T: Fn(&DAGCircuit, &PackedInstruction, &[Qubit]) -> bool,
+{
+    let circuit_to_dag = imports::CIRCUIT_TO_DAG.get_bound(py); // TODO: Take out of the recursion
+    let py_inst = py_inst.instruction.bind(py);
+
+    let raw_blocks = py_inst.getattr("blocks")?;
+    let blocks = raw_blocks.downcast::<PyTuple>()?;
+
+    for block in blocks.iter() {
```
mtreinish: You can avoid the downcast here since `iter()` works fine on a `Bound<PyAny>` the only difference is that it is fallible.

```suggestion
    for block in blocks.iter()? {
```

mtreinish (commented):
This looks good, thanks for doing this. I left a few inline comments, but at a high level this has everything needed.

crates/circuit/src/dag_circuit.rs
eliarbel: No need for Box really. Removed

crates/circuit/src/dag_circuit.rs
eliarbel: Yes, switched to the merged one.

crates/accelerate/src/gate_direction.rs
eliarbel: You're right. Switched to using `panic!`

crates/accelerate/src/gate_direction.rs
eliarbel: This was a very good suggestion, thanks! It forced me to really get my head around how `DAGCircuit` nesting is done for control flow blocks w.r.t qargs indices mapping. Changed it use a simpler, Rust-only mapping approach. It also facilitated a nice overall simplification of the code.

crates/accelerate/src/gate_direction.rs
eliarbel: Done.

crates/accelerate/src/gate_direction.rs
eliarbel: Done.

crates/accelerate/src/gate_direction.rs
eliarbel: Good point. I switched to using `HashSet<(Qubit, Qubit)>`. This required adding `FromPyObject` implementation for `Qubit` which I'd appreciate a review of. 
I didn't end up using `PhysicalQubit` after all since `DAGCircuit` doesn't use it and I wasn't clear about whether or not it's safe or effective to force-cast the `DAGCircuit` indices in pass implementation. Let me know if there is an elegant way for that.

eliarbel:
Thanks @mtreinish for the review and very helpful comments! 

Here is an updated benchmark result. The `Target` path became slower for some reason. I didn't explore why, I'd defer this for a later phase.

|Implementation|CouplingMap|Target|
|-------------|----------|------|
|Python|3.19 s ± 450|3.18 s ± 221|
|Rust|137 ms ± 6.81 ms|567 ms ± 15.6|

crates/accelerate/src/gate_direction.rs
```diff
@@ -0,0 +1,148 @@
...
+///     true iff all two-qubit gates comply with the coupling constraints
+#[pyfunction]
+#[pyo3(name = "check_gate_direction_coupling")]
+fn py_check_with_coupling_map(
+    py: Python,
+    dag: &DAGCircuit,
+    coupling_edges: HashSet<(Qubit, Qubit)>,
+) -> PyResult<bool> {
+    let coupling_map_check = |_: &PackedInstruction, op_args: &[Qubit]| -> bool {
+        coupling_edges.contains(&(op_args[0], op_args[1]))
```
mtreinish: If you change the input type here to a fixed size array you can get away without having to build the tuple. I don't think functionally it'll make a difference but it's more concise.

```suggestion
    coupling_edges: HashSet<[Qubit; 2]>,
) -> PyResult<bool> {
    let coupling_map_check = |_: &PackedInstruction, op_args: &[Qubit]| -> bool {
        coupling_edges.contains(op_args)
```

crates/circuit/src/lib.rs
```diff
@@ -37,6 +37,12 @@ pub struct Qubit(pub BitType);
 #[derive(Copy, Clone, Debug, Hash, Ord, PartialOrd, Eq, PartialEq)]
 pub struct Clbit(pub BitType);
 
+impl FromPyObject<'_> for Qubit {
+    fn extract_bound(ob: &Bound<PyAny>) -> PyResult<Self> {
+        Ok(Self(ob.extract::<BitType>()?))
+    }
+}
```
mtreinish: You can remove this manual implementation and just add `FromPyObject` to the derive macro on the newtype definition.

crates/circuit/src/lib.rs
```diff
@@ -37,6 +37,12 @@ pub struct Qubit(pub BitType);
 #[derive(Copy, Clone, Debug, Hash, Ord, PartialOrd, Eq, PartialEq)]
 pub struct Clbit(pub BitType);
 
+impl FromPyObject<'_> for Qubit {
```
mtreinish: I'm a bit on the fence about this. I know we talked about it offline, but at the time I hadn't though about the wider implications. Mostly because there is a Python class `Qubit` and this doesn't go from python `Qubit` -> Rust `Qubit`. But on the other hand in contexts coming from Rust going through python this will make sense. Also for places we take in qubit indices this will make things slightly easier. I think I'm leaning with this still being ok because it just will error if you try to pass a python `Qubit` object to something expecting a rust integer based qubit.

crates/accelerate/src/gate_direction.rs
```diff
@@ -0,0 +1,148 @@
...
+    };
+
+    check_gate_direction(
+        py,
+        dag,
+        &coupling_map_check,
+        &(0..dag.num_qubits())
+            .map(|i| Qubit(i as BitType))
+            .collect::<Vec<Qubit>>(),
```
mtreinish: I feel like there is a path where we can avoid this vec collection because `i == i` here. We can probably change `check_gate_direction` to take an `Option<&[Qubit]>` and then just use the qubit directly if it's `None`.

mtreinish (commented):
This is looking good, I think it's basically ready now thanks for making the updates. I just have a few small inline comments. But after those are addressed I think we're good to merge this.

crates/circuit/src/lib.rs
eliarbel: That was my thinking too - it only works with integers so having the convenience to directly get a Qubit object from a Python integer would be nice.

crates/accelerate/src/gate_direction.rs
eliarbel: Done. It did make the code a bit less clean since now gate directionality check and control flow blocks recursive check require now a slightly different handling of the mapping array each. But exploiting the identity mapping is better for performance, even if just marginally, but anyway it's worth it.

crates/accelerate/src/gate_direction.rs
```diff
@@ -0,0 +1,147 @@
...
+                    } else {
+                        check_gate_direction(py, &inner_dag, gate_complies, Some(inst_qargs))?
+                    };
+
+                    if !block_ok {
+                        return Ok(false);
+                    }
+                }
+            }
+        } else if inst_qargs.len() == 2
```
mtreinish: This is technically a behavior change as you won't be checking 2 qubit non-control flow `Instruction` objects anymore and the previous python version did. I'm not sure if we should care or not because at least on real backends there aren't connectivity constraints on non-gates like this pass is trying to check, but I could conceive of a backend where it would matter. The simplest fix would be to just remove the `else` here and add a `continue` like this:

```suggestion
        if let OperationRef::Instruction(py_inst) = packed_inst.op.view() {
            if py_inst.control_flow() {
                let circuit_to_dag = imports::CIRCUIT_TO_DAG.get_bound(py); // TODO: Take out of the recursion
                let py_inst = py_inst.instruction.bind(py);

                for block in py_inst.getattr("blocks")?.iter()? {
                    let inner_dag: DAGCircuit = circuit_to_dag.call1((block?,))?.extract()?;

                    let block_ok = if let Some(mapping) = qubit_mapping {
                        let mapping = inst_qargs // Create a temp mapping for the recursive call
                            .iter()
                            .map(|q| mapping[q.0 as usize])
                            .collect::<Vec<Qubit>>();

                        check_gate_direction(py, &inner_dag, gate_complies, Some(&mapping))?
                    } else {
                        check_gate_direction(py, &inner_dag, gate_complies, Some(inst_qargs))?
                    };

                    if !block_ok {
                        return Ok(false);
                    }
                }
                continue;
            }
        } 
        if inst_qargs.len() == 2
```

eliarbel: I was assuming we can skip it, but ok, that's a fair point, better keep the check more general

mtreinish (approved):
This LGTM now, thanks for making all the updates.

PR #13045, opened 2024-08-27
Oxidize `PauliFeatureMap`

Cryoris:
### Summary

This PR oxidizes the `PauliFeatureMap` (and derivatives) and implements a first step in refactoring the circuit library, where structurally fixed circuits are moved to functions. As part of this, `PauliFeatureMap` & co. are pending deprecation.

### Details and comments

Some benchmark data for `reps=10` and different numbers of qubits `n`:
```
-- n = 10, speedup 8.07x
main: 0.008 +- 0.001
this: 0.001 +- 0.000

-- n = 20, speedup 5.38x
main: 0.028 +- 0.004
this: 0.005 +- 0.004

-- n = 50, speedup 6.28x
main: 0.174 +- 0.005
this: 0.028 +- 0.006

-- n = 100, speedup 6.69x
main: 0.743 +- 0.022
this: 0.111 +- 0.007

-- n = 200, speedup 7.14x
main: 3.365 +- 0.144
this: 0.471 +- 0.015
```

* The most time is spent in parameter multiplication, which could be further improved by ~~caching the expression across the repetitions~~ (edit: tried, was slower), or (bigger scale) have the parameter logic in Rust. 
* ~~If #12767 is merged, this will be handled in a follow up.~~ Edit: Already included in this, as it was a minor update only.


crates/circuit/src/operations.rs
```diff
@@ -2046,7 +2046,31 @@ fn multiply_param(param: &Param, mult: f64, py: Python) -> Param {
...
+                    .clone_ref(py)
+                    .call_method1(py, intern!(py, "__rmul__"), (lambda,))
+                    .expect("Parameter expression multiplication failed"),
+            )
+        }
+        _ => unreachable!("Unsupported multiplication."),
+    }
+}
+
+pub fn add_param(param: &Param, summand: f64, py: Python) -> Param {
```
eliarbel, line 2085: A small suggestion at this point: consider something like this instead to remove code duplication (the `call_method1` block and the recursion):
```rust
pub fn rmultiply_param(param1: Param, param2: Param, py: Python) -> Param {
    let (p1, p2) = match (&param1, &param2) {
        (Param::Float(theta), Param::Float(lambda)) => return Param::Float(theta * lambda),
        (Param::ParameterExpression(_), Param::Float(_)) => (param1, param2),
        (Param::Float(_), Param::ParameterExpression(_)) => (param2, param1),
        (Param::ParameterExpression(_), Param::ParameterExpression(_)) => (param1, param2),
        _ => unreachable!("Unsupported multiplication."),
    };

    let Param::ParameterExpression(op1) = p1 else {unreachable!("Param::ParameterExpression is assumed")};

    Param::ParameterExpression(
    op1
    .clone_ref(py)
    .call_method1(py, intern!(py, "__rmul__"), (p2,))
    .expect("Parameter expression multiplication failed")
    )
}
```

Cryoris: I refactored the function in 71ca90b7b 🙂 I applied a slightly different version that above, which simply calls `multiply_param(Param, f64)` for the two mixed cases of `ParameterExpression, Float`. So this even further simplifies the code and also gets rid of the duplication 👍🏻

crates/accelerate/src/circuit_library/pauli_feature_map.rs
```diff
@@ -0,0 +1,317 @@
...
+///                     ┌───┐┌───────┐┌───┐
+/// 0: ─────────────────┤ X ├┤ Rz(2) ├┤ X ├──────────────────
+///    ┌──────────┐┌───┐└─┬─┘└───────┘└─┬─┘┌───┐┌───────────┐
+/// 1: ┤ Rx(pi/2) ├┤ X ├──■─────────────■──┤ X ├┤ Rx(-pi/2) ├
+///    └──────────┘└─┬─┘                   └─┬─┘└───────────┘
+/// 2: ──────────────┼───────────────────────┼───────────────
+///     ┌───┐        │                       │  ┌───┐
+/// 3: ─┤ H ├────────■───────────────────────■──┤ H ├────────
+///     └───┘                                   └───┘
+fn pauli_evolution(
```
ShellyGarion, line 55: is this function related to the `PauliEvolutionGate`? if not, perhaps it's worth to change its name so that it won't be confusing

crates/accelerate/src/circuit_library/pauli_feature_map.rs
```diff
@@ -0,0 +1,317 @@
...
+///     parameters: A parameter vector with ``feature_dimension`` elements. Taken as input
+///         here to avoid a call to Python constructing the vector.
+///     reps: The number of repetitions of Hadamard + evolution layers.
+///     entanglement: The entanglement, given as Python string or None (defaults to "full").
+///     paulis: The Pauli strings as list of strings or None (default to ["z", "zz"]).
+///     alpha: A float multiplier for rotation angles.
+///     insert_barriers: Whether to insert barriers in between the Hadamard and evolution layers.
+///     data_map_func: An accumulation function that takes as input a vector of parameters the
+///         current gate acts on and returns a scalar.
+///
```
ShellyGarion, line 141: `parameter_prefix` appears in the python code but not here. `parameters` appear here and not in the python code.

ShellyGarion: it's a bit confusing that both the python and rust functions have the same name

ShellyGarion (commented):
Very nice PR! I only have a few questions.

crates/accelerate/src/circuit_library/pauli_feature_map.rs
Cryoris: Yes, they do the same basic evolution, we should be able to use the same code for `PauliEvolutionGate` once we port it to Rust 🙂

crates/accelerate/src/circuit_library/pauli_feature_map.rs
Cryoris: Yeah, the parameters are constructed Python-side because (1) we avoid calling Python once we're in Rust space and (2) I'm not even sure we can create `ParameterVector`s Rust side 🤔 

And I gave them the same name because they do the same thing 🙂 I know we often have `function_inner` on the Rust side but IMO that makes it look like it's some internal-only function -- even though they are perfectly fine to be called from the Rust interface 😄 but we can discuss this, if you want to change it 👍🏻

test/python/circuit/library/test_pauli_feature_map.py
```diff
@@ -295,5 +304,262 @@ def test_entanglement_not_specified(self):
...
+        ref = QuantumCircuit(1)
+        ref.h(0)
+        ref.p(alpha * encoding.parameters[0], 0)
+
+        self.assertEqual(ref, encoding)
+
+    def test_zzfeaturemap_raises_if_too_small(self):
+        """Test the ``ZZFeatureMap`` raises an error if the number of qubits is smaller than 2."""
+        with self.assertRaises(QiskitError):
+            _ = zz_feature_map(1)
```
alexanderivrii, line 462: Looking at this test, the rust-based `zz_feature_map` raises `QiskitError`, whereas the python-based `ZZFeatureMap` raises`ValueError`. This is probably fine.

test/python/circuit/library/test_pauli_feature_map.py
```diff
@@ -295,5 +304,262 @@ def test_entanglement_not_specified(self):
             feat_map.count_ops()
 
 
+@ddt
+class TestPauliFeatureMap(QiskitTestCase):
+    """Test the Pauli feature map."""
```
alexanderivrii, line 309: In general, the tests for the new functionality are pretty much the same as the tests for the old functionality, however a few tests, namely `test_pauli_empty`, `test_pauli_evolution`, `test_invalid_entanglement` and `test_entanglement_not_specified` do seem missing. Was there a reason to remove these?

qiskit/circuit/library/data_preparation/_z_feature_map.py
```diff
@@ -40,38 +41,48 @@ class ZFeatureMap(PauliFeatureMap):
...
         >>> prep = ZFeatureMap(3, reps=3, insert_barriers=True)
-        >>> print(prep)
-             ┌───┐ ░ ┌──────────────┐ ░ ┌───┐ ░ ┌──────────────┐ ░ ┌───┐ ░ ┌──────────────┐
-        q_0: ┤ H ├─░─┤ U1(2.0*x[0]) ├─░─┤ H ├─░─┤ U1(2.0*x[0]) ├─░─┤ H ├─░─┤ U1(2.0*x[0]) ├
-             ├───┤ ░ ├──────────────┤ ░ ├───┤ ░ ├──────────────┤ ░ ├───┤ ░ ├──────────────┤
-        q_1: ┤ H ├─░─┤ U1(2.0*x[1]) ├─░─┤ H ├─░─┤ U1(2.0*x[1]) ├─░─┤ H ├─░─┤ U1(2.0*x[1]) ├
-             ├───┤ ░ ├──────────────┤ ░ ├───┤ ░ ├──────────────┤ ░ ├───┤ ░ ├──────────────┤
-        q_2: ┤ H ├─░─┤ U1(2.0*x[2]) ├─░─┤ H ├─░─┤ U1(2.0*x[2]) ├─░─┤ H ├─░─┤ U1(2.0*x[2]) ├
-             └───┘ ░ └──────────────┘ ░ └───┘ ░ └──────────────┘ ░ └───┘ ░ └──────────────┘
+        >>> print(prep.decompose())
```
alexanderivrii, line 44: Do now we need to call `decompose()` to produce the old result? Personally, I am fine with this since at the end we still obtain the same operator, but I am wondering if this could break some of the existing code.

crates/accelerate/src/circuit_library/pauli_feature_map.rs
```diff
@@ -0,0 +1,317 @@
...
+
+// custom math and types for a more readable code
+const PI2: f64 = PI / 2.;
+type Instruction = (
+    PackedOperation,
+    SmallVec<[Param; 3]>,
+    Vec<Qubit>,
+    Vec<Clbit>,
+);
+type StandardInstruction = (StandardGate, SmallVec<[Param; 3]>, SmallVec<[Qubit; 2]>);
```
alexanderivrii, line 37: Lol, it seems that we are defining this type in one way or another in every synthesis file. I don't suggest to do anything for this PR, but it would be nice to have this sit in a single place.

crates/accelerate/src/circuit_library/pauli_feature_map.rs
```diff
@@ -0,0 +1,317 @@
...
+///     └───┘                                   └───┘
+fn pauli_evolution(
+    pauli: &str,
+    indices: Vec<u32>,
+    time: Param,
+) -> impl Iterator<Item = StandardInstruction> + '_ {
+    // Get pairs of (pauli, qubit) that are active, i.e. that are not the identity. Note that
+    // the rest of the code also works if there are only identities, in which case we will
+    // effectively return an empty iterator.
+    let qubits = indices.iter().map(|i| Qubit(*i)).collect_vec();
```
alexanderivrii, line 63: Nice. I personally did not know about `collect_vec` before.

crates/accelerate/src/circuit_library/pauli_feature_map.rs
```diff
@@ -0,0 +1,317 @@
...
+        _ => unreachable!(),
+    });
+
+    // get the CX chain down to the target rotation qubit
+    let chain_down = active_paulis
+        .clone()
+        .into_iter()
+        .map(|(_, q)| q)
+        .tuple_windows() // iterates over (q[i], q[i+1]) windows
+        .map(|(ctrl, target)| (StandardGate::CXGate, smallvec![], smallvec![ctrl, target]));
```
alexanderivrii, line 101: `tuple_windows()` is another useful method I did not know about 👍

crates/accelerate/src/circuit_library/pauli_feature_map.rs
```diff
@@ -0,0 +1,317 @@
...
+        smallvec![time],
+        smallvec![last_qubit],
+    ));
+
+    // and finally chain everything together
+    basis_change
+        .chain(chain_down)
+        .chain(z_rotation)
+        .chain(chain_up)
+        .chain(inverse_basis_change)
```
alexanderivrii, line 125: Wow! This returns an iterator to construct the desired quantum circuit by chaining the iterators for individual parts, and yet the code is completely readable.

crates/accelerate/src/circuit_library/pauli_feature_map.rs
```diff
@@ -0,0 +1,317 @@
...
+    let entanglement = entanglement.unwrap_or(&default);
+
+    // extract the parameters from the input variable ``parameters``
+    let parameter_vector = parameters
+        .iter()?
+        .map(|el| Param::extract_no_coerce(&el?))
+        .collect::<PyResult<Vec<Param>>>()?;
+
+    // construct a Barrier object Python side to (possibly) add to the circuit
+    let packed_barrier = _get_barrier(py, feature_dimension)?;
```
alexanderivrii: I wonder if one can avoid constructing the barrier (which requires a python call) when `insert_barriers` is `false`, with possibly declaring `packed_barrier` as an `Option` to represent this.

crates/accelerate/src/circuit_library/pauli_feature_map.rs
```diff
@@ -0,0 +1,317 @@
...
+            let block_size = pauli.len() as u32;
+            let entanglement =
+                entanglement::get_entanglement(feature_dimension, block_size, entanglement, rep)
+                    .unwrap();
+            // let data_map = &mut data_map;
+            entanglement.flat_map(move |indices| {
+                // get the parameters the evolution is acting on and the corresponding angle,
+                // which is given by the data_map_func (we provide a default if not given)
+                let indices = indices.as_ref().unwrap();
+                // let indices = indices.as_ref().unwrap();
```
alexanderivrii: here and in some other places the commented out code can be removed

crates/accelerate/src/circuit_library/pauli_feature_map.rs
```diff
@@ -0,0 +1,317 @@
...
+        .collect::<PyResult<Vec<Param>>>()?;
+
+    // construct a Barrier object Python side to (possibly) add to the circuit
+    let packed_barrier = _get_barrier(py, feature_dimension)?;
+
+    // Main work: construct the circuit instructions as iterator. Each repetition is constituted
+    // by a layer of Hadamards and the Pauli evolutions of the specified Paulis.
+    // Note that we eagerly trigger errors, since the final CircuitData::from_packed_operations
+    // does not allow Result objects in the iterator.
+    let packed_insts = (0..reps).flat_map(|rep| {
```
alexanderivrii: The intrinsic beauty of this Rust code is beyond all limits (chaining iterators and flat maps -- just wow!), but (imho) makes quite a bit harder to understand what the final circuit is. The Python code is more readable (modulo the fact that one needs to switch between `pauli_feature_map` and `n_local` files trying to understand which of the `NLocal's` options are used and how). I am wondering if introducing a few auxiliary functions could help, or maybe let's just describe the circuit structure more clearly in this comment.

alexanderivrii (commented):
Thanks for the great work, the PR is in a very good shape. The only minor question that I had is why several python tests for the (older) `PauliFeatureMap` were not transferred to the (newer) `pauli_feature_map`.

The other comment is that the `pauli_feature_map` method in this PR essentially implements the very general circuit construction from `NLocal` (namely the `_build` method) in the special case of how it's called from `PauliFeatureMap`, and I am wondering if continuing this thread may lead to some code duplication in the future. However, I am completely fine with how this looks right now, since, were we to decide to do so, we could always implement the extended `build_nlocal` method as a function and let `pauli_feature_map` call this function with appropriate arguments. Personally I also find the new code easier to read because I do not need to carefully trace which arguments in `NLocal`'s `_build` are set to which values and would be against this potential future generalization, but some other people may think otherwise. You do not need to do anything, but I am leaving this as a comment to see if anyone would strongly object. 

Last but not least, I am awed by your Rust skills and have learned quite a few things by reviewing this PR. 🦀

test/python/circuit/library/test_pauli_feature_map.py
Cryoris: Ah that was because I didn't know how to raise `ValueError`s at the time, I can try changing it (and see whether the Rust compiler is happy with returning `PyValueError`s and `QiskitError`s from the same function)

qiskit/circuit/library/data_preparation/_z_feature_map.py
Cryoris: We don't change any behavior for the old classes, that was just documented wrongly. Printing the circuit without `decompose` will just print an opaque block 🙂

test/python/circuit/library/test_pauli_feature_map.py
Cryoris: Good point, some of these don't make sense with the new function:
* `test_pauli_evolution`: There's no longer a method `pauli_evolution` to get the individual Pauli evolution, so this is implicitly tested by checking the circuit structure.
* `test_pauli_empty` (& other mutable arguments): The pauli feature map is now just a function instead of a stateful object, so there are no attributes to test. I'm not sure how much sense it then makes to return an empty circuit... but we could still do that if you think it would be good to have 🙂 
* The others I think I overlooked -- I'll add them, thanks!

crates/accelerate/src/circuit_library/pauli_feature_map.rs
Cryoris: Done in 29fb437ec and d014852f8

crates/accelerate/src/circuit_library/pauli_feature_map.rs
Cryoris: Done in https://github.com/Qiskit/qiskit/commit/29fb437ecc35811f2682e14fd5affa2c8c4ef012

crates/accelerate/src/circuit_library/pauli_feature_map.rs
Cryoris: done in https://github.com/Qiskit/qiskit/commit/29fb437ecc35811f2682e14fd5affa2c8c4ef012

test/python/circuit/library/test_pauli_feature_map.py
Cryoris: The error handling is now done in https://github.com/Qiskit/qiskit/commit/d014852f8718e89fe195db704db123ef0c4d9732 which replaced the flat maps with explicit for loops. It hurt my soul a little bit to turn the intricate one-liner into loops, but the speed is the same and now errors are properly propagated 🙂

test/python/circuit/library/test_pauli_feature_map.py
Cryoris: I left them as `QiskitError` as that's coming from the entanglement functions -- we can change it there if we want to in a separate PR, but I thought it's best to not touch it in the Pauli feature map PR 🙂

alexanderivrii (approved):
Thanks for the improvements, LGTM!

PR #13048, opened 2024-08-28
Store Vars by index in `DAGCircuit`

### LINKED ISSUE #13027: Store a `Var(u32)` type for the `Wire::Var` variant in the rust DAGCircuit
### What should we add?

Right now the DAGCircuit's edge weight type is the enum `Wire` which contains a u32 newtype index for qubits and clbits, as well as a `PyObject`  for `Var` edges in the DAG. However, the u32 newtype index `Qubit` and `Clbit` are just indices for `BitData` which is a mapping of integer indices to `PyObject`. In the case of `Qubit` and `Clbit` it's for the python space qubit and clbit classes. We should be leveraging the same infrastructure for `Var`s too as this would shrink our memory overhead for `DAGCircuit` pretty substantially as the  edge size would be 50% smaller (64bits per edge instead of 128bits as it is currently).

To do this we need to add a new index newtype `Var` to indicate a var instance in the circuit, and then add a new `BitData` field to the `DAGCircuit` struct where we track new vars as they're added to the circuit. Then update the `Wire` enum to store the `Var` type instead of a `PyObject`. This will require updating anything in `DAGCircuit` that is interacting with Var wires currently.

While the plan is to port the `Var` type to rust before too long, we can still leverage the `Var` newtype as an index for edge weights after that migration occurs. Instead of using a `BitData` for the mapping we can instead just using an `IndexMap` or a `HashMap`.

jlapeyre:
Closes  #13027

### Summary

This follows the plan in #1307. This PR is centered on augmenting this
https://github.com/Qiskit/qiskit/blob/cc2edc9f7172d75ec454fdaf5748326cd3720f90/crates/circuit/src/dag_circuit.rs#L237-L240

with
```rust
pub vars: BitData<Var>,
```
That is, we use a Rust data structure `BitData` to store and manipulate variables, although the individual variables are still `PyObject`s. Previously, the Python-side containers and methods for manipulating variables were used.

### Status

Most of this is done, modulo debugging because it does not yet successfully compile. The part that is left is in
https://github.com/Qiskit/qiskit/blob/cc2edc9f7172d75ec454fdaf5748326cd3720f90/crates/circuit/src/dag_circuit.rs#L2866
This is a bit more work than all the rest because it involves iteration over the `Var`s and logic on the contents. For example, a set difference implemented as a Python call must be translated to or otherwise replaced in Rust.

It would be easier to a PR like this if there were some refactoring in dag_circuit.rs. For example, `fn substitute_node_with_dag` is over 500 lines long. One problem is the scope of local variables is not clear from the syntactic clues.

### DIFF
--- a/crates/circuit/src/bit_data.rs
+++ b/crates/circuit/src/bit_data.rs
@@ -47,6 +47,7 @@ impl BitAsKey {
     }
 }
 
+// Rust- and Python-side hash values are equal
 impl Hash for BitAsKey {
     fn hash<H: Hasher>(&self, state: &mut H) {
         state.write_isize(self.hash);
@@ -71,7 +72,7 @@ impl Eq for BitAsKey {}
 
 #[derive(Clone, Debug)]
 pub struct BitData<T> {
-    /// The public field name (i.e. `qubits` or `clbits`).
+    /// The public field name (i.e. `qubits`, `clbits`, or `vars`).
     description: String,
     /// Registered Python bits.
     bits: Vec<PyObject>,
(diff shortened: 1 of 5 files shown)


kevinhartman:
Rebased / carried forward into #13364. I did a separate PR since the merge conflicts were too hairy, but I've added @jlapeyre as a rightful co-author 🙂

Issue #12971, opened 2024-08-16
Port `QuantumCircuit.count_ops()` to rust


mtreinish:
### What should we add?

Right now the `QuantumCircuit.count_ops()` method is written in Python by iterating over the circuit data and collecting counts in a dictionary, sorting it by gate name and returning that dictionary. However, the underlying data is stored in rust's `CircuitData` struct and it would be much more efficient to do the counting and in rust. We should port this utility method to rust to provide and provide a significant speedup for the method.

melechlapson:
@mtreinish I can work on this issue

mtreinish:
Cool, thanks I've assign the issue to you.

mtreinish:
Implemented by https://github.com/Qiskit/qiskit/pull/13050

PR #13050, opened 2024-08-28
Port the count_ops method in QuantumCircuit to Rust

melechlapson:
### Summary
This PR is to address [issue 12971](https://github.com/Qiskit/qiskit/issues/12971) to port the count_ops method into Rust to improve performance. 

### Details and comments
I added a `count_ops` method to the `CircuitData` class in qiskit/crates/circuit/src/circuit_data.rs and updated the `count_ops` method in qiskit/circuit/quantumcircuit.py to call CircuitData's `count_ops` method.


melechlapson:
For a simple 2 qubit system in superposition, benchmarking shows ~3x improvement in performance on average. For a much larger, more complex circuit the improvement was ~9x.


```--------------------------------------------------------------------------------------------------- benchmark: 4 tests ----------------------------------------------------------------------------------------------------
Name (time in us)                                   Min                   Max                Mean              StdDev              Median                IQR            Outliers  OPS (Kops/s)            Rounds  Iterations
----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
test_new_count_ops_benchmark                     3.4159 (1.0)         21.2500 (1.0)        3.6377 (1.0)        0.6866 (1.0)        3.6249 (1.0)       0.0421 (1.0)         4;187      274.9019 (1.0)        1000           1
test_old_count_ops_benchmark                    10.0830 (2.95)       150.8750 (7.10)      11.2112 (3.08)       6.2859 (9.16)      10.7500 (2.97)      0.4580 (10.87)        7;27       89.1962 (0.32)       1000           1
test_new_count_ops_benchmark_large_circuit      59.9581 (17.55)      147.6670 (6.95)      65.9430 (18.13)      7.9297 (11.55)     64.3330 (17.75)     3.7500 (88.98)       62;73       15.1646 (0.06)       1000           1
test_old_count_ops_benchmark_large_circuit     529.9998 (155.16)   5,656.8750 (266.21)   593.9564 (163.28)   317.4110 (462.32)   567.1044 (156.45)   29.7080 (704.94)       8;59        1.6836 (0.01)       1000           1
----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
```

crates/circuit/src/circuit_data.rs
```diff
@@ -995,6 +996,16 @@ impl CircuitData {
         self.param_table.clear();
     }
 
+    pub fn count_ops(&mut self) -> IndexMap<String, i32> {
+        let mut ops_count: IndexMap<String, i32> = IndexMap::new();
```
mtreinish: One thing you should do here is set the hasher to `ahash::RandomState` with:

```suggestion
        let mut ops_count: IndexMap<String, i32, ::ahash::RandomState> = IndexMap::default();
```

it performs better than the default hash algorithm.

crates/circuit/src/circuit_data.rs
```diff
@@ -995,6 +996,16 @@ impl CircuitData {
...
+        let circuit: &Vec<PackedInstruction> = &self.data;
+        for instruction in circuit {
```
mtreinish: Just to make the code a bit more consise:

```suggestion
        for instruction in &self.data {
```

crates/circuit/src/circuit_data.rs
```diff
@@ -995,6 +996,17 @@ impl CircuitData {
...
+            *ops_count
+                .entry(instruction.op.view().name().to_string())
+                .or_insert(0) += 1;
+        }
+        ops_count
```
mtreinish, line 1000: The thing missing from the python implementation in this code is the final sort by counts, making the higher count instructions first. You can do that with:

```suggestion
        ops_count.par_sort_by(|_k1, v1, _k2, v2| v2.cmp(v1));
        ops_count
```

which will do a multithreaded sort inplace and put the larger counts first. You can drop the `par_` prefix to do it serially. It's probably worth benchmarking to see which is faster, because doing it in parallel doesn't always improve performance.

crates/circuit/src/circuit_data.rs
```diff
@@ -995,6 +996,17 @@ impl CircuitData {
         self.param_table.clear();
     }
 
+    pub fn count_ops(&mut self) -> IndexMap<String, i32> {
```
mtreinish: Do you mind adding a docstring with `///` (3 slashes is a docstring in rust 2 are a comment). It's not a hard requirement since this API is only used by python, but it could be called by rust.

mtreinish: Also I would change the value type to be a `usize` instead of `i32`. The `CircuitData` can contain `usize::MAX` elements and in the pathological edge case we could have an output of `{"h": usize::MAX}` assuming someone had enough local memory and constructed a circuit with 2^64 -1 hadamards.

Also I think you can avoid the string copy here because pyo3 should be able to convert a `&str` into a pystring when we return it. So it might be better to change the key type to `&str`. Although I'm not 100% certain on this, but the compiler will complain if it's not true.

```suggestion
    pub fn count_ops(&mut self) -> IndexMap<&str, usize> {
```

mtreinish (commented):
This looks good, thanks for doing this. I just have a few inline suggestions. After you apply them you might want to rerun the benchmarking too to see how it impacts the runtime performance of the method.

crates/circuit/src/circuit_data.rs
melechlapson: I saw no significant difference in benchmarking between the parallel sort, sequential sort, or default hashing algorithm.

crates/circuit/src/circuit_data.rs
```diff
@@ -995,6 +996,24 @@ impl CircuitData {
...
     }
 
+    /// Counts the number of times each operation is used in the circuit.
+    ///
+    /// # Parameters
+    /// - `self` - A mutable reference to the CircuitData struct.
+    ///
+    /// # Returns
+    /// An IndexMap containing the operation names as keys and their respective counts as values.
+    pub fn count_ops(&mut self) -> IndexMap<String, usize, ::ahash::RandomState> {
```
mtreinish: This doesn't need to be `&mut` because we're not mutating `self` here. So you can just use `&self`

crates/circuit/src/circuit_data.rs
```diff
@@ -995,6 +996,24 @@ impl CircuitData {
...
+        let mut ops_count: IndexMap<String, usize, ::ahash::RandomState> = IndexMap::default();
+        for instruction in &self.data {
+            *ops_count
+                .entry(instruction.op.view().name().to_string())
```
mtreinish: You don't actually need to build the `OperationRef` view here, `PackedOperation` has a name method you can just call. This is actually what I think blocked you from returning a `&str` as the key before.

```suggestion
                .entry(instruction.op.name().to_string())
```

crates/circuit/src/circuit_data.rs
```diff
@@ -995,6 +996,24 @@ impl CircuitData {
...
+                .or_insert(0) += 1;
+        }
+        ops_count.par_sort_by(|_k1, v1, _k2, v2| v2.cmp(v1));
+        ops_count
+    }
```
mtreinish: I took a quick look locally and if you make these changes you can use an `&str` here instead of `String`. This should actually have a noticeable impact on performance because a `String` is an owned object and with `to_string` you'll be copying it (`String` is essentially a `Vec<u8>`). While if you can use `&str` that will avoid any copies and use references for all the strings; at least until we return to Python.

```suggestion
    pub fn count_ops(&self) -> IndexMap<&str, usize, ::ahash::RandomState> {
        let mut ops_count: IndexMap<&str, usize, ::ahash::RandomState> = IndexMap::default();
        for instruction in &self.data {
            *ops_count.entry(instruction.op.name()).or_insert(0) += 1;
        }
        ops_count.par_sort_by(|_k1, v1, _k2, v2| v2.cmp(v1));
        ops_count
    }
```

crates/circuit/src/circuit_data.rs
```diff
@@ -995,6 +996,24 @@ impl CircuitData {
...
+    ///
+    /// # Returns
+    /// An IndexMap containing the operation names as keys and their respective counts as values.
+    pub fn count_ops(&mut self) -> IndexMap<String, usize, ::ahash::RandomState> {
+        let mut ops_count: IndexMap<String, usize, ::ahash::RandomState> = IndexMap::default();
+        for instruction in &self.data {
+            *ops_count
+                .entry(instruction.op.view().name().to_string())
+                .or_insert(0) += 1;
+        }
```
mtreinish: I'm not actually suggesting you make this change, as I think a for loop is probably more legible. But I was curious if you could do this with just an iterator which you can do like this:

```suggestion
        let mut ops_count =
            self.data
                .iter()
                .fold(IndexMap::default(), |mut ops_count, instruction| {
                    *ops_count.entry(instruction.op.name()).or_insert(0) += 1;
                    ops_count
                });
```

I don't think there is any advantage to writing it like this.  So it's more just an interesting musing on how to use functional programming style iterators for this.

melechlapson: I'll test it out

crates/circuit/src/circuit_data.rs
melechlapson: Ah for some reason I didn't see that and that made things more complicated for me

crates/circuit/src/circuit_data.rs
melechlapson: seemed to work

mtreinish (approved):
LGTM, thanks for doing this.

PR #13052, opened 2024-08-28
Fully port FilterOpNodes to Rust

### LINKED ISSUE #12263: Port `FilterOpNodes` to Rust

mtreinish:
### Summary

This commit ports the FilterOpNodes pass to rust. This pass is exceedingly simple and just runs a filter function over all the op nodes and removes nodes that match the filter. However, the API for the class exposes that filter function interface as a user provided Python callable. So for the current pass we need to retain that python callback. This limits the absolute performance of this pass because we're bottlenecked by calling python.

Looking to the future, this commit adds a rust native method to DAGCircuit to perform this filtering with a rust predicate FnMut. It isn't leveraged by the python implementation because of layer mismatch for the efficient rust interface and Python working with `DAGOpNode` objects. A function using that interface is added to filter labeled nodes. In the preset pass manager we only use FilterOpNodes to remove nodes with a specific label (which is used to identify temporary barriers created by qiskit). In a follow up we should consider leveraging this new function to build a new pass specifically for this use case.

### Details and comments

Fixes #12263
Part of #12208


crates/circuit/src/dag_circuit.rs
```diff
@@ -5785,6 +5785,26 @@ impl DAGCircuit {
         }
     }
 
+    // Filter any nodes that don't match a given predicate function
+    pub fn filter_op_nodes<F, E>(&mut self, mut predicate: F) -> Result<(), E>
+    where
+        F: FnMut(&PackedInstruction) -> Result<bool, E>,
```
kevinhartman: I wonder if we should allow the callable to be fallible at all. I know we have often supported this in Rustworkx core's API to make it easier to use with the wild world of Python. But in this case, maybe it's okay to require the user to give us a function that isn't expected to fail.

Otherwise, we'll be asking them to define `Infallible` result types probably more often than not.

crates/accelerate/src/filter_op_nodes.rs
```diff
@@ -0,0 +1,65 @@
...
+    Ok(())
+}
+
+/// Remove any nodes that have the provided label set
+///
+/// Args:
+///     dag (DAGCircuit): The dag circuit to filter the ops from
+///     label (str): The label to filter nodes on
+#[pyfunction]
+pub fn filter_labelled_op(dag: &mut DAGCircuit, label: String) {
```
kevinhartman: It sounds to me like `labeled` is the preferred American spelling but `labelled` is preferentially in British English.

crates/circuit/src/dag_circuit.rs
```diff
@@ -5785,6 +5785,26 @@ impl DAGCircuit {
         }
     }
 
+    // Filter any nodes that don't match a given predicate function
+    pub fn filter_op_nodes<F, E>(&mut self, mut predicate: F) -> Result<(), E>
```
kevinhartman: I think the term "filter" makes more sense in the context of a transpiler pass / pipeline than it does directly on `DAGCircuit`, but I'm not sure I can think of a better name. Maybe something like `remove_op_nodes_by_filter`.

I'm fine with this comment being closed if anyone (yourself included) prefers the current name.

kevinhartman (commented):
Looks good, just a few minor comments.

crates/circuit/src/dag_circuit.rs
mtreinish: I'm fine with dropping the result here, I added it because when I originally wrote this I intended to use this method for the python function. But because that wasn't possible because of different interfaces between python and rust. So I agree this probably doesn't need to be fallible if we're not calling this with a python callback. I'll remove the result as it'll make it easier to work with.

crates/accelerate/src/filter_op_nodes.rs
mtreinish: Hah @jakelishman will be proud of me because this looks normal to me. :gb:

I can update the spelling to use the conventional American spelling.

crates/circuit/src/dag_circuit.rs
mtreinish: Done in: https://github.com/Qiskit/qiskit/pull/13052/commits/5e00d3c416172a82d63e0ca378ef2465292edd7e

crates/accelerate/src/filter_op_nodes.rs
mtreinish: Done in: https://github.com/Qiskit/qiskit/pull/13052/commits/1a99b80736f072f9e90d07a80c1877a712ee3a1d

crates/circuit/src/dag_circuit.rs
mtreinish: Personally I thought this was pretty clear, using filter as a verb typically means to remove like this. But I'll let others chime in here, because my thoughts/reading on things aren't always clear to others.

kevinhartman: I think your thoughts on it are valid. The name just makes me think of an iterator kind of situation i.e. if you gave me just the method name I would not guess that it took a `&mut self`. But, it does. So it's up to personal preference I suppose.

jakelishman: fwiw, compare Rust's `Iterator::filter` (new iterator containing only matching items) and Python's builtin `filter` (same).  Changing to `remove_op_nodes_by_filter` would invert the polarity of the `predicate` compared to the existing code and would probably conflict with Rust and Python builtins.

jakelishman: and for the `&self` vs `&mut self` point, both Rust and Python's methods are mutations on the incoming iterator, which is consistent with `&mut self` (well, strictly `self`, but the analogy starts breaking down there

kevinhartman: Heh, the analogy breaking down there is the problem I have with this, since we're not consuming the DAG, so it's no longer a filter of some pipelined data flow, but instead removals from a sitting pool.

But, I'm fine with keeping the name as is since I'm clearly in the minority on this 🙂.

kevinhartman (approved):
Looks good. Thanks for the changes!

Issue #13053, opened 2024-08-29
"Feature Request: Extend random_circuit function to support gate number and gate set specification"


PhysicsQoo:
### What should we add?

I would like to request an enhancement to the random_circuit function in qiskit/circuit/random/utils.py. Currently, the function allows specifying only the depth of the circuit, but it would be highly useful to have additional parameters to:

- Specify the number of gates in the circuit.
- Specify a custom set of gates to be used, similar to the functionality provided by the random_clifford_circuit function.

This enhancement would allow users to have more control over the structure of the randomly generated circuit, which is particularly useful for research and testing purposes where specific gate constraints are required.

- Add an optional parameter num_gates to specify the exact number of gates in the generated circuit.
- Add an optional parameter gates to allow users to provide a list of gate types that should be randomly selected during circuit generation.

This feature would make random_circuit more versatile and better suited for a wider range of applications.
It would provide more flexibility for researchers who need to simulate or analyze circuits with specific gate constraints.

### Reference

1. [random_circuit function documentation](https://docs.quantum.ibm.com/api/qiskit/circuit#random_circuit)
2. [random_clifford_circuit function documentation](https://docs.quantum.ibm.com/api/qiskit/circuit#random_clifford_circuit)
3. [Qiskit Random Circuit Repository](https://github.com/Qiskit/qiskit/blob/cc2edc9f7172d75ec454fdaf5748326cd3720f90/qiskit/circuit/random/utils.py#L24)

Cryoris:
Specifying the gates sounds like a good idea. The `random_circuit` function should likely also use the `qiskit.circuit.library.standard_gates.get_standard_gate_name_mapping` to get the gates, instead of keeping a separate hardcoded list. I see you've already opened a first PR (thanks for that!) but maybe we can tackle the problems in a different fashion and resolve the problems together.

I would suggest we add a `gates: list[str] | None` argument to `random_circuit`, which allows selecting the gates (this would make `random_clifford` and `random_clifford_t` one-liners). We can use the `get_standard_gate_mapping` to determine the list of gates, which returns a dictionary of `{name: gate_instance}` pairs. What do you think? Let me know if you'd like some pointers on how to do that 🙂

PhysicsQoo:
I agree that adding a `gates: list[str] | None` argument is a great idea, as it would allow users to easily select their desired gate set. This would indeed make the `random_clifford` and `random_clifford_t` functions much simpler.

[...]

PR #13054, opened 2024-08-29
Update random_circuit fuction in qiskit.circuit.random.utils.py 

PhysicsQoo:
### Summary

Adding random_clifford_T_circuit to support random circuits selects T and Tdg as elements.

### Details and comments
part of [#13053](https://github.com/Qiskit/qiskit/issues/13053)

### DIFF
--- a/qiskit/circuit/random/utils.py
+++ b/qiskit/circuit/random/utils.py
@@ -356,3 +356,74 @@ def random_clifford_circuit(num_qubits, num_gates, gates="all", seed=None):
         circ.append(gate, qargs, copy=False)
 
     return circ
+
+def random_clifford_T_circuit(num_qubits, num_gates, gates="all", seed=None):
+    """Generate a pseudo-random Clifford circuit.
+
+    This function will generate a Clifford circuit by randomly selecting the chosen amount of Clifford
+    gates from the set of standard gates in :mod:`qiskit.circuit.library.standard_gates`. For example:
+
+    .. plot::
+       :include-source:
+
+       from qiskit.circuit.random import random_clifford_circuit
+
+       circ = random_clifford_circuit(num_qubits=2, num_gates=6)
+       circ.draw(output='mpl')
+
+    Args:
+        num_qubits (int): number of quantum wires.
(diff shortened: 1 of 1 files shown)


Cryoris:
Thanks for opening a PR @PhysicsQoo! But I think we could tackle your issue in a bit a more general fashion, I just left a comment on your issue.

PhysicsQoo:
### **Description:**

Added a `gates: str | list[str] ='all'` argument, replacing the hardcoded gate list with `standard_gates.get_standard_gate_name_mapping()`.
Introduced a parameterized argument, allowing users to select whether they want parameterized rotation gates.

### **Known Issue:**

The parameterized functionality currently encounters an AttributeError: 'Parameter' object has no attribute '_root_uuid'. This issue is yet to be resolved.
The feature works correctly when parameterized = False.

Cryoris:
Dear @PhysicsQoo, it's nice to see that you're motivated to implement the additions to the random circuit -- but we have to approach this a bit differently: for this PR, we cannot remove existing functionality.

We can do the following steps:
* Start by only adding support for an _additional_ `gates` argument in the `random_circuit` function. All other arguments remain in place, and we have to add the argument at the end (to not break user code that was using arguments positionally). So the new signature would be
    ```python
        def random_circuit(
        num_qubits,
        depth,
        max_operands=4,
        measure=False,
        conditional=False,
        reset=False,
        seed=None,
        num_operand_distribution: dict = None,
        gates=None,
    ):
    ```
* Then, instead of the hardcoded lists `gates_(1,2,3,4)q` we would like to use the `get_standard_gate_name_mapping` function. 
* If `gates` was given by the user, we can then filter to have only the supported gates.

Importantly, we don't change any of the existing functionality that creates the circuit with a certain depth or handles the operand distribution or parameterization. In a next step, we can have a look at supporting also `num_gates` as termination criterion or other features, but it would be good to start with a single feature and understand the process 🙂

1ucian0:
In its current state, this PR is a breaking change that do not follow the deprecation policy. I would close it for now. @PhysicsQoo please feel free to reopen a PR to address the @Cryoris comments.

PR #13055, opened 2024-08-29
Functionality to list operations in QuantumCircuit with associated Qubits

### LINKED ISSUE #12972: Add a method to compute gate counts per qubit
### What should we add?

We have a method `QuantumCircuit.count_ops` method but in some scenarios you want to know the count of gates on each qubit. You can fairly easily build this view of the circuit by iterating over `QuantumCirucit.data` but because the data is stored in rust it would be much more efficient to build this view of the circuit in Rust and return it to Python. We should add this method to the python `QuantumCircuit` class but instead call an inner method defined on the rust `CircuitData` struct.

tonmoy-b:
fixes #12972 

### Summary
Added function to tally operations in circuit with associated qubits
fixes #12972 

### Details and comments
-Added function count_ops_with_qubits() in quantumcircuit.py, this returns the operations in the QuantumCircuit along with the qubits associated with those operations.
-The abovementioned function uses the Rust function fn operations_with_qubits in circuit_data.rs
-A test (test_count_ops_with_qubits()) has been added to file test/python/circuit/test_circuit_data.py
-tox has been run -- no test failures reported
-tox -eblack run to reformat files

[Yes ] I have added the tests to cover my changes.
[Yes ] I have updated the documentation accordingly.
[Yes ] I have read the CONTRIBUTING document.


tonmoy-b:
Hi @mtreinish ,

     I have made this PR to fix issue 12972: Add a method to compute gate counts per qubit 
     
     Kindly take a look and let me know if any changes should be made.

     For a quick look, the new code works as below:
     a = QuantumCircuit(3, 3)
        a.h(0)
        a.x(1)
        a.h(1)
        a.cx(1, 0)
        b = a.count_ops_with_qubits()
        assert b == [
            ("h", "Qubit(QuantumRegister(3, 'q'), 0)"),
            ("x", "Qubit(QuantumRegister(3, 'q'), 1)"),
            ("h", "Qubit(QuantumRegister(3, 'q'), 1)"),
            ("cx", "Qubit(QuantumRegister(3, 'q'), 1)"),
            ("cx", "Qubit(QuantumRegister(3, 'q'), 0)"),
        ]

crates/circuit/src/circuit_data.rs
```diff
@@ -504,6 +504,25 @@ impl CircuitData {
         Ok(())
     }
 
+    /// Lists all operations along with the qubits involved in those operations.
+    ///
+    /// Returns:
+    ///     list.
+    #[pyo3(signature = ())]
+    pub fn operations_with_qubits(&self, py: Python<'_>) -> PyResult<Py<PyAny>> {
```
mtreinish: Instead of returning a `Py<PyAny>` here typically I'd prefer to return the `Vec` directly because then this function would be callable from rust too. The pyo3 macros build a parallel c ffi function for use that is what python is actually calling, so this is still usable rust function otherwise. If we return a Vec here the macro will do the pyobject conversion for us when called from python.

crates/circuit/src/circuit_data.rs
```diff
@@ -504,6 +504,25 @@ impl CircuitData {
...
+        let mut ret: Vec<(Option<String>, Option<String>)> = vec![];
```
mtreinish: I think I think the return type here is a bit off. What I think would be useful here is to return either a vec or hashmap of qubit indices that map to the instruction counts on them. For example something like:
```
[
    {"h": 20, "cx" 2, "s": 10},
    {"cx": 2},
]
```
for a circuit that has 2 qubits with 20 hadamards, 10 s gates, and 2 cx gates on qubit 0 and qubit 1 has 2 cx gates.

qiskit/circuit/quantumcircuit.py
```diff
@@ -3521,6 +3521,17 @@ def count_ops(self) -> "OrderedDict[Instruction, int]":
             count_ops[instruction.operation.name] = count_ops.get(instruction.operation.name, 0) + 1
         return OrderedDict(sorted(count_ops.items(), key=lambda kv: kv[1], reverse=True))
 
+    # The stringified return type is because OrderedDict can't be subscripted before Python 3.9, and
+    # typing.OrderedDict wasn't added until 3.7.2.  It can be turned into a proper type once 3.6
+    # support is dropped.
+    def count_ops_with_qubits(self) -> list:
```
mtreinish: The next qiskit release will only support python >= 3.9 so this isn't relevent. Also this isn't currently returning an ordered dict. Although I would argue we should not be using an ordered dict at all. The other methods do because they have been around since Qiskit supported much older versions of python (before dicts were sorted by default) and have retained it for backwards compatibility. But for new functionality we should just return a dict or a list.

```suggestion
```

crates/circuit/src/circuit_data.rs
```diff
@@ -504,6 +504,25 @@ impl CircuitData {
...
+        for (_index, inst) in self.data.iter().enumerate() {
+            let op_str = inst.op.view().name().to_string();
```
mtreinish: We should probably try to avoid `to_string()` here as this will make a copy of every string, and we can keep it all by reference in Rust space. When we return to Python that will do the copying for us.

crates/circuit/src/circuit_data.rs
```diff
@@ -504,6 +504,25 @@ impl CircuitData {
...
+            for b in self.qargs_interner.get(inst.qubits) {
+                ret.push((
+                    Some(op_str.to_string()),
+                    Some(self.qubits.get(*b).unwrap().to_string()),
```
mtreinish: If we were going to use qubit objects like this, I'd have suggested returning them directly instead of stringify them. Then the python user could interact with the bit object directly. However, I don't think we need to work with actual bit objects and can just use indices. If a user wants to map the index to a bit, in python that's just `circuit.qubits[index]`.

mtreinish (commented):
Thanks for taking the issue and getting an implementation started. I left a few inline comments. But I think my biggest concern right now is about the return type. Write now you're returning a `Vec<Option<String>, Option<String>>`, but instead I think it should be a `Vec<HashMap<&str, usize>>` where the vec index is the qubit index and the hashmap is gate name mapped to count. I had some other inline comments, but refactoring the code to adjust the return type will take those into account.

crates/circuit/src/circuit_data.rs
tonmoy-b: hi @mtreinish , I'm a bit stuck here, how could I get this information (the qubit numbers) without having to get substrings on the string gotten from the Qubit objects from 'Some(self.qubits.get(*b).unwrap().to_string()'? Or should I go ahead and get the substrings to obtain the qubits and populate the HasMaps from that?

mtreinish: Oh, that's not too difficult for any given `PackedInstruction` object `instr` you get while iterating over the circuit data you can call `CircuitData::get_qargs(instr.qubits)` which will return a `&[Qubit]` object. The `Qubit` object is a [newtype](https://doc.rust-lang.org/rust-by-example/generics/new_types.html) `u32`. So you can do `Qubit.0 as usize` to get the index into a vec for any `Qubit`.

tonmoy-b: Ok, will follow that procedure. Thanks

tonmoy-b:
Hi @mtreinish ,

    I have made the changes as recommended, kindly take a look.

Thanks and Regards,
Tonmoy

tonmoy-b:
Hi, I don't mean to be a nuisance but I would appreciate a review and suggestions for updates if needed. Thanks

tonmoy-b:
Hi, I'd appreciate a review of this PR. I believe it is probably close to what's required at this point.

PR #13056, opened 2024-08-29
Add `Operation::blocks` method.

kevinhartman:
### Summary
This gives us a way to get the blocks of a control flow operation as an iterator of `CircuitData`. If called on a non-control-flow instruction, it returns an empty `Vec`.

### Details and comments

This is not intended as a final API for this, which will likely instead return an iterator of `&CircuitData` once control flow operations have been fully ported to Rust and own their blocks without requiring conversion.

### DIFF
--- a/crates/circuit/src/operations.rs
+++ b/crates/circuit/src/operations.rs
@@ -12,6 +12,7 @@
 
 use approx::relative_eq;
 use std::f64::consts::PI;
+use std::vec;
 
 use crate::circuit_data::CircuitData;
 use crate::circuit_instruction::ExtraInstructionAttributes;
@@ -2085,6 +2086,27 @@ pub struct PyInstruction {
     pub op_name: String,
     pub control_flow: bool,
     pub instruction: PyObject,
+}
+
+impl PyInstruction {
+    pub fn blocks(&self) -> Option<impl Iterator<Item = CircuitData>> {
+        Python::with_gil(|py| -> Option<vec::IntoIter<CircuitData>> {
+            let raw_blocks = self.instruction.getattr(py, "blocks").ok()?;
+            let blocks: &Bound<PyTuple> = raw_blocks.downcast_bound::<PyTuple>(py).ok()?;
+            Some(
+                blocks
+                    .iter()
+                    .map(|b| {
+                        b.getattr(intern!(py, "_data"))
+                            .unwrap()
+                            .extract::<CircuitData>()
+                            .unwrap()
+                    })
+                    .collect::<Vec<_>>()
+                    .into_iter(),
+            )
+        })
+    }
 }
 
 impl Operation for PyInstruction {
--- a/crates/circuit/src/operations.rs
+++ b/crates/circuit/src/operations.rs
@@ -146,6 +146,7 @@ pub trait Operation {
     fn num_clbits(&self) -> u32;
     fn num_params(&self) -> u32;
     fn control_flow(&self) -> bool;
+    fn blocks(&self) -> Vec<CircuitData>;
     fn matrix(&self, params: &[Param]) -> Option<Array2<Complex64>>;
     fn definition(&self, params: &[Param]) -> Option<CircuitData>;
     fn standard_gate(&self) -> Option<StandardGate>;
@@ -208,6 +209,15 @@ impl<'a> Operation for OperationRef<'a> {
             Self::Gate(gate) => gate.control_flow(),
             Self::Instruction(instruction) => instruction.control_flow(),
             Self::Operation(operation) => operation.control_flow(),
+        }
+    }
+    #[inline]
+    fn blocks(&self) -> Vec<CircuitData> {
+        match self {
+            OperationRef::Standard(standard) => standard.blocks(),
+            OperationRef::Gate(gate) => gate.blocks(),
+            OperationRef::Instruction(instruction) => instruction.blocks(),
+            OperationRef::Operation(operation) => operation.blocks(),
         }
     }
     #[inline]
@@ -507,20 +517,20 @@ impl Operation for StandardGate {
         STANDARD_GATE_NUM_QUBITS[*self as usize]
     }
 
-    fn num_params(&self) -> u32 {
-        STANDARD_GATE_NUM_PARAMS[*self as usize]
-    }
-
     fn num_clbits(&self) -> u32 {
         0
+    }
+
+    fn num_params(&self) -> u32 {
+        STANDARD_GATE_NUM_PARAMS[*self as usize]
     }
 
     fn control_flow(&self) -> bool {
         false
     }
 
-    fn directive(&self) -> bool {
-        false
+    fn blocks(&self) -> Vec<CircuitData> {
+        vec![]
     }
 
     fn matrix(&self, params: &[Param]) -> Option<Array2<Complex64>> {
@@ -2020,6 +2030,10 @@ impl Operation for StandardGate {
     fn standard_gate(&self) -> Option<StandardGate> {
         Some(*self)
     }
+
+    fn directive(&self) -> bool {
+        false
+    }
 }
 
 const FLOAT_ZERO: Param = Param::Float(0.0);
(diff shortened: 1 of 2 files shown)


crates/circuit/src/operations.rs
```diff
@@ -2087,6 +2088,27 @@ pub struct PyInstruction {
     pub instruction: PyObject,
 }
 
+impl PyInstruction {
+    pub fn blocks(&self) -> Option<impl Iterator<Item = CircuitData>> {
+        Python::with_gil(|py| -> Option<vec::IntoIter<CircuitData>> {
```
jakelishman: If this needs the GIL, can we just accept the GIL token in the function call?  The PR comment says this is unlikely to be the final interface, and removing the GIL token from all the call sites won't be hard when it comes to it.

I'm concerned about how much we're hiding `with_gil` calls in Qiskit at the moment - given that the move to Rust is in part meant to make threaded parallelism easier for us, I think hiding implicit serialisation is possibly working against us.

crates/circuit/src/operations.rs
```diff
@@ -2087,6 +2088,27 @@ pub struct PyInstruction {
...
+            let raw_blocks = self.instruction.getattr(py, "blocks").ok()?;
```
jakelishman: At the point of creating a `PyInstruction`, we type-check and cache whether it's a proper `ControlFlowOp` (stored in `PyInstruction::control_flow`).  We ought to check that field before doing this, because there's nothing stopping a user-defined `Instruction` from having a `blocks` attribute that's a `tuple`, but not what we expect.

jakelishman (commented):
This seems generally fine to me, though given that `control_flow` is a part of the `Operation`, does it make sense just to similarly define this somewhere higher up as well, so we need to do less unpacking of the internal objects to get here?

kevinhartman:
> This seems generally fine to me, though given that control_flow is a part of the Operation, does it make sense just to similarly define this somewhere higher up as well, so we need to do less unpacking of the internal objects to get here?

It may. Do you have any thoughts on adding a new variant to `OperationRef` for control flow? That might not be the right choice for this PR, but I wonder if it'd be a better API long-term to be able to unpack a control flow operation and access its blocks with something like:

```rust
if let OperationRef::ControlFlow(op) = op {
    match op.op {
        ControlFlowOp::ForLoop | ControlFlowOp::WhileLoop => { todo!() },
        _ => (),
    }

    // No need to unwrap an option because we know we have a control flow op.
    for block in op.blocks() {
        todo!()
    }
}
```

I don't think we'd need any support from `PackedOperation` for this.

jakelishman:
I'm not sure what you mean by not needing support from `PackedOperation` - `OperationRef` _is_ `PackedOperation` for all intents and purposes; it's just "proper" `enum` form of a reference to it.  (There's space for 4 more variants in `PackedOperation` without any trouble at all, and the existing three Python forms could be compressed into a single one if we needed to anyway.)

I guess "longer term" depends a lot on the degree of length here.  _Long long_ term, I think control-flow operations may want to move out of being in a `PackedInstruction` at all, and be first-class in the data description (especially in `DAGCircuit`, less clearly in `CircuitData`).  Shorter term, it could be fine to separate it off, but if the primary reason to do that is to avoid an `Option` here, maybe we just get rid of the `Option` in the method and leave the PR as it is?  If it _isn't_ a control-flow operation, we can just return an empty iterator, which is an equally valid return value; the caller of this function is presumably already calling `Operation::control_flow` before anyway.

crates/circuit/src/operations.rs
kevinhartman: Done in 591b0cba56737425a2df94cae65c593d89fde389.

crates/circuit/src/operations.rs
kevinhartman: I moved `blocks` to the `Operation` trait, and it seems like that interface doesn't accept a `Python` token for any of the methods. I agree with your concern, but keeping it out of the signature feels more consistent with everything else, for now.

kevinhartman:
> Long long term, I think control-flow operations may want to move out of being in a PackedInstruction at all, and be first-class in the data description (especially in DAGCircuit, less clearly in CircuitData). Shorter term, it could be fine to separate it off, but if the primary reason to do that is to avoid an Option here, maybe we just get rid of the Option in the method and leave the PR as it is? If it isn't a control-flow operation, we can just return an empty iterator, which is an equally valid return value; the caller of this function is presumably already calling Operation::control_flow before anyway.

Heh, that's all fair. I ended up moving `blocks` to `Operation` to make it easy to access in the meantime (hopefully justified in the same way that `Operation::control_flow` is part of that interface). The return type of `blocks` is now `Vec<CircuitData>` (mostly because I think returning an abstract `Iterator` from a trait method is not stable in Rust), and the caller will get an empty `Vec` if the operation does not have any blocks for any valid reason.

crates/circuit/src/operations.rs
```diff
@@ -145,6 +146,7 @@ pub trait Operation {
     fn num_clbits(&self) -> u32;
     fn num_params(&self) -> u32;
     fn control_flow(&self) -> bool;
+    fn blocks(&self) -> Vec<CircuitData>;
```
raynelfss, line 149: Since blocks will only return a populated `Vec` when dealing with a `PyInstruction`, could we make this accept a `py` token regardless? It is very likely that the user here will have obtained the GIL whenever they want to play with an instance of `PyInstruction`, it's all within the name after all. 

I saw your previous comments when discussing with Jake about keeping this consistent. However, in this case, this method is only supposed to work with one type of  operation, one that specifically holds a `PyObject` and is tied quite closely with python, and even then it is not guaranteed to be an instance of `ControlFlowOp`. The function itself will also perform callbacks to python and obtaining the GIL manually might prove itself costly here (it would be nice to benchmark it beforehand to see if this holds true).

raynelfss (commented):
Thank you for working on this @kevinhartman, I'm sure I'm not the only one in desperate need of this method existing in Rust.

Everything looks fine but you might want to reconsider the usage of `Python::with_gil()` here, specially since this method will only work with `PyInstruction` and the user will most likely hold a `py` token when calling this method.

crates/circuit/src/operations.rs
kevinhartman: It's the case that right now `blocks` comes from Python, but once we've ported control flow operations to Rust that goes away. I don't see a compelling reason to ask for a Python token at the `Operation` level for `blocks` but not for its other methods, so I'd rather be consistent to avoid implying to code readers that there's some logical difference between `Operation::blocks` and say, `Operation::matrix`.

In the case of performance, having the GIL already means that `with_gil` will return fastest (as opposed to not holding it). And, we don't even try to acquire it unless this is a control flow operation, so the penalty to calling `with_gil` within `Operation::blocks` would scale with respect to the number of control flow operations in the circuit, rather than the number of op nodes.

crates/circuit/src/operations.rs
```diff
@@ -2103,6 +2118,24 @@ impl Operation for PyInstruction {
...
+            blocks
+                .iter()
+                .map(|b| {
+                    b.getattr(intern!(py, "_data"))
+                        .unwrap()
+                        .extract::<CircuitData>()
+                        .unwrap()
+                })
+                .collect()
+        })
```
raynelfss:
```suggestion
        Python::with_gil(|py| -> PyResult<Vec<CircuitData>> {
            let raw_blocks = self.instruction.getattr(py, "blocks")?;
            let blocks: &Bound<PyTuple> = raw_blocks.downcast_bound::<PyTuple>(py)?;
            blocks
                .iter()
                .map(|b| -> PyResult<_> {
                    b.getattr(intern!(py, "_data"))?.extract::<CircuitData>()
                })
                .collect::<PyResult<_>>()
        })
        .unwrap_or_default()
```
You might find this option more attractive. We let the iterator and the `Python::with_gil()` do their own result handling and we only unwrap at the end and use the `Default` option were it to ever fail. This option is more stable as it has zero chance of ever panicking, if it ever were to fail an empty list will be returned.

raynelfss (commented):
This looks sensible to me, and about ready to merge. I only have one inline comment here that you might want to look into :eyes:

crates/circuit/src/operations.rs
kevinhartman: This is actually very similar to what I had originally in 622509eb3e48f6a0c6bfa8f13cf7686b94c842c5 🙂 (the difference being that I was returning an `Option` instead of an empty `Vec` if something went wrong).

In the current version, I actually purposefully made this code less flexible. The reason being that if one of these expectations is not true (e.g. if `self.instruction` does not have a `blocks` attribute), then it means that there is a _**bug**_ in Qiskit. In that case, returning an empty `Vec` would only mask the underlying issue and prevent us from finding out about it / fixing it.

jakelishman: Fwiw I'm fairly sure I've done the same thing as Kevin elsewhere and taken the `control_flow` attribute to mean that it's an instance of `ControlFlowOp` and therefore `blocks` must exist (which matches the Python expectations).

jakelishman (dismissed):
This also seems fine to me now, and Ray can merge if he's happy.

crates/circuit/src/operations.rs
raynelfss: That's good then, can you add a comment explaining this? I think it's a good idea to mention that all of this unwrapping is being done with the purpose of catching bad behavior.

kevinhartman: Good idea. Done in 2a0951fd5201f4890bdb2213cb23404c77eaee17.

jakelishman (approved):

PR #13057, opened 2024-08-29
Promote `CONTROL_FLOW_OP_NAMES` to public API

jakelishman:
### Summary

This is generally useful for Qiskit's built-in transpiler passes, and we've had some interest in using it for the same reasons from outside the library.  While the control-flow structure from within Rust might change in the future, in the immediate term and from Python space, it feels fair to expose this as public API, since all the information contained in it is certainly already public.

### Details and comments

Minor request from @ihincks.

### DIFF
--- a/qiskit/circuit/__init__.py
+++ b/qiskit/circuit/__init__.py
@@ -686,6 +686,13 @@ Within :class:`QuantumCircuit`, classical control flow is represented by specifi
 
     ControlFlowOp
 
+For convenience, there is a :class:`frozenset` instance containing the :attr:`.Instruction.name`
+attributes of each of the control-flow operations.
+
+.. data:: CONTROL_FLOW_OP_NAMES
+
+    Set of the instruction names of Qiskit's known control-flow operations.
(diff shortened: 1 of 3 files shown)


mtreinish (approved):

ihincks (commented):
Thanks @jakelishman !

wshanks:
Lazy follow up request -- it would be nice to have an equivalent to `get_standard_gate_name_mapping` as well to standardize code like this:

https://github.com/Qiskit/qiskit/blob/b9d5c9c6aeb4568b6e9fba8943517ebed300886d/qiskit/transpiler/preset_passmanagers/generate_preset_pass_manager.py#L459-L464

It came up in https://github.com/Qiskit/qiskit-ibm-runtime/pull/2044#discussion_r1846328508. I will keep it on my list but I can't do it at the moment.

ElePT:
@wshanks FYI: https://github.com/Qiskit/qiskit/pull/13472. Targeted for 2.0.

PR #13058, opened 2024-08-30
Extend commutative cancellation for SX/Sdg/Tdg

Cryoris:
### Summary

The `CommmutativeCancellation` pass didn't consider cancellations of `Sdg/Tdg/SX/SXdg` gates. This PR does a minor fix to add them.

### Details and comments

There are other open points we might want to discuss:
* basic Z-sequences are not cancelled if no Z rotation is present in the circuit: E.g. `Z-Z` is not cancelled if there is no `RZ/U1/P` specified in the basis set or target
* on the other hand: X-sequences are squashed into an RX gate, even if there is no RX gate specified in the basis or target

Both these cases could be solved by e.g. always assuming an RX/RZ rotation gate to absorb the cancellations, but if finally the angles don't sum up to 0, we disregard the accumulation.


Cryoris:
Out of date

Issue #13063, opened 2024-08-30
Deprecate pulse module and related functionality for removal in 2.0


mtreinish:
### What should we add?

It's been decided that moving into Qiskit 2.0 we'll be removing the pulse module and all the related functionality (such as the [schedule](https://docs.quantum.ibm.com/api/qiskit/compiler#schedule) function along with it's backing module and calibrations in the target) as it's not in scope for the direction of the project anymore. We should ensure that we mark all these features as deprecated in 1.3.0 so there is sufficient warning prior to the release of Qiskit 2.0. These features will continue to be supported in Qiskit through the EoL with the 1.x.y release series into the 2H of 2025.

eliarbel:
I'd be happy to work on this.

Zonaldo7:
Will pulse related module become another independent package?

eliarbel:
Hi @Zonaldo7, the Pulse module will be part of [Qiskit Dynamics](https://github.com/qiskit-community/qiskit-dynamics) once removed.

chookity-pokk:
> Hi [@Zonaldo7](https://github.com/Zonaldo7), the Pulse module will be part of [Qiskit Dynamics](https://github.com/qiskit-community/qiskit-dynamics) once removed.

Just a small note here in the event others find this. The Pulse module won't be part of Qiskit Dynamics as mentioned [in this PR](https://github.com/qiskit-community/qiskit-dynamics/pull/365#issuecomment-2804010180).

PR #13064, opened 2024-08-30
Remove benchmarks for deprecated (or soon to be) features

mtreinish:
### Summary

This commit removes benchmarks for the deprecated assemble module and the soon to be (see #13063) deprecated pulse module. While these features are still supported it's not worth the spending the limited CPU time we hacve on the benchmarking system to track the performance of these components anymore.

### Details and comments


ElePT (approved):
Makes sense

PR #13065, opened 2024-09-01
Port RemoveDiagonalGatesBeforeMeasure to rust 

### LINKED ISSUE #12255: Port `RemoveDiagonalGatesBeforeMeasure` to Rust

ShellyGarion:
### Summary

close #12255


qiskit/transpiler/passes/optimization/remove_diagonal_gates_before_measure.py
```diff
@@ -48,22 +37,4 @@ def run(self, dag):
...
-            if isinstance(predecessor, DAGOpNode) and isinstance(predecessor.op, diagonal_2q_gates):
-                successors = dag.quantum_successors(predecessor)
-                if all(isinstance(s, DAGOpNode) and isinstance(s.op, Measure) for s in successors):
-                    nodes_to_remove.add(predecessor)
-
-        for node_to_remove in nodes_to_remove:
-            dag.remove_op_node(node_to_remove)
-
-        return dag
+        remove_diagonal_gates_before_measure(dag)
```
mtreinish, line 40: You need to return the dag here, the pass interface expects a dag object to returned, even if we mutate by reference.

```suggestion
        remove_diagonal_gates_before_measure(dag)
        return dag
```

crates/accelerate/src/remove_diagonal_gates_before_measure.rs
```diff
@@ -0,0 +1,96 @@
...
+
+/// Run the RemoveDiagonalGatesBeforeMeasure pass on `dag`.
+/// Args:
+///     dag (DAGCircuit): the DAG to be optimized.
+/// Returns:
+///     DAGCircuit: the optimized DAG.
+#[pyfunction]
+#[pyo3(name = "remove_diagonal_gates_before_measure")]
+fn run_remove_diagonal_before_measure(_py: Python, dag: &mut DAGCircuit) -> PyResult<()> {
+    let diagonal_1q_gates = HashSet::from([
```
mtreinish: I'm curious if using a `static` array here would be more efficient. We get constant time lookups with a hash set, but it becomes a question of the speed of the hash function vs doing 6 enum comparisons. Just using the static array might be faster. But we should benchmark to check.

If a hash set is faster then we should use a lazy static here to avoid the hash set allocation on each execution.

crates/accelerate/src/remove_diagonal_gates_before_measure.rs
```diff
@@ -0,0 +1,96 @@
...
+                    Some(gate) => {
+                        if diagonal_1q_gates.contains(&gate) {
+                            nodes_to_remove.insert(predecessor);
+                        } else if diagonal_2q_gates.contains(&gate) {
+                            let successors = dag.quantum_successors(predecessor);
+                            let mut remove_s = false;
+                            for s in successors {
+                                let node_s = &dag.dag[s];
+                                let NodeType::Operation(inst_s) = node_s else {panic!()};
+                                if inst_s.op.name() == "measure" {
```
mtreinish: I don't think think the logic here is quite right. If I'm reading this correctly we'll remove the 2q gates if they're in front of 1 measure too, but we should only be removing them if a 2q diagonal is in front of 2 measurements, one on each qubit.

You can do this check using an iterator chain (not compiled and written on my phone so there are likely typos):

```rust
let remove_s = succesors.map(|s| {
    let node_s = &dag.dag[s];
    let NodeType::Operation(inst_s) = node_s else {panic!()};
    inst_s.op.name()
})
.all(|name| name == "measure");
```

mtreinish (commented):
This looks good, thanks for doing this. I just left a few small inline comments which I think should fix the test failures too.

crates/accelerate/src/remove_diagonal_gates_before_measure.rs
```diff
@@ -0,0 +1,96 @@
...
+use qiskit_circuit::operations::StandardGate;
+
+/// Run the RemoveDiagonalGatesBeforeMeasure pass on `dag`.
+/// Args:
+///     dag (DAGCircuit): the DAG to be optimized.
+/// Returns:
+///     DAGCircuit: the optimized DAG.
+#[pyfunction]
+#[pyo3(name = "remove_diagonal_gates_before_measure")]
+fn run_remove_diagonal_before_measure(_py: Python, dag: &mut DAGCircuit) -> PyResult<()> {
```
mtreinish: Oh, the other thing is, if we're not interacting with Python at all we don't need to have a py token argument. This makes the function easier to call from rust if we need to.

```suggestion
fn run_remove_diagonal_before_measure(dag: &mut DAGCircuit) -> PyResult<()> {
```

crates/accelerate/src/remove_diagonal_gates_before_measure.rs
```diff
@@ -0,0 +1,103 @@
...
+    let diagonal_2q_gates = HashSet::from([
+        StandardGate::CZGate,
+        StandardGate::CRZGate,
+        StandardGate::CU1Gate,
+        StandardGate::RZZGate,
+        StandardGate::CPhaseGate,
+        StandardGate::CSGate,
+        StandardGate::CSdgGate,
+    ]);
+
```
ShellyGarion, line 48: Note that there are some 1-qubit and 2-qubit diagonal gates that did not appear in the original list, so I added them here (I'll add them to the tests later).

There is also a 3-qubit diagonal gate: [`CCZGate`](https://docs.quantum.ibm.com/api/qiskit/qiskit.circuit.library.CCZGate) (which was not added in this PR)

ShellyGarion:
the failing test should be fixed after https://github.com/Qiskit/qiskit/pull/13067 is merged

crates/accelerate/src/remove_diagonal_gates_before_measure.rs
ShellyGarion: the panic caused a test with a 2q gate and one measurement to fail, so @alexanderivrii suggested another fix.

crates/accelerate/src/remove_diagonal_gates_before_measure.rs
ShellyGarion: indeed, since they are constants, it makes sense to make them static arrays

crates/accelerate/src/remove_diagonal_gates_before_measure.rs
ShellyGarion: In fact, there is also an n-qubit diagonal gate: [`MCPhaseGate`](https://docs.quantum.ibm.com/api/qiskit/qiskit.circuit.library.MCPhaseGate). This gate was not handled this PR.
This is since the algorithm given here that goes over all the successors of each of the predecessors would be `O(n^2)` and not `O(n)`.

ShellyGarion:
Performance comparison of the old python code and the new rust code for a `random_circuit`:

| n_qubits | depth | time is sec of old Python code | time in sec of new rust code | 
|----------|--------|---------------------------|-------------------------|
| 5 | 1024 | 0.06 | 0.00 | 
| 14 | 1024 |  0.47 | 0.02 | 
| 20 | 1024 | 0.92 | 0.04 |

crates/accelerate/src/remove_diagonal_gates_before_measure.rs
```diff
@@ -0,0 +1,103 @@
...
+/// Run the RemoveDiagonalGatesBeforeMeasure pass on `dag`.
+/// Args:
+///     dag (DAGCircuit): the DAG to be optimized.
+/// Returns:
+///     DAGCircuit: the optimized DAG.
+#[pyfunction]
+#[pyo3(name = "remove_diagonal_gates_before_measure")]
+fn run_remove_diagonal_before_measure(dag: &mut DAGCircuit) -> PyResult<()> {
+    static DIAGONAL_1Q_GATES: [StandardGate; 8] = [
```
Cryoris, line 28: Would it be faster if these were a `HashSet` and the lookup operation is `O(1)`? We'd have an overhead to construct the HashSet, but we do check `contains` many times 🤔

crates/accelerate/src/remove_diagonal_gates_before_measure.rs
```diff
@@ -0,0 +1,103 @@
...
+        StandardGate::CZGate,
+        StandardGate::CRZGate,
+        StandardGate::CU1Gate,
+        StandardGate::RZZGate,
+        StandardGate::CPhaseGate,
+        StandardGate::CSGate,
+        StandardGate::CSdgGate,
+    ];
+
+    let mut nodes_to_remove = HashSet::new();
```
Cryoris: I'm not quite sure here, but I think this here would even be better as Vector, since we never have to search in this so we can skip calculating the hash 🤔

crates/accelerate/src/remove_diagonal_gates_before_measure.rs
ShellyGarion: with `HashSet` the time is slightly higher than with static arrays (so the current code remains unchanged)

| n_qubits | depth | time is sec with static arrays | time in sec with `HashSet` | 
|----------|--------|---------------------------|-------------------------|
| 5 | 1024 | 0.00 | 0.00 |  
| 14 | 1024 | 0.02 | 0.03 | 
| 20 | 1024 | 0.04 | 0.05 |
| 50 | 1024 |  0.31 | 0.32 | 
| 100 | 1024 | 1.52 | 1.54 |

crates/accelerate/src/remove_diagonal_gates_before_measure.rs
ShellyGarion: Indeed, we can change it to a Vec, but at the end we need to convert it to a HashSet before removing the nodes. It slightly improves the performance.

| n_qubits | depth | time is sec  | 
|----------|--------|--------------|
| 5 | 1024 | 0.00 |  
| 14 | 1024 | 0.02 |  
| 20 | 1024 | 0.04 | 
| 50 | 1024 |  0.31 |  
| 100 | 1024 | 1.48 |

crates/accelerate/src/remove_diagonal_gates_before_measure.rs
```diff
@@ -0,0 +1,105 @@
...
+                    continue;
+                }
+            }
+        }
+    }
+
+    let set_nodes_to_remove: HashSet<NodeIndex> = nodes_to_remove.into_iter().collect();
+    for node_to_remove in set_nodes_to_remove {
+        dag.remove_op_node(node_to_remove)
+    }
```
mtreinish: You can avoid the deduplication with a hash set here by just check doing a quick lookup on the graph before calling remove. Something like:

```suggestion
    for node_to_remove in nodes_to_remove {
        if dag.dag.node_weight(node_to_remove).is_some() {
            dag.remove_op_node(node_to_remove);
        }
    }
```

ShellyGarion: | n_qubits | depth | time is sec  | 
|----------|--------|--------------|
Actually, this change makes the code slower, so I didn't change the current code with the HashSet.

| n_qubits| depth | time in sec |
|----------|--------|-------------|
| 5 | 1024 | 0.01 |  
| 14 | 1024 | 0.05 |  
| 20 | 1024 | 0.08 | 
| 50 | 1024 |  0.53 |  
| 100 | 1024 | 2.85 |

crates/accelerate/src/remove_diagonal_gates_before_measure.rs
Cryoris: nice, thanks for checking! 🙂

Cryoris (dismissed):
LGTM, also great to add the phase gates 👍🏻 

As you mentioned, there is `MCPhaseGate` which we could handle, as well as `DiagonalGate` or any `AnnotatedOperation` with a diagonal base gate. But that would be something for a follow up (we would also make sure to remove these before any further decomposition occurs).

Cryoris (approved):

PR #13067, opened 2024-09-02
Fix extraction of controlled parametric gates

jakelishman:
### Summary

The `mutable` check in the controlled-gate `OperationFromPython` extraction logic to check for a mutated `base_gate` was overzealous, and would return false positives for parametric controlled gates.  The only modification to `base_gate` of a standard-library gate that would not have caused data-model problems from Python space would be setting the base-gate label, which is used for a public feature of the circuit visualisers.

The change to `get_standard_gate_name_mapping` is just a minor convenience to make the gate objects directly appendable to a circuit; previously, each `Parameter` object was distinct and had a UUID clash with others of the same name, so could not be used together.  The new behaviour is purely a convenience for tests; it largely should not be useful for users to directly append these gates.

### DIFF
--- a/crates/circuit/src/circuit_instruction.rs
+++ b/crates/circuit/src/circuit_instruction.rs
@@ -547,12 +547,18 @@ impl<'py> FromPyObject<'py> for OperationFromPython {
             // mapping to avoid an `isinstance` check on `ControlledGate` - a standard gate has
             // nonzero `num_ctrl_qubits` iff it is a `ControlledGate`.
             //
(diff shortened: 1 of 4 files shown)


mtreinish (approved):
LGTM, this is a straightforward fix and good to see. This actually fixes some edge cases I was hitting during the development of the DAGCircuit PR around things like `CXGate` ended up as a `PyGate` in rust incorrectly and complicating equality checks.

Issue #12936, opened 2024-08-09
Normalize on `py_foo()` and `foo()` function name pattern in `accelerate` crate


mtreinish:
### What should we add?

In the accelerate crate when we have a split between a python facing function and a rust entrypoint for internal reuse we typically used the pattern where the rust function was named `foo_inner()` and the `pyfunction` was named `foo()`. For example:

https://github.com/Qiskit/qiskit/blob/main/crates/accelerate/src/dense_layout.rs#L101-L130

However in the circuit crate we started using a different pattern, where the pyfunction was named `py_foo` and the name attribute on the `pyfunction` macro was the name without the `py` prefix. As we're writing an increasing amount of rust code this seems like a better pattern. So we should migrate the accelerate crate to follow the same pattern internally.

jakelishman:
Fwiw, I've been changing my naming conventions as I go, and at the moment I'm roughly around using a `py_` prefix more to mean "this function deals with / returns objects for Python-space consumption" as opposed to "this function deals only with Rust objects", and avoiding using a prefix when it wasn't necessary to disambiguate (like the function is a term that only applies to Python space, or there's no chance of confusion with a Rust-specific method), rather than it being specifically `pyfunction`s/`pymethod`s and not.

raynelfss:
Thank you for bringing this up, yes we should label the `#[pymethods]` that are python only and use the `py_` prefix for each of those. I guess my only question would be what happens when the method has a hybrid use? For example, yes the method is exposed to python through `#[pymethods]` but it's not so different in its structure that it could be left as a method that's exposed to rust as well to avoid duplicating code. 
- Do we call it `py_foo()` and create a separate duplicate rust-native method `foo()` that's called by `py_foo()`?
- Or would we be allowed to make an exception and just have one method `foo()` that's exposed through #[pymethods] but it is public within Rust?

I also think it is important to talk about methods that we'd like to keep in a private api of rust that use a `py: Python` argument and have no use in python. Do we create a third `impl Foo` to store these as well?

There's still a lot of figuring out to do when it comes to how to properly write Rust code that will be exposed to Python, but I think it is good we start thinking of these.

mtreinish:
[...]

PR #13069, opened 2024-09-02
Oxidize GateDirection pass

### LINKED ISSUE #12252: Port `GateDirection` to Rust
This is blocked on #12052 #11721

eliarbel:
### Summary
Move `GateDirection._run_coupling_map` and `GateDirection._run_target` to Rust. Also moving the code which creates replacement DAGs to Rust. Both `GateDirection` and `CheckGateDirection` will reside after this in `gate_direction.rs`.

### Details and comments
Depends on #13042 
Close #12252


eliarbel:
Will add performance benchmarking numbers soon.

crates/accelerate/src/gate_direction.rs
```diff
@@ -97,7 +112,7 @@ where
 
         if let OperationRef::Instruction(py_inst) = packed_inst.op.view() {
             if py_inst.control_flow() {
-                let circuit_to_dag = imports::CIRCUIT_TO_DAG.get_bound(py); // TODO: Take out of the recursion
+                let circuit_to_dag = imports::CIRCUIT_TO_DAG.get_bound(py);
```
raynelfss, line 115: This call here has two ways of being replaced:
    1. Using the `circuit_to_dag` rust function that would prevent you from extracting the Python `DAGCircuit`.
    2. Creating a parallel method that can do the recursion call with `QuantumCircuit` or `CircuitData` instances to skip the conversion entirely.

I'd honestly prefer option two since it doesn't seem that much is being done with the instructions that requires Python calls, and using `CircuitData::iter()` would allow you to iterate through the instructions in very much the same way.

raynelfss, line 117: This call here has two ways of being replaced:
    1. Using the `circuit_to_dag` rust function that would prevent you from extracting the Python `DAGCircuit`.
    2. Creating a parallel method that can do the recursion call with `QuantumCircuit` or `CircuitData` instances to skip the conversion entirely.

I'd honestly prefer option two since it doesn't seem that much is being done with the instructions that requires Python calls, and using `CircuitData::iter()` would allow you to iterate through the instructions in very much the same way.

crates/accelerate/src/gate_direction.rs
```diff
@@ -142,8 +157,505 @@ where
...
+                            Some(qargs),
+                            Some(
+                                get_std_gate_class(py, std_gate)
+                                    .expect("These gates should have Python classes")
+                                    .bind(py),
+                            ),
+                            Some(inst.params_view().to_vec()),
+                        )
+                        .unwrap_or(false)
+                }
```
raynelfss, line 234: For a follow up, or maybe even for #13308, we should be able to test instruction compatibility with a `Target` from a `StandardGate`.

crates/accelerate/src/gate_direction.rs
```diff
@@ -142,8 +157,505 @@ where
...
+    for node in dag.op_nodes(false) {
+        let NodeType::Operation(packed_inst) = &dag.dag()[node] else {
+            panic!("PackedInstruction is expected");
+        };
+
+        let op_args = dag.get_qargs(packed_inst.qubits);
+
+        if let OperationRef::Instruction(py_inst) = packed_inst.op.view() {
+            if py_inst.control_flow() {
+                let circuit_to_dag = imports::CIRCUIT_TO_DAG.get_bound(py);
```
raynelfss: For this I believe you can use the rust `circuit_to_dag` since it's not as lossy of a conversion. And since you're dealing with blocks you can keep the python `dag_to_circuit`. We'll improve on these structures in a follow-up :)

crates/accelerate/src/gate_direction.rs
```diff
@@ -142,8 +157,505 @@ where
...
+                packed_inst.params_view(),
+                op_args
+            )));
+        }
+    }
+
+    for (node, op_blocks) in ops_to_replace {
+        let NodeType::Operation(packed_inst) = &dag.dag()[node] else {
+            panic!("PackedInstruction is expected");
+        };
```
raynelfss: Use `unwrap_operation()` to extract a known op_node.
```suggestion
        let packed_inst: &PackedInstruction = &dag.dag()[node].unwrap_operation();
```

crates/accelerate/src/gate_direction.rs
```diff
@@ -142,8 +157,505 @@ where
...
+                    py_op: packed_inst.py_op.clone(),
+                },
+                sort_key: "".into_py(py),
+            },
+            DAGNode { node: None },
+        ),
+    )?;
+
+    dag.has_calibration_for(py, dag_op_node.borrow(py))
+}
```
raynelfss, line 414: For this function, is there a reason why `has_calibration_for_index` can't be used? From what I can remember `has_calibration_for` works for `DAGOpNodes` that are already valid within the `DAGCircuit`, so a `NodeIndex` should work just as fine.

crates/accelerate/src/gate_direction.rs
```diff
@@ -142,8 +157,505 @@ where
...
+    dag.py_apply_operation_back(
+        py,
+        op.bind(py).clone(),
+        Some(TupleLikeArg::extract_bound(&PyTuple::new_bound(py, qargs))?),
+        None,
+        false,
+    )?;
+
+    Ok(())
+}
```
raynelfss: Hmm... Could we perhaps use something like `DAGCircuit::apply_operation_back()` here? To take advantage of the rust native methods. Now I understand it makes it a little more annoying since we need to copy over the `Param`s and the `ExtraInstAttributes` but at least you remove the Python abstraction.

crates/accelerate/src/gate_direction.rs
```diff
@@ -142,8 +157,505 @@ where
...
+        py,
+        new_dag,
+        &std_gate(py, StandardGate::CXGate)?,
+        &vec![&q1, &q0],
+    )?;
+    apply_op_back(py, new_dag, &std_gate(py, StandardGate::HGate)?, &vec![&q0])?;
+    apply_op_back(py, new_dag, &std_gate(py, StandardGate::HGate)?, &vec![&q1])?;
+
+    Ok(new_dag.clone())
+}
```
raynelfss: This and the subsequent `_replacement_dag()` methods could use more rust-natiuve approaches here without so much Python abstraction. I understand this was made before the methods existed, but now that we have them available, we should use them over the python exposed ones (e.g. use `DAGCircuit::apply_operation_back()` over `DAGCircuit::py_apply_operation_back()` and so on...).

test/python/transpiler/test_gate_direction.py
```diff
@@ -65,9 +65,9 @@ def test_direction_error(self):
         """The mapping cannot be fixed by direction mapper
         qr0:---------
 
-        qr1:---(+)---
+        qr1:----.-----
                 |
-        qr2:----.----
+        qr2:---(+)----
```
raynelfss, line 70: Is there a reason why these control and target qubits are being flipped in these tests?

crates/accelerate/src/gate_direction.rs
```diff
@@ -142,8 +157,505 @@ where
...
+
+#[pymodule]
 pub fn gate_direction(m: &Bound<PyModule>) -> PyResult<()> {
     m.add_wrapped(wrap_pyfunction!(py_check_with_coupling_map))?;
     m.add_wrapped(wrap_pyfunction!(py_check_with_target))?;
+    m.add_wrapped(wrap_pyfunction!(py_fix_with_coupling_map))?;
+    m.add_wrapped(wrap_pyfunction!(py_fix_with_target))?;
```
raynelfss: Is there a reason why we're callingn these `fix_...` and not `run_...`?

crates/accelerate/src/gate_direction.rs
```diff
@@ -142,8 +157,505 @@ where
...
+    apply_op_back(py, new_dag, &std_gate(py, StandardGate::HGate)?, &vec![&q0])?;
+    apply_op_back(py, new_dag, &std_gate(py, StandardGate::HGate)?, &vec![&q1])?;
+
+    Ok(new_dag.clone())
+}
+
+#[pymodule]
 pub fn gate_direction(m: &Bound<PyModule>) -> PyResult<()> {
     m.add_wrapped(wrap_pyfunction!(py_check_with_coupling_map))?;
     m.add_wrapped(wrap_pyfunction!(py_check_with_target))?;
```
raynelfss: Are we using these two methods outside of Rust? I don't see them being used in the python `gate_direction.py` fie.

raynelfss (commented):
This looks pretty good so far, I have a couple of comments mainly suggesting the usage of rust native methods for the `DAGCircuit`. But I feel this should be ready after those corrections.

crates/accelerate/src/gate_direction.rs
eliarbel: Regarding option 2 I see your point and also prefer it, this can be a nice optimization. I also have another optimization I would like to apply, namely caching the constructed DAGs so they will not be constructed from scratch every time. So I think I'll just open a new issue for tracking these but it won't be for 1.3. Sounds OK?

crates/accelerate/src/gate_direction.rs
eliarbel: Agree, that would be nice. Any concrete suggestion though for this PR or are we OK with leaving this as is for now?

crates/accelerate/src/gate_direction.rs
eliarbel: Done in 4b363a19bd5920926d3f31748738f7b2070e4dd2

crates/accelerate/src/gate_direction.rs
eliarbel: Ah, cool! Done in 4b363a19bd5920926d3f31748738f7b2070e4dd2

crates/accelerate/src/gate_direction.rs
eliarbel: `has_calibration_for_index` uses the qubits order of the node whose `NodeIndex` is given. This is checked already in line 195. However, here we want to check if there is a calibration def for the same node type and params, but with a reversed qubits order, a combination which might not exist necessarily in the dag. We could theoretically extend the `has_calibration_for_index` function with an Option arg to indicate a custom qubits order to check on. I'm not sure it's worthwhile since the path is being deprecated anyway. What do you say?

crates/accelerate/src/gate_direction.rs
eliarbel: Done in 4b363a19bd5920926d3f31748738f7b2070e4dd2 (up to creating QuantumRegister still from Python)

crates/accelerate/src/gate_direction.rs
eliarbel: Done

crates/accelerate/src/gate_direction.rs
eliarbel: Yes, these are used in `check_gate_direction.py`. Both `CheckGateDirection` and `GateDirection` are implemented in `gate_direction.rs`. The logic is very similar for both, so I figured it makes sense to put those together and they are rather small each.

crates/accelerate/src/gate_direction.rs
eliarbel: As stated in https://github.com/Qiskit/qiskit/pull/13069#discussion_r1816967910 - the main reason being that both `GateDirection` and `CheckGateDirection` live in this file and there are the `check_...` couple and the `fix_...` couple, each pair corresponds to their respective passes. I could BTW just have one run method for each pass which could take both Target and a coupling map but I actually liked better the modularity of having 2 entry points (one for target and one for coupling map), where each defines a specialized closure but both call the same function for each pass. These are being called by `GateDirection.run()` anyway. That said, should we follow a convention that all pass entry functions should be named `run..,`? If so I'll prefix the four python functions in the file with it. Let me know.

test/python/transpiler/test_gate_direction.py
eliarbel: This is just updating the circuit drawings to match the actual test circuits in several docstrings.

crates/accelerate/src/gate_direction.rs
raynelfss: Should be left as is for now. We can address it in a follow up.

crates/accelerate/src/gate_direction.rs
raynelfss: That's fine then. I just wanted to understand the reasoning behind this, but what you said makes sense. If anything we can introduce another way to check for calibrations in a follow-up.

crates/accelerate/src/gate_direction.rs
raynelfss: It is good to have conventions to keep everything consistent. That said, we haven't had any of them apply to transpiler passes in rust yet. We do have the main run method calling upon other helper methods, so having the rust methods that represent the main `run` method of the pass have `run` somewhere in the name would be nice. I did it as a suffix for #13237, but having it as a prefix sounds more elegant. Maybe we should open an issue about these conventions here (sort of what #12936 proposes).

crates/accelerate/src/gate_direction.rs
```diff
@@ -142,8 +159,471 @@ where
...
+fn add_qreg(py: Python, dag: &mut DAGCircuit, num_qubits: u32) -> PyResult<Vec<Qubit>> {
+    let qreg = imports::QUANTUM_REGISTER
+        .get_bound(py)
+        .call1((num_qubits,))?;
+    dag.add_qreg(py, &qreg)?;
+    let mut qargs = Vec::new();
+
+    for i in 0..num_qubits {
+        let qubit = qreg.call_method1(intern!(py, "__getitem__"), (i,))?;
+        qargs.push(dag.qubits().find(&qubit).expect("Qubit should stored"));
```
raynelfss:
```suggestion
        qargs.push(dag.qubits().find(&qubit).expect("Qubit should have been stored in the DAGCircuit"));
```

crates/accelerate/src/gate_direction.rs
```diff
@@ -142,8 +159,471 @@ where
...
+    qargs: &[Qubit],
+    param: Option<SmallVec<[Param; 3]>>,
+) -> PyResult<()> {
+    dag.apply_operation_back(
+        py,
+        PackedOperation::from_standard(gate),
+        qargs,
+        &[],
+        param,
+        ExtraInstructionAttributes::new(None, None, None, None),
```
raynelfss: Use `Default` here:
```suggestion
        ExtraInstructionAttributes::default(),
```

crates/accelerate/src/gate_direction.rs
```diff
@@ -142,8 +159,471 @@ where
...
+        new_dag,
+        StandardGate::CXGate,
+        &[qargs[1], qargs[0]],
+        None,
+    )?;
+    apply_operation_back(py, new_dag, StandardGate::HGate, &[qargs[0]], None)?;
+    apply_operation_back(py, new_dag, StandardGate::HGate, &[qargs[1]], None)?;
+
+    Ok(new_dag.clone())
+}
```
raynelfss, line 503: Could we maybe optimize these circuit building functions by having them work more like `OnceCell`? Where you could create a static instance that is initialized just once, you can then just call clone each time you need a replacement `DAGCircuit`. For this you could use `GILOnceCell`, and have these function calls just call on `.get_or_init(py, |_| )` 

Here's an example of how it could work:
```rust

static FOO_DAG: GILOnceCell<PyResult<DAGCircuit>> = GILOnceCell::new();

fn foo_replacement_dag(py: Python) -> PyResult<DAGCircuit> {
    FOO_DAG.get_or_init(py, || -> PyResult<DAGCircuit> {
        let mut dag: DAGCircuit = DAGCircuit::new(py);
        // Not sure if you can call `add_qreg` from within the closure.
        let qargs = add_qreg(py, new_dag, 2)?;
        let qargs = qargs.as_slice();
        // not sure if you can call `apply_operation_back` from inside the closure.
        apply_operation_back(py, &mut dag, StandardGate::XGate, &[qargs[0]], None)?;
        Ok(dag)
    }).cloned()
}
```
This ensures that the circuit is built one and that whenever we call on this function, we just clone the circuit.

This change is merely a suggestion since #13335 makes improvements on chained additions to the `DAGCircuit` and these changes could be made in a follow up 🙃.

raynelfss (commented):
This looks about ready to merge, just a couple of suggestions that warrant discussion.

eliarbel:
Runtime comparison results:

| Before [f2e07bc5] main   | After [4b363a19] fix-gate-direction   |   Ratio | Benchmark (num_qubits, depth)                                                       |
|----------------------------|-----------------------------------------|---------|-----------------------------------------------------------------------------|
| 103±5ms                    | 5.79±0.3ms                              |    0.06 | gate_direction.GateDirectionBenchmarks.time_gate_direction_target(14, 1024) |
| 28.8±3ms                   | 1.79±0.1ms                              |    0.06 | gate_direction.GateDirectionBenchmarks.time_gate_direction_target(5, 1024)  |
| 56.8±0.9ms                 | 6.52±0.1ms                              |    0.11 | mapping_passes.RoutedPassBenchmarks.time_gate_direction(20, 1024)           |
| 6.48±0.3ms                 | 687±20μs                                |    0.11 | mapping_passes.RoutedPassBenchmarks.time_gate_direction(5, 1024)            |

crates/accelerate/src/gate_direction.rs
eliarbel: As said in https://github.com/Qiskit/qiskit/pull/13069#discussion_r1820706305 this is definitely an optimization opportunity I would like to explore.  I am actually exploring already some caching implementation, similar to what you suggested, namely generating a replacement DAG (on-demand) once and cloning when needed. This applies to non-parameterized gates only since replacements for parameterized gates (e.g. RZX) depend on the actual parameter. It also includes caching a template DAG for RZX and then when an RZX replacement is needed replacing the RZX node with a concrete parametrically-bound RZX node.  Then I also want to look at avoiding calling `.clone()` , which I believe could give even more improvement, but this will require more refactoring. But bottom-line I'm afraid I will not have time to properly explore and benchmark more optimization for the 1.3 release, so I think we should have this in a new PR. And obviously there is a risk of jut trying to micro optimize without too much benefit, but again, I'll need more time to benchmark in detail.

crates/accelerate/src/gate_direction.rs
eliarbel: Ah... good catch. Thx. Done in https://github.com/Qiskit/qiskit/pull/13069/commits/79a62174af9c6955a3d9b05adcb456d939641950

crates/accelerate/src/gate_direction.rs
eliarbel: Done in https://github.com/Qiskit/qiskit/pull/13069/commits/79a62174af9c6955a3d9b05adcb456d939641950

crates/accelerate/src/gate_direction.rs
eliarbel: On a second thought and after considering some alternatives I actually liked most this one: `py_check_direction_coupling_map` (and similarly for the `fix` and `target` variants). I find it more informational and succinct than other options like: `py_run_check_with_coupling_map` or `py_run_check_coupling_map` or `py_run_check_direction_coupling_map` etc... Also the `py_` prefix in it indicates that it's the Python entry point, hence also the pass entry point so the `run` prefix is not really required for this purpose. So unless you feel strongly against it, I would prefer to stay with this more informational naming scheme here until we decide on a convention for the pass entry point names.

crates/accelerate/src/gate_direction.rs
```diff
@@ -142,8 +159,479 @@ where
...
+                    params: packed_inst.params_view().iter().cloned().collect(),
+                    extra_attrs: packed_inst.extra_attrs.clone(),
+                    #[cfg(feature = "cache_pygates")]
+                    py_op: packed_inst.py_op.clone(),
+                },
+                sort_key: "".into_py(py),
+            },
+            DAGNode { node: None },
+        ),
+    )?;
```
raynelfss, line 411: This is something I discussed with @kevinhartman, but we will probably move to using `.into_py(py)` to convert an `PyClass` to a `PyObject` in Rust, but we want to make sure to do it in a follow up to make sure it is used everywhere, so I'm okay with leaving it as it is here but perform the change to `.into_py(py)` in a follow up.

raynelfss (approved):
LGTM! Just one inline comment that I don't think we'll address here. Great job!

PR #13071, opened 2024-09-02
docs: ensure CollectAndCollapse is part of the public API

mrossinek:
### Summary

As discussed offline, this ensures that the `CollectAndCollapse` transpiler pass is actually part of the public API.

### Details and comments

I am not sure whether this requires a release note... :thinking:

### DIFF
--- a/qiskit/transpiler/passes/__init__.py
+++ b/qiskit/transpiler/passes/__init__.py
@@ -71,6 +71,7 @@ Optimizations
    Collect1qRuns
    Collect2qBlocks
    CollectMultiQBlocks
+   CollectAndCollapse
    CollectLinearFunctions
    CollectCliffords
    ConsolidateBlocks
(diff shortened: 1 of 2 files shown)


jakelishman (approved):
Thanks Max.  If you feel like a release note would be particularly useful to people, feel free to add one, but we probably don't really need one - this was announced as if it were public API, already was intended to be, and we don't typically have release notes for documentation-only changes.

Let me know either way and I'll tag for merge.

mrossinek:
> Thanks Max. If you feel like a release note would be particularly useful to people, feel free to add one, but we probably don't really need one - this was announced as if it were public API, already was intended to be, and we don't typically have release notes for documentation-only changes.
> 
> Let me know either way and I'll tag for merge.

Sounds good :+1: I don't think a release note would add much value here so I would go with this as is

PR #13078, opened 2024-09-03
Avoid `ExtraInstructionAttributes` allocation on `unit="dt"`

### LINKED ISSUE #12900: Stop storing default `unit="dt"` in rust space extra attributes by default
### What should we add?

Right now when adding gates to a circuit from Python the default value of the `unit` field ends up meaning we populate the rust space extra attributes field in the circuit instruction and filling it with a unit of "dt". This is just wasted memory that we shouldn't need to waste since everything has an implicit default of `"dt"`, the extra attributes in the rust data model was designed only to store custom values for the mutable state for instructions. We should update the logic when adding gates from Python to Rust space to avoid setting extra attributes if `unit="dt"`.

jakelishman:
### Summary

The default value for `Instruction.unit` is `"dt"`.  Previously, the `OperationFromPython` extraction logic would only suppress allocation of the extra instruction attributes if all the contained fields were `None`, but `None` is not actually a valid value of `Instruction.unit` (which must be a string).  This meant that `OperationFromPython` would always allocate and store extra attributes, even for the default cases. This did not affect standard gates appended using their corresponding `QuantumCircuit` methods (since no Python-space extraction is performed in that case), but did affect standard calls to `append`, or anything else that entered from Python space.

This drastically reduces the memory usage of circuits built by `append`-like methods. Ignoring the inefficiency factor of the heap-allocation implementation, this saves 66 bytes plus small-allocation overhead for 2-byte heap allocations (another 14 bytes on macOS, but will vary depending on the allocator) per standard instruction, which is on the order of 40% memory-usage reduction.

### Details and comments

I'm using the same sort of microbenchmarking script I've been using since #12730, but now modified to use `append` instead of the special methods on `QuantumCircuit`:
```python
from qiskit.circuit import QuantumCircuit, library

QUBITS = 1000
REPS = 3000

def main_methods():
    qc = QuantumCircuit(QUBITS)
    for _ in range(REPS):
        for q in qc.qubits:
            qc.rz(0.0, q)
        for q in qc.qubits:
            qc.rx(0.0, q)
        for q in qc.qubits:
            qc.rz(0.0, q)
        for a, b in zip(qc.qubits[:-1], qc.qubits[1:]):
            qc.cx(a, b)


def main_append():
    qc = QuantumCircuit(QUBITS)
    rz = library.RZGate(0.0)
    rx = library.RXGate(0.0)
    cx = library.CXGate()
    for _ in range(REPS):
        for q in qc.qubits:
            qc.append(rz, (q,))
        for q in qc.qubits:
            qc.append(rx, (q,))
        for q in qc.qubits:
            qc.append(rz, (q,))
        for qs in zip(qc.qubits[:-1], qc.qubits[1:]):
            qc.append(cx, qs)
```

The memory usage of `main_append` for the parent of this PR is approximately 2.3GB on both macOS, and Linux with glibc, whereas with the PR it drops to 1.35GB on macOS and 1.06GB on Linux/glibc.  I suspect there's something additional going on in the macOS one, because while the Linux/glibc one drops to match the memory usage of `main_methods` (as expected), macOS remains 300MB higher.  It might have been different Python versions - I used 3.10 on macOS and 3.12 on Linux.

### DIFF
--- a/crates/circuit/src/circuit_instruction.rs
+++ b/crates/circuit/src/circuit_instruction.rs
@@ -17,7 +17,7 @@ use numpy::IntoPyArray;
 use pyo3::basic::CompareOp;
 use pyo3::exceptions::{PyDeprecationWarning, PyTypeError};
 use pyo3::prelude::*;
-use pyo3::types::{PyList, PyTuple, PyType};
+use pyo3::types::{PyList, PyString, PyTuple, PyType};
 use pyo3::{intern, IntoPy, PyObject, PyResult};
 
 use smallvec::SmallVec;
@@ -62,6 +62,20 @@ impl ExtraInstructionAttributes {
         } else {
             None
         }
+    }
+
+    /// Get the Python-space version of the stored `unit`.  This evalutes the Python-space default
+    /// (`"dt"`) value if we're storing a `None`.
+    pub fn py_unit(&self, py: Python) -> Py<PyString> {
+        self.unit
+            .as_deref()
+            .map(|unit| <&str as IntoPy<Py<PyString>>>::into_py(unit, py))
+            .unwrap_or_else(|| Self::default_unit(py).clone().unbind())
+    }
+
+    /// Get the Python-space default value for the `unit` field.
+    pub fn default_unit(py: Python) -> &Bound<PyString> {
+        intern!(py, "dt")
     }
 }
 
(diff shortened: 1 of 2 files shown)


crates/circuit/src/circuit_instruction.rs
```diff
@@ -63,6 +63,20 @@ impl ExtraInstructionAttributes {
...
+    /// (`"dt"`) value if we're storing a `None`.
+    pub fn py_unit(&self, py: Python) -> Py<PyString> {
+        self.unit
+            .as_deref()
+            .map(|unit| <&str as IntoPy<Py<PyString>>>::into_py(unit, py))
+            .unwrap_or_else(|| Self::default_unit(py).clone().unbind())
+    }
+
+    /// Get the Python-space default value for the `unit` field.
+    pub fn default_unit(py: Python) -> &Bound<PyString> {
```
mtreinish, line 77: I feel like I would put an `#[inline]` on this since the compiler is less likely to inline a public function. But it doesn't really matter in practice, especially since this is the python path so that won't help anything.

mtreinish (approved):
LGTM, this is a straightforward improvement and good to see the memory overhead reductions in practice.

PR #13083, opened 2024-09-04
Bug fix in `HoareOptimizer`

### LINKED ISSUE #13079: `HoareOptimizer()` changes circuit semantics
### Environment

- **Qiskit version**:  1.2.0
- **Python version**: 3.10.12
- **Operating system**: Ubuntu 23.10 LTS

### What is happening?

[Ilan](https://github.com/IlanIwumbwe) and I found this issue when testing our circuit with specific transpiler passes:

When passing a specific circuit through the `HoareOptimizer()`, there seems to be a change to the semantics of the circuit.

Running the code below will produce 2 state-vectors: one before passing the circuit through `HoareOptimizer()` and one after. They are not the same, and the subsequent circuit print shows that a CX gate was removed where it shouldn't have.

```py
from helpers.qiskit_helpers import simulate_circuit
from qiskit.transpiler.passes import *
from qiskit.transpiler.passmanager import PassManager
from numpy import vdot

def simulate_circuit(qc : QuantumCircuit):
    state = Statevector.from_int(0, 2**qc.num_qubits)
    state = state.evolve(qc)
    return state

main_circ = QuantumCircuit(3)
# Applying gates 
main_circ.cx(1, 0)
main_circ.x(2)
main_circ.x(0)
main_circ.cx(2, 0)

passes = PassManager(
    [HoareOptimizer()]
)

pass_circ = passes.run(main_circ)
sv0 = simulate_circuit(main_circ)
sv1 = simulate_circuit(pass_circ)
    
# Compare the circuit statevector before and after the pass
if(round(abs(vdot(sv0, sv1)), 6) != 1):
    print("Failed test!\n")
    print(sv0)
    print(sv1)
else:
    print("Passed test!\n")

print(main_circ)
print(pass_circ)
```

The circuit before optimization:
```bash
     ┌───┐┌───┐┌───┐
q_0: ┤ X ├┤ X ├┤ X ├
     └─┬─┘└───┘└─┬─┘
q_1: ──■─────────┼──
     ┌───┐       │  
q_2: ┤ X ├───────■──
     └───┘          
```

The circuit after:

```bash
     ┌───┐
q_0: ┤ X ├
     └───┘
q_1: ─────
     ┌───┐
q_2: ┤ X ├
     └───┘
```

### What should happen?

The circuit should have been optimized in a way that produced the same resulting state-vector. 

Might be related to this issue #4981 that was closed.

### Any suggestions?

_No response_

alexanderivrii:
### Summary

Fixes #13079.

### Detailed Comments

The problem was due to first simplifying a controlled-gate by removing its controls (in the example: the control qubit of a CX-gate is at `|1>`, allowing to replace the CX-gate by an X-gate), then incorrectly keeping track of the simplified gate (in order to store the new X-gate for later purposes, the code _mutated_ the original CX-gate in some weird way). 

The code was already using `substitute_node_with_dag` to replace the controlled-gate by the base-gate (I don't really like this, but I don't think we can use `substitute_node` as the simplified node has fewer qubits than the original node). To fix the problem, we can instead read off the simplified gate from `substitute_node_with_dag`'s output.

While this fixes the immediate problem, the code in `HoareOptimizer` is _very_ messy and quite possibly contains many other sources of bugs (for instance, I did not like the idea of iterating over the DAG's topological nodes, while removing and simplifying some of the nodes in-place) and not optimized for performance. A complete rewrite while porting to Rust would be ideal :).


qiskit/transpiler/passes/optimization/hoare_opt.py
```diff
@@ -212,11 +215,12 @@ def _traverse_dag(self, dag):
...
-                dag.substitute_node_with_dag(node, new_dag)
-                gate = gate.base_gate
-                node.op = gate.to_mutable()
-                node.name = gate.name
-                node.qargs = tuple((ctrlqb + trgtqb)[qi] for qi in qb_idx)
+                # We are replacing a node by a new node over a smaller number of qubits.
+                # This can be done using substitute_node_with_dag, which furthermore returns
+                # a mapping from old node ids to new nodes.
+                mapped_nodes = dag.substitute_node_with_dag(node, new_dag)
+                node = list(mapped_nodes.values())[0]
```
Cryoris: I think this should be slightly faster as you don't need to generate the entire list 🙂 
```suggestion
                node = next(iter(mapped_nodes.values()))
```

qiskit/transpiler/passes/optimization/hoare_opt.py
```diff
@@ -203,7 +203,10 @@ def _traverse_dag(self, dag):
         """
         import z3
 
-        for node in dag.topological_op_nodes():
+        # It does not look safe to iterate over DAG while removing and replacing
+        # some of its nodes.
```
Cryoris: Just an optional suggestion to rephrase this for outside the context of this PR 🙂 
```suggestion
        # Pre-generate all DAG nodes, since we later iterate over them while
        # potentially removing some of them.
```

qiskit/transpiler/passes/optimization/hoare_opt.py
alexanderivrii: Done in [d819a42](https://github.com/Qiskit/qiskit/pull/13083/commits/d819a424df9297eb2792dce77ef4079f9a9f8c17). Indeed, this form is better.

qiskit/transpiler/passes/optimization/hoare_opt.py
alexanderivrii: Done in [d819a42](https://github.com/Qiskit/qiskit/pull/13083/commits/d819a424df9297eb2792dce77ef4079f9a9f8c17).

Cryoris (approved):

PR #13086, opened 2024-09-04
Fix creation of registers in synthesis methods

### LINKED ISSUE #13041: `Clifford.to_circuit()` returns a circuit with emtpy `qregs`
### Environment

- **Qiskit version**: 1.2
- **Python version**: 3.10
- **Operating system**: Linux

### What is happening?

`Clifford.to_circuit()` returns a `QuantumCircuit` with empty `qregs`. I don't know if this is an expected behaviour, but it has changed since qiskit 1.1.2.

```
from qiskit.quantum_info import Clifford

qc = QuantumCircuit(2)
qc.h(0)
qc.cx(0, 1)
cli = Clifford(qc)
cli.to_circuit().qregs
```

### What should happen?

I would expect the `QuantumCircuit` to contain one `QuantumRegister` with the correct size, as it was the case in qiskit 1.1.2

```
>>> cli.to_circuit().qregs
[QuantumRegister(2, 'q')]
```

### Any suggestions?

_No response_

Cryoris:
### Summary

Fixes #13041 (and the same problem in other places, but no one complained yet 🙂).

### Details and comments

With the moving the circuit creation to Rust-space, we created circuits that did not have any quantum (or classical) registers, which led to issues as #13041. This PR adds an argument `add_regs` to `QuantumCircuit._from_circuit_data` (default `False`) which allows to add these registers and create backward-compatible circuits. In places where the final circuit was created from circuit data, this argument is set to `True`.


Cryoris:
I'm not 100% happy with this solution of `compose`-ing the circuit onto a new one (even if we can skip copying the ops), but I didn't find another solution to attach a register/index to each of the qubits (which we need for `QuantumCircuit.__eq__` to work as before). 

Another solution could be to change that `__eq__` only compares bits if they _both_ are either new style (no register/index) or old style (register/index available). But the `compose` solution seems less intrusive, especially for a bugfix PR we want to include in a bugfix release.

qiskit/circuit/quantumcircuit.py
```diff
@@ -1152,17 +1152,26 @@ def __init__(
...
         out = QuantumCircuit()
-        out._data = data
         out._qubit_indices = {bit: BitLocations(index, []) for index, bit in enumerate(data.qubits)}
         out._clbit_indices = {bit: BitLocations(index, []) for index, bit in enumerate(data.clbits)}
+        out._data = data
+
+        # To add consistent registers, we compose the circuit onto a new one with registers,
+        # as each instruction must be mapped to qubits that belong to a register (in the circuit
+        # data coming from Rust, the qubits are not attached to any register).
+        if add_regs:
```
mtreinish: Why not do something like:

```python
qr = QuantumRegister(bits=out.qubits)
out.add_register(qr)
```

Cryoris: I tried, but then the qubits in the `CircuitData` are unaware that they should belong to `qr` and doing circuit comparisons still fails. For example if we had
```python
py_circ = QuantumCircuit(1)
py_circ.x(0)

rust_circ = QuantumCircuit._from_circuit_data(data, add_regs=True)  # data just contains the X gate on qubit 0

rust_circ == py_circ
```
would still evaluate to `False`.

alexanderivrii: Would it be possible to compute `num_qubits` and `num_clbits` from `data`, and change the line `out = QuantumCircuit()` to `out = QuantumCircuit(num_qubits, num_clbits)`?

alexanderivrii:
Thanks, @Cryoris! If I recall correctly, many of the circuit definition functions in `circuit/library` also create registers named `q_*` (and honestly I never really understood why we need this). Are you keeping this convention throughout your work on circuit library refactoring?

qiskit/circuit/quantumcircuit.py
Cryoris: The problem is that the instructions in `CircuitData` use qubits, that are not attached to any registers. If we do `out = QuantumCircuit(num_qubits, num_clbits)`, then we create a register for the circuit, but the qubits in the `CircuitData` do not belong to these (and we actually end up with twice the number of qubits) 🙂

Cryoris:
It now works without an additional `compose` and explicit call to `add_register`, but there's still a tiny overhead to adding registers coming from the register creation inside `_from_circuit_data`. However this only falls into weight when the circuit creation is super fast. For example here are timings repeated over 100 trials:
```
Permutation synthesis: slower (worse as size get's larger)
-- n = 10
(main) 1.91e-05 +- 5.16e-05s
(this PR) 2.26e-05 +- 5.01e-05s
-- n = 50
(main) 4.30e-05 +- 2.63e-05s
(this PR) 6.16e-05 +- 5.36e-05s
-- n = 100
(main) 8.63e-05 +- 8.49e-05s
(this PR) 1.11e-04 +- 9.73e-05s
-- n = 1000
(main) 2.78e-04 +- 1.05e-03s
(this PR) 1.63e-03 +- 3.36e-03s

Clifford: same speed
-- n = 10
(main) 1.24e-04 +- 8.60e-05s
(this PR) 1.24e-04 +- 6.16e-05s
-- n = 50
(main)  3.05e-03 +- 1.50e-04s
(this PR) 3.05e-03 +- 1.56e-04s
-- n = 100
(main) 1.80e-02 +- 7.58e-04s
(this PR) 1.80e-02 +- 4.37e-04s
-- n = 200
(main) 1.26e-01 +- 3.85e-03s
(this PR) 1.26e-01 +- 3.78e-03s
```
I think in practice this won't be an issue as HLS doesn't need the registers to synthesize building blocks within an existing circuit. We can add this fast path in a follow up, since that will require changing the API by adding an additional argument to the synthesis method, so we should avoid that in the backport.

Cryoris:
@alexanderivrii for backward compatibility it would probably make sense to keep it, but let's see if that adds a significant overhead 🙂

alexanderivrii (approved):
LGTM, thanks!

PR #13090, opened 2024-09-05
Add qubits and clbits to dag in `DAGCircuit::with_capacity`

ElePT:
### Summary
Follow-up to #13002 that adds qubits and clbits to the dag when `num_qubits` or `num_clbits` are non-zero. This would enable, for example, directly appending instructions using an iterator (https://github.com/Qiskit/qiskit/pull/13032) after creating a new rust `DAGCircuit`. The alternative is making users add the qubits/clbits manually.

I don't think this is 100% necessary but it definitely increases the usability of `with_capacity`, in fact, this is how it's done in  `CircuitData` ([L238-251](https://github.com/Qiskit/qiskit/blob/main/crates/circuit/src/circuit_data.rs#L238-L251)).

### DIFF
--- a/crates/circuit/src/dag_circuit.rs
+++ b/crates/circuit/src/dag_circuit.rs
@@ -6195,7 +6195,7 @@ impl DAGCircuit {
             num_vars * 2 +  // One input + output node per variable
             num_ops;
 
-        Ok(Self {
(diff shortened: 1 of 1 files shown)


ElePT:
Closing after internal discussion.

PR #13091, opened 2024-09-05
Oxidize commutative cancellation

### LINKED ISSUE #12270: Port `CommutativeCancellation` to Rust

sbrandhsn:
### Summary

fixes https://github.com/Qiskit/qiskit/issues/12270

### Details and comments

This replaces the Python implementation of `CommutativeCancellation` with a Rust implementation:

- As the Rust implementation does not use the `PassManager` to run its prerequisite `CommutationAnalysis`, the test case has been `test_callback_with_pass_requires` fixed. We might be able to remove that test case as the check is already covered in https://github.com/Qiskit/qiskit/blob/52733ae09f7ac9d2d5e2a674970ff95301d34bd6/test/python/transpiler/test_pass_scheduler.py#L111
- The global phase was probably incorrectly set in the Python implementation in https://github.com/Qiskit/qiskit/blob/f4ca088644b4461b856b4b49c33e23b04a995019/qiskit/transpiler/passes/optimization/commutative_cancellation.py#L195, I changed this in the Rust implementation
- Currently, control-flow operations are handled in Python - I'm looking into a Rust implementation for that too

### Speedups (this branch vs Qiskit 1.2)

image
random_circuits
rotations_circuits)

Top to bottom Clifford circuits, random circuits, random circuits with rotation gates only

Up to 5.5x speedup for Clifford circuits where the commutation relation of two gates is always determined by a look-up in an a priori constructed commutation library. Up to 7.0 speedup for random circuits as generated by `random_circuit` and up to 8x speedup when restricting that random circuit to rotational gates only. Note this benchmark compares end-to-end performance as given by
```
            pm = PassManager([CommutativeCancellation(target=be.target)])
            start_ts = time.time()
            pm.run(qc)
            end_ts = time.time()
```
instead of only comparing a python implementation of `CommutativeCancellation` with underlying Rust code vs a rust implementation of `CommutativeCancellation` with the same underlying code.


crates/accelerate/src/commutation_cancellation.rs
```diff
@@ -0,0 +1,285 @@
...
+pub(crate) fn cancel_commutations(
+    py: Python,
+    dag: &mut DAGCircuit,
+    commutation_checker: &mut CommutationChecker,
+    basis_gates: Option<HashSet<String>>,
+    target: Option<&Target>,
+) -> PyResult<()> {
+    let basis: HashSet<String> = if let Some(tar) = target {
+        HashSet::from_iter(tar.operation_names().map(String::from))
+    } else if let Some(basis) = basis_gates {
```
mtreinish: While we can do this from Rust, this actually adds overhead to every call of the pass, while if we do this in Python we only have the cost once per initialization. Since we run the cancellation pass as part of the optimization loop this could be a performance hit. I would just do the set collection for the target in Python and only pass `basis_gates` to the rust function.

crates/accelerate/src/commutation_cancellation.rs
```diff
@@ -0,0 +1,285 @@
...
+        basis
+    } else {
+        HashSet::new()
+    };
+
+    let _var_z_map: HashMap<&str, StandardGate> =
+        HashMap::from([("rz", RZGate), ("p", PhaseGate), ("u1", U1Gate)]);
```
mtreinish: I wonder if it would make more sense to have this as a `static VAR_Z_MAP: [(&str, StandardGate); 3]`. We only are doing 3 string comparisons of at most 2 characters it feels like a hashmap here is more expensive that just checking the 3 strings.

crates/accelerate/src/commutation_cancellation.rs
```diff
@@ -0,0 +1,285 @@
...
+        the value is the list of gates that share the same gate type, qubit, commutation set.
+      - For 2qbit gates the key: (gate_type, first_qbit, sec_qbit, first commutation_set_id,
+        sec_commutation_set_id), the value is the list gates that share the same gate type,
+        qubits and commutation sets.
+    */
+    let (commutation_set, node_indices) = analyze_commutations_inner(py, dag, commutation_checker)?;
+    let mut single_q_cancellation_sets: HashMap<(String, Qubit, usize), Vec<NodeIndex>> =
+        HashMap::new();
+    let mut two_q_cancellation_sets: HashMap<(String, Qubit, Qubit, usize, usize), Vec<NodeIndex>> =
+        HashMap::new();
```
mtreinish: We're working solely with `StandardGates` here instead of storing a `String` can't we store the `StandardGate`? We'd have to change `_gates` too but I think that's ok. We also can make this all static more easily by using `StandardGate` because it's a very fast u8 comparison under the covers.

mtreinish: I think we can define better types for these that are easier to work, a bit faster, and less error prone. Something like:

```rust
    #[derive(Hash, Eq, PartialEq, Debug)]
    enum GateOrRotation {
        Gate(StandardGate),
        ZRotation,
        XRotation,
    }           
    #[derive(Hash, Eq, PartialEq, Debug)]
    struct CancellationSetKey {
        gate: GateOrRotation,
        qubits: SmallVec<[Qubit; 2]>,
        com_set_index: usize,
        second_index: Option<usize>,
    }
 
    let mut cancellation_sets: HashMap<CancellationSetKey, Vec<NodeIndex>> = HashMap::new();
```

Would let us use one hashmap instead of two and maybe simplify the logic of generating the cancellation sets. Although we could easily do two hashmaps here too if that's preferable and just have two key types. The struct will be easier to work with because everything is named access and the enum will let us use `StandardGate` instead of strings.

crates/accelerate/src/commutation_cancellation.rs
```diff
@@ -0,0 +1,285 @@
...
+                                        .entry(q2_key)
+                                        .or_insert_with(Vec::new)
+                                        .push(*node);
+                                }
+                            }
+                        })
+                    }
+                })
+        }
+    });
```
mtreinish: I was curious and sketched out using a parallel iterator chain here to build the cancellation sets hashmap using the above suggestion on changing the types:

```rust
    let cancellation_sets: HashMap<CancellationSetKey, Vec<NodeIndex>> = (0..dag.num_qubits())
        .into_par_iter()
        .filter_map(|qubit| {
            let wire = Qubit(qubit as u32);
            if let Some(wire_commutation_set) = commutation_set.get(&Wire::Qubit(wire)) {
                Some(wire_commutation_set
                    .iter()
                    .enumerate()
                    .filter_map(|(com_set_index, com_set)| {
                        // This ensures that we only have DAGOPNodes in the current com_set, yuck...
                        if let NodeType::Operation(_node0) = &dag.dag[*com_set.first().unwrap()] {
                            Some(com_set.iter().filter_map(|node| {
                                let op = match &dag.dag[*node] {
                                    NodeType::Operation(instr) => instr,
                                    _ => panic!("Unexpected type in commutation set."),
                                };
                                let num_qargs = dag.get_qargs(op.qubits).len();
                                // no support for cancellation of parameterized gates
                                if op
                                    .params_view()
                                    .iter()
                                    .all(|p| !matches!(p, Param::ParameterExpression(_)))
                                {
                                    let op_name = op.op.name().to_string();
                                    let gate = op.op.standard_gate();
                                    if num_qargs == 1usize && GATES.contains(&gate) {
                                        let key = CancellationSetKey {
                                            gate: GateOrRotation::Gate(gate),
                                            qubits: smallvec![wire],
                                            com_set_index,
                                            second_index: None,
                                        };
                                        Some((key, *node))
                                    } else if num_qargs == 1usize && _z_rotations.contains(op_name.as_str()) {
                                        let key = CancellationSetKey {
                                            gate: GateOrRotation::Z_ROTATION,
                                            qubits: smallvec![wire],
                                            com_set_index,
                                            second_index: None,
                                        };
                                        Some((key, *node))
                                    } else if num_qargs == 1usize && _x_rotations.contains(op_name.as_str()) {
                                        let key = CancellationSetKey {
                                            gate: GateOrRotation::X_ROTATION,
                                            qubits: smallvec![wire],
                                            com_set_index,
                                            second_index: None,
                                        };
                                        Some((key, *node))
                                    }
                                    // Don't deal with Y rotation, because Y rotation doesn't commute with
                                    // CNOT, so it should be dealt with by optimized1qgate pass
                                    else if num_qargs == 2usize
                                        && dag.get_qargs(op.qubits).first().unwrap() == &wire
                                    {
                                        let second_qarg = dag.get_qargs(op.qubits)[1];
                                        let key = CancellationSetKey {
                                            gate: GateOrRotation::Gate(gate),
                                            qubits: smallvec![wire, second_qarg],
                                            com_set_index,
                                            second_index: Some(*node_indices.get(&(*node, Wire::Qubit(second_qarg))).unwrap()),
                                        };
                                        Some((key, *node))
                                    } else {
                                        None
                                    }
                                } else {
                                    None
                                }
                        }).collect::<Vec<(CancellationSetKey, NodeIndex)>>())
                    } else {
                        None
                    }
                })
                .flat_map(|entries| entries.into_iter()))
            } else {
                None
            }
        })
        .fold(|| HashMap::new(), |map, item| {
            for (key, value) in item {
                map.entry(key).and_modify(|indices: &mut Vec<NodeIndex>| indices.push(value)).or_insert_with(|| vec![value]);
            }
            map
        })
        .reduce(|| HashMap::new(), |outer_map, inner_map| {
            for (key, value) in inner_map.iter_mut() {
                outer_map.entry(*key).and_modify(|indices| indices.append(value)).or_insert_with(|| *value);
            }
            outer_map
        });
```

This doesn't fully compile yet because I haven't changed the logic below it to work with a single hashmap yet. But it at least gets passed the type checking (which probably means the borrow checker will hate it :D). I'm skeptical doing this in parallel actually helps performance here though because it results in several intermediate allocations that could be avoided. It was more just an exercise for me to see how it could be done. But this might be useful as a reference for other changes so I decided to post it even though it's not complete.

crates/accelerate/src/commutation_cancellation.rs
```diff
@@ -0,0 +1,285 @@
...
+
+                let node_qargs = dag.get_qargs(node_op.qubits);
+                if node_op
+                    .extra_attrs
+                    .as_deref()
+                    .is_some_and(|attr| attr.condition.is_some())
+                    || node_qargs.len() > 1
+                    || &node_qargs[0] != run_qarg
+                {
+                    panic!("internal error");
```
mtreinish: While not helpful for debugging at all this was a Python space error before, I think it's probably worth returning a `RuntimeError` to python here.

crates/accelerate/src/commutation_cancellation.rs
```diff
@@ -0,0 +1,285 @@
...
+                }
+
+                let node_angle = if ROTATION_GATES.contains(&node_op_name) {
+                    match node_op.params_view().first() {
+                        Some(Param::Float(f)) => *f,
+                        _ => panic!(
+                            "Rotational gate with parameter expression encoutned in cancellation"
+                        ),
```
mtreinish: This should be returned to Python as a runtime error so it's "user actionable"

crates/accelerate/src/commutation_cancellation.rs
```diff
@@ -0,0 +1,285 @@
...
+                    PI / 2.0
+                } else if EIGHTH_TURNS.contains(&node_op_name) {
+                    PI / 4.0
+                } else {
+                    panic!("Angle for operation {node_op_name} is not defined")
+                };
+                total_angle += node_angle;
+
+                if let Some(definition) = node_op.op.definition(node_op.params_view()) {
+                    //TODO check for PyNone global phase?
```
mtreinish: Global phase can't be `None` in rust. It's always a Param type: https://github.com/Qiskit/qiskit/blob/main/crates/circuit/src/circuit_data.rs#L104

crates/accelerate/src/commutation_cancellation.rs
```diff
@@ -0,0 +1,285 @@
...
+                    total_phase += match definition.global_phase {Param::Float(f) => f, _ => panic!("PackedInstruction with definition has no global phase set as floating point number")};
+                }
+            }
+
+            let new_op = if cancel_key.0 == Z_ROTATION {
+                z_var_gate.unwrap()
+            } else if cancel_key.0 == X_ROTATION {
+                &RXGate
+            } else {
+                panic!("impossible case!");
```
mtreinish: If this really impossible then I'd use `unreachable!()` here.  But this is also a really good case to have the `enum`  I described above. because it makes all the possible values here very explicit.

crates/accelerate/src/commutation_cancellation.rs
```diff
@@ -0,0 +1,285 @@
...
+            */
+        }
+    }
+
+    Ok(())
+}
+
+/// Wrap angle into interval [-π,π). If within atol of the endpoint, clamp to -π
+#[inline]
+fn mod_2pi(angle: f64, atol: f64) -> f64 {
```
mtreinish: I feel like we have a function like this in a bunch of places already. One not just call it instead of copying it?

qiskit/transpiler/passes/optimization/commutative_cancellation.py
```diff
@@ -70,12 +64,11 @@ def __init__(self, basis_gates=None, target=None):
 
         # build a commutation checker restricted to the gates we cancel -- the others we
         # do not have to investigate, which allows to save time
-        commutation_checker = CommutationChecker(
+        self.commutation_checker = CommutationChecker(
```
mtreinish: Just so people don't assume this is publicly accessible:

```suggestion
        self._commutation_checker = CommutationChecker(
```

crates/accelerate/src/commutation_cancellation.rs
```diff
@@ -0,0 +1,285 @@
...
+
+            for node in cancel_set {
+                dag.remove_op_node(*node);
+            }
+
+            //TODO do we need this due to numerical instability?
+            /*
+                if np.mod(total_angle, (2 * np.pi)) < _CUTOFF_PRECISION:
+                    dag.remove_op_node(run[0])
+            */
```
mtreinish: In it's current form this isn't needed since you're removing all the gates regardless of the angle so I would just delete the comment. If this is needed we'd have to update the removal loop to account for this.

```suggestion
```

crates/accelerate/src/commutation_cancellation.rs
```diff
@@ -0,0 +1,285 @@
...
+        } else if cancel_set.len() > 1 && (cancel_key.0 == Z_ROTATION || cancel_key.0 == X_ROTATION)
+        {
+            let run_op = match &dag.dag[*cancel_set.first().unwrap()] {
+                NodeType::Operation(instr) => instr,
+                _ => panic!("Unexpected type in commutation set run."),
+            };
+
+            let run_qarg = dag.get_qargs(run_op.qubits).first().unwrap();
+            let mut total_angle: f64 = 0.0f64;
+            let mut total_phase: f64 = 0.0f64;
```
mtreinish: You don't need the suffix because you've set a type annotation here.
```suggestion
            let mut total_angle: f64 = 0.0;
            let mut total_phase: f64 = 0.0;
```

crates/accelerate/src/commutation_analysis.rs
```diff
@@ -50,7 +50,7 @@ const MAX_NUM_QUBITS: u32 = 3;
 ///     commutation_set = {0: [[0], [2, 3], [4], [1]]}
 ///     node_indices = {(0, 0): 0, (1, 0): 3, (2, 0): 1, (3, 0): 1, (4, 0): 2}
 ///
-fn analyze_commutations_inner(
+pub fn analyze_commutations_inner(
```
mtreinish: I'm fine with making this `pub` but `pub(crate)` is also an option since this is only going to be used inside the same crate for now (this will likely change in the medium term towards Qiskit 2.0).

mtreinish (commented):
Thanks for doing this, this has the potential to improve performance quite a bit, not only by speeding up the cancellation, but also by avoiding all the python overhead for the analysis. I've done a quick first pass over the code. I left a few inline comments, besides those the big thing I think we want to be careful for here is making sure we don't panic needlessly and instead return Python exceptions when there is a possible error condition. Right now the code seems a bit too likely to just call `panic!()` which won't be use actionable or catchable and also exits hard.

mtreinish:
Oh also can you please rebase this on main so the commit history is just the local changes. While the final diff is ok, it makes the commit message and authors list for this PR the sum of all the commits on the branch, which is much larger than the actual work being done here.

crates/accelerate/src/commutation_cancellation.rs
sbrandhsn: I changed this to a Python error. It was in the original code and made me chuckle - hence I thought it is worth keeping it like this. :-) Let me know if you want it to be `unreachable!` instead.

crates/accelerate/src/commutation_cancellation.rs
sbrandhsn: Yeah, I did not understand why it would be needed in the original code and thought it may be related to numerical instability. I will remove it.

crates/accelerate/src/commutation_cancellation.rs
```diff
@@ -0,0 +1,291 @@
...
+                    } else if EIGHTH_TURNS.contains(&node_op_name) {
+                        PI / 4.0
+                    } else {
+                        panic!("Angle for operation {node_op_name} is not defined")
+                    };
+                    total_angle += node_angle;
+
+                    if let Some(definition) = node_op.op.definition(node_op.params_view()) {
+                        total_phase += match definition.global_phase() {Param::Float(f) => f, _ => panic!("PackedInstruction with definition has no global phase set as floating point number")};
+                    }
```
mtreinish: As the logic is right now this should return a python error:

```suggestion
                        total_phase += match definition.global_phase() {
                            Param::Float(f) => f,
                             _ => return Err(QiskitError::new_err(format!("PackedInstruction with definition has no global phase set as floating point number")))
                        };
                    }
```

Although realistically if you make my above suggestion about how to get the definition directly from `StandardGate` I think you can change this to just use `unreachable!()` since the generator code for definitions of standard gates never has a parameterized phase, it's always a float `0.0`. Something like:

```suggestion
                        let Param::Float(new_phase) = definition.global_phase() else { unreachable!() };
                        total_phase += new_phase
```
(although maybe set an error message saying standard gate phase is never parameterized)

would do the trick. But we can't make this change unless we're working with a `StandardGate` when we call `definition()`.

crates/accelerate/src/commutation_cancellation.rs
```diff
@@ -0,0 +1,291 @@
...
+                        (*new_op, &[total_angle]),
+                        *cancel_set.first().unwrap(),
+                    );
+                    let new_node = match &dag.dag[new_index] {
+                        NodeType::Operation(instr) => instr,
+                        _ => panic!("Unexpected type in commutation set run."),
+                    };
+
+                    if let Some(definition) = new_node.op.definition(new_node.params_view()) {
+                        match definition.global_phase() {Param::Float(f) => *f, _ => panic!("PackedInstruction with definition has no global phase set as floating point number")}
```
mtreinish: Same comment above with the global phase handling.

crates/accelerate/src/commutation_cancellation.rs
```diff
@@ -0,0 +1,291 @@
...
+                        || &node_qargs[0] != run_qarg
+                    {
+                        panic!("internal error");
+                    }
+
+                    let node_angle = if ROTATION_GATES.contains(&node_op_name) {
+                        match node_op.params_view().first() {
+                            Some(Param::Float(f)) => *f,
+                            _ => return Err(QiskitError::new_err(format!(
+                                "Rotational gate with parameter expression encoutned in cancellation {:?}",
```
mtreinish:
```suggestion
                                "Rotational gate with parameter expression encountered in cancellation {:?}",
```

crates/accelerate/src/commutation_cancellation.rs
```diff
@@ -0,0 +1,291 @@
...
+    (0..dag.num_qubits() as u32).for_each(|qubit| {
+        let wire = Qubit(qubit);
+        if let Some(wire_commutation_set) = commutation_set.get(&Wire::Qubit(wire)) {
+            wire_commutation_set
+                .iter()
+                .enumerate()
+                .for_each(|(com_set_idx, com_set)| {
+                    // This ensures that we only have DAGOPNodes in the current com_set, yuck...
+                    if let NodeType::Operation(_node0) = &dag.dag[*com_set.first().unwrap()] {
+                        com_set.iter().for_each(|node| {
```
mtreinish: Besides removing the commentary on the syntax, I'm not sure exactly what the intent is with this code. Are you trying to filter sets here that don't start with an `Operation` node? There are a couple of options if so which might  be more succinct, such as doing:

```rust
for (com_set_idx, com_set) in  wire_commutation_set.iter().enumerate() {
    if Some(node) = com_set.first() {
        if !matches!(NodeType::Operation(_), dag.dag[node]) {
            continue
        }
    } else {
        continue
    }
```

or if you'd like to stick with an iterator you can do:

```rust
.filter(|(com_set_idx, com_set)| {
    if Some(node) = com_set.first() {
        if matches!(NodeType::Operation(_), dag.dag[node]) {
            true
        } else {
           false
        }
    } else {
        false
    }
})
.for_each(|(com_set_idx, com_set)| {
```

Although this depends on what the logic you're trying to do is. If you don't want to tolerate a single non-op node in the commutation set then there is different logic I'd probably use. But either way I'd suggest making either change here because besides not being a potential source of a panic it also removes one layer of nesting from the for loop here.

crates/accelerate/src/commutation_cancellation.rs
```diff
@@ -0,0 +1,291 @@
...
+                            let op = match &dag.dag[*node] {
+                                NodeType::Operation(instr) => instr,
+                                _ => panic!("Unexpected type in commutation set."),
+                            };
```
mtreinish: Related to above would it just make more sense to have this be a skip condition instead of a panic and remove the `first()` logic. So if we encounter a `non-Operation` node type we just ignore this set?

crates/accelerate/src/commutation_cancellation.rs
```diff
@@ -0,0 +1,291 @@
...
+                            let num_qargs = dag.get_qargs(op.qubits).len();
+                            // no support for cancellation of parameterized gates
+                            if op
+                                .params_view()
+                                .iter()
+                                .all(|p| !matches!(p, Param::ParameterExpression(_)))
+                            {
```
mtreinish: There is `op.is_parameterized()` here which you can probably use here instead with:

```suggestion
                            if !op.is_parameterized() {
```

crates/accelerate/src/commutation_cancellation.rs
```diff
@@ -0,0 +1,291 @@
...
+        let wire = Qubit(qubit);
+        if let Some(wire_commutation_set) = commutation_set.get(&Wire::Qubit(wire)) {
+            wire_commutation_set
+                .iter()
+                .enumerate()
+                .for_each(|(com_set_idx, com_set)| {
+                    // This ensures that we only have DAGOPNodes in the current com_set, yuck...
+                    if let NodeType::Operation(_node0) = &dag.dag[*com_set.first().unwrap()] {
+                        com_set.iter().for_each(|node| {
+                            let op = match &dag.dag[*node] {
```
mtreinish: Personally I'd name this `instr` instead of `op` because it's a bit confusing to have `op` as type `PackedInstruction`. When I saw `op` below I thought for a second it was a `PackedOperation` or something with the `Operation` trait implemented.

crates/accelerate/src/commutation_cancellation.rs
```diff
@@ -0,0 +1,291 @@
...
+                            };
+                            let num_qargs = dag.get_qargs(op.qubits).len();
+                            // no support for cancellation of parameterized gates
+                            if op
+                                .params_view()
+                                .iter()
+                                .all(|p| !matches!(p, Param::ParameterExpression(_)))
+                            {
+                                if let Some(op_gate) = op.op.try_standard_gate() {
+                                    if num_qargs == 1usize && SUPPORTED_GATES.contains(&op_gate) {
```
mtreinish: You don't actually need the `usize` suffix here, the compiler is smart enough to infer it because `num_qargs` is a usize.

```suggestion
                                    if num_qargs == 1 && SUPPORTED_GATES.contains(&op_gate) {
```

crates/accelerate/src/commutation_cancellation.rs
```diff
@@ -0,0 +1,291 @@
...
+            }
+            if matches!(cancel_key.gate, GateOrRotation::ZRotation)
+                || matches!(cancel_key.gate, GateOrRotation::XRotation)
+            {
+                let run_op = match &dag.dag[*cancel_set.first().unwrap()] {
+                    NodeType::Operation(instr) => instr,
+                    _ => panic!("Unexpected type in commutation set run."),
+                };
+
+                let run_qarg = dag.get_qargs(run_op.qubits).first().unwrap();
```
mtreinish: It's a slice so you can just get this with an index:
```suggestion
                let run_qarg = dag.get_qargs(run_op.qubits)[0];
```

mtreinish: I'd be inclined to just simplify this as:

```suggestion
                let run_qarg = match &dag.dag[*cancel_set.first().unwrap()] {
                    NodeType::Operation(instr) => dag.get_qargs(instr.qubits)[0],
                    _ => panic!("Unexpected type in commutation set run."),
                };
```

crates/accelerate/src/commutation_cancellation.rs
```diff
@@ -0,0 +1,291 @@
...
+                            )))
+                        }
+                    } else if HALF_TURNS.contains(&node_op_name) {
+                        PI
+                    } else if QUARTER_TURNS.contains(&node_op_name) {
+                        PI / 2.0
+                    } else if EIGHTH_TURNS.contains(&node_op_name) {
+                        PI / 4.0
+                    } else {
+                        panic!("Angle for operation {node_op_name} is not defined")
```
mtreinish: I feel like this should be a python error because it feels like it's user actionable.

crates/accelerate/src/commutation_cancellation.rs
```diff
@@ -0,0 +1,291 @@
...
+
+                let new_op_phase: f64 = if gate_angle.abs() > _CUTOFF_PRECISION {
+                    let new_index = dag.insert_1q_on_incoming_qubit(
+                        (*new_op, &[total_angle]),
+                        *cancel_set.first().unwrap(),
+                    );
+                    let new_node = match &dag.dag[new_index] {
+                        NodeType::Operation(instr) => instr,
+                        _ => panic!("Unexpected type in commutation set run."),
+                    };
```
mtreinish: Why do you need to do this if we defined `new_op` above? Can't we just get the definition and global phase directly from `new_op` with `new_op.definition(&[total_angle])`?

test/python/transpiler/test_commutative_cancellation.py
```diff
@@ -76,7 +77,7 @@ def test_all_gates(self):
         expected = QuantumCircuit(qr)
         expected.append(RZGate(2.0), [qr[0]])
         expected.rx(1.0, qr[0])
-
+        expected.global_phase = 0.5
```
mtreinish, line 80: I might have missed this in the PR summary message, but was this caused by the change in logic you made around the global phase handling?

mtreinish (commented):
Thanks for the quick updates. I took another pass through the code after the last updates. This is looking a lot better now and much less error prone using an enum and rust types instead of string comparisons for everything. I'm still a bit concerned how many `panic!()` calls there are though, I feel like a fair number of them can be avoided though, and either eliminated with some logical changes or replaced with Python errors. I left some inline comments with suggestions on how to do this.

crates/accelerate/src/commutation_cancellation.rs
jakelishman: Unless I'm missing something: one of Rz or P (whichever isn't exactly equal to u1) can have a parametric phase if the angle is parametric. `GlobalPhaseGate` definitely can (if the angle is parametric).

mtreinish: Ah good point, I thought there was one, but I scanned through the match statement yesterday and missed it on rzgate. But regardless this statement still holds since the angle is always a float at this point

crates/accelerate/src/commutation_cancellation.rs
sbrandhsn: Whoops, forgot to remove the comment - sorry! `com_set` only contains commuting entries if the first element is an operation. I don't think this logic is ideal but I also didn't want to refactor it to something that may take more time. I changed the code to the first example above.

crates/accelerate/src/commutation_cancellation.rs
sbrandhsn: We would deviate from the original code by doing so, but this is fine for me. Essentially, these `panic!` instructions check whether the source of `com_set` follows the contract/structure assumed in `CommutativeCancellation`. I think I would like to make the format checks explicit instead of simply passing over malformed `com_set` - what do you think?

test/python/transpiler/test_commutative_cancellation.py
sbrandhsn: Yes. The previous code would only consider the last replaced gate when setting the new `global_phase`. I think this was essentially a typo of 
```
global_phase = new_phase
```
instead of 
```
global_phase += new_phase
```

crates/accelerate/src/commutation_cancellation.rs
```diff
@@ -0,0 +1,280 @@
...
+use qiskit_circuit::dag_circuit::{DAGCircuit, NodeType, Wire};
+use qiskit_circuit::operations::StandardGate::{
+    CXGate, CYGate, CZGate, HGate, PhaseGate, RXGate, RZGate, SGate, TGate, U1Gate, XGate, YGate,
+    ZGate,
+};
+use qiskit_circuit::operations::{Operation, Param, StandardGate};
+use qiskit_circuit::Qubit;
+use rustworkx_core::petgraph::stable_graph::NodeIndex;
+use smallvec::{smallvec, SmallVec};
+use std::f64::consts::PI;
```
mtreinish: Small nit, just grouping the imports here between stdlib, external crates, and internal crates:

```suggestion
use std::f64::consts::PI;

use hashbrown::{HashMap, HashSet};
use pyo3::prelude::*;
use pyo3::{pyfunction, pymodule, wrap_pyfunction, Bound, PyResult, Python};
use rustworkx_core::petgraph::stable_graph::NodeIndex;
use smallvec::{smallvec, SmallVec};

use qiskit_circuit::dag_circuit::{DAGCircuit, NodeType, Wire};
use qiskit_circuit::operations::StandardGate::{
    CXGate, CYGate, CZGate, HGate, PhaseGate, RXGate, RZGate, SGate, TGate, U1Gate, XGate, YGate,
    ZGate,
};
use qiskit_circuit::operations::{Operation, Param, StandardGate};
use qiskit_circuit::Qubit;

use crate::commutation_analysis::analyze_commutations_inner;
use crate::commutation_checker::CommutationChecker;
use crate::{euler_one_qubit_decomposer, QiskitError};
```

crates/accelerate/src/commutation_cancellation.rs
```diff
@@ -0,0 +1,280 @@
...
+        if let Some(wire_commutation_set) = commutation_set.get(&Wire::Qubit(wire)) {
+            for (com_set_idx, com_set) in wire_commutation_set.iter().enumerate() {
+                if let Some(&nd) = com_set.first() {
+                    if !matches!(dag.dag[nd], NodeType::Operation(_)) {
+                        continue;
+                    }
+                } else {
+                    continue;
+                }
+                com_set.iter().for_each(|node| {
```
mtreinish: Personally I'd probably just use a normal for loop for this instead of `for_each()` but it functionally doesn't really matter. I typically reserve `for_each()` when I have a longer iterator chain that has maps and stuff and then need to iterate over that at the end.

If you did use a normal for loop you could reduce the nesting one layer and do:

```rust
if instr.is_parameterized() {
    continue;
}
```

which also would match the python logic more closely.

crates/accelerate/src/commutation_cancellation.rs
```diff
@@ -0,0 +1,280 @@
...
+                                        com_set_index: com_set_idx,
+                                        second_index: None,
+                                    })
+                                    .or_insert_with(Vec::new)
+                                    .push(*node);
+                            }
+                            // Don't deal with Y rotation, because Y rotation doesn't commute with
+                            // CNOT, so it should be dealt with by optimized1qgate pass
+                            if num_qargs == 2
+                                && dag.get_qargs(instr.qubits).first().unwrap() == &wire
```
mtreinish:
```suggestion
                                && dag.get_qargs(instr.qubits)[0] == &wire
```

crates/accelerate/src/commutation_cancellation.rs
```diff
@@ -0,0 +1,280 @@
...
+            if let GateOrRotation::Gate(g) = cancel_key.gate {
+                if SUPPORTED_GATES.contains(&g) {
+                    for &c_node in &cancel_set[0..(cancel_set.len() / 2) * 2] {
+                        dag.remove_op_node(c_node);
+                    }
+                }
+            }
+            if matches!(cancel_key.gate, GateOrRotation::ZRotation) && z_var_gate.is_none() {
+                continue;
+            }
```
mtreinish, line 193: In the python version this was independent of `cancel_set.len() > 1`, does this matter? I guess probably not because if we had 1 gate in the cancel set we'd skip regardless of what the key was.

crates/accelerate/src/commutation_cancellation.rs
```diff
@@ -0,0 +1,280 @@
...
+            if matches!(cancel_key.gate, GateOrRotation::ZRotation)
+                || matches!(cancel_key.gate, GateOrRotation::XRotation)
+            {
+                let run_qarg = match &dag.dag[*cancel_set.first().unwrap()] {
```
mtreinish:
```suggestion
                let run_qarg = match &dag.dag[*cancel_set[0]] {
```

crates/accelerate/src/commutation_cancellation.rs
```diff
@@ -0,0 +1,280 @@
...
+                    for &c_node in &cancel_set[0..(cancel_set.len() / 2) * 2] {
+                        dag.remove_op_node(c_node);
+                    }
+                }
+            }
+            if matches!(cancel_key.gate, GateOrRotation::ZRotation) && z_var_gate.is_none() {
+                continue;
+            }
+            if matches!(cancel_key.gate, GateOrRotation::ZRotation)
+                || matches!(cancel_key.gate, GateOrRotation::XRotation)
```
mtreinish: I think this syntax works to make it a little more concise: 
```suggestion
            if matches!(cancel_key.gate, GateOrRotation::ZRotation | GateOrRotation::XRotation)
```

crates/accelerate/src/commutation_cancellation.rs
```diff
@@ -0,0 +1,280 @@
...
+                        NodeType::Operation(instr) => instr,
+                        _ => panic!("Unexpected type in commutation set run."),
+                    };
+                    let node_op_name = node_op.op.name();
+
+                    let node_qargs = dag.get_qargs(node_op.qubits);
+                    if node_op
+                        .extra_attrs
+                        .as_deref()
+                        .is_some_and(|attr| attr.condition.is_some())
```
mtreinish:
```suggestion
                    if node_op.condition().is_some()
```

crates/accelerate/src/commutation_cancellation.rs
```diff
@@ -0,0 +1,280 @@
...
+                        || node_qargs.len() > 1
+                        || node_qargs[0] != run_qarg
+                    {
+                        return Err(QiskitError::new_err("internal error"));
```
mtreinish: For "backwards compatibility" (whatever that means in this case) we should return a `PyRuntimeError` here since that's what the python version returned. But realistically I find it super unlikely anyone was depending on this exact exception being raised. I think realistically we might be able to remove this check and error completely since none of these conditions should be possible.

crates/accelerate/src/commutation_cancellation.rs
```diff
@@ -0,0 +1,280 @@
...
+                            )))
+                        }
+                    } else if HALF_TURNS.contains(&node_op_name) {
+                        Ok(PI)
+                    } else if QUARTER_TURNS.contains(&node_op_name) {
+                        Ok(PI / 2.0)
+                    } else if EIGHTH_TURNS.contains(&node_op_name) {
+                        Ok(PI / 4.0)
+                    } else {
+                        Err(QiskitError::new_err(format!(
```
mtreinish: Similarly for backwards compatibility this would be a `PyRuntimeError` not a `QiskitError` but I also don't think this is something we could trigger. Ideally we maybe should look at using a more rust native structure that doesn't have a fall through possibility like this. But I don't have any concrete ideas so we can just leave this error path as is.

crates/accelerate/src/commutation_cancellation.rs
```diff
@@ -0,0 +1,280 @@
...
+    });
+
+    for (cancel_key, cancel_set) in &cancellation_sets {
+        if cancel_set.len() > 1 {
+            if let GateOrRotation::Gate(g) = cancel_key.gate {
+                if SUPPORTED_GATES.contains(&g) {
+                    for &c_node in &cancel_set[0..(cancel_set.len() / 2) * 2] {
+                        dag.remove_op_node(c_node);
+                    }
+                }
```
mtreinish, line 188: I think this needs a continue here because this is an xor with the other two gate fields. In the python this was if gate else if rotation.
```suggestion
                }
                continue;
```

crates/accelerate/src/commutation_cancellation.rs
```diff
@@ -0,0 +1,280 @@
...
+                    total_phase += new_phase
+                }
+
+                let new_op = if matches!(cancel_key.gate, GateOrRotation::ZRotation) {
+                    z_var_gate.unwrap()
+                } else if matches!(cancel_key.gate, GateOrRotation::XRotation) {
+                    &RXGate
+                } else {
+                    return Err(QiskitError::new_err("impossible case!"));
+                };
```
mtreinish: I think using a match here would be more natural:
```suggestion
                let new_op = match cancel_key.gate
                    GateOrRotation::ZRotation => z_var_gate.unwrap(),
                    GateOrRotation::XRotation) => &RXGate,
                    _ => unreachable!()
                };
```

I also switched this to an `unreachable!()` because while reviewing the current form of the PR it is obvious that this branch is never reachable because  after re-reading the code I realize the else arm here couldn't ever be reached since this all in a block that gate matches either rotation variant of the enum.

mtreinish (commented):
I think this almost ready now, the current state of this PR is looking much better. Thanks for making all the updates. I have another round of inline comments, but most are pretty small nits and mechanical things to make the code a bit more concise.

crates/accelerate/src/commutation_cancellation.rs
sbrandhsn: Yes, I agree

sbrandhsn (commented):
Thanks for the feedback! :-)

crates/circuit/src/dag_circuit.rs
```diff
@@ -6228,7 +6228,7 @@ impl DAGCircuit {
         &mut self,
         new_gate: (StandardGate, &[f64]),
         old_index: NodeIndex,
-    ) {
+    ) -> NodeIndex {
```
```suggestion
    ) {
```

crates/circuit/src/dag_circuit.rs
```diff
@@ -6255,6 +6255,7 @@ impl DAGCircuit {
         self.dag.add_edge(parent_index, new_index, weight.clone());
         self.dag.add_edge(new_index, old_index, weight);
         self.dag.remove_edge(edge_index);
+        new_index
```
```suggestion
```

crates/circuit/src/dag_circuit.rs
```diff
@@ -269,7 +269,7 @@ pub struct DAGCircuit {
     var_output_map: _VarIndexMap,
 
     /// Operation kind to count
-    op_names: IndexMap<String, usize, RandomState>,
+    pub op_names: IndexMap<String, usize, RandomState>,
```
mtreinish: I think instead of making this public we should be using `count_ops(py, false)`

```suggestion
    op_names: IndexMap<String, usize, RandomState>,
```

Although I can see the value here of having a no copy access method to `op_names` that returns a `&IndexMap` instead of an owned copy like `count_ops` does.

mtreinish (commented):
Sorry I glossed over the changes to `dag_circuit.rs` in my last review. We shouldn't change the interface to `insert_1q_on_incoming_qubit` unless we need to. Since nothing is using it anymore we should revert that piece. Also, I think we shouldn't make `op_names` public because that opens up the possibility of a transpiler pass corrupting the name counts accidentally by mutating it. We have the `count_ops()` method already, but we can also add a no-copy method to get an immutable reference to the inner mapping. These are so minor I'll push up a quick commit to fix both right now. Otherwise I think this is ready to merge.

mtreinish (approved):
LGTM now, thanks for all the work on this! It's going to have a very positive impact on runtime performance for the transpiler at optimization levels 2 and 3.

PR #13092, opened 2024-09-05
Fix `Interner::with_capacity` for default keys

jakelishman:
### Summary

The new method did not account for allocating the default key, causing code that then tried to use it to panic.

### Details and comments

No changelog because the bug isn't released.


ElePT (approved):
LGTM, thanks for fixing one of my sources of headache :) (I pulled the PR and confirmed it does fix it)

PR #13093, opened 2024-09-05
Clean two_qubit_decompose code

ShellyGarion:
### Summary
In this PR we remove some unused code from the file `qiskit/synthesis/two_qubit/two_qubit_decompose.py`.
And moreover, the following Python functions now calls the relevant Rust functions:
- `decompose_two_qubit_product_gate` 
- `trace_to_fid`
- `Ud`

### DIFF
--- a/crates/accelerate/src/two_qubit_decompose.rs
+++ b/crates/accelerate/src/two_qubit_decompose.rs
@@ -32,7 +32,7 @@ use ndarray::linalg::kron;
 use ndarray::prelude::*;
 use ndarray::Zip;
 use numpy::{IntoPyArray, ToPyArray};
-use numpy::{PyReadonlyArray1, PyReadonlyArray2};
+use numpy::{PyArrayLike2, PyReadonlyArray1, PyReadonlyArray2};
 
 use pyo3::exceptions::PyValueError;
 use pyo3::prelude::*;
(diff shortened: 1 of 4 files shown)


crates/accelerate/src/two_qubit_decompose.rs
```diff
@@ -182,9 +182,25 @@ fn decompose_two_qubit_product_gate(
     }
     l.mapv_inplace(|x| x / det_l.sqrt());
     let phase = det_l.arg() / 2.;
+
     Ok((l, r, phase))
 }
 
+#[pyfunction]
+#[pyo3(signature = (special_unitary))]
```
mtreinish: If you do this you can drop the `py_` prefix when calling this from Python. as it sets the name of the python side to what's specified in the macro. As for the signature field we don't really need to specify this because it being a required positional named argument is the default behavior for `pyfunction`s

```suggestion
#[pyo3(name = "decompose_two_qubit_product_gate")]
```

crates/accelerate/src/two_qubit_decompose.rs
```diff
@@ -182,9 +182,25 @@ fn decompose_two_qubit_product_gate(
...
+fn py_decompose_two_qubit_product_gate(
+    py: Python,
+    special_unitary: PyArrayLike2<Complex64>,
+) -> PyResult<(PyObject, PyObject, f64)> {
+    let view = special_unitary.as_array();
+    let (l, r, phase) = decompose_two_qubit_product_gate(view)?;
+    Ok((
+        PyObject::from(l.into_pyarray_bound(py).unbind()),
+        PyObject::from(r.into_pyarray_bound(py).unbind()),
```
mtreinish: This is fine, but you can also probably just do:

```suggestion
        l.into_pyarray_bound(py).unbind().into(),
        r.into_pyarray_bound(py).unbind().into(),
```

mtreinish (commented):
This LGTM, thanks for doing this. I have two small nits inline but otherwise I think this is good to go.

crates/accelerate/src/two_qubit_decompose.rs
```diff
@@ -301,6 +332,43 @@ fn rz_matrix(theta: f64) -> Array2<Complex64> {
...
+            (IM * c).exp() * (a - b).cos()
+        ]
+    ]
+}
+
+#[pyfunction]
+#[pyo3(name = "Ud")]
+fn py_ud(py: Python, a: f64, b: f64, c: f64) -> PyResult<Py<PyArray2<Complex64>>> {
+    let ud_mat = ud(a, b, c);
+    Ok(ud_mat.into_pyarray_bound(py).unbind())
```
mtreinish: Does this need a `PyResult`? I don't think anything here is fallible so we can simplify the code if we removed it:

```suggestion
fn py_ud(py: Python, a: f64, b: f64, c: f64) -> Py<PyArray2<Complex64>> {
    let ud_mat = ud(a, b, c);
    ud_mat.into_pyarray_bound(py).unbind()
```

mtreinish (commented):
This LGTM, just one inline question/suggestion otherwise I think this is good to go.

mtreinish (approved):
LGTM, thanks for updating.

PR #13094, opened 2024-09-05
Port `ElidePermutations` transpiler pass to Rust

### LINKED ISSUE #12336: Port `ElidePermutations` to Rust

alexanderivrii:
### Summary

Closes #12336.

This PR reimplements most of the logic of the `ElidePermutations` transpiler pass in Rust (the only thing missing is updating the passmanager's property set to account for the new mapping, this is still done in the Python space).

The Rust code also corrects a problem in the Python code where the qubit mapping was not updated correctly in the presence of permutation gates over a subset of qubits -- it's really scary how broken it was and that none of our tests were catching this (and I have added an extra python test that would exhibit the bug).

On one simple test with 100,000 CX-gates and 99,800 SWAP-gates (multiple iterating layers of CX-gates and SWAP-gates), the runtime is improved from `1.388` seconds to `0.094` seconds -- about 10x speedup; and on another simple quantum-volume-like test (multiple iterating layers of random unitary and permutation gates) the runtime is improved from `0.128` seconds to `0.020` seconds -- about 5x speedup.

### Details and comments

I have needed to expose some of the functionality from `DAGCircuit`, ~First, I needed the pure Rust version of the function `count_ops()` which I have copied from some other PR, most probably from #13013 (thus it's probably best to wait till this merges and rebase the code on top of it).~ Second, I needed to make some of the functions public, namely: ~`push_back`, `topological_op_nodes` and~ `copy_empty_like`. ~Third, I needed the setter variant `set_qargs` (and we already had `get_qargs`), and I also added `set_cargs` for consistency.~

Thanks to @jakelishman and @Cryoris  for help and suggestions!


crates/accelerate/src/elide_permutations.rs
```diff
@@ -0,0 +1,104 @@
...
+                    let index1 = qargs[1].0 as usize;
+                    let prev0 = mapping[index0];
+                    let prev1 = mapping[index1];
+                    mapping[index0] = prev1;
+                    mapping[index1] = prev0;
+                }
+                ("permutation", None) => {
+                    if let Param::Obj(ref pyobj) = inst.params.as_ref().unwrap()[0] {
+                        let pyarray: PyReadonlyArray1<i32> = pyobj.extract(py)?;
+                        let pattern = pyarray.as_array();
```
alexanderivrii, line 53: These lines together with the definition `pattern = np.array(pattern)` of a permutation gate pattern worked fine for me locally, with Python 10 and Windows, but were failing the PR's MacOS CI testing. Changing the above line to `pattern = np.array(pattern, dtype=np.int32)` seems to have fixed this, but I have some concerns whether the change is fully backward-compatible.

crates/circuit/src/dag_circuit.rs
```diff
@@ -6223,6 +6223,16 @@ impl DAGCircuit {
...
 
+    /// Insert qargs, return the intern index
+    pub fn set_qargs(&mut self, qargs: &[Qubit]) -> Interned<[Qubit]> {
+        self.qargs_interner.insert(qargs)
+    }
+
+    /// Insert cargs, return the intern index
+    pub fn set_cargs(&mut self, cargs: &[Clbit]) -> Interned<[Clbit]> {
+        self.cargs_interner.insert(cargs)
+    }
```
kevinhartman: IMO, we should favor adding a variant of `push_back` that deals in terms of native bits (in addition to the one that takes a `PackedInstruction`) rather than exposing methods like these.

(I can't think of much use for mutable interners other than the case of emplacing a new instruction into the DAG, and other passes will likely need this functionality too anyway 🙂.)

alexanderivrii: To make sure I fully understand the suggestion: you are proposing to add a function like
```
fn push_back_op(
    op: &PackedOperation, 
    qargs: &[Qubit], 
    cargs: &[Clbit], 
    params: &Option<Box<SmallVec<[Param; 3]>>>, 
    extra_attrs: &Option<Box<ExtraInstructionAttributes>>, 
    py_op: OnceCell<Py<PyAny>>
)
```
Internally the function would do the interning, create a packed instruction, and call the existing `push_back`. Is this on the right track? (Btw, I did not yet try to implement it, but the "optional" `py_op` already scares me).

Though in my particular use-case, I would like to avoid de-interning and re-interning things related to classical bits, since the clbits iterner is copied in `copy_empty_like` and the classical bits are not remapped by the pass. What do you think?

kevinhartman: I think the signature for this should be something like:

```rust
fn apply_operation_back(
    op: PackedOperation, 
    qargs: &[Qubit], 
    cargs: &[Clbit], 
    params: Option<&[Param]>,
    extra_attrs: Option<&ExtraInstructionAttributes>,
) -> NodeIndex
```

Note that I used the name `apply_operation_back` (I called the Python version `py_apply_operation_back` with anticipation of this Rust-oriented version).

I think the `py_op` cache will get updated on demand within the instruction, so perhaps we can get away with keeping it out of the interface here. Maybe @jakelishman has thoughts?

kevinhartman: Issue #13134 has been create for this work. I will quote the above discussion there

jakelishman: I'm fine with the method as described by Kevin, but I also think there ought to be ways to push operations back while passing an already-known `Interned` key.  As in Sasha's case, there's plenty of cases where we need to pass stuff in where we can know that the interned key is safe, and looking up and then re-hashing the `&[Clbit]` or `&[Qubit]` would be a waste of cycles.

_Personally_ I have no problem with exposing a method to allow manual insertions into the interners as well - I don't think it should be the standard way to do that, but you can't break data coherence by doing it, since interners are append-only, and adding the key doesn't invalidate existing keys.  That said, I don't immediately have uses for that in mind - I have a rough imagined optimisation to the Sabre post-routing rebuilder that pre-allocates the interner, but no proof it'd be a good idea - so I'm not pushing for adding them.

crates/accelerate/src/elide_permutations.rs
```diff
@@ -0,0 +1,104 @@
...
+        if let NodeType::Operation(inst) = &dag.dag[node_index] {
+            match (inst.op.name(), inst.condition()) {
+                ("swap", None) => {
+                    let qargs = dag.get_qargs(inst.qubits);
+                    let index0 = qargs[0].0 as usize;
+                    let index1 = qargs[1].0 as usize;
+                    let prev0 = mapping[index0];
+                    let prev1 = mapping[index1];
+                    mapping[index0] = prev1;
+                    mapping[index1] = prev0;
```
mtreinish: You can use https://doc.rust-lang.org/std/vec/struct.Vec.html#method.swap here:

```suggestion
                    mapping.swap(index0, index1);
```

alexanderivrii: I knew this existed! Thanks!

crates/circuit/src/dag_circuit.rs
kevinhartman: Just a heads up @alexanderivrii , #13143 is now merged which adds a native `apply_operation_back`. 

> Though in my particular use-case, I would like to avoid de-interning and re-interning things related to classical bits, since the clbits iterner is copied in copy_empty_like and the classical bits are not remapped by the pass. What do you think?

Given that you have this requirement, I'd say feel free to use `DAGCircuit::push_back` directly instead, but please also use the interner directly (e.g. `dag.qargs_interner().insert(mapped_qargs)`) rather than adding these `set_{q|c}args` methods to `DAGCircuit`.

alexanderivrii: _“If you wait by the river long enough, the bodies of your enemies will float by.”_
― Sun Tzu

@kevinhartman, @raynelfss, and @jakelishman -- many thanks for the insightful discussion and adding this feature.

alexanderivrii: @kevinhartman, actually I am still confused about the best fix: calling `dag.qargs_interner().insert(&mapped_qargs)` does not seem to work due to mutability concerns (and I am not sure we should provide the method `qargs_interner_mut`).

In fact, I like your previous suggestion to use `apply_operation_back` instead of `push_back`. Even if this currently requires a bit of unnecessary work to retrieve `cargs` from the dag's cargs interner and to load these `cargs` into the elided dag's cargs interner, this does not assume that the cargs interner is copied in `copy_empty_like` (possibly this decision may change in the future). However, I have two follow-up questions. First, how should we pass `params` to `apply_operation_back`: as simply cloning it leads to the error 
```
expected `Option<SmallVec<[Param; 3]>>`, found `Option<Box<SmallVec<[Param; 3]>>>`
```
Second, what should be `py_op` when calling `apply_operation_back`?

kevinhartman: > does not seem to work due to mutability concerns

Ah, you're right. I forgot that we don't expose a mutable reference for those. We _could_, but if you're happy to just use `apply_operation_back` let's leave that up to a future decision.

> First, how should we pass params to apply_operation_back: as simply cloning it leads to the error

I believe what you're looking for is something like: `params.as_deref().cloned()`. This takes you from an `Option<Box<SmallVec<[Param; 3]>>>` to `Option<&SmallVec<[Param; 3]>>` and finally to an owned `Option<SmallVec<[Param; 3]>>`.

> Second, what should be py_op when calling apply_operation_back?

For `py_op`, you can clone whatever is in the existing `PackedInstruction` (though you will have to do it conditionally with the macro, just like how the parameter itself is declared in `DAGCircuit::apply_operation_back`).

In general, the `py_op` thing is just a cached version of the Python-based operation, so if you have a `PyObject` corresponding to the `PackedOperation` you're pushing, you can pass it. Else, just pass `None`.

crates/accelerate/src/elide_permutations.rs
alexanderivrii: Done!

crates/circuit/src/dag_circuit.rs
alexanderivrii: Thanks, this is done. Now the only change to the dag circuit interface in this PR is making `copy_empty_like` public.

crates/accelerate/src/elide_permutations.rs
```diff
@@ -78,17 +75,26 @@ fn run(py: Python, dag: &mut DAGCircuit) -> PyResult<Option<(DAGCircuit, Vec<usi
...
                     let mapped_qargs: Vec<Qubit> = qargs
                         .iter()
                         .map(|q| q.0 as usize)
                         .map(|q| mapping[q])
                         .map(|q| Qubit(q.try_into().unwrap()))
                         .collect();
-                    let mapped_qubits = new_dag.set_qargs(&mapped_qargs);
-                    mapped_inst.qubits = mapped_qubits;
-                    new_dag.push_back(py, mapped_inst)?;
+                    let params = inst.params.clone();
```
kevinhartman: Is this still needed? Or can we directly do `inst.params.as_deref().cloned()` in the stuct initializer?

releasenotes/notes/port-elide-permutations-ed91c3d9cef2fec6.yaml
```diff
@@ -3,7 +3,3 @@ features_transpiler:
   - |
     Port most of the logic of the transpiler pass :class:`~.ElidePermutations`
     to Rust.
-fixes:
```
kevinhartman: Is there a reason to drop the message you'd added here?

kevinhartman (commented):
Just a few small comments.

crates/accelerate/src/elide_permutations.rs
alexanderivrii: Good catch, this is a remnant of my own failed attempts to clone `params` correctly. Fixed in [8209757](https://github.com/Qiskit/qiskit/pull/13094/commits/820975721a5894315640edc105114d3a58ac9d60).

releasenotes/notes/port-elide-permutations-ed91c3d9cef2fec6.yaml
alexanderivrii: Yes, we have decided to put the bug fix into a separate python-only PR so that it could be backported (please see #13186).

crates/accelerate/src/elide_permutations.rs
kevinhartman: Sounds good. We should probably (in a separate PR) expose a method on `PackedInstruction` as well which just returns `self.params.as_deref()` for convenience.

releasenotes/notes/port-elide-permutations-ed91c3d9cef2fec6.yaml
```diff
@@ -0,0 +1,5 @@
+---
+features_transpiler:
+  - |
```
kevinhartman, line 3: I think we haven't actually written release notes like this for any (most of?) the other passes being ported, have we? IMO, we can leave this in for now and the release manager (whoever that ends up being) can just remove this if it's out of place.

crates/accelerate/src/elide_permutations.rs
mtreinish: I think that exists: https://github.com/Qiskit/qiskit/blob/main/crates/circuit/src/packed_instruction.rs#L530

kevinhartman (approved):
This looks good to me. Thanks for the changes (and for trying out the new Rust-facing `DAGCircuit::apply_operation_back`!).

crates/accelerate/src/elide_permutations.rs
kevinhartman: Hmm. Perhaps what I'm suggesting would be more confusing, then. The `params_view` method returns a `&[Param]` slice, but what we'd need here is the owned representation of parameters stored by `DAGCircuit` (`Option<SmallVec<[Param; 3]>>`).

That makes me think we may want to consider changing `DAGCircuit::apply_operation_{back,front}` to take `&[Param]` instead of owned data. It'd be less efficient for callers that happen to already have owned parameters to move in during the apply, but it's certainly a cleaner interface and hides what is otherwise an implementation detail for how DAGs store parameters.

Issue #12970, opened 2024-08-16
Split2QUnitaries assumes node.op is ndarray


qci-amos:
The `TwoQubitWeylDecomposition` says it only accepts ndarray:
https://github.com/Qiskit/qiskit/blob/713ab38cccb7e55a40a87372a2590a5a4430fa8f/qiskit/synthesis/two_qubit/two_qubit_decompose.py#L184-L190

However this line is passing a node op:
https://github.com/Qiskit/qiskit/blob/713ab38cccb7e55a40a87372a2590a5a4430fa8f/qiskit/transpiler/passes/optimization/split_2q_unitaries.py#L53
some of my code is now failing in qiskit 1.2.0 because in some of my unit tests, this is an `Instruction` which is, for what it's worth, not a gate in the qiskit library.

My problem goes away if I replace this:
```python
    circuit = qiskit.transpile(circuit, basis_gates=basis_gates, optimization_level=2)
```
in my test with:
```python
    circuit = qiskit.transpile(circuit, basis_gates=basis_gates, optimization_level=1)
```

Linking associated PRs:
https://github.com/Qiskit/qiskit/pull/12727
https://github.com/Qiskit/qiskit/pull/12898

mtreinish:
Do you have example code and a traceback showing the error you're hitting? The object model/type should be protecting against this if your custom class is an `Instruction` subclass and not a `Gate` subclass I would expect the condition attempting to skip attempting the decomposition to always be triggered because the `node.matrix` will always be `None` for an `Instruction` subclass that's not a `Gate` subclass.

Looking at the code you're right there is a mismatch with the argument typing and the pass is not accounting for the case it should be passing `node.matrix` it's implicitly relying on gates with matrices also implementing `__array__` which is not a requirement of `Gate` subclasses. This will be fixed in 1.2.1, and should just call `node.matrix` instead of `node.op` as the input to the `TwoQubitWeylDecomposition`, but it would be good to confirm that this is actually what's going on in your local code because if you're running with a subclass of `Instruction` directly I think we need to do some more digging.

jakelishman:
Thanks for the report. Do you have a small producer script that shows the problem?

I think I know what's happening, but it'd be good to be certain. The type hint of the decomposer isn't really correct, and it's actually anything that can be cast to an array, which is true of most gates - but there's no requirement that a user gate has a matrix form. (Fwiw, if you want to, give your gate a `__array__` method.)

qci-amos:
[...]

PR #13095, opened 2024-09-05
Restrict `Split2QUnitaries` to run on `UnitaryGate`s only

Cryoris:
### Summary

Fixes the remaining problems with #12970.

### Details and comments

If we run `Split2QUnitaries` with an incomplete 1-q basis gate set, we can end up in the situation where a supported gate is split into 2 `UnitaryGate` objects that might not be representable in the current basis gate set. For example
```python
from qiskit.circuit.library import *

gate = PauliGate("XX")

qc = QuantumCircuit(2)
qc.append(gate, [0, 1])

basis_gates = ["x"]  # gate decomposes into XX
 
circuit = transpile(qc, basis_gates=basis_gates, optimization_level=2)  # fails -- cannot map U to the basis
```

This PR fixes the behavior by only running on `UnitaryGate` objects -- hence not changing the gate types existing in the circuit.

### DIFF
--- a/qiskit/transpiler/passes/optimization/split_2q_unitaries.py
+++ b/qiskit/transpiler/passes/optimization/split_2q_unitaries.py
@@ -9,8 +9,8 @@
 # Any modifications or derivative works of this code must retain this
 # copyright notice, and modified files need to carry a notice indicating
 # that they have been altered from the originals.
+
 """Splits each two-qubit gate in the `dag` into two single-qubit gates, if possible without error."""
-from typing import Optional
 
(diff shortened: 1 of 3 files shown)


mtreinish (approved):
LGTM, I'll make sure to update #13025 to reflect this change too.

This could arguably be considered a behavior change but I think we're in the grey area because the pass was always named `Split2QUnitaries` (I totally missed the capital `Q` which is lowercase for everything else) and the intent was always for this to be run after consolidate blocks. Also since this is actively breaking use cases that worked  in < 1.2 I would treat this as the lesser change.

PR #13098, opened 2024-09-05
Fix documentation for utils.optionals data values

Eric-Arellano:
We're adding proper support to `.. py:data::` to docs.quantum.ibm.com. However, we cannot easily handle `.. py:data::` being inside a table - it would complicate both the code and visual design.

So, this PR removes the table and uses `.. py:data::` like we use it everywhere else in Qiskit.

### DIFF
--- a/qiskit/utils/optionals.py
+++ b/qiskit/utils/optionals.py
@@ -25,172 +25,195 @@ Available Testers
 Qiskit Components
 -----------------
 
-.. list-table::
-    :widths: 25 75
+.. py:data:: HAS_AER
 
-    * - .. py:data:: HAS_AER
-      - `Qiskit Aer <https://qiskit.github.io/qiskit-aer/>` provides high-performance simulators for
-        the quantum circuits constructed within Qiskit.
+    `Qiskit Aer <https://qiskit.github.io/qiskit-aer/>` provides high-performance simulators for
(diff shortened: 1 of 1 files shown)


Eric-Arellano:
> [Eric-Arellano](https://github.com/Eric-Arellano) modified the milestones: [0.46.3](https://github.com/Qiskit/qiskit/milestone/62), [1.2.1](https://github.com/Qiskit/qiskit/milestone/65)

I made a change so that the docs conversion script can now tolerate `.. py:data::` inside tables. But it will preserve the current behavior. Rather than using the new colored bar like the first image, it will still be the table in the second image. Anchor links also will not work to these values - you'll go to the correct page (no 404), but won't go to the correct section.

image

Screenshot 2024-09-05 at 3 48 08 PM

Feel free to cherry-pick this to 0.46 and 1.2, but it's also okay if we only fix 1.3+.

jakelishman:
tbh the table form wasn't really proper Sphinx anyway from a semantic perspective. I think at the time these were added, that was the best way I could come up with something legible in the existing (pre Furo) theme. The new form in this PR is "proper" Sphinx, so good to stick with that.

jakelishman (approved):

jakelishman:
@Mergifyio backport stable/0.46 stable/1.2

Issue #12081, opened 2024-03-26
More efficient synthesis of StatePreparation and Initialize based on Isometry


ShellyGarion:
### What should we add?

The current code for [`StatePreparation`](https://docs.quantum.ibm.com/api/qiskit/qiskit.circuit.library.StatePreparation) and [`Initialize`](https://docs.quantum.ibm.com/api/qiskit/qiskit.circuit.library.Initialize) is based on [1], but there is a more efficient code in [2] which is currently implemented in [`Isometry`](https://docs.quantum.ibm.com/api/qiskit/qiskit.circuit.library.Isometry).

Therefore, it would be better to replace the internal implementation of [`StatePreparation`](https://docs.quantum.ibm.com/api/qiskit/qiskit.circuit.library.StatePreparation) and [`Initialize`](https://docs.quantum.ibm.com/api/qiskit/qiskit.circuit.library.Initialize) by the one implemented in [`Isometry`](https://docs.quantum.ibm.com/api/qiskit/qiskit.circuit.library.Isometry).

**References**
[1] Shende, Bullock, Markov. Synthesis of Quantum Logic Circuits (2004) [https://arxiv.org/abs/quant-ph/0406176v5](https://arxiv.org/abs/quant-ph/0406176v5)
[2] Iten et al., Quantum circuits for isometries (2016). [Phys. Rev. A 93, 032318](https://journals.aps.org/pra/abstract/10.1103/PhysRevA.93.032318)

For example,
```
for n_qubits in [5, 7, 10, 12, 15]:

    state = np.random.rand(2 ** n_qubits) + np.random.rand(2 ** n_qubits) * 1j
    state = state/np.linalg.norm(state)

    # State preparation
    qc = QuantumCircuit(n_qubits)
    qc.append(StatePreparation(state), range(n_qubits))

    transpiled = transpile(qc, basis_gates=['u', 'cx'], optimization_level=0)
    cx_sp = transpiled.count_ops()['cx']

    # Isometry
    reg_iso = QuantumRegister(n_qubits)
    circuit_iso = QuantumCircuit(reg_iso)
    isom = Isometry(state, 0, 0)

    qc = QuantumCircuit(n_qubits)
    qc.append(isom, range(n_qubits))

    transpiled = transpile(qc, basis_gates=['u', 'cx'], optimization_level=0)
    cx_isom = transpiled.count_ops()['cx']

[...]

PR #13099, opened 2024-09-05
Quantum Multiplexer Simplification

josejad42:
### Summary

Multiplexer optimization by reducing unnecessary controls.

### Details and comments

This method introduces a simplification in multiplexers that removes unnecessary controls and operators when a pattern of repetitions is detected. The main advantage is a significant reduction in both the number of CX gates and the overall circuit depth during separable state preparation.

### DIFF
--- a/qiskit/circuit/library/generalized_gates/uc.py
+++ b/qiskit/circuit/library/generalized_gates/uc.py
@@ -113,7 +113,9 @@ class UCGate(Gate):
         self.up_to_diagonal = up_to_diagonal
 
     def _simplify(self, gate_list, num_contr):
-        c = set()
+    """https://arxiv.org/abs/2409.05618"""
+
+        c = set()```
         nc = set()
         mux_copy = gate_list.copy()
 
--- a/qiskit/circuit/library/generalized_gates/uc.py
+++ b/qiskit/circuit/library/generalized_gates/uc.py
@@ -68,7 +68,7 @@ class UCGate(Gate):
     """
 
     def __init__(
-        self, gate_list: list[np.ndarray], up_to_diagonal: bool = False, mux_simp: bool = False
+        self, gate_list: list[np.ndarray], up_to_diagonal: bool = False, mux_simp: bool = True
     ):
         r"""
         Args:
@@ -192,6 +192,19 @@ class UCGate(Gate):
         # q[k-1],...,q[0],q_target, decreasingly ordered with respect to the
         # significance of the qubit in the computational basis
         _, diag = self._dec_ucg()
+        if self.simp_contr[1]:
+            q_controls = [self.num_qubits - i for i in self.simp_contr[1]]
+            q_controls.reverse()
+            for i in range(self.num_qubits):
+                if i not in [0] + q_controls:
+                    d = 2**i
+                    new_diag = []
+                    n = len(diag)
+                    for j in range(n):
+                        new_diag.append(diag[j])
+                        if (j + 1) % d == 0:
+                            new_diag.extend(diag[j + 1 - d : j + 1])
+                    diag = np.array(new_diag)
         return diag
 
     def _define(self):
--- a/qiskit/circuit/library/generalized_gates/uc.py
+++ b/qiskit/circuit/library/generalized_gates/uc.py
@@ -113,9 +113,9 @@ class UCGate(Gate):
         self.up_to_diagonal = up_to_diagonal
 
     def _simplify(self, gate_list, num_contr):
-    """https://arxiv.org/abs/2409.05618"""
+        """https://arxiv.org/abs/2409.05618"""
 
-        c = set()```
+        c = set()
         nc = set()
         mux_copy = gate_list.copy()
 
--- a/qiskit/circuit/library/generalized_gates/uc.py
+++ b/qiskit/circuit/library/generalized_gates/uc.py
@@ -67,7 +67,9 @@ class UCGate(Gate):
 
     """
 
-    def __init__(self, gate_list: list[np.ndarray], up_to_diagonal: bool = False):
+    def __init__(
+        self, gate_list: list[np.ndarray], up_to_diagonal: bool = False, mux_simp: bool = False
+    ):
         r"""
         Args:
             gate_list: List of two qubit unitaries :math:`[U_0, ..., U_{2^{k-1}}]`, where each
@@ -101,9 +103,64 @@ class UCGate(Gate):
             if not is_unitary_matrix(gate, _EPS):
                 raise QiskitError("A controlled gate is not unitary.")
 
+        new_controls = set()
+        if mux_simp:
+            new_controls, gate_list = self._simplify(gate_list, num_contr)
+        self.simp_contr = (mux_simp, new_controls)
+
         # Create new gate.
         super().__init__("multiplexer", int(num_contr) + 1, gate_list)
         self.up_to_diagonal = up_to_diagonal
+
+    def _simplify(self, gate_list, num_contr):
+        c = set()
+        nc = set()
+        mux_copy = gate_list.copy()
+
+        for i in range(int(num_contr)):
+            c.add(i + 1)
+
+        if len(gate_list) > 1:
+            nc, mux_copy = self._repetition_search(gate_list, num_contr, mux_copy)
+
+        new_controls = {x for x in c if x not in nc}
+        new_mux = [gate for gate in mux_copy if gate is not None]
+        return new_controls, new_mux
+
+    def _repetition_search(self, mux, level, mux_copy):
+        nc = set()
+        d = 1
+        while d <= len(mux) / 2:
+            disentanglement = False
+            if np.allclose(mux[d], mux[0]):
+                mux_org = mux_copy.copy()
+                repetitions = len(mux) / (2 * d)
+                p = 0
+                while repetitions > 0:
+                    repetitions -= 1
+                    valid, mux_copy = self._repetition_verify(p, d, mux, mux_copy)
+                    p = p + 2 * d
+                    if not valid:
+                        mux_copy = mux_org
+                        break
+                    if repetitions == 0:
+                        disentanglement = True
+
+            if disentanglement:
+                removed_contr = level - math.log2(d)
+                nc.add(removed_contr)
+            d = 2 * d
+        return nc, mux_copy
+
+    def _repetition_verify(self, base, d, mux, mux_copy):
+        i = 0
+        next_base = base + d
+        while i < d:
+            if not np.allclose(mux[base], mux[next_base]):
+                return False, mux_copy
+            mux_copy[next_base] = None
+            base, next_base, i = base + 1, next_base + 1, i + 1
+        return True, mux_copy
 
     def inverse(self, annotated: bool = False) -> Gate:
         """Return the inverse.
@@ -149,12 +206,19 @@ class UCGate(Gate):
         """
         diag = np.ones(2**self.num_qubits).tolist()
         q = QuantumRegister(self.num_qubits, "q")
-        q_controls = q[1:]
         q_target = q[0]
+        mux_simplify = self.simp_contr[0]
+
+        if mux_simplify:
+            q_controls = [q[self.num_qubits - i] for i in self.simp_contr[1]]
+            q_controls.reverse()
+        else:
+            q_controls = q[1:]
+
         circuit = QuantumCircuit(q, name="uc")
         # If there is no control, we use the ZYZ decomposition
         if not q_controls:
-            circuit.unitary(self.params[0], [q])
+            circuit.unitary(self.params[0], q[0])
             return circuit, diag
         # If there is at least one control, first,
         # we find the single qubit gates of the decomposition.
@@ -190,7 +254,10 @@ class UCGate(Gate):
             # q[k-1],...,q[0],q_target (ordered with decreasing significance),
             # where q[i] are the control qubits and t denotes the target qubit.
             diagonal = Diagonal(diag)
-            circuit.append(diagonal, q)
+            if mux_simplify:
+                circuit.append(diagonal, [q_target] + q_controls)
+            else:
+                circuit.append(diagonal, q)
         return circuit, diag
 
     def _dec_ucg_help(self):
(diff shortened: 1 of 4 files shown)


Cryoris:
Hi @josejad42, thanks for opening a PR! Could you give some more context to what you're suggesting, e.g. is there a reference that describes this simplification, and could you benchmark your improvement?

adjs:
Hi @Cryoris, the reference will appear on arxiv in the next few days.

test/python/circuit/test_initializer.py
```diff
@@ -39,6 +39,36 @@ class TestInitialize(QiskitTestCase):
...
+        qc2.initialize(statevector)
+
+        qc1 = transpile(qc1, basis_gates=["u", "cx"])
+        qc2 = transpile(qc2, basis_gates=["u", "cx"])
+
+        statevector1 = Statevector(qc1)
+        statevector2 = Statevector(qc2)
+
+        print(qc1.count_ops())
+        print(qc2.count_ops())
```
ShellyGarion: I think that these prints could be removed. In addition, is there some upper bound on the number of ops that can be added to the tests?

test/python/circuit/test_initializer.py
```diff
@@ -39,6 +39,36 @@ class TestInitialize(QiskitTestCase):
 
     _desired_fidelity = 0.99
 
+    def test_disentangled(self):
```
ShellyGarion, line 42: Please note that the current Qiskit [`Initialize`](https://docs.quantum.ibm.com/api/qiskit/qiskit.circuit.library.Initialize) code is based on [`Isometry`](https://docs.quantum.ibm.com/api/qiskit/qiskit.circuit.library.Isometry) by Iten et al. [Phys. Rev. A 93, 032318](https://journals.aps.org/pra/abstract/10.1103/PhysRevA.93.032318).
This change has been done recently in https://github.com/Qiskit/qiskit/pull/12178 for Qiskit 1.2 release.

adjs:
Hi @Cryoris and @ShellyGarion
The multiplexer simplification is described in https://arxiv.org/pdf/2409.05618

test/python/circuit/test_initializer.py
```diff
@@ -39,6 +39,37 @@ class TestInitialize(QiskitTestCase):
...
+        qc2 = transpile(qc2, basis_gates=["u", "cx"])
+
+        statevector1 = Statevector(qc1)
+        statevector2 = Statevector(qc2)
+
+        counts1 = qc1.count_ops()["cx"]
+        counts2 = qc2.count_ops()["cx"]
+
+        self.assertTrue(counts2 <= counts1)
+        self.assertTrue(np.allclose(statevector1, statevector2))
```
adjs: The UCGate simplification impacts state preparation.
If a real-valued state is separable, the number of cx gates of Initialize will be reduced. 

However, there is no guarantees that the number of cnots will be optimal.
The initialization of [0, 1, 0, 0] produces a circuit with one cx (that can be removed because the control is always in zero).

In complex-valued states, we can have some optimization. But there is no guarantee that the simplification will work.

test/python/circuit/test_initializer.py
adjs: Thanks, we removed the prints and included a test to verify if the Initialize reduced the number of cnots of a disentangled state.

test/python/circuit/test_initializer.py
adjs: Thanks, we fixed the optimization and now the test works.

ShellyGarion:
> Hi @Cryoris and @ShellyGarion The multiplexer simplification is described in https://arxiv.org/pdf/2409.05618

Could you please add some benchmarks comparing your state preparation code to the existing Qiskit code?
See e.g. the benchmarks in https://github.com/Qiskit/qiskit/issues/12081

adjs:
> > Hi @Cryoris and @ShellyGarion The multiplexer simplification is described in https://arxiv.org/pdf/2409.05618
> 
> Could you please add some benchmarks comparing your state preparation code to the existing Qiskit code? See e.g. the benchmarks in #12081

Yes, sure! The best case is a separable real-valued state.

## qiskit 1.2.0
nqubits:  3 depth:  10 time:  0.03743324100003065
nqubits:  6 depth:  116 time:  0.08239613299997472
nqubits:  9 depth:  1006 time:  0.8098489340000015
nqubits:  12 depth:  8168 time:  27.270649252000112

## this branch
nqubits:  3 depth:  2 time:  0.053231799975037575
nqubits:  6 depth:  4 time:  0.05455289990641177
nqubits:  9 depth:  10 time:  0.10728330002166331
nqubits:  12 depth:  24 time:  0.424586100038141


```python
import numpy as np

from qiskit import transpile, QuantumCircuit
from time import perf_counter

for k in range(1, 5):
    size = 2 ** k
    state1 = np.random.rand(size)
    state1 = state1 / np.linalg.norm(state1)
    state2 = np.random.rand(size)
    state2 = state2 / np.linalg.norm(state2)
    state3 = np.random.rand(size)
    state3 = state3 / np.linalg.norm(state3)

    state = np.kron(state1, state2)
    state = np.kron(state, state3)

    nqubits = int(np.log2(len(state)))

    start = perf_counter()
    qc2 = QuantumCircuit(nqubits)
    qc2.initialize(state)
    qc2 = transpile(qc2, basis_gates=['u', 'cx'], optimization_level=2)
    end = perf_counter()
    print('nqubits: ', nqubits, 'depth: ',  qc2.depth(), 'time: ', end - start)


```

qiskit/circuit/library/generalized_gates/uc.py
```diff
@@ -101,10 +103,65 @@ def __init__(self, gate_list: list[np.ndarray], up_to_diagonal: bool = False):
...
+        if mux_simp:
+            new_controls, gate_list = self._simplify(gate_list, num_contr)
+        self.simp_contr = (mux_simp, new_controls)
+
         # Create new gate.
         super().__init__("multiplexer", int(num_contr) + 1, gate_list)
         self.up_to_diagonal = up_to_diagonal
 
+    def _simplify(self, gate_list, num_contr):
+        c = set()
```
adjs:
```suggestion
    def _simplify(self, gate_list, num_contr):
    """https://arxiv.org/abs/2409.05618"""

        c = set()```

File: qiskit/circuit/library/generalized_gates/uc.py [hunk 5]
```diff
@@ -67,7 +67,9 @@ class UCGate(Gate):
 
     """
 
-    def __init__(self, gate_list: list[np.ndarray], up_to_diagonal: bool = False):
+    def __init__(
+        self, gate_list: list[np.ndarray], up_to_diagonal: bool = False, mux_simp: bool = True
+    ):
```
ShellyGarion, line 77: - why do you need the parameter `mux_simp` ? 
if it's needed - please also add it to the documentation (line 81).

- if this parameter is needed, why don't you need to call it in the functions that call `UCGate` directly or indirectly (`mcg_up_to_diagonal`, `Isometry`, `StatePreparation` and `Initialize`) ? Since it's default value is `True` it will always affect these functions.

test/python/circuit/test_uc.py
```diff
@@ -100,6 +107,21 @@ def test_inverse_ucg(self):
 
         self.assertTrue(np.allclose(unitary_desired, unitary))
 
+    def test_ucge(self):
```
ShellyGarion, line 112: why do you need a specific test? Isn't it enough to add this specific `gate_list` of hadamards to `squs` above?

test/python/circuit/test_initializer.py
ShellyGarion: in the test above both `counts1` and `counts2`  are equal (=12). Do you have an example that there is an inequality? (perhaps you would like to fix a seed).

ShellyGarion (commented):
@josejad42, @adjs - Thanks for your contribution to Qiskit!

My main question is whether you always reduce the number of CX gates in the `StatePreparation` or `Initialize` process ?
I tried the same complex random states that appears in #12081 and I see that the number of CX gates in unchanged. In the example you provided above, the number of CX gates is significantly reduced.
Are there states for which the number of CX gates may increase ?

You may also add your reference and an explanation to the API docs (line 64)

qiskit/circuit/library/generalized_gates/uc.py
josejad42: We think that some users may not want to search for simplifications or may need to use the original version of the multiplexer. Therefore, simplification is enabled by default but can be easily disabled. Documentation will be updated.

josejad42:
> @josejad42, @adjs - Thanks for your contribution to Qiskit!
> 
> My main question is whether you always reduce the number of CX gates in the `StatePreparation` or `Initialize` process ? I tried the same complex random states that appears in #12081 and I see that the number of CX gates in unchanged. In the example you provided above, the number of CX gates is significantly reduced. Are there states for which the number of CX gates may increase ?
> 
> You may also add your reference and an explanation to the API docs (line 64)

In the context of state preparation, our simplification primarily applies when the state to be prepared is separable, as the presence of a given repetition pattern in multiplexers is quite common in these cases. Randomly generated states are less likely to exhibit such repetitions, meaning that the number of CX ports remains unchanged.  The reference was updated in the documentation.

ShellyGarion:
Could you please also add release notes to your PR ?

ACE07-Sev:
Wondering if this is needed
```py
if mux_simplify:
    circuit.append(diagonal, [q_target] + q_controls)
else:
    circuit.append(diagonal, q)
```
Isn't `q` also `[q_target] + q_controls`? If we ignore `mux_simplify`, then `q_target=q[0]`, and `q_controls=q[1:]`, ergo `q`. Given the clause above, it would still be just `q` whether `mux_simplify` is `True` or not. Please correct me if I am wrong, I may be missing something.

josejad42:
> Wondering if this is needed
> 
> ```python
> if mux_simplify:
>     circuit.append(diagonal, [q_target] + q_controls)
> else:
>     circuit.append(diagonal, q)
> ```
> 
> Isn't `q` also `[q_target] + q_controls`? If we ignore `mux_simplify`, then `q_target=q[0]`, and `q_controls=q[1:]`, ergo `q`. Given the clause above, it would still be just `q` whether `mux_simplify` is `True` or not. Please correct me if I am wrong, I may be missing something.

When there's simplification, `q` is not necessarily equal to `[q_target] + q_controls`, since `q_controls` contains only the necessary control qubits. Regardless of whether simplification is applied, we can always use
```
circuit.append(diagonal, [q_target] + q_controls)
```
to add the diagonal gate in the correct position.

josejad42:
> Could you please also add release notes to your PR ?

Hi, @ShellyGarion. The release notes have been added.

ACE07-Sev:
May I ask when this will be merged with main?

ShellyGarion:
> May I ask when this will be merged with main?

I'm still reviewing it. I hope to finish it soon :)

test/python/circuit/test_uc.py
```diff
@@ -44,6 +46,7 @@ class TestUCGate(QiskitTestCase):
             [_id, _id],
             [_id, 1j * _id],
             [_id, _not, _id, _not],
+            [_rand, _had, _rand, _had, _rand, _had, _rand, _had],
```
ShellyGarion, line 49: I would also suggest to add here:
`[_had, _had, _had, _had, _had, _had, _had, _had],`

then after line 53, please also add:
` mux_simp = [True, False],`

and in line 56, call:
` def test_ucg(self, squs, up_to_diagonal, mux_simp):`

and update lines 62 and 71 to:
`UCGate(squs, up_to_diagonal=up_to_diagonal, mux_simp=mux_simp)`

test/python/circuit/test_uc.py
```diff
@@ -100,6 +107,21 @@ def test_inverse_ucg(self):
...
+        qc1 = QuantumCircuit(4)
+        uc1 = UCGate(gate_list, up_to_diagonal=False, mux_simp=False)
+        qc1.append(uc1, range(4))
+        op1 = Operator(qc1).data
+
+        qc2 = QuantumCircuit(4)
+        uc2 = UCGate(gate_list, up_to_diagonal=False, mux_simp=True)
+        qc2.append(uc2, range(4))
+        op2 = Operator(qc2).data
+        self.assertTrue(np.allclose(op1, op2))
```
ShellyGarion, line 125: When I change `up_to_diagonal=True` in both lines 115 and 120, then the test fails, any idea why? 
If `mux_simp` should not work in case that `up_to_diagonal=True` - then it's better to state it in the docs and raise a proper error that asserts that the the two variables can not both be `True`.

ShellyGarion (commented):
Thanks for the fixes. I only have a few questions and suggestions about the tests.

alexanderivrii:
This is a cool PR! Out of curiosity, how common are multiplexers with repetitions in "practical applications"?

I have not thought about the problem too much, but it strongly reminds me of the works on more concisely representing boolean functions. For instance, we can probably express the problem in terms of (multi-valued) decision diagrams, or alternatively as expressing a truth table of a boolean function as a sum-of-products with non-overlapping products. 

One more question: the simplified representation is not unique, correct? For instance, if the following 3 control paths `(a=0) & (b=0) & (c=0)`, `(a=0) & (b=0) & (c=1)`, `(a=0) & (b=1) & (c=0)` lead to the same unitary, then we can simplify either to `(a=0) & (b=0)`, `(a=0) & (b=1) & (c=0)`, or to `(a=0) & (c=0)`, `(a=0) & (b=0) & (c=1)`.

ACE07-Sev:
@alexanderivrii Multiplexors are essentially uniformly controlled gates. The main use of them for me has been in the implementation of state and unitary synthesis by Shende. If one can implement a multiplexor with less number of gates, then in those applications the overall circuit depth would be considerably smaller.

test/python/circuit/test_uc.py
josejad42: This occurs because the final operator when we have `up_to_diagonal=True` is not equal for the original and the simplified multiplexers. But when we multiply the two operators by the respective diagonal matrices obtained by `self._get_diagonal()`, the final operators (and the prepared states) are the same.

josejad42:
@alexanderivrii As described above, quantum multiplexers are common in state preparation and unitary synthesis. 

This method introduces a simplification when a specific pattern of repetitions is detected. This pattern is very common to separable states preparation and can be detected with low computational cost. So, this is not a general simplification.

test/python/circuit/test_uc.py
josejad42: Suggestions accepted and included.

ShellyGarion (approved):
LGTM! Thanks @josejad42 for your contribution to Qiskit

PR #13100, opened 2024-09-05
Fix clippy failures with Rust 1.81.0

mtreinish:
### Summary

The recently released Rust 1.81.0 introduced some new on by default clippy rules and these rules are flagging issues in the rust code in the library. While we use Rust 1.70 for clippy in CI and these won't cause failures until we raise our MSRV to >= 1.81.0 these clippy warnings/failures are still good to fix as the either make the code more consise and/or efficient. This commit fixes these issues identified by clippy.


kevinhartman (approved):
LGTM.

PR #13111, opened 2024-09-08
Fix merge conflicts and a few minor issues in optionals.py

eliarbel:
### Summary
I'm not sure how common of a workflow it is to open a PR against a Mergify backport branch, but still: 
proposing conflict resolution and a few other minor fixes (e.g. broken Aer link and typos) against #13101. Possibly also a follow-up for #13098.

### DIFF
--- a/qiskit/utils/optionals.py
+++ b/qiskit/utils/optionals.py
@@ -27,32 +27,18 @@ Qiskit Components
 
 .. py:data:: HAS_AER
 
-<<<<<<< HEAD
-    * - .. py:data:: HAS_AER
-      - `Qiskit Aer <https://qiskit.org/ecosystem/aer/>`__ provides high-performance simulators for
-        the quantum circuits constructed within Qiskit.
-
-    * - .. py:data:: HAS_IBMQ
-      - The Qiskit IBMQ Provider was historically used for accessing IBM Quantum hardware in the IBM
-        cloud, but is now deprecated.
-
-    * - .. py:data:: HAS_IGNIS
-      - Qiskit Ignis provided tools for quantum hardware verification, noise characterization, and
-        error correction, but is now deprecated.
-=======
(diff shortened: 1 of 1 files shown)


qiskit/utils/optionals.py
```diff
@@ -27,32 +27,18 @@
...
+    `Qiskit Aer <https://qiskit.github.io/qiskit-aer/>`__ provides high-performance simulators for
     the quantum circuits constructed within Qiskit.
 
 .. py:data:: HAS_IBMQ
 
-    The :mod:`Qiskit IBMQ Provider <qiskit.providers.ibmq>` is used for accessing IBM Quantum
-    hardware in the IBM cloud.
->>>>>>> 254ba83dc (Fix documentation for utils.optionals data values (#13098))
+    The Qiskit IBMQ Provider was historically used for accessing IBM Quantum hardware in the IBM
+    cloud, but is now deprecated.
```
ElePT:
```suggestion
    cloud, but is now archived.
```

qiskit/utils/optionals.py
```diff
@@ -27,32 +27,18 @@
...
 
 .. py:data:: HAS_IGNIS
 
-    :mod:`Qiskit Ignis <qiskit.ignis>` provides tools for quantum hardware verification, noise
-    characterization, and error correction.
+    Qiskit Ignis was historically used to provide tools for quantum hardware verification,
+    noise characterization, and error correction, but is now deprecated.
```
ElePT:
```suggestion
    noise characterization, and error correction, but is now archived.
```

ElePT (commented):
I tend to push directly to the mergify branch, but I think a PR works too. The only downside to opening a PR is that with it, the backport requires merging 2 PRs instead of one, which consumes more CI resources. It might make more sense in cases like these where a 3rd person is fixing the conflicts and wants approval from the original author.

Eric-Arellano (approved):
Thanks for fixing!

PR #13112, opened 2024-09-08
Pin dependency versions for reproducible installs 

### LINKED ISSUE #11349: have a "lockfile" when a major reaches end-of-life
### What should we add?

When a Qiskit major version reaches end-of-life (mid 2024 in the case of 0.*) it would be great to have a way to keep it installable with working dependencies.

eliarbel:
### Summary
This PR pins the version numbers of Qiskit 0.46 dependencies in `constraints.txt`. Dependency versions are generated using the `pip-compile` tool from the [pip-tools package](https://pypi.org/project/pip-tools/). See `constraints.txt` file for the actual parameters used to generate the dep versions. 

### Details and comments
Close #11349


eliarbel:
@mtreinish  a few questions: 

1. The initial version was generated on Linux. Running `pip-compile` on Windows yields a slightly different outcome. Mainly this: `colorama==0.4.6`. We can add this to `constraints.txt` with an environment marker `; platform_system=="Windows"`, but actually the question is: should include version constraints for all environments or only a subset? If for all envs, should we have one constraint file for all environments? or multiple, one for each environment? (which will require I think some trickery to make `pip` honor when installing)
2. Running `pip-compile` with `--all-extras` result with it adding `qiskit==1.2.0` via `qiskit-qasm3-import`. Should we pin all extras? If so, any suggestion on how to address the `qiskit-qasm3-importer` requirement?

jakelishman:
`qiskit-qasm3-import` should have at least one version that works pre-1.0, since I wrote it before 1.0 was ever conceived. It's probably just being weird because of the bi-directional dependency and the awkwardness in how we swapped from `qiskit-terra` to `qiskit` as the main dependency. I think that makes it all the more important to include a working version in the lockfile.

For the other bit: my rough feeling is that we ought to have working lockfiles for all Tier 1 targets, and not worry about others. That means having working ones for Linux, macOS and Windows, all x86-64. I don't think (but might misremember) that macOS ARM only became tier 1 later than 0.46.

eliarbel:
Thanks @jakelishman. So regarding `qiskit-qasm3-import`:
 `v0.4.0` has this: `qiskit>=0.37.0,<1.0` in `setup.cfg` while `v0.4.1` has this: `qiskit>=0.37.0`. So it seems we want `v0.4.0`.

eliarbel:
but... `qiskit-qasm3-import` `v0.3.0` has this: `qiskit-terra>=0.21.0` which seems even more suitable as it doesn't mix `qiskit` with `qiskit-terra`

mtreinish:
> I don't think (but might misremember) that macOS ARM only became tier 1 later than 0.46.

macOS arm64 became tier 1 as part of the 1.1.0 release, after github added native runners and we ported isometry to rust.

> The initial version was generated on Linux. Running pip-compile on Windows yields a slightly different outcome. Mainly this: colorama==0.4.6. We can add this to constraints.txt with an environment marker ; platform_system=="Windows", but actually the question is: should include version constraints for all environments or only a subset? If for all envs, should we have one constraint file for all environments? or multiple, one for each environment? (which will require I think some trickery to make pip honor when installing)

I think we can get away with a single file and not even need the environment marker. The constraint file should be used with `-c` on `pip install` and that means it only enforces the constraint if it would be installed. So if you have colorama in the list then that version constraint will only be set when it's needed. We can use environment markers if there is a version conflict across platforms.

eliarbel:
I think this is ready for review now

ElePT (approved):
LGTM, I tried installing 0.46.3 locally without much success, but given that CI checks pass with no issues, it is likely an issue on my side. I think we can move forward with the release.

PR #13114, opened 2024-09-09
Fix an edge case in Sabre's release valve

### LINKED ISSUE #13081: Panic in SabreLayout when running on large quantum volume circuit
### Environment

- **Qiskit version**: 1.1, 1.2, and main
- **Python version**:  Any
- **Operating system**:  Linux

### What is happening?

When running a large quantum volume circuit through the `SabreLayout` pass the individual trials start to internally panic with the following:

```
thread '<unnamed>' panicked at 'called `Option::unwrap()` on a `None` value', crates/accelerate/src/sabre/layer.rs:73:52
```

once for each thread launched. Once all 20 threads panic the pass stops.

```python
from qiskit.circuit.library import QuantumVolume
import qiskit
from qiskit.transpiler.passes import SabreLayout, Unroll3qOrMore

qv = QuantumVolume(1000, seed=42)
qv.measure_all()
qc = Unroll3qOrMore()(qv)

cmap = CouplingMap.from_heavy_hex(21)
pm = PassManager([
    SabreLayout(cmap, swap_trials=20, layout_trials=20, max_iterations=4),
])

tqc = pm.run(qc)
```

### What should happen?

The pass shouldn't panic, and should successfully apply a layout and route the circuit.

### Any suggestions?

This was most likely a regression introduced by: https://github.com/Qiskit/qiskit/pull/11977 as the example works fine with Qiskit 1.0.2.

Cryoris:
### Summary

Fixes #13081.

### Details and comments

If a Sabre trial does not find a set of Swaps to route nodes, the "release valve" adds Swaps to route the two-qubit gate in between the the closest two qubits. In rare cases, this leads to _more_ than one gate being routable, which was not handled correctly previously.

As I understand it this should only ever occur if the release-valve adds a single swap and we have a scenario like
```
A - B - A - B  --apply swap--> A - A - B - B
```
If more than one swap gate is applied it should not be possible that two gates are routable, so the code now only checks for a second routable gate if we insert a single swap. If someone can find a counterexample we can change the code to check for any routable gate after the release valve 🙂


crates/accelerate/src/sabre/route.rs
```diff
@@ -573,14 +600,16 @@ pub fn swap_map_trial(
...
             // Efficiency doesn't matter much; this path never gets taken unless we're unlucky.
             current_swaps
                 .drain(..)
                 .rev()
                 .for_each(|swap| state.apply_swap(swap));
             let force_routed = state.force_enable_closest_node(&mut current_swaps);
-            routable_nodes.push(force_routed);
+            force_routed
+                .iter()
+                .for_each(|routable_node| routable_nodes.push(*routable_node));
```
kevinhartman:
```suggestion
            routable_nodes.extend(force_routed);
```
Maybe just this?

(Also, have your Rust travels transcended you beyond the use of the 'for' loop? 🤣)

crates/accelerate/src/sabre/route.rs
```diff
@@ -573,14 +600,16 @@ pub fn swap_map_trial(
         }
         if routable_nodes.is_empty() {
```
kevinhartman, line 600: I can't comment above this line, but is there any reason we shouldn't use a `SmallVec` for `routable_nodes` itself?

And, should we increase the capacity of the pre-allocation to `3` now on line 573?

crates/accelerate/src/sabre/route.rs
```diff
@@ -328,7 +330,32 @@ impl<'a, 'b> RoutingState<'a, 'b> {
             current_swaps.push([shortest_path[end], shortest_path[end - 1]]);
         }
         current_swaps.iter().for_each(|&swap| self.apply_swap(swap));
-        closest_node
+
+        // If we apply a single swap it could be that we route 2 nodes; that is a setup like
+        //  A - B - A - B
+        // and we swap the middle two qubits. This cannot happen if we apply more than 2 swaps.
```
kevinhartman:
```suggestion
        // and we swap the middle two qubits. This cannot happen if we apply 2 or more swaps.
```
I think?

crates/accelerate/src/sabre/route.rs
jakelishman: We shouldn't necessarily, because the release valve should only be taken rarely, and we don't want to overallocate the happy path, and we should still only be routing two nodes in the new case.

crates/accelerate/src/sabre/route.rs
Cryoris: Lol yes, Rust has me thinking in one-liners where possible 😂

Cryoris:
With #13119 merged, CI should pass now I think 🙂

crates/accelerate/src/sabre/route.rs
```diff
@@ -328,7 +329,32 @@ impl<'a, 'b> RoutingState<'a, 'b> {
...
+            // check if the closest node has neighbors that are now routable -- for that we get
+            // the other physical qubit that was swapped and check whether the node on it
+            // is now routable
+            let mut possible_other_qubit = current_swaps[0]
+                .iter()
+                // check if other nodes are in the front layer that are connected by this swap
+                .filter_map(|&swap_qubit| self.front_layer.qubits()[swap_qubit.index()])
+                // remove the closest_node, which we know we already routed
+                .filter(|(node_index, _other_qubit)| *node_index != closest_node)
+                .map(|(_node_index, other_qubit)| other_qubit);
```
mtreinish, line 348: Couldn't this just be a single filter map. Something like:

```suggestion
                .filter_map(|&swap_qubit| {
                    self.front_layer.qubits()[swap_qubit.index()]).map(|(node_index, other_qubit)| {
                        // remove the closest_node, which we know we already routed
                        if node_index != closest_node {
                            Some(other_qubit)
                        } else {
                            None
                        }
                    })
                });
```
Or even:

```suggestion
                .filter_map(|&swap_qubit| self.front_layer.qubits()[swap_qubit.index()])
                // remove the closest_node, which we know we already routed
                .filter_map(|(node_index, other_qubit)| if *node_index != closest_node {
                    Some(other_qubit)
                 } else {
                    None
                 })
```

Not that this makes a big difference. I'm surprised clippy isn't complaining about the filter followed by a map though.

mtreinish (approved):
This LGTM, I have a small inline nit comment on code style but it's not significant or a blocker. I'm running the 1000q QV reproducer I put in the original issue to confirm it's fixed once I confirm it's fixed I'll add this to the merge queue.

crates/accelerate/src/sabre/route.rs
jakelishman: if you're really going hard on fancy methods, you can even do `(*node_index != closest_node).then_some(other_qubit)` haha.

mtreinish:
The 1000 qubit QV circuit reproducer from https://github.com/Qiskit/qiskit/issues/13081 works with this applied so I'm good to merge this.

PR #13117, opened 2024-09-09
Make `DAGCircuit` fields `dag`, `qubits`, and `clbits` private.

kevinhartman:
### Summary

Immutable getters are added to provide the same access we were using without the risk of data incoherence.

### Details and comments

Also adds some of the same getters we've got on `CircuitData`, as of #13006.


crates/circuit/src/dag_circuit.rs
```diff
@@ -4983,6 +4983,42 @@ def _format(operand):
...
+    #[inline(always)]
+    pub fn qargs_interner(&self) -> &Interner<[Qubit]> {
+        &self.qargs_interner
+    }
+
+    /// Returns an immutable view of the Interner used for Cargs
+    #[inline(always)]
+    pub fn cargs_interner(&self) -> &Interner<[Clbit]> {
+        &self.cargs_interner
+    }
```
mtreinish, line 5004: I'm not sure about the need for these two methods. We already have `.get_qargs()` and `.get_cargs()` which are basically the only thing people would need read only access to the interner for: to convert an `Interned<Qubit>` into `&[Qubit]`. Was there something else you thought people would need this for?

mtreinish (commented):
Overall this seems like a straightforward addition and makes the API a bit safer to access these properties of the dag which we shouldn't be mutating directly in passes. I just have one inline question about the interner methods.

crates/circuit/src/dag_circuit.rs
kevinhartman: You're probably right that it'd actually be best not to expose the interners at all since we have those methods you mentioned.

I was actually just going for symmetry with CircuitData. But perhaps we can remove the interner accessors from that as well.

jakelishman: Getting the default qargs/cargs?

(Personally I found it easier to use the interners directly than to add wrappers, but I don't mind hugely.)

kevinhartman: I suppose also access would be needed for `dag_to_circuit`, though we could in that case consider making the interner fields `pub(crate)`.

Perhaps we ought to keep these two accessor methods in this PR for symmetry with `CircuitData` and revisit later, if desired.

mtreinish: I'm fine keeping them if we think there is a use case. I'm not sure there is a use case where we need to get the default qargs, because I don't think `DAGCircuit` consumers should be in the business of getting `Interned<Qubit>` objects. Consumers can work with them or convert them to `&[Qubit]` but I'm not sure there is a use case for generating the `Intered` value. I would expect anywhere we need to insert a new qarg into the interner it should probably be done internally in `DAGCircuit` via some api adding a new op node.

mtreinish: I found a use case for them in #13126 I made the interners `pub(crate)` in that pr, but methods would be better.

mtreinish (dismissed):

mtreinish (approved):

PR #13127, opened 2024-09-10
Move `Option<Box<_>>` into `ExtraInstructionAttributes`.

kevinhartman:
### Summary
The primary motivation here is to encapsulate the hairy logic of managing memory for extra attributes. By moving the `Option<Box<_>>` into the `ExtraInstructionAttributes` struct itself, clients are no longer burdened with the responsibility of checking if the `Option` contains a value, unwrapping it etc.; the semantics of that aren't relevant to clients like `DAGCircuit` and `CircuitData`. Instead, the `ExtraInstructionAttributes` acts as a container for optional attributes, and can easily drop the `Box` internally if all attributes are cleared.

A give-away that this pattern could be helpful was `ExtraInstructionAttributes::new`'s return type, which was previously `Option<Self>`.

### Details and comments
The size of `ExtraInstructionAttributes` (now held by value by instructions) should be the same size as the old `Option<Box<ExtraInstructionAttributes>>`, so this should not have memory implications regarding the size of instructions.

Note that `ExtraAttrs` is completely private to the `circuit_instruction` module, i.e. it's only an implementation detail and in no way part of the interface.


crates/circuit/src/circuit_instruction.rs
```diff
@@ -30,16 +30,21 @@ use crate::operations::{
...
-/// typically deal with this in rust space and the majority of the time they're not used in Python
-/// space either. To save memory these are put in a separate struct and are stored inside a
-/// `Box` on `CircuitInstruction` and `PackedInstruction`.
 #[derive(Debug, Clone)]
+struct ExtraAttrs {
+    label: Option<String>,
+    duration: Option<PyObject>,
+    unit: Option<String>,
+    condition: Option<PyObject>,
+}
```
jakelishman: Two minor nits:

- imo for naming it'd be good to be consistent with either "attrs" or "attributes" between the two structs.  I know it's just internal to the module (hence just being a nit) and that we used `extra_attrs` as the attribute name, haha
- can you add a replacement doc comment explaining why the split exists, for the future?

crates/circuit/src/circuit_instruction.rs
```diff
@@ -30,16 +30,21 @@ use crate::operations::{
...
+/// space either. To save memory, the attributes are stored inside a `Box` internally, so this
+/// struct is no larger than that.
+#[derive(Default, Debug, Clone)]
 pub struct ExtraInstructionAttributes {
-    pub label: Option<String>,
-    pub duration: Option<PyObject>,
-    pub unit: Option<String>,
-    pub condition: Option<PyObject>,
+    attributes: Option<Box<ExtraAttrs>>,
 }
```
jakelishman: Any particular reason to use a record struct rather than a tuple struct?

crates/circuit/src/circuit_instruction.rs
```diff
@@ -51,32 +56,111 @@ impl ExtraInstructionAttributes {
...
+        self.attributes
+            .as_deref()
+            .and_then(|attrs| attrs.duration.as_ref())
+    }
+
+    pub fn unit(&self) -> Option<&str> {
+        self.attributes
+            .as_deref()
+            .and_then(|attrs| attrs.unit.as_deref())
+    }
```
jakelishman: Do you want to put the setters for `duration` and `unit` on now, or wait til we need them?

crates/circuit/src/operations.rs
```diff
@@ -425,29 +425,25 @@ impl StandardGate {
...
+            ("unit", extra_attrs.py_unit(py).into_any()),
+            ("duration", extra_attrs.duration().to_object(py)),
+        ]
+        .into_py_dict_bound(py);
+        let mut out = gate_class.call_bound(py, args, Some(&kwargs))?;
+        if let Some(condition) = extra_attrs.condition() {
+            out = out.call_method0(py, "to_mutable")?;
+            out.setattr(py, "condition", condition)?;
         }
+        Ok(out)
```
jakelishman: This is a potential perf regression, because it now always builds a Python `dict` when before it would attempt to skip it.  Not huge, but probably better to keep avoiding it.

jakelishman (commented):
Thanks for doing this, it looks like a very sensible idea.

crates/circuit/src/circuit_instruction.rs
kevinhartman: Done in bca33aa9443b0122a8411e057e40f63b7d5eb59f.

crates/circuit/src/circuit_instruction.rs
kevinhartman: Nope! I thought of that later, but I think it makes sense.

Done in bca33aa9443b0122a8411e057e40f63b7d5eb59f.

crates/circuit/src/circuit_instruction.rs
kevinhartman: Done in bca33aa9443b0122a8411e057e40f63b7d5eb59f.

crates/circuit/src/operations.rs
kevinhartman: Good point.

Resolved in 943b14260bb08c4413ef3a1e3cdae754bb72ab58.

jakelishman (approved):
Thanks for this and the changes.  I'll leave unmerged because I saw elsewhere that Ray was going to look too.

raynelfss (approved):
This is all sensible to me, my only comment was about the usage of a map-like struct instead of a simple tuple struct but it seems it's already been corrected :)

This should be good as is. Thanks for working on it! 🚀

PR #13133, opened 2024-09-11
Make dense layout algorithm deterministic when run in parallel

mtreinish:
### Summary

The dense layout algorithm is trying to find the densest k subgraph of the connectivity graph. To do this it performs a BFS from each node in the in graph of k nodes to determine the subgraph with the most number of edges. But in cases of ties where there are subgraphs with the same number of edges the exact output would be determined by the iteration order that we're evaluating a BFS search. However, this algorithm runs in parallel in most cases and the exact iteration order isn't going to be stable when running in parallel. It will depend on which threads finish first. This commit fixes this potential non-determinism in the algorithm by defaulting to the lower node index's trial results instead of relying on the execution order. This should mean we return identical results regardless of how many threads are run or how quickly they execute.

### Details and comments

### DIFF
--- a/crates/accelerate/src/dense_layout.rs
+++ b/crates/accelerate/src/dense_layout.rs
@@ -30,6 +30,7 @@ struct SubsetResult {
     pub error: f64,
     pub map: Vec<usize>,
     pub subgraph: Vec<[usize; 2]>,
+    pub index: usize,
 }
 
 fn bfs_sort(adj_matrix: ArrayView2<f64>, start: usize, num_qubits: usize) -> Vec<usize> {
@@ -190,6 +191,7 @@ pub fn best_subset_inner(
             error,
             map: bfs,
             subgraph,
+            index: k,
         }
     };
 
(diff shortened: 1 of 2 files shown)


henryzou50 (commented):
LGTM! This resolves the non-determinism issue in the `reduce_fn` function. By introducing the index field in the `SubsetResult` struct, the comparison now has a consistent tie-breaking mechanism when count values are equal. The index ensures that, even when multiple subgraphs have the same number of edges, the result is deterministic, as the subgraph with the lower index is always chosen.

This removes the dependency on thread execution speed and order, making the function’s behavior predictable in parallel environments. Thanks!

henryzou50:
Some notes after reviewing it again. Need to add release notes and add a check for the error aware case.

mtreinish:
> Some notes after reviewing it again. Need to add release notes and add a check for the error aware case.



Done in: https://github.com/Qiskit/qiskit/pull/13133/commits/34a828705912626f3001114bb0e0d5cc8e4556e2 and https://github.com/Qiskit/qiskit/pull/13133/commits/c3cf18e14576a9d4ad04b8674e2c2619a3ecc395

jakelishman (approved):

PR #13135, opened 2024-09-11
Use new URL for link in circuit docs page

Eric-Arellano:
I missed this when updating links for the docs reorganization around Guides. We have a redirect, but it's best to point directly to the page.


Eric-Arellano:
Imo no need to cherry-pick: the redirect means this isn't a big deal. Save the trees

1ucian0 (approved):

PR #13136, opened 2024-09-11
Fix ABI3 issues identified during 1.3.0b1 release attempt

mtreinish:
### Summary

We recently tried to publish the 1.3.0b1 release unsucessfully. During the wheel build jobs abi3audit was failing because it correctly identified a couple of abi violations in the builds. The first was we were tagging the package incorrectly as being with a minimum level of Python 3.8 instead of the correct minimum of 3.9. This commit updates the setup.py tag to match the package metadata with the binary.

The second issue is we were creating `PyInit_*` symbols for submodules which goes against the abi3 naming recomendations and abi3 audit flags it as a violation. We had this issue previously in 1.2.0 but was fixed for that release. However since 1.2.0 and now several stray usages of `#[pymodule]` on submodules slipped in during development. This commit fixes those so the bad symbol names won't be created anymore.

Once this commit merges we should tag the new commit as 1.3.0b1.

### Details and comments

Co-authored-by: Raynel Sanchez <87539502+raynelfss@users.noreply.github.com>

### DIFF
--- a/crates/accelerate/src/circuit_library/mod.rs
+++ b/crates/accelerate/src/circuit_library/mod.rs
@@ -14,7 +14,6 @@ use pyo3::prelude::*;
 
 mod entanglement;
 
-#[pymodule]
 pub fn circuit_library(m: &Bound<PyModule>) -> PyResult<()> {
     m.add_wrapped(wrap_pyfunction!(entanglement::get_entangler_map))?;
     Ok(())
--- a/crates/accelerate/src/commutation_analysis.rs
+++ b/crates/accelerate/src/commutation_analysis.rs
@@ -12,7 +12,7 @@
 
 use pyo3::exceptions::PyValueError;
 use pyo3::prelude::PyModule;
-use pyo3::{pyfunction, pymodule, wrap_pyfunction, Bound, PyResult, Python};
+use pyo3::{pyfunction, wrap_pyfunction, Bound, PyResult, Python};
 use qiskit_circuit::Qubit;
 
 use crate::commutation_checker::CommutationChecker;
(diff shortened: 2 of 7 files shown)


mtreinish:
I ran `pip wheel --no-deps .` and `abi3audit --strict --report qiskit-1.3.0b1-cp39-abi3-linux_x86_64.whl` with this PR applied and it returned:

```
{"specs": {"qiskit-1.3.0b1-cp39-abi3-linux_x86_64.whl": {"kind": "wheel", "wheel": [{"name": "_accelerate.abi3.so", "result": {"is_abi3": true, "is_abi3_baseline_compatible": true, "baseline": "3.9", "computed": "3.9", "non_abi3_symbols": [], "future_abi3_objects": {}}}]}}}
```
So I think with this in place we should be good to retry tagging 1.3.0b1.

raynelfss:
Ran the same tests locally with MacOS. Results are also good here:
```bash
% abi3audit --strict --verbose --report qiskit-1.3.0b1-cp39-abi3-macosx_14_0_arm64.whl
{"specs": {"qiskit-1.3.0b1-cp39-abi3-macosx_14_0_arm64.whl": {"kind": "wheel", "wheel": [{"name": "_accelerate.abi3.so", "result": {"is_abi3": true, "is_abi3_baseline_compatible": true, "baseline": "3.9", "computed": "3.9", "non_abi3_symbols": [], "future_abi3_objects": {}}}]}}}
```

raynelfss (approved):
LGTM! This should hopefully fix all the failures in CI. Thank you for the quick updates!

PR #12811, opened 2024-07-24
[Oxidize BasisTranslator]: Move `basis_search` and `BasisSearchVisitor` to rust.

raynelfss:
### Summary
Builds on #12585
The following commits add a rust-side version of `basis_search` that is also exposed to `BasisTranslator` in Python. Due to structural restrictions in `rustworkx-core`'s version of `dijkstra_search`, these commits also consolidate the `basis_search` method and the `BasisSearchVisitor` into one method.

### Details and comments
The `BasisTranslator` aims to map a circuit's gates into a specified target basis by performing a dijkstra search in a provided `EquivalenceLibrary`. Because of #12585 moving said structure mainly to rust, these commits aim to add the `basis_search` method of this transpiler pass to operate in rust space as well, which could save the overhead of converting the rust `EquivalenceLibrary`'s `petgraph::StableDiGraph` to a `PyDiGraph`.

It is worth to note that the following changes have been made:
- `BasisSearchVisitor` and `_basis_search` have been consolidated into one method `basis_search`, due to the visitor's data living in a different space, which would imply moving some of the `rustworkx-core`'s `dijkstra_search` retrieved graph data moving outside of its expected lifetimes.
- `basis_seach` is an entirely rust-based method. However, it is exposed temporarily to Python via `py_basis_search` to avoid cloning of `source_basis` and `target_basis` sets. This exposure however will be removed once #12246 is fully realized.

### Known issues:
~~When running the test `test.python.transpiler.test_basis_translator.TestBasisExamples.test_cx_bell_to_ecr` there's a posibility that the resulting circuit will randomly be:~~

<pre style="word-wrap: normal;white-space: pre;background: #fff0;line-height: 1.1;font-family: &quot;Courier New&quot;,Courier,monospace">        ┌────────────┐   ┌─────────────┐┌──────┐┌──────────┐
q_0: ───┤ U(π/2,0,π) ├───┤ U(0,0,-π/2) ├┤0     ├┤ U(π,0,π) ├
     ┌──┴────────────┴──┐└─────────────┘│  Ecr │└──────────┘
q_1: ┤ U(-π/2,-π/2,π/2) ├───────────────┤1     ├────────────
     └──────────────────┘               └──────┘            </pre>

~~or:~~
<pre style="word-wrap: normal;white-space: pre;background: #fff0;line-height: 1.1;font-family: &quot;Courier New&quot;,Courier,monospace">        ┌────────────┐  ┌─────────────┐┌──────────┐┌──────┐
[...]

Issue #12246, opened 2024-04-19
Port `BasisTranslator` to Rust


mtreinish:
This is blocked by #11721, #12205, and #12278.
```[tasklist]
### Parts of the BasisTranslator
- [ ] https://github.com/Qiskit/qiskit/pull/12585
- [ ] https://github.com/Qiskit/qiskit/pull/13137
- [ ] https://github.com/Qiskit/qiskit/pull/12811
- [ ] #13237
```

PR #13137, opened 2024-09-11
[Oxidize BasisTranslator] Add rust-native `compose_transforms()`

raynelfss:
### Summary
The following commits add the helper function `compose_transforms` from the `BasisTranslator` transpiler pass into rust. This is a small progression towards completing #12246. This function will be temporarily exposed to Python until #12246 is fully realized.

### Details and comments
As we continue moving parts of the `BasisTranslator` into Rust thanks to #12585 and #12811. The next logical step is to bring in those functions that relied heavily on the `DAGCircuit` and since it now lives in Rust we can now leverage its usage by moving this helper function into Rust.

`compose_transforms` basically returns a mapping of `(gate_name, num_qubits) : (parameters, gate_definition_as_dag)` for each gate present in the `DAGCircuit` that is currently being processed. It also breaks down `ControlFlowOp` to include its definitions as well.

Using the latest additions by #13036 we can perform most of these conversions from rust using both `circuit_to_dag` or `DAGCirctut::from_circuit_data`. This function will be temporarily exposed to Python until #12246 is fully realized.

### Blockers
- [x] #13143

**These are not strictly blockers but some nice-to-haves.**
- [x] #12585 
- [ ] #12811


raynelfss:
Here are some benchmarks
```bash
All benchmarks:

| Change   | Before [81063a79] <fix-abi3-issues~1>   | After [e8c957df] <rust-compose-trans>   |   Ratio | Benchmark (Parameter)                                                                                    |
|----------|-----------------------------------------|-----------------------------------------|---------|----------------------------------------------------------------------------------------------------------|
|          | 306±1ms                                 | 303±2ms                                 |    0.99 | passes.MultipleBasisPassBenchmarks.time_basis_translator(14, 1024, ['rx', 'ry', 'rz', 'r', 'rxx', 'id']) |
|          | 443±3ms                                 | 437±3ms                                 |    0.99 | passes.MultipleBasisPassBenchmarks.time_basis_translator(20, 1024, ['rx', 'ry', 'rz', 'r', 'rxx', 'id']) |
|          | 188±0.8ms                               | 182±2ms                                 |    0.97 | passes.MultipleBasisPassBenchmarks.time_basis_translator(14, 1024, ['rz', 'x', 'sx', 'cx', 'id'])        |
|          | 217±0.7ms                               | 207±0.6ms                               |    0.96 | passes.MultipleBasisPassBenchmarks.time_basis_translator(20, 1024, ['u', 'cx', 'id'])                    |
|          | 102±0.5ms                               | 98.1±1ms                                |    0.96 | passes.MultipleBasisPassBenchmarks.time_basis_translator(5, 1024, ['rx', 'ry', 'rz', 'r', 'rxx', 'id'])  |
|          | 266±4ms                                 | 253±0.5ms                               |    0.95 | passes.MultipleBasisPassBenchmarks.time_basis_translator(20, 1024, ['rz', 'x', 'sx', 'cx', 'id'])        |
|          | 155±3ms                                 | 146±0.6ms                               |    0.94 | passes.MultipleBasisPassBenchmarks.time_basis_translator(14, 1024, ['u', 'cx', 'id'])                    |
|          | 68.6±1ms                                | 62.6±0.2ms                              |    0.91 | passes.MultipleBasisPassBenchmarks.time_basis_translator(5, 1024, ['rz', 'x', 'sx', 'cx', 'id'])         |
| -        | 57.5±0.4ms                              | 51.8±1ms                                |    0.9  | passes.MultipleBasisPassBenchmarks.time_basis_translator(5, 1024, ['u', 'cx', 'id'])                     |

PERFORMANCE INCREASED.
```

```bash
All benchmarks:

| Change   | Before [81063a79] <fix-abi3-issues~1>   | After [e8c957df] <rust-compose-trans>   | Ratio   | Benchmark (Parameter)                                                                                     |
|----------|-----------------------------------------|-----------------------------------------|---------|-----------------------------------------------------------------------------------------------------------|
|          | 5.91±0s                                 | 5.13±0.01s                              | ~0.87   | randomized_benchmarking.RandomizedBenchmarkingBenchmark.time_ibmq_backend_transpile([0, 1])               |
|          | 5.93±0.02s                              | 5.11±0s                                 | ~0.86   | randomized_benchmarking.RandomizedBenchmarkingBenchmark.time_ibmq_backend_transpile_single_thread([0, 1]) |
| -        | 2.53±0.01s                              | 2.26±0.01s                              | 0.89    | randomized_benchmarking.RandomizedBenchmarkingBenchmark.time_ibmq_backend_transpile([0])                  |
| -        | 2.56±0.01s                              | 2.25±0.01s                              | 0.88    | randomized_benchmarking.RandomizedBenchmarkingBenchmark.time_ibmq_backend_transpile_single_thread([0])    |

PERFORMANCE INCREASED.
```

crates/accelerate/src/basis/basis_translator/mod.rs
```diff
@@ -0,0 +1,21 @@
+// This code is part of Qiskit.
```
kevinhartman, line 1: Is there a clear benefit to having a separate folder here for `basis_translator`?

I don't know that I mind it, but it's different from the folder structure we had in Python, and none of the other accelerate modules really go more than one module deep (except for `synthesis`, which seems to copy the organization of the Python module).

crates/accelerate/src/basis/basis_translator/compose_transforms.rs
```diff
@@ -0,0 +1,206 @@
...
+    circuit_data::CircuitData,
+    dag_circuit::{DAGCircuit, NodeType},
+    operations::{Operation, Param},
+};
+use smallvec::SmallVec;
+
+// Custom types
+// TODO: Remove these and use the version from `basis_search`
+pub type BasisTransforms = Vec<(String, u32, SmallVec<[Param; 3]>, CircuitRep)>;
+pub type MappedTransforms = HashMap<(String, u32), (SmallVec<[Param; 3]>, DAGCircuit)>;
```
kevinhartman: IMO, type aliases are generally more helpful when used to break down complex types into logical ideas, rather than as a means of typing less.

For example, this:

```rust
pub type GateIdentifier = (String, u32);
pub type BasisTransformIn = (SmallVec<[Param; 3]>, CircuitRep);
pub type BasisTransformOut = (SmallVec<[Param; 3]>, DAGCircuit);

pub(super) fn py_compose_transforms(
    py: Python,
    basis_transforms: Vec<(GateIdenfitier, BasisTransformIn)>,
    source_basis: HashSet<GateIdentifier>,
    source_dag: &DAGCircuit,
) -> PyResult<HashMap<GateIdentifier, BasisTransformOut>> 
```

gives the reader a lot more information about the call semantics of the function compared to this:

```rust
pub type BasisTransforms = Vec<(String, u32, SmallVec<[Param; 3]>, CircuitRep)>;
pub type MappedTransforms = HashMap<(String, u32), (SmallVec<[Param; 3]>, DAGCircuit)>;

pub(super) fn py_compose_transforms(
    py: Python,
    basis_transforms: BasisTransforms,
    source_basis: HashSet<(String, u32)>,
    source_dag: &DAGCircuit,
) -> PyResult<MappedTransforms> 
```

And, as a general rule of thumb, it's usually best to avoid aliasing data structures like `Vec` and `HashMap` since it hides their otherwise very familiar semantics from the reader (...though, we're guilty of this in a few places, I expect.)

crates/accelerate/src/basis/basis_translator/compose_transforms.rs
```diff
@@ -0,0 +1,206 @@
...
+/// instance of `CircuitData`.
+#[derive(Debug, Clone)]
+pub struct CircuitRep(pub CircuitData);
+
+impl FromPyObject<'_> for CircuitRep {
+    fn extract_bound(ob: &Bound<'_, PyAny>) -> PyResult<Self> {
+        if ob.is_instance(QUANTUM_CIRCUIT.get_bound(ob.py()))? {
+            let data: Bound<PyAny> = ob.getattr("_data")?;
+            let data_downcast: Bound<CircuitData> = data.downcast_into()?;
+            let data_extract: CircuitData = data_downcast.extract()?;
```
kevinhartman: I believe you can extract from the `Bound<PyAny>` without first downcasting to save an `isinstance` check.

crates/circuit/src/dag_circuit.rs
```diff
@@ -2861,7 +2861,7 @@ def _format(operand):
     /// Raises:
     ///     DAGCircuitError: if met with unexpected predecessor/successors
     #[pyo3(signature = (node, input_dag, wires=None, propagate_condition=true))]
-    fn substitute_node_with_dag(
+    pub fn substitute_node_with_dag(
```
kevinhartman: Perhaps a strange request, but I think it makes sense for us to rename any PyO3 methods to `py_` when opening them up outside of `DAGCircuit`. The reason being that it'll make it easier for us to notice calls to Python-intended methods from within Rust, which are likely to be ill-performing in contrast to anything proper Rust native.

crates/accelerate/src/basis/basis_translator/compose_transforms.rs
```diff
@@ -0,0 +1,206 @@
...
+        }
+    }
+    Ok(mapped_instructions)
+}
+
+fn get_example_gates(
+    py: Python,
+    dag: &DAGCircuit,
+    example_gates: Option<Box<HashMap<(String, u32), PackedInstruction>>>,
+) -> PyResult<Box<HashMap<(String, u32), PackedInstruction>>> {
```
kevinhartman: The original Python function is _strange_. A more appropriate way to do this in Rust would be to require that the caller allocates the map and provides a mutable reference to it so all recursive calls to the function can keep appending to it.

```suggestion
    example_gates: &mut HashMap<(String, u32), PackedInstruction>,
) -> PyResult<()> {
```

crates/accelerate/src/basis/basis_translator/compose_transforms.rs
```diff
@@ -0,0 +1,206 @@
...
+    let mut example_gates = example_gates.unwrap_or_default();
+    for node in dag.op_nodes(true) {
+        if let Some(NodeType::Operation(op)) = dag.dag().node_weight(node) {
```
kevinhartman, line 155: I think we now have `NodeType::unwrap_operation` so you can use that here instead of the `if let`.

crates/accelerate/src/basis/basis_translator/compose_transforms.rs
```diff
@@ -0,0 +1,206 @@
...
+            example_gates.insert((op.op.name().to_string(), op.op.num_qubits()), op.clone());
+            if op.op.control_flow() {
+                let OperationRef::Instruction(inst) = op.op.view() else {
+                    panic!("Control flow op can only be an instruction")
+                };
+                let inst_bound = inst.instruction.bind(py);
+                let blocks = inst_bound.getattr("blocks")?;
```
kevinhartman: And, you of course now have `Operation::blocks` :)

crates/accelerate/src/basis/basis_translator/compose_transforms.rs
```diff
@@ -0,0 +1,206 @@
...
+                for block in blocks.iter()? {
+                    let block_as_dag = circuit_to_dag(py, block?.extract()?, true, None, None)?;
+                    example_gates = get_example_gates(py, &block_as_dag, Some(example_gates))?;
```
kevinhartman: Rather than convert the `CircuitData` to a `DAGCircuit`, it'd be more efficient to handle the `CircuitData`s directly.

There's an example of this in `GatesInBasis`. For that, we have handling for the `DAGCircuit` in the root of the function toward the bottom, and define a recursive inner function to then process nested `CircuitData`s.

https://github.com/Qiskit/qiskit/pull/13034/files#diff-6cbc6dd8e1b6e781868b664eee4429814a0cedbe4a4f4d8393413e15c9ae5878R36-R102

crates/accelerate/src/basis/basis_translator/compose_transforms.rs
```diff
@@ -0,0 +1,206 @@
...
+    basis_transforms: &'a BasisTransforms,
+    source_basis: &'a HashSet<(String, u32)>,
+    source_dag: &'a DAGCircuit,
+) -> PyResult<MappedTransforms> {
+    let example_gates = *get_example_gates(py, source_dag, None)?;
+    let mut mapped_instructions: MappedTransforms = HashMap::new();
+
+    for (gate_name, gate_num_qubits) in source_basis.iter().cloned() {
+        // Need to grab a gate instance to find num_qubits and num_params.
+        // Can be removed following https://github.com/Qiskit/qiskit-terra/pull/3947 .
```
kevinhartman: This issue appears to be closed. I think the comment may be outdated in your implementation, anyhow?

crates/accelerate/src/basis/basis_translator/compose_transforms.rs
```diff
@@ -0,0 +1,206 @@
...
+// Custom types
+// TODO: Remove these and use the version from `basis_search`
+pub type BasisTransforms = Vec<(String, u32, SmallVec<[Param; 3]>, CircuitRep)>;
+pub type MappedTransforms = HashMap<(String, u32), (SmallVec<[Param; 3]>, DAGCircuit)>;
+// TODO: Remove these and use the version from `EquivalenceLibrary`
+
+/// Representation of QuantumCircuit which the original circuit object + an
+/// instance of `CircuitData`.
+#[derive(Debug, Clone)]
+pub struct CircuitRep(pub CircuitData);
```
kevinhartman: Maybe `CircuitFromPython` could be a nice name, similar to how we have `OperationFromPython`.

qiskit/transpiler/passes/basis/basis_translator.py
```diff
@@ -597,97 +594,3 @@ def _basis_search(equiv_lib, source_basis, target_basis):
...
-        placeholder_gate = Gate(gate_name, gate_num_qubits, list(placeholder_params))
-        placeholder_gate.params = list(placeholder_params)
-
-        dag = DAGCircuit()
-        qr = QuantumRegister(gate_num_qubits)
-        dag.add_qreg(qr)
-        dag.apply_operation_back(placeholder_gate, qr, (), check=False)
-        mapped_instrs[gate_name, gate_num_qubits] = placeholder_params, dag
-
-    for gate_name, gate_num_qubits, equiv_params, equiv in basis_transforms:
```
kevinhartman, line 644: It looks like the documentation is wrong then for `basis_transforms` in the original Python code, no? Since the tuple does not have the number of bits there.

crates/accelerate/src/basis/basis_translator/compose_transforms.rs
```diff
@@ -0,0 +1,206 @@
...
+                None
+            } else {
+                Some(gate_obj.params)
+            },
+            gate_obj.extra_attrs,
+            #[cfg(feature = "cache_pygates")]
+            Some(gate.into()),
+        )?;
+        mapped_instructions.insert(
+            (gate_name.clone(), gate_num_qubits),
```
kevinhartman: Do we need to clone `gate_name` here, or can we just consume it? I don't see anymore uses after this :)

crates/accelerate/src/basis/basis_translator/compose_transforms.rs
```diff
@@ -0,0 +1,206 @@
...
+            .unwrap()
+            .unbind()
+    }
+}
+
+impl ToPyObject for CircuitRep {
+    fn to_object(&self, py: Python<'_>) -> PyObject {
+        self.clone().into_py(py)
+    }
+}
```
kevinhartman: Are these used?

kevinhartman (commented):
Looks like a good start, thanks for working on this.

Unfortunately, the Python code you're starting with is probably sorely in need of a rewrite. I'm not entirely familiar with it, but it seems riddled with wrong / bad docs and tuple abuse.

I'll need to do a second pass once you've had a look at the comments thus far. I haven't spent enough time reviewing the parameter mapping stuff, specifically.

crates/accelerate/src/basis/basis_translator/compose_transforms.rs
kevinhartman: Actually, that would come as part of https://github.com/Qiskit/qiskit/pull/13034/files#diff-9d10dd27b7540092c29eb23e0fe83a90b8661fb1f452178b60f0e8fb4fd54d60R85

crates/accelerate/src/basis/basis_translator/mod.rs
raynelfss: This is just to plan ahead, since the `BasisTranslator` has a couple of moving parts and since I am separating things at the moment to make them more readable. If you take a look at #12811, that feature also exists in that same folder. but in a different file If you think it shouldn't be this way it can be changed.

crates/accelerate/src/basis/basis_translator/compose_transforms.rs
raynelfss: Yes, this is leftover code from when I was using `Borrow` instead of `extract()`, I will remove the downcast.

crates/accelerate/src/basis/basis_translator/compose_transforms.rs
raynelfss: Yes, I am planning on using it here. However, it does throw a wrench in some of my plans because I can't necessarily verify that `CircuitData` has any calibrations for a specific instruction due to that part of the circuit living entirely on Python.

EDIT: Oh I think I was wrong, I can use it without issue here.

Until then I might keep this as is, but my upcoming implementation of `BasisTranslator` does use `Operation::blocks`,  thank you for that by the way.

crates/accelerate/src/basis/basis_translator/compose_transforms.rs
raynelfss: Yes, this was a question I had. Not sure how to tackle it since it seems that issue has been abandoned.

crates/accelerate/src/basis/basis_translator/compose_transforms.rs
raynelfss: I don't think so, at least not here. This is leftover code brought over from #12585.

crates/accelerate/src/basis/basis_translator/mod.rs
kevinhartman: It suppose it doesn't really matter much, so if you think it helps to organize I'm good with it 🙂.

crates/accelerate/src/basis/basis_translator/compose_transforms.rs
kevinhartman: I would probably just remove the comment, and add a method comment to `get_example_gates` that explains that its purpose is just to get the number of parameters for a gate with a specific name and number of qubits, since that info is only found on an instruction.

(...also, we should probably rename `get_example_gates` to `get_num_params` and change the return type to  `HashMap<(String, u32), usize>`. As it is, we don't use the `PackedInstruction`s coming out of the function (do we?) and they are being cloned)

crates/accelerate/src/basis/basis_translator/compose_transforms.rs
```diff
@@ -0,0 +1,202 @@
...
+    source_basis: HashSet<GateIdentifier>,
+    source_dag: &DAGCircuit,
+) -> PyResult<HashMap<GateIdentifier, BasisTransformOut>> {
+    compose_transforms(py, &basis_transforms, &source_basis, source_dag)
+}
+
+pub(super) fn compose_transforms<'a>(
+    py: Python,
+    basis_transforms: &'a [(GateIdentifier, BasisTransformIn)],
+    source_basis: &'a HashSet<(String, u32)>,
```
kevinhartman:
```suggestion
    source_basis: &'a HashSet<GateIdentifier>,
```

crates/accelerate/src/basis/basis_translator/compose_transforms.rs
```diff
@@ -0,0 +1,202 @@
...
+    source_dag: &'a DAGCircuit,
+) -> PyResult<HashMap<GateIdentifier, BasisTransformOut>> {
+    let mut example_gates: HashMap<(String, u32), PackedInstruction> = HashMap::default();
```
kevinhartman:
```suggestion
    let mut example_gates: HashMap<GateIdentifier, PackedInstruction> = HashMap::default();
```

crates/accelerate/src/basis/basis_translator/compose_transforms.rs
```diff
@@ -0,0 +1,202 @@
...
+                        true,
+                    )?;
+                }
+            }
+        }
+    }
+    Ok(mapped_instructions)
+}
+
+fn get_example_gates(
```
kevinhartman: I made this comment elsewhere, but I think this can just be `get_num_params` and return `HashMap<GateIdentifier, usize>`, where the value is the number of parameters, since we don't actually use the `PackedInstruction` otherwise, I think.

That would save us from having to clone all of the nodes' `PackedInstruction`s.

crates/accelerate/src/basis/basis_translator/compose_transforms.rs
```diff
@@ -0,0 +1,202 @@
...
+    dag: &DAGCircuit,
+    example_gates: &mut HashMap<(String, u32), PackedInstruction>,
```
kevinhartman:
```suggestion
    example_gates: &mut HashMap<GateIdentifier, PackedInstruction>,
```

crates/accelerate/src/basis/basis_translator/compose_transforms.rs
```diff
@@ -0,0 +1,202 @@
...
+                }
+            }
+        }
+    }
+    Ok(())
+}
+
+fn get_example_gates_circuit(
+    circuit: &CircuitData,
+    example_gates: &mut HashMap<(String, u32), PackedInstruction>,
```
kevinhartman:
```suggestion
    example_gates: &mut HashMap<GateIdentifier, PackedInstruction>,
```

crates/accelerate/src/basis/basis_translator/compose_transforms.rs
```diff
@@ -0,0 +1,202 @@
...
+            },
+            gate_obj.extra_attrs,
+            #[cfg(feature = "cache_pygates")]
+            Some(gate.into()),
+        )?;
+        mapped_instructions.insert((gate_name, gate_num_qubits), (placeholder_params, dag));
+
+        for ((gate_name, gate_num_qubits), (equiv_params, equiv)) in basis_transforms {
+            for (_, dag) in &mut mapped_instructions.values_mut() {
+                let doomed_nodes = dag
```
kevinhartman: I know the name `doomed_nodes` is faithful to the original Python code, but I don't think it's particularly descriptive (even though it is **fun**) for what is happening here.

Maybe `nodes_to_replace` or something to that effect?

crates/accelerate/src/basis/basis_translator/compose_transforms.rs
```diff
@@ -0,0 +1,202 @@
...
+                    let replace_dag: DAGCircuit =
+                        DAGCircuit::from_circuit_data(py, replacement.0, true)?;
+                    let op_node = dag.get_node(py, node)?;
+                    dag.py_substitute_node_with_dag(
+                        py,
+                        op_node.bind(py),
+                        &replace_dag,
+                        None,
+                        true,
+                    )?;
```
kevinhartman, line 138: I imagine this is slower than it ought to be, but it'd be non-trivial to expose a native-Rust `substitute_node_with_dag`. Maybe we can revisit doing so in a future PR.

kevinhartman (commented):
Just a few more comments. Thanks! :)

crates/accelerate/src/basis/basis_translator/compose_transforms.rs
```diff
@@ -0,0 +1,209 @@
...
+use smallvec::SmallVec;
+
+// Custom types
+pub type GateIdentifier = (String, u32);
+pub type BasisTransformIn = (SmallVec<[Param; 3]>, CircuitFromPython);
+pub type BasisTransformOut = (SmallVec<[Param; 3]>, DAGCircuit);
+// TODO: Remove these and use the version from `EquivalenceLibrary`
+
+/// Representation of QuantumCircuit which the original circuit object + an
+/// instance of `CircuitData`.
```
kevinhartman: It might be good to phrase this to be more clear that it is just used to extract a `CircuitData` from a Python QuantumCircuit, e.g. "Used to extract a Rust `CircuitData` from the Python `QuantumCircuit`."

crates/accelerate/src/basis/basis_translator/compose_transforms.rs
```diff
@@ -0,0 +1,209 @@
...
+    compose_transforms(py, &basis_transforms, &source_basis, source_dag)
+}
+
+pub(super) fn compose_transforms<'a>(
+    py: Python,
+    basis_transforms: &'a [(GateIdentifier, BasisTransformIn)],
+    source_basis: &'a HashSet<GateIdentifier>,
+    source_dag: &'a DAGCircuit,
+) -> PyResult<HashMap<GateIdentifier, BasisTransformOut>> {
+    let mut example_gates: HashMap<GateIdentifier, usize> = HashMap::default();
```
kevinhartman: I think this should be renamed now to something like `gate_param_counts`.

kevinhartman (commented):
Last two minor comments. Otherwise, it looks reasonable to me.

crates/accelerate/src/basis/basis_translator/compose_transforms.rs
raynelfss: I will actually add those changes to the original struct which is located in `equivalence.rs`. And remove the duplicate from here.

kevinhartman (approved):
LGTM, thanks for doing this work!

PR #13139, opened 2024-09-12
Port TwoQubitControlledUDecomposer to rust

### LINKED ISSUE #12907: Port `TwoQubitControlledUDecomposer` to rust

ShellyGarion:
### Summary
close #12907

### Details and comments
depends on #13168 

Performance improvement:
Synthesizing 100 random 4x4 unitary matrices into each of the basis gates `[RXX, RYY, RZZ, RZX, CRZ, CPhase]` takes:
in Python: 20.57 sec
in Rust: 8.85


crates/accelerate/src/two_qubit_decompose.rs
```diff
@@ -2344,6 +2344,352 @@ pub fn local_equivalence(weyl: PyReadonlyArray1<f64>) -> PyResult<[f64; 3]> {
...
+            _ => panic!(),
+        })
+        .collect::<SmallVec<_>>();
+    (inv_gate.0, inv_gate_params, gate.2)
+}
+
+#[pyclass(module = "qiskit._accelerate.two_qubit_decompose", subclass)]
+pub struct TwoQubitControlledUDecomposer {
+    #[pyo3(get)]
+    rxx_equivalent_gate: StandardGate,
```
mtreinish: Do you think we should treat this class as public at all? The python space definition wasn't limited to standard gates, it would support any python space `Gate` object including a custom defined gate. This will error in the case of a custom `PyGate` though.

crates/accelerate/src/two_qubit_decompose.rs
```diff
@@ -2344,6 +2344,352 @@ pub fn local_equivalence(weyl: PyReadonlyArray1<f64>) -> PyResult<[f64; 3]> {
...
+    }
+}
+
+#[pymethods]
+impl TwoQubitControlledUDecomposer {
+    #[new]
+    #[pyo3(signature=(rxx_equivalent_gate))]
+    fn new(rxx_equivalent_gate: StandardGate) -> PyResult<Self> {
+        TwoQubitControlledUDecomposer::new_inner(rxx_equivalent_gate)
+    }
```
mtreinish: We can merge `new_inner` into this and make this public. The only reason to have a split definition like this is if the python signature needs different types than the rust caller. For example, `__call__` vs `call_inner` is a `PyReadOnlyArray2<Complex64>` vs a `ArrayView2<Complex64>`.

crates/accelerate/src/two_qubit_decompose.rs
```diff
@@ -2344,6 +2344,352 @@ pub fn local_equivalence(weyl: PyReadonlyArray1<f64>) -> PyResult<[f64; 3]> {
...
+    #[pyo3(signature=(unitary, atol))]
+    fn __call__(
+        &self,
+        unitary: PyReadonlyArray2<Complex64>,
+        atol: f64,
+    ) -> PyResult<TwoQubitGateSequence> {
```
mtreinish: Is there a reason we can't return a `CircuitData` here? That way we're building the output circuit directly instead of this intermediate vec. This should speed up the python space class's `__call__` function because we've built the circuit object in rust instead of in python.

qiskit/synthesis/two_qubit/two_qubit_decompose.py
```diff
@@ -280,157 +279,29 @@ def __init__(self, rxx_equivalent_gate: Type[Gate]):
...
-        self._weyl_gate(circ)
-        circ.compose(c1r, [0], inplace=True)
-        circ.compose(c1l, [1], inplace=True)
-        return circ
-
-    def _to_rxx_gate(self, angle: float) -> QuantumCircuit:
-        """
-        Takes an angle and returns the circuit equivalent to an RXXGate with the
-        RXX equivalent gate as the two-qubit unitary.
+        self.rxx_equivalent_gate = self._inner_decomposition.rxx_equivalent_gate
```
mtreinish: This makes a second copy of the inner rust gate object which won't be the most useful from python. I would probably store the python gate instead:

```suggestion
        self.rxx_equivalent_gate = rxx_equivalent_gate
```

qiskit/synthesis/two_qubit/two_qubit_decompose.py
```diff
@@ -280,157 +279,29 @@ def __init__(self, rxx_equivalent_gate: Type[Gate]):
...
-        c1l, c1r, c2l, c2r = (
-            oneq_decompose(k, atol=atol)
-            for k in (
-                self.decomposer.K1l,
-                self.decomposer.K1r,
-                self.decomposer.K2l,
-                self.decomposer.K2r,
-            )
+        self._inner_decomposition = two_qubit_decompose.TwoQubitControlledUDecomposer(
+            rxx_equivalent_gate._standard_gate
```
mtreinish: This is potentially a breaking API change if we treat this class as public. I think we can probably get away with considering it private since as far as I can tell we've never documented it. If we wanted to support arbitrary gates it wouldn't be too hard this is why `TwoQubitGateSequence` has an `Option<StandardGate>` in it. We'd have to pass the other details to get the matrix from the parameter value to rust but then we can keep the arbitrary gate support and when we encounter `None` we use that.

That being said if we're ok (I'm not 100% convinced one way or the other) making this API change to the class because we consider this internal only, we should change this to do something like:

```suggestion
        if (gate := rxx_equivalent_gate.getattr("_standard_gate", None)) is not None:
            self._inner_decomposition = two_qubit_decompose.TwoQubitControlledUDecomposer(
                rxx_equivalent_gate._standard_gate
        else:
            raise QiskitError("Must be initialized with a standard gate object")
```

so we raise a meaningful error if this is instantiated with a non-standard gate.

crates/accelerate/src/two_qubit_decompose.rs
```diff
@@ -2344,6 +2344,352 @@ pub fn local_equivalence(weyl: PyReadonlyArray1<f64>) -> PyResult<[f64; 3]> {
...
+            if (decomp.a * 2.0 - test_angle / scale_a).abs() > atol {
+                return Err(QiskitError::new_err(format!(
+                    "The gate {}
+                  is not equivalent to an RXXGate.",
+                    rxx_equivalent_gate.name()
+                )));
+            }
+
+            scales.push(scale_a);
+        }
```
mtreinish: I would refactor this to be an iterator chain especially since it has a fixed number of iterations this will optimize the allocation and give rust more context for optimizing the loop. Something like:

```suggestion
        let test_angles = [0.2, 0.3, PI2];
        let scales: PyResult<Vec<f64>> = test_angles.into_iter().map(|test_angle| {
            if rxx_equivalent_gate.num_params() != 1 {
                return Err(QiskitError::new_err(
                    "Equivalent gate needs to take exactly 1 angle parameter.",
                ));
            }
            let mat = rxx_equivalent_gate
                .matrix(&[Param::Float(test_angle)])
                .unwrap();
            let decomp =
                TwoQubitWeylDecomposition::new_inner(mat.view(), Some(DEFAULT_FIDELITY), None)?;

            let mat_rxx = StandardGate::RXXGate
                .matrix(&[Param::Float(test_angle)])
                .unwrap();
            let decomposer_rxx = TwoQubitWeylDecomposition::new_inner(
                mat_rxx.view(),
                None,
                Some(Specialization::ControlledEquiv),
            )?;
            let decomposer_equiv = TwoQubitWeylDecomposition::new_inner(
                mat.view(),
                None,
                Some(Specialization::ControlledEquiv),
            )?;
            let scale_a = decomposer_rxx.a / decomposer_equiv.a;

            if (decomp.a * 2.0 - test_angle / scale_a).abs() > atol {
                return Err(QiskitError::new_err(format!(
                    "The gate {}
                  is not equivalent to an RXXGate.",
                    rxx_equivalent_gate.name()
                )));
            }
            Ok(scale_a)
        })
        .collect();
        let scales = scales?;
```

crates/accelerate/src/two_qubit_decompose.rs
```diff
@@ -2344,6 +2344,352 @@ pub fn local_equivalence(weyl: PyReadonlyArray1<f64>) -> PyResult<[f64; 3]> {
...
+        let scale = scales[0];
+
+        // Check that all three tested angles give the same scale
+        for scale_val in scales.clone().into_iter() {
+            if !abs_diff_eq!(scale_val, scale, epsilon = atol) {
+                return Err(QiskitError::new_err(
+                    "Inconsistent scaling parameters in check.",
+                ));
+            }
+        }
```
mtreinish: You can remove the clone here:

```suggestion
        for scale_val in &scales {
            if !abs_diff_eq!(scale_val, &scale, epsilon = atol) {
                return Err(QiskitError::new_err(
                    "Inconsistent scaling parameters in check.",
                ));
            }
        }
```

crates/accelerate/src/two_qubit_decompose.rs
```diff
@@ -2344,6 +2344,352 @@ pub fn local_equivalence(weyl: PyReadonlyArray1<f64>) -> PyResult<[f64; 3]> {
...
+            .matrix(&[Param::Float(self.scale * angle)])
+            .unwrap();
+        let decomposer_inv =
+            TwoQubitWeylDecomposition::new_inner(mat.view(), Some(DEFAULT_FIDELITY), None)?;
+
+        let euler_basis = EulerBasis::ZYZ;
+        let mut target_1q_basis_list = EulerBasisSet::new();
+        target_1q_basis_list.add_basis(euler_basis);
+
+        let mut gates = Vec::new();
```
mtreinish: I'd use `with_capacity()` here instead of `new()` so that we pre-allocate the maximum number of gates in the sequence. 


```suggestion
        let mut gates = Vec::with_capacity(13);
```
I got 13 from 3x gates for each of the 4x 1q components + the 2q gate.

crates/accelerate/src/two_qubit_decompose.rs
```diff
@@ -2344,6 +2344,352 @@ pub fn local_equivalence(weyl: PyReadonlyArray1<f64>) -> PyResult<[f64; 3]> {
...
+        let unitary_c1r =
+            unitary_to_gate_sequence_inner(c1r, &target_1q_basis_list, 0, None, true, None);
+        let unitary_c2r =
+            unitary_to_gate_sequence_inner(c2r, &target_1q_basis_list, 0, None, true, None);
+        let unitary_c1l =
+            unitary_to_gate_sequence_inner(c1l, &target_1q_basis_list, 0, None, true, None);
+        let unitary_c2l =
+            unitary_to_gate_sequence_inner(c2l, &target_1q_basis_list, 0, None, true, None);
+
+        let mut gates = Vec::new();
```
mtreinish: Same comment here about pre-allocating:
```suggestion
        let mut gates = Vec::with_capacity(59);
```
I counted 47 max entries for `weyl_gate()` calls plus 12 (4x 3 gates) for each 1q component.

crates/accelerate/src/two_qubit_decompose.rs
```diff
@@ -2344,6 +2344,352 @@ pub fn local_equivalence(weyl: PyReadonlyArray1<f64>) -> PyResult<[f64; 3]> {
...
+                for gate in circ_c.gates.into_iter().rev() {
+                    let (inv_gate_name, inv_gate_params, inv_gate_qubits) = invert_2q_gate(gate);
+                    circ.gates
+                        .push((Some(inv_gate_name), inv_gate_params, inv_gate_qubits));
+                }
+                circ.gates
+                    .push((Some(StandardGate::HGate), smallvec![], smallvec![0]));
+                circ.gates
+                    .push((Some(StandardGate::HGate), smallvec![], smallvec![1]));
+            }
```
mtreinish: I would be slightly inclined to simplify the logic here to remove the `invert` variable and only have one if statement on `gamma > 0.0`.

mtreinish (commented):
Overall the code looks good, I left a few inline comments on how to simplify or optimize things a bit. The biggest concern I have right now though is the backwards incompatibility of the API. This PR restricts the gate objects that `TwoQubitControlledUDecomposer` class can work with to just standard gates defined in Qiskit. This wasn't a limitation before the rust rewrite and would be considered a breaking api change. However, I'm not sure the class is considered public, it's never been externally documented from what I can tell.  I'm curious of your thoughts on this though because I'm not sure what you're thinking about making this class public or not.

qiskit/synthesis/two_qubit/two_qubit_decompose.py
ShellyGarion: done in [e532c8a](https://github.com/Qiskit/qiskit/pull/13139/commits/e532c8a9e1df7dff37331b6413857c1f53a4ce36)

qiskit/synthesis/two_qubit/two_qubit_decompose.py
ShellyGarion: done in [3182838](https://github.com/Qiskit/qiskit/pull/13139/commits/31828388b0808cffd2bb7def11b70382fc50e595)
I had to change the if line to 
`if rxx_equivalent_gate._standard_gate is not None:`
otherwise the tests were failing.

crates/accelerate/src/two_qubit_decompose.rs
ShellyGarion: Currently, `rxx_equivalent_gate` could be only a `StandardGate` since in the code we need to invert it, and there is currently no inverse function for other types, see these lines:
```
                let circ_c = self.to_rxx_gate(gamma)?;
                ...
                for gate in circ_c.gates.into_iter().rev() {
                    let (inv_gate_name, inv_gate_params, inv_gate_qubits) = invert_2q_gate(gate);
```

ShellyGarion: As for making this class public, I think that it may be useful. Should we do it now or only after it's added to the `UnitarySynthesis` transpiler pass #13320?

crates/accelerate/src/two_qubit_decompose.rs
ShellyGarion: Could you please explain how to convert `TwoQubitGateSequence` to `CircuitData` ?

crates/accelerate/src/two_qubit_decompose.rs
ShellyGarion: done in [59ef018](https://github.com/Qiskit/qiskit/pull/13139/commits/59ef018f29c107d5c7cadbfa2442e5c0e94a8742)

crates/accelerate/src/two_qubit_decompose.rs
ShellyGarion: done in [59ef018](https://github.com/Qiskit/qiskit/pull/13139/commits/59ef018f29c107d5c7cadbfa2442e5c0e94a8742)

crates/accelerate/src/two_qubit_decompose.rs
ShellyGarion: Done in [1ceb94e](https://github.com/Qiskit/qiskit/pull/13139/commits/1ceb94e13e9f3abee67736b19017dc3063fdffb2)

crates/accelerate/src/two_qubit_decompose.rs
ShellyGarion: Done in [1ceb94e](https://github.com/Qiskit/qiskit/pull/13139/commits/1ceb94e13e9f3abee67736b19017dc3063fdffb2)

crates/accelerate/src/two_qubit_decompose.rs
ShellyGarion: Done in [9ac68b1](https://github.com/Qiskit/qiskit/pull/13139/commits/9ac68b1d1ead2d8ae18be96732ecf2b6b2af638d)

crates/accelerate/src/two_qubit_decompose.rs
ShellyGarion: Done in [ebc0e73](https://github.com/Qiskit/qiskit/pull/13139/commits/ebc0e73ebecd19e88dc3dc4969298f09227c9441)

ShellyGarion:
> Overall the code looks good, I left a few inline comments on how to simplify or optimize things a bit. The biggest concern I have right now though is the backwards incompatibility of the API. This PR restricts the gate objects that `TwoQubitControlledUDecomposer` class can work with to just standard gates defined in Qiskit. This wasn't a limitation before the rust rewrite and would be considered a breaking api change. However, I'm not sure the class is considered public, it's never been externally documented from what I can tell. I'm curious of your thoughts on this though because I'm not sure what you're thinking about making this class public or not.

I would appreciate @levbishop opinion regarding this question as he wrote the original Python code. 
I agree that as far as I know this class has never been published as part of Qiskit API. We can make it public if we think it will be useful. perhaps after adding it to the `UnitarySynthesis` transpiler pass https://github.com/Qiskit/qiskit/issues/13320.
In addition, in the tests we test only standard gates:
https://github.com/Qiskit/qiskit/blob/fbfe7388f3efdaaeb70dc96aa35f649c1897c5d4/test/python/synthesis/test_synthesis.py#L1429
So I wonder if this class was ever meant to be used with basic gates that are not standard gates.
Anyway, since the Pulse library is being deprecated in https://github.com/Qiskit/qiskit/pull/13164, the users will not be able to calibrate their own basic gates with Qiskit and hence this code will be useful only for standard gates.

crates/accelerate/src/two_qubit_decompose.rs
ElePT: @ShellyGarion you can iterate over the gates in the sequence and push them to a new `CircuitData` object, there is an example in `TwoQubitWeylDecomposition`: https://github.com/Qiskit/qiskit/blob/main/crates/accelerate/src/two_qubit_decompose.rs#L1151-L1170

ShellyGarion: thanks! how to convert `gate.2` which is of type `SmallVec<[u8; 2]>` into `Qubit` ?

ShellyGarion: OK, I found a solution, in [02cd73e](https://github.com/Qiskit/qiskit/pull/13139/commits/02cd73ed85dfa227892ea0c26003f478c8dbf308)

qiskit/synthesis/two_qubit/two_qubit_decompose.py
mtreinish: I restored the capability to use custom gate classes in: https://github.com/Qiskit/qiskit/pull/13139/commits/f46edbee26053d555d21283b9046b45207f863a5 It depends on calling back to python to instantiate a custom gate class but with this we don't have any concerns with any potential API change.

crates/accelerate/src/two_qubit_decompose.rs
mtreinish: I'd say lets do it later as the PR is large and intricate enough that we should review the change making it public on its own.

test/python/synthesis/test_synthesis.py
```diff
@@ -1440,6 +1440,24 @@ def test_not_rxx_equivalent(self):
             "Equivalent gate needs to take exactly 1 angle parameter.", exc.exception.message
         )
 
+    @combine(seed=range(10), name="seed_{seed}")
+    def test_correct_unitary_custom_gate(self, seed):
+        """Test synthesis with a custom controlled u equivalent gate."""
+        unitary = random_unitary(4, seed=seed)
+
+        class CustomXXGate(RXXGate):
+            """Custom RXXGate subclass that's not a standard gate"""
```
ShellyGarion, line 1453: This gate is basically the same as `RXXGate`, right?
Perhaps we can construct an `rxx_equivalent_gate` which is not a `StandardGate`?

mtreinish: Yeah, I just used an `RXXGate` for simplicity since I knew it would work. It is going through the python gate path though since this is not a standard gate, but all the math works out the same as an RXXGate. I think you're right we should have some coverage of something not equivalent to a standard gate. If you have a matrix in mind we can change this to use that instead of subclassing RXXGate.

ShellyGarion: Perhaps we can take the same matrix as `RXXGate` (or any of the other `StandardGate`s and add some constant global phase?

ShellyGarion: Also in line 1429 we can add to the list of gates:
`     for gate in [RXXGate, RYYGate, RZZGate, RZXGate, CPhaseGate, CRZGate]:`
the gates `CRXGate` and `CRYGate` .

ShellyGarion: Another possibility is this parametric iSWAP gate taken from https://threeplusone.com/pubs/on_gates.pdf (Section 6.6):

image

ShellyGarion: Also in line 1429 we can add to the list of gates:
`     for gate in [RXXGate, RYYGate, RZZGate, RZXGate, CPhaseGate, CRZGate]:`
the standard gates `CRXGate` and `CRYGate` .

done in https://github.com/Qiskit/qiskit/pull/13139/commits/ee3e88ec951598d410eaee8de8975aa8cad3d700

ShellyGarion: it seems that the XY gate suggested above is not an rxx-equivalent gate. I'll add a test that it raises an exception.
done in [1bb0616](https://github.com/Qiskit/qiskit/pull/13139/commits/1bb06160ee2fccbafe97a6a9d1ef282445b7a5de) after the fix in [06c3da4](https://github.com/Qiskit/qiskit/pull/13139/commits/06c3da4c6e44e5c19ed85d78575961b2809c1c5e)

test/python/synthesis/test_synthesis.py
```diff
@@ -1460,6 +1462,35 @@ def __init__(self, theta, label=None):
...
+                    [[1, 0, 0, 0], [0, cos, -isin, 0], [0, -isin, cos, 0], [0, 0, 0, 1]],
+                    dtype=dtype,
+                )
+
+            def inverse(self, annotated: bool = False):
+                return CustomXYGate(-self.params[0])
+
+        with self.assertRaises(QiskitError) as exc:
+            _ = TwoQubitControlledUDecomposer(CustomXYGate)
+        self.assertIn("Specialization: ControlledEquiv calculated fidelity", exc.exception.message)
```
mtreinish: FWIW, you can do this with `self.assertRaisesRegex` https://docs.python.org/3/library/unittest.html#unittest.TestCase.assertRaisesRegex by specifying a regular expression that matches the target error message. This would be something like:

```suggestion
        with self.assertRaisesRegex(QiskitError, "ControlledEquiv calculated fidelity"):
            TwoQubitControlledUDecomposer(CustomXYGate)
```

but you can get more sophisticated with the regular expression to atch more exactly on the error message if you want.

mtreinish (commented):
This LGTM now. I have one small suggestion on the test code but once that's fixed I think we're good to merge.

test/python/synthesis/test_synthesis.py
ShellyGarion: done in [8949f7f](https://github.com/Qiskit/qiskit/pull/13139/commits/8949f7f4c3a0c02318f368f82eea3d901aa5ac26)

mtreinish (approved):

PR #13140, opened 2024-09-12
Added new way to assign parameters to pulse schedule,

arthurostrauss:
### Summary

It is now possible to assign pulse parameters to a pulse schedule using a sequence
    of parameter names or parameters as keys to the value_dict argument of the `assign_parameters` method.
    This allows for more flexible assignment of parameters to pulse schedules. Notably, it is possible to directly
    assign parameters through slices of a `BindingsArray` class, used in the primitives interface for assigning
    parameters. This could be useful for future implementations custom pulse-based primitives implementations (e.g.
    Qiskit Dynamics custom primitive implementation).
    Moreover, we have a new class attribute for `Schedule` and `ScheduleBlock` enabling the disabling of parameter
    validation when assigning parameters to a schedule. This is useful for advanced users who want to integrate Qiskit
    Pulse schedule creations within a JIT compiled function (notably to accelerate simulations in Qiskit Dynamics).

### DIFF
--- a/qiskit/pulse/parameter_manager.py
+++ b/qiskit/pulse/parameter_manager.py
@@ -331,6 +331,8 @@ class ParameterManager:
     Instruction data and its location are not directly associated with this object.
     """
 
+    disable_parameter_validation = False
+
     def __init__(self):
         """Create new parameter table for pulse programs."""
         self._parameters = set()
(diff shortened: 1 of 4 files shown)


1ucian0:
Hi @arthurostrauss,

Heads up, we are on the way to [deprecate pulses](https://github.com/Qiskit/qiskit/pull/13164), so new features on those modules have very low priority.

jakelishman:
Closing as stale / obsolete since pulse was removed

PR #13143, opened 2024-09-12
Add rust-native `apply_operation` methods

### LINKED ISSUE #13134: Add methods to more easily add operations to a `DAGCircuit` from Rust
## The problem:
We crurrently have to methods that allow us to add an operation to the `DAGCircuit` from rust.
- `DAGCircuit::push_back` and `DAGCircuit::push_front`  which accepts a `PackedInstruction` instance as an argument and must have valid interned qargs and cargs.
- `DAGCircuit::py_apply_operation_back` and `DAGCircuit::py_apply_operation_front` which performs the addition of instructions in a similar way to what was done before, however this is only with python types and is not exposed to the rest of the rust api.

In each case we're either required to re-intern the qargs/cargs, or have an already valid instance of `PackedInstruction` that could work in this circuit, which is what I ultimately did for #13032. But now that we are building transpiler passes in Rust alone, having to own mutable access to the `DAGCircuit` and inserting qargs directly seems very unsafe.

## The solution
**Adding rust-native `apply_operation_back` and `apply_operation_front` to `DAGCircuit` that is exposed to the rest of the crates.**

Having said methods would allow us to:
- Insert a `PackedOperation` of any type and insert qargs and cargs by index (`Qubit` or `Clbit` as defined by the `circuit` crate), since that's how it is registered in the `DAGCircuit`.
- Avoid having mutable access to private methods in the `DAGCircuit`, such as the interners or `BitData`.
- Overall make it easier to modify and create `DAGCircuit` instances from transpiler passes that need it, such as the `BasisTranslator`.

This shouldn't be too hard to implement and would be of great help to all of the developers currently working on transpiler passes.

## Discussion
Feel free to start an educated discussion in the comments, I'd love to hear what the rest of the team thinks about this.
```[tasklist]
### Tasks
- [ ] https://github.com/Qiskit/qiskit/pull/13127
- [ ] https://github.com/Qiskit/qiskit/pull/13143
```

raynelfss:
Fixes #13134 
### Summary
The following commits add `apply_operation_front` and `apply_operation_back` as methods with rust-native arguments, and use them in the existent Python variants of these functions.

### Details and comments
With the DAGCircuit living completely in Rust as of now, we need to continue expanding its functionality across both environments (Python and Rust). As we continue to work on transpiler passes, we need methods that allow us to add operations using rust-native datatypes instead of using Python.

These commits focus on adding two methods:
- `apply_operation_front` to apply an operation to the front of the circuit.
- `apply_operation_back` which applies the operation to the back of the circuit.

These methods allow rust users to use native data types and add instructions to the circuit without needing to have valid interned indices for the Cargs/Qargs of said operations, as the current methods `push_front` and `push_back` do currently. This will decidedly enable us to build circuits more easily from Rust as transpiler passes, such as the `BasisTranslator`, need to be able to do.

### Blockers
- [x] #13127


crates/circuit/src/dag_circuit.rs
```diff
@@ -1681,28 +1682,25 @@ def _format(operand):
         let py_op = op.extract::<OperationFromPython>()?;
         let qargs = qargs.map(|q| q.value);
         let cargs = cargs.map(|c| c.value);
-        let node = {
```
kevinhartman: We may want to avoid using the Rust `apply_operation_{back, front}` from within these Python versions, and keep using the internal `push_back` for them instead.

The primary reason being that the `qarg_indices` and `carg_indices` we have on lines 1687 and 1689 are _owned_. What we had before is a bit more efficient, since it moves those owned values directly into the `PackedInstruction`, and then moves that into `DAGCircuit::push_back`.

(FWIW, the new code here looks just about the same length as what we had before, so I don't think we'd be making much of a concession in terms of code simplicity.)

crates/circuit/src/dag_circuit.rs
```diff
@@ -5263,6 +5257,123 @@ impl DAGCircuit {
...
+        &mut self,
+        py: Python,
+        op: PackedOperation,
+        qargs: &[Qubit],
+        cargs: &[Clbit],
+        params: Option<SmallVec<[Param; 3]>>,
+        extra_attrs: Option<Box<ExtraInstructionAttributes>>,
+        #[cfg(feature = "cache_pygates")] py_op: Option<PyObject>,
+        check: bool,
+        front: bool,
```
kevinhartman, line 5255: IMO, it'd make more sense to inline the contents of this method to the Rust `apply_operation_{front,back}` methods as well.

The most complex bit looks to be the bounds checking for the bits, which you should be able to simplify (see comment below). And, instead of passing `check` and `front` as a booleans here, we can avoid some indirection by leaving it up to the caller to actually do the checks and push it onto the DAG. This would save a conditional check, since the caller already knows whether to push onto the front or back. 

With all that gone, there's not much else going on here in terms of code de-duplication (there's no getting around the wad of parameters we need to pass through up to where we make the final `PackedInstruction`).

crates/circuit/src/dag_circuit.rs
```diff
@@ -5263,6 +5257,123 @@ impl DAGCircuit {
...
+        qargs.iter().try_for_each(|qarg| {
+            if !(0..self.num_qubits() as u32).contains(&qarg.0) {
+                return Err(PyValueError::new_err(format!(
+                    "Qubit index {} is out of range. This DAGCircuit currently has only {} qubits.",
+                    qarg.0,
+                    self.num_qubits()
+                )));
+            }
+            Ok(())
+        })?;
```
kevinhartman: As written, these checks iterate over all the bits for every arg, which is quadratic.

You should be able to get away with just converting each arg to a `usize` and making sure it is less than `num_{qu,cl}bits()`.

crates/circuit/src/dag_circuit.rs
raynelfss: I think this can be fixed easily considering that in most cases all the qargs/cargs will be owned as the user might construct this collection themselves. If we accepted an item that complies with the trait bounds `IntoIterator<Item = Qubit/Clbit>`, which would not discriminate the collection type as long as it has a `Bit` iterator type, it would technically achieve the same avoiding cloning the values manually.

However, I can also visualize how the user might instead want to re-use the Qargs/Cargs for something else down the line after insertion, in which case they would have to perform the copies manually. I say copies here since we've defined `Qargs/Cargs` as items that implement the `Copy` trait, which means that moving them should not be as costly.

The plan was to re-use the same methods so that they'd be tested without us having to build any new tests for them. But if a suggestion by one of your following comments (which asks to remove the insertion check from these rust native methods) is applied, we would still be required to build the PackedInstruction in this step which would require undoing the changes made.

kevinhartman: > I think this can be fixed easily considering that in most cases all the qargs/cargs will be owned as the user might construct this collection themselves. If we accepted an item that complies with the trait bounds IntoIterator<Item = Qubit/Clbit>, which would not discriminate the collection type as long as it has a Bit iterator type, it would technically achieve the same avoiding cloning the values manually.

I presume you're referring to the signature of the `DAGCircuit::apply_operation_back`, correct? The main issue with accepting a generic `IntoIterator` for `qargs` and `cargs` is that while it'd _allow_ an owned `Vec<{Qu|Cl}bit>` to be accepted, it wouldn't be possible to get it into the `Interner` without doing an allocation, since it is _more generic_ than the `Interner`'s `T::Owned` (which is exactly `Vec<T>`).

So you'd still need to collect those into `Vec<T>`, which would do an allocation.

For most use cases, it makes sense for the Rust-facing `DAGCircuit::apply_operation_back` to accept the bits as `&[{Qu,Cl}bit]`s (as you have it now) because more often than not, the `Interner` will get a cache hit during lookup, and it won't actually need to allocate. For `DAGCircuit::py_apply_operation_back` however, we already have an owned `Vec` from the Python conversion stuff, so we might as well ensure that we avoid the slim chance of that allocation happening by using the `Interner::insert_owned` methods directly. Hence why I think we ought to leave `DAGCircuit::py_apply_operation_{front,back}` as they are for now.

As @jakelishman mentioned, we might want to have some variants of `DAGCircuit::apply_operation_back` in the future that support scenarios where a Rust user already owns the bit lists, or where they pass `Interned` keys. If we do find a need for such methods, then perhaps at that point it makes sense to have `DAGCircuit::py_apply_operation_{front,back}` call those internally. But for now, `DAGCircuit::push_{front,back}` are already reasonable primitives for us to be using IMO in terms of code reuse.

> The plan was to re-use the same methods so that they'd be tested without us having to build any new tests for them. 

I like the pragmatics of this, but really we need to get Rust-based unit tests going for `DAGCircuit` sooner than later anyways. In the meantime, other Qiskit passes should exercise these at least in an integration testing capacity.

kevinhartman (approved):
This looks good to me. Thanks for doing this work and for the changes, @raynelfss!

PR #13147, opened 2024-09-13
Reformat `pyext` to avoid merge conflicts

jakelishman:
### Summary

Since we're still in a phase of adding many new extension modules to `pyext`, the current structure of the file is causing problems with most commits hitting merge conflicts in the `use` block.  This changes the style of the file so that each new module takes only a single line, which is more friendly to avoiding merge conflicts, and fixing them when they arise.

### Details and comments

I don't feel super strongly about this, just it's been a slight annoyance I've seen a few people have, and I just did too.

### DIFF
--- a/crates/pyext/src/lib.rs
+++ b/crates/pyext/src/lib.rs
@@ -12,23 +12,6 @@
 
 use pyo3::prelude::*;
 
-use qiskit_accelerate::{
-    check_map::check_map_mod, circuit_library::circuit_library,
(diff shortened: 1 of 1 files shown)


ElePT (approved):
As a merge conflict victim, I'd be in favor of the change :)

jakelishman:
Huh, that lint failure is a true lint failure - that function _doesn't_ exist, but I'm curious why lint didn't fail on #12959, which is what actually removed it.  I'll fix it in a separate PR.

jakelishman:
Ok, rebased and now this should pass lint.

PR #13148, opened 2024-09-13
Re-allow non-standard includes in OpenQASM 3 exporter

### LINKED ISSUE #13061: Using includes with the QASM3 Exporter raises instead of including definitions
### Environment

- **Qiskit version**: 1.2
- **Python version**: 3.11
- **Operating system**: Sonoma 14.6.1 (23G93) aarch64

### What is happening?

When custom includes are passed into the `qiskit/qasm3/exporter::Exporter.__init__` function, the export of QASM will always throw an error when the `dump` function is called via the `QASM3Builder`. The [docs say](https://github.com/Qiskit/qiskit/blob/stable/1.2/qiskit/qasm3/exporter.py#L143-L145):

```
includes: the filenames that should be emitted as includes.  These files will be parsed
    for gates, and any objects dumped from this exporter will use those definitions
    where possible.
```
This worked fine in Qiskit 1.1.x but 1.2 [changed the behavior](https://github.com/Qiskit/qiskit/blob/stable/1.2/qiskit/qasm3/exporter.py#L675-L682) such that any includes passed in that aren't exactly `stdgates.inc` will raise `QASM3ExporterError`. This also removed the `GlobalNamespace` type breaking consumers.

In 1.1.x the definitions from the custom includes were treated as opaque assuming any missing definitions would have been provided in the basis gates passed into the `Exporter`.

```python
from qiskit.qasm3 import Exporter, QASM3ExporterError

try:
    qc = QuantumCircuit(1, 1)
    includes = ("stdgates.inc", "custom_gates.inc")
    Exporter(includes=includes).dumps(qc)
except QASM3ExporterError as ex:
    print(ex)
```
```shell
"Unknown OpenQASM 3 include file: 'custom_gates.inc'"
```

### What should happen?

The documentation indicates that the includes will be parsed for gate definitions. The docs have not changed since 1.1.x so the docs have never matched the implementation. I think there are two primary paths which can be taken:
- Restore the old functionality and update the documentation. This would allow custom gate definitions in QASM files.
  - The `basis_gates` can be used to define opaque gate names which would be found in the external QASM files.
- Update the `Exporter`/`QASM3Builder` to match the docs. This will require updating the `crates/qasm3/build.rs` as it currently noops gate definitions when parsing the ASG.
  - How would this reconcile parsed definitions vs the `basis_gates` parameter?

These solutions would not include any files included from the supplied includes list. Instead, the supplied includes are listed at the top of the generated QASM file as was done in 1.1.x.

### Any suggestions?

_No response_

jakelishman:
### Summary

Swapping the behaviour of the exporter to error on unknown includes was a mistake, and a breaking change from previous versions of the OpenQASM 3 exporter, since the `basis_gates` argument can be used to emulate the expected behaviour of this option, even though the previously documented behaviour was never actually fulfilled.

This restores the previous non-erroring behaviour, and corrects the documentation to note how the ``includes`` argument should be used in the near term.

### Details and comments

Fix #13061

### DIFF
--- a/qiskit/qasm3/exporter.py
+++ b/qiskit/qasm3/exporter.py
@@ -140,9 +140,16 @@ class Exporter:
     ):
         """
         Args:
-            includes: the filenames that should be emitted as includes.  These files will be parsed
-                for gates, and any objects dumped from this exporter will use those definitions
-                where possible.
+            includes: the filenames that should be emitted as includes.
+
+                .. note::
+
+                    At present, only the standard-library file ``stdgates.inc`` is properly
+                    understood by the exporter, in the sense that it knows the gates it defines.
+                    You can specify other includes, but you will need to pass the names of the gates
+                    they define in the ``basis_gates`` argument to avoid the exporter outputting a
+                    separate ``gate`` definition.
+
             basis_gates: the basic defined gate set of the backend.
             disable_constants: if ``True``, always emit floating-point constants for numeric
                 parameter values.  If ``False`` (the default), then values close to multiples of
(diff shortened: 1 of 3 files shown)


kevinhartman (approved):
LGTM, thanks!

Issue #13046, opened 2024-08-27
Circuit library modernization


Cryoris:
### Summary

Modernize the circuit library to improve compiler quality and construction speed. The library circuits are classified in two categories:
* **Abstract blocks** Operations defined by an action they perform (e.g. mathematical transformation), but can be synthesized in different fashions. These will be `Instruction/Gate` objects with a detached synthesis, that the compiler can reason about. 
Examples: QFT, Pauli evolutions, arithmetic circuits, MCX gates.
* **Structural circuits** Circuit defined by their structure (the circuit is unique and cannot be built another way). These will be Python functions returning a `QuantumCircuit`.
Examples: n-local circuits, quantum volume

Changes are to be pending-deprecated for 1.3, deprecated for 1.4 and finalized for 2.0.

### Examples

The move to abstract blocks allows the compiler to reason about the circuit building blocks, where previously there was an eager construction:
```python
circuit = ...
circuit.append(MCXGate(n), range(n - 1), [n])  # compiler can now allocate auxiliary qubits for efficient synthesis
circuit.append(QFTGate(2 * n), range(2 * n))  # compiler can choose optimal method based on topology or following measurements
circuit.measure_all()
```

Python functions can be used as drop-in replacement, 
```python
circuit = two_local(num_qubits, ["rx", "ry"], "cx", entanglement="pairwise")
```
but can enable performance-improvements, e.g. by directly composing onto output circuits.

```[tasklist]
#### Abstract blocks
- [x] HLS with ancillas
- [x] `MCXGate`: synthesize via HLS & pending-deprecate MCX gate classes other than `MCXGate` (#12961)
- [x] `MCMTGate`: synthesize with HLS synthesis & pending-deprecate `MCMT` and `MCMTVChain` (#13150)
- [x] Pending deprecate circuits with existing gates: `QFT`, `Permutation`, `Diagonal`, `GMS` (@Cryoris) (#13232)
- [x] `PhaseOracle`: additionally needs a tweedledum-free synthesis method (@gadial  #13769)
- [x] `library.boolean_logic`: straightforward porting to `Gate` (@alexanderivrii) #13333
- [x] `library.arithmetic` (partially done by #13354)
- [x] `GraphState` (#13353)
- [x] Functional Pauli Rotations (#13371)
```

```[tasklist]
#### Structural circuits
- [x] `n_local` & friends @Cryoris (#13310)
- [x] `pauli_feature_map` & friends  (#13045)
[...]

Issue #10698, opened 2023-08-23
add ctrl_state to MCMT gate


Muzhou-Ma:
### What should we add?

Add ctrl_state to multi-control multi-target gates

diemilio:
I think this would be a nice feature to add. Willing to help implement it if looking for help.

Dpbm:
This is awesome, for any help, I'm here too!

ShashiQubit:
Hi! I am interested in working on this issue. Can this be assigned to me? .I would like to make my first contribution to qiskit terra.

Dpbm:
Yeah man, if @diemilio and @Marsmmz also agree, you can do this! And if during your task you find some issues, please tell us and we may help you😁

diemilio:
Sure! I had volunteered to take care of this a little while back cause I had the time, but I never got it assigned.

ShashiQubit:
@diemilio Thanks for the opportunity. I have started working on the issue following the contribution guide.

Muzhou-Ma:
Sorry for getting here late, I'll try my best to help!

ShashiQubit:
@Marsmmz I have a question. Does MCMT controls need to be custom state?. Meaning should user given freedom to specify whether the state of control is 0 or 1 or mix like '01...' control state targeting q_k number of qubits. If yes what's best way to take that input from user, dictionary or list??

Muzhou-Ma:
> @Marsmmz I have a question. Does MCMT controls need to be custom state?. Meaning should user given freedom to specify whether the state of control is 0 or 1 or mix like '01...' control state targeting q_k number of qubits. If yes what's best way to take that input from user, dictionary or list??

I personally think users should have the freedom to specify the control state. Like if a 2-C-2-T gate is needed, users can specify which of '00', '01', '10', '11' they want to be the control state.  Perhaps taking a 01 binary string as input is enough? What do you think?@ShashiQubit @diemilio @Dpbm

ShashiQubit:
@Marsmmz @diemilio @Dpbm I want feed back on this modified implementation of control function from source code of MCMT gate:
``` 
def control(self, num_ctrl_qubits=1, label=None, ctrl_state=None):
        """Return the controlled version of the MCMT circuit."""
        mcmt_circuit = QuantumCircuit(self.num_ctrl_qubits + num_target_qubits)
        if isinstance(ctrl_state, str):
[...]

PR #13150, opened 2024-09-13
`MCMTGate` & plugins

Cryoris:
### Summary

Make the MCMT circuits a gate with pluggable synthesis method, plus port the V-chain synthesis to Rust. This is part of #13046.

### Details and comments

Even though the focus of this PR is the restructuring into a block that the compiler can reason about, there's also a nice speedup due to the port to Rust:

<details closed>
<summary>New runtime / Old runtime = 0.63 (geo mean)</summary>

```
-- n = 50 ntarg = 1
main: 5.604e-04 +- 1.514e-04
this: 3.490e-04 +- 1.771e-04
fraction: 0.64

-- n = 50 targ = 5
main: 5.612e-04 +- 1.180e-04
this: 3.051e-04 +- 1.267e-04
fraction: 0.54

-- n = 50 targ = 25
main: 7.027e-04 +- 1.188e-04
this: 3.727e-04 +- 1.225e-04
fraction: 0.53

-- n = 200 targ = 1
main: 2.074e-03 +- 2.704e-04
this: 1.631e-03 +- 2.565e-03
fraction: 0.78

-- n = 200 targ = 20
main: 2.693e-03 +- 2.488e-03
this: 1.646e-03 +- 2.433e-03
fraction: 0.62

-- n = 200 targ = 100
main: 2.616e-03 +- 2.670e-04
this: 1.880e-03 +- 2.594e-03
fraction: 0.71

-- n = 1000 targ = 1
main: 1.087e-02 +- 4.227e-03
this: 6.820e-03 +- 5.478e-03
fraction: 0.62

-- n = 1000 targ = 100
main: 1.251e-02 +- 6.234e-03
this: 7.843e-03 +- 6.380e-03
fraction: 0.62

-- n = 1000 targ = 500
main: 1.415e-02 +- 5.152e-03
this: 9.360e-03 +- 6.755e-03
fraction: 0.66

```
</details>


ShellyGarion:
thanks @Cryoris for opening this PR! It also handles issue #12863.
What do you think about issue  https://github.com/Qiskit/qiskit/issues/10698 ?
I think it should be a small addition, but could also be handled in a follow-up PR.

qiskit/circuit/library/generalized_gates/mcmt.py
```diff
@@ -191,66 +134,150 @@ class MCMTVChain(MCMT):
...
+        ┤0     ├
+        │  2-H │
+        ┤1     ├
+        └──────┘
+
+    This default implementations requires no ancilla qubits, by broadcasting the target gate
+    to the number of target qubits and using Qiskit's generic control routine to control the
+    broadcasted target on the control qubits. If ancilla qubits are available, a more efficient
+    variant using the so-called V-chain decomposition can be used. This is implemented in
+    :class:`~qiskit.circuit.library.MCMTVChain`.
```
alexanderivrii: This part of the docstring probably needs to be updated.

qiskit/circuit/library/generalized_gates/mcmt.py
```diff
@@ -14,41 +14,20 @@
...
-        │  2-H │
-        ┤1     ├
-        └──────┘
-
-    This default implementations requires no ancilla qubits, by broadcasting the target gate
-    to the number of target qubits and using Qiskit's generic control routine to control the
-    broadcasted target on the control qubits. If ancilla qubits are available, a more efficient
-    variant using the so-called V-chain decomposition can be used. This is implemented in
-    :class:`~qiskit.circuit.library.MCMTVChain`.
-    """
```
alexanderivrii: I think it would be nice to keep the description till the `MCMT` circuit is fully deprecated.

qiskit/circuit/library/generalized_gates/mcmt.py
```diff
@@ -191,66 +134,150 @@ class MCMTVChain(MCMT):
...
+
+        self.definition = MCMTDefault().run(self)
+
+    @staticmethod
+    def _identify_base_gate(gate):
+        """Get the control base gate. Note this must be a single qubit gate."""
+
+        # TODO deprecate
+        if callable(gate):
+            gate = gate.__name__
```
alexanderivrii: Is this something you are planning to do for this PR?

qiskit/circuit/library/generalized_gates/mcmt.py
```diff
@@ -191,66 +134,150 @@ class MCMTVChain(MCMT):
...
 
-        if len(ancilla_qubits) < len(control_qubits) - 1:
-            raise QiskitError("Insufficient number of ancilla qubits.")
+        # extract the base gate
+        if isinstance(gate, ControlledGate):
+            base_gate = gate.base_gate
+        elif isinstance(gate, Gate):
+            base_gate = gate
+        else:
+            raise TypeError(f"Invalid gate type {type(gate)}.")
```
alexanderivrii: Would it make sense to consider a control-annotated operation as well?

qiskit/circuit/library/generalized_gates/mcmt.py
```diff
@@ -191,66 +134,150 @@ class MCMTVChain(MCMT):
...
+                This must be a single-qubit gate or a controlled single-qubit gate.
+            num_ctrl_qubits: The number of control qubits.
+            num_target_qubits: The number of target qubits.
+            ctrl_state: The control state of the control qubits. Defaults to all closed controls.
+            label: The gate label.
         """
-        if len(ancilla_qubits) == 0:
-            return
+        if num_target_qubits < 1:
+            raise ValueError("Need at least one target qubit.")
```
alexanderivrii, line 230: Do we allow MCMT gates with no controls?

qiskit/synthesis/multi_controlled/mcmt_vchain.py
```diff
@@ -0,0 +1,49 @@
...
+               │    │  ┌───┐       │    │
+        q_3: ──┼────┼──┤ H ├───────┼────┼──
+               │    │  └─┬─┘┌───┐  │    │
+        q_4: ──┼────┼────┼──┤ H ├──┼────┼──
+             ┌─┴─┐  │    │  └─┬─┘  │  ┌─┴─┐
+        q_5: ┤ X ├──■────┼────┼────■──┤ X ├
+             └───┘┌─┴─┐  │    │  ┌─┴─┐└───┘
+        q_6: ─────┤ X ├──■────■──┤ X ├─────
+                  └───┘          └───┘
+
```
alexanderivrii, line 44: A side question (not fully relevant for this PR): the definition of the MCMT gate requires that both target qubits correspond to the same single-qubit gate (e.g. H in the example above). Would the construct remain correct if the single-qubit gates were different (e.g. H on one target qubit and S on another target qubit)? 

In fact, if we were to compose two copies of the construction above 
```(CCX-chain -- U -- inverse-CCX-chain) -- (CCX-chain -- V -- inverse-CCX-chain)```,
then we would get
```(CCX-chain -- U  -- V -- inverse-CCX-chain)```, so the only benefit of defining an MCMT-gate vs. several regular multi-controlled gates is that the synthesis would necessarily create the same chain of CCX gates and allow cancellation. Am I correct about this?

File: crates/accelerate/src/synthesis/multi_controlled/mcmt.rs [hunk 7]
```diff
@@ -0,0 +1,181 @@
...
+        PackedOperation,
+        SmallVec<[Param; 3]>,
+        Vec<Qubit>,
+        Vec<Clbit>,
+    ),
+> + 'a {
+    let n = controls.len() - 1; // number of chain elements
+    let indices = std::iter::once((controls[0], controls[1], auxiliaries[0]))
+        .chain((0..n - 1).map(|i| (controls[i + 2], auxiliaries[i], auxiliaries[i + 1])));
+    indices.map(|(ctrl1, ctrl2, target)| {
```
alexanderivrii: Do we even need to create the variable `indices`, i.e. have something like this `std::iter::once(...).chain(...).map(...)`?

crates/accelerate/src/synthesis/multi_controlled/mcmt.rs
```diff
@@ -0,0 +1,181 @@
...
+    num_ctrl_qubits: usize,
+    num_target_qubits: usize,
+    control_state: Option<usize>,
+) -> PyResult<CircuitData> {
+    let packed_controlled_gate = controlled_gate.operation;
+    let num_qubits = if num_ctrl_qubits > 1 {
+        2 * num_ctrl_qubits - 1 + num_target_qubits
+    } else {
+        1 + num_target_qubits // we can have 1 control and multiple targets
+    };
```
alexanderivrii, line 115: So here it seems that 0-controlled MCMT-gates are not allowed. Is this handled somewhere?

crates/accelerate/src/synthesis/multi_controlled/mcmt.rs
```diff
@@ -0,0 +1,181 @@
...
+
+    let control_state = control_state
+        .or(Some(usize::pow(2, num_ctrl_qubits as u32) - 1))
+        .unwrap();
```
alexanderivrii: I think you can use `unwrap_or` instead.

crates/accelerate/src/synthesis/multi_controlled/mcmt.rs
```diff
@@ -0,0 +1,181 @@
...
+        .map(|index| {
+            (
+                PackedOperation::from_standard(StandardGate::XGate),
+                smallvec![] as SmallVec<[Param; 3]>,
+                vec![Qubit(index as u32)],
+                vec![] as Vec<Clbit>,
+            )
+        });
+
+    // First, we create the operations that apply the controlled base gate.
```
alexanderivrii: This can no longer be "first" (my internal comment-reader-counter raises an error) 😛

alexanderivrii (commented):
Thanks @Cryoris, this looks very nice. I have left a few minor comments and questions.

alexanderivrii:
Lol, the tests have failed due to `cargo clippy` warnings. And I caught one of them in the review. :)

qiskit/circuit/library/generalized_gates/mcmt.py
Cryoris: Good point 👍🏻 related question: do you know whether the synthesis methods like `synth_mcx_...` are API-documented?

qiskit/circuit/library/generalized_gates/mcmt.py
Cryoris: Actually this is not needed as the gate being passed as callable will already trigger a pending deprecation in `MCMT.__init__` -- I'll remove this 🙂

qiskit/circuit/library/generalized_gates/mcmt.py
Cryoris: We could if we additionally ran HLS to apply the other modifiers that are not `control`. A cleaner way might be to completely disallow 2q gates, from which we only extract the 1q base gate. That would require us to handle less cases and might be a leaner workflow.

qiskit/circuit/library/generalized_gates/mcmt.py
Cryoris: no 🙂 I'll add a check

qiskit/synthesis/multi_controlled/mcmt_vchain.py
Cryoris: Yes 🙂 As discussed offline, we might want to make the V-chain available as generic control algorithm, akin to the current `add_control`. This would then allow HLS to synthesize any controlled gates or annotated controlled gates with either of the two decompositions.

crates/accelerate/src/synthesis/multi_controlled/mcmt.rs
Cryoris: You don't count 1 1 2 3? 😛

crates/accelerate/src/synthesis/multi_controlled/mcmt.rs
Cryoris: I added a check 👍🏻

crates/accelerate/src/synthesis/multi_controlled/mcmt.rs
Cryoris: ah perfect -- thanks!

crates/accelerate/src/synthesis/multi_controlled/mcmt.rs
Cryoris: Fixed 🙂

qiskit/circuit/library/generalized_gates/mcmt.py
alexanderivrii: I think so. The synthesis methods are described in https://docs.quantum.ibm.com/api/qiskit/synthesis. Also, the synthesis plugins are described in https://docs.quantum.ibm.com/api/qiskit/transpiler_synthesis_plugins.

Cryoris: Great, I added the Sphinx doc in 646279a 👍🏻

alexanderivrii:
Thank you very much! As per your question, could you please add a table for documenting MCMT synthesis plugins (in `hls_plugins.py`, the output should be in https://docs.quantum.ibm.com/api/qiskit/transpiler_synthesis_plugins) and the new synthesis function (the output should be in https://docs.quantum.ibm.com/api/qiskit/synthesis). Also this deserves a release note, listing new functionality and deprecations.

releasenotes/notes/mcmt-gate-a201d516f05c7d56.yaml
```diff
@@ -0,0 +1,36 @@
+---
+features_circuits:
+  - |
+    Added the :class:`.MCMTGate` to represent multi-control multi-target
+    operations in a gate. This gate representation of the existing :class:`.MCMT`
+    circuit allows the compiler to select the best available implementation
+    according to the number of auxiliary qubits present in the circuit.
```
alexanderivrii: This is super-minor, please feel free to ignore it:

```suggestion
    Added the :class:`.MCMTGate` to represent a multi-control multi-target
    operation as a gate. This gate representation of the existing :class:`.MCMT`
    circuit allows the compiler to select the best available implementation
    according to the number and the state of auxiliary qubits present in the circuit.
```
Do we want to list pending deprecations?

Cryoris: I would just list the deprecations once they are not pending anymore 🙂 (I think that's what we typically have as well 🙂)

alexanderivrii (dismissed):
LGTM!

qiskit/transpiler/passes/synthesis/hls_plugins.py
```diff
@@ -214,6 +214,41 @@
...
+      - Plugin class
+      - Number of clean ancillas
+      - Number of dirty ancillas
+      - Description
+    * - ``"vchain"``
+      - :class:`.MCMTSynthesisVChain`
+      - `k-1`
+      - `0`
+      - uses a linear number of Toffoli gates
+    * - ``"noaux_v24"``
```
ShellyGarion: I think it should be only noaux (without the v24)

Cryoris: Good catch!
```suggestion
    * - ``"noaux"``
```

alexanderivrii (dismissed):
Reapproving after the symengine fix.

mtreinish:
This has a merge conflict now that https://github.com/Qiskit/qiskit/commit/642debf8b2433aa2d814341b96fdc6803b53cbc0 has merged. The interface for `CircuitData::from_packed_operations` has changed to include a result and its usage here will need to be updated.

Cryoris:
Done in 60a82a3 👍🏻

qiskit/circuit/library/__init__.py
```diff
@@ -501,6 +501,7 @@
     Permutation,
     PermutationGate,
     GMS,
+    MCMTGate,
```
ShellyGarion, line 505: You should also add it to line 188 in this file - it currently does not appear in the circuit library docs!

alexanderivrii (dismissed):

alexanderivrii (approved):
Reapproving once more now that both Matthew's and Shelly's comments have been addressed! Thanks all!

PR #13159, opened 2024-09-16
Update `cibuildwheel` to unblock 0.46.3 release

ElePT:
### Summary
An attempt to fix https://github.com/Qiskit/qiskit/actions/runs/10884382139/job/30199542822

### DIFF
--- a/.github/workflows/wheels.yml
+++ b/.github/workflows/wheels.yml
@@ -24,7 +24,7 @@ jobs:
       - uses: dtolnay/rust-toolchain@1.73
         if: runner.os == 'macOS'
       - name: Build wheels
-        uses: pypa/cibuildwheel@v2.16.5
+        uses: pypa/cibuildwheel@v2.19.2
       - uses: actions/upload-artifact@v4
         with:
           path: ./wheelhouse/*.whl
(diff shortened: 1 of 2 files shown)


kevinhartman (dismissed):
LGTM, assuming it resolves the issue 😉.

eliarbel:
I'm not sure this will work... 

Running `cibuildwheel` version 2.16.5 (the one CI used for 0.46.3) locally, I get this: 
```
Error response from daemon: pull access denied for manylinux2014-latest, repository does not exist or may require 'docker login': denied: requested access to the resource is denied
```
So it seems there is no image named `manylinux2014-latest`. 

Per `cibuildwheel` [release notes](https://github.com/pypa/cibuildwheel?tab=readme-ov-file#v2192), updating it to version 2.19.2 and restoring the image name to `manylinux2014`, the build was successful. 

So I wonder, should we instead update `cibuildwheel` version for Linux in `.github/workflows/wheels.yaml` in the `stable/0.46` branch?

@mtreinish @jakelishman

ElePT:
I applied your suggestion in [ada6eab](https://github.com/Qiskit/qiskit/pull/13159/commits/ada6eab880c4d36b2cedc45db4418509d7cd955f), @eliarbel. I think it looks very reasonable and seems to work locally.

jakelishman (commented):
This seems to be ok.  I was worried because sometimes we've had to change various bits of other configuration as we bump `cibuildwheel`, but it looks like when we did this bump on `main`, I just lazily abandoned the backport rather than fixing the minor merge conflicts that were to do with the subsequent addition of PGO: https://github.com/Qiskit/qiskit/pull/12574.

That said, Elena: did the Linux build work locally for you, or just the macOS one?

eliarbel:
> That said, Elena: did the Linux build work locally for you, or just the macOS one?

My local Linux build worked OK

jakelishman (approved):

PR #13164, opened 2024-09-17
Deprecate Pulse package and dependencies

### LINKED ISSUE #13063: Deprecate pulse module and related functionality for removal in 2.0

eliarbel:
### Summary
This PR deprecates the Pulse package and any non-pulse package code within Qiskit that uses it. 

Closes #13063


eliarbel:
I think this PR is ready for review now as it contains all the deprecations, unit-test assertions, warning filtering and release notes I thought we should have for deprecating the pulse package. A couple notes worth highlighting and getting feedback on: 

1. I've introduced a few deprecation decorators, specifically for pulse, located in `qiskit.utils.deprecate_pulse.py`. The aim of these decorators is to reuse deprecations messages and overall reduce code clutter due to the deprecation code in many places. 
2. For unit test classes which interact heavily with pulse (e.g. all test cases in `test/python/pulse`) I used a new class level decorator which is used to suppress all deprecation warnings issued by pulse classes and methods. Same goes for calibration builder passes. 
3. `warnings.catch_warnings` is used in several places which are called in common workflows such as transpile and working with `Target`. The motivation being is to prevent pulse-related deprecation from being emitted in cases where the user code does not call deprecated pulse APIs directly.

releasenotes/notes/deprecate-pulse-package-07a621be1db7fa30.yaml
```diff
@@ -0,0 +1,80 @@
+---
+deprecations:
+  - |
+    The Qiskit Pulse package is being deprecated and will be removed in Qiskit version 2.0.0. Pulse-level
+    access is currently supported only on a subset of Eagle devices and not supported on the Heron architecture
+    ones. Furthermore, newer IBM Quantum architectures will not support pulse-level access as well.
+    As a consequence, supporting Pulse as a first-class citizen frontend in the Qiskit SDK itself makes little
```
ElePT: These are mostly nitpicks:

```suggestion
    The Qiskit Pulse package is being deprecated and will be removed in Qiskit 2.0.0. Pulse-level
    access is currently only supported on a subset of Eagle devices and not supported on the Heron architecture. Furthermore, newer IBM Quantum architectures will not support pulse-level access.
    As a consequence, supporting Pulse as a first-class citizen frontend in the Qiskit SDK itself makes little
```

qiskit/utils/deprecate_pulse.py
```diff
@@ -0,0 +1,90 @@
...
+
+def deprecate_pulse_dependency(*args, **kwargs):
+    """Deprecation message for functions and classes which use or depend on Pulse"""
+    decorator = deprecate_func(
+        since="1.3",
+        package_name="Qiskit",
+        removal_timeline="in Qiskit 2.0",
+        additional_msg="The entire Qiskit Pulse package is being deprecated "
+        "and this is a dependency on the package.",
+        **kwargs,
```
ElePT: I generally try to make all deprecation messages point to an alternative. From what I have understood, some dependencies might be moved to Qiskit Dynamics but some might not? Maybe we could say something like:

```suggestion
        additional_msg="The entire Qiskit Pulse package is being deprecated "
        "and this is a dependency on the package. The relevant components ",
         "for low-level pulse control have been moved to the Qiskit Dynamics ",
         "repository:  https://github.com/qiskit-community/qiskit-dynamics",
        **kwargs,
```

qiskit/utils/deprecate_pulse.py
```diff
@@ -0,0 +1,90 @@
...
+def deprecate_pulse_arg(arg_name, **kwargs):
+    """Deprecation message for arguments related to Pulse"""
+    return deprecate_arg(
+        name=arg_name,
+        since="1.3",
+        package_name="Qiskit",
+        removal_timeline="in Qiskit 2.0",
+        additional_msg="The entire Qiskit Pulse package is being deprecated "
+        "and this argument uses a dependency on the package.",
+        **kwargs,
```
ElePT, line 91: Same comment as above:

```suggestion
        additional_msg="The entire Qiskit Pulse package is being deprecated "
        "and this argument uses a dependency on the package. The relevant components ",
         "for low-level pulse control have been moved to the Qiskit Dynamics ",
         "repository:  https://github.com/qiskit-community/qiskit-dynamics",
        **kwargs,
```

qiskit/utils/deprecate_pulse.py
```diff
@@ -0,0 +1,90 @@
...
+    """Put a given decorator on all the decorated class methods whose name starts with `test_`"""
+
+    def cls_wrapper(cls):
+        for attr in dir(cls):
+            if attr.startswith("test_") and callable(object.__getattribute__(cls, attr)):
+                setattr(cls, attr, decorator(object.__getattribute__(cls, attr)))
+
+        return cls
+
+    return cls_wrapper
```
ElePT, line 119: In my experience, we usually tackle these large deprecations adding filters to the qiskit test base class rather than defining a custom decorator for this: https://github.com/Qiskit/qiskit/blob/fbfe7388f3efdaaeb70dc96aa35f649c1897c5d4/test/utils/base.py#L171-L177. It is less granular than your approach (in the sense that we don't control which tests have raised these filtered warnings) but I think it makes the removal process smoother. This is not standardized and it probably will depend a bit on who you ask.

ElePT (commented):
Thanks @eliarbel!! This was a very large task to tackle. I added a few suggestions regarding deprecation messages and the filtering strategy. As for the use of custom decorators to reduce code repetition, we don't commonly do it but I see how in this case they are handy. Technically, as long as they don't show up in the docs, they wouldn't be part of the documented interface, so we would be able to remove them in 2.0 without issues.

releasenotes/notes/deprecate-pulse-package-07a621be1db7fa30.yaml
eliarbel: Thanks @ElePT for looking at this! A lot of files to review and I indeed contemplated at some point whether it makes sense to split this to smaller PRs. But since the vast majority of changes are trivial, I ended up leaving this as is, hoping it's manageable (and helpful) to review this as a whole. 
Anyway, suggestion looks good.

qiskit/utils/deprecate_pulse.py
eliarbel: Well, there will be no alternative in Qiskit for pulse and pulse-based things once Pulse is removed. So I think it makes sense to add your suggestion only for the things which will move with Pulse. And it can be refined to something like:  "The <X>component will be moved together with Qiskit Pulse to the Qiskit Dynamics repo...". For things which are not moving to Dynamics (e.g. the calibration passes) we can leave the original message or add something more explicit in the line of "Note that once removed in Qiskit 2.0, there will be no alternative functionality for this in Qiskit". What do you say?

qiskit/utils/deprecate_pulse.py
eliarbel: All deprecated args will be removed without alternatives, in the same spirit of the suggestion in [your comment above](https://github.com/Qiskit/qiskit/pull/13164#discussion_r1800975738), we should say this explicitly for the deprecated args?

qiskit/utils/deprecate_pulse.py
eliarbel: There is a tradeoff here... It seems that the preferred way in Qiskit to handle deprecation warnings in testing is by using `unittest.assertWarns` and that `QiskitTestCase`-level filtering is meant to be used mainly for messages emitted by 3rd party packages or otherwise in very special circumstances like deprecating a key core element which makes the use of `assertWarns` really problematic.
In this case, I wanted to avoid putting `assertWarns` everywhere in `test/python/pulse` but still use it on deprecated functions outside of Pulse like `QuantumCircuit.calibrations`. So I took a hybrid approach where I used the `@decorate_test_methods(ignore_pulse_deprecation_warnings)` only in `test/python/pulse` and `test/python/visualization_v2` which both are being deprecated as a whole, as well as on the calibration builder passes in `test/python/transpiler` which are being deprecated as well. I like this balance but I'm open to other views or ideas about how to achieve this in a different way. 
BTW, FWIW, these decorators are useful as markers for things to be removed in 2.0.

qiskit/utils/deprecate_pulse.py
ElePT: I think it makes sense

eliarbel:
Here is the result of a performance comparison of Qiskit benchmarks, showing only the change w.r.t the default ASV factor.  There isn't a clear trend and I'm curious about what these changes mean really. Anyway for now I think we should wait a bit until the PR is stable and then rerun this and decide whether or not it's worthwhile to dive into this in more detail.


| Change   | Before [54fe09b2] <main>   | After [f78a9574] <deprecate-pulse>   |   Ratio | Benchmark (Parameter)                                                                                    |
|----------|----------------------------|--------------------------------------|---------|----------------------------------------------------------------------------------------------------------|
| -        | 138±0.7μs                  | 106±0.5μs                            |    0.77 | circuit_construction.CircuitConstructionBench.time_circuit_copy(14, 8192)                                |
| -        | 1.68±0.07ms                | 1.05±0.04ms                          |    0.63 | circuit_construction.CliffordSynthesis.time_clifford_synthesis(10)                                       |
| +        | 1.02±0.02ms                | 1.69±0.2ms                           |    1.65 | mapping_passes.PassBenchmarks.time_dense_layout(14, 1024)                                                |
| +        | 3.62±0ms                   | 3.99±0.03ms                          |    1.1  | mapping_passes.PassBenchmarks.time_layout_2q_distance(14, 1024)                                          |
| +        | 1.15±0.01ms                | 1.27±0ms                             |    1.1  | mapping_passes.PassBenchmarks.time_layout_2q_distance(5, 1024)                                           |
| +        | 471±1ms                    | 525±0.5ms                            |    1.12 | passes.MultipleBasisPassBenchmarks.time_basis_translator(14, 1024, ['rx', 'ry', 'rz', 'r', 'rxx', 'id']) |
| +        | 667±3ms                    | 748±2ms                              |    1.12 | passes.MultipleBasisPassBenchmarks.time_basis_translator(20, 1024, ['rx', 'ry', 'rz', 'r', 'rxx', 'id']) |
| +        | 158±0.4ms                  | 174±0.1ms                            |    1.1  | passes.MultipleBasisPassBenchmarks.time_basis_translator(5, 1024, ['rx', 'ry', 'rz', 'r', 'rxx', 'id'])  |
| +        | 312±2ms                    | 344±6ms                              |    1.1  | quantum_info.RandomCliffordBench.time_random_clifford('5,1000')                                          |
| -        | 16.3±0.4ms                 | 5.60±0.4ms                           |    0.34 | quantum_volume.QuantumVolumeBenchmark.time_ibmq_backend_transpile(1, 'translator')                       |
| -        | 18.1±1ms                   | 6.35±0.05ms                          |    0.35 | transpiler_benchmarks.TranspilerBenchSuite.time_single_gate_compile                                      |
| +        | 54.9±0.9ms                 | 77.6±6ms                             |    1.41 | transpiler_qualitative.TranspilerQualitativeBench.time_transpile_time_cnt3_5_180(0, 'sabre', 'sabre')    |

ElePT (dismissed):
Overall, the deprecation looks good, and I think it is in a mergeable state. The only missing point I think could be improved is my previous comment on adding a link to Qiskit Dynamics for the deprecated pulse dependencies. I suggested always mentioning the link for simplicity (even for dependencies that are not being ported to Dynamics), but of course having a dedicated decorator for the elements that are being ported would be even better. I don't think it's a deal breaker though, so I will approve, and can re-approve if you decide to implement the suggestion.

ElePT:
And regarding the benchmarks, I see what you mean, and in my experience the variability of the asv results is quite high, which makes certain improvements difficult to track (I have seen this with Unitary Synthesis, where the tests that didn't run it would also change quite a lot when benchmarking the PR).

eliarbel:
> The only missing point I think could be improved is my previous comment on adding a link to Qiskit Dynamics for the deprecated pulse dependencies.

Thanks @ElePT . I have implemented your suggestion already but haven't pushed it yet since I'm waiting for input from @nkanazawa1989 and @DanPuzzuoli regarding the exact dependencies which should be moved to Dynamics. Putting the list here so we can discuss in a wider forum if needed:

`qiskit.compiler.schedule()`
`qiskit.compiler.sequence()`
`qiskit.scheduler.schedule_circuit()`
`qiskit.scheduler.sequence()`
`qiskit.scheduler.ScheduleConfig`
`qiskit.scheduler.methods.as_soon_as_possible()`
`qiskit.scheduler.methods.as_late_as_possible()`
`qiskit.visualization.pulse_v2.draw()`

qiskit/utils/deprecate_pulse.py
eliarbel: Implemented in https://github.com/Qiskit/qiskit/pull/13164/commits/040119438c9f95b7a11a383a5d7aa8ce73fdad0c

qiskit/utils/deprecate_pulse.py
```diff
@@ -32,25 +32,52 @@ def deprecate_pulse_func(func):
...
-        additional_msg="The entire Qiskit Pulse package is being deprecated "
-        "and this is a dependency on the package.",
-        **kwargs,
-    )
+def deprecate_pulse_dependency(*args, moving_to_dynamics: bool = False, **kwargs):
+    """Deprecation message for functions and classes which use or depend on Pulse
+
+    moving_to_dynamics: set to True if the dependency is moving to Qiskit Dynamics. This affects
+    the deprecation message being printed, namely saying explicitly whether the dependency will
+    be moved to Qiskit Dynamics or whether it will just be removed without an alternative.
```
ElePT: This will be my final nitpick, I know the docstring is not going to stay, but in case it's used as a reference moving forward for future deprecation decorators:

```suggestion
        Args:
            moving_to_dynamics: set to True if the dependency is moving to Qiskit Dynamics.
                This affects the deprecation message being printed, namely saying explicitly whether 
                the dependency will be moved to Qiskit Dynamics or whether it will just be removed 
                without an alternative.
```

ElePT (commented):
Thanks a lot! I agree with the changes, the PR is good to go, I just had one tiny final comment.

qiskit/utils/deprecate_pulse.py
eliarbel: Done, thanks.

ElePT (approved):

Issue #13166, opened 2024-09-17
Wrong operation names when loading/exporting circuits using QASM2


cbjuan:
### Environment

- **Qiskit version**: 1.2.1
- **Python version**: 3.11
- **Operating system**: OSX

### What is happening?

Loading/exporting QASM2 code gives sometimes wrong operation names for gates like ECR or RZX, leading to errors like

```
qiskit.qasm2.exceptions.QASM2ParseError: "<input>:11,39: 'rzx_4520860432_5206602512' is not defined in this scope"
```

or showing gate name patterns like `ecr_.*` or `rzx_.*` when listing the operations available in the circuit. 

### How can we reproduce the issue?

QASM2

```
from qiskit.circuit import QuantumCircuit
from qiskit import qasm2, transpile
from qiskit_ibm_runtime import QiskitRuntimeService

num_qubits = 16
circuit = QuantumCircuit(num_qubits)
for i in range(num_qubits-1):
    circuit.ecr(i,i+1)

runtime_service = QiskitRuntimeService(channel="ibm_quantum")
backend = runtime_service.backend("ibm_sherbrooke")

def transpile_qasm(qc):
    qc = qasm2.loads(qasm2.dumps(qc), custom_instructions=qasm2.LEGACY_CUSTOM_INSTRUCTIONS)
    tqc = transpile(qc, backend=backend, optimization_level=3, seed_transpiler=123)
    tqc = qasm2.loads(qasm2.dumps(tqc), custom_instructions=qasm2.LEGACY_CUSTOM_INSTRUCTIONS)
    return tqc

transpiled_circuit = transpile_qasm(circuit)
print({data.name for data in transpiled_circuit.data})

```

### What should happen?

Get the expected names for the different operations (`"ecr"`, `"rzx"`, etc) instead of names like `"rzx_4520860432_5206602512"` or `"ecr_140645991950416"`

cbjuan:
Related to https://github.com/Qiskit/qiskit-ibm-transpiler/issues/73

jakelishman:
[...]

PR #13167, opened 2024-09-17
Add example of `qasm2.CustomInstruction`

jakelishman:
### Summary

We didn't have an explicit usage example of this in the documentation before, and several people seem to have been confused by it.

### Details and comments

For example, #13166.


ElePT (approved):
LGTM

PR #13168, opened 2024-09-17
Add inverse function to StandardGate in rust

### LINKED ISSUE #13157: Add inverse function to StandardGate in rust
### What should we add?

Add `inverse` function to all the `StandardGate`s.
This will be useful in order to write several synthesis algorithms natively in rust, e.g. https://github.com/Qiskit/qiskit/issues/12907

ShellyGarion:
### Summary
Close #13157


ShellyGarion:
The following gates do not have an inverse which is a `StandardGate`:

- `DCXGate`
- `ISwapGate`
- `CSXGate`
- `RCCXGate`
- `C3SXGate`
- `RC3XGate`

crates/circuit/src/operations.rs
```diff
@@ -2308,4 +2495,7 @@ impl Operation for PyOperation {
             }
         })
     }
+    fn inverse(&self, _params: &[Param]) -> Option<(StandardGate, Vec<Param>)> {
+        None
+    }
```
mtreinish: Is there ever a case where a Python `Operation` will have an inverse defined? I know it's not part of the abstract interface, but I'm wondering if we should try and fallback to `None`? @alexanderivrii maybe you have some thoughts here?

crates/circuit/src/operations.rs
```diff
@@ -2043,6 +2057,155 @@ impl Operation for StandardGate {
     fn standard_gate(&self) -> Option<StandardGate> {
         Some(*self)
     }
+    fn inverse(&self, params: &[Param]) -> Option<(StandardGate, Vec<Param>)> {
```
mtreinish: Everywhere else we work with parameters in Rust we tend to use `SmallVec<[Param; 3]>` to keep things on the stack unless there are more than 3 elements. It might be worth changing this and updating the `vec!` macro usage to `smallvec!`.

crates/circuit/src/operations.rs
```diff
@@ -167,6 +167,7 @@ pub trait Operation {
     fn definition(&self, params: &[Param]) -> Option<CircuitData>;
     fn standard_gate(&self) -> Option<StandardGate>;
     fn directive(&self) -> bool;
+    fn inverse(&self, params: &[Param]) -> Option<(StandardGate, Vec<Param>)>;
```
mtreinish: Does this work in the case of `PyGate`? I don't think there is a guarantee that a Python defined gate's inverse will be a `StandardGate`. I actually think it is pretty unlikely in practice, since if a custom gate were expressible as the inverse of a `StandardGate` it'd probably just be a `StandardGate`. I think you can just do:

```suggestion
    fn inverse(&self, params: &[Param]) -> Option<(Self, Vec<Param>)>;
```
then for `impl Operation for StandardGate` `.inverse()` returns a `StandardGate` and for `impl Operator for PyGate` `.inverse()` returns a `PyGate`.

crates/circuit/src/operations.rs
```diff
@@ -2043,6 +2057,155 @@ impl Operation for StandardGate {
...
+            Self::CCXGate => Some((Self::CCXGate, vec![])),
+            Self::CCZGate => Some((Self::CCZGate, vec![])),
+            Self::CSwapGate => Some((Self::CSwapGate, vec![])),
+            Self::RCCXGate => None, // the inverse in not a StandardGate
+            Self::C3XGate => Some((Self::C3XGate, vec![])),
+            Self::C3SXGate => None, // the inverse in not a StandardGate
+            Self::RC3XGate => None, // the inverse in not a StandardGate
+        }
+    }
 }
```
mtreinish: Did you want to add a rust space test asserting that all the StandardGates multiplied with their inverse were the identity? It would be good to get in the practice of having more rust space tests, but since we have coverage from python I don't think it's strictly necessary.

@kevinhartman has been working on demonstrating a pattern of using the `py` token from rust space tests: https://github.com/Qiskit/qiskit/pull/13169 (since we still need it for `multiply_param` even if there isn't a parameter expression).

mtreinish (commented):
Thanks for getting this started it looking good. I have a couple of inline comments and questions to start.

crates/circuit/src/operations.rs
ShellyGarion: `Clifford` (which is an `Operation`) does not have an `inverse` only `adjoint`.

ShellyGarion: But there are many generalized gates (like `PermutationGate`) that have an `inverse` method

ShellyGarion: how would you suggest to change the code here?

crates/circuit/src/operations.rs
ShellyGarion: I added a test in `test_rust_equivalence` that asserts that the output of the inverse in Rust (when not None) is the same as the output of the inverse in Python

crates/circuit/src/operations.rs
ShellyGarion: In this case, it's unclear to me which type to put in line 279. Any type I try gives a mismatched type error

jakelishman: We probably can't have `Operation::inverse` return anything referring to `Self`, because that's unimplementable by the implementors of `Operation` that don't have an owned type (e.g. `OperationRef`).  Even within that, it's maybe suboptimal because not all `StandardGate`s would be able to return an inverse, even though they can be inverted (e.g. `DCX`).

The most general would potentially be to have `Operation::inverse` return some iterator type over (potentially) `PackedInstruction` (like `CircuitData`), but in practice that's probably not super performant for many simple cases.  Maybe in the near term, it's enough just to have `StandardGate::try_inverse` on `StandardGate` only?

ShellyGarion: There is a problem in implementing `inverse` in `impl<'a> Operation for OperationRef<'a>` (line 279) that cause a mismatched type error. After talking with @jakelishman it seems that the best solution would be to implement `inverse` only for `StandadGate` (maybe also for `PyGate`)

ShellyGarion: I'm a bit confused now. I can add `inverse` as a stand-alone function (only for `StandardGate`) - but then how can I test it from Python ?

```
pub fn inverse(gate:StandardGate, params: &[Param]) -> Option<(StandardGate, SmallVec<[Param; 3]>)> {
    match gate {
        StandardGate::HGate => Some((StandardGate::HGate, smallvec![])),
        StandardGate::IGate => Some((StandardGate::IGate, smallvec![])),
        ...
```

ShellyGarion: So I only replaced the function name `inverse` by `try_inverse` and the function for `PyGate` will return `None`.

crates/circuit/src/operations.rs
ShellyGarion: Following the discussion below, I think for now the main use-case is to add an `inverse` only for the `StandardGate`s (at least for most of them whose inverse is also a `StandardGate`)

test/python/circuit/test_rust_equivalence.py
```diff
@@ -146,6 +146,22 @@ def test_matrix(self):
...
+            if standard_gate is None:
+                # gate is not in rust yet
+                continue
+
+            with self.subTest(name=name):
+                params = [0.1 * (i + 1) for i in range(standard_gate._num_params())]
+                py_def = gate_class.base_class(*params).inverse()
+                rs_def = standard_gate._inverse(params)
+                if rs_def is not None:
+                    np.testing.assert_equal(py_def.name, rs_def[0].name)
```
mtreinish:
```suggestion
                    self.assertEqual(py_def.name, rs_def[0].name)
```

crates/circuit/src/operations.rs
```diff
@@ -169,6 +169,7 @@ pub trait Operation {
     fn definition(&self, params: &[Param]) -> Option<CircuitData>;
     fn standard_gate(&self) -> Option<StandardGate>;
     fn directive(&self) -> bool;
+    fn try_inverse(&self, params: &[Param]) -> Option<(StandardGate, SmallVec<[Param; 3]>)>;
```
mtreinish: I know this was a suggestion from Jake, but I'm not actually a fan of the `try_inverse` name here. In the rust stdlib the `try_methods` are used for fallible versions of a method that return an `Err` if it's not possible. In this case however, we're returning a `None` which is different behavior. It looks like the unstable `Try` trait will work with `Option` returns: https://doc.rust-lang.org/std/ops/trait.Try.html
but looking through the standard library API every example of a stable `try_` prefixed method returns a `Result`.

I know it's a bit pedantic to complain about the naming here, but I think we should either revert this name back to `inverse` or change the output type to be a `Result` instead of `Option`. Especially since the closest existing api is `StandardGate` which returns a `Option<StandardGate>`. Either way, there are enough caveats with this new method that I think it deserves a docstring to explain these limitations and the expected behavior. The others are all fairly straightforward and are more or less self documenting (e.g. num_qubits returning an integer), but this is a lot more nuanced: the inverse of an operation  when the inverse is a standard gate (except if the inverse is a standard gate but the original gate is defined in Python)

The other option is to just define this method on `StandardGate` and leave a potential broader inverse method defined in the `Operation` trait for a separate PR. Doing this though requires that as the consumer you are working with a `StandardGate` and not a generic `Operation`.

mtreinish (commented):
I know that the current state of the API in this PR was based on previous review suggestions, but I have some concerns about the naming and the exact interface here. I think the goal of making an `inverse()` method on the operation trait makes a lot of sense, but I have concerns about the current api being called `try_inverse` and it returning an `Option` and then never working for anything besides a `StandardGate` or a `OperationRef` to a `StandardGate`. I think if we are going to add this to the `Operation` trait we should attempt to support all the implementations out there. The only way that makes sense I guess is a `Box<dyn Operation>` or something similar for the return so that we can dynamically support whatever the inverse implementation returns. I"m not advocating for this though as that adds a layer of pointer indirection for what your immediate use case with `StandardGate` which is less than desirable.

All of this is to say I think to start instead of adding this to `Operation` we should just reduce the scope of this PR and define `inverse()` as a method for `StandardGate` for now. We can debate what an appropriate interface is for the `Operation` trait in a separate issue or PR that's not blocking work on something else and then just pivot the `StandardGate` method to follow that interface (or have the trait method call the `StandardGate` method inside of it).

crates/circuit/src/operations.rs
ShellyGarion: As discussed - currently `inverse` is implemented only for `StandardGate` and not for `Operation`.

mtreinish (approved):
Thanks for the quick update. This LGTM now with the reduced scope of the new API. I think we can discuss how best to add an inverse method to the wider `Operation` trait in a follow up. Thinking about how we make the trait's method work will be especially important as we add more types implementing the trait as part of #13264.

PR #13169, opened 2024-09-17
Enable PyO3 in cargo unit tests.

kevinhartman:
### Summary
At present, most of our core data structures like `DAGCircuit` internally rely on the Python interpreter in some way, which has so far prevented us from writing Rust-based unit tests. Up to this point, we've gotten away with depending on our Python-based test coverage to exercise Rust code internally, but this is quickly becoming insufficient as we add more Rust-only API surface.

This PR does a few things to enable the use of the Python interpreter and load Qiskit's Python code into it:

- The test executable is built with PyO3 feature `auto-initialize`, which initializes the free-threaded Python interpreter. This enables the use of `Python::with_gil` and thus access to a `Python` token from within tests.
- We disable `miri` for such tests, since it is not compatible (e.g. `#[cfg(not(miri))]`).
- We set the `PYTHONUSERBASE` environment variable to the active `venv`'s directory. This allows the Python interpreter baked into the test executable to locate the Python `qiskit` module.
- A new `tox` env has been added to orchestrate this. To run Rust tests, simply invoke `tox -erust`.

### Details and comments
As an example, this PR adds unit tests for `DAGCircuit::push_back` and `DAGCircuit::push_front`. We really ought to add more testing for the other new Rust-facing methods (e.g. `DAGCircuit::extend` via additional PRs).

As more of Qiskit's core is ported to Rust, our reliance on the Python interpreter from Rust code should eventually be overcome, and at that point we shouldn't need these changes anymore.


jakelishman:
If you enable the `pyo3/auto-initialize` feature during the test-suite build, then the `pyo3::prepare_freethreaded_python` calls won't be necessary.  Don't know if you had other work around the test harness planned, though.

CONTRIBUTING.md
```diff
@@ -565,11 +565,15 @@ Note: If you have run `test/ipynb/mpl_tester.ipynb` locally it is possible some
...
-where new Rust-native APIs are being added, or there are Rust-specific internal details
-to be tested, `#[test]` functions can be included inline. It's typically most
-convenient to place these in a separate inline module that is only conditionally
-compiled in, such as
+Many of Qiskit's core data structures and algorithms are implemented in Rust.
+The bulk of this code is exercised heavily by our Python-based unit testing,
+but this coverage really only provides integration-level testing from the
+perspective of Rust.
+
+To provide proper Rust unit testing, we use `cargo test`. Rust tests are
```
mtreinish: I might drop "proper"  here because I'm not sure I know what a proper unit test is :)

```suggestion
To provide Rust unit testing, we use `cargo test`. Rust tests are
```

CONTRIBUTING.md
```diff
@@ -581,18 +585,67 @@ mod tests {
...
 documentation's [guide on writing tests](https://doc.rust-lang.org/book/ch11-01-writing-tests.html).
 
-To run the Rust-space tests, do
+Rust tests are run separately from the Python tests. The easiest way to run
+them is via ``tox``, which creates an isolated venv and pre-installs ``qiskit``
+prior to running ``cargo test``:
+
+```bash
+tox -erust
+```
```
mtreinish, line 604: My only concern  here is that this results in building Qiskit from source twice on every execution. I'm not sure there is a way around this since we need the python extension built to be able to call it via python in the rust tests it just is pretty slow. We should probably build in debug mode by default in tox and for rust testing by default.

CONTRIBUTING.md
```diff
@@ -581,18 +585,67 @@ mod tests {
...
+
+You can also execute them directly in your own virtual environment with these
+commands (which is what the ``tox`` env is doing under the hood):
```
mtreinish: There is a path to running without a virtual environment, but that's probably not good to encourage because it means people will be installing a dev version of qiskit in their system site packages. It might be good to add a sentence here explaining you should make a venv for testing in this case.

CONTRIBUTING.md
```diff
@@ -581,18 +585,67 @@ mod tests {
...
 
 ```bash
-cargo test --no-default-features
+python setup.py build_rust --release --inplace
+PYTHONUSERBASE="$VIRTUAL_ENV" cargo test --no-default-features
+```
+
+The first command builds Qiskit (in release, but --debug is fine too) in editable mode,
```
mtreinish:
```suggestion
The first command builds Qiskit (in release, but `--debug` is fine too) in editable mode,
```

CONTRIBUTING.md
```diff
@@ -581,18 +585,67 @@ mod tests {
...
+```bash
+tox -erust
+```
+
+You can also execute them directly in your own virtual environment with these
+commands (which is what the ``tox`` env is doing under the hood):
 
 ```bash
-cargo test --no-default-features
+python setup.py build_rust --release --inplace
```
mtreinish: I'd actually change this to not include `--release` by default. That just increases the build time and we shouldn't be bottlenecked on the python extension's execution time in rust tests.

tox.ini
```diff
@@ -31,6 +31,14 @@ deps =
 commands =
   stestr run {posargs}
 
+[testenv:rust]
+basepython = python3
+setenv =
+  PYTHONUSERBASE={envdir}
```
mtreinish, line 38: The default tox configuration will build in release mode, I think for rust testing we should use debug mode by default.

crates/circuit/src/dag_circuit.rs
```diff
@@ -6916,3 +6916,207 @@ fn add_global_phase(py: Python, phase: &Param, other: &Param) -> PyResult<Param>
...
+                .dag
+                .edge_references()
+                .map(|e| (e.source(), e.target(), e.weight().clone()))
+                .collect();
+
+            assert_eq!(actual_wires, expected_wires, "unexpected DAG structure");
+
+            // Add measures after CX.
+            let measure_q0 = measure!(dag, 0, 0);
+            let measure_q0_node = dag.push_back(py, measure_q0)?;
```
mtreinish, line 7149: Does returning the pyresult here do the correct thing? Like would this error if push back returned an `Err`? If not I'd say we should either replace `?` with an `unwrap()` or `expect()` so the error is returned from the test.

mtreinish (commented):
Overall this looks great to me, I really am looking forward to having an established pattern for doing more involved rust testing. Especially as we're moving more into rust. I left a few inline comments, mostly nits in the docs and a question in the tests. The only concern I have right now is that we're defaulting to building the python extension in release mode for rust testing and I feel like we should be defaulting to debug mode for the extension when in a rust testing context. Not because I think we need the debug symbols or optimizations disabled, but just to reduce the compile time since the extension is only there to make python work and we shouldn't really be concerned with the runtime performance of it in a rust testing context.

crates/circuit/src/dag_circuit.rs
kevinhartman: According to the docs, it should!

https://doc.rust-lang.org/rust-by-example/testing/unit_testing.html#tests-and-

mtreinish: Hah, this is another place starting with rust in 2017 comes back to bite me, I really did think it didn't and that's because I started by using edition 2015. :)

CONTRIBUTING.md
kevinhartman: I've enabled debug mode by default for the `rust` env's package environment in 6856adbb9dc006ef0ed85f7869fd15778b470b0a.

I also documented `--skip-pkg-install` as an option for running without rebuilding Qiskit when invoking `tox -erust`, and I've explained there that this is only an option if you've already built the current working tree: 4813a7c057efe7cec1f49efe30cdbb71e4d96774

CONTRIBUTING.md
kevinhartman: Good point. I've reworded this section to guide users to first create and activate a virtual environment if they're running without tox.

234b6a56e847914cfb75c6462489bb1900bb4ea1

CONTRIBUTING.md
kevinhartman: Should be done in 234b6a56e847914cfb75c6462489bb1900bb4ea1 as well.

tox.ini
kevinhartman: Done in 6856adbb9dc006ef0ed85f7869fd15778b470b0a.

kevinhartman:
This should be ready to go, @mtreinish.

The Rust tests for `DAGCircuit` added in the PR can be seen passing in CI here: https://dev.azure.com/qiskit-ci/qiskit/_build/results?buildId=62018&view=logs&j=bf26c3d0-9b88-5f67-96f4-43488114ea6d&t=5c74c3c4-ab20-5140-0e04-11aec0505a88&l=148

CONTRIBUTING.md
```diff
@@ -581,18 +592,78 @@ mod tests {
...
+If you're not using `tox`, you can also execute Cargo tests directly in your own virtual environment.
+If you haven't done so already, [create a Python virtual environment](#set-up-a-python-venv) and
+**_activate it_**.
+
+Then, run the following commands:
 
 ```bash
-cargo test --no-default-features
+python setup.py build_rust --inplace
+PYTHONUSERBASE="$VIRTUAL_ENV" cargo test --no-default-features
```
eliarbel, line 622: Overall this PR is great, long needed for running Rust tests easily with PyO3 in Qiskit! 

My only comment is about virtual environments other than `venv`. For Conda I had to set `LD_LIBRARY_PATH` to point to the `lib` directory under the env dir so the embedded interpreter could find the required shared libraries.  I don't think we should provide examples for other custom virtual envs, but rather just put a [!note] comment that for other virtual envs a different setup might be required (and maybe mention `LD_LIBRARY_PATH` for Conda since it's common).

kevinhartman: Thanks for trying this on Conda. Any chance you could try the following instead of `LD_LIBRARY_PATH`? 

```
PYTHONUSERBASE="$CONDA_PREFIX" cargo test --no-default-features
```

eliarbel: This was actually my first attempt to make it work in Conda, following the nice documentation you put together. But it doesn't help. I didn't dig into the actual root cause but it surely something about the embedded interpreter not being aware to the virtual environment if it's Conda. I wouldn't put time solving this but rather just include a warning in this section for users to note.

kevinhartman: I'd forgotten that on Linux, the rpath of the test executable doesn't include the Python library. To work around this, the `rust` tox env now exports `LD_LIBRARY_PATH` to include the path to the current Python environment's shared Python lib before running `cargo test`. This is similar to what we do in CI.

Can you try running `tox -erust` on Linux again with these changes? I've also updated the docs for manual invocations without tox.

eliarbel: I can confirm it's working now. Thanks @kevinhartman!

mtreinish (approved):
LGTM, thanks for doing this

PR #13176, opened 2024-09-18
Avoid exporting incorrect PyInit_* symbols (0.46.3)

eliarbel:
This is a manual backport of the change introduced in PR #12889. It is done manually since in 0.46 we are using PyO3 version 0.19.2 which does not yet have the `Bound` concept PR #12889 relies on in the `add_submodule` helper function. In this branch we also need to handle just a subset of submodule functions compared to the ones in #12889.


jakelishman (approved):
Assuming you've removed all the `pymodule` tags except for accelerate's main module, this looks right to me.

Issue #13161, opened 2024-09-16
sympy2symengine error in circuit transpilation


ProThicc:
### Environment

- **Qiskit version**: 1.2.1
- **Python version**: 3.11.8
- **Operating system**: Ubuntu 22.04 LTS

### What is happening?

I am trying to transpile a circuit with CCU and CU gates with different rotation angles. In the qiskit 11.4 version the transpiler was able to optimize and transpile the circuit with the basis gate set ['sx','cz','x','rz','barrier', 'save_statevector', 'delay'], But for the current version of qiskit, the transpiler throws an error for the particular gate set.

SympifyError: sympy2symengine: Cannot convert 'None' (of type <class 'NoneType'>) to a symengine type.

### How can we reproduce the issue?

``` 
from qiskit_aer import AerSimulator
from qiskit import transpile, QuantumCircuit
from qiskit.circuit.library import UGate
from qiskit.circuit.add_control import add_control

qc = QuantumCircuit(4)
qca_U = add_control(UGate(-1.945632646294958, 1.5120405041931422, 2.4393357222229826),num_ctrl_qubits=2, ctrl_state='00', label = 'qca_u')
qca_U_red = add_control(UGate(-1.945632646294958, 1.5120405041931422, 2.4393357222229826),num_ctrl_qubits=1, ctrl_state='0', label = 'qca_u_red')
qc.append(qca_U, [0,2,1])
qc.append(qca_U_red, [0,1])
sim = AerSimulator()
# qcd = transpile(qc, sim, optimization_level=3,basis_gates=['ccx','ccz','cu','cx','u','mcphase','x', 'cx', 'save_statevector','delay', 'barrier'])
qcd = transpile(qc, sim, optimization_level=3, basis_gates=['barrier', 'save_statevector', 'delay','sx','cz','x','rz'])
# qcd = transpile(qc, sim, optimization_level=3)

```

### What should happen?
```
{
	"name": "SympifyError",
	"message": "sympy2symengine: Cannot convert 'None' (of type <class 'NoneType'>) to a symengine type.",
	"stack": "---------------------------------------------------------------------------
SympifyError                              Traceback (most recent call last)
Cell In[23], line 13
     11 sim = AerSimulator()
     12 # qcd = transpile(qc, sim, optimization_level=3,basis_gates=['ccx','ccz','cu','cx','u','mcphase','x', 'cx', 'save_statevector','delay', 'barrier'])
[...]

Issue #12625, opened 2024-06-21
Invalid Parameters with CU gate in AerSimulator


ProThicc:
### Environment

- **Qiskit version**: 1.1.0
- **Python version**: 3.11.8
- **Operating system**: Ubuntu 22.04.4 LTS

### What is happening?

Custom Single Control Unitary gate shows Invalid Operation for AerSimulator from qiskit_aer

### How can we reproduce the issue?

``` 
from qiskit import QuantumCircuit
from qiskit_aer import AerSimulator
from qiskit.circuit.add_control import add_control
from qiskit.circuit.library import UGate
qc = QuantumCircuit(3)
Sim = AerSimulator()
gate = UGate(-1.945632646294958, 1.5120405041931422, 2.4393357222229826)
qca_gate_reduced = add_control(gate, label='test', num_ctrl_qubits=1, ctrl_state=1)
qc.append(qca_gate_reduced, [1, 0])
qc.save_statevector()
result = Sim.run(qc).result().data(0)
```

### What should happen?

Simulator should function giving the statevector after the Control Operation

### Any suggestions?

Probably passing a default value of gamma for cu Gate should resolve the problem (I suspect that is the issue)

nihaocami:
@ProThicc instead of using `add_control` , have you tried using the internal `control` method? like this:
`qca_gate_reduced = gate.control(label='test', num_ctrl_qubits=1, ctrl_state=1)`

ProThicc:
> @ProThicc instead of using `add_control` , have you tried using the internal `control` method? like this: `qca_gate_reduced = gate.control(label='test', num_ctrl_qubits=1, ctrl_state=1)`


When I try that I get this error, which makes sense that the CU gate defined when num_ctrls = 1 requires an additional parameter gamma. This same issue shows up with add_control which basically doesn't include this parameter when passing this gate as an instruction to the simulator. I have added a patch which by default sets this gamma variable as 0. This issue doesn't show up when num_ctrl_qubits >1 since the additional variables shown up are implicitly set to be 0

```
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
Cell In[6], [line 8](vscode-notebook-cell:?execution_count=6&line=8)
      [6](vscode-notebook-cell:?execution_count=6&line=6) Sim = AerSimulator()
      [7](vscode-notebook-cell:?execution_count=6&line=7) gate = UGate(-1.945632646294958, 1.5120405041931422, 2.4393357222229826)
[...]

Issue #11458, opened 2023-12-22
Running a specific quantum circuit multiple times gets different state vectors


jiannanWang:
### Environment

- **Qiskit Terra version**: 0.19.1
- **Python version**: 3.8.0
- **Operating system**: Ubuntu 22.04.3

### What is happening?

When running a specific quantum circuit multiple times, I got different state vector results. To the best of my knowledge, the state vector should not change given a fixed quantum circuit.

### How can we reproduce the issue?

The below code is a minimal reproduction example for the issue. It defines a quantum circuit with four gates, runs it 1,000 times, and counts the number of different state vector results. 
```
from qiskit import Aer, transpile, execute
from qiskit import QuantumCircuit, ClassicalRegister, QuantumRegister
from qiskit.circuit.library.standard_gates import *
from qiskit.circuit.add_control import add_control
from qiskit.quantum_info import Statevector
import numpy as np

qr = QuantumRegister(6, name='qr')
cr = ClassicalRegister(6, name='cr')
qc = QuantumCircuit(qr, cr, name='qc')
qc.append(ECRGate(), qargs=[qr[1], qr[0]], cargs=[])
qc.append(C4XGate().control(1), qargs=[qr[5], qr[2], qr[0], qr[3], qr[4], qr[1]], cargs=[])
qc.append(add_control(UGate(4.993516280749269, 0.48372007916715914, 3.4370453288464713), num_ctrl_qubits=1, label=None, ctrl_state=None), qargs=[qr[0], qr[1]], cargs=[])
qc.append(add_control(UGate(5.61106634998484, 4.625022873964695, 0.6622664050777831), num_ctrl_qubits=1, label=None, ctrl_state=None), qargs=[qr[0], qr[1]], cargs=[])

qc = transpile(qc, basis_gates=None, optimization_level=1, coupling_map=None)
simulator = Aer.get_backend('statevector_simulator')

def add_or_update_count_list(st_count_list, st):
    for i, (key_st, count) in enumerate(st_count_list):
        # if np.allclose(key_st, st):
        if key_st.equiv(st):
            st_count_list[i] = (key_st, count + 1)
            return
    st_count_list.append((st, 1))

def count_result_statevector(result_statevector_list: list):
    """Count the number of different values of result statevectors."""
    result_statevecotr_count = []
    for st in result_statevector_list:
        add_or_update_count_list(result_statevecotr_count, st)
    for i, (key_st, count) in enumerate(result_statevecotr_count):
[...]

PR #13177, opened 2024-09-18
remove internal `add_control`

1ucian0:
### Summary

The internal function ``add_control`` seems to create confusion among users. Here some examples:

 * https://github.com/Qiskit/qiskit/issues/13161
 * https://github.com/Qiskit/qiskit/issues/12625
 * https://github.com/Qiskit/qiskit/issues/11458
 * https://stackoverflow.com/a/59545389/8146172
 * https://stackoverflow.com/q/71108162/8146172

I think it can be renamed to mitigate that effect.


jakelishman:
I don't think we need to do this.  It's not in the public interface, and I think changing its name like this risks suggesting that naming is what defines the public interface.

If we _were_ going to do it, it'd probably be the module we renamed - the function itself _is_ public within ourselves as a package (evidenced by how it's used in more than one file), and it's the module that's private to users.

1ucian0:
I inlined/moved the code from `qiskit/circuit/add_control.py` to `qiskit/circuit/gate.py`. I assume refactoring the control thing  makes little sense because oxidation, right? Otherwise, I can have a look on that in follow up PR.

1ucian0:
I renamed the module (and simplified one of the tests so it does not import the private module).

qiskit/circuit/_add_control.py
Cryoris, line 1: Private modules is not really a thing so far, is it? Would it be more consistent to just rename the function to `_add_control`? 

I generally agree with Jake though that it's on the user not to use arbitrary, non-API documented functions -- but with the control plugin refactorings that @alexanderivrii is doing, this might look different in the future anyways, so I don't mind much doing this renaming.

releasenotes/notes/private-add_control-2367872c0f7136c4.yaml
```diff
@@ -0,0 +1,6 @@
+---
+upgrade_circuits:
+  - |
+    The internal function ``qiskit.circuit.add_control.add_control`` was removed and inlined, as it is not part of the
```
alexanderivrii: What do you mean by "inlined" here?

releasenotes/notes/private-add_control-2367872c0f7136c4.yaml
```diff
@@ -0,0 +1,6 @@
...
+    public API. User that had been wrongly using it as ``add_control(SomeGate(...), ...)`` should change their code to
```
alexanderivrii: "User" -> "Users"

alexanderivrii (commented):
LGTM (and I have double-checked that `add_control` is not mentioned anywhere else in the Qiskit code). Just a few minor comments about the release notes.

alexanderivrii (approved):
Thanks @jakelishman for improving the release note.

PR #13178, opened 2024-09-18
Fix code examples

frankharkins:
### Summary

Minor fixes to code examples in docstrings to make sure they run without error. Mostly adding imports and defining variables. Some syntax errors too.

These were identified as part of https://github.com/Qiskit/documentation/issues/1893.

### DIFF
--- a/qiskit/pulse/schedule.py
+++ b/qiskit/pulse/schedule.py
@@ -89,7 +89,6 @@ class Schedule:
 
       .. code-block:: python
 
-          from qiskit.pulse import Schedule, Gaussian, DriveChannel, Play
           sched = Schedule()
           sched += Play(Gaussian(160, 0.1, 40), DriveChannel(0)) << 30
 
@@ -97,7 +96,6 @@ class Schedule:
 
       .. code-block:: python
 
-          from qiskit.pulse import Schedule, Gaussian, DriveChannel, Play
           sched1 = Schedule()
           sched1 += Play(Gaussian(160, 0.1, 40), DriveChannel(0))
 
--- a/qiskit/pulse/schedule.py
+++ b/qiskit/pulse/schedule.py
@@ -81,6 +81,7 @@ class Schedule:
 
       .. code-block:: python
 
+          from qiskit.pulse import Schedule, Gaussian, DriveChannel, Play
           sched = Schedule()
           sched += Play(Gaussian(160, 0.1, 40), DriveChannel(0))
 
@@ -88,6 +89,7 @@ class Schedule:
 
       .. code-block:: python
 
+          from qiskit.pulse import Schedule, Gaussian, DriveChannel, Play
           sched = Schedule()
           sched += Play(Gaussian(160, 0.1, 40), DriveChannel(0)) << 30
 
@@ -95,6 +97,7 @@ class Schedule:
 
       .. code-block:: python
 
+          from qiskit.pulse import Schedule, Gaussian, DriveChannel, Play
           sched1 = Schedule()
           sched1 += Play(Gaussian(160, 0.1, 40), DriveChannel(0))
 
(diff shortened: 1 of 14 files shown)


qiskit/pulse/schedule.py
```diff
@@ -81,20 +81,23 @@ class Schedule:
...
           sched = Schedule()
           sched += Play(Gaussian(160, 0.1, 40), DriveChannel(0)) << 30
 
     - Merge two schedules
 
       .. code-block:: python
 
+          from qiskit.pulse import Schedule, Gaussian, DriveChannel, Play
           sched1 = Schedule()
           sched1 += Play(Gaussian(160, 0.1, 40), DriveChannel(0))
```
jakelishman: It doesn't need to be right now, but for the future if we're going to expand this, we should consider what the best user experience is - repeating the same `import`s lots of times feels like it's adding noise rather than clarity to me.

qiskit/quantum_info/operators/symplectic/sparse_pauli_op.py
```diff
@@ -1053,6 +1057,8 @@ def group_commuting(self, qubit_wise: bool = False) -> list[SparsePauliOp]:
 
                 .. code-block:: python
 
+                    >>> from qiskit.quantum_info import SparsePauliOp
+
```
jakelishman: Incredibly minor, but when we're pretending to be the base Python interpreter, there's no blank lines after each `>>>` line.

qiskit/transpiler/__init__.py
```diff
@@ -135,6 +137,7 @@
         ]
     )
 
+    pass_manager = PassManager()
 
     # Add pre-layout stage to run extra logical optimization
     pass_manager.pre_layout = logical_opt
```
jakelishman: I don't think this can be correct - `pass_manager` would need to be an instance of `StagedPassManager`, and more specifically one that already has a `layout` stage.

jakelishman (commented):
Thanks for looking at this Frank.  Longer term, I think we'll want to consider when it's appropriate to show imports to users - repeating the same imports several times in quick succession over the course of a run of examples probably feels like it could get noisy quickly.

Similarly, things like `import numpy as np` and `from qiskit.circuit import QuantumCircuit` probably happen enough that if they're the only imports, or if the only import is the class/function in question (whose documentation should already clearly imply how to import it), maybe it's best not to include the noise?

Anyway, that's for the future - doesn't need thinking about right now, particularly.

qiskit/pulse/schedule.py
frankharkins: Totally agree, I'll remove the second two imports.

qiskit/quantum_info/operators/symplectic/sparse_pauli_op.py
frankharkins: Oops, thanks!

qiskit/transpiler/__init__.py
frankharkins: Thanks! Would

```
pass_manager = StagedPassManager()
```

be correct? Or should we generate a preset pass manager?

EDIT: Gone with the latter but can change if needed

jakelishman: I suspect that `generate_preset_pass_manager` is really what was implied - I can't remember off the top of my head what the semantics of a stage-less `StagedPassManager` are with regards to the magic `pre_*` and `post_*` names.

ihincks (commented):
Thanks @frankharkins , all of the primitive blocks look correct to me

wshanks:
The pulse changes look okay to me, but keep in mind that pulse is being moved out of this repo (https://github.com/Qiskit/qiskit/pull/13164).

jakelishman (approved):
Thanks Frank

PR #13179, opened 2024-09-18
Use `.. plot::` directive to test code examples

### LINKED ISSUE #1893: Add doc testing to Python client repositories
It's important that the API docs code blocks do not break with code changes. We have too many docs to manually check that there are not issues, so we should automate it.

For Qiskit SDK, they don't want to run doctests in every PR because it's too slow. So they should have a weekly (nightly?) cron job, and also be able to manually trigger it.

qiskit-ibm-runtime and qiskit-ibm-transpiler might be fast enough to run every PR.

frankharkins:
### Summary

This PR switches code examples to use Matplotlib's `.. plot::` directive. This runs those code examples during documentation builds, failing the build if any code examples raise an exception.

### Details

We replace code blocks with the [plot directive](https://matplotlib.org/stable/api/sphinxext_plot_directive_api.html), using the `:nofigs:` option. Any code inside the directive is run at build time. This doesn't change the appearance of the rendered docs.

```rst
.. plot::
   :include-source:
   :nofigs:
```

We can also use the `:context:` option to share state between code examples. This is helpful to avoid duplicating imports and for interspersing code examples with prose.

```rst
.. plot::
   :include-source:
   :nofigs:
   :context: reset

.. plot::
   :include-source:
   :nofigs:
   :context:

   qc = QuantumCircuit(1)
```

**Advantages:**

* No new tools: we already use `.. plot::` in lots of places across the docs
* Flexibility: Used like this, the directive just checks the code runs, and won't fail if there are meaningless deviations in the output.

**Potential problems:**

* The term `plot` may be confusing to contributors since we're not plotting anything.
* It's a bit noisy repeating the same options (`:include-source:`, `:nofig:`) everywhere.

**Impact on CI:**

It's hard to tell since there's a lot of variation in the docs build step (20-30mins), but this doesn't seem to add noticeable time to CI (the two builds so far took 31 and 23 minutes).

***

Fixes https://github.com/Qiskit/documentation/issues/1893


qiskit/qpy/__init__.py
```diff
@@ -57,16 +62,35 @@
...
+    :context:
+
     with open('twenty_bells.qpy', 'rb') as fd:
         twenty_new_bells = qpy.load(fd)
 
+.. plot::
+    :nofigs:
+    :context:
+
+    from pathlib import Path
```
Eric-Arellano:
```suggestion
    # This block is hidden from readers. It's cleanup code.
    from pathlib import Path
```

qiskit/qpy/__init__.py
```diff
@@ -805,18 +829,26 @@
...
+        qpy.dump(schedule, fd)
 
     with open('schedule.qpy', 'rb') as fd:
-        new_qc = qpy.load(fd)[0]
+        new_schedule = qpy.load(fd)[0]
+
+.. plot::
+   :nofigs:
+
+   from pathlib import Path
```
Eric-Arellano, line 1034:
```suggestion
   # This block is hidden from readers. It's cleanup code.
   from pathlib import Path
```

qiskit/visualization/__init__.py
```diff
@@ -78,13 +79,22 @@
...
 
    plot_histogram(data, filename='new_hist.png')
 
    hist = plot_histogram(data)
    hist.savefig('new_hist.png')
 
+.. plot::
+   :nofigs:
+
+   from pathlib import Path
```
Eric-Arellano:
```suggestion
   # This block is hidden from readers. It's cleanup code.
   from pathlib import Path
```

Eric-Arellano (dismissed):
Great work! I like this approach. Sphinx is already such a learning curve that I think there's value in limiting the number of tools.

It's be useful to put the impact on CI's build from this change by looking at the Azure logs.

qiskit/circuit/classicalfunction/__init__.py
```diff
@@ -25,7 +25,7 @@
 how to synthesize a simple boolean function defined using Python into a
 QuantumCircuit:
 
-   .. code-block::
+  .. code-block:: python
```
jakelishman, line 28: There's a space remove here, but there's still leading indentation - is that what we expect, and is it displaying correctly?

qiskit/pulse/schedule.py
```diff
@@ -1476,14 +1509,35 @@ def assign_references(
...
+
+            # This code is hidden from readers
+            # It resets the variables so the following code example runs correctly
+            from qiskit import pulse
+            with pulse.build() as nested_prog:
+                pulse.delay(10, pulse.DriveChannel(0))
+            with pulse.build() as sub_prog:
+                pulse.reference("A")
+            with pulse.build() as main_prog:
+                pulse.reference("B")
```
jakelishman, line 1537: Nice - thanks for doing this.

qiskit/qobj/pulse_qobj.py
```diff
@@ -649,7 +649,7 @@ def to_dict(self):
         for use with IBMQ you can leverage a json encoder that converts these
         as expected. For example:
 
-        .. code-block::
+        .. code-block:: python
```
jakelishman, line 652: How come this one stayed as-is?

(Fwiw, `python` should be the default language for the Qiskit build, so I think it ought to have been highlighting correctly anyway.)

qiskit/qpy/__init__.py
```diff
@@ -57,16 +62,36 @@
...
         twenty_new_bells = qpy.load(fd)
 
+.. plot::
+    :nofigs:
+    :context:
+
+    # This block is hidden from readers. It's cleanup code.
+    from pathlib import Path
+    Path("bell.qpy").unlink()
+    Path("twenty_bells.qpy").unlink()
```
jakelishman: I'm not super wild about the docs build actually writing out dirty files into the working directory while we're going.

Could we either leave the above QPY stuff as `.. code-block`, or alternatively use a prior hidden `context` block to override `open` to be something that secretly returns `io.BytesIO()` objects (potentially wrapped to avoid their context managers clearing them up prematurely).  Obviously that's a complete Python crime if it were real code, but if the context objects are localised and don't leak beyond this file, it'll keep the docs build entirely in-memory, not touching the file-system with temporaries.

qiskit/transpiler/passes/scheduling/padding/pad_delay.py
```diff
@@ -39,7 +41,7 @@ class PadDelay(BasePadding):
 
     The ASAP-scheduled circuit output may become
 
-    .. parsed-literal::
+    .. code-block::
```
jakelishman:
```suggestion
    .. code-block:: text
```

qiskit/visualization/__init__.py
```diff
@@ -78,13 +79,23 @@
...
 
    hist = plot_histogram(data)
    hist.savefig('new_hist.png')
 
+.. plot::
+   :nofigs:
+
+   # This block is hidden from readers. It's cleanup code.
+   from pathlib import Path
+   Path("new_hist.png").unlink()
```
jakelishman: Similar thing here - it feels better to me not to have the docs build actually write out to the filesystem as a side effect that we have to tidy up.

jakelishman (commented):
Thanks for doing all this, and thanks for the care you took in putting in clean-up code.  That said, I'm still a bit nervous about merging any docs-build change that causes it to dirty the working directory while it's going, even if it's ephemeral - those stray files can easily get accidentally left behind, or the dev might cancel the docs build partway through, or so on.

Are the side-effect-y blocks something we can just leave as `.. code-block::`, or is there a neater way we might trick them into not actually writing to the filesystem?

qiskit/qobj/pulse_qobj.py
frankharkins: I think I wasn't sure how to make a pulse qobj and decided to leave it since this class is deprecated.

qiskit/circuit/classicalfunction/__init__.py
frankharkins: Docs builds before and after render identically (the conversion script currently ignores the specified language, see https://github.com/Qiskit/documentation/issues/2162). I think I removed the space to keep indentation a multiple of 2.

qiskit/qpy/__init__.py
frankharkins: Yeah that's a good point. The bytes object idea is a good one and I think it should work. Otherwise, I think it's fine not to test difficult cases like these.

EDIT: Switched to bytestream in 4370d5291acbdf874849fdc95cf9d98f86d8e286

qiskit/visualization/__init__.py
frankharkins: I agree, I think this one is simple enough that we don't need to test it.

ElePT (approved):
The PR looks good to me, I know it's already gone through a pretty extensive review. I think there's a chance new code examples have been added since the PR was last updated that would have to be included too, but I am fine merging the current PR as-is and leaving potentially new examples to a follow-up. Unless you'd prefer to do the check now. What do you think @frankharkins?

frankharkins:
Thanks! Let's merge now to get this initial exploration closed off

PR #13180, opened 2024-09-18
[WIP] Oxidize qubit tracker for HighLevelSynthesis

alexanderivrii:
### Summary

In #12911, we have added the `QubitTracker` class, which tracks the state of qubits for high-level synthesis. When `HighLevelSynthesis` runs non-trivially (i.e. does not take the "fast path", see #13070), this state tracking requires a large number of _ssllllooooww_ Python set-based manipulations.

As an example, the following code
```
nq = 1000
qubits = tuple(range(nq))
qt = QubitTracker(qubits, set(qubits), set())
for i in range(nq // 2):
    for j in range(nq // 2):
        qt.used([i, j])
```
runs in about `3.2` seconds (on my laptop). The current PR almost verbatim reimplements `QubitTracker` in Rust, with the example above now running in `0.11` seconds (i.e. about 30x faster).

It also adds a few Python tests for `QubitTracker`.

This does not require release notes since `QubitTracker` is an internal class and should not be exposed to the users.

This is still work-in-progress because I would like to further revise `QubitTracker`'s API and improve the internal data structures.


crates/accelerate/src/qubit_tracker.rs
```diff
@@ -0,0 +1,233 @@
...
+use crate::QiskitError;
+use hashbrown::HashMap;
+use hashbrown::HashSet;
+use pyo3::prelude::*;
+use qiskit_circuit::Qubit;
+
+/// Track qubits by their state
+#[pyclass]
+pub struct QubitTracker {
+    qubits: Vec<Qubit>,
```
mtreinish, line 22: I think a simple optimization you can make here is to use `num_qubits: u32` instead of a `Vec`. In rust the `Qubit` object is just an integer in `0..n` so it'll always be contiguous so we don't need to store the entire list and can just store the max index and then either do a comparison if it's in the circuit or an iterator `(0..self.num_qubits)` depending on how you were using `qubits` here.

alexanderivrii:
Closed in favor of #13369.

PR #13141, opened 2024-09-12
Oxidize `UnitarySynthesis` pass

ElePT:
### Summary
Closes #12210. This is an initial effort to port UnitarySynthesis to Rust, currently limited to the target + default plugin path (custom plugins will run in Python). 

### Details and comments
TODO:
- [x]  Add unit tests for qsd path and 1q synthesis path
- [ ] Address drawer issue after calling `dag_to_circuit` (not reflected in unit tests)
- [x] Clean up comments
- [x] Review use of `dag.push_back` and see if it can be done with the new `dag.apply_operation_back` method

PR #13183, opened 2024-09-18
Add `cache_pygates` feature to `accelerate` crate.

raynelfss:
### Summary
The following commits add the `cache_pygates` feature from the `circuit` crate into `accelerate`.

### Details and comments
Since we added rust native methods to add operations into the `DAGCircuit` (thanks in part to #13143), we make use of the `cache_pygate` feature which is only available in the `circuit` crate. A couple of transpiler passes which require building instances of `DAGCircuit` from Rust entirely (mainly #13141, and #13137) are using these functions, they need to respect the bounds this feature specifies or else they would not compile.

A couple of teammates have already implemented this flag as part of their PRs but, to prevent a future merge conflict debacle, I decided to put it in this PR. These commits add the feature to `accelerate` so that people don't have to worry about that in the future.

### DIFF
--- a/crates/accelerate/Cargo.toml
+++ b/crates/accelerate/Cargo.toml
@@ -58,3 +58,6 @@ features = ["ndarray"]
 [dependencies.pulp]
 version = "0.18.22"
 features = ["macro"]
+
+[features]
+cache_pygates = ["qiskit-circuit/cache_pygates"]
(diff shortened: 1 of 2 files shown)


mtreinish (approved):
LGTM this is a straightforward addition and one I've done in some local code that isn't quite ready for a PR so glad to spin it out here even if nothing is using it yet.

PR #13184, opened 2024-09-18
Don't run ElidePermutation if routing is disabled

### LINKED ISSUE #13144: ElidePermutation undoes routing
### Environment

- **Qiskit version**:  1.3 dev
- **Python version**: 3.10
- **Operating system**: mac

### What is happening?

Circuits that are mapped and routed can become unmapped unintentionally due to ElidePermutation treating inserted SWAPs as unwanted.

Suppose i ran some routing passes and got the following circuit on a line of 4 qubits. Then i just want to bring it down to some basis gates. Optimization level 2 and 3 (due to ElidePermutation) will undo the mapping.
```py
circuit_routed = QuantumCircuit(4)
circuit_routed.cx(0, 1)
circuit_routed.h(1)
circuit_routed.swap(1, 2)
circuit_routed.cx(2, 3)
print(circuit_routed)

pm = generate_preset_pass_manager(basis_gates=['cx', 'sx', 'rz'],  routing_method='none', optimization_level=2)
circuit_basis = pm.run(circuit_routed)

circuit_basis = pm.run(circuit_routed)
print(circuit_basis)
```

gives
```
q_0: ──■───────────────
     ┌─┴─┐┌───┐        
q_1: ┤ X ├┤ H ├─X──────
     └───┘└───┘ │      
q_2: ───────────X───■──
                  ┌─┴─┐
q_3: ─────────────┤ X ├
                  └───┘
global phase: π/4
                                                                             
Qubit(QuantumRegister(4, 'q'), 0) -> 0 ──■───────────────────────────────────
                                       ┌─┴─┐┌─────────┐┌────┐┌─────────┐     
Qubit(QuantumRegister(4, 'q'), 1) -> 1 ┤ X ├┤ Rz(π/2) ├┤ √X ├┤ Rz(π/2) ├──■──
                                       └───┘└─────────┘└────┘└─────────┘  │  
Qubit(QuantumRegister(4, 'q'), 2) -> 2 ───────────────────────────────────┼──
                                                                        ┌─┴─┐
Qubit(QuantumRegister(4, 'q'), 3) -> 3 ─────────────────────────────────┤ X ├
                                                                        └───┘
```

### What should happen?

routing_method='none' should disable ElidePermutation (or anything that wants to remove swaps)

### Any suggestions?

_No response_

mtreinish:
### Summary

We have a routing plugin named "none" which is used to assert that no routing should be performed (it errors if there is routing required). We were previously running the ElidePermutations pass in this case which is not the expected or correct behavior because it causes the same kind of permutation as routing and would eliminate and swap gates in the circuit. As the typical use case for routing_method='none' is when you've already routed a circuit (or constructed it by hand to match the connectivity constraints) we shouldn't be doing this. This commit fixes this behavior so if the routing_method is set to "none" we no longer run the ElidePermutations pass.

### Details and comments

Fixes #13144


jakelishman:
I'm certainly fine with this case, but for minor discussion, an extension: should we also disable `ElidePermutations` in the cases of all-to-all connectivity, or if VF2 can find that the existing interaction graph is perfect?

(My rough initial feeling to both is "no, leave it", especially since setting `routing_method="none"` is a way to disable it, but I feel like I could be convinced.)

mtreinish:
Yeah, I think for now we should just leave it, especially for a potential backport. 

I agree it's not exactly clear what we should do in the case of a `VF2Layout` match (which will always be the case for all to all connectivity). I can go either way on it I tend to think having an explicit opt-out path makes sense. But I also think setting `routing_method="none"` is not the best UI or the most discoverable if you have this kind of edge case. Maybe this can be improved with more explicit documentation on how to do this?

If we did want to only run elide permutations when there isn't a vf2 match, we'd have to move ElidePermutations into the `layout` stage and run it conditionally after `VF2Layout` but before `SabreLayout`. Which is definitely doable, but it adds complications around how we run `ConsolidateBlocks` in the init stage too, because of this I think the only path that makes sense is to run `VF2Layout` during the `init` stage  and the layout stage. We would run it before `ElidePermutations` in it's current place to ensure we don't elide an already routed circuit. But if there isn't a match we'd want to run all the init optimizations including `ElidePermutations` and `ConsolidateBlocks`. However doing that might result in the circuit having a perfect match so we would need to run `VF2Layout` again.

Writing all of that I think I've convinced myself we should just make it an opt-out thing with the routing_method plugin.

alexanderivrii:
Just a small thought: we probably should _enable_ ElidePermutations for all-to-all connectivity, as removing swap/permutation gates keeps the circuit routable yet makes it smaller.

jakelishman (dismissed):

jakelishman (approved):

PR #13186, opened 2024-09-19
Bug fix in ElidePermutations

alexanderivrii:
### Summary

A bug fix for the `ElidePermutations` pass where the internal qubit mapping was not updated correctly after a `PermutationGate`. This fix was already present in #13094, however we have decided to separate it to an independent PR as to allow backport.

### DIFF
--- a/test/python/transpiler/test_elide_permutations.py
+++ b/test/python/transpiler/test_elide_permutations.py
@@ -207,6 +207,27 @@ class TestElidePermutations(QiskitTestCase):
         res = self.swap_pass(qc)
         self.assertEqual(res, expected)
 
+    def test_partial_permutation_in_middle(self):
+        """Test with a permutation gate in the middle of a circuit,
+        with the permutation gate defined only on a subset of qubits.
+        """
+        qc = QuantumCircuit(5)
+        qc.cx(0, 1)
+        qc.append(PermutationGate([1, 2, 0]), [0, 2, 4])
+        qc.cx(2, 3)
+
+        # The permutation corresponding to the permutation gate maps
+        #   2 -> 0, 4 -> 2, 0 -> 4, and 1 -> 1, 3 -> 3.
+        # Instead of the permutation gate, we can relabel the qubits
+        #   0 -> 2, 1 -> 1, 2 -> 4, 3 -> 3, 4 -> 2.
+        # Hence cx(2, 3) becomes cx(4, 3).
+        expected = QuantumCircuit(5)
+        expected.cx(0, 1)
+        expected.cx(4, 3)
+
+        res = self.swap_pass(qc)
+        self.assertEqual(res, expected)
+
     def test_permutation_at_beginning(self):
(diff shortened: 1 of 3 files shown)


test/python/transpiler/test_elide_permutations.py
```diff
@@ -207,6 +207,27 @@ def test_permutation_in_middle(self):
...
+        #   2 -> 0, 4 -> 2, 0 -> 4, and 1 -> 1, 3 -> 3.
+        # Instead of the permutation gate, we can relabel the qubits
+        #   0 -> 2, 1 -> 1, 2 -> 4, 3 -> 3, 4 -> 2.
+        # Hence cx(2, 3) becomes cx(4, 3).
+        expected = QuantumCircuit(5)
+        expected.cx(0, 1)
+        expected.cx(4, 3)
+
+        res = self.swap_pass(qc)
+        self.assertEqual(res, expected)
```
mtreinish, line 233: Should we add some assertions on the permutation tracking too to make sure it's correct and reverses to the original circuit?

mtreinish (commented):
This LGTM, it was clearly a mistake like this because we ended up using the elision permutation tracking instead of the actual circuit index. I just had one small inline question about the test assertions, otherwise I think this is good to merge.

test/python/transpiler/test_elide_permutations.py
alexanderivrii: This is a good idea. I have extended the test in [6791002](https://github.com/Qiskit/qiskit/pull/13186/commits/67910028e26eb8e5a013853cb31036ff9b27cca2). Do you think it's a good idea to add more "random circuits" tests, with swaps and permutations?

mtreinish: Adding more tests wouldn't hurt. I think the one you added here provides sufficient coverage to prevent a regression here. But if you wanted to add more that would give us more coverage against potential future issues.

test/python/transpiler/test_elide_permutations.py
```diff
@@ -207,6 +207,37 @@ def test_permutation_in_middle(self):
...
+
+        # Make sure that the pass removes the permutation gate and remaps
+        # the following gate
+        self.assertEqual(res, expected)
+
+        # Make sure that the transpiled circuit *with* the final permutation
+        # is equivalent to the original circuit
+        perm = pass_.property_set["virtual_permutation_layout"].to_permutation(qc.qubits)
+        res.append(PermutationGate(perm), [0, 1, 2, 3, 4])
+        self.assertEqual(Operator(res), Operator(qc))
```
mtreinish, line 239: TBH, I'm surprised you didn't just use `Operator.from_circuit` here, but adding a permutation gate like this works fine too.

mtreinish (approved):
LGTM, thanks for updating the tests. If you want to add more test coverage in the future I think you can do that in a standalone PR. It'll be an easy PR to review/merge if all it does is expand the test coverage.

PR #13187, opened 2024-09-19
Make `qasm3.CustomGate` accessible from Python space

jakelishman:
### Summary

We previously exposed the data attribute `STDGATES_INC_GATES`, but its contents are all `CustomGate`, which can't be inspected programmatically from Python space.  This makes it so you can query _what_ the gates, their number of parameters, and how to construct them.

### Details and comments

From a feature request in #13061.

### DIFF
--- a/crates/qasm3/src/circuit.rs
+++ b/crates/qasm3/src/circuit.rs
@@ -65,12 +65,23 @@ register_type!(PyClassicalRegister);
 
 /// Information received from Python space about how to construct a Python-space object to
 /// represent a given gate that might be declared.
-#[pyclass(module = "qiskit._accelerate.qasm3", frozen, name = "CustomGate")]
+#[pyclass(
+    module = "qiskit._accelerate.qasm3",
+    frozen,
+    name = "CustomGate",
+    get_all
+)]
 #[derive(Clone, Debug)]
 pub struct PyGate {
+    /// A callable Python object that takes ``num_params`` angles as positional arguments, and
+    /// returns a :class:`~.circuit.Gate` object representing the gate.
     constructor: Py<PyAny>,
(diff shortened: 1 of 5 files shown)


crates/qasm3/src/circuit.rs
```diff
@@ -65,12 +65,23 @@ register_type!(PyClassicalRegister);
 
 /// Information received from Python space about how to construct a Python-space object to
 /// represent a given gate that might be declared.
-#[pyclass(module = "qiskit._accelerate.qasm3", frozen, name = "CustomGate")]
+#[pyclass(
+    module = "qiskit._accelerate.qasm3",
+    frozen,
+    name = "CustomGate",
+    get_all
```
ElePT, line 72: So `get_all` just adds getters for all attributes? Handy.

ElePT (approved):
Looks straightforward enough :)

crates/qasm3/src/circuit.rs
jakelishman: It does, though I'm kind of trying it out to see if I like it.  It feels a bit easy to accidentally encode poor access patterns with it - this PR already isn't ideal, since it directly exposes a Rust `String`, which means all access to `gate.name` will need to allocate a new Python `str`.  Unlikely to matter, but when it's not annotated directly on the member, I think it's easier to write non-ideal behaviour.

ElePT: Yeah I also see how this could easily lead to slip-ups (for example if someone added a new member and didn't notice the "blanket" getter). In this case I don't think it would be too bad but you can let me know if you want to change it.

jakelishman: I'm fine to just leave it if you're happy - this is a super simple dataclass-like thing anyway.

PR #11019, opened 2023-10-15
Reimplement two_qubit_decompose.num_basis_gates in rust

jlapeyre:
EDIT: This PR introduces dependency on [faer]( https://github.com/sarah-ek/faer-rs) which depends on rustc 1.67, which was released in January 2023. ~This would cause difficulties for some users who must use older versions of rustc. So this PR remains in draft mode at present.~

* faer is now compatible with Qiskit's MSRV #11493

This PR does two things
* Reimplement
https://github.com/Qiskit/qiskit/blob/9843fbfe7bf96f7acdda3cc5da3bea0028377efe/qiskit/quantum_info/synthesis/two_qubit_decompose.py#L1412-L1414
in rust

I looked at profiling with @mtreinish. It shows that a not-insignificant fraction of CPU time is spent here
* At the same time, introduce a dependency on the linear algebra library [faer]( https://github.com/sarah-ek/faer-rs). We are looking for a linear library to call from rust. `faer` is one of several options. A big advantage of `faer` is that it is pure rust so it incurs no additional build/package/distribute burden. The author prioritizes performance and some benchmarks look good. The project was started, or at least became public, less than a year ago.

    This PR uses `faer` compute a determinant and eigenvalues of a 4x4 unitary matrix.

Todo:
* Done ~~make `weyl_coordinates` return correct values. Or at least agree with current Python implementation.~~
* Done ~~make `num_basis_get` return the correct values. It is currently broken. I just got the code to compile, and have not debugged. Five tests fail if you do `tox -epy -- test.python.quantum_info.test_synthesis`. No other python tests fail.~~
* Done ~~Currently this PR copies numpy complex matrices element by element to faer matrices. This should not be required. The author of faer told me that a version of faer that supports efficient conversion should be released tomorrow. To be clear, the version released tomorrow should support reading numpy arrays without copying.~~
* Release notes, formatting, etc. will be done after the above is straightened out.
* There are some inefficiencies and inelegant things in this PR as stands. We may want to correct these now, or put this off till later.

This partially addresses #8774 . Some of the obvious-looking numerical code in two-qubit synthesis is not in fact a bottleneck. However, this PR would improve performance a bit.

PR #13193, opened 2024-09-19
Remove duplicated weyl_coordinates python function

### LINKED ISSUE #8459: Use of general eigenvalues solver in weyl_coordinates sometimes throws convergence error
It looks like the optimizations in #6896 can sometimes cause problems with convergence. @albertzhu01 found several cases that triggered this problem, although there seems to be some environment-dependence that makes it hard to reproduce across machines - the resulting almost-identity-matrix up to 1.E-16 scale deviations can cause error in lapack geev.

Any thoughts @jakelishman ?

---- from qiskit-experiments issue  https://github.com/Qiskit/qiskit-experiments/issues/846#issuecomment-1205772386_

The following test case reliably reproduces the error:
```python
import numpy as np
from scipy import linalg as la
mat = np.array(
    [
        [0.9999999999999998j, 0j, -3.469446951953614e-18j, 0j], 
        [0j, 0.9999999999999998j, (8.673617379884035e-19+2.6020852139652106e-18j), -6.938893903907228e-18j], 
        [-3.469446951953614e-18j, (8.673617379884035e-19+2.6020852139652106e-18j), 0.9999999999999998j, 0j], 
        [0j, -6.938893903907228e-18j, 0j, 0.9999999999999998j]
    ]
)
la.eigvals(mat)
```
Output:
```
LinAlgError                               Traceback (most recent call last)
Input In [17], in <cell line: 9>()
      1 from scipy import linalg as la
      2 mat = np.array(
      3     [
      4         [0.9999999999999998j, 0j, -3.469446951953614e-18j, 0j], 
   (...)
      7         [0j, -6.938893903907228e-18j, 0j, 0.9999999999999998j]]
      8 )
----> 9 la.eigvals(mat)

File ~/opt/anaconda3/envs/pygsti2/lib/python3.8/site-packages/scipy/linalg/_decomp.py:880, in eigvals(a, b, overwrite_a, check_finite, homogeneous_eigvals)
    811 def eigvals(a, b=None, overwrite_a=False, check_finite=True,
    812             homogeneous_eigvals=False):
    813     """
    814     Compute eigenvalues from an ordinary or generalized eigenvalue problem.
    815 
   (...)
    878 
    879     """
--> 880     return eig(a, b=b, left=0, right=0, overwrite_a=overwrite_a,
    881                check_finite=check_finite,
    882                homogeneous_eigvals=homogeneous_eigvals)

File ~/opt/anaconda3/envs/pygsti2/lib/python3.8/site-packages/scipy/linalg/_decomp.py:247, in eig(a, b, left, right, overwrite_a, overwrite_b, check_finite, homogeneous_eigvals)
    244     w = wr + _I * wi
    245     w = _make_eigvals(w, None, homogeneous_eigvals)
--> 247 _check_info(info, 'eig algorithm (geev)',
    248             positive='did not converge (only eigenvalues '
    249                      'with order >= %d have converged)')
    251 only_real = numpy.all(w.imag == 0.0)
    252 if not (geev.typecode in 'cz' or only_real):

File ~/opt/anaconda3/envs/pygsti2/lib/python3.8/site-packages/scipy/linalg/_decomp.py:1356, in _check_info(info, driver, positive)
   1353     raise ValueError('illegal value in argument %d of internal %s'
   1354                      % (-info, driver))
   1355 if info > 0 and positive:
-> 1356     raise LinAlgError(("%s " + positive) % (driver, info,))

LinAlgError: eig algorithm (geev) did not converge (only eigenvalues with order >= 2 have converged)
```

_Originally posted by @albertzhu01 in https://github.com/Qiskit/qiskit-experiments/issues/846#issuecomment-1205772386_

mtreinish:
### Summary

This commit removes the weyl_coordinates() function from the private qiskit.synthesis.two_qubit.weyl module. This function's internal use was ported to rust as part of #11019 but it left the python implementation intact while we ensured the rust implementation was reliable longer term. Since then we've ported the majority of the two qubit synthesis to rust now and the only usage of this python implementation was the unit tests. This commit removes the python implementation and the entire internal weyl module as nothing uses it anymore. A python interface is added to the rust function and the tests are updated to call that instead.

### Details and comments

Fixes: #8459


crates/accelerate/src/two_qubit_decompose.rs
```diff
@@ -216,6 +216,22 @@ fn py_decompose_two_qubit_product_gate(
...
+///     np.ndarray: Array of the 3 Weyl coordinates.
+#[pyfunction]
+fn weyl_coordinates(py: Python, unitary: PyReadonlyArray2<Complex64>) -> PyObject {
+    let array = unitary.as_array();
+    __weyl_coordinates(array.into_faer_complex())
+        .to_vec()
+        .into_pyarray_bound(py)
+        .into()
+}
+
```
jakelishman, line 234: Did we consider this public interface in any way?  If not: is there any chance it might get called with matrices whose dtypes are not `complex`?  I'm asking because the Python code would have happily lifted the matrix to be `complex` early in the process (during the translation to the magic basis), but this will now just error immediately if anyone's doing things like `weyl_coordinates(np.eye(2))`.

ShellyGarion: Except of Jake's question, this PR looks good to me.

mtreinish: I didn't think that we considered this API public in any way. It was previously used internally for some things but was replaced by rust internals in 1.0 and 1.1. Right now it's not documented anywhere nor is the deleted internal module and this function is only used for tests now. The public interface for getting the weyl coordinates would be to use the `TwoQubitWeylDecomposition` and access the coordinate attributes. That being said if you think it's important I can change the input type to handle the casting here.

ShellyGarion (approved):

PR #13196, opened 2024-09-20
Prepare Qiskit 1.2.2

jakelishman:


mtreinish (approved):
We technically don't need to bother moving the release notes to the `1.2/` directory. I typically prefer to keep them in the root path because it makes it easier to filter for potential duplicates in the next minor version release. But it doesn't actually matter and otherwise this LGTM.

PR #13197, opened 2024-09-20
Fix future rustc lint rule on named lifetime elision

mtreinish:
### Summary

This commit fixes a future lint rule that will appear in future version of the rust compiler. The new rule will raise a warning at compile time if the code used an elided lifetime resolves to a named lifetime. While this typically would only cause an issue in unsafe rust, it can cause a confusing error message in cases when using functions like this in a safe context. This fixes the code violating this future rule to get ahead of any potential errors when compiling with a release of Rust from the near future.

This rule is documented here:

https://doc.rust-lang.org/nightly/nightly-rustc/rustc_lint/builtin/static.ELIDED_NAMED_LIFETIMES.html

### Details and comments


jakelishman (approved):
I'm unclear on why I didn't use separate lifetimes for `self` and the GIL, but this seems fine - the problematic lifetime is _within_ the bound, not the reference to the `Bound`, which clearly derives from `self`.

PR #13202, opened 2024-09-20
Skip uninstallable tags in QPY backwards compatibility tests

jakelishman:
### Summary

When a tag has been made, but the package has not yet landed on PyPI, the QPY job fails in the environment-building step.  This is not actually a failure of the QPY backwards-compatibility guarantees, and it isn't the job of the QPY tests to detect a bad tag anyway.

### Details and comments

Since this commit is an acknowledgement that not every tag might be a valid version, and my editor was complaining about the safety of the various script calls, I made the variable expansion safe against word splitting, just in case.


test/qpy_compat/process_version.sh
```diff
@@ -43,21 +43,27 @@ fi
 
 if [[ ! -d qpy_$version ]] ; then
     echo "Building venv for qiskit-terra $version"
-    python -m venv $version
+    python -m venv "$version"
     if [[ ${parts[0]} -eq 0 ]] ; then
-        ./$version/bin/pip install -c qpy_test_constraints.txt "qiskit-terra==$version"
+        if ! "./$version/bin/pip" install -c qpy_test_constraints.txt "qiskit-terra==$version"; then
```
mtreinish: I worry about this masking real issues because of a spurious network failure or something. This will make a failure to install a historical version a skip instead of a failure which has the risk of glossing over real issues. Especially if we were to introduce some incompatibility with that version. I'm not sure if there is a better way to detect a missing version on pypi though, the only thing I can think of is maybe querying the pypi api at: https://pypi.org/simple/qiskit and https://pypi.org/simple/qiskit-terra/

mtreinish (commented):
I'm a bit concerned about incorrectly skipping something we shouldn't and that masking our coverage for a version. I feel like we need to be a bit more intentional about when we're skipping as it should only be if we have yet to upload the package to pypi not for any other installation failure. The other changes here look fine to me though.

test/qpy_compat/process_version.sh
jakelishman: Most of the time we don't actually entire this block at all, because we cache the historical environments, so this should largely only trigger when we're in the intermediate period when a version is partially uploaded, and the environments haven't been created yet.

I'm fine if we want to have some other detection that a release we uploaded to PyPI fails to install, but I don't think it should be part of the QPY tests, with them being in the all-PRs merge pipeline.

mtreinish: I've seen a fair number of network issues causing QPY test failures, the caching doesn't seem to work very reliably. My concern here is in the presence of one of those network failures instead of failing we'd skip that version and then lose the coverage the testing is supposed to provide. That opens up an opportunity for a regression/backwards incompatibility to slip in.

jakelishman: That's fair enough.  I'm ok pivoting this PR to make a more informed decision about which sets of versions it's even going to attempt to install - we already do a bit of that with the pre-releases, and it's easy enough as you say to turn that logic into a slightly fancier script that asks PyPI for the list of versions (and limit it to those that have the Linux wheel), rather than us trying to infer it from the repository tags.

jakelishman: Ok, this should now hopefully be done correctly in 394b6c1.

test/qpy_compat/get_versions.py
```diff
@@ -0,0 +1,95 @@
...
+            data = json.load(fd)
+        min_version = min_version and packaging.version.parse(min_version)
+        max_version = max_version and packaging.version.parse(max_version)
+        for other_version, payload in data["releases"].items():
+            other_version = packaging.version.parse(other_version)
+            if min_version is not None and other_version < min_version:
+                continue
+            if max_version is not None and other_version >= max_version:
+                continue
+            if other_version > our_version or other_version.is_prerelease:
```
mtreinish: Is this a change in logic from before? IIRC we were skipping `a` and `b` prerelease labels but not `rc`. The `is_prerelease` attribute will return true for an RC release too.

test/qpy_compat/get_versions.py
```diff
@@ -0,0 +1,95 @@
...
+        ("qiskit-terra", version)
+        for version in available_versions_for_package("qiskit-terra", "0.18.0", "1.0.0")
+    )
+    yield from (
+        ("qiskit", version) for version in available_versions_for_package("qiskit", "1.0.0")
+    )
+
+
+def main():
+    """main"""
```
mtreinish, line 93: Hah, I like this docstring

mtreinish (commented):
Overall this looks fine to me. I personally find the new variable naming for the venvs and python in the bash scripts a bit harder to parse, but that's probably just how I think about working with this things is a bit more explicit in file paths. I only left one inline question about a change in behavior around rc releases. I'm fine if that's the direction you want to move here, but since it is a change I wanted to make sure it was intentional.

test/qpy_compat/get_versions.py
jakelishman: passive aggressive suppression of the pylint warning haha

test/qpy_compat/get_versions.py
jakelishman: Yeah, there's a small change in logic - that was a mistake.  Fixed in bf34ad5e12.

jakelishman:
For the bash scripts venv renaming: the main problem for me locally was that they were all making a lot of strong assumptions about the working directory they were executed from, and one of them was making assumptions about the existence of a Python venv at a particular relative path, which meant I had to keep recreating the entire CI setup in order to test a single script.  I generally agree the changes make it not as pleasant to read, but they decouple the scripts from particular working directories, and let me pass more information explicitly into each script, so it was easier to run things locally.

mtreinish (approved):

jakelishman:
@Mergifyio backport stable/1.2 stable/0.46

PR #13213, opened 2024-09-24
Bump macos version in wheels gha workflow 

eliarbel:
During the 1.2.0rc1 release process we had a failure to build the x86_64 macOS jobs. This was because the job was still trying to use the macos-11 runner which is no longer supported. [1] This commit fixes this by bumping the version in the wheel job to macos-12.

[1] https://github.blog/changelog/2024-05-20-actions-upcoming-changes-to-github-hosted-macos-runners/

(cherry picked from commit 6663db17a1b87e3101368002f592d37a369048a4)


eliarbel:
The QPY test fails because we already have tagged `0.46.3` which `test/qpy_compat/process_version.sh` tries to pip install. 
There is a fix here: #13202

mtreinish (approved):

PR #13214, opened 2024-09-24
Add some more gates to CollectCliffords pass

### LINKED ISSUE #13138: `CollectCliffords` pass ignores Rz's
### Environment

- **Qiskit version**: 1.2.0
- **Python version**: 3.12.3
- **Operating system**: MacOS

### What is happening?

`rz` gates are ignored by the `CollectCliffords` pass (regardless of angle), when it could potentially collect them.

```python
from qiskit.transpiler.passes import CollectCliffords

qc = QuantumCircuit(1)
qc.rz(0, 0)
qc.sx(0)

PassManager([CollectCliffords(min_block_size=1)]).run(qc).draw()
```

### What should happen?

There should be a single instruction in the resulting circuit, however I get

```
   ┌───────┐┌──────────┐
q: ┤ Rz(0) ├┤ Clifford ├
   └───────┘└──────────┘
```

### Any suggestions?

It seems [this function](https://github.com/Qiskit/qiskit/blob/56a5352c94bfa695073557a582ad5b6d0bb8815c/qiskit/transpiler/passes/optimization/collect_cliffords.py#L80-L82) is just looking at `_BASIS_1Q` which doesn't include `rz` (and shouldn't, IMO).

Presumably this issue extends to other parameterized gates.

A possible fix would be to instead have an option in `CollectCliffords` which attempts to cast it as a `Clifford`. Since this would likely come at a performance hit, it should probably be `False` by default.

E.g.

```python
def _is_clifford_gate(node, try_harder: bool = False):
    """Specifies whether a node holds a clifford gate."""
    if getattr(node.op, "condition", None) is not None:
        return False
    if node.op.name in clifford_gate_names:
        return True
    
    if not try_harder:
        return False
    
    try:
        Clifford(node.op)
        return True
    except:
        return False
```

ShellyGarion:
### Summary
Close #13138 
Add an argument `matrix_based` to the `CollectCliffords` transpiler pass.
If the new parameter `matrix_based=True`, the `CollectCliffords` transpiler pass can collect `Rz(pi/2)` gates and other unitary gates that are Clifford gates for certain parameters. 

### Details and comments
The PR is based on the suggestion of @samanthavbarron in #13138 and discussions with @alexanderivrii


releasenotes/notes/add-gates-to-collect-clifford-af88dd8f7a2a4bf9.yaml
```diff
@@ -0,0 +1,8 @@
+---
+features_transpiler:
+  - |
+    Add an argument ``matrix_based`` to the :class:`.CollectCliffords()` transpiler pass.
+    If the new parameter ``matrix_based=True``, the :class:`.CollectCliffords()` transpiler pass
+    can collect :class:`.RZGate(np.pi/2)` gates and other unitary gates that are :class:`.Clifford()`gates
+    for certain parameters.
```
alexanderivrii:
```suggestion
    can collect :class:`.RZGate(np.pi/2)` gates and other unitary gates that are :class:`.Clifford()` gates
    for certain parameters.
```
Note the extra space before "gates" -- I believe this is what's responsible for CI failures :).

ShellyGarion: good catch!
The CI failure is not related. it's due to QPY backwards tests. should be fixed after #13202 is merged.

samanthavbarron:
Awesome, thanks for doing this! :)

alexanderivrii (approved):
LGTM!

Issue #13219, opened 2024-09-24
Stop using OnceCell for `PackedInstruction.py_op` field


mtreinish:
### What should we add?

When the `cache_pygates` feature is enabled, the `py_op` field in the `PackedInstruction` struct stores a `OnceCell<PyObject>` right now. See: https://github.com/Qiskit/qiskit/blob/dcd41e9fcedd6f51fe725c25c4ed583c22bb3525/crates/circuit/src/packed_instruction.rs#L506-L517 this is used a cached python object for the underlying instruction so we don't repeatedly rebuild the objects when returning to python. However the use of a `OnceCell` in this field is preventing us from enabling the pygate cache and writing multithreaded transpiler passes as it is not thread-safe. The logical alternative is `GilOnceCell` which uses the GIL as a lock for a threadsafe version, but it isn't cloneable (which is commented on in the code). The other alternative is `OnceLock` but that likely will add some overhead to the pygate cache as we're adding an additional locking mechanism.

PR #13220, opened 2024-09-24
Fully port BarrierBeforeFinalMeasurements to rust

### LINKED ISSUE #12253: Port `BarrierBeforeFinalMeasurements` to Rust

mtreinish:
### Summary

This commit migrates the BarrierBeforeFinalMeasurements transpiler pass to operate fully in Rust. The full path of the transpiler pass now never leaves Rust until it has finished modifying the DAGCircuit. The one exception is when the `label` is not set then we still call `MergeAdjacentBarriers` in the python code of the pass.

This is the first step in the improvement of the performance of this pass. We can easily leverage multhreading to potentially parallelize the analysis portion of this pass that searches for the final operations and returns the set of indices. But this is blocked on #13219 which prevents us from accessing the PackedInstructions stored in the DAGCircuit in a multithreaded context.

This commit also fixes an issue related to shared references in the disjoint_utils module around barrier labels. The combine_barriers() function was incorrectly mutating the label by reference which wouldn't persist in the DAG, and this was causing failures after the barrier was originally generated in Rust with this pass now.

### Details and comments

Fixes #12253

### DIFF
--- a/crates/accelerate/src/barrier_before_final_measurement.rs
+++ b/crates/accelerate/src/barrier_before_final_measurement.rs
@@ -1,0 +1,118 @@
+// This code is part of Qiskit.
+//
+// (C) Copyright IBM 2024
+//
+// This code is licensed under the Apache License, Version 2.0. You may
+// obtain a copy of this license in the LICENSE.txt file in the root directory
+// of this source tree or at http://www.apache.org/licenses/LICENSE-2.0.
+//
+// Any modifications or derivative works of this code must retain this
+// copyright notice, and modified files need to carry a notice indicating
+// that they have been altered from the originals.
+
+use hashbrown::HashSet;
+use pyo3::prelude::*;
+use rustworkx_core::petgraph::stable_graph::NodeIndex;
+
+use qiskit_circuit::circuit_instruction::ExtraInstructionAttributes;
+use qiskit_circuit::dag_circuit::{DAGCircuit, NodeType};
+use qiskit_circuit::imports::BARRIER;
+use qiskit_circuit::operations::{Operation, PyInstruction};
+use qiskit_circuit::packed_instruction::{PackedInstruction, PackedOperation};
+use qiskit_circuit::Qubit;
+
+static FINAL_OP_NAMES: [&str; 2] = ["measure", "barrier"];
+
+#[pyfunction]
+pub fn barrier_before_final_measurements(
+    py: Python,
+    dag: &mut DAGCircuit,
+    label: Option<String>,
+) -> PyResult<()> {
+    let final_ops: HashSet<NodeIndex> = dag
+        .op_nodes(true)
+        .filter(|node| {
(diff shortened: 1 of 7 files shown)


mtreinish:
I ran the dedicated asv microbenchmarks on this and it yielded:

```

| Change   | Before [1344cddb] <stash~1>   | After [f62a6d79] <port-barrier-before-final-measure>   |   Ratio | Benchmark (Parameter)                                                  |
|----------|-------------------------------|--------------------------------------------------------|---------|------------------------------------------------------------------------|
| -        | 17.1±0.1ms                    | 12.3±0.1ms                                             |    0.72 | passes.PassBenchmarks.time_barrier_before_final_measurements(14, 1024) |
| -        | 7.42±0.03ms                   | 5.12±0.1ms                                             |    0.69 | passes.PassBenchmarks.time_barrier_before_final_measurements(5, 1024)  |
| -        | 29.7±0.2ms                    | 8.93±0.2ms                                             |    0.3  | passes.PassBenchmarks.time_barrier_before_final_measurements(20, 1024) |

PERFORMANCE INCREASED.
```

crates/accelerate/src/barrier_before_final_measurement.rs
```diff
@@ -0,0 +1,118 @@
...
+    let new_barrier_py_inst = PyInstruction {
+        qubits: dag.num_qubits() as u32,
+        clbits: 0,
+        params: 0,
+        op_name: "barrier".to_string(),
+        control_flow: false,
+        #[cfg(feature = "cache_pygates")]
+        instruction: new_barrier.clone().unbind(),
+        #[cfg(not(feature = "cache_pygates"))]
+        instruction: new_barrier.unbind(),
```
ElePT, line 83: What is the difference between doing this or simply omitting the `#[cfg(feature = "cache_pygates")]` comment and setting the instruction argument to `new_barrier.unbind()`?

ElePT (commented):
Thanks for taking over this pass. It looks pretty straightforward, I only had one small question.

crates/accelerate/src/barrier_before_final_measurement.rs
mtreinish: When `cache_pygates` is necessary we need to clone `new_barrier` because it's used in the `PyInstruction` here and also again in the pycache for the `PackedInstruction` that we insert into the dag. I used the cfg macros here to conditionally compile this so we only clone if the `cache_pygates` feature is enabled and otherwise skip the `clone()` where it's not needed. This would work fine if I removed the macros and just had it always do `new_barrier.clone().unbind()` regardless of whether the feature was enabled, but it was just me micro-optimizing a little as the `clone()` and `Drop` for `new_barrier` as a `Bound<PyAny>` is just python ref counting and very cheap, but I figured I could just save the overhead of that  here.

ElePT (approved):
Looks good!

PR #13222, opened 2024-09-25
unwrapped DAGCircuit::collect_runs() to not return an Option

### LINKED ISSUE #13109: Change `DAGCircuit::collect_runs()` return to not be an `Option`
> It looks like `DAGCircuit::collect_runs` returns an `Option` only because it directly returns the result of the Rustworkx core function, which uses `None` to represent a cycle in the graph (...which is questionable lol). 

> Perhaps it is beyond the scope of this PR, but it might be nice if you can change `DAGCircuit::collect_runs` to unwrap that option internally. If it fails, it should be on `DAGCircuit` for getting itself into an invalid state.

_Originally posted by @kevinhartman in https://github.com/Qiskit/qiskit/pull/13013#discussion_r1744485781_

To fix this issue we should update the `collect_runs()` method in `crates/circuit/src/dag_circuit.rs` to internally unwrap the return from rustworkx and not return an option.

cameron-d28:
### Summary
This change updates the collect_runs function to return a PyResult instead of an Option so that error can be routed by to DAGCircuit when invalid states are present.

### Details and comments
fixes #13109

### DIFF
--- a/crates/circuit/src/dag_circuit.rs
+++ b/crates/circuit/src/dag_circuit.rs
@@ -4948,7 +4948,7 @@ impl DAGCircuit {
     pub fn collect_runs(
         &self,
         namelist: HashSet<String>,
-    ) -> impl Iterator<Item = Vec<NodeIndex>> {
+    ) -> impl Iterator<Item = Vec<NodeIndex>> + '_ {
         let filter_fn = move |node_index: NodeIndex| -> Result<bool, Infallible> {
             let node = &self.dag[node_index];
             match node {
--- a/crates/circuit/src/dag_circuit.rs
+++ b/crates/circuit/src/dag_circuit.rs
@@ -4375,25 +4375,18 @@ def _format(operand):
         for name in namelist.iter() {
             name_list_set.insert(name.extract::<String>()?);
         }
-        match self.collect_runs(name_list_set) {
-            Ok(runs) => {
-                let run_iter = runs.map(|node_indices| {
-                    PyTuple::new_bound(
-                        py,
-                        node_indices
-                            .into_iter()
-                            .map(|node_index| self.get_node(py, node_index).unwrap()),
-                    )
-                    .unbind()
-                });
-                let out_set = PySet::empty_bound(py)?;
-                for run_tuple in run_iter {
-                    out_set.add(run_tuple)?;
-                }
-                Ok(out_set.unbind())
-            }
-            Err(e) => Err(PyRuntimeError::new_err(e.to_string())),
+
+        let out_set = PySet::empty_bound(py)?;
+
+        for run in self.collect_runs(name_list_set) {
+            let run_tuple = PyTuple::new_bound(
+                py,
+                run.into_iter()
+                    .map(|node_index| self.get_node(py, node_index).unwrap()),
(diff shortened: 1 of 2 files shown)


cameron-d28:
should i update all calls to collect runs? @mtreinish

crates/circuit/src/dag_circuit.rs
```diff
@@ -4967,8 +4965,13 @@ impl DAGCircuit {
...
         };
-        rustworkx_core::dag_algo::collect_runs(&self.dag, filter_fn)
-            .map(|node_iter| node_iter.map(|x| x.unwrap()))
+
+        match rustworkx_core::dag_algo::collect_runs(&self.dag, filter_fn) {
+            Some(iter) => Ok(iter.filter_map(|result| result.ok())),
+            None => Err(PyRuntimeError::new_err(
+                "Invalid DAGCircuit, cycle encountered",
+            )),
+        }
```
```suggestion
        match rustworkx_core::dag_algo::collect_runs(&self.dag, filter_fn) {
            Some(iter) => iter.map(|result| result.unwrap()),
            None => panic!("invalid DAG: cycle(s) detected!"),
        }
```

I would recommend that we instead `panic!()` in the case that the `DAGCircuit` contains a cycle. If it does, then there's definitely a bug somewhere in one of `DAGCircuit`'s other methods that led to it corrupting itself. Such a failure is non-recoverable since we'd have no way of knowing what went wrong (if we did, then we wouldn't have introduced the bug!).

Separately, while `filter_map` is semantically equivalent to the original use of `map(|x| x.unwrap())` in this case, it's comparatively a bit misleading. The only way that the inner result can fail is if `filter_fn` defined above could return an error (which it cannot, since its `Error` type is the nonconstructible `Infallible` type). So the `filter_map` would never actually be filtering anything 🙂.

crates/circuit/src/dag_circuit.rs
```diff
@@ -4957,7 +4955,7 @@ impl DAGCircuit {
     pub fn collect_runs(
         &self,
         namelist: HashSet<String>,
-    ) -> Option<impl Iterator<Item = Vec<NodeIndex>> + '_> {
+    ) -> PyResult<impl Iterator<Item = Vec<NodeIndex>> + '_> {
```
```suggestion
    ) -> impl Iterator<Item = Vec<NodeIndex>> {
```

This method shouldn't need a result type after the comment below has been addressed.

kevinhartman (commented):
Thanks for the contribution, @cameron-d28!

This looks like a good start. I've left a few comments inline, primarily regarding the way we ought to handle the cycle error case. After addressing that, you'll need to adjust `py_collect_runs` as well for the new return type.

cameron-d28:
thank you for the feedback @kevinhartman! i have improved my changes. please let me know how they look 🙂

kevinhartman (approved):
Thanks for the updates, @cameron-d28!

This is looking good to me now. I like the change you made to use a single for-loop rather than a map followed by a for-loop when adding runs to the `out_set` in `DAGCircuit::py_collect_runs`.

cameron-d28:
glad to see the update made sense @kevinhartman! i didn't see why the mapping would still be needed - (maybe for lazy eval?? not sure).

PR #13223, opened 2024-09-25
Deprecate the condition attribute and related functionality

### LINKED ISSUE #9556: Deprecate `Instruction.condition` in favor of `IfElseOp`
### What should we add?

`Instruction.condition` is Qiskit's way of representing classical execution of instructions based on the values in specified classical registers. However, due to it's location as a property on `Instruction` (but outside `params`) has historically lead to it being overlooked, resulting in a handful of bugs where this condition information is either lost or not considered e.g. when evaluating instruction order (Some examples: #6475, #6810, #8553, #8040, #7950, #7247, #7006, #6994, #6583, #6192, #6016, #4727, #4672, #3164, #3215, #2667, #2571 .) Additionally, it is one of the pieces of mutable state that is currently preventing us from moving toward more lightweight `Operation` instances ( to resolve e.g. #3800, #5895, #7624, and  #9372 ).

`IfElseOp` can currently represent a superset of what can be expressed with `Instruction.condition` and is more inline with how we expect to handle conditional and control flow operations going forward, so we should plan to deprecate `Instruction.condition` with equivalent usage of `IfElseOp`. This issue is to discuss and plan that process.

Deprecating `Instruction.condition`while maintaining full backwards compatibility seems difficult (due to both a change in `Operation` type from `Instruction` to `IfElseOp`,  and  due to `.condition` being a property of the `Instruction` without context of the `QuantumCircuit` in which it is included). The following are two use cases to consider:

```python
>>> qc = qk.QuantumCircuit(2,1)
>>> qc.h(0)
>>> qc.measure(0,0)
>>> qc.x(1).c_if(0, 0)
```

```python
qc = qk.QuantumCircuit(1, 1)
qc.x(0)
qc.data[0][0].condition = (qc.clbits[0], 1)
````

Additionally, we should:
1)Updating existing visualization to inspect `IfElseOp` operations and, when possible, draw as an equivalent `condition`
2)Updating `qiskit.assemble`, similarly to visualization, to replace applicable `IfElseOp` s with their equivalent `condition` representation

### PR DESCRIPTION
**Pull request by mtreinish**
### Summary

This commit deprecates the Instruction.condition and c_if() method for removal in 2.0. This functionality has been superseded by the more sophisiticated `IfElseOp` for some time now. Removing the condition property will simplify the Qiskit data model as it is implemented in a kind of ad-hoc manner that adds additional overhead to manually check it in many places.

For the unittest modifications the deprecation warning for the .condtion getter is suppressed for the entire suite because this gets triggered internally by the transpiler and a lot of other places, including from rust code as until it is removed we need to use it to check that piece of the data model.

### Details and comments

Fixes #9556

### CODE REVIEW

File: qiskit/circuit/instruction.py [hunk 1]
```diff
@@ -160,6 +160,7 @@ def to_mutable(self):
         return self.copy()
 
     @property
+    @deprecate_func(since="1.3.0", removal_timeline="in 2.0.0", is_property=True, stacklevel=2)
```
ElePT: I believe that 2 is the default stack level, right? It doesn't hurt but it made me think for a second, so maybe:
```suggestion
    @deprecate_func(since="1.3.0", removal_timeline="in 2.0.0", is_property=True)
```

releasenotes/notes/deprecate-condition_c_if-9548e5522814fe9c.yaml
```diff
@@ -0,0 +1,38 @@
+---
+deprecations_circuits:
+  - |
+    Deprecated the :attr:`.Instruction.condition` and the
+    :meth:`.Instruction.c_if` method. This attribute and method will be removed
+    in Qiskit 2.0 and along with any other uses of this in the Qiskit data
+    model. This functionality has been superseded by the :class:`.IfElseOp`
+    which can be used to describe a classically condition in a circuit. For
+    example, a circuit using :attr:`.Instruction.condition` like::
```
ElePT:
```suggestion
    Deprecated the :attr:`.Instruction.condition` attribute and the
    :meth:`.Instruction.c_if` method. They will be removed
    in Qiskit 2.0, along with any uses in the Qiskit data
    model. This functionality has been superseded by the :class:`.IfElseOp` class
    which can be used to describe a classical condition in a circuit. For
    example, a circuit using  :meth:`.Instruction.c_if` like::
```

releasenotes/notes/deprecate-condition_c_if-9548e5522814fe9c.yaml
```diff
@@ -0,0 +1,38 @@
...
+        from qiskit.circuit import QuantumCircuit
+
+        qc = QuantumCircuit(2, 2)
+        qc.h(0)
+        qc.x(0).c_if(0, 1)
+        qc.z(1.c_if(1, 0)
+        qc.measure(0, 0)
+        qc.measure(1, 1)
+
+    can be rewritten using as::
```
ElePT:
```suggestion
    can be rewritten as::
```

ElePT (commented):
Wow, this PR is long. I didn't get to review all tests but I left a few minor suggestions that I saw in the source code/reno.

qiskit/circuit/instruction.py
mtreinish: Oh good call, this was leftover from me playing around with the deprecation warning in different settings. We end up accessing this in a few different ways internally and I wanted to see what was user visible or not depending on circuit usage. I tried changing this value for different scenarios and settled on 2 being the best choice and left it explicit like this. I'll remove the kwarg so people realize it's the default.

test/python/transpiler/test_optimize_1q_decomposition.py
```diff
@@ -212,16 +212,18 @@ def test_ignores_conditional_rotations(self, basis):
...
         circuit.p(0.3, qr)
         circuit.p(0.4, qr)
 
         passmanager = PassManager()
         passmanager.append(Optimize1qGatesDecomposition(basis))
         result = passmanager.run(circuit)
-
-        self.assertTrue(Operator(circuit).equiv(Operator(result)))
+        with self.assertWarns(DeprecationWarning):
+            self.assertTrue(Operator(circuit).equiv(Operator(result)))
```
ElePT, line 226: Where is the warning triggered here?

test/python/transpiler/test_reset_after_measure_simplification.py
```diff
@@ -26,12 +26,13 @@ def test_simple(self):
         qc = QuantumCircuit(1, 1)
         qc.measure(0, 0)
         qc.reset(0)
-
-        new_qc = ResetAfterMeasureSimplification()(qc)
+        with self.assertWarns(DeprecationWarning):
+            new_qc = ResetAfterMeasureSimplification()(qc)
```
ElePT, line 30: Do you think this class raising a deprecation warning is a desired outcome (I'm not sure why it's raised)? Maybe the stack level of the warning isn't correct?

ElePT (commented):
LGTM! I just found a couple of deprecation warning assertions that I think would be good to clarify (maybe adding a comment explaining how they relate to this deprecation like you did in other cases).

test/python/transpiler/test_reset_after_measure_simplification.py
mtreinish: The testing harness will trigger if the warning is raised at any stack level even if it's not user code. This is to let us catch internal usage of deprecated functionality in our dependencies. In this case the `ResetAfterMeasureSimplification` pass only works with `c_if` and `condition`. so the warning is coming from inside the pass.

This does raise a good question about what we want to do with that pass, we probably should convert it to an if statement internally. I don't remember what I was thinking when I opened this 2 weeks ago, probably that I would do it in a followup PR. But it would be better if I do it here instead.

test/python/transpiler/test_optimize_1q_decomposition.py
mtreinish: `Operator(circuit)` runs `circuit.to_instruction()` which checks for `condition` iirc.

test/python/transpiler/test_reset_after_measure_simplification.py
mtreinish: When I went to do this I remember why I didn't change it. I decided at the time that leaving the pass using `c_if` for 1.x made the most sense as using an `IfElseOp` instead could have been viewed as a potential change in behavior and wanted to save that for 2.0.

ElePT: I agree with not changing the behavior yet. I brought this up thinking about downstream packages and how they sometimes get "flooded" by our deprecation warnings even if they aren't actionable. But I guess it's on them to set the stack level they want to see in their tests.

ElePT (approved):
LGTM

releasenotes/notes/deprecate-condition_c_if-9548e5522814fe9c.yaml
```diff
@@ -0,0 +1,38 @@
...
+    can be rewritten as::
+
+        qc = QuantumCircuit(2, 2)
+        qc.h(0)
+        with expected.if_test((expected.clbits[0], True)):
+            qc.x(0)
+        with expected.if_test((expected.clbits[1], False)):
+            qc.z(1)
+        qc.measure(0, 0)
+        qc.measure(1, 1)
```
wshanks, line 29: I just started addressing this deprecation warning and noticed some issues here. I think it should be `qc.z(1)` (missing parenthesis). Also, I think `expected` should be `qc`?

PR #13224, opened 2024-09-25
Deprecate the unit and duration attributes

mtreinish:
### Summary

This commit deprecates the unit and duration attributes for QuantumCircuit and Instruction. These attributes will be removed in Qiskit 2.0 as they're not needed anymore and are adding a lot of complexity to the circuit data model as they're mutable state and extra memory slots that we need to keep around. The better model for tracking instruction duration is in the Target as it's inherently a property of the backend running the instructions.

For the unittests this commit globally ignores the deprecation warning raised by this commit because internally we access the now deprecated access quite frequently as we need to check it when adding instructions to circuits to populate the rust data model.

### DIFF
--- a/releasenotes/notes/deprecate-unit-duration-48b76c957bac5691.yaml
+++ b/releasenotes/notes/deprecate-unit-duration-48b76c957bac5691.yaml
@@ -4,18 +4,18 @@ deprecations_circuits:
     The :attr:`.QuantumCircuit.unit` and :attr:`.QuantumCircuit.duration`
     attributes have been deprecated and will be removed in Qiskit 2.0.0. These
     attributes were used to track the estimated duration and unit of that
-    duration to execute on the circuit. However, the value of these attributes
-    were always limited, as it only would be properly populated if the
+    duration to execute on the circuit. However, the values of these attributes
+    were always limited, as they would only be properly populated if the
     transpiler were run with the correct settings. The duration was also only a
     guess based on the longest path on the sum of the duration of
     :class:`.DAGCircuit` and wouldn't ever correctly account for control flow
     or conditionals in the circuit.
   - |
     The :attr:`.Instruction.duration` and :attr:`.Instruction.unit` attributes
-    have been deprecated and will removed in Qiskit 2.0.0. These attributes
+    have been deprecated and will be removed in Qiskit 2.0.0. These attributes
     were used to attach a custom execution duration and unit for that duration
     to an individual instruction. However, the source of truth of the duration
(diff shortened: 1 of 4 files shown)


releasenotes/notes/deprecate-unit-duration-48b76c957bac5691.yaml
```diff
@@ -0,0 +1,25 @@
+---
+deprecations_circuits:
+  - |
+    The :attr:`.QuantumCircuit.unit` and :attr:`.QuantumCircuit.duration`
+    attributes have been deprecated and will be removed in Qiskit 2.0.0. These
+    attributes were used to track the estimated duration and unit of that
+    duration to execute on the circuit. However, the value of these attributes
+    were always limited, as it only would be properly populated if the
+    transpiler were run with the correct settings. The duration was also only a
```
ElePT:
```suggestion
    duration to execute on the circuit. However, the values of these attributes
    were always limited, as they would only be properly populated if the
    transpiler were run with the correct settings. The duration was also only a
```

releasenotes/notes/deprecate-unit-duration-48b76c957bac5691.yaml
```diff
@@ -0,0 +1,25 @@
...
+    guess based on the longest path on the sum of the duration of
+    :class:`.DAGCircuit` and wouldn't ever correctly account for control flow
+    or conditionals in the circuit.
+  - |
+    The :attr:`.Instruction.duration` and :attr:`.Instruction.unit` attributes
+    have been deprecated and will removed in Qiskit 2.0.0. These attributes
```
ElePT:
```suggestion
    have been deprecated and will be removed in Qiskit 2.0.0. These attributes
```

releasenotes/notes/deprecate-unit-duration-48b76c957bac5691.yaml
```diff
@@ -0,0 +1,25 @@
...
+    were used to attach a custom execution duration and unit for that duration
+    to an individual instruction. However, the source of truth of the duration
+    of a gate is the :class:`.BackendV2`\'s :class:`.Target` which contains
+    the duration for each instruction supported on the backend. The duration of
```
ElePT:
```suggestion
    of a gate is the :class:`.BackendV2` :class:`.Target` which contains
    the duration for each instruction supported on the backend. The duration of
```

ElePT (commented):
LGTM, I just added a few wording suggestions on the reno.

ElePT (approved):

PR #13232, opened 2024-09-27
Pending-deprecate library circuits with gate representations

Cryoris:
### Summary

Part of #13046: Library circuits that already have a gate/instruction version are pending-deprecated. These includes:
* `QFT`
* `Diagonal`
* `Permutation`
* `GMS`

### Details and comments

As part of this cleanup, `LinearFunction` is streamlined with all other gates by deprecating the custom `synthesize` method in favor of calling `definition` (or compiling the circuit).

No reno as all deprecation are only pending.


qiskit/circuit/library/basis_change/qft.py
```diff
@@ -232,30 +240,13 @@ def inverse(self, annotated: bool = False) -> "QFT":
...
-        beyond 1022 we're using fractional bits to represent leading zeros."""
-        max_num_entanglements = self.num_qubits - self.approximation_degree - 1
-        if max_num_entanglements > -np.finfo(float).minexp:  # > 1022 for doubles.
-            warnings.warn(
-                "precision loss in QFT."
-                f" The rotation needed to represent {max_num_entanglements} entanglements"
-                " is smaller than the smallest normal floating-point number.",
-                category=RuntimeWarning,
-                stacklevel=3,
-            )
```
ElePT, line 249: why was this warning removed?

qiskit/circuit/library/generalized_gates/gms.py
```diff
@@ -94,28 +98,77 @@ def __init__(self, num_qubits: int, theta: Union[List[List[float]], np.ndarray])
...
 
     def _define(self):
-        theta = self.params[0]
-        q = QuantumRegister(self.num_qubits, "q")
+        thetas = self.params[0]
+        q = QuantumRegister(self.num_qubits, name="q")
         qc = QuantumCircuit(q, name=self.name)
         for i in range(self.num_qubits):
             for j in range(i + 1, self.num_qubits):
+                # if theta was just a single angle, use that, otherwise use the correct index
```
ElePT:
```suggestion
                # if theta is just a single angle, use that, otherwise use the correct index
```

test/python/circuit/library/test_qft.py
```diff
@@ -196,10 +196,12 @@ def __init__(self, *_args, **_kwargs):
                 module=r"qiskit\..*",
                 message=r".*precision loss in QFT.*",
             )
+            warnings.filterwarnings("ignore", category=PendingDeprecationWarning)
```
ElePT, line 199: Do we fail on pending deprecation warnings? I am surprised, I thought we didn't.

qiskit/circuit/library/generalized_gates/gms.py
```diff
@@ -94,28 +98,77 @@ def __init__(self, num_qubits: int, theta: Union[List[List[float]], np.ndarray])
...
+                theta = thetas if isinstance(thetas, ParameterValueType) else thetas[i][j]
```
ElePT: I believe that it's this line that is currently failing in the unit tests. The error raised is:

`TypeError: Subscripted generics cannot be used with class and instance checks
`
Maybe you can check the shape of the input instead?

ElePT (commented):
In principle the PR looks good, I only wonder if the test with the warning filter was an oversight, and left a few tiny comments.

test/python/circuit/library/test_qft.py
Cryoris: Ah, no we don't but here we have a context that catches all warnings and checks that none are emitted (to check that there is no precision loss) -- so we have to filter out the pending deprecation warning 🙂

ElePT (approved):
Allright, LGTM!

PR #13234, opened 2024-09-27
Avoid checking if matrix is unitary in Split2qUnitaries

mtreinish:
### Summary

This commit fixes a small oversight in the Split2QUnitaries transpiler pass. When it does decide to split the unitary into two single qubit unitary gates it was creating a new UnitaryGate object. When these UnitaryGate objects were created we weren't setting the check_input flag to false. This meant we were spending time validating the matrix was a unitary, but in the context the matrix is known to be unitary already, so we don't need to spend the time doing this checking.

### Details and comments


Cryoris (approved):
Nice catch 👍🏻

PR #13238, opened 2024-09-28
Add Rust quantum volume function

mtreinish:
### Summary

This commit adds a new function quantum_volume used for generating a quantum volume model circuit. This new function is defined in Rust and multithreaded to improve the throughput of the circuit generation. This new function will eventually replace the existing QuantumVolume class as part of #13046. Since quantum volume is a circuit defined by it's structure using a generator function is inline with the goals of #13046.

Right now the performance is bottlenecked by the creation of the UnitaryGate objects as these are still defined solely in Python. We'll likely need to port the class to have a rust native representation to further speed up the construction of the circuit.

### Details and comments

### DIFF
--- a/crates/accelerate/src/circuit_library/quantum_volume.rs
+++ b/crates/accelerate/src/circuit_library/quantum_volume.rs
@@ -33,6 +33,10 @@ use qiskit_circuit::packed_instruction::PackedOperation;
 use qiskit_circuit::Qubit;
 use smallvec::smallvec;
 
+// This function's implementation was modeled off of the algorithm used in the
+// `scipy.stats.unitary_group.rvs()` function defined here:
+//
+// https://github.com/scipy/scipy/blob/v1.14.1/scipy/stats/_multivariate.py#L4224-L4256
 #[inline]
 fn random_unitaries(seed: u64, size: usize) -> impl Iterator<Item = Array2<Complex64>> {
     let mut rng = Pcg64Mcg::seed_from_u64(seed);
--- a/crates/accelerate/src/circuit_library/quantum_volume.rs
+++ b/crates/accelerate/src/circuit_library/quantum_volume.rs
@@ -172,11 +172,15 @@ pub fn quantum_volume(
             None => Pcg64Mcg::from_entropy(),
         };
         let seed: u64 = outer_rng.sample(rand::distributions::Standard);
+
+        let unitaries: Vec<Array2<Complex64>> = random_unitaries(seed, num_unitaries).collect();
+
         CircuitData::from_packed_operations(
             py,
             num_qubits,
             0,
-            random_unitaries(seed, num_unitaries)
+            unitaries
+                .into_iter()
                 .enumerate()
                 .map(|x| build_instruction(x, &mut outer_rng)),
             Param::Float(0.),
--- a/crates/accelerate/src/circuit_library/quantum_volume.rs
+++ b/crates/accelerate/src/circuit_library/quantum_volume.rs
@@ -1,0 +1,149 @@
+// This code is part of Qiskit.
+//
+// (C) Copyright IBM 2024
+//
+// This code is licensed under the Apache License, Version 2.0. You may
+// obtain a copy of this license in the LICENSE.txt file in the root directory
+// of this source tree or at http://www.apache.org/licenses/LICENSE-2.0.
+//
+// Any modifications or derivative works of this code must retain this
+// copyright notice, and modified files need to carry a notice indicating
+// that they have been altered from the originals.
+
+use std::thread::available_parallelism;
+
+use qiskit_circuit::imports::UNITARY_GATE;
+
+use pyo3::prelude::*;
+
+use crate::getenv_use_multiple_threads;
+use faer_ext::{IntoFaerComplex, IntoNdarrayComplex};
+use ndarray::prelude::*;
+use num_complex::Complex64;
+use numpy::IntoPyArray;
+use rand::prelude::*;
+use rand_distr::Normal;
+use rand_pcg::Pcg64Mcg;
+use rayon::prelude::*;
+
+use qiskit_circuit::circuit_data::CircuitData;
+use qiskit_circuit::operations::Param;
+use qiskit_circuit::operations::PyInstruction;
+use qiskit_circuit::packed_instruction::PackedOperation;
+use qiskit_circuit::Qubit;
+use smallvec::smallvec;
+
+#[inline]
+fn random_unitaries(seed: u64, size: usize) -> impl Iterator<Item = Array2<Complex64>> {
+    let mut rng = Pcg64Mcg::seed_from_u64(seed);
+    let dist = Normal::new(0., 1.0).unwrap();
+
+    (0..size).map(move |_| {
+        let mut z: Array2<Complex64> = Array2::from_shape_simple_fn((4, 4), || {
+            Complex64::new(dist.sample(&mut rng), dist.sample(&mut rng))
+        });
+        z.mapv_inplace(|x| x * std::f64::consts::FRAC_1_SQRT_2);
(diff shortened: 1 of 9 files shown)


mtreinish:
I ran some benchmarks comparing the construction speed of the quantum volume circuit using a logarithmic sweep from 10 to 1000 qubits (with equal depth for each point) using this function in parallel and serially (by setting `QISKIT_IN_PARALLEL=TRUE`) and comparing that against the Python `QuantumVolume` class:

construct_time

All these results were on my laptop with a AMD Ryzen 7 PRO 7840U.

There is potentially some tuning we might want to do about the number of threads and the chunk size for each parallel worker. But this seems like a reasonable balance and nothing I've changed so far appreciably changed the performance. It might be worth testing for the serial path though if doing the unitary generation all up front, like what is done in the parallel case, improves performance.

crates/accelerate/src/circuit_library/quantum_volume.rs
```diff
@@ -0,0 +1,149 @@
...
+use rayon::prelude::*;
+
+use qiskit_circuit::circuit_data::CircuitData;
+use qiskit_circuit::operations::Param;
+use qiskit_circuit::operations::PyInstruction;
+use qiskit_circuit::packed_instruction::PackedOperation;
+use qiskit_circuit::Qubit;
+use smallvec::smallvec;
+
+#[inline]
```
Cryoris, line 50: Would it make sense to link the paper to explain where the logic is coming from? https://arxiv.org/pdf/math-ph/0609050v2

crates/accelerate/src/circuit_library/quantum_volume.rs
```diff
@@ -0,0 +1,149 @@
...
+
+    (0..size).map(move |_| {
+        let mut z: Array2<Complex64> = Array2::from_shape_simple_fn((4, 4), || {
+            Complex64::new(dist.sample(&mut rng), dist.sample(&mut rng))
+        });
+        z.mapv_inplace(|x| x * std::f64::consts::FRAC_1_SQRT_2);
+        let qr = z.view().into_faer_complex().qr();
+        let r = qr.compute_r().as_ref().into_ndarray_complex().to_owned();
+        let mut d = r.into_diag();
+        d.mapv_inplace(|d| d / d.norm());
```
Cryoris: Does this re-compute the norm of `d` in every application? If yes would it make sense to pre-compute `d.norm()`?

crates/accelerate/src/circuit_library/quantum_volume.rs
```diff
@@ -0,0 +1,149 @@
...
+        let mut q = qr.compute_q().as_ref().into_ndarray_complex().to_owned();
+        q.axis_iter_mut(Axis(0)).for_each(|mut row| {
```
Cryoris, line 91: In the paper they return `q @ d_normalized @ q` -- is this missing the second `q` application here? 🤔

qiskit/circuit/library/quantum_volume.py
```diff
@@ -113,3 +114,37 @@ def __init__(
                     base._append(CircuitInstruction(gate, qubits[qubit : qubit + 2]))
         if not flatten:
             self._append(CircuitInstruction(base.to_instruction(), tuple(self.qubits)))
+
+
+def quantum_volume(
+    num_qubits: int,
+    depth: Optional[int] = None,
+    seed: Optional[Union[int, np.random.Generator]] = None,
```
Cryoris:
```suggestion
    depth: int | None = None,
    seed: int | np.random.Generator | None = None,
```

crates/accelerate/src/circuit_library/quantum_volume.rs
mtreinish: I actually based the logic on how the scipy function we were previously using worked internally. But sure I can add a link to that here: https://github.com/scipy/scipy/blob/v1.14.1/scipy/stats/_multivariate.py#L4224-L4256

mtreinish: Done in: https://github.com/Qiskit/qiskit/pull/13238/commits/69d4cfa50a25631a37caa12ab16a366ac48845b4

crates/accelerate/src/circuit_library/quantum_volume.rs
mtreinish: This does an in-place substitution on the diagonal `d`  so that each value also named `d` (which is probably why this is confusing) is replaced with `d / d.norm()` for each element `d`. I'll rename the diagonal to `diag` or something to make it clear which is an element and which is the array.

mtreinish: I am wondering if I should rewrite this using fixed size arrays or nalgebra's `Matrix4<Complex64>` to make this all faster. Right now I'm pretty confident the overhead is mostly dealing with allocations for 2x2 arrays.

mtreinish: I did this here: https://github.com/Qiskit/qiskit/pull/13238/commits/32b4c1f8c32a7e531c9bc4b12288be89b3cc5b38 the code is a lot more verbose now, but I think it's a bit more explicit and it'll be a bit more optimized (although the random unitary construction is not the bottleneck).

crates/accelerate/src/circuit_library/quantum_volume.rs
Cryoris: Ok no, Scipy also has `q @ d_normalized` and in the equations of the paper they describe the same. Just the pseudo-code of the paper shows the additional multiplication with `q`, maybe that's strange notation or a typo. All good then 🙂

crates/accelerate/src/circuit_library/quantum_volume.rs
Cryoris: Nice! The bottleneck is the `UnitaryGate` creation Python-side as you wrote above, right? I think a 10x speedup (maybe more now with the explicit matrix allocation?) is already great, plus we have the circuit as function reaching the goal of #13046 😄 But if you'd like to test with pre-generating the gates I'm also happy to wait 🙂

mtreinish: Yeah, we are totally bottlenecked by the `UnitaryGate` object creation, I ran a profile with the current state of the PR running the test script I was using to generate the data for the graph I shared above. In serial mode (py-spy doesn't handle the multithreading well) and the `UnitaryGate` object creation is taking ~58% of the time, while the random unitary generation is only ~30% of the runtime (almost all of that is performing the qr factorization).

I'll do a quick test of serial mode dumping the unitaries to a vec first to see if that makes a difference performance wise (I don't expect it to), and will leave a comment here with the results.

mtreinish: times

I reran the numbers collecting the unitaries into a vec and then iterating off of that instead of passing the iterator directly to the circuit constructor. It looks like it's marginally faster to collect into a vec.

qiskit/circuit/library/quantum_volume.py
```diff
@@ -113,3 +116,37 @@ def __init__(
...
+
+       from qiskit.circuit.library import quantum_volume
+       circuit = quantum_volume(5, 6, seed=10)
+       circuit.draw('mpl')
+
+    **References:**
+
+    [1] A. Cross et al. Validating quantum computers using
+    randomized model circuits, Phys. Rev. A 100, 032328 (2019).
+    [`arXiv:1811.12926 <https://arxiv.org/abs/1811.12926>`_]
```
Cryoris: I know it's like this above, but shouldn't it be the following?
```suggestion
    `arXiv:1811.12926 <https://arxiv.org/abs/1811.12926>`__
```

Cryoris (approved):
LGTM and thanks for all the detailed benchmarks! 😄

PR #13240, opened 2024-09-30
Improve qubit tracking in HighLevelSynthesis

### LINKED ISSUE #13239: Improve ancilla detection in HighLevelSynthesis
### What should we add?

In #12911 we have added the ability to track the state of each qubit within the `HighLevelSynthesis` transpiler pass (i.e. whether the qubit is "clean" (in state  $|0\rangle$) or "dirty" (in some other state)) and to automatically exploit clean/dirty ancilla qubits with the appropriate synthesis algorithms.

In the following snippet, a circuit with a single MCX-gate 
```
qc = QuantumCircuit(10)
qc.mcx([3, 4, 5, 6, 7], 0)
basis_gates = ["u", "cx"]
tqc = HighLevelSynthesis(basis_gates=basis_gates)(qc)
```    
is synthesized to a circuit with 24 CX-gates and 45 U-gates. Internally the MCX synthesis algorithm has access to $4$ clean ancilla qubits (``1, 2, 8, 9``).

However, an arguably identical circuit in the following snippet,
```
inner = QuantumCircuit(6)
inner.mcx([0, 1, 2, 3, 4], 5)
custom_gate = inner.to_gate()
qc = QuantumCircuit(10)
qc.append(custom_gate, [3, 4, 5, 6, 7, 0])
basis_gates = ["u", "cx"]
tqc = HighLevelSynthesis(basis_gates=basis_gates)(qc)
```
is synthesized to a _worse_ circuit with 84 CX-gates and 127 U-gates. This is due to how the recursion is treated with HighLevelSynthesis, with the synthesis of `custom_gate` having no access to ancilla qubits outside of its definition. 

This issue will be very soon followed by a PR addressing the problem above.

alexanderivrii:
### Summary

Fixes #13239. Now for both examples in the referenced issue, HighLevelSynthesis produces a circuit with 24 CX-gates and 45 U-gates. 

In addition, this also improves clean ancilla detection on following example:
```
inner1 = QuantumCircuit(4)
inner1.h(0)
inner1.cz(0, 2)
qc = QuantumCircuit(6)
qc.append(inner1.to_instruction(), [1, 2, 3, 4])
pass_ = HighLevelSynthesis(basis_gates=["cx", "u"])
qct = pass_(qc)
```
Even though the inner circuit `inner1` is defined over the qubits `1, 2, 3, 4` in the main circuit `qc`, the qubits `1` and `3` in the inner circuit (corresponding to the qubits `2` and `4` in the main circuit) remain clean and can be used as clean ancilla qubits in the following gates.

### Details and comments

This PR is based on numerous discussions with @Cryoris.

Recalling the second example from the referenced issue
```
inner = QuantumCircuit(6)
inner.mcx([0, 1, 2, 3, 4], 5)
custom_gate = inner.to_gate()
qc = QuantumCircuit(10)
qc.append(custom_gate, [3, 4, 5, 6, 7, 0])
basis_gates = ["u", "cx"]
tqc = HighLevelSynthesis(basis_gates=basis_gates)(qc)
```
the tricky part is that the recursive call to `HighLevelSynthesis::_run` on the inner circuit should have access to the clean "global" qubits outside of the circuit's definition. To tackle this, we introduce a class `QubitContext` which keeps the correspondence between the current DAG's qubits and the global qubits of the original circuit. The state of the global qubits (clean/dirty) is tracked by `QubitTracker`. When an internal synthesis algorithm (here for the internal MCX gate) checks how many clean/dirty ancilla qubits are available, it does so taking all of the global qubits into account. 

However, this also means that synthesizing an internal DAG may output a DAG with more qubits. In particular (demonstrating possible complications) we can have an internal DAG with say 10 qubits that contains an MCX-gate over 8 qubits that can be synthesized by the appropriate synthesis algorithm to a circuit over 14 qubits (exploiting the global clean/dirty) qubits. Hence, our internal DAG must also grow in size (fortunately all the tracking is possible using the local-to-global correspondences of all the objects involved) which might mean that the DAG higher up in the chain might need to grow as well. 

This PR also adds more HighLevelSynthesis tests exploring all these possible edge-cases.

In summary, this should handle any mixture of recursively defined circuits with annotated operations, custom gate definitions and HighLevelSynthesis plugins. For circuits containing control-flow ops, the extended functionality is not used (and will be delegated to another PR, if someone wants to tackle this). In other words, if we have a control-flow op over say 4 qubits in the bigger circuit, only these 4 qubits will be used when recursively processing the blocks in this op. 

_Update_: An additional observation is that this also improves the synthesis of _open-controlled_ MCX gates. In Qiskit, the _name_ of an open-controlled gate is not of the form "mcx", but rather of the form "mcx_o17", where "17" is (the integer representation of) the control state. When processing this gate, HLS would not immediately call a synthesis plugin (since "mcx_o17" is not in the list of gate names for which plugins exist), but recursively process the _definition_ of this gate, which is a quantum circuit consisting of a layer of X-gates, a closed-controlled MCX gate (called "mcx"), and the inverse layer of X-gates. During this recursion, HLS would now indeed call the synthesis plugin for the internal MCX-gate, and with this PR it would be able to use the ancilla qubits in the main circuit and outside of the internal open-controlled gate's definition.


qiskit/transpiler/passes/synthesis/high_level_synthesis.py
```diff
@@ -135,6 +135,58 @@ def set_methods(self, hls_name, hls_methods):
...
+    aka the "local qubits" (for instance, of the definition circuit
+    of a custom gate), and the qubit indices of the original DAG, aka the
+    "global qubits".
+
+    The local qubit indices are assumed to be consecutive integers starting at
+    zero, i.e. 0, 1, 2, etc.
+    """
+
+    def __init__(self, local_to_global: dict):
+        self.local_to_global = local_to_global
```
raynelfss: Is there a chance that maybe this could work with a `list`. From what I can tell this is adding qubit indices based on their location in the original `DAGCircuit`. So the `local_to_global` mapping would have its keys be consecutive local qubit indices, i.e. 0..n. So maybe a `list` would work the same.

qiskit/transpiler/passes/synthesis/high_level_synthesis.py
```diff
@@ -135,6 +135,58 @@ def set_methods(self, hls_name, hls_methods):
...
+        return len(self.local_to_global)
+
+    def add_qubit(self, global_qubit) -> int:
+        """Extends the correspondence by an additional qubit that
+        maps to the given global qubit. Returns the index of the
+        new local qubit.
+        """
+        new_local_qubit = len(self.local_to_global)
+        self.local_to_global[new_local_qubit] = global_qubit
+        return new_local_qubit
```
raynelfss: When providing the initial mapping, are there cases where you wouldn't provide the mapping and then you'd fill it up in-place using this function?

qiskit/transpiler/passes/synthesis/high_level_synthesis.py
```diff
@@ -135,6 +135,58 @@ def set_methods(self, hls_name, hls_methods):
...
+
+    def to_global_mapping(self):
+        """Returns the local-to-global mapping."""
+        return self.local_to_global
```
raynelfss: It seems that `local_to_global` is already visible to the public. If that's not the intention and you want this method to be the only way to check that, you should probably rename `self.local_to_global` to `self._local_to_global` and maybe make this method a `@property`.

qiskit/transpiler/passes/synthesis/high_level_synthesis.py
```diff
@@ -135,6 +135,58 @@ def set_methods(self, hls_name, hls_methods):
...
+
+    def restrict(self, qubits: list[int] | tuple[int]) -> "QubitContext":
+        """Restricts the context to a subset of qubits, remapping the indices as
+        to be consecutive integers starting at zero.
+        """
+        return QubitContext({i: self.local_to_global[qubits[i]] for i in range(len(qubits))})
+
+    def to_global(self, qubit: int) -> int:
+        """Returns the global qubits corresponding to the given local qubits."""
+        return self.local_to_global[qubit]
```
raynelfss: This also looks like you wanted to keep `local_to_global` to be private but only accessible through this API, I suggest you rename it to `self._local_to_global`.

qiskit/transpiler/passes/synthesis/qubit_tracker.py
```diff
@@ -25,108 +25,95 @@ class QubitTracker:
...
+        for q in ignored_qubits:
+            self.ignored[q] = True
+        for q in range(self.num_qubits):
+            if (not self.ignored[q]) and self.enabled[q] and self.state[q]:
+                count += 1
+        for q in ignored_qubits:
+            self.ignored[q] = False
+        return count
+
+    def num_dirty(self, ignored_qubits):
```
raynelfss, line 66: **Nitpick:**
This is a smart way of keeping `ignored_qubits: Iterable[int]` an iterable of any kind, instead of a set. But I only worry about the fact that we keep this extra `dict` in the tracker which will only be used for ignoring qubits in two methods. Is this the best approach in that case? Were we to accept sets we'd save the space of that extra `dict` and the two loops that happen in there as well.

qiskit/transpiler/passes/synthesis/qubit_tracker.py
```diff
@@ -25,108 +25,95 @@ class QubitTracker:
...
+            self.ignored[q] = True
+        for q in range(self.num_qubits):
+            if (not self.ignored[q]) and self.enabled[q] and self.state[q]:
+                res.append(q)
+        for q in range(self.num_qubits):
+            if (not self.ignored[q]) and self.enabled[q] and not self.state[q]:
+                res.append(q)
+        for q in ignored_qubits:
+            self.ignored[q] = False
+        return res[:num_qubits]
```
raynelfss, line 92: Hmm... This method feels like it could be rewritten to work a little better for a couple of reasons:
- The two loops over ignored qubits (see my previous comment).
- The multiple looping over `0..n for n = num_qubits`.
- The extra operations done when you're only looking for `num_qubit` amount of qubits to borrow. Since in the end here we will always iterate `2n for n = num_qubits`.

Perhaps this method could be re-written with a while loop to keep two collections of clean and dirty qubits as we iterate (perhaps with a data structure similar to a stack or deque). And then handle the stop condition based on the amount of clean qubits vs num_qubits.

qiskit/transpiler/passes/synthesis/qubit_tracker.py
```diff
@@ -25,108 +25,95 @@ class QubitTracker:
...
+        for q in range(self.num_qubits):
+            out += str(q) + ": "
+            if not self.enabled[q]:
+                out += "_"
+            elif self.state[q]:
+                out += "0"
+            else:
+                out += "*"
+            out += "; "
+        return out
```
raynelfss: You might want to include `QubitTracker` somewhere here in the pretty print. To keep it a little consistent.

qiskit/transpiler/passes/synthesis/high_level_synthesis.py
```diff
@@ -271,116 +323,200 @@ def run(self, dag: DAGCircuit) -> DAGCircuit:
...
-        # qubit states. It does not trade-off allocations and just gives all available qubits
-        # to the current operation (a "the-first-takes-all" approach).
+        Note that by using the auxiliary qubits to synthesize operations present in the input DAG,
+        the synthesized DAG may be defined over more qubits than the input DAG. In this case,
+        the function update in-place the global qubits tracker and extends the local-to-global
+        context.
+        """
+
+        if dag.num_qubits() != context.num_qubits():
+            raise TranspilerError("HighLevelSynthesis internal error.")
```
raynelfss, line 363: Do we know any way that this condition would be reached? If so we might wanna put a more specific error message here.

qiskit/transpiler/passes/synthesis/high_level_synthesis.py
```diff
@@ -271,116 +323,200 @@ def run(self, dag: DAGCircuit) -> DAGCircuit:
...
+        # done at the top-level since this does not update the global qubits tracker.
+        if top_level:
+            for node in dag.op_nodes():
+                qubits = tuple(dag.find_bit(q).index for q in node.qargs)
+                if not self._definitely_skip_node(node, qubits, dag):
+                    break
+            else:
+                # The for-loop terminates without reaching the break statement
+                if dag.num_qubits() != context.num_qubits():
+                    raise TranspilerError("HighLevelSynthesis internal error.")
```
raynelfss, line 375: Same here, could we be specific if we know what could cause this to happen, if ever.

qiskit/transpiler/passes/synthesis/high_level_synthesis.py
```diff
@@ -271,116 +323,200 @@ def run(self, dag: DAGCircuit) -> DAGCircuit:
...
-                    tracker.used(qubits)  # any other op used the clean state up
+            # If the synthesis did not change anything, just update the qubit tracker.
+            elif not processed:
+                tracker.set_dirty(context.to_globals(qubits))
 
-        # we did not change anything just return the input
+        # We did not change anything just return the input.
         if len(synthesized_nodes) == 0:
+            if dag.num_qubits() != context.num_qubits():
+                raise TranspilerError("HighLevelSynthesis internal error.")
```
raynelfss, line 448: Same as previous comment, do we know what could cause this condition and could we give a more detailed error message.

qiskit/transpiler/passes/synthesis/high_level_synthesis.py
```diff
@@ -135,6 +135,58 @@ def set_methods(self, hls_name, hls_methods):
...
+        self.local_to_global[new_local_qubit] = global_qubit
+        return new_local_qubit
+
+    def to_global_mapping(self):
+        """Returns the local-to-global mapping."""
+        return self.local_to_global
+
+    def to_local_mapping(self):
+        """Returns the global-to-local mapping."""
+        return {j: i for (i, j) in self.local_to_global.items()}
```
raynelfss: Maybe this is why you're keeping the `dict` for `local_to_global`, when performing this there might be a chance of it not mapping consecutive indices (or at least not have them start at 0, but I'm not sure if that's the case.

qiskit/transpiler/passes/synthesis/high_level_synthesis.py
```diff
@@ -389,89 +525,169 @@ def _synthesize_operation(
...
 
-            else:
-                raise RuntimeError(f"Unexpected synthesized type: {type(synthesized)}")
+        else:
+            raise TranspilerError(f"Unexpected synthesized type: {type(synthesized)}")
 
-        if synthesized is not None and used_qubits is None:
-            raise RuntimeError("Failed to find qubit indices on", synthesized)
+        if isinstance(synthesized, DAGCircuit) and synthesized_context is None:
+            raise TranspilerError("HighLevelSynthesis internal error.")
```
raynelfss, line 688: We could be a little more specific about this error message if the reason for this not to work is known.

raynelfss (commented):
This is great work, thank you!

I had a couple of questions about the implementation of the `QubitContext` and the refactoring of `QubitTracker`. But the addition seems to make sense so far.

qiskit/transpiler/passes/synthesis/high_level_synthesis.py
alexanderivrii: That's a really great suggestion! 

Done in [112e9e3](https://github.com/Qiskit/qiskit/pull/13240/commits/112e9e3d183b23bab83960e0050e94054b36911d).

qiskit/transpiler/passes/synthesis/high_level_synthesis.py
alexanderivrii: Yes (and this why I called this "tricky" in the description). More precisely, there are cases when you would provide some initial mapping, and then additionally call `add_qubit` (possibly multiple times).

Let's say our main circuit is over 10 qubits `0, ..., 9`, let one of its gates be a custom gate over qubits `4, 5, 6, 7, 8`, and let the _definition_ of this custom gate include the MCX-gate `mcx(0, 1, 2, 3, 4)`. This example might seem strange, but it's actually a very practical example, describing exactly what happens for open-controlled MCX-gates, arithmetic circuits, and so on.

When HLS (recursively) examines the definition of the custom gate, it sees a local DAG over 5 qubits `0, 1, 2, 3, 4`, with the qubit context initially being `{0->4, 1->5, 2->6, 3->7, 4->8}`. When synthesizing the inner MCX gate, it could choose a synthesis that needs 3 additional ancilla qubits. Note that we do not have available qubits inside the local DAG (as the mcx gate is defined over all of its qubits), however these qubits are available in the main DAG (by looking at the global qubit tracker). Let's say that we choose to use global qubits `0`, `9` and `2`. The qubit context of our local DAG thus gets updated to `{0->4, 1->5, 2->6, 3->7, 4->8, 5->0, 6->9, 7->2}`.

qiskit/transpiler/passes/synthesis/high_level_synthesis.py
alexanderivrii: My intention was to have the classes `QubitContext` and `QubitTracker` _completely internal_ (at least for now), and in fact to immediately port them to Rust (to maximally reduce the performance overhead of all of this additional bookkeeping). In particular, I did not add them to any module-level docstrings. I also do not want to commit 100% to a specific implementation, so from the perspective of the HighLevelSynthesis code itself, I do indeed want a clean separation. What do you think?

qiskit/transpiler/passes/synthesis/high_level_synthesis.py
alexanderivrii: Done in [02ca8aa](https://github.com/Qiskit/qiskit/pull/13240/commits/02ca8aa1b0f4828e6cb4cec16c4cabc9af588711).

qiskit/transpiler/passes/synthesis/high_level_synthesis.py
alexanderivrii: Sure, renaming sounds as a good idea and does not prevent me from porting this to Rust. Done in [02ca8aa](https://github.com/Qiskit/qiskit/pull/13240/commits/02ca8aa1b0f4828e6cb4cec16c4cabc9af588711).

qiskit/transpiler/passes/synthesis/qubit_tracker.py
alexanderivrii: I did run some quick-and-dirty benchmarking on circuits with many MCX gates, and found it to be a bit faster _with_ the boolean vector `ignored_qubits` in the tracker, this is why I have settled on this implementation (which has `O(n)` runtime as compared to at least `O(n * log n)` with sets). However, in all honesty, there are usually very few MCX gates in practical circuits (if at all), so in practice we may not see any runtime difference at all. (Aside: it is very important to have the function `set_dirty` as fast as possible, since this does get used after every gate, but that does not involve sets or `ignored_qubits`). I can revisit this as I port this to Rust.

qiskit/transpiler/passes/synthesis/qubit_tracker.py
alexanderivrii: Great feedback, I will think how to improve this.

qiskit/transpiler/passes/synthesis/qubit_tracker.py
alexanderivrii: Done in [60e7456](https://github.com/Qiskit/qiskit/pull/13240/commits/60e745607d27f01a16d9c8f80061bdb91ad7ccb1). Though this might become obsolete after porting to Rust.

qiskit/transpiler/passes/synthesis/high_level_synthesis.py
alexanderivrii: If there are no bugs, this should never-ever happen. However, this was very useful during PR implementation / debugging (as it allows to pinpoint the existence of a problem as early as possible), and I think keeping these checks would be good in case there are some bugs / unexpected corner-cases.  I don't know if I can think of a more meaningful message than will help debugging, modulo the fact that something is going wrong.

qiskit/transpiler/passes/synthesis/high_level_synthesis.py
alexanderivrii: Yeah, the "inverse" `global_to_local` mapping is only a partial mapping and I am returning it a `dict`. But the forward mapping can and is now a `list` based on your previous suggestion.

qiskit/transpiler/passes/synthesis/high_level_synthesis.py
alexanderivrii: Ok, I will try to see if I can think of better error messages... Or maybe I can differentiate these somehow, in one place say "something when horribly wrong", in another - "something went terribly wrong", etc.

alexanderivrii:
Note: in [7432b6b](https://github.com/Qiskit/qiskit/pull/13240/commits/7432b6b78f8b10c9cb3443478174ade6b7fb2be9) I have disabled one of the newly added MCMT tests, I am trying to decide whether it points to a real bug or HLS is accidentally doing something too smart. Please do not merge the PR before that's taken care of.

UPDATE: false alarm, the point of the test was to check that MCMT's default plugin would not use ancillas if not enough were available and thus resorted to a more naive definition consisting of two consecutive MCX gates. However, ha-ha, the MCX synthesis plugin was using these ancilla qubits. The fix in [1e10c0d](https://github.com/Qiskit/qiskit/pull/13240/commits/1e10c0d0110c005f34b7b4c348af5a3cf1181612) simply disabled MCX synthesis algorithm from using ancillas.

alexanderivrii:
Please note a small follow-up PR #13369 that ports `QubitTracker` and `QubitContext` to Rust.

raynelfss (approved):
The code here is mostly good, there are still some optimizations that can be made but since these additions are already being ported to rust by #13369 we can focus on adding those there. Great job @alexanderivrii!

PR #13241, opened 2024-09-30
IQP as circuit library function (and in Rust)

Cryoris:
### Summary

Part of #13046. Moves `IQP` (instantaneous quantum polytime) circuits to be a function, and ports the internals to Rust.

### Details and comments

* This also adds the feature of creating random IQP circuits by just setting the number of qubits, without explicitly requiring the interactions matrix.
* The speedup from porting to Rust is more noticeable for large circuits (100+ qubits), where it is ~55x faster. See the table below for more numbers 🙂 

<details closed>
<summary>IQP timings</summary>

```
-- n = 10, speedup 4.48x
main: 0.001 +- 0.000
this: 0.000 +- 0.000

-- n = 20, speedup 12.59x
main: 0.002 +- 0.000
this: 0.000 +- 0.000

-- n = 50, speedup 45.07x
main: 0.009 +- 0.000
this: 0.000 +- 0.000

-- n = 100, speedup 58.98x
main: 0.034 +- 0.002
this: 0.001 +- 0.000

-- n = 200, speedup 58.45x
main: 0.133 +- 0.004
this: 0.002 +- 0.001

-- n = 500, speedup 51.41x
main: 0.862 +- 0.037
this: 0.017 +- 0.010

```
</details>


crates/accelerate/src/circuit_library/iqp.rs
```diff
@@ -0,0 +1,159 @@
...
+fn iqp(
+    interactions: ArrayView2<i64>,
+) -> impl Iterator<Item = (StandardGate, SmallVec<[Param; 3]>, SmallVec<[Qubit; 2]>)> + '_ {
+    let num_qubits = interactions.ncols();
+
+    // The initial and final Hadamard layer.
+    let h_layer =
+        (0..num_qubits).map(|i| (StandardGate::HGate, smallvec![], smallvec![Qubit(i as u32)]));
+
+    // The circuit interactions are powers of the CSGate, which is implemented by calling
```
ShellyGarion, line 41: why not to call the `CSGate`, `CSdgGate` and `CZGate` instead of `CPhaseGate` ?

crates/accelerate/src/circuit_library/iqp.rs
```diff
@@ -0,0 +1,159 @@
...
+            .map(move |(j, value)| {
+                (
+                    StandardGate::CPhaseGate,
+                    smallvec![Param::Float(PI2 * value as f64)], // todo ignore if interactions is 0 mod 4
+                    smallvec![Qubit(i as u32), Qubit(j as u32)],
+                )
+            })
+    });
+
+    // The layer of T gates. Again we use the PhaseGate, now with powers of Pi/8. The powers
```
ShellyGarion, line 57: why not to call the `TGate`, `TdgGate`, `SGate`, `SdgGate` and `ZGate` instead of `PhaseGate` ?

qiskit/circuit/library/iqp.py
```diff
@@ -69,28 +76,79 @@ def __init__(self, interactions: list | np.ndarray) -> None:
...
+    """
+    # if no interactions are given, generate them
+    if interactions is None:
+        if num_qubits is None:
+            raise ValueError("Either interactions or num_qubits must be provided.")
+
+    # otherwise validate the interactions
+    else:
+        if num_qubits is not None:
+            raise ValueError("Only one of interactions or num_qubits can be provided, not both.")
```
ShellyGarion: perhaps both could be given, and one can assert that `num_qubits = len(interactions)` in this case ?

ShellyGarion (commented):
Looks good. I only had a few questions.

crates/accelerate/src/circuit_library/iqp.rs
Cryoris: I wanted to implement it the same fashion as the Python code did it, but we can also change it if you think that's better 🙂

ShellyGarion: I agree that probably it's better to keep the same Python behavior. Do you know how these gates are used? 
I was wondering if the original code was written before we had the `CSGate & CSdgGate` in qiskit.

qiskit/circuit/library/iqp.py
Cryoris: Yeah I thought about that, it seemed a bit strange that passing an additional argument could raise an error in a function when one could just omit it. Explicitly disallowing passing both would make sure users can only have a single way of running the code... but I'm also happy to allow `num_qubits` 🙂

crates/accelerate/src/circuit_library/iqp.rs
Cryoris: Yeah the original code (4 years ago) definitely pre-dates the CS gates (2 years ago)...

crates/accelerate/src/circuit_library/iqp.rs
```diff
@@ -0,0 +1,159 @@
...
+    // the CPhaseGate with angles of Pi/2 times the power. The gate powers are given by the
+    // upper triangular part of the symmetric ``interactions`` matrix.
+    let connections = (0..num_qubits).flat_map(move |i| {
+        (i + 1..num_qubits)
+            .map(move |j| (j, interactions[(i, j)]))
+            .filter(move |(_, value)| value % 4 != 0)
+            .map(move |(j, value)| {
+                (
+                    StandardGate::CPhaseGate,
+                    smallvec![Param::Float(PI2 * value as f64)], // todo ignore if interactions is 0 mod 4
```
ShellyGarion: do you plan to do this in a follow-up PR? or was it already done in line 46?

test/python/circuit/library/test_iqp.py
```diff
@@ -49,9 +60,20 @@ def test_iqp(self):
...
-    def test_iqp_bad(self):
+    @data(True, False)
+    def test_iqp_bad(self, use_function):
         """Test that [0,..,n-1] permutation is required (no -1 for last element)."""
-        self.assertRaises(CircuitError, IQP, [[6, 5], [2, 4]])
+        self.assertRaises(CircuitError, iqp if use_function else IQP, [[6, 5], [2, 4]])
+
+    def test_random_iqp(self):
+        """Test generating a random IQP circuit."""
+
```
ShellyGarion, line 70: perhaps also check that the interactions are symmetric?

crates/accelerate/src/circuit_library/iqp.rs
Cryoris: Oh yes that is done already, I'll remove the comment -- good catch!

test/python/circuit/library/test_iqp.py
Cryoris: Once the circuit is created there's not really a way to check whether the interactions matrix was symmetric... In fact, we could even drop the symmetry-requirement as the code only ever uses the upper triangular part of the matrix. I guess from the theory the matrix should be symmetric, so it's a good sanity check, but for the code it doesn't matter. 

Or maybe you mean something else? 🙂

ShellyGarion: I see, so in fact it could be an upper triangular matrix, since the gates are symmetric ?

test/python/circuit/library/test_iqp.py
```diff
@@ -49,9 +60,20 @@ def test_iqp(self):
         simulated = Operator(circuit)
         self.assertTrue(expected.equiv(simulated))
 
-    def test_iqp_bad(self):
+    @data(True, False)
+    def test_iqp_bad(self, use_function):
         """Test that [0,..,n-1] permutation is required (no -1 for last element)."""
-        self.assertRaises(CircuitError, IQP, [[6, 5], [2, 4]])
```
alexanderivrii: I know that the test description on line 65 existed before this PR, but I don't understand it. Is the test bad because the interaction matrix is not symmetric?

qiskit/circuit/library/iqp.py
```diff
@@ -69,28 +76,79 @@ def __init__(self, interactions: list | np.ndarray) -> None:
...
+    simulation of commuting quantum computations,
+    Phys. Rev. Lett. 117, 080501 (2016).
+    `arXiv:1504.07999 <https://arxiv.org/abs/1504.07999>`_
+
+    Args:
+        interactions: The interactions as symmetric square matrix. If ``None``, then the
+            ``num_qubits`` argument must be set and a random IQP circuit will be generated.
+        num_qubits: If no interactions are given, construct a random IQP circuit with this
+            number of qubits. This argument cannot be passed at the same time as ``interactions``.
+
```
alexanderivrii: I am wondering if it would be cleaner to have two separate functions: one that construct an IQP circuit given a list of interactions, and another that returns a random IQP circuit given the number of qubits.

crates/accelerate/src/circuit_library/iqp.rs
```diff
@@ -0,0 +1,159 @@
...
+    num_qubits: u32,
+    interactions: Option<PyReadonlyArray2<i64>>,
+) -> PyResult<CircuitData> {
+    let array = match interactions {
+        Some(matrix) => {
+            let view = matrix.as_array();
+            if !check_symmetric(&view) {
+                return Err(CircuitError::new_err("IQP matrix must be symmetric."));
+            }
+            view.to_owned()
```
alexanderivrii: Do we need to copy the interaction matrix here?

crates/accelerate/src/circuit_library/iqp.rs
```diff
@@ -0,0 +1,159 @@
...
+    // We then build the matrix of dimension NxN by reading from the ``values`` vector.
+    // * since the matrix is symmetric, we treat (i, j) and (j, i) equally, which we implement
+    //   here by sorting the indices before accessing the value vector
+    // * in each row, the offset with which we access ``values`` changes: in row K, we need to
+    //   start reading from index: \sum_{k=0}^K N-k
+    Array2::from_shape_fn((num_qubits, num_qubits), |(i, j)| {
+        let (low, high) = if i < j { (i, j) } else { (j, i) };
+        let offset = low * (2 * num_qubits - low + 1) / 2;
+        values[offset + high - low]
+    })
```
alexanderivrii: Out of curiosity, is this faster than a more naive approach in https://github.com/Qiskit/qiskit/blob/fbfe7388f3efdaaeb70dc96aa35f649c1897c5d4/crates/accelerate/src/synthesis/clifford/random_clifford.rs#L43-L53.

crates/accelerate/src/circuit_library/iqp.rs
```diff
@@ -0,0 +1,159 @@
...
+    // triangular matrix. This contains N(N+1)/2 unique random values, stored as
+    //    [v_0 v_1 ... v_{N-1}]
+    //    [    v_N ... v_{2N-2}]
+    //             ...
+    //    [            v_{-1}]
+    let num_qubits = num_qubits as usize;
+    let num_values = num_qubits * (num_qubits + 1) / 2;
+    let values: Vec<i64> = (0..num_values)
+        .map(|_| rand::thread_rng().gen_range(0..8) as i64)
+        .collect();
```
alexanderivrii: I think it would make sense to seed the random number generator, like we do for random cliffords for instance.

crates/accelerate/src/circuit_library/iqp.rs
```diff
@@ -0,0 +1,159 @@
...
+            .map(move |j| (j, interactions[(i, j)]))
+            .filter(move |(_, value)| value % 4 != 0)
+            .map(move |(j, value)| {
+                (
+                    StandardGate::CPhaseGate,
+                    smallvec![Param::Float(PI2 * value as f64)],
+                    smallvec![Qubit(i as u32), Qubit(j as u32)],
+                )
+            })
+    });
```
alexanderivrii, line 55: wow! the rustiness of this code is exquisite.

crates/accelerate/src/circuit_library/iqp.rs
alexanderivrii: and a follow-up question: where are the values below the diagonal even used?

alexanderivrii (commented):
Overall this looks great. One question is why we need the new "random IQP" mode, is this used anywhere in Qiskit? In addition, personally I would prefer to have a separate function that returns a random IQP circuit.

test/python/circuit/library/test_iqp.py
Cryoris: yep, exactly -- though that description is a bit weird, I'll update it

qiskit/circuit/library/iqp.py
Cryoris: Good point: I put them in the same function as (1) I only wanted to expose one function from Rust and (2) from reading the paper it seems like people would be mostly interested in random IQP circuits. But we can also split it into two functions Python-side, if you prefer 🙂 

Or we can promote the random circuits by changing the signature to be
```python
iqp(num_qubits, seed=None, interactions=None)
```
such that the primary use would be just `iqp(N)` for a random `N` qubit IQP circuit.

crates/accelerate/src/circuit_library/iqp.rs
Cryoris: Hmm no we shouldn't need to. I think I did that because Rust was unhappy about the match-arms returning different types (as `generate_random_interactions` returns an owned instance). But I should be able to refactor this by moving the function call into the arms 👍🏻

crates/accelerate/src/circuit_library/iqp.rs
Cryoris: * I used the `from_shape_fn` to avoid the matrix initialization with zeros, before we overwrite the values later 🙂 but I can benchmark the difference (I assume it is minimal).
* They are not, see also this comment: https://github.com/Qiskit/qiskit/pull/13241/files/84d92bc2e9ba2a3089ed8126481744d803197429#r1790125684 But since we have to fill the matrix with some values, I'm generating the full, valid IQP matrix here

Cryoris: In fact, the "naive" approach is even faster, I guess because it can omit the check "i < j"? 🤔  Here's the times (in `s`) to create random IQP circuits with `size` qubits
```
size   from_shape_fn   for-loops
----------------------------------
100    0.001           0.001
500    0.012           0.011
1000   0.057           0.048
5000   2.613           2.405
```
I'll change to the for-loops approach 🙂

crates/accelerate/src/circuit_library/iqp.rs
Cryoris: Added in 1e103807e

crates/accelerate/src/circuit_library/iqp.rs
Cryoris: Added in 1e103807e

crates/accelerate/src/circuit_library/iqp.rs
Cryoris: I tried but I couldn't get it to work without cloning... I would guess that's because we create an iterator in `iqp` that needs to own the matrix to work 🙂

test/python/circuit/library/test_iqp.py
Cryoris: Fixed in 46d9ca042

Cryoris:
As discussed offline I split the `iqp` function into `iqp` and `random_iqp` to have a separate function for the random generation of IQP circuits 🙂

crates/accelerate/src/circuit_library/iqp.rs
```diff
@@ -0,0 +1,164 @@
...
+        None => Pcg64Mcg::from_entropy(),
+    };
+
+    let mut mat = Array2::zeros((num_qubits, num_qubits));
+    for i in 0..num_qubits {
+        for j in 0..i {
+            mat[[i, j]] = rng.gen_range(0..8) as i64;
+            mat[[j, i]] = mat[[i, j]];
+        }
+    }
```
alexanderivrii: As discussed with @Cryoris, we should also allow producing non-zero elements on the diagonal.

crates/accelerate/src/circuit_library/mod.rs
```diff
@@ -13,12 +13,15 @@
...
 mod entanglement;
+mod iqp;
 mod pauli_feature_map;
 mod quantum_volume;
 
 pub fn circuit_library(m: &Bound<PyModule>) -> PyResult<()> {
     m.add_wrapped(wrap_pyfunction!(pauli_feature_map::pauli_feature_map))?;
     m.add_wrapped(wrap_pyfunction!(entanglement::get_entangler_map))?;
+    m.add_wrapped(wrap_pyfunction!(iqp::py_iqp))?;
+    m.add_wrapped(wrap_pyfunction!(iqp::py_random_iqp))?;
```
alexanderivrii, line 24: I don't think it matters too much, but this uses a slightly different naming convention from most of the other circuit library code (i.e. here the functions are prefixed with `py`).

alexanderivrii (commented):
This looks good. Just one comment about generating random IQP instances.

crates/accelerate/src/circuit_library/iqp.rs
Cryoris: Resolved with e7258c4

qiskit/circuit/library/iqp.py
Cryoris: done in e7258c4

alexanderivrii (approved):
Thanks for the quick fix, LGTM!

PR #13245, opened 2024-09-30
Change Qiskit bot for docs team

Eric-Arellano:
Abby C and Becky (technical writers) want to start copy editing release notes before the final release. We can use Qiskit Bot to notify them.

Meanwhile, it's noisy for Abby M to get notifications.

### DIFF
--- a/qiskit_bot.yaml
+++ b/qiskit_bot.yaml
@@ -31,6 +31,7 @@ notifications:
     "(?!.*pulse.*)\\bvisualization\\b":
         - "@enavarro51"
     "^docs/":
-        - "@javabster"
         - "@Eric-Arellano"
+        - "@abbycross"
+        - "@beckykd"
 always_notify: true


qiskit_bot.yaml
```diff
@@ -31,6 +31,7 @@ notifications:
     "(?!.*pulse.*)\\bvisualization\\b":
         - "@enavarro51"
     "^docs/":
```
Eric-Arellano, line 33: This regex captures release prep PRs thanks to `conf.py`'s `release` changing.

jakelishman (commented):
I'm fine if you want to go ahead with notifications for this, but beware that this is likely to be the majority of PRs that come into Terra, and that we can't hold up merging PRs to allow copy-editing of the release notes, which will mean there'll still likely need to be an all-release-notes copy-edit step later.

We do an internal kind-of copy-edit step as part of the preparations for each release - for patch releases, it's just small (since there's only a couple of release notes for a patch release) so there's a bit less space for mistakes, but your input would be appreciated, and help take the load off the Qiskit team, during the fix-up process for minor releases, which typically happens between the release candidate and the final release.

jakelishman (approved):
Oh wait sorry, I realised that I misread where you'd put this in the regex - I thought it was in `releasenotes`.

Yeah, this is probably the best notification place, if you guys are ok with it.  It might be a little noisier than ideal for you, but we can always revisit if you want.

PR #13247, opened 2024-09-30
Improve deprecation of unit and duration

mtreinish:
### Summary

The unit and duration attributes of QuantumCircuit and Instruction were introduced in #13224 however the way the PR went about this was a bit to abbreviated and it had several downsides. The most noticable was that it introduced a significant runtime regression as the rust internals were accessing the deprecated getter. This is fixed so that the rust code doesn't try to access any deprecated attributes. However, more importantly there were several edge cases not accoutned for in the PR. Mainly, DAGCircuit's matching attribute were not deprecated, but then also specific use cases of the attributes were not addressed. The primary driver of the per instruction duration and unit were the timeline visualization which didn't have a method to function without these data attributes. This commit addresses this by adding a new target argument that is used to provide a backend's target which is the source of truth for durations in a post unit/duration `Instruction`. The other notable edge case is the `Delay` instruction which will need to retain the duration and unit attributes as the fields are integral for that instruction. The `Delay` is updated to handle them explicitly outside of the `Instrution` data model to avoid complications around the deprecation.

### Details and comments


qiskit/circuit/instruction.py
```diff
@@ -103,7 +103,21 @@ def __init__(self, name, num_qubits, num_clbits, params, duration=None, unit="dt
         # list of instructions (and their contexts) that this instruction is composed of
         # empty definition means opaque or fundamental instruction
         self._definition = None
+        if duration is not None:
+            warnings.warn(
+                "Setting a custom duration per instruction is deprecated as of qiskit "
```
ElePT:
```suggestion
                "Setting a custom duration per instruction is deprecated as of Qiskit "
```

qiskit/circuit/instruction.py
```diff
@@ -103,7 +103,21 @@ def __init__(self, name, num_qubits, num_clbits, params, duration=None, unit="dt
...
+                "1.3.0. It will be removed in Qiskit 2.0.0.",
+                DeprecationWarning,
+                stacklevel=2,
+            )
         self._duration = duration
+        if unit is not None and unit != "dt":
+            warnings.warn(
+                "Setting a custom duration per instruction is deprecated as of qiskit "
```
ElePT:
```suggestion
                "Setting a custom duration per instruction is deprecated as of Qiskit "
```

qiskit/circuit/instruction.py
ElePT: Maybe we could add a line about using the target instead?

qiskit/visualization/timeline/core.py
```diff
@@ -138,11 +139,13 @@ def add_data(self, data: drawings.ElementaryData):
...
 
     # pylint: disable=cyclic-import
-    def load_program(self, program: circuit.QuantumCircuit):
+    def load_program(self, program: circuit.QuantumCircuit, target: Target | None = None):
         """Load quantum circuit and create drawing..
 
         Args:
             program: Scheduled circuit object to draw.
+            target: The target for the backend the circuit is scheduled for
+                this contains the instruction durations.
```
ElePT:
```suggestion
            target: The target the circuit is scheduled for. This contains backend information  
                including the instruction durations used in scheduling.
```

qiskit/visualization/timeline/interface.py
```diff
@@ -361,6 +365,14 @@ def draw(
     temp_style = stylesheet.QiskitTimelineStyle()
     temp_style.update(style or stylesheet.IQXStandard())
 
+    if target is None:
+        warnings.warn(
+            "Target is not specified in Qiskit 2.0.0 this will be required to get the duration of "
+            "instructions.",
```
ElePT:
```suggestion
            "Target is not specified. In Qiskit 2.0.0 this will be required to get the duration of "
            "instructions.",
```

releasenotes/notes/deprecate-unit-duration-48b76c957bac5691.yaml
```diff
@@ -1,4 +1,12 @@
 ---
+features_visualization:
+  - |
+    The :func:`.timeline_drawer` visualization function has a new argument
+    ``target``, used to specify a :class:`.Target` object for the
+    visualization. By default the function using the
+    :attr:`.Instruction.duration` to get the duration of a given isntruction,
```
ElePT:
```suggestion
    visualization. By default the function used the
    :attr:`.Instruction.duration` to get the duration of a given instruction,
```

ElePT (commented):
I see that the deprecation was trickier than expected. I left a few comments that are hopefully useful.

releasenotes/notes/deprecate-unit-duration-48b76c957bac5691.yaml
```diff
@@ -10,16 +18,28 @@ deprecations_circuits:
...
-    of a gate is the :class:`.BackendV2` :class:`.Target` which contains
-    the duration for each instruction supported on the backend. The duration of
-    an instruction is not something that's typically user adjustable and is
-    an immutable property of the backend. If you were previously using this
-    capability to experiment with different durations for gates you can
-    mutate the :attr:`.InstructionProperties.duration` field in a given
-    :class:`.Target` to set a custom duration for an instruction on a backend
-    (the unit is always in seconds in the :class:`.Target`).
+    have been deprecated and will be removed in Qiskit 2.0.0. This includes
+    setting the ``unit`` or ``duration`` arguments for any :class:`.Instruction`
```
ElePT: This should solve the ambiguity sphinx complains about (unless I didn't choose the right `Instruction`):

```suggestion
    setting the ``unit`` or ``duration`` arguments for any :class:`qiskit.circuit.Instruction`
```

ElePT (commented):
Mostly LGTM, just missing running black to fix CI. One comment is that it might be nicer to point to alternatives in the deprecation messages that say that unit/duration are deprecated. Something like "If this information is required for transpilation, it will be extracted from the target directly".

ElePT (approved):
LGTM!!

PR #13248, opened 2024-09-30
Fix broken link to Runarsson and Yao

Eric-Arellano:
We already fixed Qiskit <0.46 by manually changing the Sphinx artifacts we upload to Box.


jakelishman:
The 0.46 branch is locked now and we'll be making no more releases from it or commits to it (a month and a half after our initial deadline to drop support, but still...) - is this something you can do for 0.46 generation-side in the same way as prior versions?

Eric-Arellano:
Ah, sounds good! Yep, I can fix this directly in the HTML. That's quicker for us anyways. I didn't realize the branch is locked. Thanks!

PR #13249, opened 2024-09-30
Fabiana campanari patch 1

FabianaCampanari:


jakelishman:
This is now the third attempt you have made to falsely put your sponsorship link into the project, and I have warned you before.

This spam is not welcome here.

PR #13250, opened 2024-10-01
Temporarily pin ``symengine==0.11.0`` to unblock CI

Cryoris:
### Summary

With the release of symengine 0.13 we're running into incompatibility issues when loading/storing data in between symengine versions (e.g. in between 0.11 and 0.13). This commit pins temporarily pins the version to unblock CI, until a more permanent solution is found.

### DIFF
--- a/constraints.txt
+++ b/constraints.txt
@@ -19,3 +19,7 @@ z3-solver==4.12.2.0; platform_system=="Darwin"
 # usage, but our test suite uses an exact reference file that uses the
 # pydot 3 output, so we need to enforce that during tests.
 pydot>=3.0.0
+
+# Pin Symengine for incompatibilites loading and storing data in between
+# versions 0.13 and before.
+symengine==0.11.0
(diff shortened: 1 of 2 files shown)


ElePT (dismissed):
sounds good to me

ElePT:
Hmmm @Cryoris the fix is not fixing.

jakelishman:
Elena: that might well be CI pulling old versions from the cache, which indicates a(nother) problem in the cache key - we should have invalidated the caches if the build steps for them changed.

jakelishman:
Oh, actually, even if it's a cache miss, it'll fail - for now, also put this constraint into `test/qpy_compat/qpy_test_constraints.txt`, and we can figure out the rest of the caching separately (hopefully).

*edit*: I've pushed that change - fingers crossed I got that right, and I don't also need to invalidate the cache by other means (I'm worried that I do).

jakelishman:
Ah, I should have constrained to `<0.13` in the QPY tests, since some versions of Qiskit had upper-bounds pins.  Either way, since #13251 is up now, I'm going to shift to focussing on that.

Cryoris:
Closing in favor of #13251, which looks like it's working already 🙂

PR #10820, opened 2023-09-12
Add `use_symengine` option to `qpy.dump`

ElePT:
### Summary
This is a proposal to fix #10705 by adding the `use_symengine` flag to `qpy.dump()`. This will improve the serialization performance for both circuits containing unbound `ParameterExpressions` and pulse schedules that rely on custom symbolic expressions.

### Details and comments
The `use_symengine` flag is now stored as part of the file header, as it is a global option used by both circuit and schedule payloads. I have updated the documentation accordingly. I am not that familiar with the schedule "side" of QPY, so let me know if you can think of more cases I can add to the test, or whether I am missing anything (@nkanazawa1989).

PR #13251, opened 2024-10-01
Workaround symengine serialization payload incompatibility

mtreinish:
### Summary

In QPY we rely on symengine's internal serialization to represent the internal symbolic expression stored inside a ParameterExpression object. However, this format is nominally symengine version specific and will raise an error if there is a mismatch between the version used to generate the payload and what is trying to read it. This became an issue in the recent symengine 0.13 release which started to raise an error when people installed it and tried to load QPY payloads across the versions. This makes the symengine serialization unsuitable for use in QPY because it's supposed to be independent of these kind of concerns, especially when QPY is used in a server-client model where you don't necessarily control the installed environment of symengine.

To correctly address this issue we'll need a new version of the QPY format that owns the serialization format of ParameterExpressions directly instead of relying on symengine which doesn't offer a compatibility guarantee on the format. However this won't be quick solution and users are encountering issues since the release of 0.13. This commit introduces a workaround for this specific instance of the mismatch. It turns out the payload format between 0.11 and 0.13 is completely unchanged except for the version number. So before passing the parameter expression payload to symengine for deserialization this commit checks the versions numbers are the same, if they're not it checks that we're dealing with 0.11 or 0.13, and if so it changes the version number in the payloads header appropriately. If the version number is outside those bounds it raises an exception because while this hack is known to be safe for translating between symengine 0.11 and 0.13, it's not possible to know for a future version whether the payload format changed or not.

Longer term we will need a proper fix in qpy version 13 that introduces a qiskit native serialization format for parameter expression instead of relying on symengine or sympy to do it for us.

### DIFF
--- a/qiskit/qpy/binary_io/value.py
+++ b/qiskit/qpy/binary_io/value.py
@@ -290,6 +290,29 @@ def _read_parameter_expression_v3(file_obj, vectors, use_symengine):
 
     payload = file_obj.read(data.expr_size)
     if use_symengine:
+        # This is a horrible hack to workaround the symengine version checking
+        # it's deserialization does. There were no changes to the serialization
+        # format between 0.11 and 0.13 but the deserializer checks that it can't
+        # load across a major or minor version boundary. This works around it
+        # by just lying about the generating version.
+        symengine_version = symengine.__version__.split(".")
+        major = payload[2]
+        minor = payload[3]
+        if int(symengine_version[1]) != minor:
+            if minor not in (11, 13):
+                raise exceptions.QpyError(
+                    f"Incompatible symengine version {major}.{minor} used to generate the QPY "
+                    "payload"
+                )
+            minor_version = int(symengine_version[1])
+            if minor_version not in (11, 13):
+                raise exceptions.QpyError(
+                    f"Incompatible installed symengine version {symengine.__version__} to load "
+                    "this QPY payload"
+                )
+            payload = bytearray(payload)
+            payload[3] = minor_version
+            payload = bytes(payload)
         expr_ = load_basic(payload)
     else:
         from sympy.parsing.sympy_parser import parse_expr
--- a/qiskit/qpy/binary_io/value.py
+++ b/qiskit/qpy/binary_io/value.py
@@ -290,30 +290,7 @@ def _read_parameter_expression_v3(file_obj, vectors, use_symengine):
 
     payload = file_obj.read(data.expr_size)
     if use_symengine:
-        # This is a horrible hack to workaround the symengine version checking
-        # it's deserialization does. There were no changes to the serialization
-        # format between 0.11 and 0.13 but the deserializer checks that it can't
-        # load across a major or minor version boundary. This works around it
-        # by just lying about the generating version.
-        symengine_version = symengine.__version__.split(".")
-        major = payload[2]
-        minor = payload[3]
-        if int(symengine_version[1]) != minor:
-            if minor not in (11, 13):
-                raise exceptions.QpyError(
-                    f"Incompatible symengine version {major}.{minor} used to generate the QPY "
-                    "payload"
-                )
-            minor_version = int(symengine_version[1])
-            if minor_version not in (11, 13):
-                raise exceptions.QpyError(
-                    f"Incompatible installed symengine version {symengine.__version__} to load "
-                    "this QPY payload"
-                )
-            payload = bytearray(payload)
-            payload[3] = minor_version
-            payload = bytes(payload)
-        expr_ = load_basic(payload)
+        expr_ = common.load_symengine_payload(payload)
     else:
         from sympy.parsing.sympy_parser import parse_expr
 
--- a/qiskit/qpy/binary_io/value.py
+++ b/qiskit/qpy/binary_io/value.py
@@ -20,9 +20,6 @@ import uuid
 
 import numpy as np
 import symengine
-from symengine.lib.symengine_wrapper import (  # pylint: disable = no-name-in-module
-    load_basic,
-)
 
 
 from qiskit.circuit import CASE_DEFAULT, Clbit, ClassicalRegister
--- a/releasenotes/notes/fix-qpy-symengine-compat-858970a9a1d6bc14.yaml
+++ b/releasenotes/notes/fix-qpy-symengine-compat-858970a9a1d6bc14.yaml
@@ -1,50 +1,63 @@
 ---
 fixes:
   - |
-    Fixed an issue with :func:`.qpy.load` when loading a QPY payload that
-    contains a :class:`.ParameterExpression` that was generated when symengine
-    0.11 was installed and trying to load it in an environment where symengine
-    0.13 (the latest symengine release as of this qiskit release) or vice versa.
-    Previously, an error would have been raised by symengine around this version
-    mismatch. This has been worked around for these two specific symengine versions
-    but if you're trying to use different versions of symengine and there is a
-    mismatch with this version of Qiskit this might not work. You will need to
-    install Qiskit >1.3.0 to fix this mismatch issue more broadly for any potential
-    future version of symengine.
+    Fixed an issue with :func:`.qpy.load` when loading a QPY file containing
+    a :class:`.ParameterExpression`, if the versions of ``symengine`` installed
+    in the generating and loading environments were not the same.  For example,
+    if a QPY file containing :class:`.ParameterExpression`\ s was generated
+    using Qiskit 1.2.2 with ``symengine==0.11.0`` installed, Qiskit 1.2.2 with
+    ``syengine==0.13.0`` installed would be unable to load it.
+
+    Previously, an error would have been raised by ``symengine`` around this
+    version mismatch. This has been worked around for ``symengine`` 0.11 and
+    0.13 (there was no 0.12), but if you're trying to use different versions of
+    ``symengine`` and there is a mismatch, this version of Qiskit still might not
+    work.
 issues:
   - |
-    When dumping a QPY file using :func:`.qpy.dump` with the ``version=10`` flag and there are
-    unbound :class:`.ParameterExpression` in the circuit the output file
-    will not be loadable by Qiskit 0.45.x or 0.46.x due to incompatibilities
-    in the symengine serialization used between the symengine libraries used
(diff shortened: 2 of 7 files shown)


qiskit/qpy/binary_io/value.py
```diff
@@ -290,6 +290,29 @@ def _read_parameter_expression_v3(file_obj, vectors, use_symengine):
 
     payload = file_obj.read(data.expr_size)
     if use_symengine:
+        # This is a horrible hack to workaround the symengine version checking
+        # it's deserialization does. There were no changes to the serialization
+        # format between 0.11 and 0.13 but the deserializer checks that it can't
+        # load across a major or minor version boundary. This works around it
+        # by just lying about the generating version.
```
jakelishman: I'm roughly concerned that Symengine 0.13 might have added new opcodes or something to the serialisation format, without that necessarily needing a format change.  Do we have any knowledge of whether that's the case or not?

Of course, the worst case would be that some opcodes change in a way that _doesn't_ fail the load.  I doubt we can fully determine that right now, and since the most pressing concern is to get single `Parameter` instances transmitting properly, I'm ok to move ahead, and revisit if we need to.

qiskit/qpy/binary_io/value.py
```diff
@@ -290,6 +290,29 @@ def _read_parameter_expression_v3(file_obj, vectors, use_symengine):
...
+                raise exceptions.QpyError(
+                    f"Incompatible symengine version {major}.{minor} used to generate the QPY "
+                    "payload"
+                )
+            minor_version = int(symengine_version[1])
+            if minor_version not in (11, 13):
+                raise exceptions.QpyError(
+                    f"Incompatible installed symengine version {symengine.__version__} to load "
+                    "this QPY payload"
+                )
```
jakelishman: The ordering of this means that we can cope with generation and load from (say) a hypothetical symengine 0.14, but is it worth us risking that?  We could put `symengine>=0.11,<=0.13` in our requirements, potentially.

Not sure how I feel about that, just asking the question.

jakelishman (commented):
I'm just pulling this locally to run a test myself, since this isn't something we're going to be able to test well in CI without adding a whole other test job, but in the meantime, two questions.

I'm fine with leaving the PR as-is if it works, though.

jakelishman:
In local testing, I generated the QPY backwards compatibility tests using this branch, with Symengine 0.11 and Symengine 0.13, and then attempted to reload them with both.  I found that the schedule code (`qiskit/qpy/binary_io/schedules.py:109`) has a similar loader, which is similarly not aware of multiple symengine versions and also needs a fix.

jakelishman:
Probably we can find a way to slightly unify those two code paths, since they looked pretty similar to me.

qiskit/qpy/binary_io/value.py
mtreinish: I'm mildly optimistic that we should be ok here because we're not exposing the full set of symengine's capabilities and the pieces we do support are pretty basic to the library and likely didn't change much.

Skmming through https://github.com/symengine/symengine/blob/master/symengine/serialize-cereal.h I think it just uses the type names, I can't see how else it would work unless it uses the order the overload is added in the source file as an opcode (which we'd be fine on because there were not any substantial changes to that file). But I am not really familiar with the cereal library so this is just conjecture on my part.

mtreinish:
> In local testing, I generated the QPY backwards compatibility tests using this branch, with Symengine 0.11 and Symengine 0.13, and then attempted to reload them with both. I found that the schedule code (qiskit/qpy/binary_io/schedules.py:109) has a similar loader, which is similarly not aware of multiple symengine versions and also needs a fix.

Fixed in: https://github.com/Qiskit/qiskit/pull/13251/commits/ea69a4590ba058514cf7615e26c5113261c3e2ee

qiskit/qpy/binary_io/value.py
mtreinish: Can we, did I do the logic incorrectly? My intent was that if we have a version mismatch between the payload and the installed library and if either the payload or the installed version is not using 0.11 or 0.13 we'll raise the error. 

There is an edge case here if the major version != 0, but I figured this was unlikely to be an issue (especially with a minor version of `11` or `13`) in any reasonable timescale.

jakelishman: Yeah, I'm saying we can attempt to load if there's no payload mismatch and both were 0.14, but given we know there's problems, would we prefer just to forbid 0.14 from entering at all, rather than have it work iff there's a match?

Yeah, I saw the major version thing too, but was just assuming that we'll fix this before that happened.

jakelishman: (and by "yeah" I mean I'm agreeing with you - I think you did the logic right)

mtreinish: Ah ok, now I get what you're saying. My initial thought was if there is no mismatch I think we're fine as symengine should be totally fine if everything is the same version. I guess it's more a question of whether we think it's sufficiently a risk of foot gunning by generating a payload with 0.14.0 that will fail if someone is using any other version later

qiskit/qpy/binary_io/value.py
jakelishman: Oh yeah, I didn't really think much about how `ParameterExpression` limits the surface area we're exposed to.  That's definitely a positive. Let's move forwards with this - I'm not sure that me wringing my hands about it is going to get us any new knowledge, and in the meantime, this is hurting real deployments of Qiskit.

qiskit/qpy/binary_io/value.py
jakelishman: Yeah, I was kind of hoping you'd have a strong opinion about that last point so I'd be absolved of needing to develop one.  In lieu of anything better, let's stick with the PR as written - upper bounds aren't ideal, and I don't particularly like introducing one within a patch release.

mtreinish: Yeah I think that's what we're talking about. That's the only option really is either add a cap to the requirements file in 1.2.3, or we do it in the python code so that `qpy.dump()` fails if you have symenging > 0.14 installed. It might be prudent. I was hesitant to do that in this PR because I was worried about python 3.13 support. But since 0.13 has 3.13 support maybe it's worth it here.

jakelishman (commented):
I tested this again locally with the same procedure as above, and verified that it works.  The last question before I approve:

The `use_symengine` argument to QPY serialisation was introduced in #10820, released in Terra 0.45.0, which allowed `symengine>=0.9`.  `use_symengine=False` was the default back then, but it _was_ possible to set `use_symengine=True`, and if I set that during the QPY file generation, then this branch will fail to load the files.  I can't just add `9` to the allowed `minor` versions we can load from, because it seems like between 0.9 and 0.11 the format _did_ change, and we segfault symengine if we patch the version.

Are we prepared to accept this problem as "good enough", or do we want to look for anything deeper?  Ideologically it's not nice to call it "good enough", but pragmatically I don't think I can see an alternative solution for us in the short term.  Assuming we accept that issue, should we put a "known issues" release note in about it, or potentially document it in the QPY documentation about the backwards compatibility guarantees?

jakelishman:
To finish the thought - `use_symengine=True` became the default in 1.0.0, which also required `symengine>=0.11.0`, so we can at least say "using QPY with the defaults to `load` remains fully backwards compatible" (up to the JSON story, ofc, which has always explicitly warned about the backwards compat).

mtreinish:
Ok, I've updated the PR in: https://github.com/Qiskit/qiskit/pull/13251/commits/e089d5c2fc01508b1acf709041a799e074b611b6 to be more explicit about the symengine versions coming from 0.45.x and 0.46.x with `use_symengine=True`. I also added warnings to the docs and release notes about the known issues around this.

qiskit/qpy/binary_io/value.py
mtreinish: I opted to add the cap in: https://github.com/Qiskit/qiskit/pull/13251/commits/7aae9849ec647816a3a790a24ae6e79931122f24

jakelishman: For future posterity: Matt and I talked offline before this, and thought that pre-emptively capping is actually the lesser of the two evils.  We don't like introducing an upper bound in a patch release, but we've had upper bounds on `symengine` before, and the hack we're doing to enable compatibility is pretty dangerous, since symengine's deserialisation code can hard crash on malformed payloads.  It feels safer to try and restrict symengine 0.14 from entering (as best as we can), rather than _hoping_ that nothing will break when it's released.

jakelishman (dismissed):

releasenotes/notes/fix-qpy-symengine-compat-858970a9a1d6bc14.yaml
```diff
@@ -0,0 +1,63 @@
...
+    version mismatch. This has been worked around for ``symengine`` 0.11 and
+    0.13 (there was no 0.12), but if you're trying to use different versions of
+    ``symengine`` and there is a mismatch, this version of Qiskit still might not
+    work.
+issues:
+  - |
+    Versions of Qiskit before 1.2.3 will not be able to load QPY files dumped
+    using :func:`.qpy.dump`, even with ``version`` set appropriately, if:
+
+    * there are unbound :class:`.ParameterExpression`\ s in the circuit,
```
mtreinish: I realize this was in my text too, but it can be a scheduleblock too.

```suggestion
    * there are unbound :class:`.ParameterExpression`\ s in the QPY file,
```

releasenotes/notes/fix-qpy-symengine-compat-858970a9a1d6bc14.yaml
```diff
@@ -0,0 +1,63 @@
...
+    * the ``use_symengine=True`` flag was set (which is the default in Qiskit >=
+      1.0.0) in :func:`.qpy.dump`,
+    * the version of ``symengine`` installed in the generating and loading
+      environments are not within the same minor version (e.g. both 0.11, or
+      both 0.13).
```
mtreinish: The `e.g.` example here reads a bit unclearly to me at first because it's an example of the working condition not the not-working condition being listed here. I'd probably just remove it
```suggestion
      environments are not within the same minor version.
```

jakelishman (approved):

PR #13253, opened 2024-10-01
Change style of DAG generated in dag_drawer()

### LINKED ISSUE #13106: Allow for changing colors in dag circuit drawer
### What should we add?

It would be nice if we can change colors for the DAG drawer in a similar manner to the circuit drawer

emilkovacev:
### Summary
Closes #13106 

Adds non-breaking interface to `dag_drawer()` to allow Qiskit users to customize the colors of DAGs.

### Details and comments

I created stylesheets for `dag_drawer()`, that works similarly to how `circuit_drawer` does.

Here is an example:
```
from qiskit import QuantumRegister, ClassicalRegister, QuantumCircuit
from qiskit.visualization import dag_drawer
 

q = QuantumRegister(3, 'q')
c = ClassicalRegister(3, 'c')
circ = QuantumCircuit(q, c)
circ.h(q[0])
circ.cx(q[0], q[1])
circ.measure(q[0], c[0])
circ.rz(0.5, q[1])

dag = circuit_to_dag(circ)

style = {
    # "fontsize": 12,
    # "bgcolor": "white",
    # "dpi": 10,
    # "pad": 0,
    # "nodecolor": "green",
    "inputnodecolor": "pink",
    # "inputnodefontcolor": "white",
    "outputnodecolor": "lightblue",
    # "outputnodefontcolor": "white",
    "opnodecolor": "red",
    # "opnodefontcolor": "white",
    # "edgecolor": "black",
    # "qubitedgecolor": "red",
    # "clbitedgecolor": "black",
}

dag_drawer(dag, style=style)
```

The code above generates this DAG:

Screenshot 2024-10-09 at 11 39 46 AM


qiskit/visualization/dag_visualization.py
```diff
@@ -73,7 +74,17 @@
...
+def dag_drawer(
+    dag,
+    scale=0.7,
+    filename=None,
+    style="color",
+    node_attr_fn: Union[
+        Callable[[Union[DAGOpNode, DAGInNode, DAGOutNode]], dict[str, str]], None
+    ] = None,
+    edge_attr_fn: Union[Callable[[Union[Qubit, Clbit]], dict[str, str]], None] = None,
+    graph_attr: Union[dict[str, str], None] = None,
```
mtreinish: I'm wondering if we want to keep the interface this low level? Right now it's completely free form and this lets you override everything `dag_drawer` will do out of the box including changing node labels. But I was thinking for this issue having a bit more prescriptive of an API that allows customizing the colors or styles might be a bit better so people don't need to learn how graphviz works. One option would be something like
 ```
 input_node_color: str,
 output_node_color: str,
 op_node_color: str,
 ```
 and then document what are valid choices for each string. The other is to take a page from the matplotlib circuit drawer and use style sheets see: https://github.com/Qiskit/qiskit/blob/492005e67c8153189352ef6ef8f9fae8b5b0c407/qiskit/visualization/circuit/qcstyle.py#L106-L151 where you take an input object that defines the style customization options for the visualization and lets users build something more domain specific for customizing the output of the visualization.

mtreinish (commented):
Thanks for getting this started. I  left an inline comment about the high level interface. I'm a bit concerned with how low level the current interface is and that it would require understanding how graphviz is used internally by `dag_drawer`

qiskit/visualization/dag_visualization.py
emilkovacev: I can understand why that may be an issue. It would certainly allow people to customize the graphs more easily by giving a simpler API. However, by doing that, you lose some level of customization like in the example I provided. If that's not an issue, I can absolutely simplify the arguments or use stylesheets.

I could also restrict the **return value** of the input functions to `node_attr_fn`, `edge_attr_fn` and inputs to `graph_attr` to a certain subset of inputs from a defined style sheet. This would keep the flexibility, but add enough support documentation to prevent users from having to look elsewhere for the options needed to customize the DAG.

What do you think seems like a good approach?

Edit:
I kind of flip-flopped on this, so I am working on a style sheets implementation. I realized with this you cannot give nodes or edges a style based on the content of the node, but once again, maybe if it's not needed I can push what I have using style sheets.

emilkovacev: I sat on this for a while and changed my mind - I am in the process of implementing the stylesheet method. I like how it is similar to the matplotlib circuit drawer, and we can easily just create unique options for coloring certain nodes and edges differently.

qiskit/visualization/dag/dagstyle.py
```diff
@@ -0,0 +1,211 @@
...
+    # style={"name": "clifford", <other settings...>}
+    if isinstance(style, dict):
+        style_name = style.get("name", "color")
+    elif isinstance(style, str):
+        if style.endswith(".json"):
+            style_name = style[:-5]
+        else:
+            style_name = style
+    else:
+        raise exceptions.VisualizationError("Invalid style {style}")
```
raynelfss: I believe this might have been an f-string?
```suggestion
        raise exceptions.VisualizationError(f"Invalid style {style}")
```

qiskit/visualization/dag/dagstyle.py
```diff
@@ -0,0 +1,211 @@
...
+        path = Path(__file__).parent / "styles" / default_style_dict
+
+        with open(path, "r") as infile:
+            default_style = json.load(infile)
+
+        # set shortcuts, such as "ec" for "edgecolor"
+        self.style = StyleDict(**default_style)
+
+
+def load_style(style: dict | str = "color") -> StyleDict:
```
raynelfss: Should we instead be able to provide a file or directory here too? Since whenever we provide a string it searches for one. Perhaps we should be able to specify a specific Path or file.

qiskit/visualization/dag/dagstyle.py
```diff
@@ -0,0 +1,211 @@
...
+    elif isinstance(style, str):
+        if style.endswith(".json"):
+            style_name = style[:-5]
+        else:
+            style_name = style
+    else:
+        raise exceptions.VisualizationError("Invalid style {style}")
+
+    if style_name in ["color"]:
+        current_style = DefaultStyle().style
```
raynelfss: Maybe compere the strings directly here?
```suggestion
    if style_name == "color":
        current_style = DefaultStyle().style
```

qiskit/visualization/dag/dagstyle.py
```diff
@@ -0,0 +1,211 @@
...
+                        2,
+                    )
+                    break
+                except (OSError, FileNotFoundError):
+                    warn(
+                        f"Error loading JSON file '{path}'. Will use default style.",
+                        UserWarning,
+                        2,
+                    )
+                    break
```
raynelfss: This here can get a little convoluted. While providing a string name for a style that might be located at the styles directory is a good idea. We should be able to directly provide an instance of `File` or `Path` so that we don't need to perform this search actively and the user can just provide the Path themselves.

qiskit/visualization/dag_visualization.py
```diff
@@ -128,140 +154,185 @@ def dag_drawer(dag, scale=0.7, filename=None, style="color"):
...
+                    + cond_txt
+                    + str(args)[1:-1].replace("'", "")
+                    + ")"
                 )
-                if node.name == "barrier":
-                    n["style"] = "filled"
-                    n["fillcolor"] = "grey"
-                elif getattr(node.op, "_directive", False):
-                    n["style"] = "filled"
-                    n["fillcolor"] = "red"
```
raynelfss: Is there a reason why these two conditions were removed here?

raynelfss (commented):
This looks great, thank you for these additions. I wonder whether we should consider using `FileSpecifier` (such as File, Path, etc) as an alternative input for `load_style()`. I feel like it might benefit from not having to check whether the file exists or not. See the comments in my review.

qiskit/visualization/dag/dagstyle.py
emilkovacev: I agree, that seems a lot cleaner. This function definition was taken from the `load_style()` function in `qiskit/visualization/circuit/qcstyle.py`, should that be changed too or is that outside of the scope for this PR?

(I also noticed that the docstring here needs updating, I'll fix that too)

qiskit/visualization/dag/dagstyle.py
emilkovacev: I agree, similar to before this code uses the same logic as the other instance of load_style() in `qiskit/visualization/circuit/qcstyle.py`, but with small changes to accommodate for the different use case.

qiskit/visualization/dag_visualization.py
emilkovacev: This was a misunderstanding on my part about how `dag_drawer` behaved differently for DagDependency graphs, reintroduced rules in https://github.com/Qiskit/qiskit/pull/13253/commits/dfba4dd4db472f8e2bea3d44afdbbc86f70eb6bb.

qiskit/visualization/dag/dagstyle.py
raynelfss: Hmm... If the function definition for it is based on an existing function, then I'm okay with keeping it the way it is. After giving the old function a look it seems to be fine as long as the file exists under the styles directory. We could add support for files/paths as a follow-up, but since you still had to create a path in the method, it could've worked better with a path as an argument.

I do think that maybe this method could just expand on the original `load_style` as pointed out by @mtreinish, since it seems to be doing a lot of what's being done there already.

qiskit/visualization/dag/dagstyle.py
raynelfss: That's okay then, but what worries me now is the code duplication happening between `qiskit/visualization/circuit/qcstyle.py` `load_style()` and this. The functions don't seem to be so fundamentally different and a lot of the checks being done over there can be done pre-function call.

qiskit/visualization/dag/dagstyle.py
```diff
@@ -0,0 +1,211 @@
...
+
+import json
+import os
+from typing import Any
+from pathlib import Path
+from warnings import warn
+from qiskit.visualization import exceptions
+
+
+class StyleDict(dict):
```
raynelfss: Since there is already a StyleDict that exists in qcstyle.py, you might want to rename this into something different since it seems the fields happen to differ. Maybe go with something like `DAGStyleDict`.

qiskit/visualization/dag/dagstyle.py
```diff
@@ -0,0 +1,211 @@
...
+
+        for path in style_paths:
+            # expand ~ to the user directory and check if the file exists
+            exp_user = path.expanduser()
+            if os.path.isfile(exp_user):
+                try:
+                    with open(exp_user) as infile:
+                        json_style = json.load(infile)
+
+                    current_style = StyleDict(json_style)
```
raynelfss: This is the case where it might be a little difficult to reuse the `StyleDict` object if we plan on using `load_style` from `visualization/qc_style.py`. Perhaps we could create an inner method that is called by both this and `qc_style.py` `load_style` in which you can specify the type of `StyleDict` being used. But again, that might be a bit tricky with Python, though not impossible.

raynelfss (commented):
This looks better so far, I do have a concern about the reuse of code for the `dagstyle.py` file and some methods, so I left some tips to consider.

emilkovacev:
@raynelfss I attempted to resolve the code duplication issues in https://github.com/Qiskit/qiskit/pull/13253/commits/9d6b0ec1d23fa30ce3abca1943112a3661a656fd

* Moved `load_style`, `StyleDict`, and `DefaultStyle` to qiskit/visualization/style.py
* Formatted code that previously called `load_style` to inherit `StyleDict` and `DefaultStyle`, and passed the inherited type/object to `load_style`

These changes decouple the `load_style` code from qcstyle.py, which should also make it easier to consistently make features that load styles within the qiskit/visualization directory.

qiskit/visualization/style.py
```diff
@@ -0,0 +1,223 @@
...
+    DEFAULT_STYLE_NAME = None
+    DEFAULT_STYLE_PATH = None
+
+    def __init__(self):
+        raise NotImplementedError()
+
+
+def load_style(
+    style: Union[dict, str, None],
+    style_dict: type[StyleDict],
```
raynelfss, line 89: Maybe you'd want to rename this to `style_dict_type`? It makes more sense to have it in the name that this argument is type based.

raynelfss (dismissed):
This looks good to me, thank you for your work. I will look for a secondary reviewer to make sure visualization is in order :)

qiskit/visualization/dag/dagstyle.py
```diff
@@ -0,0 +1,103 @@
+# This code is part of Qiskit.
+#
+# (C) Copyright IBM 2019.
```
1ucian0:
```suggestion
# (C) Copyright IBM 2025.
```

qiskit/visualization/dag/styles/__init__.py
```diff
@@ -0,0 +1,13 @@
+# This code is part of Qiskit.
+#
+# (C) Copyright IBM 2014.
```
1ucian0:
```suggestion
# (C) Copyright IBM 2025.
```

qiskit/visualization/dag/__init__.py
```diff
@@ -0,0 +1,13 @@
+# This code is part of Qiskit.
+#
+# (C) Copyright IBM 2014.
```
raynelfss:
```suggestion
# (C) Copyright IBM 2025.
```

qiskit/visualization/style.py
```diff
@@ -0,0 +1,223 @@
...
+                        2,
+                    )
+                    break
+                except (OSError, FileNotFoundError):
+                    warn(
+                        f"Error loading JSON file '{path}'. Will use default style.",
+                        UserWarning,
+                        2,
+                    )
+                    break
```
raynelfss, line 199: I wonder if we gain any benefit from pre-computing the paths here, we might be able to fast-track it by performing each step separately and only continue if the next one hasn't been found. But I also understand this wasn't introduced by you, so it should stay as it is for now.

raynelfss (commented):
I quickly breezed through the code once again, just found one minor detail.

qiskit/visualization/style.py
```diff
@@ -0,0 +1,223 @@
+# This code is part of Qiskit.
+#
+# (C) Copyright IBM 2019.
```
emilkovacev: Should this copyright comment change as well?

raynelfss (dismissed):
Looks good to the best of my abilities!

qiskit/visualization/style.py
raynelfss: If it's a new file that you added, it should have the current year.

raynelfss (approved):

PR #13257, opened 2024-10-02
Replace parsed-literal with code-block

Eric-Arellano:
From `@frankharkins`:

> I'd avoid parsed-literal and just go with .. code-block:: text for outputs. Parsed literal might not behave how you expect: https://docutils.sourceforge.io/docs/ref/rst/directives.html#parsed-literal-block

`@jakelishman`'s response:

> Yeah, I don’t actually know why we ever used parsed-literal.  I assume there was a reason at some point, but just on the surface of it, I can’t imagine why we ever would have wanted to parse what’s in a text output for rst blocks

--

This change was generated with `rg 'parsed-literal::' -l | xargs sd 'parsed-literal::' 'code-block:: text'`.

This PR also fixes some circuit library docs that had module docstring that was not being built. Those docs are moved into their appropriate function docstrings.


frankharkins (commented):
Thanks for doing this. Only nit is that maybe outputs should always be text, even when they happen to be valid Python? Lack of highlighting might help users distinguish code and outputs, especially when the two blocks are back-to-back. It's also the default behaviour of ipython and jupyter notebooks.

qiskit/circuit/library/templates/nct/template_nct_2a_1.py
```diff
@@ -12,7 +12,7 @@
 
 """
 Template 2a_1:
-.. parsed-literal::
+.. code-block:: text
              ┌───┐┌───┐
         q_0: ┤ X ├┤ X ├
              └───┘└───┘
```
jakelishman: In a bunch of places, you've put an extra blank line after `.. code-block:: text` and before the start of the block, if there wasn't one there before.  Does that have an actual impact on the build / render?  If so, several of these template circuits are missing it.

If it _was_ necessary and the build it still succeeding, it's possible that all these `..` directives are failing to parse in the first place, because there's no blank line _before_ them either.

*edit*: actually, on closer inspection, all these are in the module docstring for the files, which isn't built at all.

qiskit/transpiler/passes/scheduling/base_scheduler.py
```diff
@@ -168,7 +168,7 @@ class BaseSchedulerTransform(TransformationPass):
         it may separately schedule qubit and classical register,
         insertion of the delay yields unnecessary longer total execution time.
 
-        .. parsed-literal::
+        .. code-block:: text
             : Quantum Circuit, first-xgate
             0 ░▒▒▒░░░░░░░░░░░░░░░
             1 ░▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒░
```
jakelishman: If the newlines matter, here's another file where they're missing.  If they do matter, maybe it's an indication that this file is missing from the docs?

releasenotes/notes/0.23/clifford_layered_synthesis-1a6b1038458ae8c3.yaml
```diff
@@ -8,7 +8,7 @@ features:
 
     For example, a 5-qubit Clifford circuit is decomposed into the following layers:
 
-    .. parsed-literal::
+    .. code-block:: text
              ┌─────┐┌─────┐┌────────┐┌─────┐┌─────┐┌─────┐┌─────┐┌────────┐
```
jakelishman: Newline?

test/python/transpiler/test_swap_strategy_router.py
```diff
@@ -62,7 +62,7 @@ def test_basic_zz(self):
 
         The expected circuit is:
 
-        ..parsed-literal::
+        ..code-block:: text
                                                            ┌────────────────┐
```
jakelishman: Newline?

(Not that this matters _at all_ since it's a test file.)

jakelishman (commented):
Thanks for doing this - `code-block:: text` _feels_ much more intended than `parsed-literal` for us.  I wonder what the initial reason was.

The only question is that you seem to have introduced newlines in lots of places, but there's a few missing.  I'm assuming that means that the newlines are important, which if so, means that there's a few places with substantial docstrings that don't appear to be included in the build.

I'm fine merging the PR as-is if that is the case, but we should follow up on restructuring to ensure that all the identified places _are_ built in if they're supposed to be.

Eric-Arellano:
Ah, I added the newlines in a follow up commit because the Sphinx build was failing without them. But I only fixed the places Sphinx complained about: I didn't exhaustively go through every file.

Even though Sphinx ignores those places you flagged, I'll still fix them today. We might want those docs in the future, and it reduces the risk of someone being confused when they copy-and-paste the code for new docs and it fails.

jakelishman:
Ah cool, thanks.  I suspect that the `base_scheduler` one is truly indicative of a problem - those docs look substantial, and we didn't catch the problem in the build, which I'm guessing means they're not being built in, but probably should be.  We can address that bit (adding it to the build somewhere) in a follow up.

Eric-Arellano:
@jakelishman how would you like me to fix the missing module docstring lint? Those aren't used by Sphinx, so it seems like a risk people will write good docs that end up getting ignored. Perhaps it's best to disable the lint?

jakelishman:
For the time being, I guess just put in a dummy one-line docstring or a `# pylint: disable=missing-module-docstring` suppression - it won't be the first nor the last in Qiskit.

Longer term, yeah, we should think about whether the module lint is truly adding anything - I suspect not, because most modules in Qiskit aren't public, so their docstrings aren't used.  The class and function docstring lints are probably more sensible (though it'd be better if we could blanket suppress them in tests).

Eric-Arellano:
Bump

qiskit/circuit/library/templates/rzx/rzx_zz2.py
```diff
@@ -34,7 +21,21 @@
 
 
 def rzx_zz2(theta: ParameterValueType | None = None):
-    """Template for CX - RZGate - CX."""
+    """RZX-based template for CX - RZGate - CX.
```
jakelishman:
```suggestion
    """RZX-based template for CX - PhaseGate - CX.
```

The module-level docstring of this one (correctly) says `CX - PhaseGate - CX`, but the function-level one had an existing copy-paste error.

jakelishman (commented):
Thanks for doing this.  I really don't know why the template circuits are organised like that - if we ever manage to finish overhauling the template-optimisation engine, we definitely should tidy that up.

jakelishman (approved):
Whoops, forgot to actually leave the tick.

PR #13265, opened 2024-10-03
Update version of Coveralls GitHub Action

jakelishman:
### Summary

We had this set to `master` in an attempt to use the rolling release, except Coveralls added a _new_ branch called `main` that is now the primary branch, so we had fallen out of date.  One of these effects is that the action was still attempting to use the deprecating Node 16 runners.

Since we now also have dependabot updating our Actions for us, this commit sets a fixed release of the Action for better reproducibility.


kevinhartman (approved):
LGTM!

PR #11526, opened 2024-01-09
Reconfigure CI pipelines for PR and merge queue

jakelishman:
### Summary

This commit refactors the pipeline organisation of our CI to reduce the total number of jobs run on any given PR merge, and to reduce the critical path length of merging.  In particular:

- the lint, docs and QPY tests are combined into one job for a single CI worker.  These three jobs individually add up to less than a test run, and moving QPY from the first-stage PR test run to this rebalances the (now) two jobs in the stage to be more equal in runtime.

- a new pipeline is added specifically for the merge queue.  Previously this reused the same two-stage PR pipeline.  The two-stage system unnecessarily lengthened the critical path, as when a PR enters the merge queue, it as already passed PR CI, and so is highly likely to pass all jobs.  In addition to flattening to a single stage, one each of the macOS and Windows jobs are removed to lower the amount of VMs needed and reduce the chances of a timeout (these OSes are more likely to get a dodgy VM and time out than Linux).  The only way a new test failure should appear (other than a flaky test) is by a logical merge conflict, which would be quite unlikely to only affect a particular Python version.

To make it easier to run the lint and docs jobs together, and ensure that both run even if there's a failure in one, the full lint configuration is merged back to `tox.ini`, and that's used to do the linting.  This makes it more consistent for developers, as well.

### Details and comments

This should automatically already work with the branch-protection rules, but there's like a 99% chance I've messed something up in the Azure configuration on the first attempt, so let's see if CI passes.

PR #13273, opened 2024-10-03
Move QPY tests to GitHub Actions and increase inter-symengine tests

jakelishman:
### Summary

This commit has two major goals:

- fix the caching of the QPY files for both the `main` and `stable/*` branches

- increase the number of compatibility tests between the different symengine versions that might be involved in the generation and loading of the QPY files.

Achieving both of these goals also means that it is sensible to move the job to GitHub Actions at the same time, since it will put more pressure on the Azure machine concurrency we use.

Caching
-------

The previous QPY tests attempted to cache the generated files for each historical version of Qiskit, but this was unreliable.  The cache never seemed to hit on backport branches, which was a huge slowdown in the critical path to getting releases out.  The cache restore keys were also a bit lax, meaning that we might accidentally have invalidated files in the cache by changing what we wanted to test, but the restore keys wouldn't have changed.

The cache files would fail to restore as a side-effect of ed79d42 (gh-11526); QPY was moved to be on the tail end of the lint run, rather than in a test run.  This meant that it was no longer run as part of the push event when updating `main` or one of the `stable/*` branches.  In Azure (and GitHub Actions), the "cache" action accesses a _scoped_ cache, not a universal one for the repository [^1][^2].  Approximately, base branches each have their own scope, and PR events open a new scope that is a child of the target branch, the default branch, and the source branch, if appropriate.  A cache task can read from any of its parent scopes, but write events go to the most local scope.  This means that we haven't been writing to long-standing caches for some time now.  PRs would typically miss the cache on the first attempt, hit their cache for updates, then miss again once entering the merge queue.

The fix for this is to run the QPY job on branch-update events as well. The post-job cache action will then write out to a reachable cache for all following events.

Cross-symengine tests
---------------------

We previously were just running a single test with differing versions of symengine between the loading and generation of the QPY files.  This refactors the QPY `run_tests.sh` script to run a full pairwise matrix of compatibility tests, to increase the coverage.

[^1]: https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/caching-dependencies-to-speed-up-workflows#restrictions-for-accessing-a-cache
[^2]: https://learn.microsoft.com/en-us/azure/devops/pipelines/release/caching?view=azure-devops#cache-isolation-and-security

### Details and comments

This is a CI change, so the chance I got it right first time is approximately zero.  Just need a PR to start testing it.  We'll need to update the branch-protection rules if we decide to merge this.

### DIFF
--- a/.github/workflows/qpy.yml
+++ b/.github/workflows/qpy.yml
@@ -1,0 +1,39 @@
+name: QPY
+
+on:
+  push:
+    branches:
+      - 'main'
+      - 'stable/*'
+  pull_request:
+  merge_group:
+concurrency:
+  group: ${{ github.repository }}-${{ github.ref }}-${{ github.head_ref }}-${{ github.workflow }}
+  cancel-in-progress: ${{ github.event_name == 'pull_request' }}
+
+jobs:
+  backward_compat:
+    if: github.repository_owner == 'Qiskit'
+    name: Backwards compatibility
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v4
+
+      - uses: actions/setup-python@v5
+        with:
+          python-version: '3.9'
+
+      - uses: dtolnay/rust-toolchain@stable
+
+      - uses: actions/cache@v4
+        with:
+          path: test/qpy_compat/qpy_cache
+          # The hashing is this key can be too eager to invalidate the cache,
+          # but since we risk the QPY tests failing to update if they're not in
+          # sync, it's better safe than sorry.
+          key: qpy-${{ hashFiles('test/qpy_compat/**') }}
+
+      - name: Run QPY backwards compatibility tests
+        working-directory: test/qpy_compat
+        run: ./run_tests.sh
--- a/.azure/lint_docs-linux.yml
+++ b/.azure/lint_docs-linux.yml
@@ -4,8 +4,8 @@ parameters:
     displayName: "Version of Python to use"
 
 jobs:
-  - job: 'Lint_Docs_QPY'
-    displayName: 'Lint, documentation and QPY'
+  - job: 'Lint_Docs'
+    displayName: 'Lint and documentation'
     pool: {vmImage: 'ubuntu-latest'}
 
     variables:
@@ -43,24 +43,3 @@ jobs:
           artifactName: 'html_docs'
           Parallel: true
           ParallelCount: 8
-
-      - task: Cache@2
-        inputs:
-          key: 'qpy | test/qpy_compat/test_qpy.py | "$(Build.BuildNumber)"'
-          restoreKeys: |
-            qpy | test/qpy_compat/test_qpy.py
-          path: qpy_files
-        displayName: Cache old QPY files
-
-      - bash: |
-          set -e
-          # Reuse the docs environment to avoid needing to rebuild another
-          # version of Qiskit.
-          source .tox/docs/bin/activate
-          mv qpy_files/* test/qpy_compat || :
-          pushd test/qpy_compat
-          ./run_tests.sh
-          popd
-          mkdir -p qpy_files
-          mv test/qpy_compat/qpy_* qpy_files/.
-        displayName: 'Run QPY backwards compat tests'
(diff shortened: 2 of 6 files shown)


jakelishman:
~I need to remember to add `on: merge_group` to the triggers, but I want to let the job complete first to see if the caching triggers correctly, at least within the PR scope.~

*edit*: done in 763491bd34.

jakelishman:
~Watching the logs go by, the inter-symengine builds seem to be spending a bunch of time rebuilding the dev Qiskit wheel each time.  I thought we'd get to cache all the Cargo artefacts and they'd be re-used, but maybe they're not, or maybe the linking stage is just taking all the time.  I'll refactor slightly to produce a single built wheel at the top, and then re-use that.~

*edit*: done in 0d8f905

jakelishman:
The cache size appears to be 200kB, which is slightly questionable - that's near exactly the size of a single set of QPY files.  That said, the cache action is compressing them, and there are a few places in the QPY files that will include some random bytes between versions because of randomly generated `Parameter` UUID instances, so maybe it does all add up right.

jakelishman:
I tested the caching for new PRs on my fork, and verified that it correctly restores the cache even for the initial "open a new PR" event, if the PR doesn't modify the QPY-compatibility test directory.

Building a single dev-version wheel at the top of the file means that the QPY backwards-compatibility job now takes only five minutes on a cache hit (e.g. https://github.com/jakelishman/qiskit-terra/actions/runs/11178693927/job/31076786343) despite now building an extra venv and running three more compatibility test, and it runs in parallel to all other jobs, whereas previously it was often on the critical path.  It takes about 15 minutes on a complete cache miss (the same as Azure, give or take), but the cache misses should now be much rarer, which should help a lot for the throughput of backports and releases.

This is ready for review.

.github/workflows/qpy.yml
```diff
@@ -0,0 +1,39 @@
...
+
+      - uses: dtolnay/rust-toolchain@stable
+
+      - uses: actions/cache@v4
+        with:
+          path: test/qpy_compat/qpy_cache
+          # The hashing is this key can be too eager to invalidate the cache,
+          # but since we risk the QPY tests failing to update if they're not in
+          # sync, it's better safe than sorry.
+          key: qpy-${{ hashFiles('test/qpy_compat/**') }}
```
mtreinish, line 35: Will this hash only on the files in the checkout from git, or will it include all the qpy files we generate during the run? I can't remember if the hashing is only performed at the start of the job or the end of the job.

mtreinish (commented):
Overall this LGTM, I like the changes to the run_tests.sh file so that we only build the wheel once and reuse it for all the venvs. In the original version I reused the dev venv to avoid building a second venv for symengine testing with the same version but that minimizes the overhead and makes everything more explicit which is nice.

I just had one inline question about the hashing key and when the hashing is evaluated.

.github/workflows/qpy.yml
jakelishman: If I got this right (I _think_ I did), the hash is calculated only once on the initial git checkout (so doesn't include the QPY files), but when pushing back to the cache, it _will_ include the generated files, so subsequent lookups will retrieve them.  It's important that the QPY files aren't part of the hash key because they're not fully deterministic - they include randomly generated UUID payloads in some of the parameters.

jakelishman:
Yeah, the new form will almost certainly duplicate one of the venvs we use (depends on the array of symengine specifiers we choose), but once we've got the Qiskit wheel pre-built, it ends up taking very little time to build, and it was easier to write and edit the loops that way.

jakelishman:
Rebased over `main` so the new Python 3.13 tests will run properly.

.github/workflows/qpy.yml
mtreinish: Ok, yeah that's why I was asking, especially because the docs aren't clear: https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/evaluate-expressions-in-workflows-and-actions#hashfiles but we can keep an eye on it post merge and adjust if it's a problem. The only option I can think of is manually listing out the files in the tree we care about tying to the cache.

jakelishman: Back before I opened this PR I checked it in a bunch of configurations on my fork, and the caching all seemed to be working the way I expected/hoped.

mtreinish (approved):
LGTm, thanks for doing this it should improve ci throughput quite a bit

PR #13281, opened 2024-10-04
Fix GitHub source code links for decorated functions

### LINKED ISSUE #2060: Bad GitHub source code link in API docs when decorator used
### Environment

- **Qiskit version**: NA
- **Python version**: NA
- **Operating system**: NA

### What is happening?

The link for the GitHub source code redirects to the wrong file. See [here](https://docs.quantum.ibm.com/api/qiskit/dev/qiskit.visualization.dag_drawer#qiskitvisualizationdag_drawer). 
The Github link should redirect to the source code got dag_drawer function, but it is taking us to this [link](https://github.com/Qiskit/qiskit/tree/main/qiskit/utils/lazy_tester.py#L74-L316) instead.

See the API Docs.

### What should happen?

The GitHub link in the API docs should take us to the source code of dag_drawer, which is [here](https://github.com/Qiskit/qiskit/blob/8ccbc8d28f34e3f2e1e0e8cbb66c4da95754b341/qiskit/visualization/dag_visualization.py#L76).

### Any suggestions?

Update the link to the source code.

Eric-Arellano:
Closes https://github.com/Qiskit/documentation/issues/2060. We were linking to the source code of the decorator, rather than the original function/method.

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -173,6 +173,10 @@ def linkcode_resolve(domain, info):
         except AttributeError:
             return None
 
+    # Unwrap decorators. This requires they used `functools.wrap()`.
+    while hasattr(obj, "__wrapped__"):
+        obj = getattr(obj, "__wrapped__")
+
     try:
         full_file_name = inspect.getsourcefile(obj)
     except TypeError:


jakelishman (approved):
Huh, I forgot that this script lived here in Qiskit - I thought it was in Qiskit/documentation.  Thanks for the fix.

Issue #13272, opened 2024-10-03
Add rust representation of the `UnitaryGate` class


mtreinish:
The `UnitaryGate` class is defined in Python and that means that when we need to work with it from Rust (either for transpiler passes or circuit construction) we treat it as a `PyInstruction` and that requires that we use Python for interacting with it. For the most part for reading this isn't a huge overhead because a `UnitaryGate` is a gate defined around a unitary matrix as a numpy array and we typically just need to inspect that matrix which can be done by reference without any copying or conversion. However, on writing (typically creation), which we do a fair amount of in the transpiler and the circuit library, the python overhead is significant. In rust it will be easy to build a `UnitaryGate` struct that implements the `Operation` trait and owns an `Array2<Complex64>`. We might need to think a bit about the python interface and split of responsibilities between the languages because we can generate a numpy array from a owned array but the traits in rust-numpy for doing that either consume the array or copy it.

This weakly depends on https://github.com/Qiskit/qiskit/issues/12966 as we'll need to figure out the best way to expand PackedOperation to support this and the pattern we establish in https://github.com/Qiskit/qiskit/issues/12966 will inform this.

PR #13282, opened 2024-10-04
Set num_qubits kwarg when building UnitaryGate from Rust

mtreinish:
### Summary

In the accelerate crate we currently have two places that are building `UnitaryGate`, the quantum_volume() function that builds a quantum volume model circuit and the Split2QUnitaries transpiler pass. Currently this can only be done by calling Python (until #13272 is implemented) and there was a potential optimization we could make to specify the number of qubits as an argument to the Python constructor. This skipss the need for a few python operations to compute the number of qubits from the size of the matrix. These operations are not exceedingly slow, but as these Python space constructors are often the bottleneck so it should help runtime performance of these two functions.

### DIFF
--- a/crates/accelerate/src/circuit_library/quantum_volume.rs
+++ b/crates/accelerate/src/circuit_library/quantum_volume.rs
@@ -10,7 +10,9 @@
 // copyright notice, and modified files need to carry a notice indicating
 // that they have been altered from the originals.
 
+use pyo3::intern;
 use pyo3::prelude::*;
+use pyo3::types::PyDict;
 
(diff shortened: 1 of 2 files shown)


ElePT (approved):
LGTM! One question out of curiosity, did you get to test how much faster the creation is when you give the number of qubits upfront?

mtreinish:
I did a quick test just now with `quantum_volume()`, and it's not really measurable outside of the error margin on the benchmarks I've run. Realistically all we're saving here is skipping:

```
math.log2(matrix.shape[0])
```

for each unitary which isn't that expensive.

PR #13283, opened 2024-10-04
Use Rust generator function for QuantumVolume class

mtreinish:
### Summary

In #13238 we added a new function quantum_volume for generating a quantum volume model circuit with the eventual goal of replacing the existing QuantumVolume class. This new function is ~10x faster to generate the circuit than the existing python class. This commit builds off of that to internally call the new function for generating the circuit from the class. While the plan is to deprecate and remove the class for Qiskit 2.0 until that time we can give a performance boost to users of it for the lifespan of the 1.x release series.

### Details and comments


qiskit/circuit/library/quantum_volume.py
```diff
@@ -83,39 +83,48 @@ def __init__(
         depth = depth or num_qubits  # how many layers of SU(4)
         width = num_qubits // 2  # how many SU(4)s fit in each layer
         rng = seed if isinstance(seed, np.random.Generator) else np.random.default_rng(seed)
+        seed_name = seed
         if seed is None:
```
kevinhartman: IMO it would be clearer to check if `seed_name` is `None` instead, since the purpose of this branch is to make sure it has a value.

qiskit/circuit/library/quantum_volume.py
```diff
@@ -83,39 +83,48 @@ def __init__(
...
-                    )
+        if classical_permutation:
+            if seed is not None:
+                seed = rng.integers(0, dtype=np.int64)
+            qv_circ = quantum_volume(num_qubits, depth, seed)
+            qv_circ.name = self.name
+            if flatten:
+                self.compose(qv_circ, inplace=True)
             else:
+                self._append(CircuitInstruction(qv_circ.to_instruction(), tuple(self.qubits)))
```
kevinhartman, line 107: Should we instead just early-return here instead of having the `else` on line 107? That way, it's a bit clearer that the handling of classical permutations is treated specially and forwarded to the Rust `quantum_volume`.

kevinhartman (commented):
Looks mostly good, other than the test failure regarding seeding.

qiskit/circuit/library/quantum_volume.py
mtreinish: I didn't return because `__init__` normally doesn't have a return; that's an implicit `None` but there is a pylint rule that doesn't like an explicit return in one path and not another.

jakelishman: I think pylint is fine with bare `return` for early return, it's `return None` it has a problem with (unless it's changed).

mtreinish:
I did a quick speed test of a 100q qv model circuit, with `QuantumVolume(100, 100, seed=42)` with this PR it takes 0.06909546643420132 sec and with 1.2.4 it took: 0.08987236272822026 sec (averaged over 50 runs). Most of the time is spent in `to_instruction()` for `_append`

qiskit/circuit/library/quantum_volume.py
mtreinish: Ah, that's probably it, I normally would do `return None` in this case. I can change it to a bare return if there is a strong preference here.

qiskit/circuit/library/quantum_volume.py
```diff
@@ -83,39 +83,49 @@ def __init__(
...
-            perm = rng.permutation(num_qubits)
-            if classical_permutation:
-                for w, unitary in enumerate(row):
-                    gate = UnitaryGate(unitary, check_input=False, num_qubits=2)
-                    qubit = 2 * w
-                    base._append(
-                        CircuitInstruction(gate, (qubits[perm[qubit]], qubits[perm[qubit + 1]]))
-                    )
+        if classical_permutation:
+            if seed is not None:
```
Cryoris, line 99: I might be slower than usual as I'm reviewing this before my first coffee, but should this be the following?
```suggestion
            if seed is None:
```

qiskit/circuit/library/quantum_volume.py
```diff
@@ -83,39 +83,49 @@ def __init__(
...
+                max_value = np.iinfo(np.int64).max
+                seed = rng.integers(max_value, dtype=np.int64)
+            qv_circ = quantum_volume(num_qubits, depth, seed)
+            qv_circ.name = self.name
+            if flatten:
+                self.compose(qv_circ, inplace=True)
```
Cryoris, line 105: If we `flatten` then I guess this here will take the most time... until we add some inplace functionality to the circuit functions

mtreinish: Yeah, compose takes the most time in the `flatten=True` path, and `to_instruction()` takes the most time in the `flatten=False`

qiskit/circuit/library/quantum_volume.py
mtreinish: This is actually correct. This code path is to normalize the seeds we pass to rust so it will fit in a uint64 (although I realized I used int64 by mistake, I'll fix that). The typing on the python class was a python int which was doesn't have an upper bound and it was possible to pass in an value greater than 18446744073709551615 which would cause an error when one didn't happen before. To solve that I used a numpy rng to select a random int from the provided seed value that we then pass to rust as a seed for it's internal rng.

When `seed` is `None` we can just pass that to rust because it will use system entropy to initialize the rust rng.

Cryoris (approved):

PR #13284, opened 2024-10-05
Add .index() and new() method to Qubit and Clbit

mtreinish:
### Summary

The Qubit and Clbit rust newtypes are a u32 that are used to track the Qubits or Clbits in a circuit. The u32 value represents the index of the Qubit or Clbit in the circuit and are often used as indices in Vecs or other structures where the index must be a usize. Accordingly there are a lot of times we need to convert a Qubit or Clbit object into a usize and previously the only way to do that was use `.0 as usize` on a given qubit object. To improve the ergonomics of that pattern this commit adds two new methods for converting from and to a usize new() and index() respectively. These are modeled after what petgraph does with it's NodeIndex type which is a similar pattern (although NodeIndex is generic over the inner type, although it defaults to u32).

### Details and comments


eliarbel:
This is nice. A comment though: with both `.index()` and `::new()` we can now make `BitType` private in `Qubit` and `Clbit`. But it's much nicer to write e.g. `Qubit(0)` than `Qubit::new(0)`. So maybe we only need `.index()`?

mtreinish:
> This is nice. A comment though: with both .index() and ::new() we can now make BitType private in Qubit and Clbit. But it's much nicer to write e.g. Qubit(0) than Qubit::new(0). So maybe we only need .index()?

The problem is that as the inner type is `BitType` we can't pass a `usize` like `Qubit(0usize)`, `Qubit(0)` works fine because the type inference knows 0 is a u32 in that context. But if you were to do something like: `Qubit(qubits_vec.len() + 1)` that would be a type mismatch I added the `new()` method so we could work with usize directly without needing to manually cast anything. That being said I didn't go through and update `Qubit` construction to use `new()` where applicable. I'll do that now.

crates/circuit/src/lib.rs
```diff
@@ -35,9 +35,40 @@ use pyo3::types::{PySequence, PyTuple};
 pub type BitType = u32;
 #[derive(Copy, Clone, Debug, Hash, Ord, PartialOrd, Eq, PartialEq, FromPyObject)]
 pub struct Qubit(pub BitType);
+
+impl Qubit {
+    /// Construct a new Qubit object from a usize, if you have a u32 you can
+    /// create a `Qubit` object directly with `Qubit(0u32)`.
+    #[inline(always)]
+    pub fn new(index: usize) -> Self {
+        Qubit(index as BitType)
```
kevinhartman: We might want to do something like this instead, since otherwise if an `index` is given that is too big for `BitType`, [Rust will wrap it](https://doc.rust-lang.org/rust-by-example/types/cast.html#casting).


```suggestion
        Qubit(index.try_into().expect(format!("Index value '{}' exceeds the maximum bit width!", index))
```

kevinhartman (commented):
It's certainly a nice improvement for us to be rid of the casting peppered everywhere!

My main suggestion would be to add an assertion to `{Qu|Cl}bit::new` that the `index` fits, which would make things more sound than what we have today.

I'm in favor of keeping the `BitType` field public, esp. with the new constructor, since users really need to know that whatever they're stuffing into a qubit / clbit must fit within a `u32`. Can you add an explicit warning to the `new` methods that the `usize` needs to fit within a `BitType`?

crates/circuit/src/lib.rs
mtreinish: I based the behavior here on that petgraph does for it's `NodeIndex` and `EdgeIndex` types which are essentially the same (except they're generic instead of having a fixed inner type) and they do the ceiling with `as`. But I agree panicking in case of an overflow makes more sense. I'll apply the suggestion and add a test checking for the panic.

mtreinish: Done in: https://github.com/Qiskit/qiskit/pull/13284/commits/1e6d4af8cdc8518e9b0a2ed3a7a725280ea90ab8

kevinhartman (approved):
Thanks for the changes! This is looking good from my perspective now.

PR #13289, opened 2024-10-07
Switch issue templates to use new-style issue types

jakelishman:
### Summary

Qiskit is part of the new-style GitHub issues beta[^1], which gives us access to a special kind of "super" label: the issue "type".  We've previously used two special labels ("bug" and "type: feature request") to categorise the resulting issues from the issue template.  This commit leans into this new system, setting the issue type, rather than the custom labels.

The new issues also specifies that the templates will be displayed in sorted order by file name.  This commit makes that ordering more explicit for us.

[^1]: https://github.blog/changelog/2024-10-01-evolving-github-issues-public-beta/

### Details and comments

If we choose to merge this, I can bulk update all our issues to use the new "types" rather than the magic labels, so we'll have a more consistent experience across our tracker.  The new "subissues" system is pushing us towards using these types, so it feels cleaner to centralise.

### DIFF
--- a/.github/ISSUE_TEMPLATE/02-feature.yaml
+++ b/.github/ISSUE_TEMPLATE/02-feature.yaml
@@ -1,6 +1,6 @@
 name: 🚀 Feature request
 description: Suggest an idea for this project 💡!
-labels: ["type: feature request"]
+type: Enhancement
 
 body:
   - type: markdown
--- a/.github/ISSUE_TEMPLATE/01-bug.yaml
+++ b/.github/ISSUE_TEMPLATE/01-bug.yaml
@@ -1,6 +1,6 @@
 name: 🐛 Bug report
 description: Create a report to help us improve 🤔.
-labels: ["bug"]
+type: Bug
 
 body:
   - type: markdown


eliarbel:
I think with the sub-issues feature it also makes sense to add a template and support the Epic type.

jakelishman:
Eli: is there going to be any advantage to that over just using the "blank issue"?  I think of the issue templates as being for external contributors to use slightly more than the core team doing internal tracking, and I can't imagine a situation where an external contributor would be / should be empowered to set us an epic. I'm also not sure there'd be any standard structure to an epic that the form would help with.

eliarbel:
My main motivation is to facilitate organizing tasks in a more hierarchical fashion now that we have the sub-issues feature available. So given a template like the below, I think that having a template Epic issue type in the `Create new issue` list which explicitly mentions sub-tasking would help that. Of course it's minor and just a convenience tool but still nice to have. 

**Background**
**Summary**
**Subtasks**

In any case and regardless, I agree with your statement about epics being more of the core team responsibility to set up.

jakelishman:
I don't feel strongly, and I feel like we could adjust the documentation of the template to make it clear that an "Epic" template is not intended for general use.  If you think that structure's useful, I can write up a template for it (maybe a separate PR, then we can play with it more separately?).

eliarbel:
Sounds good.

jakelishman:
Eli: do you want to approve / review this PR as-is, then, then I'll get to updating the issue metadata for the repo and make the new template?

.github/ISSUE_TEMPLATE/02-feature.yaml
```diff
@@ -1,6 +1,6 @@
 name: 🚀 Feature request
 description: Suggest an idea for this project 💡!
-labels: ["type: feature request"]
+type: Enhancement
```
eliarbel, line 3: We currently have in the label space `type: enhancement` and `type: feature request`, each with different meaning, namely "improve existing functionality" and "suggest or ask new feature", respectively. Assuming we want to have those 2 concepts captured in issue types (I would vote for it) It would be good to preserve this separation by naming the `type` here with `Feature` or something similar, but not `Enhancement`.

jakelishman: We have to set the allowed types at the organisation level (if I understand the docs correctly), but that seems easily doable.

PR #13291, opened 2024-10-07
Add missing docstrings for `EquivalenceLibrary`

### LINKED ISSUE #13258: Regression in EquivalenceLibrary documentation from oxidation
When generating the 1.3 dev docs, I noticed this diff:

<details><summary>diff</summary>

```diff
diff --git a/docs/api/qiskit/dev/qiskit.circuit.EquivalenceLibrary.mdx b/docs/api/qiskit/dev/qiskit.circuit.EquivalenceLibrary.mdx
index 9d46ec616e..8cbc564e19 100644
--- a/docs/api/qiskit/dev/qiskit.circuit.EquivalenceLibrary.mdx
+++ b/docs/api/qiskit/dev/qiskit.circuit.EquivalenceLibrary.mdx
@@ -8,40 +8,22 @@ python_api_name: qiskit.circuit.EquivalenceLibrary
 
 # EquivalenceLibrary
 
-<Class id="qiskit.circuit.EquivalenceLibrary" isDedicatedPage={true} github="https://github.com/Qiskit/qiskit/tree/main/qiskit/circuit/equivalence.py#L32-L267" signature="qiskit.circuit.EquivalenceLibrary(*, base=None)" modifiers="class">
-  Bases: [`object`](https://docs.python.org/3/library/functions.html#object "(in Python v3.12)")
+<Class id="qiskit.circuit.EquivalenceLibrary" isDedicatedPage={true} github="https://github.com/Qiskit/qiskit/tree/main/qiskit/circuit/equivalence.py#L29-L94" signature="qiskit.circuit.EquivalenceLibrary" modifiers="class">
+  Bases: `BaseEquivalenceLibrary`
 
   A library providing a one-way mapping of Gates to their equivalent implementations as QuantumCircuits.
 
-  Create a new equivalence library.
-
-  **Parameters**
-
-  **base** (*Optional\[*[*EquivalenceLibrary*](#qiskit.circuit.EquivalenceLibrary "qiskit.circuit.EquivalenceLibrary")*]*) – Base equivalence library to be referenced if an entry is not found in this library.
-
   ## Attributes
 
   ### graph
 
-  <Attribute id="qiskit.circuit.EquivalenceLibrary.graph">
-    Return graph representing the equivalence library data.
-
-    This property should be treated as read-only as it provides a reference to the internal state of the [`EquivalenceLibrary`](#qiskit.circuit.EquivalenceLibrary "qiskit.circuit.EquivalenceLibrary") object. If the graph returned by this property is mutated it could corrupt the the contents of the object. If you need to modify the output `PyDiGraph` be sure to make a copy prior to any modification.
-
-    **Returns**
-
-    A graph object with equivalence data in each node.
-
-    **Return type**
-
-    PyDiGraph
-  </Attribute>
+  <Attribute id="qiskit.circuit.EquivalenceLibrary.graph" />
 
   ## Methods
 
   ### add\_equivalence
 
-  <Function id="qiskit.circuit.EquivalenceLibrary.add_equivalence" github="https://github.com/Qiskit/qiskit/tree/main/qiskit/circuit/equivalence.py#L76-L113" signature="add_equivalence(gate, equivalent_circuit)">
+  <Function id="qiskit.circuit.EquivalenceLibrary.add_equivalence" signature="add_equivalence(gate, equivalent_circuit)">
     Add a new equivalence to the library. Future queries for the Gate will include the given circuit, in addition to all existing equivalences (including those from base).
 
     Parameterized Gates (those including qiskit.circuit.Parameters in their Gate.params) can be marked equivalent to parameterized circuits, provided the parameters match.
@@ -54,7 +36,7 @@ python_api_name: qiskit.circuit.EquivalenceLibrary
 
   ### draw
 
-  <Function id="qiskit.circuit.EquivalenceLibrary.draw" github="https://github.com/Qiskit/qiskit/tree/main/qiskit/circuit/equivalence.py#L199-L224" signature="draw(filename=None)">
+  <Function id="qiskit.circuit.EquivalenceLibrary.draw" github="https://github.com/Qiskit/qiskit/tree/main/qiskit/circuit/equivalence.py#L33-L58" signature="draw(filename=None)">
     Draws the equivalence relations available in the library.
 
     **Parameters**
@@ -78,7 +60,7 @@ python_api_name: qiskit.circuit.EquivalenceLibrary
 
   ### get\_entry
 
-  <Function id="qiskit.circuit.EquivalenceLibrary.get_entry" github="https://github.com/Qiskit/qiskit/tree/main/qiskit/circuit/equivalence.py#L156-L178" signature="get_entry(gate)">
+  <Function id="qiskit.circuit.EquivalenceLibrary.get_entry" signature="get_entry(gate)">
     Gets the set of QuantumCircuits circuits from the library which equivalently implement the given Gate.
 
     Parameterized circuits will have their parameters replaced with the corresponding entries from Gate.params.
@@ -102,7 +84,7 @@ python_api_name: qiskit.circuit.EquivalenceLibrary
 
   ### has\_entry
 
-  <Function id="qiskit.circuit.EquivalenceLibrary.has_entry" github="https://github.com/Qiskit/qiskit/tree/main/qiskit/circuit/equivalence.py#L115-L127" signature="has_entry(gate)">
+  <Function id="qiskit.circuit.EquivalenceLibrary.has_entry" signature="has_entry(gate)">
     Check if a library contains any decompositions for gate.
 
     **Parameters**
@@ -122,39 +104,15 @@ python_api_name: qiskit.circuit.EquivalenceLibrary
 
   ### keys
 
-  <Function id="qiskit.circuit.EquivalenceLibrary.keys" github="https://github.com/Qiskit/qiskit/tree/main/qiskit/circuit/equivalence.py#L180-L186" signature="keys()">
-    Return list of keys to key to node index map.
-
-    **Returns**
-
-    Keys to the key to node index map.
-
-    **Return type**
-
-    List
-  </Function>
+  <Function id="qiskit.circuit.EquivalenceLibrary.keys" signature="keys()" />
 
   ### node\_index
 
-  <Function id="qiskit.circuit.EquivalenceLibrary.node_index" github="https://github.com/Qiskit/qiskit/tree/main/qiskit/circuit/equivalence.py#L188-L197" signature="node_index(key)">
-    Return node index for a given key.
-
-    **Parameters**
-
-    **key** (*Key*) – Key to an equivalence.
-
-    **Returns**
-
-    Index to the node in the graph for the given key.
-
-    **Return type**
-
-    Int
-  </Function>
+  <Function id="qiskit.circuit.EquivalenceLibrary.node_index" signature="node_index(key)" />
 
   ### set\_entry
 
-  <Function id="qiskit.circuit.EquivalenceLibrary.set_entry" github="https://github.com/Qiskit/qiskit/tree/main/qiskit/circuit/equivalence.py#L129-L154" signature="set_entry(gate, entry)">
+  <Function id="qiskit.circuit.EquivalenceLibrary.set_entry" signature="set_entry(gate, entry)">
     Set the equivalence record for a Gate. Future queries for the Gate will return only the circuits provided.
 
     Parameterized Gates (those including qiskit.circuit.Parameters in their Gate.params) can be marked equivalent to parameterized circuits, provided the parameters match.
```

</details>

Losing the GitHub source code link doesn't seem very feasible to fix because we populate it based on `inspect.getsourcefile`. Imo it's not that big of a deal because the class still has the link, and it might confuse users to point them to Rust code.

The bigger concern is the type hints, parameters, and description.

raynelfss:
### Summary
The following commits attempt to fix #13258 by adding all of the missing docstrings that were lost after #12585.

### Details and comments
Add missing docstrings in the following methods:
- Python (PyO3):
    - `EquivalenceLibrary` struct class.
    - `EquivalenceLibrary::get_graph` method.
    - `EquivalenceLibrary::py_keys` method.
    - `EquivalenceLibrary::py_node_index` method.
- Rust:
    - `EquivalenceLibrary::keys` method.

### DIFF
--- a/crates/accelerate/src/equivalence.rs
+++ b/crates/accelerate/src/equivalence.rs
@@ -295,13 +295,15 @@ impl<'py> FromPyObject<'py> for GateOper {
     }
 }
 
-/// Used to extract an instance of [CircuitData] from a `QuantumCircuit`.
-/// It also ensures seamless conversion back to `QuantumCircuit` once sent
+/// Used to extract an instance of [CircuitData] from a [`QuantumCircuit`].
+/// It also ensures seamless conversion back to [`QuantumCircuit`] once sent
 /// back to Python.
 ///
 /// TODO: Remove this implementation once the [EquivalenceLibrary] is no longer
 /// called from Python, or once the API is able to seamlessly accept instances
 /// of [CircuitData].
+///
+/// [`QuantumCircuit`]: https://docs.quantum.ibm.com/api/qiskit/qiskit.circuit.QuantumCircuit
 #[derive(Debug, Clone)]
 pub struct CircuitFromPython(pub CircuitData);
 
@@ -341,7 +343,7 @@ type GraphType = StableDiGraph<NodeData, Option<EdgeData>>;
 type KTIType = IndexMap<Key, NodeIndex, RandomState>;
 
 /// A library providing a one-way mapping of Gates to their equivalent
-/// implementations as QuantumCircuits.
+/// implementations as :class:`.QuantumCircuit` instances.
 #[pyclass(
     subclass,
     name = "BaseEquivalenceLibrary",
@@ -426,7 +428,7 @@ impl EquivalenceLibrary {
     ///
     /// Args:
     ///     gate (Gate): A Gate instance.
-    ///     entry (List['QuantumCircuit']) : A list of QuantumCircuits, each
+    ///     entry (List['QuantumCircuit']) : A list of :class:`.QuantumCircuit` instances, each
     ///         equivalently implementing the given Gate.
     #[pyo3(name = "set_entry")]
     fn py_set_entry(
@@ -438,8 +440,8 @@ impl EquivalenceLibrary {
         self.set_entry(py, &gate.operation, &gate.params, entry)
     }
 
-    /// Gets the set of QuantumCircuits circuits from the library which
-    /// equivalently implement the given Gate.
+    /// Gets the set of :class:`.QuantumCircuit` instances circuits from the
+    /// library which equivalently implement the given Gate.
     ///
     /// Parameterized circuits will have their parameters replaced with the
     /// corresponding entries from Gate.params.
@@ -448,8 +450,8 @@ impl EquivalenceLibrary {
     ///     gate (Gate) - Gate: A Gate instance.
     ///
     /// Returns:
-    ///     List[QuantumCircuit]: A list of equivalent QuantumCircuits. If empty,
-    ///         library contains no known decompositions of Gate.
+    ///     List[QuantumCircuit]: A list of equivalent :class:`.QuantumCircuit` instances.
+    ///         If empty, library contains no known decompositions of Gate.
     ///
     ///         Returned circuits will be ordered according to their insertion in
     ///         the library, from earliest to latest, from top to base. The
@@ -672,14 +674,7 @@ impl EquivalenceLibrary {
         Ok(())
     }
 
-    /// Check if a library contains any decompositions for gate.
-    ///
-    /// # Arguments:
-    /// * `operation` [PackedOperation]: A gate instance.
-    ///
-    /// # Returns:
-    /// [bool]: `true` if gate has a known decomposition in the library.
-    ///         `false` otherwise.
+    /// Check if the [EquivalenceLibrary] instance contains any decompositions for gate.
     pub fn has_entry(&self, operation: &PackedOperation) -> bool {
         let key = Key::from_operation(operation);
         self.key_to_node_index.contains_key(&key)
@@ -705,12 +700,6 @@ impl EquivalenceLibrary {
     }
 
     /// Retrieve the [NodeIndex] that represents a [Key].
-    ///
-    /// ## Arguments:
-    /// * `key`: The [Key] to look for.
-    ///
-    /// ## Returns:
-    /// [NodeIndex]
     pub fn node_index(&self, key: &Key) -> NodeIndex {
         self.key_to_node_index[key]
     }
--- a/crates/accelerate/src/equivalence.rs
+++ b/crates/accelerate/src/equivalence.rs
@@ -340,6 +340,8 @@ impl ToPyObject for CircuitFromPython {
 type GraphType = StableDiGraph<NodeData, Option<EdgeData>>;
 type KTIType = IndexMap<Key, NodeIndex, RandomState>;
 
+/// A library providing a one-way mapping of Gates to their equivalent
+/// implementations as QuantumCircuits.
 #[pyclass(
     subclass,
     name = "BaseEquivalenceLibrary",
@@ -469,6 +471,16 @@ impl EquivalenceLibrary {
     }
 
     // TODO: Remove once BasisTranslator is in Rust.
+    /// Return graph representing the equivalence library data.
+    ///
+    /// This property should be treated as read-only as it provides
+    /// a reference to the internal state of the :class:`~.EquivalenceLibrary` object.
+    /// If the graph returned by this property is mutated it could corrupt the
+    /// the contents of the object. If you need to modify the output ``PyDiGraph``
+    /// be sure to make a copy prior to any modification.
+    ///
+    /// Returns:
+    ///     PyDiGraph: A graph object with equivalence data in each node.
     #[getter]
     fn get_graph(&mut self, py: Python) -> PyResult<PyObject> {
         if let Some(graph) = &self._graph {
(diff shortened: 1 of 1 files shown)


crates/accelerate/src/equivalence.rs
```diff
@@ -340,6 +340,8 @@ impl ToPyObject for CircuitFromPython {
 type GraphType = StableDiGraph<NodeData, Option<EdgeData>>;
 type KTIType = IndexMap<Key, NodeIndex, RandomState>;
 
+/// A library providing a one-way mapping of Gates to their equivalent
```
kevinhartman: I think we should either reference the classes with code refs or just call them "gates" and "circuits" :)

crates/accelerate/src/equivalence.rs
```diff
@@ -664,6 +687,7 @@ impl EquivalenceLibrary {
         self.key_to_node_index.contains_key(&key)
     }
 
+    /// Returns an iterator with all the Keys in the `EquivalenceLibrary`.
```
kevinhartman: For Rust doc strings, we should use the Rust-style for code refs, e.g. `[EquivalenceLibrary]`.

crates/accelerate/src/equivalence.rs
raynelfss: I'd say to go with "gates" and "circuits" because `PyO3` will reuse this docstring with Python, which may not be displayed properly.

kevinhartman: I think we're in agreement here, but I meant that it'd be better to not use symbol-sounding names like "QuantumCircuits" and "Gates" with a capital "G". Just call them "gates" and "circuits", verbatim.

kevinhartman: Also, since this is a Python class, you can use Sphinx code references if you want, instead. Like 

```
:class:`.QuantumCircuit`
```

They should turn into hyperlinks in the Python docs.

crates/accelerate/src/equivalence.rs
```diff
@@ -682,13 +704,13 @@ impl EquivalenceLibrary {
         }
     }
 
-    /// Retrieve the `NodeIndex` that represents a `Key`
+    /// Retrieve the [NodeIndex] that represents a [Key].
     ///
-    /// # Arguments:
-    /// * `key`: The `Key` to look for.
+    /// ## Arguments:
```
kevinhartman: Is there a reason to go with H2 here?

raynelfss: Nothing more than just me thinking H1 looks a bit too big here.

kevinhartman: Heh, understandable. Regardless, it looks like the Rust standard library uses H1 for a first level section, but it also looks like documenting arguments separately like we do in Python is actually not hip in Rust. 

I would recommend that we just remove the `Arguments` and `Returns` sections on all of these methods altogether. To describe arguments that aren't self-explanatory, add a second paragraph after the summary sentence to describe the usage.

Here's what the Rust standard library uses as their convention: https://github.com/rust-lang/rfcs/blob/master/text/1574-more-api-documentation-conventions.md#appendix-a-full-conventions-text

kevinhartman (commented):
The Python docs look good. I left one comment recommending we use lowercase generic terms in cases where we don't want to make a code reference to a type. In general, prefer `circuit` or `` :class:`.QuantumCircuit` `` over `QuantumCircuit`. This is a fairly standard practice, e.g. [Rust gives this guidance in its standard library style guide](https://github.com/rust-lang/rfcs/blob/master/text/1574-more-api-documentation-conventions.md#referring-to-types). The rationale that comes to mind is that if we know the exact type in question, we might as well use a code reference.

As for the Rust docs, I'd recommend we drop all of the `Args` and `Returns` sections, and stick with just a summary sentence, followed by an optional paragraph that describes any usage details. It's just not standard practice in Rust to list out arguments like we see in Python docs. Instead, the Rust community favors a cohesive usage description of the function as a whole when arguments aren't obvious, along with an `Examples` section, ideally for every method to provide usage examples. It's probably alright to skip the `Examples` section for most of the methods you've got in this PR, since the usage is self-explanatory and this is not yet a public interface.

crates/accelerate/src/equivalence.rs
mtreinish: I agree with Kevin here it should be `gates` or we do a class ref if it's a capital `G`. Although I realize the original text had `Gates` and `QuantumCircuits` without a link. We can do better here.

```suggestion
/// A library providing a one-way mapping of gates to their equivalent
```

crates/accelerate/src/equivalence.rs
```diff
@@ -436,8 +440,8 @@ impl EquivalenceLibrary {
         self.set_entry(py, &gate.operation, &gate.params, entry)
     }
 
-    /// Gets the set of QuantumCircuits circuits from the library which
-    /// equivalently implement the given Gate.
+    /// Gets the set of :class:`.QuantumCircuit` instances circuits from the
+    /// library which equivalently implement the given Gate.
```
mtreinish:
```suggestion
    /// library which equivalently implement the given :class:`.Gate`.
```

mtreinish (approved):
LGTM, thanks for fixing this.

PR #13292, opened 2024-10-07
Fix GitHub source code links on Windows docs builds

Eric-Arellano:
Per https://docs.python.org/3/library/pathlib.html#pathlib.PurePath.as_posix, calling `str()` on a Windows `Path` object uses `\` as the separator. We need to use `/` for the GitHub URL to work properly.

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -185,7 +185,7 @@ def linkcode_resolve(domain, info):
         return None
     try:
         relative_file_name = Path(full_file_name).resolve().relative_to(REPO_ROOT)
-        file_name = re.sub(r"\.tox\/.+\/site-packages\/", "", str(relative_file_name))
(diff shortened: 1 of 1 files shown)


jakelishman (approved):
It's interesting if this is the only problem with docs builds on Windows, since we don't test that at all - only Linux is tested, and pragmatically you can generally expect macOS to work because I (and a couple of other team members) build the docs from macOS.

Eric-Arellano:
> It's interesting if this is the only problem with docs builds on Windows

Indeed, I was afraid when helping Becky today to get set up that it was going to be a can of worms to get local docs working. This was the only problem we encountered! FYI Runtime docs had an issue with not setting file encoding in pathlib ops, but we don't have that script in this repo.

PR #13295, opened 2024-10-08
Prepare `PauliEvolutionGate` for Rustiq & port it to Rust

### LINKED ISSUE #12789: Add `PauliEvolutionGate` synthesis plugin using rustiq
### What should we add?

The recently released https://crates.io/crates/rustiq-core rust library contains some alternative methods for synthesis:

https://docs.rs/rustiq-core/0.0.2/rustiq_core/synthesis/index.html

which could be used to construct synthesis functions to synthesize different high level objects. Specifically we should investigate the pauli network synthesis for the `PauliEvolutionGate`.

Cryoris:
### Summary

Port the `PauliEvolutionGate` synthesis to Rust, plus, expose the Pauli network and allow other plugins to synthesize the gate. Also adds the plugin structure for the gate for #12789.

### Details and comments

The larger the Pauli network to synthesis, the better the speedup from the port to Rust. Here I measured a Heisenberg Hamiltonian (XX+YY+ZZ on interacting qubits, plus 1-qubit Z on each qubit) on a square lattice for different settings:

<details closed>
<summary>1 timestep for first order Trotter: speedup @ 100 qubits is 2.8</summary>

```
N=25: 7.0
main: 0.007 +- 0.006
this: 0.001 +- 0.000

N=100: 2.8
main: 0.042 +- 0.000
this: 0.015 +- 0.000

N=255: 2.8
main: 0.198 +- 0.040
this: 0.071 +- 0.003

N=400: 2.3
main: 0.607 +- 0.001
this: 0.260 +- 0.000
```
</details>

<details closed>
<summary>10 timestep for 4th order Trotter: speedup @ 100 qubits is 9.4</summary>

```
N=25: 6.7
main: 0.040 +- 0.003
this: 0.006 +- 0.000

N=100: 9.4
main: 0.330 +- 0.037
this: 0.035 +- 0.001

N=255: 10.8
main: 1.291 +- 0.066
this: 0.119 +- 0.007

N=400: 11.5
main: 4.052 +- 0.051
this: 0.353 +- 0.009
```
</details>

<details closed>
<summary>1 timestep for first order Trotter but with wrap=True: speedup @ 100 qubits is 3.6</summary>

```
N=25: 3.4
main: 0.034 +- 0.012
this: 0.010 +- 0.004

N=100: 3.6
main: 0.187 +- 0.017
this: 0.052 +- 0.010

N=255: 3.1
main: 0.551 +- 0.048
this: 0.178 +- 0.019

N=400: 3.0
main: 1.367 +- 0.068
this: 0.458 +- 0.027
```
</details>

Side effects of this PR:
* add `SparsePauliOp.to_sparse_list` to construct the sparse list format (i.e. `op == SparsePauliOp.from_sparse_list(op.to_sparse_list())`)
* refactor `LieTrotter` to just be an alias of `SuzukiTrotter` and reduce code duplication
* fix the docstrings to use actually non-commuting Paulis (XX and ZZ commute and would have no expansion error)


qiskit/synthesis/evolution/product_formula.py
```diff
@@ -6,7 +6,7 @@
 # obtain a copy of this license in the LICENSE.txt file in the root directory
 # of this source tree or at http://www.apache.org/licenses/LICENSE-2.0.
 #
-# Any modifications or derivative works of this code must retain this
+# typing.Any modifications or derivative works of this code must retain this
```
alexanderivrii: a random "typing" error :)

alexanderivrii: Interesting, this actually seems to cause the lint error:
> lint: commands[5]> python /home/vsts/work/1/s/tools/verify_headers.py qiskit test tools crates
qiskit/synthesis/evolution/product_formula.py failed header check because:
Header apache text string doesn't match:

Cryoris: Whoops, I was using find+replace without sufficient caution 😉 Fixed!

crates/accelerate/src/circuit_library/pauli_evolution.rs
```diff
@@ -0,0 +1,303 @@
...
+///     pauli: The Pauli string, e.g. "IXYZ".
+///     indices: The qubit indices the Pauli acts on, e.g. if given as [0, 1, 2, 3] with the
+///         Pauli "IXYZ", then the correspondence is I_0 X_1 Y_2 Z_3.
+///     time: The rotation angle. Note that this will directly be used as input of the
+///         rotation gate and not be multiplied by a factor of 2 (that should be done before so
+///         that this function can remain Rust-only).
+///     phase_gate: If ``true``, use the ``PhaseGate`` instead of ``RZGate`` as single-qubit rotation.
+///     cx_fountain: If ``true``, implement the CX propagation as "fountain" shape, where each
+///         CX uses the top qubit as target. If ``false``, uses a "chain" shape, where CX in between
+///         neighboring qubits are used.
```
alexanderivrii: This is called "cx_fountain" in the function docstring and "do_fountain" in the function arguments. :)

alexanderivrii: Out of curiosity, why do we even need both options? Or, looking at it from the other end, why do we only consider these two options out of all possible ways to connect the qubits using some kind of a tree? And here is a fun thought: metaphorically a fountains represents fluidity, while a chain represents rigidity, does this distinction apply to the two circuits at hand?

crates/accelerate/src/circuit_library/pauli_evolution.rs
```diff
@@ -0,0 +1,303 @@
...
+    let mut times: Vec<Param> = Vec::with_capacity(num_paulis);
+    let mut global_phase = Param::Float(0.0);
+
+    for el in sparse_paulis.iter() {
+        let tuple = el.downcast::<PyTuple>()?;
+        let pauli = tuple.get_item(0)?.downcast::<PyString>()?.to_string();
+        let time = Param::extract_no_coerce(&tuple.get_item(2)?)?;
+
+        if pauli.as_str().chars().all(|p| p == 'i') {
+            global_phase = radd_param(global_phase, time, py); // apply factor -1 at the end
```
alexanderivrii: I am not sure I understand the comment "apply factor -1 at the end".

crates/accelerate/src/circuit_library/pauli_evolution.rs
```diff
@@ -0,0 +1,303 @@
...
+use crate::circuit_library::utils;
+
+// custom types for a more readable code
+type StandardInstruction = (StandardGate, SmallVec<[Param; 3]>, SmallVec<[Qubit; 2]>);
+type Instruction = (
+    PackedOperation,
+    SmallVec<[Param; 3]>,
+    Vec<Qubit>,
+    Vec<Clbit>,
+);
```
alexanderivrii, line 28: These are packed operations rather than standard gates because of the barriers, right?

alexanderivrii: I just noticed that the PR defines a similar `Instruction` type in `utils.rs` as well. Should we use that type instead?

crates/accelerate/src/circuit_library/pauli_evolution.rs
```diff
@@ -0,0 +1,303 @@
...
+) -> impl Iterator<Item = StandardInstruction> {
+    let active_paulis: Vec<(char, Qubit)> = pauli
+        .into_iter()
+        .zip(indices.into_iter().map(Qubit))
+        .collect();
+
+    // get the basis change: x -> HGate, y -> SXdgGate, z -> nothing
+    let basis_change = active_paulis
+        .clone()
+        .into_iter()
```
alexanderivrii: Would it be possible to call `iter()` instead of `clone().into_iter()`? (I did not try to do this change myself, so I might be missing something obvious).

crates/accelerate/src/circuit_library/pauli_evolution.rs
```diff
@@ -0,0 +1,303 @@
...
+                        Vec::from_iter(qubits.into_iter()),
+                        Vec::new(),
+                    ))
+                },
+            );
+            as_packed.chain(utils::maybe_barrier(
+                py,
+                num_qubits as u32,
+                insert_barriers && i < (num_paulis - 1), // do not add barrier on final block
+            ))
```
alexanderivrii: The `maybe_barrier` function adds the barrier exactly when some condition holds, correct? Not too important, but you could also use `condition.then_some(...)` to construct conditional iterators

qiskit/synthesis/evolution/product_formula.py
```diff
@@ -16,16 +16,19 @@
...
-from typing import Any
-from functools import partial
+import typing
 import numpy as np
 from qiskit.circuit.parameterexpression import ParameterExpression
-from qiskit.circuit.quantumcircuit import QuantumCircuit
+from qiskit.circuit.quantumcircuit import QuantumCircuit, ParameterValueType
 from qiskit.quantum_info import SparsePauliOp, Pauli
 from qiskit.utils.deprecation import deprecate_arg
+from qiskit._accelerate.circuit_library import py_pauli_evolution
```
alexanderivrii: Just a nitpick: you call this a `py_pauli_evolution` but it's a really the fast Rust-based evolution. To emphasize this, maybe you can import it `as pauli_evolution_rs`? In addition, we probably want to call the public Rust function `pauli_evolution`, without the trailing "py",

releasenotes/notes/pauli-evo-plugins-612850146c3f7d49.yaml
```diff
@@ -0,0 +1,32 @@
+---
+features_quantum_info:
+  - |
+    Added :meth:`.SparsePauliOperator.to_sparse_list` to convert an operator into
+    a sparse list format. This works analogous to :meth:`.SparsePauliOperator.from_sparse_list`.
```
alexanderivrii: Shouldn't this be "oppositely" rather than "analogously"?

alexanderivrii (commented):
Thanks @Cryoris ! I have left a couple of tiny comments, but otherwise it looks very good. I will now try to rebase my PR on top of this and see if I have any more comments.

crates/accelerate/src/circuit_library/pauli_evolution.rs
Cryoris: * Oh whoops, I changed back and forth between the namings... 
* The idea to have both option came from considering that CX cancellations or HW mappings can differ depending on which you choose. Tbh I don't know any real case where someone would use "fountain". Seems more like a choice the compiler should make, but to keep it consistent I left it there...
* I'd argue the chain represent the interconnect of small links, whereas the fountain has individual streams 😛

crates/accelerate/src/circuit_library/pauli_evolution.rs
Cryoris: I'll elaborate in the code 👍🏻 

The coefficient of identity Paulis gets added as `-coeff` to the global phase, as we evolve with `exp(-it I)` and not with `exp(it I)`. But doing the multplication with `-1` is not for free, so we first sum all coeffs up and then do a single multiplication.

crates/accelerate/src/circuit_library/pauli_evolution.rs
Cryoris: yes and yes 🙂

crates/accelerate/src/circuit_library/pauli_evolution.rs
Cryoris: I think I tried this and it's not possible since we use `active_paulis` later on as well and cannot consume it here. I'll try again though 🙂

crates/accelerate/src/circuit_library/pauli_evolution.rs
Cryoris: You mean like
```rust
iterator1
.chain(insert_barriers.then_some(std::iter::once(barrier))
.chain(iterator2)
```
?

That would be nicer 🙂

releasenotes/notes/pauli-evo-plugins-612850146c3f7d49.yaml
Cryoris: yeah maybe 😛 I'll update

crates/accelerate/src/circuit_library/pauli_evolution.rs
```diff
@@ -0,0 +1,303 @@
...
+///
+/// Returns:
+///     A pointer to an iterator over standard instructions.
+pub fn pauli_evolution<'a>(
+    pauli: &'a str,
+    indices: Vec<u32>,
+    time: Param,
+    phase_gate: bool,
+    do_fountain: bool,
+) -> Box<dyn Iterator<Item = StandardInstruction> + 'a> {
```
raynelfss, line 53: Why are we using a `Box` here? From what I understand, using a `Box` pointer can save memory by storing things in the heap rather than the stack. But is it that much more expensive than just returning a regular iterator? Standard gates aren't supposed to be too heavy either.

This isn't critical as performance wouldn't be impacted from what I can tell since `StandardGate` instances are very light. When it comes to `Param` instances, it might be a bit trickier. But it is something to think about.

crates/accelerate/src/circuit_library/pauli_evolution.rs
```diff
@@ -0,0 +1,303 @@
...
+    let param: SmallVec<[Param; 3]> = smallvec![time.clone()];
+    let paulistring: String = pauli.iter().collect();
+
+    match paulistring.as_str() {
+        "xx" => Box::new(std::iter::once((StandardGate::RXXGate, param, qubits))),
+        "zx" => Box::new(std::iter::once((StandardGate::RZXGate, param, qubits))),
+        "yy" => Box::new(std::iter::once((StandardGate::RYYGate, param, qubits))),
+        "zz" => Box::new(std::iter::once((StandardGate::RZZGate, param, qubits))),
+        // note that the CX modes (do_fountain=true/false) give the same circuit for a 2-qubit
+        // Pauli, so we just set it to false here
```
raynelfss:
```suggestion
        // Note: The CX modes (do_fountain=true/false) give the same circuit for a 2-qubit
        // Pauli, so we just set it to false here
```

crates/accelerate/src/circuit_library/pauli_evolution.rs
raynelfss: For this I think you might need to rewrite the logic quite a bit to avoid some cloning, since you're cloning the iterator's source each time. Maybe start by using `iter().copied()` here since both structures implement the `Copy` trait.
You might also be better off collecting some of them instead of cloning the source. At least for this one I think you could get away with collecting, and maybe the `inverse_basis_change` as well since you will re-append. Then you can call into-iter() at the end to consume them.

crates/accelerate/src/circuit_library/pauli_evolution.rs
```diff
@@ -0,0 +1,303 @@
...
+///     Circuit data for to implement the evolution.
+#[pyfunction]
+#[pyo3(signature = (num_qubits, sparse_paulis, insert_barriers=false, do_fountain=false))]
+pub fn py_pauli_evolution(
+    py: Python,
+    num_qubits: i64,
+    sparse_paulis: &Bound<PyList>,
+    insert_barriers: bool,
+    do_fountain: bool,
+) -> PyResult<CircuitData> {
```
raynelfss: **Nitpick:** You could easily remove the py token here since the function takes a `Bound<_>` which themselves contain a py token.

```suggestion
pub fn py_pauli_evolution(
    num_qubits: i64,
    sparse_paulis: &Bound<PyList>,
    insert_barriers: bool,
    do_fountain: bool,
) -> PyResult<CircuitData> {
    let py: Python = sparse_paulis.py();
```

crates/accelerate/src/circuit_library/pauli_evolution.rs
```diff
@@ -0,0 +1,303 @@
...
+
+        paulis.push(pauli);
+        times.push(time); // note we do not multiply by 2 here, this is done Python side!
+        indices.push(
+            tuple
+                .get_item(1)?
+                .downcast::<PyList>()?
+                .iter()
+                .map(|index| index.extract::<u32>())
+                .collect::<PyResult<_>>()?,
```
raynelfss: Should we directly extract to a `Vec<u32>` here?
```suggestion
            tuple
                .get_item(1)?
                .extract::<Vec<u32>>()?,
```

crates/accelerate/src/circuit_library/pauli_evolution.rs
```diff
@@ -0,0 +1,303 @@
...
+) -> Box<dyn Iterator<Item = StandardInstruction> + 'a> {
+    let qubits: SmallVec<[Qubit; 2]> = smallvec![Qubit(indices[0]), Qubit(indices[1])];
+    let param: SmallVec<[Param; 3]> = smallvec![time.clone()];
+    let paulistring: String = pauli.iter().collect();
+
+    match paulistring.as_str() {
+        "xx" => Box::new(std::iter::once((StandardGate::RXXGate, param, qubits))),
+        "zx" => Box::new(std::iter::once((StandardGate::RZXGate, param, qubits))),
+        "yy" => Box::new(std::iter::once((StandardGate::RYYGate, param, qubits))),
+        "zz" => Box::new(std::iter::once((StandardGate::RZZGate, param, qubits))),
```
raynelfss, line 115: I'm not sure about the purpose of these `once` iterators. I'm assuming they're single step iterators which is a bit strange here because in most cases you are returning iterators of just one item. What is the intended effect here?

crates/accelerate/src/circuit_library/utils.rs
```diff
@@ -0,0 +1,58 @@
...
+            instruction: barrier.into(),
+        };
+        Box::new(std::iter::once(Ok((
+            barrier_inst.into(),
+            smallvec![],
+            (0..num_qubits).map(Qubit).collect(),
+            vec![] as Vec<Clbit>,
+        ))))
+    }
+}
```
raynelfss: This method is a little confusing:
- From what I can gather it returns a Barrier as an `Instruction` but then it returns a `Box<Iterator<_>>` instance, even though it is one item. Is this a memory saving measure?
- If the purpose of this operation is to just return a `Barrier` if a condition is met, you should probably have this labeled as `#[inline]` since it's only a conditional check. This prevents the compiler from seeing an extra call to this function.
- Going back to @alexanderivrii's comment, could you maybe use `insert_barrier.then_some()`.

raynelfss (commented):
I quickly reviewed the rust code. Great job. There are some questions I have that I'm fully open to discussing.

qiskit/synthesis/evolution/product_formula.py
Cryoris: A lot (most?) of the PyO3 exposed functions start with a leading `py_` and I think it's a nice convention to have in the Rust code: functions that take Python objects as input have a `py_` prefix and Rust-only functions have normal names. 

How about doing
```rust
#[pyfunction]
#[pyo3(name = "pauli_evolution", signature = (...))]
pub fn py_pauli_evolution(...)
```
which (as far as I understand it) would expose it to Python as `pauli_evolution`?

crates/accelerate/src/circuit_library/pauli_evolution.rs
Cryoris: This is because this function returns different types of `Iterator` depending on the input (e.g. `Chain<Map<...>>` vs `Empty`). The dynamic type was the only way I got to work, but I'm happy to change it if there's a better way 🙂

crates/accelerate/src/circuit_library/pauli_evolution.rs
Cryoris: Ah I just always add that token by default, but I have zero personal preference -- is there a reason to choose one over the other? 🙂

crates/accelerate/src/circuit_library/pauli_evolution.rs
Cryoris: Oh yes -- I just recently learned that one can directly extract `Vec`s -- thanks!

crates/accelerate/src/circuit_library/pauli_evolution.rs
Cryoris: This might sound stupid, but I just wanted to return an iterator of one item -- is there another way to write this? 😅

raynelfss: You could but at that point you could just return the item

crates/accelerate/src/circuit_library/pauli_evolution.rs
raynelfss: I'm not completely sure. But it's a bit redundant to have two different `py` tokens when we already get one from the `Bound` object.

crates/accelerate/src/circuit_library/pauli_evolution.rs
Cryoris: But I have another match-arm that returns an iterator (the `_` case uses `multi_qubit_evolution`), so I can't just return the item, right?

crates/accelerate/src/circuit_library/utils.rs
Cryoris: I changed to `then_some` in cf7d8b096 but I think this has one disadvantage: when we write something like
```rust
instructions.chain(
   insert_barrier.then_some(Ok(barrier.clone())
)
```
Then even if `insert_barrier` is `false`, I need to make the Python call to construct the barrier object. To avoid that, what I would like to do is 
```rust
let barrier = match insert_barrier {
    true => Some(.... get barrier instruction...),
    false => None,
}

// in my iteration loop ...
    instructions.chain(
        insert_barrier.then_some(Ok(barrier.clone().unwrap()))
    )
```
but rust won't allow this since `barrier` can be `None` and I'm not allowed to unwrap 🤔 

So it seems we have a performance vs. code legibility issue here, which is why I had the `maybe_barrier` function. How about making this into a struct with some documentation? Like
```rust
struct MaybeBarrier {
    barrier: Option<Instruction>
}

impl MaybeBarrier {
    fn from_py(...) // construct the barrier object
    fn get(condition)  // return None or the barrier based on the condition
}
```

crates/accelerate/src/circuit_library/pauli_evolution.rs
raynelfss: Oh yes, that case is definitely guaranteed to return an iterator of more than one item. The once iterators can be kept then. 🚀

qiskit/synthesis/evolution/product_formula.py
Cryoris: I changed it to the above in cf7d8b096

releasenotes/notes/pauli-evo-plugins-612850146c3f7d49.yaml
Cryoris: done in cf7d8b096

crates/accelerate/src/circuit_library/pauli_evolution.rs
Cryoris: I changed to `then_some` in cf7d8b096, see also the discussion below in https://github.com/Qiskit/qiskit/pull/13295#discussion_r1818744542

crates/accelerate/src/circuit_library/pauli_evolution.rs
Cryoris: I changed this in https://github.com/Qiskit/qiskit/commit/cf7d8b09681c400c2213c6c35722f85a7f7c63b4, let me know if this is what you had in mind 🙂

crates/accelerate/src/circuit_library/pauli_evolution.rs
Cryoris: added in https://github.com/Qiskit/qiskit/commit/cf7d8b09681c400c2213c6c35722f85a7f7c63b4

crates/accelerate/src/circuit_library/pauli_evolution.rs
Cryoris: Done in 6f4fcbad4

test/python/circuit/library/test_pauli_feature_map.py
```diff
@@ -70,24 +70,26 @@ def test_pauli_evolution(self):
...
             evo.cx(1, 2)
             evo.p(2 * time, 2)
             evo.cx(1, 2)
             evo.cx(0, 1)
-            evo.rx(-np.pi / 2, 1)
+            evo.sx(1)
             evo.h(2)
 
+            print(evo.draw())
+
```
alexanderivrii: can be removed :)

qiskit/synthesis/evolution/product_formula.py
```diff
@@ -108,8 +114,50 @@ def wrap_atomic_evolution(output, operator, time):
...
+
+        Returns:
+            A list of Pauli rotations in a sparse format, where each element is
+            ``(paulistring, qubits, coefficient)``. For example, the Lie-Trotter expansion
+            of ``H = XI + ZZ`` would return ``[("X", [1], 1), ("ZZ", [0, 1], 1)]``.
+        """
+        raise NotImplementedError(
+            f"The method ``expand`` is not implemented for {self.__class__}. Implement it to "
+            f"automatically enable the call to {self.__class__}.synthesize."
+        )
```
alexanderivrii, line 133: Can we move this abstract method to its parent class `EvolutionSynthesis`?

qiskit/synthesis/evolution/suzuki_trotter.py
```diff
@@ -105,51 +108,91 @@ def __init__(
             ValueError: If order is not even
         """
 
-        if order % 2 == 1:
+        if order > 1 and order % 2 == 1:
             raise ValueError(
                 "Suzuki product formulae are symmetric and therefore only defined "
-                "for even orders."
+                "for even orders (or order==1)."
```
alexanderivrii: Can we slightly improve this error message, maybe to "Suzuki product formulas are only defined when the order is 1 or even"?

qiskit/synthesis/evolution/product_formula.py
Cryoris: Only `ProductFormula`s can expand, general `EvolutionSynthesis` can do arbitrary things to implement the evolution (such as computing the matrix evo) and don't have `expand` 🙂

test/python/circuit/library/test_pauli_feature_map.py
Cryoris: whoops, thanks!

qiskit/synthesis/evolution/suzuki_trotter.py
Cryoris: Updated in e802c45

test/python/circuit/library/test_pauli_feature_map.py
Cryoris: Removed in e802c45

crates/accelerate/src/circuit_library/pauli_evolution.rs
raynelfss: From a recent talk I had with @alexanderivrii it seems that you want to call the `.rev` method to be able to reverse the iterator. If so, perhaps you could change the return types to be `DoubleEndedIterator` instances instead of a `Box<dyn Iterator<_>>`.

So you could change some of the return types from:
```rust
pub fn foo() -> Box<dyn Iterator<Item = StandardInstruction> + 'a> {
```
to:
```rust
pub fn foo() -> impl DoubleEndedIterator<Item = StandardInstruction> + 'a {
```
This would allow you to use any iterator type as long as it can be reversed, which seems to be the case for many of the iterators used here.

alexanderivrii:
Hmm, after switching `SX` and `SXdg` in [2b6f4eb](https://github.com/Qiskit/qiskit/pull/13295/commits/2b6f4ebc38a0eaa7bc25060985d84d8a06bedfeb), some of the tests in `pauli_feature_map` now fail. I believe the tests were incorrect before [2b6f4eb](https://github.com/Qiskit/qiskit/pull/13295/commits/2b6f4ebc38a0eaa7bc25060985d84d8a06bedfeb), so it's good that they would be fixed, yet it's scary how difficult it is to distinguish between the correct and incorrect behaviors (I was actually looking at these before and they looked perfectly fine to me). Can we add more `Operator`-based asserts? I.e. in addition to checking that the `ref` circuit looks as it's supposed to, also that `Operator(encoding) == Operator(ref)`? 

Other than that, I am very happy with this PR, thanks for the great work!

raynelfss (commented):
I had a bit of a discussion with @alexanderivrii about the usage of `Box` here, and we believe to have found a better alternative to what's being done here. I haven't tested it myself but it might be worth taking a look.

alexanderivrii:
For the record, I am no longer sure what was our conclusion with @raynelfss (Ray is so ahead of me when discussing Rust types). I know that @Cryoris used boxing to be able to return both the usual and the double-ended iterators (and maybe something else too), yet if I remember correctly the double-ended iterators are a bit heavier than the usual ones. Does it make sense for us to return the double-ended iterators everywhere? Another possibility is maybe we should return the `CircuitData` objects (instead of `Vec<Instruction>` or iterator over that).

raynelfss:
In hindsight @Cryoris, we might not be able to use a simple `impl Trait` I believe I can now see why you used so many `Box` pointers.

> This is because this function returns different types of Iterator depending on the input (e.g. Chain<Map<...>> vs Empty). The dynamic type was the only way I got to work, but I'm happy to change it if there's a better way 🙂

Even though both `Once` iterators and an example iterator such as `Chain<Map<Iterator>>` should technically both count as `Iterator` the types are different enough to require a dynamic block, and so you'd need to use the `Box` there. I can't think of another way to do this right now, so you can keep the `Box` here.

crates/accelerate/src/circuit_library/pauli_evolution.rs
```diff
@@ -0,0 +1,330 @@
...
+use pyo3::types::{PyList, PyString, PyTuple};
+use qiskit_circuit::circuit_data::CircuitData;
+use qiskit_circuit::operations::{multiply_param, radd_param, Param, PyInstruction, StandardGate};
+use qiskit_circuit::packed_instruction::PackedOperation;
+use qiskit_circuit::{imports, Clbit, Qubit};
+use smallvec::{smallvec, SmallVec};
+
+// custom types for a more readable code
+type StandardInstruction = (StandardGate, SmallVec<[Param; 3]>, SmallVec<[Qubit; 2]>);
+type Instruction = (
```
ShellyGarion, line 23: Perhaps the name "Instruction" can be confusing with Qiskit Instruction? Maybe call it something like "EvolutionIstruction" ?

crates/accelerate/src/circuit_library/pauli_evolution.rs
```diff
@@ -0,0 +1,330 @@
...
+    // exp(-i t I). We apply the negative sign here, to only do a single (-1) multiplication,
+    // instead of doing it every time we find an all-identity Pauli.
+    if modified_phase {
+        global_phase = multiply_param(&global_phase, -1.0, py);
+    }
+
+    CircuitData::from_packed_operations(py, num_qubits as u32, 0, evos, global_phase)
+}
+
+fn cx_chain(
```
ShellyGarion, line 298: could you please add some image to explain how a CX chain should look like?

crates/accelerate/src/circuit_library/pauli_evolution.rs
```diff
@@ -0,0 +1,330 @@
...
+) -> Box<dyn DoubleEndedIterator<Item = StandardInstruction>> {
+    let num_terms = active_paulis.len();
+    Box::new(
+        (0..num_terms - 1)
+            .map(move |i| (active_paulis[i].1, active_paulis[i + 1].1))
+            .map(|(target, ctrl)| (StandardGate::CXGate, smallvec![], smallvec![ctrl, target])),
+    )
+}
+
+fn cx_fountain(
```
ShellyGarion, line 322: could you please add some image to explain how a CX fountain should look like?

qiskit/circuit/library/pauli_evolution.py
```diff
@@ -85,10 +86,10 @@ class PauliEvolutionGate(Gate):
 
     def __init__(
         self,
-        operator,
-        time: Union[int, float, ParameterExpression] = 1.0,
```
ShellyGarion, line 89: why is ParamterExpression replaced by ParameterValueType ? (isn't this considered as an API change) ?

qiskit/synthesis/evolution/lie_trotter.py
```diff
@@ -100,26 +82,6 @@ def __init__(
         """
         super().__init__(1, reps, insert_barriers, cx_structure, atomic_evolution, wrap)
 
-    def synthesize(self, evolution):
```
ShellyGarion, line 103: it's not deprecated - right? will be called as SuzukiTrotter.synthesize ?

qiskit/synthesis/evolution/suzuki_trotter.py
```diff
@@ -105,51 +108,91 @@ def __init__(
...
-        if order % 2 == 1:
+        if order > 1 and order % 2 == 1:
             raise ValueError(
                 "Suzuki product formulae are symmetric and therefore only defined "
-                "for even orders."
+                f"for when the order is 1 or even, not {order}."
             )
         super().__init__(order, reps, insert_barriers, cx_structure, atomic_evolution, wrap)
 
-    def synthesize(self, evolution):
```
ShellyGarion, line 115: this is not deprecated - right, just called from ProductFormula.synthesize?

releasenotes/notes/pauli-evo-plugins-612850146c3f7d49.yaml
```diff
@@ -0,0 +1,32 @@
...
+  - |
+    Added :meth:`.ProductFormula.expand` which allows to view the expansion of a product formula
+    in a sparse Pauli format.
+  - |
+    Added the plugin structure for the :class:`.PauliEvolutionGate`. The default plugin,
+    :class:`.PauliEvolutionSynthesisDefault`, constructs circuit as before (but faster as it
+    internally uses Rust).
+
+upgrade:
+  - |
```
ShellyGarion, line 29: would you like to comment (in the release notes) on the performance improvement by porting this code to rust?

ShellyGarion (commented):
LGTM. I only have a few comments and questions.

qiskit/synthesis/evolution/lie_trotter.py
Cryoris: Correct 🙂

qiskit/synthesis/evolution/suzuki_trotter.py
Cryoris: Correct 🙂

qiskit/circuit/library/pauli_evolution.py
Cryoris: It's the same, `ParameterValueType` is a typedef used across the circuit and the gates, it's defined as 
```python
ParameterValueType = Union[ParameterExpression, float]
```
(see the `quantumcircuit.py` file)

crates/accelerate/src/circuit_library/pauli_evolution.rs
Cryoris: This typedef represents a packed instruction, which can be used as input to `CircuitData.from_packed_instructions` (so it's not specific to an evolution). The same is used in some other places (e.g. quantum volume or pauli feature map), so I'd prefer keeping as is for now and potentially adding a general typedef used across all of the rust code 🙂

releasenotes/notes/pauli-evo-plugins-612850146c3f7d49.yaml
Cryoris: done in 8d5e3f3

crates/accelerate/src/circuit_library/pauli_evolution.rs
Cryoris: done in 8d5e3f3

crates/accelerate/src/circuit_library/pauli_evolution.rs
Cryoris: done in 8d5e3f3

alexanderivrii (approved):
Thanks @Cryoris! Let me approve the PR, but let's also give a chance to @ShellyGarion and @raynelfss to see if they have additional comments.

raynelfss (approved):
Taking all previous discussions into consideration, this LGTM!

PR #13298, opened 2024-10-08
Implement arithmetic on `SparseObservable`

jakelishman:
### Summary

The simple arithmetic operators of addition, subtraction, multiplication, division and tensor product are in line with other `quantum_info` objects, including coercion from other objects (though better behaved on failed coercions, in many cases).

Where appopriate, in-place operations will re-use the existing allocations.  The tensor product cannot really be done in place, so it doesn't define a special `__rxor__` operator, but it inherits the standard Python behaviour of this being derived from `__xor__`.

There are further mathematical operations to add around composition and evolution, and to apply a `TranspileLayout` to the observable.  All of these, in the `quantum_info` style either directly deal with a layout, or take a `qargs` argument that is effectively a sublayout for the right-hand side of the operation.  These will be added in a follow-up.

### Details and comments

~Built on top of #12671, so depends on it.~

### DIFF
--- a/crates/accelerate/src/sparse_observable.rs
+++ b/crates/accelerate/src/sparse_observable.rs
@@ -1206,16 +1206,27 @@ impl SparseObservable {
         slf.borrow().eq(&other.borrow())
     }
 
-    fn __add__(&self, other: &Bound<PyAny>) -> PyResult<Py<PyAny>> {
-        let py = other.py();
+    fn __add__(slf_: &Bound<Self>, other: &Bound<PyAny>) -> PyResult<Py<PyAny>> {
+        let py = slf_.py();
+        if slf_.is(other) {
+            // This fast path is for consistency with the in-place `__iadd__`, which would otherwise
+            // struggle to do the addition to itself.
+            return Ok(<&SparseObservable as ::std::ops::Mul<_>>::mul(
+                &slf_.borrow(),
+                Complex64::new(2.0, 0.0),
+            )
+            .into_py(py));
+        }
         let Some(other) = coerce_to_observable(other)? else {
             return Ok(py.NotImplemented());
         };
+        let slf_ = slf_.borrow();
         let other = other.borrow();
-        self.check_equal_qubits(&other)?;
-        Ok((self + &other).into_py(py))
+        slf_.check_equal_qubits(&other)?;
+        Ok(<&SparseObservable as ::std::ops::Add>::add(&slf_, &other).into_py(py))
     }
     fn __radd__(&self, other: &Bound<PyAny>) -> PyResult<Py<PyAny>> {
+        // No need to handle the `self is other` case here, because `__add__` will get it.
         let py = other.py();
         let Some(other) = coerce_to_observable(other)? else {
             return Ok(py.NotImplemented());
@@ -1227,32 +1238,36 @@ impl SparseObservable {
     fn __iadd__(slf_: Bound<SparseObservable>, other: &Bound<PyAny>) -> PyResult<()> {
         if slf_.is(other) {
             *slf_.borrow_mut() *= Complex64::new(2.0, 0.0);
-        } else {
-            let mut slf_ = slf_.borrow_mut();
-            let Some(other) = coerce_to_observable(other)? else {
-                // This is not well behaved - we _should_ return `NotImplemented` to Python space
-                // without an exception, but limitations in PyO3 prevent this at the moment.  See
-                // https://github.com/PyO3/pyo3/issues/4605.
-                return Err(PyTypeError::new_err(format!(
-                    "invalid object for in-place addition of 'SparseObservable': {}",
-                    other.repr()?
-                )));
-            };
-            let other = other.borrow();
-            slf_.check_equal_qubits(&other)?;
-            *slf_ += &other;
+            return Ok(());
         }
+        let mut slf_ = slf_.borrow_mut();
+        let Some(other) = coerce_to_observable(other)? else {
+            // This is not well behaved - we _should_ return `NotImplemented` to Python space
+            // without an exception, but limitations in PyO3 prevent this at the moment.  See
+            // https://github.com/PyO3/pyo3/issues/4605.
+            return Err(PyTypeError::new_err(format!(
+                "invalid object for in-place addition of 'SparseObservable': {}",
+                other.repr()?
+            )));
+        };
+        let other = other.borrow();
+        slf_.check_equal_qubits(&other)?;
+        *slf_ += &other;
         Ok(())
     }
 
-    fn __sub__(&self, other: &Bound<PyAny>) -> PyResult<Py<PyAny>> {
-        let py = other.py();
+    fn __sub__(slf_: &Bound<Self>, other: &Bound<PyAny>) -> PyResult<Py<PyAny>> {
+        let py = slf_.py();
+        if slf_.is(other) {
+            return Ok(SparseObservable::zero(slf_.borrow().num_qubits).into_py(py));
+        }
         let Some(other) = coerce_to_observable(other)? else {
             return Ok(py.NotImplemented());
         };
+        let slf_ = slf_.borrow();
         let other = other.borrow();
-        self.check_equal_qubits(&other)?;
-        Ok((self - &other).into_py(py))
+        slf_.check_equal_qubits(&other)?;
+        Ok(<&SparseObservable as ::std::ops::Sub>::sub(&slf_, &other).into_py(py))
     }
     fn __rsub__(&self, other: &Bound<PyAny>) -> PyResult<Py<PyAny>> {
         let py = other.py();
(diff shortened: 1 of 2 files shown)


jakelishman:
Now rebased and ready for review.

crates/accelerate/src/sparse_observable.rs
```diff
@@ -712,6 +819,82 @@ impl SparseObservable {
...
+    pub fn tensor(&self, other: &SparseObservable) -> SparseObservable {
+        let mut out = SparseObservable::with_capacity(
+            self.num_qubits + other.num_qubits,
+            self.coeffs.len() * other.coeffs.len(),
+            other.coeffs.len() * self.bit_terms.len() + self.coeffs.len() * other.bit_terms.len(),
+        );
+        let mut self_indices = Vec::new();
+        for self_term in self.iter() {
+            self_indices.clear();
+            self_indices.extend(self_term.indices.iter().map(|i| i + other.num_qubits));
```
Cryoris, line 885: Is this more efficient than something like
```
let self_indices = self_term.indices.iter().map(|i| i + other.num_qubits).collect();
```
or what's the reason for the clear+extend? 🙂

crates/accelerate/src/sparse_observable.rs
```diff
@@ -998,6 +1206,133 @@ impl SparseObservable {
         slf.borrow().eq(&other.borrow())
     }
 
+    fn __add__(&self, other: &Bound<PyAny>) -> PyResult<Py<PyAny>> {
+        let py = other.py();
+        let Some(other) = coerce_to_observable(other)? else {
+            return Ok(py.NotImplemented());
```
Cryoris, line 1221: I would've expected an error here (which I guess `py.NotImplemented()` is?), why does this need to be wrapped in an `Ok`?

crates/accelerate/src/sparse_observable.rs
```diff
@@ -998,6 +1206,133 @@ impl SparseObservable {
...
+        Ok((self + &other).into_py(py))
+    }
+    fn __radd__(&self, other: &Bound<PyAny>) -> PyResult<Py<PyAny>> {
+        let py = other.py();
+        let Some(other) = coerce_to_observable(other)? else {
+            return Ok(py.NotImplemented());
+        };
+        let other = other.borrow();
+        self.check_equal_qubits(&other)?;
+        Ok((<&SparseObservable as ::std::ops::Add>::add(&other, self)).into_py(py))
```
Cryoris, line 1236: The reason to not make this `Ok((self + &other).into_py(py))` is to be consistent in how the bit terms/indices are concatenated right? Meaning that something like `A + B` will always be sorted as `<bit terms of A><bit terms of B>`?

Cryoris (commented):
This looks good to me I just have some short questions below 🙂

crates/accelerate/src/sparse_observable.rs
```diff
@@ -998,6 +1206,133 @@ impl SparseObservable {
...
+    }
+    fn __iadd__(slf_: Bound<SparseObservable>, other: &Bound<PyAny>) -> PyResult<()> {
+        if slf_.is(other) {
+            *slf_.borrow_mut() *= Complex64::new(2.0, 0.0);
```
Cryoris, line 1240: This is a bit inconsistent with `obs + obs` which does not equal `2 * obs`, would we want to also special case the standard addition?

crates/accelerate/src/sparse_observable.rs
jakelishman: Yours creates a new allocation on each loop iteration, mine re-uses the same allocation.  Most of the time (in general, not just here), re-using an allocation is faster than making a new one.

crates/accelerate/src/sparse_observable.rs
jakelishman: In PyO3, returning an `Err` variant means returning something that'll be raised as an exception.  In Python, there's two similar objects: `NotImplementedError`, which _is_ a full exception, and `NotImplemented`, which is a singleton and not an error.  In the magic arithmetic functions, you're not supposed to raise an exception yourself, but to return `NotImplemented`, which is the cue to the interpreter to try the `__radd__` method of `other` next.  The reason to have the split is that exceptions are relatively expensive to raise, but it's lightning fast to check a singleton.

Returning `NotImplemented` isn't an exception state - we're saying "we don't define `__add__` for this type of `other`", but that alone isn't an exception, because `other` might handle it.  There's a bit more detail here: https://docs.python.org/3/library/constants.html#NotImplemented.

crates/accelerate/src/sparse_observable.rs
jakelishman: What you're suggesting is what's going on here - `__radd__` is called on `self` as the handler for `other + self` iff the "forwards" `__add__` method of `other` returned `NotImplemented` (indicating that `other` doesn't know how to add `SparseObservable`).

So this case handles things like `"IXI" + SparseObservable("ZZZ")`, and makes sure that the output is like `SparseObservable.from_list([("IXI", 1.0), ("ZZZ", 1.0)])` in that order.  (Fwiw, I don't love the magic coercion of weird types to `SparseObservable`, but it's what all the other operators in `quantum_info` do, so I was matching that.)

crates/accelerate/src/sparse_observable.rs
```diff
@@ -998,6 +1206,133 @@ impl SparseObservable {
...
+        self * other
+    }
+    fn __rmul__(&self, other: Complex64) -> SparseObservable {
+        other * self
+    }
+    fn __imul__(&mut self, other: Complex64) {
+        *self *= other;
+    }
+
+    fn __truediv__(&self, other: Complex64) -> PyResult<SparseObservable> {
```
Cryoris, line 1322: Doing something ridiculous like 
```
SparseObservable.from_label("X") / 1e-200
```
gives coeffs that are `jnf+NaNj` but `(1+0j)/1e-200` gives, as I would expect, just `1e200+0j`. Is this difference in behavior of complex number handling due to how Rust represents numbers?

jakelishman: It's because in Rust space, we coerce `1e-200` up to a `Complex64` on entry to the method, whereas Python is doing the same thing as C, and treating `complex / float` by distributing the division over both components.

You can get the same behaviour in Python with `complex(1, 0)/complex(1e-200, 0)`

crates/accelerate/src/sparse_observable.rs
```diff
@@ -936,6 +1125,25 @@ impl SparseObservable {
...
+    /// substraction operations may not need to reallocate.
+    ///
+    /// Examples:
+    ///
+    ///     .. code-block:: python
+    ///
+    ///         >>> obs = SparseObservable.from_list([("IX+-rl", 2.0), ("01YZII", -1j)])
+    ///         >>> obs.clear()
+    ///         >>> assert obs == SparseObservable.zero(obs.num_qubits)
+    pub fn clear(&mut self) {
```
Cryoris, line 1140: IMO this seems a bit exotic to have on an operator, is this mainly for internal use or do you have a user use-case in mind? 🙂

crates/accelerate/src/sparse_observable.rs
jakelishman: The reason to have the special case is because you can't borrow `self` as mutable (to hold the result) _and_ immutable (to be the addand).  So there's no easy way to pass it to any existing `add` implementation here in the module, and I used this trick to avoid introducing a clone in the operation.

crates/accelerate/src/sparse_observable.rs
```diff
@@ -712,6 +819,82 @@ impl SparseObservable {
...
+        self.into()
+    }
+
+    /// Reduce the observable to its canonical form.
+    ///
+    /// This sums like terms, removing them if the final complex coefficient's absolute value is
+    /// less than or equal to the tolerance.  The terms are reordered to some canonical ordering.
+    ///
+    /// This function is idempotent.
+    pub fn canonicalize(&self, tol: f64) -> SparseObservable {
```
Cryoris, line 835: You probably were expecting this question 😛 but wouldn't it make sense to keep this in line with the Python-side name `simplify`? I guess you were renaming because the main use would be the canonical representation rather than cutting values off?

crates/accelerate/src/sparse_observable.rs
Cryoris: Ohhh gotcha, I was wondering what that difference is but never looked it up 🌚

crates/accelerate/src/sparse_observable.rs
Cryoris: ahh I see, so you would've preferred to have this be the same outcome as `obs + obs` but for performance we just do the multiplication -- but I'm also wondering whether we should then do the same for `obs + obs -> 2 * obs` for consistency? 🤔

crates/accelerate/src/sparse_observable.rs
jakelishman: It's used internally by `__isub__`, and I thought it was possibly useful, so I just exposed it.  I can avoid exposing it if you prefer.

crates/accelerate/src/sparse_observable.rs
jakelishman: I suppose yeah, it's technically a performance optimisation, though really I think that rewriting it as `2. * obs` was just the thing that came more naturally to me, because I avoid `clone` just by default whenever possible.

I can make the behaviour of the not-in-place forms the same as the in-place forms if you prefer?

crates/accelerate/src/sparse_observable.rs
jakelishman: I called the Python-space method `simplify` for consistency with `SparsePauliOp`, but I actually thought it wasn't an entirely accurate name for `SparseObservable`, because to me, `simplify` would imply a more complex operation that attempts to re-factorise expressions to reduce the number of terms (like collapsing $\lvert 0\rangle\langle 0\rvert + \lvert 1\rangle\langle 1\rvert$ to $X$).

Cryoris: Ah I see, yeah that would be more involved... let's keep `canonicalize` then 🙂

crates/accelerate/src/sparse_observable.rs
Cryoris: Yeah I think that would be nice, unless you have concerns/other expectations 🙂 (I think the same applies to `__sub__`, right?)

crates/accelerate/src/sparse_observable.rs
Cryoris: Ok yeah, I guess it doesn't hurt to have it 🙂 plus maybe we're getting to a point where users would like to focus more on writing performant code, then having this would be a benefit

Cryoris (approved):
LGTM!

Issue #13285, opened 2024-10-07
Add an efficient circuit synthesis for a special pattern `exp(-it (IZZ + ZZI + ZZZ))`


t-imamichi:
### What should we add?

[1] proposed an efficient way to synthesize the following special pattern of two-body terms (`IZZ` and `ZZI`) and three-body term (`ZZZ`) (Fig. 2 [1]).
It can halve the number of two-qubit gates (8 to 4).
This pattern is useful to handle QAOA with cubic terms as [1] did.

It would be nice if Qiskit can handle this special pattern of synthesis.

Target  (`coeffs` can be arbitrary)
```
     ┌────────────────────────────────┐
q_0: ┤0                               ├
     │                                │
q_1: ┤1 exp(-it (IZZ + ZZI + ZZZ))(x) ├
     │                                │
q_2: ┤2                               ├
     └────────────────────────────────┘
# PauliEvolutionGate(SparsePauliOp(["IZZ", "ZZI", "ZZZ"], coeffs=[1, 2, 3]), x)
```

Qiskit optimization_level=3
```
                                                        ┌───┐┌───────────┐┌───┐
q_0: ──■─────────────────■──────────────────────────────┤ X ├┤ Rz(6.0*x) ├┤ X ├─────
     ┌─┴─┐┌───────────┐┌─┴─┐                       ┌───┐└─┬─┘└───────────┘└─┬─┘┌───┐
q_1: ┤ X ├┤ Rz(2.0*x) ├┤ X ├──■─────────────────■──┤ X ├──■─────────────────■──┤ X ├
     └───┘└───────────┘└───┘┌─┴─┐┌───────────┐┌─┴─┐└─┬─┘                       └─┬─┘
q_2: ───────────────────────┤ X ├┤ Rz(4.0*x) ├┤ X ├──■───────────────────────────■──
                            └───┘└───────────┘└───┘
OrderedDict({'cx': 8, 'rz': 3})
```

Proposed circuit by [1]
```
q_0: ──■───────────────────────────────────■────────────────────
     ┌─┴─┐┌───────────┐┌───┐┌───────────┐┌─┴─┐┌───────────┐┌───┐
q_1: ┤ X ├┤ Rz(2.0*x) ├┤ X ├┤ Rz(6.0*x) ├┤ X ├┤ Rz(4.0*x) ├┤ X ├
     └───┘└───────────┘└─┬─┘└───────────┘└───┘└───────────┘└─┬─┘
[...]

PR #13301, opened 2024-10-09
Add Rustiq-based synthesis for `PauliEvolutionGate`

### LINKED ISSUE #12789: Add `PauliEvolutionGate` synthesis plugin using rustiq

alexanderivrii:
### Summary

Calls the `Rustiq's` synthesis algorithm for pauli networks natively from within Qiskit. 

Together with #13295 addresses #12789.
This should be rebased on top of #13295, in order to support the plugin interface structure for `PauliEvolutionGate`s.

Rustiq is available at https://github.com/smartiel/rustiq-core

### How to experiment

One can experiment with the new functionality by creating a `PauliEvolutionGate` and directly calling the new `PauliEvolutionSynthesisRustiq` on this gate:

```
from qiskit.transpiler.passes.synthesis.hls_plugins import PauliEvolutionSynthesisRustiq

terms = ["IX", "ZZ"]
evo_gate = PauliEvolutionGate(SparsePauliOp(terms), time=0.12)
qct = PauliEvolutionSynthesisRustiq().run(evo_gate)
```

Alternatively, one can create a circuit with `PauliEvolutionGate`s and specify "rustiq" via the synthesis config interface:

```
terms = ["IX", "ZZ"]
evo_gate = PauliEvolutionGate(SparsePauliOp(terms), time=0.12)
qc = QuantumCircuit(3)
qc.append(evo_gate, [1, 2])
config = HLSConfig(PauliEvolution=["rustiq"])
qct = transpile(qc, hls_config=config)
```

### Details:

The Qiskit Rust function `pauli_network_synthesis` calls the `greedy_pauli_network` function from the `rustiq-core` (in particular, most of the code is fully in Rust). However, the latter function only returns a sequence of Clifford gates, without the single-qubit rotations. So `pauli_network_synthesis` also adapts the code from Simon's private rustiq-plugin repository that simulates the circuit and inserts rotations wherever required.

### Preliminary Experiments:

To be updated (or maybe it would be even best to move this discussion to the issue #12789). After running some preliminary experiments on some specific examples (e.g. see #13285) and some hamlib examples (taken from https://github.com/Qiskit/benchpress/blob/main/benchpress/hamiltonian/hamlib/), `rustiq` can be amazingly powerful in many cases, greatly reducing the two-qubits gate-count and depth of the resulting circuit. On the other hand, there are also cases where Rustiq performs worse, i.e. creating a circuit with more gates.

### Some Concerns:

- On some of the hamlib examples Rustiq is extremely fast, yet there are also examples where it takes more than 1 hour to complete. 
- In some cases the resulting circuit is not fully deterministic. I am currently trying to understand why this happens. 
- So far in all the cases the circuit produced by Rustiq is equivalent to the circuit produced by the default synthesis algorithm in Qiskit, _up to the global phase_, however the global phase difference is not always 0. This is after taking care of global phase updates related to all-I pauli rotation gates. I am also trying to understand why this happens.

It is possible that some additional changes to Rustiq itself might be required. 

### ToDo:

- [x] Docstring improvements, release notes, new plugin documentation, documenting rustiq-specific options, etc. 
- [x] Tests: make sure that the output circuits are correct and are equivalent to the circuits produced by the default algorithm, make sure that the plugins run as intended


crates/accelerate/src/synthesis/evolution/mod.rs
```diff
@@ -0,0 +1,336 @@
...
+use rustworkx_core::petgraph::graph::NodeIndex;
+use rustworkx_core::petgraph::prelude::StableDiGraph;
+use rustworkx_core::petgraph::Incoming;
+
+/// TODO:
+///   * make sure that qubit indexing is correct
+///   * should we allow all-I rotations?
+
+/// A Qiskit gate
+pub type QiskitGate = (StandardGate, SmallVec<[Param; 3]>, SmallVec<[Qubit; 2]>);
```
Cryoris: Does this need to be public? 🙂

crates/accelerate/src/synthesis/evolution/mod.rs
```diff
@@ -0,0 +1,336 @@
...
+
+/// Expands the sparse pauli string representation to the full representation.
+///
+/// For example: for the input `sparse_pauli = "XY", qubits = [1, 3], num_qubits = 6`,
+/// the function returns `"IXIYII"`.
+fn expand_pauli(sparse_pauli: String, qubits: Vec<u32>, num_qubits: usize) -> String {
+    let mut v: Vec<char> = vec!["I".parse().unwrap(); num_qubits];
```
Cryoris: Does this also work? That should directly use the `char` type
```suggestion
    let mut v: Vec<char> = vec!['I'; num_qubits];
```

crates/accelerate/src/synthesis/evolution/mod.rs
```diff
@@ -0,0 +1,336 @@
...
+            // we need to multiply the angle by 2
+            // we also need to negate it when there is a phase
+            let param = match phase {
+                false => multiply_param(angle, 2.0, py),
+                true => multiply_param(angle, -2.0, py),
+            };
+            return (standard_gate, smallvec![param], smallvec![Qubit(q as u32)]);
+        }
+    }
+    unreachable!()
```
Cryoris: It might be nice to add an explanation as to why this is unreachable, e.g.
```suggestion
    unreachable!("The paulis are guaranteed to have at least one non-I term at this point.")
```

crates/accelerate/src/synthesis/evolution/mod.rs
```diff
@@ -0,0 +1,336 @@
...
+
+/// A DAG that stores ordered Paulis, up to commutativity.
+struct CommutativityDag {
+    /// Rustworkx's DAG
+    dag: StableDiGraph<usize, ()>,
+}
+
+impl CommutativityDag {
+    /// Construct a DAG based on the commutativity relations between paulis.
+    pub fn from_paulis(paulis: &PauliSet) -> Self {
```
Cryoris: Same question here, do these methods need to be public? Since the struct itself is not I would assume that we can just have these private? 🤔

crates/accelerate/src/synthesis/permutation/mod.rs
```diff
@@ -37,6 +37,7 @@ pub fn _validate_permutation(py: Python, pattern: PyArrayLike1<i64>) -> PyResult
 #[pyfunction]
 #[pyo3(signature = (pattern))]
 pub fn _inverse_pattern(py: Python, pattern: PyArrayLike1<i64>) -> PyResult<PyObject> {
+    println!("I AM HERE YES");
```
Cryoris: 🖨️

qiskit/synthesis/evolution/__init__.py
```diff
@@ -12,6 +12,9 @@
 
 """Synthesis for operator evolution gates."""
 
+# pylint: disable=unused-import
+from qiskit._accelerate.synthesis.evolution import pauli_network_synthesis
```
Cryoris: The classes in here perform the full synthesis of a `PauliEvolutionGate` to a `QuantumCircuit`,  which this function does not. Would it make more sense to expose a similar wrapper here (basically what the plugin is doing) or just to remove this function as it is more a utility for the plugin?

Cryoris (commented):
I just had a brief look, but overall this looks good to me -- I'll give a detailed review later 🙂

crates/accelerate/src/synthesis/evolution/mod.rs
alexanderivrii: fixed!

crates/accelerate/src/synthesis/evolution/mod.rs
alexanderivrii: Great, indeed!

crates/accelerate/src/synthesis/evolution/mod.rs
alexanderivrii: yes, fixed

qiskit/synthesis/evolution/__init__.py
alexanderivrii: I agree. To be consistent with the other synthesis methods, I have wrapped this into a function `synth_pauli_network_rustiq`.

crates/accelerate/src/synthesis/clifford/utils.rs
```diff
@@ -227,6 +227,24 @@ impl Clifford {
                     clifford.append_s(qubits[0].index());
                     Ok(())
                 }
+                StandardGate::SdgGate => {
+                    clifford.append_sdg(qubits[0].0 as usize);
+                    Ok(())
+                }
+                StandardGate::SXGate => {
+                    // todo: replace by a single update
```
ShellyGarion: this can be done already in this PR, see:
https://github.com/Qiskit/qiskit/blob/db0433920c86e55f4ec06b3837d10d94b040e117/qiskit/quantum_info/operators/symplectic/clifford_circuits.py#L325

crates/accelerate/src/synthesis/clifford/utils.rs
```diff
@@ -227,6 +227,24 @@ impl Clifford {
...
+                    clifford.append_sdg(qubits[0].0 as usize);
+                    clifford.append_h(qubits[0].0 as usize);
+                    clifford.append_sdg(qubits[0].0 as usize);
+                    Ok(())
+                }
+                StandardGate::SXdgGate => {
+                    // todo: replace by a single update
```
ShellyGarion: this can be done already as part of this PR, see:
https://github.com/Qiskit/qiskit/blob/db0433920c86e55f4ec06b3837d10d94b040e117/qiskit/quantum_info/operators/symplectic/clifford_circuits.py#L343

crates/accelerate/src/synthesis/evolution/mod.rs
```diff
@@ -0,0 +1,431 @@
...
+    }
+    unreachable!("The pauli rotation is guaranteed to be a single-qubit rotation.")
+}
+
+// Note: The pauli network synthesis algorithm in rustiq-core 0.0.8 only returns
+// the list of Clifford gates that, when simulated, turn every pauli rotation at
+// some point to a single-qubit rotation, in the given order, but up to commutativity.
+// Following the code in Simon's private rustiq-plugin repository, we simulate the
+// clifford gates to find where pauli rotations need to be inserted. If the future
+// the synthesis algorithm will explicitly return where the rotations need to be
```
ShellyGarion: If the future the synthesis ? --> remove one "the"

crates/accelerate/src/synthesis/evolution/mod.rs
```diff
@@ -0,0 +1,431 @@
...
+use pyo3::types::{PyList, PyString, PyTuple};
+use smallvec::{smallvec, SmallVec};
+use std::borrow::Borrow;
+
+use qiskit_circuit::circuit_data::CircuitData;
+use qiskit_circuit::operations::{multiply_param, radd_param, Param, StandardGate};
+use qiskit_circuit::Qubit;
+
+use rustiq_core::structures::{
+    CliffordCircuit, CliffordGate, IsometryTableau, Metric, PauliLike, PauliSet,
```
ShellyGarion: Is there a check needed that the user installed rustiq before using this code and raise a warning otherwise?

crates/accelerate/src/synthesis/evolution/mod.rs
```diff
@@ -0,0 +1,431 @@
...
+/// * pauli_network: pauli network represented in sparse format. It's a list
+///     of triples such as `[("XX", [0, 3], theta), ("ZZ", [0, 1], 0.1)]`.
+/// * optimize_count: if `true`, Rustiq's synthesis algorithms aims to optimize
+///     the 2-qubit gate count; and if `false`, then the 2-qubit depth.
+/// * preserve_order: whether the order of paulis should be preserved, up to
+///     commutativity. If the order is not preserved, the returned circuit will
+///     generally not be equivalent to the given pauli network.
+/// * upto_clifford: if `true`, the final Clifford operator is not synthesized
+///     and the returned circuit will generally not be equivalent to the given
+///     pauli network. In addition, the argument `upto_phase` would be ignored.
```
ShellyGarion, line 38: If it's the final Clifford, then one can have both "upto_clifford" and "upto_phase" since the global phase should not affect the final measurement?

qiskit/synthesis/evolution/pauli_network.py
```diff
@@ -0,0 +1,80 @@
...
+    preserve_order: bool,
+    upto_clifford: bool,
+    upto_phase: bool,
+    resynth_clifford_method: int,
+) -> QuantumCircuit:
+    """
+    Calls Rustiq's pauli network synthesis algorithm.
+
+    The algorithm is described in [1]. The source code (in Rust) is available at
+    https://github.com/smartiel/rustiq-core.
```
ShellyGarion, line 37: should the user install and compile rustiq-core for this code ?

ShellyGarion (commented):
LGTM. I only have a few comments and questions.

qiskit/transpiler/passes/synthesis/hls_plugins.py
```diff
@@ -1047,8 +1079,88 @@ def run(self, high_level_object, coupling_map=None, target=None, qubits=None, **
...
+
+    The plugin supports the following additional options:
+
+    * optimize_count (bool): if `True` the synthesis algorithm will try to optimize
+        the 2-qubit gate count; and if `False` then the 2-qubit depth.
+    * preserve_order (bool): whether the order of paulis should be preserved, up to
+        commutativity.
+    * upto_clifford (bool): if `True`, the final Clifford operator is not synthesized.
+    * upto_phase (bool): if `True`, the global phase of the returned circuit may
+        differ from the global phase of the given pauli network.
```
ShellyGarion, line 1503: so according to the docs, one can have both "upto_clifford=True" and "upto_phase=True" ?

qiskit/synthesis/evolution/pauli_network.py
alexanderivrii: No, the rustiq-core (version 0.0.8) is available [crates.io](https://crates.io/crates/rustiq-core), and it's "pulled in" during Qiskit's Rust compilation (as all the other Rust crates Qiskit internally uses). There is nothing specific a user needs to install.

crates/accelerate/src/synthesis/evolution/mod.rs
alexanderivrii: I don't know if it makes sense to ignore the final Clifford but to keep the global phase. And in any case currently we don't have a way to do that (without the extended Clifford simulator). So the assumption that I tried to document here is that if we ignore the final Clifford, we also automatically ignore the global phase. Anyway, for now this is really an experimental option available to users (I am pretty sure that @ajavadia would like to have it). I also expect that in future, when synthesizing up to the final Clifford, the calling application would like to obtain this Clifford as a part of the result (for instance, to modify the measurement operators that need to be performed), and we currently don't have API for any of this.

crates/accelerate/src/synthesis/clifford/utils.rs
alexanderivrii: Thanks, I planned to do it, but forgot. Added in [41ae2f3](https://github.com/Qiskit/qiskit/pull/13301/commits/41ae2f33146762e3ec0b0e456d3f63dc40e0d0c9).

crates/accelerate/src/synthesis/clifford/utils.rs
alexanderivrii: Same here. Added in [41ae2f3](https://github.com/Qiskit/qiskit/pull/13301/commits/41ae2f33146762e3ec0b0e456d3f63dc40e0d0c9).

crates/accelerate/src/synthesis/evolution/mod.rs
alexanderivrii: I changed the wording of this entire note to explain more clearly the challenges of integrating rustiq code.

crates/accelerate/src/synthesis/evolution/mod.rs
alexanderivrii: No, as I wrote somewhere else, the users do not need to install anything. Qiskit will simply not compile if the above package does not exist. Also note that in `Cargo.toml` we explicitly pin rustiq's version to "0.0.8", so we are also safe against any updates to rustiq-core.

qiskit/transpiler/passes/synthesis/hls_plugins.py
alexanderivrii: Yes, so if both are `False`, we want the exact synthesis -- this is the default behavior. If `upto_clifford=False` but `upto_phase` is `True`, then we can synthesize ignoring the global phase, in which case by default we use Qiskit's Clifford resynthesis, since it's usually faster and gives a very comparable quality of results. If `upto_clifford` it `True` then (as noted somewhere else) we don't care about the global phase either.

ShellyGarion (dismissed):
LGTM - thanks!

crates/accelerate/src/synthesis/clifford/utils.rs
```diff
@@ -149,6 +149,30 @@ impl Clifford {
         azip!((z in &mut z, &x in &x) *z ^= x);
     }
 
+    /// Modifies the tableau in-place by appending SX-gate
+    pub fn append_sx(&mut self, qubit: usize) {
```
Cryoris, line 153: These look correct, so I'm fine adding as is, but do we have a test that could specifically trigger these? Like constructing a Clifford object from a circuit with SX gates? 🙂

crates/accelerate/src/synthesis/evolution/mod.rs
Cryoris, line 1: Could we move the logic into a separate file, that is not `mod.rs`? I think in most code, the `mod.rs` only has the module imports or, at most, the `pyfunction` definition and not much logic. Otherwise we might end up with very different functionalities in the asme file 🙂

crates/accelerate/src/synthesis/evolution/mod.rs
```diff
@@ -0,0 +1,439 @@
...
+//
+// This code is licensed under the Apache License, Version 2.0. You may
+// obtain a copy of this license in the LICENSE.txt file in the root directory
+// of this source tree or at http://www.apache.org/licenses/LICENSE-2.0.
+//
+// Any modifications or derivative works of this code must retain this
+// copyright notice, and modified files need to carry a notice indicating
+// that they have been altered from the originals.
+
+#![allow(clippy::too_many_arguments)]
```
Cryoris: I would assume theres only very few functions that need this exception, could you move it to these instead of a module-level except? 🙂

crates/accelerate/src/synthesis/evolution/mod.rs
```diff
@@ -0,0 +1,439 @@
...
+
+/// A Qiskit gate. The quantum circuit data returned by the pauli network
+/// synthesis algorithm will consist of clifford and rotation gates.
+type QiskitGate = (StandardGate, SmallVec<[Param; 3]>, SmallVec<[Qubit; 2]>);
+
+/// Expands the sparse pauli string representation to the full representation.
+///
+/// For example: for the input `sparse_pauli = "XY", qubits = [1, 3], num_qubits = 6`,
+/// the function returns `"IXIYII"`.
+fn expand_pauli(sparse_pauli: String, qubits: Vec<u32>, num_qubits: usize) -> String {
```
Cryoris: Could this take a qubit slice (i.e. `qubits: &[u32]`) to avoid copying the vector itself?

crates/accelerate/src/synthesis/evolution/mod.rs
```diff
@@ -0,0 +1,439 @@
...
+    let mut v: Vec<char> = vec!['I'; num_qubits];
+    for (i, q) in qubits.iter().enumerate() {
+        v[*q as usize] = sparse_pauli.chars().nth(i).unwrap();
```
Cryoris: Just a suggestion, as I don't know if `chars().nth()` is for free
```suggestion
    for (q, p) in qubits.iter().zip(sparse_pauli.chars()) {
        v[*q as usize] = p;
```
(this might require some dereferencing or cloning -- my rust skills decrease by 98% without `rust-analyzer`)

crates/accelerate/src/synthesis/evolution/mod.rs
```diff
@@ -0,0 +1,439 @@
...
+}
+
+/// Return the Qiskit rotation gate corresponding to the single-qubit Pauli rotation.
+///
+/// # Arguments
+///
+/// * py: a GIL handle, needed to negate rotation parameters in Python space.
+/// * paulis: Rustiq's data structure storing pauli rotations.
+/// * i: index of the single-qubit Pauli rotation.
+/// * angle: - Qiskit's rotation angle.
```
Cryoris:
```suggestion
/// * angle: Qiskit's rotation angle.
```

crates/accelerate/src/synthesis/evolution/mod.rs
```diff
@@ -0,0 +1,439 @@
...
+fn qiskit_rotation_gate(py: Python, paulis: &PauliSet, i: usize, angle: &Param) -> QiskitGate {
+    let (phase, pauli_str) = paulis.get(i);
+    for (q, c) in pauli_str.chars().enumerate() {
+        if c != 'I' {
+            let standard_gate = match c {
+                'X' => StandardGate::RXGate,
+                'Y' => StandardGate::RYGate,
+                'Z' => StandardGate::RZGate,
+                _ => panic!(),
```
Cryoris: Personally, I would use `unreachable` here as we don't expect this arm can be reached, whereas `panic` is maybe less precise? Plus it might be nice adding a small message to simplify tracking down where this came from 🙂 But I'm also happy to leave as is, this is just a suggestion.
```suggestion
                _ => unreachable!("Only X, Y and Z are possible characters at this point."),
```

crates/accelerate/src/synthesis/evolution/mod.rs
```diff
@@ -0,0 +1,439 @@
...
+///
+/// # Arguments
+///
+/// * py: a GIL handle, needed to negate rotation parameters in Python space.
+/// * paulis: Rustiq's data structure storing pauli rotations.
+/// * i: index of the single-qubit Pauli rotation.
+/// * angle: - Qiskit's rotation angle.
+fn qiskit_rotation_gate(py: Python, paulis: &PauliSet, i: usize, angle: &Param) -> QiskitGate {
+    let (phase, pauli_str) = paulis.get(i);
+    for (q, c) in pauli_str.chars().enumerate() {
```
Cryoris: Is `pauli_str` guaranteed to be upper case? If not, it might be worth calling `.upper()` on it first

crates/accelerate/src/synthesis/evolution/mod.rs
```diff
@@ -0,0 +1,439 @@
...
+    /// Rustworkx's DAG
+    dag: StableDiGraph<usize, ()>,
+}
+
+impl CommutativityDag {
+    /// Construct a DAG based on commutativity relations between paulis.
+    fn from_paulis(paulis: &PauliSet) -> Self {
+        let mut dag = StableDiGraph::<usize, ()>::new();
+
+        let node_indices: Vec<NodeIndex> = (0..paulis.len()).map(|i| dag.add_node(i)).collect();
```
Cryoris: I'm surprised clippy doesn't complain about this, but is this equivalent?
```suggestion
        let node_indices: Vec<NodeIndex> = (0..paulis.len()).map(dag.add_node).collect();
```

crates/accelerate/src/synthesis/evolution/mod.rs
```diff
@@ -0,0 +1,439 @@
...
+/// # Arguments
+///
+/// * py: a GIL handle, needed to negate rotation parameters in Python space.
+/// * gates: the sequence of Rustiq's Clifford gates returned by Rustiq's
+///     pauli network synthesis algorithm.
+/// * paulis: Rustiq's data structure storing the pauli rotations.
+/// * angles: Qiskit's rotation angles corresponding to the pauli rotations.
+fn inject_rotations_unordered(
+    py: Python,
+    gates: &Vec<CliffordGate>,
```
Cryoris: From what the internet tells me "using &Vec over &[..] implies an extra pointer re-direction" and I remember clippy telling me I should use the slice over a vector -- but I'm fine as is if clippy is fine 🙂

crates/accelerate/src/synthesis/evolution/mod.rs
```diff
@@ -0,0 +1,439 @@
...
+/// The rotations are assumed to be ordered (up to commutativity).
+///
+/// # Arguments
+///
+/// * py: a GIL handle, needed to negate rotation parameters in Python space.
+/// * gates: the sequence of Rustiq's Clifford gates returned by Rustiq's
+///     pauli network synthesis algorithm.
+/// * paulis: Rustiq's data structure storing the pauli rotations.
+/// * angles: Qiskit's rotation angles corresponding to the pauli rotations.
+fn inject_rotations_ordered(
```
Cryoris: Does this only differ from the `unordered` version in the few lines that create/modify the commutativity dag? If yes, would it make sense to merge them and add a keyword `ordered: bool` to avoid code duplication?

crates/accelerate/src/synthesis/evolution/mod.rs
```diff
@@ -0,0 +1,439 @@
...
+    let mut cur_paulis = paulis.clone();
+    let mut hit_paulis: Vec<bool> = vec![false; cur_paulis.len()];
+
+    // check which paulis are hit at the very start
+    for i in 0..cur_paulis.len() {
+        let pauli_support_size = cur_paulis.support_size(i);
+        if pauli_support_size == 0 {
+            // in case of an all-identity rotation, update global phase by subtracting
+            // the angle
+            global_phase = radd_param(global_phase, multiply_param(&angles[i], -1.0, py), py);
```
Cryoris: It might be worth it to only do the multiplication by -1 after the for loop since it's not for free (but here it likely doesn't matter much since you're only dealing with `Param::Float` and no `Param::ParameterExpression` which are very expensive to multiply)

crates/accelerate/src/synthesis/evolution/mod.rs
```diff
@@ -0,0 +1,439 @@
...
+            }
+        }
+    }
+
+    (out_gates, global_phase)
+}
+
+/// Return the vector of Qiskit's gate corresponding to the given vector
+/// of Rustiq's Clifford gate.
+fn qiskit_clifford_gates(gates: &Vec<CliffordGate>) -> Vec<QiskitGate> {
```
Cryoris: Just a suggestion, since this is a transformation (I sort of expected this to return all Qiskit Clifford gates) 
```suggestion
fn to_qiskit_clifford_gates(gates: &Vec<CliffordGate>) -> Vec<QiskitGate> {
```

crates/accelerate/src/synthesis/evolution/mod.rs
```diff
@@ -0,0 +1,439 @@
...
+///     resynthesized using Qiskit's greedy Clifford synthesis algorithm. If `2`, it
+///     is resynthesized by Rustiq itself. If `upto_phase` is `false`, the naive
+///     approach is used, as neither synthesis method preserves the global phase.
+///
+/// If `preserve_order` is `true` and both `upto_clifford` and `upto_phase` are `false`,
+/// the returned circuit is equivalent to the given pauli network.
+#[pyfunction]
+#[pyo3(signature = (num_qubits, pauli_network, optimize_count=true, preserve_order=true, upto_clifford=false, upto_phase=false, resynth_clifford_method=1))]
+pub fn pauli_network_synthesis(
+    py: Python,
```
Cryoris, line 55: If you want you could skip this argument and just get the GIL handle from the bound, 
```rust
let py = pauli_network.py();
```
instead 🙂

crates/accelerate/src/synthesis/evolution/mod.rs
```diff
@@ -0,0 +1,439 @@
...
+///     the 2-qubit gate count; and if `false`, then the 2-qubit depth.
+/// * preserve_order: whether the order of paulis should be preserved, up to
+///     commutativity. If the order is not preserved, the returned circuit will
+///     generally not be equivalent to the given pauli network.
+/// * upto_clifford: if `true`, the final Clifford operator is not synthesized
+///     and the returned circuit will generally not be equivalent to the given
+///     pauli network. In addition, the argument `upto_phase` would be ignored.
+/// * upto_phase: if `true`, the global phase of the returned circuit may differ
+///     from the global phase of the given pauli network. The argument is ignored
+///     when `upto_clifford` is `true`.
```
Cryoris:
```suggestion
///     from the global phase of the given pauli network. The argument is set
///     to `true` when `upto_clifford` is `true`.
```

crates/accelerate/src/synthesis/evolution/mod.rs
```diff
@@ -0,0 +1,439 @@
...
+/// * resynth_clifford_method: describes the strategy to synthesize the final
```
Cryoris, line 42: It would be nice for this to be an enum, whose names could reflect the method. See the following link for a pyo3 enum we can expose to Python space: https://stackoverflow.com/questions/76322128/pyo3-how-to-return-enums-to-python-module for an example.

crates/accelerate/src/synthesis/evolution/mod.rs
```diff
@@ -0,0 +1,439 @@
...
+        let tuple = item.downcast::<PyTuple>()?.borrow();
+
+        let sparse_pauli: String = tuple.get_item(0)?.downcast::<PyString>()?.extract()?;
+        let angle: Param = tuple.get_item(2)?.extract()?;
+        let qubits = tuple.get_item(1)?;
+        let qubits: Vec<u32> = qubits
+            .downcast::<PyList>()?
+            .iter()
+            .map(|v| v.extract())
+            .collect::<PyResult<_>>()?;
```
Cryoris: It should be possible to directly extract a Vec, something like
```suggestion
        let qubits: Vec<u32> = tuple.get_item(1)?.extract()?;
```

crates/accelerate/src/synthesis/evolution/mod.rs
```diff
@@ -0,0 +1,439 @@
...
+    py: Python,
+    num_qubits: usize,
+    pauli_network: &Bound<PyList>,
+    optimize_count: bool,
+    preserve_order: bool,
+    upto_clifford: bool,
+    upto_phase: bool,
+    resynth_clifford_method: usize,
+) -> PyResult<CircuitData> {
+    let mut paulis: Vec<String> = Vec::new();
```
Cryoris: Could we use `Vec::with_capacity(pauli_network.len())` for both vectors here?

qiskit/synthesis/evolution/pauli_network.py
```diff
@@ -0,0 +1,80 @@
...
+            is resynthesized by Rustiq itself. If `upto_phase` is `False`, the naive
+            approach is used, as neither synthesis method preserves the global phase.
+
+    Returns:
+        A circuit implementation of the pauli network.
+
+    References:
+        1. Timothée Goubault de Brugière and Simon Martiel,
+           *Faster and shorter synthesis of Hamiltonian simulation circuits*,
+           `arXiv:2404.03280 [quant-ph] <https://arxiv.org/abs/2404.03280>`_
```
Cryoris, line 67: I think we typically don't include the category 🙂 
```suggestion
           `arXiv:2404.03280 <https://arxiv.org/abs/2404.03280>`_
```

qiskit/synthesis/evolution/pauli_network.py
```diff
@@ -0,0 +1,80 @@
...
+    out = pauli_network_synthesis_inner(
+        num_qubits=num_qubits,
+        pauli_network=pauli_network,
+        optimize_count=optimize_count,
+        preserve_order=preserve_order,
+        upto_clifford=upto_clifford,
+        upto_phase=upto_phase,
+        resynth_clifford_method=resynth_clifford_method,
+    )
+    circuit = QuantumCircuit._from_circuit_data(out)
```
Cryoris: It might be better to add the registers for consistency with the other Pauli evolution plugin 🙂 
```suggestion
    circuit = QuantumCircuit._from_circuit_data(out, add_regs=True)
```

qiskit/transpiler/passes/synthesis/hls_plugins.py
```diff
@@ -262,15 +262,46 @@
...
+    :header-rows: 1
+
+    * - Plugin name
+      - Plugin class
+      - Targeted connectivity
+    * - ``"rustiq"``
+      - :class:`~.PauliEvolutionSynthesisRustiq`
+      - all-to-all
+    * - ``"default"``
+      - :class:`~.PauliEvolutionSynthesisDefault`
```
Cryoris: Just a note: you could skip these as they just truncate the path (but there is no path)
```suggestion
      - :class:`.PauliEvolutionSynthesisRustiq`
      - all-to-all
    * - ``"default"``
      - :class:`.PauliEvolutionSynthesisDefault`
```

qiskit/transpiler/passes/synthesis/hls_plugins.py
```diff
@@ -1047,8 +1079,88 @@ def run(self, high_level_object, coupling_map=None, target=None, qubits=None, **
 
 
 class PauliEvolutionSynthesisDefault(HighLevelSynthesisPlugin):
-    """The default implementation calling the attached synthesis algorithm."""
+    """Synthesize a PauliEvolutionGate using the default synthesis algorithm.
```
Cryoris: Would it be nice to add the links (also below and in the other synthesis plugin)?
```suggestion
    """Synthesize a :class:`.PauliEvolutionGate` using the default synthesis algorithm.
```

qiskit/transpiler/passes/synthesis/hls_plugins.py
```diff
@@ -1047,8 +1079,88 @@ def run(self, high_level_object, coupling_map=None, target=None, qubits=None, **
...
+
+    def run(self, high_level_object, coupling_map=None, target=None, qubits=None, **options):
+        if not isinstance(high_level_object, PauliEvolutionGate):
+            # Don't do anything if a gate is called "evolution" but is not an
+            # actual PauliEvolutionGate
+            return None
+
+        algo = high_level_object.synthesis
+
+        if not hasattr(algo, "expand"):
```
Cryoris: Maybe this is safer?
```suggestion
        if not isinstance(algo, ProductFormula):
```

test/python/transpiler/test_high_level_synthesis.py
```diff
@@ -2607,5 +2607,57 @@ def test_annotated_mcx(self):
...
+    """Tests related to plugins for PauliEvolutionGate."""
+
+    def test_supported_names(self):
+        """Test that "default" and "rustiq" plugins do exist."""
+        supported_plugin_names = high_level_synthesis_plugin_names("PauliEvolution")
+        self.assertIn("default", supported_plugin_names)
+        self.assertIn("rustiq", supported_plugin_names)
+
+    @data("default", "rustiq")
+    def test_correctness(self, plugin_name):
```
Cryoris, line 2622: To check correctness we'd have to compare `qct` to the matrix-based evolution of the Trotter steps, see https://github.com/Qiskit/qiskit/blob/2efb59ce4cd3c2d8721390648cdca6d2fd41f792/test/python/circuit/library/test_evolution_gate.py#L42. But since that's already tested for the default decomposition, we could just check whether the `rustiq` plugin equals the `default`.

test/python/transpiler/test_high_level_synthesis.py
```diff
@@ -2607,5 +2607,57 @@ def test_annotated_mcx(self):
...
+            hls_pass = HighLevelSynthesis(hls_config=hls_config)
+            qct = hls_pass(qc)
+            cnt_ops = qct.count_ops()
+            self.assertEqual(cnt_ops["cx"], 10)
+        with self.subTest("upto_phase"):
+            hls_config = HLSConfig(PauliEvolution=[("rustiq", {"upto_phase": True})])
+            hls_pass = HighLevelSynthesis(hls_config=hls_config)
+            qct = hls_pass(qc)
+            cnt_ops = qct.count_ops()
+            self.assertEqual(cnt_ops["cx"], 9)
```
Cryoris, line 2671: Could we check that the decomposition is correct for the `upto_phase` decomposition? `Operator.equiv` should be the right method 🙂

test/python/transpiler/test_high_level_synthesis.py
```diff
@@ -2607,5 +2607,57 @@ def test_annotated_mcx(self):
...
+    def test_supported_names(self):
+        """Test that "default" and "rustiq" plugins do exist."""
+        supported_plugin_names = high_level_synthesis_plugin_names("PauliEvolution")
+        self.assertIn("default", supported_plugin_names)
+        self.assertIn("rustiq", supported_plugin_names)
+
+    @data("default", "rustiq")
+    def test_correctness(self, plugin_name):
+        """Test that all plugins return a correct Operator"""
+        op = SparsePauliOp(["XXX", "YYY", "IZZ", "XZY"], coeffs=[1, 2, 3, 4])
```
Cryoris: Could we add one test for the global phase (so with the `II...`) pauli?

Cryoris: Could we also add a check for `preserve_order=False`?

Cryoris (commented):
Looks pretty good to me! I've left a bunch of comments but they are all very small 🙂

crates/accelerate/src/synthesis/clifford/utils.rs
alexanderivrii: In general, the Rust-based Clifford code is implicitly tested when checking the correctness of synthesis results in `test_high_level_synthesis.py`, etc.. However, this is not a full-proof method since this only considers gates produced by the synthesis algorithms. This is why we did not need handling `SX` and `SXsg` gates before this PR, none of our current Clifford synthesis algorithms produced such gates. 

The best way I can think of how to apply your suggestion is to add Rust-based tests that we would then execute with `cargo test`. However, so far we have been mostly avoiding this path. Note that our internal Rust-based Clifford code is not exposed via a Python interface, this is why I don't see how to write directed tests from the Python space. Would love to hear additional suggestions.

crates/accelerate/src/synthesis/evolution/mod.rs
alexanderivrii: I though that the module-level except was the preferred way since this is how we have it in `dense_layout.rs` and many other files. But, I agree, for a single function the function-level except makes more sense. Done (the commit will soon follow).

crates/accelerate/src/synthesis/evolution/mod.rs
alexanderivrii: Done (the commit will soon follow).

crates/accelerate/src/synthesis/evolution/mod.rs
alexanderivrii: Ah, in my code I was passing the ownership to `qubits` when calling `expand_pauli`, which compiled since `qubits` were not used after that. But agree that passing this as a slice is a far better API. Done (the commit will soon follow).

crates/accelerate/src/synthesis/evolution/mod.rs
alexanderivrii: This works! I also don't know how expensive `chars().nth()` really is (after all, both approaches should just boil down to a single memory access), but your suggestion is visually better. Changed (the commit will soon follow).

crates/accelerate/src/synthesis/evolution/mod.rs
alexanderivrii: Changed (the commit will soon follow).

crates/accelerate/src/synthesis/evolution/mod.rs
alexanderivrii: Changed (the commit will soon follow).

crates/accelerate/src/synthesis/evolution/mod.rs
alexanderivrii: Yes (and I double-checked rustiq-core's `PauliSet::get` function to confirm).

crates/accelerate/src/synthesis/evolution/mod.rs
alexanderivrii: Hmm, the suggested made perfect sense to me, and I am sure I've used something similar in the past, but in fact this does not compile with the error that `dag.add_node` is a method and not a field. I guess if the function `add_node` was not a part of the DAG class, this would have worked.

crates/accelerate/src/synthesis/evolution/mod.rs
alexanderivrii: I am not sure whether `&[..]` requires less dereferencing than `&Vec<..>`, however it definitely works for a larger number of structures. Now that you mentioned it, I would have expected clippy to complain, I have seen similar complains in the past. Done (the commit will soon follow).

crates/accelerate/src/synthesis/evolution/mod.rs
alexanderivrii: The implementation that you have reviewed followed the internal rustiq repo (the python wrapper around the rustiq-core code), which has two different functions for these tasks. 

But I completely agree that we can easily combine these two functions: in the unordered case we just create a "dummy" DAG without any edges. I believe the extra cost of checking the number of predecessors of a node for an edgeless DAG is completely negligible. 

Done (the commit will soon follow).

crates/accelerate/src/synthesis/evolution/mod.rs
alexanderivrii: Hmm, the `synth_pauli_network_rustiq` function (and the underlying Rust code) work with parameterized angles. I have added a test for that. 

However, it does seem that our plugin interface for `PauliEvolutionGate`s does not accept parameterized angles. Why is this the case?

In the extremely unlikely case that we have a pauli network with multiple parametric all-identity gates, say with angles `a`, `b`, `c`, the suggestion is to replace the computation of the global phase from `0 + (-1) * a + (-1) * b + (-1) *c` to `0 + (-1) * (a + b + c)`?

Why are `ParameterExpression` more expensive to multiply than to add?

crates/accelerate/src/synthesis/evolution/mod.rs
alexanderivrii: Done (the commit will soon follow)

crates/accelerate/src/synthesis/evolution/mod.rs
alexanderivrii: nice! (but unless it's faster, let me leave this as is)

crates/accelerate/src/synthesis/evolution/mod.rs
alexanderivrii: Changed the docstring based on this suggestion. (The commit will soon follow).

crates/accelerate/src/synthesis/evolution/mod.rs
Cryoris: > In the extremely unlikely case that we have a pauli network with multiple parametric all-identity gates, say with angles a, b, c, the suggestion is to replace the computation of the global phase from 0 + (-1) * a + (-1) * b + (-1) *c to 0 + (-1) * (a + b + c)?

Yes 🙂 

> Why are ParameterExpression more expensive to multiply than to add?

Both are expensive, so we should do it as little as possible. The multiplication by -1 is something we can optimize 🙂 

> However, it does seem that our plugin interface for PauliEvolutionGates does not accept parameterized angles. Why is this the case?

Hmm it should, maybe the type hint is wrong

crates/accelerate/src/synthesis/evolution/mod.rs
alexanderivrii: While I completely agree that using enums is a great idea, here specifically I am a bit reluctant to do this, as this is an option I don't expect 99% of users to ever experiment with (and that's 99% of the users who will want to experiment with Rustiq in the first place). I have considered leaving this option out, but I myself needed it to gauge the expected performance of switching to extended Clifford simulator. 

On a related note, I do think that using enum for specifying the optimization criterion (depth, count, fidelity, or something else) is also a good idea, but I don't think it makes sense to do this just for pauli network synthesis using Rustiq. Personally, I would like to see this as an additional argument for `transpile`.

crates/accelerate/src/synthesis/evolution/mod.rs
alexanderivrii: Wow! Just wow! After all the effort I got that to work, here you effortlessly come up with a beautiful one-liner. Fixed. The commit will appear soon.

crates/accelerate/src/synthesis/evolution/mod.rs
alexanderivrii: Done.

qiskit/synthesis/evolution/pauli_network.py
alexanderivrii: I think we do, or at least we do both (`grep` found a huge number of citations that explicitly mention `[quant-ph]`).

qiskit/synthesis/evolution/pauli_network.py
alexanderivrii: done (the commit will appear soon)

qiskit/transpiler/passes/synthesis/hls_plugins.py
alexanderivrii: I actually know this one, but for some reason all of our plugin tables have the "~", including the one you have recently added for MCMT. :)

qiskit/transpiler/passes/synthesis/hls_plugins.py
alexanderivrii: I have changed this for the two PauliEvolutionGate plugins. It would not be a lot of work to do this for all the other plugins as well, but possibly best after the release.

qiskit/transpiler/passes/synthesis/hls_plugins.py
alexanderivrii: I like this, this is indeed nicer. Done.

test/python/transpiler/test_high_level_synthesis.py
alexanderivrii: I believe that this is precisely what gets done internally when calculating `Operator(qc)`.

test/python/transpiler/test_high_level_synthesis.py
alexanderivrii: Done.

test/python/transpiler/test_high_level_synthesis.py
alexanderivrii: I don't know how to add a `correctness` test: since with `preserve_order=False` the operator might be completely different, but I have added a test that checks that the argument is applied.

alexanderivrii: done

alexanderivrii:
Thanks @Cryoris for a thorough review! I think I have addressed all of your review comments (and in a few cases explained why I don't want to address it). To keep track of your comments, I was marking them as resolved as I was going through them, so you will probably want to unresolve them to see if a deeper discussion is required.

crates/accelerate/src/synthesis/evolution/mod.rs
Cryoris: Let's leave it for a follow up then 👍🏻

crates/accelerate/src/synthesis/evolution/mod.rs
Cryoris: I had the exact same process with @raynelfss, so all the praise goes to him 😛

qiskit/transpiler/passes/synthesis/hls_plugins.py
Cryoris: Oh then I was too quick to copy+paste 😂 Fair enough, let's clean this up some time later

Cryoris (approved):
LGTM thanks for the updates!

Issue #12966, opened 2024-08-16
Add rust representation for non-gate standard circuit operations


mtreinish:
### What should we add?

Now that we have a rust native representation for circuit operations we can build an work with circuits and gates directly in rust. However right now the only rust native operations defined are the standard gate library. These are the bulk of the cirucit operations used in Qiskit so there is a large gain from starting with just this set. But as we move more logic intro rust having a more complete representation of the standard circuit operations will be necessary.

Right now in rust there are 4 categories of object, `StandardGate`, `PyGate`, `PyInstruction`, and `PyOperation`. I think it would be best to add a new field `StandardInstruction` (or `StandardOperation` operation is a more descriptive term, but these will be analgous to the `Instruction` subclasses in Python) which we can use to add rust native representations of standard non-gate operations. This would house operations like `Measure` and `Reset` which aren't unitary operations but are similarly static.

The other option is to rename `StandardGate` to `StandardOperations` and add these operations to it. Personally I think there it's advantageous to leverage the type system to differentiate unitary gates from other operations like this (similar to how we use `isinstance(op, Gate)` in python). But I don't feel super strongly about it. There is also a tradeoff because we can only support so many variants in `PackedOperation` before we run out of space in the type so we do have to be careful about proliferating types there.

mtreinish:
My other thought here is when thinking about types like `UnitaryGate` and maybe other operation types that we want to move to rust like that in the future is having a `Box<dyn Operation>` variant in `PackedOperation`. It gives us the flexibility to add any type without worrying about extending `PackedOperation` explicitly, and also opens up the possibility of custom operations defined in rust or otherwise outside of Python. However it does give up having static typing for everything we put in the circuit and we'd have to rely on the operation trait (and likely the name field) to differentiate types when an operation could be this box type.

I don't think if we added this it would preclude adding a fixed type for non-gate standard operations or any other core operation type, but since this issue is where we're talking about extending the data model first I figured I'd raise it here.

PR #13310, opened 2024-10-11
Oxidize `NLocal` & family 

Cryoris:
### Summary

Part of #13046: create the `n_local` function and port the internals to Rust. This will include other parts of the `NLocal` family, such as `TwoLocal`, `EfficientSU2`, `RealAmplitudes` etc. `EvolvedOperatorAnsatz` is potentially a bit more involved and might be done in a follow-up.

### Details and comments

The new code is particularly fast when dealing with gates we already have in Rust and that don't need any special re-parameterization. In this case:
<details closed>
<summary>the new code is ~22x faster for 100 qubits, full entanglement and 10 reps</summary>

```
num qubits: 10, speedup: 9.075
main: 0.005 +- 0.004
this: 0.001 +- 0.000

num qubits: 50, speedup: 18.138
main: 0.063 +- 0.014
this: 0.003 +- 0.003

num qubits: 100, speedup: 22.011
main: 0.251 +- 0.034
this: 0.011 +- 0.007

num qubits: 200, speedup: 26.829
main: 0.957 +- 0.034
this: 0.036 +- 0.010
```
</details>

But if custom gates or special parameterizations are used (e.g. `RXGate(x + 3 * y)`) we need to go back to Python space to correctly handle the parameter logic. This slows down the code and in this case:

<details closed>
<summary>the new code is ~3.5x faster for 100 qubits, linear entanglement (full took too long) and 10 reps</summary>

```
num qubits: 10, speedup: 4.594
main: 0.027 +- 0.006
this: 0.006 +- 0.000

num qubits: 50, speedup: 3.947
main: 0.162 +- 0.006
this: 0.041 +- 0.011

num qubits: 100, speedup: 3.666
main: 0.366 +- 0.031
this: 0.100 +- 0.019
```
</details>

### To do
- [x] Improve docstrings
- [x] Make the `entanglement` handling rust-side a bit better to understand instead of a 4-times nested vec
- [ ] ~~Use `n_local` in `NLocal` (i.e. fast path for Python code)~~ This might not be possible since `NLocal` works with `QuantumCircuit`s as blocks, but `n_local` needs gates. Functionally it'd be the same but the circuits are nested differently, which is not backward compatible. We could add a fast-path in case the inputs are not circuits but gates.
- [x] Add the family members, like `efficient_su2` etc.


crates/accelerate/src/circuit_library/mod.rs
```diff
@@ -13,10 +13,15 @@
...
 
 mod entanglement;
+mod n_local;
 mod pauli_feature_map;
 mod quantum_volume;
 
 pub fn circuit_library(m: &Bound<PyModule>) -> PyResult<()> {
+    let nlocal_mod = PyModule::new_bound(m.py(), "n_local")?;
+    n_local::n_local(&nlocal_mod)?;
+    m.add_submodule(&nlocal_mod)?;
```
mtreinish: So this is adding a submodule to the library submodule, so it'll be `qiskit._accelerate.circuit_library.n_local.n_local`. Is there a reason we need the extra level of nesting here?

crates/accelerate/src/circuit_library/n_local/blocks.rs
```diff
@@ -0,0 +1,171 @@
...
+
+                let job = builder.call1(py, (py_params,))?;
+                let result = job.downcast_bound::<PyTuple>(py)?;
+
+                let operation: OperationFromPython = result.get_item(0)?.extract()?;
+                let bound_params = result
+                    .get_item(1)?
+                    .iter()?
+                    .map(|ob| Param::extract_no_coerce(&ob?))
+                    .collect::<PyResult<Vec<Param>>>()?;
```
mtreinish: Is there a reason to just collect into the smallvec here instead of using `from_vec` on the return?

crates/accelerate/src/circuit_library/n_local/blocks.rs
```diff
@@ -0,0 +1,171 @@
...
+
+type BlockEntanglement = Vec<Vec<u32>>; // vector of connections for a block (e.g. CX)
+pub(super) type LayerEntanglement = Vec<BlockEntanglement>; // entanglement of all blocks
+
+/// Represent the entanglement in an n-local circuit.
+///
+/// This eagerly expands the full entanglement for every layer.
+/// This could be done more efficiently, e.g., by creating entanglement objects that store
+/// their underlying representation (e.g. a string or a list of connections) and returning
+/// these when given a layer-index.
```
mtreinish: I assume there is a reason you didn't do this? It's a bit weird to call this out in a docstring on a public struct like this.

crates/accelerate/src/circuit_library/n_local/blocks.rs
```diff
@@ -0,0 +1,171 @@
...
+pub struct Entanglement {
+    entanglement_vec: Vec<LayerEntanglement>,
```
mtreinish: So it's `Vec<Vec<Vec<u32>>`?

crates/accelerate/src/circuit_library/n_local/multi_local.rs
```diff
@@ -0,0 +1,289 @@
...
+///     rotation_blocks: A reference to a vector containing the instructions to insert.
+///         This is a vector (sind we can have multiple rotations  operations per layer), with
+///         3-tuple elements containing (packed_operation, num_qubits, num_params).
+///     parameters: The set of parameter objects to use for the operations. This is a 3x nested
+///         vector, organized as operation -> block -> param. That means that for operation ``i``
+///         and block ``j``, the parameters are given by ``parameters[i][j]``.
+///     skipped_qubits: A hash-set containing which qubits are skipped in the rotation layer.
+///
+/// Returns:
+///     An iterator for the rotation instructions.
```
mtreinish: Heh, some old habits die hard. This is the google style python doc string format, but this is a rust doc since it's not exposed to python, and this formatting won't render well,. It might be better to use the rustdoc style instead: https://doc.rust-lang.org/rustdoc/how-to-write-documentation.html  this applies to a bunch of the functions here.

crates/accelerate/src/circuit_library/n_local/multi_local.rs
```diff
@@ -0,0 +1,289 @@
...
+    CircuitData::from_packed_operations(py, num_qubits, 0, packed_insts, Param::Float(0.0))
+}
+
+#[pyfunction]
+#[pyo3(signature = (num_qubits, reps, rotation_blocks, entanglement_blocks, entanglement, insert_barriers, skip_final_rotation_layer, skip_unentangled_qubits, parameter_prefix))]
+#[allow(clippy::too_many_arguments)]
+pub fn py_n_local(
+    py: Python,
+    num_qubits: i64,
+    reps: i64,
```
mtreinish: Can these be negative? It seems weird to use a i64 when everything else is using unsigned types for these.

crates/accelerate/src/circuit_library/n_local/multi_local.rs
```diff
@@ -0,0 +1,289 @@
...
+                    bound_params,
+                    indices.iter().map(|i| Qubit(*i)).collect(),
+                    vec![] as Vec<Clbit>,
+                ))
+            })
+    })
+}
+
+#[allow(clippy::too_many_arguments)]
+pub fn n_local(
```
mtreinish: Do you want a docstring on this, or are you assuming nobody will call this from rust? It is public so I assume you're thinking we'll want to call this from rust.

crates/accelerate/src/circuit_library/n_local/multi_local.rs
```diff
@@ -0,0 +1,289 @@
...
+struct MaybeBarrier {
+    barrier: Option<Instruction>,
+}
+
+impl MaybeBarrier {
+    fn new(py: Python, num_qubits: u32, insert_barriers: bool) -> PyResult<Self> {
+        if !insert_barriers {
+            Ok(Self { barrier: None })
+        } else {
+            let barrier_cls = imports::BARRIER.get_bound(py);
```
mtreinish: I'm looking forward to having a solution for #12966 so we can do this without needing python.

qiskit/circuit/library/n_local/efficient_su2.py
```diff
@@ -78,8 +181,17 @@ class EfficientSU2(TwoLocal):
...
+    .. seealso::
+
+        The :func:`.efficient_su2` function constructs the functionally same circuit, but faster.
+
     """
 
+    @deprecate_func(
+        since="1.3",
+        additional_msg="Use the function qiskit.circuit.library.efficient_su2 instead.",
+        pending=True,
```
mtreinish, line 194: Did you and @ElePT reach a conclusion on whether you were going to continue to do this?

qiskit/circuit/library/n_local/excitation_preserving.py
```diff
@@ -57,7 +170,7 @@ class ExcitationPreserving(TwoLocal):
     Examples:
 
         >>> ansatz = ExcitationPreserving(3, reps=1, insert_barriers=True, entanglement='linear')
-        >>> print(ansatz)  # show the circuit
+        >>> print(ansatz.decompose())  # show the circuit
```
mtreinish, line 173: Did you want to use `flatten` here instead so you don't need to add the `decompose()` call

qiskit/circuit/library/n_local/n_local.py
```diff
@@ -76,8 +294,18 @@ class NLocal(BlueprintCircuit):
 
     If specified, barriers can be inserted in between every block.
     If an initial state object is provided, it is added in front of the NLocal.
+
+    .. seealso::
+
+        The :func:`.n_local` function constructs the functionally same circuit, but faster.
```
mtreinish:
```suggestion
        The :func:`.n_local` function constructs a functionally equivalent circuit, but faster.
```

crates/accelerate/src/circuit_library/mod.rs
Cryoris: I thought there would be more submodules, but it ended up being very few. I can remove this level of nesting 🙂

crates/accelerate/src/circuit_library/n_local/blocks.rs
Cryoris: I didn't know that was possible 😬

crates/accelerate/src/circuit_library/n_local/blocks.rs
Cryoris: Mainly a time constraint and that it is already very performant as is -- but yeah a public docstring is weird, I'll move it into a TODO comment 😄

crates/accelerate/src/circuit_library/n_local/blocks.rs
Cryoris: Correct, we can think of the structure as
```
Connection: Vec<u32>  // indices that the multi-qubit gate acts on
BlockEntanglement: Vec<Connection>  // entanglement for single block
LayerEntanglement: Vec<BlockEntanglement>  // entanglements for all blocks in the layer
Entanglement: Vec<LayerEntanglement>  // entanglement for every layer
```
Maybe this would be nice to add as explanatory comment?

crates/accelerate/src/circuit_library/n_local/multi_local.rs
Cryoris: whoops, yes, I'll update!

crates/accelerate/src/circuit_library/n_local/multi_local.rs
Cryoris: No they can't, this should be unsigned -- I was just using the old habit from when I was using whatever type worked 😉

qiskit/circuit/library/n_local/excitation_preserving.py
Cryoris: I'll add an example with flatten, but would also like to show the decompose flow 😄

qiskit/circuit/library/n_local/efficient_su2.py
Cryoris: Not yet, we're currently deciding whether to play rock paper scissors for it or have a vote

qiskit/circuit/library/n_local/n_local.py
Cryoris: Done in 5fc47eec3

crates/accelerate/src/circuit_library/n_local/multi_local.rs
Cryoris: Added in 5fc47eec3

crates/accelerate/src/circuit_library/n_local/multi_local.rs
Cryoris: Changed in 5fc47eec3

crates/accelerate/src/circuit_library/n_local/multi_local.rs
Cryoris: Updated in 5fc47eec3

crates/accelerate/src/circuit_library/n_local/blocks.rs
Cryoris: Done in 5fc47eec3

crates/accelerate/src/circuit_library/n_local/blocks.rs
Cryoris: Updated in 5fc47eec3

crates/accelerate/src/circuit_library/mod.rs
Cryoris: Removed in 5fc47eec3

qiskit/circuit/library/n_local/efficient_su2.py
alexanderivrii: This is such an outdated game, you should play the "rock paper scissors lizard spock" instead.

crates/accelerate/src/circuit_library/n_local/blocks.rs
mtreinish: A comment wouldn't hurt. I assume you're using the type aliases here to make it simpler. It would be easier for me personally to understand of `Vec<u32>` `Vec<Vec<u32>>`, etc rather than `Connection`, `BlockEntanglement`, etc. But I guess cargo doesn't like that. Either way if you want to add a comment that would be nice for people reading the code.

crates/accelerate/src/circuit_library/multi_local.rs
```diff
@@ -0,0 +1,317 @@
...
+///     entanglement for that layer.
+///     Defaults to ``"full"``, meaning an all-to-all entanglement structure.
+/// - `reps`: Specifies how often the rotation blocks and entanglement blocks are repeated.
+/// - `insert_barriers`: If ``True``, barriers are inserted in between each layer. If ``False``,
+///     no barriers are inserted.
+/// - `parameter_prefix`: The prefix used if default parameters are generated.
+/// - `skip_final_rotation_layer`: Whether a final rotation layer is added to the circuit.
+/// - `skip_unentangled_qubits`: If ``True``, the rotation gates act only on qubits that
+///     are entangled. If ``False``, the rotation gates act on all qubits.
+///
```
ShellyGarion, line 154: what about the `overwrite_block_parameters, inital_state, name, flatten` parameters?

qiskit/circuit/library/__init__.py
```diff
@@ -357,6 +357,11 @@
    :toctree: ../stubs/
    :template: autosummary/class_no_inherited_members.rst
 
+   n_local
+   efficient_su2
+   real_amplitudes
+   pauli_two_design
+   excitation_preserving
```
ShellyGarion, line 364: what about the classes `TwoLocal, EvolvedOperatorAnsatz` and `QAOAAnsatz` ? they won't be converted into functions?

qiskit/circuit/library/n_local/efficient_su2.py
```diff
@@ -14,18 +14,121 @@
...
+        BlockEntanglement
+        | Iterable[BlockEntanglement]
+        | Callable[[int], BlockEntanglement | Iterable[BlockEntanglement]]
+    ) = "reverse_linear",
+    reps: int = 3,
+    skip_unentangled_qubits: bool = False,
+    skip_final_rotation_layer: bool = False,
+    parameter_prefix: str = "θ",
+    insert_barriers: bool = False,
+    name: str = "EfficientSU2",
```
ShellyGarion, line 44: what about the parameters `inital_state, flatten` ?

qiskit/circuit/library/__init__.py
ShellyGarion: It will be confusing if some of them will remain classes and not functions, or do you plan to totally deprecate them?

qiskit/circuit/library/n_local/excitation_preserving.py
```diff
@@ -13,14 +13,127 @@
...
+        | Iterable[BlockEntanglement]
+        | Callable[[int], BlockEntanglement | Iterable[BlockEntanglement]]
+    ) = "full",
+    reps: int = 3,
+    skip_unentangled_qubits: bool = False,
+    skip_final_rotation_layer: bool = False,
+    parameter_prefix: str = "θ",
+    insert_barriers: bool = False,
+    name: str = "ExcitationPreserving",
+):
```
ShellyGarion, line 40: again, what about the parameters inital_state, flatten ?

qiskit/circuit/library/n_local/n_local.py
```diff
@@ -17,28 +17,248 @@
...
+        | Callable[[int], BlockEntanglement | Iterable[BlockEntanglement]]
+    ) = "full",
+    reps: int = 3,
+    insert_barriers: bool = False,
+    parameter_prefix: str = "θ",
+    overwrite_block_parameters: bool = True,
+    skip_final_rotation_layer: bool = False,
+    skip_unentangled_qubits: bool = False,
+    name: str | None = "nlocal",
+) -> QuantumCircuit:
```
ShellyGarion, line 70: again, what about the parameters inital_state, flatten ?

qiskit/circuit/library/n_local/real_amplitudes.py
```diff
@@ -13,15 +13,123 @@
...
+        | Iterable[BlockEntanglement]
+        | Callable[[int], BlockEntanglement | Iterable[BlockEntanglement]]
+    ) = "reverse_linear",
+    reps: int = 3,
+    skip_unentangled_qubits: bool = False,
+    skip_final_rotation_layer: bool = False,
+    parameter_prefix: str = "θ",
+    insert_barriers: bool = False,
+    name: str = "RealAmplitudes",
+) -> QuantumCircuit:
```
ShellyGarion, line 40: again, what about the parameters inital_state, flatten ?

ShellyGarion (commented):
LGTM. I have a few questions ?

qiskit/circuit/library/__init__.py
ShellyGarion: Note that the functions here appear in the table with the classes, and not after it (like they do in #13361)

crates/accelerate/src/circuit_library/multi_local.rs
mtreinish: These don't have analogs in the new function. The new function doesn't expose them because they aren't really necessary in the new form of returning a bare `QuantumCircuit` directly. If you want to something like prepend an initial state you can use `compose()` or the other circuit apis for that functionality.

qiskit/circuit/library/n_local/efficient_su2.py
mtreinish: Just like for nlocal I don't these are needed for the new function, `initial_state` can be done with `compose()` and flatten is always true for this function.

test/python/circuit/library/test_nlocal.py
```diff
@@ -49,6 +54,18 @@
 from test import QiskitTestCase  # pylint: disable=wrong-import-order
 
 
+class Gato(Gate):
```
mtreinish, line 59: :cat: ?

qiskit/circuit/library/n_local/excitation_preserving.py
mtreinish: Why though? My concern here is that doing `decompose()` here is the less efficient path so we're implicitly encouraging a pattern that will result in subpar performance. But we have the `seealso` now pointing people to the faster function so that's probably enough. It's not worth blocking over either way, so I'm fine if you don't want to change this.

crates/accelerate/src/circuit_library/blocks.rs
```diff
@@ -0,0 +1,167 @@
...
+impl BlockOperation {
+    pub fn assign_parameters(
+        &self,
+        py: Python,
+        params: &[&Param],
+    ) -> PyResult<(PackedOperation, SmallVec<[Param; 3]>)> {
+        match self {
+            Self::Standard { gate } => Ok((
+                (*gate).into(),
+                SmallVec::from_iter(params.iter().map(|&p| p.clone())),
```
mtreinish, line 41: Maybe?:
```suggestion
                SmallVec::from_iter(params.iter().cloned()),
```

crates/accelerate/src/circuit_library/multi_local.rs
```diff
@@ -0,0 +1,317 @@
...
+        entanglement_blocks,
+        skip_final_rotation_layer,
+        parameter_prefix,
+    )?;
+
+    // Compute the qubits that are skipped in the rotation layer. If this is set,
+    // we skip qubits that do not appear in any of the entanglement layers.
+    let skipped_qubits = if skip_unentangled_qubits {
+        let active: HashSet<&u32> =
+            HashSet::from_iter(entanglement.iter().flatten().flatten().flatten());
```
mtreinish, line 188: Hah, are you sure you want to flatten it?

test/python/circuit/library/test_nlocal.py
```diff
@@ -470,6 +487,287 @@ def test_initial_state_as_circuit_object(self):
...
+
+        self.assertEqual(expected, circuit)
+
+    def test_gate_blocks(self):
+        """Test passing blocks as gates."""
+        x, y = Parameter("x"), Parameter("y")
+        my_gate = Gato(x, y)
+
+        circuit = n_local(4, my_gate, "cx", "linear", reps=3)
+        print(circuit.draw())
```
mtreinish: Debug print

```suggestion
```

qiskit/circuit/library/n_local/n_local.py
```diff
@@ -1069,3 +1299,172 @@ def _stdlib_gate_from_simple_block(block: QuantumCircuit) -> _StdlibGateResult |
...
+    elif isinstance(entanglement, Iterable):
+        # handle edge cases when entanglement is set to an empty list
+        if len(entanglement) == 0:
+            return [[()]]
+
+        if isinstance(entanglement[0], str):
+            # if the entanglement is given as iterable we must make sure it matches
+            # the number of entanglement blocks
+            if len(entanglement) != num_entanglement_blocks:
+                print(entanglement)
```
mtreinish: Debug print

```suggestion
```

qiskit/circuit/library/n_local/n_local.py
```diff
@@ -1069,3 +1299,172 @@ def _stdlib_gate_from_simple_block(block: QuantumCircuit) -> _StdlibGateResult |
...
+                )
+
+            return entanglement
+
+        # normalize to list[BlockEntanglement]
+        if not isinstance(entanglement[0][0], Iterable):
+            entanglement = [entanglement]
+
+        if len(entanglement) != num_entanglement_blocks:
+            print(entanglement)
```
mtreinish:
```suggestion
```

crates/accelerate/src/circuit_library/n_local/blocks.rs
Cryoris: yeah I also found myself going back checking whether `BlockEntanglement` is now a 2 or 3 times nested Vec, but I think in the end the explanation plus the types might make it a bit easier to read 🙂

mtreinish (commented):
I think I'm mostly happy with this. There is a lot of code so I'm not 100% convinced I kept the full model of everything in my head, especially between the python and rust boundary, so I don't feel super convinced it's perfect. But I think it's probably good enough to merge and we can iterate on it in bugfix releases and anything else that comes up. Especially because this is a new function it's more important that the public interface is sustainable,  and I think it looks reasonable and maintainable as a subset of the options the existing classes provided. Besides the open question about two local and other functions from @ShellyGarion I left a few small inline comments.

The only other thing I think is there are some small test coverage gaps that coveralls is flagging which would be good to exercise those paths especially as they're typical conditions for edge cases in the input which would be good to enforce the behavior of.

crates/accelerate/src/circuit_library/blocks.rs
Cryoris: Ah nice I didn't know about this! But it doesn't like that because it looks like `from_iter` expects `&Param` and `.cloned` returns `Param` types 🤔

test/python/circuit/library/test_nlocal.py
Cryoris: Si, correcto 😛

test/python/circuit/library/test_nlocal.py
Cryoris: removed in 069f45c

qiskit/circuit/library/n_local/n_local.py
Cryoris: removed in 069f45c

qiskit/circuit/library/n_local/n_local.py
Cryoris: removed in 069f45c

Cryoris:
069f45c added tests for the flagged lines in Coveralls (for the new functions, not for the existing classes), so I think we should have the 100% on that 🙂

qiskit/circuit/library/n_local/efficient_su2.py
Cryoris: The (non-game-based) outcome was to keep it 🙂

qiskit/circuit/library/n_local/excitation_preserving.py
Cryoris: I thought just one example of `decompose` would be good so users know how to handle the "vanilla" format of the circuits, where one just gives the minimal amount of argument possible 😄 Except the first example, all others now use `flatten` 🙂

qiskit/circuit/library/__init__.py
Cryoris: QAOA and EvolvedOperatorAnsatz are added in #13361 and TwoLocal is redundant since it's 100% covered by n_local 🙂

crates/accelerate/src/circuit_library/blocks.rs
mtreinish: But doesn't `p.clone()` also return a `Param` instance too? It's not a big deal, I kind of wrote this assuming clippy would be grumpy about it.

crates/accelerate/src/circuit_library/multi_local.rs
```diff
@@ -0,0 +1,317 @@
...
+        packed_insts = Box::new(packed_insts.chain(rotation_layer(
+            py,
+            num_qubits,
+            rotation_blocks,
+            ledger.get_parameters(LayerType::Rotation, reps),
+            &skipped_qubits,
+        )))
+    }
+
+    CircuitData::from_packed_operations(py, num_qubits, 0, packed_insts, Param::Float(0.0))
```
mtreinish: You don't need a box here, you can do something like:

```suggestion
    let packed_insts = (0..reps).flat_map(|layer| {
        rotation_layer(
            py,
            num_qubits,
            rotation_blocks,
            ledger.get_parameters(LayerType::Rotation, layer),
            &skipped_qubits,
        )
        .chain(maybe_barrier.get())
        .chain(entanglement_layer(
            py,
            entanglement.get_layer(layer),
            entanglement_blocks,
            ledger.get_parameters(LayerType::Entangle, layer),
        ))
        .chain(maybe_barrier.get())
    });
    if !skip_final_rotation_layer {
        let packed_insts = packed_insts.chain(rotation_layer(
            py,
            num_qubits,
            rotation_blocks,
            ledger.get_parameters(LayerType::Rotation, reps),
            &skipped_qubits,
        ));
        CircuitData::from_packed_operations(py, num_qubits, 0, packed_insts, Param::Float(0.0))
    } else {
        CircuitData::from_packed_operations(py, num_qubits, 0, packed_insts, Param::Float(0.0))
    }
```

The trick is to avoid the dynamic type when you call `from_packed_operations`

qiskit/circuit/library/n_local/n_local.py
```diff
@@ -1069,3 +1299,170 @@ def _stdlib_gate_from_simple_block(block: QuantumCircuit) -> _StdlibGateResult |
...
+        | Iterable[BlockEntanglement]
+        | Callable[[int], BlockEntanglement | Iterable[BlockEntanglement]]
+    ),
+    num_entanglement_blocks: int,
+) -> list[str | list[tuple[int]]] | Callable[[int], list[str | list[tuple[int]]]]:
+    """If the entanglement is Iterable[Iterable], normalize to list[tuple]."""
+    if isinstance(entanglement, str):
+        return [entanglement] * num_entanglement_blocks
+
+    elif isinstance(entanglement, Iterable):
```
mtreinish: So checking `isinstance` `Iterable` isn't really the ideal way to check if something is actually iterable. It only checks if the class has been registered as an Iterable using the abc mechanisms (either as a subclass or manual registration). But python will let you iterate over some objects if `__getitem__` is implemented for sequences. The only surefire way to detect if something is actually iterable is to try running `iter()`.

mtreinish (commented):
Two small inline comments, that I caught through another scan through the code. They're just small but the python side one is a potential interface issue which could cause issue depending on how things are passed in. The rust comment is more just to avoid using `Box` when it's not needed. I think I'm ready to approve once these are addressed

qiskit/circuit/library/n_local/n_local.py
Cryoris: We can just turn this around and use `not callable(entanglement)` as check, then the only valid type would be the `Iterable` 🙂

mtreinish: It's a broader thing there are a few places where `isinstance(..., Iterable)` is called and they'd all suffer from this same issue.

mtreinish (approved):

PR #13311, opened 2024-10-11
Fix `QuantumCircuit.decompose` for high-level objects

Cryoris:
### Summary

Fixed some problems with `QuantumCircuit.decompose` when dealing with high-level objects (i.e. that can get synthesized by `HighLevelSynthesis`).

Blocks #13295 and #13301.

### Details and comments

Previously, running `decompose` would first run `HighLevelSynthesis` (in case no specific gates to decompose are set) and the `Decompose`. This would end up decomposing high-level objects twice instead of once. In addition, specifying high-level objects by name, could end up not decomposing them at all. 

For example
```
Circuit at hand:
   ┌──────────┐
q: ┤ Clifford ├
   └──────────┘
Expected result when decomposing:
   ┌───┐
q: ┤ H ├
   └───┘
.decompose(['clifford']) gives:
   ┌──────────┐
q: ┤ Clifford ├
   └──────────┘
.decompose() gives:
   ┌─────────┐
q: ┤ U2(0,π) ├
   └─────────┘
where's my H gate? :(
```

This is now fixed by pulling HLS into `Decompose`, which iterates over each node and only applies HLS if an operation does not have a definition.

### DIFF
--- a/qiskit/circuit/quantumcircuit.py
+++ b/qiskit/circuit/quantumcircuit.py
@@ -3229,40 +3229,33 @@ class QuantumCircuit:
 
     def decompose(
         self,
-        gates_to_decompose: Type[Gate] | Sequence[Type[Gate]] | Sequence[str] | str | None = None,
+        gates_to_decompose: (
+            str | Type[Instruction] | Sequence[str | Type[Instruction]] | None
+        ) = None,
         reps: int = 1,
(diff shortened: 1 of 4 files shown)


alexanderivrii (approved):
Thanks @Cryoris for making these fixes!

PR #13312, opened 2024-10-11
docs: clarify method purpose

### LINKED ISSUE #13051: Clarify docstring: Pauli-types `conjugate()` is complex conjugate, not Hermitian conjugate
### Environment

- **Qiskit version**: 1.2
- **Python version**: 3.11
- **Operating system**: macOS

### What is happening?

Not strictly a bug, just a doc clarification.

The docstring for the `conjugate` method of Pauli is:
```
Return the conjugate of each Pauli in the list.
```
and for SparsePauliOp is
```
Return the conjugate of the SparsePauliOp.
```

Personally, when I hear "conjugate of an operator" I think Hermitian conjugate. To avoid confusion with Hermitian conjugate, can we change the docstring of `conjugate` for Pauli, SparsePauliOp, etc to specify explicitly:
```
Return the complex conjugate of each Pauli in the list.
```
and
```
Return the complex conjugate of the SparsePauliOp.
```

(It's also weird there's a reference to a "list" in the docstring for a single Pauli, but I'm not concerned about that here).

n/a

### What should happen?

n/a

### Any suggestions?

_No response_

boonware:
### Summary
Fixes https://github.com/Qiskit/qiskit/issues/13051.

### Details and comments
Update the doc string for the Pauli operator method `conjugate` to clarify that it returns the complex conjugate and not the conjugate transpose.

### DIFF
--- a/qiskit/quantum_info/operators/symplectic/base_pauli.py
+++ b/qiskit/quantum_info/operators/symplectic/base_pauli.py
@@ -184,7 +184,7 @@ class BasePauli(BaseOperator, AdjointMixin, MultiplyMixin):
         return BasePauli(self._z, self._x, np.mod(self._phase + phase, 4))
 
     def conjugate(self):
-        """Return the complex conjugate of each Pauli in the list."""
+        """Return the complex conjugate of the Pauli with respect to the Z basis."""
         complex_phase = np.mod(self._phase, 2)
         if np.all(complex_phase == 0):
             return self
--- a/qiskit/quantum_info/operators/symplectic/base_pauli.py
+++ b/qiskit/quantum_info/operators/symplectic/base_pauli.py
@@ -184,7 +184,7 @@ class BasePauli(BaseOperator, AdjointMixin, MultiplyMixin):
         return BasePauli(self._z, self._x, np.mod(self._phase + phase, 4))
 
     def conjugate(self):
-        """Return the conjugate of each Pauli in the list."""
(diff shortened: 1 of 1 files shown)


jakelishman (commented):
Thanks for looking at this.

I think we'll maybe need a little bit more detail than this to properly describe the function. The complex conjugate of an arbitrary linear operator is only defined with regards to a particular matrix representation (unlike the adjoint, which is what this method was confused with), so I think we should highlight that here.  (For example, if we use the $Y$-eigenstate basis, then $Y$ is real-valued and $X$ and $Z$ are complex instead.)  It's sufficient to say that the conjugation is defined in terms of the matrix returned by `to_matrix`, or we can add a short comment that it's the matrix expressed in the $Z$-eigenstate basis.

boonware:
> Thanks for looking at this.
> 
> I think we'll maybe need a little bit more detail than this to properly describe the function. The complex conjugate of an arbitrary linear operator is only defined with regards to a particular matrix representation (unlike the adjoint, which is what this method was confused with), so I think we should highlight that here. (For example, if we use the Y -eigenstate basis, then Y is real-valued and X and Z are complex instead.) It's sufficient to say that the conjugation is defined in terms of the matrix returned by `to_matrix`, or we can add a short comment that it's the matrix expressed in the Z -eigenstate basis.

@jakelishman Thanks for the feedback. I have updated the comment.

jakelishman (approved):
Nice, thanks!

PR #13313, opened 2024-10-11
Use PyO3 0.22 and rust-numpy 0.22

mtreinish:
### Summary

This commit migrates to using PyO3 0.22 and rust-numpy and updates all the deprecated and changed interfaces in the libraries to their new syntax. No new or alternative interfaces are used as part of this PR except for where deprecation warnings pointed to do so.

One thing to note is that in pyo3 0.22 a new feature py-clone was added to re-enable `Py<T>::clone()`. This was disabled by default in pyo3 0.22 because of a soundness issue around calling that in situations when the GIL was not held. [1] Right now qiskit relies on using `Clone` on `Py` objects because we have container types that own a `PyObject` that we sometimes clone at the top level so this flag needs to be set. This only occurs in the `circuit` crate. In the future we can look at adding interfaces to avoid needing this flag in the future.

Another thing to note is that rust-numpy includes support for ndarray 0.16, however this PR doesn't migrate to it because `ndarray_einsum_beta` used in the `qiskit_accelerate::unitary_compose` module still depends on ndarray 0.15 and we can't upgrade until that does.

### Details and comments

[1] https://pyo3.rs/v0.22.3/migration#pyclone-is-now-gated-behind-the-py-clone-feature

### DIFF
--- a/crates/accelerate/src/commutation_checker.rs
+++ b/crates/accelerate/src/commutation_checker.rs
@@ -613,6 +613,7 @@ impl CommutationLibrary {
 #[pymethods]
 impl CommutationLibrary {
     #[new]
+    #[pyo3(signature=(py_any=None))]
     fn new(py_any: Option<Bound<PyAny>>) -> Self {
         match py_any {
             Some(pyob) => CommutationLibrary {
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1431,12 +1431,13 @@ checksum = "2b15c43186be67a4fd63bee50d0303afffcef381492ebe2c5d87f324e1b8815c"
 
 [[package]]
 name = "rowan"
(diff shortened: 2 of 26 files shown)


crates/accelerate/src/commutation_checker.rs
```diff
@@ -613,6 +613,7 @@ impl CommutationLibrary {
 #[pymethods]
 impl CommutationLibrary {
     #[new]
+    #[pyo3(signature=(py_any=None))]
```
kevinhartman, line 616: They finally got rid of the automatic optional stuff I see!

kevinhartman (approved):
Looks good to me!

Thanks for doing this. It's always neat to see how PyO3's interface evolves through these updates 🙂.

PR #13315, opened 2024-10-14
Fix `synth_mcmt_vchain` for gates with parameters

Cryoris:
### Summary

Fixed an issue in `synth_mcmt_vchain` (and thereby also in `MCMTVChain`) where the circuit data in Rust-space was constructed without the parameters of the base gate.

### Details

~~I'm not labelling this as stable backport as there's no 1.2.5 planned, right? (And I don't think we need a release just for this bug)~~ No reno as this bug has not yet been released.

### DIFF
--- a/test/python/circuit/library/test_mcmt.py
+++ b/test/python/circuit/library/test_mcmt.py
@@ -265,6 +265,13 @@ class TestMCMT(QiskitTestCase):
                 with self.assertRaises(ValueError):
                     _ = MCMTGate(gate, 10, 2)
 
+    def test_invalid_base_gate_width_synthfun(self):
+        """Test only 1-qubit base gates are accepted."""
+        for gate in [GlobalPhaseGate(0.2), SwapGate()]:
+            with self.subTest(gate=gate):
+                with self.assertRaises(ValueError):
+                    _ = synth_mcmt_vchain(gate, 10, 2)
+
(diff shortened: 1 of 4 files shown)


test/python/circuit/library/test_mcmt.py
```diff
@@ -264,6 +265,17 @@ def test_invalid_base_gate_width(self):
                 with self.assertRaises(ValueError):
                     _ = MCMTGate(gate, 10, 2)
 
+    def test_gate_with_parameters_vchain(self):
```
ShellyGarion, line 275: would you like to add a test for `CRYGate` as well?

ShellyGarion (commented):
LGTM, thanks for the quick fix. I only have one small comment.

test/python/circuit/library/test_mcmt.py
Cryoris: `synth_mcmt_vchain` doesn't support already controlled gates 🙂 Maybe actually that's a check we should add... `CRYGate` could be passed into `MCMTGate`, but then it will just be reduced to `RYGate`, so that path is already covered 🙂

Cryoris: Additional check added in f656fa4 🙂

ShellyGarion (approved):
LGTM, thanks!

PR #13319, opened 2024-10-14
Add argument `assume_unitary` to ``Operator.power``.

### LINKED ISSUE #13307: `Operator.power` does not handle non-unitary inputs correctly
### Environment

- **Qiskit version**: 1.2.4
- **Python version**: 3.10
- **Operating system**: macOS

### What is happening?

The way `Operator.power` works, it's strongly assuming that the internal matrix is unitary, because it's assuming that it can calculate fractional matrix powers by taking the Schur decomposition taking the matrix power of the Schur form then applying the unitary transformation (ok so far), except it assumes that the matrix power of the Schur form is the power of its diagonal.  The Schur form is upper-triangular, but not necessarily diagonal (roughly, I think it's diagonal in the case that the input is a scaled unitary).  So our `Operator.power` code fails for non-unitary operators.

```python
>>> from qiskit.quantum_info import Operator
>>> import scipy.linalg
>>> # Not unitary.
>>> data = [[1, 1], [0, -1]]
>>> print(Operator(data).power(0.5).data)
[[1.000000e+00+0.j 0.000000e+00+0.j]
 [0.000000e+00+0.j 6.123234e-17+1.j]]
>>> print(scipy.linalg.fractional_matrix_power(data, 0.5)))
[[1.00000000e+00+0.j  5.00000000e-01-0.5j]
 [0.00000000e+00+0.j  4.93096709e-17+1.j ]]
```

### What should happen?

The two should give (approximately) equal results (though see #13305 for branch-cut problems).

### Any suggestions?

The ability to do fractional matrix powers of `Operator` is relatively recent (#11534), and it seems to have pulled the code from `Gate.power`, which _was_ empowered to make that assumption.

We probably should just use `scipy.linalg.fractional_matrix_power` for `Operator.power`, and specialise `Gate.power` to use a more explicit eigenvector/eigenvalue form, since it's allowed to assume that its input was unitary.

alexanderivrii:
### Summary

Fixes #13307 following @jakelishman's suggestion in the description.

### Details and comments

After a quick round of profiling on my laptop (Python 10, Windows 11), the old (Schur-based) way to raise an operator to a power is 3x-5x faster than calling `scipy.linalg.fractional_matrix_power`, so it makes sense to use it when the operator is known to be unitary.

I have renamed the old method to be called `Operator.power_if_unitary`, this **assumes** that the given operator is unitary. The `Operator.is_power` does not make this assumption (and calls `fractional_matrix_power`). The gate class uses the `power_if_unitary` method. 

P.S. In addition, for the `power` method and an integer power, I don't see much runtime difference when calling `fractional_matrix_power` and `matrix_power`. For the `power_if_unitary` method calling `matrix_power` is faster than calling the Schur-based implementation.


qiskit/quantum_info/operators/operator.py
```diff
@@ -22,6 +22,7 @@
 from typing import TYPE_CHECKING
 
 import numpy as np
+import scipy.linalg
```
alexanderivrii: Question: should we import `scipy.linalg` at the top-level, or locally within the `power`/`power_if_unitary` functions (as before)?

jakelishman: We've historically forced `scipy` imports to be runtime imports because the cost of initialising too many `scipy` submodules is very high, and it made `import qiskit` feel very sluggish for users (and developers).  We still have a lint against it, so chances are there'll be a lint failure if it's a top-level import.

That said, in the last couple of years, scipy have done a lot to make their imports much lazier and on-demand.  That means that `import scipy` now makes `scipy.linalg` available (but imported only on demand) whereas previously you had to manually import it.  I believe they may also have applied some of the lazy loading to subcomponents of the subpackages as well, so the imports might not be as costly as they used to be.

At any rate, we still have the lint right now, so it'll need to be a runtime import.

alexanderivrii: Thanks! Changed in [a82b123](https://github.com/Qiskit/qiskit/pull/13319/commits/a82b123e17ef4e4878696733683a1042adc1fbe1). 

And this also answers my would be followup question of why we call `import scipy` on the top-level in `qsd.py`.

qiskit/quantum_info/operators/operator.py
```diff
@@ -540,14 +540,17 @@ def compose(self, other: Operator, qargs: list | None = None, front: bool = Fals
         ret._op_shape = new_shape
         return ret
 
-    def power(self, n: float) -> Operator:
-        """Return the matrix power of the operator.
+    def power_if_unitary(self, exponent: float) -> Operator:
+        """Return the matrix power of the operator, provided that operator is
+        unitary.
+
+        This is faster than the ``power`` method.
```
jakelishman: I think we might want to put a bigger warning in the docstring of this method that it'll produce nonsense if the operator wasn't unitary.

Also, super minor, but ` ``power`` ` can become `` :meth:`power` `` to insert a cross-ref.

jakelishman (commented):
This generally seems sensible to me, thanks for the fix.

A minor question (and I'm fine either way): do you prefer adding a separate `power_if_unitary` method, or adding an `assume_unitary=False` flag to the existing `power` method?  Either's fine, I'm just making sure we've thought about the interface we want.

qiskit/quantum_info/operators/operator.py
```diff
@@ -556,20 +559,42 @@ def power(self, n: float) -> Operator:
...
 
             # Experimentally, for fractional powers this seems to be 3x faster than
-            # calling scipy.linalg.fractional_matrix_power(self.data, n)
+            # calling scipy.linalg.fractional_matrix_power(self.data, exponent)
             decomposition, unitary = scipy.linalg.schur(self.data, output="complex")
             decomposition_diagonal = decomposition.diagonal()
-            decomposition_power = [pow(element, n) for element in decomposition_diagonal]
+            decomposition_power = [pow(element, exponent) for element in decomposition_diagonal]
             unitary_power = unitary @ np.diag(decomposition_power) @ unitary.conj().T
             ret._data = unitary_power
```
jakelishman: I just made #13358 to sort out the SciPy 1.14 issues, which as a side-effect modified all this code to use the `fraction_matrix_power` form with a little rotation to push the branch cut somewhere less likely to cause issues (Lev's idea).  That obsoletes the "bugfix" here in `power` - sorry, I was just trying to get SciPy 1.14 working quicker.

That said, I think the new method / `assume_unitary` keyword arg to this function is still a very useful idea, so this note is just to make sure that we apply the same branch-cut protection trick here too, and ideally add some tests that it's stable for various roots.

alexanderivrii:
@jakelishman, thanks for fixing the original issue. Based on review comments, I have added the ``assume_unitary`` argument to ``Operator.power(...)`` instead of introducing a separate function (it looks much cleaner and I don't know why I did not think of this myself) and applied the "branch-cut protection trick" for the Schur-based implementation as well. Please take a look if this is what you had in mind.

alexanderivrii:
Hmm, I have made a mistake of renaming the release notes file, while keeping the hash, now I have no idea how to make the tests pass.

jakelishman:
Yeah, reno doesn't like it when it can't track what happened to a note.  Might be easiest to rebase this on top of `main`, and squash the commits together - I think that one way or another, the history needs rewriting because reno needs each commit in the series to work, and we can't fix a broken commit just by adding more children.

qiskit/quantum_info/operators/operator.py
```diff
@@ -572,6 +574,8 @@ def power(self, n: float, branch_cut_rotation=cmath.pi * 1e-12) -> Operator:
                 complex plane.  This shifts the branch cut away from the common point of :math:`-1`,
                 but can cause a different root to be selected as the principal root.  The rotation
                 is anticlockwise, following the standard convention for complex phase.
+            assume_unitary (bool): if `True`, the operator is assumed to be unitary. In this case,
+                for fractional powers we employ a faster implementation based on Schur's decomposition.
```
jakelishman:
```suggestion
            assume_unitary (bool): if ``True``, the operator is assumed to be unitary. In this case,
                for fractional powers we employ a faster implementation based on Schur's decomposition.
```

It might be worth mentioning that the output will just be wrong if this is set `True` and the matrix isn't unitary.

qiskit/quantum_info/operators/operator.py
```diff
@@ -588,11 +592,23 @@ def power(self, n: float, branch_cut_rotation=cmath.pi * 1e-12) -> Operator:
         else:
             import scipy.linalg
 
-            ret._data = cmath.rect(
-                1, branch_cut_rotation * n
-            ) * scipy.linalg.fractional_matrix_power(
-                cmath.rect(1, -branch_cut_rotation) * self.data, n
-            )
+            if not assume_unitary:
```
jakelishman: Since this is an `if/else`, it might be _slightly_ cleaner to have the positive case (`if assume_unitary`) first, so there's no "double negative" implied by the `else`.

jakelishman (commented):
This looks good thanks.  The one thing I just thought of now is that I'm not 100% certain the condition for success is whether the matrix is unitary-like - I think actually the condition is that the matrix is normal (i.e. $A A^\dagger = A^\dagger A$).  That obviously is true for unitaries, scaled unitaries and Hermitian matrices.

Maybe in that sense, the argument should be called `assume_normal`?

qiskit/quantum_info/operators/operator.py
alexanderivrii: Done in [47f7f90](https://github.com/Qiskit/qiskit/pull/13319/commits/47f7f907b51ad2404b2615e5058a9bd14cefef11).

qiskit/quantum_info/operators/operator.py
alexanderivrii: A added a note, please see [47f7f90](https://github.com/Qiskit/qiskit/pull/13319/commits/47f7f907b51ad2404b2615e5058a9bd14cefef11).

alexanderivrii:
Thanks @jakelishman, I think that you are fully correct that the matrix only needs to be _normal_ for our implementation to work. I feel that the name "assume_normal" might be a bit less intuitive to users than "assume_unitary", so I have kept the name "assume_unitary" but briefly mentioned this in the docstring. I am happy to change it, if you prefer.

ElePT (commented):
I agree with @alexanderivrii, I think it's safe to leave ``assume_unitary`` (as it's more restrictive than `normal`) and a bit more expected in this context. I don't think many users with non-unitary-normal-operators would be affected by not being fully general in the name (the output would still be correct, just a bit less efficient), and at least it's documented in the docstring. I do have one comment for @alexanderivrii and it's to add a bugfix reno on top of the feature reno mentioning that the issue with non-unitary inputs has been fixed. Other than that LGTM.

alexanderivrii:
Thanks @ElePT. Technically, the bug has been fixed by @jakelishman in #13358 (along with the branch cut improvement), and (I've double-checked) the bugfix note already appears there.

ElePT (approved):
Ahaa, noted. Then let's go!

PR #13327, opened 2024-10-15
Implement `__eq__` for remaining standard-library gates except `mcx`

jakelishman:
### Summary

Most standard-library gates had a specialised `__eq__` added in gh-11370, but a small number were left over.  This adds specialisation to almost all the stragglers, which were mostly old soft-deperecated gates.  Despite not appearing in most circuits, these gates have an outsized effect on equality comparisons if they ever _do_ appear, because they involve construction of entire `definition` fields.

### Details and comments

You can see the improvement here in the OQ3 exporter for circuits containing some of the affected gates; it uses gate equality at one part of the symbol lookup.  For example, given this circuit:
```python
from qiskit import qasm3

circ = random_circuit(200, 200, seed=42)
```
The time of `qasm3.dumps(circ)` on my machine went from about 1.15(2)s to 0.76(2)s with this PR, despite only about 9% of the gates in the circuit being affected by this PR.

### DIFF
--- a/qiskit/circuit/library/standard_gates/p.py
+++ b/qiskit/circuit/library/standard_gates/p.py
@@ -431,3 +431,11 @@ class MCPhaseGate(ControlledGate):
     def inverse(self, annotated: bool = False):
         r"""Return inverted MCPhase gate (:math:`MCPhase(\lambda)^{\dagger} = MCPhase(-\lambda)`)"""
         return MCPhaseGate(-self.params[0], self.num_ctrl_qubits)
+
+    def __eq__(self, other):
+        return (
+            isinstance(other, MCPhaseGate)
+            and self.num_ctrl_qubits == other.num_ctrl_qubits
+            and self.ctrl_state == other.ctrl_state
+            and self._compare_parameters(other)
+        )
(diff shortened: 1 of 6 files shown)


Cryoris (commented):
LGTM 👍🏻 

Just two questions:
* Would it make sense to move `__eq__` just to `Gate` and `ControlledGate` (with the two special cases of MCPhase/U1 that adds the num_ctrl_qubits check)?
* I assume you didn't add it to MCX due to all the different gate flavors, right? If yes then we should remember to add this once the flavors are gone and we just have the `MCXGate` block 🙂

jakelishman:
The trouble is that these specialised `__eq__` methods need to be opt-in, when the type is certain that there are no other modifications.  The base `Gate` and `ControlledGate` can't know that it's safe to override this; they won't know how the `_define` method is implemented, so they need to fail safer (which is actually what they do).  These checks don't need to assert tests on the definitions, because they can know equality without creating them.

For mcx: yeah, I didn't want to get into the can of worms of all the different variants, and what should be considered `==` between them.

Cryoris (approved):

PR #13309, opened 2024-10-10
Add official support for Python 3.13

mtreinish:
### Summary

This commit adds "official" support for Python 3.13 to Qiskit. We implicitly supported it previously because nothing in Qiskit itself was incompatible with 3.13 and we use the stable ABI from rust which is compatible with new releases. But to mark 3.13 as "official" we just need to add it to CI, including the cibuildwheel test phase, and add the trove classifier to the package metadata indicating 3.13 is supported.

PR #13328, opened 2024-10-15
Test wheel builds during PR CI on demand

jakelishman:
### Summary

Refactor the wheels-build and deployment workflow into one re-usable "build wheels" workflow and two orchestration workflows.  The orchestration workflows are the existing "tag push" event, which also deploys the wheels to PyPI, and a new "PR" trigger that runs the all-wheels build if the PR is labelled with `ci: test wheels`. PGO can be turned on or off based on the workflow inputs.

The deployment job is slightly reorganised so that all the "core" platforms build and deploy before the less common platforms even begin the build.  The core platforms are all tier 1 platforms, the 32-bit platforms, and the sdist.  These core platforms were already tied together, but the other platforms ran concurrently with them.  This could lead to other projects' CI failing when Qiskit was part way through a deployment.

The "other" wheels are all tied together in this PR mostly as a convenience to avoid repetition.  They could easily be untied from each other (as the parent commit does).

### Details and comments

The idea of this is that we can now use https://github.com/Qiskit/qiskit/labels/ci%3A%20test%20wheels to tag PRs that affect the deployment process (like dependabot PRs to cibuildwheel, or #13309 adding Python 3.13).  The CI run isn't required to pass to allow merge, but it significantly lowers the barrier to entry for testing a PR, compared to the prior system of somebody needing to manually push a branch that fiddled with the script triggers to test a build.

I've attached the label to this PR, so it should trigger the additional checks.  Removing the label should mean the runs stop happening on PR syncs.

We can, in a follow-up, potentially add a weekly or nightly job to build the Qiskit release artifacts (with or without PGO as we choose) for downstream consumption.  This in particular might be helpful around major-version changes, to make it easier for other projects to build against a recent Qiskit without needing to build from source.

Testing on my own fork shows that the Windows 32-bit builds are currently broken.

### DIFF
--- a/.github/workflows/wheels-build.yml
+++ b/.github/workflows/wheels-build.yml
@@ -1,0 +1,247 @@
+name: Build release artifacts
+on:
+  workflow_call:
+    inputs:
+      default-action:
+        description: >-
+          The default action for each artifact.
+          Choose from 'build' (default) or 'skip'.
+        type: string
+        default: "build"
+        required: false
+
+      sdist:
+        description: >-
+          The action to take for the sdist.
+          Choose from 'default', 'build' or 'skip'.
+        type: string
+        default: "default"
+        required: false
+
+      wheels-tier-1:
+        description: >-
+          The action to take for Tier 1 wheels.
+          Choose from 'default', 'build' or 'skip'.
+          This builds multiple artifacts, which all match 'wheels-tier-1-*'.
+        type: string
+        default: "default"
+        required: false
+
+      wheels-32bit: 
+        description: >-
+          The action to take for Tier 1 wheels.
+          Choose from 'default', 'build' or 'skip'.
+          This builds multiple artifacts, which all match 'wheels-32bit-*'.
+        type: string
+        default: "default"
+        required: false
+
+      wheels-linux-s390x: 
+        description: >-
+          The action to take for Linux s390x wheels.
+          Choose from 'default', 'build' or 'skip'.
+        type: string
+        default: "default"
+        required: false
+
+      wheels-linux-ppc64le: 
+        description: >-
+          The action to take for Linux ppc64le wheels.
+          Choose from 'default', 'build' or 'skip'.
+        type: string
+        default: "default"
+        required: false
+
+      wheels-linux-aarch64: 
+        description: >-
+          The action to take for Linux AArch64 wheels.
+          Choose from 'default', 'build' or 'skip'.
+        type: string
+        default: "default"
+        required: false
+
+      artifact-prefix:
+        description: "A prefix to give all artifacts uploaded with 'actions/upload-artifact'."
+        type: string
+        default: ""
+        required: false
+
+      python-version:
+        description: "The Python version to use to host the build runner."
+        type: string
+        default: "3.10"
+        required: false
+
+      pgo:
+        description: "Whether to enable profile-guided optimizations for supported platforms."
+        type: boolean
+        default: true
+        required: false
+      
+
+jobs:
+  wheels-tier-1:
+    name: "Wheels / Tier 1"
+    if: (inputs.wheels-tier-1 == 'default' && inputs.default-action || inputs.wheels-tier-1) == 'build'
+    runs-on: ${{ matrix.os }}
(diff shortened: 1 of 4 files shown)


jakelishman:
The failing test is the 32-bit Windows wheel, which is tracked in #13329.  The failed job there shouldn't prevent the PR from merging, because that test isn't a branch-protection requirement.

.github/workflows/wheels-build.yml
```diff
@@ -0,0 +1,247 @@
...
+        description: "Whether to enable profile-guided optimizations for supported platforms."
+        type: boolean
+        default: true
+        required: false
+      
+
+jobs:
+  wheels-tier-1:
+    name: "Wheels / Tier 1"
+    if: (inputs.wheels-tier-1 == 'default' && inputs.default-action || inputs.wheels-tier-1) == 'build'
```
raynelfss, line 85: This statement is confusing to me, I don't have much knowledge of how bashscript handles truthy string values but here is my assumption:
```txt
Run if variable wheel-tiers-1 is "default" and either default-action or wheels-tiers is provided, and the value of the one provided is "build".
```
Is this correct? I'm assuming that `true` && any string == any string, but `false` + any string == `false`.

.github/workflows/wheels-build.yml
```diff
@@ -0,0 +1,247 @@
...
+          - macos-12
+          # Used for the ARM builds.
+          - macos-14
+          - windows-latest
+    steps:
+      - uses: actions/checkout@v4
+      - uses: actions/setup-python@v5
+        with:
+          python-version: ${{ inputs.python-version }}
+          architecture: ${{ matrix.os == 'macos-14' && 'arm64' || 'x64' }}
```
raynelfss, line 102: Also here, this is how I interpret it:
```txt
If matrix.os current value is "macos-14"  and architecture is  either "arm64" or "x64".
```
Is the second clause supposed to give you the only two possible options?

raynelfss (commented):
This all looks good, I just have a couple of questions about the bash logic (which I'm still brushing up on). Once that's answered it should be good to go.

.github/workflows/wheels-build.yml
jakelishman: This isn't a bash thing actually, it's GitHub Action's built-in custom expression evaluator - the `if` YAML key is part of their stuff.

`&&` and `||` are short-circuiting logical operators who use the "truthiness" to determine their result, but return the actual object.  `&&` (logical and) binds tighter than `||` (logical or), and this is pretty directly equivalent to the Python:
```python
inputs["wheels-tier-1"] == "default" and inputs["default-action"] or inputs["wheels-tier-1"]
```
The intention is to be a proper ternary operator like C's:
```c
strcmp(INPUTS_WHEELS_TIER_1, "default") == 0 ? INPUTS_DEFAULT_ACTION : INPUTS_WHEELS_TIER_1
```
(which looks more similar) or Python's
```python
inputs["default-action"] if inputs["wheels-tier-1"] == "default" else inputs["wheels-tier-1"]
```

It's not actually exactly the same, though, because if `inputs.default-action` evaluates to the empty string (which is false-y), the Boolean logic will fall through to returning `inputs.wheels-tier-1` via the logical or.  That won't happen in this case, because the default action is never supposed to be the empty string.

That said, once you're used to ternary conditionals, this pattern is relatively legible (it's a common shorthand in a lot of older languages, or very limited ones like the GitHub Actions expression DSL), and the alternatives that make it safer are typically much messier (you can make an array with the two options, then index into it with a Boolean, which coerces to integer 0 or 1, for example).

.github/workflows/wheels-build.yml
jakelishman: This is equivalent to the Python:
```python
architecture = "arm64" if matrix.os == "macos-14" else "x64"
```
(since here there's no worries about the empty string appearing - both are string literals).

The GitHub Actions expression DSL gives us unfortunately few ways to do a _proper_ conditional within a single expression.

raynelfss (approved):
LGTM, thank you for the clarifications here.

Issue #13326, opened 2024-10-15
Transpiler support for one-in many-out circuit transformations


nonhermitian:
### What should we add?

There are several classes of circuit transformations that take a single circuit in and create output over many circuits.  Examples include Pauli Twirling and Lightcone truncation over a linear combination of operators.  It is tempting to write these circuit transformations as transpiler passes, but they do not satisfy the current model of one circuit in, one-out for passes.  Currently, the best way to implement these methods is to break transpilation into 2 pieces and perform the one-in many-out operations in-between.  It would be cool if the transpiler could accommodate these types of transformations

PR #13331, opened 2024-10-15
Add pauli_twirl_2q_gates function

### LINKED ISSUE #13325: Extend Pauli twirling to end users
### What should we add?

Qiskit has the code to Pauli twirl circuits  (up to a phase factor) within its unit tests. Other SDKs have this functionality builtin, e.g. `pytket.tailoring.PauliFrameRandomisation`.  This functionality should be exposed to the end users.

mtreinish:
### Summary

This commit adds a new function twirl_circuit to apply Pauli twirling to a given circuit. This function only works with a fixed set of two qubit gates and is not a general solution. For performance this new function is written in rust and uses static lookup tables for the twirling sets around the fixed two qubit gates that are randomly sampled. There is an option to perform twirling multiple times and return a list of circuits instead of just doing it once. Ideally we'd do this in parallel, but we're currently blocked on the use of OnceCell in the PackedInstruction and the ParameterTable from parallelizing with CircuitData. We can further improve the performance of this new function by using a parallel iterator once #13219 is resolved. The function is written in a way to make this simple in the future.

### Details and comments

Fixes #13325


mtreinish:
I ran an asv comparison (since the benchmarks needed to change to show the improvement here it's not a direct comparison) and this is showing a improvement of the previous twirling benchmark written in python from taking 254±4ms to 6.38±0.3ms with the `twirl_circuit` function in this PR:

```
| Change   | Before [54fe09b2] <lint_incr_latest~1>   | After [33b4bde2] <twirl-for-paul>   |   Ratio | Benchmark (Parameter)                                 |
|----------|------------------------------------------|-------------------------------------|---------|-------------------------------------------------------|
| x        | 254±4ms                                  | 6.38±0.3ms                          |    0.03 | manipulate.TestCircuitManipulate.time_DTC100_twirling |
```

The "Change" X is because asv detected the benchmark changed and the results may not be comparable.

qiskit/circuit/twirling.py
```diff
@@ -0,0 +1,53 @@
...
+from qiskit._accelerate.twirling import twirl_circuit as twirl_rs
+from qiskit.circuit.quantumcircuit import QuantumCircuit
+from qiskit.circuit.gate import Gate
+from qiskit.circuit.library.standard_gates.x import CXGate
+from qiskit.exceptions import QiskitError
+
+
+def twirl_circuit(
+    circuit: QuantumCircuit,
+    twirling_gate: Gate = CXGate,
```
Cryoris:
```suggestion
    twirling_gate: type[Gate] = CXGate,
```
... otherwise this would ask for `Gate` instances, I think, no? 🙂

qiskit/circuit/twirling.py
```diff
@@ -0,0 +1,53 @@
...
+    seed: int = None,
```
Cryoris:
```suggestion
    seed: int | None = None,
```

crates/accelerate/src/twirling.rs
```diff
@@ -0,0 +1,689 @@
...
+
+use qiskit_circuit::circuit_data::CircuitData;
+use qiskit_circuit::operations::{OperationRef, Param, StandardGate};
+use qiskit_circuit::Qubit;
+
+use crate::QiskitError;
+
+static ECR_TWIRL_SET: [([StandardGate; 4], f64); 16] = [
+    (
+        [
```
Cryoris: Just a suggestion to try: I wonder if these would be more easily readable/comparable if they we would omit `StandardGate::` and could write them in a single line (in case that would make it short enough)? 🤔

crates/accelerate/src/twirling.rs
```diff
@@ -0,0 +1,689 @@
...
+        None => Pcg64Mcg::from_entropy(),
+    };
+    let twirl_set: &[([StandardGate; 4], f64)] = match twirled_gate {
+        StandardGate::CXGate => &CX_TWIRL_SET,
+        StandardGate::CZGate => &CZ_TWIRL_SET,
+        StandardGate::ECRGate => &ECR_TWIRL_SET,
+        StandardGate::ISwapGate => &ISWAP_TWIRL_SET,
+        _ => {
+            return Err(QiskitError::new_err(
+                "Provided gate to twirl is not currently supported you can only use CX or CZ",
```
Cryoris:
```suggestion
                "Provided gate to twirl is not currently supported you can only use CX, CZ, ECR or iSwap.",
```

crates/circuit/src/circuit_data.rs
```diff
@@ -396,6 +397,46 @@ impl CircuitData {
...
+    /// Clone an empty CircuitData from a given reference.
+    ///
+    /// The new copy will have the global properties from the provided `CircuitData`.
+    /// The the bit data fields and interners, global phase, etc will be copied to
+    /// the new returned `CircuitData`, but the `data` field's instruction list will
+    /// be empty. This can be useful for scenarios where you want to rebuild a copy
+    /// of the circuit from a reference but insert new gates in the middle.
+    pub fn clone_empty_from(other: &Self) -> Self {
+        CircuitData {
+            data: Vec::with_capacity(other.data.len()),
```
Cryoris: This looks like a potential sharp bit, as we don't really know how large the new circuit is going to be. How about making the capacity and option than defaults to the other circuit's size?

crates/accelerate/src/twirling.rs
```diff
@@ -0,0 +1,689 @@
...
+    let generate_twirled_circuit = |rng: &mut Pcg64Mcg| {
+        let mut out_circ = CircuitData::clone_empty_from(circ);
+
+        for inst in circ.data() {
+            match inst.op.view() {
+                OperationRef::Standard(gate) => {
+                    if gate == twirled_gate {
+                        let qubits: Vec<Qubit> = out_circ.get_qargs(inst.qubits).to_vec();
+                        let (twirl, twirl_phase) = twirl_set.choose(rng).unwrap();
+                        out_circ.push_standard_gate(twirl[0], &[], &[qubits[0]])?;
```
Cryoris: Would there be a performance benefit to omitting identity gates?

Cryoris (commented):
That's a pretty neat speedup for something used _a lot_ 🙂

crates/accelerate/src/twirling.rs
mtreinish: I think it wouldn't make much of a difference for the performance of this function. The two ways to do it would be to condition the `push` call on the value of `twirl[index]` which adds a bunch of branching logic around each push call or somehow tweak the static twirling sets which would be tricky because they work well because everything is fixed sizes. There is a potential benefit downstream of this function because we've potentially used less gates

crates/circuit/src/circuit_data.rs
mtreinish: I made this configurable and explicit in: https://github.com/Qiskit/qiskit/pull/13331/commits/34c3e918e291d2a039289fc2c0d69c88721763f5

crates/accelerate/src/twirling.rs
mtreinish: This was a very good suggestion, you're going to like how it looks now: https://github.com/Qiskit/qiskit/pull/13331/commits/4fadd0dd7cdad13a1b33c656781ee751fa59f6df

qiskit/circuit/twirling.py
```diff
@@ -0,0 +1,53 @@
...
+    gate.
+
+    Args:
+        circuit: The circuit to twirl
+        twirling_gate: The gate to twirl, currently only :class:`.CXGate`, :class:`.CZGate`,
+            :class:`.ECRGate`, and :class:`.iSwapGate` are supported.
+        seed: An integer seed for the random number generator used internally.
+            If specified this must be between 0 and 8,446,744,073,709,551,615.
+        num_twirls: The number of twirling circuits to build. This defaults to 1 and will return
+            a single circuit. If it is > 1 a list of circuits will be returned.
```
jakelishman: I think this leads to an awkward interface, because the type is super dependent on the value, and something that's programmatically generating different numbers of twirls will always have to account for this.

I think having `num_twirls: None | int = None` is safer, where `None` means "not given" and defaults to 1.  If `num_twirls` is given as an int, even if it's 0 or 1, then we can return a list. (Type checkers can be taught to understand that too.)

qiskit/circuit/twirling.py
```diff
@@ -0,0 +1,53 @@
...
+    seed: int | None = None,
+    num_twirls: int = 1,
+) -> QuantumCircuit | list[QuantumCircuit]:
+    """Create a copy of a given circuit with Pauli twirling applied around a specified two qubit
+    gate.
+
+    Args:
+        circuit: The circuit to twirl
+        twirling_gate: The gate to twirl, currently only :class:`.CXGate`, :class:`.CZGate`,
+            :class:`.ECRGate`, and :class:`.iSwapGate` are supported.
```
jakelishman: The typing of this feels awkward - the restriction for it to be just a single gate feels somewhat arbitrary; since the output gates are Paulis, there's no particular guarantee that they're going to be supported on whatever hardware is being used, and so we're going to be running some sort of translation routine.  If we're running a translation routine, it feels reasonable to assume that the input circuit might not yet be mapped to a uniform 2q gate (and with the advent of heterogeneous backends, even compiled circuits might need multiple-gate twirling).

Do we think it's worth having this accept a list of gates to twirl, and some way to say "twirl everything you know" / "error if there's a gate you can't twirl"?

Also, for consistency with other parts of Qiskit: should we accept gate names as well?

qiskit/circuit/twirling.py
```diff
@@ -0,0 +1,53 @@
...
+        seed: An integer seed for the random number generator used internally.
+            If specified this must be between 0 and 8,446,744,073,709,551,615.
```
jakelishman: Why?  That's a weird number.

qiskit/circuit/twirling.py
```diff
@@ -0,0 +1,53 @@
...
+        instance of the specified twirling gate.
+    """
+    twirling_std_gate = getattr(twirling_gate, "_standard_gate", None)
+    if twirling_std_gate is None:
+        raise QiskitError("This function can only be used with standard gates")
+    new_data = twirl_rs(circuit._data, twirling_std_gate, seed, num_twirls)
+    if num_twirls > 1:
+        return [QuantumCircuit._from_circuit_data(x) for x in new_data]
+    else:
+        return QuantumCircuit._from_circuit_data(new_data[0])
```
jakelishman: What if `num_twirls` was 0?

crates/circuit/src/dag_circuit.rs
```diff
@@ -6937,7 +6937,7 @@ impl DAGCircuit {
 
 /// Add to global phase. Global phase can only be Float or ParameterExpression so this
 /// does not handle the full possibility of parameter values.
-fn add_global_phase(py: Python, phase: &Param, other: &Param) -> PyResult<Param> {
+pub(crate) fn add_global_phase(py: Python, phase: &Param, other: &Param) -> PyResult<Param> {
```
jakelishman, line 7012: If we're going to use this elsewhere, can we think about where it should live, and how we should be organising the crate?  Imports like `crate::dag_circuit::add_global_phase` are awkward to work with internally if you weren't involved with the PR that originally put the function there.

crates/circuit/src/circuit_data.rs
```diff
@@ -396,6 +397,53 @@ impl CircuitData {
...
+        CircuitData {
+            data: Vec::with_capacity(capacity.unwrap_or(other.data.len())),
+            qargs_interner: other.qargs_interner.clone(),
+            cargs_interner: other.cargs_interner.clone(),
+            qubits: other.qubits.clone(),
+            clbits: other.clbits.clone(),
+            param_table: ParameterTable::new(),
+            global_phase: other.global_phase.clone(),
+        }
+    }
```
jakelishman: In Python space we called this `copy_empty_like` - should we unify and have it be `clone_empty_like` here?  (Allowing for the `copy`/`clone` difference between Python and Rust.)

crates/accelerate/src/twirling.rs
```diff
@@ -0,0 +1,181 @@
...
+                        if *twirl_phase != 0. {
+                            out_circ.add_global_phase(py, &Param::Float(*twirl_phase))?;
+                        }
+                    } else {
+                        out_circ.push(py, inst.clone())?;
+                    }
+                }
+                _ => {
+                    out_circ.push(py, inst.clone())?;
+                }
```
jakelishman: This isn't handling control-flow operations or open-controlled CX/CZ gates, though there's no particular reason that the function couldn't do that.

crates/accelerate/src/twirling.rs
jakelishman: On average, 25% of the gates we'd be adding to the circuit will be identities, and that's a fair amount of data to reduce by not outputting it, which I'd expect to have a measurable impact on the runtime of this function.  I wouldn't imagine that the branching logic will make any huge amount of difference, because there's already a decent amount of branching and hashing work hiding within each function call here.

crates/accelerate/src/twirling.rs
```diff
@@ -0,0 +1,181 @@
...
+        }
+    };
+    let generate_twirled_circuit = |rng: &mut Pcg64Mcg| {
+        let mut out_circ = CircuitData::clone_empty_from(circ, None);
+
+        for inst in circ.data() {
+            match inst.op.view() {
+                OperationRef::Standard(gate) => {
+                    if gate == twirled_gate {
+                        let qubits: Vec<Qubit> = out_circ.get_qargs(inst.qubits).to_vec();
```
jakelishman: Why allocate a `Vec` here?

crates/circuit/src/circuit_data.rs
```diff
@@ -396,6 +397,53 @@ impl CircuitData {
...
+
+    /// Append a PackedInstruction to the circuit data.
+    pub fn push(&mut self, py: Python, packed: PackedInstruction) -> PyResult<()> {
+        let new_index = self.data.len();
+        self.data.push(packed);
+        self.track_instruction_parameters(py, new_index)
+    }
```
jakelishman: We ought to put more documentation about the preconditions on this - it's vitally important that the qubits/clbits interners have already been configured to include the indices in the `PackedInstruction`.

crates/accelerate/src/twirling.rs
```diff
@@ -0,0 +1,181 @@
...
+                        let (twirl, twirl_phase) = twirl_set.choose(rng).unwrap();
+                        out_circ.push_standard_gate(twirl[0], &[], &[qubits[0]])?;
+                        out_circ.push_standard_gate(twirl[1], &[], &[qubits[1]])?;
+                        out_circ.push(py, inst.clone())?;
+                        out_circ.push_standard_gate(twirl[2], &[], &[qubits[0]])?;
+                        out_circ.push_standard_gate(twirl[3], &[], &[qubits[1]])?;
```
jakelishman: If you're concerned about performance here, then it's probably going to be more efficient to use the interners to find the single-qubit `qargs` keys either right before the pushes, or up at the top of the function before the outermost loop at all, then build the `PackedInstruction`s manually.

crates/accelerate/src/twirling.rs
nonhermitian: The nice thing is that id gates explicitly show the added twirling gates.  Moreover, in practice one should run a 1Q gate cleanup after twirling that should take care of this e,g

```python
post_pm = PassManager([PauliTwirling('ecr'), 
                       Optimize1qGatesDecomposition(backend.target.operation_names)])
```

crates/accelerate/src/twirling.rs
```diff
@@ -0,0 +1,181 @@
...
+        // TODO: Use into_par_iter() after CircuitData is made threadsafe
+        // (see https://github.com/Qiskit/qiskit/issues/13219)
+        seed_vec
+            .into_iter()
+            .map(|seed| {
+                let mut inner_rng = Pcg64Mcg::seed_from_u64(seed);
+                generate_twirled_circuit(&mut inner_rng)
+            })
+            .collect()
+    }
```
jakelishman: The difference in seed handling here is going to have the kind-of weird property that the output lists will have the same prefix as `num_qubits` increases (so `out[0]` will be the same if `num_twirls` is 2 or 3) except that there's a cut at `num_qubits > 5` and then all _those_ different numbers of twirls will share a prefix too, but it'll be a different prefix to `num_qubits <= 4`.

That just feels a bit awkward, and maybe we should use the same strategy to generate the seed vecs in both paths?

qiskit/circuit/twirling.py
jakelishman: What interfaces should we consider around the extra metadata in the `QuantumCircuit`?  Should we be maintaining the registers / calibrations / label / metadata in the output?  I think a user might reasonably expect that the output here to be similar to having been built from `QuantumCircuit.copy_empty_like`, and it might well be surprising to them if we've erased bit information, etc.

jakelishman (commented):
If this is going to be generally useful to people, I'm ok with adding it, but I think we need to consider the interfaces a bit more, and how it fits in with the rest of Qiskit.  The benchpress version was one very specific implementation for a specific type of circuit that only works for one particular way of building the input circuit, and as long as it doesn't include quite a few of the possible features of Qiskit circuits.

In addition to the review comments, one more top-level one: I'm not really sure where in a user's compilation pipeline this is meant to be called, which also affects what I'd expect to have in the function.  The twirling is adding Pauli X/Y/Z gates (obviously), which implies to me that it needs to come _before_ compilation, but then it also makes assumptions that everything's going to be decomposed in terms of one single 2q gate, which, aside from being an awkward assumption for the future, means that we might need a better way to handle things than to output non-translated 1q gates?  If we instead want to push it more in the direction of "apply this to a virtual circuit, but still one that's decomposed into 2q operations", I think we need to consider interfaces to control the known twirling sets for additional gates.

I'm concerned that we're committing to an interface acting on `QuantumCircuit` that doesn't really act on either virtual or physical circuits, but somewhere in the middle.  That's a transpiler pass, except that the `num_twirls` thing makes that awkward, because any sort of multi-in / multi-out branching within the transpiler needs a significant rework of the pass-manager framework to become safe (speaking as somebody who already wrote such a modification once).

crates/accelerate/src/twirling.rs
jakelishman: Those kinds of nice things have a very poor effect on performance.  We've just made a big deal about Qiskit performance, and if we want to maintain it, we need to keep putting performance first, and designing our interfaces around that.

We can potentially allow it to be opt-in to show the twirling gates, but it's fairly trivial to write the twirling algorithm yourself if that's what you care about seeing, and that's always going to give you the most control over the output.  If what we're after is that kind of "utility" function rather than a high-performance one, I'm not sure it fits in Qiskit - we already shipped off bunches of `qiskit_algorithms` and the applications libraries out of the core library because they didn't fit the "highest performance core data structures and compilation" model, and decided to do that with documentation instead.  If the purpose of the function is educational, then it feels to me like it fits in that vein, and should be handled with guide-level documentation on "introduction to Pauli twirling" rather than being presented in the core library as "this is an efficient way to do it".

qiskit/circuit/twirling.py
mtreinish: It's `u64::MAX` which is what we seed the rust rng with

qiskit/circuit/twirling.py
mtreinish: The circuit isn't twirled.

qiskit/circuit/twirling.py
jakelishman: oh, it's missing a 1 off the front

jakelishman: I didn't recognise it

crates/accelerate/src/twirling.rs
mtreinish: This was necessary because the output references from `get_qargs` are from `out_circ` and we mutate `out_circ` below so the borrow checker wasn't happy because we have immutable references here when trying to get a mutable reference for the push calls.

qiskit/circuit/twirling.py
jakelishman: Sure - I mean our logic will have us index into an empty list to attempt to return a scalar, rather than return an empty-list output.

qiskit/circuit/twirling.py
mtreinish: Sure I can change this, I don't feel too strongly about it.

crates/accelerate/src/twirling.rs
nonhermitian: Ok if id gates are a performance blocker than so be it.  It also does not change the fact that I need to run a 1Q cleanup at the end, so either way, it is more important to have a function that gives users the ability to do this easily.

qiskit/circuit/twirling.py
mtreinish: I like the idea of taking multiple gates here, I also think there is value in accepting `None` here and having it twirl all the gates it can work with. I'll update the interfaces accordingly.

As for names vs class I only used class because I was a bit lazy as I wanted to work solely with `StandardGate` in rust and it was easier to query the `_standard_gate` attribute to get that. But I can make it work with names too.

crates/accelerate/src/twirling.rs
jakelishman: Sure - I commented on that in my main review.  It's not really clear to me what the purpose of this function is at all from a performance perspective.

If the purpose is to run it post-compilation, then we should design the interface of the function to be smarter about twirls in the first place, so it outputs at least hardware-compatible circuits.  Technically there'll still be possibilities for post-optimisation, since 1q runs of gates will likely be possible to be resynthesised, which again is something we might want to consider.

crates/circuit/src/dag_circuit.rs
mtreinish: I can move it somewhere, I only left it here for brevity. Where would you like me to put it?

qiskit/circuit/twirling.py
mtreinish: Oh, I see what you mean now. I'll fix this

crates/accelerate/src/twirling.rs
mtreinish: I can unify the seeding strategy. Realistically I should remove this branching, it's me just setting the ground for doing this in parallel but there is no benefit for it right now since we can't use a parallel iterator here.

crates/accelerate/src/twirling.rs
jakelishman: Ah sorry, missed that.  Since we're already making the assumption that the `qargs` interners are equivalent to the base circuit (from `clone_empty_like` and the subsequent direct `push(inst.clone()`), can we instead use the _input_ circuit to get the qargs, and avoid the vec allocation?

nonhermitian:
@jakelishman PT is not something that fits into our current circuit transformation pipelines.  It basically has to be a middle step between two passmanagers.  One to map to HW, followed by PT, and then another that does cleanup of the 1Q gates and casts back into the HW basis set (usually over a set of circuits generated by PT).  This is one of the motivations for #13326

qiskit/circuit/twirling.py
mtreinish: That's a fair point I'll come up with something to preserve the python space metadata in the output copies.

jakelishman:
If it doesn't really fit in Qiskit's models right now (and I agree, it doesn't particularly), then it feels like we're trying to merge such a function into the core library that'll never really fit in.  If we're asking users to start using this now as part of the core library, they're going to find it far more awkward in the future if we say "actually, don't do this anymore and instead use this transpiler pass".  That'll alienate our users, who already don't like the amount of changes we make.

Making the compiler multi-in, multi-out is interesting, and something we already do internally in the IBM backends (I wrote it), but it's hugely fragile in the current pass-manager system, and it's tricky to make it better without compromising performance.  We can get away with it in the backends because there's far tighter control over what passes are in the pipelines and when a pass might trigger such fan-out behaviour.

We'll have a chance to do that better when we move more of the pass-manager infrastructure into Rust; we'll have more opportunities to make prior state in the `PropertySet` copy-on-write, rather than potentially eagerly copying large amounts of data at the point of fan-out, or being smarter about invalidating stale data in the `PropertySet`.  Both of those would make an internal IR that's fan-out aware easier to make safe (the main concern) and performant (the next most important), but we're not there yet.

---

If this is only ever meant to be an intermediate-term thing, I'm most worried about how we'll present this narratively to users, especially when we need to shift them to something else.


On another narrative note: are we intending to present twirling as something that all users will need to do, and not the domain of a primitive implementer?  Adding twirling as something we do on a single circuit, outside the context of the estimator observables and potential lightcone analysis might accidentally cause primitives implementations to have to do more work and have the user do more work to reconstruct their final results, since now a user might well end up passing a much larger broadcast matrix as their estimator inputs.

crates/circuit/src/dag_circuit.rs
jakelishman: Honestly, I don't know right now.  It's probably part of a separate issue.

nonhermitian:
>If it doesn't really fit in Qiskit's models right now (and I agree, it doesn't particularly), then it feels like we're trying to merge such a function into the core library that'll never really fit in

But this is equiv to kicking the can down the road to a undetermined time frame, and features that have value will likely never see the light of day, waiting for the ideal conditions for implementation.   Perhaps latter a pass can just call this same function that remains exposed to users as well, who knows.

Twirling has value for Qiskit outside of the IBM primitives.  Note that no other provider outside of IBM has a primitives interface that supports twirling.  So there is value to other HW providers, and in addition there is value for users who want to see (and perhaps save) the exact circuits they are executing.  

I am not sure there is any major downside on the primitives side save for perhaps extra work to be done that otherwise didn't need to be done, e.g. twirl twice if that was an automatic option.  However, I am pretty sure there is an option for that.

jakelishman:
A later pass can never call this function, because it's defined on the wrong types of inputs.  Besides, that doesn't help the user, who would still have to switch what they're doing.

Yes, I am proposing step back and think through what a joined-up interface to Qiskit from the user right through to hardware looks like, rather than us adding things in that implement one overly-specific microbenchmark more efficiently.  We can't really claim either performance and stability if that's how we go about additions to the core library.  I'm not saying twirling _isn't_ important, or even that we shouldn't put it in right now, I'm saying that software engineering is more than just throwing interfaces at a wall to see what sticks.  That works well in the early days of open-source work, but not for a product whose guiding principles are meant to be stability, performance and utility.

For example, if the purpose is that this will _always_ be used as a post-compilation pipeline where it's immediately followed by an `Optimize1qGates` pass, then the interface of this function probably should change in some manner that allows the twirling itself to inline the optimal 1q gate sequence for a particular backend.  That will be _more_ performant than this current implementation, because it won't need an additional transpiler pass to come later.  Using the current interface over optimises for the microbenchmark, which isn't representative of the work that actually needs to be done - it elevates a subroutine of one particular (inefficient) choice of implementation to be a standard we should hit.

> I am not sure there is any major downside on the primitives side save for perhaps extra work to be done that otherwise didn't need to be done.

If we promote that you should do your twirling from within Qiskit, this takes away important information that a primitives implementation can use to reduce the number of QPU hours and shot-loops necessary, because now the user is supplying many different `Estimator` inputs (one per twirl).  That means that the same light-cone analysis now has to be done for each separate input, whereas with twirling done in the primitive implementations, it can be done before the twirling is.  If the primitive is doing any other additional error learning or error mitigation, it'll likely now end up doing the same learning for each of the different input circuits, which could severely impact the number of QPU hours spent.  It also means that the user has more data come back that they now have to merge together, which means more data the user has to send over the wire to us, and more data we have to send over the wire back to the user.

There are ways around these problems, but again, they mean stepping back and thinking through a joined-up interface, in conjunction with implementers of the primitives.  For example, my problem of knowing when two multi-qubit potentially parametric blocks implement the same operation can be eased by additional `box`es with annotations, which is something I'm currently in active discussions with the primitives teams about.  However, if that's the way we go, then this function needs to be designed in such a way that it outputs those boxes, and we need to build the interface around that.

I do totally agree that this could have utility outside IBM primitives implementations as well, but if so, we need to really think through _how_ those other users should use this, and how we'll design our interfaces so that users will, by default and by using the most natural forms, get the experience that is most performant and most pleasant end-to-end.

qiskit/circuit/twirling.py
mtreinish: Hah, yeah a copy and paste issue, I'll fix it.

mtreinish:
```suggestion
        seed: An integer seed for the random number generator used internally.
            If specified this must be between 0 and 18,446,744,073,709,551,615.
```

qiskit/circuit/twirling.py
mtreinish: Done in: https://github.com/Qiskit/qiskit/pull/13331/commits/b4a8aacaa75de0883726bc0e9d4bf240d235a30e

qiskit/circuit/twirling.py
mtreinish: Done in: https://github.com/Qiskit/qiskit/pull/13331/commits/392782d40406c91e35d3faa8bd65dcc567d3a42a

qiskit/circuit/twirling.py
mtreinish: This should be fixed by: https://github.com/Qiskit/qiskit/pull/13331/commits/392782d40406c91e35d3faa8bd65dcc567d3a42a

nonhermitian:
I guess if you want to future proof it a bit then you should implement as a transpiler pass that only does a single twirl.  In that case the current usage would require two pass managers still, but the input to the second would be something like `post_pm.run([qc]*num_twirls)`.  Then you could probably extend it to one-in many-out in a fairly seamless manner.  That would also leave the 1Q optimization separate, which is more in the spirit of the passes doing basic things

jakelishman:
This is just why I wanted to talk about things - if the intention of this function is to just be a short-term, low-level post-`transpile` thing, then we can still have a function like what's currently implemented, it just means we tweak the interface so it's more efficient, and we need to take care about how we teach users about it.

If I'm understanding right, then the two proposed use cases for this function are:
- illustrate what Pauli twirling does
- multiply twirl a post-compilation circuit, without concern for how efficiently that will be executable on a primitive (possibly because it'll be given to a non-primitive direct-access `BackendV2.run`, which might not be an IBM one).

In that case, maybe we work in an interface where we have an optional argument that's either `None` or an instance of `Optimize1qDecomposition`.  If we're given the optimiser, the twirler can run along the 1q runs ahead of each twirled gate, then give that run and the twirling gate to the optimiser, and inline the output of that.  That allows both use cases: for use case one, you don't pass the optimiser, and for use case two, you do.  In the second use case, we're now more efficient than anything else proposed here, because there's no need for a follow-up passmanager at all.

Also if that's the case, then that impacts the interface promises we should make about the output circuit w.r.t. propagating any stored `TranspileLayout`, the register structure, and so on.  Again, easy enough to _do_, we just needed to really hone in on what the purpose and use-case of the function is in order to know that.

The concerns about this interacting poorly in general with how efficient `Estimator`s will want to work are still there, but that can generally be handled by documentation.  That of course wouldn't be the case if this was intended to be used before input to a primitive, but that's what I want to make sure we're thinking about.

nonhermitian:
The workflow is the following:

1) Transpile circuit as best as possible

2) Take this best circuit and generate `num_twirls` versions that have been twirled, and then do a 1Q cleanup on the return set

3) Execute (perhaps with things like DD added as well after twirling) using whatever execution pipeline a provider has

4) Aggregate counts and compute desired quantities of interest.

The twirling could be a function, or a `TransformationPass` with the latter generating only a single twirl at the moment.  Then one could combine the 1Q optimization pass in a `post_pm` as done above and get the desired output.

jakelishman:
Step three is the big question to me: is this not something that narratively we promote is what a primitives implementer might do for you?  How does adding this function interact with that?  Can you expand on why you have to do this manually, when this is something that IBM primitives definitely can do?

Knowing your step 2 is important, because then we can make the function better - having "twirl" be a standalone step isn't necessarily even the best interface for your workflow, because you have to do a full 1q post-optimize, while we can do that more efficiently on-the-fly (since 1q optimisation is pretty simple and a peephole optimisation).

I'd like to know more about what other users' use-cases for manual twirling are - you're presenting one use case, but it might not be the only reason people are twirling themselves, and I wouldn't want the interface to get in the way of that.  I know that there's a few people out there who've written things approximately similar to what's here, but I don't know for what purposes they've done that.

nonhermitian:
I am not sure that Qiskit has to promote anything that a singular primitive does as its preferred workflow.  However, a quick glance at the Qiskit docs shows only Runtime examples, so perhaps Qiskit SDK IS now promoting a singular providers interface for execution.

That being said, there are several reasons why one would like to have circuits in hand.  1) Reproducibility - I want to run exactly the same circuits over again. 2) Time saving - why waste time repeatedly twirling on every call to a primitive. 3) I want to run twirled circuits on something other than IBM hardware. 4) I want to mix error suppression techniques in a manner not available within a primitive.

If making a function version, as done here, then sure one can combine the steps.  The twirling would not be explicitly clear, but it would be there logically and in a form better suited for further manipulation or execution.

crates/accelerate/src/twirling.rs
mtreinish: I added control flow support in: https://github.com/Qiskit/qiskit/pull/13331/commits/2eff03a62c06c982c4d555cbf32141653e557227 I'm not too particularly worried about the open control case, it'll complicate the logic and seems unlikely to matter.

crates/accelerate/src/twirling.rs
mtreinish: Done in: https://github.com/Qiskit/qiskit/pull/13331/commits/a85ec8f077abfac17081637a210a97fb495ddb9d

crates/accelerate/src/twirling.rs
mtreinish: Done in: https://github.com/Qiskit/qiskit/pull/13331/commits/f70566df24e7e9f51a0e6aafe04d48f4a00ab4d6

crates/circuit/src/circuit_data.rs
mtreinish: Done in: https://github.com/Qiskit/qiskit/pull/13331/commits/7112f9e6d8e4b12570673dc11af7eb64179ac390

crates/accelerate/src/twirling.rs
mtreinish: Done in: https://github.com/Qiskit/qiskit/pull/13331/commits/0b022cd3904826456424e3efef65d05a6fd975a8

crates/circuit/src/circuit_data.rs
mtreinish: I added more docs for this in: https://github.com/Qiskit/qiskit/pull/13331/commits/d6ea01f2abedc8365cf91a116b0a7644f6486bff but if you think it needs more I can expand on it further.

qiskit/circuit/twirling.py
mtreinish: Done in: https://github.com/Qiskit/qiskit/pull/13331/commits/6d6ca436ef6bde07c85e35671685b73ba5406f00

crates/accelerate/src/twirling.rs
```diff
@@ -0,0 +1,384 @@
...
+                }
+            }
+            _ => {
+                out_circ.push(py, inst.clone())?;
+            }
+        }
+    }
+    if run_pass {
+        let mut dag = DAGCircuit::from_circuit_data(py, out_circ, false)?;
+        optimize_1q_gates_decomposition(
```
mtreinish: I opted to just call `Optimize1qGatesDecomposition` from rust here. You could potentially avoid scanning the entire dag to search for runs, because you can search from where we inject the pauli gates to find new runs created by the insertions, but the do that efficiently we'd have to convert to a dag anyway and the point we're doing that it doesn't make a huge difference and it's a lot simpler to just call the existing function.

crates/accelerate/src/twirling.rs
```diff
@@ -123,12 +128,59 @@ const CZ_MASK: u8 = 4;
...
+        (YGate, y_matrix),
+        (ZGate, z_matrix),
+    ];
+    for (i, i_mat) in &iter_set {
+        for (j, j_mat) in &iter_set {
+            let before_matrix = kron(j_mat, i_mat);
+            let half_twirled_matrix = gate_matrix.dot(&before_matrix);
+            for (k, k_mat) in &iter_set {
+                for (l, l_mat) in &iter_set {
+                    let after_matrix = kron(l_mat, k_mat);
```
jakelishman: If we store the 16 possible krons outside all the loops, we can avoid recalculating the same krons 16 times over in here.  It's a factor of 17 saving in the number of krons (though the 2q matmuls will still be far more expensive - kron is really SIMD and cache-line friendly, and it's a lot fewer scalar ops than a matmul).

crates/accelerate/src/twirling.rs
```diff
@@ -123,12 +128,59 @@ const CZ_MASK: u8 = 4;
...
+                    let twirled_matrix = after_matrix.dot(&half_twirled_matrix);
+                    let norm: f64 = diff_frob_norm(twirled_matrix.view(), gate_matrix);
+                    if norm.abs() < 1e-15 {
+                        out_vec.push(([*i, *j, *k, *l], 0.));
+                    } else if (norm - 4.).abs() < 1e-15 {
+                        out_vec.push(([*i, *j, *k, *l], PI));
+                    }
```
jakelishman: We don't actually need to use the proper Frobenius norm - we can use the square of the norm and avoid a call to `sqrt` (adjust the `norm - 4.` accordingly).

Really neither here nor there, but there's only four values `norm` could have that would satisfy the check on line 168 for that tolerance - `[4.-4.*eps, 4.-2.*eps, 4., 4.+4.*eps]` where `eps` is the float epsilon ($2^{-52}\approx 2.22\times 10^{-16}$).

jakelishman (commented):
I haven't looked at the new interface, I was just commenting my nerd-sniped bits on the performance of the new twirl finder lol

crates/accelerate/src/twirling.rs
```diff
@@ -0,0 +1,472 @@
...
+        None => {
+            if custom_twirled_gates.is_none() {
+                15
+            } else {
+                0
+            }
+        }
+    };
+    let custom_gate_twirling_sets: Option<HashMap<String, Vec<([StandardGate; 4], f64)>>> =
+        custom_twirled_gates.map(|gates| {
```
mtreinish, line 430: Oops, I forgot to add checking on the qubit count and parameters to ensure we only run this with 2q gates that don't take parameters.

crates/accelerate/src/twirling.rs
mtreinish: Done in: https://github.com/Qiskit/qiskit/pull/13331/commits/ca72bd0b8686e8c83d95bde1e1dc1fb7124fa895

crates/accelerate/src/twirling.rs
mtreinish: Done in: https://github.com/Qiskit/qiskit/pull/13331/commits/aaa62951e9ddf0fca358a269a463a3e2fd42daa7

crates/circuit/src/circuit_data.rs
```diff
@@ -1268,6 +1269,11 @@ impl CircuitData {
         self.qargs_interner().get(index)
     }
 
+    /// Insert qargs into the interner and return the interned value
+    pub fn set_qargs(&mut self, qubits: &[Qubit]) -> Interned<[Qubit]> {
```
Cryoris: Would `add_qargs` be more precise? `set_` sounds like it would overwrite the current values with what value is now given 🙂

qiskit/circuit/__init__.py
```diff
@@ -1047,6 +1047,23 @@ def __array__(self, dtype=None, copy=None):
...
+---------------------------------
+
+There are two primary types of noise when executing quantum circuits. The first is stochastic,
+or incoherent, noise that is mainly due to the unwanted interaction between the quantum processor
+and the external environment in which it resides. The second is known as coherent error, and these
+errors arise due to imperfect control of a quantum system. This can be unwanted terms in a system
+Hamiltonian, i.e. incorrect unitary evolution, or errors from incorrect temporal control of the
+quantum system, which includes things like incorrect pulse-shapes for gates.
+
+Pauli-twirling is a quantum error suppression technique that uses randomization to shape coherent
```
Cryoris: Minor nitpick to be consistent with other spellings used here 🙂 
```suggestion
Pauli twirling is a quantum error suppression technique that uses randomization to shape coherent
```

qiskit/circuit/twirling.py
```diff
@@ -0,0 +1,132 @@
...
+    gate.
+
+    Args:
+        circuit: The circuit to twirl
+        twirling_gate: The gate to twirl, defaults to `None` which means twirl all default gates:
+            :class:`.CXGate`, :class:`.CZGate`, :class:`.ECRGate`, and :class:`.iSwapGate`.
+            If supplied it can either be a single gate or a list of gates either as either a gate
+            object or it's string name. Currently only the names `"cx"`, `"cz"`, `"ecr"`,  and
+            `"iswap"` are supported. If a gate object is provided outside the default gates it must
+            have a matrix defined from it's :class:`~.Gate.to_matrix` method for the gate to potentially
```
Cryoris: I'll naturally yield to english-native speakers, but shouldn't this be the following?
```suggestion
            object or its string name. Currently only the names `"cx"`, `"cz"`, `"ecr"`,  and
            `"iswap"` are supported. If a gate object is provided outside the default gates it must
            have a matrix defined from its :class:`~.Gate.to_matrix` method for the gate to potentially
```

crates/accelerate/src/twirling.rs
```diff
@@ -0,0 +1,484 @@
...
+
+fn generate_twirling_set(gate_matrix: ArrayView2<Complex64>) -> Vec<([StandardGate; 4], f64)> {
+    let mut out_vec = Vec::with_capacity(16);
+    let i_matrix = aview2(&ONE_QUBIT_IDENTITY);
+    let x_matrix = aview2(&qiskit_circuit::gate_matrix::X_GATE);
+    let y_matrix = aview2(&qiskit_circuit::gate_matrix::Y_GATE);
+    let z_matrix = aview2(&qiskit_circuit::gate_matrix::Z_GATE);
+    let iter_set = [IGate, XGate, YGate, ZGate];
+    let kron_set: [Array2<Complex64>; 16] = [
+        kron(&i_matrix, &i_matrix),
```
Cryoris, line 153: This one is redundant to check if you want to directly push it onto `out_vec` 🙂

crates/accelerate/src/twirling.rs
```diff
@@ -0,0 +1,484 @@
...
+            let before_matrix = kron_set[i_idx * 4 + j_idx].view();
+            let half_twirled_matrix = gate_matrix.dot(&before_matrix);
+            for (k_idx, k) in iter_set.iter().enumerate() {
+                for (l_idx, l) in iter_set.iter().enumerate() {
+                    let after_matrix = kron_set[k_idx * 4 + l_idx].view();
+                    let twirled_matrix = after_matrix.dot(&half_twirled_matrix);
+                    let norm: f64 = diff_frob_norm_sq(twirled_matrix.view(), gate_matrix);
+                    if norm.abs() < 1e-15 {
+                        out_vec.push(([*i, *j, *k, *l], 0.));
+                    } else if (norm - 16.).abs() < 1e-15 {
```
Cryoris, line 181: This looks like something that could give a false positive, i.e. some custom defined gate where the difference is 16 yet they differ by a relative phase or something... or is that not possible?

test/python/circuit/test_twirling.py
```diff
@@ -0,0 +1,212 @@
...
+            Operator(qc), Operator(res), err_msg=f"gate: {gate} not equiv to\n{res}"
+        )
+
+    @ddt.data(CXGate, ECRGate, CZGate, iSwapGate)
+    def test_control_flow(self, gate):
+        """Test we twirl inside control flow blocks."""
+        qc = QuantumCircuit(2, 1)
+        with qc.if_test((qc.clbits[0], 0)):
+            qc.append(gate(), [0, 1])
+        res = twirl_circuit(qc)
```
Cryoris: This should maybe also check that the circuits are not equal (maybe with a seed to avoid randomly hitting the identity-twirl case)

Cryoris (commented):
Just some minor questions, otherwise LGTM

crates/circuit/src/circuit_data.rs
mtreinish: I thought `set_qargs` was used already in `DAGCircuit`. But I don't see it in there. It made sense to me because this pairs with the `get_qargs` method used for looking up from the interner. But I can change it if you prefer `add_qargs`.

crates/accelerate/src/twirling.rs
mtreinish: Yeah, that's a good call. I guess the question is should we still do that if the output is otherwise empty? This function will always return 4 identity gates even if there is no other valid twirling combination.

I'm thinking we should probably error if 4 identities are all that matches so only add it to the output list if there are other matches.

crates/circuit/src/circuit_data.rs
Cryoris: Yeah I think `add` or `insert` would be more intuitive 🙂

crates/accelerate/src/twirling.rs
Cryoris: Yeah an error (or at least a warning) would be good to raise to let users know the code didn't work as expected 👍🏻

crates/accelerate/src/twirling.rs
Cryoris: Thanks to @anedumla I now believe this is true 😬

crates/circuit/src/circuit_data.rs
mtreinish: Done in: https://github.com/Qiskit/qiskit/pull/13331/commits/46ac27320c20cbc2ab5461b02e1e1e8c5be07679

Cryoris (approved):

PR #13335, opened 2024-10-16
Add new API for inserting multiple operations into the `DAGCircuit` in Rust.

raynelfss:
### Summary
Adds a new struct `DAGCircuitBuilder` (name has changed 😄) that enables chained addition of instructions both valid and invalid into the `DAGCircuit`. Also, adapts existing structures to those new changes.

### Details and comments
- Add struct that allows continuous addition of instructions into the back of the `DAGCircuit` by performing edge modifications in the output nodes at the end of the additions.
- The new struct includes the following new methods:
    - `push_back` which pushes a valid `PackedInstruction` into the back of the `DAGCircuit`. It behaves very similarly to `push_back`.
    - `apply_operation_back` which inserts a new operation into the back of the `DAGCircuit`. ~It has two variants which depend on the ownership of the bit indices (`Qubit`, `Clbit`)~.
    - `pack_instruction`: creates a valid instance of `PackedInstruction` based on other provided attributes.
- Modified `DAGCircuit::extend` to use this new API, without any considerable losses.

### What about front additions?
This struct currently does not support adding instructions continuously to the front of the `DAGCircuit` because the methodology for that is a little different. It requires doing much more shifting than adding to the back, therefore I suggest studying the commonality of the use case and checking if it could be done in a follow-up.

### Questions and comments?
Feel free to add these below as a review or comment 😸


raynelfss:
After rewriting the logic of the `BasisTranslator` pass to use this new API, there was some minor speedup within the pass:

| Change   | Before [f3bac5a0] <main>   | After [c91d45bd] <dag-concatenate>   |   Ratio | Benchmark (Parameter)                                                                                    |
|----------|----------------------------|--------------------------------------|---------|----------------------------------------------------------------------------------------------------------|
| -        | 17.0±0.1ms                 | 15.3±0.2ms                           |    0.9  | passes.MultipleBasisPassBenchmarks.time_basis_translator(5, 1024, ['rz', 'x', 'sx', 'cx', 'id'])         |
| -        | 15.1±0.3ms                 | 13.0±0.2ms                           |    0.86 | passes.MultipleBasisPassBenchmarks.time_basis_translator(5, 1024, ['u', 'cx', 'id'])                     |
| -        | 42.8±0.8ms                 | 36.4±0.4ms                           |    0.85 | passes.MultipleBasisPassBenchmarks.time_basis_translator(14, 1024, ['rz', 'x', 'sx', 'cx', 'id'])        |
| -        | 36.7±0.3ms                 | 30.9±0.2ms                           |    0.84 | passes.MultipleBasisPassBenchmarks.time_basis_translator(14, 1024, ['u', 'cx', 'id'])                    |
| -        | 51.1±0.4ms                 | 42.9±0.3ms                           |    0.84 | passes.MultipleBasisPassBenchmarks.time_basis_translator(20, 1024, ['u', 'cx', 'id'])                    |
| -        | 21.6±0.1ms                 | 17.7±0.1ms                           |    0.82 | passes.MultipleBasisPassBenchmarks.time_basis_translator(5, 1024, ['rx', 'ry', 'rz', 'r', 'rxx', 'id'])  |
| -        | 59.3±0.9ms                 | 48.0±0.2ms                           |    0.81 | passes.MultipleBasisPassBenchmarks.time_basis_translator(20, 1024, ['rz', 'x', 'sx', 'cx', 'id'])        |
| -        | 81.3±0.3ms                 | 64.0±1ms                             |    0.79 | passes.MultipleBasisPassBenchmarks.time_basis_translator(20, 1024, ['rx', 'ry', 'rz', 'r', 'rxx', 'id']) |
| -        | 59.2±1ms                   | 45.8±0.6ms                           |    0.77 | passes.MultipleBasisPassBenchmarks.time_basis_translator(14, 1024, ['rx', 'ry', 'rz', 'r', 'rxx', 'id']) |

crates/accelerate/src/unitary_synthesis.rs
```diff
@@ -111,16 +111,31 @@ fn apply_synth_dag(
...
+                synth_dag
+                    .cargs_interner()
+                    .get(out_packed_instr.clbits)
+                    .into(),
+            ),
+            out_packed_instr.params.clone().map(|param| *param),
+            out_packed_instr.extra_attrs.clone(),
+            #[cfg(feature = "cache_pygates")]
+            None,
+        );
```
jakelishman: On line 116: the `.dag()` call is not necessary any more (probably happened in a merge conflict).

For the whole thing: the resulting code here has ended up a lot more complex than the previous code - perhaps this was before you added the `push_back` method to your new struct?  I'd consider swapping back, even if you need a bit of extra boilerplate to expose the interners in the same way (or at least the relevant `Interner` methods).

crates/circuit/src/dag_circuit.rs
```diff
@@ -6974,6 +6861,271 @@ impl DAGCircuit {
...
+            .into_iter()
+            .enumerate()
+            .filter_map(|(var, node)| node.map(|node| (var, node)))
+        {
+            let output_node = self.dag.var_io_map[var][1];
+            self.dag
+                .dag
+                .add_edge(node, output_node, Wire::Var(Var(var as u32)));
+        }
+    }
```
jakelishman: The worry here is that if you forget to call this method, then the DAG is completely invalidated.  Making this the `Drop` implementation (the destructor) is better, but destructors aren't guaranteed to run in safe Rust either, so while you do have to try harder to forget, it's still very possible.

crates/circuit/src/dag_circuit.rs
```diff
@@ -6974,6 +6861,271 @@ impl DAGCircuit {
...
+            op,
+            qubits,
+            clbits,
+            params,
+            extra_attrs,
+            #[cfg(feature = "cache_pygates")]
+            py_op,
+        );
+        self.push_operation_back(py, instruction)
+    }
```
jakelishman: I'm not sure this offers much over making `pack_instruction` public and only exposing `push_operation_back` - the output of `pack_instruction` is already effectively public because you can retrieve the `PackedInstruction` by indexing the DAG, so I'm not sure we're gaining anything by making it private.

crates/circuit/src/dag_circuit.rs
```diff
@@ -6974,6 +6861,271 @@ impl DAGCircuit {
...
+
+    /// Pushes a valid [PackedInstruction] to the back ot the circuit.
+    pub fn push_operation_back(&mut self, py: Python<'_>, inst: PackedInstruction) -> NodeIndex {
+        let op_name = inst.op.name();
+        let (all_cbits, vars): (Vec<Clbit>, Option<Vec<Var>>) = {
```
jakelishman: There are allocation inefficiencies in this function that we can definitely improve on with this new struct, but better to leave them for a follow-up, I think.

crates/circuit/src/dag_circuit.rs
```diff
@@ -171,6 +171,23 @@ impl Wire {
...
+    fn from(value: &'a [T]) -> Self {
+        Self::Slice(value)
+    }
+}
+
+impl<'a, T> From<Vec<T>> for OwnedOrSlice<'a, T> {
+    fn from(value: Vec<T>) -> Self {
+        Self::Owned(value)
+    }
+}
```
jakelishman: If we do decide to keep the functionality, this enum is actually just a concrete case of something that exists in the standard library already, and will play nicely with the interners: `std::borrow::Cow<'a, [T]>`.

crates/circuit/src/dag_circuit.rs
```diff
@@ -6974,6 +6861,271 @@ impl DAGCircuit {
...
+        PackedInstruction {
+            op,
+            qubits,
+            clbits,
+            params: params.map(|x| x.into()),
+            extra_attrs,
+            #[cfg(feature = "cache_pygates")]
+            py_op,
+        }
+    }
```
jakelishman: While the interners do allow for switching on whether the input is owned or not, I think it's worth checking if there's actually all that much benefit to exposing it in a dynamic manner like this.  We'll be hitting the "not present" path really infrequently over all (the vast majority of insertions are cache hits), so there'll be very few actual allocations, whereas moving things in and out of the `Cow` (or `SliceOrOwned`) has a cost on every call.

I doubt `Cow` will be noticeably slower, but I'd also doubt that it'll be notably faster, and it might not be worth the code complexity at the call sites.

crates/accelerate/src/unitary_synthesis.rs
```diff
@@ -111,16 +111,31 @@ fn apply_synth_dag(
     out_qargs: &[Qubit],
     synth_dag: &DAGCircuit,
 ) -> PyResult<()> {
+    let mut out_dag_concat = out_dag.as_concat();
     for out_node in synth_dag.topological_op_nodes()? {
```
jakelishman: This function (`apply_synth_dag`) doesn't feel like it's high enough in the call tree, because we're allocating and de-allocating the concatenator for each synthesised unitary matrix, whereas in an ideal world we'd do that only once. Perhaps we should be passing the concatenator through the function calls?

jakelishman (commented):
This is a very neat idea, and something I've been interested in in the past.  Aside from the inline comments, I have one main consideration (as two chained questions):

1. Is there any way we can get by with just `extend` as is?  `UnitarySynthesis` was doing that, so can `BasisTranslator` be re-organised to do it too?  There are ways around the set-up overhead of the method.
2. Failing that: I mentioned in one of the comments that there's a big unsafety here in that if you forget to call `DAGCircuitConcat::end` but just let the concatenator go out of scope, the `DAG` is now irreparably broken (from the caller's perspective).  We definitely need to make that harder to do, ideally impossible.  Making `end` the destructor makes it harder to make the mistake, but not impossible.  I note that the only places we call `extend` (or the concatenator) are in situations where we're doing a full DAG rebuild, so we could actually take ownership of the `DAGCircuit`.  In that situation, now you can have `DAGCircuitConcat` take ownership, and `end` return the ownership, so the caller can't ever observe the `DAGCircuit` in a broken state.

crates/circuit/src/dag_circuit.rs
raynelfss: Oh, I didn't know this existed already. That's a relief!

crates/accelerate/src/unitary_synthesis.rs
raynelfss: Addressed and fixed in eb0a92f

crates/circuit/src/dag_circuit.rs
raynelfss: This, and in consideration to parts of your main review comment:
> I mentioned in one of the comments that there's a big unsafety here in that if you forget to call DAGCircuitConcat::end but just let the concatenator go out of scope, the DAG is now irreparably broken (from the caller's perspective). We definitely need to make that harder to do, ideally impossible. Making end the destructor makes it harder to make the mistake, but not impossible. I note that the only places we call extend (or the concatenator) are in situations where we're doing a full DAG rebuild, so we could actually take ownership of the DAGCircuit. In that situation, now you can have DAGCircuitConcat take ownership, and end return the ownership, so the caller can't ever observe the DAGCircuit in a broken state.

The new iteration of `DAGCircuitConcat` now takes ownership of the `DAGCircuit` instance and only exposes certain necessary elements (such as the interners and the circuit's global phase). This also involved using a hacky way of still keeping ownership during the calls to `DAGCircuit::extend` which I'd like to know your thoughts on.

All of these changes were introduced in 6735e97 and bf26f56

crates/circuit/src/dag_circuit.rs
raynelfss: Hmm... sounds good, but I did want to keep an effectively similar method to `DAGCircuit::apply_operation_back`. I think that both exposing the `pack_instruction` method and keeping the `apply_operation_back` would be a viable option.

crates/circuit/src/dag_circuit.rs
raynelfss: These now use the `Interner::insert_cow()` method that you designed :)

crates/accelerate/src/unitary_synthesis.rs
raynelfss: The call to `out_dag.as_concat()` was moved higher up in the stack in bf26f56.

crates/circuit/src/dag_circuit.rs
```diff
@@ -6534,146 +6535,15 @@ impl DAGCircuit {
...
-
+        let mut new_nodes = Vec::new();
+        // TODO: Find a less hacky way of doing this
+        let mut replacement_dag = DAGCircuit::new()?;
+        std::mem::swap(self, &mut replacement_dag);
+        let mut dag_concat = replacement_dag.into_concat();
+        for inst in iter {
+            new_nodes.push(dag_concat.push_instruction_back(py, inst?));
+        }
+        std::mem::swap(self, &mut dag_concat.end());
```
kevinhartman: I think this is the only way to do this. If we implement `Default` for `DAGCircuit`, then you could use `mem::take` instead and then `*self = dag_concat.end()` at the end. But it'd be doing the same thing, it'd just look less noticeable that a dummy value is in play here.

crates/circuit/src/dag_circuit.rs
```diff
@@ -7095,6 +6965,273 @@ impl DAGCircuit {
...
     }
+
+    /// Returns a concatenable version of the DAGCircuit. Preferably used
+    /// for adding multiple new instructions to the [DAGCircuit].
+    pub fn into_concat(self) -> DAGCircuitConcat {
+        DAGCircuitConcat::new(self)
+    }
+}
+
+pub struct DAGCircuitConcat {
```
kevinhartman: I think the term `Concat` isn't perfectly descriptive (not that naming things always can be!), since concatenation is generally a term for the joining of two sequences. Personally, I like `Writer` or `Builder` better, since they're a little more general.

What do you think about the names `DAGCircuitBuilder` and `DAGCircuit::into_builder`?

crates/circuit/src/dag_circuit.rs
```diff
@@ -7095,6 +6965,273 @@ impl DAGCircuit {
...
+    /// be added continuously into the [DAGCircuit].
+    pub fn new(dag: DAGCircuit) -> DAGCircuitConcat {
+        let num_qubits = dag.num_qubits();
+        let num_clbits = dag.num_clbits();
+        let num_vars = dag.num_vars();
+        Self {
+            dag,
+            last_qubits: (0..num_qubits).map(|_| None).collect(),
+            last_clbits: (0..num_clbits).map(|_| None).collect(),
+            last_vars: (0..num_vars).map(|_| None).collect(),
```
```suggestion
            last_qubits: vec![None; num_qubits],
            last_clbits: vec![None; num_clbits],
            last_vars: vec![None; num_vars],
```

How's this? 🙂

crates/circuit/src/dag_circuit.rs
```diff
@@ -7095,6 +6965,273 @@ impl DAGCircuit {
...
+        }
+    }
+
+    /// Finishes up the changes by re-connecting all of the ouput nodes back to the last
```
```suggestion
    /// Finishes up the changes by re-connecting all of the output nodes back to the last
```

crates/circuit/src/dag_circuit.rs
```diff
@@ -197,9 +198,9 @@ pub struct DAGCircuit {
     cregs: RegisterData<ClassicalRegister>,
 
     /// The cache used to intern instruction qargs.
-    pub qargs_interner: Interner<[Qubit]>,
+    qargs_interner: Interner<[Qubit]>,
```
kevinhartman, line 159: I agree that these should not be mutable, the reason being that it makes it possible to modify a `DAGCircuit` into an invalid state, e.g. we could `std::mem::swap` an empty `Interner` over this one, which opens us up to crashes and unpredictable behavior.

The `pub` was originally needed for #13141, which makes the valid point that they're useful when building `PackedInstruction` instances. You are still exposing them mutably from `DAGCircuitConcat`, which is moving the problem.

It may be beyond the scope of this PR to address this.

crates/circuit/src/dag_circuit.rs
```diff
@@ -7095,6 +6965,273 @@ impl DAGCircuit {
...
+            params,
+            label,
+            #[cfg(feature = "cache_pygates")]
+            py_op,
+        );
+        self.push_instruction_back(py, instruction)
+    }
+
+    /// Pushes a valid [PackedInstruction] to the back ot the circuit.
+    pub fn push_instruction_back(&mut self, py: Python<'_>, inst: PackedInstruction) -> NodeIndex {
```
kevinhartman: To be consistent with the `DAGCircuit` interface, should this be named `push_back` instead?

crates/circuit/src/dag_circuit.rs
```diff
@@ -7095,6 +6965,273 @@ impl DAGCircuit {
...
+        let op_name = inst.op.name();
+        let (all_cbits, vars): (Vec<Clbit>, Option<Vec<Var>>) = {
+            if self.dag.may_have_additional_wires(py, &inst) {
+                let mut clbits: HashSet<Clbit> =
+                    HashSet::from_iter(self.dag.cargs_interner.get(inst.clbits).iter().copied());
+                let (additional_clbits, additional_vars) = self
+                    .dag
+                    .additional_wires(py, inst.op.view())
+                    .expect("Error trying to find additional wires for instruction");
```
kevinhartman: Is there a reason not to return a `PyResult` from this method?

crates/circuit/src/dag_circuit.rs
kevinhartman: It looks like this function duplicates a lot of what's happening in `DAGCircuit::push_back`, with the exception of course of adding back the output edges.

Would it be possible to create a private function on `DAGCircuit` that we could call both from the implementation of `DAGCircuit::push_back` and `DAGCircuitConcat::push_instruction_back` that does the shared bits of logic?

crates/circuit/src/dag_circuit.rs
```diff
@@ -7095,6 +6965,273 @@ impl DAGCircuit {
...
+        &self.dag.qargs_interner
+    }
+
+    /// Returns an immutable view to the clbit interner
+    pub fn cargs_interner(&self) -> &Interner<[Clbit]> {
+        &self.dag.cargs_interner
+    }
+
+    /// Returns a mutable view to the qubit interner
+    pub fn qargs_interner_mut(&mut self) -> &mut Interner<[Qubit]> {
```
kevinhartman: As mentioned higher up, these still leak a mutable `Interner` from the `DAGCircuit` (probably something we should address separately).

kevinhartman (commented):
This looks good overall.

A few comments. I think we should probably move away from having the interners `pub` in a future PR, in favor of some way of accessing them that doesn't allow the user to replace them (i.e. make sure their semantics are append-only).

crates/circuit/src/dag_circuit.rs
raynelfss: For now, I can change the mutable views into just `insert_cargs` and `insert_qargs`, since it is the only way they're used.

crates/circuit/src/dag_circuit.rs
raynelfss: There isn't a reason currently, adding...

crates/circuit/src/dag_circuit.rs
raynelfss: I will add a `_push_back` method that does the first steps, since in the end we don't perform the same operations for modifying the nodes.

crates/accelerate/src/basis/basis_translator/mod.rs
```diff
@@ -460,7 +461,8 @@ fn apply_translation(
     >,
 ) -> PyResult<(DAGCircuit, bool)> {
     let mut is_updated = false;
-    let mut out_dag = dag.copy_empty_like(py, "alike")?;
+    let out_dag = dag.copy_empty_like(py, "alike")?;
+    let mut out_dag_concat = out_dag.into_builder();
```
kevinhartman: `out_dag_builder`? :3

crates/accelerate/src/basis/basis_translator/mod.rs
```diff
@@ -594,13 +576,13 @@ fn apply_translation(
         }
         is_updated = true;
     }
-
-    Ok((out_dag, is_updated))
+    // Kill the concatenator
```
kevinhartman:
```suggestion
```

crates/accelerate/src/unitary_synthesis.rs
```diff
@@ -374,10 +371,11 @@ fn py_run_main_loop(
                     PhysicalQubit::new(qubit_indices[out_qargs[0].0 as usize] as u32),
                     PhysicalQubit::new(qubit_indices[out_qargs[1].0 as usize] as u32),
                 ];
-                let apply_original_op = |out_dag: &mut DAGCircuit| -> PyResult<()> {
+                let apply_original_op = |out_dag: &mut DAGCircuitBuilder| -> PyResult<()> {
                     out_dag.push_back(py, packed_instr.clone())?;
                     Ok(())
                 };
+                let mut concatenable_dag = out_dag.into_builder();
```
kevinhartman:
```suggestion
                let mut builder = out_dag.into_builder();
```

crates/circuit/src/dag_circuit.rs
kevinhartman: Nice, it looks like you were able to encapsulate this again. We'll see if it stays that way, but glad to see this!

crates/circuit/src/dag_circuit.rs
```diff
@@ -6404,146 +6438,15 @@ impl DAGCircuit {
...
-        for item in vars_last_nodes.items() {
-            let (var, node): (PyObject, usize) = item.extract()?;
-            let var = self.vars.find(&var.bind(py).into()).unwrap();
-            let output_node = self.var_io_map.get(var.index()).unwrap()[1];
-            self.dag
-                .add_edge(NodeIndex::new(node), output_node, Wire::Var(var));
-        }
-
+        let mut new_nodes = Vec::new();
+        // TODO: Find a less hacky way of doing this
```
```suggestion
```

crates/circuit/src/dag_circuit.rs
```diff
@@ -7232,6 +7135,252 @@ impl DAGCircuit {
...
+
+    /// Returns an immutable view to the qubit interner
+    pub fn qargs_interner(&self) -> &Interner<[Qubit]> {
+        &self.dag.qargs_interner
+    }
+
+    /// Returns an immutable view to the clbit interner
+    pub fn cargs_interner(&self) -> &Interner<[Clbit]> {
+        &self.dag.cargs_interner
+    }
```
kevinhartman, line 7430: I recognize we have these very same methods on `DAGCircuit` which expose a non-mutable ref to the `Interner`s, but it feels a bit odd to me that we use wrapper methods for the mutable stuff (e.g. `insert_qargs`) but expose the interners directly for read-only stuff. It'd be nice if everything was at the same level, i.e. either wrapper methods `get_qargs` and `insert_qargs`, or expose the interners directly (which of course has the mutability issue).

What you've got is probably fine for now, but IMO the `InternerViewMut` stuff you mentioned offline might be a nice way to unify things in a future PR.

crates/circuit/src/dag_circuit.rs
```diff
@@ -4996,6 +4991,45 @@ impl DAGCircuit {
...
 
+    fn _push_back(&mut self, py: Python, instr: PackedInstruction) -> PyResult<DAGNodeInfo> {
+        let op_name = instr.op.name();
+        let (all_clbits, vars): (Vec<Clbit>, Option<Vec<Var>>) = {
+            if self.may_have_additional_wires(py, &instr) {
+                let mut clbits: HashSet<Clbit> =
+                    HashSet::from_iter(self.cargs_interner.get(instr.clbits).iter().copied());
+                let (additional_clbits, additional_vars) = self
+                    .additional_wires(py, instr.op.view())
+                    .expect("Error trying to find additional wires for instruction");
```
kevinhartman: Would be good to propagate via `?` instead

crates/circuit/src/dag_circuit.rs
```diff
@@ -4996,6 +4991,45 @@ impl DAGCircuit {
...
+                }
+                let additional_vars = additional_vars
+                    .into_iter()
+                    .map(|var| self.vars.find(&var.bind(py).into()).unwrap())
+                    .collect();
+                (clbits.into_iter().collect(), Some(additional_vars))
+            } else {
+                (self.cargs_interner.get(instr.clbits).to_vec(), None)
+            }
+        };
```
kevinhartman: Instead of `_push_back`, I think it might be better to just pull this chunk out into its own method, `get_classical_resources(&PackedInstruction) -> (Vec<Clbit>, Option<Vec<Var>>)` and call it from `DAGCircuit::push_back` and `DAGCircuitBuilder::push_back`. The only other thing `_push_back` is doing (other than packing up where it left off) is incrementing the `op` count, which is fine to duplicate between the two `push_back`s, IMO.

kevinhartman (commented):
Looking good, just a few more minor comments. Thanks for the changes!

crates/circuit/src/dag_circuit.rs
```diff
@@ -4947,7 +4947,9 @@ impl DAGCircuit {
     /// another that was created from the first via
     /// [DAGCircuit::copy_empty_like].
     pub fn push_back(&mut self, py: Python, instr: PackedInstruction) -> PyResult<NodeIndex> {
-        let initial_parse = self._push_back(py, instr)?;
+        // Increment the operation count
+        self.increment_op(instr.op.name());
+        let initial_parse = self.get_classical_resources(py, instr)?;
```
kevinhartman: It looks to me like `self.get_classical_resources` is the only part of this method that can fail, so it might be better to increment the op count only afterwards instead. Probably not super important since if we error, the state of the `DAGCircuit` shouldn't be assumed sound. But it might be good practice at least.

crates/circuit/src/dag_circuit.rs
```diff
@@ -4996,6 +4993,44 @@ impl DAGCircuit {
...
+
+        let result = DAGNodeInfo {
+            all_clbits,
+            vars,
+            node: self.dag.add_node(NodeType::Operation(instr)),
+            qubits_id,
+        };
+
+        // Insert op-node to graph.
+        Ok(result)
```
kevinhartman: I was thinking that this stuff should not be part of `get_classical_resources`. This method should IMO just return `(all_clbits, vars)` and should probably take `instr` by reference.

(I don't think we need `DAGNodeInfo` at all anymore if we do that.)

crates/accelerate/src/unitary_synthesis.rs
```diff
@@ -409,7 +408,9 @@ fn py_run_main_loop(
                         None,
                     )?;
                     let out_qargs = dag.get_qargs(packed_instr.qubits);
-                    apply_synth_dag(py, &mut out_dag, out_qargs, &synth_dag)?;
+                    let mut concatenable_dag = out_dag.into_builder(py);
+                    apply_synth_dag(py, &mut concatenable_dag, out_qargs, &synth_dag)?;
+                    out_dag = concatenable_dag.end();
```
kevinhartman:
```suggestion
                    let mut dag_builder = out_dag.into_builder(py);
                    apply_synth_dag(py, &mut dag_builder, out_qargs, &synth_dag)?;
                    out_dag = dag_builder.end();
```

crates/accelerate/src/unitary_synthesis.rs
```diff
@@ -1073,8 +1074,9 @@ fn reversed_synth_su4_dag(
         unreachable!("reversed_synth_su4_dag should only be called for XXDecomposer")
     };
 
-    let mut target_dag = synth_dag.copy_empty_like(py, "alike")?;
+    let target_dag = synth_dag.copy_empty_like(py, "alike")?;
     let flip_bits: [Qubit; 2] = [Qubit(1), Qubit(0)];
+    let mut target_dag_concat = target_dag.into_builder(py);
```
kevinhartman:
```suggestion
    let mut target_dag_builder = target_dag.into_builder(py);
```

crates/accelerate/src/unitary_synthesis.rs
```diff
@@ -1083,10 +1085,10 @@ fn reversed_synth_su4_dag(
             .iter()
             .map(|x| flip_bits[x.0 as usize])
             .collect();
-        inst.qubits = target_dag.qargs_interner.insert_owned(qubits);
-        target_dag.push_back(py, inst)?;
+        inst.qubits = target_dag_concat.insert_qargs(qubits.into());
+        target_dag_concat.push_back(py, inst)?;
```
kevinhartman:
```suggestion
        inst.qubits = target_dag_builder.insert_qargs(qubits.into());
        target_dag_builder.push_back(py, inst)?;
```

crates/accelerate/src/unitary_synthesis.rs
```diff
@@ -1083,10 +1085,10 @@ fn reversed_synth_su4_dag(
...
     }
-    Ok(target_dag)
+    Ok(target_dag_concat.end())
```
kevinhartman:
```suggestion
    Ok(target_dag_builder.end())
```

crates/circuit/src/dag_circuit.rs
```diff
@@ -5091,6 +5078,33 @@ impl DAGCircuit {
         Ok(new_node)
     }
 
+    fn get_classical_resources(
+        &mut self,
```
kevinhartman: Can this be `&self` now?

crates/circuit/src/dag_circuit.rs
```diff
@@ -5091,6 +5078,33 @@ impl DAGCircuit {
...
+                let additional_vars = additional_vars
+                    .into_iter()
+                    .map(|var| self.vars.find(&var.bind(py).into()).unwrap())
+                    .collect();
+                (clbits.into_iter().collect(), Some(additional_vars))
+            } else {
+                (self.cargs_interner.get(instr.clbits).to_vec(), None)
+            }
+        };
+        // Insert op-node to graph.
```
```suggestion
```

crates/circuit/src/dag_circuit.rs
```diff
@@ -6497,146 +6511,14 @@ impl DAGCircuit {
...
-        }
-
+        let mut new_nodes = Vec::new();
+        let mut replacement_dag = DAGCircuit::new()?;
+        std::mem::swap(self, &mut replacement_dag);
+        let mut dag_concat = replacement_dag.into_builder(py);
+        for inst in iter {
+            new_nodes.push(dag_concat.push_back(py, inst?)?);
+        }
+        std::mem::swap(self, &mut dag_concat.end());
```
```suggestion
        let mut dag_builder = replacement_dag.into_builder(py);
        for inst in iter {
            new_nodes.push(dag_builder.push_back(py, inst?)?);
        }
        std::mem::swap(self, &mut dag_builder.end());
```

crates/circuit/src/dag_circuit.rs
```diff
@@ -7317,6 +7199,251 @@ impl DAGCircuit {
...
+
+pub struct DAGCircuitBuilder {
+    dag: DAGCircuit,
+    last_clbits: Vec<Option<NodeIndex>>,
+    last_qubits: Vec<Option<NodeIndex>>,
+    last_vars: Vec<Option<NodeIndex>>,
+}
+
+impl DAGCircuitBuilder {
+    /// Builds a new instance of [DAGCircuitConcat] which allows instructions to
```
```suggestion
    /// Creates a new instance of [DAGCircuitBuilder] which allows instructions to
```

crates/circuit/src/dag_circuit.rs
```diff
@@ -7317,6 +7199,251 @@ impl DAGCircuit {
...
+            dag,
+            last_qubits: vec![None; num_qubits],
+            last_clbits: vec![None; num_clbits],
+            last_vars: vec![None; num_vars],
+        }
+    }
+
+    /// Finishes up the changes by re-connecting all of the output nodes back to the last
+    /// recorded nodes.
+    pub fn end(mut self) -> DAGCircuit {
```
kevinhartman: I didn't think of this originally, but now that we're calling this `DAGCircuitBuilder`, the canonical name for the method that produces the final object from a builder is `build`. It might help with API discoverability if we name this method that, but if you'd prefer to keep it as `end` I'm fine with that too.


```suggestion
    pub fn build(mut self) -> DAGCircuit {
```

kevinhartman (commented):
Almost there, just a few minor comments, mostly just correcting old variable names.

crates/circuit/src/dag_circuit.rs
raynelfss: Fixed in d128949

kevinhartman (approved):
Looks good to me!

Thanks for doing this, and for the changes. This is a neat tool.

PR #13336, opened 2024-10-16
Combine `impl CircuitData` blocks

mtreinish:
### Summary

The `CircuitData` struct had two `impl CircuitData` blocks, one before the pymethods and the other after. While there is nothing wrong with doing this it is a bit confusing because the difference between the blocks is not clear. I expect this was caused by the size of the file and people missed that it already existed and we ended up with duplicates. This commit recombines the two blocks, opting to use the one after the pymethods as it was larger.

### Details and comments

### DIFF
--- a/crates/circuit/src/circuit_data.rs
+++ b/crates/circuit/src/circuit_data.rs
@@ -106,298 +106,6 @@ pub struct CircuitData {
     global_phase: Param,
 }
 
-impl CircuitData {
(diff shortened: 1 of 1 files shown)


raynelfss (approved):
This makes sense to me, I assume the multiple `impl` were more contextual to group similar functions together. This just feels a bit more organized, so thank you for this 🚀

PR #13337, opened 2024-10-16
Fix performance regression in looped `QuantumCircuit.assign_parameters`

jakelishman:
### Summary

When calling `assign_parameters` on a heavily parametric circuit with the unusual access pattern of binding off a single parameter at a time, Qiskit 1.2 had a severe performance regression compared to Qiskit 1.1 stemming from gh-12794.  The calls to `unsorted_parameters` on each iteration were creating a new `set`, which could be huge if the number of parameters in the circuit was large.  In Qiskit 1.1 and before, that object was a direct view onto the underlying `ParameterTable` (assuming the input circuit did not have a parametric global phase), so was free to construct.

### Details and comments

For example, using a circuit like
```python
from qiskit.circuit import QuantumCircuit, Parameter

def make(num_qubits, num_reps):
    initials = [Parameter(str(i)) for i in range(num_qubits * num_reps)]
    qc = QuantumCircuit(num_qubits)
    ps = iter(initials)
    for _ in range(num_reps):
        for q in qc.qubits:
            qc.rz(next(ps), q)
    return qc

qc = make(127, 30)
```
and the not-so-great calling convention
```python
for p in list(qc.parameters):
    qc.assign_parameters({p: Parameter(p.name + "x")}, inplace=True)
```
Qiskit 1.1.2 took about 60ms, Qiskit 1.2.4 took about 1.1s, and this commit takes about 50ms.  This commit is dominated by the Python-space calls dealing with `Parameter` objects.

This is probably stable for backport, but since the immediate problem the regression was causing was mitigated by other means, and since it substantially touches a core method, I'd propose leaving this til 1.3.


qiskit/circuit/quantumcircuit.py
```diff
@@ -4348,38 +4351,57 @@ def assign_parameters(  # pylint: disable=missing-raises-doc
...
-            parameter_binds = _ParameterBindsDict(raw_mapping, our_parameters)
-            target._data.assign_parameters_mapping(parameter_binds)
+
+            def create_mapping_view():
+                return raw_mapping
+
+            target._data.assign_parameters_mapping(raw_mapping)
         else:
-            parameter_binds = _ParameterBindsSequence(target._data.parameters, parameters)
+            # This should be cause a cache retrieval, since we warmed the cache.  We need to keep
```
kevinhartman:
```suggestion
            # This should cause a cache retrieval, since we warmed the cache.  We need to keep
```

kevinhartman: Also, to make sure I understand, we want to save a reference to the initial parameters since `target._data.parameters` will be updated as we go? I think the code comment here might benefit from some more clarification. I'm not sure I understand the bit about us warming the cache. The `mapping_view` is what is cached to my knowledge, not the `initial_parameters`?

kevinhartman (commented):
This looks mostly good to me, but it might be nice if you could add a bit more explanation around `create_mapping_view` in the form of a code comment(s). Personally, I'm not used to seeing `nonlocal` and I had to look up how it works / its lifetime implications.

qiskit/circuit/quantumcircuit.py
jakelishman: I expanded all the comments in fcc0c43999.

The general idea is that if there's calibrations, Python space needs to construct a `{parameter: bound_value}` mapping if we're called in the sequence form.  However, for large numbers of parameters, that's quite expensive to construct, and the vast vast vast majority of circuits don't have parametrised calibrations, so we don't want to do it.  To construct the mapping lazily, we need to keep hold of what the parameters were initially, which is derived from the Rust-space `ParameterTable`.  Later, the `mapping_view` is that `{parameter: bound_value}` map, but it's lazily constructed in `map_calibrations` on the first access.

jakelishman:
`nonlocal` is needed because Python has implicit variable declarations on assignment.  If there's no assignment, you do regular scoped lookups.  Since we assign to `mapping_view` in the function, that would imply that we were creating a new variable (doesn't matter that we read it before we wrote to it - that just results in an `UnboundLocalError`).  `nonlocal` suppresses the implicit-declaration behaviour and instead makes the scoped lookup work as expected, though with the minor restriction that the binding must resolve at compile time to a variable in a nonlocal scope:
```python
x = 0

def f():
    # This is a syntax error because there's no enclosing scopes containing `f`.
    # (The global scope isn't "enclosing" in Python terminology).
    nonlocal x

def g():
    x = 1
    def f():
        # This is fine, and gets the `x = 1`
        nonlocal x
        
class A:
    x = 2
    def f(self):
        # Also a syntax error, because class bodies don't introduce enclosing scopes.
        nonlocal x
```

Unless you feel particularly strongly, I'd rather not add comments on how `nonlocal` works - it's not a particularly common keyword, but we do use it occasionally for things just like this, and it's one of the reasons the keyword exists.

qiskit/circuit/quantumcircuit.py
kevinhartman: Excellent, the improvements you've made to the code comments make it much clearer. Thanks!

kevinhartman (approved):
@jakelishman 

> Unless you feel particularly strongly, I'd rather not add comments on how nonlocal works - it's not a particularly common keyword, but we do use it occasionally for things just like this, and it's one of the reasons the keyword exists.

I'm in agreement with this, and I like the way you've described what's going on here in terms of "business logic" (I'd personally never argue to add commentary intended to teach the reader the specifics of a programming language heh).

Thanks for these changes, it'll be a lot more understandable now what's going on here to someone unfamiliar with this module.

PR #13340, opened 2024-10-17
Fix OpenQASM 2 `gate` definitions following a shadowed built-in `gate`

### LINKED ISSUE #13339: `from_qasm_str` not importing `Fswap` correctly
### Environment

- **Qiskit version**: 1.2.4
- **Python version**: 3.12.2
- **Operating system**: Linux

### What is happening?

When one creates a circuit with `qiskit.QuantumCircuit.from_qasm_str` an Fswap gate will sometimes be mapped to swap. This seems to only happen if there is a swap before the fswap in the qasm file.

```
swap q[0],q[1];
fswap q[0],q[1];
fswap q[0],q[1];
```

gives

```
CircuitInstruction(operation=Instruction(name='swap', num_qubits=2, num_clbits=0, params=[]), qubits=(Qubit(QuantumRegister(2, 'q'), 0), Qubit(QuantumRegister(2, 'q'), 1)), clbits=())
CircuitInstruction(operation=Instruction(name='swap', num_qubits=2, num_clbits=0, params=[]), qubits=(Qubit(QuantumRegister(2, 'q'), 0), Qubit(QuantumRegister(2, 'q'), 1)), clbits=())
CircuitInstruction(operation=Instruction(name='swap', num_qubits=2, num_clbits=0, params=[]), qubits=(Qubit(QuantumRegister(2, 'q'), 0), Qubit(QuantumRegister(2, 'q'), 1)), clbits=())
```

But 

```
fswap q[0],q[1];
swap q[0],q[1];
fswap q[0],q[1];
```

gives

```
CircuitInstruction(operation=Instruction(name='fswap', num_qubits=2, num_clbits=0, params=[]), qubits=(Qubit(QuantumRegister(2, 'q'), 0), Qubit(QuantumRegister(2, 'q'), 1)), clbits=())
CircuitInstruction(operation=Instruction(name='swap', num_qubits=2, num_clbits=0, params=[]), qubits=(Qubit(QuantumRegister(2, 'q'), 0), Qubit(QuantumRegister(2, 'q'), 1)), clbits=())
CircuitInstruction(operation=Instruction(name='fswap', num_qubits=2, num_clbits=0, params=[]), qubits=(Qubit(QuantumRegister(2, 'q'), 0), Qubit(QuantumRegister(2, 'q'), 1)), clbits=())
```

Run this code:
```py

qasm_str = """OPENQASM 2.0;

gate u3(theta,phi,lambda) q { U(theta,phi,lambda) q; }
gate u2(phi,lambda) q { U(pi/2,phi,lambda) q; }
gate u1(lambda) q { U(0,0,lambda) q; }
gate rx(theta) a { u3(theta,-pi/2,pi/2) a; }
gate ry(theta) a { u3(theta,0,0) a; }
gate rz(phi) a { u1(phi) a; }
gate cx c,t { CX c,t; }

gate swap a,b { cx a,b; cx b,a; cx a,b; }
gate fswap a,b { rz(-pi/2) a; rz(-pi/2) b; rx(pi/2) a; cx a,b; rx(-pi/2) a; ry(-pi/2) b; cx a,b; rx(-pi/2) a; }

qreg q[2];

swap q[0],q[1];
fswap q[0],q[1];
fswap q[0],q[1];"""

qiskit_circuit = QuantumCircuit.from_qasm_str(qasm_str)
for gate in qiskit_circuit:
    print(gate)
qiskit_circuit.draw()
```

It outputs

```
CircuitInstruction(operation=Instruction(name='swap', num_qubits=2, num_clbits=0, params=[]), qubits=(Qubit(QuantumRegister(2, 'q'), 0), Qubit(QuantumRegister(2, 'q'), 1)), clbits=())
CircuitInstruction(operation=Instruction(name='swap', num_qubits=2, num_clbits=0, params=[]), qubits=(Qubit(QuantumRegister(2, 'q'), 0), Qubit(QuantumRegister(2, 'q'), 1)), clbits=())
CircuitInstruction(operation=Instruction(name='swap', num_qubits=2, num_clbits=0, params=[]), qubits=(Qubit(QuantumRegister(2, 'q'), 0), Qubit(QuantumRegister(2, 'q'), 1)), clbits=())

        ┌───────┐┌───────┐
q_0: ─X─┤0      ├┤0      ├
      │ │  Swap ││  Swap │
q_1: ─X─┤1      ├┤1      ├
        └───────┘└───────┘
```

### What should happen?

I should output

```
CircuitInstruction(operation=Instruction(name='swap', num_qubits=2, num_clbits=0, params=[]), qubits=(Qubit(QuantumRegister(2, 'q'), 0), Qubit(QuantumRegister(2, 'q'), 1)), clbits=())
CircuitInstruction(operation=Instruction(name='fswap', num_qubits=2, num_clbits=0, params=[]), qubits=(Qubit(QuantumRegister(2, 'q'), 0), Qubit(QuantumRegister(2, 'q'), 1)), clbits=())
CircuitInstruction(operation=Instruction(name='fswap', num_qubits=2, num_clbits=0, params=[]), qubits=(Qubit(QuantumRegister(2, 'q'), 0), Qubit(QuantumRegister(2, 'q'), 1)), clbits=())

        ┌────────┐┌────────┐
q_0: ─X─┤0       ├┤0       ├
      │ │  Fswap ││  Fswap │
q_1: ─X─┤1       ├┤1       ├
        └────────┘└────────┘
```

### Any suggestions?

_No response_

jakelishman:
### Summary

Previously, when given an OpenQASM 2 program such as

    OPENQASM 2.0;
    gate builtin a { U(0, 0, 0) a; }
    gate not_builtin a { U(pi, pi, pi) a; }
    qreg q[1];

    not_builtin q[0];

and a set of `custom_instructions` including a `builtin=True` definition for `builtin` (but not for `not_builtin`), the Rust and Python sides would get themselves out-of-sync and output a gate that matched a prior definition, rather than `not_builtin`.

This was because the Rust side was still issuing `DefineGate` bytecode instructions, even for gates whose OpenQASM 2 definitions should have been ignored, so Python-space thought there were more gates than Rust-space thought there were.

### Details and comments

Fix #13339.

### DIFF
--- a/crates/qasm2/src/parse.rs
+++ b/crates/qasm2/src/parse.rs
@@ -827,8 +827,16 @@ impl State {
         }
         bc.push(Some(InternalBytecode::EndDeclareGate {}));
         self.gate_symbols.clear();
-        self.define_gate(Some(&gate_token), name, num_params, num_qubits)?;
-        Ok(statements + 2)
+        let num_bytecode = statements + 2;
+        if self.define_gate(Some(&gate_token), name, num_params, num_qubits)? {
+            Ok(num_bytecode)
+        } else {
+            // The gate was built-in, so we don't actually need to emit the bytecode.  This is
+            // uncommon, so it doesn't matter too much that we throw away allocation work we did -
+            // it still helps that we verified that the gate body was valid OpenQASM 2.
+            bc.truncate(bc.len() - num_bytecode);
+            Ok(0)
+        }
     }
 
     /// Parse an `opaque` statement.  This assumes that the `opaque` token is still in the token
@@ -1634,6 +1642,9 @@ impl State {
     /// bytecode because not all gate definitions need something passing to Python.  For example,
     /// the Python parser initializes its state including the built-in gates `U` and `CX`, and
     /// handles the `qelib1.inc` include specially as well.
+    ///
+    /// Returns whether the gate needs to be defined in Python space (`true`) or if it was some sort
+    /// of built-in that doesn't need the definition (`false`).
     fn define_gate(
         &mut self,
         owner: Option<&Token>,
@@ -1685,12 +1696,14 @@ impl State {
             }
             match self.symbols.get(&name) {
                 None => {
+                    // The gate wasn't a built-in, so we need to move the symbol in, but we don't
+                    // need to increment the number of gates because it's already got a gate ID
+                    // assigned.
                     self.symbols.insert(name, symbol.into());
-                    self.num_gates += 1;
-                    Ok(true)
+                    Ok(false)
                 }
                 Some(GlobalSymbol::Gate { .. }) => {
-                    self.symbols.insert(name, symbol.into());
+                    // The gate was built-in and we can ignore the new definition (it's the same).
                     Ok(false)
                 }
                 _ => already_defined(self, name),
--- a/releasenotes/notes/qasm2-builtin-gate-d80c2868cdf5f958.yaml
+++ b/releasenotes/notes/qasm2-builtin-gate-d80c2868cdf5f958.yaml
@@ -1,0 +1,7 @@
+---
+fixes:
+  - |
+    The OpenQASM 2 importer previously would output incorrect :class:`.Gate` instances for gate
+    calls referring to a ``gate`` definition that followed a prior ``gate`` definition that was
+    being treated as a built-in operation by a :class:`~.qasm2.CustomInstruction`.  See
+    `#13339 <https://github.com/Qiskit/qiskit/issues/13339>`__ for more detail.
--- a/test/python/qasm2/test_structure.py
+++ b/test/python/qasm2/test_structure.py
@@ -1648,6 +1648,47 @@ class TestCustomInstructions(QiskitTestCase):
         qc.append(MyGate(0.5), [0])
         self.assertEqual(parsed, qc)
 
+    def test_compatible_definition_of_builtin_is_ignored(self):
+        program = """
+            qreg q[1];
+            gate my_gate a { U(0, 0, 0) a; }
+            my_gate q[0];
+        """
+
+        class MyGate(Gate):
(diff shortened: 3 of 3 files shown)


crates/qasm2/src/parse.rs
```diff
@@ -1685,12 +1696,14 @@ impl State {
             }
             match self.symbols.get(&name) {
                 None => {
+                    // The gate wasn't a built-in, so we need to move the symbol in, but we don't
+                    // need to increment the number of gates because it's already got a gate ID
+                    // assigned.
                     self.symbols.insert(name, symbol.into());
-                    self.num_gates += 1;
-                    Ok(true)
+                    Ok(false)
```
ElePT, line 1703: This is probably me missing something, because it looks like the tests pass, but if the gate isn't a built-in, shouldn't the function return `Ok(True)` because it needs a definition?

jakelishman: It's just off the top of the diff, but this is in an `if` block for a gate that's already in the "overrideable gates" symbol table, which requires the user to have given a constructor function to the Python-space entry point in `custom_instructions`.  So it's already got a gate index allocated, and Python space knows how to construct it.

A "built in" here is meaning "pretend that the gate was always defined in OpenQASM 2", whereas "not a built in" means "use this function to construct the gate, but only allow it if you see a `gate` statement with its name (but ignore the body of that statement)".  We're in the second case here - we don't need to return the body of the gate statement to Python space, because we're ignoring it; we just want to assign the gate index of the overrideable gate to the now properly defined symbol.

ElePT: Thanks! That makes sense, and is what I suspected, but wanted to make sure I was understanding it right.

ElePT (approved):
LGTM for what I can see!

PR #13342, opened 2024-10-17
Use singleton canonical instances for OpenQASM 3 gate lookups

jakelishman:
### Summary

This reduces most symbol-table lookups for gate calls to simple `is` lookups, which is a sizeable increase in performance for the majority of large circuits passed to the OpenQASM 3 exporter.  These circuits are typically entirely dominated by standard-library gate calls.

### Details and comments

This is a very simple tweak with a pretty large impact on performance compared to the 1.2 series for circuits compiled into standard-library ones, since it near completely erases the overhead of the gate-based symbol lookup.


raynelfss (approved):
Looks sensible to me, since the canonical gates are already initialized when we access them now. Good job!

PR #13343, opened 2024-10-17
Improve OpenQASM 3 exporter performance for hardware-like circuits

jakelishman:
### Summary

This reorganises the variable lookups and internal AST generation slightly to improve the OpenQASM 3 exporter performance for the common case of circuits with many singly parametric 1q gates, parametrised by a single `Parameter` expression.  This is quite common in high-throughput cases, where the circuit is being used as a parametric template for a backend compiler.

### Details and comments

Similar to #13342, this was a couple of very easy tweaks I could make within Python space to slightly improve the runtime performance of the OQ3 importer.  It's still pretty slow for what it does, but there's little point focussing on its performance while it's still pure Python, and the majority of our data models are moving to Rust.


qiskit/qasm3/exporter.py
```diff
@@ -1145,9 +1145,11 @@ def _rebind_scoped_parameters(self, expression):
         # missing, pending a new system in Terra to replace it (2022-03-07).
         if not isinstance(expression, ParameterExpression):
             return expression
+        if isinstance(expression, Parameter):
+            return self.symbols.get_variable(expression).string
```
kevinhartman, line 1170: Is this a different behavior here? I've not looked too much at `ParameterExpression`, but is it odd for this to return a string rather than an `Expression`? Perhaps I'm missing something 🙂

qiskit/qasm3/exporter.py
```diff
@@ -1160,18 +1162,14 @@ def build_gate_call(self, instruction: CircuitInstruction):
...
-                for param in instruction.operation.params
-            ]
-        else:
-            parameters = [
-                ast.StringifyAndPray(pi_check(self._rebind_scoped_parameters(param), output="qasm"))
-                for param in instruction.operation.params
-            ]
-
+        qubits = [self._lookup_bit(qubit) for qubit in instruction.qubits]
+        parameters = [
```
kevinhartman, line 1187: Can we use a generator expression to lazily apply the `pi_check` to all elements when `not self.disable_constants`? As it is, it looks like we'd iterate through the parameters two times now.

kevinhartman (commented):
Looks mostly good, just a few minor comments.

Thanks for doing this :)

qiskit/qasm3/exporter.py
jakelishman: It's true, but the time of `pi_check` will absolutely dwarf the additional loop overhead. I imagine that if we used a generator expression with an extra lazy loop, the generator `map` function call on each iteration would be about the same as just iterating through again on a concrete list.

qiskit/qasm3/exporter.py
jakelishman: You're not missing anything, it's just abuse of the complete absence of handling for `ParameterExpression` in the exporter - we literally just call `str` and hope for the best.

The reason not to construct a `ParameterExpression` is because the case of a standalone `Parameter` is _very_ common on the performance critical path, and `ParameterExpression` (and `Parameter`) are both very expensive to construct.

qiskit/qasm3/exporter.py
kevinhartman: That's a fair point.

kevinhartman (approved):
Looks good.

Thanks for the comments / clarifications!

PR #13181, opened 2024-09-18
Fix qft-plugins for custom 'qft' gates

alexanderivrii:
### Summary

The ``HighLevelSynthesis`` transpiler pass no longer raises an exception when encountering a custom gate that is called "qft" but is not an instance of class ``QFTGate``. Instead, the synthesis plugins for QFT gates ignore such a gate, and the gate's definition is used (if provided).

Fixes #13120 and #13174.

The PR is very short and has backport potential (if desired).

PR #13345, opened 2024-10-19
Fix the calculation for identity operator's expectation value

### LINKED ISSUE #13029: Statevector._expectation_value_pauli returns incorrect result in certain cases
The calculation for the expectation value of the __identity operator__ seems incorrect when the statevector is __not normalized__ to 1.

I think the reason is that one should use the norm squared rather than the norm in the following lines.
https://github.com/Qiskit/qiskit/blob/1be2c5aa103f1991f2fdc75a77a09f82635b8431/qiskit/quantum_info/states/statevector.py#L478-L479

---

The whole function is attached below to provide context.

https://github.com/Qiskit/qiskit/blob/1be2c5aa103f1991f2fdc75a77a09f82635b8431/qiskit/quantum_info/states/statevector.py#L458-L491

yjh-bill:
### Summary

Fix a bug in `qiskit.quantum_info` in the calculation of expectation value of the identity operator.

Fixes #13029 

### Details and comments
Currently,  `qiskit.quantum_info` yields incorrect results for the expectation value of an identity operator for a given state vector. It returns the norm of the state vector; however, it ought to return the square of the norm.

Therefore, I am changing the following line  

https://github.com/Qiskit/qiskit/blob/1be2c5aa103f1991f2fdc75a77a09f82635b8431/qiskit/quantum_info/states/statevector.py#L478-L479

to 
```Python
   if x_mask + z_mask == 0: 
       return pauli_phase * np.linalg.norm(self.data) ** 2
```

### DIFF
--- a/releasenotes/notes/fix_identity_operator_9e2ec9770ac046a6.yaml
+++ b/releasenotes/notes/fix_identity_operator_9e2ec9770ac046a6.yaml
@@ -1,5 +1,5 @@
 ---
 fixes:
   - |
-    Fixed a bug that caused :meth:`Statevector.expectation_value` to yield incorrect results
+    Fixed a bug that caused :meth:`.Statevector.expectation_value` to yield incorrect results
     for the identity operator when the statevector was not normalized.
--- a/releasenotes/notes/fix_identity_operator_9e2ec9770ac046a6.yaml
+++ b/releasenotes/notes/fix_identity_operator_9e2ec9770ac046a6.yaml
@@ -1,5 +1,5 @@
 ---
 fixes:
   - |
-    Fix a bug that causes Statevector.expectation_value to yield incorrect result
-    for the identity operator when the statevector is not normalized to 1.
+    Fixed a bug that caused :meth:`Statevector.expectation_value` to yield incorrect results
+    for the identity operator when the statevector was not normalized.
--- a/releasenotes/notes/fix_identity_operator_9e2ec9770ac046a6.yaml
+++ b/releasenotes/notes/fix_identity_operator_9e2ec9770ac046a6.yaml
@@ -1,0 +1,5 @@
+---
+fixes:
+  - |
+    Fix a bug that causes Statevector.expectation_value to yield incorrect result
+    for the identity operator when the statevector is not normalized to 1.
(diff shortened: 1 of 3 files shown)


Cryoris:
Thanks for opening a PR to fix this! Could you add a release note and a test, to verify it now works as expected? (See e.g. [this PR](https://github.com/Qiskit/qiskit/pull/13181) for an example on how a bugfix PR should look like 🙂)

yjh-bill:
Sure, I just added a release note and a test.

releasenotes/notes/fix_identity_operator_9e2ec9770ac046a6.yaml
```diff
@@ -0,0 +1,5 @@
+---
+fixes:
+  - |
+    Fix a bug that causes Statevector.expectation_value to yield incorrect result
+    for the identity operator when the statevector is not normalized to 1.
```
Cryoris:
```suggestion
    Fixed a bug that caused :meth:`Statevector.expectation_value` to yield incorrect results
    for the identity operator when the statevector was not normalized.
```

test/python/quantum_info/states/test_statevector.py
```diff
@@ -1152,6 +1152,37 @@ def test_expval_pauli_qargs(self, qubits):
         expval = state.expectation_value(op, qubits)
         self.assertAlmostEqual(expval, target)
 
+
+    def test_expval_identity(self):
```
Cryoris, line 1155: It seems like you need to run the formatter here, you can do that e.g. by calling `make black` (you might want to make sure the right version is installed before doing so, by running `pip install -r requirements-dev.txt -c constraints.txt` before)

yjh-bill: Got it. Thanks for the instruction.

releasenotes/notes/fix_identity_operator_9e2ec9770ac046a6.yaml
```diff
@@ -0,0 +1,5 @@
+---
+fixes:
+  - |
+    Fixed a bug that caused :meth:`Statevector.expectation_value` to yield incorrect results
```
Cryoris: My bad, I missed the initial `.` here in my original suggestion -- otherwise this LGTM 🙂 
```suggestion
    Fixed a bug that caused :meth:`.Statevector.expectation_value` to yield incorrect results
```

Cryoris (approved):
Thanks for this fix!

PR #13346, opened 2024-10-19
Add `global_phase` to `CircuitData::{__reduce__|__eq__}`

eliarbel:
`global_phase` was missing in the tuple created by `CircuitData::__reduce__`, leading to partial pickling of `CircuitData`.


jakelishman (commented):
Can you add a test case so we don't drop it again?  I'm surprised we didn't already catch this in tests...

eliarbel:
> Can you add a test case so we don't drop it again?

I was going to open an issue later to discuss this. More than dropping it somehow, I'm more concerned about people adding new fields to any of our `__reduce__`'able structs in Rust and forgetting to include them for pickling. So initially I was thinking about using the `inspect` module to extract the attributes and test that all fields (modulo explicitly exempted ones) are covered by `__reduce__`, but it turned out to be a bit tricky with Rust's visibility and how PyO3 exposes class data. So I'm leaning more towards testing this in Rust maybe, but I need to think more about it. 

> I'm surprised we didn't already catch this in tests...

We actually have a test that checks this somewhat indirectly: `test/python/compiler/test_assembler.py::TestCircuitAssembler::test_circuit_with_global_phase`.  It fails on my local `stestr` runs but somehow not in all CI runs - I do see passing PR checks, but haven't drilled down on this. So maybe this is sufficient for now until we have a more comprehensive solution per the point above, what do you say?

mtreinish:
Oh, it's because you were running the tests directly on linux. The `assemble` function using multiprocessing by default when creating a qobj for more than one circuit at a time (on linux only). In our tox and ci configuration we disable multiprocessing by default to avoid oversubscribing the CPU between stestr's multiprocesing, qiskit's use of Python multiprocessing, and rust multithreading (there are also known issues with python's fork launch method and multithreading that causes deadlocking). For example:

https://github.com/Qiskit/qiskit/blob/main/tox.ini#L17

So we never were running that test in ci with multiprocessing and not testing it with pickle. But when you ran tests manually the env variables weren't set so it used multiprocessing and was using pickle to serialize the objects for IPC.

jakelishman:
For things being dropped from `__reduce__`: you can have the same trouble from a pure-Python type if `__reduce__` or `__getstate__` are defined. Typically when overriding those methods we should always be including explicit tests of the pickle protocol (and copy and deepcopy by extension), but I guess we've dropped some.

Fwiw, the sort of `inspect` thing you're talking about is actually fairly straight-up how the default implementation of pickle works (it recursively pickles everything in the slots and instance dict). If we're overriding the methods, it's usually _because_ that'll fail to work for some reason.

eliarbel:
> Fwiw, the sort of inspect thing you're talking about is actually fairly straight-up how the default implementation of pickle works (it recursively pickles everything in the slots and instance dict). If we're overriding the methods, it's usually because that'll fail to work for some reason.

I realize that my indented use of `inspect` for this use-case wasn't clear. I've opened an issue to further explain and discuss this in #13347 . Let me know if you think there is value in the proposed approach. If so we can add a test for the case here using this approach. Otherwise just an explicit test to this PR

eliarbel:
I added a pickle roundtrip test in 23449b3c123b7a85ec4f617e1148066803953f02. `global_phase` was also not looked at in the `__eq__` function so I added comparison check there as well. 

A side note: the semantics of the `__eq__` function is a bit weird in the sense that the way it is implemented now it doesn't require the RHS to be of type `CircuitData` for the comparison to evaluate to true. This is maybe intentional and in fact this behavior is taken advantage of when comparing the result of `QuantumCircuitData.copy()` (which returns a list of instructions) to the same `QuantumCircuit`'s `_data` attribute using the `==` operator. So in this use-case it's like comparing up to global phase. Although I find this weird, I kept the original semantics w.r.t the RHS type requirement. Should we change it or is that intentional? if intentional it should be at least documented.

jakelishman:
The original reason that `CircuitData::__eq__` allowed arbitrary comparisons was to match Python-space behaviour that you could compare equality of `QuantumCircuit._data` to a list, which seems like it's being used in our tests, if I understand you right?  Nowadays it includes significantly more than just the list of instructions, so it's probably right to have its `other` argument be `&Bound<Self>` (PyO3 should handle a failed downcast correctly) and to fix our tests.

I'd probably rather do that bit in a separate PR, since it's a separate change to fixing `__reduce__`.  That said, `CircuitData::__eq__` _should_ never be called, since `QuantumCircuit.__eq__` doesn't use it (it does an isomorphism check).  We might even be able to remove `CircuitData::__eq__` entirely.

eliarbel:
> seems like it's being used in our tests, if I understand you right?

Yes, there a few places that use this path. Another one now is the pickle roundtrip test that I've added. 

Ok then, removing `CircuitData::__eq__` in a separate PR sounds good. I can open one soon.

crates/circuit/src/circuit_data.rs
```diff
@@ -776,6 +777,16 @@ impl CircuitData {
...
         }
+
+        if let Ok(other_dc) = other.downcast::<CircuitData>() {
+            if !slf
+                .getattr("global_phase")?
+                .eq(other_dc.getattr("global_phase")?)?
+            {
+                return Ok(false);
+            }
+        }
```
jakelishman, line 788: This is totally fine, especially because we're probably going to remove this method in short order, but just for the future: after we've downcast the two things to `Bound<CircuitData>` (`slf` already was, and now `other_dc` is), we can `borrow` them and then use Rust-space access for the fields rather than needing Python-space access.

For example:
```rust
if let Ok(other) = other.downcast::<CircuitData>() {
    if !slf.borrow().global_phase.eq(&other.borrow().global_phase) {
        return Ok(false);
    }
}
```
should work (untested), and we don't need to check for Python-space exceptions so much.

jakelishman (approved):
Thanks - I think separating out the PRs will definitely make this easier to reason about any further changes to `CircuitData::__eq__` and when we may / may not need it.  Definitely good to get `__reduce__` fixed.

PR #13351, opened 2024-10-22
Deprecate `qiskit.result.mitigation`

### LINKED ISSUE #12832: LocalReadoutMitigator does not support BackendV2
### Environment

- **Qiskit version**: 1.3.0.dev0+6b521db
- **Python version**: 3.9
- **Operating system**: MacOS

### What is happening?

The class `qiskit.result.mitigation.local_readout_mitigator.LocalReadoutMitigator` does not support BackendV2. Given that [BackendV1 is now deprecated](https://github.com/Qiskit/qiskit/pull/12629)... it getting complicated to have this class post 2.0. 

```python
from qiskit.result import LocalReadoutMitigator

backend = GenericBackendV2(
    num_qubits=5,
    seed=42,
)
LocalReadoutMitigator(backend=backend)
```
```
AttributeError: 'GenericBackendV2' object has no attribute 'properties'
```

### What should happen?

The class works with BackendV1
```python
from qiskit.result import LocalReadoutMitigator
from qiskit.providers.fake_provider import Fake5QV1

LocalReadoutMitigator(backend=Fake5QV1())
```
Ideally, it should also work with BackendV2.

### Any suggestions?

`¯\_(ツ)_/¯`

### LINKED ISSUE #12838: Deprecate `qiskit.result.mitigation`
In Qiskit 2.0, we should remove the `qiskit.result.mitigation` module. Therefore, let's deprecate it in 1.3.

Maybe the module can be moved to `qiskit-experiments`.

ElePT:
### Summary
Closes https://github.com/Qiskit/qiskit/issues/12838


releasenotes/notes/deprecate-mitigation-f5f6ef3233b3d726.yaml
```diff
@@ -0,0 +1,7 @@
+---
+deprecations_misc:
+  - |
+    The ``qiskit.result.mitigation`` module has been deprecated and will be removed in the 2.0 release.
+    The deprecation includes the ``LocalReadoutMitigator`` and ``CorrelatedReadoutMitigator`` classes
```
mtreinish, line 5: You didn't want to use `class` refs for these? They seem to be in the docs: https://docs.quantum.ibm.com/api/qiskit/qiskit.result.LocalReadoutMitigator

qiskit/result/mitigation/utils.py
```diff
@@ -19,12 +19,20 @@
...
 logger = logging.getLogger(__name__)
 
 
+@deprecate_func(
+    since="1.3",
+    package_name="Qiskit",
+    removal_timeline="in Qiskit 2.0",
+    additional_msg="The `qiskit.result.mitigation` module is deprecated in favor of "
+    "the https://github.com/Qiskit/qiskit-addon-mthree package.",
+)
```
mtreinish, line 35: I think only the 3 classes above are actually public so these technically could just be removed. But it's good to deprecate them anyway and remove it all at once.

mtreinish (approved):
This LGTM, I had one question inline about the release note. But it's easy enough to fix in the release notes roundup as we approach the final release. Feel free to enqueue this for merge.

releasenotes/notes/deprecate-mitigation-f5f6ef3233b3d726.yaml
ElePT: I didn't remember if we would use refs for removals, I will keep it in mind and suggest the changes in the reno copy-editing phase. This way we can get the merge queue going with this PR :)

PR #13354, opened 2024-10-22
Binary arithmetic gates

Cryoris:
### Summary

Part of #13046: gate representations & plugins for the binary arithmetic operations (adder and multiplier).


qiskit/circuit/library/arithmetic/adders/adder.py
```diff
@@ -56,3 +69,133 @@ def num_state_qubits(self) -> int:
...
+
+    """
+
+    def __init__(self, num_state_qubits: int, label: str | None = None) -> None:
+        """
+        Args:
+            num_state_qubits: The number of qubits in each of the registers.
+            name: The name of the circuit.
+        """
+        super().__init__("FullAdder", 2 * num_state_qubits, [], label=label)
```
alexanderivrii: Shouldn't this be `2 * num_state_qubits + 2` (with the `+2` corresponding to the `c_in` and the `c_out` bits)?

qiskit/circuit/library/arithmetic/adders/adder.py
```diff
@@ -39,6 +42,16 @@ class Adder(QuantumCircuit):
...
+    @deprecate_func(
+        since="1.3",
+        additional_msg=(
+            "Use the adder gates provided in qiskit.circuit.library.arithmetic instead. "
+            "The gate type depends on the adder kind: fixed, half, full are represented by "
+            "ModularAdderGate, AdderGate, FullAdderGate, respectively. For different adder "
+            "implementations, see https://docs.quantum.ibm.com/api/qiskit/synthesis.",
+        ),
+        pending=True,
+    )
```
alexanderivrii: With the naming conventions introduced in this PR, "Adder" stands for (what used to be called) a "half"-adder, "ModularAdder" stands for a "fixed"-adder, and a "FullAdder" stands for a "full"-adder. Personally, I like these new names, especially "modular" instead of "fixed". My only question is whether we should use the name "Adder" or the name "HalfAdder" to represent an adder without a `c_in`. I think "HalfAdder" is a more standard terminology, and at least to me the name "Adder" can refer to all kinds of adders, not necessarily to the "half"-one. But I am fine with any decision here.

qiskit/circuit/library/arithmetic/adders/cdkm_ripple_carry_adder.py
```diff
@@ -75,6 +74,21 @@ class CDKMRippleCarryAdder(Adder):
     It has one less qubit than the full-adder since it doesn't have the carry-out, but uses
     a helper qubit instead of the carry-in, so it only has one less qubit, not two.
 
+    .. seealso::
+
```
alexanderivrii, line 78: I did not know about ".. seealso::" before, but apparently we already use it quite a bit in Qiskit.

qiskit/circuit/library/arithmetic/adders/adder.py
alexanderivrii: An alternative implementation would be to have a single "AdderGate" which internally stores the type of the adder (half/full/modular) and have adder gate plugins take this into account. I am just pointing this out but I prefer the approach taken in the PR.

qiskit/circuit/library/arithmetic/multipliers/multiplier.py
```diff
@@ -99,3 +109,84 @@ def num_result_qubits(self) -> int:
...
+        Args:
+            num_state_qubits: The number of qubits in each of the input registers.
+            num_result_qubits: The number of result qubits to limit the output to.
+                Default value is ``2 * num_state_qubits`` to represent any possible
+                result from the multiplication of the two inputs.
+            name: The name of the circuit.
+        Raises:
+            ValueError: If ``num_state_qubits`` is smaller than 1.
+            ValueError: If ``num_result_qubits`` is smaller than ``num_state_qubits``.
+            ValueError: If ``num_result_qubits`` is larger than ``2 * num_state_qubits``.
```
alexanderivrii, line 158: A small nitpick: for multipliers the check that the input data is valid (e.g. that `num_state_qubits` is at least `1`) is done as part of the `__init__` method, while for adders this check is done in the synthesis itself. For consistency, should we do the validity checks in `__init__` also for adders?

qiskit/synthesis/arithmetic/adders/cdkm_ripple_carry_adder.py
```diff
@@ -0,0 +1,154 @@
...
+
+"""Compute the sum of two qubit registers using ripple-carry approach."""
+
+from qiskit.circuit.quantumcircuit import QuantumCircuit
+from qiskit.circuit.quantumregister import QuantumRegister, AncillaRegister
+
+
+def adder_ripple_c04(num_state_qubits: int, kind: str = "half") -> QuantumCircuit:
+    r"""A ripple-carry circuit to perform in-place addition on two qubit registers.
+
```
alexanderivrii, line 21: This is nice. The actual synthesis functions still take the adder type as the parameter (i.e. we do not have individual functions `half_adder_ripple_c04`, `modular_adder_ripple_c04`, `full_adder_ripple_c04`).

qiskit/synthesis/arithmetic/adders/draper_qft_adder.py
```diff
@@ -0,0 +1,103 @@
...
+    # add registers
+    circuit = QuantumCircuit(*qr_list)
+
+    # define register containing the sum and number of qubits for QFT circuit
+    qr_sum = qr_b[:] if kind == "fixed" else qr_b[:] + qr_z[:]
+    num_sum = num_state_qubits if kind == "fixed" else num_state_qubits + 1
+
+    # build QFT adder circuit
+    qft = QFTGate(num_sum)
+    circuit.append(qft, qr_sum[:])
```
alexanderivrii, line 91: I like that the new definition now uses `QFTGate`s instead of `QFT` circuits. I guess this is also why the DraperQFTAdder circuit is *not* using this new definition (as the circuit *is* syntactically changed, especially with it using the true `QFTGate` without reversing the order of its output qubits). Semantically (as Operators) the new and the old definitions should still be equivalent.

qiskit/synthesis/arithmetic/multipliers/rg_qft_multiplier.py
```diff
@@ -0,0 +1,99 @@
...
+            f"({num_state_qubits}) and 2 * num_state_qubits ({2 * num_state_qubits})"
+        )
+
+    qr_a = QuantumRegister(num_state_qubits, name="a")
+    qr_b = QuantumRegister(num_state_qubits, name="b")
+    qr_out = QuantumRegister(num_result_qubits, name="out")
+
+    # build multiplication circuit
+    circuit = QuantumCircuit(qr_a, qr_b, qr_out)
+    qft = QFTGate(num_result_qubits)
```
alexanderivrii, line 81: I like this!

qiskit/synthesis/arithmetic/multipliers/hrs_cumulative_multiplier.py
```diff
@@ -0,0 +1,102 @@
...
+    qr_out = QuantumRegister(num_result_qubits, name="out")
+
+    circuit = QuantumCircuit(qr_a, qr_b, qr_out)
+
+    # prepare adder as controlled gate
+    # pylint: disable=cyclic-import
+    from qiskit.circuit.library.arithmetic import AdderGate, ModularAdderGate
+
+    adder = AdderGate(num_state_qubits)
+    controlled_adder = adder.control(annotated=True)
```
alexanderivrii: This is also really cool! Using a generic `AdderGate` allows users to choose different synthesis methods via the high-level-synthesis config, without having to change the multiplier gate definition. Using annotated controls opens a door to more circuit optimizations that we can make.

qiskit/transpiler/passes/synthesis/hls_plugins.py
```diff
@@ -249,6 +249,117 @@
    MCMTSynthesisNoAux
    MCMTSynthesisDefault
 
+Modular Adder Synthesis
+'''''''''''''''''''''''
```
alexanderivrii, line 266: For each adder type (full/half/modular) we have its own set of plugins -- nice!

qiskit/transpiler/passes/synthesis/hls_plugins.py
```diff
@@ -1029,3 +1147,129 @@ def run(self, high_level_object, coupling_map=None, target=None, qubits=None, **
             high_level_object.num_target_qubits,
             ctrl_state,
         )
+
+
+class ModularAdderSynthesisDefault(HighLevelSynthesisPlugin):
+    """The default modular adder (no carry in, no carry out qubit) synthesis."""
+
+    def run(self, high_level_object, coupling_map=None, target=None, qubits=None, **options):
+        if options.get("num_clean_ancillas", 0) >= 1:
```
alexanderivrii: Should we, in this and other plugins, check that the type of the high level object is really an adder, to prevent users from defining custom gates with names "Adder", "FullAdder" or "ModularAdder". I am especially worried about the more generic name "Adder".

alexanderivrii: Could you please document the option "num_clean_ancillas" in the plugin's docstring? And also to explicitly say what is the key (e.g. "ripple_c04") with which the plugin will be called? 

This comment also applies to other plugins.

alexanderivrii (commented):
I really like this PR. I have left several comments in-place. 

Now the follow-up step is to port the synthesis functions to Rust :).

qiskit/circuit/library/arithmetic/adders/adder.py
ShellyGarion: I agree with @alexanderivrii . Perhaps it's worth to have only a single `AddedGate`, with various synthesis methods in the synthesis library and several variants that can be used via the HLS plugin (similar to a single `MCXGate`)

alexanderivrii: Wait, @ShellyGarion, how do you agree with me when I said that I prefer the approach in the PR, except with a "HalfAdder" instead of "Adder"? 😛

ShellyGarion: I agreed with your "alternative implementation" suggested above. But if both you and Julien prefer to have several adders, it's OK with me.

Cryoris: Since Modular/Half/Full adder vary in the number of qubits in their mathematical definition, they cannot be represented by the same gate object -- otherwise, where would you apply it on? 🙂 

I'm fine re-re-naming to `HalfAdder` if you prefer 👍🏻

qiskit/circuit/library/arithmetic/adders/adder.py
Cryoris: Oh yes it should!

qiskit/synthesis/arithmetic/adders/draper_qft_adder.py
Cryoris: Precisely! 🙂

qiskit/synthesis/arithmetic/multipliers/hrs_cumulative_multiplier.py
Cryoris: Exactly, this was really nice to re-write now that the gate objects are just mathematical blocks and the synthesis is detached 😄

qiskit/transpiler/passes/synthesis/hls_plugins.py
Cryoris: It was a bit much to type but yeah this is neat 🙂

qiskit/circuit/library/arithmetic/adders/adder.py
ShellyGarion: For the MCXGate we managed to deal with various methods with different number of qubits (although there these were ancilla qubits). Is it possible to do something similar in this case as well ?

alexanderivrii: Yeah, I like the "HalfAdder" a bit more, and I think "HalfAdder" is the more standard terminology for this circuit: [link to wiki](https://en.wikipedia.org/wiki/Adder_(electronics))

Cryoris: @ShellyGarion: The _mathematical definition_ of MCX always has the same number of qubits; we can only deal with varying auxiliary qubits in the _decomposition_. In the adders, the mathematical definition varies, so they cannot be represented by the same gate, because in the circuit modernization `gate == mathematical operation`. 🙂

Cryoris: @alexanderivrii: renamed in a54907eec954f98f61bf723fe0e82d5f4d0fb00e

qiskit/circuit/library/arithmetic/multipliers/multiplier.py
Cryoris: Done in a54907eec954f98f61bf723fe0e82d5f4d0fb00e

qiskit/transpiler/passes/synthesis/hls_plugins.py
Cryoris: Done in a54907eec954f98f61bf723fe0e82d5f4d0fb00e

alexanderivrii (approved):
Thanks for making all these changes!

Issue #13267, opened 2024-10-03
Oxidize `ParameterExpression`


mtreinish:
Right now the `ParameterExpression` class is defined solely in Python. This has been causing a large bottleneck in the use of Rust to accelerate more of Qiskit as we have to explicitly use python any time we interact with a `ParameterExpression`.  So we should move the definition of `ParameterExpression` to Rust and expose it to python with a compatible api.

At a technical level this is a more involved issue than the other pieces from #13264 because our current interface for `ParameterExpression` is dependent on symengine, a C++ library, to handle the actual expressions. We have the option of interfacing with symengine's C api fairly easily from rust, but that adds complexity to our build system, either requiring a dynamic library is installed to use qiskit (which we can bundle in the released wheels for supported platforms) which adds some runtime complexity, or we can statically link the library into qiskit at build time, but this then adds build time complexity (either building symengine from source or pulling a precompiled library from somewhere).

mtreinish:
This isn't actually complete, #13278 moved us to a rust symbolic expression library but it didn't move the `ParameterExpression` class to rust or update the data model in `CircuitData` or `DAGCircuit` to work natively with that rust object.

Issue #13131, opened 2024-09-11
Port QPY serialization to Rust


mtreinish:
### What should we add?

Now that we have most of the circuit model in Rust it should be possible to port the QPY into Rust. We have enough of the data model in Rust now to iterate over the circuit data in rust and generate the payload as we go. There are still places where there will be python interaction, mainly around custom gates and pulse/calibrations. But we can just call out to Python functions as necessary to handle that.

The actual QPY format is packed structs with network endianess and it is pretty natural to express this in rust. The payloads defined in: https://github.com/Qiskit/qiskit/blob/main/qiskit/qpy/formats.py would be expressed as something like:

```rust
#[repr(packed)]
struct FileHeader {
    preface: [c_char; 6],
    qpy_version: c_uchar,
    major_version: c_uchar,
    minor_version: c_uchar,
    patch_version: c_uchar,
    num_programs: c_ulonglong
}
```
in rust.

Although we'll have to use something like the byteorder crate: https://docs.rs/byteorder/latest/byteorder/ to specify the endianess when reading or writing the QPY payload. Although it might be easier to use something like `serde` to handle the serialization for us and then just work with rust native structs: https://serde.rs/data-format.html

PR #13356, opened 2024-10-22
Add Qiskit native QPY ParameterExpression serialization

### LINKED ISSUE #13252: Introduce Qiskit native ParameterExpression serialization
### What should we add?

Right now QPY relies on symengine's native serialization format for serializing the internal symbolic expression used by the parameter expression. At the time this made a lot of sense because we could relying on symengine to manage how best to represent it's internal data model and it was also C++ relying on the cereal library so it was very fast. But as was recently discovered by the symengine 0.13 release the payload format is version specific, and symengine will error if you try to load a expression generated with a different symengine version. For the immediate term we were able to "workaround" this issue here: https://github.com/Qiskit/qiskit/pull/13251 but this isn't a general solution as symengine could change the format between versions, that's why they have the check.

Moving forward for qiskit 1.3.0 we must come up with a qiskit native serialization for `ParameterExpression` for QPY instead of relying on symengine or sympy to do it for us. This should be introduced in QPY version 13, this version mismatch is still a possibility for users of QPY version >=10 and < 13 when loading a historically generated payload, but in those cases the only option is to install a compatible version of symengine if the payload format is relying on it.

mtreinish:
### Summary

With the release of symengine 0.13.0 we discovered a version dependence on the payload format used for serializing symengine expressions. This was worked around in #13251 but this is not a sustainable solution and only works for symengine 0.11.0 and 0.13.0 (there was no 0.12.0). While there was always the option to use sympy to serialize the underlying symbolic expression (there is a `use_symengine` flag on `qpy.dumps` you can set to `False` to do this) the sympy serialzation has several tradeoffs most importantly is much higher runtime overhead. To solve the issue moving forward a qiskit native representation of the parameter expression object is necessary for serialization.

This commit bumps the QPY format version to 13 and adds a new serialization format for ParameterExpression objects. This new format is a serialization of the API calls made to ParameterExpression that resulted in the creation of the underlying object. To facilitate this the ParameterExpression class is expanded to store an internal "replay" record of the API calls used to construct the ParameterExpression object. This internal list is what gets serialized by QPY and then on deserialization the "replay" is replayed to reconstruct the expression object. This is a different approach to the previous QPY representations of the ParameterExpression objects which instead represented the internal state stored in the ParameterExpression object with the symbolic expression from symengine (or a sympy copy of the expression). Doing this directly in Qiskit isn't viable though because symengine's internal expression tree is not exposed to Python directly. There isn't any method (private or public) to walk the expression tree to construct a serialization format based off of it. Converting symengine to a sympy expression and then using sympy's API to walk the expression tree is a possibility but that would tie us to sympy which would be problematic for #13267 and #13131, have significant runtime overhead, and it would be just easier to rely on sympy's native serialization tools.

The tradeoff with this approach is that it does increase the memory overhead of the `ParameterExpression` class because for each element in the expression we have to store a record of it. Depending on the depth of the expression tree this also could be a lot larger than symengine's internal representation as we store the raw api calls made to create the ParameterExpression but symengine is likely simplifying it's internal representation as it builds it out. But I personally think this tradeoff is worthwhile as it ties the serialization format to the Qiskit objects and APIs instead of relying on a 3rd party library. This also gives us the flexibility of changing the internal symbolic expression library internally in the future if we decide to stop using symengine at any point.

### Details and comments

Fixes #13252

TODO:

- [x] Add QPY v13 format documentation
- [x] Add release notes
- [x] Expand test coverage. We need to ensure we have thorough coverage of all supported expression operations which we weren't previously exercising through QPY.


qiskit/qpy/formats.py
```diff
@@ -259,6 +259,18 @@
 PARAMETER_PACK = "!H16s"
 PARAMETER_SIZE = struct.calcsize(PARAMETER_PACK)
 
+## PARAMETEREXPRESSION
+# PARAM_EXPR_V2 = namedtuple("PARAM_EXPR_V2", ["size"])
+# PARAM_EXPR_V2_PACK = "!H"
+# PARAM_EXPR_V2_SIZE = struct.calcsize(PARAM_EXPR_V2_PACK)
+
```
mtreinish:
```suggestion
```

mtreinish:
The state of this is it's stuck on the overly complex expression example test I added in: https://github.com/Qiskit/qiskit/pull/13356/commits/2b352d705ab30f2106f5668ec07a3fbe0bbfcd41 I'm thinking it might be worth approaching the deserialization in a slightly different approach instead of doing nested expressions recursively like it currently is. If anyone has bandwidth and can come up with a fix before me for that test, feel free to just push to the branch. I think when that test is passing we probably have enough coverage between that new test and the expression tests that already existed for circuits and pulse schedules (we might want to convert those expression tests to circuit equivalents as part of the pulse removal in 2.0).

mtreinish:
Ok, I fixed the issue from last week with: https://github.com/Qiskit/qiskit/pull/13356/commits/a903387f435294cc5a85a2e39d119a83e4864578 which switches to a stack based approach instead of a recursive one. I think this should be ready now.

qiskit/circuit/parameterexpression.py
```diff
@@ -30,12 +33,80 @@
...
+    POW = 4
+    SIN = 5
+    COS = 6
+    TAN = 7
+    ASIN = 8
+    ACOS = 9
+    EXP = 10
+    LOG = 11
+    SIGN = 12
+    DERIV = 13
```
ElePT: Minor question, any reason why you chose `DERIV` instead of `GRAD`? Is it because it was closer to the symengine notation?

qiskit/qpy/__init__.py
```diff
@@ -322,6 +342,139 @@
...
+
+Version 13 added a native Qiskit serialization representation for :class:`.ParameterExpression`.
+Previous QPY versions relied on either sympy or symengine to serialize the underlying symbolic
+expression, while in version 13 QPY now represents the sequence of API calls used to create the
+ParameterExpression.
+
+The first change is the :ref:`qpy_param_expr_v3` payload for the ``expr_size`` bytes following
+the header is now comprised of an array of ``PARAM_EXPR_ELEM_V13`` structs. The intent is for the
+array of elements to be read one struct at a time that describe the calls to make to reconstruct
+the :class:`.ParameterExpression`.
```
ElePT:
```suggestion
Previous QPY versions relied on either ``sympy`` or ``symengine`` to serialize the underlying symbolic
expression. Starting in Version 13, QPY now represents the sequence of API calls used to create the
:class:`.ParameterExpression`.

The main change in the serialization format is in the :ref:`qpy_param_expr_v3` payload.  The ``expr_size`` bytes following
the head now contain an array of ``PARAM_EXPR_ELEM_V13`` structs. The intent is for this
array to be read one struct at a time, where each struct describes one of the calls to make to reconstruct the :class:`.ParameterExpression`.
```

qiskit/qpy/__init__.py
```diff
@@ -322,6 +342,139 @@
...
+    struct {
+        unsigned char op_code;
+        char lhs_type;
+        char lhs[16];
+        char rhs_type;
+        char rhs[16];
+    } PARAM_EXPR_ELEM_V13;
+
+The ``op_code`` field is used to define the operation used for the operation
+added to the :class:`.ParameterExpression`. The value can be:
```
ElePT:
```suggestion
The ``op_code`` field is used to define the operation added to the :class:`.ParameterExpression`. The value can be:
```

qiskit/qpy/__init__.py
```diff
@@ -322,6 +342,139 @@
...
+     - :meth:`~.ParameterExpression.abs`
+   * - 17
+     - :meth:`~.ParameterExpression.arctan`
+   * - 255
+     - NULL
+
+The ``NULL`` value of 255 is only used to fill the op code field for
+entries that are not actual operations but indicate recursive definitions.
+Then the ``lhs_type`` and ``rhs_type`` fields are used to describe
+the operand types and can be one of the following utf8 encoded
```
ElePT:
```suggestion
the operand types and can be one of the following UTF-8 encoded
```

qiskit/qpy/__init__.py
```diff
@@ -322,6 +342,139 @@
...
+    final_expr = z**2 + expr
+
+the inner ``expr`` in ``final_expr``. When ``s`` is encountered this indicates that until
+an ``e` struct is reached the next structs are used for a recursive definition. For both
+``s`` and ``e`` types the data values are 0 and not used. The type value
+of ``u`` is used to represent a substitution call. This only is used for ``lhs_type``
+and is always paired with an ``rhs_type`` of ``n``. The data value is the size in bytes of
+a :ref:`qpy_mapping` encoded mapping of :class:`.Parameter` names to their value for the
+:meth:`~.ParameterExpression.subs` call. The mapping data is immediately following the
+struct, and the next struct starts immediately after the mapping data.
```
ElePT:
```suggestion
When ``s`` is encountered, this indicates that until an ``e` struct is reached, the next structs 
are used for a recursive definition. For both
``s`` and ``e`` types, the data values are not used, and always set to 0. The type value
of ``u`` is used to represent a substitution call. This is only used for ``lhs_type``
and is always paired with an ``rhs_type`` of ``n``. The data value is the size in bytes of
a :ref:`qpy_mapping` encoded mapping of :class:`.Parameter` names to their value for the
:meth:`~.ParameterExpression.subs` call. The mapping data is immediately following the
struct, and the next struct starts immediately after the mapping data.
```

qiskit/qpy/binary_io/value.py
```diff
@@ -50,20 +56,132 @@ def _write_parameter_vec(file_obj, obj):
     file_obj.write(name_bytes)
 
 
-def _write_parameter_expression(file_obj, obj, use_symengine, *, version):
-    if use_symengine:
-        expr_bytes = obj._symbol_expr.__reduce__()[1][0]
+def _encode_replay_entry(inst, file_obj, version, side=False):
```
ElePT: If `side` is used to mark the right-side operand, I think it would be clearer to say it explicitly, maybe:

```suggestion
def _encode_replay_entry(inst, file_obj, version, r_side=False):
```
(again, minor suggestion, it just took me a bit to go and check which side it was). If I confused the side, the same would apply to the left.

qiskit/qpy/binary_io/value.py
```diff
@@ -50,20 +56,132 @@ def _write_parameter_vec(file_obj, obj):
...
+            from sympy import srepr, sympify
+
+            expr_bytes = srepr(sympify(obj._symbol_expr)).encode(common.ENCODE)
+    else:
+        with io.BytesIO() as buf:
+            extra_symbols = _write_parameter_expression_v13(buf, obj, version)
+            expr_bytes = buf.getvalue()
+    symbol_table_len = len(obj._parameter_symbols)
+    if extra_symbols:
+        symbol_table_len += 2 * len(extra_symbols)
```
ElePT, line 179: If I'm not mistaken, the `extra_symbols` represent the mapping data, right? why is the len multiplied by 2?

ElePT: Aha, I think I got it, you are iterating over `extra_symbols` twice, one time to encode the keys, and another for the values.

qiskit/qpy/binary_io/value.py
```diff
@@ -334,6 +495,138 @@ def _read_parameter_expression_v3(file_obj, vectors, use_symengine):
...
+                vectors=vectors,
+            )
+        else:
+            raise exceptions.QpyError(f"Invalid parameter expression map type: {elem_key}")
+        symbol_map[symbol] = value
+    with io.BytesIO(payload) as buf:
+        return _read_parameter_expr_v13(buf, symbol_map, version, vectors)
+
+
+def _read_parameter_expr_v13(buf, symbol_map, version, vectors):
```
ElePT, line 545: I see why this is the case (there is no previous instance of `_read_parameter_expr`), but I find it a bit difficult to track that `_read_parameter_expr_v13` is related to `_read_parameter_expression_v4`. I don't think I have a good suggestion, because randomly assigning it a v4 would also not be ideal, so I am ok with leaving it as-is.

qiskit/qpy/binary_io/value.py
```diff
@@ -334,6 +495,138 @@ def _read_parameter_expression_v3(file_obj, vectors, use_symengine):
...
+            if not isinstance(lhs, ParameterExpression) and isinstance(rhs, ParameterExpression):
+                if expression_data.OP_CODE == 0:
+                    method_str = "__radd__"
+                elif expression_data.OP_CODE == 1:
+                    method_str = "__rsub__"
+                elif expression_data.OP_CODE == 2:
+                    method_str = "__rmul__"
+                elif expression_data.OP_CODE == 3:
+                    method_str = "__rtruediv__"
+                stack.append(getattr(rhs, method_str)(lhs))
```
ElePT: why is `__rpow__` not included in these? I just noticed that we are entering the `if` for codes 4, 13, 15 too, which correspond to pow, deriv and conj, but then there are no method str changes for them.

qiskit/qpy/__init__.py
```diff
@@ -322,6 +342,139 @@
...
+   * - ``s``
+     - Recursive :class:`.ParameterExpression` definition start
+   * - ``e``
+     - Recursive :class:`.ParameterExpression` definition stop
+   * - ``u``
+     - substitution
+
+if the type value is ``f`` or ``c`` the corresponding ``lhs`` or `rhs``
+field is the literal value with 0 padding for the full 128 bit width in case of a
+float. For ``i`` it's encoded as a 64 bit signed integer with 0 padding for the full
```
ElePT: I found the wording a bit confusing, I don't know if this is too verbose but I tried to reword this part a little up to my understanding:

```suggestion
If the type value is ``f`` ,``c`` or ``i``, the corresponding ``lhs`` or `rhs``
field widths are 128 bits each. In the case of floats, the literal value is encoded as a double 
with 0 padding, while complex numbers are encoded as real part followed by imaginary part,
taking up 64 bits each. For ``i`, the value is encoded as a 64 bit signed integer with 0 padding for the full
```

qiskit/qpy/__init__.py
```diff
@@ -322,6 +342,139 @@
...
+128 bit width. ``n`` is used to represent a ``None`` and typically isn't directly used as
+it indicates an argument that's not used. For ``p`` the data is the UUID for the
+:class:`.Parameter` which can be looked up in the symbol map described in the
+``map_elements`` outer :ref:`qpy_param_expr_v3` payload. If the type value is
+``s`` this marks the start of a a new recursive section for a nested
+:class:`.ParameterExpression`. For example::
```
ElePT:
```suggestion
:class:`.ParameterExpression`. For example, in the following snippet there is an inner ``expr`` 
contained in ``final_expr``, constituting a nested expression::
```

ElePT (commented):
Thanks @mtreinish. After taking a look I think the PR looks good. I think we agree in the need for a Qiskit-native alternative for the symengine serialization even if there are some tradeoffs in terms of memory footprint. I added mostly documentation-related suggestions, some typos and some rewordings, and a few minor comments. The question I left in `read_parameter_expression_v4` might be important, but I might have missed something there. Other than that, I was wondering if we should also communicate somehow that `use_symengine`  isn't useful anymore for QPY versions over 13?

qiskit/circuit/parameterexpression.py
mtreinish: Yeah it was more or less to match the functionality to what symengine called it. I can switch it to `GRAD` or `GRADIENT` if you'd prefer, the only thing that really matters here is the `13`.

qiskit/qpy/binary_io/value.py
mtreinish: Yep, that was why I set the length to 2x here

qiskit/qpy/binary_io/value.py
mtreinish: Oh, I meant to change the v4's to v13s. v4 was just a misreading of the code when I first started this. I thought we were versioning the parameter expression independently because v3 was so low. But I realized it was really changed last in qpy v3. I'll update this.

qiskit/circuit/parameterexpression.py
ElePT: this is 100% nitpick but for some reason I would prefer `GRAD`, yes, but it's not a super strong request

qiskit/qpy/binary_io/value.py
mtreinish: For 13 and 15 we can never have a reversed version. They only work if the left hand side is a parameter expression since we can't do the gradient of substitution on an a literal type. I think you're right the lack of `__rpow__` here for `OP_CODE == 5` is probably a bug, I'll add a test case for it and fix it if so.

ElePT: That's what I thought, thanks for confirming the 13 and 15 ones, I wasn't 100% sure

mtreinish: Fixed in: https://github.com/Qiskit/qiskit/pull/13356/commits/8c2205faf7971fe281d4b82100ab8f79c5bb8d70 this turned out to be a larger bug around non-commutative operations which that commit fixes. Thanks for catching this.

qiskit/qpy/binary_io/value.py
mtreinish: Done in: https://github.com/Qiskit/qiskit/pull/13356/commits/6e75ea58fa69df4c17e5add40b86c0ac5d53c899

qiskit/qpy/binary_io/value.py
mtreinish: Done in: https://github.com/Qiskit/qiskit/pull/13356/commits/11f65adbc08690ffe3cb3587814ec1274905b111

qiskit/circuit/parameterexpression.py
mtreinish: Done in: https://github.com/Qiskit/qiskit/pull/13356/commits/8ae658932de61ad5c06e70572bb9c8e0d23a5904

mtreinish:
> Other than that, I was wondering if we should also communicate somehow that use_symengine isn't useful anymore for QPY versions over 13?

I added a sentence to the `qpy.dump` function's docstring for the `use_symengine` argument to say it doesn't do anything in qpy 13. Were you thinking we needed an upgrade release note or something to documente it further? 

We can drop the argument in 2.0 potentially since 2.0 will have a minimum compatibility version of qpy format version 13 so nothing >=2.0 emits can support the `use_symengine` flag. But my worry with that is it is a potentially breaking api change for users that might want to support dumping qpy across code run with multiple qiskit versions. So maybe it's better to just document it as not doing anything in 2.0 and remove it in 3.0. Either way that's something we can debate post 1.3, since we probably should not deprecate it here, and only save that for a future release.

ElePT (approved):
> I added a sentence to the `qpy.dump` function's docstring for the `use_symengine` argument to say it doesn't do anything in qpy 13. Were you thinking we needed an upgrade release note or something to documente it further?

I didn't have a fixed idea, but I think the docstring is a good place for the moment. I also wasn't pushing for a deprecation now, just wanted to raise the point to make sure we didn't forget to deal with it in the future.

Thanks a lot for applying the review comments so fast! I have taken a second look and cannot come up with any other potential issue, so on my side I think the PR is ready to be merged.

PR #13357, opened 2024-10-22
Add run_options option and support for processing level 1 data to BackendSamplerV2

wshanks:
This change adds support to the `BackendSamplerV2` class so that it will accept a `run_options` dict option and pass its content through to the underlying `BackendV2`'s `run()` method.

Additionally, to support `meas_level=1`, the results processing code checks for a `meas_level` option inside of `run_options` and handles `meas_level=1` data appropriately when it is set, rather than always assuming the returned data is level 2.

Finally, the behavior of `BackendSamplerV2` was changed so that it does clear the `QuantumCircuit.metadata` of its input circuits before passing them on to `backend.run` like it did previously.


wshanks:
These changes were needed to get the qiskit-experiments tests to pass in [this PR](https://github.com/qiskit-community/qiskit-experiments/pull/1470) that moves all circuit execution from `Backend.run` to `SamplerV2.run`, making use of `BackendSamplerV2` as the translation layer. We can add a release note and a test if needed but I wanted to see if these changes would be acceptable or not first. There are a couple other related issues/questions I had:

* Right now, `BackendSamplerV2` accepts an `options` dict and unpacks it into its `Options` class. These options are then passed through to `Backend.run` (except shots that is passed through only if it not overridden). This mechanism means that support for any additional arguments also requires changing the `Options` class again. Could all the options just be passed through besides shots?
* Secondary request to the previous one -- I can see why keeping to the `Options` format is more standard so it might not be acceptable to accept and pass through all options. If so, could we change the `__init__` to filter the keys in `options` and only pass the ones through to `Options` that have defined fields and warn about the ones that do not? This would make it a little easier to add new options in the future without breaking support for previous versions of Qiskit since an option not in the older `Options` definition will cause `Options(**options)` to fail and `Options` being a non-public class it is not supported to peak at its fields and filter before passing the options (edit: one could create a dummy `BackendSamplerV2` and check its options fields and use them to filter the options for the real `BackendSamplerV2` without accessing a private class).
* Caveat to the previous points -- maybe we won't need to add support for any more backend run options.
* For the qiskit-experiments tests, I also needed to delete this line:

  https://github.com/Qiskit/qiskit/blob/9a1d8d3aeaffff7e75dd3106b0317150e838aa53/qiskit/primitives/backend_estimator.py#L63

  Is there any way to work around that issue? Can we revisit if the metadata still needs to be blanked out like that or could we pass through some kind of `keep_metadata` option? The qiskit-experiments test backends use the metadata to determine what results to return for the circuits. We can look into how to refactor the experiments tests to get the metadata a different way if not.

wshanks:
https://github.com/Qiskit/qiskit-ibm-runtime/pull/1990 tries to make use of this change to translate between the IBM Quantum SamplerV2 options and common `BackendV2.run` options.

t-imamichi:
Is it possible to add some tests using `GenericBackendV2` or something? It would be nice to have some tests at least.

wshanks:
> Is it possible to add some tests using `GenericBackendV2` or something? It would be nice to have some tests at least.

There is not much support built in for testing with `meas_level=1` results. I am investigating what is feasible to add.

wshanks:
I added a simple test and also addressed the other points I had commented on in the way that works best for my case, but we can revisit them if those changes are a problem.

raynelfss:
Is this still on track for `1.3`? If so, who is interested in reviewing it? @Qiskit/qiskit-primitives

wshanks:
On my end, I think it is ready for review, so it would be great if it could still make 1.3.

qiskit/primitives/backend_sampler_v2.py
```diff
@@ -62,6 +68,16 @@ class _MeasureInfo:
     start: int
 
 
+ResultMemory = Union[list[str], list[list[float]], list[list[list[float]]]]
```
t-imamichi, line 71: I don't think you need to import `Union`. Could you verify it?
```suggestion
ResultMemory = list[str] | list[list[float]] | list[list[list[float]]]
```

t-imamichi:
Thank you for adding tests. It looks good overall. You don't have to add a test with a real device, but I would like to confirm that it works fine with real devices. Have you tested it yet?

qiskit/primitives/backend_sampler_v2.py
wshanks: I wrote it this way at first, but it raises a `TypeError` exception on Python 3.9. The type union feature was added in [Python 3.10](https://peps.python.org/pep-0604/). This syntax can be put in type signatures on 3.9 if `from __future__ import annotations` is used to skip type signature evaluation but the usage here is not inside of a type signature.

wshanks:
> Thank you for adding tests. It looks good overall. You don't have to add a test with a real device, but I would like to confirm that it works fine with real devices. Have you tested it yet?

My main interest is in using `BackendSamplerV2` with test backends in qiskit-experiments because we haven't had time to convert all the tests to use test samplers (maybe just using `BackendSamplerV2` with test backends is good enough) because support for running with `backend.run` on a real device is going away very soon. I was still able to do it though. Here is a quick test using the readout discriminator experiment from qiskit-experiments:

image

t-imamichi:
Thank you for sharing a result with a real device.

t-imamichi (approved):
LGTM

PR #13361, opened 2024-10-23
Evolved operator functions

Cryoris:
### Summary

Part of #13046: The `evolved_operator_ansatz` and `qaoa_ansatz` functions.

This is based on #13295, so that PR will need merging first.

### Details and comments

* After #13295, the `EvolvedOperatorAnsatz` is already pretty performant since it does the Pauli evolution in Rust. However, with these functions we still gain a speedup since we can do all evolution at once in Rust and don't need go back and forth.

    This benchmarks the construction of an evolved operator ansatz for an Ising Hamiltonian on a square grid with side length ``n``
(so ``n ** 2`` qubits), which has 2 operators; the ZZ-interactions and the X-field. For 100 qubits we get a speedup of 1.33x,
which is increasing with the number of qubits (e.g. we get 1.83x at 10'000 qubits).

    <details closed>

    ```
    -- n = 10, speedup 1.33x
    main: 0.033 +- 0.014
    this: 0.025 +- 0.009

    -- n = 50, speedup 1.64x
    main: 6.218 +- 0.200
    this: 3.792 +- 0.105

    -- n = 100, speedup 1.83x
    main: 100.487 +- 2.275
    this: 55.007 +- 1.139
    ```
    </details>

* This also adds a new function called `hamiltonian_variational_ansatz` (HVA), which automatically splits an input Hamiltonian into commuting terms. This is slightly different to `evolved_operator_ansatz`, which requires you to manually do this split, but IMO the HVA is the form most users would be interested in. We could've changed the default behavior of `evolved_operator_ansatz` to do the splitting, but I didn't want to change behavior for a class that users are already used to.


Cryoris:
I'm not sure why this fails, the docs build is complaining about 
```
/home/vsts/work/1/s/docs/apidoc/circuit_library.rst:: ERROR: Anonymous hyperlink mismatch: 1 references but 0 targets.
```
My best guess so far is that it doesn't like that the `evolved_operator_ansatz` function has the same name as the module its coming from?

releasenotes/notes/clib-evolved-ops-e91c00964c0209ce.yaml
```diff
@@ -0,0 +1,13 @@
+---
+features_circuits:
+  - |
+    Added :func:`.evolved_operator_ansatz`, :func:`.hamiltonian_variational_ansatz`, and
+    :func:`.qaoa_ansatz` to the circuit library to implement variational circuits based on
+    operator evolutions. :func:`.evolved_operator_ansatz` and :func:`.qaoa_ansatz` are
+    functionally equivalent to :class:`.EvolvedOperatorAnsatz` and :class:`.QAOAAnsatz`, but
+    generally more perfomant.
+
```
alexanderivrii: typo in "performant"

qiskit/circuit/library/n_local/evolved_operator_ansatz.py
```diff
@@ -15,16 +15,262 @@
...
+def hamiltonian_variational_ansatz(
+    hamiltonian: SparsePauliOp | Sequence[SparsePauliOp],
+    reps: int = 1,
+    insert_barriers: bool = False,
+    name: str = "HVA",
+    parameter_prefix: str = "t",
+):
+    r"""Construct a Hamiltonian variational ansatz.
+
+    For a Hamiltonian :math:`H = \sum_{k=1}^K H_k` where the terms :math:`H_k` consists of only
```
alexanderivrii:
```suggestion
    For a Hamiltonian :math:`H = \sum_{k=1}^K H_k` where the terms :math:`H_k` consist of only
```

alexanderivrii (commented):
I think this is in a good shape. Though I would like to take another look after #13295 is merged.

qiskit/circuit/library/n_local/evolved_operator_ansatz.py
Cryoris: Fixed in 6bf40b8d2

releasenotes/notes/clib-evolved-ops-e91c00964c0209ce.yaml
Cryoris: Fixed in 6bf40b8d2

crates/accelerate/src/circuit_library/utils.rs
```diff
@@ -0,0 +1,26 @@
...
+};
+use smallvec::SmallVec;
+
+pub(super) type StandardInstruction = (StandardGate, SmallVec<[Param; 3]>, SmallVec<[Qubit; 2]>);
+pub(super) type Instruction = (
+    PackedOperation,
+    SmallVec<[Param; 3]>,
+    Vec<Qubit>,
+    Vec<Clbit>,
+);
```
alexanderivrii: I like this very much! Do you think it would make sense / be possible to also use these definition from within the `synthesis` directory? Or should we move these to an even more central place?

qiskit/circuit/library/n_local/evolved_operator_ansatz.py
```diff
@@ -15,16 +15,262 @@
...
+    r"""Construct an ansatz out of operator evolutions.
+
+    For a set of operators :math:`[O_1, ..., O_J]` this circuit is defined as
+
+    .. math::
+
+        \prod_{r=1}^{R} \left( \prod_{j=J}^1 e^{i\theta_{j, r} O_j} \right)
+
+    where the exponentials :math:`exp(i\theta O_j)` are expanded using the product formula
+    specified by ``evolution``.
```
alexanderivrii: A few questions about this. In the formula, ``R`` corresponds to ``reps`` from the docstring? The $\theta_{j, r}$ actually depend on $r$? What is the relation between `evolved_operator_ansatz` and the `PauliEvolutionGate`? And in particular, why in `evolved_operator_ansatz` we have $e^{i\theta O_j}$ and in the definition of `PauliEvolutionGate` we have $e^{-it H}$, are the signs consistent? (My questions really show my ignorance of the subject.)

qiskit/circuit/library/n_local/evolved_operator_ansatz.py
```diff
@@ -15,16 +15,262 @@
...
+        insert_barriers: Whether to insert barriers in between each evolution.
+        name: The name of the circuit.
+        parameter_prefix: Set the names of the circuit parameters. If a string, the same prefix
+            will be used for each parameters. Can also be a list to specify a prefix per
+            operator.
+        remove_identities: If ``True``, ignore identity operators (note that we do not check
+            :class:`.Operator` inputs). This will also remove parameters associated with identities.
+        flatten: If ``True``, a flat circuit is returned instead of nesting it inside multiple
+            layers of gate objects. Setting this to ``False`` is significantly less performant,
+            especially for parameter binding, but can be desirable for a cleaner visualization.
```
alexanderivrii, line 89: Why`initial_state` is not an argument anymore?

qiskit/circuit/library/n_local/evolved_operator_ansatz.py
```diff
@@ -15,16 +15,262 @@
...
+    if isinstance(parameter_prefix, str):
+        parameters = ParameterVector(parameter_prefix, reps * num_operators)
+        param_iter = iter(parameters)
+    else:
+        # this creates the parameter vectors per operator, e.g.
+        #    [[a0, a1, a2, ...], [b0, b1, b2, ...], [c0, c1, c2, ...]]
+        # and turns them into an iterator
+        #    a0 -> b0 -> c0 -> a1 -> b1 -> c1 -> a2 -> ...
+        per_operator = [ParameterVector(prefix, reps).params for prefix in parameter_prefix]
+        param_iter = itertools.chain.from_iterable(zip(*per_operator))
```
alexanderivrii, line 124: Oh no, the iterators have spilled over from the Rust code into the Python code as well 😨.

test/python/circuit/library/test_pauli_feature_map.py
```diff
@@ -88,6 +88,8 @@ def test_pauli_evolution(self):
             evo.sxdg(1)
             evo.h(2)
 
+            print(evo.draw())
+
```
alexanderivrii: this print can be removed

qiskit/circuit/library/n_local/evolved_operator_ansatz.py
```diff
@@ -15,16 +15,262 @@
...
+            layers of gate objects. Setting this to ``False`` is significantly less performant,
+            especially for parameter binding, but can be desirable for a cleaner visualization.
+    """
+    if reps < 0:
+        raise ValueError("reps must be a non-negative integer.")
+
+    if isinstance(operators, BaseOperator):
+        operators = [operators]
+    elif len(operators) == 0:
+        return QuantumCircuit()
```
alexanderivrii, line 97: Previously having no iterators threw a `ValueError`, correct? Though I think the change is fine.

alexanderivrii (commented):
This looks good. I have left a few minor in-place comments and questions.

alexanderivrii:
@Cryoris, one question just came up in a slack discussion with @raynelfss: why do you need the Rust changes at all? I do like the (small) cleanup that moves `Instruction` and `StandardInstruction` to some common place, and I don't mind this being a part of this PR, but we technically don't need it, correct? Could you also please comment why we need the change to `radd_param`.

crates/accelerate/src/circuit_library/utils.rs
Cryoris: Yeah I think that would make sense, though we might want to discuss the name a bit if we move this to a central location, so we don't mistake it among all the other instructions 😄

qiskit/circuit/library/n_local/evolved_operator_ansatz.py
Cryoris: * yes, $R$ is `reps`: I'll add a comment
* `evolved_operator_ansatz` uses layers of `PauliEvolutionGate` to implement a sequence of parameterized evolutions. Indeed they use the _same_ sign, but this is wrongly documented 🙂 
* (you didn't ask but for completeness)`hamiltonian_variational_ansatz` is a special type of `evolved_operator_ansatz` where the operators are set to the commuting groups of a Hamiltonian

qiskit/circuit/library/n_local/evolved_operator_ansatz.py
Cryoris: Because one can simply do
```python
initial_state.compose(evolved_operator_ansatz(...))
```
Before this was an argument as the `EvolvedOperatorAnsatz` was a blueprint and could be reconstructed, in which case we needed to re-attach the initial state. Not being a blueprint anymore, this is no longer needed and we can simplify the code.

qiskit/circuit/library/n_local/evolved_operator_ansatz.py
Cryoris: Ha, nothing sneaks past you 😛  Though e.g. the n-local circuits would just be empty if no blocks are set. So I used this chance to make the functions more consistent 🙂

qiskit/circuit/library/n_local/evolved_operator_ansatz.py
Cryoris: Iter-ception! Which world do we iterate in? 🫨

Cryoris:
The changes to Rust are artifacts from developing off a non-merged PR 😄 The Rust changes are not needed, I'll revert them.

test/python/circuit/library/test_pauli_feature_map.py
Cryoris: removed in 3e009a5e7

qiskit/circuit/library/n_local/evolved_operator_ansatz.py
Cryoris: Added the above in 3e009a5e7

crates/accelerate/src/circuit_library/utils.rs
Cryoris: As per the comment below, all the Rust changes are reverted in 3e009a5e7

qiskit/circuit/library/n_local/qaoa_ansatz.py
```diff
@@ -21,8 +21,72 @@
...
+def qaoa_ansatz(
+    cost_operator: BaseOperator,
+    reps: int = 1,
+    initial_state: QuantumCircuit | None = None,
+    mixer_operator: BaseOperator | None = None,
+    insert_barriers: bool = False,
+    name: str = "QAOA",
+    flatten: bool = True,
+):
+    r"""A generalized QAOA quantum circuit with a support of custom initial states and mixers.
```
ShellyGarion, line 42: might be good to an some more explanation or example here

ShellyGarion (commented):
LGTM. I have a minor comment about the docs.

qiskit/circuit/library/n_local/qaoa_ansatz.py
ShellyGarion: Another comment - please note that these functions do not appear in the table with the classes, as in #13310

crates/accelerate/src/circuit_library/pauli_evolution.rs
```diff
@@ -27,6 +26,8 @@ type Instruction = (
     Vec<Clbit>,
 );
 
+// custom types for a more readable code
+
```
raynelfss: Maybe this line was supposed to be removed or back on line 21?

qiskit/circuit/library/n_local/evolved_operator_ansatz.py
```diff
@@ -15,16 +15,263 @@
...
+def evolved_operator_ansatz(
+    operators: BaseOperator | Sequence[BaseOperator],
+    reps: int = 1,
+    evolution: EvolutionSynthesis | None = None,
+    insert_barriers: bool = False,
+    name: str = "EvolvedOps",
+    parameter_prefix: str | Sequence[str] = "t",
+    remove_identities: bool = True,
+    flatten: bool | None = None,
+):
```
raynelfss: Since we're specifying types here, should we also say that it returns a `QuantumCircuit`?
```suggestion
def evolved_operator_ansatz(
    operators: BaseOperator | Sequence[BaseOperator],
    reps: int = 1,
    evolution: EvolutionSynthesis | None = None,
    insert_barriers: bool = False,
    name: str = "EvolvedOps",
    parameter_prefix: str | Sequence[str] = "t",
    remove_identities: bool = True,
    flatten: bool | None = None,
) -> QuantumCircuit:
```

qiskit/circuit/library/n_local/evolved_operator_ansatz.py
```diff
@@ -15,16 +15,263 @@
...
+    return circuit
+
+
+def hamiltonian_variational_ansatz(
+    hamiltonian: SparsePauliOp | Sequence[SparsePauliOp],
+    reps: int = 1,
+    insert_barriers: bool = False,
+    name: str = "HVA",
+    parameter_prefix: str = "t",
+):
```
raynelfss: You could add the return type here as well
```suggestion
) -> QuantumCircuit :
```

qiskit/circuit/library/n_local/qaoa_ansatz.py
```diff
@@ -21,8 +21,72 @@
...
+
+def qaoa_ansatz(
+    cost_operator: BaseOperator,
+    reps: int = 1,
+    initial_state: QuantumCircuit | None = None,
+    mixer_operator: BaseOperator | None = None,
+    insert_barriers: bool = False,
+    name: str = "QAOA",
+    flatten: bool = True,
+):
```
raynelfss:
```suggestion
) -> QuantumCircuit:
```

raynelfss (commented):
I'm not too familiar with PauliEvolution or what the `evolved_operator_ansatz` tries to achieve specifically but the logic here is sound to me. I'll let @alexanderivrii have the last word on approving, but great job regardless!

qiskit/circuit/library/n_local/qaoa_ansatz.py
Cryoris: done in 076f467

qiskit/circuit/library/n_local/evolved_operator_ansatz.py
Cryoris: done in 076f467

qiskit/circuit/library/n_local/evolved_operator_ansatz.py
Cryoris: done in 076f467

crates/accelerate/src/circuit_library/pauli_evolution.rs
Cryoris: done in 076f467

qiskit/circuit/library/n_local/qaoa_ansatz.py
Cryoris: done in 076f467

alexanderivrii (dismissed):
LGTM!

raynelfss (approved):
Looks okay, we'll fix the removals with #13407

Issue #13274, opened 2024-10-03
Add a Rust reperesentation of `Var` type


mtreinish:
Right now we have the `Var` type  for classical variables in a circuit defined in Python. These get used for expressions and store instructions currently, but their use will likely expand over time. However, even with the limited use currently their representation is critical in the `DAGCircuit` where they're one of the edge types so having them in Python means we relying on calling back to python to interact with them (or in the case of the `DAGCircuit` use them as keys in a hashmap).

eliarbel:
Done in #14068

PR #13364, opened 2024-10-23
Represent variables with BitData in DAGCircuit.

### LINKED ISSUE #13027: Store a `Var(u32)` type for the `Wire::Var` variant in the rust DAGCircuit

kevinhartman:
This PR builds on the work done by @jlapeyre (added as a co-author on the original commit). Thanks John! 😄

### Summary
Introduces a new `Var` type which, similarly to the existing `Qubit` and `Clbit` types, represents a variable in the DAG as an index. With this change, the size of the `Wire` type used for edge data in `DAGCircuit` should be halved. See #13027 for details.

### Details and comments
The semantics of the new `Var` type differ from those of `Qubit` and `Clbit`, in that the latter types the index has a meaning. As such, I'd think that `Var` should be completely opaque (and not constructible) outside of `qiskit-circuit`. Thoughts?

#### Todo
~- [ ] Change `BitData` to `WireData`.~
~- [ ] Resolve any visibility concerns noted above.~

> [!NOTE]
> We can resolve these to-dos in a follow-on PR, since a lot of these structures will change soon in anticipation of migrating toward a native Rust source of truth for bits.

Resolves #13027
Closes #13048

### DIFF
--- a/crates/circuit/src/dag_circuit.rs
+++ b/crates/circuit/src/dag_circuit.rs
@@ -10,7 +10,7 @@
 // copyright notice, and modified files need to carry a notice indicating
 // that they have been altered from the originals.
 
-use std::hash::{Hash, Hasher};
+use std::hash::Hash;
 
 use ahash::RandomState;
 use smallvec::SmallVec;
@@ -29,7 +29,7 @@ use crate::interner::{Interned, Interner};
 use crate::operations::{Operation, OperationRef, Param, PyInstruction, StandardGate};
 use crate::packed_instruction::{PackedInstruction, PackedOperation};
 use crate::rustworkx_core_vnext::isomorphism;
-use crate::{BitType, Clbit, Qubit, TupleLikeArg};
+use crate::{BitType, Clbit, Qubit, TupleLikeArg, Var};
 
 use hashbrown::{HashMap, HashSet};
 use indexmap::IndexMap;
@@ -79,8 +79,8 @@ pub enum NodeType {
     QubitOut(Qubit),
     ClbitIn(Clbit),
     ClbitOut(Clbit),
-    VarIn(PyObject),
-    VarOut(PyObject),
+    VarIn(Var),
+    VarOut(Var),
     Operation(PackedInstruction),
 }
 
@@ -97,45 +97,21 @@ impl NodeType {
     }
 }
 
-#[derive(Clone, Debug)]
+#[derive(Hash, Eq, PartialEq, Clone, Debug)]
 pub enum Wire {
     Qubit(Qubit),
     Clbit(Clbit),
-    Var(PyObject),
-}
-
-impl PartialEq for Wire {
-    fn eq(&self, other: &Self) -> bool {
-        match (self, other) {
-            (Wire::Qubit(q1), Wire::Qubit(q2)) => q1 == q2,
-            (Wire::Clbit(c1), Wire::Clbit(c2)) => c1 == c2,
-            (Wire::Var(v1), Wire::Var(v2)) => {
-                v1.is(v2) || Python::with_gil(|py| v1.bind(py).eq(v2).unwrap())
-            }
-            _ => false,
-        }
-    }
-}
-
-impl Eq for Wire {}
-
-impl Hash for Wire {
-    fn hash<H: Hasher>(&self, state: &mut H) {
-        match self {
-            Self::Qubit(qubit) => qubit.hash(state),
-            Self::Clbit(clbit) => clbit.hash(state),
-            Self::Var(var) => Python::with_gil(|py| var.bind(py).hash().unwrap().hash(state)),
-        }
-    }
+    Var(Var),
 }
 
 impl Wire {
     fn to_pickle(&self, py: Python) -> PyObject {
         match self {
-            Self::Qubit(bit) => (0, bit.0.into_py(py)).into_py(py),
-            Self::Clbit(bit) => (1, bit.0.into_py(py)).into_py(py),
-            Self::Var(var) => (2, var.clone_ref(py)).into_py(py),
+            Self::Qubit(bit) => (0, bit.0.into_py(py)),
+            Self::Clbit(bit) => (1, bit.0.into_py(py)),
+            Self::Var(var) => (2, var.0.into_py(py)),
         }
+        .into_py(py)
     }
 
     fn from_pickle(b: &Bound<PyAny>) -> PyResult<Self> {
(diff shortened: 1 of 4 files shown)


mtreinish:
> The semantics of the new Var type differ from those of Qubit and Clbit, in that the latter types the index has a meaning. As such, I'd think that Var should be completely opaque (and not constructible) outside of qiskit-circuit. Thoughts?

For right now I think this is fine. If we just can use it as a key to get the pyobject (or a reference to a rust struct after #13274) from a dag or get it as a return type from adding a new var to the dag that should be enough. I don't think people will be interacting with them directly or manipulating them like qubits.

crates/circuit/src/dag_circuit.rs
```diff
@@ -589,8 +503,14 @@ impl DAGCircuit {
...
-        out_dict.set_item("var_input_map", self.var_input_map.dict.clone_ref(py))?;
-        out_dict.set_item("var_output_map", self.var_output_map.dict.clone_ref(py))?;
+        out_dict.set_item(
+            "var_io_map",
+            self.var_io_map
+                .iter()
+                .enumerate()
+                .map(|(k, v)| (k, [v[0].index(), v[1].index()]))
+                .collect::<IndexMap<_, _, RandomState>>(),
+        )?;
```
raynelfss: **Nitpick:** For this part, could we instead build the `PyDict` directly? I feel like we'd avoid the conversion cost from `IndexMap` to `dict`.

crates/circuit/src/dag_circuit.rs
```diff
@@ -2618,8 +2545,18 @@ def _format(operand):
...
+                    .get(*var1)
+                    .unwrap()
+                    .bind(py)
+                    .eq(other.vars.get(*var2).unwrap()),
+                [NodeType::VarOut(var1), NodeType::VarOut(var2)] => self
+                    .vars
+                    .get(*var1)
+                    .unwrap()
+                    .bind(py)
+                    .eq(other.vars.get(*var2).unwrap()),
```
raynelfss, line 2647: Maybe I'm not getting this right so feel free to correct me, the comparison here is not done by index because `Var` objects are not opaque. Meaning that they are different fundamentally speaking and are not differentiated by index only. Is this why we're comparing them this way here?

crates/circuit/src/dag_circuit.rs
```diff
@@ -5843,12 +5786,16 @@ impl DAGCircuit {
...
+            NodeType::VarIn(var) => Py::new(
+                py,
+                DAGInNode::new(py, id, self.vars.get(*var).unwrap().clone_ref(py)),
+            )?
+            .into_any(),
+            NodeType::VarOut(var) => Py::new(
+                py,
+                DAGOutNode::new(py, id, self.vars.get(*var).unwrap().clone_ref(py)),
+            )?
+            .into_any(),
```
raynelfss, line 5889: Is there any difference between using this and `into_py()`?

crates/circuit/src/dag_circuit.rs
```diff
@@ -5945,17 +5892,19 @@ impl DAGCircuit {
...
 
         let next_layer = graph_layers.next();
         match next_layer {
             Some(layer) => Box::new(layer.into_iter().filter(|node| {
                 matches!(self.dag.node_weight(*node).unwrap(), NodeType::Operation(_))
-            })),
-            None => Box::new(vec![].into_iter()),
+            }))
+            .collect(),
+            None => vec![],
```
raynelfss: Is there any reason we're ditching the iterators here?

raynelfss (commented):
I appreciate the PR in many personal levels (look at all the `py` we got rid of). This looks great, I have a couple in-line comments/questions.

crates/circuit/src/dag_circuit.rs
kevinhartman: Pretty much, though I'd say they _are_ opaque in that they're nothing more than keys that users must exchange with the `DAGCircuit` to get information about the actual variable.

The difference between `Var` and `{Qu,Cl}bit` is pretty much that a variable's index doesn't mean _anything_. With bits, the index is always also the position within the DAG's bit list, be it qubits or clbits, so when comparing two DAGs it's enough to just compare the indices. A `Var`'s index doesn't mean anything because the relationship between `Var` and our DAG is that a DAG has a *set* of `Var`s, not a list. We just happen to be storing them in a list-like structure (`BitData`).

I'm glad you mentioned this, because I think it really means we ought to make sure that `Var` isn't `Eq` or `Ord` outside of `DAGCircuit`, since it doesn't actually have those semantics. The right thing to do is probably to make both `Wire` and `Var` private to `DAGCircuit` and if we must expose a wire-like enum, have it be an enumeration of `Qubit`, `Clbit`, and `PyObject` like it is now, and convert to it within `DAGCircuit` from the internal `Wire` enum of `Qubit`, `Clbit`, `Var` at the boundary.

I'll look at the feasibility of doing this as part of this PR and report back.

crates/circuit/src/dag_circuit.rs
kevinhartman: I think `into_any` is the best choice here from an API perspective since it doesn't need a `Python` token. If we used `into_py` instead, then we'd be calling the impl of `IntoPy` for `Py<T>` which actually just calls `into_any` anyway ^^. So there's no difference, except I guess an extra function call which would almost certainly get optimized out anyhow.

https://docs.rs/pyo3/latest/src/pyo3/instance.rs.html#1719-1726

raynelfss: Oh, my apologies for the confusion. What I meant was instead of declaring a `Py {*//*}`. We could just call `DAGOutNode.into_py(py)` I don't know if there are any limitations and why you're choosing your method. Just curious.

crates/circuit/src/dag_circuit.rs
kevinhartman: I mostly just don't like using `Box` unless it's an absolute must, and while updating this signature, I noticed that this function just emits from a single `Vec` (the second one) of `multigraph_layers`, so doing this doesn't interrupt any sort of lazy iteration that we might otherwise benefit from.

But, we can actually do this without the `collect` and without disrupting any would-be iterators, which I've just done now in 1254ba1 😄.

crates/circuit/src/dag_circuit.rs
kevinhartman: Seems like a good idea to me. I was just copying what was done above, but I made the change to all instances where we were doing that in 72170a5.

kevinhartman:
I am trying locally to provide better encapsulation for the new `Var` as described [in this PR comment thread](https://github.com/Qiskit/qiskit/pull/13364#discussion_r1817010432). Please hold off on merging until that is sorted and I've reported back 🙂.

kevinhartman:
> I am trying locally to provide better encapsulation for the new Var as described https://github.com/Qiskit/qiskit/pull/13364#discussion_r1817010432. Please hold off on merging until that is sorted and I've reported back 🙂.

The best I'm able to do at the moment is to move `Var` to `dagcircuit.rs`, which allows it to be _mostly_ opaque to clients of the DAG. I believe this is needed for a proper abstraction, but it is not perfect in its current form in that conversion into `Var` from `BitType` is public (because both `BitType` and `From` (of course) are public). In what may become a follow on PR, I'm going to explore making `BitType` private to circumvent this, but the current PR is now ready.

crates/circuit/src/dag_circuit.rs
```diff
@@ -5843,12 +5830,16 @@ impl DAGCircuit {
...
+            NodeType::VarIn(var) => Py::new(
+                py,
+                DAGInNode::new(py, id, self.vars.get(*var).unwrap().clone_ref(py)),
+            )?
+            .into_any(),
+            NodeType::VarOut(var) => Py::new(
+                py,
+                DAGOutNode::new(py, id, self.vars.get(*var).unwrap().clone_ref(py)),
+            )?
+            .into_any(),
```
raynelfss, line 5889: Bit of a nitpick since from what it seems this would have no big effects but you could rewrite it this way (which I alluded to before):
```suggestion
            NodeType::VarIn(var) => DAGInNode::new(py, id, self.vars.get(*var).unwrap().clone_ref(py)).into_py(py),
            NodeType::VarOut(var) => DAGOutNode::new(py, id, self.vars.get(*var).unwrap().clone_ref(py)).into_py(py),
```

It might be too long for a one-liner though 😅

raynelfss (approved):
LGTM! I just had one small optional nitpick.

crates/circuit/src/dag_circuit.rs
kevinhartman: I rather like that approach! Though, IMO we ought to change over to using it consistently in a separate PR (if we feel the improved readability is worth the effort).

The reason being that it keeps the diff for this PR's commit onto `main` as logically specific / isolated as possible. This way, if there's an issue in the future (or someone wants to see exactly the changes that were strictly necessary to do XYZ), there isn't any extra fluff to sift through when they're looking through the Git history.

I'd be more inclined to use the approach you mention despite this if there weren't other uses of `Py::new` inside this very method, but since there are I'd be touching lines that do not require modifications to achieve what this PR aims to do.

raynelfss: Alright, then, this looks good either way, good job!

PR #13367, opened 2024-10-24
Deprecate `BasicSimulator.configuration`

### LINKED ISSUE #12843: BasicSimulator uses deprecated BackendConfiguration
While deprecating the models in `qiskit.providers.models` I realized that BasicSimulator uses `BackendConfiguration`, which is one of those models.

ElePT:
### Summary
Closes #12843.

### DIFF
--- a/qiskit/compiler/assembler.py
+++ b/qiskit/compiler/assembler.py
@@ -189,6 +189,16 @@ def assemble(
     )
 
 
+# Note for future: this method is used in `BasicSimulator` and may need to be kept past the
+# `assemble` removal deadline (2.0). If it is kept (potentially in a different location),
+# we will need an alternative for the backend.configuration() access that currently takes
+# place in L566 (`parse_circuit_args`) and L351 (`parse_common_args`)
+# because backend.configuration() is also set for removal in 2.0.
+# The ultimate goal will be to move away from relying on any kind of `assemble` implementation
+# because of how tightly coupled it is to these legacy data structures. But as a transition step,
+# given that we would only have to support the subcase of `BasicSimulator`, we could probably just
+# inline the relevant config values that are already hardcoded in the basic simulator configuration
+# generator.
 def _assemble(
     experiments: Union[
         QuantumCircuit,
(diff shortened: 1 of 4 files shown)


qiskit/compiler/assembler.py
```diff
@@ -189,6 +189,16 @@ def assemble(
...
+# Note for future: this method is used in `BasicSimulator` and may need to be kept past the
+# `assemble` removal deadline (2.0). If it is kept (potentially in a different location),
+# we will need an alternative for the backend.configuration() access that currently takes
+# place in L566 (`parse_circuit_args`) and L351 (`parse_common_args`)
+# because backend.configuration() is also set for removal in 2.0.
+# The ultimate goal will be to move away from relying on any kind of `assemble` implementation
+# because of how tightly coupled it is to these legacy data structures. But as a transition step,
+# given that we would only have to support the subcase of `BasicSimulator`, we could probably just
+# inline the relevant config values that are already hardcoded in the basic simulator configuration
+# generator.
```
mtreinish, line 201: FWIW, it should not be hard to rewrite the `run_experiment` method of basic simulator to avoid needing qobj. The assemble usage in there is kind of a weird legacy from `BaseBackend` which required a qobj get passed to the `run()` method.

releasenotes/notes/deprecate-basic-simulator-configuration-9d782925196993e9.yaml
```diff
@@ -0,0 +1,29 @@
...
+      backend.configuration().coupling_map       backend.target.build_coupling_map()
+      backend.configuration().local              No representation
+      backend.configuration().simulator          No representation
+      backend.configuration().conditional        No representation
+      backend.configuration().open_pulse         No representation     
+      backend.configuration().memory             No representation      
+      backend.configuration().max_shots          No representation      
+
+    (*) Note that ``backend.target.operation_names`` includes ``basis_gates`` and additional 
+    non-gate instructions, in some implementations it might be necessary to filter the output.
```
mtreinish: Is it worth just linking to https://github.com/Qiskit/qiskit/blob/main/qiskit/providers/__init__.py#L679 here too?

mtreinish (dismissed):
This is a fun cycle of deprecations getting emitted between assemble and the backend. Personally I'm not sure the filterwarnings inline in code because that can hide things when we need to rely on it later. But in this case I think it's fine especially as all of this functionality will be removed at a high level in 2.0.

The only comment I really had was about maybe adding a link to the backend version migration guide in the release note. Although it is also maybe a question of whether we need to improve the docs in the migration guide too? (like is it missing something). But if you're happy with the release note or want to fix them in a follow up, potentially post rc1 that's fine and feel free to enqueue this for merging.

releasenotes/notes/deprecate-basic-simulator-configuration-9d782925196993e9.yaml
ElePT: There isn't a lot of extra information but I added it in case the migration guide gets improvements (I know there's an open issue for that).

mtreinish (approved):

Issue #12007, opened 2024-03-13
Add parallel synthesis interface to default unitary synthesis plugin


mtreinish:
Once we have #12004 (and also for #12005 too) implemented we can better leverage available CPU resources better by running synthesis in parallel using a rayon iterator. There is no data dependency between doing unitary synthesis for different unitary matrices, it should be fairly straightforward to have a function that takes in a list of unitary matrices with their target qubits and also dict/hashmap of target qubits to the appropriate decomposer class and then have the function return a list of circuit sequences which is computed using a multithreaded rayon iterator. This can be connected #12006 to have the default synthesis plugin prepare a batch of all the 2 qubit unitary matrices to exploit this function.

mtreinish:
I think this might be superseded by the suggestions in #11659. If we build out a new transpiler pass to do efficient 2q peephole unitary resynthesis we should do what this is proposing. The only case where I think this wouldn't be the case is if we ran `UnitarySynthesis` for basis translation or unrolling >=3q unitaries. So I'll keep this open just in case we want the parallel interface even though the primary utility I envisioned was for peephole optimization.

Issue #11659, opened 2024-01-28
`ConsolidateBlocks` does not have a good logic for heterogeneous gates


ajavadia:
### What should we add?

`ConsolidateBlocks` has some logic for choosing whether to collapse some blocks into a UnitaryGate. But this is pretty outdated by now. It basically checks whether number of gates in the decomposition improves. First, number of gates is not necessarily important, but rather the error is. Second, it does not currently deal with multiple (heterogeneous) possible decompositions.

But all of this is implemented correctly in UnitarySynthesis (at least for 2q blocks). So ConsolidateBlocks should just defer to UnitarySynthesis for when and how to resynthesize a sequence of 2q gates. All of its decomposition considerations should come from UnitarySynthesis.

I think it is better to write a new pass `PeepholeUnitaryResynthesis`, which does all 3 of these actions: `Collect2QBlocks`, `ConsolidateBlocks`, `UnitarySynthesis`. The logic must be consistent, so there's no point splitting these 3 stages. 
I believe this can replace the `UnitarySynthesis` pass because any Unitary can be considered a simple peephole unitary.

(** note: currently if the user knows that there's a good chance that UnitarySynthesis improves the circuit, they can force it to occur by adding `[Collect2QBlocks(target=target), ConsolidateBlocks(force_consolidate=True), UnitarySynthesis(target=target)]` to the passmanager, so it is possible to customize this by a user who knows how to use the passmanager)

mtreinish:
I like this idea in general. I'm thinking of how it relates to #8774 (specifically the #12007 sub-task) and we can sidestep the need to add a batch mode to the unitary synthesis plugin interface by doing this all at once in multithreaded rust in a new pass.

The only question I have though is in evaluating the error for the original 2q block. I agree that we should use an estimated error heuristic to evaluate a potential decompositions and select one based on that instead of the number of gates (which is just being used a proxy for estimated error rate).  But prior to synthesis there isn't a guarantee that the gates in a 2q block are in target instructions that we can query error rates on. How were you thinking we'd evaluate the block in these cases? Because I was reading this is as we we compare the error estimates for the original circuit against all the possible decompositions and pick the one which results in the lower error. I guess the answer is if the block isn't in target native instructions we always need to synthesize so in those cases we pick the lowest error decomposition?

PR #13368, opened 2024-10-24
Oxidize the ConsolidateBlocks pass

### LINKED ISSUE #12250: Port `ConsolidateBlocks` to Rust
The core linalg for this pass are already in Rust, but once we have #11721 and #12205  we should move the entire contents of the pass to Rust.

mtreinish:
### Summary

This commit ports the consolidate blocks pass to rust. The logic remains the same and this is just a straight porting. One optimization is that to remove the amount of python processing the Collect2qBlocks pass is no longer run as part of the preset pass managers and this is called directly in rust. This speeds up the pass because it avoids 3 crossing of the language boundary and also the intermediate creation of DAGNode python objects. The pass still supports running with Collect2qBlocks for backwards compatibility and will skip running the pass equivalent internally the field is present in the property set.

There are potential improvements that can be investigated here such as avoiding in place dag contraction and moving to rebuilding the dag iteratively. Also changing the logic around estimated error (see #11659) to be more robust. But these can be left for follow up PRs as they change the logic.

Realistically we should look at combining ConsolidateBlocks for it's current two usages with Split2qUnitaries and UnitarySynthesis into those passes for more efficiency. We can improve the performance and logic as part of that refactor. See #12007 for more details on this for UnitarySynthesis.

### Details and comments

Closes #12250

TODO:

- [x] Fix test failures
- [x] Add release note for support for running without Collect2qBlocks
- [x] Benchmark and tune


mtreinish:
While there are still 4 tests to fix here, I did a quick asv run to get a feel for the speedup so far and it yielded:
```

| Change   | Before [2284f192] <consolidate-blocks~1>   | After [ed2b41b5] <consolidate-blocks>   |   Ratio | Benchmark (Parameter)                                                                           |
|----------|--------------------------------------------|-----------------------------------------|---------|-------------------------------------------------------------------------------------------------|
| -        | 1.77±0s                                    | 1.60±0.01s                              |    0.91 | utility_scale.UtilityScaleBenchmarks.time_qft('cz')                                             |
| -        | 530±4ms                                    | 474±5ms                                 |    0.89 | transpiler_levels.TranspilerLevelBenchmarks.time_quantum_volume_transpile_50_x_20(3)            |
| -        | 1.77±0.01s                                 | 1.58±0.01s                              |    0.89 | utility_scale.UtilityScaleBenchmarks.time_qft('ecr')                                            |
| -        | 50.1±0.5ms                                 | 44.0±0.3ms                              |    0.88 | transpiler_levels.TranspilerLevelBenchmarks.time_transpile_qv_14_x_14(2)                        |
| -        | 75.5±0.6ms                                 | 66.5±0.4ms                              |    0.88 | transpiler_levels.TranspilerLevelBenchmarks.time_transpile_qv_14_x_14(3)                        |
| -        | 312±2ms                                    | 271±2ms                                 |    0.87 | transpiler_levels.TranspilerLevelBenchmarks.time_quantum_volume_transpile_50_x_20(2)            |
| -        | 416±2ms                                    | 359±4ms                                 |    0.86 | utility_scale.UtilityScaleBenchmarks.time_qaoa('cx')                                            |
| -        | 27.0±0.3ms                                 | 22.6±0.3ms                              |    0.84 | transpiler_levels.TranspilerLevelBenchmarks.time_transpile_from_large_qasm_backend_with_prop(2) |
| -        | 28.2±0.5ms                                 | 23.5±0.4ms                              |    0.83 | transpiler_levels.TranspilerLevelBenchmarks.time_transpile_from_large_qasm_backend_with_prop(3) |
| -        | 1.28±0.01s                                 | 1.06±0.01s                              |    0.83 | utility_scale.UtilityScaleBenchmarks.time_qv('cx')                                              |
| -        | 223±4ms                                    | 185±2ms                                 |    0.83 | utility_scale.UtilityScaleBenchmarks.time_square_heisenberg('cx')                               |
| -        | 649±3ms                                    | 531±8ms                                 |    0.82 | utility_scale.UtilityScaleBenchmarks.time_qaoa('cz')                                            |
| -        | 276±0.5ms                                  | 227±3ms                                 |    0.82 | utility_scale.UtilityScaleBenchmarks.time_square_heisenberg('cz')                               |
| -        | 279±0.9ms                                  | 228±3ms                                 |    0.82 | utility_scale.UtilityScaleBenchmarks.time_square_heisenberg('ecr')                              |
| -        | 22.3±0.3ms                                 | 18.0±0.2ms                              |    0.81 | transpiler_levels.TranspilerLevelBenchmarks.time_transpile_from_large_qasm(3)                   |
| -        | 637±4ms                                    | 512±3ms                                 |    0.8  | utility_scale.UtilityScaleBenchmarks.time_qaoa('ecr')                                           |
| -        | 1.74±0.01s                                 | 1.39±0.01s                              |    0.8  | utility_scale.UtilityScaleBenchmarks.time_qv('cz')                                              |
| -        | 1.67±0.01s                                 | 1.33±0.01s                              |    0.8  | utility_scale.UtilityScaleBenchmarks.time_qv('ecr')                                             |
| -        | 21.5±0.5ms                                 | 17.1±0.3ms                              |    0.79 | transpiler_levels.TranspilerLevelBenchmarks.time_transpile_from_large_qasm(2)                   |
| -        | 11.0±0.2ms                                 | 4.24±0.04ms                             |    0.38 | passes.Collect2QPassBenchmarks.time_consolidate_blocks(5, 1024)                                 |
| -        | 29.8±0.09ms                                | 10.2±0.04ms                             |    0.34 | passes.Collect2QPassBenchmarks.time_consolidate_blocks(14, 1024)                                |

| Change   | Before [2284f192] <consolidate-blocks~1>   | After [ed2b41b5] <consolidate-blocks>   | Ratio   | Benchmark (Parameter)                                                                                  |
|----------|--------------------------------------------|-----------------------------------------|---------|--------------------------------------------------------------------------------------------------------|
|          | failed                                     | failed                                  | n/a     | passes.Collect2QPassBenchmarks.time_consolidate_blocks(20, 1024)                                       |
|          | 0                                          | 0                                       | n/a     | utility_scale.UtilityScaleBenchmarks.track_bvlike_depth('cx')                                          |
|          | 0                                          | 0                                       | n/a     | utility_scale.UtilityScaleBenchmarks.track_bvlike_depth('cz')                                          |
|          | 0                                          | 0                                       | n/a     | utility_scale.UtilityScaleBenchmarks.track_bvlike_depth('ecr')                                         |
|          | 3.72±0.02s                                 | 3.90±0.2s                               | 1.05    | utility_scale.UtilityScaleBenchmarks.time_circSU2('cx')                                                |
|          | 9.34±0.06ms                                | 9.74±0.05ms                             | 1.04    | utility_scale.UtilityScaleBenchmarks.time_bvlike('cx')                                                 |
|          | 9.41±0.08ms                                | 9.83±0.2ms                              | 1.04    | utility_scale.UtilityScaleBenchmarks.time_bvlike('cz')                                                 |
|          | 70.9±0.7ms                                 | 72.8±0.3ms                              | 1.03    | transpiler_levels.TranspilerLevelBenchmarks.time_schedule_qv_14_x_14(1)                                |
|          | 9.62±0.06ms                                | 9.92±0.1ms                              | 1.03    | utility_scale.UtilityScaleBenchmarks.time_parse_qaoa_n100('cz')                                        |
|          | 102±1ms                                    | 105±0.4ms                               | 1.03    | utility_scale.UtilityScaleBenchmarks.time_parse_qft_n100('ecr')                                        |
|          | 397                                        | 407                                     | 1.03    | utility_scale.UtilityScaleBenchmarks.track_bv_100_depth('cz')                                          |
|          | 397                                        | 407                                     | 1.03    | utility_scale.UtilityScaleBenchmarks.track_bv_100_depth('ecr')                                         |
|          | 73.4±0.5ms                                 | 75.0±0.6ms                              | 1.02    | transpiler_levels.TranspilerLevelBenchmarks.time_schedule_qv_14_x_14(0)                                |
|          | 35.7±0.3ms                                 | 36.5±0.6ms                              | 1.02    | transpiler_levels.TranspilerLevelBenchmarks.time_transpile_qv_14_x_14(1)                               |
|          | 9.45±0.06ms                                | 9.66±0.02ms                             | 1.02    | utility_scale.UtilityScaleBenchmarks.time_bvlike('ecr')                                                |
|          | 9.70±0.02ms                                | 9.88±0.08ms                             | 1.02    | utility_scale.UtilityScaleBenchmarks.time_parse_qaoa_n100('cx')                                        |
|          | 103±0.5ms                                  | 105±0.3ms                               | 1.02    | utility_scale.UtilityScaleBenchmarks.time_parse_qft_n100('cz')                                         |
|          | 33.5±0.2ms                                 | 34.2±0.2ms                              | 1.02    | utility_scale.UtilityScaleBenchmarks.time_parse_square_heisenberg_n100('cx')                           |
|          | 33.7±0.3ms                                 | 34.2±0.3ms                              | 1.02    | utility_scale.UtilityScaleBenchmarks.time_parse_square_heisenberg_n100('cz')                           |
|          | 33.5±0.2ms                                 | 34.2±0.1ms                              | 1.02    | utility_scale.UtilityScaleBenchmarks.time_parse_square_heisenberg_n100('ecr')                          |
|          | 35.5±0.1ms                                 | 36.0±0.9ms                              | 1.01    | transpiler_levels.TranspilerLevelBenchmarks.time_transpile_from_large_qasm(1)                          |
|          | 29.5±0.1ms                                 | 29.8±0.2ms                              | 1.01    | transpiler_levels.TranspilerLevelBenchmarks.time_transpile_qv_14_x_14(0)                               |
|          | 9.73±0.07ms                                | 9.82±0.08ms                             | 1.01    | utility_scale.UtilityScaleBenchmarks.time_parse_qaoa_n100('ecr')                                       |
|          | 103±0.7ms                                  | 104±0.2ms                               | 1.01    | utility_scale.UtilityScaleBenchmarks.time_parse_qft_n100('cx')                                         |
|          | 185±2ms                                    | 185±0.4ms                               | 1.00    | transpiler_levels.TranspilerLevelBenchmarks.time_quantum_volume_transpile_50_x_20(0)                   |
|          | 43.0±0.4ms                                 | 43.0±0.3ms                              | 1.00    | transpiler_levels.TranspilerLevelBenchmarks.time_transpile_from_large_qasm_backend_with_prop(0)        |
|          | 1404                                       | 1404                                    | 1.00    | transpiler_levels.TranspilerLevelBenchmarks.track_depth_quantum_volume_transpile_50_x_20(0)            |
|          | 1403                                       | 1403                                    | 1.00    | transpiler_levels.TranspilerLevelBenchmarks.track_depth_quantum_volume_transpile_50_x_20(1)            |
|          | 1323                                       | 1323                                    | 1.00    | transpiler_levels.TranspilerLevelBenchmarks.track_depth_quantum_volume_transpile_50_x_20(2)            |
|          | 1296                                       | 1296                                    | 1.00    | transpiler_levels.TranspilerLevelBenchmarks.track_depth_quantum_volume_transpile_50_x_20(3)            |
|          | 2705                                       | 2705                                    | 1.00    | transpiler_levels.TranspilerLevelBenchmarks.track_depth_transpile_from_large_qasm(0)                   |
|          | 2005                                       | 2005                                    | 1.00    | transpiler_levels.TranspilerLevelBenchmarks.track_depth_transpile_from_large_qasm(1)                   |
|          | 7                                          | 7                                       | 1.00    | transpiler_levels.TranspilerLevelBenchmarks.track_depth_transpile_from_large_qasm(2)                   |
|          | 7                                          | 7                                       | 1.00    | transpiler_levels.TranspilerLevelBenchmarks.track_depth_transpile_from_large_qasm(3)                   |
|          | 2705                                       | 2705                                    | 1.00    | transpiler_levels.TranspilerLevelBenchmarks.track_depth_transpile_from_large_qasm_backend_with_prop(0) |
|          | 2005                                       | 2005                                    | 1.00    | transpiler_levels.TranspilerLevelBenchmarks.track_depth_transpile_from_large_qasm_backend_with_prop(1) |
|          | 7                                          | 7                                       | 1.00    | transpiler_levels.TranspilerLevelBenchmarks.track_depth_transpile_from_large_qasm_backend_with_prop(2) |
|          | 7                                          | 7                                       | 1.00    | transpiler_levels.TranspilerLevelBenchmarks.track_depth_transpile_from_large_qasm_backend_with_prop(3) |
|          | 465                                        | 465                                     | 1.00    | transpiler_levels.TranspilerLevelBenchmarks.track_depth_transpile_qv_14_x_14(0)                        |
|          | 336                                        | 336                                     | 1.00    | transpiler_levels.TranspilerLevelBenchmarks.track_depth_transpile_qv_14_x_14(1)                        |
|          | 327                                        | 327                                     | 1.00    | transpiler_levels.TranspilerLevelBenchmarks.track_depth_transpile_qv_14_x_14(2)                        |
|          | 272                                        | 272                                     | 1.00    | transpiler_levels.TranspilerLevelBenchmarks.track_depth_transpile_qv_14_x_14(3)                        |
|          | 395                                        | 395                                     | 1.00    | utility_scale.UtilityScaleBenchmarks.track_bv_100_depth('cx')                                          |
|          | 300                                        | 300                                     | 1.00    | utility_scale.UtilityScaleBenchmarks.track_circSU2_depth('cx')                                         |
|          | 300                                        | 300                                     | 1.00    | utility_scale.UtilityScaleBenchmarks.track_circSU2_depth('cz')                                         |
|          | 300                                        | 300                                     | 1.00    | utility_scale.UtilityScaleBenchmarks.track_circSU2_depth('ecr')                                        |
|          | 1607                                       | 1607                                    | 1.00    | utility_scale.UtilityScaleBenchmarks.track_qaoa_depth('cx')                                            |
|          | 1622                                       | 1622                                    | 1.00    | utility_scale.UtilityScaleBenchmarks.track_qaoa_depth('cz')                                            |
|          | 1622                                       | 1622                                    | 1.00    | utility_scale.UtilityScaleBenchmarks.track_qaoa_depth('ecr')                                           |
|          | 1954                                       | 1954                                    | 1.00    | utility_scale.UtilityScaleBenchmarks.track_qft_depth('cx')                                             |
|          | 1954                                       | 1954                                    | 1.00    | utility_scale.UtilityScaleBenchmarks.track_qft_depth('cz')                                             |
|          | 1954                                       | 1954                                    | 1.00    | utility_scale.UtilityScaleBenchmarks.track_qft_depth('ecr')                                            |
|          | 2709                                       | 2709                                    | 1.00    | utility_scale.UtilityScaleBenchmarks.track_qv_depth('cx')                                              |
|          | 2709                                       | 2709                                    | 1.00    | utility_scale.UtilityScaleBenchmarks.track_qv_depth('cz')                                              |
|          | 2709                                       | 2709                                    | 1.00    | utility_scale.UtilityScaleBenchmarks.track_qv_depth('ecr')                                             |
|          | 462                                        | 462                                     | 1.00    | utility_scale.UtilityScaleBenchmarks.track_square_heisenberg_depth('cx')                               |
|          | 462                                        | 462                                     | 1.00    | utility_scale.UtilityScaleBenchmarks.track_square_heisenberg_depth('cz')                               |
|          | 462                                        | 462                                     | 1.00    | utility_scale.UtilityScaleBenchmarks.track_square_heisenberg_depth('ecr')                              |
|          | 192±0.7ms                                  | 192±2ms                                 | 0.99    | transpiler_levels.TranspilerLevelBenchmarks.time_quantum_volume_transpile_50_x_20(1)                   |
|          | 32.2±0.3ms                                 | 31.8±0.4ms                              | 0.99    | transpiler_levels.TranspilerLevelBenchmarks.time_transpile_from_large_qasm(0)                          |
|          | 47.6±0.09ms                                | 47.3±0.5ms                              | 0.99    | transpiler_levels.TranspilerLevelBenchmarks.time_transpile_from_large_qasm_backend_with_prop(1)        |
|          | 3.88±0.06s                                 | 3.75±0.01s                              | 0.97    | utility_scale.UtilityScaleBenchmarks.time_circSU2('ecr')                                               |
|          | 168±0.9ms                                  | 158±0.9ms                               | 0.94    | utility_scale.UtilityScaleBenchmarks.time_bv_100('cx')                                                 |
|          | 179±2ms                                    | 168±0.9ms                               | 0.94    | utility_scale.UtilityScaleBenchmarks.time_bv_100('cz')                                                 |
|          | 179±1ms                                    | 168±0.7ms                               | 0.94    | utility_scale.UtilityScaleBenchmarks.time_bv_100('ecr')                                                |
|          | 3.93±0.06s                                 | 3.71±0.1s                               | 0.94    | utility_scale.UtilityScaleBenchmarks.time_circSU2('cz')                                                |
|          | 1.41±0.01s                                 | 1.32±0s                                 | 0.94    | utility_scale.UtilityScaleBenchmarks.time_qft('cx')                                                    |
```

In general there is only so much we'll be able to do on the performance here because we'll be bottlenecked on the dag manipulation and `UnitaryGate` object creation. I think we can work on fixing those in follow ups separately. The dag manipulation can be removed as part of something like: https://github.com/Qiskit/qiskit/commit/bd43c51cb7331b8a50b3ad271e575f005c67efc6 (and an equivalent for `Split2qUnitaries`) where we rebuild the dag instead of doing in place substitution.

mtreinish:
This should be good to review now. There might be some benchmarking/profiling and tuning we want to do, but it's not a blocker.

mtreinish:
After the most recent round of changes the overall benchmarking results look like:

```

| Change   | Before [f2e07bc5] <main>               | After [a4229901] <consolidate-blocks>   |   Ratio | Benchmark (Parameter)                                                                           |
|----------|----------------------------------------|-----------------------------------------|---------|-------------------------------------------------------------------------------------------------|      
| -        | 1.77±0.01s                             | 1.57±0s                                 |    0.89 | utility_scale.UtilityScaleBenchmarks.time_qft('cz')                                             |      
| -        | 310±2ms                                | 272±1ms                                 |    0.88 | transpiler_levels.TranspilerLevelBenchmarks.time_quantum_volume_transpile_50_x_20(2)            |      
| -        | 49.8±0.7ms                             | 43.8±0.6ms                              |    0.88 | transpiler_levels.TranspilerLevelBenchmarks.time_transpile_qv_14_x_14(2)                        |      
| -        | 1.77±0.01s                             | 1.56±0.01s                              |    0.88 | utility_scale.UtilityScaleBenchmarks.time_qft('ecr')                                            |      
| -        | 535±1ms                                | 463±3ms                                 |    0.87 | transpiler_levels.TranspilerLevelBenchmarks.time_quantum_volume_transpile_50_x_20(3)            |      
| -        | 27.8±1ms                               | 23.4±0.4ms                              |    0.84 | transpiler_levels.TranspilerLevelBenchmarks.time_transpile_from_large_qasm_backend_with_prop(2) |      
| -        | 76.9±2ms                               | 64.6±2ms                                |    0.84 | transpiler_levels.TranspilerLevelBenchmarks.time_transpile_qv_14_x_14(3)                        |      
| -        | 1.26±0.01s                             | 1.07±0.01s                              |    0.84 | utility_scale.UtilityScaleBenchmarks.time_qv('cx')                                              |      
| -        | 28.2±2ms                               | 23.4±0.3ms                              |    0.83 | transpiler_levels.TranspilerLevelBenchmarks.time_transpile_from_large_qasm_backend_with_prop(3) |      
| -        | 417±3ms                                | 347±0.9ms                               |    0.83 | utility_scale.UtilityScaleBenchmarks.time_qaoa('cx')                                            |      
| -        | 222±2ms                                | 183±3ms                                 |    0.83 | utility_scale.UtilityScaleBenchmarks.time_square_heisenberg('cx')                               |      
| -        | 279±4ms                                | 229±3ms                                 |    0.82 | utility_scale.UtilityScaleBenchmarks.time_square_heisenberg('ecr')                              |      
| -        | 1.72±0.01s                             | 1.38±0.02s                              |    0.81 | utility_scale.UtilityScaleBenchmarks.time_qv('cz')                                              |      
| -        | 276±2ms                                | 224±3ms                                 |    0.81 | utility_scale.UtilityScaleBenchmarks.time_square_heisenberg('cz')                               |      
| -        | 1.66±0.02s                             | 1.34±0.01s                              |    0.8  | utility_scale.UtilityScaleBenchmarks.time_qv('ecr')                                             |      
| -        | 21.1±0.2ms                             | 16.8±0.1ms                              |    0.79 | transpiler_levels.TranspilerLevelBenchmarks.time_transpile_from_large_qasm(2)                   |      
| -        | 643±4ms                                | 508±2ms                                 |    0.79 | utility_scale.UtilityScaleBenchmarks.time_qaoa('cz')                                            |      
| -        | 635±5ms                                | 500±9ms                                 |    0.79 | utility_scale.UtilityScaleBenchmarks.time_qaoa('ecr')                                           |
| -        | 24.0±0.7ms                             | 18.2±0.2ms                              |    0.76 | transpiler_levels.TranspilerLevelBenchmarks.time_transpile_from_large_qasm(3)                   |
| -        | 11.3±0.2ms                             | 3.68±0.02ms                             |    0.33 | passes.Collect2QPassBenchmarks.time_consolidate_blocks(5, 1024)                                 |
| -        | 30.5±0.2ms                             | 9.90±0.04ms                             |    0.32 | passes.Collect2QPassBenchmarks.time_consolidate_blocks(14, 1024)                                |

| Change   | Before [f2e07bc5] <main>               | After [a4229901] <consolidate-blocks>   | Ratio   | Benchmark (Parameter)                                                                                  |
|----------|----------------------------------------|-----------------------------------------|---------|--------------------------------------------------------------------------------------------------------|
|          | failed                                 | failed                                  | n/a     | passes.Collect2QPassBenchmarks.time_consolidate_blocks(20, 1024)                                       |
|          | 3.71±0.02s                             | 3.79±0.1s                               | 1.02    | utility_scale.UtilityScaleBenchmarks.time_circSU2('cx')                                                |
|          | 9.18±0.1ms                             | 9.27±0.1ms                              | 1.01    | utility_scale.UtilityScaleBenchmarks.time_bvlike('cx')                                                 |
|          | 9.21±0.05ms                            | 9.24±0.05ms                             | 1.00    | utility_scale.UtilityScaleBenchmarks.time_bvlike('cz')                                                 |
|          | 9.23±0.06ms                            | 9.25±0.07ms                             | 1.00    | utility_scale.UtilityScaleBenchmarks.time_bvlike('ecr')                                                |
|          | 4.00±0.05s                             | 3.92±0.07s                              | 0.98    | utility_scale.UtilityScaleBenchmarks.time_circSU2('cz')                                                |
|          | 35.0±1ms                               | 34.2±0.2ms                              | 0.98    | utility_scale.UtilityScaleBenchmarks.time_parse_square_heisenberg_n100('cz')                           |
|          | 35.0±1ms                               | 34.2±0.3ms                              | 0.98    | utility_scale.UtilityScaleBenchmarks.time_parse_square_heisenberg_n100('ecr')                          |
|          | 3.71±0.1s                              | 3.61±0.2s                               | 0.97    | utility_scale.UtilityScaleBenchmarks.time_circSU2('ecr')                                               |
|          | 166±2ms                                | 155±2ms                                 | 0.93    | utility_scale.UtilityScaleBenchmarks.time_bv_100('cx')                                                 |
|          | 177±1ms                                | 165±0.8ms                               | 0.93    | utility_scale.UtilityScaleBenchmarks.time_bv_100('cz')                                                 |
|          | 1.40±0.01s                             | 1.31±0s                                 | 0.93    | utility_scale.UtilityScaleBenchmarks.time_qft('cx')                                                    |
|          | 179±0.9ms                              | 165±1ms                                 | 0.92    | utility_scale.UtilityScaleBenchmarks.time_bv_100('ecr')                                                |
```
It's marginally faster than the previous run now.

mtreinish:
I ran the pgo scripts under a profiler to see where the pass is spending most of it's time after https://github.com/Qiskit/qiskit/pull/13368/commits/62df015e79c895499168c07eaaf42f4ce90aa03b and the top 4 components taking runtime are: ~42% of the time is in `TwoQubitBasisDecomposer.num_basis_gates()` ~10% is in `DAGCircuit.replace_block_with_op()`, and ~8.3% each for `DAGCircuit.collect_2q_runs()` and `is_supported()`. So I'm not sure there is a ton of extra tuning we can do without changing the behavior of the pass. I think refactoring this into something like: https://github.com/Qiskit/qiskit/commit/bd43c51cb7331b8a50b3ad271e575f005c67efc6 is going to be better path to improve the performance moving forward. The other thing I think we will want to look at is using nalgebra for it's fixed size `Array2` and `Array4` types which are stack allocated and should be faster for all of our use cases than ndarray and faer in this code path.

releasenotes/notes/rust-consolidation-a791a00380fc78b8.yaml
```diff
@@ -0,0 +1,13 @@
+---
+features_transpiler:
+  - |
+    The :class:`.ConsolidateGates` pass will now run the the equivalent of the
```
kevinhartman:
```suggestion
    The :class:`.ConsolidateGates` pass will now run the equivalent of the
```

qiskit/transpiler/passes/optimization/consolidate_blocks.py
```diff
@@ -78,9 +74,15 @@ def __init__(
         if kak_basis_gate is not None:
             self.decomposer = TwoQubitBasisDecomposer(kak_basis_gate)
         elif basis_gates is not None:
-            self.decomposer = unitary_synthesis._decomposer_2q_from_basis_gates(
-                basis_gates, approximation_degree=approximation_degree
-            )
+            kak_gates = set(basis_gates or []).intersection(KAK_GATE_NAMES.keys())
```
kevinhartman: You should be able to skip construction of the intermediate set by using the key view as the receiver. Not that it'd make much difference in this context 🙂.

```suggestion
            kak_gates = KAK_GATE_NAMES.keys() & (basis_gates or [])
```

qiskit/transpiler/passes/optimization/consolidate_blocks.py
```diff
@@ -78,9 +74,15 @@ def __init__(
...
+            kak_gate = None
+            if kak_gates:
+                kak_gate = KAK_GATE_NAMES[kak_gates.pop()]
+                self.decomposer = TwoQubitBasisDecomposer(
+                    kak_gate, basis_fidelity=approximation_degree or 1.0
+                )
+            if kak_gate is None:
+                self.decomposer = TwoQubitBasisDecomposer(CXGate())
```
kevinhartman: Can `kak_gate` actually be `None` if `kak_gates`  is truthy? If not, can we do this instead?

```suggestion
            if kak_gates:
                self.decomposer = TwoQubitBasisDecomposer(
                    KAK_GATE_NAMES[kak_gates.pop()], basis_fidelity=approximation_degree or 1.0
                )
            else:
                self.decomposer = TwoQubitBasisDecomposer(CXGate())
```

qiskit/transpiler/passes/optimization/consolidate_blocks.py
```diff
@@ -195,38 +130,15 @@ def _handle_control_flow_ops(self, dag):
         pass_manager = PassManager()
         if "run_list" in self.property_set:
             pass_manager.append(Collect1qRuns())
-        if "block_list" in self.property_set:
```
kevinhartman, line 198: Can `block_list` be specified when `run_list` is not?

qiskit/transpiler/passes/optimization/consolidate_blocks.py
```diff
@@ -195,38 +130,15 @@ def _handle_control_flow_ops(self, dag):
...
         pass_manager.append(self)
-        for node in dag.op_nodes():
-            if node.name not in CONTROL_FLOW_OP_NAMES:
-                continue
-            dag.substitute_node(
-                node,
-                node.op.replace_blocks(pass_manager.run(block) for block in node.op.blocks),
-                propagate_condition=False,
-            )
+        control_flow_nodes = dag.control_flow_op_nodes()
```
kevinhartman, line 134: I must say, I wish that `control_flow_op_nodes` just returned an empty list rather than `None`, but that is beyond the scope of this PR.

crates/circuit/src/dag_circuit.rs
```diff
@@ -6980,6 +6758,244 @@ impl DAGCircuit {
         };
         Self::from_circuit(py, circ, copy_op, None, None)
     }
+
+    // Replace a block of node indices with a new python operation
```
```suggestion
    /// Replace a block of node indices with a new python operation
```

crates/circuit/src/dag_circuit.rs
```diff
@@ -6980,6 +6758,244 @@ impl DAGCircuit {
...
+            })?;
+
+        self.increment_op(op_name.as_str());
+        for name in block_op_names {
+            self.decrement_op(name.as_str());
+        }
+        Ok(new_node)
+    }
+
+    pub fn substitute_node_with_py_op(
```
kevinhartman, line 6889: Can you add a summary comment for this method as well?

crates/accelerate/src/convert_2q_block_matrix.rs
```diff
@@ -12,102 +12,138 @@
...
 
+use crate::euler_one_qubit_decomposer::matmul_1q;
 use crate::QiskitError;
 
-fn get_matrix_from_inst<'py>(
+#[inline]
+pub fn get_matrix_from_inst<'py>(
     py: Python<'py>,
-    inst: &'py CircuitInstruction,
+    inst: &'py PackedInstruction,
```
kevinhartman: Do we need a lifetime here still?

crates/accelerate/src/convert_2q_block_matrix.rs
```diff
@@ -12,102 +12,138 @@
...
 
 /// Return the matrix Operator resulting from a block of Instructions.
-#[pyfunction]
-#[pyo3(text_signature = "(op_list, /")]
 pub fn blocks_to_matrix(
     py: Python,
-    op_list: Vec<PyRef<DAGOpNode>>,
-    block_index_map_dict: &Bound<PyDict>,
-) -> PyResult<Py<PyArray2<Complex64>>> {
-    // Build a BitData in block_index_map_dict order. block_index_map_dict is a dict of bits to
```
kevinhartman, line 63: 😄

crates/accelerate/src/consolidate_blocks.rs
```diff
@@ -0,0 +1,320 @@
...
+            if !is_supported(
+                target,
+                basis_gates.as_ref(),
+                inst.op.name(),
+                dag.get_qargs(inst.qubits),
+            ) {
+                all_block_gates.insert(inst_node);
+                let matrix = match get_matrix_from_inst(py, inst) {
+                    Ok(mat) => mat,
+                    Err(_) => continue,
```
kevinhartman, line 112: Should an error be logged or handled in some way?

crates/accelerate/src/consolidate_blocks.rs
```diff
@@ -0,0 +1,320 @@
...
+            let mut qargs: Vec<Qubit> = block_qargs.iter().copied().collect();
+            qargs.sort();
+            let block_index_map: HashMap<Qubit, usize> = qargs
+                .into_iter()
+                .enumerate()
+                .map(|(idx, qubit)| (qubit, idx))
+                .collect();
+            let circuit_data = CircuitData::from_packed_operations(
+                py,
+                block_qargs.len() as u32,
```
kevinhartman, line 150: This is perhaps more evidence that we ought to just use `usize` in any public interface of the circuit crate rather than the internal representation. 

(no action requested)

crates/accelerate/src/consolidate_blocks.rs
```diff
@@ -0,0 +1,320 @@
...
+                };
+                let array = matrix.into_pyarray_bound(py);
+                let unitary_gate = UNITARY_GATE
+                    .get_bound(py)
+                    .call1((array, py.None(), false))?;
+                dag.substitute_node_with_py_op(py, first_inst_node, &unitary_gate, false)?;
+                continue;
+            }
+            let qubit = first_qubits[0];
+            let mut matrix = [
```
kevinhartman: Could you use `ONE_QUBIT_IDENTITY` here instead?

kevinhartman (commented):
This looks good to me. Just a few comments / questions.

I'll avoid signing off since I believe @henryzou50 is also planning to look.

crates/accelerate/src/consolidate_blocks.rs
mtreinish: Done in: https://github.com/Qiskit/qiskit/pull/13368/commits/864fc51c44060eb5058017881f85ce72bfaecf0d

crates/accelerate/src/convert_2q_block_matrix.rs
mtreinish: Done in: https://github.com/Qiskit/qiskit/pull/13368/commits/11742b4653a3d0680b565c5cd5bd38b3866f1fba

crates/circuit/src/dag_circuit.rs
mtreinish: Done in: https://github.com/Qiskit/qiskit/pull/13368/commits/f1e645fa2c60ade6086d0f5b8d05770da2b3019b

qiskit/transpiler/passes/optimization/consolidate_blocks.py
mtreinish: It can be, that's arguably the normal invocation of the pass. My thinking here was that we'll implicitly run the equivalent of `Collect2qBlocks` on the control flow block regardless of the property set if we don't specify it. That's the new feature this PR adds to the pass. So we don't need to manually populate the property set with `Collect2qBlocks` unless the `run_list` is set because populating that will preclude the 2q blocks.

crates/accelerate/src/consolidate_blocks.rs
mtreinish: I don't think so, this isn't an error condition if it occurs, we simply can't process it. In practice I don't think this will occur, I'm pretty sure the collection checks filter out nodes that don't have a matrix defined. But if somehow this occurred the correct behavior is to just skip it. This is far too internal to the details of how this pass works to make logging anything worthwhile in this scenario.

crates/accelerate/src/consolidate_blocks.rs
```diff
@@ -0,0 +1,317 @@
...
+            target.instruction_supported(name, Some(&physical_qargs))
+        }
+        None => match basis_gates {
+            Some(basis_gates) => basis_gates.contains(name),
+            None => true,
+        },
+    }
+}
+
+const MAX_2Q_DEPTH: usize = 20;
```
henryzou50, line 53: It would be helpful to add documentation here explaining the purpose of MAX_2Q_DEPTH. Previously, it was clear that "If depth > 20, there will be 1q gates to consolidate," but this context isn’t immediately apparent in the current code. 

On a related note, since the value of 20 for MAX_2Q_DEPTH seems somewhat arbitrary, should we consider making it configurable?

crates/accelerate/src/consolidate_blocks.rs
```diff
@@ -0,0 +1,317 @@
...
+                let unitary_gate = UNITARY_GATE
+                    .get_bound(py)
+                    .call1((array, py.None(), false))?;
+                dag.substitute_node_with_py_op(py, inst_node, &unitary_gate, false)?;
+                continue;
+            }
+        }
+        let mut basis_count: usize = 0;
+        let mut outside_basis = false;
+        let mut block_qargs: HashSet<Qubit> = HashSet::with_capacity(2);
```
henryzou50: Instead of creating a new `HashSet` in each iteration, we could instead initialize it once outside the loop and then clear it at the beginning of each iteration. This should help optimize performance and memory, especially in cases with many iterations.

For instance we can have something like:

``` rust
let mut block_qargs: HashSet<Qubit> = HashSet::with_capacity(2);  
for block in blocks {
    block_qargs.clear();  
    // Rest of the code
}
```

crates/accelerate/src/consolidate_blocks.rs
```diff
@@ -0,0 +1,317 @@
...
+
+    let runs: Option<Vec<Vec<NodeIndex>>> = runs.map(|runs| {
+        runs.into_iter()
+            .map(|run| {
+                run.into_iter()
+                    .map(NodeIndex::new)
+                    .collect::<Vec<NodeIndex>>()
+            })
+            .collect()
+    });
```
henryzou50, line 94: To improve the memory efficiency here, we could simplify the handling of the `blocks` and `runs` by unifying them with `or_else` and `unwarp_or_else`. This approach would reduces the unnecessary temporary `Vec` allocations and leveragers iterators for more efficient memory usage. 

For instance, something like:
```suggestion
    // If `blocks` is `None` but `runs` is specified, `blocks` is set to `runs.clone()`.
    // If both are `None`, it defaults to collecting 2q runs from the DAG circuit.
    let blocks: Vec<Vec<NodeIndex>> = blocks
        .or_else(|| runs.clone())
        .unwrap_or_else(|| {
            dag.collect_2q_runs()
                .unwrap()
                .into_iter()
                .map(|run| {
                    run.into_iter()
                        .map(|node_index| node_index.index())
                        .collect()
                })
                .collect()
        })
        .into_iter()
        .map(|run| run.into_iter().map(NodeIndex::new).collect())
        .collect();

    let runs: Option<Vec<Vec<NodeIndex>>> = runs.map(|runs| {
        runs.into_iter()
            .map(|run| run.into_iter().map(NodeIndex::new).collect())
            .collect()
    });
```

crates/accelerate/src/consolidate_blocks.rs
henryzou50: I agree with Matt. Since the collection process filters out nodes without matrices, this scenario is unlikely, and skipping aligns with the intended behavior. Adding logging or handling would add unnecessary complexity for an internal detail.

henryzou50 (commented):
Overall, this looks great to me! Thanks to Kevin for also reviewing. I’ve left a few suggestions for improving efficiency and memory usage, along with some additional comments and questions. Excellent work, and thanks for all the work put into this!

crates/accelerate/src/consolidate_blocks.rs
mtreinish: Good point, I'll add the comment back. 

I'm not 100% sure, but I think the value derives from the worst case synthesis from the default two qubit decomposer. There can be at most 5 gates for the 4x 1q components in the decomposition output. It's more evidence that https://github.com/Qiskit/qiskit/issues/11659 is becoming increasingly relevant.

crates/accelerate/src/consolidate_blocks.rs
mtreinish: I'll see if I can come up with something for this. The behavior of cloning runs like this doesn't feel correct and I think probably wouldn't work correctly because it'll try to call `blocks_to_matrix` with a single qubit matrix. But I'll see if I can come up with something to avoid temporary vecs. I think I did this to make the borrow checker happy, but it was long enough ago I'm not 100% sure.

crates/accelerate/src/consolidate_blocks.rs
mtreinish: Done in: https://github.com/Qiskit/qiskit/pull/13368/commits/223bf2e4efd25011fa80320bdd26b80480fb19cd

crates/accelerate/src/consolidate_blocks.rs
mtreinish: Done in: https://github.com/Qiskit/qiskit/pull/13368/commits/b6071eecdd286d2b734d92be21aee79592b47017

crates/accelerate/src/consolidate_blocks.rs
mtreinish: I'm not sure I see a path here, the conflict is on the type mismatch between `NodeIndex` and the `usize` input. From a typing perspective rust won't be happy because `collect_2q_runs()` returns a `Vec` of `NodeIndex` and we can't take that as an input.  The best idea I had to avoid an extra allocation was to take `NodeIndex` on the input Vec but we can't define the `FromPyObject` trait for a type not defined in qiskit.

Your suggestion here doesn't actually avoid any allocations except if blocks is `Some(_)`, but at the cost of a second allocation for `blocks` is `None` path. The `blocks` is `None` path is the more common path though because that's what we run in the preset pass managers, so I'd rather stick with the bare `collect_2q_runs` call and have the conversion cost from the `blocks` list if it's specified.

henryzou50: Oh yes, I understand now, and that makes sense. You're correct about the typing conflict and the allocation cost trade-offs, especially since the blocks is `None` path is more common in the preset pass managers. Let’s stick with what you initially suggested with the bare `collect_2q_runs` call, and handle the conversion cost from the blocks list if it's specified. Thanks for clarifying!

henryzou50 (approved):
Looks good to me, Matt! Thanks for the latest changes!

PR #13369, opened 2024-10-24
Oxidize parts of HLS

alexanderivrii:
### Summary

This PR oxidize small parts of `HighLevelSynthesis`: `QubitTracker` and `QubitContext`. It is implemented on top of #13240 and replaces #13180.

### Details and comments

This is a more or less mechanical porting of the python code in #13240. Ideally, I would also like to port the auxiliary function `_definitely_skip_node` in time for 1.3.0.


crates/accelerate/src/high_level_synthesis.rs
```diff
@@ -0,0 +1,252 @@
...
+            ignored: vec![false; num_qubits],
+        }
+    }
+
+    /// Sets state of the given qubits to dirty
+    fn set_dirty(&mut self, qubits: Vec<usize>) {
+        for q in qubits {
+            self.state[q] = false;
+        }
+    }
```
raynelfss, line 50: I was initially thinking of changing the types here to `u32` instead of `usize`, however since we mostly use the types here for indexing, I believe `usize` should suffice. We'd have to convert to `usize` anyway if we wanted to index. No action required here :)

crates/accelerate/src/high_level_synthesis.rs
```diff
@@ -0,0 +1,252 @@
...
+#[derive(Clone, Debug)]
+pub struct QubitTracker {
+    /// The total number of global qubits
+    num_qubits: usize,
+    /// Stores the state for each qubit: `true` means clean, `false` means dirty
+    state: Vec<bool>,
+    /// Stores whether qubits are allowed be used
+    enabled: Vec<bool>,
+    /// Used internally for keeping the computations in `O(n)`
+    ignored: Vec<bool>,
```
raynelfss, line 30: Since this isn't used as much here, maybe we could remove it in favor of providing a `HashSet<usize>` whenever needed? Performing a hash lookup for a usize shouldn't be very expensive, and since we're only using this as a way of checking a value should be ignored, a `HashSet<usize>` as a function argument should suffice. We'd have to benchmark it to see if it affects performance in any way.

crates/accelerate/src/high_level_synthesis.rs
```diff
@@ -0,0 +1,252 @@
...
+        let count = (0..self.num_qubits)
+            .filter(|q| !self.ignored[*q] && self.enabled[*q] && self.state[*q])
+            .count();
+
+        for q in &ignored_qubits {
+            self.ignored[*q] = false;
+        }
+
+        count
+    }
```
raynelfss, line 89: Just so you're able to test quickly, see if using a hashset affects this in any way:
```suggestion
    /// Returns the number of enabled clean qubits, ignoring the given qubits
    /// ToDo: check if it's faster to avoid using ignored
    fn num_clean(&mut self, ignored_qubits: HashSet<usize>) -> usize {
        (0..self.num_qubits)
            .filter(|q| !ignored_qubits.contains(q) && self.enabled[*q] && self.state[*q])
            .count()
    }
```

crates/accelerate/src/high_level_synthesis.rs
```diff
@@ -0,0 +1,252 @@
...
+        let count = (0..self.num_qubits)
+            .filter(|q| !self.ignored[*q] && self.enabled[*q] && !self.state[*q])
+            .count();
+
+        for q in &ignored_qubits {
+            self.ignored[*q] = false;
+        }
+
+        count
+    }
```
raynelfss, line 107: Same case here (for testing):
```suggestion
    /// Returns the number of enabled dirty qubits, ignoring the given qubits
    /// ToDo: check if it's faster to avoid using ignored
    fn num_dirty(&mut self, ignored_qubits: HashSet<usize>) -> usize {
        (0..self.num_qubits)
            .filter(|q| !ignored_qubits.contains(q) && self.enabled[*q] && !self.state[*q])
            .count()
```

crates/accelerate/src/high_level_synthesis.rs
```diff
@@ -0,0 +1,252 @@
...
+        let out: Vec<usize> = clean_ancillas
+            .chain(dirty_ancillas)
+            .take(num_qubits)
+            .collect();
+
+        for q in &ignored_qubits {
+            self.ignored[*q] = false;
+        }
+        out
+    }
```
raynelfss, line 130: Here's how it would look with the Hashset:
```suggestion
    /// Get `num_qubits` enabled qubits, excluding `ignored_qubits`, and returning the
    /// clean qubits first.
    /// ToDo: check if it's faster to avoid using ignored
    fn borrow(&mut self, num_qubits: usize, ignored_qubits: HashSet<usize>) -> Vec<usize> {
        let clean_ancillas = (0..self.num_qubits)
            .filter(|q| !ignored_qubits.contains(q) && self.enabled[*q] && self.state[*q]);
        let dirty_ancillas = (0..self.num_qubits)
            .filter(|q| !self.ignored[*q] && self.enabled[*q] && !self.state[*q]);
        clean_ancillas
            .chain(dirty_ancillas)
            .take(num_qubits)
            .collect()
    }
```

raynelfss (commented):
I believe this is about ready to merge (I'll let @Cryoris have the last word). I just left some suggestions about `ignored_qubits` only to test whether a `HashSet` would be the better approach here. I left the transformed functions as suggestions here.

raynelfss (approved):
After doing some testing with sets, the performance is almost the same.  If anything I could suggest using the `HashSet` route once we have the `HighLevelSynthesis` entirely in Rust just to avoid doing the extra conversions (from `list` to `set` in Python and from `PySet` to `HashSet` in Rust).
This code looks great and is ready to merge as is, thank you for your contribution 🚀

alexanderivrii:
Thanks @raynelfss! I have just run a few experiments (after you have already merged the code) to see what happens on circuits with many small gates:
```
qc = QuantumCircuit(100)
qc.mcx([0, 1, 2, 3], 4)  # make sure that HLS does not follow fast-track
for _ in range(100):
    for i in range(49):
        qc.cx(i, i+1)
qct = HighLevelSynthesis()(qc)
```
and on circuits with many large gates that do not need to be synthesized:
```
pattern = list(range(99, -1, -1))
qc = QuantumCircuit(100)
for _ in range(1000):
    qc.append(PermutationGate(pattern), qc.qubits)
config = HLSConfig(permutation=[])  # this means: do not actually synthesize permutation gates
qct = HighLevelSynthesis(hls_config=config)(qc)
```
In fact, the performance even becomes a tiny bit better when using `hashbrown::HashSet` (despite, as you have mentioned, having to convert from `list` to `set` in Python) -- and I am actually a bit surprised by this (using the "scratch" bitvector was a very common and performance-critical trick in the C++ projects I worked on in the past).

So, indeed, I am 100% switching to the `HashSet` route in the future.

PR #13370, opened 2024-10-24
Add iterator-access methods to `SparseObservable`

jakelishman:
### Summary

This adds a new related class, `SparseObservable.Term` to be the item type of the iteration methods, and imbibes it with the simplest mathematical binary relations with `SparseObservable` as the other operand.

For some reason, in my head, the methods to convert terms and observables to their corresponding Pauli basis are related to iteration?  The `SparseObservable.pauli_bases` and `SparseObservable.Term.pauli_base` methods perform a similar bitwise trick to do this reduction as exemplified in the documentation, but return the other `quantum_info`-native classes `PauliList` and `Pauli`, respectively.  These methods are expected to be used by the primitives.

### Details and comments

Built on top of #13298.

There's still one more part of `SparseObservable` to go - `apply_layout`, `compose` and `evolve` - then potentially a follow-on PR that reorganises the class into two really separate components; a pure Rust core for Python-free consumption, and a `pyclass` wrapper around it.

### DIFF
--- a/crates/accelerate/src/sparse_observable.rs
+++ b/crates/accelerate/src/sparse_observable.rs
@@ -974,7 +974,7 @@ impl SparseObservable {
         out
     }
 
-    /// Get an owned representation of a single sparse term.
+    /// Get a view onto a representation of a single sparse term.
     ///
     /// This is effectively an indexing operation into the [SparseObservable].  Recall that two
     /// [SparseObservable]s that have different term orders can still represent the same object.
@@ -983,15 +983,15 @@ impl SparseObservable {
     /// # Panics
     ///
     /// If the index is out of bounds.
-    pub fn term(&self, index: usize) -> SparseTerm {
+    pub fn term(&self, index: usize) -> SparseTermView {
         debug_assert!(index < self.num_terms(), "index {index} out of bounds");
         let start = self.boundaries[index];
         let end = self.boundaries[index + 1];
-        SparseTerm {
+        SparseTermView {
             num_qubits: self.num_qubits,
             coeff: self.coeffs[index],
-            bit_terms: (&self.bit_terms[start..end]).into(),
-            indices: (&self.indices[start..end]).into(),
+            bit_terms: &self.bit_terms[start..end],
+            indices: &self.indices[start..end],
         }
     }
 
@@ -1314,15 +1314,15 @@ impl SparseObservable {
     }
 
     fn __getitem__(&self, py: Python, index: PySequenceIndex) -> PyResult<Py<PyAny>> {
-        match index.with_len(self.num_terms())? {
-            SequenceIndex::Int(index) => Ok(self.term(index).into_py(py)),
-            indices => Ok(PyList::new_bound(
-                py,
-                indices.iter().map(|index| self.term(index).into_py(py)),
-            )
-            .into_any()
-            .unbind()),
+        let indices = match index.with_len(self.num_terms())? {
+            SequenceIndex::Int(index) => return Ok(self.term(index).to_term().into_py(py)),
+            indices => indices,
+        };
+        let mut out = SparseObservable::zero(self.num_qubits);
+        for index in indices.iter() {
+            out.add_term(self.term(index))?;
         }
+        Ok(out.into_py(py))
     }
 
     fn __repr__(&self) -> String {
@@ -2521,6 +2521,16 @@ pub struct SparseTermView<'a> {
     pub indices: &'a [u32],
 }
 impl<'a> SparseTermView<'a> {
+    /// Convert this `SparseTermView` into an owning [SparseTerm] of the same data.
+    pub fn to_term(&self) -> SparseTerm {
+        SparseTerm {
+            num_qubits: self.num_qubits,
+            coeff: self.coeff,
+            bit_terms: self.bit_terms.into(),
+            indices: self.indices.into(),
+        }
+    }
+
     fn to_sparse_str(self) -> String {
         let coeff = format!("{}", self.coeff).replace('i', "j");
         let paulis = self
--- a/crates/accelerate/src/sparse_observable.rs
+++ b/crates/accelerate/src/sparse_observable.rs
@@ -162,13 +162,17 @@ impl BitTerm {
         }
     }
 
-    /// Is this term an operator or a projector to the X basis?
-    pub fn is_x_basis(&self) -> bool {
+    /// Does this term include an X component in its ZX representation?
+    ///
+    /// This is true for the operators and eigenspace projectors associated with X and Y.
+    pub fn has_x_component(&self) -> bool {
         ((*self as u8) & (Self::X as u8)) != 0
     }
 
-    /// Is this term an operator or a projector to the Z basis?
-    pub fn is_z_basis(&self) -> bool {
+    /// Does this term include a Z component in its ZX representation?
+    ///
+    /// This is true for the operators and eigenspace projectors associated with Y and Z.
+    pub fn has_z_component(&self) -> bool {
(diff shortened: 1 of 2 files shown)


jakelishman:
Rebased and no longer on hold.

jakelishman:
Rebased again, since it had a minor conflict with #13372.  This also uses the new `PyReadwriteArray::make_nonwriteable` method from `rust-numpy==0.22.1`, since we're using that now.

crates/accelerate/src/sparse_observable.rs
```diff
@@ -159,6 +161,16 @@ impl BitTerm {
             _ => Err(BitTermFromU8Error(value)),
         }
     }
+
+    /// Is this term an operator or a projector to the X basis?
```
Cryoris: These sound a bit misleading to me, since this would also return true for a `Y` basis, no? Maybe we could find a different name, something along the lines of `has_x_component`?

crates/accelerate/src/sparse_observable.rs
```diff
@@ -697,7 +757,7 @@ impl From<LabelError> for PyErr {
 /// observable generate only a small number of duplications, and like-term detection has additional
 /// costs.  If this does not fit your use cases, you can either periodically call :meth:`simplify`,
 /// or discuss further APIs with us for better building of observables.
-#[pyclass(module = "qiskit.quantum_info")]
+#[pyclass(module = "qiskit.quantum_info", sequence)]
```
Cryoris, line 764: Are `len` and `__getitem__` the only two methods we need to add to qualify this as a `Sequence`? 🙂

crates/accelerate/src/sparse_observable.rs
```diff
@@ -905,6 +966,19 @@ impl SparseObservable {
         out
     }
 
+    /// Get an owned representation of a single sparse term.
```
Cryoris: Would it be good to clarify here and in `iter` that the order inside the `SparseObservable` is not unique, unless one calls `canonicalize`?

crates/accelerate/src/sparse_observable.rs
```diff
@@ -1028,7 +1117,7 @@ impl SparseObservable {
             }
             return Self::py_from_label(&label).map_err(PyErr::from);
         }
-        if let Ok(observable) = data.downcast::<Self>() {
+        if let Ok(observable) = data.downcast_exact::<Self>() {
```
Cryoris, line 1136: Do we now need this because we don't want to allow casting to `Sequence` but want to exactly cast to `SparseObservable`?

crates/accelerate/src/sparse_observable.rs
```diff
@@ -1190,6 +1293,22 @@ impl SparseObservable {
...
     }
 
+    fn __len__(&self) -> usize {
+        self.num_terms()
+    }
+
+    fn __getitem__(&self, py: Python, index: PySequenceIndex) -> PyResult<Py<PyAny>> {
+        match index.with_len(self.num_terms())? {
+            SequenceIndex::Int(index) => Ok(self.term(index).into_py(py)),
+            indices => Ok(PyList::new_bound(
```
Cryoris: I sort of would've expected this to return a `SparseObservable` with terms in the expected slice -- the same way that slice-indexing a `tuple` returns a `tuple`. Is this something we can do, or did you choose a `list` on purpose? 🙂

test/python/quantum_info/test_sparse_observable.py
```diff
@@ -155,6 +155,17 @@ def test_default_constructor_copy(self):
...
+    def test_default_constructor_term(self):
+        expected = SparseObservable.from_list([("IIZXII+-", 2j)])
+        self.assertEqual(SparseObservable(expected[0]), expected)
+
+    def test_default_constructor_term_iterable(self):
+        expected = SparseObservable.from_list([("IIZXII+-", 2j), ("rlIIIIII", 0.5)])
+        terms = [expected[0], expected[1]]
+        self.assertEqual(SparseObservable(list(terms)), expected)
+        self.assertEqual(SparseObservable(tuple(terms)), expected)
+        self.assertEqual(SparseObservable(term for term in terms), expected)
```
Cryoris, line 167: Is this the same as this?
```suggestion
        self.assertEqual(SparseObservable(iter(terms)), expected)
```

Cryoris (commented):
Overall LGTM, just some questions below 🙂

crates/accelerate/src/sparse_observable.rs
jakelishman: Python actually has two concepts of "sequence" - one at the low level (which is this one) and `collections.abc.Sequence`.  The requirements of the C API sequence is just to have a `__getitem__` that acts on sequential integers and a correct `__len__`.  Here's the glossary term in their docs with a brief explanation: https://docs.python.org/3/glossary.html#term-sequence.

The C-API-level "sequence" (as here) implies that a few functions are filled in on the corresponding Python type object as C-level function pointers, rather than needing Python-space look-ups to do.  That can make it quite a bit more efficient to interact with the object from the C level.  A bit more detail here: https://docs.python.org/3/c-api/typeobj.html#c.PySequenceMethods.

crates/accelerate/src/sparse_observable.rs
jakelishman: Oh, that's a good point, and it's consistent with how `SparsePauliOp` works, I think.  I can't remember clearly if there's a reason not to do that, but if not, I'll swap it over to do that.

test/python/quantum_info/test_sparse_observable.py
jakelishman: Yeah, it'll end up being effectively the same test.  Do you find the intent of your suggestion clearer?  I'll change it if so (or you just feel free to commit the suggestion).

crates/accelerate/src/sparse_observable.rs
jakelishman: This change wasn't strictly required, and the new form was valid before, I must have just done it as I was going through.  `SparseObservable` can't be subclassed, so there's no real reason not to use `downcast_exact`, which is a bit faster.

crates/accelerate/src/sparse_observable.rs
jakelishman: Done in 79e78d87dd.

crates/accelerate/src/sparse_observable.rs
jakelishman: Done in 806f7cd946.

crates/accelerate/src/sparse_observable.rs
jakelishman: Done in 806f7cd946.

crates/accelerate/src/sparse_observable.rs
Cryoris: Ok I see! And a dumb question: in the glossary it requires `__len__` but here we're adding `len`, is that an issue?

test/python/quantum_info/test_sparse_observable.py
Cryoris: No that's perfectly fine, just to clarify 🙂 (but using `iter` would look aesthetically more pleasing 😛)

Cryoris (approved):
LGTM, feel free to put into the merge queue pending on the `len` vs `__len__` question above 🙂

crates/accelerate/src/sparse_observable.rs
jakelishman: I see us as adding `__len__`?

Cryoris: Lol me too, I don't know what happened, maybe my brain parsed the double underscore into __len__ 😂

PR #13372, opened 2024-10-25
Implement `SparseObservable.apply_layout`

jakelishman:
### Summary

This is one more usability method to bring `SparseObservable` closer inline with `SparsePauliOp`. The same functionality is relatively easily implementable by the user by iterating through the terms, mapping the indices, and putting the output back into `SparseObservable.from_sparse_list`, but given how heavily we promote the method for `SparsePauliOp`, it probably forms part of the core of user expectations.

### Details and comments

Built on top of #13298 (though mostly just because it touches similar changes to the error structs).

I only finalised this after the 1.3 feature-freeze period.  There's probably no _huge_ necessity to rush it in since it's not going to be used by the primitives before Qiskit 2.0 anyway, though it's nice to get these things early.

There's still `compose` and `evolve` to go for `SparseObservable`, but those are definitely for a later release, because I haven't written them yet, and the tests will take a good amount of time to write still.

### DIFF
--- a/crates/accelerate/src/sparse_observable.rs
+++ b/crates/accelerate/src/sparse_observable.rs
@@ -12,6 +12,7 @@
 
 use std::collections::btree_map;
 
+use hashbrown::HashSet;
 use num_complex::Complex64;
 use num_traits::Zero;
 use thiserror::Error;
@@ -263,8 +264,11 @@ impl ::std::convert::TryFrom<u8> for BitTerm {
     }
 }
 
-/// Error cases stemming from data coherence at the point of entry into `SparseObservable` from raw
-/// arrays.
+/// Error cases stemming from data coherence at the point of entry into `SparseObservable` from
+/// user-provided arrays.
+///
+/// These most typically appear during [from_raw_parts], but can also be introduced by various
+/// remapping arithmetic functions.
 ///
 /// These are generally associated with the Python-space `ValueError` because all of the
 /// `TypeError`-related ones are statically forbidden (within Rust) by the language, and conversion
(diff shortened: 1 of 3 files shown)


crates/accelerate/src/sparse_observable.rs
```diff
@@ -1454,11 +1828,482 @@ impl SparseObservable {
...
+    /// the fully transpiled *physical* circuit.
+    ///
+    /// Args:
+    ///     layout (TranspileLayout | list[int] | None): The layout to apply.  Most uses of this
+    ///         function should pass the :attr:`.QuantumCircuit.layout` field from a circuit that
+    ///         was transpiled for hardware.  In addition, you can pass a list of new qubit indices.
+    ///         If given as explicitly ``None``, no remapping is applied (but you can still use
+    ///         ``num_qubits`` to expand the observable).
+    ///     num_qubits (int | None): The number of qubits to expand the observable to.  If not
+    ///         supplied, the output will be as wide as the given :class:`TranspileLayout`, or the
```
Cryoris: Sorry if I'm asking this 100 times in this process -- but now `TranspileLayout` is not available in the same file, hence we should add a leading `.`? 😅 
```suggestion
    ///         supplied, the output will be as wide as the given :class:`.TranspileLayout`, or the
```

jakelishman: It should have had it, thanks - fixed during the rebase.

jakelishman:
Rebased and no longer on hold.

Cryoris (approved):
LGTM 👍🏻 

It's true we don't strictly require this for 1.3 but this looks solid, brief and ready, so I'd say there's no reason to put it off 🙂

PR #13374, opened 2024-10-28
Move all release notes from 1.2 into `releasenotes/notes/1.2/`

raynelfss:
### Summary

The following commits move all of the release notes from 1.2 into the `releasenotes/notes/1.2/ ` directory.

### Details and comments

After the 1.2 release, the 1.2 release notes were not separated from all the newer releasenotes. These commits aim to fix that and add all of the missing release notes into a folder dedicated to it.


raynelfss:
It seems there are more release notes that I didn't include. will continue adding more in follow-ups

raynelfss:
I was able to add all of the missing release notes from 1.2, but for some reason `reno` would not recognize them when generating the release note pages. All of them still show up under "Unreleased Notes".

raynelfss:
> but for some reason reno would not recognize them when generating the release note pages. All of them still show up under "Unreleased Notes".

Apparently due to the nature of `1.2` we are not able to show it locally on the docs. This should be good to go now :)

ElePT (approved):
LGTM! Thanks for the effort of going through the renos.

PR #13375, opened 2024-10-28
Fix performance regression in UnitarySynthesis

mtreinish:
### Summary

This commit fixes a performance regression that was introduced in PR #13141. When the pass is looking up the preferred synthesis direction for a unitary based on the connectvity constraints the connectivity was being provided as a PyList. To look up the edge in connectivity set this meant we needed to iterate over the list and then create a set that rust could lookup if it contains an edge or it's reverse. This has significant overhead because its iterating via python and also iterating per decomposition. This commit addresses this by changing the input type to be a HashSet from Python so Pyo3 will convert a pyset directly to a HashSet once at call time and that's used by reference for lookups directly instead of needing to iterate over the list each time.

### Details and comments

### DIFF
--- a/crates/accelerate/src/unitary_synthesis.rs
+++ b/crates/accelerate/src/unitary_synthesis.rs
@@ -27,7 +27,7 @@ use smallvec::{smallvec, SmallVec};
 
 use pyo3::intern;
 use pyo3::prelude::*;
-use pyo3::types::{IntoPyDict, PyDict, PyList, PyString};
+use pyo3::types::{IntoPyDict, PyDict, PyString};
 use pyo3::wrap_pyfunction;
 use pyo3::Python;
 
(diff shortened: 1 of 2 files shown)


crates/accelerate/src/unitary_synthesis.rs
```diff
@@ -268,7 +268,7 @@ fn py_run_main_loop(
                     new_ids,
                     min_qubits,
                     target,
-                    coupling_edges,
+                    coupling_edges.clone(),
```
mtreinish, line 271: We can avoid this clone if we refactor the functions a bit to split the python interface out and use a reference for the recursion. Not sure how important this is because this only comes up during control flow processing and we're already slower there because of the python usage and circuit to dag conversion. Maybe something to fix for when we make a dag native control flow op.

ElePT (approved):
LGTM. Thanks for the catch :)

crates/accelerate/src/unitary_synthesis.rs
ElePT: I think it would make sense to revisit this when we get to have rust-native control flow op.

PR #13377, opened 2024-10-28
Refactor encapsulation of Bit storage

kevinhartman:
### Summary
This PR follows up #13284 (which added an interface for converting bit types into and out of `usize`) with encapsulation of the internal representation used for bits. Previously, `Qubit` and `Clbit` were transparent wrappers around `u32`. Their internal representation is now private to encourage users to deal with bits directly or as `usize` via their `index`. As part of this, I've changed a lot of API surface in `accelerate` to deal in terms of `usize` for bit indices as well.

### Details and comments
* The `BitData` struct has also been made private to the circuit crate. It was never intended as API surface for other crates, but rather as a short term way for us to track the association between native and Python bits. To support this, a new crate private type `WireIndex` has been added which is implemented by `Qubit`, `Clbit` and `Var`. It serves as a bound for the type `T` used in a `BitData<T>` and provides crate-private conversion between our wire types and `usize`.  With these changes, it is (intentionally) also no longer possible for `Var` to be constructed outside of the circuit crate.
* A `CircuitData::from_dag` function has also been added to support this encapsulation.
* The `From<Qubit>` trait is also now implemented for `PhysicalQubit` and `VirtualQubit` (though these really also ought to move to the circuit crate soon...).
* Types like `PyInstruction` and `PyGate` still publicly store their `num_qubits` and `num_clbits` as `u32`. We should probably consider making these fields `crate(pub)` as well.

~Needs a rebase on  #13368.~


kevinhartman:
Closing this out since we didn't end up going in this direction 🙂.

PR #13378, opened 2024-10-29
Avoid unnecessary matrix construction in `Split2qUnitaries`

jakelishman:
### Summary

During the transition of this pass to Rust, the matrix creation was accidentally moved before the check of whether the pass would operate on the node.  This wastes time and allocations on creating matrices that will not be further examined.


Cryoris (approved):
Nice catch!

PR #13379, opened 2024-10-29
Improve speed of `Pauli.to_label`

### LINKED ISSUE #13355: Faster hashing for Pauli operators.
### What should we add?

Currently the hashing of a Pauli operator is computed as `hash(self.to_label())`. It is basically asking for the label of the Pauli operator which is an expensive computation to run. Instead building it directly from `self.x`, `self.z`,`self.phase` seems more reasonable.
I did a quick test where I do the hashing with `int.from_bytes(np.packbits([self.x, self.z]+[self.phase]))` and it was about 50 times faster.

jakelishman:
### Summary

This uses Numpy vectorised operations and builtin Python tooling to remove Python-space loops from the Pauli label creation.  This is a speed up of around 40x on my 2020 Intel MacBook Pro for most Paulis.

### Details and comments

Fix #13355


Cryoris (approved):
Thanks for this improvement 🚀

PR #13381, opened 2024-10-30
Bump the version of dependency oq3_semantics

### LINKED ISSUE #13096: Build warning with Rust 1.81.0
### What should we add?

When building Qiskit with the latest Rust release, 1.81.0, a future incompatibility warning is raised at the end of the build:

```
warning: the following packages contain code that will be rejected by a future version of Rust: ra_ap_stdx v0.0.188
```

Looking at the cargo report this is caused by code in the crate that might lead to undefined behavior in a future release of Rust. 

This isn't a direct issue in Qiskit, as the `ra_ap_stdx` dependency gets pulled in by: https://github.com/Qiskit/openqasm3_parser but I opened so we can track it on the Qiskit side as people building qiskit will encounter this. It also will let us track upgrading the openqasm3_parser version once this is addressed over there.

The upstream tracking issue is: https://github.com/Qiskit/openqasm3_parser/issues/229

jlapeyre:
The previous required version was 0.6.0. This commit bumps the required version to 0.7.0.
There are several changes since 0.6.0, but the impetus for bumping the version here is for compatibility with Rust 1.81.

Closes #13096

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -42,15 +42,6 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "5c6cb57a04249c6480766f7f7cef5467412af1490f8d1e243141daddada3264f"
 
 [[package]]
-name = "always-assert"
-version = "0.1.3"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "4436e0292ab1bb631b42973c61205e704475fe8126af845c8d923c0996328127"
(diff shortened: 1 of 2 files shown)


jakelishman (commented):
I think you might need to do a `cargo build` locally and commit the subsequent changes to `Cargo.lock`?

mtreinish (approved):

PR #13384, opened 2024-10-30
Avoid iterating over nodes in Split2QUnitaries if there are no unitaries

mtreinish:
### Summary

This commit adds a check to the start of the split_2q_unitaries rust function that is the core of the Split2QUnitaries transpiler pass to do a lookup if there are any `UnitaryGate` objects in the dag without iterating over the dag. The pass only will split unitary gates so we can avoid checking the name of each gate individually if we know there aren't any up front.

### DIFF
--- a/crates/accelerate/src/split_2q_unitaries.rs
+++ b/crates/accelerate/src/split_2q_unitaries.rs
@@ -28,6 +28,9 @@ pub fn split_2q_unitaries(
     dag: &mut DAGCircuit,
     requested_fidelity: f64,
 ) -> PyResult<()> {
+    if !dag.get_op_counts().contains_key("unitary") {
+        return Ok(());
+    }
     let nodes: Vec<NodeIndex> = dag.op_nodes(false).collect();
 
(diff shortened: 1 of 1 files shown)


Cryoris (commented):
Getting the op counts also iterates over the DAG though right? Would we get a similar or even better performance if we have the name-check before we get the qubits? I.e. 
```rust
for node in nodes {
   if let NodeType::Operation(inst) = &dag.dag()[node] { 
       if inst.op.name() != "unitary" { continue };
       // rest as before ...
```
This way we would keep only one loop over the DAG, no?

mtreinish:
No, it's an `O(1)` lookup that doesn't require any iteration. The `DAGCircuit` maintains the counts mapping as operations are added and removed specifically for this use case. You can just look at the method definition: https://github.com/Qiskit/qiskit/blob/main/crates/circuit/src/dag_circuit.rs#L6634-L6642

Cryoris (approved):

PR #13385, opened 2024-10-31
Fix type of `basis_gates` in `BasicSimulator` configuration

wshanks:
`basis_gates` was getting set to `Target.operation_names`, which is a `dict_keys` instance (maybe that should be changed to a more basic type?), while typically `basis_gates` is a list. `dict_keys` instances can not be deep copied, so code that usually deep copies a backend's configuration could not work. Deep copying the configuration can happen for example when using the `BasicSimulator` with sampler class like `qiskit_ibm_runtime.SamplerV2` that puts the backend into its options and then tries to deep copy the options.

### DIFF
--- a/qiskit/providers/basic_provider/basic_simulator.py
+++ b/qiskit/providers/basic_provider/basic_simulator.py
@@ -257,7 +257,7 @@ class BasicSimulator(BackendV2):
                 backend_name=self.name,
                 backend_version=self.backend_version,
                 n_qubits=self.num_qubits,
-                basis_gates=self.target.operation_names,
+                basis_gates=list(self.target.operation_names),
                 gates=gates,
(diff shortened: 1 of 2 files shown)


wshanks:
The configuration model is deprecated any way, but I think it is worth treating this as a bugfix since it blocks a usage pattern that is not trying to use the configuration. It came up in Qiskit Slack [here](https://qiskit.slack.com/archives/CGZDF48EN/p1730380335449369).

jakelishman (approved):
Thanks Will, this seems straightforward and eases a pain point in the transition.  I'll backport it to the 1.2 series, though at present we hadn't been planning to push out a further release on that branch.  If it's a big annoyance, it's easy for us to put out a new patch release.

PR #13388, opened 2024-11-01
Fix two latex issues in docs

Eric-Arellano:
These cause the docs app to crash.

### DIFF
--- a/qiskit/transpiler/passes/optimization/remove_identity_equiv.py
+++ b/qiskit/transpiler/passes/optimization/remove_identity_equiv.py
@@ -33,7 +33,7 @@ class RemoveIdentityEquivalent(TransformationPass):
 
     .. math::
 
-        \bar{F} = \frac{1 + F_{\text{process}}{1 + d}
+        \bar{F} = \frac{1 + F_{\text{process}}}{1 + d},
 
         F_{\text{process}} = \frac{|\mathrm{Tr}(G)|^2}{d^2}
 
--- a/qiskit/transpiler/passes/optimization/remove_identity_equiv.py
+++ b/qiskit/transpiler/passes/optimization/remove_identity_equiv.py
@@ -33,7 +33,7 @@ class RemoveIdentityEquivalent(TransformationPass):
 
     .. math::
 
-        \bar{F} = \frac{1 + F_{\text{process}}}{1 + d},
+        \bar{F} = \frac{1 + F_{\text{process}}}{1 + d},\ 
 
         F_{\text{process}} = \frac{|\mathrm{Tr}(G)|^2}{d^2}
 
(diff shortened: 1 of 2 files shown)


qiskit/transpiler/passes/optimization/remove_identity_equiv.py
```diff
@@ -33,7 +33,7 @@ class RemoveIdentityEquivalent(TransformationPass):
 
     .. math::
 
-        \bar{F} = \frac{1 + F_{\text{process}}{1 + d}
+        \bar{F} = \frac{1 + F_{\text{process}}}{1 + d},
```
Eric-Arellano: The comma is only to make the two equations render more nicely. 
Screenshot 2024-11-01 at 3 22 29 PM

jakelishman: You might want to put an enforced space after the comma to improve it - probably one of `\ ` or `\quad` depending on how much you want.  I'm fine to just merge if this is already good enough for you.

jakelishman (dismissed):
Thanks for this.  Any chance you know a simple way to get Sphinx to attempt the renders server side, so we can put the checks in our CI?

For my own website, I run KaTeX deployment-side as part of the page-build operation, but since we're not already running any Node-based stuff here, it'd be a bit of a faff if we had to do that via Node, especially for local contributors.

Eric-Arellano:
> Any chance you know a simple way to get Sphinx to attempt the renders server side, so we can put the checks in our CI?

I don't know of anything off the top of my head, but I added that we should investigate this as part of https://github.com/Qiskit/documentation/issues/2228. I know Kevin Sung had switched Qiskit to Katex with Sphinx, so I'm surprised the Sphinx build pass - I have not yet tried triaging what went wrong. I only just found the issue to begin with.

For what it's worth, qiskit/documentation should have caught this issue much sooner. We have a check that every changed page renders, but it didn't correctly run. That's tracked by https://github.com/Qiskit/documentation/issues/2228. So, I didn't notice the issue until doing an open-source sync in the closed source repository. We recently got the nightly dev docs job working, so qiskit/documentation in theory will catch Latex issues every night. That might be sufficient.

jakelishman (approved):
KaTeX is also client-side JS-based rendering by default like MathJax, and our Sphinx output is certainly set up for client-side rendering, which is why it's not getting caught here.

Looks like `sphinxcontrib-katex` does have a way of doing the pre-rendering if Node is installed, so maybe there's a sensible path through.  I wouldn't want to do that as part of the actual build, though, since we don't want to be making the HTML rendering decisions in this repo.

Eric-Arellano:
> Looks like sphinxcontrib-katex does have a way of doing the pre-rendering if Node is installed

Good find. For posterity, the PyPI page says

```
To enable server side pre-rendering add in addition ([nodejs](https://nodejs.org/) installation needed):
katex_prerender = True
```

We fixed the pages render check in qiskit/documentation in https://github.com/Qiskit/documentation/pull/2389. Thanks to the nightly dev sync, we'll get fast feedback on when there are issues with Latex and the docs team can report it back to you. (Exception: when there's an rc, the `/dev` docs track that rc rather than `main` so issues added to `main` won't be caught until the stable release goes out.)

> I wouldn't want to do that as part of the actual build, though, since we don't want to be making the HTML rendering decisions in this repo.

Agreed. Given that we have nine Python APIs now(!) and that will grow to 10-11 soon(!), I think it will be the simplest to rely on qiskit/documentation to catch these Latex issues instead of setting up server-side rendering in each repo as a check.

PR #13392, opened 2024-11-03
Add missing pulse-related deprecation

eliarbel:
This commit deprecates `inst_map` in `PassManagerConfig.__init__`. It should have been part of the pulse deprecation PR #13164 but I somehow missed to cover it. We're already in our feature freeze period for version 1.3 but since this is not a new feature but rather an addendum to the pulse deprecation PR  we should squeeze this in as well.


ElePT (approved):
I think that it makes a lot of sense to include this PR in the release, else the deprecation would be incomplete. I'm glad you caught this on time! It totally flew over my head.

PR #13394, opened 2024-11-05
Deprecate loose custom basis gates in preset pm pipieline

ElePT:
### Summary
This PR is a spinoff of #12850 that introduces the deprecation warning for loose custom basis gates as an input to transpile/generate_preset_pm. This is a pre-requisite for eventually merging #12850. Moving forward, users can still provide custom basis gates but they need to be included as part of a `Target` instance.

### Details and comments
The PR also modifies a few oversights found in `generate_preset_pass_manager` that didn't have user impact but should be corrected to set the stage for a target-only pipeline:

- we were "over-skipping" the target by having a condition with an `or` instead of an `and` (fixed in L453/466)
- control flow ops were not included by default in the custom name mapping (this would raise a warning if control flow ops were provided as loose basis gates). This didn't cause issues because the target was often skipped with the previous oversight.

### DIFF
--- a/qiskit/synthesis/unitary/qsd.py
+++ b/qiskit/synthesis/unitary/qsd.py
@@ -251,14 +251,13 @@ def _get_ucry_cz(nqubits, angles):
 
 
 def _apply_a2(circ):
-    from qiskit.compiler import transpile
     from qiskit.quantum_info import Operator
     from qiskit.circuit.library.generalized_gates.unitary import UnitaryGate
+    from qiskit.transpiler.passes.synthesis import HighLevelSynthesis
 
     decomposer = two_qubit_decompose_up_to_diagonal
-    ccirc = transpile(
-        circ, basis_gates=["u", "cx", "qsd2q"], optimization_level=0, qubits_initially_zero=False
-    )
+    hls = HighLevelSynthesis(basis_gates=["u", "cx", "qsd2q"], qubits_initially_zero=False)
+    ccirc = hls(circ)
     ind2q = []
     # collect 2q instrs
     for i, instruction in enumerate(ccirc.data):
--- a/qiskit/transpiler/preset_passmanagers/generate_preset_pass_manager.py
+++ b/qiskit/transpiler/preset_passmanagers/generate_preset_pass_manager.py
@@ -446,7 +446,6 @@ def generate_preset_pass_manager(
 
 
 def _parse_basis_gates(basis_gates, backend, inst_map, skip_target):
-    name_mapping = {}
     standard_gates = get_standard_gate_name_mapping()
     # Add control flow gates by default to basis set and name mapping
     default_gates = {"measure", "delay", "reset"}.union(CONTROL_FLOW_OP_NAMES)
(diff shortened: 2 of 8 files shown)


releasenotes/notes/deprecate-custom-basis-gates-transpile-e4b5893377f23acb.yaml
```diff
@@ -0,0 +1,18 @@
+---
+deprecations_transpiler:
+  - |
+    Providing custom gates through the ``basis_gates`` argument is deprecated 
+    for both :func:`.transpile` and :func:`generate_preset_pass_manager`, this functionality
```
eliarbel, line 5: The custom `basis_gates` deprecation warning does not propagate through the `transpile` function when calling it with custom basis gates. This is because of the `warnings.simplefilter` filter here: https://github.com/Qiskit/qiskit/blob/2efb59ce4cd3c2d8721390648cdca6d2fd41f792/qiskit/compiler/transpiler.py#L398 
which is too general for this warning. It should be refined with `warnings.filterwarnings` and a regex message.

qiskit/transpiler/preset_passmanagers/generate_preset_pass_manager.py
```diff
@@ -433,9 +448,14 @@ def generate_preset_pass_manager(
 def _parse_basis_gates(basis_gates, backend, inst_map, skip_target):
     name_mapping = {}
```
eliarbel: Dead code now

eliarbel (changes requested):
Overall LGTM, thanks and good catch with the target-skipping condition! I have couple small comments, one is important.

ElePT:
Thanks for the review! Really good catch with `transpile`. This actually made me notice that it was used internally in the Quantum Shannon Decomposer, so I had to do a small refactoring there. The code should look good now!

qiskit/synthesis/unitary/qsd.py
```diff
@@ -251,14 +251,13 @@ def _get_ucry_cz(nqubits, angles):
...
     from qiskit.quantum_info import Operator
     from qiskit.circuit.library.generalized_gates.unitary import UnitaryGate
+    from qiskit.transpiler.passes.synthesis import HighLevelSynthesis
 
     decomposer = two_qubit_decompose_up_to_diagonal
-    ccirc = transpile(
-        circ, basis_gates=["u", "cx", "qsd2q"], optimization_level=0, qubits_initially_zero=False
-    )
+    hls = HighLevelSynthesis(basis_gates=["u", "cx", "qsd2q"], qubits_initially_zero=False)
+    ccirc = hls(circ)
```
eliarbel, line 260: Previously `transpile` built a flow containing `UnitarySynthesis`, `HighLevelSynthesis` and `BasisTranslator`. Looking at `qs_decomposition` seems that this change is safe in the sense that `HighLevelSynthesis` should be sufficient. Can you just confirm?

eliarbel (approved):
Overall looks good, thanks for all the changes and for handling the newly exposed deprecation warnings. I've left one question for you to confirm in `qsd.py`; assuming you will, I think this PR is good to go hence already approving.

qiskit/synthesis/unitary/qsd.py
ElePT: Yes, I believe it is, I also briefly confirmed it with @Cryoris.

PR #13302, opened 2024-10-09
Parameterize Commutation Checker

MarcDrudis:
### Summary
Add support for basic parameterized gates in the commutation checker (props to @Cryoris). This is currently only done for rotation gates that have a single generator (e.g. done for `RXX` but not for `XXPlusYY`). This has the power to improve compiler quality, but also speeds up the `CommutationChecker` in cases where we have to check if gates are parameterized a lot.

### Details and Comments

This works by replacing a parameterized gate by it's generator. For example, `RZ(theta)` commutes with a gate exactly if Z commutes with that gate. The special cases of `theta mod 2pi = 0` are also handled. For compiling a 100-qubit QFT circuit (which has loads of `CPhase` gates), the runtime on my laptop was reduced as:

```
(main) Took 2.1887816429138183 +- 0.3050131081682874s
(this PR) Took 1.8230558633804321 +- 0.19875805494141444s
```

which is more than I expected given the small substitution.

PR #13399, opened 2024-11-05
Fix ``CommutationChecker`` for comparison with rotation gates on angles $k\pi$

Cryoris:
### Summary

In #13302 a mistake creeped in, accidentally stating that rotation gates commute for angles that are multiples of $\pi$ instead of $2\pi$. I would've expected this to break somewhere in the test suite, now explicit tests are added. No reno as this is not yet released.


mtreinish (approved):
Good catch, I'm surprised we didn't hit this anywhere during review of the PR. But good to catch this now.

PR #13400, opened 2024-11-05
Solve strange behavior with `qubit_properties` in `Target`.

### LINKED ISSUE #13393: Target raises exception when `qubit_properties` is used without passing `num_qubits`
### Environment

- **Qiskit version**: `main`, like 4c3f8c93037351ff47279c0e2b5211f6bc275e98
- **Python version**: 3.12
- **Operating system**: Fedora Linux

### What is happening?

`Target` raises an exception when `qubit_properties` is used without passing `num_qubits`.

```python
from qiskit.transpiler import Target, QubitProperties

target = Target(
     qubit_properties=[QubitProperties(frequency=5e9)],
     dt=1e-9,
     granularity=16,
)
print(target.num_qubits)
```

### What should happen?

The example should print `1`. Instead it `Target` raises an exception: `ValueError: The value of num_qubits specified does not match the length of the input qubit_properties list`.

### Any suggestions?

@jakelishman suggested in https://github.com/qiskit-community/qiskit-experiments/pull/1478#discussion_r1828447285 that the `Target` code should be using a default of `None` instead of 0 for `num_qubits` to signal that the number of qubits should be inferred from the qubit properties rather than be set to 0.

raynelfss:
### Summary
Fixes #13393.

The following commits address some strange behavior when deciding the `num_qubits` attribute of a Target when only `qubit_properties` is provided.

### Details and comments
The previous implementation of the Target would always make sure to receive a real value for `num_qubits` or it would default to zero. Mistakenly because of the behavior of `Target::from_configuration` it was assumed that `num_qubits` could be `None` unless otherwise specified. However, the default behavior should be to make `num_qubits` zero unless specified otherwise.

The following commits fix this behavior by fixing the signature of the rust `Target` to have default values and adding an extra check when comparing against `qubit_properties`.

A test case based on @wshanks findings in #13393 was also included in these additions.

### DIFF
--- a/crates/accelerate/src/target_transpiler/mod.rs
+++ b/crates/accelerate/src/target_transpiler/mod.rs
@@ -227,18 +227,18 @@ impl Target {
     #[new]
     #[pyo3(signature = (
         description = None,
-        num_qubits = None,
+        num_qubits = 0,
         dt = None,
-        granularity = None,
-        min_length = None,
-        pulse_alignment = None,
-        acquire_alignment = None,
+        granularity = 1,
+        min_length = 1,
+        pulse_alignment = 1,
+        acquire_alignment = 1,
         qubit_properties = None,
         concurrent_measurements = None,
     ))]
     fn new(
         description: Option<String>,
-        num_qubits: Option<usize>,
+        mut num_qubits: Option<usize>,
         dt: Option<f64>,
         granularity: Option<u32>,
         min_length: Option<usize>,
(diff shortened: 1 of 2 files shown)


ElePT (commented):
Hmmm I am a bit weary of the change in the default value of `num_qubits` because I remember playing around with it with the refactoring of `BasicSimulator` and concluding that the default should stay `None`. However, it's true that the `BasicSimulator` target is instantiated explicitly with `num_qubits=None`, so maybe I remember a different issue. In any case, wouldn't this have to be documented as an interface change? I don't think it should be something users heavily rely on but I think this is part of the user-facing API, right?

mtreinish:
I think the switch to `None` was an accidentally breaking change that I missed in the review of #12292 looking at an old version of the target code it defaulted to 0: https://github.com/Qiskit/qiskit/blob/stable/1.1/qiskit/transpiler/target.py#L247 although reviewing #12292 now it looks like the python side constructor still used 0, but the rust used `None`

ElePT:
Aha, then maybe I remembered the issue the other way round, the default was `0` and I thought of changing it to `None`, and that broke some code down the line, so I just did it explicitly in the `BasicSimulator` construction. Then @raynelfss disregard my previous comment.

ElePT (approved):
Ok, double checking the historical target version @mtreinish linked, I think it makes sense to align the default values of the constructor with the python ones, both for `num_qubits` and the pulse-related args. LGTM!

PR #13405, opened 2024-11-07
Fix clippy issues flagged by latest rust version

mtreinish:
### Summary

This commit fixes 2 issues flagged by running by clippy with Rust 1.82. We run clippy in CI with Qiskit's MSRV of 1.70, as sometimes clippy will flag failures for not using a syntax which isn't compatible with the MSRV. These issues flagged here are real, and good to fix, especially the docstring one, the `////` comment isn't a docstring and that excludes the circuit diagram from the docs which wasn't the intent there.


ElePT (dismissed):
LGTM, do we want this for 1.3 rc1? I guess the queue isn't too busy at the moment so we could squeeze it in.

raynelfss:
If anything we can enqueue it as the last item before tagging. But currently there are two items in the merge queue.

ElePT (approved):

PR #13410, opened 2024-11-08
Use OnceLock instead of OnceCell

mtreinish:
### Summary

OnceLock is a thread-safe version of OnceCell that enables us to use PackedInstruction from a threaded environment. There is some overhead associated with this, primarily in memory as the OnceLock is a larger type than a OnceCell. But the tradeoff is worth it to start leverage multithreading for circuits.

### Details and comments

Fixes #13219

PR #13411, opened 2024-11-08
Perform BarrierBeforeFinalMeasurements analysis in parallel

mtreinish:
### Summary

With #13410 removing the non-threadsafe structure from our circuit
representation we're now able to read and iterate over a DAGCircuit from
multiple threads. This commit is the first small piece doing this, it
moves the analysis portion of the BarrierBeforeFinalMeasurements pass to
execute in parallel. The pass checks every node to ensure all it's
decedents are either a measure or a barrier before reaching the end of
the circuit. This commit iterates over all the nodes and does the check
in parallel.

### Details and comments

TODO:

- [x] Rebase after #13410 merges
- [x] Benchmark to test this actually speeds the pass up
- [x] Add handling to avoid using multithreading in a multiprocessing context


mtreinish:
I ran a benchmark with a quantum volume circuit and did a sweep from 10 to 1000 qubits/depthand ran the pass on it with 1.3.2, the pass running with a serial iterator and with a parallel iterator:

barrier_pass_time

Based on these results I went with a parallel threshold of 150 qubits in: https://github.com/Qiskit/qiskit/pull/13411/commits/b89c826d3b4a9449f4a5366b43d7fcd33f51746e so when we use a parallel iterator where the performance is better. This might vary on other environments though, so it would be useful for someone else to test this and we can adjust that value if it's not a good value.

crates/circuit/src/dag_circuit.rs
```diff
@@ -5626,6 +5633,15 @@ impl DAGCircuit {
...
 
+    /// Returns an iterator of tuples of (DAGNode, [DAGNodes]) where the DAGNode is the current node
+    /// and [DAGNode] is its successors in  BFS order.
+    pub fn bfs_predecessors(
+        &self,
+        node: NodeIndex,
+    ) -> impl Iterator<Item = (NodeIndex, Vec<NodeIndex>)> + '_ {
+        core_bfs_predecessors(&self.dag, node).filter(move |(_, others)| !others.is_empty())
+    }
+
```
raynelfss: I find it quite funny that we didn't have this one before 😄, I assume it's because we didn't use anything like it. But still, nice addition :)

crates/circuit/src/dag_circuit.rs
```diff
@@ -5626,6 +5633,15 @@ impl DAGCircuit {
         core_bfs_successors(&self.dag, node).filter(move |(_, others)| !others.is_empty())
     }
 
+    /// Returns an iterator of tuples of (DAGNode, [DAGNodes]) where the DAGNode is the current node
+    /// and [DAGNode] is its successors in  BFS order.
```
raynelfss:
```suggestion
    /// Returns an iterator of tuples of (DAGNode, [DAGNodes]) where the DAGNode is the current node
    /// and [DAGNode] is its predecessors in BFS order.
```

crates/circuit/src/dag_circuit.rs
```diff
@@ -4768,6 +4768,12 @@ def _format(operand):
 }
 
 impl DAGCircuit {
+    /// Returns an immutable view of the qubit io map
+    #[inline(always)]
+    pub fn qubit_io_map(&self) -> &[[NodeIndex; 2]] {
+        &self.qubit_io_map
+    }
```
raynelfss, line 4834: Do you think it'd make sense to have one of these getters for the `clbit_io_map` too? I assume it's not needed now, but for the sake of consistency?

crates/accelerate/src/barrier_before_final_measurement.rs
```diff
@@ -30,39 +30,87 @@ pub fn barrier_before_final_measurements(
...
+                                        suc_inst.op.name() == "measure"
+                                            || suc_inst.op.name() == "barrier"
+                                    }
+                                    _ => true,
+                                })
+                            })
+                        } else {
+                            false
+                        }
+                    }
```
raynelfss: Along this method and subsequent ones, we check the same condition (``Node is OpNode && Node.op.name in {"measure", "barrier"}``). Would it be possible to make this another closure? Or would it be inaccessible from within this one?

crates/circuit/src/dag_circuit.rs
```diff
@@ -5577,7 +5583,7 @@ impl DAGCircuit {
     /// Remove an operation node n.
     ///
     /// Add edges from predecessors to successors.
```
raynelfss, line 5644: You should change the docstring to reflect that the removed `PackedInstruction` is returned by the method.

crates/accelerate/src/barrier_before_final_measurement.rs
```diff
@@ -21,7 +21,7 @@ use qiskit_circuit::operations::{Operation, PyInstruction};
 use qiskit_circuit::packed_instruction::{PackedInstruction, PackedOperation};
 use qiskit_circuit::Qubit;
 
-static FINAL_OP_NAMES: [&str; 2] = ["measure", "barrier"];
+const PARALLEL_THRESHOLD: usize = 150;
```
raynelfss, line 23: Is there any reason behind this threshold? Just trying to understand the logic behind 150 (threads I assume?).

crates/accelerate/src/barrier_before_final_measurement.rs
```diff
@@ -30,39 +30,87 @@ pub fn barrier_before_final_measurements(
...
-        .op_nodes(true)
-        .filter_map(|(node, inst)| {
-            if !is_exactly_final(inst) {
-                return None;
-            }
-            dag.bfs_successors(node)
-                .all(|(_, child_successors)| {
-                    child_successors.iter().all(|suc| match dag[*suc] {
-                        NodeType::Operation(ref suc_inst) => is_exactly_final(suc_inst),
+    let find_final_nodes = |[_in_index, out_index]: &[NodeIndex; 2]| -> Vec<NodeIndex> {
```
raynelfss, line 34: Maybe include a small comment here about what this method does?

raynelfss (commented):
This looks straightforward to me and easy to understand. I just left a couple of comments related to the documentation of the code.

crates/accelerate/src/barrier_before_final_measurement.rs
mtreinish: See the discussion of this here: https://github.com/Qiskit/qiskit/pull/13411#issuecomment-2632701335 

The way the rayon iterator works is it spawns a threadpool for the number of threads on your local system and then divides up the tasks across that threadpool. On my local system through experimentation 150 was roughly the crossover point where the parallel version was faster than the serial version.

crates/accelerate/src/barrier_before_final_measurement.rs
mtreinish: I should change this to use `StandardInstruction` matching instead.

crates/circuit/src/dag_circuit.rs
mtreinish: Probably, in practice I don't think we do that very often in the transpiler. But having the method for consistency would be good.

crates/accelerate/src/barrier_before_final_measurement.rs
mtreinish: I rewrote the logic to use StandardInstruction matching instead of string equality in: https://github.com/Qiskit/qiskit/pull/13411/commits/c3dbc724c9e063e415f969910ad5fb7ff44430df

crates/circuit/src/dag_circuit.rs
mtreinish: I updated the docstring in: https://github.com/Qiskit/qiskit/pull/13411/commits/03c5ba6c55869f4f1db6317a1c68f7b072716e8e

crates/accelerate/src/barrier_before_final_measurement.rs
mtreinish: I added comments in: https://github.com/Qiskit/qiskit/pull/13411/commits/03c5ba6c55869f4f1db6317a1c68f7b072716e8e

crates/circuit/src/dag_circuit.rs
```diff
@@ -5636,7 +5642,11 @@ impl DAGCircuit {
...
-    pub fn remove_op_node(&mut self, index: NodeIndex) {
+    ///
+    /// # Returns
+    ///
+    /// The removed PackedInstruction is returned
```
raynelfss:
```suggestion
    /// The removed [PackedInstruction] is returned
```

raynelfss (dismissed):
LGTM! Just one minor docstring comment.

raynelfss (approved):
Happy merging!

PR #13414, opened 2024-11-08
Fix filesystem case sensitivity issue with IQP docs

Eric-Arellano:
This is breaking the documentation on macOS for the 1.3.0rc1 release because `qiskit.circuit.library.iqp` and `qiskit.circuit.library.IQP` are ambiguous on case-insensitive operating systems.

I renamed `qiskit.circuit.library.iqp` because that is the old API and I don't want to break the stable URL `qiskit.circuit.library.IQP`.

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -120,6 +120,14 @@ autosummary_generate_overwrite = False
 napoleon_google_docstring = True
 napoleon_numpy_docstring = False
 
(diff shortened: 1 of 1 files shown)


raynelfss (approved):
LGTM. Can confirm this bug was happening to me as well on MacOS but assumed it was some environment error. With these commits it seems to be running just fine.

PR #13421, opened 2024-11-11
Fix twirling asv benchmark

mtreinish:
### Summary

Since #13331 merged the nightly asv runs have been failing. This is because the asv twirling benchmark code was updated in that PR to use the new twirling function instead of the embedded equivalent. However, in the PR the name of that function was renamed from twirl_circuit to pauli_twirl_2q_gates during the development of the feature. This rename was missed in the asv benchmark, and as nothing executes the asv code in CI this went uncaught until the asv nightly runs. This commit fixes this oversight and updates the function name to the correct name so that the asv benchmarks no longer fail on import.

### DIFF
--- a/test/benchmarks/manipulate.py
+++ b/test/benchmarks/manipulate.py
@@ -17,7 +17,8 @@
 import os
 
 from qiskit import QuantumCircuit
-from qiskit.circuit import twirl_circuit
+
+from qiskit.circuit import pauli_twirl_2q_gates
 from qiskit.passmanager import PropertySet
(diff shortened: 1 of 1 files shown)


test/benchmarks/manipulate.py
```diff
@@ -17,7 +17,8 @@
 import os
 
 from qiskit import QuantumCircuit
-from qiskit.circuit import twirl_circuit
+
```
mtreinish:
```suggestion
```

kevinhartman (approved):

PR #13424, opened 2024-11-12
Avoid overhead for synthesized nodes lookup

mtreinish:
### Summary

After #12550 a hash implementation was added to the implementation of DAGOpNode to be able to have identical instances of dag nodes used be usable in a set or dict. This is because after #12550 changed the DAGCircuit so the DAGOpNode instances were just a python view of the data contained in the nodes of a dag. While prior to #12550 the actual DAGOpNode objects were returned by reference from DAG methods. However, this hash implementation has additional overhead compared to the object identity based version used before. This has caused a regression in some cases for high level synthesis when it's checking for nodes it's already synthesized. This commit addresses this by changing the dict key to be the node id instead of the node object. The integer hashing is significantly faster than the object hashing.

### Details and comments

### DIFF
--- a/qiskit/transpiler/passes/synthesis/high_level_synthesis.py
+++ b/qiskit/transpiler/passes/synthesis/high_level_synthesis.py
@@ -382,7 +382,7 @@ class HighLevelSynthesis(TransformationPass):
 
             # If the synthesis changed the operation (i.e. it is not None), store the result.
             if synthesized is not None:
-                synthesized_nodes[node] = (synthesized, synthesized_context)
+                synthesized_nodes[node._node_id] = (synthesized, synthesized_context)
 
             # If the synthesis did not change anything, just update the qubit tracker.
             elif not processed:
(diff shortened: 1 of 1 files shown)


raynelfss:
Thank you for this addition! This makes more sense. Is this improvement noticeable when benchmarking runtime?

mtreinish:
I did a quick asv run and it didn't flag anything as an improvement or a regression. But it is noticeable in some benchpress benchmarks it has a noticeable improvement for example running `qiskit_gym/abstract_transpile/test_hamiltonians.py::TestWorkoutAbstractHamiltonians::test_hamiltonians[ham_ham_JW-18-all-to-all]` it went goes from 5.5106774509768 sec with 1.3.0rc1 to 3.2585 sec with this PR.

qiskit/transpiler/passes/synthesis/high_level_synthesis.py
```diff
@@ -382,7 +382,7 @@ def _run(
 
             # If the synthesis changed the operation (i.e. it is not None), store the result.
             if synthesized is not None:
-                synthesized_nodes[node] = (synthesized, synthesized_context)
+                synthesized_nodes[node._node_id] = (synthesized, synthesized_context)
```
kevinhartman, line 385: This looks fine in this case since `synthesized_nodes` is only used with one DAG (and not e.g. shared by the recursive control flow block handling).

But this optimization is something we should be careful when applying in other places, since the semantics of uniqueness in keys shifts from being global to local within a specific DAG (which means that keys will clobber each other if node indices from more than a single DAG are used in the same map). This was responsible for a bug we had in the visualization code after the initial DAG port to Rust.

kevinhartman (approved):
LGTM, thanks!

Issue #13413, opened 2024-11-08
SparsePauliOp.to_matrix() exhibits numerical instabilities


donsano33:
### Environment

- **Qiskit version**: 1.2.4
- **Python version**: 3,11
- **Operating system**: Ubuntu 24.04.1 LTS

### What is happening?

When calculating the matrix representation of a `SparsePauliOp`  consisting of multiple terms  via `to_matrix()` , numbers that should be exactly zero are not zero but something around `e-18`.
This does not happen with `qiskit<1`.

### How can we reproduce the issue?

```python
from qiskit.quantum_info import SparsePauliOp

factor = 0.025
terms = [
    factor * SparsePauliOp(data=["IIII", "IIZI", "IIIZ", "IIZZ"], coeffs=[1,-1,-1, 1]),
    factor * SparsePauliOp(data=['IIII', 'IZII', 'IIZI', 'IZZI'], coeffs=[1,-1,-1, 1])
    ]
op = SparsePauliOp.sum(terms)
op_matrix = op.to_matrix(sparse=True)
print(op_matrix.diagonal())
```

This also appears when using `sparse=False`.

### What should happen?

In `qiskit<1.0` the output looks like this:

```
[0.  0.  0.  0.1
0.  0.  0.1 0.2
0.  0.  0.  0.1
0.  0.  0.1 0.2]
```

with `qiskit>1.0`, e.g. I use `qiskit==1.2.4` the output looks somewhat like this:

```
[ 6.9388939e-18  6.9388939e-18 -6.9388939e-18  1.0000000e-01
  0.0000000e+00  0.0000000e+00  1.0000000e-01  2.0000000e-01
  6.9388939e-18  6.9388939e-18 -6.9388939e-18  1.0000000e-01
  0.0000000e+00  0.0000000e+00  1.0000000e-01  2.0000000e-01]
```
Interestingly the positions of the small numbers changes when I rerun the same script with a new call. If I run the statements in a loop the positions of the small numbers is constant.

### Any suggestions?

Just a first guess by looking at the implementation of `SparsePauliOp.to_matrix()` the issue might either be in this function

https://github.com/Qiskit/qiskit/blob/cd26d498c9b74475753b9179ac2b7f84122ffe81/qiskit/quantum_info/operators/symplectic/sparse_pauli_op.py#L956

or rather in both of the `to_matrix_sparse()` 

[...]

PR #13439, opened 2024-11-14
Fix process-based non-determinism in `SparsePauliOp.to_matrix`

jakelishman:
### Summary

The simplification step of the summed Pauli terms in `SparsePauliOp.to_matrix` had introduced a non-determinism via hash-map iteration.  In practice, the base hash seed was set per initialisation of the extension module, then stayed the same.  This is hard to detect from within tests, but caused unexpected numerical behaviour on different invocations of the same script.

### Details and comments

This fixes the bug part of #13413, but doesn't address the feature request.  It should be eligible for backport to either 1.3.0 or 1.3.1, depending on whether there's time to make 1.3.0.

Writing a test that would exercise this would be pretty complex, so I haven't done it.  I can if we feel strongly about it - we'd likely embed a little script into a test and spawn some `sys.executable` instances to run it and check the outputs all matched bit-for-bit.

### DIFF
--- a/crates/accelerate/src/sparse_pauli_op.rs
+++ b/crates/accelerate/src/sparse_pauli_op.rs
@@ -20,6 +20,7 @@ use numpy::prelude::*;
 use numpy::{PyArray1, PyArray2, PyReadonlyArray1, PyReadonlyArray2, PyUntypedArrayMethods};
 
 use hashbrown::HashMap;
+use indexmap::IndexMap;
 use ndarray::{s, ArrayView1, ArrayView2, Axis};
 use num_complex::Complex64;
 use num_traits::Zero;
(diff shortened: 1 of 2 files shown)


crates/accelerate/src/sparse_pauli_op.rs
```diff
@@ -298,7 +299,7 @@ impl MatrixCompressedPaulis {
     /// explicitly stored operations, if there are duplicates.  After the summation, any terms that
     /// have become zero are dropped.
     pub fn combine(&mut self) {
-        let mut hash_table = HashMap::<(u64, u64), Complex64>::with_capacity(self.coeffs.len());
+        let mut hash_table = IndexMap::<(u64, u64), Complex64>::with_capacity(self.coeffs.len());
```
mtreinish: Do you want to use ahash for the hasher here so the lookup performance is basically the same as `HashMap`?

jakelishman: It should have no meaningful effect on the whole algorithm's performance because this bit is immeasurable compared to the cost of constructing the matrix, but the only reason I didn't do it is because I forgot.

My kids are awake now, so if somebody wanted to push up that change in time to get this into 1.3.0, it's fine by me - I won't have time til tomorrow.

kevinhartman: Done in 23ba5f4 🙂

kevinhartman (approved):
LGTM. I don't feel strongly that this needs a test given the circumstances.

If anyone disagrees, we can add one in a separate PR.

PR #13419, opened 2024-11-10
Add new multithreaded TwoQubitPeepholeOptimization pass

mtreinish:
### Summary

This commit adds a new transpiler pass for physical optimization,
TwoQubitPeepholeOptimization. This replaces the use of Collect2qBlocks,
ConsolidateBlocks, and UnitarySynthesis in the optimization stage for
a default pass manager setup. The pass logically works the same way
where it analyzes the dag to get a list of 2q runs, calculates the matrix
of each run, and then synthesizes the matrix and substitutes it inplace.
The distinction this pass makes though is it does this all in a single
pass and also parallelizes the matrix calculation and synthesis steps
because there is no data dependency there.

This new pass is not meant to fully replace the Collect2qBlocks,
ConsolidateBlocks, or UnitarySynthesis passes as those also run in
contexts where we don't have a physical circuit. This is meant instead
to replace their usage in the optimization stage only. Accordingly this
new pass also changes the logic on how we select the synthesis to use
and when to make a substitution. Previously this logic was primarily done
via the ConsolidateBlocks pass by only consolidating to a UnitaryGate if
the number of basis gates needed based on the weyl chamber coordinates
was less than the number of 2q gates in the block (see https://github.com/Qiskit/qiskit/issues/11659 for
discussion on this). Since this new pass skips the explicit
consolidation stage we go ahead and try all the available synthesizers

Right now this commit has a number of limitations, the largest are:

- Only supports the target
- It doesn't support the XX decomposer because it's not in rust (the `TwoQubitBasisDecomposer` and `TwoQubitControlledUDecomposer` are used)

This pass doesn't support using the unitary synthesis plugin interface, since
it's optimized to use Qiskit's built-in two qubit synthesis routines written in
Rust. The existing combination of `ConsolidateBlocks` and `UnitarySynthesis`
should be used instead if the plugin interface is necessary.

### Details and comments

Fixes #12007
Fixes #11659

TODO:

- [x] Rebase after #13410 merges
- [x] Add tests
- [x] Add documentation
- [x] Benchmarking and performance tuning
- [x] Handle running serially when in multiprocessing context
- [x] Release note

PR #13450, opened 2024-11-18
Fix post-oxidization change in `ConsolidateBlocks` behavior 

### LINKED ISSUE #13438: Optimization level 2 and 3 result in a deeper circuit with 1.3.0rc1 and main branch
### Environment

- **Qiskit version**: 1.3.0rc1, main branch
- **Python version**: 3.12.7
- **Operating system**: 15.1

### What is happening?

Transpilation with optimization level 2 and 3 result in a deeper circuit than optimization level 1.
The circuit of level 2 is deeper than that of level 1.
The circuit of level 3 is deeper than that of level 2.
This does not happen with Qiskit 1.2.4.

```python
from qiskit import generate_preset_pass_manager, QuantumCircuit

qc = QuantumCircuit(2)
qc.cz(0, 1)
qc.sx([0, 1])
qc.cz(0, 1)

for level in [1, 2, 3]:
    pm = generate_preset_pass_manager(optimization_level=level, basis_gates=["rz", "rzz", "sx", "x", "rx"])
    qc2 = pm.run(qc)
    print('optimization level', level)
    print(qc2)
    print()
```

```
# 1.2.4
optimization level 1
global phase: π
     ┌─────────┐           ┌─────────┐┌─────────┐
q_0: ┤ Rz(π/2) ├─■─────────┤ Rx(π/2) ├┤ Rz(π/2) ├─■─────────
     ├─────────┤ │ZZ(-π/2) ├─────────┤├─────────┤ │ZZ(-π/2)
q_1: ┤ Rz(π/2) ├─■─────────┤ Rx(π/2) ├┤ Rz(π/2) ├─■─────────
     └─────────┘           └─────────┘└─────────┘

optimization level 2
global phase: π
     ┌─────────┐           ┌─────────┐┌─────────┐
q_0: ┤ Rz(π/2) ├─■─────────┤ Rx(π/2) ├┤ Rz(π/2) ├─■─────────
     ├─────────┤ │ZZ(-π/2) ├─────────┤├─────────┤ │ZZ(-π/2)
q_1: ┤ Rz(π/2) ├─■─────────┤ Rx(π/2) ├┤ Rz(π/2) ├─■─────────
     └─────────┘           └─────────┘└─────────┘

optimization level 3
global phase: π
     ┌─────────┐           ┌─────────┐┌─────────┐
q_0: ┤ Rz(π/2) ├─■─────────┤ Rx(π/2) ├┤ Rz(π/2) ├─■─────────
     ├─────────┤ │ZZ(-π/2) ├─────────┤├─────────┤ │ZZ(-π/2)
q_1: ┤ Rz(π/2) ├─■─────────┤ Rx(π/2) ├┤ Rz(π/2) ├─■─────────
     └─────────┘           └─────────┘└─────────┘
```

```
# 1.3.0rc1 and main branch
optimization level 1
global phase: 7π/4
               ┌──────────┐   ┌────┐            ┌──────────┐
q_0: ─■────────┤ Rz(-π/2) ├───┤ √X ├───■────────┤ Rz(-π/2) ├
      │ZZ(π/2) ├──────────┤┌──┴────┴─┐ │ZZ(π/2) ├──────────┤
q_1: ─■────────┤ Rz(-π/2) ├┤ Rx(π/2) ├─■────────┤ Rz(-π/2) ├
               └──────────┘└─────────┘          └──────────┘

optimization level 2
global phase: 3π/2
                                                     ┌──────────┐ ┌─────────┐ ┌─────────┐              ┌──────────┐  ┌─────────┐┌─────────┐
q_0: ──────────────────────────────────────■─────────┤ Rz(-π/2) ├─┤ Rx(π/2) ├─┤ Rz(π/2) ├───■──────────┤ Rz(-π/2) ├──┤ Rx(π/2) ├┤ Rz(π/2) ├
     ┌──────────┐┌─────────┐┌────────────┐ │ZZ(π/2) ┌┴──────────┴┐├─────────┤┌┴─────────┴─┐ │ZZ(π/2) ┌─┴──────────┴─┐└┬───────┬┘└─────────┘
q_1: ┤ Rz(-π/2) ├┤ Rx(π/2) ├┤ Rz(1.6719) ├─■────────┤ Rz(1.4697) ├┤ Rx(π/2) ├┤ Rz(1.6719) ├─■────────┤ Rz(-0.10111) ├─┤ Rx(π) ├────────────
     └──────────┘└─────────┘└────────────┘          └────────────┘└─────────┘└────────────┘          └──────────────┘ └───────┘

optimization level 3
global phase: π
     ┌─────────────┐┌────┐  ┌───────┐   ┌────┐┌────────────┐                                      ┌──────────┐┌─────────┐   ┌────┐  ┌──────────┐»
q_0: ┤ Rz(-1.9513) ├┤ √X ├──┤ Rz(π) ├───┤ √X ├┤ Rz(11.376) ├─────────────────────────────■────────┤ Rz(-π/2) ├┤ Rz(π/2) ├───┤ √X ├──┤ Rz(3π/2) ├»
     └─┬─────────┬─┘├────┤┌─┴───────┴──┐├────┤└┬──────────┬┘┌─────────┐┌────┐┌─────────┐ │ZZ(π/2) ├─────────┬┘└──┬────┬─┘┌──┴────┴─┐├──────────┤»
q_1: ──┤ Rz(π/2) ├──┤ √X ├┤ Rz(3.2427) ├┤ √X ├─┤ Rz(5π/2) ├─┤ Rz(π/2) ├┤ √X ├┤ Rz(π/2) ├─■────────┤ Rz(π/2) ├────┤ √X ├──┤ Rz(π/2) ├┤ Rx(-π/2) ├»
       └─────────┘  └────┘└────────────┘└────┘ └──────────┘ └─────────┘└────┘└─────────┘          └─────────┘    └────┘  └─────────┘└──────────┘»
«          ┌────┐    ┌────────┐                                                                        ┌──────────┐┌─────────┐   ┌────┐  ┌──────────┐»
«q_0: ─────┤ √X ├────┤ Rz(3π) ├───────────────────────────────────────────────────────────────■────────┤ Rz(-π/2) ├┤ Rz(π/2) ├───┤ √X ├──┤ Rz(3π/2) ├»
«     ┌────┴────┴───┐└─┬────┬─┘┌────────────┐┌────┐┌────────────┐┌─────────┐┌────┐┌─────────┐ │ZZ(π/2) ├─────────┬┘└──┬────┬─┘┌──┴────┴─┐├──────────┤»
«q_1: ┤ Rz(-3.1314) ├──┤ √X ├──┤ Rz(4.6118) ├┤ √X ├┤ Rz(10.894) ├┤ Rz(π/2) ├┤ √X ├┤ Rz(π/2) ├─■────────┤ Rz(π/2) ├────┤ √X ├──┤ Rz(π/2) ├┤ Rx(-π/2) ├»
«     └─────────────┘  └────┘  └────────────┘└────┘└────────────┘└─────────┘└────┘└─────────┘          └─────────┘    └────┘  └─────────┘└──────────┘»
«       ┌────┐  ┌────────┐
«q_0: ──┤ √X ├──┤ Rz(2π) ├────────────────────────────────
«     ┌─┴────┴─┐└─┬────┬─┘┌──────────┐┌────┐┌────────────┐
«q_1: ┤ Rz(-π) ├──┤ √X ├──┤ Rz(3π/2) ├┤ √X ├┤ Rz(7.7529) ├
«     └────────┘  └────┘  └──────────┘└────┘└────────────┘

```

### What should happen?

Circuits of level 2 and 3 should be the same size as that of level 1 to be compatible with 1.2.4.

### Any suggestions?

_No response_

ElePT:
### Summary
Fixes the behavioral difference introduced by #13368 that caused #13438. 

The original implementation of the pass used to set the `ConsolidateBlocks` internal decomposer calling `unitary_synthesis._decomposer_2q_from_basis_gates`, which, when `kak_basis` wasn't found, would return `None`. During the oxidization efforts, `unitary_synthesis._decomposer_2q_from_basis_gates` was replaced with inline code that didn't contemplate this path (always set a default decomposer), which in the reproducing example led to a longer output circuit (the gates were collected into a unitary block and then decomposed differently).

### Details and comments
Technically not a bugfix because the change wasn't released yet.

Using the example shown in the issue ->

`ConsolidateBlocks` output dag with 1.2.4:
image

`ConsolidateBlocks` output dag with 1.3.0rc1:
image

`ConsolidateBlocks` output dag after this fix:
image

### DIFF
--- a/qiskit/transpiler/passes/optimization/consolidate_blocks.py
+++ b/qiskit/transpiler/passes/optimization/consolidate_blocks.py
@@ -12,9 +12,11 @@
 
 """Replace each block of consecutive gates by a single Unitary node."""
 from __future__ import annotations
+from math import pi
 
 from qiskit.synthesis.two_qubit import TwoQubitBasisDecomposer
-from qiskit.circuit.library.standard_gates import CXGate, CZGate, iSwapGate, ECRGate
+from qiskit.circuit.library.standard_gates import CXGate, CZGate, iSwapGate, ECRGate, RXXGate
+
 from qiskit.transpiler.basepasses import TransformationPass
 from qiskit.transpiler.passmanager import PassManager
 from qiskit._accelerate.consolidate_blocks import consolidate_blocks
@@ -27,6 +29,7 @@ KAK_GATE_NAMES = {
     "cz": CZGate(),
     "iswap": iSwapGate(),
     "ecr": ECRGate(),
+    "rxx": RXXGate(pi / 2),
 }
 
 
@@ -70,7 +73,6 @@ class ConsolidateBlocks(TransformationPass):
         if basis_gates is not None:
             self.basis_gates = set(basis_gates)
         self.force_consolidate = force_consolidate
-
         if kak_basis_gate is not None:
             self.decomposer = TwoQubitBasisDecomposer(kak_basis_gate)
         elif basis_gates is not None:
--- a/qiskit/transpiler/passes/optimization/consolidate_blocks.py
+++ b/qiskit/transpiler/passes/optimization/consolidate_blocks.py
@@ -81,6 +81,10 @@ class ConsolidateBlocks(TransformationPass):
                 self.decomposer = TwoQubitBasisDecomposer(
                     KAK_GATE_NAMES[kak_gates.pop()], basis_fidelity=approximation_degree or 1.0
                 )
+            elif "rzx" in basis_gates:
+                self.decomposer = TwoQubitBasisDecomposer(
+                    CXGate(), basis_fidelity=approximation_degree or 1.0
+                )
             else:
                 self.decomposer = None
         else:
--- a/qiskit/transpiler/passes/optimization/consolidate_blocks.py
+++ b/qiskit/transpiler/passes/optimization/consolidate_blocks.py
@@ -80,7 +80,7 @@ class ConsolidateBlocks(TransformationPass):
                     KAK_GATE_NAMES[kak_gates.pop()], basis_fidelity=approximation_degree or 1.0
                 )
             else:
-                self.decomposer = TwoQubitBasisDecomposer(CXGate())
+                self.decomposer = None
         else:
             self.decomposer = TwoQubitBasisDecomposer(CXGate())
 
--- a/test/python/transpiler/test_consolidate_blocks.py
+++ b/test/python/transpiler/test_consolidate_blocks.py
@@ -16,20 +16,19 @@ Tests for the ConsolidateBlocks transpiler pass.
 
 import unittest
 import numpy as np
+from ddt import ddt, data
 
 from qiskit.circuit import QuantumCircuit, QuantumRegister, IfElseOp, Gate
 from qiskit.circuit.library import U2Gate, SwapGate, CXGate, CZGate, UnitaryGate
 from qiskit.converters import circuit_to_dag
-from qiskit.transpiler.passes import ConsolidateBlocks
 from qiskit.quantum_info.operators import Operator
 from qiskit.quantum_info.operators.measures import process_fidelity
-from qiskit.transpiler import PassManager
-from qiskit.transpiler import Target
-from qiskit.transpiler.passes import Collect1qRuns
-from qiskit.transpiler.passes import Collect2qBlocks
+from qiskit.transpiler import PassManager, Target, generate_preset_pass_manager
+from qiskit.transpiler.passes import ConsolidateBlocks, Collect1qRuns, Collect2qBlocks
 from test import QiskitTestCase  # pylint: disable=wrong-import-order
 
 
+@ddt
 class TestConsolidateBlocks(QiskitTestCase):
     """
     Tests to verify that consolidating blocks of gates into unitaries
@@ -570,6 +569,26 @@ class TestConsolidateBlocks(QiskitTestCase):
         res = pm.run(qc)
 
         self.assertEqual(res, qc)
+
+    @data(2, 3)
+    def test_no_kak_gates_in_preset_pm(self, opt_level):
+        """Test correct initialization of ConsolidateBlocks pass when kak_gates aren't found.
+        Reproduces https://github.com/Qiskit/qiskit/issues/13438."""
+
+        qc = QuantumCircuit(2)
+        qc.cz(0, 1)
+        qc.sx([0, 1])
+        qc.cz(0, 1)
+
+        ref_pm = generate_preset_pass_manager(
+            optimization_level=1, basis_gates=["rz", "rzz", "sx", "x", "rx"]
+        )
+        ref_tqc = ref_pm.run(qc)
+        pm = generate_preset_pass_manager(
+            optimization_level=opt_level, basis_gates=["rz", "rzz", "sx", "x", "rx"]
+        )
+        tqc = pm.run(qc)
+        self.assertEqual(ref_tqc, tqc)
 
 
 if __name__ == "__main__":


mtreinish (dismissed):
Ah, good catch.it was a change in behavior on my consolidate blocks pr not unitary synthesis. I remember changing that logic at the time and thinking it wouldn't impact anything. I was coming at it from the mindset that consolidate blocks is too likely to skip consolidation when it shouldn't.

Realistically we should be relying on the xx decomposer with the rzz in the basis gates list, but there isn't an interface in rust yet to tell us the estimated number of 2q basis gates needed for a unitary, which is why I changed the logic here. Also that is not really a good meteic anyway, just more reason for me to keep working on #13419 and introduce a better approach.

ElePT:
Changing the "default" to `None` now left the paths for `rzz` and `rzx` gates unaccounted for. That broke the unit test that was fixed in [e3124a6](https://github.com/Qiskit/qiskit/pull/13450/commits/e3124a615f7e89d1ae7589e48e0c860780c9272d). So, questions:
1. any specific reason we had not to have `rzz` in the list? (because I just re-added it)
2. assuming that we don't want to use the `XXDecomposer` for `rzx`. What do we want to have then, `None` or `TwoQubitBasisDecomposer(CXGate())`? It currently doesn't break any test but just to confirm.

qiskit/transpiler/passes/optimization/consolidate_blocks.py
```diff
@@ -27,6 +29,7 @@
     "cz": CZGate(),
     "iswap": iSwapGate(),
     "ecr": ECRGate(),
+    "rxx": RXXGate(pi / 2),
```
ShellyGarion, line 32: why is the angle here fixed to pi/2 ?

mtreinish: It's because that's what `UnitarySynthesis` uses if it's a `basis_gates` only setup: https://github.com/Qiskit/qiskit/blob/main/qiskit/transpiler/passes/synthesis/unitary_synthesis.py#L98 the idea is that it's trying to match the decomposer used for estimating to what the synthesis pass will use later.

The whole logic between the passes is more than a bit hacky. `ConsolidateBlocks` tries to guess how many 2q gates will be used to decide whether to consolidate the block into a unitary or not. This is strongly coupled to the behavior of `UnitarySynthesis` because that's what actually does the synthesis later. But none of this actually takes into account multiple possible decomposers/synthesis methods being supported by a target, or the presence of a non-default plugin being used.

mtreinish: FWIW, I did some testing locally with this and in some cases using `TwoQubitBasisDecomposer(RXXGate(pi/2))` under estimates what the `XXDecomposer` estimates, 3 vs 4 gates. This might result in some blocks not getting consolidated which were before, but in practice I find it probably pretty unlikely since it only counts rzz gates when evaluating if the block's gates are more than the expected synthesis output.

ElePT:
I added a small refactoring on top of the previous changes to make the new implementation hopefully closer to the old one. Now:
- ~The default decomposer is `None` when no basis_gates/kak_gates are found (same as OG). This includes the `rzz` gate from the issue~ [EDIT: didn't work. Reverted that change]
- If an `rxx` gate is found, the decomposer will be  the `TwoQubitBasisDecomposer` with a `RXX(pi/2)` (same as OG)
- If an `rzx` gate is found, the decomposer will be the `TwoQubitBasisDecomposer` with a `CXGate` (same as backup decomposer from OG implementation)

I also added a comment that mentions the previous use of the XXDecomposer for the `rzx` case for clarity.

qiskit/transpiler/passes/optimization/consolidate_blocks.py
```diff
@@ -70,19 +81,22 @@ def __init__(
...
-                )
-            else:
-                self.decomposer = TwoQubitBasisDecomposer(CXGate())
-        else:
-            self.decomposer = TwoQubitBasisDecomposer(CXGate())
+            kak_gate = KAK_GATE_NAMES.keys() & (basis_gates or [])
+            if kak_gate:
+                if isinstance(kak_gate, RZXGate):
+                    # Backup optimizer from original implementation.
+                    # Once XXDecomposer is migrated to rust, this line could be rewritten ot use it.
```
ElePT:
```suggestion
                    # Once XXDecomposer is migrated to rust, this line could be rewritten to use it.
```

mtreinish (approved):
I'm fine with this it returns the previous state of logic. There might be some subtle differences in what gets consolidated or not with rzx based on differences between `XXDecomposer.num_basis_gates` and `TwoQubitBasisDecomposer`'s logic, but it should be close enough to the 1.2.x state of the logic that I'm not worried. 

I think this whole exercise reinforces for me how we need to improve the logic around block consolidation. #13419 will do this for the optimization loop, but we will need to think what makes the most sense for when we run `ConsolidateBlocks` prior to layout in the init stage.

PR #13452, opened 2024-11-18
Uppercase `Qiskit` in deprecation warnings

Cryoris:
### Summary

This is a suggestion to change the default deprecation message to use `Qiskit 1.x` instead of `qiskit 1.x`, since we commonly refer to this package in uppercase spelling.

### Details and comments

This means the message is changed e.g. from
```
The property ``qiskit.dagcircuit.dagcircuit.DAGCircuit.unit`` is deprecated as of qiskit 1.3.0.
```
to
```
The property ``qiskit.dagcircuit.dagcircuit.DAGCircuit.unit`` is deprecated as of Qiskit 1.3.0.
```

The old format can still be obtained by explicitly setting `package_name="qiskit"` in the deprecation decorators.

### DIFF
--- a/qiskit/utils/deprecation.py
+++ b/qiskit/utils/deprecation.py
@@ -109,7 +109,7 @@ def deprecate_arg(
     additional_msg: str | None = None,
     deprecation_description: str | None = None,
     pending: bool = False,
-    package_name: str = "qiskit",
+    package_name: str = "Qiskit",
     new_alias: str | None = None,
     predicate: Callable[[Any], bool] | None = None,
     removal_timeline: str = "no earlier than 3 months after the release date",
--- a/qiskit/utils/deprecation.py
+++ b/qiskit/utils/deprecation.py
@@ -45,7 +45,7 @@ def deprecate_func(
             For example, "Instead, use the function ``new_func`` from the module
             ``<my_module>.<my_submodule>``, which is similar but uses GPU acceleration."
         pending: Set to ``True`` if the deprecation is still pending.
-        package_name: The PyPI package name, e.g. "qiskit-nature".
+        package_name: The package name shown in the deprecation message (e.g. the PyPI package name).
         removal_timeline: How soon can this deprecation be removed? Expects a value
             like "no sooner than 6 months after the latest release" or "in release 9.99".
         is_property: If the deprecated function is a `@property`, set this to True so that the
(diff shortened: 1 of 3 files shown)


ElePT (dismissed):
Yesss, been wanting to do it for a long time. LGTM

qiskit/utils/deprecation.py
```diff
@@ -26,7 +26,7 @@ def deprecate_func(
     since: str,
     additional_msg: str | None = None,
     pending: bool = False,
-    package_name: str = "qiskit",
+    package_name: str = "Qiskit",
```
kevinhartman, line 29: The docstring for this argument indicates that it ought to be the PyPI package name. Assuming we want to make this change, we ought to update the docstring (and probably the arg name as well, e.g. `package_description`).

Do we use `deprecate_func` for packages other than Qiskit at this point? If not, perhaps we should just remove the argument altogether and hard-code `"Qiskit"` within.

kevinhartman (commented):
I like this change from a user perspective.

I think we should make sure that the signature / documented use of `deprecate_func` lines up with the new expected usage.

Looks like the tests need updating to look for the new string.

qiskit/utils/deprecation.py
Cryoris: Yeah I stumbled over this too... I'd prefer not to change the argument name for backwards compatibility (and avoid using `@deprecate_arg` on this 😂), but we should change the description. Maybe 
```
package_name: The name shown in the deprecation message (use e.g. the PyPI package name).
```
?

Cryoris: Updated in 5dd8262, along with a test that setting `package_name` works as expected

1ucian0:
Super minor comment: could it be `1.3` instead of `1.3.0`?

Cryoris:
@1ucian0 whether it's 1.3 or 1.3.0 is set by the caller, we're not setting a default here 🙂 some other parts in the code also explicitly with 1.3.0 so I'd prefer not changing it, if it's fine by you 🙂

ElePT (approved):
I find the argument from @Cryoris convincing, but won't add the PR to the queue in case @1ucian0 has any last-minute objection.

1ucian0 (approved):

PR #13460, opened 2024-11-20
Fix custom basis gates deprecation warning in `generate_preset_pass_manager`

ElePT:
### Summary
Small improvements on custom basis gates deprecation message for `generate_preset_pass_manager`:

- fixed missing space that lead to "providecustom" instead of "provide custom"
- replaced "custom" with "non-standard" and added an explicit mention to the gate that triggers the deprecation warning after @Cryoris getting the message triggered with ``ccp`` and not being able to pinpoint why (it's not custom, but also not standard).


eliarbel (approved):
LGTM

PR #12571, opened 2024-06-13
Bump version of macOS used on Azure

jakelishman:
### Summary

The macOS 11 runners are deprecated pending removal.  While macOS 14 is available, it's still marked as in preview on Azure, so macOS 13 is the current "latest" stable version.

### Details and comments

macOS 11 images are pending removal on the 28th of June, and will be subject to a brownout on the 17th.

PR #13470, opened 2024-11-20
Update macOS x86_64 wheel build CI runner to 10.13

mtreinish:
### Summary

Github actions has deprecated their macOS 10.12 CI runner and will be removing it in December, 2024. We were using this macOS runner because our target OS version for intel mac is 10.12. However, with this enviornment being removed by Github we must upgrade to continue running the jobs.

This commit does not bump the MACOSX_DEPLOYMENT_TARGET environment variable (which is used to set the target OS version for intel macs) though, because we should test if we can build it compatibility mode from a newer version of macOS. If this fails then we can bump this to 10.13 in a subsequent commit on this PR and add an upgrade release note to document the change.

### Details and comments


jakelishman:
It's actually macOS 13, not 10.13 - they left the OSX numbering behind with the release of macOS 11 for Apple silicon.

jakelishman (dismissed):

jakelishman:
Oh, I think you might have made a mistake in your reasoning in the top comment - the macOS 12 OS is already way newer (Monterey, released 2021) than the 10.12 minimum version (Sierra, released 2016) we target.  We still used macOS 12 in the wheel builds only because you made the minimal bump for GHA while I made the maximal stable bump for Azure (#12882 vs #12571).

mtreinish:
> Oh, I think you might have made a mistake in your reasoning in the top comment - the macOS 12 OS is already way newer (Monterey, released 2021) than the 10.12 minimum version (Sierra, released 2016) we target. We still used macOS 12 in the wheel builds only because you made the minimal bump for GHA while I made the maximal stable bump for Azure (https://github.com/Qiskit/qiskit/pull/12882 vs https://github.com/Qiskit/qiskit/pull/12571).

You're right I got confused by all the 12s in the version numbers. I was confusing 10.12 and macOS 12 (which in my head is still mac osx 12 and I have to remember to translate). That being said I think we found that we're not even setting the deployment target correctly with the qiskit 1.3.0 release and this will be important to combine with #13503. I guess the question for me is do you want me to just combine this with #13503 or should we just merge this as is (I can update the commit message to not refer to the dpeloyment target at all) and then tackle the target version as a follow up.

Cryoris:
IMO it would be nicer to combine this with #13503 if it's important to go together, such that we have a single commit touching the CI config and we can run the wheel tests together.

Cryoris (approved):
The wheels seem to be building 👍🏻

kevinhartman (approved):
LGTM, thanks!

kevinhartman:
As it turns out, we can't use `MACOSX_DEPLOYMENT_TARGET` to get a `10.12` build:

Screenshot 2024-12-12 at 4 18 23 PM

We should open a follow-up PR to bump the deployment targets as well, and possibly add a release note to document the change.

PR #13472, opened 2024-11-21
Add new `get_control_flow_name_mapping` function

### LINKED ISSUE #13468: Equivalent to `get_standard_gate_name_mapping` for control-flow ops
Lazy follow up request -- it would be nice to have an equivalent to `get_standard_gate_name_mapping` as well to standardize code like this:

https://github.com/Qiskit/qiskit/blob/b9d5c9c6aeb4568b6e9fba8943517ebed300886d/qiskit/transpiler/preset_passmanagers/generate_preset_pass_manager.py#L459-L464

It came up in https://github.com/Qiskit/qiskit-ibm-runtime/pull/2044#discussion_r1846328508. I will keep it on my list but I can't do it at the moment.

_Originally posted by @wshanks in https://github.com/Qiskit/qiskit/issues/13057#issuecomment-2489568891_

ElePT:
### Summary
Addresses #13468 by exposing the mapping done in `generate_preset_pass_manager` as a public function for convenience.

### DIFF
--- a/qiskit/circuit/__init__.py
+++ b/qiskit/circuit/__init__.py
@@ -693,8 +693,8 @@ attributes of each of the control-flow operations.
 
     Set of the instruction names of Qiskit's known control-flow operations.
 
-In addition to this, the function :func:`.get_control_flow_name_mapping` returns a dictionary
-mapping the names of the control-flow operations to their corresponding classes."
+The :func:`.get_control_flow_name_mapping` function allows to access the control-flow operation
+classes associated to each name.
 
 .. autofunction:: get_control_flow_name_mapping
 
--- a/qiskit/circuit/__init__.py
+++ b/qiskit/circuit/__init__.py
@@ -1307,6 +1307,7 @@ from .controlflow import (
     BreakLoopOp,
     ContinueLoopOp,
     CONTROL_FLOW_OP_NAMES,
+    get_control_flow_name_mapping,
 )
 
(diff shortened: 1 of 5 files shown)


qiskit/circuit/__init__.py
jakelishman, line 1: Don't forget to add it to the documentation.

wshanks:
The way this is implemented matches how it has been used so I think it is good. One point that we should note at least is that the standard gate version of this function returns a mapping to gate instances while the control flow function here returns a mapping to classes. The gates version gets away with this by using `Parameter`s to fill in the gate parameters, but I don't think there are reasonable equivalents for the conditions and bodies of the conditionals (could use size 1 registers and empty circuits but it seems weird).

By the way, you can also replace this case:

https://github.com/Qiskit/qiskit/blob/5f3f59469fcd9edae2feb80f74026265ebceb29a/qiskit/providers/backend_compat.py#L94-L99

ElePT:
Thanks @wshanks!  I applied your suggestion in [5434583](https://github.com/Qiskit/qiskit/pull/13472/commits/54345832bc005333fc92734f2dd8641716f8661f). you're right, the function doesn't match the gate version exactly, but I also didn't find any reasonable way to instantiate the gates, and tried to avoid making the parallelism in the docs for that reason. There is a risk that users might still associate them but well... we could always highlight the difference in documentation if necessary.

mtreinish (approved):

PR #13474, opened 2024-11-21
Migrate CI to Github Actions

### LINKED ISSUE #12807: Migrate CI to github actions

raynelfss:
### Summary
Fixes #12807. 
The following commits move existing workflows from Azure into Github Actions.

### Details and comments
As simple as it may seem, our CI Pipeline goes through several workflows, some of which are split between Github Actions and Azure Workflows. While Azure Workflows provided us with many extra features and an isolated quota, tests seem to run much slower than they would with Github Actions (_exact numbers to be examined_). This move will (_most likely_) ensure that workflows run faster increasing our overall commit throughput.

The moved workflows include:
- Pull Requests:
    - Docs and Lint
    - Python and partial rust tests.
- Commits:
    - On push
    - Nightly tests
    - Merge queue

#### New Pull Request workflow

The new PR workflow looks as such:
image
Where if both `Docs and Lint` and `Preliminary Tests` pass, then the rest of the test suite runs concurrently.
For a successful run of all workflows please refer to [this example run](https://github.com/Qiskit/qiskit/pull/13474/checks?check_run_id=39780574632) from this PR.
### Tasks
- [x] Move Lint and Docs
- [x] Move Python Tests
- [x] Organize concurrency
- [x] Add nightly tests
- [x] Merge queue tests
- [ ] ~~Run nightly matrix tests when code changes.~~

### Known issues
- [x] How to properly test merge-queue, push, nightly tests.
- [ ] ~~Main nightly tests should only run when new changes in the code are introduced (as per the previous azure workflow).~~


raynelfss:
I have yet to test the nightly, merge-queue, and push tests. But I believe they should work without issues since they're mostly reusing the same workflows we use here for Pull Requests CI. I'm going to open this up for review :)

.github/workflows/lint_docs.yml
```diff
@@ -0,0 +1,58 @@
...
+          fetch-depth: 0
+      - name: Set up Python
+        uses: actions/setup-python@v5
+        with:
+          python-version: "3.9"
+      - name: Install ubuntu docs dependencies
+        run: tools/install_ubuntu_docs_dependencies.sh
+      - name: Run docs
+        run: |
+          make cformat
```
mtreinish: This is a lint task not a docs one. It's running clang format.

.github/workflows/test-linux.yml
```diff
@@ -0,0 +1,172 @@
...
+          # This variable should become redundant on release of jupyter-core 6.
+          JUPYTER_PLATFORM_DIRS: 1
+      - name: Store image tests diff
+        uses: actions/upload-artifact@v4
+        if: ${{ inputs.test-images && failure()}}
+        with:
+          name: image-test-failure-img-diffs
+          path: |
+            ./test/visual/mpl/graph/graph_results
+          if-no-files-found: error
```
mtreinish: We should have enough job parallelism on linux to be able to split these  out into separate jobs instead of serially executed tasks. This would reduce the runtime on this job which will be the critical path for the preliminary stage.

.github/workflows/test-linux.yml
```diff
@@ -0,0 +1,172 @@
...
+          virtualenv test-job
+      - name: Install Qiskit from sdist
+        if: ${{ inputs.install-from-sdist }}
+        run: |
+          set -e
+          # Use stable Rust, rather than MSRV, to spot-check that stable builds properly.
+          rustup override set stable
+          source test-job/bin/activate
+          python -m pip install -U pip
+          python -m pip install -U build
```
mtreinish: In general it's better to bundle pip commands into a single execution so the depsolver can work with the shared dependencies.

.github/workflows/test-linux.yml
```diff
@@ -0,0 +1,172 @@
...
+          restore-keys: |
+            stestr | "${{ runner.os }}" | "${{ inputs.python-version }}"
+            stestr | "${{ runner.os }}"
+            stestr
+          path: .stestr
+      - name: Prepare venv
+        run: |
+          set -e
+          python -m pip install --upgrade pip setuptools wheel virtualenv
+          virtualenv test-job
```
mtreinish: I would drop virtualenv here and just do `python -m venv test-job`

.github/workflows/test-linux.yml
```diff
@@ -0,0 +1,172 @@
...
+    runs-on: ${{ matrix.os }}
+    env:
+      QISKIT_SUPPRESS_PACKAGING_WARNINGS: Y
+      PIP_CACHE_DIR: ${{ github.workspace }}/.pip
+      QISKIT_TEST_CAPTURE_STREAMS: 1
+      HAVE_VISUAL_TESTS_RUN: false
+    strategy:
+      fail-fast: false
+      matrix:
+        os: ["ubuntu-latest", "ubuntu-24.04-arm"]
```
mtreinish: I'm fine with this, but does this mean the preliminary stage for 3.9 is run in duplicate on arm and x86? I think that's actually fine but I want to know if that was actually what's going to happen and if that's your intent.

mtreinish (commented):
I've only gotten through the linux tests so far, but some preliminary comments.

.github/workflows/test-linux.yml
```diff
@@ -33,19 +38,17 @@ on:
   merge_group:
 
 jobs:
-  tests-linux:
+  unit-tests:
     if: github.repository_owner == 'Qiskit'
-    name: ${{ matrix.os }}-tests-Python-${{ inputs.python-version }}
-    runs-on: ${{ matrix.os }}
+    name: ubuntu-${{ inputs.system-suffix }}-tests-Python-${{ inputs.python-version }}
+    runs-on: ubuntu-${{ inputs.system-suffix }}
```
mtreinish: Instead of doing this as a suffix, I'd say change this to just be the image-type or something like that. It's not a guarantee we'll always be using ubuntu images and it's better to be explicit here.

.github/workflows/nightly.yml
```diff
@@ -0,0 +1,36 @@
...
+    # file needs to match on _both_ `main` and the branch itself.
+    if: ${{ github.repository_owner == 'Qiskit'}}
+    uses: ./.github/workflows/on-nightly.yml
+  nightly-failure:
+    name: Create comment on failed test run
+    if: ${{ github.repository_owner == 'Qiskit' && failure() }}
+    runs-on: ubuntu-latest
+    needs: [nightly-tests]
+    steps:
+      - uses: peter-evans/create-or-update-comment@v4
```
mtreinish: There's not a more official action for this?

.github/workflows/on-merge-queue.yml
```diff
@@ -0,0 +1,49 @@
...
+# flaky VMs timing out (which we can't do much about), or a merge conflict
+# with another PR that is also in the merge queue.
+#
+# There's no reason to have multiple stages in this case, because we're
+# expecting it to pass.  Having more than one stage frustrates parallel
+# throughput in low-contention cases, and guarantees a longer critical path.
+name: Merge queue
+on:
+  # remove when proven to work
+  pull_request:
```
mtreinish: Should we mark this as on hold until you remove this?

mtreinish (commented):
This is looking good. A couple of inline comments.

.github/workflows/on-merge-queue.yml
raynelfss: Agreed, so that I don't randomly forget :)

.github/workflows/nightly.yml
raynelfss: This is what was being used by other workflows that do the same. I could change it here but the other ones will need to be changed on a separate PR

.github/workflows/test-linux.yml
raynelfss: Runs as duplicate as of now.

.github/workflows/nightly.yml
raynelfss: Switched to gh CLI in 0466f92

.github/workflows/test-linux.yml
raynelfss: Added runner arguments for all workflows in 2be0b9b

.github/workflows/on-pull-request.yml
```diff
@@ -0,0 +1,95 @@
...
+concurrency:
+  group: ${{ github.repository }}-${{ github.ref }}-${{ github.head_ref }}-${{ github.workflow }}
+  cancel-in-progress: true
+jobs:
+  # The preliminary stage should be small in both total runtime (including
+  # provisioning) and resources required.  About half of PR commits result in
+  # a CI failure, and over 90% of these are in linting, documention or a test
+  # failure that would affect _any_ OS or Python version.  The goal in the
+  # first stage is to catch the vast majority of failures with minimal cost.
+  lint-docs:
```
mtreinish, line 14: We should set `timeout-minutes` for these jobs and limit them to an hour. That was our time budget on azure and something is fundamentally wrong if it takes that long for anything in these jobs and on the jobs on `push` and in the merge queue too.

.github/workflows/nightly.yml
```diff
@@ -0,0 +1,37 @@
...
+# want to be enabled, then particular branches can override it to exclude
+# themselves by removing themselves from the trigger list.  For example, old
+# stable branches can remove `stable/*` from their copy of this file once they
+# reach their end-of-life.
+name: Nightly Main tests
+on:
+  # remove when proven to work
+  # pull_request:
+  schedule:
+    - cron: "20 6 * * *"
```
mtreinish: How did you settle on this time?

.github/workflows/nightly.yml
```diff
@@ -0,0 +1,37 @@
...
+    needs: [nightly-tests]
+    steps:
+      - name: "Create comment on failed test run"
+        run: gh issue comment "$NUMBER" --body "$BODY"
+        env:
+          NUMBER: 7864
+          BODY: >
+            Randomized tests failed at commit ${{ github.sha }}.
+
+            _Logs_: ${{ github.server_url }}/${{ github.repository }}/actions/runs/${{ github.run_id }}.
```
mtreinish: Don't you need to specify the github token here?

See: https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/using-github-cli-in-workflows

.github/workflows/on-merge-queue.yml
```diff
@@ -0,0 +1,57 @@
...
+# flaky VMs timing out (which we can't do much about), or a merge conflict
+# with another PR that is also in the merge queue.
+#
+# There's no reason to have multiple stages in this case, because we're
+# expecting it to pass.  Having more than one stage frustrates parallel
+# throughput in low-contention cases, and guarantees a longer critical path.
+name: Merge queue
+on:
+  # remove when proven to work
+  # pull_request:
```
mtreinish:
```suggestion
```

.github/workflows/nightly.yml
```diff
@@ -0,0 +1,37 @@
...
+# A schedule only runs on branches that match include rules from _both_ `main`
+# and the branch itself.  On `main`, we blanket include all branches that might
+# want to be enabled, then particular branches can override it to exclude
+# themselves by removing themselves from the trigger list.  For example, old
+# stable branches can remove `stable/*` from their copy of this file once they
+# reach their end-of-life.
+name: Nightly Main tests
+on:
+  # remove when proven to work
+  # pull_request:
```
mtreinish:
```suggestion
```

.github/workflows/on-pull-request.yml
```diff
@@ -0,0 +1,95 @@
+---
+# Full PR suite.  PRs need to pass this pipeline in order to be moved to the
+# merge queue, where they'll use the next rule as the branch-protection rule
+# for the final merge to the base branch.
+name: Pull Requests
+on:
+  pull_request:
+    branches: [main, "stable/*"]
+  merge_group:
```
mtreinish: Is this is needed?

.github/workflows/on-push.yml
```diff
@@ -0,0 +1,34 @@
+---
+# Push to main or the stable branches.  The triggering branches also need to
+# be in the triggers at the top of this file.
+name: Push
+on:
+  # Remove when proven to work
+  # pull_request:
```
mtreinish:
```suggestion
```

.github/workflows/test-linux.yml
```diff
@@ -0,0 +1,264 @@
...
+        type: boolean
+        default: false
+
+      install-from-sdist:
+        description: >
+          Decides whether we perform rust tests
+        type: boolean
+        default: false
+
+  merge_group:
```
mtreinish: Is this needed?

.github/workflows/test-linux.yml
```diff
@@ -0,0 +1,264 @@
...
+      PIP_CACHE_DIR: ${{ github.workspace }}/.pip
+      QISKIT_TEST_CAPTURE_STREAMS: 1
+      HAVE_VISUAL_TESTS_RUN: false
+    strategy:
+      fail-fast: false
+    steps:
+      - uses: actions/checkout@v4
+      - name: Install Rust toolchain
+        uses: dtolnay/rust-toolchain@1.79
+        if: inputs.python-version == '3.9'
```
mtreinish: Is this needed if we have the `rustup override stable` manually called below?

.github/workflows/on-pull-request.yml
```diff
@@ -0,0 +1,95 @@
...
+    if: github.repository_owner == 'Qiskit'
+    name: Main Tests (ubuntu)
+    strategy:
+      matrix:
+        python_version: ["3.13"]
+        runner: ["ubuntu-latest", "ubuntu-24.04-arm"]
+    uses: ./.github/workflows/test-linux.yml
+    with:
+      python-version: ${{ matrix.python_version }}
+      test-rust: false
```
mtreinish: It looks like we're running the rust tests on 3.13: https://github.com/Qiskit/qiskit/actions/runs/15005241942/job/42162089199?pr=13474 which is duplicated work. We only need to run it once since the coverage is the same.

.github/workflows/nightly.yml
raynelfss: This was the same time used in Azure. Do you know if there's a way of making it only run if the code has changed? The same way `schedule/always` worked in Azure.

.github/workflows/on-pull-request.yml
raynelfss: I don't think so, this workflow should only run during pull-requests, the one for the merge queue is different and defined in `on-merge-queue`

.github/workflows/on-pull-request.yml
raynelfss: This is due to a missing if statement in the test-rust job. Fixing...

.github/workflows/nightly.yml
mtreinish: I don't see anything in the docs for that in the trigger conditions. You could probably hack it together with runtime checks and storing the last sha1 run somewhere. But I wouldn't worry about that. I was more just curious if you picked this time or just copied it. It's reasonable to run things at 6:20 UTC, activity is typically pretty slow then.

.github/workflows/test-linux.yml
raynelfss: This is only done when building from sdist, I believe we override to make sure our sdists work correctly with the latest stable version of rust.

.github/workflows/on-pull-request.yml
raynelfss: Removed in eb9cf50

.github/workflows/test-linux.yml
raynelfss: Removed in eb9cf50

.github/workflows/on-pull-request.yml
raynelfss: Fixed in eb9cf50

.github/workflows/on-push.yml
raynelfss: Removed in 708be7e

.github/workflows/nightly.yml
raynelfss: Removed in 0d98e72

.github/workflows/on-pull-request.yml
raynelfss: Added in 44b6757

.github/workflows/nightly.yml
raynelfss: Added in e09daea

.github/workflows/on-pull-request.yml
```diff
@@ -0,0 +1,94 @@
...
+  # provisioning) and resources required.  About half of PR commits result in
+  # a CI failure, and over 90% of these are in linting, documention or a test
+  # failure that would affect _any_ OS or Python version.  The goal in the
+  # first stage is to catch the vast majority of failures with minimal cost.
+  lint-docs:
+    if: github.repository_owner == 'Qiskit'
+    name: Docs and Lint
+    uses: ./.github/workflows/lint_docs.yml
+  preliminary-tests:
+    name: Preliminary Tests
```
mtreinish: If we've dropped the two stage approach and are going with maximum parallelism we should use consistent naming here. There is no preliminary or main. We also should just flatten the structure here so it's grouped by operating system/file. Right now the mix of the old and the new is confusing. Also none of the comments apply anymore.

mtreinish (commented):
I think this is pretty close. The big thing right now is the entire structure of the files is setup to document a two stage approach but because of the increased concurrency limits that no longer applies. We should rename the job definitions and remove all the stale comments. Or if you want to be conservative and keep a staged approach to start you can add back the dependency and save the flattening for a follow up PR.

.github/workflows/test-linux.yml
```diff
@@ -0,0 +1,266 @@
...
+          python -m venv image_tests
+          image_tests/bin/python -m pip install -U \
+            -c constraints.txt \
+            -r requirements.txt \
+            -r requirements-dev.txt \
+            -r requirements-optional.txt \
+            -e .
+          sudo apt-get update
+          sudo apt-get install -y graphviz pandoc
+          image_tests/bin/pip check
```
raynelfss: Is there a reason why we don't just do `source image_tests/bin/activate` here?

.github/workflows/on-push.yml
```diff
@@ -0,0 +1,32 @@
...
+    uses: ./.github/workflows/test-linux.yml
+    strategy:
+      matrix:
+        runner: ["ubuntu-latest", "ubuntu-4.04-arm"]
+    with:
+      python-version: "3.9"
+      test-rust: true
+      test-images: true
+      install-optionals: false
+      runner: ${{ matrix.runner }}
```
raynelfss: Do we want to test Windows here too?

.github/workflows/on-pull-request.yml
raynelfss: Fixed by a067d09 and 80c3268

.github/workflows/test-linux.yml
mtreinish: It's just a habit I have when writing bash scripts. I don't like calling `activate` because of how it mangles `$PS1` and also might interact with other shell command scripting.

.github/workflows/on-push.yml
mtreinish: on azure we only run the linux 3.9 job because it's duplicated effort with the merge queue. The push job is just a sanity check that the merge was fine and doesn't really provide value so we should probably keep it minimal. I'd remove macOS and the arm job from this list too.

raynelfss: Removed in ad8cd2e

.github/workflows/image-tests.yml
```diff
@@ -0,0 +1,97 @@
...
+        description: >
+          Decides whether we install optyional dependencies
+        type: boolean
+        default: false
+
+      install-from-sdist:
+        description: >
+          Decides whether we perform rust tests
+        type: boolean
+        default: false
```
mtreinish: Does this need to be an option here? Like we only run this once so it seems unecessary to make this switchable, just pick sdist or normal install and we can simplify the definition here.

.github/workflows/image-tests.yml
```diff
@@ -0,0 +1,97 @@
...
+        description: >
+          Describes the system this workflow should run on.
+        type: string
+        required: true
+
+      install-optionals:
+        description: >
+          Decides whether we install optyional dependencies
+        type: boolean
+        default: false
```
mtreinish: We always need the optionals here because we're testing visualization. This shouldn't be an option and just default to true.

.github/workflows/image-tests.yml
```diff
@@ -0,0 +1,97 @@
...
+      PIP_CACHE_DIR: ${{ github.workspace }}/.pip
+      QISKIT_TEST_CAPTURE_STREAMS: 1
+      HAVE_VISUAL_TESTS_RUN: false
+    strategy:
+      fail-fast: false
+    steps:
+      - uses: actions/checkout@v4
+      - name: Install Rust toolchain
+        uses: dtolnay/rust-toolchain@1.79
+        if: inputs.python-version == '3.9'
```
mtreinish: I feel like building with the MSRV or not doesn't give us any value for the image tests we should just pick either stable or msrv and not make this an option. We should keep the python version config just so the config is generic for when we bump the MSPV after 3.9 goes EoL.

.github/workflows/image-tests.yml
```diff
@@ -0,0 +1,97 @@
...
+jobs:
+  image-tests:
+    name: Test-images-${{ inputs.python-version }}
+    runs-on: ${{ inputs.runner }}
+    timeout-minutes: 60
+    env:
+      QISKIT_SUPPRESS_PACKAGING_WARNINGS: Y
+      PIP_CACHE_DIR: ${{ github.workspace }}/.pip
+      QISKIT_TEST_CAPTURE_STREAMS: 1
+      HAVE_VISUAL_TESTS_RUN: false
```
mtreinish: What is this for?

.github/workflows/image-tests.yml
```diff
@@ -0,0 +1,97 @@
...
+        run: |
+          set -e
+          source test-job/bin/activate
+          python -m pip install -U \
+            -c constraints.txt \
+            -r requirements.txt \
+            -r requirements-dev.txt \
+            -e .
+          # Build and install both qiskit and qiskit-terra so that any optionals
+          # depending on `qiskit` will resolve correctly.
```
mtreinish: See my above comment I would just delete this

.github/workflows/on-merge-queue.yml
```diff
@@ -0,0 +1,67 @@
...
+      install-optionals: ${{ matrix.python-version == '3.9' }}
+      runner: ${{ matrix.runner }}
+  test-rust:
+    if: github.repository_owner == 'Qiskit'
+    uses: ./.github/workflows/rust-tests.yml
+    with:
+      python-version: "3.9"
+      runner: "ubuntu-latest"
+      install-optionals: true
+  test-images:
```
mtreinish, line 52: Not for this PR since we shouldn't change the pipeline makeup like this. But I'm not sure it's worth running the image tests in the merge queue. But we should revisit this after the migration is complete.

.github/workflows/rust-tests.yml
```diff
@@ -0,0 +1,90 @@
...
+      runner:
+        description: >
+          Describes the system this workflow should run on.
+        type: string
+        required: true
+      install-optionals:
+        description: >
+          Decides whether we install optyional dependencies
+        type: boolean
+        default: false
```
mtreinish: This will always be false we don't need to waste time installing unused dependencies in the rust tests.

.github/workflows/rust-tests.yml
```diff
@@ -0,0 +1,90 @@
...
+
+      install-from-sdist:
+        description: >
+          Decides whether we perform rust tests
+        type: boolean
+        default: false
```
mtreinish: This should also be false we don't need this to be selectable we're not testing python here or the python build system this is for running rust unit tests.

.github/workflows/rust-tests.yml
```diff
@@ -0,0 +1,90 @@
...
+      - name: Set up Python ${{ inputs.python-version }}
+        id: python-outputs
+        uses: actions/setup-python@v5
+        with:
+          python-version: ${{ inputs.python-version }}
+      - name: Prepare venv
+        run: |
+          set -e
+          python -m pip install --upgrade pip setuptools wheel
+          python -m venv test-job
```
mtreinish: Do we need to do this here? We're not testing python packaging we only install qiskit to cover some edge cases in our rust testing where python is still in the data model. But besides running `pip install -e .` we shouldn't need anything else.

.github/workflows/rust-tests.yml
```diff
@@ -0,0 +1,90 @@
...
+          # depending on `qiskit` will resolve correctly.
+      - name: Install Qiskit directly
+        if: ${{ !inputs.install-from-sdist }}
+        run: |
+          set -e
+          source test-job/bin/activate
+          python -m pip install -U \
+              -c constraints.txt \
+              -r requirements.txt \
+              -r requirements-dev.txt \
```
mtreinish: You can just do `pip install -c constraints.txt -e .` here We don't need optionals or dev dependencies here we only need to install qiskit so something trying to import qiskit in rust code doesn't error.

.github/workflows/image-tests.yml
raynelfss: I'll go with normal here

.github/workflows/image-tests.yml
raynelfss: These flags are mostly no longer used, it seems. Tox will already enable `QISKIT_SUPPRESS_PACKAGING_WARNINGS` and `QISKIT_TEST_CAPTURE_STREAMS` to the same values here. The only reason to keep them, I assume, would be with anything that runs with stestr, unless we don't care about these variables anymore (one of which is mentioned in our [readme](https://github.com/Qiskit/qiskit/blob/main/CONTRIBUTING.md#stdoutstderr-and-logging-capture)).

If anything we should keep the `QISKIT_TEST_CAPTURE_STREAMS` whenever we test in Python so we can capture errors. The rest of them I'm not quite sure.

.github/workflows/test-linux.yml
```diff
@@ -0,0 +1,122 @@
...
+          set -e
+          source test-job/bin/activate
+          python -m pip install -r requirements-optional.txt -c constraints.txt
+          python -m pip check
+        if: ${{ inputs.install-optionals }}
+      - name: Install optional non-Python dependencies
+        run: |
+          set -e
+          sudo apt-get update
+          sudo apt-get install -y graphviz
```
raynelfss, line 90: Do we still need these here? I feel like they can be removed.

mtreinish: I think we have regular unit tests that run graphviz for visualization testing so we will need this on jobs that run with optionals enabled.

.github/workflows/image-tests.yml
raynelfss: Switched to stable

.github/workflows/image-tests.yml
raynelfss: Removed all env variables except for `QISKIT_TEST_CAPTURE_STREAMS` on main tests.

mtreinish:
There are syntax errors in the current version:

https://github.com/Qiskit/qiskit/actions/runs/15122611633
https://github.com/Qiskit/qiskit/actions/runs/15122611426

raynelfss:
> There are syntax errors in the current version:
> 
> https://github.com/Qiskit/qiskit/actions/runs/15122611633 https://github.com/Qiskit/qiskit/actions/runs/15122611426

Fixed in 768694d

.github/workflows/image-tests.yml
```diff
@@ -0,0 +1,61 @@
...
+  image-tests:
+    name: Test-images-${{ inputs.python-version }}
+    runs-on: ${{ inputs.runner }}
+    timeout-minutes: 60
+    strategy:
+      fail-fast: false
+    steps:
+      - uses: actions/checkout@v4
+      - name: Install Rust toolchain
+        uses: dtolnay/rust-toolchain@stable
```
mtreinish: This doesn't do anything in practice but will expose another failure point. The `rust-toolchain.toml` will override the configured default version when `cargo` is called from the repo root since rustup is installed in the runner images. This will result in rust rustup being run twice here installing stable and then the pip call at line 42 that will call rustup internally to install the MSRV from the `rust-toolchain.toml`

```suggestion
```

.github/workflows/lint_docs.yml
```diff
@@ -0,0 +1,60 @@
...
+      - uses: actions/checkout@v4
+        with:
+          # We need to fetch the whole history so 'reno' can do its job and we can inspect tags.
+          fetch-depth: 0
+      - name: Set up Python
+        uses: actions/setup-python@v5
+        with:
+          python-version: "3.9"
+      - name: Install ubuntu docs dependencies
+        run: tools/install_ubuntu_docs_dependencies.sh
```
mtreinish: Do we need this for the lint job?

.github/workflows/nightly.yml
```diff
@@ -0,0 +1,37 @@
+---
+# A schedule only runs on branches that match include rules from _both_ `main`
+# and the branch itself.  On `main`, we blanket include all branches that might
+# want to be enabled, then particular branches can override it to exclude
+# themselves by removing themselves from the trigger list.  For example, old
+# stable branches can remove `stable/*` from their copy of this file once they
+# reach their end-of-life.
```
mtreinish: Is this true? IIRC the schedule trigger only runs on the default branch.

.github/workflows/nightly.yml
```diff
@@ -0,0 +1,37 @@
...
+  workflow_dispatch:
+
+jobs:
+  nightly-tests:
+    # Nightly cron job.
+    #
+    # For this to run on a branch, the `schedules` trigger  up at the top of this
+    # file needs to match on _both_ `main` and the branch itself.
+    if: ${{ github.repository_owner == 'Qiskit'}}
+    uses: ./.github/workflows/on-nightly.yml
```
mtreinish: Why does this need to be a separate file. I can understand on the other triggers where there is a clear divide between multiple job types. But for this it just seems like unnecessary layer because this is already defined in using the split job definitions. I would flatten this so there is only one nightly file to limit how much jumping around you need to do to understand the configs.

.github/workflows/nightly.yml
```diff
@@ -0,0 +1,37 @@
...
+  schedule:
+    - cron: "20 6 * * *"
+  workflow_dispatch:
+
+jobs:
+  nightly-tests:
+    # Nightly cron job.
+    #
+    # For this to run on a branch, the `schedules` trigger  up at the top of this
+    # file needs to match on _both_ `main` and the branch itself.
```
mtreinish: What does this mean?

.github/workflows/on-pull-request.yml
```diff
@@ -0,0 +1,90 @@
...
+    name: Main Tests
+    uses: ./.github/workflows/test-linux.yml
+    if: github.repository_owner == 'Qiskit'
+    with:
+      python-version: "3.9"
+      install-optionals: true
+      runner: ubuntu-latest
+      use-msrv: true
+  main-tests-rust:
+    name: Extra Tests
```
mtreinish: What does extra tests mean? These seem like all part of the main workflow. Do you mean non-python unit test jobs? I think that should be the distinction then. But I think instead of just saying extra tests here "Rust unit tests" and "Image comparison tests" for the names here would be better.

.github/workflows/on-pull-request.yml
```diff
@@ -0,0 +1,90 @@
...
+concurrency:
+  group: ${{ github.repository }}-${{ github.ref }}-${{ github.head_ref }}-${{ github.workflow }}
+  cancel-in-progress: true
+jobs:
+  lint-docs:
+    if: github.repository_owner == 'Qiskit'
+    name: Docs and Lint
+    uses: ./.github/workflows/lint_docs.yml
+  main-tests-ubuntu-3-9:
+    name: Main Tests
```
mtreinish: Maybe "Python unit tests" instead?

.github/workflows/test-linux.yml
```diff
@@ -0,0 +1,125 @@
...
+    timeout-minutes: 60
+    env:
+      QISKIT_TEST_CAPTURE_STREAMS: 1
+    strategy:
+      fail-fast: false
+    steps:
+      - uses: actions/checkout@v4
+      - name: Install Rust toolchain
+        uses: dtolnay/rust-toolchain@1.79
+        if: ${{ inputs.use-msrv }}
```
mtreinish: This is actually just likely to get out of sync, if `use-msrv` is set to true you're potentially installing rust twice if install from sdist is used and not actually running with a different rust versions. You would need to also set the appropriate override when calling cargo (via pip) if you wanted `false` case to be respected fully.

I would just remove this flag and option from the test jobs. This is different from the azure config as that always used MSRV and never ran with Rust stable (it did over a year ago but we moved away from that). We can discuss doing this in the future, but we shouldn't make changes like this as part of a migration PR.

.github/workflows/test-linux.yml
```diff
@@ -0,0 +1,125 @@
...
+        run: |
+          set -e
+          python -m pip install --upgrade pip setuptools wheel
+          python -m venv test-job
+      - name: Install Qiskit from sdist
+        if: ${{ inputs.install-from-sdist }}
+        run: |
+          set -e
+          # Use stable Rust, rather than MSRV, to spot-check that stable builds properly.
+          rustup override set stable
```
mtreinish: If we were doing this (which we shouldn't in this PR) wouldn't this need to be conditioned on `use-msrv` not install from sdist?

.github/workflows/test-mac.yml
```diff
@@ -0,0 +1,99 @@
...
+          Describes the system this workflow should run on.
+        type: string
+        required: true
+      
+      use-msrv:
+        description: >
+          Decides whether we use MSRV or stable for rust toolchain.
+        required: false
+        type: boolean
+        default: false
```
mtreinish: Same comment from linux remove this option and always use msrv.

.github/workflows/test-windows.yml
```diff
@@ -0,0 +1,100 @@
...
+          Decides whether we install optyional dependencies
+        type: boolean
+        default: false
+
+      use-msrv:
+        description: >
+          Decides whether we use MSRV or stable for rust toolchain.
+        required: false
+        type: boolean
+        default: false
```
mtreinish: Same here remove this option and always use msrv.

.github/workflows/image-tests.yml
raynelfss: Removed in 8d54c24

.github/workflows/test-linux.yml
raynelfss: Removed in 8d54c24

.github/workflows/test-linux.yml
raynelfss: Removed in 8d54c24

.github/workflows/test-mac.yml
raynelfss: Removed in 8d54c24

.github/workflows/test-windows.yml
raynelfss: Removed in 8d54c24

.github/workflows/lint_docs.yml
raynelfss: Removed in b7f717b

.github/workflows/on-pull-request.yml
raynelfss: Changed to "Rust Unit Tests" and "Image Comparison Tests" in b7f717b

.github/workflows/on-pull-request.yml
raynelfss: Changed in b7f717b

.github/workflows/nightly.yml
raynelfss: This was an old comment, this behavior probably works in Azure but not in GH Actions as it should only run scheduled workflows on the main branch. Removed in 4655adc.

.github/workflows/nightly.yml
raynelfss: Removed "nightly.yml" in favor of "on-nightly" in 4655adc, also updated that comment.

.github/workflows/nightly.yml
raynelfss: More old comments that were removed by 4655adc. Many comments were brought over thinking behavior would be similar in GH Actions, but that was not the case :)

.github/workflows/image-tests.yml
```diff
@@ -0,0 +1,59 @@
...
+          # a new internal interface that will be the default in jupyter-core 6.x.
+          # This variable should become redundant on release of jupyter-core 6.
+          JUPYTER_PLATFORM_DIRS: 1
+      - name: Store image tests diff
+        uses: actions/upload-artifact@v4
+        if: ${{ failure() }}
+        with:
+          name: image-test-failure-img-diffs
+          path: |
+            ./test/visual/mpl/graph/graph_results
```
mtreinish, line 57: This is missing results, it needs to include the circuit results directory too. There are two sets of tests that write out the results to different paths.

.github/workflows/lint_docs.yml
```diff
@@ -0,0 +1,60 @@
...
+          fetch-depth: 0
+      - name: Set up Python
+        uses: actions/setup-python@v5
+        with:
+          python-version: "3.9"
+      - name: Install dependencies
+        run: python -m pip install --upgrade tox
+      - name: Run lint
+        run: |
+          make cformat
```
mtreinish, line 59: For those following along in azure we ran a script `tools/install_ubuntu_c_dependencies.sh` that installs clang format which is missing here. But it doesn't matter because the gha runner image already includes clang format: https://github.com/actions/runner-images/blob/main/images/ubuntu/Ubuntu2404-Readme.md#language-and-runtime which is the same version. We can probably delete that bash script but lets save that for a followup.

.github/workflows/test-mac.yml
```diff
@@ -0,0 +1,88 @@
...
+      QISKIT_TEST_CAPTURE_STREAMS: 1
+    strategy:
+      fail-fast: false
+    steps:
+      - uses: actions/checkout@v4
+      - name: Set up Python ${{ inputs.python-version }}
+        uses: actions/setup-python@v5
+        with:
+          python-version: ${{ inputs.python-version }}
+          architecture: arm64
```
mtreinish: Is this correct? Shouldn't we only use the arm64 version of python on the macOS-14 runner?

.github/workflows/test-mac.yml
```diff
@@ -0,0 +1,88 @@
...
+          python -m venv test-job
+          source test-job/bin/activate
+          python -m pip install -U pip setuptools wheel
+          python -m pip install -U \
+            -c constraints.txt \
+            -r requirements.txt \
+            -r requirements-dev.txt \
+            -e .
+            # Build and install both qiskit and qiskit-terra so that any optionals
+            # depending on `qiskit` will resolve correctly.
```
mtreinish:
```suggestion
```

.github/workflows/test-windows.yml
```diff
@@ -0,0 +1,89 @@
...
+          python -m venv test-job
+          .\test-job\Scripts\activate
+          python -m pip install -U pip setuptools wheel
+          python -m pip install -U `
+            -c constraints.txt `
+            -r requirements.txt `
+            -r requirements-dev.txt `
+            -e .
+          # Build and install both qiskit and qiskit-terra so that any optionals
+          # depending on `qiskit` will resolve correctly.
```
mtreinish:
```suggestion
```

.github/workflows/test-linux.yml
```diff
@@ -0,0 +1,114 @@
...
+        run: |
+          set -e
+          source test-job/bin/activate
+          python -m pip install -U \
+            -c constraints.txt \
+            -r requirements.txt \
+            -r requirements-dev.txt \
+            -e .
+          # Build and install both qiskit and qiskit-terra so that any optionals
+          # depending on `qiskit` will resolve correctly.
```
mtreinish:
```suggestion
```

.github/workflows/test-linux.yml
```diff
@@ -0,0 +1,114 @@
...
+          source test-job/bin/activate
+          python -m pip install -U pip build
+          python -m build --sdist .
+          python -m pip install -U \
+            -c constraints.txt \
+            -r requirements.txt \
+            -r requirements-dev.txt \
+            dist/qiskit-*.tar.gz
+          # Build and install both qiskit and qiskit-terra so that any optionals
+          # depending on `qiskit` will resolve correctly.
```
mtreinish:
```suggestion
```

.github/workflows/test-mac.yml
raynelfss: You are correct! Do I make it do so conditionally? Like e.g.
```yaml
          # This may not wokr
          architecture: ${{ inputs.runner == "macos-14" && "arm64" ||  "x64" }}
```

mtreinish: Does it work out of the box if you just remove this flag? Like I feel like the action should just use the native architecture if you don't specify anything.

raynelfss: I don't see any reason why it shouldn't work. It defaults to the architecture of the runner as per [documentation](https://github.com/actions/setup-python/tree/v5/?tab=readme-ov-file#supported-architectures).

.github/workflows/image-tests.yml
raynelfss: Fixed in c1f1a8a

.github/workflows/test-mac.yml
raynelfss: Fixed in 4dc9eae

.github/workflows/test-mac.yml
raynelfss: Removed in 9ee9683

.github/workflows/test-windows.yml
raynelfss: Removed in 9ee9683

.github/workflows/test-linux.yml
raynelfss: Removed in 9ee9683

.github/workflows/test-linux.yml
raynelfss: Removed in 9ee9683

.github/workflows/on-merge-queue.yml
```diff
@@ -0,0 +1,65 @@
...
+# There's no reason to have multiple stages in this case, because we're
+# expecting it to pass.  Having more than one stage frustrates parallel
+# throughput in low-contention cases, and guarantees a longer critical path.
+name: Merge queue
+on:
+  merge_group:
+    branches: [main, "stable/*"]
+concurrency:
+  group: ${{ github.repository }}-${{ github.ref }}-${{ github.head_ref }}-${{ github.workflow }}
+  cancel-in-progress: true
```
mtreinish, line 16: This doesn't hurt, but it really shouldn't do anything because we should ever have 2 groups at the same time for commits that have entered the merge queue. If someone pushes a new commit to a enqueued PR it wouldn't be mergeable anymore and a new merge queue job wouldn't start. Lets just leave it though because it doesn't hurt to have this protection in place here too.

.github/workflows/on-merge-queue.yml
```diff
@@ -0,0 +1,65 @@
...
+# flaky VMs timing out (which we can't do much about), or a merge conflict
+# with another PR that is also in the merge queue.
+#
+# There's no reason to have multiple stages in this case, because we're
+# expecting it to pass.  Having more than one stage frustrates parallel
+# throughput in low-contention cases, and guarantees a longer critical path.
+name: Merge queue
+on:
+  merge_group:
+    branches: [main, "stable/*"]
```
mtreinish, line 13: I'm wondering what's going to happen after we merge this. We probably aren't going to backport this PR and we'll still use azure on stable branches until they age out (`stable/2.0` goes EoL in about a month). We might want to backport it to `stable/1.4` since that's supported longer. I'm wonder if we will need to remove this until we are using gha on all the supported stable branches.

.github/workflows/on-merge-queue.yml
```diff
@@ -0,0 +1,65 @@
...
+    with:
+      python-version: "3.13"
+      install-optionals: false
+      runner: ${{ matrix.runner }}
+  test-windows:
+    if: github.repository_owner == 'Qiskit'
+    uses: ./.github/workflows/test-windows.yml
+    strategy:
+      matrix:
+        runner: ["windows-latest"]
```
mtreinish: Using a matrix here seems a bit unnecessary since there isn't more than 1 configuration but it's fine for now. I can see a world where we add windows arm support in several years so lets leave the flexibility for the future.

.github/workflows/on-pull-request.yml
```diff
@@ -0,0 +1,88 @@
+---
+# Full PR suite.  PRs need to pass this pipeline in order to be moved to the
+# merge queue, where they'll use the next rule as the branch-protection rule
+# for the final merge to the base branch.
```
mtreinish:
```suggestion
# Full PR suite. PRs need to pass this pipeline in order to be moved to the
# merge queue, where they'll use the on-merge-queue workflow for the final
# merge to the base branch.
```

.github/workflows/on-pull-request.yml
```diff
@@ -0,0 +1,88 @@
...
+      runner: ubuntu-latest
+  # The rest of the PR pipeline is to test the oldest and newest supported
+  # versions of Python.
+  main-tests-ubuntu-3-9-arm:
+    if: github.repository_owner == 'Qiskit'
+    name: Python Unit Tests
+    uses: ./.github/workflows/test-linux.yml
+    with:
+      python-version: "3.9"
+      install-from-sdist: true
```
mtreinish: Should this be false to match the other 3.9 job?

.github/workflows/on-pull-request.yml
```diff
@@ -0,0 +1,88 @@
...
+      install-optionals: false
```
mtreinish: Shouldn't this be true to match the other 3.9 job? Or do the optional packages not all work on aarch64? (I vaguely remember that might have been the case).

.github/workflows/on-push.yml
```diff
@@ -0,0 +1,28 @@
+---
+# Push to main or the stable branches.  The triggering branches also need to
+# be in the triggers at the top of this file.
```
mtreinish: This is the top of the file...
```suggestion
# Run on push to main or the stable branches.
```

.github/workflows/rust-tests.yml
```diff
@@ -0,0 +1,45 @@
...
+jobs:
+  unit-tests:
+    name: Tests-Rust-${{ inputs.python-version }}
+    runs-on: ${{ inputs.runner }}
+    timeout-minutes: 60
+    env:
+      QISKIT_SUPPRESS_PACKAGING_WARNINGS: Y
+      PIP_CACHE_DIR: ${{ github.workspace }}/.pip
+      QISKIT_TEST_CAPTURE_STREAMS: 1
+      HAVE_VISUAL_TESTS_RUN: false
```
mtreinish: None of these will have any effect here
```suggestion
```

.github/workflows/rust-tests.yml
```diff
@@ -0,0 +1,45 @@
...
+        uses: actions/setup-python@v5
+        with:
+          python-version: ${{ inputs.python-version }}
+      - name: Install Qiskit directly
+        run: |
+          set -e
+          python -m pip install -c constraints.txt -e .
+      - name: Run Rust Tests
+        # Note that we use the virtual env with Qiskit installed to run the Rust
+        # tests since some of them still depend on Qiskit's Python API via PyO3.
```
mtreinish: There is no virtual environment used in this job.
```suggestion
```

.github/workflows/test-linux.yml
```diff
@@ -0,0 +1,110 @@
...
+          python tools/report_numpy_state.py
+          mkdir -p /tmp/terra-tests
+          cp -r test /tmp/terra-tests/.
+          cp .stestr.conf /tmp/terra-tests/.
+          cp -r .stestr /tmp/terra-tests/. || :
+          pushd /tmp/terra-tests
+          export PYTHONHASHSEED=$(python -S -c "import random; print(random.randint(1, 4294967295))")
+          echo "PYTHONHASHSEED=$PYTHONHASHSEED"
+          stestr run
+          popd
```
mtreinish, line 126: This is missing a filter step post-run that is used to the filter the stestr history. This is what: https://github.com/Qiskit/qiskit/blob/main/.azure/test-linux.yml#L135-L137 does. We'll need this on all platform jobs if we're caching the stestr run times for better scheduling.

Without this you'll have continual growth of the stestr history because the run history will be incremented on each run growing to use all the cache space and being very slow to upload and download.

raynelfss: Would we also need https://github.com/Qiskit/qiskit/blob/22367fca2dfa7cc22d5b4167b74279697f838f96/.azure/test-linux.yml#L138-L140 (except line 139) here too?

mtreinish: Yes, you will need the copy so the cache action has the stestr timing data to upload. This copy is only needed when you're not running in the repo like this job is. In the other OSes you don't need the copy because .stestr in the repo root is where you have the cache action pointing to.

mtreinish: Oh actually that raises a question why are we running the tests in `/tmp/terra-tests` unconditionally. I would expect this to only be for the sdist job where we want to test the install from a package installation and avoid issues with the CWD getting imported by mistake.

.github/workflows/on-merge-queue.yml
raynelfss: Removed in 75eb49c

.github/workflows/rust-tests.yml
raynelfss: Removed in fb635da

.github/workflows/rust-tests.yml
raynelfss: Removed in fb635da

.github/workflows/on-pull-request.yml
raynelfss: Fixed in 3a00da3

.github/workflows/on-pull-request.yml
raynelfss: Fixed in 3a00da3, merged into a matrix job

.github/workflows/on-pull-request.yml
raynelfss: Fixed in 3a00da3, merged into a matrix job

.github/workflows/test-linux.yml
raynelfss: Tried to fix in 1e0466d

.github/workflows/image-tests.yml
```diff
@@ -0,0 +1,60 @@
...
+            ./test/visual/mpl/circuit/circuit_results
```
mtreinish, line 58: This is also missing: `test/visual/mpl/visual_test_failures`

.github/workflows/test-linux.yml
```diff
@@ -0,0 +1,146 @@
...
+    runs-on: ${{ inputs.runner }}
+    timeout-minutes: 60
+    env:
+      QISKIT_TEST_CAPTURE_STREAMS: 1
+    strategy:
+      fail-fast: false
+    steps:
+      - uses: actions/checkout@v4
+      - name: Set up Python ${{ inputs.python-version }}
+        id: python-outputs
```
mtreinish: What is this used for?

.github/workflows/test-mac.yml
```diff
@@ -0,0 +1,91 @@
...
+            -c constraints.txt \
+            -r requirements.txt \
+            -r requirements-dev.txt \
+            -e .
+          pip check
+      - name: "Install optionals"
+        run: |
+          set -e
+          source test-job/bin/activate
+          pip install -r requirements-optional.txt -c constraints.txt
```
mtreinish, line 65: Should you `pip check` here instead of above?

.github/workflows/test-windows.yml
```diff
@@ -0,0 +1,92 @@
...
+            stestr | "${{ runner.os }}"
+            stestr
+          path: .stestr
+      - name: Install dependencies
+        run: |
+          python -m pip install --upgrade pip setuptools wheel
+          python -m venv test-job
+          .\test-job\Scripts\activate
+          python -m pip install -U pip setuptools wheel
+          python -m pip install -U `
```
mtreinish, line 54: I guess \` is the right syntax on windows cmd/powershell for the bash equivalent of `\`? I didn't realize that

.github/workflows/test-windows.yml
```diff
@@ -0,0 +1,92 @@
...
+            -c constraints.txt `
+            -r requirements.txt `
+            -r requirements-dev.txt `
+            -e .
+          pip check
```
mtreinish, line 59: This probably isn't needed here because it's a single pip command. This was originally needed before pip added a new dependency solver to validate your weren't relying on implicit behavior that would break. It's good to run on two successive pip calls still to check that you didn't break a dependency from the first install with the second (because the depsolver is only invoked per session and not shared across multiple calls).

.github/workflows/test-linux.yml
raynelfss: Nothing :) Removed in 50b4883

.github/workflows/test-mac.yml
raynelfss: Added to the line after in cf2b8d3

.github/workflows/test-windows.yml
raynelfss: Yeah, apparently this is the powershell syntax for continued line [(source).](https://devblogs.microsoft.com/scripting/powertip-line-continuation-in-powershell/)

mtreinish (dismissed):
I think this LGTM now, lets wait for the test PR to pass the jobs to make sure and then we can make the migration.

mtreinish (approved):

PR #13475, opened 2024-11-21
Do not unroll gates in basis in `add_control`

### LINKED ISSUE #13473: The multi-controlled gate by RZGate().control() is less optimized.
### Environment

- **Qiskit version**: 1.3.0rc2
- **Python version**: 3.12.6
- **Operating system**: Sonoma 14.5

### What is happening?

The multi-controlled gate by RZGate().control() is less optimized.

We create a circuit with a 6-qubit multi-controlled RZGate.  We then transpile it to an Eagle device, and count the number of entangling gates.

```
backend = service.backend("ibm_sherbrooke")
pm = generate_preset_pass_manager(optimization_level=3, backend=backend, seed_transpiler=12345)

import numpy as np
from qiskit import QuantumCircuit, QuantumRegister
from qiskit.circuit.library import RZGate
n=6;  ang=np.pi/7
qr = QuantumRegister(n)
qc = QuantumCircuit(qr)
qc.append(RZGate(ang).control(n-1), qr)

isa_qc=pm.run(qc)
print(isa_qc.count_ops()['ecr'])
```

### What should happen?

The above code outputs 266 in 1.3.0rc2, while 65 in 1.2.4.

### Any suggestions?

_No response_

Cryoris:
### Summary

The `add_control` mechanism currently does some redundant unrolling and re-wrapping of gates that are already in the basis of gates we know how to control. This has knock-on effects and causes #13473, by translating an RZ operation (already supported) to a Phase gate. 

This PR avoids this unrolling. As discussed with @alexanderivrii, in the longer term we should consider rewriting this logic entirely and have a controlled gate plugin, which also allows users to supply their own control mechanism.

### Details and comments

- [x] There's a failing test (likely since the decomposition now uses the RX-path instead of wrongly the RX->R->U3->U path before), but the unitary is still correct and there are only dangling `U(0,0,0)` gates.
- [x] Need to add a test for #13473


qiskit/circuit/add_control.py
```diff
@@ -101,22 +101,32 @@ def control(
...
-    ):
-        controlled_circ.mcx(q_control[:] + q_target[:-1], q_target[-1], q_ancillae)
-        if operation.definition is not None and operation.definition.global_phase:
+
+    basis = ["p", "u", "x", "z", "rx", "ry", "rz", "cx"]
+
+    if operation.name in basis:
+        apply_basic_controlled_gate(controlled_circ, operation, q_control, q_target[0])
+
+        # TODO: double check if this is necessary
```
ElePT: Is this todo completed? if not, how could we check if this is necessary?

test/python/circuit/test_controlled_gate.py
```diff
@@ -1560,6 +1561,9 @@ def test_single_controlled_rotation_gates(self, gate, cgate):
             op_mat = (np.cos(0.5 * self.theta) * iden - 1j * np.sin(0.5 * self.theta) * zgen).data
         else:
             op_mat = Operator(gate).data
+
+        print(self.ugu1)
+        print(cgate)
```
ElePT: sneaky prints!

```suggestion
```

test/python/circuit/test_controlled_gate.py
```diff
@@ -1577,7 +1581,7 @@ def test_single_controlled_rotation_gates(self, gate, cgate):
         elif gate.name == "rz":
             self.assertLessEqual(uqc.size(), 43, f"\n{uqc}")
         else:
-            self.assertLessEqual(uqc.size(), 20, f"\n{uqc}")
+            self.assertLessEqual(uqc.size(), 23, f"\n{uqc}")
```
ElePT: Why does the size increase in this test? shouldn't the decomposition be more efficient? (I admit I haven't looked in depth so I might be missing something)

test/python/circuit/test_controlled_gate.py
Cryoris: heh classic -- sorry about this one, I think I should add a pre-commit hook to check for these 😂

qiskit/circuit/add_control.py
Cryoris: Not yet: I was wondering whether this was necessary because here we don't decompose the gate which we want to control and shouldn't be able to accumulate any global phase. I left it for the first version since the previous code had it, but I'll try removing and see if everything passes 🙂

test/python/circuit/test_controlled_gate.py
```diff
@@ -1595,6 +1599,25 @@ def test_composite(self):
         self.log.info("%s gate count: %d", uqc.name, uqc.size())
         self.assertLessEqual(uqc.size(), 96, f"\n{uqc}")  # this limit could be changed
 
+    def test_mcrz_complexity(self):
+        """Test MCRZ is decomposed using the efficient MC-SU(2) algorithm.
+
+        Regression test of #13427.
```
ShellyGarion: what is 13427 ?

test/python/circuit/test_controlled_gate.py
```diff
@@ -1595,6 +1599,25 @@ def test_composite(self):
...
+        """
+        basis_gates = ["sx", "x", "rz", "ecr"]
```
ShellyGarion, line 1606: since rx / ry / rz are synthesized to optimize CX gates, perhaps the test should be with "cx" and not "ecr"

test/python/circuit/test_controlled_gate.py
```diff
@@ -1595,6 +1599,25 @@ def test_composite(self):
...
+        )
+
+        num_qubits = 6
+        angle = np.pi / 7
+        qc = QuantumCircuit(num_qubits)
+        qc.append(RZGate(angle).control(num_qubits - 1), qc.qubits)
+
+        isa_qc = pm.run(qc)
+
+        self.assertLessEqual(isa_qc.count_ops().get("ecr", 0), 65)
```
ShellyGarion: Any reason that the bound here is 65? for mcrz this bound is in fact 40!

test/python/circuit/test_controlled_gate.py
ShellyGarion: should we use size? perhaps it's better to count cx-count and u-count desperately, and check that cx-count does not increase

ShellyGarion: btw, is there some randomization in this test that can explain this difference?

test/python/circuit/test_controlled_gate.py
Cryoris: That's a typo -- this should be https://github.com/Qiskit/qiskit/issues/13473 🙂

test/python/circuit/test_controlled_gate.py
Cryoris: I used this specific basis set to mimic the code in https://github.com/Qiskit/qiskit/issues/13473, which uses `ibm_sherbrooke`'s basis gate set with `ecr` 🙂 I can update the `test_single_controlled_rotation_gates` test to check for the CX count though, as you suggested above.

test/python/circuit/test_controlled_gate.py
Cryoris: Hm maybe it was 65 before the fix, but you're right: it is 40. I'll update!

test/python/circuit/test_controlled_gate.py
Cryoris: The 3 additional gates were `U(0,0,0)` gates, it's not very clear to me why that would've changed, most likely a different decomposition path.. but @ShellyGarion's suggestion sounds good, what we care about are the CX gates, I'll update the test to check that.

test/python/circuit/test_controlled_gate.py
Cryoris: Fixed in ffbb955c3

test/python/circuit/test_controlled_gate.py
Cryoris: done in ffbb955c3

test/python/circuit/test_controlled_gate.py
Cryoris: Fixed in ffbb955c3

test/python/circuit/test_controlled_gate.py
Cryoris: done in ffbb955c3

test/python/circuit/test_controlled_gate.py
Cryoris: removed in ffbb955c3

qiskit/circuit/add_control.py
Cryoris: I double checked and it is not necessary, since we're not decomposing, which could add a global phase, but directly implement the controlled gate. This is removed in ffbb955c3.

qiskit/circuit/add_control.py
```diff
@@ -101,22 +101,23 @@ def control(
...
+
     global_phase = 0
-    if operation.name == "x" or (
-        isinstance(operation, controlledgate.ControlledGate) and operation.base_gate.name == "x"
-    ):
-        controlled_circ.mcx(q_control[:] + q_target[:-1], q_target[-1], q_ancillae)
-        if operation.definition is not None and operation.definition.global_phase:
-            global_phase += operation.definition.global_phase
+
+    basis = ["p", "u", "x", "z", "rx", "ry", "rz", "cx"]
```
ShellyGarion, line 112: Any reason for this fixed list of gates? what about other gates like: 
y / s / t / sx / cz / cp ?

ShellyGarion: for example, it seems that mcy is not as efficiently decomposed as mcx or mcz, see https://github.com/Qiskit/qiskit/issues/13514

test/python/circuit/test_controlled_gate.py
ShellyGarion: I think the original issues had a CX count of 65, since it was transpiling for the connectivity map of the device, but for all-to-all connectivity the CX-count is lower

qiskit/circuit/add_control.py
Cryoris: This was the original list in the code, this PR ensures we don't unroll unnecessarily but doesn't change _what_ we unroll. For a bugfix PR I'd keep it this way, but we should add the cases above for `main`!

qiskit/circuit/add_control.py
```diff
@@ -250,15 +182,101 @@ def control(
...
+                circuit.mcp(lamb, controls, target)
+            else:
+                circuit.mcp(lamb, controls, target)
+                circuit.mcry(
+                    theta,
+                    controls,
+                    target,
+                    use_basis_gates=True,
+                )
+                circuit.mcp(phi, controls, target)
```
ShellyGarion, line 260: why do you use here mcp-mcry-mcp and not mcrz-mcry-mcrz (which should have less CX gates) ?

Cryoris: I didn't touch any of the synthesis logic in this PR, just how we unroll 🙂 

We can change this in a PR to `main`, but MCP-MCRY-MCP is not the same as MCRZ-MCRY-MCRZ due to the phase difference, so wouldn't that implement a different unitary?

ShellyGarion (approved):
LGTM, thanks for the fix

Cryoris:
@mergifyio backport stable/1.4

PR #13479, opened 2024-11-22
Fix typing of inserted UnitaryGates in quantum_volume

mtreinish:
### Summary

There was a small typing issue in the quantum_volume implementation where there was a mismatch in the Python type and the rust type. The UnitaryGate were being added to the circuit as a rust space PyInstruction instead of a PyGate. This was incorrect as UnitaryGate is unitary and a gate type in python, this mismatch was causing subsequent transpilation or anything working with the unitaries in the quantum volume circuit from rust to mischaracterize the operations as non-unitary so things like getting the matrix would fail. This commit corrects the typing so the gates are added as a unitary operation in the rust typing.

This will get much simpler and less error prone when #13272 is implemented and we have a rust native UnitaryGate type.

### Details and comments

### DIFF
--- a/crates/accelerate/src/circuit_library/quantum_volume.rs
+++ b/crates/accelerate/src/circuit_library/quantum_volume.rs
@@ -27,7 +27,7 @@ use rayon::prelude::*;
 use qiskit_circuit::circuit_data::CircuitData;
 use qiskit_circuit::imports::UNITARY_GATE;
 use qiskit_circuit::operations::Param;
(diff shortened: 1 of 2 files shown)


Cryoris (approved):
LGTM, nice catch 👍🏻 Personally I might have put the test to the quantum volume tests in the circuit library, since it was a bug in the implementation of the quantum volume function, but I'm also fine keeping it here 🙂

PR #13482, opened 2024-11-22
Fix parameter handling for `NLocal(..., flatten=True)` and standard gates

### LINKED ISSUE #13478: Parameter assignment of `QuantumCircuit`does not update the data attribute when inplace=True
### Environment

- **Qiskit version**: 1.2.4
- **Python version**: 3.10
- **Operating system**: Mac OS Sequoia

### What is happening?

When writing the following: 

```python
import numpy as np

from qiskit.circuit.library import TwoLocal

num_qubits = 6
n_layers = 4

qc = TwoLocal(num_qubits, "rx", "cx", entanglement=[[(0, 1), (2, 3), (4, 5)], [(1, 2), (3, 4)]], reps=n_layers, flatten=True, skip_unentangled_qubits=True, skip_final_rotation_layer=True)
qc.assign_parameters(rotation_angles, inplace=True)
print(qc.data)
```

we get:

```python
[CircuitInstruction(operation=Instruction(name='rx', num_qubits=1, num_clbits=0, params=[ParameterVectorElement(θ[0])]), qubits=(Qubit(QuantumRegister(6, 'q'), 0),), clbits=()), CircuitInstruction(operation=Instruction(name='rx', num_qubits=1, num_clbits=0, params=[ParameterVectorElement(θ[1])]), ...]
```

Meaning that instructions still store the reference to the original Parameters, which should have been replaced by the values.

This behavior does not appear when doing 

```python
qc=qc.assign_parameters(rotation_angles, inplace=False)
```

where in this case the binding is clearly visible on the instructions

```python
import numpy as np

from qiskit.circuit.library import TwoLocal

num_qubits = 6
n_layers = 4

qc = TwoLocal(num_qubits, "rx", "cx", entanglement=[[(0, 1), (2, 3), (4, 5)], [(1, 2), (3, 4)]], reps=n_layers, flatten=True, skip_unentangled_qubits=True, skip_final_rotation_layer=True)

rotation_angles = np.random.random(qc.num_parameters)
qc.assign_parameters(rotation_angles, inplace=True)
print(qc.data)
```

### What should happen?

Instructions should be binded with the numeric values in both cases. One should therefore take care of the `inplace=True` case to match the other case.

### Any suggestions?

_No response_

Cryoris:
### Summary

Fixes #13478. 

The fast-path for parameterizing standard gates caused an inconsistent state of the circuit, which is now fixed by using the explicit `circuit._append_standard_gate` method.

### Details and comments

I'm not 100% sure what exactly the inconsistency was. Another way to fix this issue was to have the Rust-side `assign_parameter_inner` to explicitly re-assign the `params` attribute of the cached Python gate. So I'm assuming the old path didn't correctly register the gate in the cache, but I couldn't pin point the exact location in the code. If someone has more insight that would be great to know 🙂


mtreinish (changes requested):
I don't think this actually fixes the issue. This is sidestepping the underlying bug by changing the construction of the n local circuit to not initially populate the py gate cache anymore. This means that the params in the cached python gate objects and the what's stored in circuit data/circuit instruction/packed instruction don't go out of sync because there isn't anything cached yet. If you updated the test to read every gate object from the circuit before assigning the params I think this would fail in the same way reported in the issue.

To fix this fully I'm pretty sure you need to update: https://github.com/Qiskit/qiskit/blob/main/crates/circuit/src/circuit_data.rs#L1382-L1389 to update the cached py gate, if the cache has been set.

mtreinish (approved):
LGTM now, this looks the correct fix, thanks for updating.

garrison:
This change seems to have broken the cutting addon's CI.  I don't understand why yet: https://github.com/Qiskit/qiskit-addon-cutting/issues/714.

PR #10772, opened 2023-09-05
Kraus operator documentation formatting

1ucian0:
There is a small issue in https://qiskit.org/documentation/stubs/qiskit.quantum_info.Kraus.html related with the way that the type enumeration is done in `qiskit.quantum_info.Kraus`:
screencapture-qiskit-org-documentation-dev-stubs-qiskit-quantum-info-Kraus-html-2023-09-05-15_32_24

PR #9705, opened 2023-03-01
Remove dependency on sphinx-autodoc-typehints

jakelishman:
### Summary

This replaces the whole usage of the separate package `sphinx-autodoc-typehints` with the built-in support from `autodoc`, available (with this combination of options) since Sphinx 5.0.  This removes a docs dependency that has caused us significant problems before due to its instability, and the built-in handling in `autodoc` is much cleaner as well; rather than trying to textually modify the docstring, this version (correctly) mutates the parsed rST structure, which also naturally it 100% compatible with the Napoleon processing.

### Details and comments

I set the type-hint location to be both in the signatures and descriptions (our current behaviour is only in the descriptions, unless there's a `typing.overload`).  I'm happy to make it either just signatures or just descriptions, as people prefer.  Here's an example of the new output for `QuantumCircuit.if_else` (a function that currently only has type hints in the signature):

Screenshot 2023-03-01 at 23 35 22

(The slightly buggy output is because of Qiskit/qiskit_sphinx_theme#183; it's pre-existing for  this change.)

For reference, [here's what it looks like in the built documentation](https://qiskit.org/documentation/stable/0.41/stubs/qiskit.circuit.QuantumCircuit.if_else.html).

Thanks to @levbishop for pointing out that `autodoc` does this natively now.

PR #13483, opened 2024-11-22
Show missing type hints & inline class base class in API docs

Eric-Arellano:
The API docs were leaving off some important information:

* For inlined classes, we weren't showing the parent class. This impacts when using `.. autoclass::` on a module page; it was already fine for `.. autosummary::`.
* Generally, we left off type hints for parameters without docstring, along with some return types like `None`. These type hints are useful.

To land this PR, I had to make a fix similar to https://github.com/Qiskit/qiskit/pull/10772 to fix ambiguous type hints.

I also set up aliases for the extremely long types for `EstimatorPubLike` and `SamplerPubLike`. See https://github.com/Qiskit/qiskit-ibm-runtime/issues/1877 for more context on that problem.

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -102,14 +102,21 @@ html_last_updated_fmt = "%Y/%m/%d"
 # documentation created by autosummary uses a template file (in autosummary in the templates path),
 # which likely overrides the autodoc defaults.
 
+# These options impact when using `.. autoclass::` manually.
+# They do not impact the `.. autosummary::` templates.
+autodoc_default_options = {
+    "show-inheritance": True,
+}
+
 # Move type hints from signatures to the parameter descriptions (except in overload cases, where
 # that's not possible).
 autodoc_typehints = "description"
-# Only add type hints from signature to description body if the parameter has documentation.  The
-# return type is always added to the description (if in the signature).
-autodoc_typehints_description_target = "documented_params"
-
 autoclass_content = "both"
+# Some type hints are too long to be understandable. So, we set up aliases to be used instead.
+autodoc_type_aliases = {
+    "EstimatorPubLike": "EstimatorPubLike",
+    "SamplerPubLike": "SamplerPubLike",
+}
 
 autosummary_generate = True
 autosummary_generate_overwrite = False
--- a/qiskit/quantum_info/operators/channel/chi.py
+++ b/qiskit/quantum_info/operators/channel/chi.py
@@ -36,6 +36,7 @@ from qiskit.quantum_info.operators.base_operator import BaseOperator
 if TYPE_CHECKING:
     from qiskit import circuit
 
+
 class Chi(QuantumChannel):
     r"""Pauli basis Chi-matrix representation of a quantum channel.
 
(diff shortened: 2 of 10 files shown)


docs/conf.py
```diff
@@ -102,14 +102,21 @@
...
-# Only add type hints from signature to description body if the parameter has documentation.  The
-# return type is always added to the description (if in the signature).
-autodoc_typehints_description_target = "documented_params"
-
 autoclass_content = "both"
+# Some type hints are too long to be understandable. So, we set up aliases to be used instead.
+autodoc_type_aliases = {
+    "EstimatorPubLike": "EstimatorPubLike",
+    "SamplerPubLike": "SamplerPubLike",
+}
```
Eric-Arellano, line 119: Feel free to disagree on this. The Runtime devs thought in https://github.com/Qiskit/qiskit-ibm-runtime/issues/1877 that the full type is way too long to actually be understandable. I agree with them, and I think it's better to explain pubs in guides like https://docs.quantum.ibm.com/guides/primitive-input-output. But there's an argument for having the full type.

Eric-Arellano (commented):
See https://github.com/Qiskit/documentation/pull/2377 for a preview of how this changes *the Git diff for the 1.3 dev docs.

docs/conf.py
ElePT: I agree with the aliases, the full type is not really helpful.

ElePT (commented):
Thanks Eric! I tried accessing the preview but I didn't see any in the PR you linked. Did I maybe miss something? It's been a while that I haven't checked a docs preview.

Eric-Arellano:
> I tried accessing the preview but I didn't see any in the PR you linked. Did I maybe miss something? It's been a while that I haven't checked a docs preview.

Oops, I meant to say how it changes *the Git diff to see a before and after.

docs/conf.py
```diff
@@ -102,14 +102,21 @@
...
+    "show-inheritance": True,
+}
+
 # Move type hints from signatures to the parameter descriptions (except in overload cases, where
 # that's not possible).
 autodoc_typehints = "description"
-# Only add type hints from signature to description body if the parameter has documentation.  The
-# return type is always added to the description (if in the signature).
-autodoc_typehints_description_target = "documented_params"
-
```
ElePT, line 111: Is there a chance we are exposing parameters that were being kept private on purpose with this change? Else I wonder why this variable would have been set originally.

jakelishman: When we swapped from using the third-party `sphinx-autodoc-typehints` to the built-in `sphinx.ext.autodoc` support for this, I think I just set the options that caused the most similar behaviour to before (#9705).

Eric-Arellano: > Is there a chance we are exposing parameters that were being kept private on purpose with this change?

It wouldn't make sense to me to have certain parameters be private and others public. I suspect certain parameters aren't being documented simply due to oversight, or perhaps people thinking it's intuitive so it doesn't warrant docstring.

Also, the parameters _are_ already in the docs as part of the signature. The only change this PR makes is showing their type hints.

ElePT (approved):
Thanks for the context, @Eric-Arellano and @jakelishman. I am in favor of merging the PR after the discussion in the dev meeting. I don't think we have any strong reason not to. But let me know if I missed anything.

Eric-Arellano:
> I don't think we have any strong reason not to. But let me know if I missed anything.

Thanks! Yes, to confirm, this should be safe and uncontroversial to merge.

PR #13484, opened 2024-11-24
Fix `CollectCliffords` to index nodes using their ids

### LINKED ISSUE #13457: `CollectCliffords` pass fails for circuit with large `PauliLindbladError`
### Environment

- **Qiskit version**: 1.3.0rc1
- **Python version**: 3.11
- **Operating system**: macos

### What is happening?

`qiskit.transpiler.passes.CollectCliffords` goes through a circuit and merges recognized clifford gates into combined Clifford operations.

However, this fails if we've appended a many-qubit `PauliLindbladError` (`qiskit_aer.noise.errors.PauliLindbladError`) to the circuit. The `PauliLindbladError` is represented in the circuit as a `quantum_channel` instruction. At some point, `CollectCliffords` attempts to convert this channel to a `Kraus` object, which involves creating a many-qubit `SuperOp` which naturally crashes with this error:
```ValueError: array is too big; `arr.size * arr.dtype.itemsize` is larger than the maximum possible size.```

The attempt to convert to a `Kraus` object apparently happens here: 
https://github.com/Qiskit/qiskit/blob/cd26d498c9b74475753b9179ac2b7f84122ffe81/qiskit/dagcircuit/collect_blocks.py#L168
It looks like it encounters the `PauliLindbladError` (as a `quantum_channel`) as a successor node (of something), and tries to use that `quantum_channel` as a dictionary key (`suc`). Attempting to check for equality with other dict keys apparently involves converting the `quantum_channel` to a many-qubit `SuperOp`, which fails.

```
from qiskit.transpiler.passes import CollectCliffords
from qiskit.transpiler.passmanager import PassManager
from qiskit_aer.noise.errors import PauliLindbladError

pm = PassManager(CollectCliffords(
    do_commutative_analysis = False,
    split_blocks = False,
    split_layers = False,
    collect_from_back = False,
))

qc = QuantumCircuit(20)
qc.cx(0,1)
qc.cx(0,1)
qc.append(PauliLindbladError(generators=['X'+'I'*19],rates=[0.1]),range(20))

qc_transpiled = pm.run(qc)
```

### What should happen?

Personally I don't want `CollectCliffords` to attempt to do anything with the `quantum_channel` instructions. It should leave those instructions unchanged, without raising any errors.

### Any suggestions?

It looks like internally, the low-level `collect` functions use a `filter_fn` in order to ignore some instructions, avoiding the dictionary lookup:
https://github.com/Qiskit/qiskit/blob/cd26d498c9b74475753b9179ac2b7f84122ffe81/qiskit/dagcircuit/collect_blocks.py#L163-L172

Can we modify the existing `filter_fn` used by `CollectCliffords` so that it ignores `quantum_channel` instructions?

### PR DESCRIPTION
**Pull request by alexanderivrii**
### Summary

Fixes #13457.

This PR changes the internal field ``in_degree`` in ``BlockCollector`` to be indexed by the ids of the nodes rather than by nodes themselves. See also #13424 for a related discussion.

In particular, this avoids hashing the nodes as python objects, both making the code faster and avoiding problems for things like `PauliLindbladError` coming from `qiskit_aer`.

### Details and comments

I am not sure if this change really deserves a release note, but I have decided to add one, given that it does address a reported bug.

### DIFF
--- a/qiskit/dagcircuit/collect_blocks.py
+++ b/qiskit/dagcircuit/collect_blocks.py
@@ -80,7 +80,7 @@ class BlockCollector:
         self._in_degree = {}
         for node in self._op_nodes():
             deg = len(self._direct_preds(node))
-            self._in_degree[node._node_id] = deg
+            self._in_degree[self._get_node_id(node)] = deg
             if deg == 0:
                 self._pending_nodes.append(node)
 
@@ -90,6 +90,13 @@ class BlockCollector:
             return self.dag.op_nodes()
         else:
             return self.dag.get_nodes()
+
+    def _get_node_id(self, node):
+        """Returns the id of the given node."""
+        if not self.is_dag_dependency:
+            return node._node_id
+        else:
+            return node.node_id
 
     def _direct_preds(self, node):
         """Returns direct predecessors of a node. This function takes into account the
@@ -165,8 +172,8 @@ class BlockCollector:
 
                     # update the _in_degree of node's successors
                     for suc in self._direct_succs(node):
-                        self._in_degree[suc._node_id] -= 1
-                        if self._in_degree[suc._node_id] == 0:
+                        self._in_degree[self._get_node_id(suc)] -= 1
+                        if self._in_degree[self._get_node_id(suc)] == 0:
                             new_pending_nodes.append(suc)
                 else:
                     self._pending_nodes.append(node)
--- a/qiskit/dagcircuit/collect_blocks.py
+++ b/qiskit/dagcircuit/collect_blocks.py
@@ -54,7 +54,7 @@ class BlockCollector:
 
         self.dag = dag
         self._pending_nodes: list[DAGOpNode | DAGDepNode] | None = None
-        self._in_degree: dict[DAGOpNode | DAGDepNode, int] | None = None
+        self._in_degree: dict[int, int] | None = None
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: qiskit/dagcircuit/collect_blocks.py [hunk 1]
```diff
@@ -54,7 +54,7 @@ def __init__(self, dag: DAGCircuit | DAGDependency):
 
         self.dag = dag
         self._pending_nodes: list[DAGOpNode | DAGDepNode] | None = None
-        self._in_degree: dict[DAGOpNode | DAGDepNode, int] | None = None
+        self._in_degree: dict[int, int] | None = None
```
mtreinish, line 57: This is problematic, using the node ids is only safe for a single dag. If we want to switch this to be keyed on node ids we can **not** cache it. We could get away with the cache with the `DAGNode` storage because the hash and eq implementations are unique; either pre-rust dagcircuit where it was object id based or post-rust dag where this is an actual implementation. But node ids are not unique between dags and the indices are reused between dags and will refer to different nodes.

alexanderivrii: I see the point, but a single `BlockCollector` class only makes sense for a single DAG, in which case we should be fine using node ids, correct?

alexanderivrii:
Closing this.

PR #13486, opened 2024-11-25
Port standard instructions to Rust.

### LINKED ISSUE #12966: Add rust representation for non-gate standard circuit operations

kevinhartman:
### Summary
Adds a new Rust enum `StandardInstruction` representation for standard instructions (i.e. `Barrier`, `Delay`, `Measure` and `Reset`) and updates the encoding of `PackedOperation` to support storing it compactly at rest.

The size of a `PackedOperation` has now been anchored to always be 64 bits ~, even on 32-bit systems~ _**EDIT: we [no longer support 32-bit](https://github.com/Qiskit/qiskit/pull/13650) anyway.**_ This is necessary to encode the data payload of standard instructions like `Barrier` and `Delay`. See the docstrings within `packed_instruction.rs` for details.

### Details and comments
The implementation works very similarly to what we currently do for `StandardGate`, but with a bit of extra consideration for an additional data payload used by variadic/template instructions like `Barrier` and `Delay`.

Similarly to `StandardGate`, the newly added `StandardInstruction` enum serves as the first-class Rust interface for working with standard instructions. The existing `OperationRef` enum returned by `PackedOperation::view` has a new variant `StandardInstruction` which exposes this.

Unlike `StandardGate`, `StandardInstruction` is not a `pyclass` because it is not directly used to tag instructions as standard in our Python class definitions. Instead, the simple integral enum `StandardInstructionType` is used as the tag, and `OperationFromPython::extract_bound` further queries the incoming Python object for variadic/template data when building the complete `StandardInstruction`.

#### Encoding
The `PackedOperation` encoding has been updated as follows:

* A `PackedOperation` is now always 64 bits wide
* The discriminant has been widened from 2 to 3 bits (it is now at its maximum width, but we still have room for 3 more variants).
* The discriminant now has additional variant `StandardInstruction`.
* A packed standard gate's enum is now byte-aligned.


jakelishman:
I'm concerned that the `LoHi` struct and `PackedOperation` union are making it easier to make mistakes on BE and/or 32-bit systems, because the encodings now change between them.

I feel like the bitshift and masking operations could be extended over the whole u64, and that'll all automatically work on BE and 32-bit systems, especially with the size of `PackedOperation` now fixed to 64 bits - we can even have everything be aligned. The shifts of the payload of instructions can be set such that a u32 payload is always in the high bits, with the padding bits between it and the rest of the discriminants if you're concerned about aligned access, because then the bitshifts will be compiled out into a single register load (just read the u32 from the right place) and the compiler will handle endianness for us. The pointer doesn't need to be handled any differently to any other payload - everything's just shifts and masks anyway (`LoHi` can't avoid that), and I think introducing the `union` makes that harder to follow.

The `LoHi` struct to me seems to be forcing every method of `PackedOperation` to think about the partial register reads/loads. If we use shifts and masks on a u64 with `const` masks and shifts, there's no logic split where some of it is done with shifts and some with endianness-dependent partial register access in our own code.

jakelishman:
I guess my main point is this: the 64-bit `PackedOperation` can always be thought about as a collection of payloads each of which is stored in a particular `mask` and needs a particular `shift`.  These are:

- the `PackedOperation` discriminant at (`0b111`, 0)
- a pointer at (`usize::MAX as u64 & 0b000`, 0) (regardless of 32-bit or 64-bit)
- a `StandardGate` discriminant at (`0xff << 3`, 3)
- a `StandardInstruction` discriminant at (`0xf << 3`, 3)
- a `DelayUnit` payload at (for example) (`0x7 << 32`, 32)
- a `u32` payload at (`u32::MAX as u64 << 32`, 32)

Introducing `LoHi` doesn't remove the need to bitshift and mask for most items, it just means that some of my above list are done one way, some are done another, and the union means that the pointer now has more ways to access it.  `LoHi` also restricts the payload size to u32, when we already have payloads that exceed that (the pointers).

If you want a shade more encapsulation than my loose `const` shift and mask associated items, would something like this look better to you?

```rust
struct PayloadLocation<T: SomeBytemuckOrCustomTrait> {
    mask: u64,
    shift: u32,
    datatype: PhantomData<T>,
}
impl PayloadLocation {
    // Note we _mustn't_ accept `PackedOperation` until we've got a valid one,
    // because holding a `PackedOperation` implies ownership over its stored
    // pointer, and it mustn't `Drop` or attempt to interpret partial data.

    #[inline]
    fn store(&self, src: T, target: u64) -> u64 {
        let src: u64 = bytemuck::cast(src);
        target | ((src << self.shift) & self.mask)
    }
    #[inline]
    fn load(&self, target: u64) -> T {
        bytemuck::cast((target & self.mask) >> self.shift)
    }
}

const PACKED_OPERATION_DISCRIMINANT = PayloadLocation { mask: 0b111, shift: 0 };
const STANDARD_GATE_DISCRIMINANT = PayloadLocation { mask: 0bff << 3, shift: 3 };
const POINTER_PAYLOAD = PayloadLocation { mask: usize::MAX as u64 & 0b000, shift: 0 };
// and so on
```

(Bear in mind I just typed that raw without trying it, and I didn't think all that hard about what the trait bound should be.)

If everything's inlined and constant, the compiler absolutely should be able to compile out 32 bit shifts and all-ones masks into partial register reads/loads itself, so there oughtn't to be any overhead.

crates/circuit/src/circuit_instruction.rs
```diff
@@ -695,6 +701,37 @@ impl<'py> FromPyObject<'py> for OperationFromPython {
                 extra_attrs: extract_extra()?,
             });
         }
+        'standard_instr: {
```
jakelishman, line 726: If you're adding this label as a separate one, you probably want to rename the `'standard` label above to `'standard_gate`.

crates/circuit/src/circuit_instruction.rs
```diff
@@ -695,6 +701,37 @@ impl<'py> FromPyObject<'py> for OperationFromPython {
...
+            // Our Python standard instructions have a `_standard_instruction_type` field at the
+            // class level so we can quickly identify them here without an `isinstance` check.
+            // Once we know the type, we query the object for any type-specific fields we need to
+            // read (e.g. a Barrier's number of qubits) to build the Rust representation.
+            let Some(standard_type) = ob_type
+                .getattr(intern!(py, "_standard_instruction_type"))
```
jakelishman, line 732: Why `_standard_instruction_type` and not `_standard_instruction`, which is consistent with `_standard_gate`?

crates/circuit/src/operations.rs
```diff
@@ -269,13 +280,218 @@ impl<'a> Operation for OperationRef<'a> {
...
+
+/// An internal type used to further discriminate the payload of a `PackedOperation` when its
+/// discriminant is `PackedOperationType::StandardInstruction`.
+///
+/// This is also used to tag standard instructions via the `_standard_instruction_type` class
+/// attribute in the corresponding Python class.
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+#[pyclass(module = "qiskit._accelerate.circuit", eq, eq_int)]
+#[repr(u8)]
+pub(crate) enum StandardInstructionType {
```
jakelishman, line 337: Personally, I don't think I've seen any compelling reasons to use `pub(crate)` yet.  Either it should be entirely private to this module, or it's useful outside the module, in which case it's likely to be useful outside the crate too.  Feels to me that almost every time we've introduced a `pub(crate)`, we've had a PR shortly after making it a `pub`, and in this case, it's logically exposed publicly to Python space anyway.

Should this type live here at all, or is it just an implementation detail of `PackedOperation`, and could move there?

(Though see other questions about the new `PackedOperation` implementation - maybe other stuff should move _here_.)

crates/circuit/src/operations.rs
```diff
@@ -269,13 +280,218 @@ impl<'a> Operation for OperationRef<'a> {
...
+                }
+                out.setattr("_duration", duration)?;
+            }
+            if let Some(unit) = unit {
+                if !mutable {
+                    out = out.call_method0("to_mutable")?;
+                }
+                out.setattr("_unit", unit)?;
+            }
+        }
```
jakelishman, line 482: How does the `unit`/`duration` stuff here interact with `Delay`, which expects those fields already?

crates/circuit/src/operations.rs
```diff
@@ -332,13 +548,15 @@ pub enum StandardGate {
     C3XGate = 49,
     C3SXGate = 50,
     RC3XGate = 51,
+    // Remember to update StandardGate::is_valid_bit_pattern below
+    // if you add or remove this enum's variants!
```
jakelishman, line 544: Weird that you added a comment here, but not on the `StandardInstructionType` introduced in this PR with the same consideration, but thanks for catching it!

crates/circuit/src/operations.rs
```diff
@@ -269,13 +280,218 @@ impl<'a> Operation for OperationRef<'a> {
...
+pub(crate) enum StandardInstructionType {
+    Barrier = 0,
+    Delay = 1,
+    Measure = 2,
+    Reset = 3,
+}
+
+#[derive(Clone, Debug, Copy, Eq, PartialEq, Hash)]
+pub enum StandardInstruction {
+    Barrier(usize),
```
jakelishman: Why `usize` here?  We use `u32` for qubit counts in most places, and the details of `PackedOperation` require any payload here to be `u32` anyway.

crates/circuit/src/circuit_instruction.rs
```diff
@@ -695,6 +701,37 @@ impl<'py> FromPyObject<'py> for OperationFromPython {
...
+                    StandardInstruction::Delay(unit)
+                }
+                StandardInstructionType::Measure => StandardInstruction::Measure,
+                StandardInstructionType::Reset => StandardInstruction::Reset,
+            };
+            return Ok(OperationFromPython {
+                operation: PackedOperation::from_standard_instruction(standard),
+                params: extract_params()?,
+                extra_attrs: extract_extra()?,
+            });
```
jakelishman: Should `Delay` have different handling of its extras, since `unit` and `duration` are required fields for that?  Have we thought about how we'll handle that in the future?

crates/circuit/src/packed_instruction.rs
```diff
@@ -39,132 +42,375 @@ enum PackedOperationType {
     // will make it appear as a standard gate, which will never allow accidental dangling-pointer
     // dereferencing.
     StandardGate = 0,
-    Gate = 1,
-    Instruction = 2,
-    Operation = 3,
+    StandardInstruction = 1,
+    PyGatePointer = 2,
+    PyInstructionPointer = 3,
+    PyOperationPointer = 4,
```
jakelishman: Personally I think the `_Pointer` suffixes are a bit unnecessarily noisy, but `PyGate`, `PyInstruction`, etc is better (*edit*: better than what I'd written before).

crates/circuit/src/packed_instruction.rs
```diff
@@ -39,132 +42,375 @@ enum PackedOperationType {
...
 ///
-/// There is currently one spare bit that could be used for additional metadata, if required.
+/// PointerBits (32-bit systems)
+/// 0b_PPPPPPPP_PPPPPPPP_PPPPPPPP_PPPPPPPP_xxxxxxxx_xxxxxxxx_xxxxxxxx_xxxx011
+///    |---------------------------------|                                |-|
+///                    |                                                   |
+///    On 32-bit systems, the upper 32-bits are used to store the full     +-- Discriminant. This is
+///    address. This gives the compiler the best chance at eliminating         0b100, so in this
+///    the otherwise required bit-shift needed to read the pointer.            case, this points to
+///                                                                            a `PyOperation`.
```
jakelishman: I stand by the previous comment that I don't think it's a good idea to have the explicitly platform-dependent code here for pointers.  Seems like the only reason to attempt to separate them is to then try and make 32-bit pointers have one fewer bitshift?  But we don't care as much about performance of 32-bit platforms, and even if we did, I think the masking out of three lower bits in a pointer on access would be entirely imperceptible compared to the memory indirection penalty that's almost certainly going to be incurred just by following the pointer.

crates/circuit/src/packed_instruction.rs
```diff
@@ -39,132 +42,375 @@ enum PackedOperationType {
...
+                << StandardInstructionBits::DISCRIMINANT_OFFSET
+                == Self::DISCRIMINANT_MASK,
+            "(StandardInstructionBits) discriminant MUST be the 3 lowest bits!"
+        );
+        assert!(
+            ((1 << PointerBits::DISCRIMINANT_BITS) - 1) << PointerBits::DISCRIMINANT_OFFSET
+                == Self::DISCRIMINANT_MASK,
+            "(PointerBits) discriminant MUST be the 3 lowest bits!"
+        );
+    };
```
jakelishman: I'm not really sure what this is buying over a comment on the `discriminant` member of each bitfield saying "this _must_ be the lowest three bits".  Feels a bit like boiler plate to make a testing harness for the sake of it, when the comment would have the same result - this won't catch problems in new bitfields because you'd have to remember to write the test, and it moves the "comment" away from the place where you'd actually have introduced the bug, which would also mean that if the test gets broken, it's less likely to be caught by the PR reviewer.

It's a cool trick, but I think it might be a bit overengineering here.

crates/circuit/src/packed_instruction.rs
```diff
@@ -39,132 +42,375 @@ enum PackedOperationType {
...
+    }
+    const fn from_bits(value: u8) -> Self {
+        if value as usize >= STANDARD_GATE_SIZE {
+            panic!("unexpected standard gate type!")
+        }
+        // It would've been nice to use bytemuck here, but it doesn't support const
+        // casting for enums.
+        unsafe { std::mem::transmute(value) }
+    }
+}
```
jakelishman: I don't love that this is an `impl` from miles away from where `StandardGate` is defined - this really feels a bit spaghetti-ish to be introducing new `impl` functions.  At a minimum, can we use the `from` and `into` macro rules to the bitfield derivation mapped to stand-alone functions?

Why is the `const` casting necessary?  I don't love re-introducing `transmute` here, and having the `from_bits` check defined a long way away from the type itself.

crates/circuit/src/packed_instruction.rs
```diff
@@ -39,132 +42,375 @@ enum PackedOperationType {
...
-    /// low pointer bits that are guaranteed to be 0 by alignment, so no shifting is required.
-    const POINTER_MASK: usize = usize::MAX ^ Self::DISCRIMINANT_MASK;
-
-    /// Extract the discriminant of the operation.
     #[inline]
     fn discriminant(&self) -> PackedOperationType {
-        ::bytemuck::checked::cast((self.0 & Self::DISCRIMINANT_MASK) as u8)
+        // SAFETY: we read (just!) the discriminant from any of the union's members,
+        // since we guarantee it is found in the same place for all bitfields.
+        unsafe { self.gate.discriminant() }
```
jakelishman: I'm really nervous about this, because I think this is already invoking undefined behaviour, which wasn't happening before the bitfield structs were introduced.

Not all 8-bit patterns are valid `StandardGate`, and doing `self.gate` interprets bits 3 to 11 directly as a `StandardGate` in order to produce the `StandardGateBits` struct.  If this object is actually a pointer, those 8 bits may well _not_ be a `StandardGate`, and that's UB, even if we don't explicitly access them.

crates/circuit/src/packed_instruction.rs
```diff
@@ -39,132 +42,375 @@ enum PackedOperationType {
...
+    fn u32(&self) -> u32 {
+        self.0
     }
+}
+
+/// The bitfield layout used by pointer types, like `PyGate`, `PyInstruction`, and `PyOperation`.
+///
+/// On 64-bit systems, the address is stored in the upper 62 bits, and is converted into and out of
+/// its packed form using `unpack_address` and `pack_address`. On 32-bit systems, the address is
+/// simply stored in the upper 32 bits.
```
jakelishman, line 338: I think we could do "simply" better if we just do the same on both platforms, and have the `pointer` method cast the `u64` to `usize`.

crates/circuit/src/packed_instruction.rs
```diff
@@ -39,132 +42,375 @@ enum PackedOperationType {
...
+#[cfg(target_pointer_width = "64")]
+#[bitfield(u64, new = false)]
+struct PointerBits {
+    #[bits(3, access = RO)]
+    discriminant: PackedOperationType,
+    #[bits(61, from = PointerBits::unpack_address, into = PointerBits::pack_address)]
+    address: usize,
+}
```
jakelishman: I believe that `bitfield_struct` automatically impls `Copy`, and there's no way to disable this.  I'm really nervous about that, because this PR is introducing a lot more types into the system here, and I think it's really easy to lose the ownership semantics in the middle of all of them.

crates/circuit/src/packed_instruction.rs
```diff
@@ -39,132 +42,375 @@ enum PackedOperationType {
...
+struct StandardInstructionBits {
+    #[bits(3, default = PackedOperationType::StandardInstruction, access = RO)]
+    discriminant: PackedOperationType,
+    #[bits(8)]
+    standard_instruction: StandardInstructionType,
+    #[bits(21)]
+    _pad1: u32,
+    #[bits(32)]
+    value: ImmediateValue,
+}
```
jakelishman: It's odd to me that you went to all the trouble with the 32-bit pointers to make them aligned, even though they're the "terrible performance" path on platforms we don't even care much about, yet `StandardGate` and `StandardInstruction` (very much meant to be the absolute fastest paths) have their internal discriminant misaligned.

A second consideration: an alternative to having `StandardInstructionBits` forcibly define the bitlayout of `StandardInstruction` itself would be to have `StandardInstruction` have a `to_bits` method that is required to return a 61-bit value, and leave it up to `StandardInstruction`.  That would let the code be a lot more local to the rest of the type, while in this form, there's still co-operation necessary between `StandardInstruction` and `StandardInstructionBits`, but they're defined far away from each other.

crates/circuit/src/packed_instruction.rs
```diff
@@ -39,132 +42,375 @@ enum PackedOperationType {
...
+    #[inline]
+    fn delay_unit(&self) -> DelayUnit {
+        ::bytemuck::checked::cast(self.0 as u8)
+    }
+
+    #[inline]
+    fn u32(&self) -> u32 {
+        self.0
     }
+}
```
jakelishman: This `ImmediateValue` is largely indicative of the problem I meant above.  I feel like there's assumptions and invariants that need to be upheld between two places now, and that makes things a lot more difficult:

1. the maximum width of a `StandardInstruction` payload must be the same size as `ImmediateValue`.  This PR already violates that from a typing perspective (`Barrier(usize)`).
2. `ImmediateValue` has been marked as deriving `Clone` and `Copy`, except the actual type that's been punned to a `u32` might not logically allow the `Clone` and `Copy`.  Obviously `u32` would let us clone it unfairly too, but it's easier to reason about ownership semantics if the _actual_ owner is the one producing the new type.
3. (much more minor and future looking) why do we need to apply an arbitrary limit of 32 bits anyway?  What if there's a payload in the future that's 2x16 bits, and those don't need to be pushed up to the 32-bit boundary for aligned access anyway.  It feels a bit neater to me to give `StandardInstruction` the whole payload if it wants it, and say it's required to return a 64-bit value with the low three bits 0, then let it put things where it wants them.

crates/circuit/src/packed_instruction.rs
```diff
@@ -210,51 +487,73 @@ impl PackedOperation {
...
+    }
+
+    /// Create a `PackedOperation` from a `StandardInstruction`.
+    pub fn from_standard_instruction(instruction: StandardInstruction) -> Self {
+        let mut bits = StandardInstructionBits::new();
+        match instruction {
+            StandardInstruction::Barrier(num_qubits) => {
+                let num_qubits: u32 = num_qubits.try_into().expect(
+                    "The PackedOperation representation currently requires barrier size to be <= 32 bits."
+                );
```
jakelishman: Then why don't we enforce that in the typing?

jakelishman (commented):
I haven't thought hugely hugely about all the implications of this all the way through, but here's comments on the implementation.

The rough summary is:

- this does look like a successful way of doing it
- I agree that you can kind of view these as bitfield structs
- I'm very nervous that `bitfield-struct` is causing us to invoke undefined behaviour, introduce more `unsafe`, and I'm not certain it's making the code a huge amount clearer.

My biggest worry about `bitfield-struct` as written here is about accessing the `PackedOperation` discriminant - getting it from the `StandardGateBits` representation is, I believe, UB.  We can swap that to the pointer form, but the fact that it's so _easy_ to produce UB like that makes me nervous.

My other hesitation is that generally these aren't entirely bitfields, because of the overlap.  All the extra code needed to support the masking out of the low bits of the pointer feels more complex to me than it was before, and I'm not convinced that `bitfield_struct` is offering us enough in return.  If anything, it feels like we might have spread the dangerous assumptions further around across more types and more (type-level) indirection than there was before.

That said, I'm not really strongly against it.  I worry that it's using a pneumatic drill to solve a hammer-and-nail problem here since there really are not many fields at all in our structs, and these structs should not need much manipulation other than construction. My perception of the complexity of manual bitshift-and-mask operations compared to this might be different to other people's.

kevinhartman:
Apologies for the delay in responding to your initial comments, @jakelishman! I did read them, and they were ultimately what led me to rewrite the initial draft using bitfields. I meant to reply to your comments at the same time I pushed up these changes, but didn't have enough time yesterday.

#### Regarding the initial draft:

> The shifts of the payload of instructions can be set such that a u32 payload is always in the high bits, with the padding bits between it and the rest of the discriminants if you're concerned about aligned access, because then the bitshifts will be compiled out into a single register load (just read the u32 from the right place) and the compiler will handle endianness for us.

This is exactly what I was concerned about, though I was also trying to accomplish it in a way that the alignment of `PackedOperation` could be 4 on 32-bit systems, which is partly why I'd ended up with `LoHi`. In the current approach with bitfields, I've forgone this. The alignment of `PackedOperation` is now always 8 bytes, even on 32-bit.

> Introducing LoHi doesn't remove the need to bitshift and mask for most items, it just means that some of my above list are done one way, some are done another, and the union means that the pointer now has more ways to access it. LoHi also restricts the payload size to u32, when we already have payloads that exceed that (the pointers).

This was true, and ultimately indicative of me trying to jam everything into a single instruction encoding / layout.

I liked the `PayloadLocation` example you provided, and I actually spent some time locally building it out. I couldn't shake the feeling though that we were implementing a poor-man's bitfield library (and, one without static layout validation at that), so I started poking around and found the `bitfield-struct` crate. It's _really excellent_. The author has also already merged a PR I made to optionally disable the derived `Clone` implementation on its bitfields.

#### Regarding the current version:

> My biggest worry about bitfield-struct as written here is about accessing the PackedOperation discriminant - getting it from the StandardGateBits representation is, I believe, UB. We can swap that to the pointer form, but the fact that it's so easy to produce UB like that makes me nervous.

This I'm not convinced of. Type-punning is allowed in Rust and is not UB, as long as the access doesn't produce an invalid value, if I understand correctly. It's also of course important that we're not reading uninitialized memory, which we shouldn't be here.

From https://doc.rust-lang.org/reference/items/unions.html#reading-and-writing-union-fields

> Unions have no notion of an “active field”. Instead, every union access just interprets the storage as the type of the field used for the access. Reading a union field reads the bits of the union at the field’s type. [..] It is the programmer’s responsibility to make sure that the data is valid at the field’s type. Failing to do so results in [undefined behavior](https://doc.rust-lang.org/reference/behavior-considered-undefined.html).

If you're worried about the representations of the fields comprising the union, they are all `#[repr(transparent)]` structs around a `u64`. So no matter which field we access, the data gets interpreted as a `u64`, which is valid. And, the `::new` generated by `bitfield-struct` starts off by setting the inner `u64` to `0`, so there shouldn't be any case where we are interpreting uninitialized data. The discriminant is located (and statically asserted to be) in the same place for all bitfields, so that should be safe too.

> My other hesitation is that generally these aren't entirely bitfields, because of the overlap. All the extra code needed to support the masking out of the low bits of the pointer feels more complex to me than it was before, and I'm not convinced that bitfield_struct is offering us enough in return. If anything, it feels like we might have spread the dangerous assumptions further around across more types and more (type-level) indirection than there was before.

In the 64-bit pointer case, I was also thinking about it as "overlapping" a pointer with a discriminant originally, and that was the reason I hadn't thought to use bitfields. But really, there's no overlapping here. We're simply storing the upper 61 bits (the interesting ones!) of an address inside of a 61 bit field. I'd argue that the code is less complex in terms of bit manipulation: there's no explicit masking at all, we just shift an address down by 3 and store it.

> I worry that it's using a pneumatic drill to solve a hammer-and-nail problem here since there really are not many fields at all in our structs, and these structs should not need much manipulation other than construction. My perception of the complexity of manual bitshift-and-mask operations compared to this might be different to other people's.

The original approach was certainly the right one at the time. In earlier commits, I explored extending it for `StandardInstruction`, but it felt very easy to read/write the wrong part of the field or make the wrong cast with the added complexity. The bitfield approach validates at compile-time that fields aren't overlapping and lets us model `PackedOperation`'s layouts declaratively. I think this is actually quite elegant, overall.

crates/circuit/src/circuit_instruction.rs
kevinhartman: I can change this to `_standard_instruction`, if you prefer. I did this because there's also a Rust-side `StandardInstruction` enum, and even though it isn't exposed to Python, my thinking was that this would be more consistent to readers of the Qiskit codebase if the name here aligned to the `StandardInstructionType` Rust enum that backs it.

crates/circuit/src/operations.rs
kevinhartman: It is `pub(crate)` here only so that we can expose it to Python from `lib.rs`, IIRC. Perhaps that is not necessary and I was mistaken.

crates/circuit/src/operations.rs
kevinhartman: This is more an artifact of my workflow, which tends to be fairly iterative. I noticed these were out of sync, so I left a comment and assumed I'd fix up the rest of the docs / comments for the PR in accordance once everything else was fleshed out.

crates/circuit/src/operations.rs
kevinhartman: I will change this to `u32` for this PR.

This was around the time I was working on https://github.com/Qiskit/qiskit/pull/13377, which I thought might be the direction we wanted to go (using `usize` to represent bit indices, but failing at runtime if internal data structures were to attempt storage of something bigger than a `u32`).

crates/circuit/src/packed_instruction.rs
kevinhartman: If anything, 32-bit CPUs with memory indirect addressing modes would be able to more efficiently dereference the pointer if it doesn't need shifting and masking, right?

And, we can't use a 3-bit mask on 32-bit systems since the addresses would be 4-byte aligned.

The idea here was that we're making life worse for 32-bit systems by making `PackedOperation` always 64 bits. I wanted to therefore do as much as easily possible to curb that. I don't think the separate layout is a big deal in terms of complexity, and it doesn't need any special consideration externally to `PointerBits`.

crates/circuit/src/packed_instruction.rs
kevinhartman: I did this so they'd be private to this file. I agree that it would be safer for these to be defined beside `StandardGate`, like we do with the `bytemuck` stuff.

crates/circuit/src/packed_instruction.rs
kevinhartman: The `StandardGateBits` struct is actually just a transparent tuple struct around a `u64`. And that `u64` is zero-initialized, so we should be safe here.

jakelishman:
For the UB: `bytemuck` made it near impossible to get this wrong.  This implementation with `bitfield_struct` may be right, because we don't call the problematic getter, and the value isn't constructed til then, but it's nowhere near as obvious to me with the new form, and it feels a lot more dangerous to even access an incorrect `union` type - much easier to change the code to get it wrong when you've already "constructed" the wrapper type that would allow the invalid access, as opposed to still explicitly treating it as "this is a bag of bits".

I was concerned about the UB because I know (as your quote says) that even constructing an `enum` value with an invalid discriminant is UB, even if you never read the value.  Just constructing it in memory is UB.  I guess that probably doesn't happen here, but to know that, you need to really think about how the `bitfield_struct` crate works, which is more mental load - from the `struct` definition, you're defining a struct that _looks like_ it's automatically constructed, so you have to fight standard Rust intuition to realise that it's not UB (which obviously I didn't get right the first time round).

---

Re the pointer overlap: a pointer is 64 bits, not 61 bits, so what the bitfield is storing isn't a pointer, it's as you say, the informative bits of the pointer - the true value of the actual pointer really is overlapped with the discriminant.  Strictly, x86_64 pointers are only 48 bits wide and required to be all 0s in the top 16 bits, so we _might_ even want to use those bits too in the future, and they'd be overlapping.  Then it'd really be clearer as a single mask, not a bitfield.

---

> The bitfield approach validates at compile-time that fields aren't overlapping and lets us model `PackedOperation`'s layouts declaratively. I think this is actually quite elegant, overall

I mentioned before that imo, the fields do overlap for pointers, and we're fighting `bitfield_struct` in those cases because it's trying to assert that nothing actually overlaps.  About elegance: I do appreciate that you can think about it as an extension of bitfields, but my reading of the resulting code here is that there's more boilerplate accessors/whatever than there was before, and there's a lot more complexity in the type system which is requiring more assumptions to be spread between different objects in the code.  To me, the reasoning about this new system is less local (it's the subject of a few of my comments).

I can totally believe that the `bitfield_struct` crate is really neat.  My problem is that the pointer object just isn't a true bitfield - it's a really simple "mask out the low bits" trick, and we're adding a lot of complexity around that to use an abstraction that doesn't fit quite right.  It might be super obvious to you, having played around with the `bitfield_struct` crate a load, but it adds more complexity for me reading the code, and a whole other library that I need to know the intricacies of.

Fwiw, my preferred form probably isn't the `PayloadLocation` encapsulation I attempted above.  As a rough sketch, I was thinking more like: define an `unsafe` sealed trait that `StandardGate` and `StandardInstruction` implement, which supertraits `Copy + !Drop`, and has methods that encode them into a `u64` and back again, that's defined on them (so the reasoning about the data representation is local to those types, including the choices of how to encode themselves), and has a safety comment that they must carry no information in the low three bits.  The `Copy + !Drop` requirement is what `PackedOperation` really requires for those right now (though tbh there's probably ways to avoid that), and other than that, the "low 3 bits will be overwritten" is the only thing that the type itself needs to know to implement things.  In that form, there's no shift operations needed within `PackedOperation` itself, so little to get wrong - `PackedOperation` does
```rust
struct PackedOperation(u64);

fn do_something(&self) {
    let discriminant = self.0 & Self::DISCRIMINANT_MASK;
    let payload = self.0 & !Self::DISCRIMINANT_MASK;
    match discriminant {
        PackedOperationType::StandardGate => <StandardGate as Packable>::unpack(payload),
        PackedOperationType::StandardInstruction => <StandardInstruction as Packable>::unpack(payload),
        PackedOperationType::PyGate => ...,
    }
```
and all the other bit-twiddling / whatever is local to the types themselves.  In cases where a bitfield representation is appropriate (certainly I could be convinced by `StandardInstruction` in that case), then we could do that.  I don't think the current implementations of the pointer-like objects or `StandardGate` are served by the bitfields, though.

crates/circuit/src/packed_instruction.rs
kevinhartman: I agree that it ought not to. The author has merged a PR I've made which allows the derived `Clone` (and thus `Copy`) to be disabled. Once included in a release, we should be able to get rid of the inner `BitField` and make `PackedOperation` the union type instead, if we should want to.

crates/circuit/src/packed_instruction.rs
kevinhartman: > It's odd to me that you went to all the trouble with the 32-bit pointers to make them aligned, even though they're the "terrible performance" path on platforms we don't even care much about, yet StandardGate and StandardInstruction (very much meant to be the absolute fastest paths) have their internal discriminant misaligned.

I did consider this, and we can certainly add a bit of padding to get them over the byte boundary. The 32-bit case was just what I was focusing on initially since I understood we would be impeding performance for that case the most. You didn't pad `StandardGate` originally, so I figured I wasn't making anything worse by deferring that.

crates/circuit/src/packed_instruction.rs
jakelishman: If the dtructs being pointed to aren't already forcibly 8-byte aligned, you can stick an `align(8)` on them - I thought I did it already, but if I didn't, we'd always planned to.

Yeah, I'm not arguing that this wouldn't be ever so slightly more efficient for 32-bit. I'm just saying that it comes with a lot of complexity to achieve that, and I don't think it's justified. All the pointer forms are the Python-interaction ones already, which means the performance is already going to be naff.

crates/circuit/src/packed_instruction.rs
jakelishman: If you do keep these and want them private to the file, can you make them stand-alone functions (not `impl`) and use the `from` and `into` macro attributes to do it instead? Probably just minor style, but it'd make it easier to parse for me.

crates/circuit/src/packed_instruction.rs
jakelishman: I can buy that it isn't UB, though I don't think it's because of 0 initialisation (the u64 of `self` here isn't 0 - there's no `new` to be called, since we're interpreting a union field). If it's not UB, it's because the `StandardGate` type isn't constructed by the (internally generated) mask and shift of the getter, but the definition of the `struct` certainly makes it _look_ like you should think about the value as having been constructed.

I think I was wrong initially and it's not UB, but I'm still concerned about how non-obvious it is without having really thought about how `bitfield_struct` must work internally, which is more mental load.

crates/circuit/src/operations.rs
jakelishman: If it's needed to export for Python, it's fine (though if it's public to Python, is there no chance that it's useful elsewhere in Rust too?). I mostly just knee-jerk against `pub(crate)`, I think, because it's caused us a fair amount of churn in the last few months.

crates/circuit/src/packed_instruction.rs
jakelishman: Yeah, I'm onboard with any potential modification to `StandardGate` being done in a separate PR - if nothing else, it'll let us see if the change is actually worth it.

When I wrote the initial PR, I think my thought process was "a bitshift will be so fast we don't notice it", and "we might need those padding bits later". Certainly worth challenging those assumptions - I didn't test them.

crates/circuit/src/packed_instruction.rs
jakelishman: With points 1 and 2: my main problem is that it's not clear at the site of definition of `StandardInstruction` that these needs to be upheld, which I imagine is how `Barrier(usize)` ended up like the PR state in the first place.

I think something like a trait implementation of the packing might help that a bit, because then the trait defines the interface the type is required to fulfil, and whether the type fulfils them can be audited and understood more locally. The need to have the unsafe bit saying "the low three bits must be 0" isn't super ideal, but it reduces the amount of co-operation we have to enforce.

crates/circuit/src/circuit_instruction.rs
jakelishman: I can only speak to my interpretation, but to me, the mismatched attributed names jumped out as odd.  I get that it's not super pleasant with needing the `StandardInstruction`/`StandardInstructionType` split anyway, and I don't feel strongly about which way you want to keep it.

kevinhartman:
Thanks for the review! I ended up reworking this a bit to address your concerns:

### Stuff that looks like UB
I'm still using `#[bitfield]` from the bitfield-struct crate for standard gates and instructions, but only as an encoding implementation detail of their respective private modules; the `PackedOperation` itself is now back to being a wrapper around an unsigned int (`u64` now), and the bitfield structs are used just as internal helpers. I've also switched to just using built-in types for their fields, and using `bytemuck` to do any sort of checked conversions. It should also be very easy for us to migrate away from bitfield-struct if we need to in the future since it's not part of `PackedOperation`'s type hierarchy.

Also, the pointer types no longer use bitfields.

###  Localization of conversions
I've created private modules `standard_gate`, `standard_instruction`, and `pointers` to fully encapsulate packing/unpacking of their corresponding types. This localizes all of the conversion stuff and makes `PackedOperation`'s impl a good bit simpler. I ended up using `From` and `TryFrom` for packing/unpacking, which I believe works nicely since these conversions appear useful globally as well. Let me know if you think we should prefer a separate trait(s).

---

I'll do another pass this week over your review comments to make sure I haven't missed any.

crates/circuit/src/operations.rs
```diff
@@ -249,14 +260,230 @@ impl Operation for OperationRef<'_> {
...
+    Reset = 3,
+}
+
+unsafe impl ::bytemuck::CheckedBitPattern for StandardInstructionType {
+    type Bits = u8;
+
+    fn is_valid_bit_pattern(bits: &Self::Bits) -> bool {
+        *bits < 4
+    }
+}
```
jakelishman, line 350: Any chance you've thought of a nicer way to do this?  I know we've got a few instances of this kind of check in the library by now, and I'm getting more nervous about making mistakes in them, especially since we already did it once).

No need to change this PR about it (and if we do have any ideas, let's do it in a follow-up anyway), just asking if anything's occurred to you?

crates/circuit/src/dag_circuit.rs
```diff
@@ -7064,21 +7069,17 @@ mod test {
 
     macro_rules! measure {
```
jakelishman, line 7130: Completely besides the point of this PR, but do you know why these are macros?  I don't remember seeing them before, and they just look like regular functions.

crates/circuit/src/operations.rs
```diff
@@ -249,14 +260,230 @@ impl Operation for OperationRef<'_> {
...
+
+impl fmt::Display for DelayUnit {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        write!(
+            f,
+            "{}",
+            match self {
+                DelayUnit::NS => "ns",
+                DelayUnit::PS => "ps",
+                DelayUnit::US => "us",
```
jakelishman, line 300: It would be nice (if it doesn't break anything) to return "µs" rather than "us" if possible, since that's the more correct spelling.

crates/circuit/src/operations.rs
```diff
@@ -249,14 +260,230 @@ impl Operation for OperationRef<'_> {
...
+    }
+}
+
+impl<'py> FromPyObject<'py> for DelayUnit {
+    fn extract_bound(b: &Bound<'py, PyAny>) -> Result<Self, PyErr> {
+        let str: String = b.extract()?;
+        Ok(match str.as_str() {
+            "ns" => DelayUnit::NS,
+            "ps" => DelayUnit::PS,
+            "us" => DelayUnit::US,
```
jakelishman, line 315: Here I'm relatively sure we need to accept "µs", because that's a valid spelling in Python space.

qiskit/circuit/measure.py
```diff
@@ -16,18 +16,27 @@
...
             label: optional string label for this instruction.
         """
         super().__init__("measure", 1, 1, [], label=label, duration=duration, unit=unit)
 
+    def __init_subclass__(cls, **kwargs):
+        super().__init_subclass__(**kwargs)
+        # Subclasses of Measure are not "standard", so we set this to None to
+        # prevent the Rust code from treating them as such.
+        cls._standard_instruction_type = None
+
```
jakelishman, line 39: I think we're going to have to revisit this in the future to more properly support specialised subclasses of standard instructions - we're almost certain to miss things in the compiler if we're encoding two different things that are actually measures in two very different forms (`StandardInstruction` and `Instruction`).  That said, we _already_ have that problem with the subclassing, so I don't think this isn't going to cause any new problems.

I mostly just mean that I think the need to do this is indicative that we don't have the story properly formalised yet, which was true before anyway.

crates/circuit/src/packed_instruction.rs
```diff
@@ -58,203 +60,380 @@ unsafe impl ::bytemuck::NoUninit for PackedOperationType {}
...
+    impl From<StandardGate> for PackedOperation {
+        fn from(value: StandardGate) -> Self {
+            Self(
+                StandardGateBits::new()
+                    .with_discriminant(bytemuck::cast(PackedOperationType::StandardGate))
+                    .with_standard_gate(bytemuck::cast(value))
+                    .into_bits(),
+            )
+        }
     }
```
jakelishman: I'd be a little more comfortable if we make any changes to the representation of `StandardGate` in a separate PR so we can more properly evaluate its performance impact.  That said, if you're able to run the full benchmark suite and we show zero regressions, I'm ok to merge with this in place rather than separating it out now that it's written.

If there's any questions about regressions on this PR, even if we think we can explain them or they're prima facie not related to `StandardGate`, I'd rather separate this off just to be sure.

crates/circuit/src/packed_instruction.rs
```diff
@@ -58,203 +60,380 @@ unsafe impl ::bytemuck::NoUninit for PackedOperationType {}
...
-    /// low pointer bits that are guaranteed to be 0 by alignment, so no shifting is required.
-    const POINTER_MASK: usize = usize::MAX ^ Self::DISCRIMINANT_MASK;
-
-    /// Extract the discriminant of the operation.
-    #[inline]
-    fn discriminant(&self) -> PackedOperationType {
-        ::bytemuck::checked::cast((self.0 & Self::DISCRIMINANT_MASK) as u8)
-    }
+/// A private module to encapsulate the encoding of [StandardGate].
+mod standard_gate {
```
jakelishman, line 134: Minor question, but what encapsulation do you mean here?  The trait implementations aren't private, because visibility is tied to the trait and the types involved (and everything here is public), and the structs aren't public because they're not exported anyway.

I'm fine if it's just for logical grouping within the file, I'm just confused by the comment.

crates/circuit/src/packed_instruction.rs
```diff
@@ -58,203 +60,380 @@ unsafe impl ::bytemuck::NoUninit for PackedOperationType {}
...
+        discriminant: u8,
+        #[bits(5)]
+        _pad0: u8,
+        #[bits(8)]
+        standard_instruction: u8,
+        #[bits(16)]
+        _pad1: u32,
+        #[bits(32)]
+        payload: u32,
+    }
```
jakelishman, line 201: This does look a lot neater now, thank you - I guess some of it is us dropping 32-bit support, but this removes all my hang-ups about the prior implementations.

crates/circuit/src/packed_instruction.rs
```diff
@@ -58,203 +60,380 @@ unsafe impl ::bytemuck::NoUninit for PackedOperationType {}
...
+    const POINTER_MASK: u64 = !PackedOperation::DISCRIMINANT_MASK;
+
+    /// A private trait used to associate a [PackedOperationType] with
+    /// a supported pointer type.
+    trait PackablePointer {
+        const OPERATION_TYPE: PackedOperationType;
+    }
+
+    macro_rules! impl_packable_pointer {
+        ($type:ty, $operation_type:path) => {
```
jakelishman: More strictly, I think `$operation_type` is meant to be an `expr` not a `path` (since we're referring to an enum variant as a value of a type, and it's just coincidence that that happens to be a valid path), but it hardly matters for our limited use-cases here.

crates/circuit/src/packed_instruction.rs
```diff
@@ -58,203 +60,380 @@ unsafe impl ::bytemuck::NoUninit for PackedOperationType {}
...
+            }
+
+            match self.discriminant() {
+                PackedOperationType::StandardGate | PackedOperationType::StandardInstruction => (),
+                PackedOperationType::PyGate => drop_pointer_as::<PyGate>(self),
+                PackedOperationType::PyInstruction => drop_pointer_as::<PyInstruction>(self),
+                PackedOperationType::PyOperation => drop_pointer_as::<PyOperation>(self),
+            }
+        }
+    }
```
jakelishman: I think this would be better in the main scope, since the `Drop` impl is a function of the entirety of `PackedOperation`, it's just coincidental (of sorts) that the main logic of it is tied to the pointer variants.  With `UnitaryGate`, I'm strongly expecting that `StandardGate` or `StandardInstruction` gains a variant that's also `Drop` in the near future (likely an owned heap-allocated matrix), so I'd rather not have to disentangle the module-level encapsulation at that point, since that PR will already be complex.

crates/circuit/src/packed_instruction.rs
```diff
@@ -58,203 +60,380 @@ unsafe impl ::bytemuck::NoUninit for PackedOperationType {}
...
+
+    impl<T: PackablePointer> From<Box<T>> for PackedOperation {
+        fn from(value: Box<T>) -> Self {
+            let discriminant = T::OPERATION_TYPE as u64;
+            let ptr = NonNull::from(Box::leak(value)).cast::<()>();
+            let addr = ptr.as_ptr() as u64;
+            assert!((addr & PackedOperation::DISCRIMINANT_MASK == 0));
+            Self(discriminant | addr)
         }
     }
```
jakelishman: Given we're already doing all this by macro invocation anyway, could it be slightly cleaner to avoid the trait and just code-gen all the impl blocks anyway?  It would probably be slightly neater dev experience, since code completions (and hypothetical dev docs) would spell out all the `From<Box<PyGate>>` etc implementations explicitly, rather than it being hidden behind an indirect `trait` that wouldn't be documented, and (iiuc) is mostly just to "store" the result of the `OPERATION_TYPE` binding.

The trait lets us operate generically, except it's private and encapsulated to a module where we code-gen it anyway, so we could just move it all into code gen, and then the private `trait` doesn't leak out into public blanket impls.

I guess the problem you were trying to solve is with `try_pointer`, since you're trying to avoid ever having a type-erased `NonNull<()>` that we were using before.  Personally I wouldn't mind just having to be a little bit more explicit / repetitive in the `drop` impl by repeating the pointer-cast code (or tbh using the `NonNull<()>`-and-cast approach just localised to that function), in favour of getting that `impl` into the top-level scope.

jakelishman (commented):
Thanks for this - this latest implementation looks very nice and clean to me.  I have a couple of questions about the private modules inside `packed_instruction`, but they're relatively minor.  In general for the implementation I'm happy with this and would like to get it merged sooner rather than later.

It would be really good to get a benchmark comparison for this before we merge, to check there's no expected problems.

jakelishman:
Ah, I see that #13759 is actually built on top of this (I hadn't looked at it properly yet), so my concerns about the `Drop` impl aren't really realised in the same way.

Putting the `Drop` impl inside `mod pointer {}` still _feels_ wrong to me, and I'd rather we didn't do it, but given that the way `UnitaryGate` is implemented in Rust doesn't actually have a problem with it, it's not as critical to change.

crates/circuit/src/packed_instruction.rs
jakelishman: (as it turns out, #13759 did things a bit differently to how I expected, so the worry about it is unfounded, though the other component about the encapsulation of `Drop` into "pointer-y stuff" still feels uncomfortable to me)

crates/circuit/src/operations.rs
kevinhartman: The only thing that has _loosely_ occurred to me here (other than using nightly features) is to define these enum types with some kind of recursive macro that generates each variant's index and implements `CheckedBitPattern` after visiting all members. It feels like that would be possible, but I'm not well-versed in Rust macros yet, so I could be wrong.

crates/circuit/src/dag_circuit.rs
kevinhartman: I don't fully recall why I did that originally, but it may have been a readability thing. To me, these being macros communicate that they're just shortcuts for something you could do manually inline, and the distinct syntax highlighting makes it easy to see which operations are being applied during the test. I don't have a strong preference, though.

crates/circuit/src/operations.rs
kevinhartman: For better or (probably) worse, it at least looks to me like it's not valid for our Python-space `Delay`:

https://github.com/Qiskit/qiskit/blob/main/qiskit/circuit/delay.py#L35

crates/circuit/src/operations.rs
kevinhartman: I'm currently using `to_string` to convert from this enum back to Python space, so if we make the Rust `Display` look nice (using "µs"), then I'll need to write a custom `IntoPyObject`. Maybe that's the right thing to do anyway since we could use the string interner...

crates/circuit/src/operations.rs
jakelishman: oh, in that case ignore this comment and the previous one - I think I must have mixed `Delay` up with something else

crates/circuit/src/operations.rs
jakelishman: ignore this comment chain - I was mistaken about Python space

crates/circuit/src/operations.rs
jakelishman: yeah no worries - I also had no better ideas, I'm just really hoping someone comes up with something workable haha.

We _could_ write a proc macro to do it, but I don't know what effect that'll have on our build times.  I suppose we're already in that boat via PyO3?

crates/circuit/src/dag_circuit.rs
jakelishman: fair enough - it confused me a bit because I assume that a macro means there's something that _can't_ be done with a regular function call, but it's not a huge deal anyway

crates/circuit/src/packed_instruction.rs
jakelishman: (follow-up to the last sentence): but if it's _super_ gross any other way, then leaving it here is ok.  If we did move everything into code-gen, we could potentially remove the inline module, since there'd be nothing to encapsulate anymore, and then `Drop` wouldn't be in a submodule either.

I do like the nice grouping in the modules in principle, the only reason I'm grating a bit is that `Drop` feels (quite strongly to me) like it should be top-level

crates/circuit/src/packed_instruction.rs
kevinhartman: I can remove the comment since it's likely self-documenting anyway and therefore confusing to readers. I really only meant that `StandardInstructionBits` and our reliance on `bitfield_struct` is private to the inline `standard_gate` module. I was trying to emphasize this to show that the new implementation shouldn't lock us in to using an external crate, if it becomes inconvenient.

crates/circuit/src/operations.rs
kevinhartman: I suppose this will just go away anyway after Matt's removal of `ExtraInstructionAttributes`. But I suppose it would be better to skip it for `Delay`. I can do it for this PR, or we can just assume the unnecessary attributes will be dropped once that merges.

crates/circuit/src/packed_instruction.rs
kevinhartman: Done in 1c89bffd4bfcfb3d7f32b6fbf4ce1f2838392fa7

crates/circuit/src/packed_instruction.rs
kevinhartman: Hopefully addressed in e89cf473620d469ea7ee4547155e39da75ba07be.

Specifically, I've switched to codegen for the `From<Box<...>>` impls, and I've moved the `Drop` impl for `PackedOperation` back out to the root. I did this by making the `PackablePointer` type public (though still only visible to this file) and adding a `drop_packed` function to it to be called by the outer `Drop`. Hopefully that's good enough at least for the moment 🙂.

crates/circuit/src/operations.rs
kevinhartman: It's not _supposed_ to be useful outside of this crate (and Python where we use it to avoid an `isinstance` as an optimization), but I wouldn't be surprised if it becomes useful for the C API stuff eventually. I can make it `pub` if you prefer, just let me know!

crates/circuit/src/operations.rs
kevinhartman: This was done in 1f2c181c

crates/circuit/src/packed_instruction.rs
kevinhartman: Done in 93317388556bb4a87f4a205f1808c51dadfe80a0.

crates/circuit/src/packed_instruction.rs
kevinhartman: I didn't like it either, but couldn't think of an easy alternative at the time.

Resolved by e89cf473620d469ea7ee4547155e39da75ba07be

kevinhartman:
Thanks for the review!

> Putting the Drop impl inside mod pointer {} still feels wrong to me, and I'd rather we didn't do it, but given that the way UnitaryGate is implemented in Rust doesn't actually have a problem with it, it's not as critical to change.

As mentioned in a few of my recent comment replies, this should be resolved now 🙂.

Separately, I'll run some benchmarks for this against `main`, and then against a copy that reverts the layout change to a packed `StandardGate`. I sort of doubt the compiled output will be any different, tbh, but you're right we should at least make sure it isn't _worse_.

jakelishman (commented):
Ah nice nice nice, yeah, that last change to the `Drop` is nice - pointer-y stuff can stay in the pointer code group, but `Drop` itself isn't hidden.

Thanks for all this - I think it's good just to verify with the benchmarks, but assuming no regressions, I'm happy to merge this now. I'll tick once we've got the benchmarks.

kevinhartman:
There were a few benchmarks that got worse, so I switched us back to using the same padding for `StandardGate` for the sake of isolation, even though I don't believe the pad difference is responsible. These benchmarks compare the merge-base of `main` and this branch, before I added a few more `#[inline]` directives (these ended up doing nothing for the smaller sample of tests I ran locally, and I'd rather not run all the circuit construction and pass tests again).

Here's what improved:

| Change   | Before [8095ace7] <standard-op^2>   | After [c85be7a9] <standard-op~1>   |   Ratio | Benchmark (Parameter)                                                                        |
|----------|-------------------------------------|------------------------------------|---------|----------------------------------------------------------------------------------------------|
| -        | 44.6±3μs                            | 39.0±1μs                           |    0.87 | circuit_construction.CircuitConstructionBench.time_circuit_extend(2, 128)                    |
| -        | 56.1±5μs                            | 45.2±2μs                           |    0.81 | circuit_construction.CircuitConstructionBench.time_circuit_extend(8, 128)                    |
| -        | 40.3±1ms                            | 35.9±0.7ms                         |    0.89 | passes.MultiQBlockPassBenchmarks.time_collect_multiq_block(5, 1024, 1)                       |
| -        | 736±20ms                            | 662±7ms                            |    0.9  | passes.MultipleBasisPassBenchmarks.time_optimize_1q_commutation(14, 1024, ['u', 'cx', 'id']) |
| -        | 37.5±0.7μs                          | 34.0±0.7μs                         |    0.91 | passes.PassBenchmarks.time_remove_diagonal_gates_before_measurement(5, 1024)                 |

And here's what got worse:

| Change   | Before [8095ace7] <standard-op^2>   | After [c85be7a9] <standard-op~1>   |   Ratio | Benchmark (Parameter)                                           |
|----------|-------------------------------------|------------------------------------|---------|-----------------------------------------------------------------|
| +        | 40.7±2ms                            | 47.2±2ms                           |    1.16 | mapping_passes.PassBenchmarks.time_apply_layout(14, 1024)       |
| +        | 52.1±2μs                            | 61.7±1μs                           |    1.18 | passes.PassBenchmarks.time_remove_barriers(14, 1024)            |
| +        | 64.1±0.6μs                          | 75.1±3μs                           |    1.17 | passes.PassBenchmarks.time_remove_barriers(20, 1024)            |
| +        | 32.7±0.2μs                          | 36.8±0.4μs                         |    1.12 | passes.PassBenchmarks.time_remove_barriers(5, 1024)             |
| +        | 1.55±0.03ms                         | 1.80±0.05ms                        |    1.16 | passes.PassBenchmarks.time_remove_reset_in_zero_state(14, 1024) |
| +        | 634±10μs                            | 713±70μs                           |    1.12 | passes.PassBenchmarks.time_remove_reset_in_zero_state(5, 1024)  |
| +        | 420±20μs                            | 531±7μs                            |    1.27 | passes.PassBenchmarks.time_unroll_3q_or_more(5, 1024)           |

I believe the passes that have gotten worse are all Python-based and do things like iterate over the DAG (which requires Python `Barrier`, `Delay`, `Reset`, and `Measure` to be re-inflated). So I don't think there's anything to be concerned about here, though I'm happy to dig further if anyone feels differently.

jakelishman:
Yeah I'd agree - all those passes that got worse look to be because we've just shifted the Python-space construction cost into the passes themselves.  The total time is probably the same, just some cost moved from setup to benchmark.  I also highly doubt it's anything to do with `StandardGate` changes, but since it's not proven safe, it's good to move it, and we can just do that move in a separate PR if you still want it, where it'd be easier to revert if the overnight benchmarking shows a regression.

jakelishman (approved):
Thanks for all the work on this!

PR #13494, opened 2024-11-26
fixing docstrings for PauliEvolutionGate plugins

alexanderivrii:
### Summary

Matthew has noticed that the descriptions for the two PauliEvolutionGate plugins were swapped -- this commit fixes this. In addition, Matthew commented that it's better to provide a link to the rustiq-core library than to say the "rustiq-core library", as this probably doesn't make too much sense for the users.

### DIFF
--- a/qiskit/transpiler/passes/synthesis/hls_plugins.py
+++ b/qiskit/transpiler/passes/synthesis/hls_plugins.py
@@ -263,11 +263,11 @@ Pauli Evolution Synthesis
       - Targeted connectivity
     * - ``"rustiq"``
       - :class:`~.PauliEvolutionSynthesisRustiq`
-      - use a diagonalizing Clifford per Pauli term
+      - use the synthesis method from https://github.com/smartiel/rustiq-core
       - all-to-all
     * - ``"default"``
       - :class:`~.PauliEvolutionSynthesisDefault`
-      - use ``rustiq_core`` synthesis library
+      - use a diagonalizing Clifford per Pauli term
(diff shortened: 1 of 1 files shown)


qiskit/transpiler/passes/synthesis/hls_plugins.py
```diff
@@ -263,11 +263,11 @@
       - Targeted connectivity
     * - ``"rustiq"``
       - :class:`~.PauliEvolutionSynthesisRustiq`
-      - use a diagonalizing Clifford per Pauli term
+      - use the synthesis method from https://github.com/smartiel/rustiq-core
       - all-to-all
```
alexanderivrii: I am not 100% sure that this is the right way to include a link. Should I do something like `<...>__`?

qiskit/transpiler/passes/synthesis/hls_plugins.py
```diff
@@ -263,7 +263,7 @@
       - Targeted connectivity
     * - ``"rustiq"``
       - :class:`~.PauliEvolutionSynthesisRustiq`
-      - use the synthesis method from https://github.com/smartiel/rustiq-core
+      - use the synthesis method from `this quantum circuit synthesis library <https://github.com/smartiel/rustiq-core>`_
       - all-to-all
```
ShellyGarion, line 268: perhaps say:
this quantum circuit synthesis library --> rustiq circuit synthesis library

ShellyGarion (approved):

PR #13503, opened 2024-11-28
Fix macOS deployment target config in cibuildwheel

mtreinish:
### Summary

Looking at the 1.3.0 release build the build failed during the audit stage checking the shared object file that was built. This was failing because the macOS target version in the binary was 10.12 but the environment variable to override the default target of 10.9 was not getting set properly. We set this in the pyproject.toml but it is getting overridden by the PGO settings via CIBW_ENVIRONMENT in the github actions config. If you look at the recent builds we were publishing 10.9 target wheels on pypi despite documenting we target 10.12. The CI job logs also show that env variable is never being set during the cibuildwheel process.

The release of rust 1.83 in the last 24 hrs probably is what changed the output here causing the failure, although rust was supposed to raise it's minimum target to 10.12 in 1.74 [1] something probably changed in the release today which is tripping up the audit now.

This PR updates the CI configuration as was done to hotfix the deployment job temporarily during the 1.3.0 release process [2] to ensure we're setting our macOS deployment target correctly on x86_64 macOS builds to 10.12.

### Details and comments

[1] https://blog.rust-lang.org/2023/09/25/Increasing-Apple-Version-Requirements.html
[2] https://github.com/Qiskit/qiskit/actions/runs/12075613320

### DIFF
--- a/.github/workflows/wheels-build.yml
+++ b/.github/workflows/wheels-build.yml
@@ -121,6 +121,7 @@ jobs:
           cat >>"$GITHUB_ENV" <<EOF
           CIBW_BEFORE_BUILD=bash ./tools/build_pgo.sh $PGO_WORK_DIR $PGO_OUT_PATH
           CIBW_ENVIRONMENT=RUSTUP_TOOLCHAIN=stable RUSTFLAGS='-Cprofile-use=$PGO_OUT_PATH -Cllvm-args=-pgo-warn-missing-function'
+          CIBW_ENVIRONMENT_MACOS=MACOSX_DEPLOYMENT_TARGET='10.12' RUSTUP_TOOLCHAIN=stable RUSTFLAGS='-Cprofile-use=$PGO_OUT_PATH -Cllvm-args=-pgo-warn-missing-function'
           CIBW_ENVIRONMENT_LINUX=RUSTUP_TOOLCHAIN=stable RUSTFLAGS='-Cprofile-use=$PGO_OUT_PATH -Cllvm-args=-pgo-warn-missing-function' PATH="\$PATH:\$HOME/.cargo/bin" CARGO_NET_GIT_FETCH_WITH_CLI="true"
           EOF
         env:
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -184,7 +184,6 @@ environment = 'PATH="$PATH:$HOME/.cargo/bin" CARGO_NET_GIT_FETCH_WITH_CLI="true"
 repair-wheel-command = "auditwheel repair -w {dest_dir} {wheel} && pipx run abi3audit --strict --report {wheel}"
 
 [tool.cibuildwheel.macos]
-environment = "MACOSX_DEPLOYMENT_TARGET=10.12"
 repair-wheel-command = "delocate-wheel --require-archs {delocate_archs} -w {dest_dir} -v {wheel} && pipx run abi3audit --strict --report {wheel}"
 
 [tool.cibuildwheel.windows]


.github/workflows/wheels-build.yml
```diff
@@ -121,6 +121,7 @@ jobs:
           cat >>"$GITHUB_ENV" <<EOF
           CIBW_BEFORE_BUILD=bash ./tools/build_pgo.sh $PGO_WORK_DIR $PGO_OUT_PATH
           CIBW_ENVIRONMENT=RUSTUP_TOOLCHAIN=stable RUSTFLAGS='-Cprofile-use=$PGO_OUT_PATH -Cllvm-args=-pgo-warn-missing-function'
+          CIBW_ENVIRONMENT_MACOS=MACOSX_DEPLOYMENT_TARGET='10.12' RUSTUP_TOOLCHAIN=stable RUSTFLAGS='-Cprofile-use=$PGO_OUT_PATH -Cllvm-args=-pgo-warn-missing-function'
```
jakelishman, line 124: Can we set the `MACOSX_DEPLOYMENT_TARGET` variable in the general `CIBW_ENVIRONMENT` var to avoid duplicating the options?  I think it should just be safely ignored on Linux.

jakelishman (commented):
Other than the minor question (which isn't super important), this looks fine, except I think we still need it in `pyproject.toml` in case we trigger the build without PGO?

.github/workflows/wheels-build.yml
mtreinish: It actually would only get shared between macOS and windows since we have a separate linux one already. But it should be ignored on all the other platforms. I'll update it in a sec.

mtreinish:
Do we ever trigger a macOS build without PGO now? I guess you might be thinking longer term if we downgrade intel mac to tier 2 or 3 in the future?

.github/workflows/wheels-build.yml
jakelishman: Oh so we do - in that case I withdraw the concern. I'm so used to it being Windows that has the special case I misread.

Now you mention it, I remember reorganising the setup and environment here such that Windows no longer needed special handling.

jakelishman:
We don't, but the workflow that you're changing here has the option - the PGO environment is only compiled in conditionally, so it shouldn't be relied on or we'll have a problem down the line some way or another.

jakelishman:
From offline conversations: we realised in the Actions log that something is bumping the minimum `MACOSX_DEPLOYMENT_TARGET` to 10.13 for Intel Macs.  We think that this is rustc or Apple LLVM doing it, because Xcode 14 removed the ability to target 10.12 and before.  As long as the `macos-12` image remains available, it's possible for us to drop the version of Xcode to 13.4, which would let us target macOS 10.12, but the `macos-12` image is scheduled to become unavailable soon (see #13470), and that doesn't include any Xcode prior to 14 in its image.

This might mean that we need to put out a notice saying that Qiskit 1.3 to 1.4 _might_ not support macOS 10.12 - there's still a chance that it works correctly (and `delocate` seems to suggest we'll be fine with 1.3.0, at least), but the toolchains we have available to us no longer support targetting 10.12, so there's not a vast amount we can do.  We can potentially see if we can install Xcode 13.4 on a newer macOS image, but it might not be possible.

Fwiw, I think the command to switch Xcode version on the `macos-12` image should be
```
sudo xcode-select --switch /Applications/Xcode_13.4.1.app/Contents/Developer
```

Cryoris:
Note: Merged into #13470 via https://github.com/Qiskit/qiskit/pull/13470/commits/9c228445ed6d4d3537bc6512703b9ae8db447fb0.

kevinhartman:
Closing in favor of #13470

PR #13505, opened 2024-11-30
Fix clippy warnings with latest rust release

mtreinish:
### Summary

The recent rust release 1.83 introduced a few new clippy warnings and tweaks to existing checks that are triggering warnings in Qiskit's source. While we run clippy in CI it is with our MSRV of 1.70 for stability to avoid periodic CI outages when a new rust reversion is released. But for those who develop locally using the latest stable version of rust these warnings can make it hard to find issues in under development code. The new clippy rules are also real issues in the code that would be good to fix even if it's not a hard requirement yet.


jakelishman (approved):
As is probably fairly clear from this diff, I didn't know you could elide lifetimes in the `impl` introduction.

PR #13506, opened 2024-12-01
Remove condition/c_if, duration, and unit from instructions

### LINKED ISSUE #13700: Deprecate `propagate_condition` argument in `DAGCircuit` substitutions
Once #13506 merges, the `propagate_condition` argument to `substitute_node_with_dag` will no longer have any effect.  We will be able to remove the code, and to deprecate the argument.

We _could_ remove the argument directly in 2.0, but that would make it unnecessarily difficult to support 1.x and 2.y simultaneously, and it costs us near enough nothing to issue a Python-space deprecation on it.

### LINKED ISSUE #13859: Require a `target` in the timeline drawer
### What should we add?

https://github.com/Qiskit/qiskit/pull/13601 deprecates usage of the timeline drawer without a target. 2.0 should remove the non-target usage.

mtreinish:
### Summary

This commit removes the per instruction attributes for condition, duration, and unit. These attributes were deprecated in 1.3.0. Besides the label these attributes were the only mutable state for singleton instructions and removing them simplifies what needs to be tracked for instructions in both Python and rust. The associated methods and classes that were previously dependent on these attributes are also removed as they no longer serve a purpose. The removal of condition in particular removes a lot of logic from Qiskit especially from the transpiler because it was a something that needed to be checked outside of the normal data model for every operation.

This PR simplifies the representation of this extra mutable state in Rust by removing the `ExtraInstructionAttributes` struct and replacing it with a `Option<Box<String>>`. The `Option<Box<String>>` is used instead of the simpler `Option<String>` because this this reduces the size of the label field from 24 bytes for `Option<String>` to 8 bytes for `Option<Box<String>>` with an extra layer of pointer indirection and a second heap allocation. This will have runtime overhead when labels are set, but because the vast majority of operations don't set labels the tradeoff for optimizing for the None case makes more sense. Another option would have been to use `Box<str>` here which is the equivalent of `Box<[u8]>` (where `String` is the equivalent to `Vec<u8>`) and from a runtime memory tradeoff would be a better choice for this application if labels were commonly used as labels aren't growable. But that requires 16 bytes instead of 8 and we'd be wasting an additional 8 bytes for each instruction in the circuit which isn't worth the cost.

### Details and comments

TODO:

- [x] Fix scheduling pass updates
- [x] Remove remaining `c_if` usage in tests (there's >200 instances of it still)
- [x] Fix any failing tests not caused by the above 2 (hard to see with ~300-400 failures)
- [x] Finish writing release notes
- [x] Do some tuning/profiling to see if we can optimize things a bit more after the removal


nonhermitian:
Is there a replacement method for getting circuit timing if `duration` is removed?  It is somewhat critical to have a way to determine circuit timing for a variety of reasons.

mtreinish:
> Is there a replacement method for getting circuit timing if `duration` is removed? It is somewhat critical to have a way to determine circuit timing for a variety of reasons.

This PR doesn't remove `circuit.duration` (although it is deprecated and will likely be removed for 2.0 in a separate PR), this PR is only removing the per instruction `Instruction.duration` along with `.condition` and `.unit`. The timing data is owned by a backend/target so it didn't make sense to require label each individual instruction with timing information. Also this level of mutability in the data model is a constant source of bugs, and limits performance which is why it's being removed.

mtreinish:
Aside from the failing tests in `test.python.circuit.test_control_flow_builders.TestControlFlowBuilders` (which I'll need a little more time to fix) everything on the Qiskit side should be working now. Although there are more failures reported here, those are coming form aer's usage of the deprecated functionality. I pushed up a PR to aer to fix that here: https://github.com/Qiskit/qiskit-aer/pull/2301 we'll need that released or start building aer from source in CI to unblock this.

mtreinish:
I ran a quick asv run to see if there was a performance impact to this change, and it does speed things up by about 10%:
```

| Change   | Before [86f203d0] <drop-instruction-dead-weight~12^2>   | After [7b4fd309] <drop-instruction-dead-weight>   |   Ratio | Benchmark (Parameter)                                                         |
|----------|---------------------------------------------------------|---------------------------------------------------|---------|-------------------------------------------------------------------------------|
| -        | 9.83±0.1ms                                              | 8.85±0.1ms                                        |    0.9  | utility_scale.UtilityScaleBenchmarks.time_parse_qaoa_n100('ecr')              |
| -        | 9.86±0.09ms                                             | 8.78±0.1ms                                        |    0.89 | utility_scale.UtilityScaleBenchmarks.time_parse_qaoa_n100('cz')               |
| -        | 104±2ms                                                 | 93.0±0.6ms                                        |    0.89 | utility_scale.UtilityScaleBenchmarks.time_parse_qft_n100('cx')                |
| -        | 105±1ms                                                 | 92.8±1ms                                          |    0.89 | utility_scale.UtilityScaleBenchmarks.time_parse_qft_n100('cz')                |
| -        | 104±2ms                                                 | 92.7±0.9ms                                        |    0.89 | utility_scale.UtilityScaleBenchmarks.time_parse_qft_n100('ecr')               |
| -        | 33.9±0.4ms                                              | 30.0±0.2ms                                        |    0.88 | utility_scale.UtilityScaleBenchmarks.time_parse_square_heisenberg_n100('cx')  |
| -        | 33.7±0.5ms                                              | 29.7±0.3ms                                        |    0.88 | utility_scale.UtilityScaleBenchmarks.time_parse_square_heisenberg_n100('ecr') |
| -        | 34.0±0.1ms                                              | 29.6±0.2ms                                        |    0.87 | utility_scale.UtilityScaleBenchmarks.time_parse_square_heisenberg_n100('cz')  |

| Change   | Before [86f203d0] <drop-instruction-dead-weight~12^2>   | After [7b4fd309] <drop-instruction-dead-weight>   | Ratio   | Benchmark (Parameter)                                                                                  |
|----------|---------------------------------------------------------|---------------------------------------------------|---------|--------------------------------------------------------------------------------------------------------|
|          | 0                                                       | 0                                                 | n/a     | utility_scale.UtilityScaleBenchmarks.track_bvlike_depth('cx')                                          |
|          | 0                                                       | 0                                                 | n/a     | utility_scale.UtilityScaleBenchmarks.track_bvlike_depth('cz')                                          |
|          | 0                                                       | 0                                                 | n/a     | utility_scale.UtilityScaleBenchmarks.track_bvlike_depth('ecr')                                         |
|          | 10.4±0.04ms                                             | 10.9±0.1ms                                        | 1.04    | utility_scale.UtilityScaleBenchmarks.time_bvlike('cx')                                                 |
|          | 10.4±0.06ms                                             | 10.8±0.06ms                                       | 1.04    | utility_scale.UtilityScaleBenchmarks.time_bvlike('cz')                                                 |
|          | 115±1ms                                                 | 119±1ms                                           | 1.03    | transpiler_levels.TranspilerLevelBenchmarks.time_quantum_volume_transpile_50_x_20(1)                   |
|          | 19.2±0.2ms                                              | 19.6±0.4ms                                        | 1.02    | transpiler_levels.TranspilerLevelBenchmarks.time_transpile_from_large_qasm_backend_with_prop(2)        |
|          | 20.0±0.1ms                                              | 20.5±0.3ms                                        | 1.02    | transpiler_levels.TranspilerLevelBenchmarks.time_transpile_from_large_qasm_backend_with_prop(3)        |
|          | 18.1±0.1ms                                              | 18.3±0.2ms                                        | 1.01    | transpiler_levels.TranspilerLevelBenchmarks.time_transpile_from_large_qasm(2)                          |
|          | 37.4±0.3ms                                              | 37.7±0.3ms                                        | 1.01    | transpiler_levels.TranspilerLevelBenchmarks.time_transpile_from_large_qasm_backend_with_prop(1)        |
|          | 122±1ms                                                 | 123±0.6ms                                         | 1.01    | utility_scale.UtilityScaleBenchmarks.time_bv_100('cx')                                                 |
|          | 127±0.8ms                                               | 127±0.7ms                                         | 1.01    | utility_scale.UtilityScaleBenchmarks.time_bv_100('cz')                                                 |
|          | 126±0.6ms                                               | 127±0.5ms                                         | 1.01    | utility_scale.UtilityScaleBenchmarks.time_bv_100('ecr')                                                |
|          | 90.0±0.3ms                                              | 90.4±0.9ms                                        | 1.00    | transpiler_levels.TranspilerLevelBenchmarks.time_quantum_volume_transpile_50_x_20(0)                   |
|          | 181±0.9ms                                               | 181±2ms                                           | 1.00    | transpiler_levels.TranspilerLevelBenchmarks.time_quantum_volume_transpile_50_x_20(2)                   |
|          | 31.3±0.5ms                                              | 31.4±0.3ms                                        | 1.00    | transpiler_levels.TranspilerLevelBenchmarks.time_transpile_from_large_qasm(0)                          |
|          | 19.5±0.4ms                                              | 19.6±0.2ms                                        | 1.00    | transpiler_levels.TranspilerLevelBenchmarks.time_transpile_from_large_qasm(3)                          |
|          | 19.5±0.2ms                                              | 19.6±0.2ms                                        | 1.00    | transpiler_levels.TranspilerLevelBenchmarks.time_transpile_qv_14_x_14(0)                               |
|          | 34.7±0.4ms                                              | 34.6±0.3ms                                        | 1.00    | transpiler_levels.TranspilerLevelBenchmarks.time_transpile_qv_14_x_14(2)                               |
|          | 1404                                                    | 1404                                              | 1.00    | transpiler_levels.TranspilerLevelBenchmarks.track_depth_quantum_volume_transpile_50_x_20(0)            |
|          | 1403                                                    | 1403                                              | 1.00    | transpiler_levels.TranspilerLevelBenchmarks.track_depth_quantum_volume_transpile_50_x_20(1)            |
|          | 1323                                                    | 1323                                              | 1.00    | transpiler_levels.TranspilerLevelBenchmarks.track_depth_quantum_volume_transpile_50_x_20(2)            |
|          | 1296                                                    | 1296                                              | 1.00    | transpiler_levels.TranspilerLevelBenchmarks.track_depth_quantum_volume_transpile_50_x_20(3)            |
|          | 2705                                                    | 2705                                              | 1.00    | transpiler_levels.TranspilerLevelBenchmarks.track_depth_transpile_from_large_qasm(0)                   |
|          | 2005                                                    | 2005                                              | 1.00    | transpiler_levels.TranspilerLevelBenchmarks.track_depth_transpile_from_large_qasm(1)                   |
|          | 7                                                       | 7                                                 | 1.00    | transpiler_levels.TranspilerLevelBenchmarks.track_depth_transpile_from_large_qasm(2)                   |
|          | 7                                                       | 7                                                 | 1.00    | transpiler_levels.TranspilerLevelBenchmarks.track_depth_transpile_from_large_qasm(3)                   |
|          | 2705                                                    | 2705                                              | 1.00    | transpiler_levels.TranspilerLevelBenchmarks.track_depth_transpile_from_large_qasm_backend_with_prop(0) |
|          | 2005                                                    | 2005                                              | 1.00    | transpiler_levels.TranspilerLevelBenchmarks.track_depth_transpile_from_large_qasm_backend_with_prop(1) |
|          | 7                                                       | 7                                                 | 1.00    | transpiler_levels.TranspilerLevelBenchmarks.track_depth_transpile_from_large_qasm_backend_with_prop(2) |
|          | 7                                                       | 7                                                 | 1.00    | transpiler_levels.TranspilerLevelBenchmarks.track_depth_transpile_from_large_qasm_backend_with_prop(3) |
|          | 465                                                     | 465                                               | 1.00    | transpiler_levels.TranspilerLevelBenchmarks.track_depth_transpile_qv_14_x_14(0)                        |
|          | 336                                                     | 336                                               | 1.00    | transpiler_levels.TranspilerLevelBenchmarks.track_depth_transpile_qv_14_x_14(1)                        |
|          | 327                                                     | 327                                               | 1.00    | transpiler_levels.TranspilerLevelBenchmarks.track_depth_transpile_qv_14_x_14(2)                        |
|          | 272                                                     | 272                                               | 1.00    | transpiler_levels.TranspilerLevelBenchmarks.track_depth_transpile_qv_14_x_14(3)                        |
|          | 820±2ms                                                 | 818±9ms                                           | 1.00    | utility_scale.UtilityScaleBenchmarks.time_qv('cz')                                                     |
|          | 133±1ms                                                 | 133±1ms                                           | 1.00    | utility_scale.UtilityScaleBenchmarks.time_square_heisenberg('cx')                                      |
|          | 390                                                     | 390                                               | 1.00    | utility_scale.UtilityScaleBenchmarks.track_bv_100_depth('cx')                                          |
|          | 397                                                     | 397                                               | 1.00    | utility_scale.UtilityScaleBenchmarks.track_bv_100_depth('cz')                                          |
|          | 397                                                     | 397                                               | 1.00    | utility_scale.UtilityScaleBenchmarks.track_bv_100_depth('ecr')                                         |
|          | 300                                                     | 300                                               | 1.00    | utility_scale.UtilityScaleBenchmarks.track_circSU2_depth('cx')                                         |
|          | 300                                                     | 300                                               | 1.00    | utility_scale.UtilityScaleBenchmarks.track_circSU2_depth('cz')                                         |
|          | 300                                                     | 300                                               | 1.00    | utility_scale.UtilityScaleBenchmarks.track_circSU2_depth('ecr')                                        |
|          | 1607                                                    | 1607                                              | 1.00    | utility_scale.UtilityScaleBenchmarks.track_qaoa_depth('cx')                                            |
|          | 1622                                                    | 1622                                              | 1.00    | utility_scale.UtilityScaleBenchmarks.track_qaoa_depth('cz')                                            |
|          | 1622                                                    | 1622                                              | 1.00    | utility_scale.UtilityScaleBenchmarks.track_qaoa_depth('ecr')                                           |
|          | 1954                                                    | 1954                                              | 1.00    | utility_scale.UtilityScaleBenchmarks.track_qft_depth('cx')                                             |
|          | 1954                                                    | 1954                                              | 1.00    | utility_scale.UtilityScaleBenchmarks.track_qft_depth('cz')                                             |
|          | 1954                                                    | 1954                                              | 1.00    | utility_scale.UtilityScaleBenchmarks.track_qft_depth('ecr')                                            |
|          | 2709                                                    | 2709                                              | 1.00    | utility_scale.UtilityScaleBenchmarks.track_qv_depth('cx')                                              |
|          | 2709                                                    | 2709                                              | 1.00    | utility_scale.UtilityScaleBenchmarks.track_qv_depth('cz')                                              |
|          | 2709                                                    | 2709                                              | 1.00    | utility_scale.UtilityScaleBenchmarks.track_qv_depth('ecr')                                             |
|          | 462                                                     | 462                                               | 1.00    | utility_scale.UtilityScaleBenchmarks.track_square_heisenberg_depth('cx')                               |
|          | 462                                                     | 462                                               | 1.00    | utility_scale.UtilityScaleBenchmarks.track_square_heisenberg_depth('cz')                               |
|          | 462                                                     | 462                                               | 1.00    | utility_scale.UtilityScaleBenchmarks.track_square_heisenberg_depth('ecr')                              |
|          | 36.3±0.4ms                                              | 35.8±0.3ms                                        | 0.99    | transpiler_levels.TranspilerLevelBenchmarks.time_transpile_from_large_qasm(1)                          |
|          | 33.2±0.3ms                                              | 32.8±0.2ms                                        | 0.99    | transpiler_levels.TranspilerLevelBenchmarks.time_transpile_from_large_qasm_backend_with_prop(0)        |
|          | 57.7±0.4ms                                              | 57.0±0.6ms                                        | 0.99    | transpiler_levels.TranspilerLevelBenchmarks.time_transpile_qv_14_x_14(3)                               |
|          | 863±20ms                                                | 855±20ms                                          | 0.99    | utility_scale.UtilityScaleBenchmarks.time_circSU2('ecr')                                               |
|          | 271±2ms                                                 | 267±1ms                                           | 0.99    | utility_scale.UtilityScaleBenchmarks.time_qaoa('cx')                                                   |
|          | 356±2ms                                                 | 352±2ms                                           | 0.99    | utility_scale.UtilityScaleBenchmarks.time_qaoa('cz')                                                   |
|          | 336±3ms                                                 | 334±2ms                                           | 0.99    | utility_scale.UtilityScaleBenchmarks.time_qaoa('ecr')                                                  |
|          | 639±3ms                                                 | 629±3ms                                           | 0.99    | utility_scale.UtilityScaleBenchmarks.time_qft('cz')                                                    |
|          | 650±3ms                                                 | 646±3ms                                           | 0.99    | utility_scale.UtilityScaleBenchmarks.time_qft('ecr')                                                   |
|          | 690±8ms                                                 | 682±5ms                                           | 0.99    | utility_scale.UtilityScaleBenchmarks.time_qv('cx')                                                     |
|          | 809±8ms                                                 | 800±10ms                                          | 0.99    | utility_scale.UtilityScaleBenchmarks.time_qv('ecr')                                                    |
|          | 152±1ms                                                 | 150±2ms                                           | 0.99    | utility_scale.UtilityScaleBenchmarks.time_square_heisenberg('cz')                                      |
|          | 150±1ms                                                 | 149±0.9ms                                         | 0.99    | utility_scale.UtilityScaleBenchmarks.time_square_heisenberg('ecr')                                     |
|          | 26.2±0.2ms                                              | 25.7±0.2ms                                        | 0.98    | transpiler_levels.TranspilerLevelBenchmarks.time_transpile_qv_14_x_14(1)                               |
|          | 862±10ms                                                | 847±20ms                                          | 0.98    | utility_scale.UtilityScaleBenchmarks.time_circSU2('cx')                                                |
|          | 510±5ms                                                 | 500±3ms                                           | 0.98    | utility_scale.UtilityScaleBenchmarks.time_qft('cx')                                                    |
|          | 391±2ms                                                 | 380±2ms                                           | 0.97    | transpiler_levels.TranspilerLevelBenchmarks.time_quantum_volume_transpile_50_x_20(3)                   |
|          | 10.8±0.04ms                                             | 10.5±0.09ms                                       | 0.97    | utility_scale.UtilityScaleBenchmarks.time_bvlike('ecr')                                                |
|          | 874±20ms                                                | 852±20ms                                          | 0.97    | utility_scale.UtilityScaleBenchmarks.time_circSU2('cz')                                                |
|          | 68.7±0.3ms                                              | 63.6±0.3ms                                        | 0.93    | transpiler_levels.TranspilerLevelBenchmarks.time_schedule_qv_14_x_14(0)                                |
|          | 69.5±0.4ms                                              | 64.5±0.3ms                                        | 0.93    | transpiler_levels.TranspilerLevelBenchmarks.time_schedule_qv_14_x_14(1)                                |
|          | 9.79±0.1ms                                              | 8.90±0.05ms                                       | 0.91    | utility_scale.UtilityScaleBenchmarks.time_parse_qaoa_n100('cx')                                        |

PERFORMANCE INCREASED.
```

crates/circuit/src/dag_circuit.rs
```diff
@@ -2828,26 +2807,19 @@ def _format(operand):
     ///         and the order gets matched to the node wires by qargs first, then cargs, then
     ///         conditions.  If a dictionary, then a mapping of bits in the ``input_dag`` to those
     ///         that the ``node`` acts on.
-    ///     propagate_condition (bool): If ``True`` (default), then any ``condition`` attribute on
-    ///         the operation within ``node`` is propagated to each node in the ``input_dag``.  If
-    ///         ``False``, then the ``input_dag`` is assumed to faithfully implement suitable
-    ///         conditional logic already.  This is ignored for :class:`.ControlFlowOp`\\ s (i.e.
-    ///         treated as if it is ``False``); replacements of those must already fulfill the same
-    ///         conditional logic or this function would be close to useless for them.
```
jakelishman, line 2896: I commented elsewhere (#13700) that I feel we could leave this as a no-effect option for an extra major cycle at little-to-no cost to us, and might help ease the transition.  Just flagging it again in this review for visibility.

crates/circuit/src/dag_circuit.rs
```diff
@@ -3030,166 +3001,17 @@ def _format(operand):
             );
         }
 
-        let mut new_input_dag: Option<DAGCircuit> = None;
+        let new_input_dag: Option<DAGCircuit> = None;
         // It doesn't make sense to try and propagate a condition from a control-flow op; a
         // replacement for the control-flow op should implement the operation completely.
```
jakelishman, line 3084: The comment is surely out of date too if `propagate_condition` is gone?

crates/circuit/src/dag_circuit.rs
```diff
@@ -3263,36 +3092,34 @@ def _format(operand):
...
-                                    py_inst
-                                        .instruction
-                                        .setattr(py, "condition", new_condition)?;
-                                }
-                                OperationRef::Gate(py_gate) => {
-                                    py_gate.gate.setattr(py, "condition", new_condition)?;
-                                }
-                                OperationRef::Operation(py_op) => {
-                                    py_op.operation.setattr(py, "condition", new_condition)?;
+                            if old_inst.op.name() != "switch_case" {
```
jakelishman, line 3177: If you have to have this line anyway, feels like the `if switch` may as well be protected in it too, since checking for `op.control_flow()` is faster than a string comparison?

That said, switch can't have a `condition` - it had a property override to make it unsettable already.

crates/circuit/src/dag_circuit.rs
```diff
@@ -3318,38 +3145,33 @@ def _format(operand):
     ///     inplace (bool): Optional, default False. If True, existing DAG node
     ///         will be modified to include op. Otherwise, a new DAG node will
     ///         be used.
-    ///     propagate_condition (bool): Optional, default True.  If True, a condition on the
-    ///         ``node`` to be replaced will be applied to the new ``op``.  This is the legacy
-    ///         behaviour.  If either node is a control-flow operation, this will be ignored.  If
-    ///         the ``op`` already has a condition, :exc:`.DAGCircuitError` is raised.
```
jakelishman, line 3386: Same comment re #13700 about `propagate_condition`.

crates/circuit/src/dag_circuit.rs
```diff
@@ -5062,7 +4877,7 @@ impl DAGCircuit {
         qargs: &[Qubit],
         cargs: &[Clbit],
         params: Option<SmallVec<[Param; 3]>>,
-        extra_attrs: ExtraInstructionAttributes,
+        label: Option<String>,
```
jakelishman, line 5055: Since we have to `Box` it within the function anyway, does it make sense to accept `Option<Box<String>>` instead, or is that less ergonomic / less impactful?

crates/circuit/src/dag_circuit.rs
```diff
@@ -5306,33 +5117,35 @@ impl DAGCircuit {
...
-                }
-            }
-        }
 
         if let OperationRef::Instruction(inst) = op {
             let op = inst.instruction.bind(py);
             if inst.control_flow() {
+                let condition: Option<Bound<'_, PyAny>> = op.getattr(intern!(py, "condition")).ok();
+                if condition.is_some() {
+                    let condition = condition.unwrap();
```
jakelishman: This looks pretty funky - the `.ok()` is throwing away the potential Python-space exception, and `if condition.is_some() { let condition = condition.unwrap() }` would have been weird regardless.

crates/circuit/src/packed_instruction.rs
```diff
@@ -550,14 +549,9 @@ impl PackedInstruction {
...
-    #[inline]
-    pub fn condition(&self) -> Option<&Py<PyAny>> {
-        self.extra_attrs.condition()
-    }
-
     #[inline]
     pub fn label(&self) -> Option<&str> {
-        self.extra_attrs.label()
+        self.label.as_ref().map(|label| label.as_str())
     }
```
jakelishman, line 739: I feel like there's places in this PR where using this method might have ended up in cleaned code than accessing the property directly, but not worth going back and changing anything.

qiskit/circuit/controlflow/builder.py
```diff
@@ -284,7 +284,9 @@ def _copy_mutable_properties(self, instruction: Instruction) -> Instruction:
             The same instruction instance that was passed, but mutated to propagate the tracked
             changes to this class.
         """
-        instruction._condition = self._condition
+        cond = getattr(self, "_condition", None)
+        if cond is not None:
+            instruction._condition = cond
         return instruction
```
jakelishman: iirc, this method was only for `break_loop` and `continue_loop` in conjunction with `c_if`.  I _think_ it's dead code now?

qiskit/circuit/quantumcircuit.py
```diff
@@ -2085,13 +2085,7 @@ def recurse_block(block):
...
                 n_op = op
                 is_control_flow = isinstance(n_op, ControlFlowOp)
-                if (
-                    not is_control_flow
-                    and (condition := getattr(n_op, "_condition", None)) is not None
-                ):
-                    n_op = n_op.copy() if n_op is op and copy else n_op
-                    n_op.condition = variable_mapper.map_condition(condition)
-                elif is_control_flow:
+                if is_control_flow:
```
jakelishman: The `is_control_flow` variable is superfluous now - we can just inline the `isinstance` check into the `if`.

qiskit/circuit/duration.py
```diff
@@ -63,21 +63,6 @@ def convert_durations_to_dt(qc: QuantumCircuit, dt_in_sec: float, inplace=True):
...
-            raise CircuitError(f"Invalid time unit: '{operation.unit}'")
-
-        duration = operation.duration
-        if operation.unit != "s":
-            duration = apply_prefix(duration, operation.unit)
-
-        operation.duration = duration_in_dt(duration, dt_in_sec)
-        operation.unit = "dt"
-
     if circ.duration is not None and circ.unit != "dt":
```
jakelishman, line 66: Do we have tracking for a follow-up that removes circuit duration?  I couldn't immediately see one.

qiskit/circuit/instructionset.py
```diff
@@ -50,9 +49,9 @@ def __init__(  # pylint: disable=bad-docstring-quotes
 
                 .. note::
 
-                    The callback ``resource_requester`` is called once for each call to
-                    :meth:`.c_if`, and assumes that a call implies that the resource will now be
-                    used.  It may throw an error if the resource is not valid for usage.
+                    The callback ``resource_requester`` assumes that a call implies that the
+                    resource will now be used.  It may throw an error if the resource is not valid
+                    for usage.
```
jakelishman, line 55: I think the whole `resource_requester` argument is superfluous now and can be deleted, and the handling removed from `QuantumCircuit.append`.  Happy to wait for a follow-up for that.

qiskit/circuit/library/standard_gates/rxx.py
```diff
@@ -79,10 +79,12 @@ class RXXGate(Gate):
     _standard_gate = StandardGate.RXXGate
 
     def __init__(
-        self, theta: ParameterValueType, label: Optional[str] = None, *, duration=None, unit="dt"
+        self,
+        theta: ParameterValueType,
+        label: Optional[str] = None,
```
jakelishman: How did this diff get _longer_?  Maybe an errant comma tricked black?

qiskit/circuit/library/standard_gates/ry.py
```diff
@@ -55,10 +55,12 @@ class RYGate(Gate):
     _standard_gate = StandardGate.RYGate
 
     def __init__(
-        self, theta: ParameterValueType, label: Optional[str] = None, *, duration=None, unit="dt"
+        self,
+        theta: ParameterValueType,
+        label: Optional[str] = None,
```
jakelishman: More errant commas spoiling your diffstat ;)

qiskit/circuit/singleton.py
```diff
@@ -589,17 +583,17 @@ def stdlib_singleton_key(*, num_ctrl_qubits: int = 0):
 
     if num_ctrl_qubits:
 
-        def key(label=None, ctrl_state=None, *, duration=None, unit="dt", _base_label=None):
-            if label is None and duration is None and unit == "dt" and _base_label is None:
+        def key(label=None, ctrl_state=None, _base_label=None):
```
jakelishman: The `*` was still valid despite the removals (and probably right - `_base_label` shouldn't be exposed as settable to users).

qiskit/primitives/statevector_sampler.py
```diff
@@ -288,7 +288,4 @@ def _final_measurement_mapping(circuit: QuantumCircuit) -> dict[tuple[ClassicalR
 
 
 def _has_control_flow(circuit: QuantumCircuit) -> bool:
-    return any(
-        isinstance((op := instruction.operation), ControlFlowOp) or op._condition
-        for instruction in circuit
-    )
+    return any(isinstance(instruction.operation, ControlFlowOp) for instruction in circuit)
```
jakelishman: I think we allow `any(instruction.is_control_flow() for instruction in circuit)` (maybe without the function call)?

qiskit/qasm2/export.py
```diff
@@ -246,14 +246,13 @@ def _instruction_call_site(operation):
...
+    if cond is not None:
+        if not isinstance(cond[0], ClassicalRegister):
             raise QASM2ExportError(
                 "OpenQASM 2 can only condition on registers, but got '{operation.condition[0]}'"
             )
-        qasm2_call = (
-            f"if({operation._condition[0].name}=={operation._condition[1]:d}) " + qasm2_call
-        )
+        qasm2_call = f"if({cond[0].name}=={cond[1]:d}) " + qasm2_call
     return qasm2_call
```
jakelishman: I don't see how the `cond is not None` branch can ever be entered validly here - control-flow operations would already have exploded.

That said: I think we ought to improve the handling of `IfElseOp` in the OQ2 exporter such that it can successfully output to OQ2, otherwise there's no way for Qiskit to generate the `if (cr==0) h q[0];` stuff any more, even though we support it.

test/python/qasm2/test_export.py
```diff
@@ -68,9 +62,6 @@ def test_basic_output(self):
 barrier qr2[0],qr2[1];
 cx qr2[1],qr1[0];
 h qr2[1];
-if(cr==0) x qr2[1];
-if(cr==1) y qr1[0];
-if(cr==2) z qr1[0];
```
jakelishman, line 73: This removal is perhaps not great: Qiskit does still support conditionals, but this test is removing the tests of the OQ2 export on the grounds that we don't support `c_if` any more.

I think we need to get a feature in place that we export `IfElseOp` on register conditions to sequences of `if(...)` OQ2 instructions.  Can be done in a separate PR.

qiskit/qasm3/exporter.py
```diff
@@ -1006,13 +1006,14 @@ def build_current_scope(self) -> List[ast.Statement]:
...
                 statements.extend(nodes)
             else:
                 body = ast.ProgramBlock(nodes)
                 statements.append(
                     ast.BranchingStatement(
-                        self.build_expression(_lift_condition(instruction.operation._condition)),
+                        self.build_expression(_lift_condition(cond)),
                         body,
                     )
                 )
```
jakelishman: I think this code is dead now - we should already have left this bit when checking whether it's a control-flow op.

qiskit/qpy/binary_io/circuits.py
```diff
@@ -319,13 +319,9 @@ def _read_instruction(
...
-            warnings.warn(
-                f"The .condition attribute on {gate_name} will be loaded as an IfElseOp "
-                "starting in Qiskit 2.0",
-                FutureWarning,
-                stacklevel=3,
-            )
-        inst_obj.condition = condition
+            body = QuantumCircuit(qargs, cargs)
+            body.append(inst_obj, qargs, cargs)
+            inst_obj = IfElseOp(condition, body)
```
jakelishman: Could a `QPYWarning` to indicate the inaccurate recreation still be appropriate here?

qiskit/qpy/binary_io/circuits.py
```diff
@@ -421,13 +417,9 @@ def _read_instruction(
...
-                warnings.warn(
-                    f"The .condition attribute on {gate} will be loaded as an "
-                    "IfElseOp starting in Qiskit 2.0",
-                    FutureWarning,
-                    stacklevel=3,
-                )
-                gate = gate.c_if(*condition)
+                body = QuantumCircuit(qargs, cargs)
+                body.append(gate, qargs, cargs)
+                gate = IfElseOp(condition, body)
```
jakelishman: Same question about maintaining a `QPYWarning`?

qiskit/transpiler/passes/scheduling/asap.py
```diff
@@ -75,40 +75,9 @@ def run(self, dag):
...
-                        #
-                        t0c = max(t0q - self.conditional_latency, t0c)
-                    t1c = t0c + self.conditional_latency
-                    for bit in node.op.condition_bits:
-                        # Lock clbit until state is read
-                        idle_after[bit] = t1c
-                    # It starts after register read access
-                    t0 = max(t0q, t1c)
-                else:
-                    t0 = t0q
```
jakelishman: Looking at the drawings in these removals: I think there's maybe some documentation text in the docstring of `qiskit.transpiler` that might refer to scheduling conditions as well?

qiskit/transpiler/passes/scheduling/base_scheduler.py
```diff
@@ -271,7 +271,10 @@ def _get_node_duration(
             cal_key = tuple(indices), tuple(float(p) for p in node.op.params)
             duration = dag._calibrations_prop[node.op.name][cal_key].duration
         else:
-            duration = node.op.duration
+            try:
+                duration = self.durations.get(node.op, indices)
+            except TranspilerError:
+                duration = None
```
jakelishman: This feels not identical: what if `node.op` is a delay?

qiskit/transpiler/passes/scheduling/base_scheduler.py
jakelishman, line 1: In all scheduling passes and classes, is `conditional_latency` now a dead argument?  Also, at least _this_ class docstring (but I think others?) has large amounts of text referring to the conditions, which is obsolete now too.

qiskit/transpiler/passes/scheduling/padding/base_padding.py
```diff
@@ -57,6 +59,7 @@ class BasePadding(TransformationPass):
     def __init__(
         self,
         target: Target = None,
+        durations: InstructionDurations = None,
     ):
```
jakelishman, line 63: Adding this in and making it required is definitely a breaking API change against the deprecation policy - we're allowed to add it in, but the necessity makes it clear that there was not an alternative in place properly in Qiskit 1.x.

I'm worried that this could cause a lot more pain than we expected, owing to our unfamiliarity with the scheduling passes.  I get that it's only if `Target` wasn't passed, but still.

qiskit/transpiler/passes/scheduling/padding/pad_delay.py
```diff
@@ -56,7 +57,12 @@ class PadDelay(BasePadding):
     See :class:`BasePadding` pass for details.
     """
 
-    def __init__(self, fill_very_end: bool = True, target: Target = None):
+    def __init__(
+        self,
+        fill_very_end: bool = True,
+        target: Target = None,
+        durations: InstructionDurations = None,
+    ):
```
jakelishman, line 65: Same comment about adding `durations` to the interface here (though this is the real part of it, I guess, since this is the public interface).

qiskit/transpiler/passes/scheduling/scheduling/base_scheduler.py
```diff
@@ -56,29 +56,35 @@ def __init__(self, durations: InstructionDurations = None, target: Target = None
...
     ) -> int:
         """A helper method to get duration from node or calibration."""
         indices = [dag.find_bit(qarg).index for qarg in node.qargs]
 
-        if dag._has_calibration_for(node):
+        if node.name == "delay":
+            if self.durations.dt is None:
+                duration = self.durations._convert_unit(node.op.duration, node.op.unit, "s")
+            else:
+                duration = self.durations._convert_unit(node.op.duration, node.op.unit, "dt")
```
jakelishman: Where did this conversion logic come from?  Does it indicate that the `TimeUnitConversion` (or whatever it's called) has got gutted a bit _too_ much?  This code being here feels like it's moving logic around more than strictly necessary.

qiskit/transpiler/passes/scheduling/scheduling/base_scheduler.py
```diff
@@ -56,29 +56,35 @@ def __init__(self, durations: InstructionDurations = None, target: Target = None
...
-            dag.substitute_node(node, op, propagate_condition=False)
         else:
-            duration = node.duration
+            try:
+                if self.durations.dt is None:
+                    duration = self.durations.get(node.name, indices, unit="s")
+                else:
+                    duration = self.durations.get(node.name, indices)
+            except TranspilerError:
+                duration = None
```
jakelishman: Seems weird that `TranspilerError` would be raised here and not `KeyError`?

qiskit/transpiler/passes/utils/convert_conditions_to_if_ops.py
```diff
@@ -1,93 +0,0 @@
...
-    Clbit,
-    IfElseOp,
-    QuantumCircuit,
-)
-from qiskit.dagcircuit import DAGCircuit
-from qiskit.transpiler.basepasses import TransformationPass
-from qiskit.utils import deprecate_func
-
-
-class ConvertConditionsToIfOps(TransformationPass):
```
jakelishman, line 28: Similar to the `propagate_condition` thing, I wouldn't have minded if we'd just left this in as a no-op pass for the Qiskit 2.x series (since its primary use-case was already for code to start eagerly migrating forwards, which is still useful as a Qiskit 1.x compatibility layer), though handling a missing import is much easier than a missing function argument, so I'm not very bothered by it at all.

qiskit/transpiler/target.py
```diff
@@ -969,6 +970,18 @@ def __setstate__(self, state: tuple):
...
+
+        Args:
+            duration: The duration in seconds, such as in an :class:`.InstructionProperties`
+                field for an instruction in the target.
+
+        Returns
+            duration: The duration in units of dt
+        """
+        return duration_in_dt(duration, self.dt)
+
```
jakelishman: This would need a feature release note.  Perhaps it ought to be `seconds_to_dt` so it's obvious at the call site what unit it expects too, rather than being documentation only.

qiskit/visualization/timeline/core.py
jakelishman, line 1: I have no feel for the effects of the changes in this file, but since there's no maintainer of this visualiser, I think the best we can do is potentially just merge and hope - I don't have time to learn enough about the subsystem to properly review it.

test/python/circuit/classical/test_expr_constructors.py
```diff
@@ -14,67 +14,13 @@
...
+from qiskit.circuit import Clbit, ClassicalRegister
 from qiskit.circuit.classical import expr, types
 from test import QiskitTestCase  # pylint: disable=wrong-import-order
 
 
 @ddt.ddt
 class TestExprConstructors(QiskitTestCase):
-    def test_lift_legacy_condition(self):
-        cr = ClassicalRegister(3, "c")
-        clbit = Clbit()
```
jakelishman: All of these tests ought to stay - they're still valid and the function still exists.  `IfElseOp` and `WhileLoopOp` still allows their conditions to be in the legacy format, so the tests could swap to those, which is spiritually what the function is now for (or just write the conditions without an instruction).

jakelishman (commented):
Thanks for doing all this work.

This isn't a complete review; I haven't looked at the vast majority of the test files or the release note, but I did flick through all the code changes.  Elena's still assigned and should probably be the main reviewer, I just wanted to have a skim through myself so I knew what was going on (especially since I wasn't involved in the deprecation and didn't know what was happening there).

qiskit/transpiler/target.py
mtreinish: I thought in this context of `Target` all being in terms of seconds it was implicit. But I can change the name if you think it's more clear.

qiskit/transpiler/passes/scheduling/scheduling/base_scheduler.py
mtreinish: Yeah, I would have expected a key error, but it's what was expected so I just went with it than change more apis.

qiskit/transpiler/passes/scheduling/scheduling/base_scheduler.py
mtreinish: Heh, I didn't actually touch the pass much at all. But I think you might be right the lack of per instruction duration might be breaking that pass and that's what is requiring this logic here too. I'm surprised all its tests would pass if that was the case though.

qiskit/transpiler/passes/scheduling/padding/base_padding.py
mtreinish: Realistically this was just for the tests. In the real world this almost always will be a target. The only way to get an instruction durations in 2.0 with the removal of `BackendV1` is to either build it by hand (which is what our tests are doing), or call the method on the target to generate it from a target. Maybe I was just lazy but I didn't feel that I should rewrite nearly every scheduling test in this PR to avoid adding this option. I can mark it as private if you think that's better.

qiskit/transpiler/passes/scheduling/base_scheduler.py
mtreinish: It is, it was only used to handle adding a delay to account for `c_if` evaluation time. I can remove it if you think that's correct here. Or should I raise a deprecation warning.

qiskit/transpiler/passes/scheduling/asap.py
mtreinish: Oh, you're right there is a whole section on the logic around conditional latency and the scheduling passes. I will remove that.

qiskit/transpiler/passes/scheduling/base_scheduler.py
mtreinish: That's a good point, it passed all the tests so I missed that. But we should check if it's a delay and pull the duration from that.

That being said the legacy scheduling are staged to be removed too. But we should probably fix this until this module is deleted.

qiskit/qpy/binary_io/circuits.py
mtreinish: Yeah we probably should leave the warning in place to tell people we're transforming the circuit from the previous form. I'll restore it and change it to a `QPYWarning` from a future warning.

test/python/qasm2/test_export.py
mtreinish: I'll open an issue about this so we can track it after this merges.

qiskit/circuit/duration.py
mtreinish: I don't have a PR for it yet, I was waiting for #13783 to merge before pushing it. But I'll open a tracking issue.

1ucian0:
`on hold` until Aer release a PR that is blocking

mtreinish:
Aer 0.16.1 was released that included the fix removing on hold

crates/circuit/src/dag_circuit.rs
mtreinish: Done in: https://github.com/Qiskit/qiskit/pull/13506/commits/9bf0500580f92ef3f7156823cfe415ff690cadbb

crates/circuit/src/dag_circuit.rs
mtreinish: Removed in: https://github.com/Qiskit/qiskit/pull/13506/commits/9bf0500580f92ef3f7156823cfe415ff690cadbb

crates/circuit/src/dag_circuit.rs
mtreinish: Done in: https://github.com/Qiskit/qiskit/pull/13506/commits/9bf0500580f92ef3f7156823cfe415ff690cadbb

crates/circuit/src/packed_instruction.rs
mtreinish: To be honest I forgot we had the method when I was in the depths of rewriting things. I'm pretty sure I wrote the rust bits on a long flight and just was turning my brain off and mechanically fixing the compiler errors after changing the core representation.

qiskit/circuit/controlflow/builder.py
mtreinish: Yeah I can drop the method completely, you're correct. I'll push up the commit in a sec.

qiskit/qpy/binary_io/circuits.py
mtreinish: Added a warning for this in: https://github.com/Qiskit/qiskit/pull/13506/commits/cd28d592bde6ff9c332b85589472c1c64dd987c9 we were using `UserWarning` in other places for similar warnings.

qiskit/circuit/singleton.py
mtreinish: Fixed in: https://github.com/Qiskit/qiskit/pull/13506/commits/e3255269a20360874d50ccba02ed5d0fd3633564

qiskit/circuit/quantumcircuit.py
mtreinish: Simplified in: https://github.com/Qiskit/qiskit/pull/13506/commits/e3255269a20360874d50ccba02ed5d0fd3633564

qiskit/circuit/controlflow/builder.py
mtreinish: Done in: https://github.com/Qiskit/qiskit/pull/13506/commits/1ae2e9169a715966bf9643c4f3d6cf350d9fb2d2

test/python/circuit/classical/test_expr_constructors.py
mtreinish: Fixed in: https://github.com/Qiskit/qiskit/pull/13506/commits/9ad7e9eae0dd8aeacab4dffe22344f696569748c

qiskit/transpiler/target.py
mtreinish: Renamed in: https://github.com/Qiskit/qiskit/pull/13506/commits/78a7ee49b8b67e0c7a4776380fc4ace65e1ea419

qiskit/transpiler/passes/scheduling/asap.py
mtreinish: This should be fixed now. I think I pulled all of this out in some of my docs fixing efforts

qiskit/qasm2/export.py
mtreinish: Removed in: https://github.com/Qiskit/qiskit/pull/13506/commits/e3255269a20360874d50ccba02ed5d0fd3633564 I will open an issue to track adding IfElseOp -> qasm2 export path.

mtreinish: The tracking issue is: https://github.com/Qiskit/qiskit/issues/13848

test/python/qasm2/test_export.py
mtreinish: I opened an issue for this here: https://github.com/Qiskit/qiskit/issues/13848

qiskit/circuit/duration.py
mtreinish: Tracked in: https://github.com/Qiskit/qiskit/issues/13849

crates/circuit/src/dag_circuit.rs
```diff
@@ -2888,27 +2867,38 @@ def _format(operand):
...
-        propagate_condition: bool,
+        propagate_condition: Option<bool>,
     ) -> PyResult<Py<PyDict>> {
+        if propagate_condition.is_some() {
+            imports::WARNINGS_WARN.get_bound(py).call1((
+                intern!(
+                    py,
+                    concat!(
+                        "The propagate_condition argument is deprecated as of Qiskit 2.0.0.",
+                        "It will be removed in Qiskit 3.0.0.",
```
ElePT: When I started the review I was wondering why you were changing propagate condition from `true` or `false` to `None` (I hadn't reached this file yet). Then I read the docstring change saying that it doesn't work anymore. I think it's a good information to add to the warning (deprecated for me means everything stays as it is but will change in the future, but this is a change to a dummy argument):

```suggestion
                        "The propagate_condition argument is deprecated as of Qiskit 2.0.0.",
                        "It has no effect anymore and will be removed in Qiskit 3.0.0.",
```

crates/circuit/src/dag_circuit.rs
```diff
@@ -3090,166 +3079,17 @@ def _format(operand):
...
-                    &qubit_wire_map,
-                    &clbit_wire_map,
-                    &var_map,
-                )?
-            }
-        } else {
-            self.substitute_node_with_subgraph(
-                py,
-                node_index,
-                input_dag,
```
ElePT, line 3247: nice! 156 lines in one go

crates/circuit/src/dag_circuit.rs
```diff
@@ -3380,38 +3223,53 @@ def _format(operand):
...
-        propagate_condition: bool,
+        propagate_condition: Option<bool>,
     ) -> PyResult<Py<PyAny>> {
+        if propagate_condition.is_some() {
+            imports::WARNINGS_WARN.get_bound(py).call1((
+                intern!(
+                    py,
+                    concat!(
+                        "The propagate_condition argument is deprecated as of Qiskit 2.0.0.",
+                        "It will be removed in Qiskit 3.0.0.",
```
ElePT:
```suggestion
                        "The propagate_condition argument is deprecated as of Qiskit 2.0.0.",
                        "It has no effect anymore and will be removed in Qiskit 3.0.0.",
```

qiskit/circuit/classical/expr/constructors.py
```diff
@@ -63,21 +63,7 @@ def _coerce_lossless(expr: Expr, type: types.Type) -> Expr:
...
-            from qiskit.circuit import ClassicalRegister
-            from qiskit.circuit.library import HGate
-            from qiskit.circuit.classical import expr
-
-            cr = ClassicalRegister(2)
-            instr = HGate().c_if(cr, 3)
-
-            lifted = expr.lift_legacy_condition(instr.condition)
-    """
+    """Lift a legacy two-tuple equality condition into a new-style :class:`Expr`."""
```
ElePT, line 66: Does keeping `lift_legacy_condition` make sense now that the legacy conditions are gone? is there any other use case for it?

qiskit/circuit/controlflow/builder.py
```diff
@@ -205,10 +205,12 @@ class InstructionPlaceholder(Instruction, abc.ABC):
...
             qc.measure(0, 0)
-            qc.break_loop().c_if(0, 0)
+            with qc.if_test((0, 0)):
+                qc.break_loop()
 
-    since ``qc.break_loop()`` needs to return a (mostly) functional
-    :obj:`~qiskit.circuit.Instruction` in order for :meth:`.InstructionSet.c_if` to work correctly.
+    ``qc.break_loop()`` needs to return a (mostly) functional
+    :obj:`~qiskit.circuit.Instruction` in order for the historical ``.InstructionSet.c_if``
+    to work correctly.
```
ElePT: was this only necessary for c_if? or is it also important for if_test? If it was only for c_if I would at least speak about it in past tense, given that it's dead:

```suggestion
    ``qc.break_loop()`` needed to return a (mostly) functional
    :obj:`~qiskit.circuit.Instruction` in order for the historical ``.InstructionSet.c_if``
    to work correctly.
```

qiskit/transpiler/passes/scheduling/base_scheduler.py
ElePT: Still a TODO, right?

qiskit/transpiler/passes/scheduling/base_scheduler.py
ElePT: Hmm I would be in favor of removal now and deprecation in 1.4. But it wouldn't be too bad to leave it around if we can't.

qiskit/transpiler/passes/scheduling/padding/base_padding.py
ElePT: I think that it's not that we are "creating" a new required input, is that the information that could be in the dag is no longer there, right? so in my opinion the breaking API change is the removal of `duration` that causes the `target` argument to be required, not the addition of the `durations` input. The API is equally broken with and without `durations`, because the information that wasn't required is now required, so I'd say let's keep `durations`.

ElePT (commented):
Thanks for the titanic removal!! I have taken a look up to `qiskit/transpiler/passes/scheduling/scheduling/asap.py` and left a couple of minor comments and replied to a few of the discussion points Jake opened. Overall I think it looks good. I will continue with part 2 of the review after a break.

qiskit/transpiler/passes/scheduling/base_scheduler.py
mtreinish: Oops, yep I still need to do this. It got lost in all the other comments.

qiskit/circuit/classical/expr/constructors.py
mtreinish: We can still have a legacy condition structure. It's basically just a tuple of `(register, val)`, it just won't come from a `.condition` anymore, so we still need this function

releasenotes/notes/remove-deadweight-5d68e3fa42fc19bd.yaml
```diff
@@ -0,0 +1,112 @@
...
+  - |
+    Added a new method to the :class:`.Target` class has a new method, :meth:`~.Target.seconds_to_dt`.
+    This is used to translate a duration in seconds to a number of discretized
+    time steps of the system time resolution specified in the :attr:`.Target.dt` attribute.
+    This is typically useful for converting the :attr:`.InstructionProperties.duration` value
+    to units of ``dt``.
+upgrade_circuits:
+  - |
+    The deprecated attributes for :class:`~qiskit.circuit.Instruction` and :class:`.Gate`: ``duration`` and ``unit``. This includes
+    setting the ``unit`` or ``duration`` arguments for any :class:`qiskit.circuit.Instruction`
```
ElePT:
```suggestion
    The deprecated attributes for :class:`~qiskit.circuit.Instruction` and :class:`.Gate`: ``duration`` and ``unit`` have been removed. This includes
    setting the ``unit`` or ``duration`` arguments for any :class:`qiskit.circuit.Instruction`
```

releasenotes/notes/remove-deadweight-5d68e3fa42fc19bd.yaml
```diff
@@ -0,0 +1,112 @@
...
+    or subclass. These attributes were deprecated in Qiskit 1.3.0. These attributes were used to attach a custom execution
+    duration and unit for that duration to an individual instruction. However,
```
ElePT:
```suggestion
    or subclass. These attributes were deprecated in Qiskit 1.3.0 and 
    were used to attach a custom execution
    duration and unit for that duration to an individual instruction. However,
```

releasenotes/notes/remove-deadweight-5d68e3fa42fc19bd.yaml
```diff
@@ -0,0 +1,112 @@
...
+        qc.h(0)
+        with expected.if_test((expected.clbits[0], True)):
+            qc.x(0)
+        with expected.if_test((expected.clbits[1], False)):
+            qc.z(1)
+        qc.measure(0, 0)
+        qc.measure(1, 1)
+
+  - |
+    The deprecated methods ``InstructionSet.c_if`` has been removed. This method
```
ElePT:
```suggestion
    The deprecated method ``InstructionSet.c_if`` has been removed. This method
```

ElePT (commented):
Second part of the review completed! I didn't see any issues stand out from the tests/ final passes. I started leaving reno suggestions, changed my mind (it will be easier to edit in the release preparation PR), but the interface wasn't responding so I couldn't delete them. My only question is whether we should comment in this collection of renos the fact that BasicSimulator no longer supports control flow. I didn't add it at the time, and I could edit that reno too, I just thing it would be good to have it documented somewhere.

mtreinish:
> I started leaving reno suggestions, changed my mind (it will be easier to edit in the release preparation PR), but the interface wasn't responding so I couldn't delete them. 

Don't worry I'll ignore them then. :) 

> My only question is whether we should comment in this collection of renos the fact that BasicSimulator no longer supports control flow. I didn't add it at the time, and I could edit that reno too, I just thing it would be good to have it documented somewhere.

I can add a note about it, it feels a bit weird to call it out on it's own. But you're right we should mention it.

qiskit/transpiler/passes/scheduling/base_scheduler.py
mtreinish: Fixed in: https://github.com/Qiskit/qiskit/pull/13506/commits/c8086ef65f1b2a4d7743a5241914e09a97618998

qiskit/transpiler/passes/scheduling/padding/base_padding.py
mtreinish: I missed which scheduler pass this is in. This pass is deprecated so there is nothing to do here besides remove it. See: https://github.com/Qiskit/qiskit/pull/13876

qiskit/transpiler/passes/scheduling/base_scheduler.py
mtreinish: I missed which scheduler pass this is in. This pass is deprecated so there is nothing to do here besides remove it. See: https://github.com/Qiskit/qiskit/pull/13876

crates/circuit/src/dag_circuit.rs
jakelishman: This one seems like an actual potential bug we ought to fix - it's silencing a Python exception in that `.ok()`.

qiskit/circuit/instructionset.py
jakelishman: Tracked in #13907.

qiskit/transpiler/passes/utils/convert_conditions_to_if_ops.py
jakelishman: We already deprecated this in 1.3, so the ship sailed.

qiskit/transpiler/passes/scheduling/padding/pad_delay.py
jakelishman: Same resolution reason as https://github.com/Qiskit/qiskit/pull/13506#discussion_r1960019724.

I wouldn't say I'm entirely sold on Matt and Elena's logic, but I don't have anything better to suggest, so pragmatically this needs to move forwards.

qiskit/transpiler/passes/scheduling/scheduling/base_scheduler.py
jakelishman: I think we're wrong to have time-unit conversion logic here.  `TimeUnitConversion` is supposed to do this, and it's even set as a `self.requires` pass.  I'll have a play locally, because I think there must some tricks in how the scheduling passes have been done.

crates/circuit/src/dag_circuit.rs
jakelishman: On reflection: actually I see that the purpose here is that we need to skip `ForLoop`, `SwitchCaseOp` etc, and suppressing the attribute error is faster than explicitly type checking.

Clarified (mostly to myself) in 505d28b449.

qiskit/circuit/library/standard_gates/rxx.py
jakelishman: Done in 3e47130923.

qiskit/circuit/library/standard_gates/ry.py
jakelishman: Done in 3e47130923.

qiskit/qasm3/exporter.py
jakelishman: Done in aea12188f2.

jakelishman:
I pushed up a few commits, but right now, I think there's trouble in the scheduling transpiler passes, in that some of the logic around converting `Delay`'s time units seems to be in the wrong place now (and tbh, I don't immediately understand how it was working in the 1.x series).  I'll have a bit more of a look into that.

qiskit/transpiler/passes/scheduling/scheduling/base_scheduler.py
jakelishman: `TimeUnitConversion` was doing this bit of the delay-unit conversion (though the logic is weirdly hidden in `InstructionDurations.get`), so these lines were unnecessary.

Tidied up in 13b68142c8.

test/python/circuit/test_circuit_operations.py
```diff
@@ -1030,31 +1030,6 @@ def test_reverse_with_standlone_vars(self):
...
-        with self.subTest("repeat 3 times"):
-            with self.assertWarns(DeprecationWarning):
-                inst = qc.to_instruction()
-            ref = QuantumCircuit(qr, cr)
-            for _ in range(3):
-                ref.append(inst, ref.qubits, ref.clbits)
-            with self.assertWarns(DeprecationWarning):
-                rep = qc.repeat(3)
-            self.assertEqual(rep, ref)
-
```
jakelishman: I think most of this test still ought to be present, it's just the `c_if` bits that should be removed?

test/python/circuit/test_circuit_operations.py
```diff
@@ -1289,28 +1264,6 @@ def test_reverse_bits_with_overlapped_registers(self):
...
-        expected = QuantumCircuit([c1, c0], [q1, q0])
-        expected.h(1)
-        expected.cx(1, 0)
-        with self.assertWarns(DeprecationWarning):
-            expected.x(1).c_if(0, True)
-        expected.measure(1, 1)
-
-        self.assertEqual(qc.reverse_bits(), expected)
-
```
jakelishman: Same comment here: I think the test is still valid and unduplicated, just doesn't need the `c_if`?

test/python/circuit/test_control_flow_builders.py
```diff
@@ -1958,346 +1877,162 @@ def test_break_continue_deeply_nested(self, loop_operation):
...
-                loop_body.h(qubits[6]).c_if(clbits[10], 0)
+            with loop_body.if_test((clbits[10], 0)):
+                loop_body.h(qubits[6])
 
             expected = QuantumCircuit(qubits, clbits)
             expected.while_loop(cond_loop, loop_body, qubits[:7], clbits[:11])
 
             self.assertEqual(canonicalize_control_flow(test), canonicalize_control_flow(expected))
 
-        with self.subTest("while/else/else"):
```
jakelishman, line 1969: I like how you just gave up on this one haha

qiskit/qpy/binary_io/circuits.py
```diff
@@ -320,12 +320,16 @@ def _read_instruction(
         )
         if condition is not None:
             warnings.warn(
-                f"The .condition attribute on {gate_name} will be loaded as an IfElseOp "
-                "starting in Qiskit 2.0",
-                FutureWarning,
+                f"The .condition attribute on {gate_name} can not be "
+                "represented in this version of Qiskit. It will be "
+                "represented as an IfElseOp instead.",
+                UserWarning,
```
jakelishman, line 326: Maybe this ought to be `QPYLoadingDeprecatedFeatureWarning` from `qiskit.qpy.exceptions`?  Same for the other warning in this file.

test/python/visualization/test_dag_drawer.py
```diff
@@ -84,34 +84,6 @@ def test_dag_drawer_no_register(self):
...
-        qc.barrier(0, 1)
-        qc.measure(0, 0)
-        dag = circuit_to_dagdependency(qc)
-        with tempfile.TemporaryDirectory() as tmpdirname:
-            tmp_path = os.path.join(tmpdirname, "dag_d.png")
-            dag_drawer(dag, filename=tmp_path)
-            image_ref = path_to_diagram_reference("dag_dep.png")
-            image = Image.open(tmp_path)
-            self.assertImagesAreEqual(image, image_ref, 0.1)
-
```
jakelishman: This test should stay, just the `c_if` needs to go - we'd be losing the test coverage.

test/python/visualization/test_circuit_text_drawer.py
```diff
@@ -353,56 +353,6 @@ def test_text_measure_2_reverse_bits(self):
...
-                circuit_drawer(
-                    circuit,
-                    output="text",
-                    initial_state=True,
-                    cregbundle=False,
-                    wire_order=[2, 1, 3, 0, 6, 8, 9, 5, 4, 7],
-                )
-            ),
-            expected,
-        )
```
jakelishman: This test looks still correct, just we need to use something else for the conditional bits (`if_test`?)

test/python/visualization/test_circuit_text_drawer.py
```diff
@@ -5975,145 +3737,6 @@ def test_if_op_bundle_true(self):
...
-                    circuit,
-                    output="text",
-                    fold=77,
-                    initial_state=False,
-                    wire_order=[2, 0, 3, 1, 4, 5, 6],
-                )
-            ),
-            expected,
-        )
-
```
jakelishman: I think these tests are still necessary?

test/python/visualization/test_circuit_drawer.py
```diff
@@ -141,74 +141,6 @@ def test_latex_output_file_correct_format(self):
                     self.assertIn(im.format.lower(), filename.rsplit(".", maxsplit=1)[-1])
             os.remove(filename)
 
-    def test_wire_order(self):
```
jakelishman: Still a valid test, I think?

test/python/compiler/test_transpiler.py
```diff
@@ -2393,23 +2395,6 @@ def test_qasm3_output(self, optimization_level):
...
-                self._regular_circuit(),
-                backend=backend,
-                optimization_level=optimization_level,
-                seed_transpiler=2022_10_17,
-            )
-        # TODO: There's not a huge amount we can sensibly test for the output here until we can
-        # round-trip the OpenQASM 3 back into a Terra circuit.  Mostly we're concerned that the dump
-        # itself doesn't throw an error, though.
-        self.assertIsInstance(qasm3.dumps(transpiled).strip(), str)
-
```
jakelishman, line 2412: Should this have been deleted?

test/python/primitives/test_backend_estimator.py
```diff
@@ -504,7 +504,7 @@ def test_dynamic_circuit(self):
             estimator = BackendEstimator(backend, skip_transpilation=True)
             estimator.set_transpile_options(seed_transpiler=15)
             result = estimator.run(qc, observable).result()
-        self.assertAlmostEqual(result.values[0], 0, places=1)
+        self.assertAlmostEqual(float(result.values[0]), -1.0, places=1)
```
jakelishman: Should this have changed?  I feel like it still ought to be averaging to 0.  Maybe a bug in Aer?

test/python/primitives/test_backend_sampler.py
```diff
@@ -394,16 +394,16 @@ def test_circuit_with_dynamic_circuit(self):
...
+                qc.break_loop()
 
         with self.assertWarns(DeprecationWarning):
             backend = Aer.get_backend("aer_simulator")
             sampler = BackendSampler(backend, skip_transpilation=True)
             sampler.set_options(seed_simulator=15)
             sampler.set_transpile_options(seed_transpiler=15)
             result = sampler.run(qc).result()
-        self.assertDictAlmostEqual(result.quasi_dists[0], {0: 0.5029296875, 1: 0.4970703125})
+        self.assertDictAlmostEqual(result.quasi_dists[0], {0: 0.021484375, 1: 0.978515625})
```
jakelishman: This also looks incorrect now - maybe an Aer bug?

jakelishman (commented):
13b68142c8 simplifies the modifications to the scheduling passes, which pulls them back a bit more to what I'd expect.  Somebody else ought to review that commit, though.

Since I hadn't last time, I looked through all the tests too.  I think there's a couple that were maybe removed by accident, but not certain.

I'm still not wild about the stability around the scheduling passes suddenly needing `InstructionDurations` (or `Target`) in 2.0 when it wasn't necessary in 1.x.  I get _why_ we have to do that, and I guess I'm saying I'm ok to accept it as a break/exception to the deprecation-policy, but only really because we don't have a choice at this point, other than keeping around state that we _really_ don't want to keep alive any longer.

Other than these last few comments, I'm ok to approve.

test/python/circuit/test_control_flow_builders.py
mtreinish: Lol, yeah I totally did, this one in particular I couldn't make happy. I can give it a try again if you think it's important to keep it without the c_if. But this was the only one I really struggled to rewrite and keep working.

qiskit/qpy/binary_io/circuits.py
mtreinish: I looked at the description of the warning class and it felt more like a warning raised on loading a deprecated but not yet removed function. I saw `UserWarning` being used for this kind of warning in other places in QPY so I went with it. If we want a custom warning type here I can add a new type that is specific to this situation.

test/python/circuit/test_control_flow_builders.py
jakelishman: I don't think it's important - I overdid this file when writing it

test/python/compiler/test_transpiler.py
mtreinish: I may have been jumping the gun a bit on this one since `BackendV1` is deprecated and being removed I opted to just remove this test. `BackendV1` doesn't support control flow instruction (as it can't represent them as allowed) and `_regular_circuit()`  adds an if test. So I opted to just remove this test because we cover the qasm and transpiler pieces in `test_qasm3_output` and it'll be removed anyway in the `BackendV1` removal.

test/python/visualization/test_circuit_drawer.py
mtreinish: Restored in: https://github.com/Qiskit/qiskit/pull/13506/commits/b47f82c00ca9c384f04be7d9655502d7e666e1a1

test/python/visualization/test_circuit_text_drawer.py
mtreinish: Restored in: https://github.com/Qiskit/qiskit/pull/13506/commits/6f3861acbf80a115679ffa7f66f55a108b3c6af8

test/python/visualization/test_dag_drawer.py
mtreinish: Restored in: https://github.com/Qiskit/qiskit/pull/13506/commits/720dbe8246fff5e1b6cacb0f8016003a60dcfde6

test/python/visualization/test_circuit_text_drawer.py
mtreinish: Restored in: https://github.com/Qiskit/qiskit/pull/13506/commits/6f3861acbf80a115679ffa7f66f55a108b3c6af8

test/python/circuit/test_circuit_operations.py
mtreinish: Restored in: https://github.com/Qiskit/qiskit/pull/13506/commits/95581c17513b5e41ddc9603922f7b05477a10493

test/python/circuit/test_circuit_operations.py
mtreinish: Restored in: https://github.com/Qiskit/qiskit/pull/13506/commits/95581c17513b5e41ddc9603922f7b05477a10493

test/python/primitives/test_backend_estimator.py
mtreinish: It shouldn't have changed as the condition should be identical. It would either be an aer bug or an issue in the primitive implementation (since the `BackendEstimatorV2` version didn't require any update). I opted to just roll will it here because this test will also be removed with the BackendEstimator v1 estimator implementation and the v2 tests didn't need any changes.

test/python/primitives/test_backend_sampler.py
mtreinish: I did the same thing here as for the estimator implementation.

test/python/compiler/test_transpiler.py
jakelishman: Ah, I'd not realised that `_regular_circuit` included control flow.  Yeah, I can live with just removing this, even though it's marginally out of order.

qiskit/qpy/binary_io/circuits.py
jakelishman: I thought we'd overhauled the warnings a little more than that, but if there's still ones that just emit `UserWarning`, I'm fine leaving it in this PR.

test/python/primitives/test_backend_estimator.py
jakelishman: Sounds like it must be a bug - I feel like we should either delete this test (and the sampler one) slightly early, or put a big comment on it saying that we're asserting something incorrect.  I don't like merging a PR that encodes an incorrect test, even if we're expecting it to be ephemeral.

jakelishman:
My only remaining comment is about the buggy backend primitives V1 tests - I don't mind if we eagerly remove them or put in a comment explaining that they're bugged, but I don't want to merge a broken test, just out of principle.

Otherwise: somebody else should confirm that my commit 13b68142c8754b814a766bedf16a87a151e90e44 is fine, and then I can approve.

mtreinish:
> My only remaining comment is about the buggy backend primitives V1 tests - I don't mind if we eagerly remove them or put in a comment explaining that they're bugged, but I don't want to merge a broken test, just out of principle.
> 

I'll remove them then, I have the PR removing the backend primitive v1 open already here: #13877 which deletes both test files in their entirety.

> Otherwise: somebody else should confirm that my commit [13b6814](https://github.com/Qiskit/qiskit/commit/13b68142c8754b814a766bedf16a87a151e90e44) is fine, and then I can approve.

It looks fine to me, I think that pass will need more of a refactor pretty soon. The whole mutating the local durations object based on the dag doesn't really work after this PR and the removal of pulse, but for now it should be fine.

jakelishman (approved):
Thanks for all the changes on this.  Let's get this merged, and hopefully there's not too much of a tail afterwards.  I suspect we'll at least need to port some changes over to `qiskit-ibm-runtime` in the scheduling passes.

Issue #13192, opened 2024-09-19
Help to transpile a parameterized circuit with more than 3 controls


Augusto12:
### Environment

- **Qiskit version**: 1.2.1
- **Python version**: 3.10.14
- **Operating system**: Ubuntu 20.04.6 LTS (WSL2)

### What is happening?

I can't transpile a parameterized circuit with more than 3 controls

### How can we reproduce the issue?

```python
from qiskit.compiler import transpile
from qiskit.circuit import QuantumCircuit, ParameterVector

p = ParameterVector('p',4)

pqc = QuantumCircuit(4)
for i in range(4):
    pqc.ry(p[i],i)

pqc = pqc.control(4,0,annotated=True)
pqc = transpile(pqc, basis_gates=['u1', 'u2', 'u3', 'cx'], optimization_level=3)
```

### What should happen?

I got the error: 
**QiskitError: 'Cannot synthesize MCRY with unbound parameter: p[0].'**

ACE07-Sev:
Apparently, it's because they haven't implemented the `q_ancillae` in `add_control()` method of `qiskit.circuit.add_control.py`.
```py
def control(
    operation: Gate | ControlledGate,
    num_ctrl_qubits: int | None = 1,
    label: str | None = None,
    ctrl_state: str | int | None = None,
) -> ControlledGate:
    """Return controlled version of gate using controlled rotations. This function
    first checks the name of the operation to see if it knows of a method from which
    to generate a controlled version. Currently, these are ``x``, ``rx``, ``ry``, and ``rz``.
    If a method is not directly known, it calls the unroller to convert to `u1`, `u3`,
    and `cx` gates.

    Args:
        operation: The gate used to create the ControlledGate.
        num_ctrl_qubits: The number of controls to add to gate (default=1).
        label: An optional gate label.
        ctrl_state: The control state in decimal or as
            a bitstring (e.g. '111'). If specified as a bitstring the length
            must equal num_ctrl_qubits, MSB on left. If None, use
            2**num_ctrl_qubits-1.

    Returns:
        Controlled version of gate.

    Raises:
        CircuitError: gate contains non-gate in definition
    """
    from math import pi

    # pylint: disable=cyclic-import
    from qiskit.circuit import controlledgate

    ctrl_state = _ctrl_state_to_int(ctrl_state, num_ctrl_qubits)

    q_control = QuantumRegister(num_ctrl_qubits, name="control")
[...]

Issue #11142, opened 2023-10-30
TypeError: ParameterExpression with unbound parameters ({ParameterVectorElement()}) cannot be cast to a float.


NahidaNahida:
### Environment

- **Qiskit Terra version**: 0.25.3
- **Python version**: 3.9.18
- **Operating system**: ubuntu

### What is happening?

When I tried to construct a parameterized quantum circuit with multiple-controlled quantum gates with `RYGate(theta).control(num_ctrl_qubits=5, ctrl_state='01101'), [0,1,2,3,4,5])` in ubuntu OS, it turns to be a warning that `TypeError: ParameterExpression with unbound parameters ({ParameterVectorElement(θ[0])}) cannot be cast to a float.`. If I applied this controlled gate with fewer control qubits, the warning would not occur. By contrast, as I ran the codes in windows OS, nothing unexpected happened, in the case of the same version of qiskit.

I don't know what is wrong with that and I wonder whether ubuntu OS triggers this bug.

### How can we reproduce the issue?

Run the below codes, especially in ubuntu
```
from qiskit.circuit import QuantumCircuit, ParameterVector
from qiskit.circuit.library import RYGate

theta_list = ParameterVector('θ', 6)                                   
circuit = QuantumCircuit(10)
for theta in theta_list:                              
    circuit.append(RYGate(theta).control(num_ctrl_qubits=5,
                                        ctrl_state='01101'), 
                                        [0,1,2,3,4,5])
circuit.draw('mpl')
```

### What should happen?

The program crash should not happen like that in windows OS.

Raghav-Bell:
Please refer to issue #10697 .

zhiyaol:
I'm experiencing this exact issue.

NahidaNahida:
> I'm experiencing this exact issue.

My temporary solution is to give up the parameterized circuit and directly bind parameters when applying the rotation gates.

SimoneGasperini:
This is related to issue #10311.

ShellyGarion:
This should be fixed now that #13507 is merged. If it's not so - please re-open this issue.

Issue #7975, opened 2022-04-21
MCRZ gate does not alllow to bind parameters


danperazzo:
### Environment

- **Qiskit Terra version**:  0.19.2
- **Python version**:  3.8
- **Operating system**:  Windows 10/Ubuntu 20.04

### What is happening?

If I use the multi-controlled Z gate rotation I cannot use parameter binds to run a circuit. It fails to create the QOBJ.

### How can we reproduce the issue?

``` 
from qiskit.circuit import QuantumCircuit, Parameter
from qiskit import QuantumCircuit, QuantumRegister, AncillaRegister, ClassicalRegister
from qiskit import Aer
from qiskit import transpile, assemble


states = QuantumRegister(5, 'states') 
ancilla = QuantumRegister(1, 'ancilla')
output = ClassicalRegister(1, 'reg_class')

# create the parameter
phi = Parameter('phi')

qc = QuantumCircuit(states,ancilla,output)

# parameterize the rotation
qc.mcrz(phi, states[:-1],states[-1])
qc.measure(ancilla,output)

simulator_backend = Aer.get_backend('aer_simulator')

qc = transpile(qc,simulator_backend)

qobj = assemble(qc,
                        shots=2,
                        parameter_binds = [{phi: 3.14}])


results = simulator_backend.run(qobj).result()
        
all_counts = results.get_counts()

print(all_counts)
```

### What should happen?

If we ran the code, the exped result would be the counts of the outputs of the measured value. However, we get a message saying it failed to convert from python to C++ and returns a empty string.

`ERROR: Failed to load qobj: Unable to cast Python instance to C++ type (compile in debug mode for details)`

### Any suggestions?

The controlled Z-gste seems to be working fine, maybe inspect it.

jakelishman:
Thanks for the report!  It's pretty arcane, but this is actually caused by `CUGate` rather than the multi-controlled rz - it's to do with misalignments of expectations between how `CUGate.params` works, and how other parts of Terra expect to be able to mutate those lists.  There's more information in #7326 - we don't have a work-around yet unfortunately, but it is a bug we need to squash.

ShellyGarion:
I think that #13507 should solve the binding issues of `mcrz`. Please reopen this issue if this is not so.

PR #13507, opened 2024-12-01
Allow mcrx, mcry and mcrz gate methods to accept angles of ParameterValueType

ShellyGarion:
### Summary

Address https://github.com/Qiskit/qiskit/issues/12863

There are currently several open issues that fail with errors like:
`QiskitError: 'Cannot synthesize MCRY with unbound parameter:`

https://github.com/Qiskit/qiskit/issues/7975
https://github.com/Qiskit/qiskit/issues/11142
https://github.com/Qiskit/qiskit/issues/13192

This PR aims to fix them


ShellyGarion:
The failing tests were added as part of https://github.com/Qiskit/qiskit/pull/12752.
Perhaps they could be removed?

qiskit/circuit/library/standard_gates/multi_control_rotation_gates.py
```diff
@@ -337,14 +418,20 @@ def mcry(
...
+                self.compose(cgate, control_qubits + [target_qubit], inplace=True)
 
-            cgate = _mcsu2_real_diagonal(
-                RYGate(theta).to_matrix(),
-                num_controls=len(control_qubits),
-                use_basis_gates=use_basis_gates,
-            )
-            self.compose(cgate, control_qubits + [target_qubit], inplace=True)
+            else:
+                cgate = _mcsu2_real_diagonal(
```
Cryoris: Is there a reason not to use `_mcsu2_with_params` here as well? The code looks like it should work for unparameterized cases, too

qiskit/circuit/library/standard_gates/multi_control_rotation_gates.py
```diff
@@ -205,6 +205,86 @@ def _mcsu2_real_diagonal(
...
+
+    """
+    from .x import MCXVChain
+    from qiskit.circuit.library.standard_gates.ry import RYGate
+    from qiskit.circuit.library.standard_gates.rz import RZGate
+
+    if unitary_type == "ry":
+        s_gate = RYGate(-theta / 4)
+    elif unitary_type in ("rz", "rx"):
+        s_gate = RZGate(-theta / 4)
```
Cryoris: Is the `s_gate` for `RXGate` also `RXGate(-theta / 4)`?

Cryoris (commented):
This looks great, I'm glad that we can support parameterized multicontrolled rotations! 

I'm wondering whether we can consolidate the method you added with the existing `_mcsu2_real_diagonal` to ensure we don't duplicate the logic. One option could be to take the unitary as `Gate` instead of `np.ndarray`:
```python
def _mcsu2(gate: Gate, num_controls: int, use_basis_gates: bool) -> QuantumCircuit:
    if isinstance(gate, RYGate):
        s_gate = # ...
    elif # the other Pauli rotation cases, RX and RZ
    else:
       unitary = gate.to_matrix()
       s_gate = # compute s_gate
       
   # common logic here
```
do you think that might work?

Cryoris:
>The failing tests were added as part of https://github.com/Qiskit/qiskit/pull/12752.
Perhaps they could be removed?

You refer to `test_mc_failure_without_annotation`? We can remove `RX/RY/RZ` in that test, but we likely want to keep the others, I'd think

qiskit/circuit/library/standard_gates/multi_control_rotation_gates.py
ShellyGarion: I think that perhaps `_mcsu2_real_diagonal` should generally be faster ?

qiskit/circuit/library/standard_gates/multi_control_rotation_gates.py
ShellyGarion: Yes, the difference is at lines 258 & 282:
```
   if unitary_type == "rx":
        circuit.h(target)
```

ShellyGarion:
> > The failing tests were added as part of #12752.
> > Perhaps they could be removed?
> 
> You refer to `test_mc_failure_without_annotation`? We can remove `RX/RY/RZ` in that test, but we likely want to keep the others, I'd think

These are the failing tests, but all of them are failing (not only for RX / RY / RZ)

qiskit/circuit/library/standard_gates/multi_control_rotation_gates.py
Cryoris: So would it be cleaner to write
```python
if unitary_type == "rx":
    s_gate = RXGate(-theta / 4)
elif unitary_type == "ry":
    s_gate = RYGate(-theta / 4)
else:
    s_gate = RZGate(-theta / 4)
```
and skip the bit that does
```python
   if unitary_type == "rx":
        circuit.h(target)
```
?

qiskit/circuit/library/standard_gates/multi_control_rotation_gates.py
```diff
@@ -205,6 +205,86 @@ def _mcsu2_real_diagonal(
...
+        ctrl_state_k_2 = str_ctrl_state[::-1][k_1:][::-1]
+
+    circuit = QuantumCircuit(num_controls + 1, name="MCSU2")
+    controls = list(range(num_controls))  # control indices, defined for code legibility
+    target = num_controls  # target index, defined for code legibility
+
+    if unitary_type == "rx":
+        circuit.h(target)
+
+    mcx_1 = MCXVChain(num_ctrl_qubits=k_1, dirty_ancillas=True, ctrl_state=ctrl_state_k_1)
```
Cryoris: Just a reminder from our offline discussions that these should use the non-deprecated synthesis methods 🙂

qiskit/circuit/library/standard_gates/multi_control_rotation_gates.py
Cryoris: Hmm why? Isn't `_mcsu2_real_diagonal` maybe even slower because it has to do some extra calculations to get the matrix and for the `s_gate` whereas we here know what it is?

releasenotes/notes/fix-multi-controlled-rotation-gates-with-parameter-12a04701d0cd095b.yaml
```diff
@@ -0,0 +1,8 @@
+---
+fixes:
+  - |
+    Fix a bug in the multi-controlled rotation circuit methods `QuantumCircuit.mcrx`, `QuantumCircuit.mcry,
+    and `QuantumCircuit.mcrz`, when the user provides an unbounded parameter, as well as when calling
+    `RXGate(theta).control()`, `RYGate(theta).control()` or `RZGate(theta).control()` for some parameter theta.
```
Cryoris:
```suggestion
    Fix a bug in the multi-controlled rotation circuit methods :meth:`.QuantumCircuit.mcrx`, :meth:`.QuantumCircuit.mcry`,
    and :meth:`.QuantumCircuit.mcrz`, when the user provides an unbounded parameter, as well as when calling
    :meth:`.RXGate.control`, :meth:`.RYGate.control` or :meth:`.RZGate.control` where the rotation
    angle is a :class:`.ParameterExpression`.
```

qiskit/circuit/library/standard_gates/multi_control_rotation_gates.py
ShellyGarion: You are right, we can consolidate the methods into a single method, and not do all the s_gate calculations in the RX / RY / RZ cases.

qiskit/circuit/library/standard_gates/multi_control_rotation_gates.py
ShellyGarion: I don't think this is equivalent, since we need to add the H Gates at the beginning and the end of the entire circuit.
(see now with the merged code)

qiskit/circuit/library/standard_gates/multi_control_rotation_gates.py
ShellyGarion: Yes. I also plan to move this code to the synthesis library.

qiskit/synthesis/multi_controlled/multi_control_rotation_gates.py
```diff
@@ -0,0 +1,205 @@
...
+
+    circuit = QuantumCircuit(num_controls + 1, name="MCSU2")
+    controls = list(range(num_controls))  # control indices, defined for code legibility
+    target = num_controls  # target index, defined for code legibility
+
+    if not is_secondary_diag_real:
+        circuit.h(target)
+
+    mcx_1 = synth_mcx_n_dirty_i15(num_ctrl_qubits=k_1).to_gate()
+    circuit.append(mcx_1, controls[:k_1] + [target] + controls[k_1 : 2 * k_1 - 2])
```
ShellyGarion: `to_gate()` is needed here (and at mcx_2, mcx_3 and mcx_4) since otherwise this test:

```
def test_to_gate():
    num_ctrls = 6
    theta = 0.3
    qc = QuantumCircuit(num_ctrls+1)
    qc.mcrx(theta, range(num_ctrls), num_ctrls)
    print (qc.to_gate())
```

is failing with the following error:
`qiskit.exceptions.QiskitError: 'One or more instructions cannot be converted to a gate. "mcx_vchain" is not a gate instruction'`

qiskit/circuit/quantumcircuit.py
```diff
@@ -4683,6 +4685,202 @@ def mcp(
             copy=False,
         )
 
+    def mcrx(
+        self,
+        theta: ParameterValueType,
+        q_controls: Union[QuantumRegister, List[Qubit]],
```
Cryoris: It would be nice to use the native typehints we started using since Python 3.7 is the minimal support version, i.e.
```suggestion
        q_controls: QuantumRegister | list[Qubit],
```

qiskit/circuit/quantumcircuit.py
```diff
@@ -4683,6 +4685,202 @@ def mcp(
...
+    ):
+        """
+        Apply Multiple-Controlled X rotation gate
+
+        Args:
+            self (QuantumCircuit): The QuantumCircuit object to apply the mcrx gate on.
+            theta (float): angle theta
+            q_controls (QuantumRegister or list(Qubit)): The list of control qubits
+            q_target (Qubit): The target qubit
+            use_basis_gates (bool): use p, u, cx
```
Cryoris: `self` is implicit, we don't need to list it 🙂 

Also we typically don't duplicate the type hints here, if they are already in the signature we can skip them in the docstring:
```suggestion
            theta: angle theta
            q_controls: The list of control qubits
            q_target: The target qubit
            use_basis_gates: use p, u, cx
```

qiskit/circuit/quantumcircuit.py
```diff
@@ -4683,6 +4685,202 @@ def mcp(
...
+        # auto-select the best mode
+        if mode is None:
+            # if enough ancillary qubits are provided, use the 'v-chain' method
+            additional_vchain = MCXGate.get_num_ancilla_qubits(len(control_qubits), "v-chain")
+            if len(ancillary_qubits) >= additional_vchain:
+                mode = "basic"
+            else:
+                mode = "noancilla"
+
+        if mode == "basic":
```
Cryoris, line 4796: This is not something to fix in this PR: but do you know why we have this special case only for `MCRY` and not for the others? If it's more efficient we could also use it there. 

This is something we can fix very cleanly if we synthesize controlled gates via HLS, which knows the number of free auxiliaries.

qiskit/synthesis/multi_controlled/mcx_synthesis.py
```diff
@@ -54,6 +54,8 @@ def synth_mcx_n_dirty_i15(
     """
 
     num_qubits = 2 * num_ctrl_qubits - 1
+    if num_ctrl_qubits == 1:
+        num_qubits = 2
```
Cryoris: Just a suggestion, but this looks like the nicer flow which avoid re-assigning the `num_qubits` variable
```suggestion
    if num_ctrl_qubits == 1:
        num_qubits = 2
    else:
        num_qubits = 2 * num_ctrl_qubits - 1
```

qiskit/synthesis/multi_controlled/multi_control_rotation_gates.py
Cryoris: Yeah this is because `append` will cast the circuit into an `Instruction`. Alternatively you could use
```
mcx_1 = synth_mcx_n_dirty_i15(k_1)
circuit.compose(mcx_1, ..., inplace=True)
```
This will end up being faster because we don't have to create a new `Gate` and upon transpilation we save unwrapping the gate.

qiskit/circuit/quantumcircuit.py
ShellyGarion: The use of ancillas was done in a PR that you implemented https://github.com/Qiskit/qiskit/pull/4787 following issue https://github.com/Qiskit/qiskit/issues/4786. I'm not sure if there is some improvement anymore.

ShellyGarion: after checking it seems that adding ancillas can improve mcry in certain cases (num_controls >= 7). This is an optimization that should be explored in a follow-up work.

qiskit/circuit/quantumcircuit.py
```diff
@@ -4751,7 +4749,7 @@ def mcrx(
     def mcry(
         self,
         theta: ParameterValueType,
-        q_controls: Union[QuantumRegister, List[Qubit]],
+        q_controls: QuantumRegister | list[Qubit],
         q_target: Qubit,
         q_ancillae: Optional[Union[QuantumRegister, Tuple[QuantumRegister, int]]] = None,
         mode: Optional[str] = None,
```
Cryoris: Sorry I didn't tag these before 🙂 I changed the typehint from `tuple[QuantumRegister, int]` to `list[Qubit]` because it looks like it should be a list of qubits, no?
```suggestion
        q_ancillae: QuantumRegister | list[Qubit] | None = None,
        mode: str | None = None,
```

qiskit/circuit/quantumcircuit.py
```diff
@@ -4683,6 +4684,199 @@ def mcp(
             copy=False,
         )
 
+    def mcrx(
```
Cryoris, line 4687: It would be nice to polish these docs a bit, see e.g. the docstring for `mcp` above, which looks a bit nicer in the docs 🙂 Should the type of the qubits also be `QubitSpecifier` here?

qiskit/circuit/quantumcircuit.py
```diff
@@ -4683,6 +4684,199 @@ def mcp(
...
+        control_qubits = self._qbit_argument_conversion(q_controls)
+        target_qubit = self._qbit_argument_conversion(q_target)
+        if len(target_qubit) != 1:
+            raise QiskitError("The mcrx gate needs a single qubit as target.")
+        all_qubits = control_qubits + target_qubit
+        target_qubit = target_qubit[0]
+        self._check_dups(all_qubits)
+
+        n_c = len(control_qubits)
+        if n_c == 1:  # cu
```
Cryoris, line 4720: With @alexanderivrii's PR on the annotated gate plugins, could we change this method in future to just append an annotated Pauli rotation? It would be nice if we can move this logic to the plugin so it doesn't need to be duplicated for the annotated rotation.

qiskit/synthesis/multi_controlled/multi_control_rotation_gates.py
```diff
@@ -0,0 +1,207 @@
...
+import numpy as np
+
+from qiskit.circuit import QuantumCircuit, Gate
+from qiskit.circuit.library.standard_gates.u3 import _generate_gray_code
+from qiskit.exceptions import QiskitError
+
+
+def _apply_cu(circuit, theta, phi, lam, control, target, use_basis_gates=True):
+    if use_basis_gates:
+        # pylint: disable=cyclic-import
```
Cryoris:
```suggestion
```

Cryoris (commented):
One small comment on the docs and one question that doesn't need to be resolved in this PR, otherwise LGTM 👍🏻

qiskit/circuit/quantumcircuit.py
ShellyGarion: Indeed. I think this could be done in a follow-up PR.

Cryoris (approved):
Thanks for the improvement! I'm not 100% sure whether this should be a bugfix or a new feature -- do you remember whether parameterized rotations were supported before the changes to `mcsu2_real_diag` etc. ?

qiskit/circuit/quantumcircuit.py
ShellyGarion: done in [db31ab2](https://github.com/Qiskit/qiskit/pull/13507/commits/db31ab23b38fc78833595293eecddcd5f3372962)

qiskit/synthesis/multi_controlled/multi_control_rotation_gates.py
ShellyGarion: done in [c62b90b](https://github.com/Qiskit/qiskit/pull/13507/commits/c62b90bb78bd1efae907e35e49c9693ab480bb77)

alexanderivrii (approved):
This looks good on my side as well. For posteriority, could you please add the number of 2-qubit gates we get for multiply-controlled parametric p gates before and after this PR (since we now use the same presumably better algorithm both for parametric and non-parametric gates)? It would be also nice to record somewhere further follow-ups discussed in this PR.

PR #13508, opened 2024-12-01
Add global barrier in qasm3 dump

### LINKED ISSUE #13485: More efficient QASM3 dumping of global barriers
### What should we add?

Qiskit let's the user use global barriers in a circuit, for example:
```python

qc= QuantumCircuit(3)
qc.x(0)
qc.barrier()
qc.x(1)
```

However, when these barriers get dumped into QASM3, all qubits are explicitly listed:
```python
from qiskit import qasm3

qasm3.dumps(qc)
```
```
OPENQASM 3.0;
include "stdgates.inc";
qubit[3] q;
x q[0];
barrier q[0], q[1], q[2];
x q[1];
```
While this is a valid QASM3 output, it doesn't use the [declared support](https://openqasm.com/language/delays.html#barrier-instruction) of global barriers in QASM3 via `barrier;`. For wide circuits this increases the output's size, parsing time, and even compile time down the line (as the global nature of the barrier needs to be checked for, or ignored).

kibrahim757:
### Summary
Using <code>barrier;</code> in a circuit and then  dumping it in QASM3,  results in the explicit listing of qubits, as noted in issue #13485. This pull request enables the use of a barrier on all qubits in the QASM3 dump by incorporating a <code>GLOBAL_BARRIER</code> flag within ExperimentalFeatures. When this flag is activated while invoking the <code>dump</code> function, it allows for a barrier to be applied to all qubits.

### Details and comments

1.Implemented <code>GLOBAL_BARRIER</code> in <code>ExperimentalFeatures</code>.
2. Created a new class, <code>QuantumGlobalBarrier</code>.
3. Added two test cases: one for barriers applied to all qubits and another for those applied to specific qubits.

Fix #13485


qiskit/qasm3/ast.py
```diff
@@ -478,6 +478,15 @@ def __init__(self, indexIdentifierList: List[Identifier]):
         self.indexIdentifierList = indexIdentifierList
 
 
+class QuantumGlobalBarrier:
+    """
+    quantumBarrier
+        : 'barrier'
+    """
+
+    __slots__ = ()
```
jakelishman, line 527: Could this be unified sensibly with `QuantumBarrier`, so we just represent `qargs = []` if it's meant to be the empty list?  It could simplify processing a little.

qiskit/qasm3/experimental.py
```diff
@@ -68,3 +68,7 @@ class ExperimentalFeatures(enum.Flag):
             }
         }
     """
+    GLOBAL_BARRIER = enum.auto()
+    """Support exporting using golobal barrier as proposed in
+        https://github.com/Qiskit/qiskit/issues/13485
+    """
```
jakelishman, line 74: It'd be a little cleaner to just explain inline what's meant, rather than linking to a Qiskit issue.  The "global barrier" is already a well-defined feature of OpenQASM 3, so the documentation might more appropriately note why it's not the default behaviour (not every consumer of Qiskit's OQ3 output supports it), and say something like

> Export `` :class:`.Barrier` `` objects that act on all qubit circuit qubits as global ``barrier;`` statements, rather than explicitly listing qubits.

It's a bit easier for a user to follow, without needing to click links and read technical discussions.

test/python/qasm3/test_export.py
```diff
@@ -2665,6 +2665,40 @@ def test_switch_v1_expr_target(self):
...
+OPENQASM 3.0;
+include "stdgates.inc";
+qubit[3] q;
+x q[0];
+barrier q[0], q[1];
+x q[1];
+"""
+        test = dumps(qc, experimental=ExperimentalFeatures.GLOBAL_BARRIER)
+        self.assertEqual(test, expected)
+
```
jakelishman, line 2642: Perhaps we could combine the two programs into a single OpenQASM 3 string, so we test both at once, and make sure we have coverage of when there's more than one barrier in a circuit?

Could you also add a test that the handling works when there's a `barrier` inside a control-flow statement, and the `barrier` affects all the qubits touched by the control-flow operation, but _not_ all the qubits in the circuit?  In this case, we mustn't write out a global barrier.

jakelishman (commented):
Thank you for this, and sorry for the delay in reviewing - I'd been off on paternity leave.

Please could you also add a "feature" release note about the new OpenQASM 3 experimental flag?  There's details here: https://github.com/Qiskit/qiskit/blob/main/CONTRIBUTING.md#release-notes.

jakelishman:
Oh, and btw, you don't need to press the "update branch" button unless there's actual merge conflicts to be resolved.  The branch will be automatically brought up-to-date when the PR enters the merge queue.

1ucian0:
Hi @kibrahim757 ,

Are you still working on this PR. There seems to be several comments that are not addressed yet.

jakelishman:
I'll just close this, and the underlying issue - I don't think there's a strong motivating for adding the extra surface.

jakelishman:
Oh sorry, I didn't realise Michael had commented on the base issue.

I'll leave this PR closed for now, but feel free to re-open if you want to work on it.

PR #13517, opened 2024-12-02
fix missing inverse definitions in `generate_basis_approximations`

burgholzer:
### Summary

This tiny PR fixes an oversight in the generation routine for basis approximations that is used in the SolovayKitaev synthesis pass. 

### Details and comments

While the dictionary of single-qubit gates listed the `sx` and `sxdg` gates, the dictionary of inverses was missing those definitions.
This PR simply adds the respective definitions.

Should this receive a release note?


Cryoris:
Nice catch! Did you notice this because you saw decompositions with `SX-SXdg` sequences, which should've been cancelled? If yes, it would be nice if we could add a regression test that ensures this behavior is now fixed.

burgholzer:
> Nice catch! Did you notice this because you saw decompositions with `SX-SXdg` sequences, which should've been cancelled? If yes, it would be nice if we could add a regression test that ensures this behavior is now fixed.

Not really. I was trying to set up a transpilation pipeline for compiling arbitrary circuits to the Clifford+T gate-set. I wanted to throw all possible single-qubit Clifford gates into the basis approximations method, including the `sx` and `sxdg` gates. Doing so made the function error out.
I believe this was missed as it was not tested in the existing unit test. The addition in https://github.com/Qiskit/qiskit/pull/13517/commits/4e8b90d496abdb61eab8cf362364c6aef8b110d1 should make sure this works as expected.
Do you believe additional tests are necessary here?

Cryoris:
Ah so the test you added fails without `sx` and `sxdg` in the `_1q_inverses` dictionary? If yes, then we don't need any further tests 🙂 However, as you suggested above, this merits a bugfix release note.

burgholzer:
> Ah so the test you added fails without `sx` and `sxdg` in the `_1q_inverses` dictionary? If yes, then we don't need any further tests 🙂 However, as you suggested above, this merits a bugfix release note.

Yeah, exactly.
Added a release note just now. Sorry that took so long... these days are quite busy.
Let me know if you want anything changed 😌

alexanderivrii:
Hi @burgholzer, a question not directly related to the PR: what is the best way to transpile into Clifford+T (i.e. how did you set up the transpilation pipeline for that)? Thx!

burgholzer:
> Hi @burgholzer, a question not directly related to the PR: what is the best way to transpile into Clifford+T (i.e. how did you set up the transpilation pipeline for that)? Thx!

Hi 👋🏼 
That's a good question. Right now, the respective code is still a bit messy and not as clean as I would have liked it to be.

```python
from qiskit import transpile
from qiskit.converters import circuit_to_dag, dag_to_circuit
from qiskit.transpiler.passes.synthesis import SolovayKitaev

# The quantum circuit to transpile
qc = ...

# Clifford+T gateset
gateset = [
    "i",
    "x",
    "y",
    "z",
    "h",
    "s",
    "sdg",
    "t",
    "tdg",
    "sx",
    "sxdg",
    "cx",
    "cy",
    "cz",
    "swap",
    "iswap",
    "dcx",
    "ecr",
    "measure",
    "barrier",
]

# Transpile the circuit to single- and two-qubit gates including rotations
compiled_for_sk = transpile(
    qc,
    basis_gates=[*gateset, "rx", "ry", "rz"],
)
# Synthesize the rotations to Clifford+T gates
# Measurements are removed and added back after the synthesis to avoid errors in the Solovay-Kitaev pass
pass_ = SolovayKitaev()
new_qc = dag_to_circuit(pass_.run(circuit_to_dag(compiled_for_sk.remove_final_measurements(inplace=False))))
new_qc.measure_all()
# Transpile once more to remove unnecessary gates and optimize the circuit
compiled_qc = transpile(
    new_qc,
    basis_gates=gateset,
)
```

While one could use the `"sk"` unitary synthesis method with the `transpile` command, I could not make that work in any combination I tried, as it would simply not transpile the circuit to the respective gate-set. So something like
```python

import numpy as np

qc = QuantumCircuit(2)
qc.cp(np.pi/4, 0, 1)

gateset = [
    "i",
    "x",
    "y",
    "z",
    "h",
    "s",
    "sdg",
    "t",
    "tdg",
    "sx",
    "sxdg",
    "cx",
    "cy",
    "cz",
    "swap",
    "iswap",
    "dcx",
    "ecr",
    "measure",
    "barrier",
]

compiled_qc = transpile(
    qc,
    basis_gates=gateset,
    unitary_synthesis_method = "sk",
)
```
would not work an error out because the synthesis does nothing, and then the basis gate decomposer cannot find a decomposition into the discrete basis.

However, I haven't had the time to fully work this out and report it as an issue.

releasenotes/notes/fix-missing-inverse-definition-af7fe8d9f2193308.yaml
```diff
@@ -0,0 +1,3 @@
+fixes:
+  - |
+    Generating discrete basis approximations for a basis containing `sx` or `sxdg` gates would fail due to missing inverse definitions of these gates in the :mod:`qiskit.synthesis.discrete_basis/generate_basis_approximations` module. This has been fixed by adding the missing inverse definitions.
```
Cryoris: The `generate_basis_approximations` function is currently not API documented, so let's just refer to `SolovayKitaev`, whose docs page explains how to call this function.
```suggestion
    Fixed a bug in the basis approximation generation for :class:`.SolovayKitaev`. Previously, generating discrete basis approximations using ``generate_basis_approximations`` for a basis containing ``"sx"`` or ``"sxdg"` gates would fail. This has now been fixed.
```

Cryoris (dismissed):
LGTM thanks for the fix!

releasenotes/notes/fix-missing-inverse-definition-af7fe8d9f2193308.yaml
```diff
@@ -0,0 +1,3 @@
+fixes:
+  - |
+    Fixed a bug in the basis approximation generation for :class:`.SolovayKitaev`. Previously, generating discrete basis approximations using ``generate_basis_approximations`` for a basis containing ``"sx"`` or ``"sxdg"` gates would fail. This has now been fixed.
```
Cryoris:
```suggestion
    Fixed a bug in the basis approximation generation for :class:`.SolovayKitaev`. 
    Previously, generating discrete basis approximations using 
    ``generate_basis_approximations`` for a basis containing ``"sx"`` or 
    ``"sxdg"`` gates would fail. This has now been fixed.
```

alexanderivrii (approved):
LGTM!

Issue #13518, opened 2024-12-03
README typo


jrzingel:
Hi all

I'm new to Qiskit and was following the guide in the README but some of the code was not working.

I believe that line 73 is incorrect and should change from

```python
print(f" > Counts: {result[0].meas.get_counts()}")
```

to

```python
print(f" > Counts: {result[0].data["meas"].get_counts()}")
```

Cheers

1ucian0:
closed via https://github.com/Qiskit/qiskit/pull/13519

PR #13519, opened 2024-12-03
Fix .data typo in README (#13518)

jrzingel:
### Summary
Small typo in the README as referenced in issue #13518

### Details and comments
This commit implements the changes as suggested in the issue


Cryoris:
Thank you for catching this! We should test this to ensure the readme is always correct, it seems to me that we had it at some point but I can't find this test anymore 🤔

Cryoris (approved):
I double checked this worked, let's add a test for this in a follow up.

Cryoris:

PR #13530, opened 2024-12-05
Minor fixes for new adder and multiplier gates classes

alexanderivrii:
### Summary

This PR adds several fixes and improvements for the new adder and multiplier gate classes
``FullAdderGate``, ``HalfAdderGate``, ``ModularAdderGate`` and ``MultiplierGate``.

First, it adds the ``_define`` method to each of these classes, populating ``self.definition`` with some default decomposition of the gate that does not use any ancilla qubits and in particular allows to construct ``Operator``s from quantum circuits containing these gates.

Second, it fixes the plugins for adder gates based on the "ripple_v95" synthesis function that requires ``n-1`` ancilla qubits. Previously the function could be called with fewer qubits available and the synthesized quantum circuit was not guaranteed to fit into the original quantum circuit leading to internal errors in ``HighLevelSynthesis`` pass.

Third, it adds the default synthesis plugin for ``FullAdderGate`` and improves the default synthesis plugins for ``HalfAdderGate`` and ``ModularAdderGate`` to always choose the best synthesis method depending on the number of ancillas available. Previously, we were getting less optimal circuits for certain small values of ``n``.


kevinhartman:
@Cryoris should we include this for 1.3.1? If so, could you take a look?

qiskit/transpiler/passes/synthesis/hls_plugins.py
```diff
@@ -1309,10 +1340,24 @@ def run(self, high_level_object, coupling_map=None, target=None, qubits=None, **
...
+        # The next best option is to use ripple_c04 (if there are enough ancilla qubits)
+        if (
+            decomposition := HalfAdderSynthesisC04().run(
+                high_level_object, coupling_map, target, qubits, **options
+            )
+        ) is not None:
+            return decomposition
+
+        # The QFT-based adder does not require ancilla qubits and should always succeed
+        return HalfAdderSynthesisD00.run(high_level_object, coupling_map, target, qubits, **options)
```
Cryoris: This seems to miss `()` -- we should make sure to add a test going here. Does the `coverage` report help here?

qiskit/circuit/library/arithmetic/adders/adder.py
```diff
@@ -208,3 +226,9 @@ def num_state_qubits(self) -> int:
             The number of state qubits.
         """
         return self._num_state_qubits
+
+    def _define(self):
+        """Populates self.definition with a decomposition of this gate."""
+        from qiskit.synthesis.arithmetic import adder_ripple_c04
+
+        self.definition = adder_ripple_c04(self.num_state_qubits, kind="full")
```
Cryoris, line 235: It would be nice to add a comment here that in the special case of a "full" adder, this method doesn't use ancillas -- otherwise it seems a bit surprising that this is the only one not using the QFT-based decomp

qiskit/transpiler/passes/synthesis/hls_plugins.py
```diff
@@ -1212,10 +1227,26 @@ def run(self, high_level_object, coupling_map=None, target=None, qubits=None, **
         if not isinstance(high_level_object, ModularAdderGate):
             return None
 
-        if options.get("num_clean_ancillas", 0) >= 1:
-            return adder_ripple_c04(high_level_object.num_state_qubits, kind="fixed")
+        # For up to 5 qubits, the QFT-based adder is best
```
Cryoris, line 1230: How did you determine the "best" here? 🙂

qiskit/transpiler/passes/synthesis/hls_plugins.py
```diff
@@ -1381,18 +1426,38 @@ def run(self, high_level_object, coupling_map=None, target=None, qubits=None, **
...
 
-    The plugin supports the following plugin-specific options:
+        # FullAdderSynthesisC04 requires no ancilla qubits and returns better results
+        # than FullAdderSynthesisV95 in all cases except for n=1.
+        if high_level_object.num_state_qubits == 1:
+            decomposition = FullAdderSynthesisV95().run(
+                high_level_object, coupling_map, target, qubits, **options
+            )
+            if decomposition is not None:
+                return decomposition
```
Cryoris, line 1449: Is there a case where this can be none, given that `num_state_qubits==1`?

releasenotes/notes/fix-adder-gates-39cf3d5f683e8880.yaml
```diff
@@ -0,0 +1,20 @@
+---
+features_transpiler:
+  - |
+    Added :class:`.FullAdderSynthesisDefault` plugin that chooses the best decomposition
+    for :class:`.FullAdderGate` based on the number of clean ancilla qubits 
+    available. 
+  - |
+    Improved :class:`.HalfAdderSynthesisDefault` and :class:`.ModularAdderSynthesisDefault`
+    plugins for :class:`.HalfAdderGate` and :class:`.ModularAdderGate` respectively,
+    choosing the best decomposition based on the number of clean ancilla qubits available.
```
Cryoris: If we want to include this in 1.3.1 these should be bugfixes, otherwise these improvements would have to be split off for 2.0 (these were likely oversights from before and valid as bugfix... 😉)

qiskit/transpiler/passes/synthesis/hls_plugins.py
alexanderivrii: Fixed and added tests that exercise different behaviors for the default plugins. Though we don't really have a robust way of knowing which synthesis method was actually used, so the new tests only check that quantum circuits with various adder types and number of state and ancilla qubits run successfully.

The coverage report is really amazing, and I am planning to use it significantly more in the future.

qiskit/circuit/library/arithmetic/adders/adder.py
alexanderivrii: Added this comment.

qiskit/transpiler/passes/synthesis/hls_plugins.py
alexanderivrii: A small python script that simply runs all available plugins for each of the adder gates, while varying the number of control and ancillary qubits, and manually choosing the method that minimizes the CX-count. Note that we have an option to automatically run all of the available plugins and choose the decomposition that provides the best possible results. 

Actually, it might be nice to think if the above behavior should be used by default.

qiskit/transpiler/passes/synthesis/hls_plugins.py
alexanderivrii: Probably not, however as more synthesis methods may get added in the future, and the default code may get shuffled around, it might be good to keep the if-check just in case.

releasenotes/notes/fix-adder-gates-39cf3d5f683e8880.yaml
alexanderivrii: I have rephrased this as fixes ;). And indeed I actually do think now that including one missing default plugin and fixing the logic of the other default plugins should be considered as fixes.

test/python/circuit/library/test_adders.py
```diff
@@ -233,6 +233,57 @@ def test_plugins_when_do_not_apply(self):
...
             self.assertEqual(synth.count_ops(), {"ModularAdder": 1})
 
+    def test_default_plugins(self):
+        """Tests covering different branches in the default synthesis plugins.."""
+        with self.subTest(name="HalfAdderTest1"):
+            adder = HalfAdderGate(3)
+            circuit = QuantumCircuit(9)
+            circuit.append(adder, range(7))
+            hls = HighLevelSynthesis()
+            _ = hls(circuit)
```
Cryoris: Could you add some assertion here that makes sure HLS ran as expected? It can be something simple, like checking for a gate you expect to exist here 🙂

releasenotes/notes/fix-adder-gates-39cf3d5f683e8880.yaml
```diff
@@ -0,0 +1,19 @@
+---
+fixes:
+  - |
+    Added default definitions for :class:`.FullAdderGate`, :class:`.HalfAdderGate`,
+    :class:`.ModularAdderGate` and :class:`.MultiplierGate` gates, allowing to 
+    contruct :class:`qiskit.quantum_info.Operator`\s from quantum circuits
```
Cryoris:
```suggestion
    contruct :class:`.Operator`\s from quantum circuits 
```

Cryoris (commented):
Two tiny comments on the test/reno, but the fix LGTM 👍🏻

releasenotes/notes/fix-adder-gates-39cf3d5f683e8880.yaml
alexanderivrii: Thanks! Done in [c39d08e](https://github.com/Qiskit/qiskit/pull/13530/commits/c39d08e0f6f91b1d80ddead3f15078b46219670a).

test/python/circuit/library/test_adders.py
alexanderivrii: Done in [c39d08e](https://github.com/Qiskit/qiskit/pull/13530/commits/c39d08e0f6f91b1d80ddead3f15078b46219670a).

Cryoris (approved):
LGTM thanks for the updates!

PR #13531, opened 2024-12-05
Update Split2QUnitaries to handle SWAP unitary gates

### LINKED ISSUE #13476: Add support for SWAP specialization in Split2qUnitaries
### What should we add?

In the split 2q unitaries transpiler pass right now we only look for identity equivalent unitaries where we return the two single qubit components from the two qubit weyl decomposition. But there is an extension to the pass where we can look at swap equivalent unitaries and simply swap the qubits the single qubit components execute on. There is a todo comment for this in the pass's code:

https://github.com/Qiskit/qiskit/blob/5f3f59469fcd9edae2feb80f74026265ebceb29a/crates/accelerate/src/split_2q_unitaries.rs#L79-L81

The trick here is if we do this we basically need to mirror the logic of elide permutations both by swapping the virtual qubits in the tracked permutation but also ensure that we swap the subsequent gates on those qubits.

gadial:
### Summary
Updates the `Split2QUnitaries` transpiler pass to handle 2-qubit unitary gates which can be decomposed as a `swap` gate and 1-qubit gates.

Closes #13476 

### Details and comments
This PR extends the `Split2QUnitaries` in the following manner:
1. During the initial pass `Split2QUnitaries` performs, it checks whether any unitary encountered is a SWAP gate.
2. If SWAP gates were found, another pass is performed on the DAG, replacing SWAP gates by the 1-qubit unitary gates in their decomposition (as is currently done in `Split2QUnitaries`) and instead of adding SWAP gates, performs a virtual swap by dynamically changing the qubit layout (affecting subsequent gates and the final layout) similar to what the `ElidePermutations` does.

Since step 2 requires generating a new DAG (as opposed to what `Split2QUnitaries` does to identity-equivalent unitary gates), it is not performed during the initial pass, only if SWAP gates were detected.


test/python/transpiler/test_split_2q_unitaries.py
```diff
@@ -266,3 +266,49 @@ def to_matrix(self):
...
+        qc.global_phase += 1.2345
+        qc_split = QuantumCircuit(2)
+        qc_split.append(UnitaryGate(Operator(qc)), [0, 1])
+
+        pm = PassManager()
+        pm.append(Collect2qBlocks())
+        pm.append(ConsolidateBlocks())
+        pm.append(Split2QUnitaries())
+        res = pm.run(qc_split)
+        res_op = Operator.from_circuit(res, final_layout=res.layout.final_layout)
```
mtreinish: Can you add a test that has this in it as part of a larger circuit that runs through the full transpilation pipeline? I want to make sure that we validating that we're tracking and composing the final permutations correctly.

test/python/transpiler/test_split_2q_unitaries.py
```diff
@@ -266,3 +266,49 @@ def to_matrix(self):
...
+        expected_op = Operator(qc_split)
+
+        self.assertTrue(expected_op.equiv(res_op))
+        self.assertTrue(
+            matrix_equal(expected_op.data, res_op.data, ignore_phase=False)
+        )
```
mtreinish: Can we add an assertion here that `swap` has been removed from the output?

test/python/transpiler/test_split_2q_unitaries.py
```diff
@@ -266,3 +266,49 @@ def to_matrix(self):
...
```
mtreinish: I would add an assertion that we've removed the swap gate from the circuit. This test could still pass if there was no transformation on the input circuit.

mtreinish (commented):
Thanks for doing this, from a quick glance the code looks good (but I only quickly skimmed it on my phone, I'll do a deeper reviewer later). But I had some quick comments on the tests to start.

alexanderivrii:
If I recall correctly, we don't run the `ElidePermutations` pass if the user explicitly sets the initial layout (to prevent the transpiler from messing up circuits that are already routed). Is this something we should worry about for this pass as well? Would it make sense to have an additional flow where the pass would replace a 2-qubit unitary by a SWAP + 1-qubit unitaries without merging the SWAP into the final permutation?

test/python/transpiler/test_split_2q_unitaries.py
gadial: I've added `self.assertNotIn("swap", res.count_ops())` but I don't think it's enough - there's also the possibility that we didn't split the unitary at all (note that it's also not being checked in the tests relating to the original split unitary code).

Any idea how to easily verify that the unitary was removed? Note that just using `count_ops` is tricky since we have a unitary both before (as the 2-qubit unitary) and after (as two individual 1-qubit unitaries). I can compare the count itself (1 before vs 2 after) but it feel less robust than it should.

test/python/transpiler/test_split_2q_unitaries.py
gadial: Done

test/python/transpiler/test_split_2q_unitaries.py
gadial: Done. Also added a smaller test which does non-trivial swapping (with swap gates removed using the elide permutation pass) before the unitary test which surfaced a possible bug in the layout composition phase. We need to discuss whether in this phase we should compose on the existing layout or (as was the case before my change) on its inverse.

crates/accelerate/src/split_2q_unitaries.rs
```diff
@@ -19,20 +21,73 @@ use qiskit_circuit::circuit_instruction::OperationFromPython;
 use qiskit_circuit::dag_circuit::{DAGCircuit, NodeType, Wire};
 use qiskit_circuit::imports::UNITARY_GATE;
 use qiskit_circuit::operations::{Operation, Param};
+use qiskit_circuit::packed_instruction::PackedInstruction;
+use qiskit_circuit::Qubit;
 
 use crate::two_qubit_decompose::{Specialization, TwoQubitWeylDecomposition};
 
+fn create_k1_gates<'a>(
+    decomp: &'a TwoQubitWeylDecomposition,
```
ShellyGarion: could you write a rust docstring for this function?

crates/accelerate/src/split_2q_unitaries.rs
```diff
@@ -19,20 +21,73 @@ use qiskit_circuit::circuit_instruction::OperationFromPython;
...
+        .get_bound(py)
+        .call((k1r_arr, py.None(), false), Some(&kwargs))?;
+    let k1l_gate = UNITARY_GATE
+        .get_bound(py)
+        .call((k1l_arr, py.None(), false), Some(&kwargs))?;
+    Ok((k1r_gate, k1l_gate))
+}
+
+fn add_new_op(
+    new_dag: &mut DAGCircuit,
```
ShellyGarion: could you write a rust docstring for this function?

crates/accelerate/src/split_2q_unitaries.rs
```diff
@@ -76,12 +125,78 @@ pub fn split_2q_unitaries(
...
+                let matrix = inst
+                    .op
+                    .matrix(inst.params_view())
+                    .expect("'unitary' gates should always have a matrix form");
+                let decomp = TwoQubitWeylDecomposition::new_inner(
+                    matrix.view(),
+                    Some(requested_fidelity),
+                    None,
+                )?;
+                if matches!(decomp.specialization, Specialization::SWAPEquiv) {
```
ShellyGarion, line 120: why do you need to check it again instead of checking that `has_swaps == true` ?

test/python/transpiler/test_split_2q_unitaries.py
```diff
@@ -266,3 +266,99 @@ def to_matrix(self):
...
+        pm.append(Split2QUnitaries())
+        res = pm.run(qc_split)
+        res_op = Operator.from_circuit(res)
+        expected_op = Operator(qc_split)
+
+        self.assertNotIn("swap", res.count_ops())
+        self.assertTrue(expected_op.equiv(res_op))
+        self.assertTrue(matrix_equal(expected_op.data, res_op.data, ignore_phase=False))
+
+    def test_2q_swap_with_non_unitary_swaps(self):
```
ShellyGarion, line 320: what is "non unitary" here?

qiskit/transpiler/passes/optimization/split_2q_unitaries.py
```diff
@@ -36,5 +37,21 @@ def __init__(self, fidelity: float = 1.0 - 1e-16):
 
     def run(self, dag: DAGCircuit) -> DAGCircuit:
         """Run the Split2QUnitaries pass on `dag`."""
-        split_2q_unitaries(dag, self.requested_fidelity)
-        return dag
+        result = split_2q_unitaries(dag, self.requested_fidelity)
```
ShellyGarion: Perhaps the docstring of the `Split2QUnitaries` class (lines 22-28) should also be updated?

ShellyGarion (commented):
Thanks for your contribution. I have a few minor comments.

crates/accelerate/src/split_2q_unitaries.rs
gadial: `has_swaps` is a non-gate specific boolean that tells us whether the circuit has swap-equivalent unitaries. If such gates are present, we iterate again over the entire circuit, rebuilding it as we go; as part of this process we need to identify the swap-equivalent gates as we iterate over them.

test/python/transpiler/test_split_2q_unitaries.py
gadial: Explicit `swap` gates, which are not handled by the new pass (they are handled by the elide permutation pass) as opposed to swap-equivalent unitary gates which are the target of the new pass.

qiskit/transpiler/passes/optimization/split_2q_unitaries.py
gadial: Done in 1d7b4f61a53c96a56090d392a6ca0423dc48ea20

crates/accelerate/src/split_2q_unitaries.rs
gadial: Done in 1d7b4f61a53c96a56090d392a6ca0423dc48ea20

crates/accelerate/src/split_2q_unitaries.rs
gadial: Done in 1d7b4f61a53c96a56090d392a6ca0423dc48ea20

crates/accelerate/src/split_2q_unitaries.rs
```diff
@@ -19,20 +21,77 @@ use qiskit_circuit::circuit_instruction::OperationFromPython;
...
+    py: Python,
+) -> PyResult<()> {
+    let inst = PackedInstruction {
+        op: new_op.operation,
+        qubits: new_dag.qargs_interner.insert_owned(qargs),
+        clbits: new_dag.cargs_interner.get_default(),
+        params: (!new_op.params.is_empty()).then(|| Box::new(new_op.params)),
+        extra_attrs: new_op.extra_attrs,
+        #[cfg(feature = "cache_pygates")]
+        py_op: std::sync::OnceLock::new(),
```
mtreinish: You do have the py op available for caching here since you are generating this via python. But in this case it doesn't actually matter too much because this is always going to be a `PyGate` and if the cache is `None` the python object reference will just be taken from that.

crates/accelerate/src/split_2q_unitaries.rs
```diff
@@ -76,12 +129,78 @@ pub fn split_2q_unitaries(
...
+                        &mut new_dag,
+                        k1l_gate.extract()?,
+                        vec![qubits[1]],
+                        &mapping,
+                        py,
+                    )?;
+                    new_dag.add_global_phase(py, &Param::Float(decomp.global_phase + PI4))?;
+                }
+            }
+            if !is_swap {
```
mtreinish: Instead of doing this with a mutable `is_swap` boolean variable, couldn't you just add a `continue` inside the if block for a swap equivalent unitary? and remove this `!is_swap` if statement. I feel like that might be a bit preferable because it reduce a layer of indentation here if nothing else. Although I suspect the compiler might be able to reason about the flow a bit better with a continue, but that would need to be validated and if true is of very minimal relevance to not be worth validating or benchmarking.

qiskit/transpiler/passes/optimization/split_2q_unitaries.py
```diff
@@ -24,17 +25,39 @@ class Split2QUnitaries(TransformationPass):
...
     simplified into two single qubit gates and this pass will perform this optimization and will
     replace the two qubit gate with two single qubit :class:`.UnitaryGate`.
+
+    If some of the gates can be viewed as a swap joined by the product of 2 single qubit gates,
+    the pass will recreate the DAG, permuting the swapped qubits similar
+    to how it's done in :class:`ElidePermutations`.
     """
 
-    def __init__(self, fidelity: float = 1.0 - 1e-16):
+    def __init__(self, fidelity: float = 1.0 - 1e-16, split_swap: bool = False):
```
mtreinish, line 34: By default this means we don't run the swap splitting, are you planning a follow-up where you update https://github.com/Qiskit/qiskit/blob/main/qiskit/transpiler/preset_passmanagers/builtin_plugins.py#L192-L194 to set it to True unless `routing=None` is specified?

qiskit/transpiler/passes/optimization/split_2q_unitaries.py
```diff
@@ -24,17 +25,39 @@ class Split2QUnitaries(TransformationPass):
...
         """Run the Split2QUnitaries pass on `dag`."""
-        split_2q_unitaries(dag, self.requested_fidelity)
-        return dag
+        result = split_2q_unitaries(dag, self.requested_fidelity, self.split_swap)
+        if result is None:
+            return dag
+
+        (new_dag, qubit_mapping) = result
+        input_qubit_mapping = {qubit: index for index, qubit in enumerate(dag.qubits)}
+        self.property_set["original_layout"] = Layout(input_qubit_mapping)
```
mtreinish, line 52: Does this behave correctly if we run elide permutations first? I realize this is just a straight copy of what elide permutations is doing, but I'm wondering if we need to check that we're not doing multiple permutations before applying the layout or do some other storing if we run these together.

releasenotes/notes/update-split-2q-unitaries-with-swap-557a1252e3208257.yaml
```diff
@@ -0,0 +1,8 @@
+---
+features_transpiler:
+  - |
+    The :class:`.Split2QUnitaries` transpiler pass has been upgraded to
+    handle the case where the unitary in consideration can be written
+    as a SWAP gate and two 1-qubit gates. In this case, it splits the
+    unitary and also applies virtual swapping similar to what is done in
+    :class:`.ElidePermutations`.
```
mtreinish: I would probably mention the new argument here to document that it is being introduced. Something like:

```suggestion
    :class:`.ElidePermutations`. This functionality can be controlled with a new argument,
    ``split_swap``, on the constructor of :class`.Split2QUnitaries` which can be used to disable
    splitting swap equivalent gates.
```
```

**Review by mtreinish — COMMENTED** [review]
Overall this looks excellent thanks for doing this, this looks about ready now. I just have a few small comments and questions inline. The biggest one I think is over how/when you're thinking we should integrate this new feature into the preset pass managers. It's fine if you'd like to do that in a separate follow-up PR to keep this PR to a single logical change (that's probably preferable actually). I just wanted to understand your thinking on that.

File: crates/accelerate/src/split_2q_unitaries.rs (hunk 9 above)
Comment (gadial) [reply crates/accelerate/src/split_2q_unitaries.rs]: So, do you think something needs to change here?

File: qiskit/transpiler/passes/optimization/split_2q_unitaries.py (hunk 11 above)
Comment (gadial) [reply qiskit/transpiler/passes/optimization/split_2q_unitaries.py:34]: That's exactly what I'm wondering about. If you think the specific logic you suggested here suffices, I'll add this as part of this PR.

Comment (mtreinish) [reply qiskit/transpiler/passes/optimization/split_2q_unitaries.py:34]: Yeah I think that's sufficient. You can mirror the exact logic that conditions the execution of `ElidePermutations` .

File: crates/accelerate/src/split_2q_unitaries.rs (hunk 9 above)
Comment (mtreinish) [reply crates/accelerate/src/split_2q_unitaries.rs]: I think this is fine as is. This was just more my idle musings.

File: qiskit/transpiler/passes/optimization/split_2q_unitaries.py (hunk 12 above)
Comment (gadial) [reply qiskit/transpiler/passes/optimization/split_2q_unitaries.py:52]: It should work - in theory. in `split_2q_unitaries.py` we have
```
if current_layout := self.property_set["virtual_permutation_layout"]:
            self.property_set["virtual_permutation_layout"] = new_layout.compose(
                current_layout, dag.qubits
            )
```

And `compose` works, if I recall correctly. However, this is not covered by the current test suite, and that's one of the things I wanted to ask about - should the "joint" test be placed in `TestSplit2QUnitaries` or perhaps a different location? What flows should we check, other than adding `Split2QUnitaries` before and after `ElidePermutations`?

File: crates/accelerate/src/split_2q_unitaries.rs (hunk 10 above)
Comment (gadial) [reply crates/accelerate/src/split_2q_unitaries.rs]: Done in a9bc3e1ea

File: qiskit/transpiler/passes/optimization/split_2q_unitaries.py (hunk 11 above)
Comment (gadial) [reply qiskit/transpiler/passes/optimization/split_2q_unitaries.py:34]: Done in a9bc3e1ea

File: releasenotes/notes/update-split-2q-unitaries-with-swap-557a1252e3208257.yaml (hunk 13 above)
Comment (gadial) [reply releasenotes/notes/update-split-2q-unitaries-with-swap-557a1252e3208257.yaml]: Done in a9bc3e1ea

File: qiskit/transpiler/passes/optimization/split_2q_unitaries.py (hunk 12 above)
Comment (mtreinish) [reply qiskit/transpiler/passes/optimization/split_2q_unitaries.py:52]: You can put it in `TestSplit2QUnitaries` that's fine if the intent is to validate the `Split2QUnitaries` logic is correct that's fine. But having the coverage in general is more important than where it's located specifically.

**Review by mtreinish — DISMISSED** [review]
I think the PR looks good now, thanks for all the work on it. I'll hold off on adding it to the merge queue if you want to push the extra test case in this PR. If you'd rather add it to a follow up PR though that's fine too since I think we have indirect coverage of it now that we're running with the new option in the preset pass manager. Having a direct test of the combination of two permutation passes back to back would be good to have direct validation though.

**Comment by gadial** [comment]
> I think the PR looks good now, thanks for all the work on it. I'll hold off on adding it to the merge queue if you want to push the extra test case in this PR. If you'd rather add it to a follow up PR though that's fine too since I think we have indirect coverage of it now that we're running with the new option in the preset pass manager. Having a direct test of the combination of two permutation passes back to back would be good to have direct validation though.

Thanks Matthew. I'll try adding something useful tomorrow and see if I succeed.

File: crates/accelerate/src/split_2q_unitaries.rs [hunk 14]
```diff
@@ -19,20 +21,77 @@ use qiskit_circuit::circuit_instruction::OperationFromPython;
...
+    let k1l_arr = decomp.K1l(py);
+    let kwargs = PyDict::new(py);
+    kwargs.set_item(intern!(py, "num_qubits"), 1)?;
+    let k1r_gate = UNITARY_GATE
+        .get_bound(py)
+        .call((k1r_arr, py.None(), false), Some(&kwargs))?;
+    let k1l_gate = UNITARY_GATE
+        .get_bound(py)
+        .call((k1l_arr, py.None(), false), Some(&kwargs))?;
+    Ok((k1r_gate, k1l_gate))
```
mtreinish: This actually needs an update now since #13759 I'll push up a small commit to fix it, since I caused the conflict.

mtreinish: Done in: https://github.com/Qiskit/qiskit/pull/13531/commits/0f84fc46abd2092521789b61996b28174061c9a3

mtreinish (approved):
I pushed up a fix for the merge conflicts caused by using the rust native UnitaryGate in the pass. As part of that process I caught a bug in the indexing in the pass. So I'm going to merge this without that one test. I'll open a tracking issue to add the test coverage so you can push a new PR when you have the test ready.

PR #13543, opened 2024-12-10
Invalidate `PackedInstruction` cache upon modification

### LINKED ISSUE #13504: Basis translation can fail if run in parallel

raynelfss:
### Summary
This should supercede #13532, and should correctly fix #13504
A previous mis-implementation allowed shared references between cloned instances of `PackedInstruction` when it came to their Python counterparts. The following commits re-implement the `Clone` trait and makes `PackedInstruction` attributes accessible only through views. This is done to allow the cached `PyGate` to be discarded upon modification of `op`, `params` or `extra_attrs` attributes of said instance of `PackedInstruction`.

### Details and comments
As referenced by @Cryoris in #13504, we were finding unknown parameters when performing translations in parallel. This was due to modifications coming from `BasisTranslator` helper function `compose_transforms` which would modify certain parameters from a circuit. The reason for this was that we had been unintentionally copying references to the cached python gates stored within these operation instances.

The following commits perform a couple of changes on the `PackedInstruction` class to prevent this from happening:
- Implement the `Clone` trait for `PackedInstruction` such that the cached `py_gate` reference is discarded.
- Control any modifications of the `cached_pygate` by discarding the cache whenever these attributes are borrowed mutably. This was done by:
    - Making all attributes of the `PackedInstruction` private.
    - Create getters and setters (more accurately mutable getters) that control when it is proper to discard the cached gate.

These changes had a propagated effect in the codebase and a lot of other pieces of code had to be modified for it to work with it.

### Notes:
- There are a couple of performance regressions related to circuit construction time. This is a tradeoff from having to rebuild the cached gate many more times than it was required before.

Co-authored-by: Julien Gacon <gaconju@gmail.com>


crates/circuit/src/packed_instruction.rs
```diff
@@ -518,6 +518,75 @@ pub struct PackedInstruction {
...
+    pub fn qubits(&self) -> Interned<[Qubit]> {
+        self.qubits
+    }
+
+    /// Retrieves an immutable reference to the index under which the interner has stored `clbits`.
+    pub fn clbits(&self) -> Interned<[Clbit]> {
+        self.clbits
+    }
+
+    pub fn extra_attrs(&self) -> &ExtraInstructionAttributes {
```
Cryoris: Did you want to add a docstring on this as well? (same for the `_mut` version)

crates/circuit/src/packed_instruction.rs
```diff
@@ -628,3 +715,23 @@ impl PackedInstruction {
...
 }
+
+// When cloning a `PackedInstruction` we will need to delete the old cached
+// `Gate` to avoid having duplicated references to the same python object.
+impl Clone for PackedInstruction {
+    fn clone(&self) -> Self {
+        Self {
+            op: self.op.clone(),
+            qubits: self.qubits,
+            clbits: self.clbits,
```
Cryoris, line 968: These don't have to be cloned because they are interned, right?

crates/circuit/src/packed_instruction.rs
Cryoris, line 1: `py_op` should also be private, or is there a reason to keep it `pub`?

crates/circuit/src/packed_instruction.rs
```diff
@@ -493,15 +493,15 @@ impl Drop for PackedOperation {
...
 /// A `PackedInstruction` in general cannot be safely mutated outside the context of its
 /// `CircuitData`, because the majority of the data is not actually stored here.
-#[derive(Clone, Debug)]
+#[derive(Debug)]
 pub struct PackedInstruction {
-    pub op: PackedOperation,
+    op: PackedOperation,
     /// The index under which the interner has stored `qubits`.
-    pub qubits: Interned<[Qubit]>,
+    qubits: Interned<[Qubit]>,
```
Cryoris: As discussed we could also keep qubits/clbits public, but I think I agree with your choice here to add methods, such that all members are used in the same fashion.

Cryoris:
While I'm not sure how to test the original bug in CI (since we turn Qiskit in parallel off) we can test @jakelishman's reproducer from: https://github.com/Qiskit/qiskit/pull/13532#issuecomment-2521565790 🙂

crates/circuit/src/packed_instruction.rs
raynelfss: They don't have to be cloned because I believe they implement the `Copy` trait, so it's just a matter of re-using the bits, no need to re-build them.

raynelfss:
These changes do fix things but there is a big performance regression in circuit construction. Since I removed a lot of the pre-caching that was being done prior to introducing these changes, I thought adding those back would fix the performance regression, however, local testing proved that including those made things even slower. **I will post those numbers later, but slowdowns were as high as 2x slower.** This is the current regression:

---
| Change   | Before [841cb29d]    | After [b337cf3e] <rebalance-packed-instructions>   |   Ratio | Benchmark (Parameter)                                                             |
|----------|----------------------|----------------------------------------------------|---------|-----------------------------------------------------------------------------------|
| +        | 60.9±2μs             | 93.0±4μs                                           |    1.53 | circuit_construction.CircuitConstructionBench.time_circuit_copy(8, 8192)          |
| +        | 215±4μs              | 315±6μs                                            |    1.47 | circuit_construction.CircuitConstructionBench.time_circuit_copy(20, 32768)        |
| +        | 216±9μs              | 314±9μs                                            |    1.46 | circuit_construction.CircuitConstructionBench.time_circuit_copy(8, 32768)         |
| +        | 838±30μs             | 1.21±0.01ms                                        |    1.44 | circuit_construction.CircuitConstructionBench.time_circuit_copy(1, 131072)        |
| +        | 60.9±1μs             | 87.3±2μs                                           |    1.43 | circuit_construction.CircuitConstructionBench.time_circuit_copy(14, 8192)         |
| +        | 223±10μs             | 315±10μs                                           |    1.41 | circuit_construction.CircuitConstructionBench.time_circuit_copy(1, 32768)         |
| +        | 215±6μs              | 301±8μs                                            |    1.4  | circuit_construction.CircuitConstructionBench.time_circuit_copy(5, 32768)         |
| +        | 61.4±1μs             | 85.5±7μs                                           |    1.39 | circuit_construction.CircuitConstructionBench.time_circuit_copy(2, 8192)          |
| +        | 1.32±0.02ms          | 1.80±0.06ms                                        |    1.37 | circuit_construction.CircuitConstructionBench.time_circuit_copy(14, 131072)       |
| +        | 220±6μs              | 300±8μs                                            |    1.36 | circuit_construction.CircuitConstructionBench.time_circuit_copy(2, 32768)         |
| +        | 19.3±0.4μs           | 26.1±0.5μs                                         |    1.35 | circuit_construction.CircuitConstructionBench.time_circuit_copy(1, 2048)          |
| +        | 223±6μs              | 301±10μs                                           |    1.35 | circuit_construction.CircuitConstructionBench.time_circuit_copy(14, 32768)        |
| +        | 20.6±0.4μs           | 27.9±1μs                                           |    1.35 | circuit_construction.CircuitConstructionBench.time_circuit_copy(5, 2048)          |
| +        | 22.1±0.9μs           | 29.8±1μs                                           |    1.35 | circuit_construction.CircuitConstructionBench.time_circuit_copy(8, 2048)          |
| +        | 60.0±3μs             | 79.3±4μs                                           |    1.32 | circuit_construction.CircuitConstructionBench.time_circuit_copy(1, 8192)          |
| +        | 1.33±0.02ms          | 1.75±0.09ms                                        |    1.32 | circuit_construction.CircuitConstructionBench.time_circuit_copy(2, 131072)        |
| +        | 61.0±2μs             | 80.4±4μs                                           |    1.32 | circuit_construction.CircuitConstructionBench.time_circuit_copy(5, 8192)          |
| +        | 64.7±1μs             | 84.5±0.9μs                                         |    1.31 | circuit_construction.CircuitConstructionBench.time_circuit_copy(20, 8192)         |
| +        | 24.0±0.8μs           | 31.0±2μs                                           |    1.29 | circuit_construction.CircuitConstructionBench.time_circuit_copy(14, 2048)         |
| +        | 20.9±0.6μs           | 27.0±0.7μs                                         |    1.29 | circuit_construction.CircuitConstructionBench.time_circuit_copy(2, 2048)          |
| +        | 1.44±0.08ms          | 1.83±0.07ms                                        |    1.27 | circuit_construction.CircuitConstructionBench.time_circuit_copy(8, 131072)        |
| +        | 1.34±0.02ms          | 1.67±0.03ms                                        |    1.25 | circuit_construction.CircuitConstructionBench.time_circuit_copy(20, 131072)       |
| +        | 26.0±0.5μs           | 32.0±0.7μs                                         |    1.23 | circuit_construction.CircuitConstructionBench.time_circuit_copy(20, 2048)         |
| +        | 1.37±0.06ms          | 1.68±0.1ms                                         |    1.23 | circuit_construction.CircuitConstructionBench.time_circuit_copy(5, 131072)        |
| +        | 5.92±0.07ms          | 6.90±0.3ms                                         |    1.16 | circuit_construction.ParamaterizedDifferentCircuit.time_DTC100_set_build(100, 10) |
| +        | 7.73±0.06ms          | 8.56±0.2ms                                         |    1.11 | circuit_construction.ParamaterizedDifferentCircuit.time_DTC100_set_build(50, 50)  |
| +        | 6.34±0.09μs          | 6.98±0.1μs                                         |    1.1  | circuit_construction.CircuitConstructionBench.time_circuit_copy(1, 128)           |

PERFORMANCE DECREASED.
---

With this comes the question of whether we should continue to explore ways of optimizing this and, were we not to find a way of optimizing, should we choose having a more stable solution in exchange of a potentially slower circuit builder?

raynelfss:
> While I'm not sure how to test the original bug in CI (since we turn Qiskit in parallel off) we can test @jakelishman's reproducer from: https://github.com/Qiskit/qiskit/pull/13532#issuecomment-2521565790 🙂

It seems to work well with this based on local tests.

raynelfss:
Other than the circuit instruction benchmarks, not much seems to have changed in the utility scale, although a couple benchmarks seem to have slowed down.

---
# Utility scale benchmarks:
| Change   | Before [841cb29d]    | After [b337cf3e] <rebalance-packed-instructions>   | Ratio   | Benchmark (Parameter)                                                         |
|----------|----------------------|----------------------------------------------------|---------|-------------------------------------------------------------------------------|
|          | 0                    | 0                                                  | n/a     | utility_scale.UtilityScaleBenchmarks.track_bvlike_depth('cx')                 |
|          | 0                    | 0                                                  | n/a     | utility_scale.UtilityScaleBenchmarks.track_bvlike_depth('cz')                 |
|          | 0                    | 0                                                  | n/a     | utility_scale.UtilityScaleBenchmarks.track_bvlike_depth('ecr')                |
|          | 328±8ms              | 343±10ms                                           | 1.05    | utility_scale.UtilityScaleBenchmarks.time_qaoa('ecr')                         |
|          | 345±8ms              | 357±4ms                                            | 1.03    | utility_scale.UtilityScaleBenchmarks.time_qaoa('cz')                          |
|          | 500±10ms             | 513±10ms                                           | 1.03    | utility_scale.UtilityScaleBenchmarks.time_qft('cx')                           |
|          | 7.12±0.1ms           | 7.24±0.2ms                                         | 1.02    | utility_scale.UtilityScaleBenchmarks.time_bvlike('cz')                        |
|          | 2.77±0.01s           | 2.82±0.02s                                         | 1.02    | utility_scale.UtilityScaleBenchmarks.time_circSU2('cz')                       |
|          | 2.77±0.04s           | 2.79±0.05s                                         | 1.01    | utility_scale.UtilityScaleBenchmarks.time_circSU2('ecr')                      |
|          | 4.86±0.2ms           | 4.93±0.1ms                                         | 1.01    | utility_scale.UtilityScaleBenchmarks.time_parse_qaoa_n100('ecr')              |
|          | 51.7±0.8ms           | 52.1±0.9ms                                         | 1.01    | utility_scale.UtilityScaleBenchmarks.time_parse_qft_n100('cz')                |
|          | 51.5±1ms             | 52.2±0.6ms                                         | 1.01    | utility_scale.UtilityScaleBenchmarks.time_parse_qft_n100('ecr')               |
|          | 16.5±0.4ms           | 16.6±0.1ms                                         | 1.01    | utility_scale.UtilityScaleBenchmarks.time_parse_square_heisenberg_n100('cx')  |
|          | 279±3ms              | 283±4ms                                            | 1.01    | utility_scale.UtilityScaleBenchmarks.time_qaoa('cx')                          |
|          | 608±20ms             | 614±20ms                                           | 1.01    | utility_scale.UtilityScaleBenchmarks.time_qft('cz')                           |
|          | 103±1ms              | 103±2ms                                            | 1.00    | utility_scale.UtilityScaleBenchmarks.time_bv_100('cz')                        |
|          | 2.79±0.03s           | 2.78±0.01s                                         | 1.00    | utility_scale.UtilityScaleBenchmarks.time_circSU2('cx')                       |
|          | 51.6±0.8ms           | 51.6±1ms                                           | 1.00    | utility_scale.UtilityScaleBenchmarks.time_parse_qft_n100('cx')                |
|          | 16.6±0.5ms           | 16.6±0.4ms                                         | 1.00    | utility_scale.UtilityScaleBenchmarks.time_parse_square_heisenberg_n100('cz')  |
|          | 16.4±0.4ms           | 16.5±0.4ms                                         | 1.00    | utility_scale.UtilityScaleBenchmarks.time_parse_square_heisenberg_n100('ecr') |
|          | 661±20ms             | 659±10ms                                           | 1.00    | utility_scale.UtilityScaleBenchmarks.time_qv('cx')                            |
|          | 733±20ms             | 737±20ms                                           | 1.00    | utility_scale.UtilityScaleBenchmarks.time_qv('ecr')                           |
|          | 390                  | 390                                                | 1.00    | utility_scale.UtilityScaleBenchmarks.track_bv_100_depth('cx')                 |
|          | 397                  | 397                                                | 1.00    | utility_scale.UtilityScaleBenchmarks.track_bv_100_depth('cz')                 |
|          | 397                  | 397                                                | 1.00    | utility_scale.UtilityScaleBenchmarks.track_bv_100_depth('ecr')                |
|          | 300                  | 300                                                | 1.00    | utility_scale.UtilityScaleBenchmarks.track_circSU2_depth('cx')                |
|          | 300                  | 300                                                | 1.00    | utility_scale.UtilityScaleBenchmarks.track_circSU2_depth('cz')                |
|          | 300                  | 300                                                | 1.00    | utility_scale.UtilityScaleBenchmarks.track_circSU2_depth('ecr')               |
|          | 1607                 | 1607                                               | 1.00    | utility_scale.UtilityScaleBenchmarks.track_qaoa_depth('cx')                   |
|          | 1622                 | 1622                                               | 1.00    | utility_scale.UtilityScaleBenchmarks.track_qaoa_depth('cz')                   |
|          | 1622                 | 1622                                               | 1.00    | utility_scale.UtilityScaleBenchmarks.track_qaoa_depth('ecr')                  |
|          | 1954                 | 1954                                               | 1.00    | utility_scale.UtilityScaleBenchmarks.track_qft_depth('cx')                    |
|          | 1954                 | 1954                                               | 1.00    | utility_scale.UtilityScaleBenchmarks.track_qft_depth('cz')                    |
|          | 1954                 | 1954                                               | 1.00    | utility_scale.UtilityScaleBenchmarks.track_qft_depth('ecr')                   |
|          | 2709                 | 2709                                               | 1.00    | utility_scale.UtilityScaleBenchmarks.track_qv_depth('cx')                     |
|          | 2709                 | 2709                                               | 1.00    | utility_scale.UtilityScaleBenchmarks.track_qv_depth('cz')                     |
|          | 2709                 | 2709                                               | 1.00    | utility_scale.UtilityScaleBenchmarks.track_qv_depth('ecr')                    |
|          | 462                  | 462                                                | 1.00    | utility_scale.UtilityScaleBenchmarks.track_square_heisenberg_depth('cx')      |
|          | 462                  | 462                                                | 1.00    | utility_scale.UtilityScaleBenchmarks.track_square_heisenberg_depth('cz')      |
|          | 462                  | 462                                                | 1.00    | utility_scale.UtilityScaleBenchmarks.track_square_heisenberg_depth('ecr')     |
|          | 98.2±3ms             | 97.7±1ms                                           | 0.99    | utility_scale.UtilityScaleBenchmarks.time_bv_100('cx')                        |
|          | 103±1ms              | 102±2ms                                            | 0.99    | utility_scale.UtilityScaleBenchmarks.time_bv_100('ecr')                       |
|          | 7.20±0.2ms           | 7.13±0.2ms                                         | 0.99    | utility_scale.UtilityScaleBenchmarks.time_bvlike('cx')                        |
|          | 4.90±0.1ms           | 4.87±0.08ms                                        | 0.99    | utility_scale.UtilityScaleBenchmarks.time_parse_qaoa_n100('cz')               |
|          | 125±3ms              | 124±1ms                                            | 0.99    | utility_scale.UtilityScaleBenchmarks.time_square_heisenberg('cz')             |
|          | 124±3ms              | 123±2ms                                            | 0.99    | utility_scale.UtilityScaleBenchmarks.time_square_heisenberg('ecr')            |
|          | 5.00±0.04ms          | 4.89±0.07ms                                        | 0.98    | utility_scale.UtilityScaleBenchmarks.time_parse_qaoa_n100('cx')               |
|          | 611±20ms             | 602±10ms                                           | 0.98    | utility_scale.UtilityScaleBenchmarks.time_qft('ecr')                          |
|          | 760±20ms             | 739±3ms                                            | 0.97    | utility_scale.UtilityScaleBenchmarks.time_qv('cz')                            |
|          | 7.48±0.1ms           | 7.19±0.2ms                                         | 0.96    | utility_scale.UtilityScaleBenchmarks.time_bvlike('ecr')                       |
|          | 115±3ms              | 110±2ms                                            | 0.96    | utility_scale.UtilityScaleBenchmarks.time_square_heisenberg('cx')             |

---
# Transpiler levels

| Change   | Before [841cb29d]    | After [b337cf3e] <rebalance-packed-instructions~1>   |   Ratio | Benchmark (Parameter)                                                                           |
|----------|----------------------|------------------------------------------------------|---------|-------------------------------------------------------------------------------------------------|
| +        | 40.3±2ms             | 49.5±0.2ms                                           |    1.23 | transpiler_levels.TranspilerLevelBenchmarks.time_schedule_qv_14_x_14(0)                         |
| +        | 40.5±1ms             | 49.2±0.2ms                                           |    1.21 | transpiler_levels.TranspilerLevelBenchmarks.time_schedule_qv_14_x_14(1)                         |
| -        | 12.3±0.2ms           | 11.1±0.2ms                                           |    0.9  | transpiler_levels.TranspilerLevelBenchmarks.time_transpile_from_large_qasm_backend_with_prop(3) |

PERFORMANCE DECREASED.
___

raynelfss:
There seems to be a strange effect in which some benchmarks improve while others get worse, which is strange:
| Change   | Before [841cb29d]    | After [590b4a3a]    |   Ratio | Benchmark (Parameter)                                                                                  |
|----------|----------------------|---------------------|---------|--------------------------------------------------------------------------------------------------------|
| +        | 16.7±0.5ms           | 22.4±0.3ms          |    1.35 | transpiler_qualitative.TranspilerQualitativeBench.time_transpile_time_qft_16(2, 'sabre', 'dense')      |
| +        | 38.5±0.2ms           | 50.2±0.5ms          |    1.3  | transpiler_levels.TranspilerLevelBenchmarks.time_schedule_qv_14_x_14(0)                                |
| +        | 40.9±0.9ms           | 49.6±0.2ms          |    1.21 | transpiler_levels.TranspilerLevelBenchmarks.time_schedule_qv_14_x_14(1)                                |
| -        | 24.0±0.5ms           | 16.5±0.4ms          |    0.69 | transpiler_qualitative.TranspilerQualitativeBench.time_transpile_time_qft_16(3, 'sabre', 'sabre')      |
| -        | 23.3±0.2ms           | 15.7±0.1ms          |    0.67 | transpiler_qualitative.TranspilerQualitativeBench.time_transpile_time_qft_16(3, 'stochastic', 'dense') |
| -        | 24.4±0.5ms           | 15.8±0.2ms          |    0.65 | transpiler_qualitative.TranspilerQualitativeBench.time_transpile_time_qft_16(3, 'sabre', 'dense')      |

PERFORMANCE DECREASED.

crates/circuit/src/circuit_data.rs
```diff
@@ -398,17 +391,18 @@ impl CircuitData {
...
-            inst.params = (!result.params.is_empty()).then(|| Box::new(result.params));
-            inst.extra_attrs = result.extra_attrs;
+            *inst.op_mut() = result.operation;
+            *inst.params_mut() =
+                (!result.params.is_empty()).then_some(result.params.clone().into());
+            *inst.extra_attrs_mut() = result.extra_attrs;
             #[cfg(feature = "cache_pygates")]
             {
-                inst.py_op = py_op.unbind().into();
+                *inst.py_op_mut() = py_op.unbind().into();
```
kevinhartman, line 639: If we consider constructing a new `PackedInstruction` with a user-provided Python op for the cache as unsafe, then we should also consider this to be unsafe, correct? Since the instance we get by calling `func` is something the user may have a reference to.

crates/circuit/src/dag_circuit.rs
```diff
@@ -1762,15 +1762,13 @@ def _format(operand):
...
                 .cargs_interner
                 .insert_owned(self.clbits.map_bits(cargs.iter().flatten())?.collect());
-            let instr = PackedInstruction {
-                op: py_op.operation,
-                qubits: qubits_id,
-                clbits: clbits_id,
-                params: (!py_op.params.is_empty()).then(|| Box::new(py_op.params)),
-                extra_attrs: py_op.extra_attrs,
-                #[cfg(feature = "cache_pygates")]
-                py_op: op.unbind().into(),
```
kevinhartman, line 1483: Do we actually need to throw away the `py_op` when creating new `PackedInstruction` instances? 

Surely we anticipated that it would be an issue for Python-side code to modify the cached objects at the time we introduced the `py_op` cache. That makes me think we should preserve `py_op` as part of `PackedOperation::new`. Isn't the issue only via the default `Clone` for `PackedInstruction`?

I realize that @jakelishman is on holiday, but perhaps if he sees this he can clarify.

crates/circuit/src/packed_instruction.rs
```diff
@@ -514,10 +514,93 @@ pub struct PackedInstruction {
...
+    /// Retrieves the cached py_gate immutably.
+    #[cfg(feature = "cache_pygates")]
+    pub fn py_op(&self) -> &OnceLock<PyObject> {
+        &self.py_op
+    }
+
+    // Setters
+
+    /// Retrieves an immutable reference to the instruction's underlying operation.
+    pub fn op_mut(&mut self) -> &mut PackedOperation {
```
kevinhartman, line 756: It looks to me that all of these `_mut` methods are only used within the `qiskit-circuit` crate. Should we introduce these with visibility `pub(crate)` instead?

Originally `PackedInstruction` was more of a Rust POD, but now it feels like we should lock down its internals where possible since it is managing its own cache invalidation.

crates/circuit/src/packed_instruction.rs
```diff
@@ -514,10 +514,93 @@ pub struct PackedInstruction {
...
+    /// Retrieves a mutable reference to the extra_attributes of this instruction.
+    pub fn extra_attrs_mut(&mut self) -> &mut ExtraInstructionAttributes {
+        #[cfg(feature = "cache_pygates")]
+        {
+            self.py_op.take();
+        }
+        &mut self.extra_attrs
+    }
+
+    /// Retrieves the cached py_gate immutably.
```
kevinhartman:
```suggestion
    /// Retrieves the cached py_gate mutably.
```

crates/circuit/src/packed_instruction.rs
```diff
@@ -514,10 +514,93 @@ pub struct PackedInstruction {
...
+    /// Retrieves an immutable reference to the instruction's underlying operation.
+    pub fn op_mut(&mut self) -> &mut PackedOperation {
+        #[cfg(feature = "cache_pygates")]
+        {
+            self.py_op.take();
+        }
+        &mut self.op
+    }
+
+    /// Retrieves an immutable reference to the index under which the interner has stored `qubits`.
```
kevinhartman:
```suggestion
    /// Retrieves a mutable reference to the index under which the interner has stored `qubits`.
```

crates/circuit/src/packed_instruction.rs
```diff
@@ -514,10 +514,93 @@ pub struct PackedInstruction {
...
+
+    /// Retrieves the cached py_gate immutably.
+    #[cfg(feature = "cache_pygates")]
+    pub fn py_op(&self) -> &OnceLock<PyObject> {
+        &self.py_op
+    }
+
+    // Setters
+
+    /// Retrieves an immutable reference to the instruction's underlying operation.
```
kevinhartman:
```suggestion
    /// Retrieves a mutable reference to the instruction's underlying operation.
```

crates/circuit/src/packed_instruction.rs
```diff
@@ -514,10 +514,93 @@ pub struct PackedInstruction {
...
+            qubits,
+            clbits,
+            params: params.map(Box::new),
+            extra_attrs,
+            #[cfg(feature = "cache_pygates")]
+            py_op: OnceLock::new(),
+        }
+    }
+
+    // Getters
```
kevinhartman: I don't think we have a pattern for labeling sections, and it's probably easy for these to fall out of sync with the code. I'd recommend we drop them ^^

crates/circuit/src/packed_instruction.rs
```diff
@@ -628,3 +729,23 @@ impl PackedInstruction {
         }
     }
 }
+
+// When cloning a `PackedInstruction` we will need to delete the old cached
+// `Gate` to avoid having duplicated references to the same python object.
```
kevinhartman: It might be something other than a `Gate` as well, so I think we should use the term "operation" here instead. And, perhaps it'd be better to say that we "explicitly implement `Clone` to avoid copying the source operation's cached Python op", since the term "delete" may not be specific enough for an unfamiliar reader.

crates/circuit/src/packed_instruction.rs
```diff
@@ -628,3 +729,23 @@ impl PackedInstruction {
...
+impl Clone for PackedInstruction {
+    fn clone(&self) -> Self {
+        Self {
+            op: self.op.clone(),
+            qubits: self.qubits,
+            clbits: self.clbits,
+            params: self.params.clone(),
+            extra_attrs: self.extra_attrs.clone(),
+            #[cfg(feature = "cache_pygates")]
+            py_op: OnceLock::new(),
```
kevinhartman, line 972: Toward the beginning of the discussion [here](https://github.com/Qiskit/qiskit/pull/13532#issuecomment-2521505541), @jakelishman says:

> the right solution is to manually impl Clone for PackedInstruction to not propagate the cache for gates that aren't Python-space singletons.

Should we be preserving the `py_op` in the case of singleton gates to avoid a more significant performance regression?

kevinhartman (commented):
This looks good so far, but I've noted a few concerns inline.

Specifically, it isn't entirely clear to me when we can preserve the `py_op`, if at all. In its current state, it looks like this PR always throws it out, even when building new instances of `PackedInstruction` via `::new`, and even when cloning a singleton gate. If we're doing so unnecessarily, that is certainly performance we're leaving (/putting back) on the table. Do you know if that's the case @raynelfss?

It'd be great if @jakelishman or @mtreinish can confirm prior to the potential inclusion of this PR in 1.3.1 this week, though they're both on holiday.

raynelfss:
@kevinhartman I'd honestly be okay with not keeping an initialized `py_op` in any case whatsoever. We can have it be a cached instance that's only built when needed and discard it anytime we modify a `parameter`, `op`, or `extra_attrs`.  I feel this makes much more sense as it would produce more stable results. I wonder how the rest of the team feels about this.

As for what happens if we bring back the deleted assigned `py_op` instances, the performance is actually way worse (though I'm not quite sure why that is). I made a divergent branch not too long ago in which I added `py_op` as an argument for `PackedInstruction::new()` and it resulted in an even greater performance regression during circuit construction:

---
| Change   | Before [84def520] <rebalance-packed-instructions>   | After [d6923131] <packed-with-cache>   |   Ratio | Benchmark (Parameter)                                                       |
|----------|-----------------------------------------------------|----------------------------------------|---------|-----------------------------------------------------------------------------|
| +        | 121±10μs                                            | 187±40μs                               |    1.54 | circuit_construction.CircuitConstructionBench.time_circuit_copy(8, 8192)    |
| +        | 110±2μs                                             | 138±10μs                               |    1.25 | circuit_construction.CircuitConstructionBench.time_circuit_copy(5, 8192)    |
| +        | 373±20μs                                            | 454±20μs                               |    1.22 | circuit_construction.CircuitConstructionBench.time_circuit_copy(20, 32768)  |
| +        | 114±3μs                                             | 133±0.6μs                              |    1.17 | circuit_construction.CircuitConstructionBench.time_circuit_copy(14, 8192)   |
| +        | 112±4μs                                             | 131±3μs                                |    1.17 | circuit_construction.CircuitConstructionBench.time_circuit_copy(20, 8192)   |
| +        | 33.8±2μs                                            | 38.8±0.6μs                             |    1.15 | circuit_construction.CircuitConstructionBench.time_circuit_copy(14, 2048)   |
| +        | 37.4±1μs                                            | 43.0±2μs                               |    1.15 | circuit_construction.CircuitConstructionBench.time_circuit_copy(20, 2048)   |
| +        | 3.19±0.05ms                                         | 3.56±0.03ms                            |    1.12 | circuit_construction.CircuitConstructionBench.time_circuit_copy(20, 131072) |
| +        | 3.22±0.01ms                                         | 3.57±0.03ms                            |    1.11 | circuit_construction.CircuitConstructionBench.time_circuit_copy(1, 131072)  |
| +        | 3.15±0.03ms                                         | 3.50±0.01ms                            |    1.11 | circuit_construction.CircuitConstructionBench.time_circuit_copy(5, 131072)  |
| +        | 3.21±0.05ms                                         | 3.54±0.03ms                            |    1.1  | circuit_construction.CircuitConstructionBench.time_circuit_copy(2, 131072)  |
| +        | 3.18±0.03ms                                         | 3.51±0.04ms                            |    1.1  | circuit_construction.CircuitConstructionBench.time_circuit_copy(8, 131072)  |

PERFORMANCE DECREASED.

---

It's very puzzling as to why this is happening, but if we could make the attribute fully private to the `PackedInstruction` instance and have it only generate it when needed, it would make a lot more sense to keep this around.

crates/circuit/src/circuit_data.rs
raynelfss: I agree with your reasoning, I'm okay with making the `py_op` attribute fully invisible since it is still unsafe (even when immutably borrowed).

crates/circuit/src/dag_circuit.rs
raynelfss: Yes, but also think of it this way. 

If you look at the `PackedInstruction::unpack_py_op()` method, in which a reference is built to a python object (basically what we do when the `cache_pygate` flag is disabled), in the case of a standard gate, we build a new instance of py_op based on the current parameters or the extra attributes of our gate, and not only that, but the instance we get from this function whenever dealing with any other type of gate comes directly from the `OperationRef` type if it contains a python object within and isn't really affected by changes unless our `op` type is `StandardGate`.

Why do I mention this? Well, whenever a gate is initialized without a cached pygate, this is the function in charge of building that gate for us. Since this gate is disconnected and changes won't propagate, we'd need to update it by creating a new one whenever our params or attributes are modified (or at least that's what I understand from here).

We could technically optimize further by only re-building if we have an instance of `Standard` gate, since what we do in other cases is a call to `clone_ref(py)`. What do you think?

releasenotes/notes/fix-packed-instruction-f2620d5aff61269d.yaml
```diff
@@ -0,0 +1,7 @@
+---
+fixes:
+  - |
+    ``PackedInstruction``, the principal representation of an instance of :class:`.CircuitInstruction`
+    in Rust, has been re-organized to discard its cached instance of a python :class:`.Gate` whenever its
+    operation, parameters, or extra attributes are modified. This results in ``PackedInstruction`` no
+    no longer having shared references between them.
```
alexanderivrii: "no no"

crates/circuit/src/packed_instruction.rs
raynelfss: Sure, this makes more sense

crates/circuit/src/packed_instruction.rs
raynelfss: How would we be able to identify which ones are singletons?

crates/circuit/src/dag_circuit.rs
jakelishman: Python doesn't enforce ownership in the way Rust does, but we can still think of it in similar terms. `DAGCircuit.apply_operation_back` has always behaved _as if_ calling it transferred ownership of the gate object, as opposed to `QuantumCircuit.append`, which copied the object if it detected that you might want to modify it (i.e. it was parametric).

Since the Python object coming in is logically "owned" by the function now, I think it's fine to put it in the cache.

With all the Python stuff, we can't compile-time (or even, in most cases, run-time) forbid modifications to Python objects, so there's always going to be the possibility for a user to break us a bit, and some of our Python-based will be pragmatics.  I think in this case, saving the incoming object in the cache is no more dangerous than having the cache at all, so it should be fine to keep.

crates/circuit/src/packed_instruction.rs
jakelishman: I think making the methods `pub(crate)` is a bit needlessly restrictive.  Anywhere you can get a `&mut PackedInstruction`, you can effectively change any fields you need to, because you can `*instruction = my_new_packed_instruction`, so the actual external access restriction for the circuit needs to be purely managed by when you can get a `&mut PackedInstruction`.

Otherwise, it's just needlessly restrictive - if you can create a new `PackedInstruction` with `pub` visibility, then making these methods `pub(crate)` doesn't actually restrict any capability.

crates/circuit/src/packed_instruction.rs
jakelishman: I was talking mostly about what's _necessary_, rather than what's easy.  Technically it's safe to propagate the cache if the Python object itself is immutable.  If it's too hard to tell, or we deem it too risky, we can just zero them all out.

Fwiw, we'd basically just apply a Rust-only heuristic: check the `PackedOperation` is a `Standard` and it's a 0-parameter gate with `extra_attrs == None`.  That won't catch every single singleton, but it'll do the vast majority and fails safe.

jakelishman:
I'm very surprised to see performance dropping if some of the caches are being propagated but not all - it might be a good idea to try and dig heavily into that and really understand _why_ it's happened, especially since this is being considered for a bugfix.

Fwiw, this feels like a very major change to be targetted at a bugfix to me, especially a release that might go out tomorrow.  It might worth considering squashing the `BasisTranslator` problem in a less ideal way for the bugfix release (like just having `assign_parameters` in Rust space clear the pyop cache rather than attempting to mutate the Python object) that we know is safe and doesn't have the same not-quite-understood performance drawbacks.  But of course, I'm still on leave and don't have anywhere near the full context for anything going on here, so you guys should actually make the decision, not me.

kevinhartman:
> Fwiw, this feels like a very major change to be targetted at a bugfix to me, especially a release that might go out tomorrow. It might worth considering squashing the BasisTranslator problem in a less ideal way for the bugfix release (like just having assign_parameters in Rust space clear the pyop cache rather than attempting to mutate the Python object) that we know is safe and doesn't have the same not-quite-understood performance drawbacks. But of course, I'm still on leave and don't have anywhere near the full context for anything going on here, so you guys should actually make the decision, not me.

We're all in agreement that this PR should not merge until after 1.3.1. @raynelfss is working on a separate more localized hotfix for #13504 for this release, and we'll revisit this one in the future when we have time to understand the performance characteristics.

1ucian0:
Given https://github.com/Qiskit/qiskit/issues/13504, removing the `priority:high` tag

raynelfss:
It seems I resolved some of the issues with the `circuit_construction` benchmarks by cloning the parameters raw. Apparently, cloning the parameters from the views seems to be slower, although I can't quite comprehend the exact reason causing this, but the iteration seemed to take longer. However, `transpiler_levels` benchmarks seem to still be doing not too well. Now the performance hit is not as big.

## `transpiler_level`

All benchmarks:

| Change   | Before [94504f21] <rebalance-packed-instructions^2>   | After [c818a4ff] <rebalance-packed-instructions>   |   Ratio | Benchmark (Parameter)                                                                                  |
|----------|-------------------------------------------------------|----------------------------------------------------|---------|--------------------------------------------------------------------------------------------------------|
| +        | 50.7±0.2ms                                            | 68.6±0.2ms                                         |    1.35 | transpiler_levels.TranspilerLevelBenchmarks.time_schedule_qv_14_x_14(0)                                |
| +        | 50.8±0.8ms                                            | 65.7±0.2ms                                         |    1.29 | transpiler_levels.TranspilerLevelBenchmarks.time_schedule_qv_14_x_14(1)                                |
|          | 27.6±0.1ms                                            | 28.3±0.8ms                                         |    1.03 | transpiler_levels.TranspilerLevelBenchmarks.time_transpile_qv_14_x_14(2)                               |
|          | 18.6±0.1ms                                            | 18.7±0.3ms                                         |    1.01 | transpiler_levels.TranspilerLevelBenchmarks.time_transpile_qv_14_x_14(1)                               |

PERFORMANCE DECREASED.

---
## `circuit_construction`

All benchmarks:

| Change   | Before [94504f21] <rebalance-packed-instructions^2>   | After [c818a4ff] <rebalance-packed-instructions>   | Ratio   | Benchmark (Parameter)                                                                                           |
|----------|-------------------------------------------------------|----------------------------------------------------|---------|-----------------------------------------------------------------------------------------------------------------|
|          | 306±60μs                                              | 721±200μs                                          | ~2.36   | circuit_construction.CircuitConstructionBench.time_circuit_copy(8, 32768)                                       |
|          | 559±200μs                                             | 724±200μs                                          | ~1.30   | circuit_construction.CircuitConstructionBench.time_circuit_copy(2, 32768)                                       |
|          | 28.0±1μs                                              | 32.8±2μs                                           | ~1.17   | circuit_construction.CircuitConstructionBench.time_circuit_copy(5, 2048)                                        |
|          | 28.1±5μs                                              | 25.2±0.4μs                                         | ~0.90   | circuit_construction.CircuitConstructionBench.time_circuit_copy(1, 2048)                                        |
| +        | 159±10μs                                              | 181±3μs                                            | 1.14    | circuit_construction.CircuitConstructionBench.time_circuit_copy(1, 8192)                                        |
| +        | 2.32±0.02ms                                           | 2.58±0.04ms                                        | 1.11    | circuit_construction.ParamaterizedDifferentCircuit.time_DTC100_set_build(50, 10)                                |
| -        | 177±4μs                                               | 97.0±2μs                                           | 0.55    | circuit_construction.CircuitConstructionBench.time_circuit_copy(5, 8192)                                        |

PERFORMANCE DECREASED.

raynelfss:
It seems that the `transpiler_level` issues have also been fixed by cloning the parameters directly from their `Box`. The one non deterministic test that can perform either better or worse in some cases is circuit_construction. Still quite unsure about what causes it to behave this way 🤔. However, the performance hit is minimal as it is only off by a couple microseconds.

## `circuit_construction`

| Change   | Before [1c213c2e] <rebalance-packed-instructions>   | After [bff7017a] <fix-asv-transpiler-level>   |   Ratio | Benchmark (Parameter)                                                              |
|----------|-----------------------------------------------------|-----------------------------------------------|---------|------------------------------------------------------------------------------------|
| +        | 56.8±0.7μs                                          | 67.2±2μs                                      |    1.18 | circuit_construction.CircuitConstructionBench.time_circuit_copy(8, 8192)           |
| +        | 213±9μs                                             | 238±8μs                                       |    1.12 | circuit_construction.CircuitConstructionBench.time_circuit_copy(8, 32768)          |
| -        | 311±4ms                                             | 282±5ms                                       |    0.91 | circuit_construction.CircuitConstructionBench.time_circuit_construction(5, 131072) |
| -        | 139±5μs                                             | 126±5μs                                       |    0.91 | circuit_construction.ParamaterizedDifferentCircuit.time_QV100_build(50, 10)        |
| -        | 377±8μs                                             | 341±3μs                                       |    0.9  | circuit_construction.ParamaterizedDifferentCircuit.time_QV100_build(50, 50)        |
| -        | 801±10μs                                            | 693±5μs                                       |    0.87 | circuit_construction.ParamaterizedDifferentCircuit.time_QV100_build(50, 150)       |
| -        | 146±4μs                                             | 115±8μs                                       |    0.79 | circuit_construction.ParamaterizedDifferentCircuit.time_QV100_build(10, 50)        |
| -        | 278±7μs                                             | 207±8μs                                       |    0.75 | circuit_construction.ParamaterizedDifferentCircuit.time_QV100_build(10, 150)       |

PERFORMANCE DECREASED.

raynelfss:
Here's one uninterrupted run of asv with anything that has a circuit in it

| Change   | Before [547a9295] <rebalance-packed-instructions^2>   | After [b54265f1] <rebalance-packed-instructions>   |   Ratio | Benchmark (Parameter)                                                         |
|----------|-------------------------------------------------------|----------------------------------------------------|---------|-------------------------------------------------------------------------------|
| +        | 213±3μs                                               | 286±7μs                                            |    1.34 | circuit_construction.ParamaterizedDifferentCircuit.time_QV100_build(10, 150)  |
| +        | 691±5μs                                               | 796±5μs                                            |    1.15 | circuit_construction.ParamaterizedDifferentCircuit.time_QV100_build(50, 150)  |
| +        | 1.33±0.02ms                                           | 1.52±0.05ms                                        |    1.14 | circuit_construction.ParamaterizedDifferentCircuit.time_QV100_build(100, 150) |
| -        | 51.3±0.7μs                                            | 46.4±0.4μs                                         |    0.91 | converters.ConverterBenchmarks.time_circuit_to_dag(5, 128)                    |
| -        | 23.4±0.08μs                                           | 21.2±0.2μs                                         |    0.91 | converters.ConverterBenchmarks.time_dag_to_circuit(2, 128)                    |
| -        | 189±0.4μs                                             | 170±0.2μs                                          |    0.9  | converters.ConverterBenchmarks.time_circuit_to_dag(1, 2048)                   |
| -        | 512±10μs                                              | 461±0.6μs                                          |    0.9  | converters.ConverterBenchmarks.time_circuit_to_dag(5, 2048)                   |
| -        | 782±4μs                                               | 701±4μs                                            |    0.9  | converters.ConverterBenchmarks.time_circuit_to_dag(8, 2048)                   |
| -        | 16.7±0.1μs                                            | 15.1±0.1μs                                         |    0.9  | converters.ConverterBenchmarks.time_dag_to_circuit(1, 128)                    |
| -        | 1.53±0.02ms                                           | 1.38±0.05ms                                        |    0.9  | converters.ConverterBenchmarks.time_dag_to_circuit(14, 2048)                  |
| -        | 288±6μs                                               | 258±1μs                                            |    0.9  | converters.ConverterBenchmarks.time_dag_to_circuit(32, 128)                   |
| -        | 465±20μs                                              | 419±2μs                                            |    0.9  | converters.ConverterBenchmarks.time_dag_to_circuit(53, 128)                   |
| -        | 280±1μs                                               | 250±1μs                                            |    0.89 | converters.ConverterBenchmarks.time_circuit_to_dag(2, 2048)                   |
| -        | 116±0.6μs                                             | 103±6μs                                            |    0.89 | converters.ConverterBenchmarks.time_dag_to_circuit(1, 2048)                   |
| -        | 223±8μs                                               | 196±2μs                                            |    0.88 | converters.ConverterBenchmarks.time_dag_to_circuit(2, 2048)                   |
| -        | 66.9±0.5μs                                            | 59.0±0.3μs                                         |    0.88 | converters.ConverterBenchmarks.time_dag_to_circuit(8, 128)                    |
| -        | 119±2μs                                               | 103±4μs                                            |    0.87 | converters.ConverterBenchmarks.time_dag_to_circuit(14, 128)                   |
| -        | 3.41±0.04ms                                           | 2.95±0.06ms                                        |    0.87 | converters.ConverterBenchmarks.time_dag_to_circuit(8, 8192)                   |
| -        | 878±9μs                                               | 748±4μs                                            |    0.85 | converters.ConverterBenchmarks.time_dag_to_circuit(2, 8192)                   |
| -        | 513±8μs                                               | 437±5μs                                            |    0.85 | converters.ConverterBenchmarks.time_dag_to_circuit(5, 2048)                   |
| -        | 2.11±0.01ms                                           | 1.79±0.02ms                                        |    0.85 | converters.ConverterBenchmarks.time_dag_to_circuit(5, 8192)                   |
| -        | 433±7μs                                               | 361±5μs                                            |    0.83 | converters.ConverterBenchmarks.time_dag_to_circuit(1, 8192)                   |
| -        | 848±40μs                                              | 700±20μs                                           |    0.83 | converters.ConverterBenchmarks.time_dag_to_circuit(8, 2048)                   |

PERFORMANCE DECREASED.

**Utility, transpiler_qualitative, transpiler_levels remain unchanged.**

crates/circuit/src/packed_instruction.rs
```diff
@@ -702,10 +702,94 @@ pub struct PackedInstruction {
...
+    }
+
+    /// Retrieves the cached py_gate immutably.
+    #[cfg(feature = "cache_pygates")]
+    #[inline]
+    pub fn py_op(&self) -> &OnceLock<PyObject> {
+        &self.py_op
+    }
+
+    /// Retrieves a mutable reference to the instruction's underlying operation.
```
kevinhartman, line 754: Perhaps we should document the cache invalidation here?

crates/circuit/src/packed_instruction.rs
```diff
@@ -681,14 +681,14 @@ impl Drop for PackedOperation {
...
 pub struct PackedInstruction {
-    pub op: PackedOperation,
+    op: PackedOperation,
     /// The index under which the interner has stored `qubits`.
     pub qubits: Interned<[Qubit]>,
     /// The index under which the interner has stored `clbits`.
     pub clbits: Interned<[Clbit]>,
-    pub params: Option<Box<SmallVec<[Param; 3]>>>,
+    params: Option<Box<SmallVec<[Param; 3]>>>,
     pub label: Option<Box<String>>,
```
kevinhartman, line 692: If this is public, then someone can mutate it if they have a `&mut PackedInstruction` (which it looks like we don't want for `label` since `label_mut` invalidated the cache.

Perhaps it's cleanest to just make all of these private?

crates/circuit/src/packed_instruction.rs
```diff
@@ -702,10 +702,94 @@ pub struct PackedInstruction {
...
+        {
+            self.py_op.take();
+        }
+        self.label.as_deref_mut()
+    }
+
+    /// Retrieves the cached py_gate mutably.
+    #[cfg(feature = "cache_pygates")]
+    #[inline]
+    pub fn py_op_mut(&mut self) -> &mut OnceLock<PyObject> {
```
kevinhartman, line 789: If the user can get the cache mutably, then they can also set it to be something that doesn't match with the other fields. I suppose they can also pass a cached object that doesn't match to `PackedInstruction::new` as well.

crates/circuit/src/packed_instruction.rs
```diff
@@ -848,3 +957,23 @@ impl PackedInstruction {
...
+impl Clone for PackedInstruction {
+    fn clone(&self) -> Self {
+        Self {
+            op: self.op.clone(),
+            qubits: self.qubits,
+            clbits: self.clbits,
+            params: self.params.clone(),
+            label: self.label.clone(),
+            #[cfg(feature = "cache_pygates")]
+            py_op: OnceLock::new(),
```
kevinhartman, line 972: @raynelfss I think it's probably worth doing this. Maybe something like this (untested):

```suggestion
            py_op: if matches!(self.op.view(), OperationRef::StandardGate(_))
                && self.params.is_none() && self.label.is_none() {
                self.py_op.clone()
            } else {
                OnceLock::new(),
            }
```

crates/circuit/src/packed_instruction.rs
```diff
@@ -848,3 +957,23 @@ impl PackedInstruction {
...
+        }
+    }
+
+    fn clone_from(&mut self, source: &Self) {
+        *self = source.clone()
+    }
```
kevinhartman, line 978:
```suggestion
```

I think this is the default implementation provided by the `Clone` trait itself, so no need to provide.

releasenotes/notes/fix-packed-instruction-f2620d5aff61269d.yaml
```diff
@@ -0,0 +1,7 @@
+---
+fixes:
+  - |
+    ``PackedInstruction``, the principal representation of an instance of :class:`.CircuitInstruction`
+    in Rust, has been re-organized to discard its cached instance of a python :class:`.Gate` whenever its
+    operation, parameters, or extra attributes are modified. This results in ``PackedInstruction`` no
+    longer having shared references between them.
```
kevinhartman, line 7: For the changelog, we should keep it relevant to the user impact (i.e. explain the bug we're fixing) and avoid details about the Rust-side implementation. None of these Rust structs are part of our public interface, so we shouldn't mention them.

Edit: the original bug is fixed, so I think we'd probably just delete this release note 🙂.

kevinhartman (commented):
@raynelfss and I have chatted a bit offline. I've suggested the following:

> But basically, I think the clearing of the cache on clone is good and we should absolutely do that. But, since we’re giving `PackedInstruction` the responsibility of managing the validity of its own cache, it feels a little weird to me that we’ve got a method like `py_op_mut` that lets the user change the cache, and I’m also not especially keen on the other `_mut` methods “taking” the cache on access (it feels a little obscure).
>
> What I’m starting to wonder is if we really need `PackedInstruction` to be mutable _at all_. Obviously, we need to do things like set the `params` during parameter assignment, but I wonder if it’d be better to have a method like `PackedInstruction::with_params(self, params: Box<...>) -> PackedInstruction` that consumes the original and returns a new `PackedInstruction` with the correct parameters and cleared cache. Same for other methods, e.g. `PackedInstruction::with_label(self, …)`.
>
> So the “contract” would be that the user uses `PackedInstruction::new(…, py_op)` to initially create a `PackedInstruction` with an assumed valid cache, and then if they want to modify anything, they call one of these `with_xyz` methods which creates a new `PackedInstruction` with just `xyz` changed and the cache reset. They would then need to overwrite the original packed instruction, e.g. `*instr = instr.with_params(smallvec![Param::Float(2.0)])`.
>
> The benefit to all of this would be that `PackedInstruction` is immutable once it gets created, so no need to fuss with the cache or have all of these different `_mut` methods. If it gets cloned, then the cloned version won’t have a cache (unless perhaps its a standard gate without params and label). If a `with_xyz` method is called, then the result won’t have a cache either.

I think if we're going to do anything for this release, we might want to just do the invalidation of the cache on clone, and leave the rest for a follow up. The control flow PR series I'm working on also makes some significant changes to how we track `Param`s, so it might be better for that to land first.

raynelfss:
Moving this to `to_defer` as it is in risk of not being included in 2.1

raynelfss:
Closing this as a new approach might require a full overhaul. Splitting off to different series of PRs starting with #14680.

PR #13561, opened 2024-12-12
Drop support for macOS 10.12 from wheel build.

kevinhartman:
### Summary
We'd originally expected to build wheels for x86_64 macOS 10.12 in #13470, but it turns out that macOS 13 cannot build anything older than `10.13`:

Screenshot 2024-12-12 at 4 18 23 PM

The automatic bump-up to `11` on ARM is presumably something we expect. Correct me if I'm wrong.

### Details and comments
This was noticed during the release of Qiskit 1.3.1.

Oddly, the wheel built for macOS ARM uses the bumped-up version of `11`:

https://pypi.org/project/qiskit/#qiskit-1.3.1-cp39-abi3-macosx_11_0_arm64.whl

But the x86_64 wheel uses the version `10.12` (even though supposedly it would've been built for `10.13`). I'm not sure what will happen, then, if this wheel gets pulled on macOS `10.12`. Unfortunately, the included release note should be present in the 1.3.1 release, but is not. Perhaps we can add it manually.

https://pypi.org/project/qiskit/#qiskit-1.3.1-cp39-abi3-macosx_10_12_x86_64.whl

### DIFF
--- a/.github/workflows/wheels-build.yml
+++ b/.github/workflows/wheels-build.yml
@@ -121,7 +121,7 @@ jobs:
           cat >>"$GITHUB_ENV" <<EOF
           CIBW_BEFORE_BUILD=bash ./tools/build_pgo.sh $PGO_WORK_DIR $PGO_OUT_PATH
           CIBW_ENVIRONMENT=RUSTUP_TOOLCHAIN=stable RUSTFLAGS='-Cprofile-use=$PGO_OUT_PATH -Cllvm-args=-pgo-warn-missing-function'
-          CIBW_ENVIRONMENT_MACOS=MACOSX_DEPLOYMENT_TARGET='10.12' RUSTUP_TOOLCHAIN=stable RUSTFLAGS='-Cprofile-use=$PGO_OUT_PATH -Cllvm-args=-pgo-warn-missing-function'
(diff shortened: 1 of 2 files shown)


jakelishman:
There's a bit more info on why this happens here: https://github.com/Qiskit/qiskit/pull/13503#issuecomment-2507718533.

kevinhartman:
> There's a bit more info on why this happens here: https://github.com/Qiskit/qiskit/pull/13503#issuecomment-2507718533.

@jakelishman thanks! Any idea why the wheel itself built with the 10.13 target still comes out as 10.12+ compatible? Will that actually work if someone running 10.12 pulls it down?

jakelishman:
I don't, unfortunately - it probably just still happens to only choose APIs that were still present in 10.12.  Given that we have very little interaction with the OS from our extension modules, I wouldn't be surprised if our exposure is mostly just limited to the particular formats / initialisation routines in the dynamic libraries, and it's unlikely that macOS 10.13 _completely_ overhauled everything, I guess?

jakelishman:
Or of course, it's not actually compatible with 10.12, and the audit tool just doesn't catch it.

kevinhartman:
Closing this out since we haven't ended up needing it.

PR #13567, opened 2024-12-16
Fix a docstring rst link rendering within an inline literal

eliarbel:
Changing from this: 

image

to this: 

image

If someone knows if there is a way to properly render an rst link within an inline literal text, I'd love to know.

### DIFF
--- a/qiskit/transpiler/target.py
+++ b/qiskit/transpiler/target.py
@@ -725,7 +725,7 @@ class Target(BaseTarget):
 
     @property
     def instructions(self):
-        """Get the list of tuples ``(:class:`~qiskit.circuit.Instruction`, (qargs))``
+        """Get the list of tuples (:class:`~qiskit.circuit.Instruction`, (qargs))
         for the target
 
         For globally defined variable width operations the tuple will be of the form


qiskit/transpiler/target.py
```diff
@@ -725,7 +725,7 @@ def operation_names(self):
 
     @property
     def instructions(self):
-        """Get the list of tuples ``(:class:`~qiskit.circuit.Instruction`, (qargs))``
+        """Get the list of tuples (:class:`~qiskit.circuit.Instruction`, (qargs))
```
ElePT, line 728: Thanks for addressing this! What do you think of maintaining the code formatting for the rest of the tuple? It would look like this:

Screenshot 2024-12-16 at 13 23 06

```suggestion
        """Get the list of tuples ``(``:class:`~qiskit.circuit.Instruction` ``, (qargs))``
```

It's not perfect but I find it a bit easier to read, probably personal choice.

eliarbel: Bikeshedding alert :-) before I opened the PR we had a little poll (of 3) here at the lab which included your suggestion and the applied fix was chosen unanimously. Personally I find it slightly cleaner visually.

ElePT: ok, I am not to stand against unanimous votes. This change also looks good :)

ElePT (approved):

Issue #13428, opened 2024-11-13
Optimize consecutive RZZ gates into a single RZZ gate


t-imamichi:
### What should we add?

Since the fractional gates are available, RZZ can be used as a basis gate.
But, optimization of consecutive RZZ gates does not seem to be enough as follows.
It would be nice to merge them into a single RZZ.

```python
from qiskit import generate_preset_pass_manager, QuantumCircuit

qc = QuantumCircuit(2)
qc.rzz(0.1, 0, 1)
qc.rzz(0.2, 0, 1)

pm = generate_preset_pass_manager(optimization_level=3, basis_gates=["rzz"])
qc2 = pm.run(qc)
print(qc2)
```

output (both 1.2.4 and main branch)
```
q_0: ─■─────────■────────
      │ZZ(0.1)  │ZZ(0.2)
q_1: ─■─────────■────────
```

melechlapson:
Is this issue limited to the rzz gates and not other gates like ryy rzx etc.?

t-imamichi:
Not limited to RZZ. But I don't mind RYY and RZX because RZZ becomes a basis gate of devices that supports fractional gates. RYY and RZX are not basis gates of any device.
https://docs.quantum.ibm.com/guides/fractional-gates

ShellyGarion:
There is a synthesis algorithm `TwoQubitControlledUDecomposer` that takes a 2q-unitary and synthesizes it using the 2q basis gate `RZZGate` (with angles in the range [-pi/2, pi/2]. This algorithm has recently been ported to rust in #13139. 

Here is an example of the output of this algorithm on the circuit suggested below:
```
        from qiskit.quantum_info import Operator
        from qiskit.synthesis.two_qubit.two_qubit_decompose import TwoQubitControlledUDecomposer
        unitary = Operator(qc)
        decomposer = TwoQubitControlledUDecomposer(RZZGate)
        qc3 = decomposer(unitary)
        print (qc3)
```
outputs:
```
global phase: -π
     ┌─────────┐┌──────────┐                       ┌─────────┐┌──────────┐»
q_0: ┤ Ry(π/2) ├┤ Ry(-π/2) ├─────────────■─────────┤ Ry(π/2) ├┤ Ry(-π/2) ├»
     └┬────────┤├──────────┤┌──────────┐ │ZZ(-0.3) ├─────────┤└┬────────┬┘»
[...]

PR #13568, opened 2024-12-16
Add 2q fractional gates to the `UnitarySynthesis` transpiler pass

### LINKED ISSUE #13320: Add `TwoQubitControlledUDecomposer` to the `UnitarySynthesis` tranpiler pass
### What should we add?

Add the `TwoQubitControlledUDecomposer` to the `UnitarySynthesis` tranpiler pass.
This should be the default pass for certain gates (`RXX, RYY, RZZ, CRZ, RZX, CRZ, CPhase`) with an arbitrary angle.
This depends on PRs https://github.com/Qiskit/qiskit/pull/13139 and https://github.com/Qiskit/qiskit/pull/13141.

ShellyGarion:
### Summary
close #13320

With the introduction of new 2q-fractional gates, such as `RZZGate` as part of the QPU, we add them to the `UnitarySynthesis` transpiler pass.
https://www.ibm.com/quantum/blog/fractional-gates


qiskit/transpiler/passes/synthesis/unitary_synthesis.py
```diff
@@ -107,7 +129,12 @@ def _choose_kak_gate(basis_gates):
     """Choose the first available 2q gate to use in the KAK decomposition."""
```
gadial, line 128: Question: It seems we are not choosing the element according to some order, but using set `pop`, which is random, leading to nondeterministic behavior. Can this be avoided (or already avoided and I'm missing something)? e.g.

```python
kak_gate = sorted([KAK_GATE_PARAM_NAMES[kak_gate_param] for kak_gate_params in kak_gates_params])[0]
```

ElePT: This is a good point @gadial, I had always assumed that this randomness was a reflection of the fact that there can be several equally valid gates. And if that's case, I wonder if we should favor the one that comes first in alphabetical order? At the same time, I am not sure how conscious some of these decisions were. There are other parts of the code (the parts that were migrated to Rust) where we actually rely on insertion order when there are several valid decompositions, so at that point I guess it's as arbitrary as sorting alphabetically. It might be also worth noting that these cases don't come up often in real scenarios because we rarely see a basis set with more than one 2q gate (although it can definitely happen).

qiskit/synthesis/two_qubit/two_qubit_decompose.py
```diff
@@ -290,7 +290,9 @@ def __init__(self, rxx_equivalent_gate: Type[Gate]):
         self.rxx_equivalent_gate = rxx_equivalent_gate
         self.scale = self._inner_decomposition.scale
 
-    def __call__(self, unitary: Operator | np.ndarray, *, atol=DEFAULT_ATOL) -> QuantumCircuit:
+    def __call__(
+        self, unitary: Operator | np.ndarray, approximate=False, use_dag=False, atol=DEFAULT_ATOL
+    ) -> QuantumCircuit:
```
ElePT: I wonder what's the motivation for this change.  Are `approximate` or `use_dag` used in the code? Could it be an oversight?

qiskit/transpiler/passes/synthesis/unitary_synthesis.py
```diff
@@ -151,14 +178,17 @@ def _decomposer_2q_from_basis_gates(basis_gates, pulse_optimize=None, approximat
...
+    #    backup_optimizer = TwoQubitBasisDecomposer(
+    #        CXGate(),
+    #        basis_fidelity=basis_fidelity,
+    #        euler_basis=euler_basis,
+    #        pulse_optimize=pulse_optimize,
+    #    )
+    #    decomposer2q = XXDecomposer(euler_basis=euler_basis, backup_optimizer=backup_optimizer)
+
+    if kak_gate in KAK_GATE_PARAM_NAMES.values():
+        decomposer2q = TwoQubitControlledUDecomposer(kak_gate)
```
ElePT: Double checking our discussion from earlier in the week: for the target-aware code we discussed adding a third check (apart from `is_controlled` or `is_supercontrolled`) to differentiate cases for the 3 available decomposers: `TwoQubitDecomposer`, `XXDecomposer`, and `TwoQubitControlledUDecomposer`. However, for the Python code (just basis gates), it looks like we are simply substituting the `XXDecomposer` for the `TwoQubitControlledUDecomposer`. Is this the plan?

ElePT (commented):
Hi @ShellyGarion! thanks for opening the PR, I left a couple of preliminary comments. I hope they make sense :)

qiskit/transpiler/passes/synthesis/unitary_synthesis.py
ShellyGarion: Thanks for the comment. I'll look into it.

qiskit/synthesis/two_qubit/two_qubit_decompose.py
ShellyGarion: when calling the `decomposer2q` later in the code it assumes to have both parameters, see e.g 
https://github.com/Qiskit/qiskit/blob/17648ebb030c90fa7a595333b61823735275f68f/qiskit/transpiler/passes/synthesis/unitary_synthesis.py#L1046

qiskit/transpiler/passes/synthesis/unitary_synthesis.py
gadial: Thank you @ElePT. My approach is usually to prefer the arbitrary to the nondeterminisitic, since nondeterministic behavior makes testing more difficult (since it's harder to reproduce problems) but also less resilient (since we are only testing the conditions generated nondeterminisically on the current testing machines, whereas a future version of Python might behave differently).

qiskit/transpiler/passes/synthesis/unitary_synthesis.py
ShellyGarion: In the case of basis gates, I'm not sure if we can have 3 options or only 2 (this is why this code is only commented out and not removed yet).
If a user provides the basis gates `[rzz, rz, ry]` how can we tell if they want an `RZZGate` with an arbitrary angle or an `RZZGate` with a fixed angle (and if so - which fixed value)?

ShellyGarion: Another comment: I ran the tests in `qiskit/test/python/transpiler/test_unitary_synthesis.py` (main branch), and couldn't find any test that goes into the following "if" statement (lines 181-188 above), so I wonder if the `XXDecomposer` here has ever been used.

qiskit/transpiler/passes/synthesis/unitary_synthesis.py
ShellyGarion: I agree with @gadial's comment. If the user provides several 2q KAK gates, this "pop" just randomly chooses one of them. We would probably want a more deterministic behavior (perhaps favoring more "commonly" used 2q gates).

qiskit/transpiler/passes/synthesis/unitary_synthesis.py
ElePT: I am not surprised, the coverage of these tests is not good because a lot of the tests are in the synthesis folder and not testing the transpilation pipeline, but calling the synthesis methods directly. Maybe this is a good opportunity to do some test refactoring before merging this PR. During the Rust migration I also noticed there weren't any tests for the `UnitarySynthesis` class that used the QSD method with a target and had to add a couple of them. I think it would make sense to ensure that we have a couple of tests per decomposer, and that there are sufficient tests for the target (HW-aware) path (this I could do in a separate PR because I didn't have time to look more into it during the Rust migration).

qiskit/synthesis/two_qubit/two_qubit_decompose.py
ElePT: I see. I keep forgetting that we still have python code for the target aware case (this is kept in case someone uses a non-default plugin). In that case I think we should respect the keyword-only argument:

```suggestion
    def __call__(
        self, unitary: Operator | np.ndarray, approximate=False, use_dag=False, *, atol=DEFAULT_ATOL
    ) -> QuantumCircuit:
```

I also have realized that the `TwoQubitControlledUDecomposer` is not documented. If we are going to add it to the default pipeline, I believe we should, as right now it's not part of the public interface. So we should probably make sure to add it to the API docs with the other decomposers and add a release note listing the class as a new feature.

qiskit/transpiler/passes/synthesis/unitary_synthesis.py
ElePT: I think that sounds reasonable, it helps with benchmarking and bug-catching.

qiskit/transpiler/passes/synthesis/unitary_synthesis.py
ElePT: > If a user provides the basis gates [rzz, rz, ry] how can we tell if they want an RZZGate with an arbitrary angle or an RZZGate with a fixed angle?

I think this is up to us to decide and document the behavior change. I am not sure how many users actually rely or expect the fixed angle?

qiskit/synthesis/two_qubit/two_qubit_decompose.py
ShellyGarion: I plan to add ` TwoQubitControlledUDecomposer` to the API docs following this PR (as I may need to update the API as part of adding it to the transpiler pass)

qiskit/transpiler/passes/synthesis/unitary_synthesis.py
ElePT: FYI, I started the test cleanup with this PR: https://github.com/Qiskit/qiskit/pull/13582, that is not much but I think can be helpful for future development. Let me know if you have any thoughts on it!

qiskit/transpiler/passes/synthesis/unitary_synthesis.py
ShellyGarion: done in [eeff4cd](https://github.com/Qiskit/qiskit/pull/13568/commits/eeff4cd23503c0cd9ffbb678286b9c14891c69d7)

qiskit/transpiler/passes/synthesis/unitary_synthesis.py
ShellyGarion: Another question is what to expect when given a parametrized gate in the target? see e.g this [test](https://github.com/Qiskit/qiskit/blob/079f69f7dd906299dfdf8c1262fa9bc1d75e478b/test/python/transpiler/test_unitary_synthesis.py#L1026).

Currently, we replace the parametrized gates by gates with a constant angle, without even checking if this angle is actually calibrated in the target:
https://github.com/Qiskit/qiskit/blob/079f69f7dd906299dfdf8c1262fa9bc1d75e478b/qiskit/transpiler/passes/synthesis/unitary_synthesis.py#L801

Again, I could not find any test that checks the `XXDecomposer` with a given target, namely the tests don't go into [here](https://github.com/Qiskit/qiskit/blob/079f69f7dd906299dfdf8c1262fa9bc1d75e478b/qiskit/transpiler/passes/synthesis/unitary_synthesis.py#L941).

ShellyGarion: It seems that [`_decomposer_2q_from_target`](https://github.com/Qiskit/qiskit/blob/8d8da66272b44f0868106142999ea348188545ad/qiskit/transpiler/passes/synthesis/unitary_synthesis.py#L786) is not being tested at all, since the tests for `UnitarySynthesis` pass for the target use the rust code.
Should this function be removed or updated to include `TwoQubitControlledUDecomposer` ? (and then some tests should be added).

crates/accelerate/src/unitary_synthesis.rs
```diff
@@ -687,19 +745,48 @@ fn get_2q_decomposers_from_target(
...
     }
 
     if check_goodbye(&available_basis_set) {
         return Ok(Some(decomposers));
     }
 
+    for basis_1q in &available_1q_basis {
+        for (_basis_2q, gate) in available_2q_param_basis.iter() {
+            let decomposer = TwoQubitControlledUDecomposer::new(
+                py,
```
ShellyGarion: is there a way to avoid using `py` here? 
This is since `TwoQubitControlledUDecomposer` does not have a purely rust `new_inner`, only `new` that requires `py` in order to handle `rxx_equivalent_gate` of type `RXXEquivalent`.

crates/accelerate/src/unitary_synthesis.rs
```diff
@@ -881,6 +968,7 @@ fn preferred_direction(
 }
 
 fn synth_su4_sequence(
+    py: Python,
```
ShellyGarion: is there a way to avoid using `py` here?

crates/accelerate/src/unitary_synthesis.rs
```diff
@@ -920,6 +1011,7 @@ fn synth_su4_sequence(
                     };
                     if synth_dir != preferred_dir {
                         reversed_synth_su4_sequence(
+                            py,
```
ShellyGarion, line 1014: is there a way to avoid using `py` here?

crates/accelerate/src/unitary_synthesis.rs
```diff
@@ -934,6 +1026,7 @@ fn synth_su4_sequence(
 }
 
 fn reversed_synth_su4_sequence(
+    py: Python,
```
ShellyGarion: is there a way to avoid using `py` here?

crates/accelerate/src/unitary_synthesis.rs
```diff
@@ -889,6 +977,9 @@ fn synth_su4_sequence(
     let is_approximate = approximation_degree.is_none() || approximation_degree.unwrap() != 1.0;
     let synth = if let DecomposerType::TwoQubitBasisDecomposer(decomp) = &decomposer_2q.decomposer {
         decomp.call_inner(su4_mat.view(), None, is_approximate, None)?
+    } else if let DecomposerType::TwoQubitControlledUDecomposer(decomp) = &decomposer_2q.decomposer
+    {
+        decomp.call_inner(py, su4_mat.view(), DEFAULT_ATOL)?
```
ShellyGarion: is there a way to avoid using `py` here?
This is since `TwoQubitControlledUDecomposer` has `py` in `call_inner`, in order to handle `rxx_equivalent_gate` of type `RXXEquivalent`.

qiskit/synthesis/two_qubit/two_qubit_decompose.py
```diff
@@ -270,32 +270,46 @@ class TwoQubitControlledUDecomposer:
     :math:`U \sim U_d(\alpha, 0, 0) \sim \text{Ctrl-U}`
     gate that is locally equivalent to an :class:`.RXXGate`."""
 
-    def __init__(self, rxx_equivalent_gate: Type[Gate]):
+    def __init__(self, rxx_equivalent_gate: Type[Gate], euler_basis: str = "ZYZ"):
         r"""Initialize the KAK decomposition.
```
ShellyGarion, line 274: should we replace the default value to "ZXZ" to match the current fractional gates `rx` and `rz` ?
(If so, we should do this more generally in other places of the code in this file)

crates/accelerate/src/two_qubit_decompose.rs
```diff
@@ -2751,11 +2751,17 @@ impl TwoQubitControlledUDecomposer {
...
     ///      rxx_equivalent_gate: Gate that is locally equivalent to an :class:`.RXXGate`:
     ///      :math:`U \sim U_d(\alpha, 0, 0) \sim \text{Ctrl-U}` gate.
+    ///     euler_basis: Basis string to be provided to :class:`.OneQubitEulerDecomposer`
+    ///     for 1Q synthesis.
     ///  Raises:
     ///      QiskitError: If the gate is not locally equivalent to an :class:`.RXXGate`.
     #[new]
-    #[pyo3(signature=(rxx_equivalent_gate))]
-    pub fn new(py: Python, rxx_equivalent_gate: RXXEquivalent) -> PyResult<Self> {
+    #[pyo3(signature=(rxx_equivalent_gate, euler_basis="ZYZ"))]
```
ShellyGarion: should we replace the default value here to "ZXZ"? If so, we should better do this in other places of this file.

crates/accelerate/src/two_qubit_decompose.rs
```diff
@@ -2583,14 +2584,14 @@ impl TwoQubitControlledUDecomposer {
...
-            global_phase += unitary_k1r.global_phase;
+            global_phase -= unitary_k1r.global_phase;
             for gate in unitary_k1r.gates.into_iter().rev() {
                 let (inv_gate_name, inv_gate_params) = invert_1q_gate(gate);
                 gates.push((Some(inv_gate_name), inv_gate_params, smallvec![0]));
             }
         }
         if let Some(unitary_k1l) = unitary_k1l {
-            global_phase += unitary_k1l.global_phase;
+            global_phase -= unitary_k1l.global_phase;
```
ElePT, line 2612: Was this sign change from an oversight from the previous code? If it is, it would be good to have a reproducing test to make sure the behavior is tracked.

crates/accelerate/src/two_qubit_decompose.rs
```diff
@@ -2728,13 +2728,13 @@ impl TwoQubitControlledUDecomposer {
...
         if let Some(unitary_c1r) = unitary_c1r {
-            global_phase -= unitary_c1r.global_phase;
+            global_phase += unitary_c1r.global_phase;
             for gate in unitary_c1r.gates.into_iter() {
                 gates1.gates.push((Some(gate.0), gate.1, smallvec![0]));
             }
         }
         if let Some(unitary_c1l) = unitary_c1l {
-            global_phase -= unitary_c1l.global_phase;
+            global_phase += unitary_c1l.global_phase;
```
ElePT, line 2812: I imagine this sign change is related to the previous one?

qiskit/transpiler/passes/synthesis/unitary_synthesis.py
ShellyGarion: Since the function [_decomposer_2q_from_target](https://github.com/Qiskit/qiskit/blob/8d8da66272b44f0868106142999ea348188545ad/qiskit/transpiler/passes/synthesis/unitary_synthesis.py#L786) is part of the `DefaultUnitarySynthesis` plugin we don't update it as part of this PR. We should consider porting this code to rust in a follow-up PR.

crates/accelerate/src/two_qubit_decompose.rs
ShellyGarion: I replaced the default to "ZXZ". Should we do it in other classes as well?

crates/accelerate/src/unitary_synthesis.rs
ShellyGarion: I replaced `new` by `TwoQubitControlledUDecomposer::new_inner` but still `py` is needed

qiskit/synthesis/two_qubit/two_qubit_decompose.py
ShellyGarion: I replaced "ZYZ" to "ZXZ". Should we also do it in other classes ?

t-imamichi:
I think this closes #13428. How about adding a test case of consecutive RZZ (RXX, and RYY) gates?

crates/accelerate/src/two_qubit_decompose.rs
```diff
@@ -2743,19 +2743,13 @@ impl TwoQubitControlledUDecomposer {
...
-    #[new]
-    #[pyo3(signature=(rxx_equivalent_gate))]
-    pub fn new(py: Python, rxx_equivalent_gate: RXXEquivalent) -> PyResult<Self> {
+    /// Initialize the KAK decomposition.
+    pub fn new_inner(
+        py: Python,
+        rxx_equivalent_gate: RXXEquivalent,
+        euler_basis: &str,
+    ) -> PyResult<Self> {
         let atol = DEFAULT_ATOL;
```
ElePT: I know we discussed the addition of a `new_inner` class, but the current implementation doesn't seem to add (or remove) anything to the existing `new` class, so I would just keep and use `new` in the rust code.

crates/accelerate/src/two_qubit_decompose.rs
```diff
@@ -2445,14 +2445,16 @@ impl RXXEquivalent {
...
 #[pyclass(module = "qiskit._accelerate.two_qubit_decompose", subclass)]
 pub struct TwoQubitControlledUDecomposer {
     rxx_equivalent_gate: RXXEquivalent,
+    euler_basis: EulerBasis,
     #[pyo3(get)]
     scale: f64,
 }
 
-const DEFAULT_ATOL: f64 = 1e-12;
+pub const DEFAULT_ATOL: f64 = 1e-12;
```
ElePT: I wonder if we could make the tolerance maybe an optional argument for `TwoQubitControlledUDecomposer` with an internal default, I might be wrong but I don't see the point of exposing a default to just use it by default, we might as well use it internally and avoid the users the hassle.

crates/accelerate/src/unitary_synthesis.rs
```diff
@@ -474,7 +502,44 @@ fn run_2q_unitary_synthesis(
...
+                                        .standard_gate()
+                                        .name()
+                                        .to_string(),
+                                    Some(params.iter().map(|p| Param::Float(*p)).collect()),
+                                    inst_qubits,
+                                ),
+                            }
+                        });
+                let synth_error_from_target = synth_error(py, scoring_info, target);
+                synth_errors_sequence.push((sequence, synth_error_from_target));
```
ElePT: I should have come up with a nicer solution at the time, but back when we only had 2 decomposer types and the code for the `scoring_info` was slightly different, I was relatively ok with the code duplication. Now that two of the three arms look almost exactly the same, I wonder if we could abstract the code in a smarter way to avoid so much repetition. I don't think this will affect performance so it can also be left to a follow-up.

crates/accelerate/src/unitary_synthesis.rs
ElePT: See my comment above, but I think that if you need the `py` (it looks like you do) it isn't really necessary to have different functions for `new` and `new_inner`.

crates/accelerate/src/unitary_synthesis.rs
```diff
@@ -565,28 +632,15 @@ fn get_2q_decomposers_from_target(
...
-                        op.params[0] = Param::Float(PI2)
-                    }
-                }
-                "rzx" => {
-                    if let Param::ParameterExpression(_) = op.params[0] {
-                        op.params[0] = Param::Float(PI4)
-                    }
+            if PARAM_SET.contains(&std_gate.name()) {
+                if let Param::ParameterExpression(_) = op.params[0] {
+                    return true;
```
ElePT: I think we might be missing out a group of gates here. As far as I understand it, the `TwoQubitControlledUDecomposer` works not only for standard gates (the ones defined in `PARAM_SET`), but should also be at least a candidate for user-defined parametrized unitary gates that match the input conditions. In that case, we should probably not check `PARAM_SET.contains(&std_gate.name())`, and instead consider any 2q unitary parametrized gate for the "`available_2q_param_basis`".

crates/accelerate/src/unitary_synthesis.rs
ElePT: If you need it to instantiate the decomposer, I'm afraid there's no way to avoid it.

crates/accelerate/src/unitary_synthesis.rs
ElePT: See answer above.

qiskit/synthesis/two_qubit/two_qubit_decompose.py
ElePT: Mmm I don't really know what motivated the original choice of default gates, maybe it was done on purpose not to match the current fractional gates? @mtreinish do you have any insight?

crates/accelerate/src/unitary_synthesis.rs
ElePT: See comment above

ElePT (commented):
Hi @ShellyGarion, thanks for your PR!! I think that this is a very good start, I left some comments that are mostly non-crucial improvements, as well as one important question on the `check_parametrized_gate` function. I will make sure to review the new unit tests when you add them, and feel free to reach out if you have any question (or objection) :)

crates/accelerate/src/two_qubit_decompose.rs
ShellyGarion: The sign change here is since this PR extends `TwoQubitControlledUDecomposer` to any given `euler_basis`, where previously "ZYZ" was hard coded. In the case of "ZYZ" the global_phase of `k1l` and `k1r` was 0, so we didn't get an error before. The test `test_correct_unitary` was extended in the PR to include all possible values of `euler_basis`.

crates/accelerate/src/two_qubit_decompose.rs
ShellyGarion: Indeed, see my previous comment.

crates/accelerate/src/two_qubit_decompose.rs
ShellyGarion: the `new` function is `pyo3` method -- isn't the `new_inner` function faster? if not, I can change it back.

crates/accelerate/src/unitary_synthesis.rs
ShellyGarion: Note that not any parametrized gate can fit here, only certain parametrized gates which are equivalent (up to single qubit gates) to an `RXXGate`. The `TwoQubitControlledUDecomposer` works for such gates if the user was able to calibrate them somehow. Should the `UnitarySynthesis` transpiler pass fit any calibrated gates?
So I'm not sure what exactly should be checked here. See the function `TwoQubitControlledUDecomposer.to_rxx_gate`

ShellyGarion:
> I think this closes #13428. How about adding a test case of consecutive RZZ (RXX, and RYY) gates?

@t-imamichi - unfortunately this PR does not close #13428 yet, since the `ConsolidateBlocks` transpiler pass does not consolidate the consecutive RZZ gates into a single unitary. 
This should be fixed in https://github.com/Qiskit/qiskit/pull/13419.
I will therefore transfer your comment there.

crates/accelerate/src/two_qubit_decompose.rs
ShellyGarion: I managed to remove the need of `py` in [231af7f](https://github.com/Qiskit/qiskit/pull/13568/commits/231af7f18b4ce6b69d82c526d1c3d2488c899f91)
by replacing it with gil, as suggested by Matthew in https://github.com/Qiskit/qiskit/pull/13419.
I still think we need both functions `new` and `new_inner` since one of them is in pure rust and the other is pyo3.

crates/accelerate/src/unitary_synthesis.rs
ShellyGarion: I managed to remove the need of py in [231af7f](https://github.com/Qiskit/qiskit/pull/13568/commits/231af7f18b4ce6b69d82c526d1c3d2488c899f91)
by replacing it with gil, as suggested by Matthew in https://github.com/Qiskit/qiskit/pull/13419.

crates/accelerate/src/unitary_synthesis.rs
ShellyGarion: I managed to remove the need of py in [231af7f](https://github.com/Qiskit/qiskit/pull/13568/commits/231af7f18b4ce6b69d82c526d1c3d2488c899f91)
by replacing it with gil, as suggested by Matthew in https://github.com/Qiskit/qiskit/pull/13419.

crates/accelerate/src/unitary_synthesis.rs
ShellyGarion: I managed to remove the need of py in [231af7f](https://github.com/Qiskit/qiskit/pull/13568/commits/231af7f18b4ce6b69d82c526d1c3d2488c899f91)
by replacing it with gil, as suggested by Matthew in https://github.com/Qiskit/qiskit/pull/13419.

crates/accelerate/src/unitary_synthesis.rs
ShellyGarion: I managed to remove the need of py in [231af7f](https://github.com/Qiskit/qiskit/pull/13568/commits/231af7f18b4ce6b69d82c526d1c3d2488c899f91)
by replacing it with gil, as suggested by Matthew in https://github.com/Qiskit/qiskit/pull/13419.

crates/accelerate/src/unitary_synthesis.rs
ElePT: We currently support non-standard gates too, so the pass should work if there is a non-standard gate that meets the condition of rxx equivalence. `TwoQubitControlledUDecomposer::new` already checks if the gate is rxx equivalent:https://github.com/Qiskit/qiskit/blob/e9ccd3f374fd5424214361d47febacfa5919e1e3/crates/accelerate/src/two_qubit_decompose.rs#L2777-L2822

So I was thinking that we could abstract that logic and add a check to `UnitarySynthesis` similar to `is_controlled` or `is_supercontrolled`, but in this case it could be called `is_rxx_equivalent`(for example). The idea for the check is to filter out the non-rxx-equivalent gates.

crates/accelerate/src/two_qubit_decompose.rs
ShellyGarion: Done in [c2a9ad4](https://github.com/Qiskit/qiskit/pull/13568/commits/c2a9ad452984a4880badf746e28a97d66ffd7ea3)

crates/accelerate/src/two_qubit_decompose.rs
ElePT: Yes, let's keep them.

test/python/transpiler/test_unitary_synthesis.py
```diff
@@ -843,19 +844,39 @@ def test_iswap_no_cx_synthesis_succeeds(self):
         result_qc = dag_to_circuit(result_dag)
         self.assertTrue(np.allclose(Operator(result_qc.to_gate()).to_matrix(), cxmat))
 
-    def test_parameterized_basis_gate_in_target(self):
-        """Test synthesis with parameterized RXX gate."""
+    @data(True, False)
+    def test_parameterized_basis_gate_in_target(self, is_random):
+        """Test synthesis with parameterized RZZ gate."""
```
ElePT: Would it make sense to keep both the test for RXX and RZZ? Given that we used to have it...

test/python/transpiler/test_unitary_synthesis.py
```diff
@@ -843,19 +844,39 @@ def test_iswap_no_cx_synthesis_succeeds(self):
...
+        self.assertTrue(set(opcount).issubset({"rz", "rx", "rzz"}))
+        self.assertTrue(np.allclose(Operator(qc_transpiled), Operator(qc)))
+
+    @data(
+        ["rx", "ry", "rxx"],
+        ["rx", "rz", "rzz"],
+    )
+    def test_parameterized_backend(self, basis_gates):
+        """Test synthesis with parameterized backend."""
+        backend = GenericBackendV2(3, basis_gates=basis_gates)
```
ElePT: I don't think this will have a big effect on this particular test, but as a good practice, I recommend always setting the seed of the generic backend to make sure the errors and durations are always the same (this affects especially the routing tests)

```suggestion
        backend = GenericBackendV2(3, basis_gates=basis_gates, seed=0)
```

ElePT (commented):
Thanks for the tests! I took a look and they look good to me.

qiskit/synthesis/two_qubit/two_qubit_decompose.py
ShellyGarion: done in [14c7aac](https://github.com/Qiskit/qiskit/pull/13568/commits/14c7aacea7b3f0e5be8b9338ad21eecd558bf0c4)

test/python/transpiler/test_unitary_synthesis.py
ShellyGarion: done in [6a82649](https://github.com/Qiskit/qiskit/pull/13568/commits/6a82649709a9ee23f7eef241c055f0f483f00b79)

test/python/transpiler/test_unitary_synthesis.py
ShellyGarion: done in [6a82649](https://github.com/Qiskit/qiskit/pull/13568/commits/6a82649709a9ee23f7eef241c055f0f483f00b79)

crates/accelerate/src/two_qubit_decompose.rs
```diff
@@ -2760,19 +2757,9 @@ impl TwoQubitControlledUDecomposer {
...
-    ///  Args:
-    ///      rxx_equivalent_gate: Gate that is locally equivalent to an :class:`.RXXGate`:
-    ///      :math:`U \sim U_d(\alpha, 0, 0) \sim \text{Ctrl-U}` gate.
-    ///  Raises:
-    ///      QiskitError: If the gate is not locally equivalent to an :class:`.RXXGate`.
-    #[new]
-    #[pyo3(signature=(rxx_equivalent_gate))]
-    pub fn new(py: Python, rxx_equivalent_gate: RXXEquivalent) -> PyResult<Self> {
+    /// Initialize the KAK decomposition.
+    pub fn new_inner(rxx_equivalent_gate: RXXEquivalent, euler_basis: &str) -> PyResult<Self> {
```
mtreinish, line 2823: If you're using `new_inner` here is there a reason to just take the `EulerBasis` object directly?

crates/accelerate/src/two_qubit_decompose.rs
```diff
@@ -2516,7 +2517,7 @@ impl TwoQubitControlledUDecomposer {
                         .collect::<SmallVec<_>>();
                     Ok((Some(inv_gate.0), inv_gate_params, qubits))
                 }
-                RXXEquivalent::CustomPython(gate_cls) => {
+                RXXEquivalent::CustomPython(gate_cls) => Python::with_gil(|py: Python| {
```
mtreinish, line 2522: I'm fine with this changes since longer term we'll need to do this when running this in a parallel context. But I'm curious what prompted these changes?

qiskit/synthesis/two_qubit/two_qubit_decompose.py
mtreinish: I would avoid changing the defaults for any other classes it causes necessary churn. I think the choice was more just out of habit/standard gates to use. In this case using zxz as the default is fine since it's new, (although I'd be inclined to use `zsxx` as the default). But it doesn't really matter because either you don't care about the euler basis, or you do and you will specify it.

ShellyGarion: I think that with RZZ gates the one-qubit gates are RZ and RX, so it makes sense to choose the default ZXZ here: https://docs.quantum.ibm.com/guides/fractional-gates
For other 2-qubit gates (CX/CZ/ECR) the one-qubit gates are still ZX, SX and X so it makes sense to choose other defaults. So I agree that we don't need to change default values in other parts of the code.

crates/accelerate/src/two_qubit_decompose.rs
ShellyGarion: See the discussion here: https://github.com/Qiskit/qiskit/pull/13568#discussion_r1918690246
I wanted to have a purely-rust `new_inner` function, so I copied the solution from your PR :)

crates/accelerate/src/two_qubit_decompose.rs
ShellyGarion: it's similar to ‎`TwoQubitBasisDecomposer::new_inner` which also gets `euler_basis: &str`
https://github.com/Qiskit/qiskit/blob/75436a485bfe25e8732fbb9a6cc8305275ecbfba/crates/accelerate/src/two_qubit_decompose.rs#L1724

test/python/transpiler/test_unitary_synthesis.py
```diff
@@ -863,6 +866,78 @@ def test_parameterized_basis_gate_in_target(self, is_random, param_gate):
...
+        target.add_instruction(CustomXXGate(theta))
+
+        qc = QuantumCircuit(2)
+        qc.unitary(random_unitary(4, seed=1234), [0, 1])
+        qc_transpiled = UnitarySynthesis(target=target)(qc)
+        # TODO: fix this assertion
+        # opcount = qc_transpiled.count_ops()
+        # print(opcount)
+        # self.assertTrue(set(opcount).issubset({"rz", "rx", CustomXXGate.name}))
+        # self.assertTrue(np.allclose(Operator(qc_transpiled), Operator(qc)))
```
ElePT: This test has unveiled an issue with the integration of the `TwoQubitControlledUDecomposer`. The expectation for the unitary synthesis pass is that the decomposed circuit will respect the user-given basis gates (directly or through a target). However, the  `TwoQubitControlledUDecomposer` will always append a series of `h`, `s` and `sdg` to convert from the canonical form to the given basis. We could only run the decomposer if the target basis gates contain `h`, `s` and `sdg`, but this would leave out essentially all IBM backends. After discussing it with @Cryoris  we think it might make sense to express this conversion from canonical form in terms of ibm-supported basis gates, which doesn't fix the issue completely, but would make the pass more usable in the current scenarios we know. If we did that, we could follow this simple logic where we run the decomposer when the gates used match the given basis set. What do you think?

ElePT (commented):
Hi @ShellyGarion, I have pushed a commit that includes the work I had been doing on integrating non-standard gates. As you know, I included a couple of unit tests which have pointed to a new issue we hadn't realized until now. Even though the CI checks might not fully pass I wanted to expose this commit to make sure we could all look for solutions with the latest information.

test/python/transpiler/test_unitary_synthesis.py
mtreinish: As there are only 12 euler bases to pick from in the 1q decomposer I think the approach I would look at here would be to just make a small lookup table in for each of the euler basis values and then substitute in the appropriate gates based on the selected euler basis. For example in the two qubit decomposer define something like:

```rust
static S_GATE_LUT = [[Option<(StandardGate, [f64; 3])>; 3]; EULER_BASIS_SIZE] = [[Some((StandardGate::U3, [0.,0.,PI_2]), None, None], ...]
```
for all 3 gates. Then in the current code where it's adding the gates just do a lookup with `S_GATE_LUT[self.euler_basis as usize]` and add the gates to the output accordingly. That should minimize the overhead here but meet the expectations of the transpiler pass.

ShellyGarion: Thanks @ElePT for adding the code and tests, and @mtreinish for your suggestion!
So this update should be done in the file `two_qubit_decompose.rs` when we add the gates S, Sdg and H?
Is there a reason not to use the `unitary_to_gate_sequence_inner` to decompose the matrices of the S, Sdg and H gates? (so we wouldn't need to update the static array if we add another euler basis)

ShellyGarion: BTW, why does the new test `test_custom_parameterized_gate_in_target` fail, while the test `test_parameterized_basis_gate_in_target` goes well (without adding the extra S/Sdg/H gates)?

ElePT: The new test runs `UnitarySynthesis` in isolation while `test_parameterized_basis_gate_in_target ` runs the full transpiler pipeline, it looks like these gates are translated in later stages and that's why they don't show up in the final result. The test with the custom gate fails if you try to run `transpile` because the basis translator doesn't know how to handle the circuit. However, even if the pipeline compensates the extra gates, I think the pass in isolation should work following the assumptions (i.e. it should follow the specified basis).

ShellyGarion: see [5bb55b5](https://github.com/Qiskit/qiskit/pull/13568/commits/5bb55b581f0852e464399e9e916eb2f2a7d3430a)
I decomposed the S, Sdg and H gates using `unitary_to_gate_sequence_inner` (which is more general then having lookup tables)

test/python/transpiler/test_unitary_synthesis.py
```diff
@@ -843,19 +847,110 @@ def test_iswap_no_cx_synthesis_succeeds(self):
...
+        target.add_instruction(RXGate(phi))
+        target.add_instruction(CustomXXGate(theta))
+
+        qc = QuantumCircuit(2)
+        qc.unitary(random_unitary(4, seed=1234), [0, 1])
+        qc_transpiled = UnitarySynthesis(target=target)(qc)
+        # TODO: fix this assertion
+        opcount = qc_transpiled.count_ops()
+        self.assertTrue(set(opcount).issubset({"rz", "rx", "MyCustomXXGate"}))
+        # self.assertTrue(np.allclose(Operator(qc_transpiled), Operator(qc)))
```
ShellyGarion: Note that this test is still failing, since in the decomposed circuit I still see `MyCustomXXGate(theta)` instead of the relevant angles.

ShellyGarion: these is the place to fix:
https://github.com/Qiskit/qiskit/blob/f4da2dac064a0442bf1ed091679e16bc11af3cc8/crates/accelerate/src/unitary_synthesis.rs#L157-L164

ElePT: Fixed in [1bd71c5](https://github.com/Qiskit/qiskit/pull/13568/commits/1bd71c58ba9333f7cd72eaf2809df07986452b42)

releasenotes/notes/add-2q-fractional-gates-to-unitarysynthesis-pass-f66eee29903f5639.yaml
```diff
@@ -0,0 +1,10 @@
+---
+features_synthesis:
+  - |
+    Add a :class:`.TwoQubitControlledUDecomposer` that decomposes any two-qubit unitary
+    in terms of basis two-qubit fractional gates, such as :class:`.RZZGate`
+    (or two-gates gates which are locally equivalent to :class:`.RZZGate` up to single qubit gates).
```
ElePT, line 6: I think we need to show an example here of how the class is used from python.

releasenotes/notes/add-2q-fractional-gates-to-unitarysynthesis-pass-f66eee29903f5639.yaml
```diff
@@ -0,0 +1,10 @@
...
+features_transpiler:
+  - |
+    Add two-qubit fractional basis gates, such as :class:`.RZZGate`, to the
+    :class:`.UnitarySynthesis` transpiler pass.
```
ElePT: Same comment as above, we should add an example usage.

```suggestion
    Added support for two-qubit fractional basis gates, such as :class:`.RZZGate`, to the
    :class:`.UnitarySynthesis` transpiler pass. The decomposition is done using the :class:`.TwoQubitControlledUDecomposer`, and supports both standard and custom basis gates.
```

ElePT (commented):
I think that the PR is in a good state, the functionality is covered and tested. Some parts of the code could be cleaned up (like my `score_instruction` comment), but this could be done in a follow-up (I will have to refactor the code a bit to add support for `basis_gates` inputs anyway). The only open point would be the performance of the pass, and the use of the lookup table vs the euler decomposer. I believe that the lookup table will be faster, @ShellyGarion, and the likelihood of adding a new basis to the euler list I think is pretty low. Would you be willing to give it a try?

test/python/transpiler/test_unitary_synthesis.py
ShellyGarion: @ElePT - following your question below. I think that the correct place to add lookup tables for non-parametric 1-qubit gates (like S, Sdg and H) is in the function `unitary_to_gate_sequence_inner`, since the `two_qubit_decompose` code should handle 2-qubit gates and not 1-qubit gates. 
I'm not sure how much it will impact the performance here (since we need to decompose other 1-qubit and 2-qubit gates). We can consider it in a follow-up PR.

releasenotes/notes/add-2q-fractional-gates-to-unitarysynthesis-pass-f66eee29903f5639.yaml
ShellyGarion: done in [55556f8](https://github.com/Qiskit/qiskit/pull/13568/commits/55556f8f4762d960391311a76d2ecc4ee2d3222c)

releasenotes/notes/add-2q-fractional-gates-to-unitarysynthesis-pass-f66eee29903f5639.yaml
ShellyGarion: done in [55556f8](https://github.com/Qiskit/qiskit/pull/13568/commits/55556f8f4762d960391311a76d2ecc4ee2d3222c)

test/python/transpiler/test_unitary_synthesis.py
ShellyGarion: after talking with @mtreinish I don't think that the one-qubit synthesis is a bottleneck since it should be negligible compared to the two-qubit synthesis. We can still improve it in a follow-up PR if needed.

ElePT (approved):
Approving, as this will unblock follow-up work.

PR #13577, opened 2024-12-18
Bump PyO3 and rust-numpy to 0.23.x

mtreinish:
### Summary

This commit bumps the version of pyo3 and rust-numpy used by qiskit to the latest release 0.23. The largest change by volume of code is the deprecation of all the `*_bound()` methods. These are just warnings but they would be fatal to our CI so it needs to be updated. The larger functional change that required updating the code is the change in the traits around converting to Python objects. This actually found a bug in the target where we were not returning a proper instruction type for standard gates. This also opens up the opportunity to update our hashbrown version to 0.15.x, but we can't do that until the next rustworkx-core release.

### Details and comments


crates/accelerate/src/consolidate_blocks.rs
```diff
@@ -180,10 +180,11 @@ pub(crate) fn consolidate_blocks(
                     dag.remove_op_node(node);
                 }
             } else {
-                let unitary_gate =
-                    UNITARY_GATE
-                        .get_bound(py)
-                        .call1((array.to_object(py), py.None(), false))?;
+                let unitary_gate = UNITARY_GATE.get_bound(py).call1((
+                    <pyo3::Bound<'_, PyArray<Complex<f64>, ndarray::Dim<[usize; 2]>>> as Clone>::clone(&array).into_pyobject(py)?,
```
kevinhartman: 👀

crates/accelerate/src/sabre/heuristic.rs
```diff
@@ -159,9 +174,11 @@ impl DecayHeuristic {
 
     pub fn __repr__(&self, py: Python) -> PyResult<Py<PyAny>> {
         let fmt = "DecayHeuristic(increment={!r}, reset={!r})";
-        Ok(PyString::new_bound(py, fmt)
+        Ok(PyString::new(py, fmt)
             .call_method1("format", (self.increment, self.reset))?
-            .into_py(py))
+            .into_pyobject(py)?
+            .into_any()
+            .unbind())
```
kevinhartman: Is this potentially interesting for these cases?

https://pyo3.rs/main/doc/pyo3/conversion/trait.intopyobjectext#method.into_py_any

crates/accelerate/src/target_transpiler/mod.rs
```diff
@@ -112,6 +95,26 @@ pub(crate) struct NormalOperation {
...
+    type Error = PyErr;
+
+    fn into_pyobject(self, py: Python<'py>) -> Result<Self::Output, Self::Error> {
+        Ok(self.op_object.bind(py).clone())
+    }
+}
+
+impl<'py> IntoPyObject<'py> for &NormalOperation {
+    type Target = PyAny;
+    type Output = Bound<'py, Self::Target>;
```
kevinhartman: From the migration guide, should this be a `Borrowed` instead?

https://pyo3.rs/v0.23.3/migration.html#intopyobject-manual-implementation

crates/circuit/src/operations.rs
```diff
@@ -72,25 +73,35 @@ impl<'py> FromPyObject<'py> for Param {
...
-impl ToPyObject for Param {
-    fn to_object(&self, py: Python) -> PyObject {
-        match self {
-            Self::Float(val) => val.to_object(py),
-            Self::ParameterExpression(val) => val.clone_ref(py),
-            Self::Obj(val) => val.clone_ref(py),
-        }
-    }
-}
+//impl<'py> IntoPyObject<'py> for Param {
```
kevinhartman: 👀

kevinhartman (commented):
Thanks for doing this! I've done just one quick pass for now, and will unassign myself since I'm unlikely to have another look before I'm out on vacation.

crates/circuit/src/operations.rs
mtreinish: Oops this was leftover from me playing around before I derived the implementation

mtreinish: Removed in: https://github.com/Qiskit/qiskit/pull/13577/commits/c32f590d0a106a0d96601918244d22c383f196c4

crates/accelerate/src/sabre/heuristic.rs
mtreinish: Good idea, I updated this in: https://github.com/Qiskit/qiskit/pull/13577/commits/dd1ce36d1bae71e381e25f10d825e249ea2cda00

crates/accelerate/src/target_transpiler/mod.rs
mtreinish: Done in: https://github.com/Qiskit/qiskit/pull/13577/commits/17210ee76c8c4395b69d709184d5a286c069fe60

crates/accelerate/src/basis/basis_translator/compose_transforms.rs
```diff
@@ -106,7 +106,7 @@ pub(super) fn compose_transforms<'a>(
                 for (node, params) in nodes_to_replace {
                     let param_mapping: HashMap<ParameterUuid, Param> = equiv_params
                         .iter()
-                        .map(|x| ParameterUuid::from_parameter(x.to_object(py).bind(py)))
+                        .map(|x| ParameterUuid::from_parameter(&x.into_pyobject(py).unwrap()))
```
kevinhartman, line 103: For others, the `unwrap` is needed here because PyO3's `IntoPyObject::into_pyobject` now returns a `Result<T, E>` type. The error type is `PyErr` in this case (dictated by the `#[derive(IntoPyObjectRef)]` attribute on `Param`).

crates/accelerate/src/commutation_checker.rs
```diff
@@ -725,6 +721,31 @@ pub enum CommutationLibraryEntry {
...
+    type Target = PyAny;
+    type Output = Bound<'py, Self::Target>;
+    type Error = PyErr;
+
+    fn into_pyobject(self, py: Python<'py>) -> Result<Self::Output, Self::Error> {
+        let py_out = match self {
+            CommutationLibraryEntry::Commutes(b) => {
+                let temp = b.into_pyobject(py)?;
+                <pyo3::Bound<'_, PyBool> as Clone>::clone(&temp).into_any()
+            }
```
kevinhartman: I think this would be the expected way to do this. The impl of `IntoPyObject` for `bool` gives us a `Borrowed<PyBool>`, and then by calling trait method `BoundObject::into_bound` on that, we get an owned `Bound<PyBool>`.

```suggestion
            CommutationLibraryEntry::Commutes(b) => b.into_pyobject(py)?.into_bound().into_any(),
```

crates/accelerate/src/consolidate_blocks.rs
kevinhartman: Maybe this?

```suggestion
                    array.as_ref().into_pyobject(py)?,
```

crates/accelerate/src/sparse_observable.rs
```diff
@@ -1315,14 +1318,21 @@ impl SparseObservable {
...
     fn __getitem__(&self, py: Python, index: PySequenceIndex) -> PyResult<Py<PyAny>> {
         let indices = match index.with_len(self.num_terms())? {
-            SequenceIndex::Int(index) => return Ok(self.term(index).to_term().into_py(py)),
+            SequenceIndex::Int(index) => {
+                return Ok(self
+                    .term(index)
+                    .to_term()
+                    .into_pyobject(py)?
+                    .into_any()
+                    .unbind())
```
kevinhartman: Looks like you can use `into_py_any(py)` here if you `use pyo3::IntoPyObjectExt;`.

crates/accelerate/src/target_transpiler/mod.rs
```diff
@@ -57,30 +58,12 @@ type GateMapState = Vec<(String, Vec<(Option<Qargs>, Option<InstructionPropertie
 
 /// Represents a Qiskit `Gate` object or a Variadic instruction.
 /// Keeps a reference to its Python instance for caching purposes.
-#[derive(FromPyObject, Debug, Clone)]
+#[derive(FromPyObject, Debug, Clone, IntoPyObjectRef)]
```
kevinhartman, line 61: Should we also derive `IntoPyObject`?

crates/circuit/src/interner.rs
```diff
@@ -206,7 +206,8 @@ mod test {
     fn default_key_exists() {
         let mut interner = Interner::<[u32]>::new();
         assert_eq!(interner.get_default(), interner.get_default());
-        assert_eq!(interner.get(interner.get_default()), &[]);
+        let res: &[u32] = &[];
+        assert_eq!(interner.get(interner.get_default()), res);
```
kevinhartman, line 210: What's going on here?

crates/qasm2/src/bytecode.rs
```diff
@@ -211,89 +211,108 @@ pub enum InternalBytecode {
     },
 }
 
-impl IntoPy<Bytecode> for InternalBytecode {
+impl<'py> IntoPyObject<'py> for InternalBytecode {
```
kevinhartman, line 214: Is this not derivable because the enum variants have named fields?

kevinhartman (commented):
Just a few suggestions, otherwise this looks ready to me. Thanks!

crates/accelerate/src/sparse_observable.rs
mtreinish: I knew I would miss at least one of these when I updated this

crates/accelerate/src/target_transpiler/mod.rs
mtreinish: Probably, I probably didn't do it because nothing used it, I'll add this. There might havee been a reason I didn't add it, but I'll find out when I try this (and will report back if it can't be added).

crates/circuit/src/interner.rs
mtreinish: TBH I can't remember. It's either the compiler started complaining here and suggested I do this to make it work, or I was debugging something and did this to do `println!("{:?}", res)` before the assertion and forgot to change it back but removed the print. I'll try reverting this and see what happens.

crates/qasm2/src/bytecode.rs
mtreinish: Yeah when I did the derive approach it didn't work, I don't remember the exact details because I worked on this > 1 month ago. But I had to go this route.

crates/circuit/src/interner.rs
mtreinish: After trying it locally I remember now. The compiler was complaining it can't infer the type of `&[]` anymore so I used the res var to define it as a `&[u32]`.

crates/accelerate/src/target_transpiler/mod.rs
mtreinish: Ah, so trying this locally adding the definition confuses the return of `self._gate_name_map[x].into_py_object()` below because it tries to do it with an owned version instead of the ref.

crates/accelerate/src/sparse_observable.rs
mtreinish: Done in: https://github.com/Qiskit/qiskit/pull/13577/commits/389f6e7a4fcd92e11e00b57a57d57471fee499b1

crates/circuit/src/interner.rs
jakelishman: This seems weird that the PyO3 update would cause the inference here to suddenly fail.  The types of `Interner` and `get_default()` shouldn't have changed.

crates/qasm2/src/bytecode.rs
jakelishman: Using the semantics of the derive macros would substantially change how the Rust/Python boundary layer is handled here.  Even if PyO3 didn't error, the types would be entirely different to what Python is expecting.

crates/accelerate/src/target_transpiler/mod.rs
jakelishman: I imagine that's expected / intended - I'd assume the spelling would need to become `(&self._gate_name_map[x]).into_py_object()`.  I think in general `t.into_py()` goes to `t.into_py_object()` and `t.to_object()` goes to `(&t).into_py_object()`?

crates/accelerate/src/basis/basis_translator/compose_transforms.rs
jakelishman: It's not clear to me why the `Param` needs converting into a Python object at all here.  If it isn't a `Param::Parameter` variant, the whole thing will fail, so it seems like we could have sorted out that typing a while prior.

kevinhartman (dismissed):

kevinhartman (dismissed):

crates/accelerate/src/error_map.rs
```diff
@@ -101,14 +102,19 @@ impl ErrorMap {
...
+        py: Python,
+        key: [PhysicalQubit; 2],
+        default: Option<PyObject>,
+    ) -> PyResult<PyObject> {
+        Ok(match self.error_map.get(&key).copied() {
+            Some(val) => val.into_py_any(py)?,
             None => match default {
                 Some(val) => val,
                 None => py.None(),
             },
```
jakelishman, line 116: Seems weird that we do `default: Option<PyObject>` when we just recreate the `None` singleton anyway, but that was pre-existing.

crates/accelerate/src/euler_one_qubit_decomposer.rs
```diff
@@ -109,11 +110,13 @@ impl OneQubitGateSequence {
 
     fn __getitem__(&self, py: Python, idx: PySequenceIndex) -> PyResult<PyObject> {
         match idx.with_len(self.gates.len())? {
-            SequenceIndex::Int(idx) => Ok(self.gates[idx].to_object(py)),
-            indices => Ok(PyList::new_bound(
+            SequenceIndex::Int(idx) => Ok(self.gates[idx].clone().into_py_any(py)?),
```
jakelishman: The `clone` and move seems weird here.  Does it work with `(&self.gates[idx]).into_py_any(py)` to do it implicitly (and potentially make it easier to elide the clone/move)?

crates/accelerate/src/euler_one_qubit_decomposer.rs
```diff
@@ -109,11 +110,13 @@ impl OneQubitGateSequence {
...
+            indices => Ok(PyList::new(
                 py,
-                indices.iter().map(|pos| self.gates[pos].to_object(py)),
-            )
+                indices
+                    .iter()
+                    .map(|pos| self.gates[pos].clone().into_pyobject(py).unwrap()),
```
jakelishman: Same here with the `clone`.

crates/accelerate/src/euler_one_qubit_decomposer.rs
```diff
@@ -712,12 +715,8 @@ impl EulerBasis {
...
 #[pymethods]
 impl EulerBasis {
-    fn __reduce__(&self, py: Python) -> Py<PyAny> {
-        (
-            py.get_type_bound::<Self>(),
-            (PyString::new_bound(py, self.as_str()),),
-        )
-            .into_py(py)
+    fn __reduce__(&self, py: Python) -> PyResult<Py<PyAny>> {
+        (py.get_type::<Self>(), (PyString::new(py, self.as_str()),)).into_py_any(py)
```
jakelishman, line 720: This `__reduce__` is actually a fancy way of spelling `__getnewargs__`, but I think I might have been the culprit here.

crates/accelerate/src/nlayout.rs
```diff
@@ -123,12 +115,11 @@ impl NLayout {
     }
 
     fn __reduce__(&self, py: Python) -> PyResult<Py<PyAny>> {
-        Ok((
-            py.get_type_bound::<Self>()
-                .getattr("from_virtual_to_physical")?,
-            (self.virt_to_phys.to_object(py),),
+        (
+            py.get_type::<Self>().getattr("from_virtual_to_physical")?,
+            (self.virt_to_phys.clone().into_pyobject(py).unwrap(),),
```
jakelishman:
```suggestion
            (self.virt_to_phys.clone().into_pyobject(py)?,),
```
since we're already in a `PyResult`-returning function.

crates/accelerate/src/nlayout.rs
```diff
@@ -143,8 +134,8 @@ impl NLayout {
     ///     where the virtual qubit is the index in the qubit index in the circuit.
     ///
     #[pyo3(text_signature = "(self, /)")]
-    fn layout_mapping(&self, py: Python<'_>) -> Py<PyList> {
-        PyList::new_bound(py, self.iter_virtual()).into()
+    fn layout_mapping(&self, py: Python<'_>) -> PyResult<Py<PyList>> {
+        PyList::new(py, self.iter_virtual()).map(|x| x.unbind())
```
jakelishman, line 138: Maybe better to change the return type to `PyResult<Bound<'py, PyList>>` rather than add an explicit unbind?

crates/accelerate/src/sparse_observable.rs
```diff
@@ -1385,15 +1391,20 @@ impl SparseObservable {
                 &slf_.borrow(),
                 Complex64::new(2.0, 0.0),
             )
-            .into_py(py));
+            .into_pyobject(py)?
+            .into_any()
+            .unbind());
```
jakelishman, line 1398: Spiritually, I feel like these magic methods should have just started reuning `Bound<...>` instead of `Py<...>`, but we don't need to change it again in this PR.

crates/accelerate/src/sparse_observable.rs
```diff
@@ -1547,7 +1579,7 @@ impl SparseObservable {
     /// Examples:
     ///
     ///     .. code-block:: python
-    ///         
+    ///
```
jakelishman, line 1584: Huh, weird to me that rustfmt doesn't tidy these up in comments.

crates/accelerate/src/two_qubit_decompose.rs
```diff
@@ -1295,11 +1297,16 @@ impl TwoQubitGateSequence {
 
     fn __getitem__(&self, py: Python, idx: PySequenceIndex) -> PyResult<PyObject> {
         match idx.with_len(self.gates.len())? {
-            SequenceIndex::Int(idx) => Ok(self.gates[idx].to_object(py)),
-            indices => Ok(PyList::new_bound(
+            SequenceIndex::Int(idx) => {
+                let item = &self.gates[idx];
+                (item.0, PyList::new(py, &item.1)?, PyList::new(py, &item.2)?).into_py_any(py)
+            }
+            indices => Ok(PyList::new(
```
jakelishman, line 1305: Is this working around the `[u8]`-is-now-`bytes` conversion trait weirdness?

crates/circuit/src/dag_circuit.rs
```diff
@@ -910,9 +905,10 @@ impl DAGCircuit {
         }
 
         let params_tuple = if let Some(operands) = params {
-            let add_calibration = PyModule::from_code_bound(
+            let add_calibration = PyModule::from_code(
                 py,
-                r#"
+                std::ffi::CString::new(
+                    r#"
```
jakelishman, line 911: If only we were MSRV 1.77 and could just write `cr#"`...

crates/circuit/src/dag_node.rs
```diff
@@ -249,21 +252,17 @@ impl DAGOpNode {
...
-                &slf.instruction.clbits,
-            ),
-            state,
-        )
-            .into_py(py))
+        let temp = (
+            slf.instruction.get_operation(py)?,
+            &slf.instruction.qubits,
+            &slf.instruction.clbits,
+        );
```
jakelishman, line 264: I wouldn't mind calling `temp` something like `new_args` - the `__reduce__` tuple is complex enough as it is.

crates/circuit/src/lib.rs
```diff
@@ -33,7 +33,7 @@ use pyo3::prelude::*;
 use pyo3::types::{PySequence, PyTuple};
 
 pub type BitType = u32;
-#[derive(Copy, Clone, Debug, Hash, Ord, PartialOrd, Eq, PartialEq, FromPyObject)]
+#[derive(Copy, Clone, Debug, Hash, Ord, PartialOrd, Eq, PartialEq, FromPyObject, IntoPyObject)]
 pub struct Qubit(pub BitType);
```
jakelishman: This looks like a behaviour change - `Qubit` and `Clbit` didn't previously have `IntoPyObject` defined, I think?

crates/circuit/src/operations.rs
```diff
@@ -688,9 +663,12 @@ impl StandardGate {
...
-    pub fn _to_matrix(&self, py: Python, params: Vec<Param>) -> Option<PyObject> {
-        self.matrix(&params)
-            .map(|x| x.into_pyarray_bound(py).into())
+    pub fn _to_matrix<'py>(
+        &'py self,
+        py: Python<'py>,
+        params: Vec<Param>,
+    ) -> Option<Bound<'py, PyArray2<Complex64>>> {
+        self.matrix(&params).map(|x| x.into_pyarray(py))
     }
```
jakelishman: I don't think the lifetime bound on `self` needs to live as long as `'py`?  Not that I think it matters, since this method is marked "just for testing".

crates/circuit/src/packed_instruction.rs
```diff
@@ -111,7 +111,7 @@ unsafe impl ::bytemuck::NoUninit for PackedOperationType {}
 /// * The pointed-to data must match the type of the discriminant used to store it.
 /// * `PackedOperation` must take care to forward implementations of `Clone` and `Drop` to the
 ///   contained pointer.
-#[derive(Debug)]
+#[derive(Debug, IntoPyObject, IntoPyObjectRef)]
 #[repr(transparent)]
 pub struct PackedOperation(usize);
```
jakelishman: This really feels wrong - `PackedOperation` definitely should not ever be exposed to Python.  Aside from the data being completely meaningless to Python, implementing `IntoPyObjectRef` for it allows the holder to `Clone` the inner `usize`, which might represent ownership of a `Box` pointer.

crates/circuit/src/parameter_table.rs
```diff
@@ -10,7 +10,7 @@
 // copyright notice, and modified files need to carry a notice indicating
 // that they have been altered from the originals.
 
-use std::cell::OnceCell;
+use std::sync::OnceLock;
```
jakelishman, line 13: Is this because of free-threading support in PyO3, or a separate behavioral change?

crates/qasm2/src/expr.rs
```diff
@@ -146,57 +146,76 @@ pub enum Expr {
...
 }
 
-impl IntoPy<PyObject> for Expr {
-    fn into_py(self, py: Python<'_>) -> PyObject {
-        match self {
-            Expr::Constant(value) => bytecode::ExprConstant { value }.into_py(py),
-            Expr::Parameter(index) => bytecode::ExprArgument { index }.into_py(py),
+impl<'py> IntoPyObject<'py> for Expr {
+    type Target = PyAny; // the Python type
+    type Output = Bound<'py, Self::Target>; // in most cases this will be `Bound`
```
jakelishman, line 151: lol, this isn't the first time I've seen the copy/paste comments from the guide

crates/qasm2/src/lex.rs
```diff
@@ -337,7 +337,7 @@ pub struct TokenStream {
     /// backing file or other named resource.
     pub filename: std::ffi::OsString,
     strict: bool,
-    source: Box<dyn std::io::BufRead + Send>,
+    source: Box<dyn std::io::BufRead + Send + Sync>,
```
jakelishman, line 340: Is the `Sync` bound now required because of the free-threading support?

crates/qasm3/src/circuit.rs
```diff
@@ -86,30 +83,38 @@ pub struct PyGate {
...
+    >(
+        py: Python<'py>,
         constructor: T,
         name: S,
         num_params: usize,
         num_qubits: usize,
-    ) -> Self {
+    ) -> Self
+    where
+        <T as pyo3::IntoPyObject<'py>>::Error: std::fmt::Debug,
```
jakelishman, line 98: This bound feels pretty questionable - should we just make the constructor fallible and have callers bubble it up?

crates/qasm3/src/circuit.rs
```diff
@@ -86,30 +83,38 @@ pub struct PyGate {
...
-    pub fn construct<A>(&self, py: Python, args: A) -> PyResult<Py<PyAny>>
+    pub fn construct<'py, A>(&'py self, py: Python<'py>, args: A) -> PyResult<Py<PyAny>>
     where
-        A: IntoPy<Py<PyTuple>>,
+        A: pyo3::IntoPyObject<'py, Target = PyTuple, Output = Bound<'py, PyTuple>>,
+        <A as pyo3::IntoPyObject<'py>>::Error: std::fmt::Debug,
     {
-        let args = args.into_py(py);
-        let received_num_params = args.bind(py).len();
+        let args = args.into_pyobject(py).unwrap();
```
jakelishman: This function already returns `PyResult`, so no need for the crazy trait bounds and `unwrap`, I think?

crates/qasm3/src/circuit.rs
```diff
@@ -264,24 +270,36 @@ impl PyCircuitModule {
...
-        self.circuit_instruction
-            .call1(py, (operation, qubits.into_py(py), clbits.into_py(py)))
+        self.circuit_instruction.call1(
+            py,
+            (
+                operation,
+                qubits.into_pyobject(py)?,
+                clbits.into_pyobject(py)?,
+            ),
+        )
```
jakelishman: I think this could be `IntoPyObjectExt::into_pyobject_or_pyerr`, which would automatically use the trait bounds from the `IntoPyObject::Error` associated type to avoid needing to spell them out again.

jakelishman (commented):
Thanks for doing this.  I think there's a fair number of function returns that spiritually ought to swap from `Py<PyAny>` (or the `PyObject` spelling) into `Bound<PyAny>` - I think making that change was one of the driving reasons to make the trait change in the first place.  We have a lot of explicit `unbind`s in places where we're returning the object itself to Python space, not hiding it in a Rust object that needs to be GIL-independent (could be `into_bound_py_any` in many cases?).  No need to go hunting for them, though, for this PR.

(Side note: it's weird to me that `into_bound_py_any` was chosen, rather than having `into_py_any` getting called `into_unbound_py_any`, given the shift of focus away from `Py` in those conversions.)

I feel like there's at least a couple of places I've written that does the same thing as `IntoPyObjectExt::into_pyobject_or_pyerr`, but I don't really remember them off the top of my head.

It's most likely lack of experience, but I've got no intuition for what the `IntoPyObject` derive macro will do to enums with compound variants.  Personally, I'm not sure I like the derive macros for record structs much at all - turning stuff into `dict` feels icky - but that's just a knee-jerk reaction right now and I'll see what I think with more experience.

I didn't manage to 100% convince myself that all the `Target` changes produce visibly equivalent code, so I'm pretty much banking on the tests there.

crates/accelerate/src/two_qubit_decompose.rs
mtreinish: Yes this is how I had to workaround that. See: https://github.com/Qiskit/qiskit/pull/13577/commits/c59dc9c071d002fcc5288b0ba3c55a9979a19b3e where I made this change

crates/qasm2/src/lex.rs
mtreinish: Yeah, it was a requirement that the members of a pyclass were all Sync

crates/circuit/src/lib.rs
mtreinish: Removed in: https://github.com/Qiskit/qiskit/pull/13577/commits/132de0ad23929c2f8ff3c1296872eec5a1b075be

crates/circuit/src/packed_instruction.rs
mtreinish: Removed in: https://github.com/Qiskit/qiskit/pull/13577/commits/132de0ad23929c2f8ff3c1296872eec5a1b075be

crates/circuit/src/parameter_table.rs
mtreinish: Yeah, because we need the members of a pyclass need to support be used in a threading context this was needed to make `CircuitData` as a pyclass work with pyo3 0.23

crates/qasm3/src/circuit.rs
mtreinish: Done in: https://github.com/Qiskit/qiskit/pull/13577/commits/0577f48a7814455551ec2cda3540a9d1a7ad9847

crates/qasm3/src/circuit.rs
mtreinish: Done in: https://github.com/Qiskit/qiskit/pull/13577/commits/0577f48a7814455551ec2cda3540a9d1a7ad9847

crates/qasm3/src/circuit.rs
mtreinish: Removed in: https://github.com/Qiskit/qiskit/pull/13577/commits/0577f48a7814455551ec2cda3540a9d1a7ad9847

crates/accelerate/src/nlayout.rs
mtreinish: Done in: https://github.com/Qiskit/qiskit/pull/13577/commits/0577f48a7814455551ec2cda3540a9d1a7ad9847

crates/circuit/src/operations.rs
mtreinish: Fixed in: https://github.com/Qiskit/qiskit/pull/13577/commits/a2813401119cd80360cfb6891ea30c41525d60f8

jakelishman:
I'm sorting out the merge conflicts and looking into fixing up a couple of the comments I left in the last review, then I'll merge.

crates/accelerate/src/euler_one_qubit_decomposer.rs
jakelishman: I solved this by implementing `IntoPyObject for &T` for all our local `T: PyClass + Clone` (since I can't do the blanket impl before of coherence rules).  I might ask over on PyO3 if it's something that should have a blanket impl, be part of the `pyclass` generation, or something else.

jakelishman (approved):
I'm happy with this now, though somebody else ought to sign off on my commit cd99d9fc88e28125374a2ef90ea9ddb87b915ab5.

mtreinish:
[That](https://github.com/Qiskit/qiskit/pull/13577/commits/cd99d9fc88e28125374a2ef90ea9ddb87b915ab5) looks good to me, so I went ahead and added it to merge queue.

PR #13579, opened 2024-12-18
Import CircuitInstruction to fix Pylance complaint

kevinsung:
### Summary

Using `CircuitInstruction` in a type annotation causes Pylance to complain
```
Variable not allowed in type expression
```
This fixes that.


Cryoris:
If we change this, it would be good to have a guideline and a check by CI/linter. Otherwise we cannot ensure we're consistent here and this change seems a bit arbitrary. 

I've used Pylance as well, aren't there also other things it complains about? E.g. using `(TypeA, TypeB)` instead of `tuple[TypeA, TypeB]` or something?

qiskit/circuit/quantumcircuitdata.py
```diff
@@ -16,15 +16,13 @@
 from collections.abc import MutableSequence
 
 import qiskit._accelerate.circuit
+from qiskit._accelerate.circuit import CircuitInstruction  # pylint: disable=unused-import
```
jakelishman: Technically I think the right way to do this is to add an `__all__` variable, and include `"CircuitInstruction"` in it (along with the other things that should be there, which I think is just `"QuantumCircuitData"`) - that marks it as "explicitly re-exported".

kevinsung: I made a mistake, it was actually the preceding import that was unused, and I simply removed it.

kevinsung:
>If we change this, it would be good to have a guideline and a check by CI/linter. Otherwise we cannot ensure we're consistent here and this change seems a bit arbitrary.

Does the change still seem arbitrary after the latest commit? I think it brings it more in line with the rest of Qiskit, since we normally import stuff rather than setting a variable like it was before.

>I've used Pylance as well, aren't there also other things it complains about? E.g. using (TypeA, TypeB) instead of tuple[TypeA, TypeB] or something?

I don't know, this is the only one that I noticed.

PR #13580, opened 2024-12-18
Fix ``SparsePauliOp`` initialization with dense ``Y`` labels

### LINKED ISSUE #13522: `SparsePauliOp` incorrectly sets non-zero imaginary component
### Environment

- **Qiskit version**: 1.3.0
- **Python version**: 3.11.7
- **Operating system**: Fedora 40

### What is happening?

When a `SparsePauliOp` object is instantiated with a long string of Y's (greater than 100), the imaginary component of the coefficients is non-zero (but trivially small, ~1e-15). This results in errors when submitting jobs using `qiskit-ibm-runtime` which throws an error when using the `SparsePauliOp` as the expectation value to estimate.

```python
# produces non-zero imaginary component
yyy = SparsePauliOp('Y'*100)
print(yyy.coeffs)
# correctly contains no imaginary component
yy = SparsePauliOp('Y'*99)
print(yy.coeffs)
```

### What should happen?

A high-weight `SparsePauliOp`, when no coefficients are specified, should possess coeffcients = `[1.0, 0.0]`

### Any suggestions?

_No response_

trigpolynom:
### Summary
Fixes #13522 

"When a SparsePauliOp object is instantiated with a long string of Y's (greater than 100), the imaginary component of the coefficients is non-zero (but trivially small, ~1e-15). This results in errors when submitting jobs using qiskit-ibm-runtime which throws an error when using the SparsePauliOp as the expectation value to estimate." 

The fix to the above problem involved using modular arithmetic to calculate coeffs for the SparsePauliOp when coeffs is None. This avoids the floating point error that occurs for instantiating heavy-weight SparsePauliOps.

### Details and comments
Added a test method `test_sparse_pauli_op_init_long_ys` to ensure that the error causing PauliList ('Y"*100) results with the same coeffs as PauliList ("Y"*99)


ShellyGarion:
Hi @trigpolynom - thanks for fixing this bug!
After you complete the tests and fix the lint error, could you please add release notes?

qiskit/quantum_info/operators/symplectic/sparse_pauli_op.py
```diff
@@ -167,7 +167,16 @@ def __init__(
             # move the phase of `pauli_list` to `self._coeffs`
             phase = pauli_list._phase
             count_y = pauli_list._count_y()
-            self._coeffs = np.asarray((-1j) ** (phase - count_y) * coeffs, dtype=coeffs.dtype)
+
+            # Compute exponentiation via integer arithmetic and lookup table to avoid floating point errors
+            exp_val = (phase - count_y) % 4
+            p = (3 * exp_val) % 4  # (−i)^(phase - count_y) = i^(3*(phase - count_y))
+            lookup = np.array([1 + 0j, 1j, -1 + 0j, -1j], dtype=coeffs.dtype)
```
Cryoris: That's a nice solution! Couldn't we avoid using `p` here and just adjust the lookup accordingly?
```suggestion
            lookup = np.array([1, -1j, -1j, 1j], dtype=coeffs.dtype)  # (-1j) ** [0, 1, 2, 3]
```

qiskit/quantum_info/operators/symplectic/sparse_pauli_op.py
```diff
@@ -167,7 +167,16 @@ def __init__(
             # move the phase of `pauli_list` to `self._coeffs`
             phase = pauli_list._phase
             count_y = pauli_list._count_y()
-            self._coeffs = np.asarray((-1j) ** (phase - count_y) * coeffs, dtype=coeffs.dtype)
+
+            # Compute exponentiation via integer arithmetic and lookup table to avoid floating point errors
+            exp_val = (phase - count_y) % 4
```
Cryoris: Also a tiny nitpick -- `exp_val` to me sounds like an expectation value, as we're in the realm of quantum computing. Would just `exponent` be a more precise name? 🙂

trigpolynom:
Hi @ShellyGarion and @Cryoris thanks for taking a look at this! I'll make the updates and requested changes tonight EST.

qiskit/quantum_info/operators/symplectic/sparse_pauli_op.py
trigpolynom: yes makes complete sense. See update.

qiskit/quantum_info/operators/symplectic/sparse_pauli_op.py
trigpolynom: Oh I like this approach much better thank you! See updated code.

test/python/quantum_info/operators/symplectic/test_sparse_pauli_op.py
```diff
@@ -140,6 +140,15 @@ def test_sparse_pauli_op_init(self):
...
             self.assertEqual(spp_op, ref_op)
 
+    def test_sparse_pauli_op_init_long_ys(self):
+        """Test heavy-weight SparsePauliOp initialization."""
+        with self.subTest(msg="make SparsePauliOp from string of Y's"):
+            yyy = SparsePauliOp("Y" * 100)
+            print(yyy.coeffs)
+            yy = SparsePauliOp("Y" * 99)
+            print(yy.coeffs)
+            self.assertEqual(yyy.coeffs, yy.coeffs)
```
Cryoris: We don't need a `subTest` since there's only a single assertion statement in the function, also I think it would be better to use a fixed comparison value -- theoretically both 99 and 100 Y terms could be equal but wrong 🙂 We could e.g. just do
```suggestion
        y = SparsePauliOp("Y" * 1000) 
        self.assertEqual(1, y.coeffs[0])
```

releasenotes/notes/sparse-pauli-op-heavy-weight-fix-aa822428643d642a.yaml
```diff
@@ -0,0 +1,7 @@
+fixes:
+  - |
+    Fixed a bug where a floating point arithmetic error caused a heavy-weight 
+    initialization of :class:`.SparsePauliOp` (pauli list of size > 99) without any 
+    specified coeffs, to have non-zero imaginary components. The fix involved using
+    modular arithmetic to set the correct coefficients based on the phase and count_y of
+    the :class:`.PauliList`.
```
Cryoris: Thanks for adding a reno! Note that we don't typically explain how the bug was fixed if it is a technical detail. However we do include the issue that has been fixed, so users can have a look at more detailed discussions, if there are any.
```suggestion
    Fixed a bug where a initializing :class:`.SparsePauliOp` with a large
    number of Pauli-``Y`` terms (typically :math:`\geq 100`) and no explicit
    ``coeffs`` would result in a coefficient close to 1 but with a floating point
    error. The coefficient is now correctly 1 per default.
    Fixed `#13522 <https://github.com/Qiskit/qiskit/issues/13522>`__.
```

qiskit/quantum_info/operators/symplectic/sparse_pauli_op.py
ShellyGarion: note that there is still a lint error (see [here](https://dev.azure.com/qiskit-ci/qiskit/_build/results?buildId=62994&view=logs&j=ee6a574e-a884-556d-e11f-f4e0408370dd&t=3f2257ea-a3bb-5388-9567-23f2497491da))
```
************* Module qiskit.quantum_info.operators.symplectic.sparse_pauli_op
qiskit/quantum_info/operators/symplectic/sparse_pauli_op.py:171:0: C0301: Line too long (107/105) (line-too-long)
```

test/python/quantum_info/operators/symplectic/test_sparse_pauli_op.py
```diff
@@ -140,6 +140,14 @@ def test_sparse_pauli_op_init(self):
             coeffs[:] = 0
             self.assertEqual(spp_op, ref_op)
 
+    def test_sparse_pauli_op_init_long_ys(self):
+        """Test heavy-weight SparsePauliOp initialization."""
+        yyy = SparsePauliOp("Y" * 100)
+        print(yyy.coeffs)
+        yy = SparsePauliOp("Y" * 99)
+        print(yy.coeffs)
```
ShellyGarion: you should remove the "print" statements from the test

qiskit/quantum_info/operators/symplectic/sparse_pauli_op.py
```diff
@@ -167,7 +167,15 @@ def __init__(
             # move the phase of `pauli_list` to `self._coeffs`
             phase = pauli_list._phase
             count_y = pauli_list._count_y()
-            self._coeffs = np.asarray((-1j) ** (phase - count_y) * coeffs, dtype=coeffs.dtype)
+
+            # Compute exponentiation via integer arithmetic and lookup table to avoid floating point errors
```
ShellyGarion: this line is too long and fails formatting checks. Please split it

qiskit/quantum_info/operators/symplectic/sparse_pauli_op.py
trigpolynom: addressed

test/python/quantum_info/operators/symplectic/test_sparse_pauli_op.py
trigpolynom: addressed

releasenotes/notes/sparse-pauli-op-heavy-weight-fix-aa822428643d642a.yaml
trigpolynom: addressed

test/python/quantum_info/operators/symplectic/test_sparse_pauli_op.py
trigpolynom: addressed

qiskit/quantum_info/operators/symplectic/sparse_pauli_op.py
trigpolynom: addressed

test/python/quantum_info/operators/symplectic/test_sparse_pauli_op.py
ShellyGarion: not exactly, you still check that both values are equal, but not that they equal to 1 as suggested above

trigpolynom: you're right im sorry i missed this completely!, I just processed the subTest isssue.

ShellyGarion:
there are some formatting issues. other than that, I think this PR is ready to go!

ShellyGarion (approved):
LGTM.  Thanks for your contribution to Qiskit @trigpolynom !
I'll wait with merging it to see if @Cryoris has any further comments.

Cryoris (approved):

ShellyGarion:
Congrats @trigpolynom for making the first PR being merged in 2025!

PR #13582, opened 2024-12-19
Small refactoring of unitary synthesis tests

ElePT:
### Summary
This is a small refactoring PR that proposes splitting the `UnitarySynthesis` test class into two separate classes, one for tests that follow the `basis_gates` path, which is still in Python, and one for the `target` path, which is now in Rust. With the current interleaving of target and basis gates tests it's easy to oversee what is covered where, and I believe that as a result we have an improvable coverage of both the Python and Rust `UnitarySynthesis` code. This PR doesn't fix that but sets the stage for reviewing and adding new tests in an organized way as we work on PRs like [ #13568](https://github.com/Qiskit/qiskit/pull/13568).

### Details and comments
As I was moving the tests I took the time to squash a few of them into one, as they often just varied in one parameter that could be handled by ddt.

### DIFF
--- a/test/python/transpiler/test_unitary_synthesis.py
+++ b/test/python/transpiler/test_unitary_synthesis.py
@@ -110,8 +110,8 @@ class FakeBackend5QV2(GenericBackendV2):
 
 
 @ddt
-class TestUnitarySynthesis(QiskitTestCase):
-    """Test UnitarySynthesis pass."""
+class TestUnitarySynthesisBasisGates(QiskitTestCase):
+    """Test UnitarySynthesis pass with basis gates."""
 
     def test_empty_basis_gates(self):
         """Verify when basis_gates is None, we do not synthesize unitaries."""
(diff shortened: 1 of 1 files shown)


test/python/transpiler/test_unitary_synthesis.py
```diff
@@ -673,38 +569,6 @@ def test_coupling_map_transpile_with_backendv2(self, opt_level, bidirectional):
                 (0, 1), (circ_01_index[instr.qubits[0]], circ_01_index[instr.qubits[1]])
             )
 
-    @data(1, 2, 3)
-    def test_coupling_map_unequal_durations(self, opt):
```
ElePT, line 677: I removed this test because we originally had it as the backend V1 version of `test_coupling_unequal_duration_with_backendv2`, it was then migrated to backend v2 and became essentially a duplicate of the other one.

ShellyGarion (approved):
LGTM. Thanks @ElePT for the tests refactoring, it's very helpful!

It's not part of this PR, but I would just like to point out that it seems that the `XXDecomposer` lines inside the `unitary_synthesis` transpiler pass are not being covered by any of the tests, see the coverage report:
https://coveralls.io/builds/71414685/source?filename=qiskit%2Ftranspiler%2Fpasses%2Fsynthesis%2Funitary_synthesis.py

PR #13584, opened 2024-12-19
Deprecation of MCMT in favor of MCMTGate

1ucian0:
The Multiple-Control-Multiple-Target in `qiskit.circuit.library.generalized_gates.MCMT` was pending for deprecation in https://github.com/Qiskit/qiskit/pull/13150 . This PR moves it to deprecated.


Cryoris:
I think we want to deprecate the circuit library objects for 1.4 and 2.0, but might not want to remove them for 2.0. Even though this PR only changes from `PendingDeprecationWarning` to `DeprecationWarning`, we probably should discuss what the exact plan is 🙂

releasenotes/notes/followup_13150-5bd0c77248601e1a.yaml
```diff
@@ -0,0 +1,6 @@
+---
+deprecations_circuits:
+  - |
+    The Multiple-Control-Multiple-Target in :class:`~qiskit.circuit.library.generalized_gates.MCMT` is now deprecated
+    and replaced by :class:`MCMTGate`, which is a proper :class:`Gate` subclass.
```
Cryoris:
```suggestion
    and replaced by :class:`.MCMTGate`, which is a proper :class:`.Gate` subclass. Using
    a gate instead of a circuit allows the compiler to reason about the object at a higher
    level of abstraction and allows for multiple synthesis plugins.
```

Cryoris (approved):

Cryoris (changes requested):
Sorry I was too eager to approve -- this should actually be merged both into 2.0 and 1.4 since we want to keep the `MCMT` circuit (and all other library objects) deprecated throughout the 2.0 cycle and only remove them in 3.0.

1ucian0:
@mergifyio backport stable/1.4

Cryoris (dismissed):
Thanks!

Cryoris (approved):

PR #13591, opened 2024-12-20
Fix `UnitarySynthesis` for 3+ qubits when compiled for a backend

### LINKED ISSUE #13586: Rust exeception raised when transpiling PaulEvolutionGate with measurements against backend
### Environment

- **Qiskit version**: 1.3.1
- **Python version**: 
- **Operating system**: linux

### What is happening?

```python

backend = provider.backend('ibm_nazca')

hamiltonian = SparsePauliOp.from_list([('IXX', 1), ('IYY', 1), ('IZZ', 1),
                                       ('XXI', 1), ('YYI', 1), ('ZZI', 1)])

N = 8
qc = QuantumCircuit(3)
# Initial state 110
qc.x([1, 2])
op = PauliEvolutionGate(hamiltonian, time=1)
qc.append(op.power(N), [0,1,2])
qc.measure_all()

transpile(qc, backend)
```

gives

thread '<unnamed>' panicked at crates[/circuit/src/interner.rs:147:14](http://localhost:8888/circuit/src/interner.rs#line=146):
the caller is responsible for only using interner keys from the correct interner
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace

PanicException: the caller is responsible for only using interner keys from the correct interner

run above

### What should happen?

it should work.  It does if I first compile against no backend, and then use those circuits as input to the backend transpile

### Any suggestions?

No clue

Cryoris:
### Summary

When encountering a 3+ qubit `UnitaryGate`, `UnitarySynthesis` currently replaces the whole dag with the definition of the block instead of appending it, see
https://github.com/Qiskit/qiskit/blob/079f69f7dd906299dfdf8c1262fa9bc1d75e478b/crates/accelerate/src/unitary_synthesis.rs#L375

### Details and comments

This also leads to #13586 because the qubit/clbit interners are overwritten, but maybe more importantly this can lead to issues when multiple 3-qubit blocks are present in the circuit. E.g. running
```python
import numpy as np
from qiskit_ibm_runtime.fake_provider import FakeSherbrooke

backend = FakeSherbrooke()

x = QuantumCircuit(3)
x.x(2)
x_mat = Operator(x)

qc = QuantumCircuit(3, 1)
qc.unitary(x_mat, range(3))
qc.unitary(np.eye(2 ** 3), range(3))

tqc = transpile(qc, backend, optimization_level=0)
print(tqc.count_ops())
```
produces an _empty_ circuit, since the last block implements the identity.

Some additional tests, such as the precise case of #13586 and a series of 3q `UnitaryGate`s, should be added.

### DIFF
--- a/releasenotes/notes/fix-unitary-synthesis-3q-2b2de5305bfd11ff.yaml
+++ b/releasenotes/notes/fix-unitary-synthesis-3q-2b2de5305bfd11ff.yaml
@@ -1,0 +1,8 @@
+---
+fixes:
+  - |
+    Fixed a bug in the :class:`.UnitarySynthesis` transpiler pass, where blocks of
+    :class:`.UnitaryGate`\s of 3 qubits or more were not correctly synthesized.
+    This lead e.g. to the circuit being overwritten with the last processed block or
+    to internal panics when encountering measurements after a such a block.
+    Fixed `#13586 <https://github.com/Qiskit/qiskit/issues/13586>`__.
--- a/releasenotes/notes/fix-unitary-synthesis-3q-2b2de5305bfd11ff.yaml
+++ b/releasenotes/notes/fix-unitary-synthesis-3q-2b2de5305bfd11ff.yaml
@@ -2,7 +2,7 @@
 fixes:
   - |
     Fixed a bug in the :class:`.UnitarySynthesis` transpiler pass, where blocks of
-    :class:`.UnitaryGate`\s of 3 qubits or more were not correctly synthesized.
-    This lead e.g. to the circuit being overwritten with the last processed block or
-    to internal panics when encountering measurements after a such a block.
+    :class:`.UnitaryGate`\s on 3 qubits or more were not correctly synthesized.
+    This led, e.g., to the circuit being overwritten with the last processed block or
+    to internal panics when encountering measurements after such a block.
     Fixed `#13586 <https://github.com/Qiskit/qiskit/issues/13586>`__.
--- a/test/python/transpiler/test_unitary_synthesis.py
+++ b/test/python/transpiler/test_unitary_synthesis.py
@@ -57,7 +57,9 @@ from qiskit.circuit.library import (
     RYYGate,
     RZZGate,
     RXXGate,
+    PauliEvolutionGate,
 )
+from qiskit.quantum_info import SparsePauliOp
 from qiskit.circuit import Measure
 from qiskit.circuit.controlflow import IfElseOp
 from qiskit.circuit import Parameter, Gate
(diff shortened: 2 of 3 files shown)


releasenotes/notes/fix-unitary-synthesis-3q-2b2de5305bfd11ff.yaml
```diff
@@ -0,0 +1,8 @@
+---
+fixes:
+  - |
+    Fixed a bug in the :class:`.UnitarySynthesis` transpiler pass, where blocks of
+    :class:`.UnitaryGate`\s of 3 qubits or more were not correctly synthesized.
+    This lead e.g. to the circuit being overwritten with the last processed block or
+    to internal panics when encountering measurements after a such a block.
+    Fixed `#13586 <https://github.com/Qiskit/qiskit/issues/13586>`__.
```
alexanderivrii:
```suggestion
    Fixed a bug in the :class:`.UnitarySynthesis` transpiler pass, where blocks of
    :class:`.UnitaryGate`\s on 3 qubits or more were not correctly synthesized.
    This led, e.g., to the circuit being overwritten with the last processed block or
    to internal panics when encountering measurements after such a block.
    Fixed `#13586 <https://github.com/Qiskit/qiskit/issues/13586>`__.
```

alexanderivrii:
Thanks for the fix, @Cryoris! 

Looks good! (modulo some minor spell-checking nitpicks about the release notes)

> Some additional tests, such as the precise case of https://github.com/Qiskit/qiskit/issues/13586 and a series of 3q UnitaryGates, should be added.

Are you planning to add a test for that (like the one you have in the code snippet in the description)?

Cryoris:
@alexanderivrii more tests added in 583ec3d 🙂

test/python/transpiler/test_unitary_synthesis.py
```diff
@@ -1062,6 +1064,43 @@ def test_3q_with_measure(self):
...
+
+        qc = QuantumCircuit(3)
+        qc.unitary(x_mat, range(3))
+        qc.unitary(np.eye(2**3), range(3))
+
+        tqc = transpile(qc, backend, optimization_level=0, initial_layout=[0, 1, 2])
+
+        expected = np.kron(np.eye(2**2), x_mat)
+        self.assertTrue(Operator(tqc).equiv(expected))
+
```
alexanderivrii: Is there a reason that we are checking that the two operators are equivalent vs. being exactly the same?

Cryoris: Mainly because I'm not sure if `UnitarySynthesis` is guaranteed to have the correct global phase 🙂 but if it is, then we can use `==` instead

alexanderivrii: Yes, having the correct global phase after transpilation is one of the guarantees that the transpiler/each individual pass must uphold.

alexanderivrii (approved):

Issue #13391, opened 2024-11-01
Implement the `qargs` temporary state for `SparseObservable`


jakelishman:
### What should we add?

Most `quantum_info` operators, including `SparsePauliOp`, have a `qargs` argument in several of their mathematical methods, and have a `__call__` method that creates a shallow copy of the `SparsePauliOp` with the `qargs` temporarily set, for use with the Python binary operators (like `+`).  The Python version of `SparseObservable` should also gain this functionality.

A `qargs` keyword argument to certain methods is straightforwards enough.  Making the shallow-copy-on-`__call__` behaviour is more challenging because the data is owned by Rust, and can't be easily shared like it can in Python.  Likely we will want to make an entirely clean split between the Rust-space `SparseObservable` and the Python-exposed version (which I'll call `PySparseObservable`, but its Python name will also be `SparseObservable`) that contains something like an `Arc<RwLock<SparseObservable>>` and whatever shallow additional Python state we want.  Something like:

```rust
// The same as it is now.  *Not* a pyclass.
pub struct SparseObservable {
    num_qubits: u32,
    coeffs: Vec<Complex64>,
    bit_terms: Vec<BitTerm>,
    indices: Vec<u32>,
    boundaries: Vec<usize>,
}
impl SparseObservable {
    // All the actual worker methods on the struct.
}

#[pyclass(name="SparseObservable")]
struct PySparseObservable {
    inner: Arc<RwLock<SparseObservable>>,
    qargs: Option<Vec<u32>>,
}
#[pymethods]
impl PySparseObservable {
    // All the Python methods.  These should generally be thin wrappers around proper
    // `SparseObservable` Rust-space methods, and the pymethods only handle interaction
    // with Python.
}
```

Shobhit21287:
I'd like to take this up

PR #13595, opened 2024-12-23
Split ``PySparseObservable`` off ``SparseObservable``

### LINKED ISSUE #13594: Separate ``SparseObservable`` into a Rust-only core and Python interface
As discussed in #13445  and #13391, separate ``SparseObservable`` into a Rust-only core, that does not have any PyO3/Python dependency, and Python interface.

Cryoris:
### Summary

Closes #13594 to prepare for ``SparseObservable``'s C API. This change has been tested with our basic C API for ``SparseObservable``, which will come in a separate PR to keep the review load in balance 🙂 

### Details and comments

This PR splits the sparse observable class into a Rust-only ``SparseObservable`` struct and a ``PySparseObservable``, which serves as Python interface. As suggested in #13391, the Python interface keeps an ``Arc`` to a read-write-locked ``SparseObservable``. The ``API Change`` label is only due to some minuscule change in an error message, the Python interface remains unchanged.

The implementation is based on
```rust
#[pyclass(name = "SparseObservable", ...)]  // exposed as qiskit.quantum_info.SparseObservable, as before
struct PySparseObservable {
    // This class keeps a pointer to a pure Rust-SparseTerm and serves as interface from Python.
    inner: Arc<RwLock<SparseObservable>>,
}
```
and methods on ``PySparseObservable`` first acquire the read- or write-lock to perform actions on the `inner` data. For example, implementing `transpose` becomes
```rust
    fn transpose(&self) -> PyResult<Self> {
        // acquire the read lock, mapping the PoisonError into our own error that can be cast to a PyErr
        let inner = self.inner.read().map_err(|_| InnerReadError)?;

        // perform the action
        let result = inner.transpose();  
        
        // return a new Arc<RwLock> (if we did an inplace operation, we would just return nothing)
        Ok(Self { inner: Arc::new(RwLock::new(result)) })
    }
```

Some notes/questions:
* For ``SparseTerm`` we analogously split off ``PySparseTerm``, since it can be returned to Python. The view/mutable view versions are not returned to Python and don't need a specific interface.
* We couldn't implement `IntoPy` to `PoisonError` (coming from RwLock::read/write), so as solution we introduced custom ``InnerReadError``s and ``InnerWriteError``s.
* We moved some methods from the `pymethods` into the core Rust object and restricted direct access to the inner data, in favor of using public getters/methods.
* The ``SparseObservable`` docstring is moved to the Python interface for now, though we might want to add a bit more Rust-specific info.


crates/accelerate/src/py_sparse_observable.rs
```diff
@@ -0,0 +1,2409 @@
...
+/// These are typically created by indexing into or iterating through a :class:`SparseObservable`.
+#[pyclass(name = "Term", frozen, module = "qiskit.quantum_info")]
+#[derive(Clone, Debug)]
+struct PySparseTerm {
+    // This class keeps a pointer to a pure Rust-SparseTerm and serves as interface from Python.
+    // Note that we can use the standard clone, since the PySparseTerm is immutable from Python
+    // space, hence we can just clone the pointer. Otherwise we'd have to impl Clone manually,
+    // like for the PySparseObservable.
+    inner: Arc<RwLock<SparseTerm>>,
+}
```
jakelishman: Why does `PySparseTerm` need to include an `Arc<RwLock<SparseTerm>>`, rather than just being a wrapper around `SparseTerm`?

crates/accelerate/src/py_sparse_observable.rs
```diff
@@ -0,0 +1,2409 @@
...
+    ///
+    ///         >>> SparseObservable.zero(100)
+    ///         <SparseObservable with 0 terms on 100 qubits: 0.0>
+    #[pyo3(signature = (/, num_qubits))]
+    #[staticmethod]
+    pub fn zero(num_qubits: u32) -> Self {
+        let inner = SparseObservable::zero(num_qubits);
+        Self {
+            inner: Arc::new(RwLock::new(inner)),
+        }
```
jakelishman: In normal situations, this kind of thing is what the `IntoPy`/`IntoPyObject` trait is for.  I think we can do that in a coherent manner, even with a clean crate structure, since `IntoPyObject` takes a type argument that will be `Py<PySparseObservable>`.

crates/accelerate/src/py_sparse_observable.rs
```diff
@@ -0,0 +1,2409 @@
...
+                .into());
+            }
+            Ok(inferred)
+        };
+
+        // get the number of qubits in the layout and map the layout to Vec<u32> to
+        // call SparseObservable.apply_layout
+        let (num_qubits, layout): (u32, Option<Vec<u32>>) = if layout.is_none() {
+            // if the layout is none,
+            (num_qubits.unwrap_or(inner.num_qubits()), None)
```
jakelishman: Some of the comments left around the file seem like they're old helper reminders?  Might be good to run through and clean them up.

crates/accelerate/src/py_sparse_observable.rs
```diff
@@ -0,0 +1,2409 @@
...
+        Ok(format!(
+            "<SparseObservable with {} on {}: {}>",
+            str_num_terms, str_num_qubits, str_terms
+        ))
+    }
+
+    fn __reduce__(&self, py: Python) -> PyResult<Py<PyAny>> {
+        // we acquire the read lock once here and use the internal methods
+        // to avoid checking and acquiring the lock in every method call
+        let inner = self.inner.read().map_err(|_| InnerReadError)?;
```
jakelishman: If this comment is helpful to you, feel free to leave it, but this is really the base expectation for how to handle the `RwLock` - it would be weird to keep locking and unlocking.

crates/accelerate/src/py_sparse_observable.rs
```diff
@@ -0,0 +1,2409 @@
...
+impl Clone for PySparseObservable {
+    // we need to implement clone specifically to ensure the underlying inner data
+    // is cloned, not the reference to it
+    fn clone(&self) -> Self {
+        let inner = self.inner.read().unwrap();
+        Self {
+            inner: Arc::new(RwLock::new(inner.clone())),
+        }
+    }
+}
```
jakelishman: What's the reason for this implementation?  In what situations are you thinking that this will be called from Rust space, and how sure are you that these semantics are always what would be expected?  I'm pretty uncomfortable with a `Clone` implementation that can panic.

Since this is object intended for consumption only from Python space, perhaps it would be cleaner to override `__copy__` and `__deepcopy__`, if `copy.copy` and `copy.deepcopy` aren't doing what you expect (though I think they probably should already work).

crates/accelerate/src/py_sparse_observable.rs
```diff
@@ -0,0 +1,2409 @@
...
+            // struggle to do the addition to itself.
+            let slf_ = slf_.borrow();
+            let inner = slf_.inner.read().map_err(|_| InnerReadError)?;
+            let doubled =
+                <&SparseObservable as ::std::ops::Mul<_>>::mul(&inner, Complex64::new(2.0, 0.0));
+            return Ok(Self {
+                inner: Arc::new(RwLock::new(doubled)),
+            }
+            .into_py(py));
+        }
```
jakelishman: The `is` check here, and everywhere, quite possibly needs rethinking now, because two non-referentially equal Python objects can (in theory) contain pointers to the same Rust-space `SparseObservable`, which will cause `RwLock` problems in `__iadd__`.

I think this PR doesn't _expose_ anyway to produce two such objects, but the `PySparseObservable` type makes it very possible, and the shallow-copy stuff we're also planning to do for `qargs` etc will make it very likely to happen.

crates/accelerate/src/py_sparse_observable.rs
```diff
@@ -0,0 +1,2409 @@
...
+        let other = other.borrow();
+        let other_inner = other.inner.read().map_err(|_| InnerReadError)?;
+        slf_inner.check_equal_qubits(&other_inner)?;
+        slf_inner.sub_assign(&other_inner);
+        Ok(())
+    }
+
+    fn __pos__(&self) -> Self {
+        self.clone()
+    }
```
jakelishman: I would suggest that this is a dangerous call - to me, `self.clone()` should be shallow because this is an `Arc<RwLock<T>>` wrapper.  `self.clone()` can also panic, but we should be returning a proper Python-space error to users.

I'd suggest `self.inner.read().map_err(|_| InnerReadError)?.clone().into_py(py)` (along with various comments from above about `into_py`) or some helper wrapper around that if it's common.

crates/accelerate/src/sparse_observable.rs
```diff
@@ -110,7 +90,7 @@ impl BitTerm {
     /// Get the label of this `BitTerm` used in Python-space applications.  This is a single-letter
     /// string.
     #[inline]
-    fn py_label(&self) -> &'static str {
+    pub(crate) fn py_label(&self) -> &'static str {
```
jakelishman: If these are purely for Python consumption and unrelated to the Rust code (which I believe is the case), perhaps consider moving them to be regular functions that live with the rest of the Python-specific code?

crates/accelerate/src/sparse_observable.rs
```diff
@@ -146,7 +126,7 @@ impl BitTerm {
     /// returning `Ok(None)` for it.  All other letters outside the alphabet return the complete
     /// error condition.
     #[inline]
-    fn try_from_u8(value: u8) -> Result<Option<Self>, BitTermFromU8Error> {
+    pub(crate) fn try_from_u8(value: u8) -> Result<Option<Self>, BitTermFromU8Error> {
```
jakelishman: If this needs to be `pub` beyond this module, I don't see much reason to make it `pub(crate)` and not just `pub`, looking to a future.

crates/accelerate/src/sparse_observable.rs
```diff
@@ -327,11 +201,8 @@ pub enum CoherenceError {
     DuplicateIndices,
     #[error("the provided qubit mapping does not account for all contained qubits")]
     IndexMapTooSmall,
-}
-impl From<CoherenceError> for PyErr {
-    fn from(value: CoherenceError) -> PyErr {
-        PyValueError::new_err(value.to_string())
-    }
+    #[error("cannot shrink the qubit count in an observable from {current} to {target}")]
+    NotEnoughQubits { current: usize, target: usize },
```
jakelishman, line 234: I think this is just a straight-up code change?  Might be better to do that in a separate PR, if you think it's necessary.

crates/accelerate/src/sparse_observable.rs
```diff
@@ -348,420 +219,16 @@ pub enum LabelError {
...
-/// observable using ``+=`` is amortised linear in the total number of terms added, rather than
-/// the quadratic complexity that the binary ``+`` would require.
-///
-/// Additions and subtractions are implemented by a term-stacking operation; there is no automatic
-/// "simplification" (summing of like terms), because the majority of additions to build up an
-/// observable generate only a small number of duplications, and like-term detection has additional
-/// costs.  If this does not fit your use cases, you can either periodically call :meth:`simplify`,
-/// or discuss further APIs with us for better building of observables.
-#[pyclass(module = "qiskit.quantum_info", sequence)]
+/// See PySparseObservable in crates/accelerate/src/py_sparse_observable for detailed docs.
```
jakelishman: Possibley

> See `[PySparseObservable]` for detailed docs.

might be clearer that it's just implied you should go find that struct, and less likely for the file to get out of sync.

crates/accelerate/src/sparse_observable.rs
```diff
@@ -905,6 +372,83 @@ impl SparseObservable {
...
+
+    #[inline]
+    pub fn bit_terms(&self) -> &Vec<BitTerm> {
+        &self.bit_terms
+    }
+
+    #[inline]
+    pub fn bit_terms_mut(&mut self) -> &mut Vec<BitTerm> {
+        &mut self.bit_terms
+    }
```
jakelishman: Returning `&Vec<T>` is almost never correct; you almost invariably mean `&[T]` (and you do, in every case here).  `&mut Vec` can be correct in some circumstances, but in this case is very not - all these methods allow the `Vec`-resizing methods to be called, which easily makes the data incoherent.  For ones we definitely need to expose, it should be `&mut [T]`.

`boundaries_mut` and `indices_mut` might very well be considered `unsafe`, since you can easily break data coherence by writing bad values to them.  I can't remember what the old code did around this, but if they are actually needed and used, then they may well want to be `unsafe` in Rust interfaces.

releasenotes/notes/update-sparse-observable-error-2bb4b9e678675eee.yaml
```diff
@@ -0,0 +1,8 @@
+---
+upgrade:
+  - |
+    Binary arithmetic operations on :class:`.SparseObservable` have a slightly updated
+    error message, if the two observables in the operation have incompatible numberss of qubits.
+    Previously, the resulting ``ValueError`` contained the message
+    ``incompatible numbers of qubits: A and B``, which now is
+    ``mismatched numbers of qubits: A, B``.
```
jakelishman: This is good thinking, but I don't think we really need to consider this an interface break.

jakelishman (commented):
Thanks for doing this.

This is just a quick high-level overview - I'll look more in detail in the new year, especially since I'll have to use a lot more local tools to do a good comparison - with the file move and changes to the code, it's hard to see what's gone on here.

Top level questions:

- Why split `py_sparse_observable` into a separate flat file?  I'd have expected any of:
  * keep both in the same file
  * make a `sparse_observable` module to put them in
  * make a separate crate that contains only the C component
  
  with a rough preference to just keeping everything in the same file for now.  This form to me has meant that a lot of logically private functions have had to become `pub(crate)`, and now there's more places to look to understand the code.
- For everything that's become `pub(crate)`: in some cases, I think `pub(crate)` just indicates that a function is defined in the wrong file.  In many others, since this PR is looking to a future when `SparseObservable` is consumable by non-Qiskit crates directly from Rust, I suspect that anything that became `pub(crate)` should be either private or fully `pub`.  If it's useful for the Python wrapper, feels highly likely it ought to be a proper public interface.

crates/accelerate/src/py_sparse_observable.rs
Cryoris: Ah yes, this is actually one of the to-dos we didn't try in the end, that would be much nicer 🙂

crates/accelerate/src/sparse_observable.rs
Cryoris: Hmmm I think we added this because it comes up in `apply_layout`, which was previously only a Python method which directly triggered a `PyErr`. Now that `apply_layout` is moved into the Rust methods, we needed a new error -- so it's linked to this PR due to the move of the `apply_layout` logic 🙂 (which could ofc also be separated out if you prefer)

crates/accelerate/src/sparse_observable.rs
Cryoris: We might want to make it `pub` with some of the C functionality, but I wanted to be as cautious as possible -- we can just make it `pub` pending the structural question you raised above 🙂

crates/accelerate/src/sparse_observable.rs
Cryoris: Yeah on this bit we were initially wondering whether we want the same lookups for C, to have a nice print for the `BitTerm` enum (not expose a "C string" but just print the label from Rust side). But in that case we ought to remove the `py_` prefix!

crates/accelerate/src/py_sparse_observable.rs
Cryoris: The reason is that we didn't think of putting this into `__copy__` and we wanted to ensure we clone the object, not the ref 🙂 You're right, it's not really intended to be called from Rust space

> (though I think they probably should already work).

I.e. that Python's native `__copy__` would already copy the `inner` data instead of the reference to it?

crates/accelerate/src/py_sparse_observable.rs
Cryoris: I'll remove it, it was mainly helpful for my own sanity starting to use these objects 😜

crates/accelerate/src/py_sparse_observable.rs
Cryoris: You mean because it's immutable anyways?

Cryoris:
Thanks for the comments! I think they all make sense but I'll read them more carefully next year as well 🙂 Regarding

> Why split py_sparse_observable into a separate flat file? I'd have expected any of:
> * keep both in the same file
> * make a sparse_observable module to put them in
> * make a separate crate that contains only the C components
> 
> with a rough preference to just keeping everything in the same file for now. This form to me has meant that a lot of logically private functions have had to become pub(crate), and now there's more places to look to understand the code.

To me, having a separate crate (I assume into `py_ext`?) sounds the cleanest, but I didn't want just move it w/o discussion, so I moved it into a separate file to facilitate that process 😛 I'm fine with keeping it in the same file as well for now too, though.

releasenotes/notes/update-sparse-observable-error-2bb4b9e678675eee.yaml
Cryoris: Ah even better -- I was just being double safe 😉

crates/accelerate/src/py_sparse_observable.rs
jakelishman: I can't imagine a situation where there'd be shared ownership of an underlying `SparseTerm` - the only reason the `Arc<RwLock<T>>` is useful for `SparseObservable` is because of the shallow-copy `qargs`-after-`__call__` behaviour of the QI classes (so strictly it's not even needed for this PR).

crates/accelerate/src/py_sparse_observable.rs
jakelishman: `copy.copy` and `copy.deepcopy` don't _need_ `__copy__` etc to be defined - they'll use the pickle-protocol methods (`__reduce__` or  `__getstate__`/`__statestate__`) if they're not, and that causes `SparseObservable` to integrate with them already. I'm relatively sure I wrote a test.

crates/accelerate/src/sparse_observable.rs
jakelishman: Also fine, yeah. I guess it's important to note that the actual values here are part of the Python interface and need to be stable.

crates/accelerate/src/sparse_observable.rs
jakelishman: Personally I've seen very few reasons for `pub(crate)` so far - most of them have either been that the function is misplaced, an API boundary has got very muddled, or they should just be `pub`. We've had quite a lot of churn turning `pub(crate)` into `pub`.

crates/accelerate/src/sparse_observable.rs
jakelishman: Ok sure, if it's actually an important part of the split, it's fine to do here. (Part of the difficulty in reviewing this kind of PR!)

jakelishman:
btw should Max be a co-author, or is this all you so far?

crates/accelerate/src/sparse_observable.rs
Cryoris: Oh right, it should definitely not be `Vec` here -- I'll fix that 🙂 

Regarding necessity of these methods: this mutable access is used in `ArrayView::__setitem__`, which previously had a `SparseObservable` as base and simply used mutable access to the internals. But now `PySparseObservable` it the base and we don't have this access anymore. Using `unsafe` seems fine, but maybe we should add some validation in `__setitem__` 😄

crates/accelerate/src/py_sparse_observable.rs
Cryoris: Right, we could also check whether the `inner` observables are the same (if we only need read access we should be able to do that even if `other` is `slf_`) but then we'll get into troubles with different `qargs` (which we would need to also check, but that's possible).

The way it is right now it is more conservative, in the fact that adding two `PySparseObservable` that link to the same `SparseObservable` would not return 2x the observable. I thought this is the safer way, but indeed this might give rise to some unexpected behavior...

The last option is to just remove this special case.

Writing this down the first or the last maybe seem the most reasonable options, or what do you think?

jakelishman: The reason to have the special case at all is because we cannot modify the inner `SparseObservable` with an in-place operation if the right-hand side of the operation involves the same `SparseObservable`, even if the two have different Python wrappers have different `qargs`; it would be memory unsafe.

So one way or another, we're protecting against runtime borrow-check errors in the `RwLock` from killing the program.  We can either do that by actually inspecting the output of `RwLock::try_borrow` rather than panicking, though it's probably faster and at the correct level of abstraction for the dispatch if we check `Arc::ptr_eq` (which is effectively the Python `is` check).

releasenotes/notes/update-sparse-observable-error-2bb4b9e678675eee.yaml
Cryoris: removed in 4aff73b

crates/accelerate/src/sparse_observable.rs
Cryoris: done in 4aff73b

crates/accelerate/src/py_sparse_observable.rs
Cryoris: Done in c790cff

crates/accelerate/src/py_sparse_observable.rs
Cryoris: The `Clone` implementation is removed in c790cff

crates/accelerate/src/py_sparse_observable.rs
Cryoris: removed in 4aff73b

crates/accelerate/src/py_sparse_observable.rs
Cryoris: removed in 4aff73b

crates/accelerate/src/py_sparse_observable.rs
Cryoris: Added in e4a4db8

crates/accelerate/src/py_sparse_observable.rs
Cryoris: Updated to just keep a `SparseTerm` in fd2f3ee

crates/accelerate/src/sparse_observable.rs
Cryoris: 3a4b5f0 merged the `py_sparse_observable.rs` file back into `sparse_observable.rs` so we don't need to touch the visibility of these methods anymore (though we likely will for the C API later)

Cryoris:
With 3a4b5f0, all code is in a single file for now, but it should be easy to separate if we want to later on (basically the lower half is the Python wrapper) 

> btw should Max be a co-author, or is this all you so far?

This was my bit so far, since I was reviewing the `SparseObservable` PRs initially I volunteered to do the split 🙂

crates/accelerate/src/py_sparse_observable.rs
Cryoris: Using `Arc::ptr_eq` sounds good, though there's no way we can test this case (two `PySparseObservable` referencing the same `SparseObservable`) in the Python tests right now, right?

jakelishman: Right now no, because we don't have a Python-space way of creating two `PySparseObservable`s that point to the same `SparseObservable`.  We can test it from Rust space following #13169, though.

qiskit/__init__.py
```diff
@@ -78,7 +78,7 @@
 sys.modules["qiskit._accelerate.results"] = _accelerate.results
 sys.modules["qiskit._accelerate.sabre"] = _accelerate.sabre
 sys.modules["qiskit._accelerate.sampled_exp_val"] = _accelerate.sampled_exp_val
-sys.modules["qiskit._accelerate.sparse_observable"] = _accelerate.sparse_observable
+sys.modules["qiskit._accelerate.sparse_observable"] = _accelerate.py_sparse_observable
```
jakelishman: Let's set the name of the generated Python module so that this is called the correct thing here as well, so `from qiskit._accelerate import sparse_observable` would work (in principle) too.

crates/accelerate/src/sparse_observable.rs
```diff
@@ -348,1190 +248,1905 @@ pub enum LabelError {
...
+
+    /// The coefficients of each abstract term in in the sum.  This has as many elements as terms in
+    /// the sum.
+    #[getter]
+    fn get_coeffs(slf_: Py<Self>) -> ArrayView {
+        ArrayView {
+            base: slf_,
+            slot: ArraySlot::Coeffs,
+        }
     }
```
jakelishman: Strictly for these, I think we could store a clone of the `Arc<RwLock<SparseObservable>>` in the `ArrayView::base` - we were using `Py` for the ref-counted smart pointer part, but now we've got `Arc` in the mix, we could do it raw in Rust.

I don't think it's very important, though - the existing way is fine too.

jakelishman (commented):
This looks like a really nice, clean split, thank you!

I pushed up one minor stylistic commit (just moving trait impls closer to the relevant structs, except when doing so would move "Python" code into "Rust" code), and I sorted out the merge conflict by upgrading the `PySparseObservable` conversions into the new PyO3 form.  Feel free to mess with those two commits if I've done something against what you wanted.

I had one non-actionable comment and one pretty minor nitpick, but other than that I'm very happy with this and ready to merge!

crates/accelerate/src/py_sparse_observable.rs
Cryoris: I opened #13726 to keep track of this

qiskit/__init__.py
Cryoris: done in a18a5a5

crates/accelerate/src/sparse_observable.rs
Cryoris: done in a18a5a5

crates/accelerate/src/sparse_observable.rs
Cryoris: I added an inline comment on this in a18a5a5

crates/pyext/src/lib.rs
```diff
@@ -55,7 +55,7 @@ fn _accelerate(m: &Bound<PyModule>) -> PyResult<()> {
     add_submodule(m, ::qiskit_accelerate::results::results, "results")?;
     add_submodule(m, ::qiskit_accelerate::sabre::sabre, "sabre")?;
     add_submodule(m, ::qiskit_accelerate::sampled_exp_val::sampled_exp_val, "sampled_exp_val")?;
-    add_submodule(m, ::qiskit_accelerate::sparse_observable::py_sparse_observable, "py_sparse_observable")?;
+    add_submodule(m, ::qiskit_accelerate::sparse_observable::sparse_observable, "sparse_observable")?;
```
jakelishman: Technically this was the only line that actually needed to chance to implement my query (and then, just the string name), but giving them the same name is better anyway.

At some point in the near-ish future when I've lost my rage at my initial failed attempt, I'll post a PR that turns the submodule-construction logic back into more native PyO3 tooling, since PyO3 0.23 gave us the APIs we need, and `abi3audit` removed the failure we were getting from it anyway (for reasons I don't 100% agree with).

jakelishman (dismissed):
Thanks for all the work on this!

jakelishman:
Wait, no - one second, I think your last merge undid a bunch of stuff.

jakelishman (approved):

jakelishman:
I straightened out the history to remove the second merge, but actually I think I was mistaken anyway - I think I just confused myself because the merge commit you pushed up had the same diff as the one I'd made (since those _were_ the changes you were merging).

Cryoris:
Thanks to you for the help! I'm very happy to see this merged and move forward with the C API bit 😁

PR #13596, opened 2024-12-23
Only apply MCMT plugin on ``MCMTGate``

### LINKED ISSUE #13563: Error transpiling circuit containing `MCMT` gate
### Environment

- **Qiskit version**: 1.3.1
- **Qiskit_aer version**: 0.15.1
- **Python version**: 3.12.3
- **Operating system**: Linux

### What is happening?

Transpiling a circuit containing an `MCMT` gate with the `aer_simulator_unitary` fails. A very long exception trace ends with: `AttributeError: 'Instruction' object has no attribute 'num_ctrl_qubits'`.

This is a regression from qiskit 1.2.4.

```python
from qiskit import QuantumCircuit, QuantumRegister, transpile
from qiskit.circuit.library import MCMT, RYGate
from qiskit_aer import Aer


qr = QuantumRegister(2, "q")
qcirc = QuantumCircuit(qr)
ry_gate = RYGate(0.1)
cry_gate = MCMT(gate=ry_gate, num_ctrl_qubits=1, num_target_qubits=1)
qcirc.append(cry_gate, [qr[0], qr[1]])

backend = Aer.get_backend("aer_simulator_unitary")

transpile(qcirc, backend)
```

### What should happen?

Transpilation should succeed (as it does with qiskit 1.2.4).

### Any suggestions?

_No response_

Cryoris:
### Summary

Fixes #13563, by enabling the high-level synthesis plugins of MCMT only for the `MCMTGate`, not any gate called `"mcmt"`.

### Details and comments

While we often let it be the user's responsibility to be mindful with gate names, this is a special case as we provide a circuit called `"mcmt"`. This can cause innocent-looking code to break, so we should add this check at least until the `MCMT` circuit is removed 🙂

### DIFF
--- a/qiskit/transpiler/passes/synthesis/hls_plugins.py
+++ b/qiskit/transpiler/passes/synthesis/hls_plugins.py
@@ -420,6 +420,7 @@ from qiskit.circuit.library import (
     C3XGate,
     C4XGate,
     PauliEvolutionGate,
+    MCMTGate,
     ModularAdderGate,
     HalfAdderGate,
     FullAdderGate,
@@ -1156,6 +1157,9 @@ class MCMTSynthesisDefault(HighLevelSynthesisPlugin):
 
     def run(self, high_level_object, coupling_map=None, target=None, qubits=None, **options):
         # first try to use the V-chain synthesis if enough auxiliary qubits are available
+        if not isinstance(high_level_object, MCMTGate):
+            return None
+
         if (
             decomposition := MCMTSynthesisVChain().run(
                 high_level_object, coupling_map, target, qubits, **options
@@ -1170,6 +1174,9 @@ class MCMTSynthesisNoAux(HighLevelSynthesisPlugin):
     """A V-chain based synthesis for ``MCMTGate``."""
 
     def run(self, high_level_object, coupling_map=None, target=None, qubits=None, **options):
+        if not isinstance(high_level_object, MCMTGate):
+            return None
+
         base_gate = high_level_object.base_gate
         ctrl_state = options.get("ctrl_state", None)
 
@@ -1195,6 +1202,9 @@ class MCMTSynthesisVChain(HighLevelSynthesisPlugin):
     """A V-chain based synthesis for ``MCMTGate``."""
 
     def run(self, high_level_object, coupling_map=None, target=None, qubits=None, **options):
+        if not isinstance(high_level_object, MCMTGate):
+            return None
+
         if options.get("num_clean_ancillas", 0) < high_level_object.num_ctrl_qubits - 1:
             return None  # insufficient number of auxiliary qubits
 
--- a/qiskit/transpiler/passes/synthesis/hls_plugins.py
+++ b/qiskit/transpiler/passes/synthesis/hls_plugins.py
@@ -420,13 +420,13 @@ from qiskit.circuit.library import (
     C3XGate,
     C4XGate,
     PauliEvolutionGate,
+    PermutationGate,
     MCMTGate,
     ModularAdderGate,
     HalfAdderGate,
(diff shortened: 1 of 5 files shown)


test/python/circuit/library/test_mcmt.py
```diff
@@ -285,6 +285,19 @@ def test_gate_with_parameters_vchain(self):
...
 
+    def test_mcmt_circuit_as_gate(self):
+        """Test the MCMT plugin is only triggered for the gate, not the same-named circuit.
+
+        Regression test of #13563.
+        """
+        circuit = QuantumCircuit(2)
+        gate = RYGate(0.1)
+        mcmt = MCMT(gate=gate, num_ctrl_qubits=1, num_target_qubits=1)
+        circuit.append(mcmt, circuit.qubits)  # append the MCMT circuit as gate called "MCMT"
```
alexanderivrii, line 296: A bit off-topic. Here we are appending the _quantum circuit_ ``mcmt``  to the quantum circuit ``circuit``, yet the docstring for ``append`` states ``instruction: Operation | CircuitInstruction``. The code does work, since a ``QuantumCircuit`` defines a ``to_instruction`` method, which ``append`` tries to use when the instruction is not of type ``Operation``. Would it be slightly cleaner to change to ``circuit.append(mcmt.to_instruction(), ...)``, or should we change the typehint for the ``append`` method?

alexanderivrii (dismissed):
The fix looks straightforward, LGTM!

I had one tiny off-topic question.

qiskit/transpiler/passes/synthesis/hls_plugins.py
```diff
@@ -420,6 +420,7 @@
     C3XGate,
     C4XGate,
     PauliEvolutionGate,
+    MCMTGate,
     ModularAdderGate,
```
ShellyGarion, line 425: Are there other gates that might be missing here? what about `PermutationGate`, `MultiplierGate`  ?

test/python/circuit/library/test_mcmt.py
Cryoris: Updating the docs for `append` sounds good, since we use `append(QuantumCircuit)` all over the place 🤔

qiskit/transpiler/passes/synthesis/hls_plugins.py
Cryoris: `MultiplierGate` already has this check but indeed `PermutationGate` and `Clifford` don't -- I'll add them 🙂

ShellyGarion: Thanks for fixing it. This PR looks OK to me now.

test/python/circuit/library/test_mcmt.py
```diff
@@ -285,6 +285,19 @@ def test_gate_with_parameters_vchain(self):
...
+
+        transpiled = transpile(circuit, basis_gates=["u", "cx"])
+        self.assertTrue(Operator(transpiled).equiv(gate.control(1)))
+
```
alexanderivrii: Just one super-tiny question: why do we check for equivalence rather than equality?

Cryoris: Changed to equality in 0d26bc0e6 🙂

alexanderivrii (approved):
Thanks @Cryoris for also updating our included plugins so that they all follow the same convention. And you have taken the art of writing tests to a completely new level -- now they read like a fiction story (unfortunate names, intruder circuits, etc) 😄.

Cryoris:
Haha I'm hoping to provide some joy for people reviewing the code 😛

Cryoris:
@mergifyio backport stable/1.3

PR #13600, opened 2024-12-23
More conservative caching in the ``CommutationChecker``

### LINKED ISSUE #13570: Commutation Checker with PauliGate
### Environment

- **Qiskit version**:  1.0, 1.2 and 1.3
- **Python version**:  3.12.7
- **Operating system**: Linux

### What is happening?

The commutation checker always fails to commute a PauliGate with any other gate. I get this error message.
```python
return self.cc.commute(op1, qargs1, cargs1, op2, qargs2, cargs2, max_num_qubits)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
qiskit.exceptions.QiskitError: 'Unable to hash a non-float instruction parameter.'
``` 
I think it is because the parameter in a `PauliGate` is the `Pauli` operator that it implements rather than a `Parameter` or a `float`. I realized that `CommutatorChecker` did not have a unittest for this particular case, and also I have tried it for different python versions 1.0, 1.2 and 1.3. 1.0 and 1.2 work fine but 1.3 does not.

2 things have changed.
1. The way that parameters are hashed has been ported to rust (the parameter in `PauliGate` is a string so I don't see why we can't hash it).
2. Before if one of the gates was parameterized `commute` would return `False`. Now it does some more invoved logic.

```python
from qiskit.circuit.commutation_library import SessionCommutationChecker as scc
from qiskit.circuit.library.generalized_gates.pauli import PauliGate
from qiskit.circuit.library.standard_gates.rx import RXGate
from qiskit.circuit.library.standard_gates.x import XGate

pauli_gate = PauliGate("XX")
# pauli_gate = XGate()
rx_gate_theta = RXGate(Parameter("Theta"))
print(scc.commute(pauli_gate, [0, 1], [], rx_gate_theta, [0], []))
print(scc.commute(rx_gate_theta, [0], [], pauli_gate, [0, 1], [])) 
``` 

### What should happen?

For versions up to 1.2 we get `False` (which is not the correct answer). And for version 1.3 we get the error message.

### Any suggestions?

I think the best option would be to fix the hashing. If there is a deeper reason why that is not possible then we need to change the logic of the function.

Cryoris:
### Summary

Fixes #13570. 

This PR might require some discussion on what assumptions we have on custom `Instruction`s. However, if we cannot assume that an instruction's definition is fully specified by it's ``params``, then we cannot use the ``params`` as key in the commutation library. (We also sometimes put additional information outside of ``params``.)
This PR makes the commutation cache more conservative, by only additionally caching standard gate relations, where we know the instructions are defined by it's ``params``.

### Details and comments

The commutation checker currently caches commutations of any instruction with float-only ``params`` (with some hardcoded exceptions). This leads to problems if the instruction's definition depends on more than just the ``params``, e.g. we could have a custom gate defined as 
```python
class MyEvilRXGate(Gate):
    """A RX gate designed to annoy the caching mechanism (but a realistic gate nevertheless)."""

    def __init__(self, evil_input_not_in_param: float):
        """
        Args:
            evil_input_not_in_param: The RX rotation angle.
        """
        self.value = evil_input_not_in_param
        super().__init__("<evil laugh here>", 1, [])

    def _define(self):
        self.definition = QuantumCircuit(1)
        self.definition.rx(self.value, 0) 
```
Calling the commutation checker with an instance of this gate will then cache the first commutation, under the key `[]` (since the ``params`` are empty) and subsequent queries of the cache return the first commutation, regardless of the evil rotation angle. For example
```python

all_commuter = MyEvilRXGate(0)  # this will with anything
some_rx = MyEvilRXGate(1.6192)  # this should not commute with H

print(scc.commute(all_commuter, [0], [], HGate(), [0], []))  # True, as we expect
# this should be False, but queries the first occurence of MyEvilRXGate(0) and is True!
print(scc.commute(some_rx, [0], [], HGate(), [0], []))
```

Another solution could be to explicitly state the assumption that the gate must be fully specified by the ``params`` attribute, and we'd have to update some gates in the library. 

I ran the `test/benchmarks/utility_scale.py` benchmarks and could not see a regression on my laptop.

### DIFF
--- a/test/python/circuit/test_commutation_checker.py
+++ b/test/python/circuit/test_commutation_checker.py
@@ -27,6 +27,7 @@ from qiskit.circuit import (
     Parameter,
     QuantumRegister,
     Qubit,
+    QuantumCircuit,
 )
 from qiskit.circuit.commutation_library import SessionCommutationChecker as scc
 from qiskit.circuit.library import (
@@ -37,9 +38,11 @@ from qiskit.circuit.library import (
     CRYGate,
     CRZGate,
     CXGate,
+    CUGate,
     LinearFunction,
     MCXGate,
     Measure,
+    PauliGate,
     PhaseGate,
     Reset,
     RXGate,
@@ -80,6 +83,22 @@ class NewGateCX(Gate):
 
     def to_matrix(self):
         return np.array([[1, 0, 0, 0], [0, 0, 0, 1], [0, 0, 1, 0], [0, 1, 0, 0]], dtype=complex)
+
+
+class MyEvilRXGate(Gate):
+    """A RX gate designed to annoy the caching mechanism (but a realistic gate nevertheless)."""
+
+    def __init__(self, evil_input_not_in_param: float):
+        """
+        Args:
+            evil_input_not_in_param: The RX rotation angle.
+        """
+        self.value = evil_input_not_in_param
+        super().__init__("<evil laugh here>", 1, [])
+
+    def _define(self):
+        self.definition = QuantumCircuit(1)
+        self.definition.rx(self.value, 0)
 
 
 @ddt
@@ -137,7 +156,7 @@ class TestCommutationChecker(QiskitTestCase):
     def test_caching_positive_results(self):
         """Check that hashing positive results in commutativity checker works as expected."""
         scc.clear_cached_commutations()
-        self.assertTrue(scc.commute(ZGate(), [0], [], NewGateCX(), [0, 1], []))
+        self.assertTrue(scc.commute(ZGate(), [0], [], CUGate(1, 2, 3, 0), [0, 1], []))
         self.assertGreater(scc.num_cached_entries(), 0)
 
     def test_caching_lookup_with_non_overlapping_qubits(self):
@@ -150,16 +169,17 @@ class TestCommutationChecker(QiskitTestCase):
     def test_caching_store_and_lookup_with_non_overlapping_qubits(self):
         """Check that commutations storing and lookup with non-overlapping qubits works as expected."""
         scc_lenm = scc.num_cached_entries()
-        self.assertTrue(scc.commute(NewGateCX(), [0, 2], [], CXGate(), [0, 1], []))
-        self.assertFalse(scc.commute(NewGateCX(), [0, 1], [], CXGate(), [1, 2], []))
-        self.assertTrue(scc.commute(NewGateCX(), [1, 4], [], CXGate(), [1, 6], []))
-        self.assertFalse(scc.commute(NewGateCX(), [5, 3], [], CXGate(), [3, 1], []))
+        cx_like = CUGate(np.pi, 0, np.pi, 0)
+        self.assertTrue(scc.commute(cx_like, [0, 2], [], CXGate(), [0, 1], []))
+        self.assertFalse(scc.commute(cx_like, [0, 1], [], CXGate(), [1, 2], []))
+        self.assertTrue(scc.commute(cx_like, [1, 4], [], CXGate(), [1, 6], []))
+        self.assertFalse(scc.commute(cx_like, [5, 3], [], CXGate(), [3, 1], []))
         self.assertEqual(scc.num_cached_entries(), scc_lenm + 2)
 
     def test_caching_negative_results(self):
         """Check that hashing negative results in commutativity checker works as expected."""
         scc.clear_cached_commutations()
-        self.assertFalse(scc.commute(XGate(), [0], [], NewGateCX(), [0, 1], []))
+        self.assertFalse(scc.commute(XGate(), [0], [], CUGate(1, 2, 3, 0), [0, 1], []))
         self.assertGreater(scc.num_cached_entries(), 0)
 
     def test_caching_different_qubit_sets(self):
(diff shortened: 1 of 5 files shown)


alexanderivrii:
Thanks @Cryoris, I agree that it's better to not cache a commutation relationship rather than to cache a wrong one. Nevertheless, I thought we had a way both to check and to store that $RZ(\theta)(0)$ commutes with $CX(0, 1)$ for _every_ possible value of $\theta$, including the parametric case -- do I remember this correctly?

Cryoris:
Yes, we cover the parametric $RZ(\theta)$ case -- but only for the _Qiskit-native_ `RZGate`. This PR disables `params`-based caching for any non-native gate, since we cannot be certain that the gate instances only differ in the `params` attribute. 

We can also discuss this with the team, whether this is the right approach or we should rather make the assumption that gate instances do only differ by their `params` (which currently is not the case for some library gates). Personally, I would asusme that the commutation checker by far works on Qiskit-native gates so using this more conservative approach likely doesn't affect performance and is safer 🙂

test/python/circuit/test_commutation_checker.py
```diff
@@ -150,27 +169,29 @@ def test_caching_lookup_with_non_overlapping_qubits(self):
...
         scc_lenm = scc.num_cached_entries()
-        self.assertTrue(scc.commute(NewGateCX(), [0, 2], [], CXGate(), [0, 1], []))
-        self.assertFalse(scc.commute(NewGateCX(), [0, 1], [], CXGate(), [1, 2], []))
-        self.assertTrue(scc.commute(NewGateCX(), [1, 4], [], CXGate(), [1, 6], []))
-        self.assertFalse(scc.commute(NewGateCX(), [5, 3], [], CXGate(), [3, 1], []))
+        cx_like = CUGate(np.pi, 0, np.pi, 0)
+        self.assertTrue(scc.commute(cx_like, [0, 2], [], CXGate(), [0, 1], []))
+        self.assertFalse(scc.commute(cx_like, [0, 1], [], CXGate(), [1, 2], []))
+        self.assertTrue(scc.commute(cx_like, [1, 4], [], CXGate(), [1, 6], []))
+        self.assertFalse(scc.commute(cx_like, [5, 3], [], CXGate(), [3, 1], []))
```
alexanderivrii, line 176: Please tell me if I understand correctly. ``NewGateCX`` is not a standard gate, so with this PR we will no longer cache its commutation relations. And ``cx_like`` is a standard gate (a ``CUGate``). This is why it makes sense to update the tests to reason about ``cx_like`` instead of ``NewGateCX``, correct? However, the `scc.commute` method should still be able to handle ``NewGateCX``. So maybe it makes sense to leave a test that deals with custom gates. Do we still have such a test?

test/python/circuit/test_commutation_checker.py
```diff
@@ -82,6 +85,22 @@ def to_matrix(self):
...
+class MyEvilRXGate(Gate):
+    """A RX gate designed to annoy the caching mechanism (but a realistic gate nevertheless)."""
+
+    def __init__(self, evil_input_not_in_param: float):
+        """
+        Args:
+            evil_input_not_in_param: The RX rotation angle.
+        """
+        self.value = evil_input_not_in_param
+        super().__init__("<evil laugh here>", 1, [])
```
alexanderivrii, line 97: I love this. This is what I meant earlier by saying that you are taking the art of writing tests to a whole new level 😄.

test/python/circuit/test_commutation_checker.py
```diff
@@ -430,6 +453,25 @@ def test_rotation_mod_2pi(self, gate_cls):
...
+        self.assertTrue(scc.commute(all_commuter, [0], [], HGate(), [0], []))
+        self.assertFalse(scc.commute(some_rx, [0], [], HGate(), [0], []))
+
+    def test_nonfloat_param(self):
+        """Test commutation-checking on a gate that has non-float ``params``."""
+        pauli_gate = PauliGate("XX")
+        rx_gate_theta = RXGate(Parameter("Theta"))
+        self.assertTrue(scc.commute(pauli_gate, [0, 1], [], rx_gate_theta, [0], []))
+        self.assertTrue(scc.commute(rx_gate_theta, [0], [], pauli_gate, [0, 1], []))
+
```
alexanderivrii, line 485: How does this work, i.e. how do we infer that a parametric RXGate commutes with a Pauli-XX gate: as we can't use matrix multiplication for that, right? 

Could you also add a test for some other Pauli gate that does not commute with RX?

alexanderivrii (commented):
I think that caching/checking commutation relations for standard gates only makes a lot of sense. I had a few in-line questions (probably showing that I have not fully followed the recent updates to commutation checking). One other question, if I remember correctly, at some point @brandhsn has generated a database of commutation relations, are we still using it, and if so, should it be regenerated for standard gates only?

I agree that a more long-term plan might be to rethink the role of ``params`` when defining custom gates, but let's leave this as a possible followup.

test/python/circuit/test_commutation_checker.py
Cryoris: Correct to all of the above 👍🏻 

We don't have a test using `NewGateCX` concretely, only with the `EvilRXGate`, which should test the same thing. I can add a test with `NewGateCX` though just to be safe.

test/python/circuit/test_commutation_checker.py
Cryoris: RXGate get's reduced to XGate, so we can use the matrix multiplication to check for commutation with Pauli-XX. Note that if the angle is explicitly 0 (mod 2pi) then we don't reduce to X but to I.

alexanderivrii: Ah, I remember now, thanks for explaining that!

test/python/circuit/test_commutation_checker.py
alexanderivrii: Thanks, having an additional test for a benign custom gate makes sense.

Cryoris: Added in 44457f7 👍🏻

alexanderivrii (approved):
LGTM, thanks for the fix and updating the tests.

Cryoris:
@mergifyio backport stable/1.3

PR #13605, opened 2024-12-25
Improvements to `HighLevelSynthesis` transpiler pass

### LINKED ISSUE #13565: Move handling of `AnnotatedOperations` in `HighLevelSynthesis` to a plugin
### What should we add?

An ``AnnotatedOperation`` is a just-yet-another high-level object on a circuit, consisting of the base operation and a list of control, power and inverse modifiers. The handling of all the other high-level objects (e.g., Cliffords, Permutations, MCX gates, HalfAdder gates, etc.) is done using the plugin mechanism, and it makes sense to use the plugin mechanism for annotated operations as well. The idea was initially suggested by @jakelishman and recently discussed in depth with @Cryoris and @ShellyGarion.

There are several benefits of doing the above. First, it unifies and simplifies the ``HighLevelSynthesis`` transpiler pass, making it easier to port it to Rust. Second, it allows for introducing different implementation for (especially) control-annotated gates. In particular, this would disentangle the ``HighLevelSynthesis`` transpiler pass  and the mechanism for synthesizing a controlled-circuit, making it easier to improve the controlled-circuit synthesis code in the future.

On the other hand, the current handling of annotated operations inside the ``HighLevelSynthesis`` requires a lot more of additional information than available to other plugins. This is because we first use the full power of ``HighLevelSynthesis`` to recursively synthesize the base operation (which itself may have a complex definition containing other high-level and annotated operations), and in particular we need access to ``qubit_tracker``, ``qubit_context`` and some other internal ``HighLevelSynthesis`` data. So this is also a great time to completely rethink the high-level-synthesis plugin interface (with the extra complication that whatever we do should be backward compatible).

Historically, the `run` method of a "regular" high-level-synthesis plugin accepted the ``high-level-object`` to synthesize, the ``coupling_map`` and then later also ``target`` (it probably makes sense to deprecate ``coupling_map`` as done elsewhere), and the list of ``qubits`` over which the high-level-object is defined, with the convention that ``qubits=None`` indicates the synthesis runs pre-layout, and while the coupling map is available and can guide the plugin to which synthesis algorithm to choose, the physical locations of the qubits are still not chosen and so we still don't know which pairs of qubits are connected and which are not. In addition each plugin accepts an arbitrary dictionary of plugin-specific options, and we started to use this to pass the number of clean and dirty qubits available to the plugin. However, as mentioned above, synthesizing annotated operations also requires the qubits of the operation, the qubit context (the mapping between the current DAG's qubits and the qubits of the original circuit), the qubit tracker (which qubits in the original circuit and clean or dirty) and other data. I am now fairly convinced that even "regular" plugins may wish to have access to this information. In addition, it should also be a plugin's responsibility to update the lists of clean and dirty ancilla qubits after the plugin is run. 

I am planning to open a draft PR shortly, using it as a way to discuss the desired interface.

alexanderivrii:
### Summary

This PR reimplements the internals of the `HighLevelSynthesis` transpiler pass, still in Python. The pass is now arguably much simpler and cleaner. 

Features:
- It moves the handling of annotated operations into a separate plugin, addressing #13565.
- If the `HighLevelSynthesis` transpiler pass needs to do something nontrivial, then except for the top-level `circuit_to_dag` and `dag_to_circuit` conversions, all the internal recursive functions work only with `QuantumCircuit`s, greatly simplifying the API and avoiding constant back-and-forth conversions.
- It will be now much easier to port the main functionality to Rust.

The PR is based on multiple discussions with @Cryoris.

Since most of the changes are internal implementation details, the only thing I am emphasizing in the release notes is the plugin interface for annotated operations.

### Details and comments

One particular implementation detail that has changed significantly is how we reason about ancilla qubits in combination with the highly recursive nature of the pass. Take for example a quantum circuit containing a custom gate whose definition involves another custom gate whose definition involves an MCX gate that we can synthesize more efficiently using the auxiliary qubits from the global circuit. This was already possible, and to support this we have introduced the ``QubitContext`` class that represents a mapping from the _current_ (internal) circuit's qubits to the _global_ qubits (i.e. qubits of the original circuit). If synthesizing an operation required more auxiliary variables, these mappings were extended to include these new variables. With this PR, all the internal functions directly work with the _global qubits_ instead of these mappings. For instance, the internal ``_synthesize_operation`` function now receives an ``Operation`` and a list of global qubits it's defined on and returns.a ``QuantumCircuit`` and a possibly extended list of global qubits on which the new quantum circuit is defined. This makes the booktracking significantly simpler. (In particular, the internal `QubitContext` class will probably be deleted, or repurposed to speed up certain internal computations).

A lot of this refactoring was needed to turn handling of annotated operations into a plugin. The way this was done before (and is pretty much the same way it's still done now) is first to synthesize the base operation of such an annotated operation using the full power of recursive synthesis, and then apply control, power and inverse modifiers to the obtained circuit. However, synthesizing the base operation requires the internal data and options of the `HighLevelSynthesis` pass itself, the qubits on which the base operation is defined, and the qubit tracker -- the last two are needed to allow using ancilla qubits. This motivated making the function ``_run``, ``_synthesize_operation``, etc. as global functions rather than belonging to a specific ``HighLevelSynthesis`` object (alternatively we could make these staticfunctions). In addition, this flattening also simplifies moving these function to Rust in a followup.

Alternatively, above we could have instantiate a new `HighLevelSynthesis` pass from within the current pass, however the initialization (i.e. `HighLevelSynthesis__init__` ) incurs a certain overhead in the (getting the lists of available plugins, etc.) . Instead, all the immutable entries are stored in the new `HLSData` struct which is passed from one recursive function to another (and also to the plugin for annotated operations using the `plugin_args` dict).

### Follow-up plans:

There are several follow-up items that will be handled in follow-up PRs. 

- Port the main functionality of the pass to Rust 
- Rethink the input/output API for plugins (while preserving backward-compatibility). Ideally, I would now like each plugin method to take in a triple (operation, global qubits on which this operation is defined, state of the qubits) and return either `None` or the synthesized result in the form of another triple (synthesized circuit, possibly extended list of global qubits on which this circuit is defined, possibly updated state of the qubits). Right now all the plugins only know how many clean/dirty ancilla qubits are available, but not what these qubits actually are, and only return the synthesized quantum circuit, without specifying ancilla qubits actually used. This might be important if we want to _resynthesize_ a quantum circuit using ancilla qubits, when the connectivity map is provided, and the choice of an actual ancilla qubit actually matters. 
- Somewhere along the way, we should rethink the API for the qubit tracking mechanism. Right now I am trying to avoid copying it when unnecessary (i.e. changing the state of qubits in-place), but maybe an API where the function both accepts and returns a qubit tracker might be cleaner (and in Rust, we can possibly avoid copying, i.e. just moving).


test/python/transpiler/test_high_level_synthesis.py
```diff
@@ -1106,10 +1106,8 @@ def test_control_clifford(self):
         cliff = Clifford(qc)
         circuit = QuantumCircuit(4)
         circuit.append(AnnotatedOperation(cliff, ControlModifier(2)), [0, 1, 2, 3])
-        transpiled_circuit = HighLevelSynthesis()(circuit)
-        expected_circuit = QuantumCircuit(4)
-        expected_circuit.append(cliff.to_instruction().control(2), [0, 1, 2, 3])
-        self.assertEqual(transpiled_circuit, expected_circuit)
+        transpiled_circuit = HighLevelSynthesis(basis_gates=["cx", "u"])(circuit)
+        self.assertEqual(transpiled_circuit.count_ops().keys(), {"cx", "u"})
```
alexanderivrii, line 1147: This actually fixes a problem: previously the circuit was not fully synthesized, now it is.

test/python/circuit/library/test_multipliers.py
```diff
@@ -143,7 +142,7 @@ def test_plugins(self):
         """Test setting the HLS plugins for the modular adder."""
 
         # all gates with the plugins we check, including an expected operation
-        plugins = [("cumulative_h18", "ccircuit-.*"), ("qft_r17", "qft")]
+        plugins = [("cumulative_h18", "ch"), ("qft_r17", "mcphase")]
```
alexanderivrii, line 145: When simplifying the code for handling controlled-annotated operations (in particular wrapping a single operation into a circuit so that we only need to consider circuits and not both circuits and operations), I have changed the code on how a `control` is applied to a circuit: instead of first converting the circuit into a gate and applying control to this gate, we create a new quantum circuit with the controlled version of each gate. This makes sure that we do not deteriorate the results in the case of a single iteration (and may actually improve the results in the case of a circuit). Hmm, now that I am writing this, they may be a slight problem when controlling a circuit that has a global phase, I will need to fix this.

UPDATE: that was indeed a bug, which is fixed in [c5f0f0e](https://github.com/Qiskit/qiskit/pull/13605/commits/c5f0f0e21a46184f8157cb76d48705823e410de1).

ShellyGarion:
Could you please add some examples of how this improved HLS should be used now?
(as well as in the docs and release notes)

alexanderivrii:
@ShellyGarion, most of the changes in this PR involve an updated implementation of the HLS pass, with no impact on HLS documentation or on how HLS is used. The only arguably new functionality is that objects of type ``AnnotatedOperation`` are now handled via a dedicated plugin, which is already mentioned in the release notes.

qiskit/transpiler/passes/synthesis/high_level_synthesis.py
```diff
@@ -135,6 +125,34 @@ def set_methods(self, hls_name, hls_methods):
         self.methods[hls_name] = hls_methods
 
 
+class HLSData:
```
Cryoris, line 130: You could use a `dataclass` here 🙂

qiskit/transpiler/passes/synthesis/high_level_synthesis.py
```diff
@@ -270,585 +291,530 @@ def run(self, dag: DAGCircuit) -> DAGCircuit:
...
+
+        # op's qubits as viewed globally
+        op_qubits = [input_qubits[input_circuit.find_bit(q).index] for q in inst.qubits]
+
+        # Start by handling special operations.
+        # In the future, we can also consider other possible optimizations, e.g.:
+        #   - improved qubit tracking after a SWAP gate
+        #   - automatically simplify control gates with control at 0.
+        if op.name in ["id", "delay", "barrier"]:
+            output_circuit.append(op, inst.qubits, inst.clbits)
```
Cryoris, line 402: We might want to consider using `_append` where possible, since these operations are coming from a valid circuit so we can skip the checks

qiskit/transpiler/passes/synthesis/hls_plugins.py
```diff
@@ -1624,3 +1633,145 @@ def run(self, high_level_object, coupling_map=None, target=None, qubits=None, **
...
+
+        operation = high_level_object
+        modifiers = high_level_object.modifiers
+
+        # The plugin needs additional information that is not yet passed via the run's method
+        # arguments: namely high-level-synthesis data and options, the global qubits over which
+        # the operation is defined, and the initial state of each global qubit.
+        tracker = options.get("qubit_tracker", None)
+        data = options.get("hls_data")
+        input_qubits = options.get("input_qubits")
```
Cryoris: It might be worth to fail with a nice message for the users here 🤔

qiskit/transpiler/passes/synthesis/hls_plugins.py
```diff
@@ -1624,3 +1633,145 @@ def run(self, high_level_object, coupling_map=None, target=None, qubits=None, **
...
+
+            # The base operation cannot use control qubits as auxiliary qubits.
+            # In addition, when we have power or inverse modifiers, we need to set all of
+            # the operation's qubits to dirty. Note that synthesizing the base operation we
+            # can use additional auxiliary qubits, however they would always be returned to
+            # their previous state, so clean qubits remain clean after each for- or while- loop.
+            annotated_tracker = tracker.copy()
+            annotated_tracker.disable(input_qubits[:num_ctrl])  # do not access control qubits
+            if total_power != 0 or is_inverted:
+                annotated_tracker.set_dirty(input_qubits)
```
Cryoris, line 1722: Why do we need to set all of them to dirty if we have a power or inverse? If we have the inverse of a controlled gate, then the control qubits still don't need to change the state or do they?

qiskit/transpiler/passes/synthesis/hls_plugins.py
```diff
@@ -1624,3 +1633,145 @@ def run(self, high_level_object, coupling_map=None, target=None, qubits=None, **
...
+            # return the actual ancilla qubits used.
+            synthesized_base_op, _ = _synthesize_operation(
+                operation.base_op, input_qubits[num_ctrl:], data, annotated_tracker
+            )
+
+            # The base operation does not need to be synthesized.
+            # For simplicity, we wrap the instruction into a circuit. Note that
+            # this should not deteriorate the quality of the result.
+            if synthesized_base_op is None:
+                synthesized_base_op = _instruction_to_circuit(operation.base_op)
```
Cryoris, line 1738: I'm wondering whether we can keep this a `Gate | Instruction`, all methods inside `apply_annotation` should also work then no? One less conversion 🙂

Cryoris (commented):
Some very brief comments, I'm looking forward to the discussion!

qiskit/transpiler/passes/synthesis/hls_plugins.py
alexanderivrii: For the inverse modifier, currently we (i) synthesize the circuit for the "base operator", and (ii) call ``circuit.inverse()`` to invert the result. Trying to think of an example of what may go wrong, suppose that we are synthesizing `CustomGate().inverse(annotated=True)`, where `CustomGate` has the definition ``cx(0, 1) h(0) h(1)``. The expected result should then be the circuit ``h(1) h(0) cx(0, 1)``, even if all of the qubits are initially at $|0\rangle$, because the ``h``-gates bring the qubits to non- $|0\rangle$ states. Hence in step (i) we cannot assume that the qubits are at $|0\rangle$, potentially removing the leading ``cx``-gate, because in step (ii) we will need to reverse the order of operations. I might very well be missing some optimization opportunities, trying to error only on the safe side.

qiskit/transpiler/passes/synthesis/hls_plugins.py
alexanderivrii: The problem that some of the methods in `apply_annotations` will return you a circuit (and not just an instruction), so we already have to support quantum circuits. In addition, the final output from the plugin should be a quantum circuit as well. Previously we supported `Instruction | QuantumCircuit` and the code was significantly uglier. I would be happy to discuss it further,

qiskit/transpiler/passes/synthesis/hls_plugins.py
alexanderivrii: I am still not settled what the plugin API should look like and whether we can avoid passing all this information or pass this in some other cleaner way. So I would like to keep it internal, at least for now.

qiskit/transpiler/passes/synthesis/high_level_synthesis.py
```diff
@@ -270,585 +291,530 @@ def run(self, dag: DAGCircuit) -> DAGCircuit:
...
-                        context=inner_context,
-                        use_ancillas=False,
-                        top_level=False,
-                    ),
-                    node.op,
-                )
+        # If the synthesized circuit uses (auxiliary) global qubits that are not in the output circuit,
+        # we add these qubits to the output circuit.
+        for q in synthesized_circuit_qubits:
+            if q not in global_to_local:
```
Cryoris, line 465: I'm a bit skeptical about using `Qubit`s here. Are all qubits in `synthesized_circuit_qubits` guaranteed to be part of the `input_circuit` or can they be generated by a synthesis method and be other instances?

alexanderivrii: Let me explain what's going on. We are recursively running HLS on some "internal" circuit. For example, say that the global circuit `C` is over 10 qubits, and contains a custom gate `G` over the 5 (global) qubits 0, 2, 4, 6, 8 with a given definition circuit `Gdef` for `G`. When synthesizing `G`, the HLS pass will recursively process `Gdef` -- this is what I mean by an internal circuit. So, say HLS is now running on the circuit `Gdef`. The qubits of `Gdef` are labeled consecutively starting from 0 (i.e. 0, 1, 2, 3, 4), while the additional parameter `qubits` (passed to and returned from every function) specifies the corresponding global indices (i.e. 0, 2, 4, 6, 8). Let's say that the circuit `Gdef` contains an MCX gate `H` over 4 of its qubits (let's say 0, 1, 2, 3 as viewed from within `Gdef`, correponding to global qubits 0, 2, 4, 6), and suppose that the synthesis algorithm for this MCX gate `H` produces a circuit `H_synthesized` that additionally uses the global qubits `8` and `1` as ancilla qubits (here, the global qubit 8 corresponds to a qubit that is already in `Gdef` but 1 is outside). In this case, the result of synthesizing `Gdef` will be a circuit `Gdef_synthesized` over 6 qubits with internal indices 0, 1, 2, 3, 4, 5 and global indices 0, 2, 4, 6, 8, 1. This is what the selected lines are doing, extending `Gdef_synthesized`'s map from global to internal indices to include `6 -> 1` (and we need this in order to compute the indices within `Gdef_synthesized` of the gates of the `H_synthesized` circuit).

The short answer to the question is that `synthesized_circuit_qubits` do not need to be a part of `input_circuit` -- and this is precisely where we handle it.

This is actually a huge simplification from how it used to be before. 😅

qiskit/transpiler/passes/synthesis/high_level_synthesis.py
alexanderivrii: Did this in [e92ac9d](https://github.com/Qiskit/qiskit/pull/13605/commits/e92ac9d93da845470a5e778722518c683b7f2ff3). Also removed the argument ``qubits_initially_zero`` per your offline comments.

qiskit/transpiler/passes/synthesis/hls_plugins.py
alexanderivrii: Update: I am adding an error message as a sanity check, however the ``data`` and ``input_qubits`` fields should be set by the ``HighLevelSynthesis`` pass itself, which is out of the users' control.

mtreinish:
I'm going to close this PR as it's superseded by https://github.com/Qiskit/qiskit/pull/13813 which ports the equivalent logic to rust. We can still refer to this PR for reference of the evolution of the logic in Python space, but there isn't a reason to merge this if it'll immediately be replaced by a rust implementation.

PR #13606, opened 2024-12-26
Fix `random_clifford`

### LINKED ISSUE #13590: `random_clifford` does not produce random Cliffords on 2 qubits
### Environment

- **Qiskit version**: 1.3
- **Python version**: 
- **Operating system**: 

`random_clifford` does not produce random Cliffords on 2 qubits. 

### What is happening?

`random_clifford(2)` produces only Cliffords with 0,1,2 CX gates and not 3 CX gates.

```
for seed in range(100):
    cliff = random_clifford(2, seed)
    qc = cliff.to_circuit()
    cnt = qc.count_ops().get('cx', 0)
    print (cnt)
```

### What should happen?

`random_clifford(2)`  should also produce Cliffords with 3 CX's.

### Any suggestions?

This is probably following this PR: https://github.com/Qiskit/qiskit/pull/12695

alexanderivrii:
### Summary

Fixes #13590.

### Details and comments

The bug was introduced when porting `random_clifford` to Rust in https://github.com/Qiskit/qiskit/pull/12695.

I have added tests that ensure that for 2-qubits Cliffords, `random_clifford` produces Cliffords with all possible symplectic matrices (there should be 720) and all possible CX counts (should be {0, 1, 2, 3}). @ShellyGarion, can you think of other tests that we may wish to add?

### DIFF
--- a/crates/accelerate/src/synthesis/clifford/random_clifford.rs
+++ b/crates/accelerate/src/synthesis/clifford/random_clifford.rs
@@ -125,9 +125,7 @@ pub fn random_clifford_tableau_inner(num_qubits: usize, seed: Option<u64>) -> Ar
 
     // Compute the full stabilizer tableau
 
-    // The code below is identical to the Python implementation, but is based on the original
-    // code in the paper.
-
+    // The code below is based on the original code in the referenced paper.
     let mut table = Array2::from_elem((2 * num_qubits, 2 * num_qubits), false);
 
     // Apply qubit permutation
--- a/crates/accelerate/src/synthesis/clifford/random_clifford.rs
+++ b/crates/accelerate/src/synthesis/clifford/random_clifford.rs
@@ -132,10 +132,10 @@ pub fn random_clifford_tableau_inner(num_qubits: usize, seed: Option<u64>) -> Ar
 
     // Apply qubit permutation
     for i in 0..num_qubits {
-        replace_row_inner(table.view_mut(), i, table2.slice(s![i, ..]));
+        replace_row_inner(table.view_mut(), i, table2.slice(s![perm[i], ..]));
         replace_row_inner(
             table.view_mut(),
-            perm[i] + num_qubits,
+            i + num_qubits,
             table2.slice(s![perm[i] + num_qubits, ..]),
         );
     }
--- a/test/python/quantum_info/operators/test_random.py
+++ b/test/python/quantum_info/operators/test_random.py
@@ -181,7 +181,7 @@ class TestRandomClifford(QiskitTestCase):
         self.assertEqual(value1, value2)
 
     def test_not_global_seed(self):
-        """Test fixing random_hermitian seed is locally scoped."""
+        """Test fixing random_seed is locally scoped."""
         seed = 314159
         test_cases = 100
         random_hermitian(2, seed=seed)
@@ -189,6 +189,26 @@ class TestRandomClifford(QiskitTestCase):
         random_hermitian(2, seed=seed)
         rng_after = np.random.randint(1000, size=test_cases)
         self.assertFalse(np.all(rng_before == rng_after))
+
+    def test_cliffords_2q(self):
+        """Test that we get all 2-qubit Cliffords (actually symplectic
+        matrices) with sufficiently many trials.
+        """
+        seen = set()
+        for seed in range(10000):
+            cliff = random_clifford(2, seed)
+            seen.add(cliff.symplectic_matrix.tobytes())
+        self.assertEqual(len(seen), 720)
+
+    def test_clifford_2q_decompositions(self):
+        """Test that we get all possible CX-counts for 2q-random cliffords
+        with sufficiently many trials.
+        """
+        seen = set()
+        for seed in range(100):
+            cliff = random_clifford(2, seed)
+            seen.add(cliff.to_circuit().count_ops().get("cx", 0))
+        self.assertEqual(seen, {0, 1, 2, 3})
 
 
 @ddt
--- a/test/python/quantum_info/operators/test_random.py
+++ b/test/python/quantum_info/operators/test_random.py
@@ -210,6 +210,16 @@ class TestRandomClifford(QiskitTestCase):
             seen.add(cliff.to_circuit().count_ops().get("cx", 0))
         self.assertEqual(seen, {0, 1, 2, 3})
 
+    def test_clifford_3q_decompositions(self):
+        """Test that we get all possible CX-counts for 3q-random cliffords
+        with sufficiently many trials.
+        """
+        seen = set()
+        for seed in range(10000):
+            cliff = random_clifford(3, seed)
+            seen.add(cliff.to_circuit().count_ops().get("cx", 0))
+        self.assertEqual(seen, {0, 1, 2, 3, 4, 5, 6})
+
 
 @ddt
 class TestRandomPauliList(QiskitTestCase):
--- a/test/python/quantum_info/operators/test_random.py
+++ b/test/python/quantum_info/operators/test_random.py
@@ -181,7 +181,7 @@ class TestRandomClifford(QiskitTestCase):
         self.assertEqual(value1, value2)
 
     def test_not_global_seed(self):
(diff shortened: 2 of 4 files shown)


ShellyGarion (commented):
Thanks @alexanderivrii for fixing this bug!
The tests for 2-qubit Cliffords look good to me.
I wonder if we can also test larger Cliffrods, say with 3 and 4 qubits. We can't check that we generate the entire group (it's just too large), but perhaps we can check that they at least produce distinct CX counts (according to their popularity)?

See this table:

<html xmlns:o="urn:schemas-microsoft-com:office:office"
xmlns:w="urn:schemas-microsoft-com:office:word"
xmlns:m="http://schemas.microsoft.com/office/2004/12/omml"
xmlns="http://www.w3.org/TR/REC-html40">

<head>

<meta name=ProgId content=Word.Document>
<meta name=Generator content="Microsoft Word 15">
<meta name=Originator content="Microsoft Word 15">
<link rel=File-List
href="file:///C:/Users/SHELLY~1/AppData/Local/Temp/msohtmlclip1/01/clip_filelist.xml">

<link rel=themeData
href="file:///C:/Users/SHELLY~1/AppData/Local/Temp/msohtmlclip1/01/clip_themedata.thmx">
<link rel=colorSchemeMapping
href="file:///C:/Users/SHELLY~1/AppData/Local/Temp/msohtmlclip1/01/clip_colorschememapping.xml">

<style>

</style>

</head>

<body lang=en-IL style='tab-interval:36.0pt;word-wrap:break-word'>



Num of CNOTS | 1 qubit | 2 qubits | 3 qubits | 4 qubits
-- | -- | -- | -- | --
Paulis | 4 | 16 | 64 | 256
0 | 6 | 36=6^2 | 216=6^3 | 1296=6^4
1 | - | 324=9*36 | 27*216 | 54*1296
2 | - | 324=9*36 | 432*216 | 1917*1296
3 | - | 36 | 2784*216 | 47904*1296
4 | - | - | 3042*216 | 849438*1296
5 | - | - | 432*216 | 8192583*1296
6 | - | - | 2*216 | 22623581*1296
7 | - | - | - | 4809996*1296
8 | - | - | - | 31320*1296
9 | - | - | - | 6*1296
Total | 24=4x6 | 11,520=   16x36x20 | 6720x216x64=   92,897,280 | 36,556,800x1296x256   =12,128,668,876,800




</body>

</html>

synthesis0x42 (approved):

test/python/quantum_info/operators/test_random.py
```diff
@@ -190,6 +190,26 @@ def test_not_global_seed(self):
         rng_after = np.random.randint(1000, size=test_cases)
         self.assertFalse(np.all(rng_before == rng_after))
 
+    def test_cliffords_2q(self):
+        """Test that we get all 2-qubit Cliffords (actually symplectic
+        matrices) with sufficiently many trials.
+        """
+        seen = set()
+        for seed in range(10000):
```
gadial, line 198: Are 10000 trials negligible in terms of run time in this case? We want to avoid unit tests taking too much time. Perhaps using one-time preprocessing it's possible to find a small set of seeds such that a small number of iterations on each one is enough to guarantee full coverage?

alexanderivrii:
Yes, this is fast -- the full test takes 0.17 seconds on my laptop. Regarding your suggestion, we probably don't want to update the test each time that we slightly change the algorithm. Of course, there is a tiny chance that after a certain hypothetical change even 10,000 random Cliffords would not be enough to hit all of 720 symplectic matrices, however for a true random Clifford generation algorithm this chance would be bounded above by $720 * (719/720)^{10000}$, which is 0.067%. In other words, if the test does fail after a future hypothetical change, it would most probably point to a bug in the code.

test/python/quantum_info/operators/test_random.py
alexanderivrii: Yes, this is fast -- the full test takes 0.17 seconds on my laptop. Regarding your suggestion, we probably don't want to update the test each time that we slightly change the algorithm. Of course, there is a tiny chance that after a certain hypothetical change even 10,000 random Cliffords would not be enough to hit all of 720 symplectic matrices, however for a true random Clifford generation algorithm this chance would be bounded above by $720 * (719/720)^{10000}$, which is 0.067%. In other words, if the test does fail after a future hypothetical change, it would most probably point to a bug in the code.

crates/accelerate/src/synthesis/clifford/random_clifford.rs
```diff
@@ -132,10 +132,10 @@ pub fn random_clifford_tableau_inner(num_qubits: usize, seed: Option<u64>) -> Ar
 
     // Apply qubit permutation
     for i in 0..num_qubits {
-        replace_row_inner(table.view_mut(), i, table2.slice(s![i, ..]));
+        replace_row_inner(table.view_mut(), i, table2.slice(s![perm[i], ..]));
```
Cryoris, line 133: Maybe it's obvious and my brain is still on holiday, but why does this swapping of `perm[i]` and `i` fix the issue?

crates/accelerate/src/synthesis/clifford/random_clifford.rs
```diff
@@ -132,10 +132,10 @@ pub fn random_clifford_tableau_inner(num_qubits: usize, seed: Option<u64>) -> Ar
```
Cryoris, line 130: I can't comment there but above is a comment saying that "this code is identical to the Python implementation" -- we could probably remove that since the Python code is long gone (it's not related to this PR but maybe we can take the chance to clean it up 🙂)

crates/accelerate/src/synthesis/clifford/random_clifford.rs
alexanderivrii: This is because with the _correct_ algorithm we should permute the rows of `table` using the permutation `perm` (see the last page https://arxiv.org/pdf/2003.09412).

crates/accelerate/src/synthesis/clifford/random_clifford.rs
alexanderivrii: Sure, it makes sense to improve the comment (done in [8954d20](https://github.com/Qiskit/qiskit/pull/13606/commits/8954d20f361610edef4b86d3972ecda41777390b)), especially that despite porting the code from Python to Rust literally line by line, I have managed to introduce an error.

crates/accelerate/src/synthesis/clifford/random_clifford.rs
Cryoris: Ah great, thanks for the link 🙂

alexanderivrii:
The failing test seems unrelated to this PR: ``qiskit_neko.tests.machine_learning.test_neural_network_classifier.TestNeuralNetworkClassifierOnPrimitives.test_neural_network_classifier_2_aer[aer,machine_learning,terra]``

alexanderivrii:
@ShellyGarion, based on your suggestion, I have added a test (see [97b3cc9](https://github.com/Qiskit/qiskit/pull/13606/commits/97b3cc93c468324f8d2fcd14a980ab25f3c4ed75)) that checks that we also get all possible 2-qubit depths for random 3-qubit cliffords, also note that in this case for decomposing Cliffords we use the optimal Bravyj-Maslov algorithm, so the CX-counts are exact. The test considers 10,000 random Cliffords and is still reasonably fast, taking 1s on my laptop. 

However, for 4-qubit Cliffords, the chosen decomposition method is only heuristic and we don't necessarily get the exact counts. In addition, it requires significantly more tests (and more time) to hit the low-popularity counts. I have locally verified that we seem to get the approximately expected counts, but for the reasons above I don't think we should add an explicit test for that.

ShellyGarion (approved):
LGTM. Thanks @alexanderivrii for fixing this bug!

wshanks:
It looks like this missed Qiskit 1.3.2. Should it still be backported to 1.3?

jakelishman:
Yeah, looks like it didn't get tagged for backport despite the milestone. I'll shift to hopefully 1.3.3.

PR #13612, opened 2025-01-06
Add option ``collect_from_back`` to ``CollectMultiQBlocks``

alexanderivrii:
### Summary

This PR adds a new option ``collect_from_back`` to the transpiler pass ``CollectMultiQBlocks``, leading to an increased flexibility of the pass. As an example, on the following circuit
```
             ┌───┐
        q_0: ┤ H ├──■────■────■───────
             └───┘┌─┴─┐  │    │
        q_1: ─────┤ X ├──┼────┼───────
                  └───┘┌─┴─┐  │
        q_2: ──────────┤ X ├──┼───────
                       └───┘┌─┴─┐┌───┐
        q_3: ───────────────┤ X ├┤ H ├
                            └───┘└───┘
```
running the pass with ``max_block_size=3`` and ``collect_from_back=False`` (default), collects the first three gates (the H-gate and the two following CX-gates) into the first block, and the two remaining gates into the second block, while running the pass with ``max_block_size=3`` and ``collect_from_back=True`` (new), collects the last three gates into the second block, and the remaining two gates into the first block.

### DIFF
--- a/test/python/transpiler/test_collect_multiq_blocks.py
+++ b/test/python/transpiler/test_collect_multiq_blocks.py
@@ -319,8 +319,8 @@ class TestCollect2qBlocks(QiskitTestCase):
 
         block_list = pass_manager.property_set["block_list"]
         self.assertEqual(len(block_list), 2)
-        self.assertEq(len(block_list[0]), 3)
-        self.assertEq(len(block_list[1]), 2)
+        self.assertEqual(len(block_list[0]), 3)
+        self.assertEqual(len(block_list[1]), 2)
 
         # When collecting blocks of size-3 using the opposite direction,
         # the first block should contain the H-gate and a single CX-gate,
--- a/test/python/transpiler/test_collect_multiq_blocks.py
+++ b/test/python/transpiler/test_collect_multiq_blocks.py
@@ -310,29 +310,25 @@ class TestCollect2qBlocks(QiskitTestCase):
         qc.cx(0, 3)
         qc.h(3)
 
+        dag = circuit_to_dag(qc)
+        # For the circuit above, the topological order is unique
+        topo_ops = list(dag.topological_op_nodes())
+
         # When collecting blocks of size-3 using the default direction,
         # the first block should contain the H-gate and two CX-gates,
         # and the second block should contain a single CX-gate and an H-gate.
-        pass_manager = PassManager()
-        pass_manager.append(CollectMultiQBlocks(max_block_size=3, collect_from_back=False))
-        pass_manager.run(qc)
-
-        block_list = pass_manager.property_set["block_list"]
-        self.assertEqual(len(block_list), 2)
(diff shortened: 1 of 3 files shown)


test/python/transpiler/test_collect_multiq_blocks.py
```diff
@@ -290,6 +290,50 @@ def test_larger_blocks(self):
...
+        # the first block should contain the H-gate and a single CX-gate,
+        # and the second block should contain two CX-gates and an H-gate.
+        pass_manager = PassManager()
+        pass_manager.append(CollectMultiQBlocks(max_block_size=3, collect_from_back=True))
+        pass_manager.run(qc)
+
+        block_list = pass_manager.property_set["block_list"]
+        self.assertEqual(len(block_list), 2)
+        self.assertEqual(len(block_list[0]), 2)
+        self.assertEqual(len(block_list[1]), 3)
```
jakelishman: Is there any assertion we can add that will make sure we've collected the correct blocks, not just the right lengths?

jakelishman (commented):
This looks like a straightforward new feature and implementation, thanks.  One minor testing question only.

test/python/transpiler/test_collect_multiq_blocks.py
alexanderivrii: That's a good idea, done in [9bc44fb](https://github.com/Qiskit/qiskit/pull/13612/commits/9bc44fba05afde65e2bba88c1dc95535ced34748).

test/python/transpiler/test_collect_multiq_blocks.py
```diff
@@ -310,29 +310,25 @@ def test_collect_from_back(self):
...
-        pass_manager.run(qc)
-
-        block_list = pass_manager.property_set["block_list"]
-        self.assertEqual(len(block_list), 2)
-        self.assertEqual(len(block_list[0]), 3)
-        self.assertEqual(len(block_list[1]), 2)
+        pass_ = CollectMultiQBlocks(max_block_size=3, collect_from_back=False)
+        pass_.run(dag)
+        expected_blocks = [[topo_ops[0], topo_ops[1], topo_ops[2]], [topo_ops[3], topo_ops[4]]]
+        self.assertTrue(pass_.property_set["block_list"], expected_blocks)
```
jakelishman: Presumably you mean `self.assertEqual` (and in the other place)?  This is asserting non-emptiness of the `pass_.property_set["block_list"]`, with a "message" field of `expected_blocks`.

alexanderivrii: Oops, of course! Fixed in [65a1aba](https://github.com/Qiskit/qiskit/pull/13612/commits/65a1abad1e1a23509eca6612e683666e871c4d23).

jakelishman (approved):
Cool, thanks!

PR #13616, opened 2025-01-06
Remove gif

### LINKED ISSUE #2533: Deal with large and inaccessible Qiskit GIF
See https://github.com/Qiskit/documentation/blob/main/public/images/api/qiskit/depth.gif and its usage:

https://github.com/Qiskit/documentation/blob/be6687be304ffc3262c6c21bb6117d21531821f6/docs/api/qiskit/qiskit.circuit.QuantumCircuit.mdx?plain=1#L3490-L3500

Issues with GIFs:
1. GIFs violate [accessibility](https://accessibleweb.com/question-answer/animated-gifs-okay/) because of flashing automatically and not having a way to pause them
2. The GIF is 2.7mb, which makes the page slower to download for users, even with the lazy loading Next.js uses. It also bloats our Git repository size in qiskit/documentation

Ideas for workarounds:

1. Replace it with a link to our 1 minute video "What is circuit depth": https://www.youtube.com/watch?v=7AVIc7SkX3M
2. Use text-based only explanation
3. Remove it entirely

Migrating the file from GIF to AVIF only reduced the image from 2.7MB to 2.3MB, and it corrupted the GIF.

Note that this GIF is in our historical API docs. We should probably rewrite them to not have the image so that those pages load quicker & our repo size is smaller.

--

Response from Jake L in Slack, Dec 27

> I tried to kill it once (for one, I disagree that it’s anything to do with tetris!), but there was some sentiment from within the qiskit-compiler team that it had helped with the understanding.
>
> Fwiw, I feel like the entire “simple circuit metrics” section could do with a rework.  It’s generally seemed to me to be targeted at a (now 6+ years ago) QC theorist who doesn’t know how to program, which probably was a core demographic at the time when Qiskit was in its infancy, but it’s not the best pedagogy now.  (Fwiw, that section was mostly unaltered when the rest of the QuantumCircuit page was rewritten, because it was already good enough compared to the complete absence of everything else.)
>
> That is to say, I’m strongly in favour of removing the GIF, and I feel like the description of “circuit depth” should have a text overview in the QuantumCircuit.depth and DAGCircuit.depth docstrings.  If people think it merits a particular further explanation, the youtube link is fine to include for me, though equally that content to me is a bit more about the theory of QC and doesn’t necessarily belong in the API docs.

abbycross:
Closes https://github.com/Qiskit/documentation/issues/2533. See discussion on this issue for context.


qiskit/circuit/quantumcircuit.py
```diff
@@ -887,25 +887,6 @@ class QuantumCircuit:
 
        assert qc.size() == 19
 
-    A particularly important circuit property is known as the circuit :meth:`depth`.  The depth
-    of a quantum circuit is a measure of how many "layers" of quantum gates, executed in
-    parallel, it takes to complete the computation defined by the circuit.  Because quantum
-    gates take time to implement, the depth of a circuit roughly corresponds to the amount of
-    time it takes the quantum computer to execute the circuit.  Thus, the depth of a circuit
-    is one important quantity used to measure if a quantum circuit can be run on a device.
```
Eric-Arellano, line 897: This paragraph might be worth preserving? It's helpful to mention we have the method `depth`.

abbycross: Ah, I thought we were removing entirely. I  can move this part of the explanatory info to the description in https://docs.quantum.ibm.com/api/qiskit/qiskit.circuit.QuantumCircuit#depth like Jake suggested

abbycross: Drafted [here](548fa8e4199bc7085633dbdbc88fe88f57b29ecb)

qiskit/circuit/quantumcircuit.py
```diff
@@ -3483,6 +3464,13 @@ def depth(
         ),
     ) -> int:
         """Return circuit depth (i.e., length of critical path).
+        The depth of a quantum circuit is a measure of how many
```
Eric-Arellano: Formatters won't like the lack of space

```suggestion
        """Return circuit depth (i.e., length of critical path).

        The depth of a quantum circuit is a measure of how many
```

ElePT (approved):
LGTM

PR #13624, opened 2025-01-08
Fix phase of `pauli_list.insert(..., qubit=True)` for length-1 `pauli_list`

### LINKED ISSUE #13623: `pauli_list.insert(..., qubit=True)` with length-1 `pauli_list` sets `phase` to wrong shape
### Environment

- **Qiskit version**: 1.3.1
- **Python version**: 3.11
- **Operating system**: macos

### What is happening?

I believe the `phase` attribute of any `PauliList` should always be a 1D numpy array.

For instance, this creates a `PauliList` where the `phase` attribute is a 1D array of length 1:
```
plist = PauliList(['XYZ'])
```

However, this code:
```
plist = plist.insert(1, PauliList(['X']), qubit=True)
```
makes `plist.phase` a 2D numpy array (shape = 1 x 1).

This causes problems when further manipulating the object. For example, trying to pass it to `SparsePauliOp()` fails.

```

plist1 = qi.PauliList(['XX'])

plist2 = qi.PauliList(['X'])
plist2 = plist2.insert(1, qi.PauliList(['X']), qubit=True)

# qiskit thinks the two PauliLists are equivalent:
print(f'{(plist1 == plist2) = }') # np.True_

# but their phase attributes are different shapes:
print(f'{plist1.phase.shape = }') # (1,)
print(f'{plist2.phase.shape = }') # (1, 1)

# and plist1 can be used to construct a SparsePauliOp:
qi.SparsePauliOp(plist1)

# whereas plist2 cannot:
qi.SparsePauliOp(plist2) # raises an error
```

### What should happen?

The `phase` attribute of the output of `insert` should be a 1D array, not a 2D array.
Perhaps the `PauliList.from_symplectic()` constructor (called at the end of `insert`) should also verify/coerce that the `phase` argument is 1D.

(If that's somehow actually the correct behavior, then maybe we'd instead need to fix the SparsePauliOp constructor to handle that input.)

### Any suggestions?

The issue may be how these lines in `PauliList.insert()` treat `phase` on the same footing as `x` and `z`, which are 2D arrays:
https://github.com/Qiskit/qiskit/blob/e33da2ad22b1202fb8c96736416ae585d7b2c593/qiskit/quantum_info/operators/symplectic/pauli_list.py#L448-L452

aeddins-ibm:
### Summary
Fix #13623.

### Details and comments
Fixes bug in `PauliList.insert(..., qubit=True)` method, where the output `PauliList` would have a 2D `phase` attribute (should be 1D).

The bug was in a line of code manually broadcasting the `phase` of the `PauliList` being inserted to match the shape of the `phase` of the original `PauliList`. This PR removes this line of code, and lets numpy handle the broadcasting of `phase`.

This PR also adds a check in `PauliList.from_symplectic(..., phase)` to reject `phase` if it is a multi-dimensional array, and a test for issue #13623.

(For the record, I messed up the commit message in 92161ec. The simpler `if/elif` case is indeed moved to be first, but in the commit message I mixed up the two cases).

### DIFF
--- a/releasenotes/notes/fix-paulilist-length1-phase-688d0e3a64ec9a9f.yaml
+++ b/releasenotes/notes/fix-paulilist-length1-phase-688d0e3a64ec9a9f.yaml
@@ -1,6 +1,6 @@
 ---
 fixes:
   - |
-    Fixed a bug that caused :meth:`.PauliList.insert(..., qubit=True)` to produce a `phase`
+    Fixed a bug that caused :meth:`.PauliList.insert` with ``qubit=True`` to produce a `phase`
     attribute with the wrong shape when the original object was length 1.
     Fixed `#13623 <https://github.com/Qiskit/qiskit/issues/13623>`__.
--- a/releasenotes/notes/fix-paulilist-length1-phase-688d0e3a64ec9a9f.yaml
+++ b/releasenotes/notes/fix-paulilist-length1-phase-688d0e3a64ec9a9f.yaml
@@ -1,0 +1,6 @@
+---
+fixes:
+  - |
+    Fixed a bug that caused :meth:`.PauliList.insert(..., qubit=True)` to produce a `phase`
+    attribute with the wrong shape when the original object was length 1.
+    Fixed `#13623 <https://github.com/Qiskit/qiskit/issues/13623>`__.
--- a/qiskit/quantum_info/operators/symplectic/pauli_list.py
+++ b/qiskit/quantum_info/operators/symplectic/pauli_list.py
@@ -455,7 +455,6 @@ class PauliList(BasePauli, LinearMixin, GroupMixin):
             #  Blocks are already correct size
             value_x = value.x
             value_z = value.z
-            value_phase = value.phase
         elif len(value) == 1:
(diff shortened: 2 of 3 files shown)


releasenotes/notes/fix-paulilist-length1-phase-688d0e3a64ec9a9f.yaml
```diff
@@ -0,0 +1,6 @@
+---
+fixes:
+  - |
+    Fixed a bug that caused :meth:`.PauliList.insert(..., qubit=True)` to produce a `phase`
```
jakelishman:
```suggestion
    Fixed a bug that caused :meth:`.PauliList.insert` with ``qubit=True`` to produce a `phase`
```

Thanks for the great cross-ref - unfortunately Sphinx isn't smart enough to be able to handle it with keyword examples in the middle, so we have to separate them out.

jakelishman (dismissed):
Thanks for this fix - it looks sensible to me (though I'll admit I'm not at all familiar with the code here).

jakelishman (approved):

PR #13633, opened 2025-01-08
Fix an error in qasm3 exporter when operating on unitary gates

### LINKED ISSUE #13362: `qiskit.qasm3.dumps` raise TypeError: only length-1 arrays can be converted to Python scalars
### Environment

- **Qiskit version**: 
```log
qiskit                    1.2.4
qiskit-aer                0.15.0
qiskit-aer-gpu            0.14.2
qiskit-algorithms         0.3.0
qiskit-ibm-provider       0.11.0
qiskit-ibm-runtime        0.30.0
qiskit-qasm3-import       0.5.0
```

- **Python version**: 
```log
Python 3.12.2 | packaged by conda-forge | (main, Feb 16 2024, 20:50:58) [GCC 12.3.0] on linux
```

- **Operating system**: 
```log
Ubuntu 24.04.1 LTS x86_64
```

### What is happening?

A numpy type error raised during the execution of `qiskit.qasm3.dumps` .

Image
Image

```python
from qiskit.qasm3 import dumps

qc = QuantumCircuit(4)
for i in range(4):
    qc.h(i)
    qc.append(random_unitary(2),[i])
    
print(qc.draw())

dumps(qc)
```

or 

```python
from qiskit.qasm3 import dumps

qc = QuantumCircuit(4)
for i in range(4):
    qc.h(i)
    qc.append(random_unitary(2).to_instruction(),[i])
    
print(qc.draw())

dumps(qc)
```

### What should happen?

I should be able to dump this circuit to QASM3 because `qiskit.qasm2.dumps` can successfully dump this circuit.

### Any suggestions?

This issue appears to be mentioned in https://github.com/Qiskit/qiskit/issues/12013#issuecomment-2038064846

gadial:
### Summary
Fixes a bug causing the qasm3 exporter to crash on circuits containing a unitary gate.

Fixes #13362.

### Details and comments
The bug results from the `params` field of the unitary `operation` being handled incorrectly, due to it being a somewhat nonstandard field (as opposed to the `params` field of `u` gate, for instance, which is used differently).

This problem was already handled in the `qasm2` exporter by a call to `_qasm2_decomposition` in the operation, if present. This PR integrates the same solution in the qasm3 exporter, renaming the method to `_qasm_decomposition`.

### DIFF
--- a/releasenotes/notes/fix-qasm-3-unitary-2da190be6ba25bbd.yaml
+++ b/releasenotes/notes/fix-qasm-3-unitary-2da190be6ba25bbd.yaml
@@ -1,5 +1,5 @@
 ---
 fixes:
   - |
-    Fix a bug in :class:`qasm3.Exporter` that caused the exporter to crash when
-    handling a unitary gate due to incorrect processing of its `params` field.
+    Fix a bug in :class:`.qasm3.Exporter` that caused the exporter to crash when
+    handling a unitary gate due to incorrect processing of its ``params`` field.
--- a/releasenotes/notes/fix-qasm-3-unitary-2da190be6ba25bbd.yaml
+++ b/releasenotes/notes/fix-qasm-3-unitary-2da190be6ba25bbd.yaml
@@ -1,0 +1,5 @@
+---
+fixes:
+  - |
+    Fix a bug in :class:`qasm3.Exporter` that caused the exporter to crash when
+    handling a unitary gate due to incorrect processing of its `params` field.
--- a/test/python/qasm3/test_export.py
+++ b/test/python/qasm3/test_export.py
@@ -2665,6 +2665,24 @@ switch (switch_dummy_0) {
         test = dumps(qc, experimental=ExperimentalFeatures.SWITCH_CASE_V1)
         self.assertEqual(test, expected)
 
+    def test_circuit_with_unitary(self):
+        """Test that circuits with `unitary` gate are correctly handled"""
+        matrix = [[0, 1], [1, 0]]
+        qc = QuantumCircuit(1)
+        qc.unitary(matrix, [0])
+        expected = """\
+OPENQASM 3.0;
+include "stdgates.inc";
+gate unitary _gate_q_0 {
+  U(pi, -pi, 0) _gate_q_0;
+}
+qubit[1] q;
+unitary q[0];
+"""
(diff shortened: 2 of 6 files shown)


releasenotes/notes/fix-qasm-3-unitary-2da190be6ba25bbd.yaml
```diff
@@ -0,0 +1,5 @@
+---
+fixes:
+  - |
+    Fix a bug in :class:`qasm3.Exporter` that caused the exporter to crash when
+    handling a unitary gate due to incorrect processing of its `params` field.
```
jakelishman:
```suggestion
    Fix a bug in :class:`.qasm3.Exporter` that caused the exporter to crash when
    handling a unitary gate due to incorrect processing of its ``params`` field.
```

test/python/qasm3/test_export.py
```diff
@@ -2665,6 +2665,24 @@ def test_switch_v1_expr_target(self):
...
+OPENQASM 3.0;
+include "stdgates.inc";
+gate unitary _gate_q_0 {
+  U(pi, -pi, 0) _gate_q_0;
+}
+qubit[1] q;
+unitary q[0];
+"""
+        test = dumps(qc)
+        print(test)
```
jakelishman:
```suggestion
```

If the test fails, the assertion should print out both strings - it's the job of the test runner to display those nicely (if it was a problem for you, likely we can tweak some settings somewhere).

jakelishman (commented):
I think this is a fine way to go about solving the problem, thanks.  At first I was a bit nervous because the gate models of OQ2 and OQ3 aren't necessarily identical (OQ3 has _scope_ to be more permissive), but in practice I doubt it'll come up any time soon, if ever.  If it does, we'll deal with it then.

test/python/qasm3/test_export.py
gadial: This one's probably a debug print I forgot to remove - my bad.

jakelishman: Oh yeah, sure, my thing is that the test runner should show the exact same string on an assertion failure, so the debug print shouldn't be necessary.  If it's still necessary, it probably indicates that we need better configuration exposed for the test runner or something.

jakelishman (approved):

PR #13635, opened 2025-01-08
Fix repeated plots in docs

ElePT:
### Summary
The use of `:context:` in pages with several plots was causing an accumulation of figures, this PR fixes this issue by adding `:context: close-figs`. For example, the `grover_operator` docs before this PR showed:
Screenshot 2025-01-08 at 17 37 30

While after the fix there is only one figure per snippet (as expected):

 
Screenshot 2025-01-08 at 17 36 34

### DIFF
--- a/qiskit/circuit/library/grover_operator.py
+++ b/qiskit/circuit/library/grover_operator.py
@@ -118,7 +118,7 @@ def grover_operator(
 
         .. plot::
             :include-source:
-            :context:
+            :context: close-figs
 
             oracle = QuantumCircuit(1)
             oracle.z(0)  # the qubit state |1> is the good state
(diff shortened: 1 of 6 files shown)


jakelishman (approved):
Wow, in some ways I'm surprised that this is the default behaviour.  I guess I can just about see the purpose if you're showing how a plot builds up in the matplotlib documentation, but for most uses, it seems like `close-figs` should be the default.

ElePT:
[@Mergifyio](https://github.com/Mergifyio) backport stable/1.3

ElePT:

ElePT:
@Mergifyio backport stable/1.3

PR #13649, opened 2025-01-10
Improve 2q block collection via 1q quaternion-based collection

jakelishman:
### Summary

This is a small series of patches, which could be split into 2-3 separate PRs if preferred.  There are two main goals:

- introduce a quaternion-based mechanism for working with the $U(2)$ group (1q gates) matrix-free (uses 5 floats, rather than 8 + nd-matrix overhead).
- improve the collection speed of `ConsolidateBlocks`

This doesn't do everything that could be done for `ConsolidateBlocks`, but I've stopped at the point where the most natural changes to me now are quite a bit larger.

Using this script:
```python
from qiskit import transpile
from qiskit.circuit import library as lib
from qiskit.transpiler.passes import ConsolidateBlocks

pass_ = ConsolidateBlocks(basis_gates=["rz", "sx", "ecr"], force_consolidate=True)

# Some arbitary circuit with lots of runs that use both 1q runs and 2q runs
# with gates in both directions (transpile will put them all in one direction).
qc_base = transpile(
    lib.quantum_volume(100, 100, seed=0),
    basis_gates=["rz", "sx", "ecr"],
    seed_transpiler=0,
)
qc = qc_base.copy_empty_like()
flip = False
for inst in qc_base.data:
    if len(inst.qubits) == 2:
        if flip:
            inst = inst.replace(qubits=inst.qubits[::-1])
        flip = not flip
    qc._append(inst)

dags = [circuit_to_dag(qc, copy_operations=False) for _ in [None]*100]
%time for dag in dags: pass_.run(dag)
```
I see a modest (~10%) improvement in the pass time, going from ~117ms to ~106ms.

### Details and comments

Individual commit notes:
#### Add versor-based representation of 1q gates

1q gates are members of the group U(2), which we can represent as a scalar phase term and a member of SU(2).  The members of SU(2) can be represented by versors (also called unit quaternions, but I got tired of typing that all the time...).

This adds a representation of versors and the group action to the Rust code, and ways to convert from matrix-based forms to the them.

This commit introduces `nalgebra` as a dependency, to use its quaternion logic.  This is a relatively heavy dependency, especially for something as simple as quaternions, but some of this is in anticipation of moving more matrix code to the static matrices of `nalgebra`, rather than the too-dynamic-for-our-needs ones of `ndarray`; `faer` also offers static matrices, but its APIs continue to heavily fluctuate between versions, and it requires ever higher MSRVs.

#### Use quaternions in 1q block collection

Switch the inner algorithm of `ConsolidateBlocks` to use the quaternion form for single-qubit matrix multiplications.  This offered a few percentage-points speedup for the block collection for typical `rz-sx-rz-sx-rz`-type runs in quantum-volume-like collections.

#### Avoid unnecessary allocations in qargs lookup

Switch the lookup logic for the qargs in the block collector to determine the qargs ordering without additional heap allocations.  This offers another modest (~4%) improvement in collection performance for large runs.

#### Avoid allocations in simple matrix operations

Producing the Kronecker product of the two single-qubit matrices from the versor representation is trivially calculable, and can be written into an existing allocation.  Similarly, switching the qubit order of a 2q matrix involves only six swaps, and does not need to allocate a new matrix if one is already available.

There are lots of places remaining in this code where more matrix allocations could be avoided.  If nothing else, it should be possible to allocate only three 2q matrices in total, and keep shuffling the labelling of them when doing `A.B -> C`.  `ndarray` does not make this easy, though; `nalgebra` and `faer` both have better interfaces for doing this, but currently all our matrix code in the `Operation` trait is in terms of `ndarray`.


jakelishman:
There's also more places we could make use of the versor representation, like in 1q gate optimisation, but that pass currently involves passing matrices through many different parts of its interface with itself, so it'll be more complicated to modify.

crates/accelerate/src/qi/mod.rs
```diff
@@ -0,0 +1,18 @@
+// This code is part of Qiskit.
```
mtreinish: Should we call this directory `quantum_info` to match what we do in python?

crates/accelerate/src/qi/mod.rs
```diff
@@ -0,0 +1,18 @@
...
+//
+// This code is licensed under the Apache License, Version 2.0. You may
+// obtain a copy of this license in the LICENSE.txt file in the root directory
+// of this source tree or at http://www.apache.org/licenses/LICENSE-2.0.
+//
+// Any modifications or derivative works of this code must retain this
+// copyright notice, and modified files need to carry a notice indicating
+// that they have been altered from the originals.
+
+//! Quantum-information and linear-algebra related functionality, typically used as drivers for
```
mtreinish: We might want to move over some linear algebra functionality like: https://github.com/Qiskit/qiskit/blob/main/crates/accelerate/src/utils.rs and https://github.com/Qiskit/qiskit/blob/main/crates/accelerate/src/synthesis/linear/utils.rs (although that first one might not be needed anymore).

crates/accelerate/src/qi/versor_gate.rs
```diff
@@ -0,0 +1,445 @@
...
+
+use nalgebra::{Quaternion, UnitQuaternion};
+use ndarray::ArrayView2;
+use num_complex::Complex64;
+use thiserror::Error;
+
+use qiskit_circuit::operations::{Param, StandardGate};
+
+const COS_PI_8: f64 = 0.9238795325112867;
+const SIN_PI_8: f64 = 0.3826834323650898;
```
mtreinish: Heh, I thought you didn't like `PI_8` variable naming for PI / 8 :)

mtreinish (commented):
Just some quick high level comments. I missed the update and don't want these lost by github weirdness. I'll review in more depth later.

crates/accelerate/src/qi/versor_gate.rs
jakelishman: I was aiming for consistency, really, but looking again I should have called it `COS_FRAC_PI_8`, perhaps?

crates/accelerate/src/qi/mod.rs
jakelishman: Yeah, I'm happy in a follow-up to move a few other bits over.  I think there's other loose files and bits and bobs that could probably move into it too, just to keep things a bit more localised.

crates/accelerate/src/qi/mod.rs
jakelishman: I don't mind particularly - I read `quantum_info` in my head as `qi` most of the time anyway haha.  I don't think we necessarily _must_ match Python space, but if you prefer it for consistency I don't have any issues changing it.

crates/accelerate/src/qi/versor_gate.rs
mtreinish: Yeah, that's probably the more consistent name with the built-in f64 consts and is harder to mess up by mistake.

crates/accelerate/src/qi/mod.rs
jakelishman: Coming back to this again: how strongly do we feel about consistency with Python space?  We're already not consistent elsewhere in Rust (`dag_circuit.rs` vs `dagcircuit.py`, for example), and for Rust-internal stuff, there's not much imperative to match Python.

If people prefer `quantum_info` in general, I'm still happy to swap to that, I just wanted to challenge the "consistency" argument.

(@Cryoris, for visibility.)

crates/accelerate/src/qi/versor_gate.rs
jakelishman: Done in [24bddd7](https://github.com/Qiskit/qiskit/pull/13649/commits/24bddd7f7bc3b78870943fd9d9a5226e1f9fc532).

jakelishman:
I fixed the merge conflicts, and also I pivoted the structs a little bit to expose both $SU(2)$ and $U(2)$ representations, since I was already effectively using the $SU(2)$ representation in the block collector.

crates/accelerate/src/qi/mod.rs
mtreinish: I think I still prefer `quantum_info` the `qi` abbreviation is still hard for me to parse at first glance.

crates/accelerate/src/qi/versor_u2.rs
```diff
@@ -0,0 +1,547 @@
...
+// obtain a copy of this license in the LICENSE.txt file in the root directory
+// of this source tree or at http://www.apache.org/licenses/LICENSE-2.0.
+//
+// Any modifications or derivative works of this code must retain this
+// copyright notice, and modified files need to carry a notice indicating
+// that they have been altered from the originals.
+
+use std::f64::consts::{FRAC_1_SQRT_2, FRAC_PI_2, FRAC_PI_4, FRAC_PI_8};
+
+use nalgebra::{Quaternion, Unit, UnitQuaternion};
```
mtreinish: There are some advantages to using a library that has some support for computer graphics. It's nice we don't have to write all of this out.

crates/accelerate/src/qi/versor_u2.rs
```diff
@@ -0,0 +1,547 @@
...
+    pub fn from_standard(gate: StandardGate, params: &[Param]) -> Result<Self, VersorU2Error> {
+        match gate {
+            StandardGate::GlobalPhaseGate => {
+                let &[Param::Float(phase)] = params else {
+                    return Err(VersorU2Error::Symbolic);
+                };
+                Ok(VersorSU2::identity().with_phase(phase))
+            }
+            StandardGate::HGate => {
+                debug_assert!(params.is_empty());
```
mtreinish: We probably should get in the habit of doing this more for underlying assumptions like this.. Not that we build a ton in debug mode, but we do in some places so this would be good to sanity check.

crates/accelerate/src/qi/versor_u2.rs
```diff
@@ -0,0 +1,547 @@
...
+    }
+
+    /// Get the versor representation of a 1q [StandardGate] without constructing a matrix.
+    ///
+    /// Returns the error state if `gate` is not 1q, or if any of the parameters are symbolic.
+    pub fn from_standard(gate: StandardGate, params: &[Param]) -> Result<Self, VersorU2Error> {
+        match gate {
+            StandardGate::GlobalPhaseGate => {
+                let &[Param::Float(phase)] = params else {
+                    return Err(VersorU2Error::Symbolic);
```
mtreinish: Won't this identify the error class as having symbolic parameters if we somehow had a mismatch between param length and the number of params?

Maybe we should add a `debug_assert!(params.len(), StandardGate::GlobalPhaseGate.num_params())` (and for all the other arms. I don't think we want the runtime overhead of doing this everytime, but for a debug build this will at least catch the pathological case.

crates/accelerate/src/qi/versor_u2.rs
```diff
@@ -0,0 +1,547 @@
...
+    fn get(&self, row: usize, col: usize) -> Complex64 {
+        self[row][col]
+    }
+}
+impl Matrix1q for ArrayView2<'_, Complex64> {
+    #[inline(always)]
+    fn get(&self, row: usize, col: usize) -> Complex64 {
+        self[(row, col)]
+    }
+}
```
mtreinish: Do you want to add a `Matrix2` impl here too? I guess nothing needs it now because you only used static arrays and ndarray here. But I feel now that we have `UnitaryGate` that can be built from a `Matrix2<Complex64>` it might be good to have the impl here so we can call `VersorU2::from_matrix_unchecked` directly on that.

mtreinish (commented):
I've only reviewed the quantum info piece so far. But it looks good to me. I'll follow up tomorrow reviewing the changes to the block converter.

crates/accelerate/src/qi/versor_u2.rs
jakelishman: Yeah, I'll update this in the face of the new `UnitaryGate` for sure.

crates/accelerate/src/convert_2q_block_matrix.rs
```diff
@@ -53,94 +52,147 @@ pub fn get_matrix_from_inst(py: Python, inst: &PackedInstruction) -> PyResult<Ar
...
+                Qarg::Q0
+            } else {
+                Qarg::Q1
+            }
+        } else if interned.len() == 2 {
+            if interned[0] == block_index_map[0] {
+                Qarg::Q01
+            } else {
+                Qarg::Q10
+            }
```
mtreinish: Not that I think it matters very much, but I feel like we could probably maybe map the value of `qargs` directly to `Q0`, `Q1`, `Q01`, and `Q10` and avoid the interner lookup. There should only be 4 interned ids in the context of a 2q block so it feels like once we see an id once we don't need to do the full get_qargs() call on every instruction. But I highly doubt this is a bottlneck at all, so probably not worth the extra complexity.

crates/accelerate/src/convert_2q_block_matrix.rs
```diff
@@ -53,94 +52,147 @@ pub fn get_matrix_from_inst(py: Python, inst: &PackedInstruction) -> PyResult<Ar
...
+        qubits_1q.map(|sep| sep.matrix_into(&mut work)),
+        output_matrix,
+    ) {
+        (Some(sep), Some(state)) => Ok(sep.dot(&state)),
+        (None, Some(state)) => Ok(state),
+        (Some(sep), None) => Ok(sep.to_owned()),
+        // This shouldn't actually ever trigger, because we expect blocks to be non-empty, but it's
+        // trivial to handle anyway.
+        (None, None) => Ok(arr2(&TWO_QUBIT_IDENTITY)),
+    }
```
mtreinish, line 201: Not for this PR, but it'll be great to have this all done via `Matrix4` after this merges, so we can just consolidate to a `UnitaryGate` with `ArrayType::TwoQ` and avoid allocating a matrix altogether.

crates/accelerate/src/convert_2q_block_matrix.rs
```diff
@@ -53,94 +52,147 @@ pub fn get_matrix_from_inst(py: Python, inst: &PackedInstruction) -> PyResult<Ar
...
+                target[out_row][out_col] = q1_row[(out_col & 2) >> 1] * q0[out_row][out_col & 1];
+            }
+        }
+        let q1_row = [Complex64::new(-iy, ix), Complex64::new(scalar, -iz)];
+        for out_row in 0..2 {
+            for out_col in 0..4 {
+                target[out_row + 2][out_col] =
+                    q1_row[(out_col & 2) >> 1] * q0[out_row][out_col & 1];
+            }
+        }
```
mtreinish, line 103: Is there a nalgebra method to do the kronecker product like this? I'm guessing not because we're not building the matrix representation for q1. But figured there might be a small chance.

mtreinish (commented):
Following up with the `convert_2q_block_matrix.rs`  file this looks good. Just a couple more small comments.

crates/accelerate/src/convert_2q_block_matrix.rs
jakelishman: Yeah absolutely - I stopped in this PR at the point that I started interacting with the 2q matrices because I wasn't prepared to put too much more time into things I knew were both in flux and needed more major changes.

crates/accelerate/src/convert_2q_block_matrix.rs
jakelishman: I think I did it like this because I didn't write out the 2x2 matrix of one of the quaternions while doing it (only a 2x1 row twice).  I don't really know _why_ I did that - probably faster to write out the two matrices and use `kron`.

jakelishman: There's no `kronecker_to` method in `nalgebra`, which is I think what I needed here.  For the time being, I've just left this as-is along with all the other 2q inefficiencies, which I think are best left for follow-ups.

crates/accelerate/src/qi/mod.rs
jakelishman: Done in d2f58d9bba.

crates/accelerate/src/qi/versor_u2.rs
jakelishman: I just moved the logic to be in front of the whole `match` in c196226d77, since it's a `debug_assert!` anyway, so now _every_ branch tests it precisely.

crates/accelerate/src/qi/versor_u2.rs
jakelishman: Done in 3f20b0fe75.

crates/accelerate/src/convert_2q_block_matrix.rs
jakelishman: Done in 3bfb8f7428, though I don't think it has much impact on the timings.

jakelishman:
Should be updated and comments addressed now.  There's still quite a lot of inefficiency in how this PR handles 2q matrices, but I thought it's better to leave that til a follow-up and do it more completely throughout `ConsolidateBlocks`.

crates/accelerate/src/convert_2q_block_matrix.rs
mtreinish: Maybe something to open an issue or contribute in the future then.

crates/accelerate/src/convert_2q_block_matrix.rs
```diff
@@ -53,94 +52,153 @@ pub fn get_matrix_from_inst(py: Python, inst: &PackedInstruction) -> PyResult<Ar
...
+}
+
+/// Extract a versor representation of an arbitrary 1q DAG instruction.
+fn versor_from_1q_gate(py: Python, inst: &PackedInstruction) -> PyResult<VersorU2> {
+    let tol = 1e-12;
+    match inst.op.view() {
+        OperationRef::StandardGate(gate) => VersorU2::from_standard(gate, inst.params_view()),
+        OperationRef::Unitary(gate) => match &gate.array {
+            ArrayType::NDArray(arr) => VersorU2::from_ndarray(&arr.view(), tol),
+            ArrayType::OneQ(arr) => VersorU2::from_nalgebra(arr, tol),
```
mtreinish: You didn't want to use `from_naglebra_unchecked` here. I think we can probably safely assume a `UnitaryGate`'s matrix is unitary. The python space constructor has a check on the constructor, and while you can opt-out of the check it's the user's burden to ensure it is then. The rust space direct construction of the object only ever creates a unitary (it's only consolidate blocks or qv iirc).

crates/accelerate/src/quantum_info/versor_u2.rs
```diff
@@ -0,0 +1,554 @@
...
+    use super::*;
+
+    use approx::AbsDiffEq;
+    use ndarray::aview2;
+    use qiskit_circuit::operations::{Operation, Param, StandardGate, STANDARD_GATE_SIZE};
+
+    fn all_1q_gates() -> Vec<StandardGate> {
+        (0..STANDARD_GATE_SIZE as u8)
+            .filter_map(|x| {
+                ::bytemuck::checked::try_cast::<_, StandardGate>(x)
```
mtreinish, line 488: I actually have code somewhere that I added a `impl TryFrom<u8> for StandardGate` and just made a static  `[StandardGate; STANDARD_GATE_SIZE]` array to do the conversion. It was part of some code I didn't end up finish or using for anything. But I can dig up the branch and push that as a standalone if you think it would be useful (I added it for a similar use case to this).

mtreinish (dismissed):
This LGTM now thanks for doing this, I'm looking forward to using this and also the better usage of `UnitaryGate` in the pass in a follow-up. I left two questions inline one isn't relevant for this PR itself but more about a separate follow-up. The other is about the code, but feel free to enqueue this to merge if you don't want to change it.

crates/accelerate/src/quantum_info/versor_u2.rs
jakelishman: `TryFrom<u8> for StandardGate` _feels_ wrong to me, but not super strongly.  It feels nicer somehow using `bytemuck` to say "I know I'm doing something a little bit odd", since it feels like we're kind of conjuring semantics out of nowhere - the fact that `StandardGate` is backed by integers is (imo) an implementation detail, albeit a fixed one.

If we wanted to expose an iterator over all gates, I'd just expose your `[StandardGate; Standard_GATE_SIZE]` as a `static` publicly, then you can iterate through it directly.  That's what I did with `BitTerm` in the tests of `SparseObservable`'s lookup tables.

crates/accelerate/src/convert_2q_block_matrix.rs
jakelishman: Done in ac469a2e70.

mtreinish (approved):

PR #13650, opened 2025-01-10
Drop support for 32bit x86 linux and windows

mtreinish:
### Summary

Going into Qiskit 2.0 we decided to drop support for windows 32 bit and linux i686 platforms. Right now the maintainence burden for continuing to support these platforms is becoming increasingly high, as we're increasing writing low level rust code to improve performance where we have to limit ourselves to continue to ensure 32bit platforms work. The user base for these platforms is quite small small; in the last 150 days there were only 7 downloads for i686 out of 221,992 total downloads from PyPI (there is no metadata field in PyPI's database to differentiate between windows 32bit and 64bit if the CPU is 64bit capable it uses AMD64 for the CPU field), 5 downloads where the CPU type was: "x86" (I'm not sure which platform uses that string, linux uses x86_64 for 64bit intel, but i686, i586, i486, etc for 32bit systems), and 35 downloads were armv7l which is not an official supported platform but is 32bit and you could build from source (this is typically raspberry pi users). When combining this with the lack of wider ecosystem support for 32bit platforms, numpy, scipy, etc all have dropped 32bit support already, keeping support for 32bit platforms is no longer tenable. This commit updates the wheel building configuration to drop the 32bit platform builds and adds a release note documenting the change in status for these platforms in 2.0.

A separate Qiskit/documentation PR will be opened to update our platform support documentation here:

https://docs.quantum.ibm.com/guides/install-qiskit#operating-system-support

closer to the final 2.0 release to remove the 32bit platforms from the list. Although we likely should update that page to have a subsection for 2.x and 1.x while both are supported.

### Details and comments

### DIFF
--- a/releasenotes/notes/64-bit-only-4132f330ec7804b3.yaml
+++ b/releasenotes/notes/64-bit-only-4132f330ec7804b3.yaml
@@ -10,5 +10,7 @@ upgrade:
     scientific/numeric Python community's trend away from 32 bit platform
     support maintaining support for 32bit platforms is no longer viable.
     Qiskit 1.x will still continue to support 32 bit platforms, but starting
-    in this 2.0.0 release Qiskit no longer supports these platforms, and will
-    not publish pre-compiled binaries for these platforms any longer.
+    in this 2.0.0 release Qiskit no longer supports these platforms, will
+    not publish pre-compiled binaries for these platforms any longer, and
+    there is no longer any guarantee of being able to build from source on
+    32 bit platforms.
--- a/releasenotes/notes/64-bit-only-4132f330ec7804b3.yaml
+++ b/releasenotes/notes/64-bit-only-4132f330ec7804b3.yaml
@@ -1,0 +1,14 @@
+---
+upgrade:
+  - |
+    Qiskit no longer supports Linux i686 and 32 bit Windows. Starting in Qiskit
+    2.0 a 64 bit platform is needed to run Qiskit. The user base for 32bit
+    architectures is relatively small and as Qiskit continues to focus on
+    improving performance to handle the increased scale in the complexity
+    of quantum computing hardware maintaining support for 32 bit platforms
+    is proving increasingly difficult. This coupled with the larger
+    scientific/numeric Python community's trend away from 32 bit platform
+    support maintaining support for 32bit platforms is no longer viable.
+    Qiskit 1.x will still continue to support 32 bit platforms, but starting
+    in this 2.0.0 release Qiskit no longer supports these platforms, and will
+    not publish pre-compiled binaries for these platforms any longer.
--- a/.github/workflows/wheels-build.yml
+++ b/.github/workflows/wheels-build.yml
@@ -27,15 +27,6 @@ on:
         default: "default"
         required: false
 
-      wheels-32bit:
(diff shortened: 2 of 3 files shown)


releasenotes/notes/64-bit-only-4132f330ec7804b3.yaml
```diff
@@ -0,0 +1,14 @@
...
+    2.0 a 64 bit platform is needed to run Qiskit. The user base for 32bit
+    architectures is relatively small and as Qiskit continues to focus on
+    improving performance to handle the increased scale in the complexity
+    of quantum computing hardware maintaining support for 32 bit platforms
+    is proving increasingly difficult. This coupled with the larger
+    scientific/numeric Python community's trend away from 32 bit platform
+    support maintaining support for 32bit platforms is no longer viable.
+    Qiskit 1.x will still continue to support 32 bit platforms, but starting
+    in this 2.0.0 release Qiskit no longer supports these platforms, and will
+    not publish pre-compiled binaries for these platforms any longer.
```
jakelishman: On the last sentence: do we want to drive home even further that we will not even commit to Qiskit _building_ on 32-bit platforms at all?  Your note does already imply it, in some ways.

In particular, I fairly strongly expect that as soon as we commit to dropping 32-bit support, I'll rejig `PackedOperation` a little bit to explicitly store a pointer-like thing within itself rather than a `u64` (assuming #13486), so we can start moving in the Miri-checkable direction of strict pointer providence, and at that point we'll be requiring that the width of `*mut ()` is 64 bits.

jakelishman (commented):
I'll be happy to see the back of the 32-bit platforms now; it's rather easier to reason about when everything's the same, especially since there's ever more places where we're getting very concerned about the exact alignments, padding bits, and whatnot implicit in pointers.

releasenotes/notes/64-bit-only-4132f330ec7804b3.yaml
mtreinish: When I wrote this I thought about that, but I didn't want to call it out explicitly because we can still build on 32bit platforms right now and I wasn't 100% sure we would land the changes in 2.0 that would break this. But I can say something like "and we're no longer providing any guarantee of being able to build from source on 32 bit platforms."

jakelishman: Yeah, that's enough for me. I don't necessarily want to say "it 100% won't build" before we've actually landed it, but I think it's probably worth really mentioning that this really is the end of the road - not even Tier 3 or anything.

mtreinish:
```suggestion
    Qiskit 1.x will still continue to support 32 bit platforms, but starting
    in this 2.0.0 release Qiskit no longer supports these platforms, will
    not publish pre-compiled binaries for these platforms any longer, and
    there is no longer any guarantee of being able to build from source on
    32 bit platforms.
```

kevinhartman (approved):
Looks good to me, and should simplify #13486 and our future bit-packing needs.

I'm glad we have the metrics to show that almost nobody is using Qiskit on 32-bit platforms. Thanks for including them here 🙂.

PR #13651, opened 2025-01-10
Fix `UnitarySynthesis` pass bug when target contains global gates

ElePT:
### Summary
While working on https://github.com/Qiskit/qiskit/pull/12850 I realized that the `UnitarySynthesis` Rust implementation  would sometimes append non-2q-gates to the basis set and panic over the dimension mismatch. I had assumed that `target.operation_names_for_qargs(Some(&qubits))` would only return gates whose number of qubits matched `len(qubits)`, but this assumption is incorrect. It will return any gate with **some** overlap with the provided qargs, independently of the number of qubits of the gate. This bug would come up often when the target contained global gates. 

I believe that the fix is as simple as making sure the 2q basis contains only 2q gates, at least for the cases I found.

### DIFF
--- a/crates/accelerate/src/unitary_synthesis.rs
+++ b/crates/accelerate/src/unitary_synthesis.rs
@@ -600,7 +600,7 @@ fn get_2q_decomposers_from_target(
                     }
                     // Filter out non-2q-gate candidates
                     if op.operation.num_qubits() != 2 {
-                        continue
+                        continue;
                     }
                     available_2q_basis.insert(key, replace_parametrized_gate(op.clone()));
 
--- a/crates/accelerate/src/unitary_synthesis.rs
+++ b/crates/accelerate/src/unitary_synthesis.rs
@@ -599,9 +599,8 @@ fn get_2q_decomposers_from_target(
                         _ => continue,
                     }
                     // Filter out non-2q-gate candidates
-                    match op.operation.num_qubits() {
-                        2 => (),
-                        _ => continue,
+                    if op.operation.num_qubits() != 2 {
+                        continue
                     }
                     available_2q_basis.insert(key, replace_parametrized_gate(op.clone()));
 
--- a/crates/accelerate/src/unitary_synthesis.rs
+++ b/crates/accelerate/src/unitary_synthesis.rs
@@ -545,6 +545,7 @@ fn get_2q_decomposers_from_target(
     let mut available_2q_props: IndexMap<&str, (Option<f64>, Option<f64>)> = IndexMap::new();
 
     let mut qubit_gate_map = IndexMap::new();
+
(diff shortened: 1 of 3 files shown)


crates/accelerate/src/unitary_synthesis.rs
```diff
@@ -597,7 +598,11 @@ fn get_2q_decomposers_from_target(
                         OperationRef::Standard(_) => (),
                         _ => continue,
                     }
-
+                    // Filter out non-2q-gate candidates
+                    match op.operation.num_qubits() {
+                        2 => (),
+                        _ => continue,
+                    }
```
mtreinish:
```suggestion
                    if op.operation.num_qubits() != 2 {
                        continue
                    }
```

mtreinish (commented):
Looks good to me, but just one small code comment

crates/accelerate/src/unitary_synthesis.rs
```diff
@@ -597,7 +598,10 @@ fn get_2q_decomposers_from_target(
                         OperationRef::Standard(_) => (),
                         _ => continue,
                     }
-
+                    // Filter out non-2q-gate candidates
+                    if op.operation.num_qubits() != 2 {
+                        continue
```
ElePT:
```suggestion
                        continue;
```

releasenotes/notes/fix-unitary-synthesis-global-gates-19b93840b28cfcf7.yaml
```diff
@@ -0,0 +1,7 @@
+---
+fixes:
+  - |
+    Fixed a bug in the :class:`.UnitarySynthesis` transpiler pass where 
+    non-2-qubit gates would be incuded in the available 2 qubit basis,
```
ShellyGarion: incuded --> included ?

crates/accelerate/src/unitary_synthesis.rs
```diff
@@ -597,7 +598,10 @@ fn get_2q_decomposers_from_target(
                         OperationRef::Standard(_) => (),
                         _ => continue,
                     }
-
+                    // Filter out non-2q-gate candidates
+                    if op.operation.num_qubits() != 2 {
+                        continue;
```
ShellyGarion, line 603: it's the same behavior both for 1-qubit and n-qubits gates (n>2)?

ShellyGarion (commented):
Good catch! Thanks!

crates/accelerate/src/unitary_synthesis.rs
ElePT: yes, the gate matrix needs to be 4x4 in this part of the code, any n!=2 qubit gate causes the dimension mismatch panic

mtreinish (approved):

PR #13652, opened 2025-01-10
Add test case for specialized measurement with dynamic circuits

mtreinish:
### Summary

In the target model we have had support for a long time for creating specialized measurements by subclassing the Measure class and then adding that subclass to your target. However, this functionality was never tested explicitly. This commit closes that coverage gap by adding a dedicated test for this full path transpilation with the preset pass managers ensuring the operations pass through the transpiler and are preserved as measurements.

### Details and comments


ElePT (approved):
I had never thought of this functionality, cool! LGTM

PR #13654, opened 2025-01-13
Port linear_depth_lnn to rust

### LINKED ISSUE #12228: Port `synth_cnot_depth_line_kms` to Rust

gadial:
### Summary
Ports the `optimize_cx_circ_depth_5n_line` method to Rust.

Closes #12228 

### Details and comments
This is synthesis pass for constructing general linear reversible circuits given by an invertible boolean matrix. The code generating the instruction list was ported to Rust in a straightforward manner; construction of the circuit itself is done in Python.

A quick benchmarking was performed by using `random_invertible_binary_matrix(n, seed=seed)` and timing the application of `synth_cnot_depth_line_kms` on the result, showing consistent improvement for the rust version.

```
╒═════════════════════╤═════════════╤═════════════╤═══════════════════════╕
│ Test Case           │        rust │      python │   Ratio (python/rust) │
╞═════════════════════╪═════════════╪═════════════╪═══════════════════════╡
│ 10x10 (seed 42)     │ 8.58307e-05 │ 0.00221419  │              25.7972  │
├─────────────────────┼─────────────┼─────────────┼───────────────────────┤
│ 5x5 (seed 42)       │ 9.48906e-05 │ 0.000352144 │               3.71106 │
├─────────────────────┼─────────────┼─────────────┼───────────────────────┤
│ 20x20 (seed 55)     │ 0.000299549 │ 0.00747766  │              24.9631  │
├─────────────────────┼─────────────┼─────────────┼───────────────────────┤
│ 50x50 (seed 13)     │ 0.00281119  │ 0.0665068   │              23.6579  │
├─────────────────────┼─────────────┼─────────────┼───────────────────────┤
│ 60x60 (seed 55)     │ 0.00581484  │ 0.119858    │              20.6124  │
├─────────────────────┼─────────────┼─────────────┼───────────────────────┤
│ 70x70 (seed 13)     │ 0.00826478  │ 0.168697    │              20.4115  │
├─────────────────────┼─────────────┼─────────────┼───────────────────────┤
│ 100x100 (seed 1089) │ 0.0262463   │ 0.439383    │              16.7407  │
├─────────────────────┼─────────────┼─────────────┼───────────────────────┤
│ 120x120 (seed 17)   │ 0.0428403   │ 0.675542    │              15.7688  │
├─────────────────────┼─────────────┼─────────────┼───────────────────────┤
│ 150x150 (seed 99)   │ 0.0835741   │ 1.47674     │              17.6699  │
╘═════════════════════╧═════════════╧═════════════╧═══════════════════════╛
```


crates/accelerate/src/synthesis/linear/lnn.rs
```diff
@@ -0,0 +1,321 @@
...
+    mat: ArrayViewMut2<bool>,
+    a: usize,
+    b: usize,
+) {
+    cx_instructions.push((a, b));
+    _row_op(mat, a, b);
+}
+
+///     Get the instructions for a lower triangular basis change of a matrix mat.
+///     See the proof of Proposition 7.3 in [1].
```
ShellyGarion: there are extra spaces in the docstring

crates/accelerate/src/synthesis/linear/lnn.rs
```diff
@@ -0,0 +1,321 @@
...
+        } else {
+            empty_layers = 0;
+        }
+        first_qubit = 1 - first_qubit;
+    }
+    cx_instructions_rows
+}
+
+///    Transform a north-west triangular matrix to identity in depth 3*n by Proposition 7.4 of [1]
+fn _north_west_to_identity(n: usize, mut mat: ArrayViewMut2<bool>) -> InstructionList {
```
ShellyGarion, line 202: extra spaces in docstring

crates/accelerate/src/synthesis/linear/lnn.rs
```diff
@@ -0,0 +1,321 @@
...
+fn _row_sum(row_1: ArrayView1<bool>, row_2: ArrayView1<bool>) -> Option<Array1<bool>> {
+    let n = row_1.len();
+    if row_2.len() != n {
+        return None;
+    }
+    let result: Array1<bool> = (0..n).map(|i| row_1[i] ^ row_2[i]).collect();
+    Some(result)
+}
+/// Perform ROW operation on a matrix mat
+fn _row_op(mat: ArrayViewMut2<bool>, ctrl: usize, trgt: usize) {
```
ShellyGarion: we already have the function row_op here:
https://github.com/Qiskit/qiskit/blob/b98e0d0cba7311caa2704a97518509f0fd75161d/crates/accelerate/src/synthesis/linear/mod.rs#L135
if you need a non pyo3 function, please also add it to the util file.

crates/accelerate/src/synthesis/linear/lnn.rs
```diff
@@ -0,0 +1,321 @@
...
+    _add_row_or_col(mat, &false, ctrl, trgt);
+}
+
+/// Perform COL operation on a matrix mat (in the inverse direction)
+fn _col_op(mat: ArrayViewMut2<bool>, ctrl: usize, trgt: usize) {
```
ShellyGarion: we already have the function col_op here:
https://github.com/Qiskit/qiskit/blob/b98e0d0cba7311caa2704a97518509f0fd75161d/crates/accelerate/src/synthesis/linear/mod.rs#L143
 if you need a non pyo3 function, please also add it to the util file.

crates/accelerate/src/synthesis/linear/lnn.rs
```diff
@@ -0,0 +1,321 @@
...
+/// linear nearest neighbor (LNN) connectivity.
+/// The depth of the circuit is bounded by 5*n, while the gate count is approximately 2.5*n^2
+///
+/// References:
+///     [1]: Kutin, S., Moulton, D. P., Smithline, L. (2007).
+///          Computation at a Distance.
+///          `arXiv:quant-ph/0701194 <https://arxiv.org/abs/quant-ph/0701194>`_.
+
+type InstructionList = Vec<(usize, usize)>;
+fn _row_sum(row_1: ArrayView1<bool>, row_2: ArrayView1<bool>) -> Option<Array1<bool>> {
```
ShellyGarion: I think this function should go into the util file:
https://github.com/Qiskit/qiskit/blob/main/crates/accelerate/src/synthesis/linear/utils.rs

could you write a docstring explaining what this function does?

crates/accelerate/src/synthesis/linear/lnn.rs
```diff
@@ -0,0 +1,321 @@
...
+///     See the proof of Proposition 7.3 in [1].
+fn _get_lower_triangular(
+    n: usize,
+    mat: ArrayView2<bool>,
+    mat_inv: ArrayView2<bool>,
+) -> (Array2<bool>, Array2<bool>) {
+    let mut mat = mat.to_owned();
+    let mut mat_t = mat.to_owned();
+    let mut mat_inv_t = mat_inv.to_owned();
+
```
ShellyGarion, line 56: are the to_owned needed ?

crates/accelerate/src/synthesis/linear/lnn.rs
```diff
@@ -0,0 +1,321 @@
...
+    let mut cx_instructions_rows: InstructionList = Vec::new();
+    // Use the instructions in U, which contains only gates of the form cx(a,b) a>b
+    // to transform the matrix to a permuted lower-triangular matrix.
+    // The original Matrix is unchanged.
+    for i in (0..n).rev() {
```
ShellyGarion, line 62: can we replace some for loops by iterators? (I don't know how, just asking...)

crates/accelerate/src/synthesis/linear/lnn.rs
```diff
@@ -0,0 +1,321 @@
...
+        _row_op(mat_t.view_mut(), ctrl, trgt);
+        _col_op(mat_inv_t.view_mut(), ctrl, trgt);
+    }
+    (mat_t, mat_inv_t)
+}
+
+/// For each row in mat_t, save the column index of the last "1"
+fn _get_label_arr(n: usize, mat_t: ArrayView2<bool>) -> Vec<usize> {
+    let mut label_arr: Vec<usize> = Vec::new();
+    for i in 0..n {
```
ShellyGarion: can we replace some for loops by iterators? (I don't know how, just asking...)

ShellyGarion (commented):
Thanks @gadial for your contribution! I only have some minor comments and questions.

ShellyGarion:
I also think we can add some release notes saying that there was a performance improvement in synth_cnot_depth_line_kms since it's implementation was ported to rust

crates/accelerate/src/synthesis/linear/lnn.rs
```diff
@@ -0,0 +1,321 @@
...
+use pyo3::prelude::*;
+
+/// Optimize the synthesis of an n-qubit circuit contains only CX gates for
+/// linear nearest neighbor (LNN) connectivity.
+/// The depth of the circuit is bounded by 5*n, while the gate count is approximately 2.5*n^2
+///
+/// References:
+///     [1]: Kutin, S., Moulton, D. P., Smithline, L. (2007).
+///          Computation at a Distance.
+///          `arXiv:quant-ph/0701194 <https://arxiv.org/abs/quant-ph/0701194>`_.
```
alexanderivrii: I think you mean this as a regular comment, and not as a docstring comment for `InstructionList`. If so, it should preceded by `//`, not by `///`. In addition, it may make sense to move the reference to the paper to the docstring of `optimize_cx_circ_depth_5n_line`.

crates/accelerate/src/synthesis/linear/lnn.rs
alexanderivrii: A small question: In the case that the two rows have different lengths, would it make sense to return an `Error` rather than `None` (that is, returning a `Result` rather than `Option`)?

crates/accelerate/src/synthesis/linear/lnn.rs
alexanderivrii: I agree, we aggressively call `copy` in the Python code, but it might be a good time to revisit if this excessive copying is really required.

crates/accelerate/src/synthesis/linear/lnn.rs
alexanderivrii: We definitely can, but I think this is a good tradeoff between code clarity and code "rustiness". 😄

crates/accelerate/src/synthesis/linear/lnn.rs
alexanderivrii: just for fun, you could probably do something like this (I am not sure this compiles or returns the correct result):

```rust
fn _get_label_arr(n: usize, mat_t: ArrayView2<bool>) -> Vec<usize> {
    mat
      .rows()
      .into_iter()
      .map(|i|i.iter().rposition(|&r| r == true).unwrap())
      .collect()
}
```

crates/accelerate/src/synthesis/linear/lnn.rs
```diff
@@ -0,0 +1,321 @@
...
+///     [1, 0, 0, 0]]
+
+///     According to [1] the synthesis is done on the inverse matrix
+///     so the matrix mat is inverted at this step
+#[pyfunction]
+#[pyo3(signature = (mat))]
+pub fn optimize_cx_circ_depth_5n_line(
+    _py: Python,
+    mat: PyReadonlyArray2<bool>,
+) -> PyResult<(InstructionList, InstructionList)> {
```
alexanderivrii: For the rust function to be entirely self-contained and for consistency with other synthesis functions, it would be nice to return `PyResult<CircuitData>`.

qiskit/synthesis/linear_phase/cx_cz_depth_lnn.py
```diff
@@ -33,7 +33,7 @@
 
 from qiskit.circuit import QuantumCircuit
 from qiskit.synthesis.linear.linear_matrix_utils import calc_inverse_matrix
-from qiskit.synthesis.linear.linear_depth_lnn import _optimize_cx_circ_depth_5n_line
+from qiskit._accelerate.synthesis.linear import optimize_cx_circ_depth_5n_line
```
alexanderivrii: is this change needed, i.e. shouldn't the relevant function be importable from ``qiskit.synthesis.linear.linear_depth_lnn``?

qiskit/synthesis/linear/linear_depth_lnn.py
```diff
@@ -267,7 +58,7 @@ def synth_cnot_depth_line_kms(mat: np.ndarray[bool]) -> QuantumCircuit:
     # Returns the quantum circuit constructed from the instructions
     # that we got in _optimize_cx_circ_depth_5n_line
     num_qubits = len(mat)
-    cx_inst = _optimize_cx_circ_depth_5n_line(mat)
+    cx_inst = optimize_cx_circ_depth_5n_line(mat)
     qc = QuantumCircuit(num_qubits)
```
alexanderivrii: it would be nice to have all the code to construct the quantum circuit be entirely in rust.

qiskit/synthesis/linear_phase/cx_cz_depth_lnn.py
```diff
@@ -245,7 +245,7 @@ def synth_cx_cz_depth_line_my(mat_x: np.ndarray, mat_z: np.ndarray) -> QuantumCi
     n = len(mat_x)
     mat_x = calc_inverse_matrix(mat_x)
 
-    cx_instructions_rows_m2nw, cx_instructions_rows_nw2id = _optimize_cx_circ_depth_5n_line(mat_x)
+    cx_instructions_rows_m2nw, cx_instructions_rows_nw2id = optimize_cx_circ_depth_5n_line(mat_x)
```
alexanderivrii: Ah, I see the problem: we actually want to have a method that returns both lists. So maybe we can expose two different functions from the Rust crate (the one that returns a pair of lists and the one that returns the full circuit), what do you think?

alexanderivrii (commented):
Thanks Gadi, the PR looks very nice and is quite close to the original Python code. I have left a few comments in-place. The main one is probably to rethink which functions we want to expose from the rust side. You probably want to add release notes as well.

crates/accelerate/src/synthesis/linear/lnn.rs
gadial: I think my main question here is - should I use pyo3 function? It seems like adding overhead where it's not needed (but sure, I'll move to the utils file).

crates/accelerate/src/synthesis/linear/lnn.rs
gadial: I have also thought about it. On the surface of it we need all those copies, but then again - maybe not.

`mat_inv`, in this stage, is a clone of the original matrix passed to the algorithm (it's cloned in line 305, in `optimize_cx_circ_depth_5n_line`). We pass it to `_matrix_to_north_west` and don't use it again in `_north_west_to_identity`. We also don't use it inside  `_matrix_to_north_west` other than as the input to `_get_lower_triangular` so it is possible to have `_get_lower_triangular` altering it instead of creating a copy (this is less "clean", code-wise, so I didn't rush to do it).

As to the the clonings in 65 and 66 - I don't see a way around it. `mat` is changed in `_matrix_to_north_west` and used again later so it must be cloned prior to the changes done in `_get_lower_triangular`. Inside `_get_lower_triangular` itself, we modify `mat` to generate the instruction list, and then need to perform the row operations on `mat_t` in order to obtain the label array; this cannot be done on the original `mat` or the one after the changes.

crates/accelerate/src/synthesis/linear/lnn.rs
ShellyGarion: you don't have to use the pyo3 function, but please add your new function to the util file, and see that the naming conventions are consistent (there are many functions there that have both rust and pyfunctions)

crates/accelerate/src/synthesis/linear/lnn.rs
ShellyGarion: it's worth to do it for better performance, for the clarity one can add a comment.

qiskit/synthesis/linear_phase/cx_cz_depth_lnn.py
gadial: This is not the "main" function in the sense that it is not the one that generates the circuit, but the helper method that generates the instruction lists (as lists of pairs), so I'm not sure if we should wrap it as well.

qiskit/synthesis/linear_phase/cx_cz_depth_lnn.py
gadial: Yes, I'll do this

crates/accelerate/src/synthesis/linear/lnn.rs
gadial: I actually think the iterator-based version can be easier to understand. compare
```rust
(0..i)
.rev()
.filter(|k| mat[[k, first_j]])
.for_each(|k| {
    _row_op_update_instructions(&mut cx_instructions_rows, mat.view_mut(), i, k);
});
 ```
 
 With
 ```rust
 for k in (0..i).rev() {
    if mat[[k, first_j]] {
        _row_op_update_instructions(&mut cx_instructions_rows, mat.view_mut(), i, k)
    }
```

While the latter is more concise, it's less explicit in describing what it actually does.

crates/accelerate/src/synthesis/linear/lnn.rs
gadial: Done in 4902815e1. Note I still have concerns on why `col_op` inverts the order of the ctrl and trgt; I'm not sure why it was done like this in the first place, and since we add it to the utils file I think we should avoid it altogether.

crates/accelerate/src/synthesis/linear/lnn.rs
gadial: Done in 4902815e1. See also my concern about `col_op` above.

crates/accelerate/src/synthesis/linear/lnn.rs
gadial: Done to the best of my ability in 4902815e1

crates/accelerate/src/synthesis/linear/lnn.rs
gadial: Done in 4902815e1

crates/accelerate/src/synthesis/linear/lnn.rs
gadial: Thanks. Changed in 4902815e1

crates/accelerate/src/synthesis/linear/lnn.rs
gadial: Done in 4902815e1

qiskit/synthesis/linear/linear_depth_lnn.py
gadial: Done in 4902815e1

qiskit/synthesis/linear_phase/cx_cz_depth_lnn.py
gadial: Done in 4902815e1

crates/accelerate/src/synthesis/linear/lnn.rs
gadial: I don't blindly trust my benchmarking results, but it seems this had a significant impact (now we're around 20 times better than the Python version, compared to 4 previously).

crates/accelerate/src/synthesis/linear/lnn.rs
gadial: I don't blindly trust my benchmarking results, but it seems this had a significant impact (now we're around 20 times better than the Python version, compared to 4 previously).

crates/accelerate/src/synthesis/linear/lnn.rs
ShellyGarion: I agree. Perhaps you could provide this function a better name to indicate that it also reverts the order.

crates/accelerate/src/synthesis/linear/lnn.rs
```diff
@@ -0,0 +1,334 @@
...
+    cx_instructions.push((a, b));
+    _row_op(mat, a, b);
+}
+
+/// Get the instructions for a lower triangular basis change of a matrix mat.
+/// See the proof of Proposition 7.3 in [1].
+fn _get_lower_triangular<'a>(
+    n: usize,
+    mat: ArrayView2<bool>,
+    mut mat_inv_t: ArrayViewMut2<'a, bool>,
```
alexanderivrii: Thanks for improving this function! A small thing that looks a bit strange is that `mat_inv_t` is both passed as a mutable input to the function and is returned as part of the output, we probably do not need both.

Not important for the PR, but tf you remember the details of the algorithm (I personally need to take another look at the paper), could you please add a short sentence on the preconditions/postconditions of this function? Like what is the relationship between the input `mat` and `mat_inv_t` -- are these the inverses of each other? What is the relationship between the outputs?

crates/accelerate/src/synthesis/linear/lnn.rs
```diff
@@ -0,0 +1,334 @@
...
+    let mut cx_instructions_rows: InstructionList = Vec::new();
+    // Use the instructions in U, which contains only gates of the form cx(a,b) a>b
+    // to transform the matrix to a permuted lower-triangular matrix.
+    // The original Matrix mat is unchanged, but mat_inv_t is
+
+    for i in (0..n).rev() {
+        // Find the last "1" in row i, use COL operations to the left in order to
+        // zero out all other "1"s in that row.
+        let cols_to_update: Vec<usize> = (0..n).rev().filter(|&j| mat[[i, j]]).collect();
+        let (first_j, cols_to_update) = cols_to_update.split_first().unwrap();
```
alexanderivrii, line 66: oh nice! I didn't know about `split_first` before :)

crates/accelerate/src/synthesis/linear/lnn.rs
```diff
@@ -0,0 +1,334 @@
...
+        cols_to_update.iter().for_each(|j| {
+            _col_op_inv(mat.view_mut(), *j, *first_j);
```
alexanderivrii: Would it make sense to replace `_col_op_inv` by `_col_op` and swap the order of the arguments? (I know this is how it was done in the Python code, and badly named at that).

crates/accelerate/src/synthesis/linear/lnn.rs
```diff
@@ -0,0 +1,334 @@
...
+    }
+    (mat_t, mat_inv_t)
+}
+
+/// For each row in mat_t, save the column index of the last "1"
+fn _get_label_arr(n: usize, mat_t: ArrayView2<bool>) -> Vec<usize> {
+    (0..n)
+        .map(|i| (0..n).find(|&j| mat_t[[i, n - 1 - j]]).unwrap_or(n))
+        .collect()
+}
```
alexanderivrii, line 90: nice!

crates/accelerate/src/synthesis/linear/lnn.rs
```diff
@@ -0,0 +1,334 @@
...
+    row: ArrayView1<bool>,
+    k: usize,
+) -> bool {
+    // Find the linear combination of mat_t rows which produces "row"
+    !(0..row.len())
+        .filter(|&row_l| row[row_l])
+        .fold(Array1::from_elem(row.len(), false), |w_needed, row_l| {
+            _row_sum(w_needed.view(), mat_inv_t.row(row_l)).unwrap()
+        })[label_arr_t[k]]
+}
```
alexanderivrii, line 105: This is really cool use of the `fold` method, out of curiosity, does this also improve the performance of the code compared to the `for-loop`?

crates/accelerate/src/synthesis/linear/lnn.rs
```diff
@@ -0,0 +1,334 @@
...
+
+/// Returns label_arr_t = label_arr^(-1)
+fn _get_label_arr_t(n: usize, label_arr: &[usize]) -> Vec<usize> {
+    let mut label_err_t: Vec<usize> = vec![0; n];
```
alexanderivrii: super minor: you probably want to call this `label_arr_t` (`a` instead of `e`). 

also, is there any case when `n` is not equal to `label_arr.len()`, in which case we don't need to pass `n` explicitly to the function?

crates/accelerate/src/synthesis/linear/lnn.rs
```diff
@@ -0,0 +1,334 @@
...
+/// so the matrix mat is inverted at this step
+
+/// References:
+/// [1]: Kutin, S., Moulton, D. P., Smithline, L. (2007).
+/// Computation at a Distance.
+/// `arXiv:quant-ph/0701194 <https://arxiv.org/abs/quant-ph/0701194>`_.
+#[pyfunction]
+#[pyo3(signature = (mat))]
+pub fn optimize_cx_circ_depth_5n_line(
+    _py: Python,
```
alexanderivrii: We can probably skip passing `_py` to the function.

alexanderivrii: The function name is not particularly representative of what the function is doing (in particular, it does not `optimize` anything). Maybe something like `synth_cnot_depth_line_pair` (but please find a better name than that). Also it seems that our new agreed on convention is to prefix python interface functions with `py`, i.e. let's change the names to `py_synth_cnot_depth_line_pair` and to `py_synth_cnot_depth_line_kms`. We don't follow this convention anywhere else in the synthesis code yet, but I believe someone is working on that already.

alexanderivrii (commented):
This looks great! I have left a few minor in-place suggestions, but I don't feel strongly about any of these, so feel free to ignore, I am completely happy to approve this PR just as it is right now. @ShellyGarion, please take a look to see if you have any more comments. I am not sure if we need release notes: on the one hand, we can skip release notes if there is no API change, but on the other hand it might be nice to highlight that a certain algorithm was ported to Rust, giving 20x performance speedup.

crates/accelerate/src/synthesis/linear/lnn.rs
gadial: Yes, returning it this way was intentional so the line `let (mat_t, mat_inv_t) = _get_lower_triangular(n, mat.view(), mat_inv.view_mut());` would be clearer about what we're actually getting from `_get_lower_triangular`.

crates/accelerate/src/synthesis/linear/lnn.rs
gadial: Passing n around is pure convenience - find it once, use it in the rest instead of re-computing it via some matrix access.

crates/accelerate/src/synthesis/linear/lnn.rs
gadial: doing some test benchmarking run on my side, it doesn't seem conclusive one way or the other.

crates/accelerate/src/synthesis/linear/lnn.rs
```diff
@@ -0,0 +1,336 @@
...
+/// According to [1] the synthesis is done on the inverse matrix
+/// so the matrix mat is inverted at this step
+
+/// References:
+/// [1]: Kutin, S., Moulton, D. P., Smithline, L. (2007).
+/// Computation at a Distance.
+/// `arXiv:quant-ph/0701194 <https://arxiv.org/abs/quant-ph/0701194>`_.
+#[pyfunction]
+#[pyo3(signature = (mat))]
+pub fn py_synth_cnot_depth_line_kms(
```
ShellyGarion, line 306: this is a minor comment: both these pyfunctions have exactly the same docstring. could you add an explanation? it seems that the difference is in the args+returns. Could you please add them too (at least to the pyfunctions)?

ShellyGarion (commented):
Generally it looks good. I only have a minor comment on the docs.

ShellyGarion (approved):
LGTM, thanks @gadial !

PR #13655, opened 2025-01-13
Fix `Target.instruction_supported` when `target.num_qubits == None`

ElePT:
### Summary
This PR fixes a small bug found in `Target.instruction_supported`, where  `target.num_qubits == None` would cause `Target.instruction_supported` to always evaluate to `False`. This was handled in the [original Python implementation by artificially overwriting `qargs`](https://github.com/Qiskit/qiskit/blob/53667d167e2de2f841d3b781877427f0b459289b/qiskit/transpiler/target.py#L786-L788), but this would require a mutable variable in Rust, so I just added an additional condition to account for this case.

### Details and comments
Bug discovered through [#12850](https://github.com/Qiskit/qiskit/pull/12850).

### DIFF
--- a/crates/accelerate/src/target_transpiler/mod.rs
+++ b/crates/accelerate/src/target_transpiler/mod.rs
@@ -1163,6 +1163,7 @@ impl Target {
                         match obj {
                             TargetOperation::Variadic(_) => {
                                 return qargs.is_none()
+                                    || self.num_qubits == None
                                     || _qargs.iter().all(|qarg| {
                                         qarg.index() <= self.num_qubits.unwrap_or_default()
                                     }) && qarg_set.len() == _qargs.len();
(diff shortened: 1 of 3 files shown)


raynelfss (commented):
LGTM but before I can approve it, I have one question. Would it be easier to just discard the `qargs` argument at the beginning the way it was done in Python? You could avoid the mutable reference by just re-defining the variable, for example:

```rust
// Handle case where num_qubits is None by always checking globally supported operations
// I don't think it has to be mut, but in case it's needed.
let mut qargs: Option<Qargs> = if self.num_qubits.is_none() {
    None
} else {
    qargs
};
```

I do see why this would be less clean but it would avoid having to check the same condition multiple times.

ElePT:
> LGTM but before I can approve it, I have one question. Would it be easier to just discard the qargs argument at the beginning the way it was done in Python? You could avoid the mutable reference by just re-defining the variable.

Sure, I was in between the two approaches, and I don't really have a strong preference, so I can totally change it.

raynelfss (approved):
LGTM, I wanted to add that this was being done in `py_instruction_supported` already, but when I split off these functions to have a rust-native parallel, the condition got lost for the rust-native one.

PR #13661, opened 2025-01-14
Add option ``max_block_width`` to ``CollectLinearFunctions`` and ``CollectClifford`` passes

alexanderivrii:
### Summary

This PR adds a new argument ``max_block_width`` to the class ``BlockCollector`` and to the transpiler passes ``CollectLinearFunctions`` and ``CollectCliffords`` that are based on this class. This argument enforces block collection strategies to restrict the maximum number of qubits over which a block of nodes is defined.

This option is similar to the option ``max_block_size`` in the ``CollectMultiQBlocks`` transpiler pass (though naming-wise, I am calling this _"width"_ instead of _"size"_, which is how it should be using the terminology of ``DAGCircuit`` and ``QuantumCircuit``). 

The implementation is based on the adaptation of ``BlockCollector`` in AI resynthesis passes (including a small cleanup for updating the frontier nodes).

### Details and comments

Somewhat related to the PR, here is a more refined difference between ``BlockCollector`` and ``CollectMultiQBlocks``: 

1. ``BlockCollector`` allows to only collect nodes that match a given "filter function" and treat other nodes as "should not be collected", while ``CollectMultiQBlocks`` collects all nodes (in particular, ``CollectMultiQBlocks`` cannot be used to only collect say Clifford gates within in a larger circuit). 

2. ``BlockCollector`` can collect nodes exploiting commutativity of nodes within the DAG;  ``CollectMultiQBlocks`` does not have this functionality. ``BlockCollector`` has other options that ``CollectMultiQBlocks`` does not have.

3. Both can now collect nodes "forward-wise" from the inputs towards the ouputs of a circuit, and ``backward-wise`` from the outputs towards the inputs.

4.  Both ``BlockCollector`` and ``CollectMultiQBlocks`` are greedy, however ``CollectMultiQBlocks`` is smarter, in the following way. ``BlockCollector`` greedily collects one block at a time, greedily trying to add nodes into the block being collected. In particular it may create "disconnected" blocks . On the other hand, ``CollectMultiQBlocks`` (with the implementation based on the disjoint set union data structure) produces connected blocks by construction.


test/python/dagcircuit/test_collect_blocks.py
```diff
@@ -878,6 +878,112 @@ def test_collect_blocks_backwards_dagdependency(self):
...
+
+        # When max_block_width is None, we should still obtain 1 block
+        blocks = block_collector.collect_all_matching_blocks(
+            lambda node: True,
+            min_block_size=1,
+            max_block_width=None,
+        )
+        self.assertEqual(len(blocks), 1)
+
+        # When max_block_width=2, we should still obtain 5 blocks
```
ShellyGarion, line 910: remove "still"?

test/python/dagcircuit/test_collect_blocks.py
```diff
@@ -878,6 +878,112 @@ def test_collect_blocks_backwards_dagdependency(self):
...
+        blocks = block_collector.collect_all_matching_blocks(
+            lambda node: True,
+            min_block_size=1,
+            max_block_width=2,
+        )
+        self.assertEqual(len(blocks), 5)
+
+        # When max_block_width=3, we should still obtain 3 blocks
```
ShellyGarion: remove "still"?

test/python/dagcircuit/test_collect_blocks.py
```diff
@@ -878,6 +878,112 @@ def test_collect_blocks_backwards_dagdependency(self):
...
+        blocks = block_collector.collect_all_matching_blocks(
+            lambda node: True,
+            min_block_size=1,
+            max_block_width=3,
+        )
+        self.assertEqual(len(blocks), 3)
+
+        # When max_block_width=4, we should still obtain 2 blocks
```
ShellyGarion: remove "still"?

test/python/dagcircuit/test_collect_blocks.py
ShellyGarion: remove "still"?

test/python/dagcircuit/test_collect_blocks.py
```diff
@@ -878,6 +878,112 @@ def test_collect_blocks_backwards_dagdependency(self):
...
+        # original circuit
+        circuit = QuantumCircuit(6)
+        circuit.h(0)
+        circuit.cx(0, 1)
+        circuit.cx(1, 2)
+        circuit.cx(2, 3)
+        circuit.cx(3, 4)
+        circuit.cx(4, 5)
+
+        block_collector = BlockCollector(circuit_to_dag(circuit))
```
ShellyGarion: is this the only line which is different between this test and the next one? maybe you can write a single test with args?

ShellyGarion (commented):
LGTM. I only have some minor comments on the tests.

ShellyGarion:
what happens if you take this circuit:

```
        circuit = QuantumCircuit(5)
        circuit.cx(0, 1)
        circuit.cx(3, 4)
        circuit.cx(1, 3)
        circuit.cx(2, 4)
```

with `max_block_width=4`?
do you get 3 blocks [0,1,3], [3,4],[2,4] or 2 block [0,1][1,2,3,4]? 
do we care about this?

test/python/dagcircuit/test_collect_blocks.py
alexanderivrii: Great suggestion, I have refactored the tests in [9b4af62](https://github.com/Qiskit/qiskit/pull/13661/commits/9b4af620d6e8a50543854b35450e7e629eaf9871).

test/python/dagcircuit/test_collect_blocks.py
alexanderivrii: Regarding this and other usages of "still" -- this is implicitly done after refactoring tests in [9b4af62](https://github.com/Qiskit/qiskit/pull/13661/commits/9b4af620d6e8a50543854b35450e7e629eaf9871).

alexanderivrii:
> what happens if you take this circuit:
> 
> ```
>     circuit = QuantumCircuit(5)
>     circuit.cx(0, 1)
>     circuit.cx(3, 4)
>     circuit.cx(1, 3)
>     circuit.cx(2, 4)
> ```
> with max_block_width=4?
> do you get 3 blocks [0,1,3], [3,4],[2,4] or 2 block [0,1][1,2,3,4]?
> do we care about this?

In general, the problem of "optimally" dividing a circuit into blocks is "hard" (and in fact NP-hard for suitable notions of "optimality"). In practice, which blocks get collected depends on the order in which the nodes are processed and on other options given to ``BlockCollector`` (like exploiting commutativity between nodes or collecting from the back of the circuit). So the result is generally not unique. In fact, for the example above, with the default options, we would collect the first 3 gates into one block, and the remaining gate into the second block, and with ``collect_from_back=True``, we would collect the last 3 gates into one block, and the first gate into the second block.

I would say we should care about this when we think about _resynthesis_, as collecting different blocks gives rise to different resynthesis opportunities.

test/python/transpiler/test_clifford_passes.py
```diff
@@ -400,6 +400,29 @@ def test_collect_cliffords_default(self):
         self.assertEqual(qct.size(), 1)
         self.assertIn("clifford", qct.count_ops().keys())
 
+    def test_collect_cliffords_max_width(self):
+        """Make sure that collecting Clifford gates and replacing them by Clifford
+        works correctly when the option ``max_width`` is specified."""
```
ShellyGarion: very minor: should be "max_block_width"

alexanderivrii: done

ShellyGarion (approved):
LGTM. Thanks @alexanderivrii

PR #13667, opened 2025-01-15
Reduce non-determinism in `UnitarySynthesis` pass

ElePT:
### Summary
This PR proposes the use of an `IndexMap` in the Rust implementation of `UnitarySynthesis` to allow for a more deterministic output of the pass, and to ensure that the Python and Rust codes produce the same outputs when facing the same inputs. While this is not a guarantee we make, I believe it is good to have, makes testing much easier, and will smoothen the path for the transition to always using the target as a transpilation constraint "vehicle" (this PR fixes an issue found in a test in #12850).

### DIFF
--- a/crates/accelerate/src/unitary_synthesis.rs
+++ b/crates/accelerate/src/unitary_synthesis.rs
@@ -17,7 +17,7 @@ use std::sync::OnceLock;
 
 use approx::relative_eq;
 use hashbrown::{HashMap, HashSet};
-use indexmap::IndexMap;
+use indexmap::{IndexMap, IndexSet};
 use itertools::Itertools;
 use ndarray::prelude::*;
 use num_complex::{Complex, Complex64};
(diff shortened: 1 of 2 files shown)


ShellyGarion (approved):

crates/accelerate/src/unitary_synthesis.rs
```diff
@@ -627,8 +627,8 @@ fn get_2q_decomposers_from_target(
     }
 
     let target_basis_set = get_target_basis_set(target, qubits[0]);
-    let available_1q_basis: HashSet<&str> =
-        HashSet::from_iter(target_basis_set.get_bases().map(|basis| basis.as_str()));
+    let available_1q_basis: IndexSet<&str> =
```
mtreinish, line 630: You should use use the ahash hasher here otherwise this will cause a slowdown on lookups in the hash set.

PR #13670, opened 2025-01-15
Fix 4-pi periodicity of controlled rotations in the `CommutationChecker`

Cryoris:
### Summary

The commutation checker currently wrongly assumes that the controlled Pauli rotations CRX/Y/Z are equivalent to the identity for multiples of $2\pi$. However, this is only true for $4\pi$. Note that CP is indeed $2\pi$ periodic. 

### Details and comments

The tests were modified to compare the output of the commutation checker to a numeric result instead of a hardcoded one, such that this error shouldn't arise in the future again.


crates/accelerate/src/commutation_checker.rs
```diff
@@ -50,23 +48,37 @@ static SUPPORTED_OP: Lazy<HashSet<&str>> = Lazy::new(|| {
     ])
 });
 
-// Map rotation gates to their generators, or to ``None`` if we cannot currently efficiently
-// represent the generator in Rust and store the commutation relation in the commutation dictionary
-static SUPPORTED_ROTATIONS: Lazy<HashMap<&str, Option<OperationRef>>> = Lazy::new(|| {
+// Map rotation gates to their generators (or to ``None`` if we cannot currently efficiently
+// represent the generator in Rust and store the commutation relation in the commutation dictionary)
+// and their pi-periodicity. E.g. RX is generated by X and 2-pi periodic, while CRX is generated
+// by CX and 4-pi periodic.
```
alexanderivrii: As discussed offline, can you please add that periodicity refers to "equal to the identity up to a global phase"

crates/accelerate/src/commutation_checker.rs
```diff
@@ -636,12 +648,11 @@ fn map_rotation<'a>(
     tol: f64,
 ) -> (&'a OperationRef<'a>, &'a [Param], bool) {
     let name = op.name();
-    if let Some(generator) = SUPPORTED_ROTATIONS.get(name) {
-        // If the rotation angle is below the tolerance, the gate is assumed to
-        // commute with everything, and we simply return the operation with the flag that
-        // it commutes trivially.
```
alexanderivrii: I think this is a helpful comment, why was it removed?

test/python/circuit/test_commutation_checker.py
```diff
@@ -433,26 +434,25 @@ def test_cutoff_angles(self, gate_cls):
                 self.assertFalse(scc.commute(generic_gate, [0, 1], [], gate, qargs, []))
 
     @idata(ROTATION_GATES)
-    def test_rotation_mod_2pi(self, gate_cls):
-        """Test the rotations modulo 2pi commute with any gate."""
+    def test_controlled_rotation_mod_4pi(self, gate_cls):
+        """Test the rotations modulo 2pi (controlled 4pi) commute with any gate."""
```
ElePT: Nitpick, maybe we could specify/hint that cp is still 2pi:

```suggestion
        """Test the rotations modulo 2pi (controlled rx, ry, rz -> 4pi) commute with any gate."""
```

ElePT (commented):
LGTM! This is the typical tricky bug, I just left a tiny suggestions for the test docstring.

releasenotes/notes/fix-4pi-periodic-commutations-3b89d1813513f613.yaml
```diff
@@ -0,0 +1,9 @@
+---
+fixes:
+  - |
+    The :class:`.CommutationChecker` did not handle commutations of the :class:`.CRXGate`,
+    :class:`.CRYGate` and :class:`.CRZGate` correctly for angles
+    :math:`\pi(4k + 2)` for :math:`k \in \mathbb Z`.
+    In these cases, the controlled rotations were falsely assumed to commute with any gate.
+    Now these gates correctly commute with any gate if the rotation angle is a multiple of
+    :math:`4\pi`.
```
alexanderivrii, line 9: Now we are using the generator to check commutativity correct?

BTW, is it possible that a CRZ-gate with a very specific angle of :math:`2*\pi` commutes with some other gate, but we don't see it when we just examine the generator?

alexanderivrii (commented):
Thank you for identifying and fixing the bug. I have left a few tiny comments.

crates/accelerate/src/commutation_checker.rs
Cryoris: this happened by accident 🙂

releasenotes/notes/fix-4pi-periodic-commutations-3b89d1813513f613.yaml
Cryoris: Yes, we compare to a matrix-based commutation check 🙂

crates/accelerate/src/commutation_checker.rs
Cryoris: added again in 59dbf24

crates/accelerate/src/commutation_checker.rs
Cryoris: added in 59dbf24

test/python/circuit/test_commutation_checker.py
Cryoris: added in 59dbf24

alexanderivrii (approved):
LGTM! Thanks for addressing the comments.

PR #13679, opened 2025-01-17
Fixes issue #12977

### LINKED ISSUE #12977: Update MCX synthesis to use `PhaseGate` instead of `U1Gate`
### Summary

We should update the synthesis methods to use the `PhaseGate` instead of the IBM-legacy `U1Gate`, if possible.
See https://github.com/Qiskit/qiskit/pull/12961#discussion_r1719719621.

Ak-ash22:
### Summary
Updates the synthesis methods to use the PhaseGate instead of the IBM-legacy U1Gate.

### Details and comments
Substituted the `CU1/MCU1Gate` for the `CPhase/MCPhaseGate` in the `qiskit/synthesis/multi_controlled/mcx_synthesis.py` file.


Cryoris:
Thanks for opening the draft 🙂 The tests are failing because they were expecting the `u1` gates (or `cu1`/`mcu1`) and now the phase gates are used. You can just update those tests to now expect the phase gates.

Could you include a releasenote which mentions this gate change for certain MCX gates under the `upgrades` section?

Ak-ash22:
Hey @Cryoris ... can you help me navigate through these tests that are breaking? For ex:
1) One of the test that breaks is `test_decompose_mixture_of_names_and_labels()` , and the only possible line of code I could track here is this `self.assertEqual(dag.op_nodes()[5].name, "cu1")`, am I just supposed to change "cu1" to "cphase"?
2) Also in `test_mcx_gate_variants()`, you would expect me to change the `expected_qasm()` line, to expect cphase instead, right?

I just wanted to confirm before I went ahead with these, as this is my second time contributing.

alexanderivrii:
Hi @Ak-ash22, you are correct: you need to modify the failing tests to refer to CPhase instead of CU1. 

For `test_decompose_mixture_of_names_and_labels()`: the name of the CPhase gate is `"cp"`, so you should change the test to `self.assertEqual(dag.op_nodes()[5].name, "cp")`. 

For `test_mcx_gate_variants()`: you should indeed alter the `expected_qasm` string.

In case you do not know this already, you can click on "Details" right next to the "Tests/macOS-..." line above, taking you to the "Run tests" log, which shows both how many and which tests have failed, and also the specific error messages (right at the bottom of the log). For `test_decompose_mixture_of_names_and_labels()` you see
```
    AssertionError: 'cp' != 'cu1'
- cp
+ cu1
```
which indeed indicates that the gate name's is `"cp"`.

Ak-ash22:
Hello @alexanderivrii @Cryoris .. Can you help me solve this particular issue?... I have modified the failed tests to use `cp` instead of `cu1` which was successful for 2 of the tests. 
The other 2 tests, where I have to alter the `expected_qasm` string is not being solved. 

The `expected_qasm` string has `cu1` and also `mcu1`. I tried changing both to `cp` and `mcp` which didnt work. I also tried changing only `cp` to `cu1`, which is not working as well.

Cryoris:
These QASM output files can change substantially because it now dumps the decomposition via the MC-Phase path. So we don't only have to change the definition of `mcu1` to `mcp` but also the entire decomposition of the `mcp`, which is not a native gate in QASM. Since all other tests pass, you can just print the QASM output of the current circuit and update the expected QASM accordingly.

Ak-ash22:
Hello @alexanderivrii  @Cryoris .. It took me a while but I managed to correct the `expected_qasm` string.
The main tests all seem to be running successfully, but now there is an issue with the lint test, although it rates 10/10 for the code.

The issue seems to be that some lines of expected_qasm strings are too long. Can you help me fix this one last issue?

ShellyGarion:
@Ak-ash22 - since it has been a while, could you please rebase your PR on qiskit main branch and we'll see if there are still some lint/docs errors?

ShellyGarion:
I'm closing this since it seems that this work was already done in other PRs.

PR #13680, opened 2025-01-17
Unbox the `DAGCircuit::op_nodes` iterator

jakelishman:
This allows the iterator to be (reliably) properly monomorphised into its usage sites.  Since the dynamic `include_directives` call is almost certainly given as a literal, I would strongly expect that even the trivial overhead of the short-circuiting `||` operator will be compiled out in all cases.


ElePT (approved):
The change makes sense to me, especially in the context of #13681, but I will not merge it to give the chance to other reviewers (with stronger opinions) to take a look.

kevinhartman (approved):
Looks good to me as well.

PR #13681, opened 2025-01-17
Include `&PackedInstruction` in `DAGCircuit::op_nodes` iteration

jakelishman:
Uses of the `DAGCircuit::op_nodes` iterator are almost invariably followed by indexing into the DAG to retrieve the node weight, which is dynamically asserted to be `NodeType::Operation`.  Since the iterator already verifies this, we can just include the unwrapped `&PackedInstruction` in the iterator return to simply calling code and reduce the number of `unreachable!()`, `panic!()`, etcs.

A new `op_node_indices` is provided for ease for cases where the iterator needs to be consumed to avoid being lifetime-tied to the DAG, such as when the DAG is going to be mutated based on the nodes.

`topological_op_nodes` is left for a follow-up; it might be slightly more challenging because of the interaction with rustworkx, but it can also be done separately anyway.

### Details and comments

Built on top of #13680.

`DAGCircuit::op_nodes_by_py_type` is unused since #13040, so I just removed the method rather than modifying it.

### DIFF
--- a/crates/accelerate/src/barrier_before_final_measurement.rs
+++ b/crates/accelerate/src/barrier_before_final_measurement.rs
@@ -30,24 +30,21 @@ pub fn barrier_before_final_measurements(
     dag: &mut DAGCircuit,
     label: Option<String>,
 ) -> PyResult<()> {
+    let is_exactly_final = |inst: &PackedInstruction| FINAL_OP_NAMES.contains(&inst.op.name());
     let final_ops: HashSet<NodeIndex> = dag
         .op_nodes(true)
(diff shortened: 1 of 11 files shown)


kevinhartman (commented):
Looks great! This improves the ergonomics of `op_nodes` substantially, and is something that got lost amongst other technical debts in the original port of DAGCircuit. Thanks for addressing it.

I'll sign off once the PR this depends on goes in.

jakelishman:
Now rebased over `main`.

kevinhartman (approved):

PR #13682, opened 2025-01-17
Build Linux AArch64 wheels natively without QEMU

jakelishman:
GitHub Actions now provides Linux runner images on native arm64 hardware[^1], which means we should be able to build the wheels natively, without requiring QEMU.  This potentially paves the way (pending a more complete move of CI to GHA) for Linux on AArch64 to potentially gain Tier 1 platforma support.

[^1]: https://github.blog/changelog/2025-01-16-linux-arm64-hosted-runners-now-available-for-free-in-public-repositories-public-preview/

### DIFF
--- a/.github/workflows/wheels-build.yml
+++ b/.github/workflows/wheels-build.yml
@@ -181,8 +181,6 @@ jobs:
           python-version: ${{ inputs.python-version }}
       - uses: dtolnay/rust-toolchain@stable
       - uses: pypa/cibuildwheel@v2.22.0
-        env:
-          CIBW_ARCHS_LINUX: aarch64
       - uses: actions/upload-artifact@v4
         with:
           path: ./wheelhouse/*.whl
--- a/.github/workflows/wheels-build.yml
+++ b/.github/workflows/wheels-build.yml
@@ -173,16 +173,13 @@ jobs:
   wheels-linux-aarch64:
     name: "Wheels / Linux AArch64"
     if: (inputs.wheels-linux-aarch64 == 'default' && inputs.default-action || inputs.wheels-linux-aarch64) == 'build'
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-24.04-arm
     steps:
       - uses: actions/checkout@v4
       - uses: actions/setup-python@v5
         with:
           python-version: ${{ inputs.python-version }}
       - uses: dtolnay/rust-toolchain@stable
-      - uses: docker/setup-qemu-action@v3
-        with:
-          platforms: all
       - uses: pypa/cibuildwheel@v2.22.0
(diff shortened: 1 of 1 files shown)


.github/workflows/wheels-build.yml
```diff
@@ -173,16 +173,13 @@ jobs:
...
     name: "Wheels / Linux AArch64"
     if: (inputs.wheels-linux-aarch64 == 'default' && inputs.default-action || inputs.wheels-linux-aarch64) == 'build'
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-24.04-arm
     steps:
       - uses: actions/checkout@v4
       - uses: actions/setup-python@v5
         with:
           python-version: ${{ inputs.python-version }}
       - uses: dtolnay/rust-toolchain@stable
```
mtreinish, line 182: I can't comment on L186 below but I think we can remove that line and use the default test command. Tier 2 we define as running the tests during build/release but not on PR CI. We limited it to just `test.python.compiler.test_transpiler` in the interest of runtime because the emulation was so slow. But if we have a native host we should have the time budget to run the full test suite which is more in spirit of what we have defined as tier 2 support.

jakelishman: Done in 2b394e4099, let's see what happens.

.github/workflows/wheels-build.yml
```diff
@@ -173,20 +173,16 @@ jobs:
...
-      - uses: docker/setup-qemu-action@v3
-        with:
-          platforms: all
       - uses: pypa/cibuildwheel@v2.22.0
         env:
           CIBW_ARCHS_LINUX: aarch64
```
mtreinish: We probably can also remove this line since it's implicitly the host arch. We only needed this to enable the emulation when running from x86.

jakelishman: The tests of Pythons 3.9 and 3.10 had passed, so I've pushed up cd85bf783d that removes this explicit selection.

jakelishman:
It's all good - Linux AArch64 passed all tests.  After #13474, there's a high chance that we can promote Linux AArch64 to Tier 1 platform support.

mtreinish (approved):
This LGTM, I'm excited to potentially promote linux aarc64 to tier 1 in 2.0. I think when we do decided to bump it up to tier 1 we should just add this runner to the tier 1 job definition to enable pgo etc since the config is now essentially the same as the normal ubuntu latest job. But for the time being keeping it separate to not enable pgo makes sense to me.

jakelishman:
Yeah, I deliberately didn't move it have PGO in the build process here because I somewhat considered that a Tier 1 thing.

PR #13683, opened 2025-01-17
Implement `Index<NodeIndex>` for `DAGCircuit`

jakelishman:
### Summary

This removes the syntax noise of the `dag.dag()` calls when indexing by `NodeIndex`.  As it happens, this is _almost_ all of the reason we even use the underlying graph object in `accelerate`. The only exceptions are some needless defensive programming in `RemoveDiagonalGatesBeforeMeasure` (which really is the same thing underneath anyway), and in the graphviz utilities, which is legitimate.

### Details and comments

Built on #13681, since I'm trying to incrementally improve the ergonomics of DAG access here.


jakelishman:
Now rebased over main.

kevinhartman (approved):
LGTM, thanks!

PR #13685, opened 2025-01-17
Stop shuffling coupling map node indices in VF2 passes

mtreinish:
### Summary

This commit updates the preset pass manager construction usage of the VF2Layout and VF2PostLayout to stop shuffling the coupling map nodes by default. The theory behind the node shuffling is that since we limit the search space in the interest of runtime shuffling the node indices would change the search order to hopefully find a match that would be otherwise missed because we hit the internal state visit limit. However, this is showing in practice not to having a huge impact, especially since we're using the ordering heuristic from vf2++ that orders nodes by degree for the vf2 search. However, in the case of a circuit that was hardware efficient this can have the negative effect of making it harder for vf2 to find potential matches, especially on regular lattices. For example, in cases of a square grid lattice coupling map, and a path interaction graph (e.g. 0->1->2->3) the shuffling makes it much harder to find the mapping. This is because the lattice graphs the node degree is the same (or fall into the same few types of nodes) so the influence of the vf2++ heuristic isn't as significant and the index order has a larger impact because it is the search order for vf2. For smaller graphs this wasn't as noticeable but as devices scale up this effect has more of an impact.

Since we rely solely on VF2 to find a perfect layout at higher optimization levels this shuffling is not desirable because we always want to find the perfect layout if it exists, especially for hardware efficient circuits that are constructed to not require swaps. So prioritizing the results for hardware efficient circuits is desirable by default. Especially since most connectivity graphs are lattices and will exhibit the negative impacts for hardware efficient circuits.

From an API impact perspective this doesn't change any of the interfaces or defaults for the VF2 passes in the interest of backwards compatibility. The only change is that this updates how we instantiate the VF2 passes to always use a deterministic node ordering independent of any user specified seed. This will be fully deterministic even in cases the user specifies a seed value for the transpilation, the output just might not be the same as before with the fixed seed; which is not guaranteed between releases.

### Details and comments

This is a partial backport of #13492 without the API implications on the generate function from the original PR because that is only valid for 2.0. The change in what the preset pass managers do when running VF2 is valid though because it doesn't change our API guarantees around the transpilation, it just increases the likelihood of a match being found.

### DIFF
--- a/qiskit/transpiler/preset_passmanagers/builtin_plugins.py
+++ b/qiskit/transpiler/preset_passmanagers/builtin_plugins.py
@@ -238,7 +238,6 @@ class BasicSwapPassManager(PassManagerStagePlugin):
 
     def pass_manager(self, pass_manager_config, optimization_level=None) -> PassManager:
         """Build routing stage PassManager."""
-        seed_transpiler = pass_manager_config.seed_transpiler
         target = pass_manager_config.target
         coupling_map = pass_manager_config.coupling_map
         backend_properties = pass_manager_config.backend_properties
@@ -257,7 +256,7 @@ class BasicSwapPassManager(PassManagerStagePlugin):
                 routing_pass,
                 target,
                 coupling_map=coupling_map,
-                seed_transpiler=seed_transpiler,
+                seed_transpiler=-1,
                 use_barrier_before_measurement=True,
             )
         if optimization_level == 1:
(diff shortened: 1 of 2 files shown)


alexanderivrii (commented):
LGTM. My only question is we want release notes to highlight this change (I believe we might not need these, given that there was no API change, but I would like to double-check that this is indeed the case).

ElePT (approved):
I think that the release note is not necessary if there isn't any API change, so LGTM

PR #13686, opened 2025-01-17
Use native bits and registers for `CircuitData`.

### LINKED ISSUE #13269: Make rust Qubit and Clbit the source of truth in a circuit
Currently we have a Python space `Qubit` and `Clbit` class which is the source of truth of bits in a circuit. We also have a rust space `Qubit` and `Clbit` struct which are indices representing the order of qubits in the circuit but those indices map back to a `Qubit` python space object which is used from python to address the bits in the circuit. But this relationship is backwards from a performance perspective because it means we always need to create a python object to handle bits, even when creating a circuit in rust. This leads to a significant performance overhead for things like synthesis functions that are generating circuits where the overhead to create bit objects and map them to indices so we have a rust native bit object to use. We should invert this relationship so we have the pure rust bit as the source of truth and the python space object serves as a reference to that.

raynelfss:
### Summary

The following commits add native bits and register usage to `CircuitData`. This is the first public attempt at implementing these structures and is open to feedback.

### Details and comments

As we continue to move towards having more rust-native components as part of our core data model, we need to continue to promote more functionality to be able to work without needing intervention from Python. #13269 describes the cost paid whenever an instance of `CircuitData` and `DAGCircuit` needs to be constructed from Rust. There is a constant need for calling back to Python to create instances of `Qubit` and `Clbit` as needed.

An important point made towards this is to transfer the ownership of `Bits` and `Registers` to the circuit, but also to keep the current functionality of this API. Meaning that, instead of removing the dependency completely, we had to shift the source of truth to rust. The following commits achieve this by performing the following:

#### Adding representations of `QuantumRegister` and `ClassicalRegister` in Rust.

A simple representation using an `IndexSet`, and attributes such as `name` and `size`.

#### Modifying `BitData` currently `NewBitData` to also store register information and dynamically initialize the Python counter parts as requested:

- Registers are now kept in a `Vec<_>` and can be accessed by index or by key (`RegisterKey`).
- Bits are still referred to as either `Qubit(u32)` and `Clbit(u32)` and there exists a mapping between Python `Bits` and the specified index type.
    - The mapping has to be inside of an `RWLock` due to the dynamic initialization of Python `Bit`s.
- A new attribute: `BitInfo` keeps a list, organized by the bit's index, which keeps track of:
    - Whether a bit was added by a register into the circuit or not.
    - Which registers and which index within said register(s) does that bit belong in.
- Separate python specific and rust methods:
    - Python methods will perform dynamic initialization of bits and registers in Python when needed.
- Both `Bit`s and `Register`s from Python are stored as a `OnceLock` which gets dynamically initialized upon request based on the information stored here.
#### Changes to `CircuitData`

A new Python-free constructor `CircuitData::new()` was created to allow initialization of Rust instances without any need for interaction with Python unless otherwise requested. 
- Modifications were also added to the `CircuitData::with_capacity` constructor to also add the bits dynamically without accessing Python.
    - This constructor will still need a `py` token due to assignments to the `ParameterTable`.

### Known issues and drawbacks
- `copy()` can take up to twice as long to perform a copy of the circuit compared to previous iterations. This most likely due re-mapping of the registers.
- Risks with using `OnceLock` and `RWLock` as these can panic under the wrong circumstances.
- Python `Bit`s are fully opaque to the Rust ones and there's no validation other than the mapping and information stored in Rust.
- `DAGCircuit` does not use any of the new additions.
    - This is mostly due to the robust nature of the changes and a follow-up PR will add these changes to the `DAGCircuit`.
- Testing in Rust does not allow us to check if conversion works as needed into a `QuantumCircuit`.

### Tasks:
- [x] Implement a new `BitData` struct to store registers.
- [x] Implement the new `BitData` in `CircuitData`.
- [x] Fix all the broken tests.
- [x] Add rust-native tests to check if this is working properly.
- [ ] Implement partial python cross testing.

### Questions or comments?
Feel free to leave a review with any questions or concerns you might have.


crates/circuit/src/bit.rs
```diff
@@ -0,0 +1,65 @@
+use std::fmt::Debug;
+
+/// Keeps informatyion about where a bit is located within the circuit.
```
to24toro:
```suggestion
/// Keeps information about where a bit is located within the circuit.
```

crates/circuit/src/bit.rs
```diff
@@ -0,0 +1,65 @@
...
+///
+/// This information includes whether the bit was added by a register,
+/// which register it belongs to and where it is located within it.
+#[derive(Debug, Clone, PartialEq, PartialOrd, Ord, Eq, Hash)]
+pub struct BitInfo {
+    added_by_reg: bool,
+    registers: Vec<BitLocation>,
+}
```
to24toro, line 23: The `BitInfo` struct and its `registers` field may be somewhat ambiguous in conveying their true purpose.It appears that `BitInfo` primarily tracks how bits are associated with registers.

You might consider renaming this `BitInfo` to something more indicative of its focus on bit-register relationships or placements.

However, if you assume how to use it in another way, it is not necessary to rename this field.

crates/circuit/src/bit_data.rs
```diff
@@ -230,3 +235,602 @@ where
         self.bits.clear();
     }
 }
+
+#[derive(Clone, Debug)]
+pub struct NewBitData<T: From<BitType>, R: Register + Hash + Eq> {
```
to24toro, line 241: [Q]  Both `BitData<T>` and `NewBitData<T, R>`, which both manage and track bit-related information. However, it’s not entirely clear how they are intended to coexist. Are they meant to be used in parallel for different scenarios(is `BitData<T>` for dag-circuit?), or is there a plan to eventually replace `BitData<T>` with `NewBitData<T, R>`? Any clarification on whether `BitData<T>` is deprecated or if both are needed for distinct reasons would be greatly appreciated.

crates/circuit/src/bit_data.rs
```diff
@@ -230,3 +235,602 @@ where
...
+        else if let Some(bit_info) = self.bit_info[index_as_usize].orig_register_index() {
+            // If it is not initalized and has a register, initialize the original register
+            // and retrieve it from there the first time
+            if self.bits[index_as_usize].get().is_none() {
+                // A register index is guaranteed to exist in the instance of `BitData`.
+                let py_reg = self.py_get_register(py, bit_info.register_index())?;
+                let res = py_reg.unwrap().bind(py).get_item(bit_info.index())?;
+                self.bits[index_as_usize]
+                    .set(res.into())
+                    .map_err(|_| PyRuntimeError::new_err("Could not set the OnceCell correctly"))?;
```
to24toro: Instead of simply saying "Could not set the OnceCell correctly", you might include any relevant internal state or the specific condition that led to the failure. This added detail will greatly aid in diagnosing issues.

to24toro (commented):
Thank you for sharing this pull request. I understand that this is currently marked as a WIP and a draft, so I was initially hesitant to leave comments. However, I do have a few questions and suggestions that I hope will be helpful as the work progresses.

crates/circuit/src/bit.rs
raynelfss: You got the intended use correctly. I'm open to renaming this, since it is WiP you can leave any suggestions.

crates/circuit/src/bit_data.rs
raynelfss: `NewBitData` here is meant to replace `BitData`, the only reason why I'm keeping the old one around still is because the new replacement methods will need a bigger rewrite of `DAGCircuit`'s methods. Which I plan on doing on a separate PR due to how big it is.

crates/circuit/src/bit_data.rs
raynelfss: I agree with this.

1ucian0:
closing in favor of #13860

PR #12779, opened 2024-07-17
[WIP] Improve controlled UnitaryGate() (replace #10320)

ShellyGarion:
### Summary
close #10313 
this PR replaces #10320
co-authored by @ewinston 

### Details and comments
In PR #10320 there were some tests failures due to numerical errors. We start a new PR to see if these problems occur again.

PR #10320, opened 2023-06-21
improve controlled UnitaryGate (issue #10313)

ewinston:
### Summary
resolves #10313

PR #13688, opened 2025-01-18
block diagonal optimized quantum shannon decomposition

### LINKED ISSUE #10313: UnitaryGate.control implementation high gate count
### What should we add?

It seems UnitaryGate.control is producing higher gate counts than is necessary. The implementation is currently using Isometry for the definition but should probably switch to `qs_decomposition`.  ~Using qs_decomposition may also help with resolve issue #10311.~

| | nqubits | Isometry| QSD|
| ---- | ---- | ---- | --- |
|N(u) | 4|224| 163|
|N(cx)|4 | 192| 100|
|N(u) | 5|697| 1017|
|N(cx)|5 | 928| 444|
|N(u)| 6 |4381| 2886|
|N(cx)|6 | 4147| 1868|

ewinston:
### Summary
This PR attempts to detect when the unitary matrix has a block diagonal representation with respect to one of it's qubits at each step of the recursion in the QSD. If this occurs, one may skip the cosine-sine decomposition and therefore reduce the CX gate count. In addition this may help avoid numerical errors the underlying scipy.linalg.cossin function seems to have for block-diagonal input matrices.


ShellyGarion:
should hopefully close #10313 and replace #12779 (#10320)

jakelishman:
Glancing quickly, I think this PR is implementing a special case of a question something like "is there a subset of target qubits for which the partial trace commutes with $Z^{(i)}$?".  That _feels_ like something we could tackle by Hilbert-space re-ordering at the top level of the decomposition, so we can get them out of the way earlier and reduce the difficulty at _all_ levels.

I want to stress how little I've thought this through, so I might just be chatting nonsense.

ShellyGarion:
> Glancing quickly, I think this PR is implementing a special case of a question something like "is there a subset of target qubits for which the partial trace commutes with Z ( i ) ?". That _feels_ like something we could tackle by Hilbert-space re-ordering at the top level of the decomposition, so we can get them out of the way earlier and reduce the difficulty at _all_ levels.
> 
> I want to stress how little I've thought this through, so I might just be chatting nonsense.

@jakelishman - We have tried to extend this suggestion to the case of multi-controlled gates, but it seems that the algorithm of Shende et. al. decomposes the unitary gate using uniformly-controlled Rz or Ry gates, and this structure is not preserved in the induction step. Any concrete suggestions would be helpful.

ShellyGarion:
Comparing QSD to multi-controlled synthesis for 1-qubit gates (see: `test_mc_1qubit_opt`):
Usually it's better to use the multi-controlled synthesis directly, except of a few cases (with 3,4,5 qubits and p or u basis gate).

```
num_qubits= 3 base_gate= x cx count using mc: OrderedDict({'u': 8, 'cx': 6}) cx count using qsd: OrderedDict({'u': 15, 'cx': 8})
num_qubits= 3 base_gate= p cx count using mc: OrderedDict({'u': 7, 'cx': 6}) cx count using qsd: OrderedDict({'u': 10, 'cx': 6})
num_qubits= 3 base_gate= u cx count using mc: OrderedDict({'u': 25, 'cx': 22}) cx count using qsd: OrderedDict({'u': 20, 'cx': 10})
num_qubits= 4 base_gate= x cx count using mc: OrderedDict({'u': 16, 'cx': 14}) cx count using qsd: OrderedDict({'u': 77, 'cx': 47})
num_qubits= 4 base_gate= p cx count using mc: OrderedDict({'u': 24, 'cx': 20}) cx count using qsd: OrderedDict({'u': 18, 'cx': 14})
num_qubits= 4 base_gate= u cx count using mc: OrderedDict({'u': 56, 'cx': 51}) cx count using qsd: OrderedDict({'u': 87, 'cx': 51})
num_qubits= 5 base_gate= x cx count using mc: OrderedDict({'u': 41, 'cx': 36}) cx count using qsd: OrderedDict({'u': 396, 'cx': 238})
num_qubits= 5 base_gate= p cx count using mc: OrderedDict({'u': 53, 'cx': 44}) cx count using qsd: OrderedDict({'u': 34, 'cx': 30})
num_qubits= 5 base_gate= u cx count using mc: OrderedDict({'u': 109, 'cx': 92}) cx count using qsd: OrderedDict({'u': 388, 'cx': 236})
num_qubits= 6 base_gate= x cx count using mc: OrderedDict({'u': 97, 'cx': 84}) cx count using qsd: OrderedDict({'u': 1642, 'cx': 1012})
num_qubits= 6 base_gate= p cx count using mc: OrderedDict({'u': 97, 'cx': 84}) cx count using qsd: OrderedDict({'u': 1686, 'cx': 1046})
num_qubits= 6 base_gate= u cx count using mc: OrderedDict({'u': 181, 'cx': 164}) cx count using qsd: OrderedDict({'u': 1668, 'cx': 1030})
num_qubits= 7 base_gate= x cx count using mc: OrderedDict({'u': 155, 'cx': 140}) cx count using qsd: OrderedDict({'u': 6794, 'cx': 4244})
num_qubits= 7 base_gate= p cx count using mc: OrderedDict({'u': 155, 'cx': 140}) cx count using qsd: OrderedDict({'u': 2364, 'cx': 1488})
num_qubits= 7 base_gate= u cx count using mc: OrderedDict({'u': 267, 'cx': 252}) cx count using qsd: OrderedDict({'u': 3436, 'cx': 2156})
num_qubits= 8 base_gate= x cx count using mc: OrderedDict({'u': 242, 'cx': 220}) cx count using qsd: OrderedDict({'u': 27734, 'cx': 17494})

```

ShellyGarion:
Comparing QSD to multi-controlled synthesis for 2-qubit gates (see: test_mc_2qubit_opt):
 Usually it's better to use the multi-controlled synthesis directly, except when the number of qubits is 4.
```
num_qubits= 3 cx count using mc: OrderedDict({'u': 20, 'cx': 10}) cx count using qsd: OrderedDict({'u': 20, 'cx': 10})
num_qubits= 4 cx count using mc: OrderedDict({'u': 82, 'cx': 48}) cx count using qsd: OrderedDict({'u': 78, 'cx': 46})
num_qubits= 5 cx count using mc: OrderedDict({'u': 396, 'cx': 238}) cx count using qsd: OrderedDict({'u': 406, 'cx': 246})
num_qubits= 6 cx count using mc: OrderedDict({'u': 1668, 'cx': 1030}) cx count using qsd: OrderedDict({'u': 1686, 'cx': 1046})
num_qubits= 7 cx count using mc: OrderedDict({'u': 6836, 'cx': 4278}) cx count using qsd: OrderedDict({'u': 6870, 'cx': 4310})
```

test/python/circuit/test_controlled_gate.py
```diff
@@ -844,15 +844,17 @@ def test_controlled_unitary(self, num_ctrl_qubits):
...
-        base_gate = random_unitary(2**num_target).to_instruction()
+        base_gate = random_unitary(
+            2**num_target, seed=10 * num_ctrl_qubits + num_target
+        ).to_instruction()
         base_mat = base_gate.to_matrix()
         cgate = base_gate.control(num_ctrl_qubits)
         test_op = Operator(cgate)
         cop_mat = _compute_control_matrix(base_mat, num_ctrl_qubits)
-        self.assertTrue(matrix_equal(cop_mat, test_op.data))
+        self.assertTrue(matrix_equal(cop_mat, test_op.data, atol=num_ctrl_qubits * 1e-6))
```
ShellyGarion, line 858: I made a few changes in the test, due to numerical issues in some processors, including: 
fixing a seed, reducing the number of controls and fixing a tolerance bound.

test/python/synthesis/test_qsd_synthesis.py
```diff
@@ -0,0 +1,392 @@
...
+#
+# This code is licensed under the Apache License, Version 2.0. You may
+# obtain a copy of this license in the LICENSE.txt file in the root directory
+# of this source tree or at http://www.apache.org/licenses/LICENSE-2.0.
+#
+# Any modifications or derivative works of this code must retain this
+# copyright notice, and modified files need to carry a notice indicating
+# that they have been altered from the originals.
+
+"""Tests for QS decomposition synthesis methods."""
```
ShellyGarion, line 13: I've used this PR for moving the QSD synthesis tests to a separate file.

qiskit/circuit/library/generalized_gates/unitary.py
```diff
@@ -185,14 +183,16 @@ def control(
         if not annotated:
             mat = self.to_matrix()
             cmat = _compute_control_matrix(mat, num_ctrl_qubits, ctrl_state=None)
-            iso = Isometry(cmat, 0, 0)
+            from qiskit.synthesis.unitary.qsd import qs_decomposition
+
+            cmat_def = qs_decomposition(cmat, opt_a1=True, opt_a2=False)
```
alexanderivrii, line 199: Why is it fine to apply the `opt_a1` optimization and not `opt_a2`?

qiskit/synthesis/unitary/qsd.py
```diff
@@ -122,6 +122,20 @@ def decomp_2q(mat):
...
+                if _off_diagonals_are_zero(um01, um10):
+                    decirc = _demultiplex(
+                        um00,
+                        um11,
+                        opt_a1=opt_a1,
+                        opt_a2=opt_a2,
+                        _depth=_depth,
+                        _ctrl_index=ctrl_index,
+                    )
+                    return decirc
```
alexanderivrii: I am somewhat confused about different optimizations that take place. Why can we only consider the "block diagonal" optimization when `opt_a2=False`? Additionally, when this code is called from `unitary.py`, why do we need to set `opt_a2=False`? Here is a more precise question: is the problem that `opt_a2` does not apply, or that `opt_a2` and "block diagonal" can't be applied together? 

Depending on the answer to the previous question, the following questions might also make sense. If both optimizations can be applied, does the "block diagonal" optimization always lead to a better circuit? If the block diagonal optimization does not apply for a particular unitary, can we still enable `opt_a2`?

qiskit/synthesis/unitary/qsd.py
```diff
@@ -183,12 +197,19 @@ def _demultiplex(um0, um1, opt_a1=False, opt_a2=False, *, _depth=0):
...
         QuantumCircuit: decomposed circuit
     """
     dim = um0.shape[0] + um1.shape[0]  # these should be same dimension
-    nqubits = int(math.log2(dim))
+    nqubits = dim.bit_length() - 1
+    layout = list(range(nqubits))
+    if _ctrl_index:
+        rindex = np.arange(nqubits)[::-1][_ctrl_index]
+        pop_element = layout.pop(rindex)
+        layout = layout + [pop_element]
```
alexanderivrii: Could you simplify this code to something like:
```
if _ctrl_index:
    layout.pop(_ctrl_index)
    layout = layout + [_ctrl_index]
```

qiskit/synthesis/unitary/qsd.py
```diff
@@ -199,24 +220,25 @@ def _demultiplex(um0, um1, opt_a1=False, opt_a2=False, *, _depth=0):
     dmat = np.diag(dvals)
     wmat = dmat @ vmat.T.conjugate() @ um1
 
-    circ = QuantumCircuit(nqubits)
+    qr = [Qubit() for _ in range(nqubits)]
+    circ = QuantumCircuit(qr)
```
alexanderivrii: What is the reason for this change? Personally, I find the older way
```
qc = QuantumCircuit(5)
qc.x(3)
```
a bit nicer than
```
qr = [Qubit() for _ in range(5)]
qc = QuantumCircuit(qr)
qc.x(qr[3])
```

qiskit/synthesis/unitary/qsd.py
```diff
@@ -286,3 +308,54 @@ def _apply_a2(circ):
...
+       um00 (ndarray): upper left block
+       um01 (ndarray): upper right block
+       um10 (ndarray): lower left block
+       um11 (ndarray): lower right block
+    """
+    dim = umat.shape[0]
+    nqubits = dim.bit_length() - 1
+    halfdim = dim // 2
+
+    utensor = umat.reshape((2,) * nqubits + (2,) * nqubits)
```
alexanderivrii: minor: instead of ``(2,) * nqubits + (2,) * nqubits`` you could also write ``(2,) * (2 * nqubits)``.

releasenotes/notes/improve-controlled-UnitaryGate-5f3aa3d44034cfa7.yaml
```diff
@@ -0,0 +1,9 @@
+---
+upgrade_circuits:
+  - |
+    Previously the `control` method of UnitaryGate was utilizing `Isometry` for decomposition.
+    This improvement switches the synthesis to `qs_decomposition` which approximately halves the gate count.
```
alexanderivrii: What about something like this:

```suggestion
    The `control` method of  :class:`.UnitaryGate` now uses :func:`.qs_decomposition` instead of :class:`.Isometry` for decomposition. 
    This change reduces the number of gates by approximately 2x.
```

releasenotes/notes/improve-controlled-UnitaryGate-5f3aa3d44034cfa7.yaml
```diff
@@ -0,0 +1,9 @@
...
+upgrade_synthesis:
+  - |
+    An improvement to the Quantum Shannon Decomposition `qs_decomposition` was added which reduces the CNOT gate count
+    in the special case that the input unitary appears to be a controlled unitary.
```
alexanderivrii: I am not sure, but it seems that this should be a new feature and not an upgrade -- unless you think that someone would want to call the `qs_decomposition` code without the new "block diagonal" optimization.

```suggestion
    The Quantum Shannon Decomposition (:func:`.qs_decomposition`) includes an optimization that reduces the CNOT gate count in the case that the input unitary happens to be a controlled unitary.
```

test/python/circuit/test_controlled_gate.py
alexanderivrii: What is the reason for reducing the number of controls? Is this because the tests become slow or because the matrices are not equal within the given tolerance? As discussed offline, would it make sense to let `atol` to depend on the number of control/target qubits?

test/python/synthesis/test_qsd_synthesis.py
```diff
@@ -0,0 +1,390 @@
...
+
+        gate = UnitaryGate(mat)
+        qc = QuantumCircuit(3)
+        qc.append(gate, range(3))
+        try:
+            qc.to_gate().control(1)
+        except UnboundLocalError as uerr:
+            self.fail(str(uerr))
+
+    def _create_random_multiplexed_gate(self, num_qubits):
```
alexanderivrii, line 254: As it's a bit hard to see what exactly has changed after the tests were moved (at least I don't know how to do this): all the tests before this point are copied without changes and after this point are new?

test/python/synthesis/test_qsd_synthesis.py
```diff
@@ -0,0 +1,390 @@
...
+            for j in range(num_qubits):
+                um00, um11, um01, um10 = qsd._extract_multiplex_blocks(hidden_mat, k=j)
+                # Check the off-diagonal
+                is_mult = qsd._off_diagonals_are_zero(um01, um10)
+                if is_mult:
+                    qc_uc = QuantumCircuit(num_qubits)
+                    ucgate = UCGate([um00, um11])
+                    qc_uc.append(ucgate, layout)
+                    uc_op = Operator(qc_uc)
+                    self.assertEqual(Operator(qc), uc_op)
```
alexanderivrii: Can we also check that for the random multiplexed gate we have created, there is at least one qubit for which `is_mult` is true?

test/python/synthesis/test_qsd_synthesis.py
```diff
@@ -0,0 +1,390 @@
...
+        for layout in itertools.permutations(range(num_qubits)):
+            # create gate with "control" on different qubits
+            qc = QuantumCircuit(num_qubits)
+            qc.append(gate, layout)
+            hidden_mat = Operator(qc).data
+            for j in range(num_qubits):
+                um00, um11, um01, um10 = qsd._extract_multiplex_blocks(hidden_mat, k=j)
+                # Check the off-diagonal
+                is_mult = qsd._off_diagonals_are_zero(um01, um10)
+                if is_mult:
```
alexanderivrii: Same comment as for the previous test: can we also check that `is_mult` is true at least once? Also, since `is_mult` is not used anywhere else, why not write ``if qsd._off_diagonals_are_zero(um01, um10) ...``?

test/python/synthesis/test_qsd_synthesis.py
```diff
@@ -0,0 +1,390 @@
...
+        layout = tuple(np.random.permutation(range(num_qubits)))
+        # create gate with "control" on different qubits
+        qc = QuantumCircuit(num_qubits)
+        qc.append(gate, layout)
+        hidden_op = Operator(qc)
+        hidden_mat = hidden_op.data
+        cqc = transpile(qc, basis_gates=["u", "cx"])
+
+        qc2 = qsd.qs_decomposition(hidden_mat, opt_a2=False)
+        cqc2 = transpile(qc2, basis_gates=["u", "cx"])
```
alexanderivrii: Not sure: maybe you can specify `optimization_level=0` to avoid extra cancellations?

alexanderivrii (commented):
Thanks Shelly (and Erick!), I am eager to see this code finally in Qiskit. 

The PR is in a good shape, I have left a few questions and comments.

If I understand correctly,  for controlled unitaries this PR implements two improvements:  (1) replaces `Isometry` by `qsd` and (2) implements the "block diagonal" optimization in `qsd`. Do we have some data on how much each optimization helps independently?

This is not blocking this PR, but we should understand the bounds on numerical precision errors for QSD (probably depending on the number of qubits).

qiskit/circuit/library/generalized_gates/unitary.py
ShellyGarion: `opt_a2=True` extracts the diagonal elements and does not keep the block structure, so we can't do both optimizations.
`opt_a1=True` keeps the block structure.

qiskit/synthesis/unitary/qsd.py
ShellyGarion: `opt_a2=True` extracts the diagonal elements and does not keep the block structure, so we can't do both optimizations.
Since choosing opt_a1 and opt_a2 is part of the API I'm not sure what is the best way to choose here.
Perhaps we can write a separate `unitary_decomposition` function or HLS plugin that chooses the best unitary synthesis option?

qiskit/synthesis/unitary/qsd.py
ShellyGarion: no, this code is not correct.

qiskit/synthesis/unitary/qsd.py
ShellyGarion: I think you're correct. we can change it back.

qiskit/synthesis/unitary/qsd.py
ShellyGarion: done in [0738249](https://github.com/Qiskit/qiskit/pull/13688/commits/073824919a85274f93e67b1fef13abd1db417a12)

releasenotes/notes/improve-controlled-UnitaryGate-5f3aa3d44034cfa7.yaml
ShellyGarion: done in [70dfc17](https://github.com/Qiskit/qiskit/pull/13688/commits/70dfc170d2b192cc6b51597c54147292292ad17a)

releasenotes/notes/improve-controlled-UnitaryGate-5f3aa3d44034cfa7.yaml
ShellyGarion: done in [70dfc17](https://github.com/Qiskit/qiskit/pull/13688/commits/70dfc170d2b192cc6b51597c54147292292ad17a)

test/python/synthesis/test_qsd_synthesis.py
ShellyGarion: I copied all the tests "as is". I only replaced a hard-coded constant by a function `_qsd_l2_a1a2_mod`

test/python/circuit/test_controlled_gate.py
ShellyGarion: let's discuss how to set the tolerance bound (since locally everything works OK for me, but there are still CI failures)

test/python/synthesis/test_qsd_synthesis.py
ShellyGarion: yes, that's a great idea!

test/python/synthesis/test_qsd_synthesis.py
ShellyGarion: yes, that's a good idea!

test/python/synthesis/test_qsd_synthesis.py
ShellyGarion: OK. [1597516](https://github.com/Qiskit/qiskit/pull/13688/commits/1597516bc45b4c126567b2ddc0d43608b7f884c2)

qiskit/synthesis/unitary/qsd.py
ShellyGarion: done in [0738249](https://github.com/Qiskit/qiskit/pull/13688/commits/073824919a85274f93e67b1fef13abd1db417a12)

test/python/synthesis/test_qsd_synthesis.py
ShellyGarion: done in [1597516](https://github.com/Qiskit/qiskit/pull/13688/commits/1597516bc45b4c126567b2ddc0d43608b7f884c2)

test/python/synthesis/test_qsd_synthesis.py
ShellyGarion: done [1597516](https://github.com/Qiskit/qiskit/pull/13688/commits/1597516bc45b4c126567b2ddc0d43608b7f884c2)

test/python/synthesis/test_qsd_synthesis.py
ShellyGarion: yes, from here onward these are new tests for this PR

qiskit/synthesis/unitary/qsd.py
alexanderivrii: Hmm, I am trying to make sense of what's going on by looking at this code 
```
circ.append(left_gate, layout[: nqubits - 1])
circ.append(ucrz, [layout[-1]] + layout[: nqubits - 1])
circ.append(right_gate, layout[: nqubits - 1])
```
and at this picture
```
               ┌───┐
        ───────┤ Rz├──────
          ┌───┐└─┬─┘┌───┐
        /─┤ w ├──□──┤ v ├─
          └───┘     └───┘
```
Let's assume that `w` and `v` are 3-qubit unitaries. From what I understand, `nqubits = 4`, `layout=[0, 1, 2, 3]`, `w` is defined over qubits `0, 1, 2`, and the `Rz`-box is over qubit `3`. 

What should be `_ctrl_index` then? From this description: 
```_ctrl_index (int): The index wrt which um0 and um1 are controlled.```
I would expect that `_ctrl_index` is the qubit of the `Rz` box, but it looks like that this is not actually the case.

Just to make clear: I believe that the current code is correct, but the description of what `_ctrl_index` is not accurate.

In any case, I now think that you can simplify the original code as:
```
layout = list(range(nqubits))
rindex = num_qubits - 1 - _ctrl_index
layout.pop(rindex)
layout = layout + [rindex]
```
or even as 
```
rindex = num_qubits - 1 - _ctrl_index
layout = list(range(0, rindex)) + list(range(rindex+1,num_qubits)) + [rindex]
```

File: qiskit/circuit/library/generalized_gates/unitary.py (hunk 3 above)
Comment (ShellyGarion) [reply qiskit/circuit/library/generalized_gates/unitary.py:199]: I'll add some benchmarks comparing all approaches (isometry, qsd with opt_a2=True, qsd with opt_a2=False)

File: qiskit/synthesis/unitary/qsd.py (hunk 5 above)
Comment (ShellyGarion) [reply qiskit/synthesis/unitary/qsd.py]: thanks, done in [9dd2af2](https://github.com/Qiskit/qiskit/pull/13688/commits/9dd2af2c1a9cd4af014f058c3afbf524cab5a254)

**Comment by ShellyGarion** [comment]
Here are a few benchmarks. I took a random unitary matrix on `num_qubits=(3,4,5,6)` and `num_controls=(0,1,2,3)` and tried the following synthesis algorithms (transpiling with `optimization_level=0`):
1. `Isometry`  https://docs.quantum.ibm.com/api/qiskit/qiskit.circuit.library.Isometry
2. QSD with `opt_a1=False` and `opt_a2=False`
3. QSD with `opt_a1=True` and `opt_a2=True`
4. QSD with `opt_a1=True` and `opt_a2=False` (so we can do the multiplexed reduction for controlled gates)

Here are the CX counts results:

| num_qubits | num_controls | Isometry | QSD: `opt_a1=False`, `opt_a2=False` | QSD: `opt_a1=True`, `opt_a2=True` | QSD: `opt_a1=True`, `opt_a2=False` | 
|--------------|----------------|-----------|-----------------------------------------|----------------------------------------|---------------------------------------------------|
| 3 | 0 | 41 | 24 | 20 | 23 | 
| 3 | 1 | 144 | 56 | 99 | 54 |
| 3 | 2 | 922 | 252 | 444 | 244 |
| 3 | 3 | 1741 | 1088 | 1868 | 1046 | 
| 4 | 0 | 218 | 120 | 100 | 115 | 
| 4 | 1 | 928 | 256 | 444 | 246 |
| 4 | 2 | 4279 | 1088 | 1868 | 1046 |
| 4 | 3 | 18348 | 4480 | 7660 | 4310 | 
| 5 | 0 | 1025 | 528 | 444 | 507 |
| 5 | 1 | 2811 | 1088 | 1868 | 1046 | 
| 5 | 2 | 18131 | 4480 | 7660 | 4310 | 
| 5 | 3 | 26022 | 18176 | 31020 | 17494 | 
| 6 | 0 | 4474 | 2208 | 1868 | 2123 |
| 6 | 1 | 11658 | 4480 | 7660 | 4310 | 
| 6 | 2 | 72595 | 18176 | 31020 | 17494 | 

In conclusion, for a random unitary (non-controlled) it is most beneficial to use QSD with `opt_a1=True` and `opt_a2=True`.
For controlled unitary, it is better to use QSD with `opt_a1=True` and `opt_a2=False`, thus allowing to do the multiplexer trick.

**Comment by alexanderivrii** [comment]
Thanks for providing the table (the second last column should read "opt_a2 = True", correct?

**Comment by ShellyGarion** [comment]
> Thanks for providing the table (the second last column should read "opt_a2 = True", correct?

yes, fixed. thanks!

File: qiskit/circuit/library/generalized_gates/unitary.py [hunk 14]
```diff
@@ -161,6 +162,10 @@ def _define(self):
             )
 
             self.definition = qs_decomposition(self.to_matrix())
+            if not (
+                matrix_equal(Operator(self.definition).to_matrix(), self.to_matrix(), atol=1e-7)
+            ):
+                self.definition = Isometry(self.matrix, 0, 0).definition
```
ShellyGarion, line 172: code updated to use Isometry the calculated Operator is different than the original matrix.

qiskit/circuit/library/generalized_gates/unitary.py
```diff
@@ -185,14 +190,19 @@ def control(
         if not annotated:
             mat = self.to_matrix()
             cmat = _compute_control_matrix(mat, num_ctrl_qubits, ctrl_state=None)
-            iso = Isometry(cmat, 0, 0)
+            from qiskit.synthesis.unitary.qsd import qs_decomposition
+
+            cmat_def = qs_decomposition(cmat, opt_a1=True, opt_a2=False)
+            if not matrix_equal(Operator(cmat_def).to_matrix(), cmat, atol=1e-7):
+                self.definition = Isometry(cmat, 0, 0).definition
```
ShellyGarion, line 206: code updated to use Isometry the calculated Operator is different than the original matrix.

test/python/circuit/test_controlled_gate.py
ShellyGarion: code updated to use Isometry the calculated Operator is different than the original matrix.

qiskit/circuit/library/generalized_gates/unitary.py
alexanderivrii: This is not ideal as this adds additional overhead and does not guarantee that we always get the better definition based on QSD. However, this is the best we can do for now and this does significantly improve the gate counts when this does work. Thus I am in favor of merging this in the current form, but we should open an issue to investigate and improve numerical decomposition errors in `cossin`.

qiskit/circuit/library/generalized_gates/unitary.py
alexanderivrii: Do you know what are the typical numerical errors when running the `Isometry` code?

qiskit/synthesis/unitary/qsd.py
alexanderivrii: Thanks for the explanation.

qiskit/synthesis/unitary/qsd.py
```diff
@@ -183,12 +197,18 @@ def _demultiplex(um0, um1, opt_a1=False, opt_a2=False, *, _depth=0):
           unitaries into a diagonal gate and a two cx unitary and reduces overall cx count by
           4^(n-2) - 1.
        _depth (int): This is an internal variable to track the recursion depth.
+       _ctrl_index (int): The index wrt which um0 and um1 are controlled.
```
alexanderivrii, line 201: I have mentioned previously that the explanation that the explanation ``_ctrl_index (int): The index wrt which um0 and um1 are controlled`` is not fully accurate: the control index is rather what you call `rindex` below, do you agree? Could you please update the docstring or, better yet, call the function with `nqubits - 1 - _ctrl_index` (and simplify the code for computing `layout` accordingly)?

test/python/circuit/test_controlled_gate.py
alexanderivrii: Hmm, one more question: why is the tolerance `1e-6` and not `1e-7`?

alexanderivrii (commented):
Thanks for doing this, this looks good. I have left just one nitpick about the argument `_ctrl_index`. Can you also open an issue to investigate/fix numerical errors in the cossin decomposition?

test/python/circuit/test_controlled_gate.py
ShellyGarion: the current test is 1e-7 (and not 1e-6)

alexanderivrii: You are right. This is a strange glitch: when "reviewing changes" or looking at the file, I see ``1e-7``. But from the PR page I see ``1e-6``.

qiskit/circuit/library/generalized_gates/unitary.py
ShellyGarion: I don't think there are special numerical errors. QSD is done recursively and this is the reason that we get numeric errors. However, Isometry is not iterative. I can add similar tests to Isometry.

qiskit/synthesis/unitary/qsd.py
ShellyGarion: done in [6fffbd9](https://github.com/Qiskit/qiskit/pull/13688/commits/6fffbd9838e78b6e6829d99845ddf0b28492c71e)

ShellyGarion:
See: https://github.com/Qiskit/qiskit/issues/14422

alexanderivrii (approved):
Thanks @ShellyGarion, LGTM!

PR #13690, opened 2025-01-19
Ignoring unsupported gates for Solovay-Kitaev transpiler pass

alexanderivrii:
### Summary

The Solovay-Kitaev transpiler pass (i.e. the standalone pass and not the unitary synthesis plugin) previously considered all operations over $1$ qubit and ignored all operations over $k$ qubits for $k\ne 1$. In doing so, it raised a transpiler error for circuits with single-qubit operations that are either parameterized or don't have a ``to_matrix`` method (these include barriers, measures, control-flow ops and more). This PR changes the default behavior to ignore such operations instead of raising an error.

### DIFF
--- a/releasenotes/notes/sk-ignore-unsupported-ops-8d7d5f6fca255ffb.yaml
+++ b/releasenotes/notes/sk-ignore-unsupported-ops-8d7d5f6fca255ffb.yaml
@@ -1,0 +1,6 @@
+---
+upgrade_transpiler:
+  - |
+    The :class:`.SolovayKitaev` transpiler pass no longer raises an exception on circuits
+    that contain single-qubit operations without a ``to_matrix`` method (such as measures,
+    barriers, control-flow operations) or parameterized single-qubit operations.
--- a/releasenotes/notes/sk-ignore-unsupported-ops-8d7d5f6fca255ffb.yaml
+++ b/releasenotes/notes/sk-ignore-unsupported-ops-8d7d5f6fca255ffb.yaml
@@ -3,4 +3,5 @@ upgrade_transpiler:
   - |
     The :class:`.SolovayKitaev` transpiler pass no longer raises an exception on circuits
     that contain single-qubit operations without a ``to_matrix`` method (such as measures,
-    barriers, control-flow operations) or parameterized single-qubit operations.
+    barriers, control-flow operations) or parameterized single-qubit operations,
+    but will leave them unchanged.
--- a/qiskit/transpiler/passes/synthesis/solovay_kitaev_synthesis.py
+++ b/qiskit/transpiler/passes/synthesis/solovay_kitaev_synthesis.py
@@ -170,7 +170,7 @@ class SolovayKitaev(TransformationPass):
 
             # ignore operations on which the algorithm cannot run
             if (
-                (not node.op.num_qubits == 1)
+                (node.op.num_qubits != 1)
                 or node.is_parameterized()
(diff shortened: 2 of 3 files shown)


Cryoris:
I agree it should leave `Measure` and `Reset` as is, but I'm not sure if we should just ignore gates that don't have `to_matrix`. I always thought of SK like a kind of basis translator that takes you to the target 1q basis and, like `BasisTranslator`, it would fail if it can't 🤔

alexanderivrii:
> I agree it should leave Measure and Reset as is, but I'm not sure if we should just ignore gates that don't have to_matrix. I always thought of SK like a kind of basis translator that takes you to the target 1q basis and, like BasisTranslator, it would fail if it can't 🤔

This is indeed a valid alternative. I like the analogy with the BasisTranslator pass, though (unlike BT) SK already ignores all operations on 2 or more qubits, and it feels strange that the pass should ignore, for instance, the 2-qubit `initialize` instruction while throw an error for the 1-qubit `initialize` instruction.

The other thing is that probably we need to make the pass recursive for control-flow ops (and not just ignore these as it happens right now), what do you think?

Maybe we should first decide on the typical usages of the SK pass, when and how it would be used in a typical transpiler pipeline.

jakelishman:
The SK algorithm only applies to decompositions of $U(2)$, so it makes sense to ignore anything that isn't in that.  Under that rule, ignoring both `Measure` and _any_ `Initialize` is consistent, because neither is unitary.

Agree that the pass should recurse over control-flow operations. 

If we're intending to use the SK algorithm as part of a translation stage, we might want to consider it as an entire pipeline.  To use the `BasisTranslator` example: that alone isn't an entire translation stage; the stage also includes `Unroll3qOrMore`.  If we want to make an SK-based translation plugin, I'd suggest we make an entire plugin, rather than trying to make the pass try to guess - the pipeline can be something like
- unroll 3q or more
- consolidate blocks
- unitary synthesis (min_qubits=2) to a target basis `[u, <2q gates>]`
- SK

In an ideal world, we'd want `UnitarySynthesis` to output the 1q unitaries without bothering to find their Euler-angle decomposition, but we could potentially work out how to improve that story later.

I wouldn't necessarily make the SK pass error out if it can't translate, since it's conceivable to me that you might want to use it in other contexts - we can make the translation _stage_ error out in that case, potentially, where it's clear that the stage is required to fail if translation can't be done.

alexanderivrii:
Thanks @jakelishman, @Cryoris! I have added recursion for control-flow operators. If we agree that the SK pass should not error out if it can't translate, is there more you suggest to do here, modulo correcting the class documentation and polishing the release notes?

Let's take the discussion of the transpilation pipeline to #13695.

releasenotes/notes/sk-ignore-unsupported-ops-8d7d5f6fca255ffb.yaml
```diff
@@ -0,0 +1,6 @@
+---
+upgrade_transpiler:
+  - |
+    The :class:`.SolovayKitaev` transpiler pass no longer raises an exception on circuits
+    that contain single-qubit operations without a ``to_matrix`` method (such as measures,
+    barriers, control-flow operations) or parameterized single-qubit operations.
```
Cryoris: It might be good to state what happens with them 🙂 
```suggestion
    barriers, control-flow operations) or parameterized single-qubit operations,
    but will leave them unchanged.
```

Cryoris (approved):
LGTM, thanks!

PR #13694, opened 2025-01-20
C API for ``SparseObservable``

### LINKED ISSUE #13445: Add C API for SparseObservable
### What should we add?

The recently added `SparseObservable` class was added as a Rust native object in #12671 . To facilitate different use cases for generating and interacting with this new structure from compiled languages including other Python libraries' compiled extensions we should add a lower level C API around the structure. This will likely involve splitting the current `SparseObservable` structure into a pure rust struct and a wrapping python interface struct. Then we can add a new C FFI wrapper for it, likely should be put in a new crate `c_ext` crate or something like that to define the C API to qiskit. This sparse observable interface will likely be a good way to PoC the packaging and structure for other C API's that have a longer on-ramp, mainly: #13276.

Cryoris:
### Summary

This PR introduces a C API for the ``SparseObservable``. These changes are based on top of #13595, which separates the Python and Rust implementations. While there are still some open question (see below), this first implementation is functioning and can serve as basis for discussions. See e.g. the test file (`test/c/test_sparse_observable.c`) for example usage.

Closes #13445.

### Details and comments

Open points (that we know of, certainly there are more 😄):
- [x] Merge #13595 
- [x] Add the tests to CI
- [x] ~~How to add the docs?~~ We have a working concept with doxygen & breathe, let's add as follow-up
- [x] Should structs/enums have prefixes? I.e. should `SparseObservable` become something like `QkSparseObservable`, or are the names sufficiently unique.
- [x] Reno


.clang-format
```diff
@@ -0,0 +1,7 @@
+---
+BasedOnStyle: LLVM
+ColumnLimit: '100'
+IndentWidth: '4'
+Language: Cpp
+TabWidth: '4'
+UseTab: Never
```
jakelishman: Super minor: can you fix the line ending of the file?

.gitignore
```diff
@@ -19,6 +19,9 @@ __pycache__/
 # C extensions
 *.so
 
+# The C API header is generated automatically, therefore we do not upload it.
+qiskit.h
+
```
jakelishman: I feel like we should build this into some sort of sensible deployment location, rather than being loose in the repo root.  For most Python stuff, we build it into `/dist` for distribution.  Perhaps we make a folder `/dist/c` (if it doesn't already exist), then output header files into `/dist/c/include`?

CONTRIBUTING.md
```diff
@@ -736,6 +780,15 @@ conform to the style guidelines. This is very similar to running `tox -eblack` f
...
 
+### C style and lint
+
+Qiskit uses [clang-format](https://clang.llvm.org/docs/ClangFormat.html) to format C code.
+The style is based on LLVM, with some few Qiskit-specific adjustments. 
+To check whether the C code conforms to the style guide, you can run `make cformat`. This check
+will need to execute without any warnings or errors for CI to pass.
+
+Automatic formatting will not be applied by `make cformat`, but instead you can call
+`./tools/run_clang_format.sh apply`.
```
jakelishman: Maybe we could make `make cformat-fix` do this if we're centralising on the Makefile?

tools/install_ubuntu_docs_dependencies.sh
```diff
@@ -8,4 +8,4 @@ python -m pip install --upgrade pip setuptools wheel
 python -m pip install --upgrade tox
 
 sudo apt-get update
-sudo apt-get install -y graphviz pandoc
+sudo apt-get install -y graphviz pandoc clang-format
```
jakelishman: This isn't a requirement for the docs build, so the script name (and comment at the top) are getting a bit misleading - might be cleaner to put it in a separate script.

tools/run_clang_format.sh
```diff
@@ -0,0 +1,23 @@
...
+    args="-i"  # inplace change of the files
+else
+    # any other argument is invalid
+    echo "Invalid argument, either no arguments or 'apply' is supported, not: $1"
+    exit 1
+fi
+
+# this is the style file -- note that this script should
+# be run from root, such that this file is correctly found
+style=".clang-format"
```
jakelishman: If you look in `tests/qpy_compat/run_tests.sh` (for example), there's some ways you can make the script work independently of the working directory it's launched from.

tools/run_clang_format.sh
```diff
@@ -0,0 +1,23 @@
+#! /bin/bash
+
```
jakelishman, line 2: Can we add at least
```bash
set -e
```
(and possibly all of `set -exo pipefail`) to the setup at the top to ensure we fail the script if intermediate commands file?

tools/run_clang_format.sh
```diff
@@ -0,0 +1,23 @@
...
+# be run from root, such that this file is correctly found
+style=".clang-format"
+
+# get all tracked files in HEAD, and filter for files ending in .c or .h
+files=$(git ls-tree --name-only -r HEAD | grep ".*\.[c,h]$")
+
+# apply clang format on all files
+for file in $files; do
+    clang-format --style="file:$style" $args $file
+done
```
jakelishman: At the moment, this won't actually fail the script if `clang-format` in check mode fails, I don't think?

If we `set -e` at the top, then the script will exit on the first error, but we probably want it to run through all the files and report an error condition based on if any of them failed.  Something like
```bash
status = 0
for file in $files; do
    clang-format --style="file:$style" $args $file || :
    status|=$?;
done
exit $status
```
is _possibly_ along the right lines, but I'm not expert at typing `bash` scripts without testing them lol.

Second point: it's probably pointless for now, but in the future we might want to avoid spinning up too many different `clang-format` processes, and instead do something along the lines of
```bash
find -print0 -name '*.[ch]' "$repo_root/test/c" "$repo_root/crates" | xargs -0 clang-format --style="file:$style" $args 
```
which would let `xargs` distribute the arguments to `clang-format` so we don't have to make as many subprocesses as there are files (or potentially we write out the files into a tmpfile that we pass to `clang-format --files=`, though imo it sucks that `clang-format` can't accept file-name input from `stdin` with `--files=-`).

tools/verify_headers.py
```diff
@@ -37,11 +37,14 @@ def discover_files(code_paths):
...
                 for subfile in directory[2]:
+                    # pylint: disable=too-many-boolean-expressions
                     if (
                         subfile.endswith(".py")
                         or subfile.endswith(".pyx")
                         or subfile.endswith(".pxd")
                         or subfile.endswith(".rs")
+                        or subfile.endswith(".c")
+                        or subfile.endswith(".h")
                     ):
```
jakelishman: Or alternatively `if subfile.rsplit(".", 1)[-1] in ("py", "rs", "c", "h")` if you want to avoid the pylint supression (we can remove the Cython extensions, since we don't use it any more).

tox.ini
```diff
@@ -84,6 +84,11 @@ commands =
   python {toxinidir}/tools/verify_images.py
   reno -q lint
 
+[testenv:lint-c]
+allowlist_externals = bash
+commands =
+  bash {toxinidir}/tools/run_clang_format.sh
+
```
jakelishman: I'm not sure there's much point putting this in `tox` - we don't need an isolated Python environment, and we already have `make clint`, which is even faster to type.  `tox` just makes things more awkward and more likely to take forever building an environment / wasting space / whatever.

test/c/common.h
jakelishman, line 1: I am very unfamiliar with testing in C, but having to write our own runner harness like this feels weird to me - I'd have expected us to get a bit more help from the test framework.

test/c/CMakeLists.txt
```diff
@@ -0,0 +1,42 @@
...
+
+# All files in this directory (non-recursively! use GLOB_RECURSE if that becomes
+# necessary) which match the pattern `test_*.c` are discovered as test files by
+# the command below. All paths are stored relative to the location of this file.
+file (
+    GLOB discovered_tests
+    LIST_DIRECTORIES FALSE
+    RELATIVE ${CMAKE_CURRENT_SOURCE_DIR}
+    test_*.c
+)
```
jakelishman, line 14: Is there any harm in just making it recursive immediately?  Seems like a footgun waiting to go off, since practically all our other test suites involve subdirectories.

releasenotes/notes/c-sparse-observable-c14b17a9303b1c34.yaml
```diff
@@ -0,0 +1,48 @@
+---
+features:
+  - |
+    Introduced a C API to build and interact with sparse observables.
+    This API does not yet allow for an end-to-end workflow in C, but rather lays the 
+    foundation for Qiskit's C interface. The (dynamic) C library is compiled by Rust and 
+    the header file can be automatically generated using ``cbindgen``, both of which is 
+    done by running ``make cheader``.
```
jakelishman: Probably let's link to more complete documentation - feels like this summary of the generation doesn't tell the casual reader anything much, and doesn't give the advanced reader enough information.

releasenotes/notes/c-sparse-observable-c14b17a9303b1c34.yaml
```diff
@@ -0,0 +1,48 @@
...
+      }
+
+    To compile the above, a Rust compiler is required to build the dynamic library and 
+    ``cbindgen`` to build the header. Then the C script can be compiled by specifying the library's 
+    location, e.g., using ``gcc``, we have:
+
+    .. code::
+
+      make cheader  # build's the library into ./target/release and the header into root
+      gcc <my_script.c> -lqiskit_cext -L./target/release -I./  # compile and link the library
```
jakelishman: This really ought to be in our public docs somewhere, right?

For the `-I` command: `-I.` looks super gnarly to me, because `.` is intended here to be the repo root, which is a very dirty directory - there's all sorts of stuff in it.  `make cheader` should almost certainly be writing out to some clean build directory like `/dist/c/include`, which will be much easier to package for distribution.

crates/cext/Cargo.toml
jakelishman, line 1: Can you also inherit the lints from the workspace (examples in the other `Cargo.toml` files)?

Makefile
```diff
@@ -83,3 +83,21 @@ coverage_erase:
 	coverage erase
 
 clean: coverage_erase ;
+
+# Run clang-format
+cformat:
+	sh tools/run_clang_format.sh
```
jakelishman: Minor: the shebang of the file claims it needs `bash`, but you're running it with `sh` here.

Makefile
```diff
@@ -83,3 +83,21 @@ coverage_erase:
...
+
+# Build C API crate and header
+cheader:
+	cargo build --release --no-default-features --features cbinding
+	cbindgen --crate qiskit-cext --output qiskit.h --lang C
```
jakelishman: It'd be good to have a) a centralised output here so there's a clean output directory for the include files, so we can tell people where to point the `-I` flag and b) so we can add a `clean` (or `cclean`) rule that cleans the generated files.

_In theory_, `cheader` could have proper dependencies (or something using CMake) so we only rebuild a given header file when it's stale rather than always, but it's not big deal.

releasenotes/notes/c-sparse-observable-c14b17a9303b1c34.yaml
```diff
@@ -0,0 +1,48 @@
...
+    the header file can be automatically generated using ``cbindgen``, both of which is 
+    done by running ``make cheader``.
+
+    The detailed syntax and more information is available under the C API docs, 
+    but a minimal example to construct the 100-qubit observable ``X0 Y1 Z1`` is:
+
+    .. code:: c
+
+      #include <stdio.h>
+      #include "qiskit.h"
```
jakelishman: Is the `#include "qiskit.h"` convention (quoted include) definitely the right one for us to push people to? Genuine question: I _very_ vaguely thought that quoted imports are slightly more for files local to the project itself, whereas it's more typical to use `<>` imports when compiling against a library distributed from elsewhere, since `""` searches the working directory before the `-I` paths.

Completely willing to be told I'm wrong on that, though - I've little experience.

releasenotes/notes/c-sparse-observable-c14b17a9303b1c34.yaml
```diff
@@ -0,0 +1,48 @@
...
+
+      int main() {
```
jakelishman: _technically_ I think this isn't valid C code since C99 - it should be `int main(void)` at a minimum, but `int main(int argc, char *argv[])` would be preferable.  In practice I imagine every C compiler we care about would accept it anyway, but probably be good to show best practices.

releasenotes/notes/c-sparse-observable-c14b17a9303b1c34.yaml
```diff
@@ -0,0 +1,48 @@
...
+        uint32_t indices[3] = {0, 1, 2};  // indices: 0 1 2
+        QkSparseTerm term = {&coeff, 3, bit_terms, indices, num_qubits};
+        qk_obs_add_term(obs, &term);  // append the term
+
+        // print some properties
+        printf("num_qubits: %i\n", qk_obs_num_qubits(obs));
+        printf("num_terms: %i\n", qk_obs_num_terms(obs));
+
+        // free the memory allocated for the observable
+        obs_free(obs);
```
jakelishman: Should this be `qk_obs_free`?

crates/cext/cbindgen.toml
```diff
@@ -0,0 +1,30 @@
+include_version = true
+style = "type" 
+sys_includes = ["complex.h"]
+after_includes = """
+// Complex number typedefs -- note these are memory aligned but
+// not calling convention compatible.
+typedef float complex QkComplex32; 
+typedef double complex QkComplex64;
```
jakelishman, line 8: Is there much advantage to these over using the types raw?

crates/cext/cbindgen.toml
```diff
@@ -0,0 +1,30 @@
...
+
+[enum]
+prefix_with_name = true
+
+[export]
+prefix = "Qk"
+renaming_overrides_prefixing = false
+
+[export.rename]
+"CSparseTerm" = "SparseTerm"
```
jakelishman: Minor: can you fix the line ending?

crates/cext/src/sparse_observable.rs
```diff
@@ -0,0 +1,749 @@
...
+        if value.bit_terms.is_null() || value.indices.is_null() {
+            return Err(CInputError::NullPointerError);
+        }
+
+        if !value.bit_terms.is_aligned() || !value.indices.is_aligned() {
+            return Err(CInputError::AlignmentError);
+        }
+
+        let bit_terms = unsafe { ::std::slice::from_raw_parts(value.bit_terms, value.len) };
+        let indices = unsafe { ::std::slice::from_raw_parts(value.indices, value.len) };
```
jakelishman, line 51: Can we get some `// SAFETY:` comments on these to explain why they're certainly valid?  I wouldn't mind moving up the `let bit_terms` to be right below the pointer checks as well to keep things even more localised.

crates/cext/src/sparse_observable.rs
```diff
@@ -0,0 +1,749 @@
...
+use qiskit_accelerate::sparse_observable::{BitTerm, SparseObservable, SparseTermView};
+
+/// @ingroup QkSparseTerm
+/// A term in a [SparseObservable].
+///
+/// This term does not own the data, but only contains pointers (and a size) to the
+/// bit terms, indices and the coefficient.
+#[repr(C)]
+pub struct CSparseTerm {
+    coeff: *mut Complex64,
```
jakelishman: The trouble with `Complex64` in interfaces is only in the calling conventions of functions - we can't use a raw `Complex64` in them because the conventions of Rust structs aren't necessarily the same as for C's `double complex` on all platforms and compilers.  I think it's fine to have `coeff: Complex64` inline here, because it's memory-layout compatible with `double complex`, and that'll simplify creating one of these.

crates/cext/src/sparse_observable.rs
```diff
@@ -0,0 +1,749 @@
...
+
+use crate::exit_codes::{CInputError, ExitCode};
+use num_complex::Complex64;
+use qiskit_accelerate::sparse_observable::{BitTerm, SparseObservable, SparseTermView};
+
+/// @ingroup QkSparseTerm
+/// A term in a [SparseObservable].
+///
+/// This term does not own the data, but only contains pointers (and a size) to the
+/// bit terms, indices and the coefficient.
```
jakelishman: We might want to expand the documentation on this a little bit, to make it more explicit that it's the lengths of `bit_terms` and `indices` that `len` refers to, and that it's the responsibility of C code to ensure that those pointers are non-null, aligned and are valid to read `len` elements from them.

crates/cext/src/sparse_observable.rs
```diff
@@ -0,0 +1,749 @@
...
+
+/// Check the pointer is not null and is aligned.
+fn check_ptr<T>(ptr: *const T) {
+    if ptr.is_null() {
+        panic!("Unexpected null pointer.");
+    };
+    if !ptr.is_aligned() {
+        panic!("Pointer is not properly aligned.");
+    };
+}
```
jakelishman: Perhaps this should return `Result`, then we can use it in more places - when panicking is appropriate, we can `Result.unwrap` it to panic, and when we're using it fallibly (like above in the `SparseTerm` wrapper), we can match on the error result.

crates/cext/src/sparse_observable.rs
```diff
@@ -0,0 +1,749 @@
...
+/// # Safety
+///
+/// This function is unsafe since it is reading from a raw pointer. Even if the pointer
+/// is not null and aligned, the memory could still be an invalid representation of the
+/// target type ``T``.
+unsafe fn mut_ptr_as_ref<'a, T>(ptr: *mut T) -> &'a mut T {
+    check_ptr(ptr);
+    let as_mut_ref = unsafe { ptr.as_mut() };
+    as_mut_ref.unwrap() // we know the pointer is not null, hence we can safely unwrap
+}
```
jakelishman: Typically the "Safety" section tells you what you must do to call the function safely.  In this case, the comments could make it clearer that the null and aligned checks are handled safely, and the only `unsafe` bit of it is that the data _must_ be valid.  The only data that's truly unsafe in this sense is `BitTerm`, since all bit patterns of `u32`, `double complex` and `usize` are valid values of those types, they just might not be _correct_ values of those types.

crates/cext/src/sparse_observable.rs
```diff
@@ -0,0 +1,749 @@
...
+///
+/// Behavior is undefined if any of the following conditions are violated:
+///
+///   * ``coeffs`` is a pointer to a ``complex double`` array of length ``num_terms``
+///   * ``bit_terms`` is a pointer to a ``QkBitTerm`` array of length ``num_bits``
+///   * ``indices`` is a pointer to a ``uint32_t`` array of length ``num_bits``, which is
+///     term-wise sorted in strict ascending order, and every element is smaller than ``num_qubits``
+///   * ``boundaries`` is a pointer to a ``size_t`` array of length ``num_terms + 1``, which is
+///     sorted in ascending order, the first element is 0 and the last element is
+///     smaller than ``num_terms``
```
jakelishman: This is really a comment for the follow-up documentation PR, but since I'm here now:

I'm not super wild about introducing new Doxygen syntax to the docs build - we already have to deal with Markdown and Sphinx-flavoured rST, and I'd prefer it if there's a sensible way not to introduce another language.

I know we've talked about this elsewhere, and you've heard this concern of mine before, but just commenting again for the record.  We can take it up in the docs PR more.

crates/cext/src/sparse_observable.rs
```diff
@@ -0,0 +1,749 @@
...
+    indices: *mut u32,
+    boundaries: *mut usize,
+) -> *mut SparseObservable {
+    let num_terms = num_terms as usize;
+    let num_bits = num_bits as usize;
+
+    let coeffs = unsafe { ::std::slice::from_raw_parts(coeffs, num_terms).to_vec() };
+    let bit_terms = unsafe { ::std::slice::from_raw_parts(bit_terms, num_bits).to_vec() };
+    let indices = unsafe { ::std::slice::from_raw_parts(indices, num_bits).to_vec() };
+    let boundaries = unsafe { ::std::slice::from_raw_parts(boundaries, num_terms + 1).to_vec() };
```
jakelishman, line 202: We should have at least a `SAFETY` comment on this block.

I think it's worth discussing what level of fallibility checks we want to have here.

Given that we already have to copy the buffers across to form the Rust-space `SparseObservable`, I feel like this function _could_ do entirely complete error-checking of the unsafety (except for read-length validity).  The biggest one is `BitTerm` - we could accept `bit_terms: *mut u8`, do the non-null and alignment checks (as for all pointers), returning `null_ptr` on failure, then turn it into a `&[u8]`, and then do a `bytemuck` checked cast to turn it into a `&[BitTerm]`.  Since we have to iterate through the whole slice anyway, the overhead isn't huge.

On the other hand: the expectations of C programmers are different to Rust programmers, and it might be acceptable to say "this should be as fast as possible, and if you invoke undefined behaviour, it's your own problem - we're not going to help you".

I don't think I have a fully formed opinion about it yet, but would appreciate us discussing / deciding as a team (and potentially encoding it somewhere into a design document).

crates/cext/src/sparse_observable.rs
```diff
@@ -0,0 +1,749 @@
...
+    };
+
+    match obs.add_term(view) {
+        Ok(_) => ExitCode::Success,
+        Err(err) => ExitCode::from(err),
+    }
+}
+
+/// @ingroup QkSparseObservable
+/// Get an observable term by reference. This can modify the underlying observable.
```
jakelishman: Should maybe comment that it's typically a _really bad_ idea to actually modify it in place, or at least spell out all the reasons it's hard (e.g. can't change the length of indices or terms, the `coeff` won't update (subject to my other comment about `*mut Complex64` maybe not being necessary), etc) - the fact that you _can_ mutate the observable in place through these pointers is more of a side-effect of being no-copy than a desirable feature.

crates/cext/src/sparse_observable.rs
```diff
@@ -0,0 +1,749 @@
...
+///
+/// # Safety
+///
+/// Behavior is undefined ``obs`` is not a valid, non-null pointer to a ``QkSparseObservable``.
+#[no_mangle]
+#[cfg(feature = "cbinding")]
+pub unsafe extern "C" fn qk_obs_num_terms(obs: *const SparseObservable) -> u64 {
+    let obs = unsafe { const_ptr_as_ref(obs) };
+    obs.num_terms() as u64
+}
```
jakelishman: Why do we cast this one to `u64` when we expose `usize` (as `size_t`) elsewhere?

crates/cext/src/sparse_observable.rs
```diff
@@ -0,0 +1,749 @@
...
+///
+/// # Safety
+///
+/// Behavior is undefined ``obs`` is not a valid, non-null pointer to a ``QkSparseObservable``.
+#[no_mangle]
+#[cfg(feature = "cbinding")]
+pub unsafe extern "C" fn qk_obs_len(obs: *const SparseObservable) -> u64 {
+    let obs = unsafe { const_ptr_as_ref(obs) };
+    obs.bit_terms().len() as u64
+}
```
jakelishman: Same question about `u64` vs `usize`.

crates/cext/src/sparse_observable.rs
```diff
@@ -0,0 +1,749 @@
...
+
+/// @ingroup QkSparseObservable
+/// Get a pointer to the coefficients.
+///
+/// This can be used to read and modify the observable's coefficients.
```
jakelishman: Would be good to comment about the data-length validity of the pointer (here and for other similar methods).

crates/cext/src/sparse_observable.rs
```diff
@@ -0,0 +1,749 @@
...
+///
+/// Behavior is undefined ``obs`` is not a valid, non-null pointer to a ``QkSparseObservable``.
+#[no_mangle]
+#[cfg(feature = "cbinding")]
+pub unsafe extern "C" fn qk_obs_indices(obs: *mut SparseObservable) -> *mut u32 {
+    let obs = unsafe { mut_ptr_as_ref(obs) };
+
+    // this is unsafe as we can no longer ensure the indices are within
+    // range of the observable's number of qubits
+    unsafe { &mut obs.indices_mut()[0] }
```
jakelishman: Since `obs` becomes a `&mut SparseObservable`, we've already assumed that the `indices` field as-pointed-to is valid, so the unsafe-ness shouldn't be about taking the index, it should only be about retrieving a mutable reference to `indices`.  Accessing the zeroth element of that will be bounds checked by Rust, but also, we should just be able to do `unsafe { obs.indices_mut() }.as_mut_ptr()`.

(Comment repeats elsewhere in similar methods.)

crates/cext/src/sparse_observable.rs
```diff
@@ -0,0 +1,749 @@
...
+///     QkSparseObservable *canonical = qk_obs_canonicalize(two);
+///
+/// # Safety
+///
+/// Behavior is undefined ``obs`` is not a valid, non-null pointer to a ``QkSparseObservable``.
+#[no_mangle]
+#[cfg(feature = "cbinding")]
+pub unsafe extern "C" fn qk_obs_canonicalize(
+    obs: *const SparseObservable,
+    tol: f64, // no optional arguments in C -- welcome to the ancient past
```
jakelishman: Nor in Rust lol

crates/cext/src/sparse_observable.rs
```diff
@@ -0,0 +1,749 @@
...
+///
+/// # Safety
+///
+/// Behavior is undefined ``obs`` is not a valid, non-null pointer to a ``QkSparseObservable``.
+#[no_mangle]
+#[cfg(feature = "cbinding")]
+pub unsafe extern "C" fn qk_obs_print(obs: *const SparseObservable) {
+    let obs = unsafe { const_ptr_as_ref(obs) };
+    println!("{:?}", obs);
+}
```
jakelishman: We probably need to think a lot more about this interface.  I don't know how Rust's buffering wrapper around `Stdout` is going to interact with other things, like whatever the C program is doing with stdout, and we also probably need to consider how we define the interfaces here such that the output of this string-ification can be properly integrated with the slightly more general stream-based handlers, like `sprintf`, `fprintf` or potentially (in the future) C++'s stream operators.

crates/cext/src/sparse_observable.rs
```diff
@@ -0,0 +1,749 @@
...
+/// # Example
+///     
+///     QkBitTerm bit_term = QkBitTerm_Y;
+///     // cast the uint32_t to char
+///     char label = qk_bitterm_label(bit_term);
+#[no_mangle]
+#[cfg(feature = "cbinding")]
+pub extern "C" fn qk_bitterm_label(bit_term: BitTerm) -> u32 {
+    // BitTerm is implemented as u8, which is calling convention compatible with C,
+    // hence we can pass ``bit_term`` by value
```
jakelishman: We can pass `BitTerm` by value, but this is super unsafe if C passes us a `u8` that isn't a valid representation of a `BitTerm` lol.  That's part of the larger conversation about how much error-checking we want to do, though.

crates/cext/src/sparse_observable.rs
```diff
@@ -0,0 +1,749 @@
...
+    bit_term
+        .py_label()
+        .chars()
+        .next()
+        .expect("Label has exactly one character") as u32
+}
```
jakelishman: C's `char` is probably the most natural thing to return out of this, which is a _super_ awkward type.  Perhaps it might be a bit easier to turn the `&str` of the `py_label` into a `&[u8]`, using the `expect` to enforce that that's only a single byte (since we know they all are - they're ASCII characters) and returning the only one as a `u8`?

jakelishman (commented):
Thanks to both of you for all this work.

I wouldn't say this is a fully in-depth review - I didn't even attempt to build it locally myself yet - I just wanted to start getting some questions out there to move it along.  I will pull it down locally in the coming days and try to do some local playing myself, which hopefully won't lead to _too_ many further questions.

I wouldn't pretend to have much familiarity with CMake or really making large-scale C projects at all, so at least some of my review on those stuff is just going to be vibes and not knowledge - I'll ask some questions, but you guys probably have more experience from looking into it than I do, so they're just questions.

I wouldn't mind a bit of name bikeshedding on the rewrite rules, if anyone's up for it: I've always been partial to the CPython API conventions of `<Project><Type>_<Function>` for functions where `<Function>` is PascalCase, e.g. `PyList_New` (c.f. `QkObs_New`) - it's a little less noisy than full snake_case, which matters more when there's no module-based namespacing in C.  If we're using `obs` as the prefix for `SparseObservable`-related functions, I like the idea of the class just being `QkObs` (and `QkObsTerm`) in a similar vein, though I'm also fine with `QkSparseObservable`.  I don't feel super strongly about any of this, just loose preference and thought I'd bring it up.

For all `unsafe` blocks: it'd be good to get into the habit of having `// SAFETY:` comments on top of all of them: the more we write, the more likely it is that we forget some of the invariants, or inadvertently add to them and think it's obvious from context, etc.  I think there's a clippy lint against it we might consider turning on - it means a bit more boilerplate to write for us, but my rough feeling atm is it's better to have it.  Rust `unsafe` is even harder to write than C code, because there's more invariants necessary to satisfy the higher-level Rust objects.

tools/run_clang_format.sh
Cryoris: it took some back and forth but the following now works
```
set -e  # set this in case one of the earlier commands fails

# ... get the files...

status=0
for file in $files; do
    # we don't want to exit prematurely but process all files
    if ! clang-format --style="file:$style" $args $file; then
        status=1
    fi
done
exit $status
```

Makefile
Cryoris: I'll add the `cclean` though the dependencies seems a little hacky to do since we'd like to depend on everything inside `cargo/cext` right? Maybe something like 
```
cheader: $(find crates/cext)
```
? I'll leave as is for now 🙂

crates/cext/cbindgen.toml
Cryoris: Summarizing our offline discussion, we use this since cbinding doesn't automatically translate `Complex64` to `double complex`.

releasenotes/notes/c-sparse-observable-c14b17a9303b1c34.yaml
Cryoris: I think you're right -- also `<>` seems more professional 🙂

crates/cext/src/sparse_observable.rs
Cryoris: I'll add some more info, let me know if this is what you had in mind (once committed) 

_Edit:_ It's updated now 🙂

crates/cext/src/sparse_observable.rs
Cryoris: That's legacy code -- initially we tried to avoid `usize` and tried to only use explicitly sized things objects. I'll update

crates/cext/src/sparse_observable.rs
Cryoris: Yeah that's a good point, it doesn't interact well I think and things are not printed in order if one does
```c
qk_obs_print(obs);
printf("Some print statements are more equal than others");
```

crates/cext/src/sparse_observable.rs
Cryoris: Regarding the `BitTerm` value safety: If the C signature shows `BitTerm` in the signature, which is a well-defined enum, it seems legit to ask of the user that it should be such a value enum instead of passing arbitrary `u8`s 🤔

crates/cext/src/sparse_observable.rs
Cryoris: How about something like
```
/// # Safety
///
/// This function requires ``ptr`` to be non-null, aligned, initialized object of
/// type ``T``. While the resulting reference exists, the memory pointed to must not
/// be accessed otherwise.
```
(potentially strongly motivated of the `as_mut()` docstring) 😄

crates/cext/Cargo.toml
```diff
@@ -0,0 +1,26 @@
...
+[lints]
+workspace = true
+
+[dependencies]
+thiserror.workspace = true
+num-complex.workspace = true
+qiskit-accelerate.workspace = true
+
+[build-dependencies]
+cbindgen = "0.26"
```
mtreinish: Our MSRV is now 1.79 for 2.0, so we can upgrade this to the latest release (I assume that's why you held this back).
```suggestion
cbindgen = "0.28"
```

crates/cext/src/sparse_observable.rs
```diff
@@ -0,0 +1,801 @@
...
+/// # Safety
+///
+/// Behavior is undefined if ``obs`` is not either null or a valid pointer to a
+/// [SparseObservable].
+#[no_mangle]
+#[cfg(feature = "cbinding")]
+pub unsafe extern "C" fn qk_obs_free(obs: *mut SparseObservable) {
+    if !obs.is_null() {
+        if !obs.is_aligned() {
+            panic!("Attempted to free a non-aligned pointer.")
```
mtreinish, line 230: Should we have a panic if it's a `null` pointer too with an error message that says as much?

crates/cext/src/sparse_observable.rs
```diff
@@ -0,0 +1,801 @@
...
+#[no_mangle]
+#[cfg(feature = "cbinding")]
+pub unsafe extern "C" fn qk_obs_print(obs: *const SparseObservable) {
+    // SAFETY: Per documentation, the pointer is non-null and aligned.
+    let obs = unsafe { const_ptr_as_ref(obs) };
+    println!("{:?}", obs);
+}
+
+/// @ingroup QkSparseTerm
+/// Print the sparse term.
```
mtreinish: I would probably add:

```suggestion
/// Print the sparse term to stdout.
```

To make it clear that it's printing to stdout.

crates/cext/src/sparse_observable.rs
```diff
@@ -0,0 +1,801 @@
...
+) -> bool {
+    // SAFETY: Per documentation, the pointers are non-null and aligned.
+    let obs = unsafe { const_ptr_as_ref(obs) };
+    let other = unsafe { const_ptr_as_ref(other) };
+
+    obs.eq(other)
+}
+
+/// @ingroup QkSparseObservable
+/// Print the observable.
```
mtreinish: I would probably add:

```suggestion
/// Print the observable to stdout.
```

To make it clear that it's printing to stdout.

crates/cext/README.md
mtreinish, line 1: I would expand this a bit to include some of the details from the release notes. Basic building and usage instructions would be really good to have here.

releasenotes/notes/c-sparse-observable-c14b17a9303b1c34.yaml
mtreinish: Yeah, I think it should be in the public docs somewhere, but also we should put this in the cext crate readme. Right now we don't have a readme for the c api either.

releasenotes/notes/c-sparse-observable-c14b17a9303b1c34.yaml
```diff
@@ -0,0 +1,48 @@
+---
+features:
+  - |
+    Introduced a C API to build and interact with sparse observables.
+    This API does not yet allow for an end-to-end workflow in C, but rather lays the 
+    foundation for Qiskit's C interface. The (dynamic) C library is compiled by Rust and
```
mtreinish: I really dislike the phrasing "end-to-end workflow", it's not clear what that is in Python so it's not really somethign we have as a goal for the C API.

```suggestion
    While the API surface in this release is fairly small, just covering the :class:`.SparseObservable` class
    this new API lays the foundation for Qiskit's C interface. The (dynamic) C library is compiled by Rust and 
```

test/c/CMakeLists.txt
```diff
@@ -0,0 +1,42 @@
+# This file is inspired by the description of handling a larger number of tests
+# provided here:
+# https://cmake.org/cmake/help/book/mastering-cmake/chapter/Testing%20With%20CMake%20and%20CTest.html
+cmake_minimum_required (VERSION 3.27)
```
mtreinish, line 4: Just curious how you picked 3.27 here? It's not a big deal since this is just for testing.

tools/run_clang_format.sh
```diff
@@ -0,0 +1,31 @@
+#! /bin/bash
```
mtreinish, line 1: This should probably have an apache header too right?

mtreinish (commented):
Overall this looks really good. Thanks a lot for all the work on this @Cryoris and @mrossinek.

My main comment after reading through this is that I think we need a bit more user documentation, but that's something we can do in a follow up. I left a few small comments inline, but nothing major on the code front.

crates/cext/Cargo.toml
mrossinek: This was probably just generated at the time of the `0.26` release (i.e. in December which is prior to both later releases). But yes, upgrading

test/c/CMakeLists.txt
mrossinek: @Cryoris can you comment on this? Looks like you added that here: https://github.com/Cryoris/qiskit-terra/commit/029018843b01a50ab92a0cf4d384105827e780be

Cryoris: Oh that was just the version I happened to have installed locally -- we can also change this to something lower, not sure at which point we start losing features we need 🙂

crates/cext/src/sparse_observable.rs
jakelishman: We should most likely remove this function and only have it return a string - having multiple places access `stdout` through different runtimes is going to be a nightmare of locking and buffering, I'd think.  Unless you immediately know best practices here, might be easiest to just return allocated strings, and add to the API later with some stream printer, if we really need it.

tools/install_ubuntu_docs_dependencies.sh
jakelishman: This comment is still valid - `clang-format` isn't a docs-dependency requirement, so I don't think it should be in this script.

test/c/CMakeLists.txt
jakelishman: This comment is still valid.

Cryoris: Good point 👍🏻 Changing this currently breaks on my machine with
```
$ make ctest 
[.... compiling Rust, build header, then CMake starts... ]
Undefined symbols for architecture arm64:
  "_build_test_c_test_driver", referenced from:
      _cmakeGeneratedFunctionMapEntries in test_driver.c.o
ld: symbol(s) not found for architecture arm64
clang: error: linker command failed with exit code 1 (use -v to see invocation)
```
I'll have another look after addressing the remaining changes 🤔

crates/cext/src/sparse_observable.rs
Cryoris: I didn't add this since it seems valid C code to free null pointers (so nothing happens), but we can also add it for code validation

crates/cext/src/sparse_observable.rs
Cryoris: Yeah this bit maybe needs some updating... we currently (.. or ever?) cannot coherently print from Rust and C: interleaving print statements like
```
printf("observable a:\n");
qk_obs_print(a);
printf("observable b:\n");
qk_obs_print(b);
```
prints first all Rust messages and then C (or possibly the other way around I can't quite remember right now). So with @jakelishman's comments from above it might be the better solution to just have a function to get the C-string representation of the observable, which can be printed with `printf` from C

Makefile
jakelishman: Probably just discard the second comment - it somewhat depends on what `cbindgen` is doing anyway.  We're going to have these sorts of slightly inefficiencies when we're mixing build systems.

In Make land, you'd specify the set of intermediate objects that are directly input into building the `cheader` (no matter how large that was, though if it's huge, you'd use phony subtargets), but since the actual rule here is probably doing _something_ and eventually calling `cargo` (another build system), there's likely nothing sensible to do.

crates/cext/src/sparse_observable.rs
mtreinish: I just don't like leaving it undefined behavior. I'd rather just explicitly error even if C doesn't care.

releasenotes/notes/c-sparse-observable-c14b17a9303b1c34.yaml
Cryoris: Yeah we plan on adding detailed information on building in the docs, done in a follow-up. It sounds good to also put this info in the `crates/cext` readme, I'll add it 👍🏻 

> Right now we don't have a readme for the c api either.

Here you mean a section in the root readme?

crates/cext/Cargo.toml
Cryoris: Updated 👍🏻

crates/cext/src/sparse_observable.rs
Cryoris: Updated to `u8` in bc376c6

crates/cext/src/sparse_observable.rs
Cryoris: Updated in bc376c6

tools/install_ubuntu_docs_dependencies.sh
Cryoris: I moved it into a separate file in a2ea67e (and I hope I updated the azure workflow accordingly, let's see if CI is still happy)

tools/verify_headers.py
Cryoris: This one should be addressed in https://github.com/Qiskit/qiskit/commit/bc376c6f27b20c932c7eb4e3da7e63ffb9cbf2ea

releasenotes/notes/c-sparse-observable-c14b17a9303b1c34.yaml
Cryoris: I removed this bit and added some explanation in `crates/cext/README.md` but yes, this will go into the docs 👍🏻

crates/cext/Cargo.toml
Cryoris: Done in https://github.com/Qiskit/qiskit/commit/bc376c6f27b20c932c7eb4e3da7e63ffb9cbf2ea

crates/cext/src/sparse_observable.rs
Cryoris: Updated to use the value, not the pointer, in https://github.com/Qiskit/qiskit/commit/bc376c6f27b20c932c7eb4e3da7e63ffb9cbf2ea

crates/cext/src/sparse_observable.rs
Cryoris: Added in https://github.com/Qiskit/qiskit/commit/bc376c6f27b20c932c7eb4e3da7e63ffb9cbf2ea

crates/cext/src/sparse_observable.rs
Cryoris: Changed in https://github.com/Qiskit/qiskit/commit/bc376c6f27b20c932c7eb4e3da7e63ffb9cbf2ea

crates/cext/src/sparse_observable.rs
Cryoris: Updated in https://github.com/Qiskit/qiskit/commit/bc376c6f27b20c932c7eb4e3da7e63ffb9cbf2ea

crates/cext/src/sparse_observable.rs
Cryoris: Done in https://github.com/Qiskit/qiskit/commit/bc376c6f27b20c932c7eb4e3da7e63ffb9cbf2ea

crates/cext/src/sparse_observable.rs
Cryoris: Added in https://github.com/Qiskit/qiskit/commit/bc376c6f27b20c932c7eb4e3da7e63ffb9cbf2ea, let me know if this is what you had in mind 🙂

crates/cext/src/sparse_observable.rs
Cryoris: We can add some error checking -- or can we argue that the signature expects a `BitTerm` (which is a defined enum) so passing something else is undefined? Or doesn't that work because of the `typedef uint_8 QkBitTerm`?

Cryoris:
Thanks for all the comments! We addressed almost all of them, the main ones that are still open I think are:
* don't print anything from Rust, instead return a string
* documentation (do only README in this PR, full docs will be a follow up)
* check if everyone is happy with the `SAFETY` bits and validation

Plus some small pieces like updates to the CMakeLists, or checking if we can put the library into `dist/c/lib`?

Exposing the C bindings from the same shared library as the Python extension is something we plan as follow up, right? Or would you prefer this to be in the same PR?

mtreinish:
> Thanks for all the comments! We addressed almost all of them, the main ones that are still open I think are:
> 
>     * don't print anything from Rust, instead return a string

I pushed up a small commit adding this: https://github.com/Qiskit/qiskit/pull/13694/commits/603ae56df0038d97b371ea720eed75226c48c9b4 someone should check it to make sure it's sane.

> 
> Plus some small pieces like updates to the CMakeLists, or checking if we can put the library into `dist/c/lib`?
> 
> Exposing the C bindings from the same shared library as the Python extension is something we plan as follow up, right? Or would you prefer this to be in the same PR?

I think lets do this as a small follow-up. We co-authored the commit already to do this already, and it's like 3 lines in `pyext` plus the python helper function(s). It's should be quick to review and I think this is large enough already as a standalone.

test/c/CMakeLists.txt
mtreinish: It doesn't necessarily need to be lower. I was just curious how you settled on that. I'd say we should be conservative if this was needed for building more broadly, but since it's only for testing it should be fine as is.

crates/cext/src/sparse_observable.rs
mtreinish: I pushed up a commit modifying this methods to return strings: https://github.com/Qiskit/qiskit/pull/13694/commits/603ae56df0038d97b371ea720eed75226c48c9b4 someone else should validate they look ok though.

crates/cext/src/sparse_observable.rs
jakelishman: `free(NULL)` is well defined to do nothing in C.  It's not so much "doesn't care" as "explicitly not an error".

test/c/CMakeLists.txt
jakelishman: Fwiw, [the CMake documentation says](https://cmake.org/cmake/help/latest/command/file.html#glob):

> We do not recommend using GLOB to collect a list of source files from your source tree. If no CMakeLists.txt file changes when a source is added or removed then the generated build system cannot know when to ask CMake to regenerate.

Let's not worry about this for now, but I'll open an issue to track as a future improvement to the build system.  #13932.

crates/cext/src/sparse_observable.rs
jakelishman: in our case, it's safe to call the function with a null or misaligned pointer, because those will error safely. It's only the "must point to an initialised valid object of type `T`" bit that's unsafe in our version

crates/cext/src/sparse_observable.rs
jakelishman: The members of an `enum` in C are declared as constants with type `int`.  I _think_ it's "unspecified behaviour", _not_ "undefined behaviour", to pass a value that's not an enum member.  A compiler may warn against it, but it's not required to, and it's not invalid ("undefined") behaviour.

In Rust, it is absolutely critical that we never construct a `BitTerm` out of a bit pattern that isn't a valid member of it - it's instant undefined behaviour, even if we don't read it.  We need to make that 200% clear in the C-API interface: it's undefined behaviour to ever write an invalid `BitTerm` value into the data.

jakelishman: Looking at the generated cbindgen file: `QkBitTerm` is overridden to be `typedef uint8_t QkBitTerm` after the `enum` keyword is used to define the constants.  So from C's perspective, at the time of declaration, the type `QkBitTerm` just straight-up _is_ a `u8`, so we really need to make this clear.

crates/cext/src/sparse_observable.rs
jakelishman: I think the safety comment you added is fine for now - it's clearly documenting that it's undefined behaviour to pass an invalid value.

Makefile
```diff
@@ -83,3 +83,28 @@ coverage_erase:
...
+ctest: cheader
+	# -S specifically specifies the source path to be the current folder
+	# -B specifically specifies the build path to be inside test/c/build
+	cmake -S. -Btest/c/build
+	cmake --build test/c/build
+	# -V ensures we always produce a logging output to indicate the subtests
+	ctest -V --test-dir test/c/build
+
+cclean:
+	rm -r dist/c
```
jakelishman: Possibly this should clean `test/c/build` too?

Makefile
```diff
@@ -12,7 +12,7 @@
 
 OS := $(shell uname -s)
 
-.PHONY: default ruff env lint lint-incr style black test test_randomized pytest pytest_randomized test_ci coverage coverage_erase clean
+.PHONY: default ruff env lint lint-incr style black test test_randomized pytest pytest_randomized test_ci coverage coverage_erase clean cheader ctest
```
jakelishman:
```suggestion
.PHONY: default ruff env lint lint-incr style black test test_randomized pytest pytest_randomized test_ci coverage coverage_erase clean cheader ctest cclean cformat fix_cformat
```

crates/cext/cbindgen.toml
jakelishman, line 8: So in theory we can have `cbindgen` generate the _correct_ definitions of these two types if we pass `parse.includes = ["num-complex"]`.

In practice, that's going to make the C API a right pain to use for complex numbers, because then you have to do everything in terms of the structs.  I think we're probably right to do some sort of hackery like this, but we need to take absolute care that we never accept or return. one of these by value.

crates/cext/README.md
```diff
@@ -0,0 +1,60 @@
+# `qiskit-cext`
+
+This crate contains the bindings for Qiskit's C API. 
+
+## Building the library
+
+The C bindings are compiled into a shared library, which can be built along with the header file
+by running
+```
+$ make cheader
```
jakelishman:
```suggestion
```bash
make cheader
```

crates/cext/README.md
```diff
@@ -0,0 +1,60 @@
...
+return 0;
+}
+```
+Refer to the C API documentation for more information and examples.
+
+## Compiling
+
+The above program can be compiled by including the header and linking to the `qiskit_cext` library,
+which is located in `target/release`. (The exact name depends on the platform, e.g.,
+`target/release/libqiskit_cext.dylib` on MacOS.)
```
jakelishman: We can follow up on this, but we should have a Makefile command that builds `libqiskit_cext.so` and then copies it into `dist/c/lib` with the name `libqiskit.{so,dylib,dll}` as appropriate.  The "cext" name is an internal implementation detail, and we should be building the object into the distribution location, along with the header file.

requirements-dev.txt
```diff
@@ -1,4 +1,4 @@
-# Requirements to develop Terra, and the minimum needed to run its CI.  All
+# Requirements to develop Qiskit, and the minimum needed to run its CI.  All
```
jakelishman, line 1: boooooooooooooo

test/c/test_sparse_observable.c
```diff
@@ -0,0 +1,650 @@
...
+int test_zero() {
+    QkSparseObservable *obs = qk_obs_zero(100);
+    uint64_t num_terms = qk_obs_num_terms(obs);
+    uint32_t num_qubits = qk_obs_num_qubits(obs);
+    qk_obs_free(obs);
+
+    if (num_terms != 0 || num_qubits != 100) {
+        return EqualityError;
+    }
+    return Ok;
```
jakelishman: Longer term we might want to think more about patterns here - maybe we make `TestResult` a flags enum, and update some exit-code variable on the fly, like
```c
int test_zero()
{
    TestResult exit = Ok;
    QkSparseObservable *obs = qk_obs_zero(100);
    if (qk_obs_num_terms(obs) != 0) {
        exit |= EqualityError;
    }
    if (qk_obs_num_qubits(obs) != 100) {
        exit |= EqualityError;
    }

cleanup:
    qk_obs_free(obs);
    return exit;
}
```
or similar, so the test logic can stay all up top, next to where it matters.  This simple example obviously doesn't need it, but if we needed early termination from the code, we can then `goto cleanup;` after setting the return code, etc.

No need to edit now.

test/c/test_sparse_observable.c
```diff
@@ -0,0 +1,650 @@
...
+    qk_obs_free(obs100);
+
+    return result;
+}
+
+/**
+ * Test adding an individual term to an observable.
+ */
+int test_custom_build() {
+    u_int32_t num_qubits = 100;
```
jakelishman: Pretty sure that `u_int32_t` is some weird BSDish extension.  I think we might just be getting lucky that Apple clang and the Ubuntu gcc support it as well - I don't think it's a required part of `stdint.h`.

test/c/test_sparse_observable.c
```diff
@@ -0,0 +1,650 @@
...
+        qk_obs_term(obs, i, &view);
+        size_t nni = view.len;
+        nnis[i] = nni; // store to compare later
+
+        for (uint32_t n = 0; n < nni; n++) {
+            // this loop is only called once, so we can use ``n`` to index here
+            bits[n] = view.bit_terms[n];
+            indices[n] = view.indices[n];
+        }
+    }
```
jakelishman: There's some pretty weird typing going on here.  Everything's `int` up top, then it's assigned various other values, etc.  Might be good to get some linting set up on this.

Can be a follow up.

test/c/test_sparse_observable.c
```diff
@@ -0,0 +1,650 @@
...
+    QkBitTerm bit_terms[3] = {QkBitTerm_X, QkBitTerm_Y, QkBitTerm_Z};
+    uint32_t qubits[3] = {0, 1, 2};
+    complex double coeff = 1 + I;
+
+    QkSparseTerm term = {coeff, 3, bit_terms, qubits, num_qubits};
+    int err = qk_obs_add_term(obs, &term);
+
+    if (err != 0) {
+        return RuntimeError;
+    }
```
jakelishman: This is one place where we early return without cleaning up, so the `goto cleanup;` pattern I'd suggested earlier _might_ be appropriate.

That said: do we know how many separate processes CTest runs? Does it do every function in a separate process?  If it does: it's maybe not really any trouble if we "leak" memory, since it's all reclaimed on process termination anyway, and the memory use of the test is clearly well bounded.  That said, it's quite conceivable we might want to run the test suite under very address / memory / behaviour sanitisers in the future, which would fail us if we leak.

test/c/test_sparse_observable.c
```diff
@@ -0,0 +1,650 @@
...
+
+    // copy the term so we can safely modify it and add it onto the observable
+    size_t len = borrowed.len;
+    QkBitTerm *copied_bits = (QkBitTerm *)malloc(len * sizeof(QkBitTerm));
+    uint32_t *copied_indices = (uint32_t *)malloc(len * sizeof(uint32_t));
+    for (size_t i = 0; i < len; i++) {
+        copied_bits[i] = borrowed.bit_terms[i];
+        copied_indices[i] = borrowed.indices[i];
+    }
```
jakelishman, line 314: typically in C you use `memcpy` for this, which is in `string.h` for some reason

No need to change, just saying.

test/c/test_sparse_observable.c
```diff
@@ -0,0 +1,650 @@
...
+    QkSparseTerm borrowed;
+    int error = qk_obs_term(obs, 0, &borrowed); // get view on 0th term
+    if (error > 0) {
+        return RuntimeError;
+    }
+
+    // copy the term so we can safely modify it and add it onto the observable
+    size_t len = borrowed.len;
+    QkBitTerm *copied_bits = (QkBitTerm *)malloc(len * sizeof(QkBitTerm));
+    uint32_t *copied_indices = (uint32_t *)malloc(len * sizeof(uint32_t));
```
jakelishman, line 310: Given we know the expected sizes here, if the dynamic memory allocation gets too fiddly with cleanups, etc, we could move towards patterns where we declare `QkBitTerm copied_bits[4]` or so on the stack, then assert that `borrowed.len` is what we expect, and then we don't need to bother with `free` because it's stack-allocated.

Just commenting for consideration in future testing patterns.

test/c/test_sparse_observable.c
```diff
@@ -0,0 +1,650 @@
...
+    qk_obs_add_term(expected, &term2);
+
+    bool equal = qk_obs_equal(expected, obs);
+    qk_obs_free(obs);
+    qk_obs_free(expected);
+
+    if (!equal) {
+        return EqualityError;
+    }
+    return Ok;
```
jakelishman: Super minor, but if you like: if you're doing this a lot, C has a ternary conditional operator that works similar to Python's `a if x else b`.  It looks like `x ? a : b`, so you can have `return equal ? Ok : EqualityError;`.

No need to change.

crates/cext/src/sparse_observable.rs
```diff
@@ -0,0 +1,849 @@
...
+    Ok(())
+}
+
+/// Check a const pointer is non-null and aligned, then cast it to a reference.
+///
+/// # Safety
+///
+/// This function requires ``ptr`` to be non-null, aligned, initialized object of
+/// type ``T``. While the resulting reference exists, the memory pointed to must not
+/// be mutated.
```
jakelishman: It's not unsafe to pass a null pointer or a misaligned pointer - we check those and handle them safely.  Worth documenting, but those two bits don't need to be in the safety section.

(Comment repeats wherever this comment does.)

crates/cext/src/sparse_observable.rs
```diff
@@ -0,0 +1,849 @@
...
+///     uint64_t num_bits = 4; // we have 4 non-identity bits: 0, 1, +, -
+///
+///     complex double coeffs[2] = {1, -1};
+///     QkBitTerm bits[4] = {QkBitTerm_Zero, QkBitTerm_One, QkBitTerm_Plus, QkBitTerm_Minus};
+///     uint32_t indices[4] = {0, 1, 98, 99};  // <-- e.g. {1, 0, 99, 98} would be invalid
+///     size_t boundaries[3] = {0, 2, 4};
+///
+///     QkSparseObservable *obs = qk_obs_new(
+///         num_qubits, num_terms, num_bits, coeffs, bits, indices, boundaries
+///     );
```
jakelishman: Super minor documentation stuff (no need to change now): it'd be cool to use GitHub-style fenced code blocks, assuming Doxygen supports them (I think it does), if not only because it's easier to see where the code is, and my editor will highlight the internal code for me ;)

crates/cext/src/sparse_observable.rs
```diff
@@ -0,0 +1,849 @@
...
+///
+/// Behavior is undefined if any of the following conditions are violated:
+///
+///   * ``coeffs`` is a pointer to a ``complex double`` array of length ``num_terms``
+///   * ``bit_terms`` is a pointer to a ``QkBitTerm`` array of length ``num_bits``
+///   * ``indices`` is a pointer to a ``uint32_t`` array of length ``num_bits``, which is
+///     term-wise sorted in strict ascending order, and every element is smaller than ``num_qubits``
+///   * ``boundaries`` is a pointer to a ``size_t`` array of length ``num_terms + 1``, which is
+///     sorted in ascending order, the first element is 0 and the last element is
+///     smaller than ``num_terms``
```
jakelishman: It would be good to make clear in here that all the items in `bit_terms` _must_ be valid `BitTerm`s, since in C land, the type is just an alias to `uint8_t`, so tooling won't help.

crates/cext/src/sparse_observable.rs
```diff
@@ -0,0 +1,849 @@
...
+///
+/// Behavior is undefined ``obs`` is not a valid, non-null pointer to a ``QkSparseObservable``.
+#[no_mangle]
+#[cfg(feature = "cbinding")]
+pub unsafe extern "C" fn qk_obs_bit_terms(obs: *mut SparseObservable) -> *mut BitTerm {
+    // SAFETY: Per documentation, the pointer is non-null and aligned.
+    let obs = unsafe { mut_ptr_as_ref(obs) };
+
+    obs.bit_terms_mut().as_mut_ptr()
+}
```
jakelishman: We need to be super clear in here that you _must not_ write invalid `BitTerm` values into the pointer - it's pretty immediate undefined behaviour.

crates/cext/src/sparse_observable.rs
```diff
@@ -0,0 +1,849 @@
...
+/// # Safety
+///
+/// Behavior is undefined if any of the following is violated
+/// * ``obs`` is a valid, non-null pointer to a ``QkSparseObservable``
+/// * ``coeff`` is a valid, non-null pointer to a ``complex double``
+#[no_mangle]
+#[cfg(feature = "cbinding")]
+pub unsafe extern "C" fn qk_obs_multiply(
+    obs: *const SparseObservable,
+    coeff: *const Complex64,
```
jakelishman, line 592: A shame that Rust doesn't expose a `std::ffi::c_complex_double`, but I guess it's not possible since there's no built-in complex type.

crates/cext/src/sparse_observable.rs
```diff
@@ -0,0 +1,849 @@
...
+/// @ingroup QkSparseObservable
+/// Multiply the observable by a complex coefficient.
+///
+/// @param obs A pointer to the observable.
+/// @param coeff The coefficient to multiply the observable with.
+///
+/// # Example
+///
+///     QkSparseObservable *obs = qk_obs_identity(100);
+///     QkSparseObservable *result = qk_obs_multiply(obs, 2);
```
jakelishman: I don't think this example is correct.  You're passing an integer 2 as a `*const Complex64` - in C land, you'll be passing the pointer with value 2, which is neither aligned nor valid.

Perhaps we ought to consider having `_multiply_real(*const SparseObservable, f64)` and `_multiply_complex(*const SparseObservable, f64, f64)` or something, where in the latter case you pass the real and imaginary parts separately?  I don't love it.

crates/cext/src/sparse_observable.rs
```diff
@@ -0,0 +1,849 @@
...
+///
+/// # Safety
+///
+/// Behavior is undefined ``obs`` is not a valid, non-null pointer to a ``QkSparseObservable``.
+///
+/// The string must not be freed with the normal C free, you must use ``qk_free_obs_str`` to
+/// free the memory consumed by the String. Not calling ``qk_free_obs_str`` will lead to a
+/// memory leak.
+///
+/// Do not change the length of the string after it's returned, although values can be mutated
```
jakelishman: @mtreinish: Did you mean "don't put an earlier `nul` in"?  I didn't quite understand this at first.

crates/cext/src/sparse_observable.rs
```diff
@@ -0,0 +1,849 @@
...
+#[no_mangle]
+#[cfg(feature = "cbinding")]
+pub unsafe extern "C" fn qk_obs_str(obs: *const SparseObservable) -> *mut c_char {
+    // SAFETY: Per documentation, the pointer is non-null and aligned.
+    let obs = unsafe { const_ptr_as_ref(obs) };
+    let string: String = format!("{:?}", obs);
+    CString::new(string).unwrap().into_raw()
```
jakelishman, line 754: Using debug printing makes quite a mess of the complex numbers.  We should probably refactor the Rust stuff a little so we can use the same machinery as the Python `__repr__`.

Can be a follow-up.

crates/cext/src/sparse_observable.rs
jakelishman: This documentation needs changing now because the function doesn't print any more haha

crates/cext/src/sparse_observable.rs
```diff
@@ -0,0 +1,849 @@
...
+///
+/// @return The function exit code. This is ``>0`` if reading the term failed.
+///
+/// # Example
+///
+///     QkSparseObservable *obs = qk_obs_identity(100);
+///     QkSparseTerm term;
+///     qk_obs_term(obs, 0, &term);
+///     char *string = qk_obsterm_print(&term);
+///     qk_free_obsterm_str(string);
```
jakelishman: The example code isn't right here - wrong function names.

crates/cext/src/sparse_observable.rs
```diff
@@ -0,0 +1,849 @@
...
+/// Free a string representation of a ``SparseObservable``.
+///
+///  @param str A pointer to the returned string representation from ``qk_obs_str``
+///
+/// # Safety
+///
+/// Behavior is undefined if ``str`` is not a pointer returned by ``qk_obs_str``
+#[no_mangle]
+#[cfg(feature = "cbinding")]
+pub unsafe extern "C" fn qk_free_obs_str(string: *mut c_char) {
```
jakelishman: This arguably breaks the naming convention of `qk_<obj>_<fn>` - it should be probably be `qk_obs_free_str`?

crates/cext/src/sparse_observable.rs
```diff
@@ -0,0 +1,849 @@
...
+/// Free a string representation of a ``SparseObservable``.
+///
+///  @param str A pointer to the returned string representation from ``qk_obsterm_str``
+///
+/// # Safety
+///
+/// Behavior is undefined if ``str`` is not a pointer returned by ``qk_obsterm_str``
+#[no_mangle]
+#[cfg(feature = "cbinding")]
+pub unsafe extern "C" fn qk_free_obsterm_str(string: *mut c_char) {
```
jakelishman: If we're using `obsterm` for the object name of `SparseTerm` (fine by me, just pointing it out), the `free` here is similarly breaking the naming convention.

jakelishman (commented):
Thanks for this, Julien and Max!

A couple of other questions beyond those in the review:
- do we have any plan for / idea how to measure coverage of our C API?  It would be good to know if we're exercising it all in tests.
- what plans do we have for linting the test suite?
- do we have plans for running the test suite under some combination of behaviour sanitisers?  I don't know if we'd be able to run the Rust stuff under Miri's interpreter simultaneously from the C test suite, but it would be really good if we could.  If not: perhaps we could look into writing tests in Rust that use the C API itself, so we _can_ run them under Miri?

I think most of the comments I left here say they can be for follow ups.

I pushed up a few bits and bobs of fixes that I saw while I was at it too.

test/c/test_sparse_observable.c
jakelishman: I fixed this in one of my commits.

crates/cext/cbindgen.toml
Cryoris: We investigated this in the beginning, but it seemed less convenient than just the typedef (which I believe is also used in the C++ project)

requirements-dev.txt
Cryoris: hahahaha we can keep the easter egg 😛

test/c/test_sparse_observable.c
Cryoris: Ohh right, I forgot about that one -- thanks!

crates/cext/src/sparse_observable.rs
Cryoris: It seems it does https://www.doxygen.nl/manual/markdown.html ! Also in a fashion that I've never seen code blocks before lol but the triple tick also works 👍🏻

crates/cext/src/sparse_observable.rs
Cryoris: Since this frees any string, maybe we could just have this be `qk_str_free`? Or is this tied to the object the string came from for some reason?

crates/cext/src/sparse_observable.rs
Cryoris: Yeah that is for sure wrong, the correct way would be 
```c
complex double c = 2;
qk_obs_multiply(obs, &c);
```

I agree it would be nice being able to pass by value, and I would assume there's a few other places where we can find syntax improvements 🙂

crates/cext/src/sparse_observable.rs
jakelishman: We can unify them to that - it's not dependent.

crates/cext/cbindgen.toml
jakelishman: The trade-off for convenience is just a lot of discipline - it would be easy to accidentally introduce the type into a by-value argument or return, and that would break because we're lying about the true type here.

Still the right call, I think, but I don't love how we generate some of the file and hard code other bits. Nothing to do here, though.

crates/cext/src/sparse_observable.rs
mtreinish: Or by removing the final `nul`. It's a warning from the `into_raw()` and `from_raw` docs: https://doc.rust-lang.org/std/ffi/struct.CString.html#method.into_raw

jakelishman: the parenthetical expression in that link

> (by writing a nul byte somewhere inside the string or removing the final one)

makes it a fair amount easier to read, I think.

test/c/test_sparse_observable.c
Cryoris: oh yeah, I'll add the missing free here -- the `cleanup` sounds like a good pattern though we'd have to check what to do in places where we allocate more objects during the test, then we don't always have to free the same variables

crates/cext/src/sparse_observable.rs
Cryoris: Good point 👍🏻

crates/cext/src/sparse_observable.rs
Cryoris: This is removed in [624c3dd](https://github.com/Qiskit/qiskit/pull/13694/commits/624c3ddbc49b69caecdd010f721d0acdf91cbb20)

crates/cext/src/sparse_observable.rs
Cryoris: Done in 624c3dd

crates/cext/src/sparse_observable.rs
Cryoris: fixed in 624c3dd

crates/cext/src/sparse_observable.rs
Cryoris: Yeah let's fix this in a follow up 👍🏻

crates/cext/src/sparse_observable.rs
Cryoris: I added that info in 624c3dd

crates/cext/src/sparse_observable.rs
Cryoris: Extended the `Safety` section on this in 624c3dd

crates/cext/src/sparse_observable.rs
Cryoris: Added in 624c3dd

crates/cext/src/sparse_observable.rs
Cryoris: Updated in 624c3dd

test/c/test_sparse_observable.c
Cryoris: This cleaned the code up a bit in 624c3dd 🙂

test/c/test_sparse_observable.c
Cryoris: Updated in all occurences in 624c3dd

test/c/test_sparse_observable.c
Cryoris: Fixed in 624c3dd

crates/cext/src/sparse_observable.rs
Cryoris: Updated in 624c3dd

Makefile
Cryoris: Added in 624c3dd

Makefile
Cryoris: Added in 624c3dd

crates/cext/README.md
Cryoris: Done in 624c3dd

crates/cext/README.md
Cryoris: Let's do this in a follow up 👍🏻

jakelishman: I did this in 6869d1b4bf, so the usage instructions can all be `-lqiskit` etc.

jakelishman (approved):
Thanks for all the work on this everybody.  I'm going to add it to the queue now - I think the general principles are all ok now, the build instructions present a sensible initial platform to build from, and everything else we can go from here.

We've got quite a few follow-ups and ways to improve from here, but I think the right thing to do is get it merged at this point.

PR #13696, opened 2025-01-20
Brand operation indices with `OperationIndex`

jakelishman:
### Summary

Operations on the `DAGCircuit` that return `NodeIndex`s that are known to refer to `Operation` instances now return a transparent wrapper type, `OperationIndex`.  `DAGCircuit` can be indexed with an `OperationIndex`, which returns the `&PackedInstruction` directly, rather than requiring the caller to indirect through the `OperationType` enum, which 99% of the time they don't care about.

`DAGCircuit` methods that are required to act on operations now take `OperationIndex` arguments, shifting the typing validation left; the compiler now effectively tracks provenance of indices in the type system. You can convert a `NodeIndex` into an `OperationIndex` as part of a DAG lookup by using `DAGCircuit::get_operation`, which returns the `&PackedInstruction` (if available), and upgrades the `NodeIndex` into an `OperationIndex` at the same time.  Most of the time this is not necessary, as the methods that can only return operation nodes now return `OperationIndex`.

Most of the changes in this commit are simplifying relevant call sites, and removing `if let NodeType::Operation(inst) = &dag[node_index] { }` constructs, since they were automatically flattened into `let inst = &dag[op_index]`.

### Details and comments

This was the goal I had in mind at the end of my little sequence of patches late last week (#13680, #13681, #13682, #13683).  I don't love the `unsafe` casting of the `Vec` for `collect_1q_runs` etc - perhaps we can extend the rustworkx API to allow a mapping function, so we can do it all in safe no-copy Rust.

A little over one third of the diff of this PR is whitespace-only changes where the level of indent is reduced.  For example, `elide_permutations.rs` claims a diffstat of +49,-52, but with `git diff -b` it reduces to +2,-5.

### DIFF
--- a/crates/circuit/src/dag_circuit.rs
+++ b/crates/circuit/src/dag_circuit.rs
@@ -10,7 +10,8 @@
 // copyright notice, and modified files need to carry a notice indicating
 // that they have been altered from the originals.
 
-use std::hash::Hash;
+use std::hash::{Hash, Hasher};
+use std::ops::Deref;
 
 use ahash::RandomState;
 use smallvec::SmallVec;
@@ -49,7 +50,6 @@ use rustworkx_core::dag_algo::layers;
 use rustworkx_core::err::ContractError;
 use rustworkx_core::graph_ext::ContractNodesDirected;
 use rustworkx_core::petgraph;
-use rustworkx_core::petgraph::prelude::StableDiGraph;
 use rustworkx_core::petgraph::prelude::*;
 use rustworkx_core::petgraph::stable_graph::{EdgeReference, NodeIndex};
 use rustworkx_core::petgraph::unionfind::UnionFind;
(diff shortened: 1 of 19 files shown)


crates/circuit/src/dag_circuit.rs
```diff
@@ -7017,3 +7163,59 @@ fn emit_pulse_dependency_deprecation(py: Python, msg: &str) {
...
     ));
 }
+
+/// Convert a `Vec<NodeIndex>` to a `Vec<OperationIndex>` without copying.
+///
+/// # Safety
+///
+/// The safety guarantees of [OperationIndex::new] must be upheld for each of the values in `val`.
+#[inline]
+unsafe fn cast_node_vec_to_operation_vec(nodes: Vec<NodeIndex>) -> Vec<OperationIndex> {
```
kevinhartman, line 7173: Out of curiosity, is there any chance Rust is smart enough that we don't actually need this? I.e., if we do `nodes.into_iter().map(|n| OperationIndex(n)).collect::<Vec<_>>()`, does that in effect do this anyway?

It looks like `Vec`'s `FromIterator` does something really similar: https://doc.rust-lang.org/beta/src/alloc/vec/spec_from_iter.rs.html#49-58

crates/circuit/src/dag_circuit.rs
```diff
@@ -125,28 +125,86 @@ pub enum NodeType {
...
+///
+/// Constructing this struct is a promise that the `NodeIndex` was checked and is _known_ to refer
+/// to an operation.  Usually, you use [DAGCircuit::get_operation] to do that, though if you
+/// can be certain by other means that it is correct, you may use the `unsafe` function
+/// [`OperationIndex::new`].
+#[derive(Clone, Copy, PartialEq, Eq, PartialOrd, Ord)]
+#[repr(transparent)]
+pub struct OperationIndex(NodeIndex);
+impl OperationIndex {
+    /// Brand a [NodeIndex] as referring to an operation.
```
kevinhartman, line 147: The word "brand" feels a little weird IMO (I think of cattle). Can we pick something else? Perhaps "designate", "mark", or even "downcast"?

crates/circuit/src/dag_circuit.rs
```diff
@@ -125,28 +125,86 @@ pub enum NodeType {
...
+    }
+}
+impl ::hashbrown::Equivalent<NodeIndex> for OperationIndex {
+    fn equivalent(&self, key: &NodeIndex) -> bool {
+        self.0 == *key
     }
 }
 
-#[derive(Hash, Eq, PartialEq, Clone, Debug)]
+#[derive(Hash, Eq, PartialEq, Clone, Copy, Debug)]
```
kevinhartman, line 199: Nice! I think I forgot to add that when doing the `Var` stuff.

crates/circuit/src/dag_circuit.rs
```diff
@@ -4805,11 +4855,51 @@ impl DAGCircuit {
...
+    ///
+    /// Returns `None` if the index is either not present, or not an operation.
+    ///
+    /// The resulting index can be used to index the DAG (`dag[op_index]`) to retrieve the
+    /// same [PackedInstruction] reference returned by this method, rather than the caller needing
+    /// to unwrap the enum.
+    #[inline]
+    pub fn get_operation(&self, index: NodeIndex) -> Option<(OperationIndex, &PackedInstruction)> {
+        match self.dag.node_weight(index) {
+            Some(NodeType::Operation(inst)) => Some((OperationIndex(index), inst)),
```
kevinhartman, line 4878: Should we put `OperationIndex` into its own inner module to force `DAGCircuit` implementors to also reason about using the unsafe `OperationIndex::new`?

kevinhartman (commented):
This is a neat idea and makes working with the op node API surface of `DAGCircuit` much cleaner.

Just a few minor comments.

crates/circuit/src/dag_circuit.rs
jakelishman: I don't know, honestly.  Even if it does on one compiler, I'm not sure if it'd feel comfortable relying on the behaviour, though equally, I don't love this `unsafe` function at all either.

This function is one of my biggest problems with this PR - I think _overall_ it's decent, but the end result isn't fully as nice as I'd hoped.  Not sure what exactly I have a problem with, but it definitely didn't feel like a slam-dunk when I looked at the final diff.

crates/circuit/src/dag_circuit.rs
jakelishman: Yeah, whatever you like, really.  "Mark" is nice and short, "downcast" is consistent with type theory and PyO3.

crates/circuit/src/dag_circuit.rs
jakelishman: Honestly, that bit probably didn't need to be in this same PR - I think I did it because one of the code collapses became a lot neater if `Wire` was copy, but really I could have done that separately.

crates/circuit/src/dag_circuit.rs
jakelishman: Oh yeah, I think that's a really good idea.

Let's hold this PR for a sec, then, and I'll push up another one that makes `qiskit_circuit::dag` and moves the relevant existing files into that, since this change would mean we'll be up to four files that are specific to the DAG (`dag_node.rs`, `dag_circuit.rs`, `rustworkx_core_vnext.rs` and the new index file), and we can start making the imports in `qiskit_accelerate` a little cleaner too.

crates/circuit/src/dag_circuit.rs
jakelishman: > I think _overall_ it's decent, but the end result isn't fully as nice as I'd hoped.

I mean the entire PR, btw.

crates/circuit/src/dag_circuit.rs
jakelishman: Ok, that PR is #13721, so this PR can wait for that one.

Issue #13687, opened 2025-01-18
can remove a little overhead from `SparsePauliOp.copy()`


aeddins-ibm:
`SparsePauliOp.copy()` is inherited from `BaseOperator`, which calls `deepcopy`:
https://github.com/Qiskit/qiskit/blob/524fb4759ee5b45aeb588886b33f6c6472069d08/qiskit/quantum_info/operators/base_operator.py#L143-L145

For small `SparsePauliOp` objects, `deepcopy` seems to have a little more overhead compared to copying the arrays.

The overhead is usually marginal, but it seems like we may as well avoid it by adding these two lines to sparse_pauli_op.py, using the copy routine already present in `__init__`:
```
def copy(self):
    return SparsePauliOp(data=self, copy=True)
```

Quick comparison of repeatedly copying a 2-qubit, 2-term object (e.g. some two-qubit gates could be this size):

```
## Setup:
import numpy as np
from qiskit import quantum_info as qi
from copy import deepcopy
np.random.seed(0)
num_qubits = 2
num_paulis = 2
num_trials = 200000
spo = qi.SparsePauliOp(qi.random_pauli_list(num_paulis, size=num_qubits, seed=0, phase=False),
                    coeffs = np.random.random(num_paulis) + 1j*np.random.random(num_paulis))

## This took 2.8 seconds:
for _ in range(num_trials):
    x = deepcopy(spo)

## This took 0.8 seconds:
for _ in range(num_trials):
    y = qi.SparsePauliOp(data=spo, copy=True)
```

It may be good also to check performance when the SparsePauliOp contains parameters.

jakelishman:
This seems very sensible to me. I _really_ don't like the use of `copy.deepcopy` when it's just a shortcut in library code, since it shifts loads of work onto generic runtime algorithms that have 2 copies rather than one (typically you pickle to byte stream and unpickle that to an object, so you create two things).

I'm not so concerned about performance of the `Parameter` case - using those objects is already rather slow, and I doubt this would make it worse. We can always fix it if it does.

Would you be able to make the PR?

PR #13697, opened 2025-01-21
Add `copy` method to SparsePauliOp, replacing inherited generic `copy` method

aeddins-ibm:
Addresses issue #13687. Shadows generic inherited `copy` method. This avoids overhead of using `deepcopy`.

### DIFF
--- a/qiskit/quantum_info/operators/symplectic/sparse_pauli_op.py
+++ b/qiskit/quantum_info/operators/symplectic/sparse_pauli_op.py
@@ -232,6 +232,10 @@ class SparsePauliOp(LinearOp):
             atol = self.atol
         return np.allclose((self - other).simplify().coeffs, 0.0, atol=atol)
 
+    def copy(self):
+        """Return a copy of the SparsePauliOp."""
+        return SparsePauliOp(self, copy=True)
+
     @property
     def settings(self) -> dict:
         """Return settings."""


jakelishman:
Thanks for doing this Andrew.  Looks like the test failure is indicating that there's some amount of state in the `SparsePauliOp` that we're presumably failing to pass on through the new copy constructor.  If you've got time to try and figure it out, please feel free to, otherwise I can look into it.

aeddins-ibm:
Hmm, this seems more involved than I'd hoped. I proposed this change when I was surprised to see `deepcopy` show up when profiling some code, but since the payoff is marginal (and can be largely avoided once you're aware of it) I'm not sure it's worth putting time into merging this since to me it seems hard to diagnose why the test fails.

For the record, the failing test is in the Neko integration tests (the main tests pass):
```
qiskit_neko.tests.machine_learning.test_neural_network_classifier.TestNeuralNetworkClassifierOnPrimitives.test_neural_network_classifier_1_reference[aer,machine_learning,terra]
```
```
      File "/home/runner/work/qiskit/qiskit/qiskit-neko/qiskit_neko/tests/machine_learning/test_neural_network_classifier.py", line 61, in test_neural_network_classifier
    self.assertGreaterEqual(score, 0.5)
```
```
    AssertionError: 0.45 not greater than or equal to 0.5
```

The score is almost good enough, so I wonder if it could just be some non-deterministic failure with a random seed not being set somewhere, and this test run got unlucky?

Or else some issue inside the Sampler -- the last "warning" message in the test output suggests using SparsePauliOp.apply_layout, so evidently something related to SparsePauliOp is happening, which could plausibly be affected by this PR.

jakelishman:
I suspect that the true problem is that something isn't getting properly copied over: either there's some data being shared between the two and it's getting mutated later (so affects an object it shouldn't), or there's something that isn't getting copied across at all, changing the meaning of the operator.  I'll try and have a look into it, because 3x runtime improvement in copy performance is definitely worth merging if we can.

aeddins-ibm:
Thanks Jake. Just to be clear since I don't want to waste your time, I think the time improvement is only for very small SparsePauliOps, becoming negligible for medium-to-large SparsePauliOps. By the example I posted in the issue, it's roughly 2 seconds of cumulative overhead from 200,000 repetitions, so like ~10us of overhead per call. I think that overhead is roughly independent of object size. So it can matter when copying many small objects, where the more efficient copy operation apparently takes only ~4us, but won't always matter.

PR #11684, opened 2024-01-31
Remove `Unroller` entry-point registration

jakelishman:
### Summary

The `Unroller` pass and its associated plugin was removed in gh-11581, but this registration was overlooked (not surprisingly - it's very easy to miss!).  This is causing the test suite to log warnings at the moment during plugin evaluation.

PR #11581, opened 2024-01-17
Remove deprecated feature in 0.45 in unroller

henryzou50:
### Summary
Removes the class deprecated in 0.45 in qiskit/transpiler/passes/basis/unroller.py

### Details and comments
`Unroller` is going to be removed in Qiskit 1.0 and has been replaced by the `BasisTranslator` pass. However, using `BasisTranslator` as a direct replacement for `Unroller` is a bit unclear. `BasisTranslator` takes in an `EquivalenceLibrary` as one of its arguments, where `Unroller` does not. To adjust the tests, I mainly used:
`from qiskit.circuit.library.standard_gates.equivalence_library import (
    StandardEquivalenceLibrary as std_eqlib,
)`
as the `EquivalenceLibrary`. While most of the tests were successful, a few tests fail as there needs to be additional equivalence rules, so I am not sure what is the best way to approach this.

PR #13698, opened 2025-01-21
remove mentioning ``unroller`` translation in docstrings

alexanderivrii:
### Summary

The ``Unroller`` pass was removed in #11581, its registration as an entrypoint was removed in #11684, it's probably time to remove it from being mentioned in the documentation (e.g. see https://docs.quantum.ibm.com/api/qiskit/transpiler_plugins).

### DIFF
--- a/qiskit/compiler/transpiler.py
+++ b/qiskit/compiler/transpiler.py
@@ -191,7 +191,7 @@ def transpile(  # pylint: disable=too-many-return-statements
             This can also be the external plugin name to use for the ``routing`` stage.
             You can see a list of installed plugins by using :func:`~.list_stage_plugins` with
             ``"routing"`` for the ``stage_name`` argument.
-        translation_method: Name of translation pass ('translator', 'synthesis')
+        translation_method: Name of translation pass (``"translator"`` or ``"synthesis"``)
             This can also be the external plugin name to use for the ``translation`` stage.
             You can see a list of installed plugins by using :func:`~.list_stage_plugins` with
             ``"translation"`` for the ``stage_name`` argument.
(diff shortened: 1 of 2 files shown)


gadial:
Just to clarify things for me. In this one:
```
class Instruction(Operation):
    """Generic quantum instruction."""

    # Class attribute to treat like barrier for transpiler, unroller, drawer
    # NOTE: Using this attribute may change in the future (See issue # 5811)
```

Is the "unroller" referring to what is now the basis translator?

Also, in `circuit.add_control` we have
```
        unrolled_gate = _unroll_gate(operation, basis_gates=basis)
        if unrolled_gate.definition.global_phase:
            global_phase += unrolled_gate.definition.global_phase
```
Where `_unroll_gates` calls both `BasisTranslator` and `UnrollCustomDefinitions`. Should the latter be referred to somehow in the documentaion? It is mentioned in the docstring of `control`: `it calls the unroller to convert to u1, u3,and cx gates.`

alexanderivrii:
Yeah, I've also done `grep` (😄) and thought that while the above comment and internal function docstring (it's internal because afaik it does not appear anywhere in the documentation) are not the most accurate, fixing these has no effect on the documentation.

In fact, the docstrings for most of the functions in `add_control.py` seem quite outdated (e.g., "Currently, these are ``x``, ``rx``, ``ry``, and ``rz``.") and I agree that it would be nice to eventually improve these.

The two usages of ``unroller`` that I removed correspond to ``"unroller"`` previously being the name of a transpiler translation stage plugin (which was removed in the two PRs linked above).

qiskit/compiler/transpiler.py
```diff
@@ -191,7 +191,7 @@ def transpile(  # pylint: disable=too-many-return-statements
             This can also be the external plugin name to use for the ``routing`` stage.
             You can see a list of installed plugins by using :func:`~.list_stage_plugins` with
             ``"routing"`` for the ``stage_name`` argument.
-        translation_method: Name of translation pass ('unroller', 'translator', 'synthesis')
+        translation_method: Name of translation pass ('translator', 'synthesis')
```
Cryoris: While we're at it, we could make it consistent with the rest of the docstring
```suggestion
        translation_method: Name of translation pass (``"translator"`` or ``"synthesis"``)
```

alexanderivrii: Done!

Cryoris (approved):

PR #13702, opened 2025-01-21
Remove errant `print` statements from tests

jakelishman:
If any `print` is necessary for debugging failed test cases, we probably ought to reconfigure the test runner's display mechanisms, or use the built-in configuration of the `unittest` assert methods to produce better results.


ElePT (approved):
Absolutely. Probably would be good to backport too?

PR #13704, opened 2025-01-21
Oxidize `UnitarySynthesis` path using `basis_gates`

ElePT:
### Summary

This PR is a compilation of small improvements to the `UnitarySynthesis` pass, including:

1. **Performance** -> Addition of fast path (Rust) for the use case where `target is None`, all input combinations can now be run using the fast path.

2. **Fixes** ->  Fixed an oversight where the `pulse_optimize`  argument wasn't added to the fast path and therefore ignored (discovered after expanding the unit tests).

These set the stage for future refactorings of the `UnitarySynthesis` class to take better advantage of the plugin model (the default plugin is currently ignored in the fast path). 

Note that this PR touches on code that is currently actively worked on in:

- #12850 
- #13568 

And builds on top of #13706 (will need to be rebased after it's merged).


crates/accelerate/src/unitary_synthesis.rs
```diff
@@ -59,48 +59,61 @@ enum DecomposerType {
...
+    decomp_gate: StandardGate,
+    decomp_params: SmallVec<[Param; 3]>,
 }
 
 // Used in get_2q_decomposers. If the found 2q basis is a subset of GOODBYE_SET,
 // then we know TwoQubitBasisDecomposer is an ideal decomposition and there is
 // no need to bother trying the XXDecomposer.
 static GOODBYE_SET: [&str; 3] = ["cx", "cz", "ecr"];
 
+fn get_euler_basis_set(basis_list: IndexSet<&str>) -> EulerBasisSet {
```
alexanderivrii, line 86: A small question about the type of `basis_list`: why does it need to be a `set` (and not say a `vec`) and why does it need to be `index` (the order is not important here, right?)

crates/accelerate/src/unitary_synthesis.rs
```diff
@@ -354,26 +379,33 @@ fn py_run_main_loop(
...
-                let synth_dag = circuit_to_dag(
-                    py,
-                    QuantumCircuitData::extract_bound(&synth_circ)?,
-                    false,
-                    None,
-                    None,
-                )?;
-                let out_qargs = dag.get_qargs(packed_instr.qubits);
-                apply_synth_dag(py, &mut out_dag, out_qargs, &synth_dag)?;
+                if basis_gates.is_empty() && matches!(target, None) {
```
alexanderivrii: Minor question: is there a benefit of writing `matched!(target, None)` over `target.is_none()`?

crates/accelerate/src/unitary_synthesis.rs
```diff
@@ -460,18 +516,13 @@ fn run_2q_unitary_synthesis(
...
-                                        .name()
-                                        .to_string(),
+                                    sequence.decomp_gate.name().to_string(),
                                     Some(params.iter().map(|p| Param::Float(*p)).collect()),
                                     inst_qubits,
                                 ),
                             }
                         });
-                let synth_error_from_target = synth_error(py, scoring_info, target);
+                let synth_error_from_target = synth_error(py, scoring_info, target.unwrap());
```
alexanderivrii: Why `target` can't be `None` here?

qiskit/transpiler/passes/synthesis/unitary_synthesis.py
```diff
@@ -350,8 +349,6 @@ def __init__(
                 directionality of the coupling_map will be taken into
                 account if ``pulse_optimize`` is ``True``/``None`` and ``natural_direction``
                 is ``True``/``None``.
-            backend_props (BackendProperties): Properties of a backend to
-                synthesize for (e.g. gate fidelities).
```
alexanderivrii: Ah, I see, the `BackendProperties` should be removed by the "prequel" PR #13706, correct?

alexanderivrii (commented):
I have taken a quick look at the PR, and it seems to be in a very good shape. I have left a few minor questions, and will take a deeper look once the PRs it's based on are merged. And of course you are missing release notes.

crates/accelerate/src/unitary_synthesis.rs
ElePT: I don't think so, this was me just being rusty in Rust when I wrote it :) I actually prefer `target.is_none()`

1ucian0:
Removing `on hold` as  #13706 merged.

crates/accelerate/src/unitary_synthesis.rs
ElePT: I went for `IndexSet` because:
- the `set` allows to compute the intersection (used in `get_2q_decomposer_from_basis`)
- the `index` ensures that the behavior of the pass is reproducible, when we "pop" an item from the set (done in `get_2q_decomposer_from_basis` and a few more places), the item is always the same, so that the behavior is reproducible. We had a couple of issues with unit tests because the previous behavior of UnitarySynthesis was unpredictable and we decided to make it all reproducible.

If you have any suggestions I am happy to look into them.

crates/accelerate/src/unitary_synthesis.rs
ElePT: Good question. If `target` is `None`, we would be in the `basis_gates` path, which proposes a single decomposer, so we would avoid computing the synthesis score (https://github.com/Qiskit/qiskit/pull/13704/files#diff-c37693939c4bc977d8d0d823e462802e87616250233241a010553702515f4da3R494). Maybe I can try to make the behavior more explicit...

ElePT:
Sorry @alexanderivrii if I have complicated the review with the last couple of commits. It might look like a lot of changes but I mostly moved around existing code (the one you already reviewed). As a summary:

- I restructured the Rust file to try to make it more understandable and reduce code repetition, I also added comments. 
- On the Python side, I moved the `DefaultUnitarySynthesis` plugin code to an independent file. This plugin duplicates the Rust logic, which unfortunately was developed in parallel to the plugin infrastructure. The Rust code combines the pass `run_main_loop` logic with the plugin `run` and currently cannot simply replace the plugin `run`. Ideally, we could yet again refactor the code to match the plugin infrastructure, but I don't think this will be possible before 2.0 (in terms of time).

For reference, the Python `DefaultUnitarySynthesis` class is still set as a fallback for the case where a custom plugin fails, which is the reason why I haven't removed it either. At least separating it from `UnitarySynthesis` might explain a bit better the situation.

crates/accelerate/src/unitary_synthesis.rs
```diff
@@ -62,53 +62,76 @@ enum DecomposerType {
...
+            }
+            let basis = EULER_BASIS_NAMES[idx];
+            Some(basis)
+        })
+        .for_each(|basis| euler_basis_set.add_basis(basis));
+
+    if euler_basis_set.basis_supported(EulerBasis::U3)
+        && euler_basis_set.basis_supported(EulerBasis::U321)
+    {
+        euler_basis_set.remove(EulerBasis::U3);
```
ShellyGarion, line 103: what happens if a user provides both 'U' and 'U3'?

ShellyGarion (commented):
@ElePT - thanks for refactoring the `unitary_synthesis.rs` code, as well as separating `unitary_synthesis.py`  and the `DefaultUnitarySynthesis` plugin. 
So it seems that the plugin is not being tested, and also behaves differently than the `UnitarySynthesis` pass?

crates/accelerate/src/unitary_synthesis.rs
ElePT: They are both included in the target `euler_basis_set` and given to the 1q decomposer to deal with.

ElePT:
Thanks @ShellyGarion, I guess that the strictly correct thing here would be to duplicate the synthesis unit tests for the plugin. I can do that.

crates/accelerate/src/unitary_synthesis.rs
ShellyGarion: if `euler_basis_set.basis_supported(EulerBasis::U3) == true` and `if euler_basis_set.basis_supported(EulerBasis::U) == true `
should we remove one of them?

alexanderivrii (dismissed):
Thanks @ElePT, I have carefully reviewed the code and it looks perfect to me, either because it truly is or because I lack the expertise to spot any issues, probably both are true :). 

One question (and I have a feeling that I may have already asked it at some point, but then forgot the answer) is why we need some code duplication between rust and python sides, i.e. why can't `DefaultSynthesisPlugin` just call the relevant functionality implemented in rust? I might be missing something even more basis here, doesn't UnitarySynthesis pass by default call the DefaultSynthesisPlugin?

And special thanks adding code comments, this helped me a lot to understand what each different function does. 

I am approving it, but I think there were also some review comments from @ShellyGarion.

crates/accelerate/src/unitary_synthesis.rs
ElePT: This part of the code literally follows the logic from the 1q euler decomposer... I am not sure if there was an initial motivation not to remove one of the two (U/U3) or if it was an oversight. I can't think of a reason from the top of my head.

ElePT:
Thanks @alexanderivrii! To answer your question, the code shouldn't be duplicated. The reason why it is is that at the time the unitary synthesis code was originally oxidized, we didn't have any reference of how to interact with the plugin infrastructure, and because we only had 1 unitary synthesis plugin (the infrastructure wasn't really taken advantage of), we went for the path of replicating the whole flow in rust. In hindsight it would be way more modular to have only the plugin `run` method in rust, but implementing this now requires a refactoring that I didn't have the time to prioritize yet. Once you finish the HLS oxidization it might also be a good idea to try to make unitary synthesis is pluggable in the HLS workflow. This is something I briefly discussed with @Cryoris. The TL:DR is: not enough information at the time, not enough time to fix it when the information was clear.

ElePT:
Adding the plugin unit tests will probably take a bit more effort than originally planned. If you are fine with merging the PR as-is, I can address the test situation in a follow-up (it's anyways a non user-facing change).

ShellyGarion (approved):
LGTM. Thanks!

PR #13705, opened 2025-01-21
Remove need for mutable reference to static

jakelishman:
Mutable references to static data are inherently unsafe and typically unsound in Rust, because statics are implicitly shared between threads, and the borrow checker can only enforce the shared/exclusive reference limitations within a single thread here.

This just moves the thread-exclusion logic into the individual elements of the `static`, where the `GILOnceCell` can correctly handle the runtime exclusion of multiple threads. (The GIL is no longer suitable for thread exclusion if we were doing a freethreaded Python build, but that's a problem we have all over Qiskit, and would need to change to `OnceLock` or the like.)

### Details and comments

Minor conflict with the update to PyO3 0.23, but only a small syntactic one - the `import_bound` becomes `import`.  The logic still works.

### DIFF
--- a/crates/circuit/src/imports.rs
+++ b/crates/circuit/src/imports.rs
@@ -255,42 +255,65 @@ static STDGATE_IMPORT_PATHS: [[&str; 2]; STANDARD_GATE_SIZE] = [
 ///
 /// NOTE: the order here is significant it must match the StandardGate variant's number must match
 /// index of it's entry in this table. This is all done statically for performance
-static mut STDGATE_PYTHON_GATES: GILOnceCell<[Option<PyObject>; STANDARD_GATE_SIZE]> =
-    GILOnceCell::new();
+static STDGATE_PYTHON_GATES: [GILOnceCell<PyObject>; STANDARD_GATE_SIZE] = [
+    GILOnceCell::new(),
+    GILOnceCell::new(),
+    GILOnceCell::new(),
+    GILOnceCell::new(),
+    GILOnceCell::new(),
+    GILOnceCell::new(),
+    GILOnceCell::new(),
+    GILOnceCell::new(),
+    GILOnceCell::new(),
+    GILOnceCell::new(),
+    GILOnceCell::new(),
+    GILOnceCell::new(),
+    GILOnceCell::new(),
+    GILOnceCell::new(),
+    GILOnceCell::new(),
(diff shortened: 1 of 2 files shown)


crates/circuit/src/imports.rs
```diff
@@ -255,42 +255,65 @@ static STDGATE_IMPORT_PATHS: [[&str; 2]; STANDARD_GATE_SIZE] = [
...
+    GILOnceCell::new(),
+    GILOnceCell::new(),
+    GILOnceCell::new(),
+    GILOnceCell::new(),
+    GILOnceCell::new(),
+    GILOnceCell::new(),
+    GILOnceCell::new(),
+    GILOnceCell::new(),
+    GILOnceCell::new(),
+];
```
mtreinish, line 311: This is pretty ugly. But I don't think there is a syntax that we can use here with GilOnceCell given the constraints defined in: https://doc.rust-lang.org/std/primitive.array.html so this is fine.

mtreinish (approved):
This LGTM. I think I found the previous code cleaner but rust 1.84 warns 3 times about this being unsafe, and rust edition 2024 will actually make the mut ref static stuff unsafe. While we won't be using that for 2.x, it's good to remove the mut ref static preemptively.

crates/circuit/src/imports.rs
jakelishman: Yeah, it's ugly.  For anybody from the future looking: the trouble is that there's not enough available in `const` functions to manage this.  The closest I managed to get was
```rust
use std::mem::MaybeUninit;
use pyo3::prelude::*;
use pyo3::sync::GILOnceCell;

const STANDARD_GATE_SIZE: usize = 52;

static STDGATE_PYTHON_GATES: [GILOnceCell<Py<PyAny>>; STANDARD_GATE_SIZE] = const {
    let mut out: [MaybeUninit<GILOnceCell<Py<PyAny>>>; STANDARD_GATE_SIZE] =
        [const { MaybeUninit::uninit() }; STANDARD_GATE_SIZE];
    let mut i = 0;
    while i < STANDARD_GATE_SIZE {
        out[i].write(GILOnceCell::new());  // error: `MaybeUninit::write` isn't const stable
        i += 1;
    }
    unsafe { ::std::mem::transmute(out) }
};
```

PR #13706, opened 2025-01-21
Remove use of `BackendProperties` (BackendV1) in `UnitarySynthesis` pass

ElePT:
### Summary

This PR includes:

- **Test upgrade** -> Upgrade of class unit tests to avoid use of Backend V1. Some tests were also modified to use `target` instead of `basis_gates`, as the functionality they were testing was dependant of the `backend_props` input (deprecated)
- **Removal** -> Removed the `backend_props` argument from `UnitarySynthesis`, as it relied on the deprecated `BackendProperties` class. The argument itself was never deprecated, so it might be necessary to open a deprecation PR against 1.4. 

Note that this PR touches on code that is currently actively worked on in:

- #12850 
- #13568 
- #13704 

And there will be conflicts.

### Details and comments

- Currently blocking https://github.com/Qiskit/qiskit/pull/13722 and # https://github.com/Qiskit/qiskit/pull/13704
- Contributes to #13708


ElePT:
Marked as on hold to make sure I have a 1.4 deprecation ready before merging this PR.

ElePT:
No longer on hold after https://github.com/Qiskit/qiskit/pull/13719

releasenotes/notes/remove-backend-props-unitary-synth-0dca60b56ccd9ee3.yaml
```diff
@@ -0,0 +1,11 @@
+---
+upgrade_transpiler:
+  - |
+    The ``backend_properties`` input argument has been removed from the following objects:
+    
+    * :class:`.UnitarySynthesis` transpiler pass
+    * :func:`.generate_translation_passmanager` function
+
+    This argument required a ``BackendProperties`` object, deprecated since qiskit 1.2, as it
+    belonged to the deprecated BackendV1 workflow. Constraints such as gate error and durations
```
Cryoris: Do you want to mention that `BackendProperties` is removed in this release?
```suggestion
    This argument required a ``BackendProperties`` object, deprecated since Qiskit 1.2, as it
    belonged to the deprecated ``BackendV1`` workflow. Constraints such as gate error and durations 
```

test/python/transpiler/test_unitary_synthesis.py
```diff
@@ -952,6 +849,88 @@ def test_determinism(self):
             for basis_gate in basis_gates:
                 self.assertLessEqual(out.count_ops()[basis_gate], gate_counts[basis_gate])
 
+    @combine(gate=["unitary", "swap"], natural_direction=[True, False])
+    def test_two_qubit_synthesis_to_directional_cx_target(self, gate, natural_direction):
```
Cryoris, line 853: Could we merge these tests with the target-free versions above by adding e.g. a third argument `via_target=True/False` that then builds a Target or not inside? Then we could reduce the duplication 🙂

Cryoris (commented):
Overall LGTM, I just have two small, non-logic-touching comments 🙂

test/python/transpiler/test_unitary_synthesis.py
ElePT: We could, but not all the tests have an equivalent on the other class, so the unification would take a bit of work. I think that would be out of scope for this PR.

releasenotes/notes/remove-backend-props-unitary-synth-0dca60b56ccd9ee3.yaml
ElePT: I wrote the reno before #13722, which completes the removal... it is a bit weird to have UnitarySynthesis singled out, so I am thinking... should we remove the reno alltogether?

ElePT: I decided to remove the reno, as the changes are already reflected in the reno of https://github.com/Qiskit/qiskit/pull/13722. I will change the label to Changelog:None to make sure there are no issues with the release bot.

ShellyGarion:
LGTM (except of the failing tests that should be updated). 
I found the word "backend" still mentioned in some docs - should it be "target"?
https://github.com/Qiskit/qiskit/blob/00bee06d4fdab1d18dfc832ef0a6abac1f67df46/qiskit/transpiler/passes/synthesis/unitary_synthesis.py#L131
https://github.com/Qiskit/qiskit/blob/00bee06d4fdab1d18dfc832ef0a6abac1f67df46/qiskit/transpiler/passes/synthesis/unitary_synthesis.py#L346

ElePT:
Yeah I don't know why the tests started failing on CI, they all pass locally and on previous CI runs, the lines that are reported to fail don't exist so I think it must be a small glitch. I re-pushed things to trigger the tests again.

ElePT:
@jakelishman helped me understand the mysterious source of the errors: the CI jobs don’t actually run on the PR `HEAD`, they run on the merge of the PR `HEAD` and `main` (unless there's a conflict), and this was causing the strange untraceable failures. Should be fixed now.

releasenotes/notes/remove-backend-props-unitary-synth-0dca60b56ccd9ee3.yaml
Cryoris: Agreed, it should all be in a single reno to understand what's going on :+1:

Cryoris (approved):

qiskit/transpiler/preset_passmanagers/common.py
```diff
@@ -429,7 +429,6 @@ def generate_translation_passmanager(
     method="translator",
     approximation_degree=None,
     coupling_map=None,
-    backend_props=None,
```
jakelishman, line 432: This was a breaking change to a public API - it breaks `qiskit-ibm-runtime`'s translation plugins.  The option wasn't deprecated previously, so we ought to at a minimum issue a deprecation warning.  Since it's necessary to pass for the entirety of Qiskit 1.x, I wouldn't be opposed to us leaving it as a deprecated no-op argument for the duration of Qiskit 2.x.

jakelishman: Oh, I see now that `qiskit-ibm-runtime` already linked a PR above.

PR #13709, opened 2025-01-21
Bump MSRV to 1.79

mtreinish:
### Summary

This commit bumps the minimum supported rust version (MSRV) from 1.70 to 1.79. The newer rust version lets us leverage new language features in Qiskit's rust code which will be increasingly important as we use more rust inside of Qiskit. This commit raises the MSRV proactively to align with the 2.0.0 release so we don't have to potentially raise it during a minor version release (like 2.1.0 or 2.2.0) if we start using these language features later.

### Details and comments

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -37,9 +37,9 @@ dependencies = [
 
 [[package]]
 name = "allocator-api2"
-version = "0.2.18"
+version = "0.2.21"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "5c6cb57a04249c6480766f7f7cef5467412af1490f8d1e243141daddada3264f"
+checksum = "683d7910e743518b0e34f1186f92494becacb047c7b6bf616c96772180fef923"
 
 [[package]]
 name = "approx"
@@ -85,9 +85,9 @@ checksum = "bef38d45163c2f1dde094a7dfd33ccf595c92905c8f8f4fdc18d06fb1037718a"
 
 [[package]]
 name = "bitflags"
-version = "2.6.0"
+version = "2.8.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "b048fb63fd8b5923fc5aa7b340d8e156aec7ec02f0c78fa8a6ddc2613f6f71de"
+checksum = "8f68f53c83ab957f72c32642f3868eec03eb974d1fb82e453128456482613d36"
 
 [[package]]
 name = "block-buffer"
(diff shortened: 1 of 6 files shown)


Cargo.lock
```diff
@@ -261,6 +261,15 @@ dependencies = [
...
 
+[[package]]
+name = "dyn-stack"
+version = "0.13.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "490bd48eb68fffcfed519b4edbfd82c69cbe741d175b84f0e0cbe8c57cbe0bdd"
+dependencies = [
+ "bytemuck",
+]
+
```
raynelfss, line 272: This seems to be used to dynamically allocate stack space. Did this package get added due to some sub-dependency introduced when upgrading the MSRV?

Cargo.lock
```diff
@@ -1056,6 +1066,20 @@ dependencies = [
...
+checksum = "95fb7a99b37aaef4c7dd2fd15a819eb8010bfc7a2c2155230d51f497316cad6d"
+dependencies = [
+ "bytemuck",
+ "cfg-if",
+ "libm",
+ "num-complex",
+ "reborrow",
+ "version_check",
+]
+
```
raynelfss, line 1082: This also seems to be more related to SIMD and multithreading if I read correctly, do we know where it's being used?

Cargo.lock
```diff
@@ -1550,22 +1574,42 @@ checksum = "f18aa187839b2bdb1ad2fa35ead8c4c2976b64e4363c386d45ac0f7ee85c9233"
...
+version = "1.0.69"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "b6aaf5339b578ea85b50e080feb250a3e8ae8cfcdff9a461c9ec2904bc923f52"
+dependencies = [
+ "thiserror-impl 1.0.69",
+]
+
+[[package]]
+name = "thiserror"
+version = "2.0.11"
```
raynelfss, line 1586: Is there anything preventing us from moving to "2.0.11" entirely here? Having two different versions of a package can be a bit tricky at times, though I think we've done something like this before.

raynelfss (approved):
LGTM, just a couple of inline questions about packages that got added.

Cargo.lock
mtreinish: As part of bumping the msrv I ran `cargo update` which updates all of the dependencies in the lock file with the latest compatible version. This got pulled in because it's a dependency on the new version of `gemm` (which is a dependency of faer-rs) that got updated by this.

Cargo.lock
mtreinish: Pulp is a simd acceleration library we use it a little bit in some of the numeric code in rust. But this got bumped because of the new version of `gemm` that was pulled in by running `cargo update`.

Cargo.lock
mtreinish: We can bump our `Cargo.toml` dependency to 2.0 assuming our usage in Qiskit is compatible (the breaking changes in thiserror 2.0 seem fairly small). But I'd suggest doing that in a dedicated PR as this is mostly related to the MSRV. The addition here is cargo's default behavior and it's not anything to be concerned about. If a dependency has a specific differing version requirement on a crate it will use that different version when building the dependency and it doesn't interact with the version we use in Qiskit itself (the only exception to that is if we share objects between crates, for example passing a hashbrown hashmap to rustworkx needs to use the same version in Qiskit and rustworkx). This is perfectly normal and the only real downside is longer build times because we have to build multiple versions of a crate.

That all being said this is also coming from a bump in `pest` which is dependency of `npyz` which faer-rs uses. This just got pulled in because of the cargo update.

raynelfss: Sounds good then, I believe if it's a dependency not being up to date, once it does update, cargo would take care of that.

kevinhartman (approved):
Awesome! Thanks for doing this.

Issue #13665, opened 2025-01-14
Migrate linear algebra code from faer-rs to nalgebra


mtreinish:
### What should we add?

In qiskit 1.0 we started using the [faer-rs](https://github.com/sarah-quinones/faer-rs) library for linear algebra from rust. This enabled us to migrate a lot of  python/numpy numeric code into Rust during the development of 1.x which greatly accelerated a lot of transpiler code in the 1.x cycle. But with the recent releases of the faer it's become clear the project isn't really in a place where we can rely on it as a stable upstream dependency for Qiskit. The development of the library is very rapid and it's hard to keep track of what is going on, and the MSRV changes are incompatible with our requirements for Qiskit.

For our use cases most of the linear algebra we do is with small matrices, especially in the transpiler, typically with 2x2 and 4x4 matrices, and nalgebra is a good fit for that because we can used fixed size stack allocated types for the arrays. This will also provide better performance in most cases for what we're currently using.  It's also a more stable/established library that is better aligned with our requirements for dependencies in Qiskit.

tonmoy-b:
Hi, I would love to contribute to this if there's still any space for a contributor for these tasks. Kindly let me know if I can join. Thanks, Tonmoy

eliarbel:
Current status is that we have 2 places that still use faer:
`uc_gate.rs` - this is being addressed in #13956 
2q decompose - @gadial to investigate

mtreinish:
To update here we decided to keep using faer moving forward we're going to need it for the quantum shannon decomposition and also the linalg decomposition we're using in places like the two qubit decomposer we will still benefit from nalgebra's fixed sized matrix types but we'll likely still use faer's linalg with those. I'm going to keep this issue open until https://github.com/Qiskit/qiskit/pull/15017 merges because that is updating the two qubit decomposer to use nalgebra's fixed sized matrices instead of ndarray.

ShellyGarion:
is this issue still relevant after you've handled the two-qubit decomposition code?

mtreinish:
I've marked #15960 as closing this. I think there is probably more optimization we'll want to do once we're using nalgebra for the matrix type in the two qubit decomposers (like updating the `TwoQubitControlledDecomposer` to use nalgebra for `Matrix2` and matmul after #16036 is fixed). But the bulk of what this issue was tracking is done with #15960.

PR #13713, opened 2025-01-21
Remove unused qiskit_accelerate::utils module

mtreinish:
### Summary

The utils module was added to have a place for some common utilities and expose them to other rust modules or also to python. This was originally added to expose faer's eigensolver as an alternative to numpy for Python, but nothing ever used it and we didn't properly expose the python submodule so it wasn't easily usable via qiskit._accelerate. Since it's never used this opts to just remove it, if we have a future need we can just add the function back. With the eigensolver removed the only function that remained in the module was the arg_sort() function which was only used by the two qubit decomposer. To simplify the code base this is just made into a private function for that module and we can remove the utils module altogether.

### Details and comments

Related to #13665


ElePT (approved):
LGTM

PR #13714, opened 2025-01-21
Remove faer usage from quantum_volume

mtreinish:
### Summary

As part of #13665 this migrates the faer-rs usage in quantum_volume.rs to just use nalgebra instead. The only faer usage there was for the qr decomposition and everything else was done with fixed sized arrays or Array2. This migration is fairly simple but it's a good start towards the goal of eventually removing our faer-rs usage in Qiskit. It might improve performance because we remove one layer of heap allocated objects by moving from Array2 to Matrix4 for rust operations. Although we do need an equivalent heap allocation for generating the numpy arrays used in `UnitaryGate` so it might not really change anything significantly.

### Details and comments

Part of #13665


ElePT (approved):
LGTM! Not an expert by any means on linear algebra in rust but this change seems straightforward enough.

mtreinish:
This is largely duplicated in #13765 but I used nalgebra slightly differently in that PR (I forgot I had this PR open when I wrote it). I'm fine merging this as is, and just rebasing #13765 on top of this. Or we can just close this and move proceed solely with #13765. I'll defer to what you think is best @ElePT.

mtreinish:
This has been superseded by: https://github.com/Qiskit/qiskit/pull/13765

PR #13717, opened 2025-01-22
Configure workspace-level Rust lints

jakelishman:
This configures each of the workspace packges to inherit the `lints` table from the top-level `Cargo.toml`.  This feature is only available from Rust 1.74+, which is now within our MSRV.

This commit is mostly setting up the infrastructure; we can discuss more lints as a team if/when they come up.  The two configured in this commit are an "allow" that is already global over `accelerate`, and a "deny" that will become a warning with Rust 2024 (it's allow-by-default in Rust 2021).


kevinhartman (approved):
Nice, LGTM.

PR #13719, opened 2025-01-22
Deprecate use of `BackendProperties` (BackendV1) in transpilation pipeline

ElePT:
### Summary
This PR deprecates the following uses of `BackendProperties` to be removed in 2.0:

- `backend_prop` input argument in `DenseLayout`
- `properties` input argument in `VF2Layout`
- `properties` input argument in `VF2PostLayout`
- `backend_props` input argument in `UnitarySynthesis`
- `backend_properties` in `PassManagerConfig`
- `backend_properties` in `generate_routing_passmanager`
- `backend_properties` in `generate_translation_passmanager`
- `backend_properties` in  `Target.from_configuration`

### Details and comments
This PR unblocks #13706 and further 2.0 removals.

### DIFF
--- a/qiskit/transpiler/passes/layout/dense_layout.py
+++ b/qiskit/transpiler/passes/layout/dense_layout.py
@@ -21,6 +21,8 @@ from qiskit.transpiler.basepasses import AnalysisPass
 from qiskit.transpiler.exceptions import TranspilerError
 from qiskit.transpiler.passes.layout import disjoint_utils
 
+from qiskit.utils import deprecate_arg
+
 from qiskit._accelerate.dense_layout import best_subset
 
 
@@ -36,6 +38,13 @@ class DenseLayout(AnalysisPass):
         by being set in ``property_set``.
     """
 
+    @deprecate_arg(
+        name="backend_prop",
+        since="1.4",
+        package_name="Qiskit",
+        removal_timeline="in Qiskit 2.0",
+        additional_msg="The `target` parameter should be used instead.",
+    )
     def __init__(self, coupling_map=None, backend_prop=None, target=None):
         """DenseLayout initializer.
 
--- a/qiskit/transpiler/passes/layout/dense_layout.py
+++ b/qiskit/transpiler/passes/layout/dense_layout.py
@@ -43,7 +43,10 @@ class DenseLayout(AnalysisPass):
         since="1.4",
         package_name="Qiskit",
         removal_timeline="in Qiskit 2.0",
-        additional_msg="The `target` parameter should be used instead.",
+        additional_msg="The BackendProperties data structure has been deprecated and will be "
+        "removed in Qiskit 2.0. The `target` input argument should be used instead. "
+        "You can use Target.from_configuration() to build the target from the properties "
+        "object, but in 2.0 you will need to generate a target directly.",
     )
     def __init__(self, coupling_map=None, backend_prop=None, target=None):
         """DenseLayout initializer.
--- a/qiskit/transpiler/passes/layout/vf2_layout.py
+++ b/qiskit/transpiler/passes/layout/vf2_layout.py
@@ -79,7 +79,10 @@ class VF2Layout(AnalysisPass):
         since="1.4",
         package_name="Qiskit",
         removal_timeline="in Qiskit 2.0",
-        additional_msg="The `target` parameter should be used instead.",
+        additional_msg="The BackendProperties data structure has been deprecated and will be "
+        "removed in Qiskit 2.0. The `target` input argument should be used instead. "
+        "You can use Target.from_configuration() to build the target from the properties "
+        "object, but in 2.0 you will need to generate a target directly.",
     )
     def __init__(
         self,
--- a/qiskit/transpiler/passes/layout/vf2_layout.py
+++ b/qiskit/transpiler/passes/layout/vf2_layout.py
@@ -24,6 +24,7 @@ from qiskit.transpiler.basepasses import AnalysisPass
 from qiskit.transpiler.exceptions import TranspilerError
 from qiskit.transpiler.passes.layout import vf2_utils
 
+from qiskit.utils import deprecate_arg
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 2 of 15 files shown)


qiskit/transpiler/preset_passmanagers/common.py
jakelishman, line 1: There isn't a reasonable migration path for plugin authors in the 1.x series here - you're given a `PassManagerConfig`, which has this field, and it would be a mistake not to pass it along.  `qiskit-ibm-runtime` does this (as should other plugin authors).  I don't think these are eligible for deprecation until 2.0, which is when they become no-ops.  That's indicated by all the `filter_warnings` stuff in the built-in plugins too.

ElePT: Following the discussion in https://github.com/Qiskit/qiskit/issues/13831, I singled out these cases in the reno to try to clarify that the `target` was already in place as alternative path for the standalone properties argument. See [a4cc3c4](https://github.com/Qiskit/qiskit/pull/13719/commits/a4cc3c499d88933e59bcb92b74762e5e6fdb7eb6). I am happy to add tweaks if needed.

qiskit/transpiler/passes/layout/dense_layout.py
jakelishman, line 1: In `DenseLayout`, there's a path where a disagreement between the `Target` and `BackendProperties` can cause the `backend_props` argument to be used in place of the `Target` (if `Target.qargs is None`).  That might be a bug - perhaps in `_build_error_matrix` we should have
```python
if target is not None:
    if target.qargs is None:
        return zeros
    # ... build errors
elif backend_prop and coupling_map:
    # ... build errors
return errors
```
instead of what we have right now, which is the `target.qargs is not None` being in the top-line if.

If a bug, it can be a follow-up.

qiskit/transpiler/passes/layout/vf2_layout.py
jakelishman, line 1: We have the same `if target.is not None and target.qargs is not None` maybe-problem in `vf2_utils.build_average_error_map`.  Also affects `VF2PostLayout`.

qiskit/transpiler/passes/synthesis/unitary_synthesis.py
jakelishman, line 1: In `_build_gate_errors` in this file, we use _both_ `Target` and `BackendProperties` if they're both given. The other similar `_build_*` functions use only `Target` if given. Don't know if it's a bug or intentional.

That's probably not a huge deal for the deprecation - the alternate path would just be to update the `Target` if it was intended behaviour anyway.

qiskit/transpiler/preset_passmanagers/builtin_plugins.py
jakelishman, line 1: I'm personally not keen on using `filter_warnings` in library code, but probably the ship's sailed on that, and these at least aren't inner-loop locations.

releasenotes/notes/deprecate-backend-v1-props-09481526448e6bfd.yaml
```diff
@@ -0,0 +1,25 @@
...
+deprecations_transpiler:
+  - |
+    The following uses of the ``BackendProperties`` object in the transpilation
+    pipeline have been deprecated as of Qiskit 1.4 and will be removed in Qiskit 2.0:
+
+    * ``backend_prop`` input argument in :class:`.DenseLayout`
+    * ``properties`` input argument in :class:`.VF2Layout`
+    * ``properties`` input argument in :class:`.VF2PostLayout`
+    * ``backend_props`` input argument in :class:`.UnitarySynthesis`
+    * ``backend_properties`` in :class:`.PassManagerConfig`
```
jakelishman:
```suggestion
    * ``backend_props`` input argument in :class:`.UnitarySynthesis`
    * ``backend_properties`` input argument in :class:`.PassManagerConfig`
```

just for consistency - all the other `__init__`s say this.

releasenotes/notes/deprecate-backend-v1-props-09481526448e6bfd.yaml
```diff
@@ -0,0 +1,25 @@
...
+    * ``backend_properties`` in  :meth:`.Target.from_configuration`
+    * ``backend_properties`` in :func:`.generate_routing_passmanager`
+    * ``backend_properties`` in :func:`.generate_translation_passmanager`
+
+    The :class:`.BackendProperties` class has been deprecated since Qiskit 1.2, as it was part
+    of the :class:`.BackendV1` workflow, and will be removed in Qiskit 2.0. Specific instruction 
+    properties such as gate errors or durations can be added to a :class:`.Target` upon 
+    construction through the :meth:`.Target.add_instruction` method, and communicated to 
+    the relevant transpiler passes through the `target` input argument.
```
jakelishman:
```suggestion
    the relevant transpiler passes through the ``target`` input argument.
```

jakelishman (dismissed):
Overall, this looks about as clean as we'll manage, thanks Elena.  I think none of my comments are actionable for this PR.

I wish that we'd handled this move a little cleaner overall - if the `Target` was supposed to supersede `BackendProperties`, I wish we'd had `generate_preset_pass_manager` and `transpile` set that option to `None` the second a `Target` was passed (or at least overwrite them with the data re-derived from the `Target` so we knew it was identical), which we could have introduced at any point, especially once all the transpiler passes were migrated, but oh well, too late for that now.

As it stands, we're relying on each individual pass to make sure it prioritises the `Target`, and of the ones touched by this PR, either there's (minor) bugs in their implementations of that, or we're interpreting "priority" different in some places.

qiskit/transpiler/passes/layout/dense_layout.py
ElePT: Hmmm this does look like a bug, but the bug would only affect 1.4 because the `elif backend_props` branch is anyways going away in 2.0. Given that 1.4 is supposed to not include fixes, and we are probably too late for 1.3.3, should we just... let the "legacy" behavior be even if it's inconsistent with the prioritization of the target?

qiskit/transpiler/passes/synthesis/unitary_synthesis.py
ElePT: This is surely a bug, but the rust path currently skips these gate errors so it will not affect the code when the plugin is set to `"default"` (yes, a cleanup of this python code is pending).

jakelishman (dismissed):

jakelishman (approved):

Issue #13708, opened 2025-01-21
Remove BackendV1 and related dependencies, deprecated in 1.2


ElePT:
This issue tracks the BackendV1 removal efforts, this includes backendV1-dependant provider models, fake backends, etc.

Related PRs:

- #13706 
- #13722

The removal is blocked by the corresponding deprecation PR in 1.4:

- #13719

PR #13722, opened 2025-01-22
Remove use of `BackendProperties` (BackendV1) in transpiler pipeline

ElePT:
### Summary
This PR removes the following uses of `BackendProperties` deprecated in #13719:

- backend_prop input argument in DenseLayout
- properties input argument in VF2Layout
- properties input argument in VF2PostLayout
- backend_props input argument in UnitarySynthesis
- backend_properties in PassManagerConfig
- backend_properties in generate_routing_passmanager
- backend_properties in generate_translation_passmanager
- backend_properties in Target.from_configuration 

as well as the previously deprecated input arguments of `transpile`, `generate_preset_pass_manager`.

### Details and comments
Currently on hold as it contains the changes from https://github.com/Qiskit/qiskit/pull/13706, should be rebased after it's merged.

Contributes to #13708.


test/python/transpiler/test_vf2_post_layout.py
```diff
@@ -587,30 +446,6 @@ def test_no_constraints(self):
...
             empty_pass.run(circuit_to_dag(qc))
 
-    def test_no_backend_properties(self):
-        """Test we raise at runtime if no properties are provided with a coupling graph."""
-        qc = QuantumCircuit(2)
-        empty_pass = VF2PostLayout(
-            coupling_map=CouplingMap([(0, 1), (1, 2)]), strict_direction=False
-        )
-        with self.assertRaises(TranspilerError):
-            empty_pass.run(circuit_to_dag(qc))
```
mtreinish, line 597: Isn't this test still valid without backend properties? Shouldn't the pass fail if we provide a coupling map with no error rates to use?

qiskit/transpiler/preset_passmanagers/generate_preset_pass_manager.py
```diff
@@ -27,7 +27,7 @@
 from qiskit.transpiler.instruction_durations import InstructionDurations
 from qiskit.transpiler.layout import Layout
 from qiskit.transpiler.passmanager_config import PassManagerConfig
-from qiskit.transpiler.target import Target, target_to_backend_properties
+from qiskit.transpiler.target import Target
```
mtreinish, line 30: Are we planning to remove `target_to_backend_properties` in the PR removing the class?

mtreinish (commented):
This LGTM, one inline question about a removed test but outside of that I think this is good to merge.

test/python/transpiler/test_vf2_post_layout.py
ElePT: I thought about removing `coupling_map` from the `VF2PostLayout` pass because of this test, because it really cannot be used without backend properties. Then I went for not erroring when only a coupling map is provided, I think I checked that it simply didn't return a solution. I am now standing somewhere between the two, the only advantage of still accepting the argument I guess is some kind of compatibility angle?

mtreinish: Yeah it'd mostly be a compatibility question. From a practical PoV `VF2Layout` works with just a coupling map and for the error heuristic it uses node degree in the connectivity graph as a proxy for error (so a more connected node is avoided). For `VF2PostLayout` that is unlikely to have an impact because most coupling maps are regular lattices so the degrees won't differ that much. 

The one case I could see it making a difference is like if the interaction graph of the post routing circuit was a line you could maybe snake the line over boundary nodes in the lattice that have a smaller degree. Something like:

im graph:
```
0 - 1 - 2 - 3 -4
```

cmap:
````
0 - 1 - 2 - 3
|   |    |  |
4 - 5 - 6 - 7
|   |    |  |
8 - 9 - 10 -11
````

if you used the node degree the mapping would favor putting the line that maximized the number of nodes on 0, 3, 8, and 11 in the cmap. Whether this is good behavior for VF2PostLayout is debatable though.

Comment (ElePT) [reply test/python/transpiler/test_vf2_post_layout.py:597]: Hmm ok, even though that use case doesn't come up much in practice now, I think the most "behavior preserving" path regarding this removal is erroring on the coupling map.

**Comment by ElePT** [comment]
After a bit of back and forth between the different removal alternatives in `VF2PostLayout`, I settled for removing the `coupling_map` input argument. Unlike `VF2Layout`, which might use `coupling_map` in the presence of a `target` if the target is missing the connectivity information, `VF2PostLayout` had no use for it as a standalone input. The version change seems like a good opportunity to get rid of both inputs as a "pack". I updated the deprecation in #13719 to include the `coupling_map` and removed it in [fe772b9](https://github.com/Qiskit/qiskit/pull/13722/commits/fe772b970926b8119bb5119ee58917f9c8dbfde5).

File: qiskit/transpiler/passes/layout/vf2_post_layout.py [hunk 3]
```diff
@@ -153,16 +142,12 @@ def __init__(
 
     def run(self, dag):
         """run the layout method"""
-        if self.target is None and (self.coupling_map is None or self.properties is None):
-            raise TranspilerError(
-                "A target must be specified or a coupling map and properties must be provided"
-            )
+        if self.target is None:
+            raise TranspilerError("A target must be specified or a coupling map must be provided")
```
mtreinish, line 146: This threw me for a second until I remember these passes should be instantiable even if the required target isn't specified.

mtreinish (approved):
This LGTM now thanks for doing all of the work on this.

qiskit/transpiler/passes/layout/vf2_post_layout.py
ElePT: Yes I also spent a long time thinking if I should make it a required argument.

PR #13733, opened 2025-01-24
Bump rustworkx-core and hashbrown to latest releases

mtreinish:
### Summary

This commit bumps the rustworkx-core and hashbrown dependencies to the latest releases. We were previously blocked on upgrading hashbrown because we use it to exchange data with rustworkx-core and rustworkx-core 0.15.x depended on the previous release of hashbrown. But with the new release of rustworkx-core 0.16.0 it is using the latest release of hashbrown and we can bump both at the same time.

One thing we'll want to do is upgrade our ndarray version in the future because rustworkx-core uses 0.16 now, but we're blocked on that by our einsum library at the momemnt. Luckily we don't exchange arrays with rustworkx at the moment in rust. But this will become an issue if we do.

### Details and comments


jakelishman (dismissed):
This pulls in a second version of `ndarray`, which is a nuisance because Qiskit can't upgrade to 0.16.1 because the version of faer we use has a pin on 0.15.  Fine to merge for now, but that's an annoyance for sure.

jakelishman:
Oh, maybe it's not fine given the test failure haha.

raynelfss:
After taking a closer look, I think the error is happening in a couple of lines where we are trying to obtain unique and ordered `qargs` during the `apply_translation` and `extract_basis` stages of the `BasisTranslator` using a `HashSet`. What's probably happening is that, when iteraring, the order is different because of the Hasher, which impacts the results for these functions. We might want to consider using `BTreeSet` instead of `HashSet` for these.

For example, in line 434 of `basis/basis_translator/mod.rs` we generate a unique `qubit_set` which should also be ordered.
```rs
// Assume `node_qarg` is [0,1,2]
let qubit_set: HashSet<Qubit> = HashSet::from_iter(node_qarg.iter().copied());
```
Further down, in line 512, we'd generate qargs based on the qubit set above:
```rs
let node_qarg_as_physical: Option<Qargs> =
            Some(node_qarg.iter().map(|x| PhysicalQubit(x.0)).collect());
/// Could randomly result in [1,2,0], [0,2,1], [2,1,0], etc...
```

Which we later use for a lookup which could fail if the order is flipped.

If we switched this to a `BTreeSet` we wouldn't have this issue since the `BTreeSet` preserves the order of the tree when iterating.

What worries me about this is that this failure had an exact 50% chance of being triggered if a certain `qargs` would've gotten flipped during iteration (from `(2,3)` to `(3,2)`). It could've gone undetected given the right circumstances 😆

jakelishman:
Ray: do we need numerical order in the qargs, insertion order, or just any deterministic order? `BTreeSet` is for numerical order, `IndexSet` is for insertion order. There's typically a bit more overhead with `BTreeSet` because of the sorting.

raynelfss:
@jakelishman: for this I believe it is numerical order. We only want to preserve the combination of values rather than their permutation, and then build a `SmallVec<_>` with it. The problem with insertion order is that, for example `(2,3)` and `(3,2)`, we need to just account for the combination of the values, so we'd still end up with two choices instead of one. And since the Python implementation for it used `set()` which inherently sorts the values, `BTreeSet` seems like the logical choice.

jakelishman:
It's ok if there's two choices, as long as we'll entirely reliably choose the same one. If we'll always encounter the qubits in the same order while adding them to the set, then `IndexSet` would still maintain determinism.

Python's `set` doesn't sort, and it doesn't maintain insertion order - it's got different properties to `dict`. For small ints, it might appear to sort because of how Python hashes ints (take the low bits mod a large mersenne prime), and because there's no hash-seed randomisation for ints.  If you take ints that will have some hash collisions (say 0, 64, 256), though, you'll see that the order isn't sorted, it's just some arbitrary deterministic one that's a function only of the ints and the order they're inserted. (For most other Python objects it's a different story - hash-seed randomisation means their hashes change non-deterministically in new processes.)

raynelfss:
> If we'll always encounter the qubits in the same order while adding them to the set, then IndexSet would still maintain determinism.

That's the problem here, we don't always find them in the same order, therefore using an `IndexSet` will produce a different outcome for `(3,2)` and `(2,3)`, whereas `BTreeSet` would always produce `(2,3)` which is what we want whenever we end up casting to a `SmallVec<_>`. The focus was to produce the unique combination of qargs, not a unique permutation.

While it is true that Python sets are unorderded, when casting them to lists they produce an ordered list, at least in the case of numbers, due to the nature of hashing in python.

We may instead use Sets by default in these cases as keys during `apply_translation` and `extract_basis` stages of the `BasisTranslator` for which case, sticking to `HashSet<_>` should be fine, but since it is a bigger change out of the scope of this PR, switching to `BTreeSet` for now should be okay.

jakelishman:
We only need to hit them in the same order for a given set of inputs to `BasisTranslator`, and in a way that doesn't change between invocations or processes because of hash-seed randomisation.  Python has never sorted the inputs but was still deterministic, so it would be highly surprising to me if we need to introduce the overhead of sorting a set in Rust space.

jakelishman:
It might be that the trouble is that there's a non-deterministic `HashMap` iteration somewhere higher up in the `BasisTranslator` logic that ought to move to an `IndexMap` instead - it might not be this exact `HashMap`/`IndexMap` that's the problem, but the iteration that builds it.

mtreinish:
I opted to just use indexmap with ahash everywhere in the basis translator to solve the iteration determinism issues there when hashbrown version bump (see: https://github.com/Qiskit/qiskit/pull/13733/commits/fd318fd63449615d34608e61064c7f061fb3c4a0 ). This is probably overkill, but it seemed simpler than futzing around with the code to find the exact objects that needed changes. So this fixes all the test failures and to verify there are no performance impacts I ran asv and it's not flagging any regressions from this:

```

| Change   | Before [26906d33] <lint_incr_latest>   | After [a519f352] <bump-rx-core>   | Ratio   | Benchmark (Parameter)                                                                                    |
|----------|----------------------------------------|-----------------------------------|---------|----------------------------------------------------------------------------------------------------------|
|          | 0                                      | 0                                 | n/a     | utility_scale.UtilityScaleBenchmarks.track_bvlike_depth('cx')                                            |
|          | 0                                      | 0                                 | n/a     | utility_scale.UtilityScaleBenchmarks.track_bvlike_depth('cz')                                            |
|          | 0                                      | 0                                 | n/a     | utility_scale.UtilityScaleBenchmarks.track_bvlike_depth('ecr')                                           |
|          | 18.9±0.7ms                             | 19.9±0.5ms                        | 1.05    | transpiler_levels.TranspilerLevelBenchmarks.time_transpile_from_large_qasm(2)                            |
|          | 35.4±0.6ms                             | 37.2±0.8ms                        | 1.05    | transpiler_levels.TranspilerLevelBenchmarks.time_transpile_qv_14_x_14(2)                                 |
|          | 9.72±0.2ms                             | 10.1±0.3ms                        | 1.04    | utility_scale.UtilityScaleBenchmarks.time_bvlike('cx')                                                   |
|          | 9.76±0.2ms                             | 10.1±0.3ms                        | 1.04    | utility_scale.UtilityScaleBenchmarks.time_bvlike('ecr')                                                  |
|          | 69.5±0.9ms                             | 71.8±0.5ms                        | 1.03    | transpiler_levels.TranspilerLevelBenchmarks.time_schedule_qv_14_x_14(1)                                  |
|          | 19.2±0.5ms                             | 19.8±0.5ms                        | 1.03    | transpiler_levels.TranspilerLevelBenchmarks.time_transpile_qv_14_x_14(0)                                 |
|          | 27.3±2ms                               | 28.2±0.4ms                        | 1.03    | transpiler_levels.TranspilerLevelBenchmarks.time_transpile_qv_14_x_14(1)                                 |
|          | 56.9±0.8ms                             | 58.3±2ms                          | 1.03    | transpiler_levels.TranspilerLevelBenchmarks.time_transpile_qv_14_x_14(3)                                 |
|          | 817±20ms                               | 840±20ms                          | 1.03    | utility_scale.UtilityScaleBenchmarks.time_qv('cz')                                                       |
|          | 113±0.9ms                              | 116±2ms                           | 1.03    | utility_scale.UtilityScaleBenchmarks.time_square_heisenberg('cx')                                        |
|          | 67.7±3ms                               | 68.9±2ms                          | 1.02    | passes.MultipleBasisPassBenchmarks.time_basis_translator(14, 1024, ['u', 'cx', 'id'])                    |
|          | 33.9±0.9ms                             | 34.5±0.9ms                        | 1.02    | transpiler_levels.TranspilerLevelBenchmarks.time_transpile_from_large_qasm_backend_with_prop(1)          |
|          | 20.2±0.8ms                             | 20.6±0.7ms                        | 1.02    | transpiler_levels.TranspilerLevelBenchmarks.time_transpile_from_large_qasm_backend_with_prop(2)          |
|          | 456±6ms                                | 463±8ms                           | 1.02    | utility_scale.UtilityScaleBenchmarks.time_qft('cx')                                                      |
|          | 777±30ms                               | 790±10ms                          | 1.02    | utility_scale.UtilityScaleBenchmarks.time_qv('ecr')                                                      |
|          | 134±0.8ms                              | 137±3ms                           | 1.02    | utility_scale.UtilityScaleBenchmarks.time_square_heisenberg('ecr')                                       |
|          | 76.2±2ms                               | 77.0±1ms                          | 1.01    | passes.MultipleBasisPassBenchmarks.time_basis_translator(14, 1024, ['rz', 'x', 'sx', 'cx', 'id'])        |
|          | 93.7±2ms                               | 94.8±1ms                          | 1.01    | passes.MultipleBasisPassBenchmarks.time_basis_translator(20, 1024, ['u', 'cx', 'id'])                    |
|          | 28.9±0.2ms                             | 29.2±0.7ms                        | 1.01    | passes.MultipleBasisPassBenchmarks.time_basis_translator(5, 1024, ['u', 'cx', 'id'])                     |
|          | 33.4±0.9ms                             | 33.6±0.4ms                        | 1.01    | transpiler_levels.TranspilerLevelBenchmarks.time_transpile_from_large_qasm(1)                            |
|          | 29.0±0.5ms                             | 29.1±0.3ms                        | 1.01    | transpiler_levels.TranspilerLevelBenchmarks.time_transpile_from_large_qasm_backend_with_prop(0)          |
|          | 82.4±0.8ms                             | 83.6±1ms                          | 1.01    | utility_scale.UtilityScaleBenchmarks.time_bv_100('ecr')                                                  |
|          | 244±2ms                                | 248±2ms                           | 1.01    | utility_scale.UtilityScaleBenchmarks.time_qaoa('cx')                                                     |
|          | 320±9ms                                | 323±7ms                           | 1.01    | utility_scale.UtilityScaleBenchmarks.time_qaoa('ecr')                                                    |
|          | 648±10ms                               | 655±8ms                           | 1.01    | utility_scale.UtilityScaleBenchmarks.time_qv('cx')                                                       |
|          | 132±1ms                                | 133±1ms                           | 1.01    | utility_scale.UtilityScaleBenchmarks.time_square_heisenberg('cz')                                        |
|          | 38.1±0.3ms                             | 38.3±0.2ms                        | 1.00    | passes.MultipleBasisPassBenchmarks.time_basis_translator(5, 1024, ['rx', 'ry', 'rz', 'r', 'rxx', 'id'])  |
|          | 88.6±1ms                               | 88.4±1ms                          | 1.00    | transpiler_levels.TranspilerLevelBenchmarks.time_quantum_volume_transpile_50_x_20(0)                     |
|          | 362±4ms                                | 363±3ms                           | 1.00    | transpiler_levels.TranspilerLevelBenchmarks.time_quantum_volume_transpile_50_x_20(3)                     |
|          | 66.5±3ms                               | 66.5±0.9ms                        | 1.00    | transpiler_levels.TranspilerLevelBenchmarks.time_schedule_qv_14_x_14(0)                                  |
|          | 27.8±1ms                               | 27.7±0.8ms                        | 1.00    | transpiler_levels.TranspilerLevelBenchmarks.time_transpile_from_large_qasm(0)                            |
|          | 20.6±0.5ms                             | 20.5±0.6ms                        | 1.00    | transpiler_levels.TranspilerLevelBenchmarks.time_transpile_from_large_qasm(3)                            |
|          | 21.1±0.3ms                             | 21.0±0.4ms                        | 1.00    | transpiler_levels.TranspilerLevelBenchmarks.time_transpile_from_large_qasm_backend_with_prop(3)          |
|          | 1429                                   | 1429                              | 1.00    | transpiler_levels.TranspilerLevelBenchmarks.track_depth_quantum_volume_transpile_50_x_20(0)              |
|          | 1380                                   | 1380                              | 1.00    | transpiler_levels.TranspilerLevelBenchmarks.track_depth_quantum_volume_transpile_50_x_20(1)              |
|          | 1362                                   | 1362                              | 1.00    | transpiler_levels.TranspilerLevelBenchmarks.track_depth_quantum_volume_transpile_50_x_20(2)              |
|          | 1490                                   | 1490                              | 1.00    | transpiler_levels.TranspilerLevelBenchmarks.track_depth_quantum_volume_transpile_50_x_20(3)              |
|          | 2705                                   | 2705                              | 1.00    | transpiler_levels.TranspilerLevelBenchmarks.track_depth_transpile_from_large_qasm(0)                     |
|          | 2005                                   | 2005                              | 1.00    | transpiler_levels.TranspilerLevelBenchmarks.track_depth_transpile_from_large_qasm(1)                     |
|          | 7                                      | 7                                 | 1.00    | transpiler_levels.TranspilerLevelBenchmarks.track_depth_transpile_from_large_qasm(2)                     |
|          | 7                                      | 7                                 | 1.00    | transpiler_levels.TranspilerLevelBenchmarks.track_depth_transpile_from_large_qasm(3)                     |
|          | 2705                                   | 2705                              | 1.00    | transpiler_levels.TranspilerLevelBenchmarks.track_depth_transpile_from_large_qasm_backend_with_prop(0)   |
|          | 2005                                   | 2005                              | 1.00    | transpiler_levels.TranspilerLevelBenchmarks.track_depth_transpile_from_large_qasm_backend_with_prop(1)   |
|          | 7                                      | 7                                 | 1.00    | transpiler_levels.TranspilerLevelBenchmarks.track_depth_transpile_from_large_qasm_backend_with_prop(2)   |
|          | 7                                      | 7                                 | 1.00    | transpiler_levels.TranspilerLevelBenchmarks.track_depth_transpile_from_large_qasm_backend_with_prop(3)   |
|          | 402                                    | 402                               | 1.00    | transpiler_levels.TranspilerLevelBenchmarks.track_depth_transpile_qv_14_x_14(0)                          |
|          | 396                                    | 396                               | 1.00    | transpiler_levels.TranspilerLevelBenchmarks.track_depth_transpile_qv_14_x_14(1)                          |
|          | 341                                    | 341                               | 1.00    | transpiler_levels.TranspilerLevelBenchmarks.track_depth_transpile_qv_14_x_14(2)                          |
|          | 374                                    | 374                               | 1.00    | transpiler_levels.TranspilerLevelBenchmarks.track_depth_transpile_qv_14_x_14(3)                          |
|          | 77.9±2ms                               | 77.9±0.9ms                        | 1.00    | utility_scale.UtilityScaleBenchmarks.time_bv_100('cx')                                                   |
|          | 82.7±1ms                               | 83.0±0.6ms                        | 1.00    | utility_scale.UtilityScaleBenchmarks.time_bv_100('cz')                                                   |
|          | 9.77±0.08ms                            | 9.80±0.09ms                       | 1.00    | utility_scale.UtilityScaleBenchmarks.time_parse_qaoa_n100('cx')                                          |
|          | 9.83±0.09ms                            | 9.81±0.1ms                        | 1.00    | utility_scale.UtilityScaleBenchmarks.time_parse_qaoa_n100('cz')                                          |
|          | 34.0±0.4ms                             | 34.0±0.7ms                        | 1.00    | utility_scale.UtilityScaleBenchmarks.time_parse_square_heisenberg_n100('ecr')                            |
|          | 387                                    | 387                               | 1.00    | utility_scale.UtilityScaleBenchmarks.track_bv_100_depth('cx')                                            |
|          | 400                                    | 400                               | 1.00    | utility_scale.UtilityScaleBenchmarks.track_bv_100_depth('cz')                                            |
|          | 400                                    | 400                               | 1.00    | utility_scale.UtilityScaleBenchmarks.track_bv_100_depth('ecr')                                           |
|          | 300                                    | 300                               | 1.00    | utility_scale.UtilityScaleBenchmarks.track_circSU2_depth('cx')                                           |
|          | 300                                    | 300                               | 1.00    | utility_scale.UtilityScaleBenchmarks.track_circSU2_depth('cz')                                           |
|          | 300                                    | 300                               | 1.00    | utility_scale.UtilityScaleBenchmarks.track_circSU2_depth('ecr')                                          |
|          | 1727                                   | 1727                              | 1.00    | utility_scale.UtilityScaleBenchmarks.track_qaoa_depth('cx')                                              |
|          | 1734                                   | 1734                              | 1.00    | utility_scale.UtilityScaleBenchmarks.track_qaoa_depth('cz')                                              |
|          | 1734                                   | 1734                              | 1.00    | utility_scale.UtilityScaleBenchmarks.track_qaoa_depth('ecr')                                             |
|          | 2484                                   | 2484                              | 1.00    | utility_scale.UtilityScaleBenchmarks.track_qft_depth('cx')                                               |
|          | 2484                                   | 2484                              | 1.00    | utility_scale.UtilityScaleBenchmarks.track_qft_depth('cz')                                               |
|          | 2484                                   | 2484                              | 1.00    | utility_scale.UtilityScaleBenchmarks.track_qft_depth('ecr')                                              |
|          | 2694                                   | 2694                              | 1.00    | utility_scale.UtilityScaleBenchmarks.track_qv_depth('cx')                                                |
|          | 2694                                   | 2694                              | 1.00    | utility_scale.UtilityScaleBenchmarks.track_qv_depth('cz')                                                |
|          | 2694                                   | 2694                              | 1.00    | utility_scale.UtilityScaleBenchmarks.track_qv_depth('ecr')                                               |
|          | 558                                    | 558                               | 1.00    | utility_scale.UtilityScaleBenchmarks.track_square_heisenberg_depth('cx')                                 |
|          | 558                                    | 558                               | 1.00    | utility_scale.UtilityScaleBenchmarks.track_square_heisenberg_depth('cz')                                 |
|          | 558                                    | 558                               | 1.00    | utility_scale.UtilityScaleBenchmarks.track_square_heisenberg_depth('ecr')                                |
|          | 130±4ms                                | 129±1ms                           | 0.99    | passes.MultipleBasisPassBenchmarks.time_basis_translator(20, 1024, ['rx', 'ry', 'rz', 'r', 'rxx', 'id']) |
|          | 32.6±0.6ms                             | 32.2±0.5ms                        | 0.99    | passes.MultipleBasisPassBenchmarks.time_basis_translator(5, 1024, ['rz', 'x', 'sx', 'cx', 'id'])         |
|          | 117±2ms                                | 115±2ms                           | 0.99    | transpiler_levels.TranspilerLevelBenchmarks.time_quantum_volume_transpile_50_x_20(1)                     |
|          | 9.85±0.2ms                             | 9.72±0.07ms                       | 0.99    | utility_scale.UtilityScaleBenchmarks.time_parse_qaoa_n100('ecr')                                         |
|          | 104±0.9ms                              | 103±0.8ms                         | 0.99    | utility_scale.UtilityScaleBenchmarks.time_parse_qft_n100('cx')                                           |
|          | 104±2ms                                | 103±0.4ms                         | 0.99    | utility_scale.UtilityScaleBenchmarks.time_parse_qft_n100('cz')                                           |
|          | 33.7±0.5ms                             | 33.2±0.6ms                        | 0.99    | utility_scale.UtilityScaleBenchmarks.time_parse_square_heisenberg_n100('cz')                             |
|          | 97.2±1ms                               | 95.1±1ms                          | 0.98    | passes.MultipleBasisPassBenchmarks.time_basis_translator(14, 1024, ['rx', 'ry', 'rz', 'r', 'rxx', 'id']) |
|          | 179±2ms                                | 175±2ms                           | 0.98    | transpiler_levels.TranspilerLevelBenchmarks.time_quantum_volume_transpile_50_x_20(2)                     |
|          | 9.95±0.4ms                             | 9.72±0.2ms                        | 0.98    | utility_scale.UtilityScaleBenchmarks.time_bvlike('cz')                                                   |
|          | 340±5ms                                | 333±5ms                           | 0.98    | utility_scale.UtilityScaleBenchmarks.time_qaoa('cz')                                                     |
|          | 613±10ms                               | 600±10ms                          | 0.98    | utility_scale.UtilityScaleBenchmarks.time_qft('cz')                                                      |
|          | 634±10ms                               | 623±9ms                           | 0.98    | utility_scale.UtilityScaleBenchmarks.time_qft('ecr')                                                     |
|          | 104±1ms                                | 101±1ms                           | 0.97    | utility_scale.UtilityScaleBenchmarks.time_parse_qft_n100('ecr')                                          |
|          | 34.5±0.7ms                             | 33.5±0.4ms                        | 0.97    | utility_scale.UtilityScaleBenchmarks.time_parse_square_heisenberg_n100('cx')                             |
|          | 107±2ms                                | 103±3ms                           | 0.96    | passes.MultipleBasisPassBenchmarks.time_basis_translator(20, 1024, ['rz', 'x', 'sx', 'cx', 'id'])        |
|          | 889±10ms                               | 850±10ms                          | 0.96    | utility_scale.UtilityScaleBenchmarks.time_circSU2('cx')                                                  |
|          | 905±10ms                               | 857±20ms                          | 0.95    | utility_scale.UtilityScaleBenchmarks.time_circSU2('ecr')                                                 |
|          | 913±10ms                               | 855±20ms                          | 0.94    | utility_scale.UtilityScaleBenchmarks.time_circSU2('cz')                                                  |

```

crates/accelerate/src/basis/basis_translator/compose_transforms.rs
```diff
@@ -33,12 +33,14 @@ pub type BasisTransformOut = (SmallVec<[Param; 3]>, DAGCircuit);
...
     source_dag: &'a DAGCircuit,
-) -> PyResult<HashMap<GateIdentifier, BasisTransformOut>> {
-    let mut gate_param_counts: HashMap<GateIdentifier, usize> = HashMap::default();
+) -> PyResult<IndexMap<GateIdentifier, BasisTransformOut, ahash::RandomState>> {
+    let mut gate_param_counts: IndexMap<GateIdentifier, usize, ahash::RandomState> =
+        IndexMap::default();
     get_gates_num_params(source_dag, &mut gate_param_counts)?;
-    let mut mapped_instructions: HashMap<GateIdentifier, BasisTransformOut> = HashMap::new();
+    let mut mapped_instructions: IndexMap<GateIdentifier, BasisTransformOut, ahash::RandomState> =
+        IndexMap::with_hasher(ahash::RandomState::default());
```
jakelishman, line 43: lol at sometimes using `Default::default` and sometimes `with_hasher`

crates/accelerate/src/basis/basis_translator/mod.rs
```diff
@@ -64,23 +65,40 @@ fn run(
...
+        ahash::RandomState,
+    > = qargs_with_non_global_operation
+        .into_iter()
+        .map(|(k, v)| {
+            (
+                k,
+                v.into_iter().collect::<IndexSet<_, ahash::RandomState>>(),
+            )
+        })
+        .collect();
```
jakelishman, line 80: There's no impl for `IndexSet` in PyO3?  We ought to contribute that upstream to avoid the double copy-in here.

jakelishman (approved):
This looks fine, thanks - probably we don't need _everything_ to be `IndexMap`/`IndexSet`, but I have no desire to track down the minimal set of changes.

Surprised that PyO3 seems to be missing the `impl FromPyObject` for `IndexSet` - we ought to contribute that upstream.

crates/accelerate/src/basis/basis_translator/mod.rs
jakelishman: Turns out that Ivan was the one who added the `IndexMap` impl in PyO3/pyo3#1728, and deliberately cut `IndexSet` for the totally valid reason that `set` doesn't preserve insertion order, so the semantics wouldn't be the same.

In our case, `set[int]` has a run-independent iteration order as an implementation detail of CPython, whereas `HashSet<int>` _doesn't_, and I suspect that "implementation detail of CPython" isn't sufficient motivation to include it here.

Also worth pointing out that because this `IndexSet` is built _from_ a `HashSet`, the iteration of the `IndexSet` is actually not really defined either, because the insertion order isn't defined.  So this change to an `IndexSet` internally is a bit of a misnomer.

IvanIsCoding:
> This looks fine, thanks - probably we don't need _everything_ to be `IndexMap`/`IndexSet`, but I have no desire to track down the minimal set of changes.
> 
> Surprised that PyO3 seems to be missing the `impl FromPyObject` for `IndexSet` - we ought to contribute that upstream.

That one is on me! https://github.com/PyO3/pyo3/pull/1728#issuecomment-881776843

At the time we thought there was no use for IndexSet as Python’s set would mess with the ordering regardless.

jakelishman:
Ivan: yeah - I realised a couple of minutes later haha (https://github.com/Qiskit/qiskit/pull/13733#discussion_r1965805861)

IvanIsCoding:
No worries. Personally, I am not aware of the requirements for this code but if you want a consistent set input from Python, `BTreeSet` will implicitly sort the items.

For returning the value to Python though, there is no work around AFAIK.

PR #13727, opened 2025-01-23
Fix parameter deserialization bugs in `QPY`

ElePT:
### Summary
This PR fixes a series of reported issues in QPY:

- https://github.com/Qiskit/qiskit/issues/13720
- https://github.com/Qiskit/qiskit/issues/14088
- https://github.com/Qiskit/qiskit/issues/14089

The common denominator for these issues is the assumption that parameter names are unique within a circuit. This may not be true in cases where parameter re-assignment has taken place and both the old and new parameters share a name (but have different uuids). In these cases, the substitution operation is recorded in the `qpy_replay` field and should be taken into account during serialization. The bug materialized in two flavors:

1. name clash when deserializing the `ParameterVector` component of a `ParameterExpression` (for which a new `from_expr` flag has been added to `read_parameter_vec`). This came up with `NLocal` implementations (including subclasses).
2. name clash when deserializing the `ParameterExpression` directly. This use case is tested in `test_duplicated_param_name`, and fix by tweaking the encoded mapping to use uuids as keys instead of names.

This fix tweaks the way `ParameterExpressions` are stored, so it probably needs a QPY version bump.

PR #13735, opened 2025-01-25
Reduce parallelism for qpy tests

mtreinish:
### Summary

In #13506 the QPY backwards compatibility tests are failing because the image is running out of disk space and github actions is killing the runner. The qpy files themselves should be fairly small and will be dwarfed by the virtual environments created to install old versions of qiskit to generate old QPY payloads. We use GNU parallel to speed up the testing, however by doing this we end up having multiple venvs at once which increases the disk space usage. This commit attempts to reduce the pressure on the disk space usage by decreasing the parallelism from N cpus (which is currently 4 for linux runners according to [1]) to 2. This will increase the runtime for this job but if we have insufficient disk space on the image to generate new qpy files there isn't really a choice.

### Details and comments

[1] https://docs.github.com/en/actions/using-github-hosted-runners/using-github-hosted-runners/about-github-hosted-runners#standard-github-hosted-runners-for-public-repositories

### DIFF
--- a/test/qpy_compat/run_tests.sh
+++ b/test/qpy_compat/run_tests.sh
@@ -42,7 +42,7 @@ python -m venv "$qiskit_venv"
 "$qiskit_venv/bin/pip" install -c "$repo_root/constraints.txt" "$qiskit_dev_wheel" packaging
 
 # Run all of the tests of cross-Qiskit-version compatibility.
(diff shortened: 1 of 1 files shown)


ElePT (approved):
LGTM. I ran into the same issue when I added a qpy backwards compatibility test in #13727

PR #4040, opened 2020-03-29
Cache lexicographical topological sort key in DAGNode

mtreinish:
### Summary

This commit caches the string sort key used for lexicographical
topological sort as an attribute in DAGNode objects. Currently when
retworkx's lexicographical topological sort call is called it needs to
call the python callable provided to get a string used for sorting.
Prior to this commit about ~70% of the cumulative time spent in the
lexicographical topological sort function is spent in python. This is
because every time it is sorting a DAGNode it has to call
str(x.qargs) which spends all that time constructing the string
including a dict.get() and bit and register repr() calls, etc. To avoid
this unnecessary overhead at run time this commit adds a sort_key
attribute to the DAGNode class which is set during `__init__()` and
updates to the qargs property. Then the DAGCircuit lexicographical
topological sort wrapper method changes the callable to use this
attribute. Doing this eliminates all the overhead from calling back to
python because it simply becomes attribute access.

### Details and comments

PR #13736, opened 2025-01-26
Remove sort_key attribute from DAGNode

mtreinish:
### Summary

Prior to having the DAGCircuit in rust the `sort_key` attribute was added as a cache to speed up the lexicographical topological sort. Prior to having this attribute the topological sort method would compute the sort key as a string on the fly which ended up being a large bottleneck in transpilation (see: #4040 for more details). However, since migrating the DAGCircuit implementation to Rust this sort_key attribute isn't needed anymore because we call the rustworkx-core function with a tuple of bit objects instead of a string. The `sort_key` atribute was left on in place for backwards compatibility (it shouldn't have been public, this was a mistake in #4040) and when we create a python DAGNode object that will be returned to Python the creation of the sort key is unnecessary overhead (it takes roughly 1% of profile time to format the sort_key string during transpilation). Since nothing in DAGCircuit is actually using it this commit removes it to save the CPU cycles creating the string on each dag creation. We will need to add a deprecation to 1.4.0 to mark this removal for 2.0.0 since this was missed in 1.3.0.

### Details and comments

Deprecation in: https://github.com/Qiskit/qiskit/pull/13744

### DIFF
--- a/crates/circuit/src/dag_node.rs
+++ b/crates/circuit/src/dag_node.rs
@@ -114,8 +114,6 @@ impl DAGNode {
 #[pyclass(module = "qiskit._accelerate.circuit", extends=DAGNode)]
 pub struct DAGOpNode {
     pub instruction: CircuitInstruction,
-    #[pyo3(get)]
-    pub sort_key: PyObject,
 }
 
 #[pymethods]
@@ -131,7 +129,6 @@ impl DAGOpNode {
     ) -> PyResult<Py<Self>> {
         let py_op = op.extract::<OperationFromPython>()?;
         let qargs = qargs.map_or_else(|| PyTuple::empty(py), |q| q.value);
-        let sort_key = qargs.str().unwrap().into();
         let cargs = cargs.map_or_else(|| PyTuple::empty(py), |c| c.value);
         let instruction = CircuitInstruction {
             operation: py_op.operation,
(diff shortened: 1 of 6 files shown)


crates/circuit/src/dag_node.rs
```diff
@@ -114,8 +114,6 @@ impl DAGNode {
 #[pyclass(module = "qiskit._accelerate.circuit", extends=DAGNode)]
 pub struct DAGOpNode {
     pub instruction: CircuitInstruction,
```
kevinhartman, line 116: Nice! Maybe this should become a tuple struct now. Feel free to leave as is if that's your preference.

qiskit/transpiler/passes/routing/star_prerouting.py
```diff
@@ -331,7 +339,15 @@ def star_preroute(self, dag, blocks, processing_order):
         }
 
         def tie_breaker_key(node):
-            return processing_order_index_map.get(node, node.sort_key)
+            processing_order = processing_order_index_map.get(node, None)
+            if processing_order is not None:
+                return processing_order
+            else:
```
kevinhartman: This `else` block would be redundant, I suppose

kevinhartman (dismissed):
It's nice to see this removed 🙂 (it was a headache for me when we began porting `DAGCircuit` to Rust, initially).

Can you also link the deprecation PR in the description once it is out?

crates/circuit/src/dag_node.rs
mtreinish: I kept it this way for simplicity since I just deleted the lines with the previous attribute. I did overlook there is only a single field now. But I think I'd probably rather keep it the same because I feel like there are a few places we do `node.instruction`/`self.instruction` and that feels a bit more ergonomic than `node.0`

qiskit/transpiler/passes/routing/star_prerouting.py
mtreinish: Simplified in: https://github.com/Qiskit/qiskit/pull/13736/commits/55b0ddfc1f94e51cb36fef228f025fdba12b5c26

kevinhartman (approved):

PR #13737, opened 2025-01-26
Promote linux aarch64 to tier 1 support

mtreinish:
### Summary

With native linux aarch64 runners available from github now available [1] we can run tests on every PR on the platform and promote it to tier 1 support [2]. This commit adds a github action test job that runs on PRs to test that the platform works on every commit. Then the wheel job is promoted to the tier 1 job which includes PGO which will improve the performance on the wheels we publish for the platform.

While there is a larger effort to migrate our entire CI to github actions in #13474 expanding our test matrix prior to that won't cause a conflict as we can easily expand the test matrix in that PR if this merges first.

### Details and comments

[1]: https://github.blog/changelog/2025-01-16-linux-arm64-hosted-runners-now-available-for-free-in-public-repositories-public-preview/
[2]: https://docs.quantum.ibm.com/guides/install-qiskit#operating-system-support


.github/workflows/tests.yml
```diff
@@ -13,19 +13,17 @@ concurrency:
...
+    name: ${{ matrix.os }}-arm64-tests-Python-${{ matrix.python-version }}
+    runs-on: ${{ matrix.os }}
     strategy:
       fail-fast: false
       matrix:
         # Normally we test min and max version but we can't run python
         # 3.9 on arm64 until actions/setup-python#808 is resolved
-        python-version: ["3.10", "3.13"]
+        python-version: ["3.9", "3.13"]
+        os: ["macOS-14", "ubuntu-24.04-arm"]
```
jakelishman: This comment must be out-of-date now?

.github/workflows/wheels-build.yml
```diff
@@ -123,6 +114,52 @@ jobs:
...
+          mkdir -p "$PGO_WORK_DIR"
+
+          cat >>"$GITHUB_ENV" <<EOF
+          CIBW_BEFORE_ALL=yum install -y wget && {package}/tools/install_rust_msrv.sh
+          CIBW_BEFORE_BUILD=bash ./tools/build_pgo.sh $PGO_WORK_DIR $PGO_OUT_PATH
+          CIBW_ENVIRONMENT_LINUX=RUSTFLAGS='-Cprofile-use=$PGO_OUT_PATH -Cllvm-args=-pgo-warn-missing-function' PATH="\$PATH:\$HOME/.cargo/bin" CARGO_NET_GIT_FETCH_WITH_CLI="true"
+          EOF
+        env:
+          PGO_WORK_DIR: ${{ github.workspace }}/pgo-data
+          PGO_OUT_PATH: ${{ github.workspace }}/merged.profdata
```
jakelishman: I'm fine to live this in a separate job for the initial support just to get it merged after I know how much of a slog it just was, but I might like to follow up with a little bit of logic that moves the AArch64 config into the regular Tier 1 matrix. Now that we've got the working config, and the difference seems to be only in the `CIBW_BEFORE_ALL` and not setting `RUSTUP_TOOLCHAIN` (for the env config, I mean), I think it's possible to de-duplicate the job a bit more, which is just a nice-to-have - we've definitely missed updating configs in the past in the CI jobs when we had them duplicated (on Aer more than Terra, but still).

releasenotes/notes/arm-linux-tier1-fa79edaabf565685.yaml
```diff
@@ -0,0 +1,8 @@
+---
+features:
+  - |
+    Support for the Linux aarch64 platform has been promoted to tier 1 support as documented in:
+
+    https://docs.quantum.ibm.com/guides/install-qiskit#operating-system-support
+
+    from it's previous support level of tier 2 in the 1.x release series.
```
jakelishman: "its", no apostrophe

tools/install_rust_msrv.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/sh
+if [ ! -d rust-installer ]; then
+    mkdir rust-installer
+    wget https://sh.rustup.rs -O rust-installer/rustup.sh
+    sh rust-installer/rustup.sh -y -c llvm-tools
+    rustup default 1.79
+    rustup component add llvm-tools
+fi
+. "$HOME/.cargo/env"
```
jakelishman: Since this is set to the MSRV, we probably also should include a link to it in `Cargo.toml`'s `rust-version` field, where we keep track of all the places that need updating.  I could easily see this getting forgotten on an MSRV bump, and I don't think we habitually build the wheels on those commits.

tools/build_pgo.sh
```diff
@@ -34,4 +34,8 @@ python tools/pgo_scripts/test_utility_scale.py
 
 deactivate
 
-${HOME}/.rustup/toolchains/*$arch*/lib/rustlib/$arch*/bin/llvm-profdata merge -o "$out_path" "$work_dir"
+if [[ `uname -m` == "aarch64" ]]; then
+    ${HOME}/.rustup/toolchains/1.79-aarch64-unknown-linux-gnu/lib/rustlib/$arch*/bin/llvm-profdata merge -o "$out_path" "$work_dir"
+else
+    ${HOME}/.rustup/toolchains/*$arch*/lib/rustlib/$arch*/bin/llvm-profdata merge -o "$out_path" "$work_dir"
+fi
```
jakelishman: I'm not a big fan of how we're hard-coding the paths to the Rust toolchain in this file, making assumptions that we used MSRV for AArch64 - feels very easy to get out-of-sync and harder to update in the future.  Can we refactor slightly so that the `install_rust_{,_msrv}.sh` or relevant GitHub Actions jobs detect the installation location of `llvm-profdata` at install time, and feed the path into this script by command-line argument?

jakelishman (commented):
Thanks for the giant slog getting this to a point where the PGO finally passes.  I'm excited to get AArch64 up to Tier 1.

*edit*: I posted this review pretty immediately in excitement when I saw the PGO pass - sorry if I jumped the gun and you hadn't had time to do any tidy up you were planning.

tools/build_pgo.sh
mtreinish: I fixed this in: https://github.com/Qiskit/qiskit/pull/13737/commits/fae956b52c0198c73d587a816d2101635c945b9e I just made sure in the msrv path we only install a single toolchain so the glob find the correct path.

releasenotes/notes/arm-linux-tier1-fa79edaabf565685.yaml
mtreinish: Fixed in: https://github.com/Qiskit/qiskit/pull/13737/commits/b19817f4426415b6f3a5d7c748a256177b02609c

tools/install_rust_msrv.sh
mtreinish: Added in: https://github.com/Qiskit/qiskit/pull/13737/commits/20f7e0a820678b21efc5f081adab96e42f943add

.github/workflows/tests.yml
mtreinish: Removed in: https://github.com/Qiskit/qiskit/pull/13737/commits/962594e1de794ea04072d2b47720e1f2953325e6

.github/workflows/wheels-build.yml
mtreinish: I've deduplicated the job to do the toolchain switching logic in the run step (assuming I got the logic and syntax correct in the current iteration)

mtreinish:
This should be ready to go now. 

For those following along, the issue seems to be a build issue in the llvm-tools libraries bundled with rustup on aarch64 linux in both 1.84.x and 1.85. It is trying to find non-portable symbols when linking the libraries used for PGO. The MSRV version works fine so for the time being this PR opts to use that. I'm working on bisecting the failure a bit more locally on an rpi 5 (which is quite slow for repeated compilations) but plan to report it upstream once I get a few more details.

.github/workflows/wheels-build.yml
```diff
@@ -108,12 +100,19 @@ jobs:
...
+              RUST_TOOLCHAIN=stable
+          fi
           cat >>"$GITHUB_ENV" <<EOF
+          CIBW_BEFORE_ALL_LINUX=yum install -y wget && {package}/$INSTALL_RUST_PATH
           CIBW_BEFORE_BUILD=bash ./tools/build_pgo.sh $PGO_WORK_DIR $PGO_OUT_PATH
           CIBW_ENVIRONMENT=RUSTUP_TOOLCHAIN=stable RUSTFLAGS='-Cprofile-use=$PGO_OUT_PATH -Cllvm-args=-pgo-warn-missing-function'
           CIBW_ENVIRONMENT_MACOS=MACOSX_DEPLOYMENT_TARGET='10.12' RUSTUP_TOOLCHAIN=stable RUSTFLAGS='-Cprofile-use=$PGO_OUT_PATH -Cllvm-args=-pgo-warn-missing-function'
-          CIBW_ENVIRONMENT_LINUX=RUSTUP_TOOLCHAIN=stable RUSTFLAGS='-Cprofile-use=$PGO_OUT_PATH -Cllvm-args=-pgo-warn-missing-function' PATH="\$PATH:\$HOME/.cargo/bin" CARGO_NET_GIT_FETCH_WITH_CLI="true"
+          CIBW_ENVIRONMENT_LINUX=RUSTUP_TOOLCHAIN=$RUST_TOOLCHAIN RUSTFLAGS='-Cprofile-use=$PGO_OUT_PATH -Cllvm-args=-pgo-warn-missing-function' PATH="\$PATH:\$HOME/.cargo/bin" CARGO_NET_GIT_FETCH_WITH_CLI="true"
           EOF
```
jakelishman, line 116: This is really nice - the logic's super clear to me now.

.github/workflows/tests.yml
```diff
@@ -13,19 +13,15 @@ concurrency:
 jobs:
   tests:
     if: github.repository_owner == 'Qiskit'
-    name: macOS-arm64-tests-Python-${{ matrix.python-version }}
-    runs-on: macOS-14
+    name: ${{ matrix.os }}-arm64-tests-Python-${{ matrix.python-version }}
+    runs-on: ${{ matrix.os }}
```
jakelishman, line 17: We branch-protect on these names - we'll need to update them in the protection settings before this can merge.

jakelishman (approved):
Thanks for all the work - this looks good to me to merge now.  I'll approve, but we need to take care with the branch protection rules.  We either need to:
- temporarily disable the `macOS-arm64-tests-Python-3.*` requirements for branch protection until this has merged, then enable protection on the four new test names
- immediately switch the protection rules over, hold the queue so this merges clean first, then update pending PRs so they get the new test runs

I need to leave now, so I'll leave it up to you to do whichever you want.

PR #13743, opened 2025-01-27
Remove use of deprecated objects in `BasicSimulator`

ElePT:
### Summary
This PR refactors `BasicSimulator` not to rely on deprecated objects such as:

- `_assemble`
- ` qobj`
- `BackendConfiguration`

Only the last removal is considered user facing. 

### Details and comments
Note that because of the removal of `c_if` (documented in #13506), there is currently no support for control flow operations in `BasicSimulator`. The "new" control flow operations were never supported in `BasicSimulator`, and this PR removes the outdated `c_if` support. The potential implementation of proper control flow is left as a future improvement as users can always resort to `AerSimulator` for this. 

Similarly, as a side effect of the detanglement of `assemble`, the `BasicSimulator` class no longer supports providing parameter bounds to its `run` method. This functionality was not explicitly documented and is likely not expected by users, as it has not been part of the "standard" backend.run workflow for a while. However, a few legacy tests made use of it together with `assemble`. The tests are removed as part of the PR.

TODO:
- [x] clean up code
- [x] reno


qiskit/providers/basic_provider/basic_simulator.py
```diff
@@ -555,55 +458,52 @@ def run(
...
             "success": True,
             "time_taken": (end - start),
-            "header": qobj.header.to_dict(),
         }
 
         return Result.from_dict(result)
 
-    def run_experiment(self, experiment: QasmQobjExperiment) -> dict[str, ...]:
-        """Run an experiment (circuit) and return a single experiment result.
+    def _run_circuit(self, circuit) -> dict:
```
ElePT, line 506: This method was part of the public API so unfortunately we will need a deprecation PR in 1.4 prior to its "privatization" 🥲🥲

qiskit/providers/basic_provider/basic_simulator.py
```diff
@@ -432,38 +357,36 @@ def _validate_initial_statevector(self) -> None:
...
         if "backend_options" in backend_options and backend_options["backend_options"]:
             backend_options = backend_options["backend_options"]
 
         # Check for custom initial statevector in backend_options first,
         # then config second
-        if (
-            "initial_statevector" in backend_options
-            and backend_options["initial_statevector"] is not None
-        ):
+        if getattr(backend_options, "initial_statevector", None) is not None:
```
eliarbel: `backend_options` is a dictionary not `Options`,  so `getattr` shouldn't work here checking for a contained value. 
Worth adding BTW, a test in `test_basic_simulator.py` with `initial_statevector` to make sure it's honored

qiskit/providers/basic_provider/basic_simulator.py
```diff
@@ -529,7 +436,7 @@ def run(
             BasicProviderJob: derived from BaseJob
 
         Additional Information:
-            backend_options: Is a dict of options for the backend. It may contain
+            backend_options: Is a dict of options for the backend. It may contain:
                 * "initial_statevector": vector_like
 
             The "initial_statevector" option specifies a custom initial
```
eliarbel:
```suggestion
            The "initial_statevector" option specifies a custom
```
duplicate _initial_

qiskit/providers/basic_provider/basic_simulator.py
```diff
@@ -555,55 +458,52 @@ def run(
...
+        self._set_options(backend_options=backend_options)
         job_id = str(uuid.uuid4())
-        job = BasicProviderJob(self, job_id, self._run_job(job_id, qobj))
+        job = BasicProviderJob(self, job_id, self._run_job(job_id, run_input))
         return job
 
-    def _run_job(self, job_id: str, qobj: QasmQobj) -> Result:
-        """Run experiments in qobj
+    def _run_job(self, job_id: str, run_input) -> Result:
+        """Run circuits in job.
```
eliarbel:
```suggestion
        """Run circuits in run_input.
```

qiskit/providers/basic_provider/basic_simulator.py
```diff
@@ -555,55 +458,52 @@ def run(
...
-        self._validate(qobj)
+        if isinstance(run_input, QuantumCircuit):
+            run_input = [run_input]
+
+        self._validate(run_input)
         result_list = []
-        self._shots = qobj.config.shots
-        self._memory = getattr(qobj.config, "memory", False)
-        self._qobj_config = qobj.config
+        self._memory = True
```
eliarbel: Why is this set to True rather than using the value given in `self.options`? 
Another thing BTW worth testing in `test_basic_simulator.py` (that `memory` is honored)

qiskit/providers/basic_provider/basic_simulator.py
```diff
@@ -555,55 +458,52 @@ def run(
...
-        for experiment in qobj.experiments:
-            result_list.append(self.run_experiment(experiment))
+        for circuit in run_input:
+            result_list.append(self._run_circuit(circuit))
         end = time.time()
         result = {
             "backend_name": self.name,
             "backend_version": self.backend_version,
-            "qobj_id": qobj.qobj_id,
+            "qobj_id": None,
```
eliarbel, line 496: I assume this expected by Result's spec, can you confirm? or otherwise can we just remove `qobj_id`?

qiskit/providers/basic_provider/basic_simulator.py
```diff
@@ -555,55 +458,52 @@ def run(
...
             "time_taken": (end - start),
-            "header": qobj.header.to_dict(),
         }
 
         return Result.from_dict(result)
 
-    def run_experiment(self, experiment: QasmQobjExperiment) -> dict[str, ...]:
-        """Run an experiment (circuit) and return a single experiment result.
+    def _run_circuit(self, circuit) -> dict:
+        """Simulate a single circuit run.
```
eliarbel, line 507: In the Returns section below, it would be nice to mention the `header` field which was added

eliarbel (changes requested):
Overall seems good, thanks @ElePT! 

I've left some minor line comments which need addressing. 

The only concern I have is the possibe impact of removing support for the legacy `c_if` on projects which for some reason still use it. I guess we'll just have to rely on testing against `main` before the 2.0 release to reveal that, if at all.

qiskit/providers/basic_provider/basic_simulator.py
ElePT: Done in https://github.com/Qiskit/qiskit/pull/13750

qiskit/providers/basic_provider/basic_simulator.py
ElePT: Yes, the idea would be to remove the field altogether from the `Result` class in a followup (when the rest of `qobj` gets removed). Maybe I could add a todo to make sure it doesn't slip between the cracks.

qiskit/providers/basic_provider/basic_simulator.py
ElePT: Added in [e14ae50](https://github.com/Qiskit/qiskit/pull/13743/commits/e14ae5058884646e3275130c2b3addd8f942f080)

qiskit/providers/basic_provider/basic_simulator.py
ElePT: Yes, this was an oversight, I didn't realize the original code set self._memory twice. Fixed in [8370c17](https://github.com/Qiskit/qiskit/pull/13743/commits/8370c1780ca5f63d60f37857702e6189af244bee), where the default option is changed to `True` to reflect the actual usage.

qiskit/providers/basic_provider/basic_simulator.py
ElePT: Thanks for raising this! I realized a series of oversights in the option-setting part of the code, starting from the fact that the types were wrongly documented, that the options attribute wasn't really reflecting the internal variables, that `BackendV2` was using outdated methods... The documentation now is more consistent with the aer simulator too after [8370c17](https://github.com/Qiskit/qiskit/pull/13743/commits/8370c1780ca5f63d60f37857702e6189af244bee)

qiskit/providers/basic_provider/basic_simulator.py
```diff
@@ -215,76 +204,13 @@ def _build_basic_target(self) -> Target:
...
     def _default_options(cls) -> Options:
         return Options(
             shots=1024,
-            memory=False,
+            memory=True,
             initial_statevector=None,
-            chop_threshold=1e-15,
-            allow_sample_measuring=True,
             seed_simulator=None,
-            parameter_binds=None,
```
ElePT, line 287: I realized that these options weren't used anywhere so I removed the fields.

1ucian0: Maybe extend the reno with these removals? Should they be deprecated in 1.4?

1ucian0:
This PR is blocking #13748. It would be great to merge it this week.

qiskit/providers/basic_provider/basic_simulator.py
ElePT: These are not part of the documented API, they were actually not even documented informally, so the deprecation isn't needed. I added the "surviving" options to the docstring of the run method to have them documented somewhere. I could add that to the reno, but I don't think it's required.

qiskit/providers/basic_provider/basic_simulator.py
```diff
@@ -512,102 +419,113 @@ def _validate_measure_sampling(self, experiment: QasmQobjExperiment) -> None:
...
         Returns:
             BasicProviderJob: derived from BaseJob
 
         Additional Information:
-            backend_options: Is a dict of options for the backend. It may contain
-                * "initial_statevector": vector_like
+            * kwarg options specified in ``run_options`` will temporarily override
+              any set options of the same name for the current run. These may include:
+
+                * "initial_statevector": vector_like. The "initial_statevector"
```
eliarbel:
```suggestion
                * "initial_statevector": vector-like. The "initial_statevector"
```

qiskit/providers/basic_provider/basic_simulator.py
```diff
@@ -512,102 +419,113 @@ def _validate_measure_sampling(self, experiment: QasmQobjExperiment) -> None:
...
+                  the number of qubits in the ``run_input`` argument.
 
-            The "initial_statevector" option specifies a custom initial
-            initial statevector for the simulator to be used instead of the all
-            zero state. This size of this vector must be correct for the number
-            of qubits in ``run_input`` parameter.
+                * "seed_simulator": int. This is the internal seed for sample
+                  generation.
+
+                * "shots": int. number of shots used in the simulation.
```
eliarbel:
```suggestion
                * "shots": int. Number of shots used in the simulation.
```

qiskit/providers/basic_provider/basic_simulator.py
```diff
@@ -512,102 +419,113 @@ def _validate_measure_sampling(self, experiment: QasmQobjExperiment) -> None:
...
 
         Returns:
              A result dictionary which looks something like::
-
                 {
-                "name": name of this experiment (obtained from qobj.experiment header)
+                "name": name of this experiment
                 "seed": random seed used for simulation
                 "shots": number of shots used in the simulation
+                "header: {
```
eliarbel:
```suggestion
                "header": {
```

qiskit/providers/basic_provider/basic_simulator.py
```diff
@@ -784,39 +643,54 @@ def run_experiment(self, experiment: QasmQobjExperiment) -> dict[str, ...]:
...
+            "n_qubits": circuit.num_qubits,
+            "qreg_sizes": [[qreg.name, qreg.size] for qreg in circuit.qregs],
+            "creg_sizes": [[creg.name, creg.size] for creg in circuit.cregs],
+            "qubit_labels": [[qreg.name, j] for qreg in circuit.qregs for j in range(qreg.size)],
+            "clbit_labels": [[creg.name, j] for creg in circuit.cregs for j in range(creg.size)],
+            "memory_slots": circuit.num_clbits,
+            "global_phase": circuit.global_phase,
+            "metadata": circuit.metadata if circuit.metadata is not None else {},
+        }
+        print(header)
```
eliarbel:
```suggestion
```

eliarbel (changes requested):
Thanks @ElePT for the fixes and adding the tests. Just a few tiny doc touches and a clean and this should be ready to go.

eliarbel (approved):
Cheers!

PR #13752, opened 2025-01-28
Add merging of `Interner`s

jakelishman:
### Summary

This adds ways of merging two interners together, including ways that re-use temporary allocations.  The intention here is to make it much cheaper to rebuild DAGs and circuits from many small parts (like synthesis passes, basis-translation, etc), and to avoid many small-scale allocations when doing qarg remapping.  The small-scale allocations are comparatively costly of themselves, and can often have amplified effects in threaded contexts (though it depends on the system allocator in use).

There's quite a few places that this could be used, such as `UnitarySynthesis` and `BasisTranslator`, but both of those will be substantial amounts of work to convert, and more appropriate for separate patches.  Given the performance improvements in this PR, I suspect that they should both see decent improvements too.

### Details and comments

Commit messages:

#### Add merging of `Interner` instances
    
This makes it possible to merge one interner into another, using a map-and-filter function to explain how the values should be considered afterwards.  The main use-case for this is in DAG and circuit substitution methods, where we want to be able to map `PackedInstruction`s from a small circuit to new ones on a larger circuit, without having to unwrap and re-hash the interned slices once for instruction; instead, we do that once _per interner key_, and then have a no-hash direct lookup table to remap the interner keys directly.
    
The methods here are designed to directly allow the reuse of the heap allocations used to track the interner key mappings, to help reduce the number of small-scale allocations done in substitution methods.
    
The filtering part of the mapping is because one cannot typically be certain that an `Interner` contains the minimally necessary keys, especially for DAGs; it can well be that substitutions and bit removals cause old data to be present.

#### Use new `Interner` merging in `DAGCircuit::from_circuit`
    
This converts one obvious candidate for the new interner-merging functionality.  There are several others in the codebase (all uses of `DAGCircuit::extend`, for example, and several other full rebuilds), but the structure of their code wasn't designed with this in mind, so the modifications would typically be far more involved and more suitable for separate patches.
    
Using a timing script:
    
```python
import itertools

qc = QuantumCircuit(100, 100)
for pair in itertools.permutations(qc.qubits, 2):
    qc.rz(0, pair[0])
    qc.sx(pair[0])
    qc.rz(0, pair[0])
    qc.rz(0, pair[1])
    qc.sx(pair[1])
    qc.rz(0, pair[1])
    qc.cx(*pair)
    
%timeit circuit_to_dag(qc, copy_operations=False)
```
    
this patch showed an improvement from 18.5(6)ms to 14.4(5)ms on my machine, which is a 22% speedup, though admittedly this particular function was doing larger-scale allocation work that could be removed than other candidate replacements are.

### DIFF
--- a/crates/circuit/src/dag_circuit.rs
+++ b/crates/circuit/src/dag_circuit.rs
@@ -25,7 +25,7 @@ use crate::dag_node::{DAGInNode, DAGNode, DAGOpNode, DAGOutNode};
 use crate::dot_utils::build_dot;
 use crate::error::DAGCircuitError;
 use crate::imports;
-use crate::interner::{Interned, Interner};
+use crate::interner::{Interned, InternedMap, Interner};
 use crate::operations::{Operation, OperationRef, Param, PyInstruction, StandardGate};
 use crate::packed_instruction::{PackedInstruction, PackedOperation};
 use crate::rustworkx_core_vnext::isomorphism;
@@ -4810,6 +4810,76 @@ impl DAGCircuit {
         &self.vars
     }
 
+    /// Merge the `qargs` in a different [Interner] into this DAG, remapping the qubits.
+    ///
+    /// This is useful for simplifying the direct mapping of [PackedInstruction]s from one DAG to
+    /// another, like in substitution methods, or rebuilding a new DAG out of a lot of smaller ones.
+    /// See [Interner::merge_map_slice] for more information on the mapping function.
+    ///
+    /// The input [InternedMap] is cleared of its previous entries by this method, and then we
+    /// re-use the allocation.
+    pub fn merge_qargs_using(
+        &mut self,
+        other: &Interner<[Qubit]>,
+        map_fn: impl FnMut(&Qubit) -> Option<Qubit>,
+        map: &mut InternedMap<[Qubit]>,
+    ) {
+        // 4 is an arbitrary guess for the amount of stack space to allocate for mapping the
+        // `qargs`, but it doesn't matter if it's too short because it'll safely spill to the heap.
+        self.qargs_interner
+            .merge_map_slice_using::<4>(other, map_fn, map);
+    }
+
+    /// Merge the `qargs` in a different [Interner] into this DAG, remapping the qubits.
+    ///
+    /// This is useful for simplifying the direct mapping of [PackedInstruction]s from one DAG to
+    /// another, like in substitution methods, or rebuilding a new DAG out of a lot of smaller ones.
+    /// See [Interner::merge_map_slice] for more information on the mapping function.
+    pub fn merge_qargs(
+        &mut self,
+        other: &Interner<[Qubit]>,
+        map_fn: impl FnMut(&Qubit) -> Option<Qubit>,
+    ) -> InternedMap<[Qubit]> {
+        let mut out = InternedMap::new();
+        self.merge_qargs_using(other, map_fn, &mut out);
+        out
+    }
+
+    /// Merge the `cargs` in a different [Interner] into this DAG, remapping the clbits.
+    ///
+    /// This is useful for simplifying the direct mapping of [PackedInstruction]s from one DAG to
+    /// another, like in substitution methods, or rebuilding a new DAG out of a lot of smaller ones.
+    /// See [Interner::merge_map_slice] for more information on the mapping function.
+    ///
+    /// The input [InternedMap] is cleared of its previous entries by this method, and then we
+    /// re-use the allocation.
+    pub fn merge_cargs_using(
+        &mut self,
+        other: &Interner<[Clbit]>,
+        map_fn: impl FnMut(&Clbit) -> Option<Clbit>,
+        map: &mut InternedMap<[Clbit]>,
+    ) {
+        // 4 is an arbitrary guess for the amount of stack space to allocate for mapping the
+        // `cargs`, but it doesn't matter if it's too short because it'll safely spill to the heap.
+        self.cargs_interner
+            .merge_map_slice_using::<4>(other, map_fn, map);
+    }
+
+    /// Merge the `cargs` in a different [Interner] into this DAG, remapping the clbits.
+    ///
+    /// This is useful for simplifying the direct mapping of [PackedInstruction]s from one DAG to
+    /// another, like in substitution methods, or rebuilding a new DAG out of a lot of smaller ones.
+    /// See [Interner::merge_map_slice] for more information on the mapping function.
+    pub fn merge_cargs(
+        &mut self,
+        other: &Interner<[Clbit]>,
+        map_fn: impl FnMut(&Clbit) -> Option<Clbit>,
+    ) -> InternedMap<[Clbit]> {
+        let mut out = InternedMap::new();
+        self.merge_cargs_using(other, map_fn, &mut out);
+        out
+    }
+
     /// Return an iterator of gate runs with non-conditional op nodes of given names
     pub fn collect_runs(
         &self,
@@ -6383,10 +6453,24 @@ impl DAGCircuit {
         &self.op_names
     }
 
-    /// Extends the DAG with valid instances of [PackedInstruction]
+    /// Extends the DAG with valid instances of [PackedInstruction].
     pub fn extend<I>(&mut self, py: Python, iter: I) -> PyResult<Vec<NodeIndex>>
     where
         I: IntoIterator<Item = PackedInstruction>,
+    {
+        self.try_extend(
+            py,
+            iter.into_iter()
+                .map(|inst| -> Result<PackedInstruction, Infallible> { Ok(inst) }),
+        )
+    }
+
+    /// Extends the DAG with valid instances of [PackedInstruction], where the iterator produces the
+    /// results in a fallible manner.
+    pub fn try_extend<I, E>(&mut self, py: Python, iter: I) -> PyResult<Vec<NodeIndex>>
+    where
+        I: IntoIterator<Item = Result<PackedInstruction, E>>,
+        PyErr: From<E>,
     {
         // Create HashSets to keep track of each bit/var's last node
         let mut qubit_last_nodes: HashMap<Qubit, NodeIndex> = HashMap::default();
@@ -6400,6 +6484,7 @@ impl DAGCircuit {
         // Store new nodes to return
         let mut new_nodes = Vec::with_capacity(iter.size_hint().1.unwrap_or_default());
         for instr in iter {
+            let instr = instr?;
             let op_name = instr.op.name();
             let (all_cbits, vars): (Vec<Clbit>, Option<Vec<PyObject>>) = {
                 if self.may_have_additional_wires(py, &instr) {
@@ -6571,8 +6656,8 @@ impl DAGCircuit {
 
         new_dag.metadata = qc.metadata.map(|meta| meta.unbind());
 
-        // Add the qubits depending on order.
-        let qubit_map: Option<Vec<Qubit>> = if let Some(qubit_ordering) = qubit_order {
+        // Add the qubits depending on order, and produce the qargs map.
+        let qarg_map = if let Some(qubit_ordering) = qubit_order {
             let mut ordered_vec = Vec::from_iter((0..num_qubits as u32).map(Qubit));
             qubit_ordering
                 .into_iter()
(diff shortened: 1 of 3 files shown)


crates/circuit/src/interner.rs
```diff
@@ -150,6 +295,75 @@ where
...
+    pub fn try_key(&self, value: &T) -> Option<Interned<T>> {
+        self.0.get_index_of(value).map(|index| Interned {
+            index: index as u32,
+            _type: PhantomData,
+        })
+    }
+
+    /// Return whether this value is already in the [Interner].
+    ///
+    /// Typically you want to use [try_get], which returns the key if present, or [insert], which
```
kevinhartman:
```suggestion
    /// Typically you want to use [try_key], which returns the key if present, or [insert], which
```

crates/circuit/src/interner.rs
```diff
@@ -181,26 +414,138 @@ where
...
+        let mut work = SmallVec::<[T; N]>::with_capacity(N);
+        target.map.clear();
+        target.map.reserve(other.0.len());
+        for slice in other.0.iter() {
+            let new_slice = 'slice: {
+                work.clear();
+                work.reserve(slice.len());
+                for value in slice {
+                    let Some(scalar) = scalar_map_fn(value) else {
+                        break 'slice None;
```
kevinhartman, line 508: Cool!

crates/circuit/src/interner.rs
```diff
@@ -181,26 +414,138 @@ where
...
+        mut scalar_map_fn: impl FnMut(&T) -> Option<T>,
+        target: &mut InternedMap<[T]>,
+    ) {
+        // Workspace for the mapping function. The aim here is that we're working on the stack, so
+        // the mapping doesn't need to make heap allocations.  We could either guess (which the
+        // higher-level `merge_slice_map` does), or force the caller to tell us how much stack space
+        // to allocate.  This method is lower leveel, so in this case we ask them to tell us; if
+        // they're optimising to the point of re-using the return allocations, they probably have a
+        // good idea about the maximum slice size of the interner they'll be merging in.
+        let mut work = SmallVec::<[T; N]>::with_capacity(N);
```
kevinhartman, line 499: This is very neat. From what I can tell, we only do a heap allocation if every scalar maps and the slice doesn't already exist? And the latter is possible because of `Equivalent` working for `Vec<T>` and `&[T]`.

crates/circuit/src/interner.rs
```diff
@@ -181,26 +414,138 @@ where
...
+    pub fn merge_map_slice(
+        &mut self,
+        other: &Self,
+        scalar_map_fn: impl FnMut(&T) -> Option<T>,
+    ) -> InternedMap<[T]> {
+        let mut out = InternedMap::new();
+        // We're specifying the stack space here.  This is just a guess, but it's not hugely
+        // important; we'll safely spill from the stack to the heap if needed, and this function is
+        // an API convenience at the cost of optimal allocation performance anyway.
+        self.merge_map_slice_using::<4>(other, scalar_map_fn, &mut out);
```
kevinhartman, line 540: Since this is stack space, are you choosing a low count like `4` for cache locality concerns?

crates/circuit/src/dag_circuit.rs
```diff
@@ -6587,7 +6672,11 @@ impl DAGCircuit {
                     ordered_vec[qubit_index.index()] = new_dag.add_qubit_unchecked(py, &qubit)?;
                     Ok(())
                 })?;
-            Some(ordered_vec)
+            // The `Vec::get` use is because an arbitrary interner might contain old references to
+            // bit instances beyond `num_qubits`, such as if it's from a DAG that had wires removed.
+            new_dag.merge_qargs(qc_data.qargs_interner(), |bit| {
```
kevinhartman, line 6677: At this point, the `new_dag` does not yet contain any entries (other than the default), unless I'm mistaken. So it's less of a "merge" here and more of an efficient clone that skips over now-dead keys in the source interner, correct?

Separately, I just noticed that `DAGCircuit::with_capacity` preallocates its interners to have `num_qubits` and `num_clbits` entries. I'm not sure if that was an intentional choice, or due to a misread of the docstring when `DAGCircuit::with_capacity` was added.

kevinhartman (commented):
This one definitely seems like a slam dunk to me 😄.

Most of my comments are just for clarification.

The `FromIterator` implementation for `Interner<T>` is interesting. If it were implemented for `T` instead, would there really be any downside to that? It seems like implementing for `::Owned` would mean equivalent values would still need to be duplicated by the caller.

crates/circuit/src/interner.rs
jakelishman: We do heap allocations in two situations:
- we have to insert into the interner (the situation you described) - this `work` thing here is a tacit assumption that that happens relatively rarely, though the cost of being wrong about that is very low in practice - it's a small memcpy.
- we need more space for a slice than we allocated on the stack (so the `SmallVec` spills to the heap)

The `Vec` and the slice are `Equivalent`, but in this case, they also support `PartialEq` between borrows of each other, so it'd work in the default hashmaps without `Equivalent` too, I think.

crates/circuit/src/interner.rs
jakelishman: Yeah, sort of, though I didn't really put much thought into the exact value - it's unlikely to matter in this method, because of the cost of allocating the map anyway.

crates/circuit/src/dag_circuit.rs
jakelishman: Yeah exactly, this particular merge isn't so much a merge as an efficient mapping.  In other situations we might want it (like `substitute_node_with_dag`), it's much more of an actual merge-with-map.

crates/circuit/src/interner.rs
jakelishman: It's a feature to be careful of (feels like it could easily get spaghetti-ish), but when it works, it can _really_ tidy up code.  I'd originally tried to write this as a closure in order to use `return` in the same way (via `?`), but it ended up really gross and I swapped to this.

crates/circuit/src/interner.rs
jakelishman: Thanks yeah - I went through a bunch of different names for things.

jakelishman:
> The `FromIterator` implementation for `Interner<T>` is interesting. If it were implemented for `T` instead, would there really be any downside to that? It seems like implementing for `::Owned` would mean equivalent values would still need to be duplicated by the caller.

In general (and in almost all the cases we care about), `T` isn't `Sized` (it's usually `[T]`), so you can't hold direct values of it - you've got to indirect somehow.  So then we have to choose either the borrowed type (e.g. `&[T]`) or the owned (e.g. `Vec<T>`), or something that can `Into<<T as ToOwned>::Owned>`.  The `FromIterator` impl takes ownership of the values coming in, so _we_ don't need to clone the values in the way I wrote it (we just move them into the `IndexSet`), but the caller of `FromIterator` might in order to produce the iterator.

In practice, I mostly wrote that `impl` to make the tests shorter to write, and mostly I thought that the chances were that if you wanted to use that method, you'd have the owned types anyway.

raynelfss (commented):
This is a very neat idea that should hopefully make conversion between circuits much easier. I will try to do a more in depth review of this tomorrow. But what I've seen so far looks great!

crates/circuit/src/interner.rs
```diff
@@ -43,6 +44,128 @@ impl<T: ?Sized> Copy for Interned<T> {}
...
+/// This is created by the interner-merging functions like [Interner::merge_map] and
+/// [Interner::merge_slice_map].
+///
+/// This map can be indexed by the [Interned] keys of the smaller [Interner], and returns [Interned]
+/// keys that work on the larger [Interner] (the one that expanded itself).
+///
+/// The indexing implementation panics if asked for the new key for an object that was filtered out
+/// during the merge.
+#[derive(Clone, Debug, Default)]
+pub struct InternedMap<S: ?Sized, T: ?Sized = S> {
```
raynelfss, line 58: Is there a reason why we require two different generics for this? I was assuming that mapping was only possible between interners of the same type.

crates/circuit/src/interner.rs
```diff
@@ -150,6 +295,75 @@ where
...
+        self.0.iter().enumerate().map(|(i, v)| {
+            (
+                Interned {
+                    index: i as u32,
+                    _type: PhantomData,
+                },
+                v.borrow(),
+            )
+        })
+    }
```
raynelfss, line 335: Thank you for adding these by the way. These are very useful for some of the things I have been working on :)

crates/circuit/src/interner.rs
```diff
@@ -181,26 +414,138 @@ where
...
+        // you want to set `N`, which is just an API annoyance since we'll probably never need the
+        // two interners to be different types.
+        other: &Self,
+        mut scalar_map_fn: impl FnMut(&T) -> Option<T>,
+        target: &mut InternedMap<[T]>,
+    ) {
+        // Workspace for the mapping function. The aim here is that we're working on the stack, so
+        // the mapping doesn't need to make heap allocations.  We could either guess (which the
+        // higher-level `merge_slice_map` does), or force the caller to tell us how much stack space
+        // to allocate.  This method is lower leveel, so in this case we ask them to tell us; if
```
raynelfss:
```suggestion
        // to allocate.  This method is lower level, so in this case we ask them to tell us; if
```

crates/circuit/src/interner.rs
```diff
@@ -181,26 +414,138 @@ where
...
+        // they're optimising to the point of re-using the return allocations, they probably have a
```
raynelfss: Hate to be doing this but...
```suggestion
        // they're optimizing to the point of re-using the return allocations, they probably have a
```
🇺🇸

crates/circuit/src/dag_circuit.rs
```diff
@@ -6383,10 +6453,24 @@ impl DAGCircuit {
...
+        self.try_extend(
+            py,
+            iter.into_iter()
+                .map(|inst| -> Result<PackedInstruction, Infallible> { Ok(inst) }),
+        )
+    }
+
+    /// Extends the DAG with valid instances of [PackedInstruction], where the iterator produces the
+    /// results in a fallible manner.
+    pub fn try_extend<I, E>(&mut self, py: Python, iter: I) -> PyResult<Vec<NodeIndex>>
```
raynelfss, line 6470: Thank you for separating the extend method to be able to throw its own type of `Error` result. Do you think this is a pattern that we should be following with fallible code from now on?

raynelfss (commented):
After a quick passthrough of the code, this looks like a very useful addition that I can appreciate. Other than that, the additions to the Interner methods are fairly welcome as these were issues I was running into myself, and had to, in come cases, add them too. I like everything being done here. LGTM!

I left a couple of questions and comments, would suggest that we also wait for @kevinhartman's in depth review.

kevinhartman:
> would suggest that we also wait for @kevinhartman's in depth review

@raynelfss I've already reviewed to a suitable depth. Feel free to sign off when your comments are addressed 🙂.

crates/circuit/src/interner.rs
jakelishman: No worries - I'd seen you were using something like them, and I needed them to write test cases too.

crates/circuit/src/dag_circuit.rs
jakelishman: It depends - the downside this form that I've moved it into is that the DAG gets partially constructed before a potential failure, so it's not easily possible to recover from.  The upside (and why I did it) is that there's no longer a need to collect into an intermediate `Vec` before extending.

I thought that it made sense to do here, because whenever Python's involved, we typically can't _ever_ be certain that things will go off infallibly, and we already have lots of DAG mutation methods that leave the DAG in a partially mutated state if an exception occurs.

crates/circuit/src/interner.rs
jakelishman: We don't require two, but the mapping can in general be between two different types, providing the mapping function has the signature `&S -> T`.  I think there's a test case of it.

I didn't make the specialised mapping method for slice-based `Interner`s doubly generic in the same way, because those take a `const N: usize` generic argument, which you have to specify by hand already, so I didn't want to make them more awkward by then requiring you to spell the call as `merge_map_slice_using::<N, _>(...)` instead of just `merge_map_slice_using::<N>`.

jakelishman:
> Hate to be doing this but...

you'll be here forever if you check the spelling in my code comments haha - I try and write public documentation with American spelling, but I write comments straight from my head.

(I don't mind changing it to American, I just won't spot it myself easily.)

raynelfss (approved):
LGTM! Thank you for adding these changes and making the Interner much more flexible. It will be a helpful addition to #13686 and others.

PR #13753, opened 2025-01-28
Make sure that 'tox -erust' fails on bad RC.

kevinhartman:
### Summary
Previously, a non-zero exit code from the subprocess would not fail the tox run like we'd expect.

### DIFF
--- a/tools/run_cargo_test.py
+++ b/tools/run_cargo_test.py
@@ -1,0 +1,40 @@
+#!/usr/bin/env python3
+
+# This code is part of Qiskit.
+#
+# (C) Copyright IBM 2025.
+#
+# This code is licensed under the Apache License, Version 2.0. You may
+# obtain a copy of this license in the LICENSE.txt file in the root directory
+# of this source tree or at http://www.apache.org/licenses/LICENSE-2.0.
+#
+# Any modifications or derivative works of this code must retain this
+# copyright notice, and modified files need to carry a notice indicating
+# that they have been altered from the originals.
+
+"""
+Utility script to invoke cargo test within the current Python environment.
+
+Notably, this sets up the environment variables necessary for Qiskit to be
+found by PyO3 / our Rust test executable.
+"""
+
+import os
+import subprocess
+import sys
+import sysconfig
+
+# This allows the Python interpreter baked into our test executable to find the
+# Qiskit installed in the active environment.
+os.environ["PYTHONUSERBASE"] = sys.prefix
+
+# On Linux, the test executable's RPATH doesn't contain libpython, so we add it
+# to the dlopen search path here.
+os.environ["LD_LIBRARY_PATH"] = os.pathsep.join(
+    filter(None, [sysconfig.get_config_var("LIBDIR"), os.getenv("LD_LIBRARY_PATH")])
+)
+
+# The '--no-default-features' flag is used here to disable PyO3's
+# 'extension-module' when running the tests (which would otherwise cause link
+# errors).
+subprocess.run(["cargo", "test", "--no-default-features"], check=True)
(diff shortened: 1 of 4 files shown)


tox.ini
```diff
@@ -41,7 +41,7 @@ commands =
   python -c '\
   import os, subprocess, sysconfig;\
   os.environ["LD_LIBRARY_PATH"] = os.pathsep.join(filter(None, [sysconfig.get_config_var("LIBDIR"), os.getenv("LD_LIBRARY_PATH")]));\
-  subprocess.run(["cargo", "test", "--no-default-features"])'
+  subprocess.run(["cargo", "test", "--no-default-features"]).check_returncode()'
```
mtreinish: Not that it particularly matters in this context. But what I normally do is:

```suggestion
  subprocess.run(["cargo", "test", "--no-default-features"], check=True)'
```

mtreinish (dismissed):
LGTM, just left a small comment inline. It doesn't really matter in this context though.

tox.ini
jakelishman: I think pylint has a lint against failing to check the return code of `subprocess.run` - it might be a motivation to move this inline script into a proper file in the `tools` directory so it's subject to lint?

jakelishman: It does: https://pylint.readthedocs.io/en/latest/user_guide/messages/warning/subprocess-run-check.html

kevinhartman: Seems like a good idea, and should let us use the same script from tox and CI.

Done in 663eaa8.

tools/run_cargo_test.py
```diff
@@ -0,0 +1,40 @@
...
+
+# This allows the Python interpreter baked into our test executable to find the
+# Qiskit installed in the active environment.
+os.environ["PYTHONUSERBASE"] = sys.prefix
+
+# On Linux, the test executable's RPATH doesn't contain libpython, so we add it
+# to the dlopen search path here.
+os.environ["LD_LIBRARY_PATH"] = os.pathsep.join(
+    filter(None, [sysconfig.get_config_var("LIBDIR"), os.getenv("LD_LIBRARY_PATH")])
+)
```
mtreinish, line 38: Is there a reason to not just do?:
```suggestion
os.environ["LD_LIBRARY_PATH"] = os.pathsep.join([
    sysconfig.get_config_var("LIBDIR"),
    os.getenv("LD_LIBRARY_PATH")
])
```

I've always found the `filter(None, ...)` pattern a bit odd when you can just use a generator expression.

mtreinish (commented):
This LGTM for the most part. I like having the script in a dedicated file now. Just one question/suggestion inline.

tools/run_cargo_test.py
kevinhartman: The reason I did that was to filter out `None` in the case that `"LD_LIBRARY_PATH"` isn't actually set to anything. I can do something different if you prefer!

kevinhartman:
While testing this on his machine, @raynelfss discovered an issue with using the existing approach with `PYTHONUSERBASE` to get the virtual environment's `site-packages` into the embedded Python interpreter's path. The issue was that `venvs` use a potentially different folder layout when installing packages than the system's Python will use when looking for them, so the path returned by [site.USER_SITE](https://docs.python.org/3/library/site.html#site.USER_SITE) was incorrect.

Instead, we now set the `PYTHONPATH` to directly be the actual full path to the active Python interpreter's `site-packages`, which is more natural anyways. This wasn't done originally because before this PR, we weren't actually invoking `cargo test` from within a Python script.

Anyway, this should be good to go now.

jakelishman (approved):
Given that Matt's question was minor, and from offline conversations, this is important for people, let's merge now.  If there's more to tweak, we can follow up.

PR #13754, opened 2025-01-29
Use ``Arc::ptr_eq`` over ``.is()`` in ``SparseObservable``

Cryoris:
### Summary

Follow up on #13595: a commit got lost in the final merge, this PR adds it again 🙂 

### Details

We should compare the inner objects of `PySparseObservable`,  not the observable itself, since in theory multiple `PySparseObservable`s could point to the same inner `SparseObservable`.


jakelishman (approved):
I think this is right, and that it's just the missing bits from the merged I bust up on the main merge, thanks.

If there's anything else lurking with the new form, we'll catch/fix it when we add more use of these methods.

PR #13758, opened 2025-01-30
Support `SparseObservable` to `SparsePauliOp` conversions

Cryoris:
### Summary

Add `SparseObservable.to_sparse_list` to generate a sparse Pauli list representation, which is used to implement `SparsePauliOp.from_sparse_observable`. This method will also allow using `SparseObservable`s in a `PauliEvolutionGate` (follow up PR).

### Details and comments

In as many places as I could, I added a warning that this conversion will blow up if one attempts to convert a large number of projectors to a Pauli representation.

### DIFF
--- a/crates/accelerate/src/sparse_observable.rs
+++ b/crates/accelerate/src/sparse_observable.rs
@@ -11,6 +11,7 @@
 // that they have been altered from the originals.
 
 use hashbrown::HashSet;
+use itertools::Itertools;
 use ndarray::Array2;
 use num_complex::Complex64;
 use num_traits::Zero;
@@ -23,7 +24,7 @@ use pyo3::{
     intern,
     prelude::*,
     sync::GILOnceCell,
-    types::{IntoPyDict, PyList, PyTuple, PyType},
+    types::{IntoPyDict, PyList, PyString, PyTuple, PyType},
     IntoPyObjectExt, PyErr,
 };
 use std::{
@@ -183,6 +184,24 @@ impl BitTerm {
     /// This is true for the operators and eigenspace projectors associated with Y and Z.
     pub fn has_z_component(&self) -> bool {
         ((*self as u8) & (Self::Z as u8)) != 0
+    }
+
+    pub fn is_projector(&self) -> bool {
+        !matches!(self, BitTerm::X | BitTerm::Y | BitTerm::Z)
+    }
+}
+
+fn bit_term_as_pauli(bit: &BitTerm) -> Vec<(bool, Option<BitTerm>)> {
+    match bit {
+        BitTerm::X => vec![(true, Some(BitTerm::X))],
+        BitTerm::Y => vec![(true, Some(BitTerm::Y))],
+        BitTerm::Z => vec![(true, Some(BitTerm::Z))],
+        BitTerm::Plus => vec![(true, None), (true, Some(BitTerm::X))],
+        BitTerm::Minus => vec![(true, None), (false, Some(BitTerm::X))],
+        BitTerm::Left => vec![(true, None), (true, Some(BitTerm::Y))],
+        BitTerm::Right => vec![(true, None), (false, Some(BitTerm::Y))],
+        BitTerm::Zero => vec![(true, None), (true, Some(BitTerm::Z))],
+        BitTerm::One => vec![(true, None), (false, Some(BitTerm::Z))],
     }
 }
 
@@ -639,6 +658,46 @@ impl SparseObservable {
             bit_terms: &self.bit_terms[start..end],
             indices: &self.indices[start..end],
         }
+    }
+
+    /// Get a sparse Pauli representation of the observable.
+    ///
+    /// This returns an iterator over the terms in form of tuples ``(paulis, indices, coeff)``,
+    /// where ``paulis`` is a vector of **Pauli** [BitTerm]s X, Y, Z.
+    ///
+    /// # Warning
+    ///
+    /// This representation is highly inefficient for projectors. For example, a term with
+    /// :math:`n` projectors :math:`|+\rangle\langle +|` will use :math:`2^n` Pauli terms.
+    pub fn to_paulis(&self) -> impl Iterator<Item = (Vec<BitTerm>, Vec<u32>, Complex64)> + '_ {
+        self.iter().flat_map(|view| {
+            let num_projectors = view
+                .bit_terms
+                .iter()
+                .filter(|&bit| bit.is_projector())
+                .count();
+            let div = 2_f64.powi(num_projectors as i32);
+
+            view.bit_terms
+                .iter()
+                .map(bit_term_as_pauli)
+                .multi_cartesian_product()
+                .map(move |combination| {
+                    let mut positive = true;
+                    let mut paulis: Vec<BitTerm> = Vec::new();
+                    let mut indices: Vec<u32> = Vec::new();
+
+                    for (index, (sign, bit)) in combination.iter().enumerate() {
+                        positive &= sign;
+                        if let Some(bit) = bit {
+                            paulis.push(*bit);
+                            indices.push(view.indices[index]);
+                        }
+                    }
+                    let coeff = if positive { view.coeff } else { -view.coeff };
+                    (paulis, indices, coeff / div)
+                })
+        })
     }
 
     /// Add the term implied by a dense string label onto this observable.
@@ -2335,6 +2394,29 @@ impl PySparseObservable {
         }
         let inner = SparseObservable::new(num_qubits, coeffs, bit_terms, indices, boundaries)?;
         Ok(inner.into())
+    }
+
+    #[pyo3(signature = ())]
+    fn to_sparse_list(&self, py: Python) -> PyResult<Py<PyList>> {
+        let inner = self.inner.read().map_err(|_| InnerReadError)?;
+
+        let sparse_list = inner
+            .to_paulis()
+            .map(move |(bits, indices, coeff)| {
+                let mut pauli_string = String::new();
+                for bit in bits {
+                    pauli_string.push_str(bit.py_label());
+                }
+                let py_string = PyString::new(py, &pauli_string).unbind();
+                let py_indices = PyList::new(py, indices)?.unbind();
+                let py_coeff = coeff.into_py_any(py)?;
+
+                PyTuple::new(py, vec![py_string.as_any(), py_indices.as_any(), &py_coeff])
+            })
+            .collect::<PyResult<Vec<_>>>()?;
+
+        let out = PyList::new(py, sparse_list)?;
+        Ok(out.unbind())
     }
 
     /// Construct a :class:`.SparseObservable` from a :class:`.SparsePauliOp` instance.
--- a/crates/accelerate/src/sparse_observable.rs
+++ b/crates/accelerate/src/sparse_observable.rs
@@ -2396,24 +2396,50 @@ impl PySparseObservable {
         Ok(inner.into())
     }
 
-    #[pyo3(signature = ())]
-    fn to_sparse_list(&self, py: Python) -> PyResult<Py<PyList>> {
+    /// Express the observable in terms of a sparse list format.
+    ///
+    /// This is the counter-operation of :meth:`.SparseObservable.from_sparse_list`.
+    ///
+    /// Args:
+    ///     only_pauli: If ``True``, express the observable only in terms of non-identity Paulis,
+    ///         :math:`X`, :math:`Y`, and :math:`Z`. Beware that this will use at least :math:`2^n`
+    ///         terms if there are :math:`n` single-qubit projectors present, which can lead
+    ///         to an exponentially expensive representation. Defaults to ``False``.
+    ///
+    /// Examples:
+    ///     
+    ///     >>> obs = SparseObservable.from_list([("IIXIZ", 2j), ("IIZIX", 2j)])
+    ///     >>> reconstructed = SparseObservable.from_sparse_list(obs.to_sparse_list(), obs.num_qubits)
+    ///     >>> assert obs == reconstructed
+    #[pyo3(signature = (only_paulis=false))]
+    fn to_sparse_list(&self, py: Python, only_paulis: bool) -> PyResult<Py<PyList>> {
         let inner = self.inner.read().map_err(|_| InnerReadError)?;
 
-        let sparse_list = inner
-            .to_paulis()
-            .map(move |(bits, indices, coeff)| {
-                let mut pauli_string = String::new();
-                for bit in bits {
-                    pauli_string.push_str(bit.py_label());
-                }
-                let py_string = PyString::new(py, &pauli_string).unbind();
-                let py_indices = PyList::new(py, indices)?.unbind();
-                let py_coeff = coeff.into_py_any(py)?;
+        // turn a 3-tuple of (bit terms, indices, coeff) into a Python tuple
+        let to_py_tuple = |bits: &[BitTerm], indices: &[u32], coeff: Complex64| {
+            let mut pauli_string = String::new();
+            for bit in bits {
+                pauli_string.push_str(bit.py_label());
+            }
+            let py_string = PyString::new(py, &pauli_string).unbind();
+            let py_indices = PyList::new(py, indices)?.unbind();
+            let py_coeff = coeff.into_py_any(py)?;
 
-                PyTuple::new(py, vec![py_string.as_any(), py_indices.as_any(), &py_coeff])
-            })
-            .collect::<PyResult<Vec<_>>>()?;
+            PyTuple::new(py, vec![py_string.as_any(), py_indices.as_any(), &py_coeff])
+        };
+
+        // to map onto a Pauli list, we first have to expand all projectors, otherwise
+        // we can just directly iterate over the view
+        let sparse_list = match only_paulis {
+            false => inner
+                .iter()
+                .map(|view| to_py_tuple(view.bit_terms, view.indices, view.coeff))
+                .collect::<PyResult<Vec<_>>>()?,
+            true => inner
+                .to_paulis()
+                .map(move |(bits, indices, coeff)| to_py_tuple(&bits, &indices, coeff))
+                .collect::<PyResult<Vec<_>>>()?,
+        };
 
         let out = PyList::new(py, sparse_list)?;
         Ok(out.unbind())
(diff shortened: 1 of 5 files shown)


crates/accelerate/src/sparse_observable.rs
```diff
@@ -641,6 +660,46 @@ impl SparseObservable {
...
+    ///
+    /// This returns an iterator over the terms in form of tuples ``(paulis, indices, coeff)``,
+    /// where ``paulis`` is a vector of **Pauli** [BitTerm]s X, Y, Z.
+    ///
+    /// # Warning
+    ///
+    /// This representation is highly inefficient for projectors. For example, a term with
+    /// :math:`n` projectors :math:`|+\rangle\langle +|` will use :math:`2^n` Pauli terms.
+    pub fn to_paulis(&self) -> impl Iterator<Item = (Vec<BitTerm>, Vec<u32>, Complex64)> + '_ {
+        self.iter().flat_map(|view| {
```
jakelishman: How would you feel about having the `to_paulis` method return another `SparseObservable`, just with the `BitTerm`s rewritten in the way you have now?  We could expose it to Python in that form.

I'm just thinking: this could simplify the `to_sparse_list` method, because then we wouldn't need the `only_paulis` option at all: it's the composition `obs.to_paulis().to_sparse_list()`.

The `to_paulis()` method would also have a secondary use: expensive as it is, `to_paulis().simplify()` produces a unique canonical form for any given `SparseObservable`; everything representing the same observable has the same form under that method, because the Paulis are a precise basis, rather than overcomplete.

qiskit/quantum_info/operators/symplectic/sparse_pauli_op.py
```diff
@@ -929,6 +931,45 @@ def from_sparse_list(
...
+            num_projectors = sum(bit not in [bterm.X, bterm.Y, bterm.Z] for bit in obs.bit_terms)
+
+            # 16 is an arbitrary threshold, but at this point please use a SparseObservable
+            if num_projectors > 16:
+                warnings.warn(
+                    f"SparseObservable contains a lot of projectors ({num_projectors}), which will "
+                    f"lead to at least {2 ** num_projectors} terms.",
+                    stacklevel=2,
+                    category=RuntimeWarning,
+                )
```
jakelishman: The actual runtime warnings might be a shade too far, maybe - I think the `.. warning::` note you've got in the documentation is probably enough.

test/python/quantum_info/operators/symplectic/test_sparse_pauli_op.py
```diff
@@ -361,6 +367,36 @@ def test_to_list_parameters(self):
...
 
+    def test_from_sparse_observable(self):
+        """Test from a SparseObservable."""
+        obs = SparseObservable.zero(10)
+        expected = SparsePauliOp(["I" * 10], coeffs=[0])
+        self.assertEqual(expected, SparsePauliOp.from_sparse_observable(obs))
+
+        obs = SparseObservable("XrZ")
+        expected = SparsePauliOp(["XIZ", "XYZ"], coeffs=[0.5, -0.5])
+        self.assertEqual(expected, SparsePauliOp.from_sparse_observable(obs))
```
jakelishman: This test is simple enough to be ok, but just note that technically it's potentially coupling to an internal detail of `SparsePauliOp.from_sparse_observable`: `SparsePauliOp(["XYZ", "XIZ"], coeffs=[-0.5, 0.5]` would also be a correct answer, but the test would reject it.

We could say:
- `SparseObservable.to_paulis()` will be in fully sorted canonical order.  This could be quite a lot of additional overhead.
- `SparseObservable.to_paulis()` will always expand each individual projector `BitTerm` in canonical order (sorts to IXYZ, etc), in which case the order is completely defined, but not fully canonical.
- there is no defined order from `SparseObservable.to_paulis()`, in which case this test should include a canonicalisation step after `SparsePauliOp.from_sparse_observable(obs)` to isolate the test (and the others in this file) from changes in the order.

We ought to pick one of those - do you have a preference?

test/python/quantum_info/test_sparse_observable.py
```diff
@@ -2008,3 +2008,71 @@ def test_term_pauli_base(self):
...
+        """Test converting to a sparse list."""
+        obs = SparseObservable.zero(100)
+        with self.subTest(msg="zero"):
+            self.assertEqual(0, len(obs.to_sparse_list()))
+
+        obs = SparseObservable.identity(100)
+        sparse_list = obs.to_sparse_list()
+        with self.subTest(msg="identity"):
+            self.assertEqual(1, len(sparse_list))
+            self.assertEqual("", sparse_list[0][0])
```
jakelishman: Can we write these equality tests (etc) using the actual expected lists - that'll test length and full correctness in a single step, and give better error messages.

test/python/quantum_info/test_sparse_observable.py
```diff
@@ -2008,3 +2008,71 @@ def test_term_pauli_base(self):
...
+
+        obs = SparseObservable("IXYZ")
+        sparse_list = obs.to_sparse_list()
+        with self.subTest(msg="IXYZ"):
+            self.assertEqual(1, len(sparse_list))
+            self.assertEqual("ZYX", sparse_list[0][0])
+            self.assertListEqual([0, 1, 2], sparse_list[0][1])
```
jakelishman: You typically don't need/want to call `assertListEqual` (just `assertEqual`).  If `sparse_list[0][1]` is some list-like type and not necessarily _precisely_ a list, then `self.assertEqual([0, 1, 2], list(x))` is fine.

test/python/quantum_info/operators/symplectic/test_sparse_pauli_op.py
```diff
@@ -361,6 +367,36 @@ def test_to_list_parameters(self):
         target = list(zip(labels, coeffs))
         self.assertEqual(op.to_list(), target)
 
+    def test_from_sparse_observable(self):
+        """Test from a SparseObservable."""
+        obs = SparseObservable.zero(10)
```
jakelishman: Can we also add a test of the zero-qubit zero operator and zero-qubit identity as edge cases?

jakelishman (commented):
Thanks for doing this - one main question about the exposed interfaces, and a couple of minor bits on testing.

crates/accelerate/src/sparse_observable.rs
Cryoris: Yeah I also considered that, the downside is that we need to take a detour via `SparsePauliOp` if we want the sparse list for the Pauli evolution. So we'd always have to go via the "dense" format where we store all identities explicitly, which would be nice to avoid 🙂 

The alternative would be to handle projectors in the Pauli evolution directly, but from our offline discussion yesterday I'm not sure if we're overcomplicating things then 🤔

qiskit/quantum_info/operators/symplectic/sparse_pauli_op.py
Cryoris: Sounds good -- I just really wanted to make sure people don't do this 😛

test/python/quantum_info/operators/symplectic/test_sparse_pauli_op.py
Cryoris: I would have a preference for the last option -- if there's no clear practical benefit in guaranteeing a format then I'd rather not commit to one 🙂

crates/accelerate/src/sparse_observable.rs
jakelishman: I mean having `to_paulis` return `SparseObservable`, not `SparsePauliOp` - still qubit-sparse, but with a promise that we only use alphabet letters corresponding to Pauli ops?

test/python/quantum_info/operators/symplectic/test_sparse_pauli_op.py
jakelishman: Fine by me - in that case let's update the tests to decouple them.

crates/accelerate/src/sparse_observable.rs
jakelishman: Not sure if you misunderstood me before, or if I misunderstood you.  Returning `SparseObservable` from this method is actually slightly better in terms of memory use than the collection into a list/vec of this output, and I'm not super concerned about the intermediate memory usage of the `SparseObservable` object here, because the usage of the resulting `SparsePauliOp` is already going to be about the same (depends on how dense the terms are).

Cryoris: Ah! I misunderstood you! Oh yeah having `to_paulis` return a `SparseObservable` (not `SparsePauliOp`) would be an elegant solution indeed. I'll update that 🙂

test/python/quantum_info/operators/symplectic/test_sparse_pauli_op.py
Cryoris: Added in 358575b

test/python/quantum_info/test_sparse_observable.py
Cryoris: Done in 358575b

test/python/quantum_info/test_sparse_observable.py
Cryoris: Done in 358575b

test/python/quantum_info/operators/symplectic/test_sparse_pauli_op.py
Cryoris: done in 358575b

qiskit/quantum_info/operators/symplectic/sparse_pauli_op.py
Cryoris: Changed in 358575b 👍🏻

crates/accelerate/src/sparse_observable.rs
Cryoris: Changed in 358575b, seems much nicer now 🙂

crates/accelerate/src/sparse_observable.rs
```diff
@@ -641,6 +660,52 @@ impl SparseObservable {
...
+                }
+                boundaries.push(paulis.len());
+
+                let coeff = if positive { view.coeff } else { -view.coeff };
+                coeffs.push(coeff / div)
+            }
+        }
+
+        SparseObservable::new(self.num_qubits, coeffs, paulis, indices, boundaries)
+    }
```
jakelishman: This function should never return a `CoherenceError` - the input `SparseObservable` is guaranteed valid, and all valid `SparseObservable`s have a valid representation in the pure-Pauli-operator basis.  We can construct the output `SparseObservable` directly using the record-struct constructor (`SparseObservable {num_qubits: self.num_qubits, ..}`) or using `SparseObservable::new_unchecked` with a mild preference on my side for the record-struct constructor.

crates/accelerate/src/sparse_observable.rs
```diff
@@ -184,6 +185,24 @@ impl BitTerm {
...
+        BitTerm::Y => vec![(true, Some(BitTerm::Y))],
+        BitTerm::Z => vec![(true, Some(BitTerm::Z))],
+        BitTerm::Plus => vec![(true, None), (true, Some(BitTerm::X))],
+        BitTerm::Minus => vec![(true, None), (false, Some(BitTerm::X))],
+        BitTerm::Left => vec![(true, None), (true, Some(BitTerm::Y))],
+        BitTerm::Right => vec![(true, None), (false, Some(BitTerm::Y))],
+        BitTerm::Zero => vec![(true, None), (true, Some(BitTerm::Z))],
+        BitTerm::One => vec![(true, None), (false, Some(BitTerm::Z))],
+    }
 }
```
jakelishman: Rather unimportant, but these could all return `&'static [(bool, Option<BitTerm>)]`, rather than eagerly allocating a `Vec`.

crates/accelerate/src/sparse_observable.rs
```diff
@@ -2337,6 +2402,63 @@ impl PySparseObservable {
...
+    fn to_paulis(&self) -> PyResult<Self> {
+        let inner = self.inner.read().map_err(|_| InnerReadError)?;
+        Ok(inner.to_paulis()?.into())
+    }
+
+    /// Express the observable in terms of a sparse list format.
+    ///
+    /// This can be seen as counter-operation of :meth:`.SparseObservable.from_sparse_list`, however
+    /// the order of terms is not guaranteed to be the same at after a roundtrip to a sparse
+    /// list and back. Use :meth:`SparseObservable.simplify` to obtain a canonical representation.
```
jakelishman: I think this note about `SparseObservable.simplify` actually belongs in the `to_paulis` method above - the recipe for a completely canonical representation of a `SparseObservable` is `obs.to_paulis().simplify()`, nothing to do with `to_sparse_list`.

crates/accelerate/src/sparse_observable.rs
```diff
@@ -2337,6 +2402,63 @@ impl PySparseObservable {
...
+            PyTuple::new(py, vec![py_string.as_any(), py_indices.as_any(), &py_coeff])
+        };
+
+        let sparse_list = inner
+            .iter()
+            .map(|view| to_py_tuple(view.bit_terms, view.indices, view.coeff))
+            .collect::<PyResult<Vec<_>>>()?;
+
+        let out = PyList::new(py, sparse_list)?;
+        Ok(out.unbind())
```
jakelishman: Two points here, both pretty minor:

1. why deconstruct the `SparseView` _before_ `to_py_tuple` and not just accept that in the closure?
2. it'd probably be a bit more efficient to construct a `let out = PyList::empty(py)?;` and do `for term in inner.iter(): out.append(to_py_tuple(term)?)?;` than to collect into a `Vec` and then convert that into a list

qiskit/quantum_info/operators/symplectic/sparse_pauli_op.py
```diff
@@ -929,6 +930,27 @@ def from_sparse_list(
         paulis = PauliList(labels)
         return SparsePauliOp(paulis, coeffs, copy=False)
 
+    @staticmethod
+    def from_sparse_observable(obs: SparseObservable) -> SparsePauliOp:
```
jakelishman, line 934: Do you want to add a variant to the `__init__` method that handles this case too?  `quantum_info` types (for better or worse) check everything under the sun in their default constructors.

crates/accelerate/src/sparse_observable.rs
```diff
@@ -2337,6 +2402,63 @@ impl PySparseObservable {
...
+
+        // turn a 3-tuple of (bit terms, indices, coeff) into a Python tuple
+        let to_py_tuple = |bits: &[BitTerm], indices: &[u32], coeff: Complex64| {
+            let mut pauli_string = String::new();
+            for bit in bits {
+                pauli_string.push_str(bit.py_label());
+            }
+            let py_string = PyString::new(py, &pauli_string).unbind();
+            let py_indices = PyList::new(py, indices)?.unbind();
+            let py_coeff = coeff.into_py_any(py)?;
```
jakelishman: What do you think about reversing the bit terms and the indices on output, so the Pauli string comes out in "reading" order?

Minor, but the necessary capacity of `pauli_string` is known immediately; all the `py_label` return values are (necessarily) a single ASCII letter, so `pauli_string` needs a capacity of `bits.len()`, and we could use `String::with_capacity` on it.

crates/accelerate/src/sparse_observable.rs
```diff
@@ -2337,6 +2402,63 @@ impl PySparseObservable {
...
+    ///
+    /// Returns:
+    ///     The same observable, but expressed in Pauli terms only.
+    ///
+    /// Examples:
+    ///
+    ///     >>> obs = SparseObservable("+")
+    ///     >>> obs.to_paulis()
+    ///     <SparseObservable with 2 terms on 1 qubit: (0.5+0j)() + (0.5+0j)(X_0)>
+    fn to_paulis(&self) -> PyResult<Self> {
```
jakelishman: Very minor: the name `to_paulis` in the context of Qiskit makes me think it might come out related to the `Pauli` class.  _Perhaps_ `as_paulis` makes it clearer from reading that it remains a `SparseObservable`?

jakelishman (commented):
Thanks for doing this - it'll be a big help for tests, and it's good to add the conversion, even if it's expensive in some cases.

We need to add the new methods into sensible places into the class-level documentation - they're not linked at the moment.

crates/accelerate/src/sparse_observable.rs
Cryoris: Good point, I'll add it to the `as_paulis` method. For `to_sparse_list` I also thought we didn't want to guarantee an order, so I left the comment that doing the roundtrip `SparseObservable -> sparse list -> SparseObservable` might not maintain the order of terms.

crates/accelerate/src/sparse_observable.rs
Cryoris: 1. that's legacy from when `to_paulis` returned the bits/indices/coeffs separately
2. good idea, I was looking for a way to avoid this!

qiskit/quantum_info/operators/symplectic/sparse_pauli_op.py
Cryoris: I think we said we don't want to add this because it would give a very easy path to a potentially very expensive representation -- so it's after a "you-need-an-explicit-constructor"-wall 😄

jakelishman: Ok sure, that's valid too.  I forgot we'd said that haha

crates/accelerate/src/sparse_observable.rs
Cryoris: Sounds good, that'd be a bit more intuitive 🙂 Also that'll verify that the tests are indeed order-independent 😉 

Good call on the `with_capacity` 👍🏻

Cryoris: Done in 984a75c54

crates/accelerate/src/sparse_observable.rs
Cryoris: Done in 984a75c54

crates/accelerate/src/sparse_observable.rs
Cryoris: Done in 984a75c54

jakelishman (approved):
Thanks for this.  I tweaked the docs slightly to add a couple more examples, and modify the cross-links a little bit, but other than that it looks good to go to me.

PR #13759, opened 2025-01-30
Add unitary gate representation to rust

### LINKED ISSUE #13272: Add rust representation of the `UnitaryGate` class

mtreinish:
### Summary

This commit expands the rust circuit data model to include a definition
of UnitaryGate. This is a more abstract operation than what we've
defined so far in Rust, a gate represented solely by it's unitary matrix
but during normal transpilation we create and interact with these
operations for optimization purposes so having a native representation
in rust is necessary to avoid calling to Python when working with them.
    
This introduces a new UnitaryGate struct which represents the unpacked
operation. It has 3 internal storage variants based on either an ndarray
arbitrary sized array, a 2x2 nalgebra fixed sized array, or a 4x4
nalgebra fixed sized array. From the python perspective these all look
the same, but being able to internally work with all 3 variants lets us
optimize the code paths for 1q and 2q unitary gates (which are by far
more common) and avoid one layer of pointer indirection.
    
When stored in a circuit the packed representation is just a pointer to
the actual UnitaryGate which we put in a Box. This is necessary because
the struct size is too large to fit in our compact representation of
operations. Even the ndarray which is a heap allocated type requires more
than our allotted space in a PackedOperation so we need to reduce it's
size by putting it in a Box.
    
The one major difference here from the previous python based
representation the unitary matrix was stored as an object type parameter
in the PackedInstruction.params field, which now it is stored internally
in the operation itself. There is arguably a behavior change around
this because it's no longer possible to mutate the array of a
UnitaryGate in place once it's inserted into the circuit. While doing
this was horribly unsound, because there was no guardrails for doing it
a release note is added because there is a small use case where it would
have worked and it wasn't explicitly documented.

### Details and comments

Closes #13272

A follow-up PR to this one will start to use UnitaryGate directly in rust instead of using the Python version.

~This is currently based on top of #13486 and will need to be rebased when that merges. To view the contents of just this PR without #13486 you can look at the HEAD commit: https://github.com/Qiskit/qiskit/commit/9bfe14bf95626812f9bd8a11c4bf57646f166255~


mtreinish:
I've rebased this now that #13486 has merged and it shouldn't be blocked anymore.

crates/circuit/src/dag_circuit.rs
```diff
@@ -2631,6 +2632,96 @@ def _format(operand):
                         | [OperationRef::Instruction(_), OperationRef::StandardInstruction(_)] => {
                             Ok(inst1.py_op_eq(py, inst2)? && check_args() && check_conditions()?)
                         }
+                        [OperationRef::Unitary(op_a), OperationRef::Unitary(op_b)] => {
+                            match [&op_a.array, &op_b.array] {
+                                [ArrayType::NDArray(a), ArrayType::NDArray(b)] => {
+                                    Ok(relative_eq!(a, b, max_relative = 1e-5, epsilon = 1e-8))
```
kevinhartman, line 2638: Less of a comment on this PR and more of a general question, but what is `max_relative` vs `epsilon`, exactly?

I spent some time looking at the docs for the `approx` crate, but I could quite grok it.

crates/circuit/src/dag_circuit.rs
```diff
@@ -2631,6 +2632,96 @@ def _format(operand):
...
+                                                    return Ok(false);
+                                                }
+                                            }
+                                        }
+                                        Ok(true)
+                                    } else {
+                                        Ok(false)
+                                    }
+                                }
+                                [ArrayType::OneQ(a), ArrayType::NDArray(b)] => {
```
kevinhartman: Can you combine arms like this to avoid duplication? I think Rust should be happy with this since the names will have the same types for both matches

```rust
[ArrayType::OneQ(one_q), ArrayType::NDArray(nd)] | [ArrayType::NDArray(nd), ArrayType::OneQ(one_q)]
```

crates/circuit/src/operations.rs
```diff
@@ -2792,3 +2805,101 @@ impl Operation for PyOperation {
...
+#[derive(Clone, Debug)]
+pub enum ArrayType {
+    NDArray(Array2<Complex64>),
+    OneQ(Matrix2<Complex64>),
+    TwoQ(Matrix4<Complex64>),
+}
+
+/// This class is used to wrap a Python side Operation that is not in the standard library
+#[derive(Clone, Debug)]
+// We bit-pack pointers to this, so having a known alignment even on 32-bit systems is good.
```
kevinhartman: Since we don't support 32-bit anymore, should we adjust the comment?

crates/circuit/src/circuit_instruction.rs
```diff
@@ -762,6 +766,45 @@ impl<'py> FromPyObject<'py> for OperationFromPython {
...
         }
 
+        // We need to check by name here to avoid a circular import during initial loading
+        if ob.getattr(intern!(py, "name"))?.extract::<String>()? == "unitary" {
+            let params = extract_params()?;
+            if let Param::Obj(data) = &params[0] {
+                let py_matrix: PyReadonlyArray2<Complex64> = data.extract(py)?;
+                let matrix: Option<MatrixView2<Complex64>> = py_matrix.try_as_matrix();
+                if let Some(x) = matrix {
+                    let unitary_gate = UnitaryGate {
```
kevinhartman: Is it better to make this variable be the `Box<UnitaryGate>`? I imagine the compiler will figure out that you want this to end up on the heap and thus avoid a copy from stack to heap, but maybe having it locally opens us up to someone in the future using / modifying it on the stack, preventing that optimization.

kevinhartman (commented):
This is so _beautiful_ (lol). It's nice to see how easily it fits in with the rest of our `PackedOperation` data model.

Perhaps worth noting, the `impl_packable_pointer!` macro also implements `From<UnitaryGate> for PackedOperation`, and that will automatically move the unitary into a `Box` and pack it into `PackedOperation`. But, I think what you're doing in `FromPyObject` for `OperationFromPython` is more likely to be optimized to elide a copy from stack to heap, so no changes are requested.

Just a few small comments, otherwise!

crates/circuit/src/dag_circuit.rs
mtreinish: Hah, what docs? It's very sparse from what I remember. 

The `relative_eq!` macro is basically checking something like: `abs(a - b) <= epsilon + max_relative * abs(b)`. The equivalent in python land would be `numpy.is_close()` where `epsilon` is the  `atol` argument and `max_relative` is `rtol`. You can look at the impl for the `RelativeEq` trait for f64 to get a feeling for how it works.

crates/circuit/src/dag_circuit.rs
mtreinish: I wasn't sure if the compiler was smart enough to resolve the typing in cases like this. But I can give a try, I'll update it if it's happy and report back if it's not.

crates/circuit/src/operations.rs
mtreinish: Oops this is a copy and paste error, I'm not sure how this ended up here. I'll clean this up.

crates/circuit/src/circuit_instruction.rs
mtreinish: At least in this context since we immediately pack it I wasn't thinking there was anything we could do locally in this context. I was thinking the compiler would see that nothing is touching the matrix after it's created. But I can immediately box it here if you think that's better.

kevinhartman: I'm not especially worried if you leave it as is, though I believe we create the box immediately for the other pointer codepaths.

mtreinish: Done in: https://github.com/Qiskit/qiskit/pull/13759/commits/ffb2ecf703ac776ffcc0fc6663b34dee81242530

crates/circuit/src/dag_circuit.rs
mtreinish: Yeah it works, done in: https://github.com/Qiskit/qiskit/pull/13759/commits/e1d6b9b2d3c0bd2f13bd7bb00836056092552d70

crates/circuit/src/operations.rs
mtreinish: Fixed in: https://github.com/Qiskit/qiskit/pull/13759/commits/66d397d35aca15d35536971e309db52101656f64

kevinhartman (approved):

PR #13761, opened 2025-01-30
Increase documentation of `ParameterVector` and `ParameterVectorElement`

jakelishman:
### Summary

These were somewhat lacking before, despite being fairly heavily used. `ParameterVectorElement` was not publicly documented, despite adding additional properties that were intended for public use.


Cryoris (approved):
LGTM but I won't put it on the merge queue since Luciano is assigned 🙂

jakelishman:
Oh, Luciano's just assigned because this just came out of a meeting he and I were in together. It was more an FYI, and I assumed others were busy.

Cryoris:
In that case....

PR #13765, opened 2025-01-30
Leverage native UnitaryGate from rust

mtreinish:
### Summary

This commit builds off of the native rust representation of a
UnitaryGate added in #13759 and uses the native representation
everywhere we were using UnitaryGate in rust via python previously:
the quantum_volume() function and consolidate blocks.

One future item is consolidate blocks can be updated to use nalgebra
types internally instead of ndarray as for the 1 and 2q cases we know
the fixed size of the array ahead of time. However the block
consolidation code is built using ndarray currently and later synthesis
code also works in ndarray so there isn't any real benefit yet, and we'd
ust add unecessary conversions and allocations. However, once #13649
merges this will change and it would make more sense to add the unitary
gate with a Matrix4. But this can be handled separately after this
merges.

### Details and comments

~This is based on top of #13759 and will need to be rebased once that merges. Until then you can see the contents of this PR by looking at the HEAD commit on the PR branch: https://github.com/Qiskit/qiskit/pull/13765/commits/8c36ffd10380e19891c186ea64b08b369b45ebbb~

### DIFF
--- a/crates/accelerate/src/consolidate_blocks.rs
+++ b/crates/accelerate/src/consolidate_blocks.rs
@@ -11,18 +11,22 @@
 // that they have been altered from the originals.
 
 use hashbrown::{HashMap, HashSet};
+use nalgebra::Matrix2;
 use ndarray::{aview2, Array2};
 use num_complex::Complex64;
-use numpy::{IntoPyArray, PyReadonlyArray2};
+use numpy::PyReadonlyArray2;
 use pyo3::intern;
 use pyo3::prelude::*;
 use rustworkx_core::petgraph::stable_graph::NodeIndex;
+use smallvec::smallvec;
 
 use qiskit_circuit::circuit_data::CircuitData;
+use qiskit_circuit::circuit_instruction::ExtraInstructionAttributes;
 use qiskit_circuit::dag_circuit::DAGCircuit;
 use qiskit_circuit::gate_matrix::{ONE_QUBIT_IDENTITY, TWO_QUBIT_IDENTITY};
-use qiskit_circuit::imports::{QI_OPERATOR, QUANTUM_CIRCUIT, UNITARY_GATE};
-use qiskit_circuit::operations::{Operation, Param};
+use qiskit_circuit::imports::{QI_OPERATOR, QUANTUM_CIRCUIT};
+use qiskit_circuit::operations::{ArrayType, Operation, Param, UnitaryGate};
+use qiskit_circuit::packed_instruction::PackedOperation;
 use qiskit_circuit::Qubit;
 
 use crate::convert_2q_block_matrix::{blocks_to_matrix, get_matrix_from_inst};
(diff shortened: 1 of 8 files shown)


mtreinish:
I ran a quick benchmark of `quantum_volume()` with this PR and compared it against 1.3.2 from PyPI. 

qv_construct_time

The release from PyPI will have PGO too which likely speeds it up a little bit.

mtreinish:
This should be ready for review now, I rebased this on top on main now that https://github.com/Qiskit/qiskit/pull/13759 has merged.

crates/accelerate/src/consolidate_blocks.rs
```diff
@@ -293,17 +308,19 @@ pub(crate) fn consolidate_blocks(
                     dag.remove_op_node(node);
                 }
             } else {
-                let array = aview2(&matrix).to_owned().into_pyarray(py);
-                let unitary_gate = UNITARY_GATE
-                    .get_bound(py)
-                    .call1((array, py.None(), false))?;
+                let array: Matrix2<Complex64> =
+                    Matrix2::from_row_iterator(matrix.into_iter().flat_map(|x| x.into_iter()));
```
Cryoris, line 312: It seems we'll end up needing a lot of these converters... it's a bit a bummer that `nalgebra` doesn't implement a `From<Array2>` or something

crates/circuit/src/dag_circuit.rs
```diff
@@ -6972,6 +6981,47 @@ impl DAGCircuit {
...
+            extra_attrs,
+            #[cfg(feature = "cache_pygates")]
+            py_op: OnceLock::new(),
+        });
+        if let Some(weight) = self.dag.node_weight_mut(node_index) {
+            *weight = new_weight;
+        }
+
+        // Update self.op_names
+        self.decrement_op(op_name.as_str());
```
Cryoris, line 7020: These look like they might do a roundtrip `&str -> String -> &str`?

Cryoris (commented):
Overall LGTM, just a teeny tiny comment below 🙂

crates/circuit/src/dag_circuit.rs
mtreinish: I did just copy what was there before I split this out to a method for working with PackedInstruction directly. I'll play with it a little bit and see if I can remove the double conversion (I have a feeling it's acting as a copy to avoid a borrow checker failure).

mtreinish: Yeah I just looked at it the `to_string()` is to copy the string. There is a shared reference issue if a copy isn't done here because everything is based on a references to dag and we're mutating it. So we need to copy the name string before we mutate. But the `decrement_op()` and `increment_op()` methods require `&str` as the input type.

Cryoris (approved):
Thanks for the inve`str`igation 🙂

PR #13766, opened 2025-01-30
Add `SparseObservable.compose`

### LINKED ISSUE #13390: Implement `SparseObservable.compose`
### What should we add?

The initial implementation of `SparseObservable` in Qiskit 1.3 doesn't include these core mathematical methods (mostly because I ran out of time).  They should be implemented later.

jakelishman:
### Summary

This adds matrix multiplication between `SparseObservable`s (or one `SparseObservable` and anything that can be coerced to an observable), using the same signature as similar methods in `quantum_info`.

The current implementation attempts to be relatively memory efficient, though there are specialisations that may be implemented if we wanted to be faster:

- if all the contained terms are Paulis, so each pairwise multiplication produces either zero or one alphabet letters, we currently have a redundant copy-in and copy-out of the Cartesian-product generator that effectively doubles the overhead.  We could check whether each term contains only Paulis once, and then avoid that overhead in those cases.

- alternatively, the Cartesian-product iterator could be taught to use the buffers of the output `SparseObservable` directly as its output space, rather than requiring copy-out of its buffers, which would remove the overhead from all calls, not just the single-Pauli ones, but requires a bit more bookkeeping, since the references would keep need to be updated.

- the case of `qargs` being not `None` and `front=True` currently involves an extra copy step to simplify the logic.  If this is an important case, we could add a right-matmul specialisation of `compose_map`.

### Details and comments

This commit currently does not contain tests, because we currently don't have suitable methods to test the validity of the output without coupling to the particular choice of factorisation of the matrix multiplication.  We need either a way to convert to a Pauli-only representation (with the exponential explosion that entails) or to a matrix; either of these can be uniquely canonicalised.

~WIP until it has tests.~

Close #13390


crates/accelerate/src/sparse_observable/mod.rs
```diff
@@ -620,6 +687,128 @@ impl SparseObservable {
...
+    /// finding an efficient re-factorization of the sum is generally equally computationally hard.
+    /// It's better to use domain knowledge of your observables to minimize the number of terms that
+    /// ever exist, rather than trying to simplify them after the fact.
+    ///
+    /// # Panics
+    ///
+    /// If `self` and `other` have different numbers of qubits.
+    pub fn compose(&self, other: &SparseObservable) -> SparseObservable {
+        if other.num_qubits != self.num_qubits {
+            panic!(
```
Cryoris, line 747: Theres the `ArithmeticError::MismatchedQubits` which we could return here

crates/accelerate/src/sparse_observable/lookup.rs
```diff
@@ -0,0 +1,294 @@
...
+    //
+    // These rules were generated by a brute-force search over increasing-n-tuples of (coeff, term)
+    // pairs, where `coeff` was restricted to magnitudes `(0.25, 0.5, 1.0)` at complex phases
+    // `(1, i, -1, -i)`.  Under those conditions, the rules are minimal in number of sum terms, but
+    // don't have any further logic to "curate" the choice of output factorisation.
+    match (left, right) {
+        (X, X) => Some(const { &[] }),
+        (X, Plus) => Some(const { &[(c64(1., 0.), Plus)] }),
+        (X, Minus) => Some(const { &[(c64(-1., 0.), Minus)] }),
+        (X, Y) => Some(const { &[(c64(0., 1.), Z)] }),
```
Cryoris, line 83: Just to be sure; the tuples are arranged in matrix-composition order (`result = tuple.0.dot(tuple.1)`) and not in the `.compose` order (`result = tuple.0.compose(tuple.1)`) you describe in the docstring of the compose method, right?

jakelishman:
Ok, I've written all the tests as being against the `SparsePauliOp` equivalent and subsequent canonicalisation, so we don't have to commit to specific factorisations.

crates/accelerate/src/sparse_observable/lookup.rs
```diff
@@ -0,0 +1,294 @@
...
+}
+
+/// The allowable items of `BitTerm`.  This is used by the lookup expansion; we need a const-safe
+/// way of iterating through all of the variants.
+static BIT_TERMS: [BitTerm; 9] = [X, Plus, Minus, Y, Right, Left, Z, Zero, One];
+
+/// Create a full binary lookup table for all valid combinations of `BitTerm`.
+macro_rules! expand_lookup_binary {
+    (static $table:ident: [$ty:ty], $generate:ident, $dummy:expr) => {
+        static $table: [$ty; 256] = {
```
Cryoris, line 29: Just for my own sanity here: you use 256=2^8 here because you want to use the 4 bits of two bit terms to index -- but there are only 81 entries, right?

crates/accelerate/src/sparse_observable/mod.rs
Cryoris, line 1: Is there some norm/guidelines in Rust as to what goes into `mod.rs` and what goes in separate files? To me its a bit confusing having a lot of logic in `mod.rs` as I'd expect this to live in some `sparse_observable.rs` file or something -- but if this is common practice I'm also fine

crates/accelerate/src/sparse_observable/mod.rs
```diff
@@ -1026,6 +1233,181 @@ impl ::std::ops::Neg for SparseObservable {
...
+        /// a given iteration, this never changes; it's just stored to make it easier to memcpy
+        /// out of.
+        indices: Vec<u32>,
+        /// The full set of [BitTerm]s (including ones that don't correspond to multiples).  The
+        /// multiple ones get updated inplace during iteration.
+        bit_terms: Vec<BitTerm>,
+        /// Whether iteration has started.
+        started: bool,
+        /// We always have at least one item to put out, but we're exhausted once we've iterated
+        /// through all our product terms.
```
Cryoris: Might be nice to explicitly something like this, I was a little confused when reading the case where the `matmul` is None and `exhausted` was directly set 🙂 
```suggestion
        /// We always have at least one item to put out. We're exhausted once we've iterated
        /// through all our product terms, or the product is zero.
```

crates/accelerate/src/sparse_observable/mod.rs
```diff
@@ -1263,6 +1645,24 @@ impl From<ArithmeticError> for PyErr {
     }
 }
 
+/// The single-character string label used to represent this term in the :class:`SparseObservable`
+/// alphabet.
+#[pyfunction]
+#[pyo3(name = "label")]
+fn bit_term_label(py: Python, slf: BitTerm) -> &Bound<PyString> {
```
Cryoris, line 1653: What's the difference of this to `BitTerm.py_label`?

crates/accelerate/src/sparse_observable/mod.rs
```diff
@@ -2941,6 +3354,118 @@ impl PySparseObservable {
...
+                other.get_type().repr()?
+            )));
+        };
+        let other = other.borrow();
+        let inner = self.inner.read().map_err(|_| InnerReadError)?;
+        let other_inner = other.inner.read().map_err(|_| InnerReadError)?;
+        if let Some(order) = qargs {
+            if other_inner.num_qubits() > inner.num_qubits() {
+                return Err(PyValueError::new_err(format!(
+                    "argument ({}) has more qubits than the base ({})",
```
Cryoris: Would this be clearer to read? 
```suggestion
                    "argument has more qubits ({}) than the base ({})",
```

crates/accelerate/src/sparse_observable/mod.rs
```diff
@@ -2941,6 +3354,118 @@ impl PySparseObservable {
...
+    /// observable, as the alphabet of :class:`SparseObservable` is not closed under composition;
+    /// the composition of two single-bit terms can be a sum, which multiplies the total number of
+    /// terms.  This memory usage is not _necessarily_ inherent to the resultant observable, but
+    /// finding an efficient re-factorization of the sum is generally equally computationally hard.
+    /// It's better to use domain knowledge of your observables to minimize the number of terms that
+    /// ever exist, rather than trying to simplify them after the fact.
+    ///
+    /// Args:
+    ///     other: the observable used to left-multiply ``self``.
+    ///     qargs: if given, the qubits in ``self`` to associated with the qubits in ``other``.  Put
```
Cryoris:
```suggestion
    ///     qargs: if given, the qubits in ``self`` to be associated with the qubits in ``other``.  Put
```

test/python/quantum_info/test_sparse_observable.py
```diff
@@ -241,6 +241,10 @@ def test_bit_term_enum(self):
             "1": SparseObservable.BitTerm.ONE,
         }
         self.assertEqual({label: SparseObservable.BitTerm[label] for label in labels}, labels)
+        # The `label` property retusnknown values.
```
Cryoris: Not actually sure what this should be, maybe "returns known"? 😛

Cryoris (commented):
This looks very nice, and given that it also worked in the Jordan-Wigner demo and the tests seems to cover a lot of edge cases I'm very happy with this PR. I just have a few question below, otherwise we can merge 🙂

crates/accelerate/src/sparse_observable/lookup.rs
jakelishman: Yeah, exactly. There are 81 explicit entries (though some are `None` themselves, so there's slightly fewer than 81 `Some` values), but the highest populated index is `0b10111011` (`(Left, Left)`, I think?), so most of the way to `0xff`.

crates/accelerate/src/sparse_observable/lookup.rs
jakelishman: Yeah, I called the function `matmul`, so it's matrix-multiplication order.

crates/accelerate/src/sparse_observable/mod.rs
jakelishman: it's in the comment 1 line before this diff haha:

> This doesn't use `py_label` so we can use `intern!`.

crates/accelerate/src/sparse_observable/mod.rs
jakelishman: I panicked in the Rust-space methods by analogy to how the other mathematical functions work, and how `ndarray` works, etc - it's a Rust programmer logic error to pass bad inputs to `compose`, and it's a simple invariant to uphold.

If you want, we can have `try_compose` and `try_compose_map` for `Result` variants (there's a couple more errors we'd have to add), and `compose`, `compose_map` for panicking ones.

test/python/quantum_info/test_sparse_observable.py
jakelishman: Fixed in 5dd0666c99.

crates/accelerate/src/sparse_observable/mod.rs
jakelishman: Done in 5dd0666c99.

crates/accelerate/src/sparse_observable/mod.rs
jakelishman: Sure, if you prefer.  Done in 5dd0666c99.

crates/accelerate/src/sparse_observable/mod.rs
jakelishman: Oh, the comment is actually just wrong - we don't always have at least one term to put out.  Changed in 5dd0666c99.

crates/accelerate/src/sparse_observable/mod.rs
Cryoris: Ah, yeah, that also makes sense. This comment mainly comes from the fact that we raise errors in the Python function, but we can add `try_` variants also later 👍🏻

crates/accelerate/src/sparse_observable/mod.rs
Cryoris: Ahh, this reads as "we would like to use `intern!` hence we don't use `py_label`"

Cryoris (approved):
LGTM, thank you!

PR #13520, opened 2024-12-03
Update the circuit library docs

Cryoris:
e.g. why we use different gate types and what's the advantages

### Summary

With #13046 the designated workflow with the circuit library changed. This PR adds a section to the docs explaining this new structure. While not fully done, I'm already opening this draft PR for people to have a look at the text.

### Details and comments

This includes some other minor changes:
* add a TOC for the different sections
* consistently sort the standard gates (before this was partially sorted, partially chaotic)
* move the documentation of `get_standard_gate_name_mapping` under `Standard gates`
* always mention the circuit functions and gates _before_ the old circuits

### To do

- [x] Add some more examples to the sections 
- [ ] Move the template circuits to a separate page
- [ ] Can we find a better name/structure for "particular quantum circuits"?

PR #13769, opened 2025-01-30
Remove dependance of `BooleanExpression` on `tweedledum`

### LINKED ISSUE #13755: Remove tweedledum from qiskit-sdk
### What should we add?

The [`tweedledum`](https://github.com/boschmitt/tweedledum) external library is currently used inside `qiskit` extensively in the `circuit.classical` library. It is used when parsing boolean and Python expressions to generate an inner representation (`ClassicalFunction`), to read cnf representations from DIMACS files (`BooleanExpression`), and to simulate and synthesize quantum circuits for those functions, especially as part of `PhaseOracle`.

However, it seems the library is no longer maintained and we need to drop it as part of the 2.0.0 release.

Essentially, this requires us to provide different ways to:
1. Parse Python expressions.
2. Parse DIMACS file (or drop support)
3. Simulate (i.e. evaluate) the resulting expressions on bitstrings.
4. Synthesize the expressions to `QuantumCircuit`.

Doing 1 and 2 is very simply and can be done with not major changes to the code; doing 1 requires some modification of how `ClassicalFunctionVisitor` works since currently it relies on tweedledum's `LogicNetwork` class. For 2, since DIMACS is a very simply file format, parsing can be done via a small standalone function

Given convenient representations resulting from 1+2, doing 3 is also fairly simple, leaving 4 as the only real obstacle. The synthesis of a quantum circuit for a given logical expression is highly dependent on the complexity of the representation of the expression. Currently, tweedledum uses a method that relies on the ESOP (exclusive sum of products) for small number of variables, and a method based on XAG (xor-and graph) for more complex expressions; see Bruno Schmitt's description [here](https://quantumcomputing.stackexchange.com/questions/22173/phase-oracle-in-qiskit-solving-satisfiability-problems-using-grovers-algorithm).

Obtaining a convenient ESOP representation in cases where the truth table for the expression can be conveniently computed (i.e. given a way to do 3) is straightforward, and given such an ESOP, synthesizing a circuit is immediate. However, dropping tweedledum's support means we'll encounter difficulty in handling more complex expressions where there are too many variables to compute the truth table. In the future we can explore the idea of using more robust tools such as Z3 for optimizing this stage. For the 2.0.0 release I suggest we focus on removing the reliance on tweedledum with minimal interface changes, in the cost of being unusable for some cases.

Note, however, that since synthesizing ESOP is always easy, we can add a new feature for the user to supply the ESOP directly, in addition to 1+2, so the expression optimization stage can be done by the user with his libraries of choice.

gadial:
### Summary
Rewrites the `BooleanExpression` class such that expression parsing and circuit synthesis does not rely on the `tweedledum` library, and moves it to the `synthesis` library. Adds a `BitFlipOracle` class to preserve the user-facing usage of `BooleanExpression`.

Partially addresses #13755.

### Details and comments
`BooleanExpression` is used to describe classical boolean functions and synthesize phase/bit-flip oracles for them. Currently it is used directly within qiskit only for the `PhaseOracle` class.

The current implementation of `BooleanExpression` relies on the `tweedledum` library for parsing and synthesizing. Since qiskit 2.0 will stop using `tweedledum`, this PR makes the code of `BooleanExpression` independent, in the cost of lower capabilities.

These are the main changes:
1. Parsing is now done using the `ast` library, using a custom `boolean_expression_visitor`. Parsing DIMACS files is done directly using regexp.
2. Oracle synthesis is now performed directly (within the new `boolean_expression_synth` module) using a straightforward approach seemingly also employed by `tweedledum`, given a representation of the function as ESOP (Exclusive sum of squares).
3. A rudimentary procedure for converting a `BooleanExpression` into ESOP representation was also added. While it features basic optimizations to reduce the size of the ESOP, it needs to generate the full truth table of the represented function, making it a viable approach only for functions on a relatively low number of variables. This can be improved upon in the future if required, using more sophisticated ESOP generation and minimization techniques (such as using BDDs or SAT-solvers).
4. All the synthesis-related code was moved to a new directory, `synthesis/boolean`. A new class, `BitFlipOracle` was added to the circuit library. The old `BooleanExpression` code remains unchanged and will be deprecated in a separate PR along with the rest of the `classicalfunction` library.

Non comprehensive benchmarking indicates a factor-10 runtime blowup for the non-tweedledum version, but oracle generation for functions of around 10 variables is still viable.

```
╒═══════════════════════════════╤═════════════╤═════════════╤═══════════════════╕
│ Test Case                     │         new │         old │   Ratio (new/old) │
╞═══════════════════════════════╪═════════════╪═════════════╪═══════════════════╡
│ 3 vars, 9 clauses (seed 42)   │ 0.000578117 │ 0.00033474  │          1.72707  │
├───────────────────────────────┼─────────────┼─────────────┼───────────────────┤
│ 2 vars, 6 clauses (seed 42)   │ 0.000580263 │ 0.0016449   │          0.352766 │
├───────────────────────────────┼─────────────┼─────────────┼───────────────────┤
│ 4 vars, 12 clauses (seed 42)  │ 0.00112677  │ 0.000320101 │          3.52004  │
├───────────────────────────────┼─────────────┼─────────────┼───────────────────┤
│ 5 vars, 15 clauses (seed 42)  │ 0.00388117  │ 0.00120397  │          3.22365  │
├───────────────────────────────┼─────────────┼─────────────┼───────────────────┤
│ 6 vars, 18 clauses (seed 42)  │ 0.0112587   │ 0.00248165  │          4.5368   │
├───────────────────────────────┼─────────────┼─────────────┼───────────────────┤
│ 7 vars, 21 clauses (seed 42)  │ 0.0248424   │ 0.0142735   │          1.74045  │
├───────────────────────────────┼─────────────┼─────────────┼───────────────────┤
│ 8 vars, 24 clauses (seed 42)  │ 0.0663635   │ 0.0208194   │          3.18758  │
├───────────────────────────────┼─────────────┼─────────────┼───────────────────┤
│ 9 vars, 27 clauses (seed 42)  │ 0.156358    │ 0.0419133   │          3.73051  │
├───────────────────────────────┼─────────────┼─────────────┼───────────────────┤
│ 10 vars, 30 clauses (seed 42) │ 0.406802    │ 0.071683    │          5.67501  │
├───────────────────────────────┼─────────────┼─────────────┼───────────────────┤
│ 11 vars, 33 clauses (seed 42) │ 0.863733    │ 0.156799    │          5.50854  │
├───────────────────────────────┼─────────────┼─────────────┼───────────────────┤
│ 12 vars, 36 clauses (seed 42) │ 2.25606     │ 0.234907    │          9.60408  │
╘═══════════════════════════════╧═════════════╧═════════════╧═══════════════════╛
```

Comparing the CX gate count in the result of transpiling the oracle for `GenericBackendV2(num_qubits=27)` with pass manager `generate_preset_pass_manager(optimization_level=2, backend=backend)` yields a slight advantage to the new implementation:

```
╒═══════════════════════════════╤═══════╤═══════╤═══════════════════╕
│ Test Case                     │   new │   old │   Ratio (new/old) │
╞═══════════════════════════════╪═══════╪═══════╪═══════════════════╡
│ 2 vars, 6 clauses (seed 42)   │     0 │     0 │        nan        │
├───────────────────────────────┼───────┼───────┼───────────────────┤
│ 3 vars, 9 clauses (seed 42)   │     0 │     0 │        nan        │
├───────────────────────────────┼───────┼───────┼───────────────────┤
│ 4 vars, 12 clauses (seed 42)  │     0 │     0 │        nan        │
├───────────────────────────────┼───────┼───────┼───────────────────┤
│ 5 vars, 15 clauses (seed 42)  │    36 │    36 │          1        │
├───────────────────────────────┼───────┼───────┼───────────────────┤
│ 6 vars, 18 clauses (seed 42)  │   218 │   244 │          0.893443 │
├───────────────────────────────┼───────┼───────┼───────────────────┤
│ 7 vars, 21 clauses (seed 42)  │   694 │  1088 │          0.637868 │
├───────────────────────────────┼───────┼───────┼───────────────────┤
│ 8 vars, 24 clauses (seed 42)  │  1842 │  3125 │          0.58944  │
├───────────────────────────────┼───────┼───────┼───────────────────┤
│ 9 vars, 27 clauses (seed 42)  │  6954 │  8238 │          0.844137 │
├───────────────────────────────┼───────┼───────┼───────────────────┤
│ 10 vars, 30 clauses (seed 42) │ 21927 │ 25955 │          0.844808 │
╘═══════════════════════════════╧═══════╧═══════╧═══════════════════╛
```


qiskit/circuit/classicalfunction/boolean_expression.py
```diff
@@ -12,16 +12,19 @@
...
+from qiskit.circuit import Gate
+from .boolean_expression_visitor import (
+    BooleanExpressionEvalVisitor,
+    BooleanExpressionArgsCollectorVisitor,
+)
 
 
-@HAS_TWEEDLEDUM.require_in_instance
-class BooleanExpression(ClassicalElement):
+class BooleanExpression(Gate):
```
ShellyGarion: Note the API change here (changed types)

qiskit/circuit/classicalfunction/boolean_expression.py
```diff
@@ -58,72 +63,123 @@ def simulate(self, bitstring: str) -> bool:
...
+
+        By default the bit-flip oracle is generated.
 
         Args:
-            registerless: Default ``True``. If ``False`` uses the parameter names
-                to create registers with those names. Otherwise, creates a circuit with a flat
-                quantum register.
-            synthesizer: A callable that takes self and returns a Tweedledum
-                circuit.
+            circuit_type: which type of oracle to create, 'bit' or 'phase' flip oracle.
```
ShellyGarion: Note the API change in this function

qiskit/circuit/library/phase_oracle.py
```diff
@@ -50,40 +44,18 @@ class PhaseOracle(QuantumCircuit):
 
     def __init__(
         self,
-        expression: Union[str, ClassicalElement],
-        synthesizer: Optional[Callable[[BooleanExpression], QuantumCircuit]] = None,
+        expression: str,
         var_order: list = None,
```
ShellyGarion: Note the API change in the init

ShellyGarion (commented):
Thanks @gadial for taking care of this - it looks great!

Note that there are some necessary API changes here, which makes sense and it's also allowed for Qiskit 2.0, but probably need better release notes, as well as raise deprecation warnings in Qiskit 1.4. 

Did you make some experiments to check the performance compared to Tweedledum?

qiskit/circuit/classicalfunction/boolean_expression.py
Cryoris: The easiest deprecation path would be to just deprecate `BooleanExpression` in place and have the new version in the circuit library. 

It might also be worth considering a renaming to avoid too much confusion with the new object. What does this circuit implement exactly? Does it flip a target bit if the boolean expression is satisfied or does it something in-place? If it's the first, maybe `BitflipOracle` could be a good name.

gadial: This seems like a sound strategy - we already have a `PhaseOracle` so we can add `BitFlipOracle` which will be similar but with a different synthesis, and both gates would rely on a non-gate version of `BooleanExpression` (maybe with a different name) which handles parsing, simulating (to create the truth table) and synthesizing (since we might want to invest in optimized algorithms later.

The main question is where to put those files. We can put `BitFlipOracle` next to `PhaseOracle` in `qiskit.circuit.library` but we still need to put the `NewBooleanExpression` module (which would consist at least of `boolean_expression`, `boolean_expression_visitor` and `boolean_expression_synth` files) somewhere that makes sense.

gadial: Done in e403028

qiskit/circuit/library/bit_flip_oracle.py
```diff
@@ -0,0 +1,127 @@
...
+    A bit-flip oracle for a boolean function `f(x)` performs the following
+    quantum operation:
+
+    .. math::
+
+            |x\rangle|y\rangle \mapsto |x\rangle|f(x)\oplus y\rangle
+
+    For convenience, this oracle, in addition to parsing arbitrary logical expressions,
+    also supports input strings in the `DIMACS CNF format
+    <http://www.satcompetition.org/2009/format-benchmarks2009.html>`__,
```
ShellyGarion: the link gives 404 error (note that it also appears in other places)

qiskit/circuit/library/__init__.py
```diff
@@ -345,6 +345,7 @@
    QuantumVolume
    PhaseEstimation
    GroverOperator
+   BitFlipOracle
```
ShellyGarion: perhaps this should be done in another PR, like @Cryoris's https://github.com/Qiskit/qiskit/pull/13520, but I think that the 3 oracles (grover, phase and bitflip) deserve their own section in the circuit library docs.

qiskit/synthesis/boolean/boolean_expression.py
```diff
@@ -0,0 +1,185 @@
...
+from os.path import basename, isfile
+
+from qiskit.circuit import Gate
+from .boolean_expression_visitor import (
+    BooleanExpressionEvalVisitor,
+    BooleanExpressionArgsCollectorVisitor,
+)
+
+
+class BooleanExpression(Gate):
```
ShellyGarion: we usually don't have gates in the synthesis library, but I don't know where to put it.
this won't be publicly available?
does it have to be a gate? perhaps it can just be an abstract class?

releasenotes/notes/boolean_expression_update-39c19edbbe71ba0d.yaml
```diff
@@ -0,0 +1,8 @@
+---
+upgrade_circuits:
+  - |
+    The :class:`.PhaseOracle` no longer relies on the `tweedledum` library
+    but might not be synthesize as effectively as before.
```
ShellyGarion: synthesize --> synthesized

Do you have some performance estimation?

test/python/circuit/library/test_phase_oracle.py
```diff
@@ -19,11 +19,9 @@
 from qiskit.circuit import QuantumCircuit
 from qiskit.circuit.library import PhaseOracle
 from qiskit.quantum_info import Statevector
-from qiskit.utils.optionals import HAS_TWEEDLEDUM
 from test import QiskitTestCase  # pylint: disable=wrong-import-order
 
 
-@unittest.skipUnless(HAS_TWEEDLEDUM, "Tweedledum is required for these tests")
 @ddt
 class TestPhaseOracle(QiskitTestCase):
```
ShellyGarion: Is there also a test for the BitFlipOracle?

ShellyGarion (commented):
Thanks @gadial for taking care of this! I have a few comments and questions.

releasenotes/notes/boolean_expression_update-39c19edbbe71ba0d.yaml
```diff
@@ -0,0 +1,8 @@
...
+    :class:`.BitFlipOracle` was added as an alternative to
+    directly synthesizing :class:`.BooleanExpression`, as this
+    class will be removed in Qiskit 2.0.
```
ShellyGarion: will be --> is

qiskit/circuit/library/__init__.py
gadial: Seems like a good idea but I agree it belongs in another PR.

qiskit/synthesis/boolean/boolean_expression.py
gadial: It should not be a gate, this was a leftover from the old `BooleanExpression`. Fixed in 91a8c47c8

ShellyGarion (commented):
LGTM, thanks @gadial !

Thanks for the comparison of synthesis time compared to tweedledum.
I also wonder what is the CX-cost comparison of the transpiled circuits.

ShellyGarion (dismissed):
LGTM. @Cryoris - do you have any further comments?

qiskit/circuit/library/bit_flip_oracle.py
```diff
@@ -0,0 +1,127 @@
...
+
+"""Bit-flip Oracle object."""
+
+from qiskit.circuit import QuantumCircuit
+
+from qiskit.synthesis.boolean.boolean_expression import BooleanExpression
+
+
+class BitFlipOracle(QuantumCircuit):
+    r"""Bit-flip Oracle.
```
alexanderivrii: In the spirit of circuit library refactoring, do we want `BitFlipOracle` and `PhaseOracle` to also have `Gate` variants? @Cryoris, what are your thoughts on this? What about `BooleanExpression`?

gadial: Personally I'm againt `BooleanExpression` being a `Gate` (as is currently the situation) since it's "interface" is not clear, as is reflected in the difference between phase oracle (n qubits, applies Z conditioned on f(x)) and bit flip oracle (n+1 qubits, applies X on a specific qubit conditioned on f(x)). Currently the somewhat arbitrary choice was to have `BooleanExpression` as a gate act as a bit-flip oracle.

qiskit/circuit/library/phase_oracle.py
```diff
@@ -36,9 +27,16 @@ class PhaseOracle(QuantumCircuit):
     For example, `'a & b'`, and `(v0 | ~v1) & (~v2 & v3)`
     are both valid string representation of boolean logical expressions.
 
+    A phase oracle for a boolean function `f(x)` performs the following
+    quantum operation:
```
alexanderivrii, line 34: From the description above these lines, it seems that ``'a AND b'`` is also allowed as an input expression, though I don't believe this is the case. Maybe something like this would be a bit more clearer:
```
`&` (logical AND), `|` (logical OR), `~` (logical NOT), and `^` (logical XOR).
```

releasenotes/notes/boolean_expression_update-39c19edbbe71ba0d.yaml
```diff
@@ -0,0 +1,8 @@
+---
+upgrade_circuits:
+  - |
+    The :class:`.PhaseOracle` no longer relies on the `tweedledum` library
+    but might not be synthesized as effectively as before.
+    :class:`.BitFlipOracle` was added as an alternative to
+    directly synthesizing :class:`.BooleanExpression`, as this
+    class is removed in Qiskit 2.0.
```
alexanderivrii: It would be nice to expand the release notes a bit.

First, it would be nice to emphasize the API change when calling `PhaseOracle`: i.e. the argument `expression` can now only be a `str`, and the argument `synthesizer` is no longer supported. It would be nice to give an example of using `PhaseOracle`.

Second, it would be nice to add a more detailed description of the new `BitFlipOracle` in `features_circuits` part of the release notes. It would also be nice to give an example on using `BitFlipOracle` (including using `evaluate_bitstring`).

qiskit/circuit/library/bit_flip_oracle.py
alexanderivrii: I completely agree that `BooleanExpression` should not be a `Gate`, thanks! I also think that we probably want to have the new `BitFlipOracle` available as a subclass of a quantum circuit, at the least to be consistent with `PhaseOracle`. However, for the sake of circuit library refactoring, in a follow-up we should also expose phase oracle and bit flip oracles as gates.

qiskit/synthesis/boolean/boolean_expression.py
```diff
@@ -0,0 +1,223 @@
...
+
+
+class TruthTable:
+    """A simple implementation of a truth table for a boolean function"""
+
+    EXPLICIT_REP_THRESHOLD = (
+        12  # above this number of bits, do not explicitly save the values in a list
+    )
+
+    def __init__(self, func, num_bits):
```
alexanderivrii: Even though it's an internal class, it would be nice to describe the arguments, possibly using type hints.

qiskit/synthesis/boolean/boolean_expression.py
```diff
@@ -0,0 +1,223 @@
...
+            self.values = np.array(
+                [self.func(assignment) for assignment in self.all_assignments()],
+                dtype=bool,
+            )
+        else:
+            self.implicit_values = {}
+
+    def all_assignments(self):
+        """Return an ordered list of all assignments, ordered from left to right
+        i.e. 000, 100, 010, 110, 001, 101, 011, 111"""
```
alexanderivrii: Shouldn't this be "from right to left"? Or maybe something like "from least significant bit to most significant bit"?

qiskit/synthesis/boolean/boolean_expression_synth.py
```diff
@@ -0,0 +1,126 @@
...
+                diff = self.clause_diff(clause_1, clause_2)
+                if len(diff) == 1:
+                    new_clause = clause_1[: diff[0]] + "-" + clause_1[diff[0] + 1 :]
+                    if new_clause not in combined_clauses:
+                        combined_clauses.remove(clause_1)
+                        combined_clauses.remove(clause_2)
+                        combined_clauses.append(new_clause)
+                        something_changed = True
+                        break
+        return combined_clauses
```
alexanderivrii, line 48: It seems that one can significantly optimize this function, but it's better to do this in a follow-up.

alexanderivrii (commented):
This is a really-really cool PR. Thanks for making the phase oracle (and its friend, bit-flip oracle) available to those who can no longer use tweedledum.

My only suggestions are around documentation: expanding the release notes and adding python type hints to various functions.

qiskit/synthesis/boolean/boolean_expression_synth.py
```diff
@@ -0,0 +1,126 @@
...
+        self.truth_table = truth_table
+        self.cofactor_table = {}
+        self.esop = self.generate_esop(assignment=())
+
+    def clause_diff(self, clause1, clause2):
+        """The indices of variables where the clauses differ"""
+        return [i for i in range(len(clause1)) if clause1[i] != clause2[i]]
+
+    def combine_and_optimize(self, clauses_1, clauses_2):
+        """Combining clauses of distance 1 until no more combinations can be performed"""
```
alexanderivrii, line 33: One minor question: why do you pass `clauses_1` and `clauses_2` and not just `clauses_1 + clauses_2`?

qiskit/synthesis/boolean/boolean_expression.py
```diff
@@ -0,0 +1,223 @@
...
+        ]
+
+    def __getitem__(self, key):
+        if isinstance(key, str):
+            key = (bit != "0" for bit in key)
+        if self.explicit_storage:
+            key = sum(2**n for n, bit in enumerate(key) if bit)
+            return self.values[key]
+        else:
+            return self.implicit_values.setdefault(key, self.func(key))
```
alexanderivrii, line 70: Why do we need to insert `(key, value)` to `implicit_values`? I can see how this can be an optimization when we need the same value multiple times, but if the calling algorithm goes over all assignments, wouldn't this build the full truth table?

qiskit/synthesis/boolean/boolean_expression_synth.py
```diff
@@ -0,0 +1,126 @@
...
+
+    def generate_esop(self, assignment):
+        """Recursively generates an ESOP for a partially determined boolean expression
+        'assignment' is a partial assignment to the expression's variables
+        """
+        # this base case is non-optimized;
+        # we may be able to terminate even before having a full assignment
+        if len(assignment) == self.truth_table.num_bits:
+            if self.truth_table[assignment]:
+                return ["-" * self.truth_table.num_bits]  # True; a don't care clause
```
alexanderivrii, line 58: I don't understand this line. Let's say the assignment is 101. Why does this return "---" and not "101"?

qiskit/synthesis/boolean/boolean_expression_synth.py
```diff
@@ -0,0 +1,126 @@
...
+            else:
+                return []  # False
+        if assignment in self.cofactor_table:
+            return self.cofactor_table[assignment]
```
alexanderivrii: With the current approach of building the ESOP bottom-up, how can we examine the same assignment multiple times?

qiskit/circuit/library/phase_oracle.py
gadial: It **is** allowed. It's not a correct syntax, Python-wise, since AND is for booleans and & is for bitwise, but it was like this before and I don't see why not to keep it, since it can only help users. You can see [here](https://github.com/gadial/qiskit/blob/5f4d12f40a779855b6fefbfc7470a3fc1bba5e0f/qiskit/synthesis/boolean/boolean_expression_visitor.py#L26) how both cases are handled.

qiskit/synthesis/boolean/boolean_expression.py
gadial: You are correct, it's simply "from right to left". Will fix.

qiskit/synthesis/boolean/boolean_expression_synth.py
gadial: I'm open to suggestions but indeed, the goal here was to keep it simple and establish a baseline.

qiskit/synthesis/boolean/boolean_expression_synth.py
gadial: That's mostly a leftover from a previous version, in which I tried optimizing based on the fact that at the beginning, we are only interested in clauses obtained by `clause_1` from `clauses_1` and `clause_2` from `clauses_2`. It still makes sense to keep the interface as it is now, in my opinion; it gives more information than passing the combined clauses.

qiskit/synthesis/boolean/boolean_expression.py
gadial: The `TruthTable` object is created to be somewhat generic, without a specific algorithm in mind. If we switch to an algorithm that doesn't have to read the whole truth table for whatever reason, I still want it to function correctly and with relative efficiency (i.e. never use `func` twice on the same input if possible).

qiskit/synthesis/boolean/boolean_expression_synth.py
gadial: This is the base case for a function that was already fully specified, assignment-wise, so now it's a constant. If it's the constant `T`, then the assignment doesn't matter. Later, when returning from the recursion base, we add specifications into the clause. See [here](https://people.csail.mit.edu/costis/papers/esop.pdf) for more details (what we do here is basically what the paper refers to as the Shannon expansion).

qiskit/synthesis/boolean/boolean_expression_synth.py
gadial: I don't think it can happen in the original algorithm, so although I think the code is more robust this way, it can only slow down the runtime. I'll remove this.

qiskit/synthesis/boolean/boolean_expression.py
gadial: Done in 831ec93

qiskit/synthesis/boolean/boolean_expression.py
gadial: Done in 831ec93

releasenotes/notes/boolean_expression_update-39c19edbbe71ba0d.yaml
gadial: Done in 831ec93

qiskit/synthesis/boolean/boolean_expression_synth.py
gadial: Done in 831ec93

qiskit/circuit/library/bit_flip_oracle.py
Cryoris: I would like to make a point for `Gate`s 🙂 in the circuit library modernization (#13046) we define that all objects are either gates or functions -- gates, if their action is defined by a mathematical operation and we are free to synthesize with different algorithms, and functions, if they are defined by their structure.

To me, this is therefor a clear case of a `Gate`. We can discuss what interface makes most sense, if there are multiple options. What you suggested above makes sense to me:
* a `BitFlipOracle` flips a target qubit if `f(x)` is satisfied
* a `PhaseOracle` flips the phase of states satisfying `f(x)` -- note that it is fine if the actual implementation uses an auxiliary qubit.

I would prefer not introducing a `BitFlipOracle(QuantumCircuit)` if we add a gate. The `PhaseOracle` should also be updated to be a gate once we agree on the interface, to match the rest of the library and enable different synthesis algorithms.

qiskit/circuit/library/bit_flip_oracle.py
```diff
@@ -0,0 +1,127 @@
...
+        self,
+        expression: str,
+        var_order: list = None,
+    ) -> None:
+        """Creates a BitFlipOracle object
+
+        Args:
+            expression: A Python-like boolean expression.
+            var_order(list): A list with the order in which variables will be created.
+               (default: by appearance)
```
Cryoris: Some minor comments on docstrings (which might be off in the original class already, not that you added this 🙂)
*  if `None` is supported we can annotate it with `OtherType | None` (this might require `from __future__ import annotations`)
* you could also add the type of the list elements -- I'm not sure but this might be `list[str]`?
* if the type is in the signature it can be skipped in the docstring
* we typically skip the initial sentence in the initializer
```suggestion
        var_order: list[str] | None = None,
    ) -> None:
        """
        Args:
            expression: A Python-like boolean expression.
            var_order: A list with the order in which variables will be created.
               (default: by appearance)
```

releasenotes/notes/boolean_expression_update-39c19edbbe71ba0d.yaml
```diff
@@ -0,0 +1,66 @@
...
+
+    The interface of :class:`.PhaseOracle` was simplified; it no longer
+    accepts a `synthesizer` paramter, and the `expression` parameter
```
alexanderivrii: Typo in the word "parameter".

releasenotes/notes/boolean_expression_update-39c19edbbe71ba0d.yaml
```diff
@@ -0,0 +1,66 @@
...
+    can only be a string; `ClassicalElement` have been deprecated.
```
alexanderivrii: should it be "has been deprecated"?

releasenotes/notes/boolean_expression_update-39c19edbbe71ba0d.yaml
```diff
@@ -0,0 +1,66 @@
...
+      q_4: ┤ X ├┤ X ├┤ X ├
+           └───┘└───┘└───┘
+
+    .. code-block:: python
+
+      print(oracle.evaluate_bitstring("1010"))
+
+    .. code-block:: text
+
+      True
```
alexanderivrii: You might want to add one (empty) line at the end of the file

releasenotes/notes/boolean_expression_update-39c19edbbe71ba0d.yaml
```diff
@@ -0,0 +1,66 @@
...
+    accepts a `synthesizer` paramter, and the `expression` parameter
+    can only be a string; `ClassicalElement` have been deprecated.
+
+    :class:`.PhaseOracle` is used exactly as before:
+    .. code-block:: python
+      
+      from qiskit.circuit.library.phase_oracle import PhaseOracle
+      bool_expr = "(x0 & x1 | ~x2) & x4"
+      oracle = PhaseOracle(bool_expr)
+      print(oracle))
```
alexanderivrii: You have an extra right parenthesis in `print(oracle)`

qiskit/synthesis/boolean/boolean_expression_synth.py
alexanderivrii: Thanks, I see, all these `-` values will be later replaced by either `0` or `1`, so we could have just as well put any other symbol (e.g., `*`) there instead of `-`. So my only (minor) concern is that if in the future the algorithm for generating ESOPs changes, we don't mix up between `-` symbols that mean "the value will later be overwritten" and between `-` symbols that mean "the corresponding variable has already been expanded and the value can be both `0` and `1`". Maybe you can put indeed `"*"` there to avoid such potential future, or add a comment in code that explains this point.

qiskit/synthesis/boolean/boolean_expression_synth.py
alexanderivrii: I agree. I was not suggesting to do in for this PR.

qiskit/circuit/library/phase_oracle.py
alexanderivrii: As we briefly discussed in-person, the code raises an error if we put either "AND" or "and" instead of "&" (the two errors are actually a bit different). I think it's completely fine not to support python-like boolean syntax (i.e. using "and" instead of "&"), but it's up to you. However, if the capital-case "AND" is not supported, I am suggesting to fix the description above.

qiskit/circuit/library/bit_flip_oracle.py
alexanderivrii: On the point of interface to the future `PhaseOracleGate` and `BitFlipOracleGate`, do you think these gates should have the "evaluate" and "from_dimacs" methods?

qiskit/synthesis/boolean/boolean_expression.py
alexanderivrii: In this case, I would suggest to reduce the size of the cache if it gets too large (for example, simply remove all cached values).

qiskit/synthesis/boolean/boolean_expression_synth.py
gadial: On the contrary; I think that using `*` might cause future problems, that is why I used `-`.

In the Shannon expansion, the idea is to recursively generate ESOP's for partially-assigned functions ("subfunctions"). For the subfunction which is always `T`, the correct ESOP is indeed one clause of `-` for all variables. So it's not here as a placeholder but as the actual base case.

Note, if you look at the note I refenced above, you can see that one might use other expansions (e.g. positive and negative Davio) which use the ESOP for a subfunction but does not fill it completely. Hence, if we extend the current code to handle Davio as well, the correct way to go is to have `-` clauses.

qiskit/circuit/library/phase_oracle.py
gadial: The description might be somewhat misleading, but it never actually meant to say that you can use `AND` and such, it's simply an explanation of the meaning of the symbol `&`. I won't change this description since we've had it for a long time and it's factually correct. What I'll change is switch `AND` to `logical AND` so it's obvious this is not a syntax but a description.

qiskit/circuit/library/bit_flip_oracle.py
gadial: We need to clarify this point before finalizing the PR. What is the exact difference between `QuantumCircuit` subclass and `Gate` subclass, other than the actual class declaration? Can we switch `PhaseOracle` to be `Gate` without additional deprecations?

Also, `PhaseOracle` won't need an auxiliary qubit. Although indeed the conversion of bit-flip oracle to phase oracle uses such a qubit, it's not needed in direct synthesis from the function `f(x)` itself.

qiskit/synthesis/boolean/boolean_expression_synth.py
alexanderivrii: I agree now -- thanks for clarifying this!

qiskit/circuit/library/bit_flip_oracle.py
Cryoris: > We need to clarify this point before finalizing the PR.

I'm fine keeping the `from_dimacs` as method, we commonly have separate constructors. I'm less sure about `evaluate`, this seems more a simulation tool. How about having this as standalone function that takes a `PhaseOracle` as input?

> What is the exact difference between QuantumCircuit subclass and Gate subclass, other than the actual class declaration?

`Gate`s are abstract blocks that the compiler can reason about on a higher level. We define a synthesis in the compilation process that best fits the hardware and the circuit context. `QuantumCircuit`s contain an eager definition, which makes it hard to resynthesize or do high-level optimizations. This layer of abstraction is why we move towards Gate for objects with a mathematical definition.

> Can we switch PhaseOracle to be Gate without additional deprecations?

No, we need to properly deprecate `PhaseOracle` in favor of `PhaseOracleGate`. We can't just change the underlying type since that would break user code that relies on `PhaseOracle` being a `QuantumCircuit`, that the `Gate` doesn't have. E.g. you wouldn't be able to attach new gates or directly pass it into a simulator.

> Also, PhaseOracle won't need an auxiliary qubit. Although indeed the conversion of bit-flip oracle to phase oracle uses such a qubit, it's not needed in direct synthesis from the function f(x) itself.

In this case, we would probably define `PhaseOracleGate` without an additional qubit and have the synthesis via bit-flip oracle use an auxiliary qubit. Or what do you think @alexanderivrii ?

alexanderivrii: From the docstrings,
* The phase oracle implements $|x\rangle \mapsto (-1)^{f(x)}|x\rangle$, and thus does not involve an additional qubit,
* The bitflip oracle implements $|x\rangle|y\rangle \mapsto |x\rangle|f(x)\oplus y\rangle$, and thus involves an additional "target" qubit $y$.

@Cryoris, I think the questions are: would you like `PhaseOracleGate` and `BitFlipOracleGate` be a part of this PR and/or would you like to pending-deprecate `PhaseOracle` as a part of this PR? 

Alternatively, maybe it's fine to introduce a `BitFlipOracle` quantum circuit (as this PR is doing), and add the two gates in a follow-up?

gadial: My concrete question is whether I can simply change the line `class BitFlipOracle(QuantumCircuit)` into `class BitFlipOracleGate(Gate)` and be done with it, or whether that means I should implement additional methods currently not implemented.

As for `evaluate_bitstring` I don't mind removing it from bit flip oracle, and after deprecating `PhaseOracle` in favor of `PhaseOracleGate` we can also drop it from `PhaseOracleGate`. It's not a functionality that a gate should supply, and anyway the user can always do `oracle.boolean_expression.simulate`.

qiskit/circuit/library/bit_flip_oracle.py
gadial: Done in [d94f490](https://github.com/Qiskit/qiskit/pull/13769/commits/d94f490eebf36759ec2f645516c72f0fdad746e2)

qiskit/circuit/library/bit_flip_oracle.py
ShellyGarion: If we are already introducing `BitFlipOracleGate` as part of this PR, should we also add `PhaseOracleGate` ?
(and `PhaseOracle` will be deprecated at Qiskit 3.0)

gadial: Seems reasonable.

gadial: Done in [991ded1](https://github.com/Qiskit/qiskit/pull/13769/commits/991ded1d0a05fd9237f86f0127aab95b3b557a51)

I think we can merge.

releasenotes/notes/boolean_expression_update-39c19edbbe71ba0d.yaml
```diff
@@ -0,0 +1,120 @@
...
+      q_3: ─■──■──■─
+                      
+features_circuits:
+  - |
+    The new :class:`.BitFlipOracleGate` and :class:`.PhaseOracleGate`
+    have the same interface as :class:`.PhaseOracle` 
+    (except the `evaluate_bitstring` method). Bit-flip oracle gate
+    synthesizes a bit flip oracle instead
+    of a phase flip oracle, meaning it acts on one additional qubit
+    and can be seen a applying a controlled X operation, where the
```
ShellyGarion: can be seen a --> can be seen as

releasenotes/notes/boolean_expression_update-39c19edbbe71ba0d.yaml
```diff
@@ -0,0 +1,120 @@
...
+    The :class:`.PhaseOracle` no longer relies on the `tweedledum` library
+    but might not be synthesized as effectively as before.
+    :class:`.BitFlipOracleGate` was added as an alternative to
+    directly synthesizing :class:`.BooleanExpression`, as this
+    class is removed in Qiskit 2.0. A :class:`.PhaseOracleGate`
+    was added and will replace :class:`.PhaseOracle` in Qiskit 3.0.
+
+    The interface of :class:`.PhaseOracle` was simplified; it no longer
+    accepts a `synthesizer` parameter, and the `expression` parameter
+    can only be a string; `ClassicalElement` has been deprecated.
```
ShellyGarion: deprecated in Qiskit 1.4.

releasenotes/notes/boolean_expression_update-39c19edbbe71ba0d.yaml
```diff
@@ -0,0 +1,120 @@
...
+
+    :class:`.PhaseOracle` is used exactly as before:
```
ShellyGarion, line 15: need a space here (after this line), since the code doesn't look well

qiskit/circuit/library/bit_flip_oracle.py
```diff
@@ -0,0 +1,127 @@
...
+
+    For convenience, this oracle, in addition to parsing arbitrary logical expressions,
+    also supports input strings in the `DIMACS CNF format
+    <https://web.archive.org/web/20190325181937/https://www.satcompetition.org/2009/format-benchmarks2009.html>`__,
+    which is the standard format for specifying SATisfiability (SAT) problem instances in
+    `Conjunctive Normal Form (CNF) <https://en.wikipedia.org/wiki/Conjunctive_normal_form>`__,
+    which is a conjunction of one or more clauses, where a clause is a disjunction of one
+    or more literals.
+    See :meth:`qiskit.circuit.library.bit_flip_oracle.BitFlipOracleGate.from_dimacs_file`.
+
```
ShellyGarion, line 48: would you like to add a link to the method here?

qiskit/circuit/library/phase_oracle.py
```diff
@@ -147,7 +125,112 @@ def from_dimacs_file(cls, filename: str):
...
+            |x\rangle \mapsto (-1)^{f(x)}|x\rangle
+
+    For convenience, this oracle, in addition to parsing arbitrary logical expressions,
+    also supports input strings in the `DIMACS CNF format
+    <https://web.archive.org/web/20190325181937/https://www.satcompetition.org/2009/format-benchmarks2009.html>`__,
+    which is the standard format for specifying SATisfiability (SAT) problem instances in
+    `Conjunctive Normal Form (CNF) <https://en.wikipedia.org/wiki/Conjunctive_normal_form>`__,
+    which is a conjunction of one or more clauses, where a clause is a disjunction of one
+    or more literals. See :meth:`qiskit.circuit.library.phase_oracle.PhaseOracleGate.from_dimacs_file`.
+
```
ShellyGarion, line 157: would you like to add a link to the method here?

qiskit/synthesis/boolean/boolean_expression.py
```diff
@@ -0,0 +1,223 @@
...
+
+from .boolean_expression_visitor import (
+    BooleanExpressionEvalVisitor,
+    BooleanExpressionArgsCollectorVisitor,
+)
+
+
+class TruthTable:
+    """A simple implementation of a truth table for a boolean function"""
+
```
ShellyGarion, line 38: would you like to add some explanation or example?

ShellyGarion (commented):
I have some minor comments on the docs. it's worth to build it on the main branch after 1.4 release to see that there are no conflicts or other problems.

qiskit/circuit/library/bit_flip_oracle.py
```diff
@@ -0,0 +1,127 @@
...
+        var_order: list[str] | None = None,
+    ) -> None:
+        """
+        Args:
+            expression: A Python-like boolean expression.
+            var_order: A list with the order in which variables will be created.
+               (default: by appearance)
+        """
+        self.boolean_expression = BooleanExpression(expression, var_order=var_order)
+        self.oracle = self.boolean_expression.synth(circuit_type="bit")
```
Cryoris: This runs the synthesis, right? If yes, this should be moved into `_define` to avoid eagerly synthesizing the circuit upon gate creation.

qiskit/circuit/library/bit_flip_oracle.py
```diff
@@ -0,0 +1,127 @@
...
+        )
+
+    def _define(self):
+        """
+        Defined by the synthesized bit-flip oracle
+        """
+        self.definition = self.oracle
+
+    @classmethod
+    def from_dimacs_file(cls, filename: str):
```
Cryoris: In case you'd like to include it:
```suggestion
    def from_dimacs_file(cls, filename: str) -> BitFlipOracleGate:
```

qiskit/circuit/library/phase_oracle.py
```diff
@@ -147,7 +125,112 @@ def from_dimacs_file(cls, filename: str):
         Returns:
             PhaseOracle: A quantum circuit with a phase oracle.
         """
-        from qiskit.circuit.classicalfunction.boolean_expression import BooleanExpression
+        expr = BooleanExpression.from_dimacs_file(filename)
+        return cls(expr)
+
+
+class PhaseOracleGate(Gate):
+    r"""Phase PhaseOracleGate.
```
Cryoris:
```suggestion
    r"""Implements a phase oracle.
```

qiskit/circuit/library/phase_oracle.py
```diff
@@ -147,7 +125,112 @@ def from_dimacs_file(cls, filename: str):
...
+        var_order: list[str] | None = None,
+    ) -> None:
+        """
+        Args:
+            expression: A Python-like boolean expression.
+            var_order: A list with the order in which variables will be created.
+               (default: by appearance)
+        """
+        self.boolean_expression = BooleanExpression(expression, var_order=var_order)
+        self.oracle = self.boolean_expression.synth(circuit_type="phase")
```
Cryoris: Same comment on the eager synthesis as above

qiskit/circuit/library/phase_oracle.py
```diff
@@ -147,7 +125,112 @@ def from_dimacs_file(cls, filename: str):
...
+    or more literals. See :meth:`qiskit.circuit.library.phase_oracle.PhaseOracleGate.from_dimacs_file`.
+
+    From 16 variables on, possible performance issues should be expected when using the
+    default synthesizer.
+    """
+
+    def __init__(
+        self,
+        expression: str,
+        var_order: list[str] | None = None,
```
Cryoris, line 165: Could you add an optional `label: str | None = None` here, which allows the user to overwrite the default label? All gates typically have this 🙂 (also for the `BitFlipOracleGate`)

qiskit/circuit/library/phase_oracle.py
```diff
@@ -147,7 +125,112 @@ def from_dimacs_file(cls, filename: str):
...
+        )
+
+    def _define(self):
+        """
+        Defined by the synthesized phase oracle
+        """
+        self.definition = self.oracle
+
+    @classmethod
+    def from_dimacs_file(cls, filename: str):
```
Cryoris:
```suggestion
    def from_dimacs_file(cls, filename: str) -> PhaseOracleGate:
```

test/python/circuit/library/test_phase_and_bitflip_oracles.py
```diff
@@ -0,0 +1,234 @@
...
+        result_valid = [isclose(statevector.data[state], valid_state) for state in states]
+
+        expected_invalid = [state not in good_states for state in states]
+        result_invalid = [isclose(statevector.data[state], invalid_state) for state in states]
+        self.assertListEqual(expected_valid, result_valid)
+        self.assertListEqual(expected_invalid, result_invalid)
+
+
+@ddt
+class TestPhaseOracleGate(QiskitTestCase):
```
Cryoris: Could you merge this test with the one for the circuit? They seem to be the same 🙂 You could e.g. do
```python
@data( ... )
@unpack
def test_statevector(self, expression, truth_table):
    for use_gate in [True, False]:
        oracle = PhaseOracleGate(expression) if use_gate else PhaseOracle(expression)
        # ... rest as before ...
        
        with self.subTest(use_gate=use_gate):
            self.assertListEqual( ... )
```

Cryoris (commented):
Thanks for all the effort, the result looks very nice! I have a few small comments, the main one about the fact that we don't synthesize `Gate` objects upon creation, but only when `_define` is called. Otherwise this LGTM 👍🏻

test/python/circuit/library/test_phase_and_bitflip_oracles.py
gadial: I actually prefer to avoid this since we are deprecating `PhaseOracle` and I want to keep is separate from `PhaseOracleGate`.

Cryoris: We're only going to remove it for 3.0 though, so it would be nice to have them together since bugfixes and new checks would likely need to cover both. But we can also keep them separate and copy these, if they occur in the future, in case you prefer 🙂

gadial: I'm convinced. Done in 454ecc5ee

qiskit/circuit/library/phase_oracle.py
gadial: Done in 454ecc5ee

qiskit/circuit/library/bit_flip_oracle.py
gadial: Done in 454ecc5ee

qiskit/synthesis/boolean/boolean_expression.py
gadial: Done in 454ecc5ee

Cryoris (approved):
LGTM, but I won't put it on the merge queue yet in case @ShellyGarion has further comments 🙂

PR #13774, opened 2025-01-31
Add `QuantumCircuit.noop` for instructionless qubit use

### LINKED ISSUE #13767: Add convenience method to include an idle bit in control-flow builders
### What should we add?

Currently in the control-flow builders, there is no simple way to cause the block to contain a qubit that idles for the duration of the block.  There is not always a major reason to want the additional data dependency on an idle qubit, but this becomes more useful for the upcoming `box`, where one might want to manually specify the grouping of certain idle qubits for the purposes of configuring error mitigation.

We can add a `QuantumCircuit.noop` method that accepts any `Qubit` specifier and considers them to be "used" by the current active block.  It is a standard `CircuitError` to `noop` a qubit that is not in the circuit, as it is for regular operations.

*edit*: medium priority because it's wanted by internal teams for demos, but not a 100% blocker.

jakelishman:
### Summary

This closes a usability gap between the low-level construction of control-flow ops and the builder interface.  In the builder interface, there was previously no way to consider a qubit to be "used" by the scope without adding some corresponding instruction on it.  It was possible to express this already by manually constructing the blocks.

In general, this is not so useful for the control-flow operations that we already have present because the additional dependency is spurious and simply stymies some ability to perform optimisations.  It is also a fair optimisation to remove the spurious data dependency in the transpiler.  It becomes more useful with the upcoming `box`, however; this has additional semantics around its incident data dependencies.

### Details and comments

Close #13767.

### DIFF
--- a/qiskit/circuit/quantumcircuit.py
+++ b/qiskit/circuit/quantumcircuit.py
@@ -769,6 +769,10 @@ class QuantumCircuit:
     ``with`` statement.  It is far simpler and less error-prone to build control flow
     programmatically this way.
 
+    When using the control-flow builder interface, you may sometimes want a qubit to be included in
+    a block, even though it has no operations defined.  In this case, you can use the :meth:`noop`
+    method.
+
     ..
         TODO: expand the examples of the builder interface.
 
@@ -779,6 +783,7 @@ class QuantumCircuit:
     .. automethod:: if_test
     .. automethod:: switch
     .. automethod:: while_loop
+    .. automethod:: noop
 
 
     Converting circuits to single objects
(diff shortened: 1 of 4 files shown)


jakelishman:
Medium priority because it's part of the high-priority `box` epic, but technically not an absolutely hard requirement for fulfilment of the minimal viable implementation.

qiskit/circuit/quantumcircuit.py
```diff
@@ -6176,6 +6181,42 @@ def unitary(
...
+        any additional operations on it.
+
+        For example::
+
+            from qiskit.circuit import QuantumCircuit
+
+            qc = QuantumCircuit(3)
+            with qc.box():
+                # This control-flow block will only use qubits 0 and 1.
+                qc.cx(0, 1)
```
alexanderivrii, line 6199: To make sure, we don't have the functionality `qc.box()` is Qiskit yet, correct? So this code example would not be runnable, would this be a problem?

alexanderivrii (commented):
This PR is well-motivated and well-executed. My only (minor) question is whether we should use "future" examples in the documentation. @jakelishman. are you planning to add the `QuantumCircuit.box` method for `2.0.0`?

qiskit/circuit/quantumcircuit.py
jakelishman: Oh yeah sorry, I wrote the example speculatively, but it's written against the spec of Qiskit/RFCs#76.  I can change it to `if_test` for now if you like, but `box` is the bigger motivation.

jakelishman:
Yeah, `box` is a high-priority item for the 2.0 release - I'll be posting a PR for it in relatively short order.

qiskit/circuit/quantumcircuit.py
alexanderivrii: Thanks, since `box` will make it into 2.0 release, I am completely happy to keep this example as is.

alexanderivrii (approved):

PR #13777, opened 2025-02-02
Fixing definition method for ``PyGate`` and ``PyInstruction``

alexanderivrii:
### Summary

This commit fixes a problem when calling ``definition`` for a ``PyGate`` and ``PyInstruction``. Update: for ``UnitaryGate`` we have decided to returning ``None`` is the best option.

### Details and comments

I have stumbled on this problem while working on porting ``HighLevelSynthesis`` to Rust, with the ``definition`` method for a custom gate incorrectly returning ``None`` (instead of the expected ``CircuitData``), e.g., on the following circuit:
```
qc_inner = QuantumCircuit(3)
qc_inner.cx(0, 2)
qc_inner.h(1)

qc = QuantumCircuit(5)
qc.x(0)
qc.append(qc_inner.to_gate(), [1, 4, 2])
```

Looking at the code for ``PyGate``
```
match self.gate.getattr(py, intern!(py, "definition")) {
    Ok(definition) => {
        let res: Option<PyObject> = definition.call0(py).ok()?.extract(py).ok();
        ...
```                    
the problem is that ``circuit.definition`` [in the Python space] is a ``QuantumCircuit`` and not a callable, with ``definition.call0(...)`` returning precisely this error message. 

The only question is how can I add a test for this fix? I don't see how I can add something to the Rust space and I don't see how I can add it to the Python space.


Cryoris (commented):
Going via `QuantumCircuitData` seems like it would work (though I would ask whether the case of `definition` being `None` still returns `None` with this new solution?) but this will populate a lot fields we don't return in the end (since we only return `.data` but e.g. `.qregs` and many others are being built). 

Couldn't we fix the previous code by removing the `.call0` and directly trying to cast the definition into a `CircuitData`? This way we'd only build the data we actually return from the function.

alexanderivrii:
Thanks @Cryoris: the fix looks even better after applying your suggestion to avoid constructing the full `QuantumCircuitData`. It should indeed return `None` when the definition is `None`. 

Unfortunately, creating a test from within Rust space seems impossible before the problems mentioned in https://github.com/Qiskit/qiskit/pull/13776 are solved. The only thing I managed to do is creating a regular function in `operations.rs` that does the checking, exposing it to Python, and calling it from one of the python tests, but this solution is horrendously ugly [so not included in this fix].

crates/circuit/src/operations.rs
```diff
@@ -2860,7 +2848,19 @@ impl Operation for UnitaryGate {
         }
     }
     fn definition(&self, _params: &[Param]) -> Option<CircuitData> {
-        None
+        Python::with_gil(|py| -> Option<CircuitData> {
+            let py_op = self
+                .create_py_op(py, &ExtraInstructionAttributes::default())
+                .unwrap();
```
alexanderivrii: I am a bit unsure whether I should pass ``ExtraInstructionAttributes::default()``, or modify the ``definition`` method to include the ``extra_attrs: &ExtraInstructionAttributes`` which can then be passed to ``create_py_op``. Note that this changes the signature of the ``definition`` method and I would need to propagate it to other functions. I am also a bit unsure about the ``unwrap`` here, should I change the returned value of the function to ``Result`` or would it be better to return ``None`` when ``create_py_op`` produces an ``Err``  variant?

crates/circuit/src/operations.rs
```diff
@@ -2860,7 +2848,19 @@ impl Operation for UnitaryGate {
...
+            match py_op.getattr(py, intern!(py, "definition")) {
+                Ok(definition) => definition
+                    .getattr(py, intern!(py, "_data"))
+                    .ok()?
+                    .extract::<CircuitData>(py)
+                    .ok(),
+                Err(_) => None,
```
mtreinish: Do we need a definition for UnitaryGate at all? I purposely didn't define it here because the python space one is to call unitary synthesis. In rust it's a bit of a layer violation to do that since it lives in the accelerate crate. While going through python avoids that issue, I'm wondering if it's really necessary because it's not a definition in the strict sense like it is for a concrete gate.

alexanderivrii: I need this to handle certain corner-cases arising when porting `HighLevelSynthesis` to Rust. Even if `HighLevelSynthesis` runs right after `UnitarySynthesis`, there may be gates whose plugins / definitions include `UnitaryGates`, which then need to be recursively synthesized (this is for instance the case for `Isometry` gates). The current `HighLevelSynthesis` code in Python handles this by extracting `definition` for these unitary gates, and proceeding recursively, so it made sense for the Rust code to do so as well. I would be happy to hear alternative approaches.

mtreinish: Why not just call the two qubit decomposer on the matrix if you encounter a unitary gate in hls during recursion? Realistically that's what you are doing here, just via a very inefficient path that goes back and forth through python.

alexanderivrii: To make sure I understand this correctly, for 1- or 2- qubit unitaries we can use the decomposers already available in rust, and so we can completely avoid going through python; but for larger unitaries the `_define` method of `UnitaryGate` uses `qs_decomposition`, which is only available in python, and so we would still have to go through the python space.

mtreinish: Sure, we will have to rely on qsd in python for 3 or more qubit `UnitaryGate` instances. But that's easy to both encapsulate to only use python when we need it and also substitute the python usage for a rust function if/when we have an implementation of qsd in rust. You can just do something like:

```rust
if matches!(inst.view(), OperationRef::Unitary(gate) {
    match gate.num_qubits() {
        0 => panic!(),
        1 => call_euler_decomposer(),
        2 => call_two_qubit_basis_decomposer_with_cx(),
        _ => py_call_qsd(),
    }
}
```

in hls.

alexanderivrii: This makes sense. So I should revert the change to handle unitary gates? Alternatively, I could close this PR, and only keep the HLS PR, since it's the only consumer for the definition method.

mtreinish: I think having a standalone PR to fix definition is good. Just revert the unitary gate piece and I think we can merge this.

alexanderivrii: Reverted in [5079541](https://github.com/Qiskit/qiskit/pull/13777/commits/50795419a05907c01982f4b9488c76fe0c0a3734).

crates/circuit/src/operations.rs
alexanderivrii: This question is no longer relevant.

mtreinish (approved):
LGTM thanks for fixing this.

Issue #13768, opened 2025-01-30
Control-flow routing causes idle qubits to be stripped from control-flow blocks


jakelishman:
### What should we add?

Any qubits that are idling within a control-flow block will be stripped from the block if routing runs on it.  This is a side effect of how control-flow routing is done; we first expand the block to all physical qubits, recurse the router, then contract it back to non-idling qubits.

This can in many ways be considered a fair optimisation, especially for the control-flow operations currently implemented, since it reduces spurious data dependencies.  However, it does change the semantics of the circuit in a way that is not entirely appropriate for routing to do, and it becomes more of a problem when explicit no-ops are given higher status as in #13767.  For the upcoming `box`, removing the qubit dependency is a violation of the contract, since the data dependency is a part of the `box` semantics (certainly for some "verbatim"-like annotations).

There are two steps to this:
- make the routers contract to at least the physical qubits corresponding to incident virtual qubits that were originally incident, plus any additional physical qubits now touched by routing, such that the virtual-qubit dependencies on input and output are non-strict supersets of the pre-routing state.
- add a transpiler pass to relevant optimisation loops that performs legal contractions of control-flow operation bodies (probably both abstract and hardware-lowered optimisations).

PR #13779, opened 2025-02-03
Add `ContractIdleWiresInControlFlow` optimisation pass

jakelishman:
### Summary

This transpiler pass removes data dependencies on idle qubits from control-flow operations.  For example, given a circuit such as::

    qc = QuantumCircuit(1, 1)
    qc.x(0)
    with qc.if_test((qc.clbits[0], True)):
        qc.x(0)
        qc.x(0)
    qc.x(0)

the current optimisation passes will collapse the inner control-flow block to the identity, but the qubit dependency will remain, preventing the outer two X gates from being cancelled.  This pass removes the now-spurious dependency, making it possible to detect and remove the two X gates in a follow-up loop iteration.

As an accidental side-effect of their algorithms, the control-flow-aware routing passes currently do this when they run.  This aims to move the logic into a suitable place to run before routing (so the spurious dependency never arises in routing in the first place) and in the low-level optimisation stage.

The aim of this pass is also to centralise the logic, so when the addition of the new `box` scope with different semantics around whether a wire is truly idle in the box or not, the routers aren't accidentally breaking them, and it's clearer when the modifications happen.

### Details and comments

Built on top of #13774, since it's very convenient to have that method for writing the tests.  This is one part of #13768.

I haven't hooked this up to the preset pass-managers in this commit; I wanted to do that after #13768 is fixed properly and potentially once `box` itself has landed.


jakelishman:
Rebased over main and no longer on hold.

qiskit/transpiler/passes/optimization/contract_idle_wires_in_control_flow.py
```diff
@@ -0,0 +1,104 @@
...
+from qiskit.dagcircuit import DAGCircuit, DAGOpNode
+from qiskit.transpiler.basepasses import TransformationPass
+
+
+class ContractIdleWiresInControlFlow(TransformationPass):
+    """Remove idle qubits from control-flow operations of a :class:`.DAGCircuit`."""
+
+    def run(self, dag):
+        # `control_flow_op_nodes` is eager and doesn't borrow; we're mutating the DAG in the loop.
+        for node in dag.control_flow_op_nodes() or []:
```
kevinhartman, line 25: You'll probably remember, but adding a note here to remove the `or []` once it is no longer needed.

qiskit/transpiler/passes/optimization/contract_idle_wires_in_control_flow.py
```diff
@@ -0,0 +1,104 @@
...
+                if isinstance(wire, Qubit):
+                    qubits[wire] = wire
+                elif isinstance(wire, Clbit):
+                    clbits[wire] = wire
+                else:
+                    vars_[wire] = wire
+            replacement.add_qubits(list(qubits))
+            replacement.add_clbits(list(clbits))
+            for var in vars_:
+                replacement.add_captured_var(var)
```
kevinhartman, line 45: Do we need to add local variables too?

qiskit/transpiler/passes/optimization/contract_idle_wires_in_control_flow.py
```diff
@@ -0,0 +1,104 @@
...
+            replacement._apply_op_node_back(DAGOpNode.from_instruction(new_inst))
+            # The replacement DAG is defined over all the same qubits, but with the correct
+            # qubits now explicitly marked as idle, so everything gets linked up correctly.
+            dag.substitute_node_with_dag(
```
kevinhartman: I was a bit confused at first reading through this until I realized that `replacement` is a temporary DAG that gets its contents unwrapped / flattened into `dag`. Do we not have a simpler way to replace a node with another node in `DAGCircuit`? Presumably there's more going on here than I'm understanding!

test/python/transpiler/test_contract_idle_wires_in_control_flow.py
```diff
@@ -0,0 +1,183 @@
...
+
+from test import QiskitTestCase  # pylint: disable=wrong-import-order
+
+
+class TestContractIdleWiresInControlFlow(QiskitTestCase):
+    def test_simple_body(self):
+        qc = QuantumCircuit(3, 1)
+        with qc.while_loop((qc.clbits[0], False)):
+            qc.cx(0, 1)
+            qc.noop(2)
```
kevinhartman, line 27: Certainly beyond the scope of this PR, but I find it a bit strange that a `noop` doesn't add an operation (I did not see the introductory PR for it). I was surprised to see that there's no `noop` in the `expected` circuit of this test, and thought I may have missed something in this new pass where they get removed.

test/python/transpiler/test_contract_idle_wires_in_control_flow.py
```diff
@@ -0,0 +1,183 @@
...
+            with qc.if_test((cr2, 3)):
+                qc.noop(0, 1, 2)
+        expected = QuantumCircuit(qr, cr1, cr2)
+        with expected.if_test((cr1, 3)):
+            with expected.if_test((cr2, 3)):
+                pass
+
+        actual = ContractIdleWiresInControlFlow()(qc)
+        self.assertNotEqual(qc, actual)  # Smoke test.
+        self.assertEqual(actual, expected)
```
kevinhartman, line 183: Perhaps it'd be nice to have a test like the one in this PR's description that demonstrates commutation along a now-idle bit through a control flow op?

kevinhartman (commented):
This looks good! It's nice to have this logic in one place.

Presumably you were missing something in Rust and thus decided to do the initial implementation in Python?

test/python/transpiler/test_contract_idle_wires_in_control_flow.py
kevinhartman: Ah, I see you have such a test here: https://github.com/Qiskit/qiskit/pull/13891/files#diff-df5e0c6b04d9ea56420653f11b66b75d78856b979e1e43abe8cf4e1b78ed9a4aR2255

qiskit/transpiler/passes/optimization/contract_idle_wires_in_control_flow.py
jakelishman: The trick is around the node contraction - `substitute_node_with_dag` has enough information available to it that it can link the now-unused wires together.  We have `substitute_node`, but it requires that the two nodes act on the same set of wires.

test/python/transpiler/test_contract_idle_wires_in_control_flow.py
jakelishman: The entire point of `noop` is _not_ to add an operation haha.  `id` is the mathematical expression of "unitary identity", whereas `noop` is "mark this qubit as used, but do explicitly do not do anything with it".

qiskit/transpiler/passes/optimization/contract_idle_wires_in_control_flow.py
jakelishman: Good to highlight thanks, because it's now been long enough that I probably would have forgotten to do it haha.

jakelishman:
I did this in Python because it was faster and cleaner to write for the initial implementation, and there won't be any appreciable slowdown because all the logic is Python-space anyway - control-flow ops are still represented by Python-space objects.

The loop `dag.control_flow_op_nodes()` happens in Rust space, which is the only bit where the Python overhead would have hurt, I think.

qiskit/transpiler/passes/optimization/contract_idle_wires_in_control_flow.py
jakelishman: In the replacement they're all "capturing" from the circuit they're about to be inlined into - the outer circuit defines the true "type" of them.

qiskit/transpiler/passes/optimization/contract_idle_wires_in_control_flow.py
kevinhartman: Ah, thanks for clarifying 🙂.

test/python/transpiler/test_contract_idle_wires_in_control_flow.py
kevinhartman: Heh, I suppose I'm thinking of it more like a noop instruction in classical computing, which does end up getting executed by the CPU and takes some number of cycles to do nothing.

kevinhartman (dismissed):

test/python/transpiler/test_contract_idle_wires_in_control_flow.py
jakelishman: It kind of does here too, but qubit operations are all inherently super parallel.  Putting a `noop` on a qubit to bring it into a box causes it to idle for the duration of the box (unless you apply other things to it).

qiskit/transpiler/passes/optimization/contract_idle_wires_in_control_flow.py
```diff
@@ -0,0 +1,104 @@
...
+                node, replacement, wires=qubits | clbits | vars_, propagate_condition=False
+            )
```
jakelishman:
```suggestion
            dag.substitute_node_with_dag(node, replacement, wires=qubits | clbits | vars_)
```

kevinhartman (approved):

PR #13782, opened 2025-02-03
Have `BlueprintCircuit.copy_empty_like` return an empty `QuantumCircuit`

### LINKED ISSUE #13535: Unexpected behavior of `copy_empty_like()` for `BlueprintCircuit`s
### Environment

- **Qiskit version**: 1.3.0
- **Python version**: 3.12.5
- **Operating system**: MacOS

### What is happening?

`QuantumCircuit.copy_empty_like()` copies `RealAmplitudes` instruction to the clone.

```
>>> from qiskit.circuit.library import RealAmplitudes
>>> circuit = RealAmplitudes(num_qubits=2, reps=2)
>>> circuit.draw()
     ┌────────────────────────────────────────────────┐
q_0: ┤0                                               ├
     │  RealAmplitudes(θ[0],θ[1],θ[2],θ[3],θ[4],θ[5]) │
q_1: ┤1                                               ├
     └────────────────────────────────────────────────┘
>>> circuit.copy_empty_like().draw()
     ┌────────────────────────────────────────────────┐
q_0: ┤0                                               ├
     │  RealAmplitudes(θ[0],θ[1],θ[2],θ[3],θ[4],θ[5]) │
q_1: ┤1                                               ├
     └────────────────────────────────────────────────┘
>>> circuit.copy_empty_like() == circuit
True
>>> 

```

### What should happen?

The copy should not have a gate

### Any suggestions?

_No response_

Cryoris:
### Summary

Fixes #13535.

### Details and comments

The `BlueprintCircuit.copy_empty_like` method currently returns another instance of a `BlueprintCircuit` with it's definition wiped. That means that the definition would just be re-built if any circuit method was called, and the circuit wasn't really empty. As discussed in #13535 fixes this by returning a `QuantumCircuit` type, not a `BlueprintCircuit` anymore.

### DIFF
--- a/test/python/circuit/library/test_blueprintcircuit.py
+++ b/test/python/circuit/library/test_blueprintcircuit.py
@@ -24,7 +24,7 @@ from qiskit.circuit import (
     Instruction,
     CircuitInstruction,
 )
-from qiskit.circuit.library import BlueprintCircuit, XGate
+from qiskit.circuit.library import BlueprintCircuit, XGate, EfficientSU2
 from test import QiskitTestCase  # pylint: disable=wrong-import-order
 
 
@@ -208,6 +208,34 @@ class TestBlueprintCircuit(QiskitTestCase):
             parametric.copy().assign_parameters({a: math.pi / 2}).global_phase,
             math.pi / 2,
         )
+
+    def test_copy_empty_like(self):
+        """Test copying empty-like.
+
+        Regression test of #13535.
+        """
+        circuit = EfficientSU2(2)
+        circuit.global_phase = -2
+        circuit.metadata = {"my_legacy": "i was a blueprintcircuit once"}
+
+        cpy = circuit.copy_empty_like()
+        expected = QuantumCircuit(
+            circuit.num_qubits, global_phase=circuit.global_phase, metadata=circuit.metadata
+        )
+        self.assertEqual(cpy, expected)
+        self.assertNotIsInstance(cpy, BlueprintCircuit)
+
+    def test_copy_empty_like_post_build(self):
+        """Test copying empty-like after building the circuit."""
+        circuit = EfficientSU2(2)
+        _ = circuit.count_ops()  # trigger building the circuit
+
+        cpy = circuit.copy_empty_like()
+
+        circuit.num_qubits = 3  # change the original circuit
+
+        expected = QuantumCircuit(2)  # we still expect an empty 2-qubit circuit
+        self.assertEqual(cpy, expected)
 
 
 if __name__ == "__main__":
--- a/test/python/circuit/library/test_blueprintcircuit.py
+++ b/test/python/circuit/library/test_blueprintcircuit.py
@@ -219,23 +219,30 @@ class TestBlueprintCircuit(QiskitTestCase):
         circuit.metadata = {"my_legacy": "i was a blueprintcircuit once"}
 
         cpy = circuit.copy_empty_like()
+        cpy.x(0)  # if it were a BlueprintCircuit, this would trigger the building
+
         expected = QuantumCircuit(
             circuit.num_qubits, global_phase=circuit.global_phase, metadata=circuit.metadata
         )
+        expected.x(0)
+
         self.assertEqual(cpy, expected)
         self.assertNotIsInstance(cpy, BlueprintCircuit)
 
     def test_copy_empty_like_post_build(self):
         """Test copying empty-like after building the circuit."""
         circuit = EfficientSU2(2)
-        _ = circuit.count_ops()  # trigger building the circuit
+        num_params = circuit.num_parameters  # trigger building the circuit
 
         cpy = circuit.copy_empty_like()
+        cpy_num_params = cpy.num_parameters  # if it were a BP circuit, this would trigger the build
 
         circuit.num_qubits = 3  # change the original circuit
 
         expected = QuantumCircuit(2)  # we still expect an empty 2-qubit circuit
         self.assertEqual(cpy, expected)
+        self.assertEqual(cpy_num_params, 0)
+        self.assertEqual(num_params, 16)
 
 
 if __name__ == "__main__":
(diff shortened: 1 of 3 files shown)


test/python/circuit/library/test_blueprintcircuit.py
```diff
@@ -209,6 +209,34 @@ def _build(self):
...
+        circuit = EfficientSU2(2)
+        circuit.global_phase = -2
+        circuit.metadata = {"my_legacy": "i was a blueprintcircuit once"}
+
+        cpy = circuit.copy_empty_like()
+        expected = QuantumCircuit(
+            circuit.num_qubits, global_phase=circuit.global_phase, metadata=circuit.metadata
+        )
+        self.assertEqual(cpy, expected)
+        self.assertNotIsInstance(cpy, BlueprintCircuit)
```
jakelishman: I don't think the metadata is included in the equality check, so I don't think we'll actually be testing that.

I think we need to trigger the build of `circuit` and do something that would trigger the build of `cpy` (if it were a `BlueprintCircuit`) as well like adding a single gate to it - the `not isinstance` check is ok, but it's a round-about way of testing the behaviour we actually want, which is "the circuit should actually be empty".

test/python/circuit/library/test_blueprintcircuit.py
```diff
@@ -209,6 +209,34 @@ def _build(self):
...
+        """Test copying empty-like after building the circuit."""
+        circuit = EfficientSU2(2)
+        _ = circuit.count_ops()  # trigger building the circuit
+
+        cpy = circuit.copy_empty_like()
+
+        circuit.num_qubits = 3  # change the original circuit
+
+        expected = QuantumCircuit(2)  # we still expect an empty 2-qubit circuit
+        self.assertEqual(cpy, expected)
```
jakelishman: It's good to test mutating `circuit` to ensure no modifications sneak through in shared state, but it's the `_build` method getting called on `cpy` which would be the full test, I think?

jakelishman (commented):
thanks, I hate it (lol)

This definitely does seem to be the right way to go - it's a very awkward interaction with a method called `copy_empty_like` and the `BlueprintCircuit`, because you _could_ argue that you'd expect to receive an unbuilt object of the same class, but in practice, I think (almost) nobody actually _intends_ to use the `BlueprintCircuit` for the blueprint-y stuff, it just gets used because of an implementation detail of the circuit library.

Regardless, 2.0 is probably the right time to make that change.

jakelishman:
I've untagged the "stable backport potential" for now, unless there's a pressing reason to really want to backport this, because it does feel like changing one _technically_ valid interpretation of `copy_empty_like` into another.  We're calling it a "bugfix", but it _could_ be considered an API break, so unless it's critical to backport (which I doubt, since I think this has been around for ever), I think the backport risk is greater than the benefit.

test/python/circuit/library/test_blueprintcircuit.py
Cryoris: There should be no `_build` on `cpy` since that's just a plain `QuantumCircuit` 🙂

test/python/circuit/library/test_blueprintcircuit.py
Cryoris: The `assertEqual` check should trigger the build -- with the old version of the `copy_empty_like` that test failed 🙂 The final instance check was just to be very clear it is no longer a blueprint circuit 😛 I can add a method call that would trigger the build explicitly, if it were a blueprint circuit

Cryoris:
> This definitely does seem to be the right way to go - it's a very awkward interaction with a method called copy_empty_like and the BlueprintCircuit, because you could argue that you'd expect to receive an unbuilt object of the same class, but in practice, I think (almost) nobody actually intends to use the BlueprintCircuit for the blueprint-y stuff, it just gets used because of an implementation detail of the circuit library.

Yep, the `copy_empty_like` is very confusing here -- I'd also think it works correctly now, it's just very unintuitive and very unexpected if one is expecting a plain circuit 😅 But I also don't care so much since we'll deprecate the blueprint anyways... 

> I've untagged the "stable backport potential" for now, unless there's a pressing reason to really want to backport this, because it does feel like changing one technically valid interpretation of copy_empty_like into another. We're calling it a "bugfix", but it could be considered an API break, so unless it's critical to backport (which I doubt, since I think this has been around for ever), I think the backport risk is greater than the benefit.

Sounds good, it does feel more like an API change than a bugfix tbh -- if we don't backport it, I'd be happy to just label it that.

test/python/circuit/library/test_blueprintcircuit.py
jakelishman: Yeah I know, because you've fixed it, but the point is that this test doesn't _necessarily_ test the regression behaviour, because there's no code in here that super obviously would have caused a build if your PR didn't actually fix the problem.  `__eq__` _might_ do it accidentally, but you could easily imagine an optimisation in `BlueprintCircuit` that would cause it not to.

Might be good to do something to `cpy` which is API guaranteed to cause a build _if it were_ a `BlueprintCircuit` before testing for emptiness.

test/python/circuit/library/test_blueprintcircuit.py
jakelishman: (I answered the other chain before I read this one.)

Ok, if `assertEqual` did actually trigger the build, then I guess it is testing it.  I guess my now-implied point is that it wasn't obvious to me that `assertEqual` would necessarily have triggered the build, so it wasn't 100% clear to me that the test was correct, and having that certainty is a nice property of tests.

Cryoris: I'll add a bit to ensure it would trigger the building and add a comment 🙂

Cryoris:
I added explicit triggers to the building process (with comments) and changed to `API Upgrade` in c3bd386 🙂

jakelishman (approved):
Solid, this looks much more obvious to me now, thanks!

PR #13783, opened 2025-02-03
Add method to compute estimated duration of scheduled circuit

mtreinish:
### Summary

This commit adds a new QuantumCircuit method to compute the estimated duration of a scheduled circuit. This is to replace the deprecated duration attribute that the transpiler was potentially setting during the scheduling stage. This method computes the longest duration path in the dag view of the circuit internally.

This method should have been included in the 1.2.0 release prior to the deprecation of the `QuantumCircuit.duration` attribute in 1.3.0. But, this was an oversight in the path to deprecation, as it was part of larger deprecation of numerous scheduling pieces in the 1.3.0. We should definitely backport this for the 1.4.0 release for inclusion in that release prior to the Qiskit 2.0.0 release which removes the deprecated attribute

### Details and comments

TODO:

- [x] Add tests (and fix issues as this is untested so far)


mtreinish:

crates/accelerate/src/circuit_duration.rs
```diff
@@ -0,0 +1,105 @@
...
+use rustworkx_core::petgraph::visit::{EdgeRef, IntoEdgeReferences};
+
+/// Estimate the duration of a scheduled circuit in seconds
+#[pyfunction]
+pub(crate) fn compute_estimated_duration(dag: &DAGCircuit, target: &Target) -> PyResult<f64> {
+    let dt = target.dt;
+
+    let get_duration =
+        |edge: <&StableDiGraph<NodeType, Wire> as IntoEdgeReferences>::EdgeRef| -> PyResult<f64> {
+            let node_weight = &dag.dag()[edge.target()];
```
jakelishman: You can just do `&dag[edge.target()]` now

qiskit/circuit/quantumcircuit.py
```diff
@@ -904,8 +907,9 @@ class QuantumCircuit:
 
     If a :class:`QuantumCircuit` has been scheduled as part of a transpilation pipeline, the timing
     information for individual qubits can be accessed.  The whole-circuit timing information is
-    available through the :attr:`duration`, :attr:`unit` and :attr:`op_start_times` attributes.
+    available through the :meth:`estimate_duration` method and :attr:`op_start_times` attribute.
```
jakelishman, line 915: Are we leaving `op_start_times` as a valid attribute of `QuantumCircuit` post deprecation?

qiskit/circuit/quantumcircuit.py
```diff
@@ -6877,6 +6881,46 @@ def qubit_stop_time(self, *qubits: Union[Qubit, int]) -> float:
 
         return 0  # If there are no instructions over bits
 
+    def estimate_duration(self, target, unit: str = "s") -> int | float:
+        """Estimate the duration of a scheduled circuit
+
+        This method computes the estimate of the circuit duration by finding
+        the longest duration path in the circuit based on the durations
+        provided by a given target.
```
jakelishman: We should include some description in this of the limitations of the function, beyond just the summary in the "raises" documentation - something like "this only works for simple circuits with no control flow or other classical feed-forwards operations".

crates/accelerate/src/circuit_duration.rs
```diff
@@ -0,0 +1,105 @@
...
+            match node_weight {
```
jakelishman: The nesting's got a bit out of control in this function - I think we reach 11 levels of indentation at the deepest point?  Some of the `if let`s might want to become `let ... else` or anything else we can do.

jakelishman (commented):
Thanks for doing this.

This feels like we probably ought to have a migration guide or something to both explain ourselves a bit more and to include a couple more worked examples to help the transition?

qiskit/circuit/quantumcircuit.py
mtreinish: I didn't think that we deprecated it as part of 1.3.0. There are also a bunch of associated methods with it too. We can deprecate it for 1.4.0 as falls into the same category as duration though and we should probably remove it from the circuit object. But we can look at that in a separate pr.

1ucian0 (commented):
~`compute_estimated_duration` sounds like a verbose name. @nonhermitian suggested `estimate_duration` and I think it is much better.~ ups.. it was already that name. my bad.

crates/accelerate/src/circuit_duration.rs
mtreinish: Done in: https://github.com/Qiskit/qiskit/pull/13783/commits/6c56c3eca4d03a326002c1b570b0d2391476d22b

qiskit/circuit/quantumcircuit.py
mtreinish: Added in: https://github.com/Qiskit/qiskit/pull/13783/commits/af6c739743060c3275fe418c84c12dab61830577

crates/accelerate/src/circuit_duration.rs
mtreinish: I dropped an indent level as part of refactoring for the rust standard instruction in: https://github.com/Qiskit/qiskit/pull/13783/commits/0fe8e0d6814d0b138a1f06e138eb6f64b696cc31 (I'll have to rewrite it for 1.4 backport though since StandardInstruction is only in 2.0).

mtreinish:
This should be ready for review now. I'll have to rewrite the rust implementation when it's backported to 1.4 because the rust internals on `main` are different than in 1.3, but it shouldn't be too hard it's basically the implementation before: https://github.com/Qiskit/qiskit/pull/13783/commits/0fe8e0d6814d0b138a1f06e138eb6f64b696cc31

qiskit/circuit/quantumcircuit.py
mtreinish: I looked into doing this and don't think we should for 2.0. The timeline visualizer depends on this data and methods. I need to figure out a more complete mechanism to rewrite the drawer without this, and rushing it in the week of 1.4.0 doesn't seem ideal. We should update the docstrings in 2.0 though to make it more explicit that this is an estimate etc, like what this method does for the duration.

ElePT: I added an issue to keep track of the docstring update: https://github.com/Qiskit/qiskit/issues/13880, it's very small but I am afraid that without it we might forget to do it

ElePT (approved):
LGTM! My only concern here is how to make sure that this new utility reaches users I think that at least we can make sure to include it in the 2.0 migration guide (added a [comment](https://github.com/Qiskit/documentation/issues/2632#issuecomment-2668012412) in the issue).

PR #13785, opened 2025-02-04
Correctly updating global phase when removing gates that are identity up to a global phase

### LINKED ISSUE #13778: incorrect statevector simulation using qiskit & qiskit-aer
### Environment

- **Qiskit version**: 1.3.0 (I believe the bug still exists in 1.3.2)
- **Python version**: 3.12
- **Operating system**: linux

mirror ticket from https://github.com/Qiskit/qiskit-aer/issues/2271 for visibility

### What is happening?

For a fixed QuantumCircuit with custom unitary gate matrix, QiskitAer is producing different statevector simulation results for qiskit v.1.2.4 and v.1.3.0

Using `qiskit-aer=0.15.1`, run the following script with `qiskit==1.2.4` and `qiskit==1.3.0` will reproduce different statevector results
```python
import qiskit

import numpy as np

def get_qiskit_unitary_gate(rng, control=None):
    rand_unitary = random_unitary(4, seed=rng)
    gate = UnitaryGate(rand_unitary)
    print(f"unitary gate matrix:")
    print(gate)
    if control is None:
        return gate
    else:
        return gate.control(control)
    
def get_qiskit_multi_control_circuit():
    qubits = qiskit.QuantumRegister(5)
    circuit = qiskit.QuantumCircuit(qubits)
    for q in qubits:
        circuit.h(q)
    qs = list(qubits)

    rng = np.random.default_rng(1234)
    for i in range(2):
        rng.shuffle(qs)
        ccu_gate = get_qiskit_unitary_gate(rng, control=2)
        print(f"{i}th gate on {qs}")
        circuit.append(ccu_gate, qs[:4])
    circuit.p(0.1, qubits[0])
    return circuit

print(f"{qiskit.__version__=}")

circuit = get_qiskit_multi_control_circuit()
print(circuit)

circuit.save_statevector()

# Transpile for simulator
simulator = AerSimulator(method='statevector')
circ = qiskit.transpile(circuit, simulator)

# Run and get statevector
result = simulator.run(circ).result()
sv = np.asarray(result.get_statevector(circ))
print("Sum of state vector:", sv.sum())
```

### What should happen?

The gate matrices defining the custom unitary gates are identical between `qiskit==1.2.4` and `qiskit==1.3.0` and therefore the returned statevector should be the same. On my end I have the following results:

`qiskit==1.2.4`
```
qiskit.__version__='1.2.4'
unitary gate matrix:
Instruction(name='unitary', num_qubits=2, num_clbits=0, params=[array([[ 0.26093746+0.18318812j,  0.12558674-0.38936694j,
         0.56486648-0.53659333j, -0.03587912+0.35025885j],
       [-0.5208327 +0.41955884j,  0.38293134-0.20373037j,
        -0.26997299-0.01611016j,  0.47100825+0.26375647j],
       [-0.18047959-0.0114164j ,  0.42182388+0.05310412j,
         0.01892911-0.48459796j, -0.10334574-0.73530097j],
       [-0.44557601+0.46884564j, -0.63238151+0.25262043j,
         0.15882664-0.24378088j, -0.17446941-0.05317653j]])])
0th gate on [Qubit(QuantumRegister(5, 'q1'), 0), Qubit(QuantumRegister(5, 'q1'), 4), Qubit(QuantumRegister(5, 'q1'), 1), Qubit(QuantumRegister(5, 'q1'), 3), Qubit(QuantumRegister(5, 'q1'), 2)]
unitary gate matrix:
Instruction(name='unitary', num_qubits=2, num_clbits=0, params=[array([[-0.41490662-0.1027918j , -0.45229939-0.45892536j,
         0.1965442 +0.27907236j,  0.53283662+0.04090535j],
       [-0.54960898-0.38886747j, -0.18060452+0.35532657j,
        -0.42299907-0.41452735j,  0.00811512+0.19238008j],
       [-0.48710529+0.16105765j,  0.30802545-0.50934349j,
         0.16188328-0.37911161j, -0.36497996-0.28166686j],
       [ 0.16591044+0.27066769j,  0.26600357-0.02948626j,
         0.10303047-0.5934482j ,  0.6357136 +0.24628756j]])])
1th gate on [Qubit(QuantumRegister(5, 'q1'), 1), Qubit(QuantumRegister(5, 'q1'), 4), Qubit(QuantumRegister(5, 'q1'), 2), Qubit(QuantumRegister(5, 'q1'), 0), Qubit(QuantumRegister(5, 'q1'), 3)]                      
Sum of state vector: (3.450103311811794-0.308774856832824j)
```

`qiskit==1.3.0`
```
qiskit.__version__='1.3.0'
unitary gate matrix:
Instruction(name='unitary', num_qubits=2, num_clbits=0, params=[array([[ 0.26093746+0.18318812j,  0.12558674-0.38936694j,
         0.56486648-0.53659333j, -0.03587912+0.35025885j],
       [-0.5208327 +0.41955884j,  0.38293134-0.20373037j,
        -0.26997299-0.01611016j,  0.47100825+0.26375647j],
       [-0.18047959-0.0114164j ,  0.42182388+0.05310412j,
         0.01892911-0.48459796j, -0.10334574-0.73530097j],
       [-0.44557601+0.46884564j, -0.63238151+0.25262043j,
         0.15882664-0.24378088j, -0.17446941-0.05317653j]])])
0th gate on [Qubit(QuantumRegister(5, 'q1'), 0), Qubit(QuantumRegister(5, 'q1'), 4), Qubit(QuantumRegister(5, 'q1'), 1), Qubit(QuantumRegister(5, 'q1'), 3), Qubit(QuantumRegister(5, 'q1'), 2)]
unitary gate matrix:
Instruction(name='unitary', num_qubits=2, num_clbits=0, params=[array([[-0.41490662-0.1027918j , -0.45229939-0.45892536j,
         0.1965442 +0.27907236j,  0.53283662+0.04090535j],
       [-0.54960898-0.38886747j, -0.18060452+0.35532657j,
        -0.42299907-0.41452735j,  0.00811512+0.19238008j],
       [-0.48710529+0.16105765j,  0.30802545-0.50934349j,
         0.16188328-0.37911161j, -0.36497996-0.28166686j],
       [ 0.16591044+0.27066769j,  0.26600357-0.02948626j,
         0.10303047-0.5934482j ,  0.6357136 +0.24628756j]])])
1th gate on [Qubit(QuantumRegister(5, 'q1'), 1), Qubit(QuantumRegister(5, 'q1'), 4), Qubit(QuantumRegister(5, 'q1'), 2), Qubit(QuantumRegister(5, 'q1'), 0), Qubit(QuantumRegister(5, 'q1'), 3)]                        
Sum of state vector: (2.6579282427026834+2.2212546524499133j)
```

From our experiment, I believe the results from `qiskit=1.2.4` are correct. There might be some parsing issue between `qiskit-aer` and `qiskit==1.3.0`

### Any suggestions?

_No response_

alexanderivrii:
### Summary

This fixes a bug in the ``RemoveIdentityEquivalent`` transpiler pass (the pass was introduced in #12384), where unitary gates close to identity _up to a global phase_ were removed from the circuit but the global phase of the circuit was not updated. This includes, for instance, single-qubit unitary gates with matrices close to [[ $e^{i\theta}$, 0], [0, $e^{i\theta}$]] for fix floating-point values of $\theta$, standard gates $R_X(2\pi)$, $R_Y(2\pi)$, $R_Z(2\pi)$ (whose matrices are close to $-I$), and some others. Now the pass removes these gates and correctly updates the global phase of the circuit. In particular, the pass now handles non-parameterized ``GlobalPhaseGates`` as well.  

Fixes #13778.

### Details and comments:

The pass ignores parameterized gates. We could probably relatively easily extend it to handle parameterized global phase gates.

The thing to notice [in the current documentation of the pass](https://docs.quantum.ibm.com/api/qiskit/qiskit.transpiler.passes.RemoveIdentityEquivalent) is that the current condition for removing a gate is indeed satisfied for gates of the form $e^{i\theta} I$. And in fact the opposite is also true (thanks to @ShellyGarion for helping me work out the math): if the removal condition is satisfied, then a gate is necessarily of the form $e^{i\theta} I$.

Lol, yet another change after the #13759 was merged is to now also explicitly handle Unitary gates.

Thanks to @Cryoris and @ShellyGarion for joint debugging of the problem.


alexanderivrii:
Update: there is another part of the pass dealing with RX, RY, RZ and RXXGate, RYYGate, RZZGate, RZXGate gates ~that I should look at to see if a similar problem is possible~ that is similarly fixed in [b7713a1](https://github.com/Qiskit/qiskit/pull/13785/commits/b7713a17e42ec11d89cc7dccae81cf872431f682).

Cryoris (commented):
Thanks for tackling this!

I would expect this needs to cover the full spectrum of operations that are the identity up to a global phase, i.e. `exp(i phi) * I` 🙂 This could be done e.g. by just looking at the first entry of the matrix and extracting the phase (similar to how it is done in `Operator.equiv`).

alexanderivrii:
@Cryoris, indeed, you are absolutely correct, we have the same problem with all of the operations that are identity up to a global phase, I will need to check in an updated fix shortly. A related question is do you know why the global phase gates were ignored by RemoveIdentityEquivalent pass?

alexanderivrii:
Hmm, after merging with main (and in particular with #13759), I had to do a small fix to support unitary matrices coming from unitary gates. However, this means that the pass now can't be backported as is, in the backported version the block for unitary matrices needs to be removed.

ShellyGarion (dismissed):
LGTM. Thanks @alexanderivrii for fixing this bug. I'm not sure to which branch it should be merged.

crates/accelerate/src/remove_identity_equiv.rs
```diff
@@ -91,46 +96,53 @@ fn remove_identity_equiv(
...
-                    remove_list.push(op_node)
+                    remove_list.push(op_node);
+                    global_phase_update += (trace / dim).arg();
                 }
             }
             OperationRef::Gate(gate) => {
-                // Skip global phase like gate
-                if gate.num_qubits() < 1 {
-                    continue;
+                if let Some(matrix) = gate.matrix(inst.params_view()) {
```
Cryoris: I wonder if the `Gate` and `UnitaryGate` branches could be merged, as they are identical (right?) and both are `OperationRef` and provide the `matrix` method... maybe something like
```rust
let view = inst.op.view();

match view { 
    // existing parts...
    OperationRef::Gate | OperationRef::Unitary => {
        if let Some(matrix) = view.matrix(inst.params_view()) {
            // existing parts...
        }
    }
 }
```

crates/accelerate/src/remove_identity_equiv.rs
```diff
@@ -140,6 +152,11 @@ fn remove_identity_equiv(
     for node in remove_list {
         dag.remove_op_node(node);
     }
+
+    if global_phase_update != 0. {
+        dag.add_global_phase(py, &Param::Float(global_phase_update))
+            .unwrap();
```
Cryoris: The addition can only fail if the Parameter is not a float, so we can give a reason why this shouldn't fail (and if it does fail, then we know what assumption we had that's wrong 🙂)
```suggestion
            .unwrap("The global phase is guaranteed to be a float");
```

releasenotes/notes/fix-phase-in-remove-id-equivalent-6480da0c62f20df1.yaml
```diff
@@ -0,0 +1,8 @@
+---
+fixes:
+  - |
+    Fixed a bug in :class:`.RemoveIdentityEquivalent` transpiler pass, where the
+    unitary gates close to identity up to a global phase were removed from the circuit,
+    but the global phase of the circuit was not updated. In particular,
+    :class:`.RemoveIdentityEquivalent` now removes non-parameterized :class:`.GlobalPhaseGate`
+    gates.
```
Cryoris: To me the adjective "unitary" here could mean we refer to `UnitaryGate` objects, maybe removing it is better (since it could have been any other gate)?
```suggestion
    Fixed a bug in the :class:`.RemoveIdentityEquivalent` transpiler pass, where
    gates close to identity up to a global phase were removed from the circuit,
    but the global phase of the circuit was not updated. In particular,
    :class:`.RemoveIdentityEquivalent` now removes non-parameterized :class:`.GlobalPhaseGate`
    gates.
```

Cryoris (commented):
Two tiny comments but the logic LGTM!

crates/accelerate/src/remove_identity_equiv.rs
alexanderivrii: This is a nice suggestion, but the compiler does not like it. The problem is that `gate` has different types in the two arms: in `OperationRef::Gate(gate)` it's a `&PyGate`, and in `OperationRef::UnitaryGate(gate)`  it's a `&UnitaryGate`.

releasenotes/notes/fix-phase-in-remove-id-equivalent-6480da0c62f20df1.yaml
alexanderivrii: Sure, done in [68e6a7f](https://github.com/Qiskit/qiskit/pull/13785/commits/68e6a7fca205e994ea7e1b0d95a8c082a9644235).

crates/accelerate/src/remove_identity_equiv.rs
alexanderivrii: Done in [68e6a7f](https://github.com/Qiskit/qiskit/pull/13785/commits/68e6a7fca205e994ea7e1b0d95a8c082a9644235). And of course, you mean `expect` instead of `unwrap`.

Cryoris (approved):
This looks great, thank you! 🙂

PR #13786, opened 2025-02-05
Deprecate `qiskit.circuit.classicalfunction`

gadial:
### Summary
This PR deprecates the `ClassicalFunction` class and related `classical_function` method, to be removed in 2.0.

Partially addresses #13755 

### Details and comments
`ClassicalFunction` is a useful class, but it relies on the `tweedledum` library which is dropped from qiskit in 2.0. Most of the functionality enabled by `ClassicalFunction` is also supplied by `BooleanExpression` that is made independent from `tweedledum` in #13769 and it's not clear if there is demand for the extra functionality at this stage.

### DIFF
--- a/qiskit/circuit/classicalfunction/__init__.py
+++ b/qiskit/circuit/classicalfunction/__init__.py
@@ -128,7 +128,9 @@ from .boolean_expression import BooleanExpression
 @deprecate_func(
     since="1.4",
     removal_timeline="in Qiskit 2.0",
-    additional_msg="Use `PhaseOracle` or `BitFlipOracle` instead",
+    additional_msg="Use `PhaseOracle` instead, which can be turned into a "
+    "bit-flip oracle by applying Hadamard gates on the target "
+    "qubit before and after the instruction.",
 )
 def classical_function(func):
     """
--- a/qiskit/circuit/classicalfunction/__init__.py
+++ b/qiskit/circuit/classicalfunction/__init__.py
@@ -130,7 +130,8 @@ from .boolean_expression import BooleanExpression
     removal_timeline="in Qiskit 2.0",
     additional_msg="Use `PhaseOracle` instead, which can be turned into a "
     "bit-flip oracle by applying Hadamard gates on the target "
-    "qubit before and after the instruction.",
+    "qubit before and after the instruction, and conditioning."
+    "the instruction on the target qubit.",
 )
 def classical_function(func):
     """
--- a/qiskit/circuit/classicalfunction/__init__.py
+++ b/qiskit/circuit/classicalfunction/__init__.py
@@ -20,6 +20,10 @@ ClassicalFunction compiler (:mod:`qiskit.circuit.classicalfunction`)
 Overview
 ========
 
+.. warning::
+    
+    This module is deprecated as of qiskit 1.4.0 version. It will be removed in the Qiskit 2.0 release.
+
 The classical function compiler provides the necessary tools to map a classical
 potentially irreversible functions into quantum circuits.  Below is a simple example of
 how to synthesize a simple boolean function defined using Python into a
--- a/qiskit/circuit/classicalfunction/__init__.py
+++ b/qiskit/circuit/classicalfunction/__init__.py
@@ -128,7 +128,7 @@ from .boolean_expression import BooleanExpression
 @deprecate_func(
     since="1.4",
     removal_timeline="in Qiskit 2.0",
-    additional_msg="Use `BooleanExpression` instead",
+    additional_msg="Use `PhaseOracle` or `BitFlipOracle` instead",
 )
 def classical_function(func):
     """
--- a/qiskit/circuit/classicalfunction/__init__.py
+++ b/qiskit/circuit/classicalfunction/__init__.py
@@ -22,7 +22,7 @@ Overview
 
 .. warning::
     
-    This module is deprecated as of qiskit 1.4.0 version. It will be removed in the Qiskit 2.0 release.
+    This module is deprecated as of Qiskit 1.4.0. It will be removed in the Qiskit 2.0 release.
 
 The classical function compiler provides the necessary tools to map a classical
 potentially irreversible functions into quantum circuits.  Below is a simple example of
--- a/qiskit/circuit/classicalfunction/__init__.py
+++ b/qiskit/circuit/classicalfunction/__init__.py
@@ -111,7 +111,7 @@ Exceptions
    ClassicalFunctionCompilerTypeError
 
 """
-
+from qiskit.utils.deprecation import deprecate_func
 from .classicalfunction import ClassicalFunction
 from .exceptions import (
     ClassicalFunctionParseError,
@@ -121,6 +121,11 @@ from .exceptions import (
 from .boolean_expression import BooleanExpression
 
 
+@deprecate_func(
+    since="1.4",
+    removal_timeline="in Qiskit 2.0",
+    additional_msg="Use `BooleanExpression` instead",
+)
 def classical_function(func):
     """
     Parses and type checks the callable ``func`` to compile it into an ``ClassicalFunction``
--- a/qiskit/circuit/classicalfunction/exceptions.py
+++ b/qiskit/circuit/classicalfunction/exceptions.py
@@ -10,7 +10,13 @@
 # copyright notice, and modified files need to carry a notice indicating
 # that they have been altered from the originals.
 
-"""Exceptions for ClassicalFunction compiler"""
+"""Exceptions for ClassicalFunction compiler
+
+.. warning::
+    
+    This module is deprecated as of qiskit 1.4.0 version. It will be removed in the Qiskit 2.0 release.
+
+"""
 
 from qiskit.exceptions import QiskitError
 
--- a/qiskit/circuit/library/phase_oracle.py
+++ b/qiskit/circuit/library/phase_oracle.py
@@ -63,6 +63,11 @@ class PhaseOracle(QuantumCircuit):
             var_order(list): A list with the order in which variables will be created.
                (default: by appearance)
         """
+        # ignore deprecation warning for BooleanExpression; implementation is changing in 2.0
+        import warnings
+
+        warnings.simplefilter("ignore", DeprecationWarning)
+
         from qiskit.circuit.classicalfunction.boolean_expression import BooleanExpression
         from qiskit.circuit.classicalfunction.classical_element import ClassicalElement
 
--- a/qiskit/circuit/library/phase_oracle.py
+++ b/qiskit/circuit/library/phase_oracle.py
@@ -66,33 +66,35 @@ class PhaseOracle(QuantumCircuit):
         # ignore deprecation warning for BooleanExpression; implementation is changing in 2.0
         import warnings
 
-        warnings.simplefilter("ignore", DeprecationWarning)
+        with warnings.catch_warnings():
+            warnings.simplefilter("ignore", DeprecationWarning)
+            from qiskit.circuit.classicalfunction.boolean_expression import BooleanExpression
+            from qiskit.circuit.classicalfunction.classical_element import ClassicalElement
 
-        from qiskit.circuit.classicalfunction.boolean_expression import BooleanExpression
-        from qiskit.circuit.classicalfunction.classical_element import ClassicalElement
+            if not isinstance(expression, ClassicalElement):
+                expression = BooleanExpression(expression, var_order=var_order)
 
-        if not isinstance(expression, ClassicalElement):
-            expression = BooleanExpression(expression, var_order=var_order)
+            self.boolean_expression = expression
 
-        self.boolean_expression = expression
+            if synthesizer is None:
 
-        if synthesizer is None:
+                def synthesizer(boolean_expression):
+                    from tweedledum.synthesis import pkrm_synth  # pylint: disable=import-error
+                    from qiskit.circuit.classicalfunction.utils import tweedledum2qiskit
 
-            def synthesizer(boolean_expression):
-                from tweedledum.synthesis import pkrm_synth  # pylint: disable=import-error
-                from qiskit.circuit.classicalfunction.utils import tweedledum2qiskit
+                    truth_table = boolean_expression._tweedledum_bool_expression.truth_table(
+                        output_bit=0
+                    )
+                    tweedledum_circuit = pkrm_synth(
+                        truth_table, {"pkrm_synth": {"phase_esop": True}}
+                    )
+                    return tweedledum2qiskit(tweedledum_circuit)
 
-                truth_table = boolean_expression._tweedledum_bool_expression.truth_table(
-                    output_bit=0
-                )
-                tweedledum_circuit = pkrm_synth(truth_table, {"pkrm_synth": {"phase_esop": True}})
-                return tweedledum2qiskit(tweedledum_circuit)
+            oracle = expression.synth(synthesizer=synthesizer)
 
-        oracle = expression.synth(synthesizer=synthesizer)
+            super().__init__(oracle.num_qubits, name="Phase Oracle")
 
-        super().__init__(oracle.num_qubits, name="Phase Oracle")
-
-        self.compose(oracle, inplace=True, copy=False)
+            self.compose(oracle, inplace=True, copy=False)
 
     def evaluate_bitstring(self, bitstring: str) -> bool:
         """Evaluate the oracle on a bitstring.
--- a/releasenotes/notes/deprecate-classical-function-3bf44ef26d984366.yaml
+++ b/releasenotes/notes/deprecate-classical-function-3bf44ef26d984366.yaml
@@ -2,10 +2,11 @@
 deprecations_circuits:
   - |
     The :mod:`qiskit.circuit.classicalfunction` and with it
-    the :class:`.ClassicalFunction` and its related :func:`.classical_function`
-    have been deprecated as of Qiskit 1.4 and will be removed in Qiskit 2.0.
-    This change is done to avoid a dependency on the external library `tweedledum`,
-    which is no longer compatible with all of Qiskit's supported platforms and 
-    Python versions.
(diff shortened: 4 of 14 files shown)


releasenotes/notes/deprecate-classical-function-3bf44ef26d984366.yaml
```diff
@@ -0,0 +1,5 @@
+---
+deprecations_circuits:
+  - |
+    The :class:`.ClassicalFunction` and its related :meth:`.classical_function`
```
Cryoris:
```suggestion
    The :class:`.ClassicalFunction` and its related :func:`.classical_function`
```

releasenotes/notes/deprecate-classical-function-3bf44ef26d984366.yaml
```diff
@@ -0,0 +1,5 @@
+---
+deprecations_circuits:
```
Cryoris, line 2: It would be nice to mention why they are deprecated and also mention in the reno that the `BooleanExpression` can be used as replacement 🙂

qiskit/circuit/classicalfunction/__init__.py
Cryoris, line 1: At this point, it would make sense deprecate the whole module (and add a warning in the module description) and move the `BooleanExpression` to the circuit library. 

The errors are technically also API documented, but I'm not sure if there's such a thing as deprecation warnings for errors... maybe we can ask offline!

Cryoris (commented):
Thanks for taking this on, I'm happy to see this cleaned up for 2.0! 🙂 

It seems there is still one call to classical function in the visualization tests, which is not caught and causes the CI to fail.

releasenotes/notes/deprecate-classical-function-3bf44ef26d984366.yaml
gadial: Done in 5b9e39044

releasenotes/notes/deprecate-classical-function-3bf44ef26d984366.yaml
gadial: Done in 5b9e39044

qiskit/circuit/classicalfunction/__init__.py
gadial: Done in 5b9e39044

releasenotes/notes/deprecate-classical-function-3bf44ef26d984366.yaml
```diff
@@ -1,5 +1,9 @@
 ---
 deprecations_circuits:
   - |
-    The :class:`.ClassicalFunction` and its related :meth:`.classical_function`
+    The :mod:`qiskit.circuit.classicalfunction` and with it
+    the :class:`.ClassicalFunction` and its related :func:`.classical_function`
     have been deprecated as of Qiskit 1.4 and will be removed in Qiskit 2.0.
+    This change is done to stop relying on the `tweedledum` external library.
+    For a similar functionality please use the :class:`BooleanExpression` which
```
Cryoris:
```suggestion
    For a similar functionality please use the :class:`.BooleanExpression` which
```

qiskit/circuit/classicalfunction/__init__.py
```diff
@@ -20,6 +20,10 @@
 Overview
 ========
 
+.. warning::
+    
+    This module is deprecated as of qiskit 1.4.0 version. It will be removed in the Qiskit 2.0 release.
```
Cryoris: For consistency 🙂 
```suggestion
    This module is deprecated as of Qiskit 1.4.0. It will be removed in the Qiskit 2.0 release.
```

qiskit/circuit/classicalfunction/exceptions.py
```diff
@@ -10,7 +10,13 @@
 # copyright notice, and modified files need to carry a notice indicating
 # that they have been altered from the originals.
 
-"""Exceptions for ClassicalFunction compiler"""
+"""Exceptions for ClassicalFunction compiler
+
+.. warning::
+    
+    This module is deprecated as of qiskit 1.4.0 version. It will be removed in the Qiskit 2.0 release.
```
Cryoris, line 17: This module isn't included in a toctree, so the module level docs will not show up in the API docs (see https://docs.quantum.ibm.com/api/qiskit/classicalfunction). So you don't really need the warning here, but it also doesn't hurt for the devs that pass by and see it in the source code 🙂

releasenotes/notes/deprecate-classical-function-3bf44ef26d984366.yaml
```diff
@@ -1,5 +1,9 @@
 ---
 deprecations_circuits:
   - |
-    The :class:`.ClassicalFunction` and its related :meth:`.classical_function`
+    The :mod:`qiskit.circuit.classicalfunction` and with it
+    the :class:`.ClassicalFunction` and its related :func:`.classical_function`
     have been deprecated as of Qiskit 1.4 and will be removed in Qiskit 2.0.
+    This change is done to stop relying on the `tweedledum` external library.
```
Cryoris:
```suggestion
    This change is done to avoid a dependency on the external library `tweedledum`,
    which is no longer compatible with all of Qiskit's supported platforms and 
    Python versions.
```

Cryoris (commented):
Some small comments, otherwise LGTM 👍🏻

releasenotes/notes/deprecate-classical-function-3bf44ef26d984366.yaml
```diff
@@ -0,0 +1,11 @@
+---
+deprecations_circuits:
+  - |
+    The :mod:`qiskit.circuit.classicalfunction` and with it
+    the :class:`.ClassicalFunction` and its related :func:`.classical_function`
+    have been deprecated as of Qiskit 1.4 and will be removed in Qiskit 2.0.
+    This change is done to avoid a dependency on the external library `tweedledum`,
+    which is no longer compatible with all of Qiskit's supported platforms and 
+    Python versions.
+    For a similar functionality please use the :class:`.BooleanExpression` which
```
ShellyGarion: but the BooleanExpression class will be part of the circuit/synthesis library and not of classicalfucntion?

ShellyGarion (commented):
Just making sure: in this PR we raise deprecation warnings, and the new API will be introduced in qiskit 2.0 ?
Should there be another PR for removing ClassicalFunction for qiskit 2.0, since it's not part of https://github.com/Qiskit/qiskit/pull/13769?

releasenotes/notes/deprecate-classical-function-3bf44ef26d984366.yaml
ShellyGarion: actually, will the BooleanExpression class be publicly available? or only the oracles?

ShellyGarion:
I also found tweedledum here:
https://github.com/Qiskit/qiskit/blob/9ae0a808b45249d6f185baf54f078cb1ed00648a/qiskit/utils/optionals.py#L186
should it be removed / deprecated? 

See also this test file:
https://github.com/Qiskit/qiskit/blob/main/test/python/circuit/test_extensions_standard.py

gadial:
> Just making sure: in this PR we raise deprecation warnings, and the new API will be introduced in qiskit 2.0 ? Should there be another PR for removing ClassicalFunction for qiskit 2.0, since it's not part of #13769?

Yes, we'll add the new API in 2.0.

I think we *can* use #13769 to remove `ClassicalFunction`, even though I usually try to keep PR's centered around one change.

releasenotes/notes/deprecate-classical-function-3bf44ef26d984366.yaml
gadial: Currently we only intend to expose the oracles, this release note should be updated to reflect this.

gadial:
> I also found tweedledum here:
> 
> https://github.com/Qiskit/qiskit/blob/9ae0a808b45249d6f185baf54f078cb1ed00648a/qiskit/utils/optionals.py#L186
> 
> 
> should it be removed / deprecated?
> See also this test file: https://github.com/Qiskit/qiskit/blob/main/test/python/circuit/test_extensions_standard.py

It should definitely be removed, but I don't think we need deprecation warning here.

1ucian0:
This one should be against `stable/1.4`

ShellyGarion (dismissed):
LGTM. are there any further comments @Cryoris ?

qiskit/circuit/classicalfunction/__init__.py
```diff
@@ -121,6 +125,11 @@ def grover_oracle(a: Int1, b: Int1, c: Int1, d: Int1) -> Int1:
 from .boolean_expression import BooleanExpression
 
 
+@deprecate_func(
+    since="1.4",
+    removal_timeline="in Qiskit 2.0",
+    additional_msg="Use `PhaseOracle` or `BitFlipOracle` instead",
```
Cryoris: The `BitFlipOracle` doesn't exist in Qiskit 1.4, we need to remove mentioning it from the deprecation messages. Since the `PhaseOracle` is not a direct replacement of the classical functions but requires extra steps, it might be good to tell the users how to get there
```suggestion
    additional_msg=(
        "Use the PhaseOracle instead, which can be turned into a "
        "bit-flip oracle by applying Hadamard gates on the target "
        "qubit before and after the instruction."
    ),
```

releasenotes/notes/deprecate-classical-function-3bf44ef26d984366.yaml
```diff
@@ -0,0 +1,12 @@
...
+  - |
+    The :mod:`qiskit.circuit.classicalfunction` and with it
+    the :class:`.ClassicalFunction`, its related :func:`.classical_function`
+    and :class:`.BooleanExpression` have been deprecated as of Qiskit 1.4 
+    and will be removed in Qiskit 2.0. This change is done to avoid a 
+    dependency on the external library `tweedledum`, which is no longer 
+    compatible with all of Qiskit's supported platforms and Python versions.
+    For a similar functionality please use the :class:`.PhaseOracle` which
+    is going to have a `tweedledum` independent implementation, and
+    the :class:`.BitFlipOracle` which will be added in Qiskit 2.0.
```
Cryoris, line 12: We could also add a code snippet as example if we want to be extra nice to the users 😄

test/python/circuit/library/test_phase_oracle.py
```diff
@@ -37,7 +37,11 @@ class TestPhaseOracle(QiskitTestCase):
     @unpack
     def test_evaluate_bitstring(self, expression, input_bitstring, expected):
         """PhaseOracle(...).evaluate_bitstring"""
-        oracle = PhaseOracle(expression)
+        with self.assertWarnsRegex(
+            DeprecationWarning,
+            expected_regex="BooleanExpression`` is deprecated as of qiskit 1.4",
+        ):
+            oracle = PhaseOracle(expression)
```
Cryoris: The `PhaseOracle` is not getting deprecated, it should not emit a deprecation warning. Instead we can filter the warning in the `PhaseOracle` code in 1.4 until we get the tweedledum-free version in 2.0.

Cryoris (commented):
This deprecation is a bit finnicky. Our deprecation rules are that 
1. an alternative implementation must be available for 1 minor release, where no `DeprecationWarning` is issued (`PendingDeprecationWarning` is fine)
2. then, the `DeprecationWarning` must be issued for 1 minor release
3. finally, we can remove the feature

Here we cannot point users to the `BitFlipOracle`, which doesn't exist yet, and we cannot issue a deprecation warning inside `PhaseOracle`, which doesn't have a deprecation-free alternative. In a perfect world we would have introduced a tweedledum-free version of `PhaseOracle` and the `BitFlipOracle` in 1.3 already, deprecate the `classicalfunction` in 1.4 and remove in 2.0.

To move forward while still being true to the deprecation policy, I'd suggest we'd 
* just point to `PhaseOracle` as alternative path for everything (this has existed for long enough)
* don't issue any warnings when using `PhaseOracle`
* tell users how they could get a boolean expression/bitflip oracle from a `PhaseOracle`

test/python/circuit/library/test_phase_oracle.py
gadial: Done in 4510aee

qiskit/circuit/classicalfunction/__init__.py
gadial: Done in 4510aee

qiskit/circuit/classicalfunction/__init__.py
```diff
@@ -121,6 +125,13 @@ def grover_oracle(a: Int1, b: Int1, c: Int1, d: Int1) -> Int1:
 from .boolean_expression import BooleanExpression
 
 
+@deprecate_func(
+    since="1.4",
+    removal_timeline="in Qiskit 2.0",
+    additional_msg="Use `PhaseOracle` instead, which can be turned into a "
+    "bit-flip oracle by applying Hadamard gates on the target "
+    "qubit before and after the instruction.",
+)
```
alexanderivrii: It would be nice to include somewhere an explicit example that shows how to do this. In fact, I don't quite understand how to turn a phase-oracle (defined over `n` qubits) to a bit-flip-oracle (defined over `n+1` qubits).

releasenotes/notes/deprecate-classical-function-3bf44ef26d984366.yaml
alexanderivrii: I second that, and maybe the release notes would be a perfect place to include an example on how to convert the phase oracle to the bit-flip oracle.

qiskit/circuit/classicalfunction/__init__.py
gadial: It's slightly more tricky than what we currently write in the deprecation message, since you also need to condition the whole phase-flip oracle on the additional qubit. It basically looks like this:
```
from qiskit import QuantumRegister, QuantumCircuit
from qiskit.circuit.classicalfunction import BooleanExpression
from qiskit.circuit.library.phase_oracle import PhaseOracle
from qiskit.quantum_info.operators import Operator

bool_expr = "(x0 & x1 | ~x2) ^ x4" # Play with this to check different oracles
qr_x = QuantumRegister(4, "x")
qr_y = QuantumRegister(1, "y")
qc = QuantumCircuit(qr_x, qr_y)

bit_flip_oracle = BooleanExpression(bool_expr)
qc.compose(bit_flip_oracle, qubits=[*qr_x, *qr_y], inplace=True)
bit_flip_op = Operator(qc)

qc = QuantumCircuit(qr_x, qr_y)
phase_flip_oracle = PhaseOracle(bool_expr)
controlled_phase_flip_oracle = phase_flip_oracle.control(1)
qc.h(qr_y)
qc.compose(controlled_phase_flip_oracle, qubits=[*qr_y, *qr_x], inplace=True)
qc.h(qr_y)
phase_flip_op = Operator(qc)

print(bit_flip_op == phase_flip_op) # should always return True
```

The math is relatively simple; the tricky part is conditioning the oracle, but thankfully `qiskit` has a general mechanism for that.

Do you think a code based on the above would make a good example? (no need for the comparison part)

ShellyGarion: I think that it's too complicated. It makes more sense not to mention it at all or to refer the users to the Qiskit 2.0 BitFlipGate.

gadial: If I understand the problem correctly, we **must** supply a viable alternative. Since the relevant code is simply
```
phase_flip_oracle = PhaseOracle(bool_expr)
controlled_phase_flip_oracle = phase_flip_oracle.control(1)
qc.h(qr_y)
qc.compose(controlled_phase_flip_oracle, qubits=[*qr_y, *qr_x], inplace=True)
qc.h(qr_y)
```

I think it's a good enough solution.

ShellyGarion (dismissed):
LGTM. do you have any last comments @Cryoris ?

qiskit/circuit/library/phase_oracle.py
```diff
@@ -63,6 +63,11 @@ def __init__(
             var_order(list): A list with the order in which variables will be created.
                (default: by appearance)
         """
+        # ignore deprecation warning for BooleanExpression; implementation is changing in 2.0
+        import warnings
+
+        warnings.simplefilter("ignore", DeprecationWarning)
```
Cryoris: This will enable a global filter, we should use 
```python
with warnings.catch_warnings(action="ignore", category=DeprecationWarning):
   # code here...
```
🙂

releasenotes/notes/deprecate-classical-function-3bf44ef26d984366.yaml
```diff
@@ -0,0 +1,25 @@
...
+
+    Until :class:`.BitFlipOracle` is added, a phase flip oracle can be converted
+    to a bit flip oracle by conditioning it on the result qubit, and applying 
+    Hadamard gate before and after the application of the oracle, as in the
+    following example (where the oracle is on `qr_x` and the result is on `qr_y`):
+
+    .. code-block:: python
+    
+        phase_flip_oracle = PhaseOracle(bool_expr)
```
Cryoris: That's nice! What do you think of making this a fully executable snippet with some example boolean expression?

gadial: Done in b4878c62c

qiskit/circuit/library/phase_oracle.py
gadial: Done in b4878c62c

Cryoris (approved):
Thanks for all the work!

Cryoris:
Third try to really be sure it's not a flake..

ElePT:
I just inspected the backwards compatibility tests and found the source of the problems we are seeing: 
`qiskit.qpy.exceptions.QpyError: 'Incompatible symengine version 0.14 used to generate the QPY payload'`. We need to see why symengine 0.14 is getting installed where it shouldn't be. [Edit: earlier qiskit versions don't have the pin on the symengine version, that's why].

PR #13790, opened 2025-02-05
Do not contract control-flow operations during `SabreSwap`

### LINKED ISSUE #13768: Control-flow routing causes idle qubits to be stripped from control-flow blocks

jakelishman:
### Summary

The previous implementation of control-flow handling for `SabreSwap` caused control-flow blocks to contract away from idle qubit wires as a side effect of its algorithm.  This will no longer be always valid with the addition of `Box`, and besides, that optimisation isn't part of a routing algorithm's responsibilities, so it's cleaner to have it in a separate pass (now `ContractIdleWiresInControlFlow`) where it can be applied at times other than routing, like in the optimisation loop.

### Details and comments

Fix #13768.

`ContractIdleWiresInControlFlow` is in #13779.

### DIFF
--- a/qiskit/transpiler/passes/routing/sabre_swap.py
+++ b/qiskit/transpiler/passes/routing/sabre_swap.py
@@ -416,6 +416,12 @@ def _apply_sabre_result(
                 block_root_logical_map = {
                     inner: root_logical_map[outer] for inner, outer in zip(block.qubits, node.qargs)
                 }
+                # The virtual qubits originally incident to the block should be retained even if not
+                # actually used; the user might be marking them out specially (like in `box`).
+                # There are other transpiler passes to remove those dependencies if desired.
+                incident_qubits = {
+                    layout.virtual_to_physical(block_root_logical_map[bit]) for bit in block.qubits
+                }
                 block_dag, block_layout = recurse(
                     empty_dag(block),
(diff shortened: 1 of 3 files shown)


kevinhartman (approved):
Looks good to me. Thanks for the changes!

PR #13620, opened 2025-01-07
Increase documentation of built-in transpiler plugins

jakelishman:
### Summary

This overhauls how the `qiskit.transpiler` documentation talks about the transpiler plugins.  All of the built-in plugins now have a decent amount of overview documentation, and the requirements and expectations for each stage of the preset pipelines is explained in more detail.

This form of the documentation makes the distinction between "a compilation routine in general" and "Qiskit's specific choice of default pipeline" clearer, to avoid confusion for advanced users.

Much of the guide-level explanations of the different preset pipeline stages moved to https://docs.quantum.ibm.com some time ago, so this PR removes those, in favour of focussing on the actual API, and inserts links to learn more about the principles elsewhere.

The guide-level explanation of scheduling is left in-place for now, because the content on the other parts of the IBM documentation isn't as complete for that.

### Details and comments

I've attempted to draw a good line between explaining what Qiskit actually does, so users know, and maintaining that we reserve the right to modify specifics of things (especially `default` plugins) between minor versions.

This isn't a full rewrite of everything - I feel like we could probably do with revisiting the documentation of _a lot_ of individual transpiler passes' classes - but I was writing my own plugin over the last month or so, and this is documentation I personally wanted to see.

PR #13792, opened 2025-02-05
Ensure `translation` stage always outputs ISA circuits

### LINKED ISSUE #13787: Move `GateDirection` responsibility into `translation` preset stage
At the moment, the `GateDirection` fix-up pass is run in the "pre-optimisation" stage as a fix-up method, mostly because `BasisTranslator` isn't direction-aware.

Instead, we should shift the responsibility of the translation stage so that a translation plugin is required to put out an ISA circuit, such that the optimisation loop can then be certain it takes in a hardware-lowered circuit.  This makes it possible for translation plugins that _are_ direction-aware (like the upcoming `BasisConstructor`) to avoid the extra cost, and it removes the need for the `pre_optimization` stage in the default `StagedPassManager` construction, which is a de facto stage of its own accord right now, since it's populated by default.

This is an API change, since it's adding to the responsibilities of external translation plugins, but the mitigation strategy is already available: add the `GateDirection` pass to plugins that might need it.

jakelishman:
### Summary

This shifts the responsibility of the `translation` stage slightly to always output ISA circuits.  It previously was not 100% required that they did this, at least implicitly because Qiskit's built-in plugins didn't always respect 2q direction.

The builtin translation plugins now always respect 2q direction, which removes the need for the `pre_optimization` implicit stage, freeing it up for better user customisation.

This (in theory) shouldn't have runtime impacts because the optimisation loop was already having to do this afterwards anyway.  For potential plugins that _do_ respect gate direction (like the upcoming `BasisConstructor`), it can be a speedup in the default pipeline, since they won't need to run the gate-direction check any more.

### Details and comments

Built on top of #13620, since it will need to change documentation in that PR.

Fix #13787.

### DIFF
--- a/qiskit/transpiler/preset_passmanagers/common.py
+++ b/qiskit/transpiler/preset_passmanagers/common.py
@@ -52,6 +52,7 @@ from qiskit.transpiler.passes.layout.vf2_layout import VF2LayoutStopReason
 from qiskit.transpiler.passes.layout.vf2_post_layout import VF2PostLayoutStopReason
 from qiskit.transpiler.exceptions import TranspilerError
 from qiskit.transpiler.layout import Layout
+from qiskit.utils import deprecate_func
 from qiskit.utils.deprecate_pulse import deprecate_pulse_arg
 
 
@@ -379,11 +380,19 @@ def generate_routing_passmanager(
     return routing
 
 
+@deprecate_func(
+    since="2.0",
+    additional_msg=(
+        "Translation plugins are now required to respect ISA directionality,"
+        " so typically no replacement is necessary."
+    ),
+    removal_timeline="in Qiskit 3.0",
+)
 def generate_pre_op_passmanager(target=None, coupling_map=None, remove_reset_in_zero=False):
     """Generate a pre-optimization loop :class:`~qiskit.transpiler.PassManager`
 
     This pass manager will check to ensure that directionality from the coupling
-    map is respected
+    map is respected.
 
     Args:
         target (Target): the :class:`~.Target` object representing the backend
@@ -458,6 +467,7 @@ def generate_translation_passmanager(
         TranspilerError: If the ``method`` kwarg is not a valid value
     """
     if method == "translator":
+        translator = BasisTranslator(sel, basis_gates, target)
         unroll = [
             # Use unitary synthesis for basis aware decomposition of
             # UnitaryGates before custom unrolling
@@ -479,8 +489,9 @@ def generate_translation_passmanager(
                 basis_gates=basis_gates,
                 qubits_initially_zero=qubits_initially_zero,
             ),
-            BasisTranslator(sel, basis_gates, target),
+            translator,
         ]
+        fix_1q = [translator]
     elif method == "synthesis":
         unroll = [
             # # Use unitary synthesis for basis aware decomposition of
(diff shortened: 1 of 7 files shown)


jakelishman:
Now rebased over `main`.  I'll run the benchmark suite overnight.

jakelishman:
I ran the full benchmark suite, and there (as mostly expected) doesn't seem to be much change from this.  There's a small positive improvement in a couple of places of runtime, though I'm not convinced it wasn't just fluctuations on my machine.  There's depth improvements in QUEKO, perhaps because in cases where the optimisation loop is entered twice, there's a bit more information available to it?  Or maybe it's just because there's the knock-on effect of some changes to the randomisation.

```text
| Change   | Before [4d2265f6] <dag-module^2>   | After [92f30cdf] <translation-stage-direction>   |   Ratio | Benchmark (Parameter)                                                    |
|----------|------------------------------------|--------------------------------------------------|---------|--------------------------------------------------------------------------|
| -        | 1.24±0.04ms                        | 1.07±0.04ms                                      |    0.86 | quantum_info.CliffordDecomposeBench.time_decompose('5,10')               |
| -        | 4.53±0.4ms                         | 3.33±0.07ms                                      |    0.73 | quantum_info.SparsePauliOpBench.time_tensor(150, 100)                    |
| -        | 18.0±0.4ms                         | 13.5±0.3ms                                       |    0.75 | queko.QUEKOTranspilerBench.time_transpile_bigd(3, None)                  |
| -        | 81.9±1ms                           | 61.9±1ms                                         |    0.76 | queko.QUEKOTranspilerBench.time_transpile_bntf(0, 'sabre')               |
| -        | 138±6ms                            | 88.2±1ms                                         |    0.64 | queko.QUEKOTranspilerBench.time_transpile_bntf(0, None)                  |
| -        | 114±2ms                            | 99.8±1ms                                         |    0.87 | queko.QUEKOTranspilerBench.time_transpile_bntf(2, 'sabre')               |
| -        | 165±0.7ms                          | 123±2ms                                          |    0.74 | queko.QUEKOTranspilerBench.time_transpile_bntf(3, 'sabre')               |
| -        | 74                                 | 43                                               |    0.58 | queko.QUEKOTranspilerBench.track_depth_bigd_optimal_depth_45(2, 'sabre') |
| -        | 74                                 | 45                                               |    0.61 | queko.QUEKOTranspilerBench.track_depth_bigd_optimal_depth_45(2, None)    |
| -        | 74                                 | 43                                               |    0.58 | queko.QUEKOTranspilerBench.track_depth_bigd_optimal_depth_45(3, 'sabre') |
| -        | 74                                 | 45                                               |    0.61 | queko.QUEKOTranspilerBench.track_depth_bigd_optimal_depth_45(3, None)    |
| -        | 255                                | 216                                              |    0.85 | queko.QUEKOTranspilerBench.track_depth_bntf_optimal_depth_25(2, 'sabre') |
| -        | 45                                 | 30                                               |    0.67 | queko.QUEKOTranspilerBench.track_depth_bntf_optimal_depth_25(2, None)    |
| -        | 45                                 | 30                                               |    0.67 | queko.QUEKOTranspilerBench.track_depth_bntf_optimal_depth_25(3, None)    |
| -        | 28.7±1ms                           | 22.5±0.7ms                                       |    0.79 | random_circuit_hex.BenchRandomCircuitHex.time_ibmq_backend_transpile(10) |
| -        | 39.3±2ms                           | 28.7±1ms                                         |    0.73 | random_circuit_hex.BenchRandomCircuitHex.time_ibmq_backend_transpile(12) |
| -        | 20.4±0.8ms                         | 17.3±0.6ms                                       |    0.85 | random_circuit_hex.BenchRandomCircuitHex.time_ibmq_backend_transpile(8)  |
```

qiskit/transpiler/preset_passmanagers/common.py
```diff
@@ -528,8 +539,40 @@ def generate_translation_passmanager(
...
+                plugin_config=unitary_synthesis_plugin_config,
+                method=unitary_synthesis_method,
+                target=target,
+            ),
+        ]
     else:
         raise TranspilerError(f"Invalid translation method {method}.")
+    # Our built-ins don't 100% guarantee that 2q gate direction is respected, so we might need to
+    # run a little bit of fix up on them.  `GateDirection` doesn't guarantee that 1q gates are
+    # ISA safe after it runs, so we need another run too.
```
ElePT, line 561: Just to make sure I understood this sentence, is it `GateDirection` or `CheckGateDirection` that doesn't guarantee that 1q gates are ISA?

ElePT (commented):
I am absolutely in favor of this change, I think it's much cleaner and flexible, and the changes look good! I just left a question for my own understanding of the passes.

qiskit/transpiler/preset_passmanagers/common.py
jakelishman: It's `GateDirection` - that's the one that actually inserts gates.  `CheckGateDirection` just says whether or not we need to run `GateDirection` and the subsequent fixer.

ElePT (approved):
LGTM!

PR #13794, opened 2025-02-05
Add `default` translation stage plugin

### LINKED ISSUE #13788: Add `default` translation plugin
At the moment, our available translation stages are called `translator` and `synthesis`, both of which are fine names.  We default to `translator` unless overridden by a backend, or the user.

We may start to add more support for discrete-basis backend compilation over the course of Qiskit 2.x, or wish to modify the default method in the presence of fractional 2q gates, or so son.  Adding a new default called `default`, which simply wraps `translator` (to begin with) gives us the semantic freedom to make some of those changes within the major version without a hard API break.

jakelishman:
### Summary

This is added to give us semantic room to play with the default translation algorithm during the 2.x series.  It starts life as an alias to `translator`, but may change to improve support for fractional-gate translation, more optimized synthesis routines, or discrete-basis translations (like a SK-based 1q translator for logical backends).

This makes it programmatically possible to distinguish the "default" plugin choice from an explicit "translator" choice, even when (as right now) those are the same thing.  Note that `pass_manager_config` uses `None` as an implicit default, but we active suggest using the plugin-stage getters, which need an actual plugin name.

### Details and comments

Close #13788.

Built on top of the enhanced documentation #13620.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -122,6 +122,7 @@ sk = "qiskit.transpiler.passes.synthesis.solovay_kitaev_synthesis:SolovayKitaevS
 default = "qiskit.transpiler.preset_passmanagers.builtin_plugins:DefaultInitPassManager"
 
 [project.entry-points."qiskit.transpiler.translation"]
+default = "qiskit.transpiler.preset_passmanagers.builtin_plugins:DefaultTranslationPassManager"
 synthesis = "qiskit.transpiler.preset_passmanagers.builtin_plugins:UnitarySynthesisPassManager"
 translator = "qiskit.transpiler.preset_passmanagers.builtin_plugins:BasisTranslatorPassManager"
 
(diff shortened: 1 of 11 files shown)


jakelishman:
Now rebased over `main`.

ElePT (approved):
I am usually triggered by things like this but even after reviewing the new docs I didn't realize that the translation stage was missing a default plugin. I think the change looks good and makes sense (but I will leave the PR around for a bit in case anyone has further opinions)

Cryoris (approved):
Nice, I'm looking forward to finally transpiling to a discrete basis out of the box 😄

jakelishman:
Ha, I think we've got quite some way to go still til then, but this at least gives us semantic leeway to make changes to the algorithm we use by default to move that way.

PR #13545, opened 2024-12-10
Fix `.decompose` on control flow

Cryoris:
### Summary

This PR fixes two behaviors when decomposing control flow:
* `c_if` on an operation without definition (i.e. `None`) would fail #13493)
* control flow blocks were unaffected by `.decompose` (#13544)

### Details and comments

Even though `c_if` is deprecated we should still handle this case, but the logic for this can be removed once `c_if` is gone. There's a comment specifying which part is `c_if`-specific.

PR #13796, opened 2025-02-05
Pylint configuration to prevent accidentally committing debugging print statements

1ucian0:
### Summary

Adds the `bad_builtin` extension to pylint and raise a warning when `print` is used. Add a `disable` for current cases.

### Details and comments

It happens from time to time that we accidentally push for code with debugging `print` statements. [Many](https://github.com/Qiskit/qiskit/pull/12850/commits/6204af45fa43e6219fdeb842f31211e1679b6143) [times](https://github.com/Qiskit/qiskit/pull/13545/commits/a9c5bc7e644ebc0d7de8947ebbe6670fa080ddd6), they are captured during revision. However, [sometimes](https://github.com/Qiskit/qiskit/pull/13702) they end up in `main`. Maybe it helps if pylint tell us.

out-of-scope: evaluate if the current prints are justify. Issues can be submitted for them independently.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -215,7 +215,7 @@ extension-pkg-allow-list = [
     "rustworkx",
     "tweedledum",
 ]
-load-plugins = ["pylint.extensions.docparams", "pylint.extensions.docstyle"]
+load-plugins = ["pylint.extensions.docparams", "pylint.extensions.docstyle", "pylint.extensions.bad_builtin"]
 py-version = "3.9"  # update it when bumping minimum supported python version
 
 [tool.pylint.basic]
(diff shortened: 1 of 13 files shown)


jakelishman (approved):
This seems like a reasonable and easy catch for the test suite.  We only _really_ rarely use it in library code (mostly just visualisations), so I think it'll probably help more than hurt - it's easy to blanket allow in relevant tools (though it'd be cool if we could more easily configure pylint per directory).

1ucian0:
> though it'd be cool if we could more easily configure pylint per directory.

I fully agree. However, I'm hesitant if introducing [a plugin](https://pypi.org/project/pylint-per-file-ignores/) just for this is worthy.

jakelishman:
Oh yeah, I don't want a plugin really, I kind of wanted it as a pylint builtin.  We should probably put more effort in moving to Ruff properly than trying to fix up pylint at this stage - I think I was the last major hold-out a year ago or so.

I have a semi-decent configuration for the basis-constructor plugin repo that we might be able to bring over to Qiskit, though I'm still nervous that it's not as effective as pylint (but it is a hell of a lot faster).

PR #13797, opened 2025-02-06
Remove dangling prints

Cryoris:
### Summary

The latest merge from main triggered my pre-commit hook and found some dangling print statements -- unfortunately #13796 only covers Python 🙂


jakelishman (approved):
haha whoops, my bad!

Issue #13662, opened 2025-01-14
Remove pulse and its dependencies in Qiskit 2.0


eliarbel:
### Background
This issue is for tracking the work required for removing pulse and all its dependencies within Qiskit. For convenience, `qiskit/utils/deprecate_pulse.py` defines a few decorators for deprecating pulse code, namely `deprecate_pulse_func`, `deprecate_pulse_dependency` and `deprecate_pulse_arg`. All pulse deprecations were marked with these decorators (mostly in https://github.com/Qiskit/qiskit/pull/13164), so it should be rather easy to locate the code that need to be removed by searching for `deprecate_pulse` within the codebase. 

### Details
Most of the changes here are just code removal, but certain cases require special handling for supporting the fact that we don't have pulse anymore. For example, we would want QPY to still be able to load a payload with pulse data in it (likely leading to a partially-specified circuit), possibly notifying the user that pulse data cannot be represented in Qiskit thus it will be ignored. So I think it makes sense to expect that we'll end up having a few PRs for handling all the removal with the vast majority done in one PR with few additional ones to implement specific changes. Furthermore, I think it would help to work against a dedicated branch so we can make progress in parallel without breaking anything in Qiskit `main`. 

Here are the major things that we need to address. If we work against a single branch, we can mark the progress against that branch: 

- [ ] Pulse itself  - `qiskit/pulse`
- [ ] Pulse visualization - `visualization/pulse_v2`
- [x] Transpiler and circuit related
    - [x] Passes: `PulseGates`, `RXCalibrationBuilder`, `RZXCalibrationBuilder`, `EchoRZXWeylDecomposition`, `ValidatePulseGates`
    - [x] Deprecated args e.g. in `generate_preset_pass_manager()`, `PassManagerConfig`, `generate_scheduling()`, `transpile`
    - [x] Target - calibration and instruction schedule map support
    - [x] `DAGCircuit` and `DagDependency` - calibrations support (also in Rust)
    - [x] `QuantumCircuit` - calibrations support
- [x] qpy
    - [x] Drop support for `ScheduleBlock` in `dump()`
    - [x] Gracefully ignore schedule blocks when loading payloads containing pulse data
- [ ] qobj - remove `QobjToInstructionConverter`
- [x] assembler/scheduler/compiler
    - [x] Remove `sequence()`, `schedule()`, `schdule_circuit()`, `assemble_schedules()`, `ScheduleConfig()` and scheduling related methods
- [ ] Providers 
[...]

PR #13793, opened 2025-02-05
Remove deprecated `BackendV1` elements, including base class, `providers.models` and `qobj` modules

ElePT:
### Summary
This PR builds upon (~and is BLOCKED BY~) the changes in #13805.
This PR removes `BackendV1` and all related modules and utils. This includes:

- from `qiskit/providers`:
    - `BackendV1 `
    - `BackendV2Converter `
    - `convert_to_target `
    - `BackendPropertyError `
    - `BackendConfigurationError `
   
- from `qiskit/providers/models`: these were BackendV1 data structures. Full module removed
    - `BackendConfiguration`
    - `BackendProperties`
    - `BackendStatus`
    - `QasmBackendConfiguration`
    - `PulseBackendConfiguration `
    - `UchannelLO`
    - `GateConfig `
    - `PulseDefaults `
    - `Command`
    - `JobStatus`
    - `GateProperties`
    - `Nduv`

- Removed `PulseQobjDef ` from `qiskit/providers/models` (this steps a bit into pulse removal realm)
- from `qiskit/qobj`: Full module removed.
    - QobjExperimentHeader
    - QobjHeader
    - QasmQobj
    - QasmQobjInstruction
    - QasmQobjExperimentConfig
    - QasmQobjExperiment
    - QasmQobjConfig
    - QasmExperimentCalibrations
    - GateCalibration
    - PulseQobj
    - PulseQobjInstruction
    - PulseQobjExperimentConfig
    - PulseQobjExperiment
    - PulseQobjConfig
    - QobjMeasurementOption
    - PulseLibraryItem

Note that there were two classes defined in the qobj utils (`MeasLevel` and `MeasReturnType`) used in the result class, so they have been moved to `result/models.py`.
### TODOs

- [x] Reno

### Details

- [x] What should we do with the "Migrate from BackendV1 to BackendV2" guide? Check with docs team to make sure it's covered somewhere else. -> Update: opened https://github.com/Qiskit/documentation/pull/2705
- [x] This removal breaks Aer, currently seen in image tests: 

```
qiskit.exceptions.OptionalDependencyImportWarning: While trying to import 'Qiskit Aer', some components were located but raised other errors during import. You might have an incompatible version installed. Qiskit will continue as if the optional is not available.
module 'qiskit_aer' failed to import with: ImportError("cannot import name 'convert_to_target' from 'qiskit.providers' (/home/vsts/work/1/s/qiskit/providers/__init__.py)")
```

PR #13798, opened 2025-02-06
Remove pulse-related passes

eliarbel:
This commit removes the calibration builder and pulse-gate passes which were deprecated in Qiskit 1.3, together with the pulse package itself. Related unit tests have also been removed.

### Details and comments
Part of #13662


releasenotes/notes/remove-pulse-passes-3128f27ed7e42bf6.yaml
```diff
@@ -0,0 +1,7 @@
+---
+upgrade_transpiler:
+  - |
+    The ``PulseGates``, ``ValidatePulseGates``, ``RXCalibrationBuilder``, ``RZXCalibrationBuilder``,
+    ``RZXCalibrationBuilderNoEcho`` and ``EchoRZXWeylDecomposition`` passes have been removed,
+    following their deprecation in Qiskit 1.3. These passes depend on and relate to the Pulse
+    package which is also being removed in Qiskit 2.0.
```
ElePT, line 7: Lint is indicating that the `inst_map` argument in `generate_scheduling` is no longer used, if we were to remove it (why keep it), the reno should also mention it.

ElePT (commented):
The PR looks good, and it should help unblock the situation in #13793.  There are only a couple of lint complaints left that unfortunately github doesn't let me suggest changes for.

releasenotes/notes/remove-pulse-passes-3128f27ed7e42bf6.yaml
eliarbel: Yes, the argument is no longer used within `generate_scheduling` itself indeed (and I addressed the lint warning about that in this PR), but I was thinking about mentioning it in a different reno bullet that removes `inst_map` from all the other functions (e.g. `generate_preset_pass_manager`), together possibly with more removals in the transpiler/circuit area rather than here. So if you're OK, I would keep this reno bullet about passes

ElePT: As long as we don't forget about it, it definitely makes more sense in an independent reno. I sometimes create the reno bullet in one PR and extend it in following ones, but this is a small enough change (and close enough time-wise) that we probably don't need to do that.

ElePT (dismissed):

ElePT (approved):

PR #13800, opened 2025-02-06
Deny use of `println!` and related write macros

jakelishman:
### Summary

We lint against the `print` builtin in Python space, since it's a debugging artifact almost all of the time.  This is the equivalent lint, but for Rust space.

### Details and comments

Goes on top of #13797 to enforce the lint properly, once that's fixed.


ElePT (approved):
Yes please

PR #13802, opened 2025-02-06
Add `default` routing stage plugin

jakelishman:
### Summary

This is added for a similar reason as the recent `default` translation-stage plugin (gh-13794).  We currently have no plans to change the default routing method, if not only because the `default` layout plugin does routing most of the time, but this makes the semantics of "just use the default method" clearer.

### Details and comments

Co-authored-by: Elena's neuroses <ocd@peña-tapia.es>


ElePT (approved):
`ocd@peña-tapia.es` is satisfied with this PR.

PR #13803, opened 2025-02-06
Fix the sign in `SparseObservable.as_paulis`

Cryoris:
### Summary

`echo $issue_title`

### Details and comments

In my mind `{1, -1} *= {1, -1}` translated to `bool &= bool` -- but it doesn't. Instead it should be `bool ^= !bool`. This would come up when an observable has more than a single -1 eigenvalue projector (e.g. `"11"`). Explicit tests for this are added, while a reno is not, since it's not yet released 🙂

### DIFF
--- a/crates/accelerate/src/sparse_observable.rs
+++ b/crates/accelerate/src/sparse_observable.rs
@@ -687,10 +687,10 @@ impl SparseObservable {
                 .multi_cartesian_product();
 
             for combination in combinations {
-                let mut positive = true;
+                let mut positive = true; // keep track of the global sign
 
                 for (index, (sign, bit)) in combination.iter().enumerate() {
-                    positive &= sign;
+                    positive ^= !sign; // accumulate the sign; global_sign *= local_sign
                     if let Some(bit) = bit {
                         paulis.push(*bit);
                         indices.push(view.indices[index]);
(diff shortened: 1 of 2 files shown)


crates/accelerate/src/sparse_observable.rs
```diff
@@ -2499,11 +2499,11 @@ impl PySparseObservable {
             // we reverse the order of bits and indices so the Pauli string comes out in
             // "reading order", consistent with how one would write the label in
             // SparseObservable.from_list or .from_label
-            for bit in view.bit_terms.iter().rev() {
+            for bit in view.bit_terms.iter() {
                 pauli_string.push_str(bit.py_label());
             }
             let py_string = PyString::new(py, &pauli_string).unbind();
-            let py_indices = PyList::new(py, view.indices.iter().rev())?.unbind();
+            let py_indices = PyList::new(py, view.indices.iter())?.unbind();
```
jakelishman: Didn't we deliberately reverse these, according to the comment above?

jakelishman (commented):
Thanks for the extra tests - not a great look we didn't spot it first time round haha!

crates/accelerate/src/sparse_observable.rs
Cryoris: Whoops no -- on top of catching debugging prints, we now need something to catch debugging-undos 😂 Fixed in 3a5a5ad28

jakelishman (approved):
The tests look right to me, thanks.

PR #13807, opened 2025-02-07
Remove redundant `CODEOWNERS` entries

jakelishman:
### Summary

Shelly and Sasha are both part of terra-core.  Rules that "specialise" a directory to terra-core are redundant, since the blanket rule catches that case.

Naoki is moved from being terra-core to instead being a specialised code-owner of the pulse and scheduling sections, which is more representative of the current reality.

### Details and comments

I also removed team members who have moved away from the core team from terra-core as well.

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -24,11 +24,8 @@
 
 # Qiskit folders (also their corresponding tests)
 providers/            @Qiskit/terra-core @jyu00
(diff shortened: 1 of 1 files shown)


ElePT (approved):
LGTM, the changes in `terra-core` I assume are handled somewhere in the settings, and not via the codeowners file, right?

jakelishman:
Yeah, it's GitHub access management.

PR #13809, opened 2025-02-07
Remove `schedule`, `sequence` and related unit tests

ElePT:
### Summary
This PR removes the `schedule` and `sequence` functions deprecated in qiskit 1.3 as part of the Pulse deprecation effort.

### Details and comments
Part of https://github.com/Qiskit/qiskit/issues/13662


test/python/transpiler/test_calibrationbuilder.py
```diff
@@ -323,60 +224,6 @@ def test_raise(self):
                 qubits = [qubit_map[q] for q in node.qargs]
                 _pass.get_calibration(node.op, qubits)
 
-    def test_ecr_cx_forward(self):
```
mtreinish: These tests could arguably be kept and be rewritten to assert based on the the pulse gates more directly. But since the pass deprecation is in the merge queue currently I think it's fine to just drop this now.

releasenotes/notes/remove-schedule-sequence-a6249577da8d1c86.yaml
```diff
@@ -0,0 +1,7 @@
+---
+upgrade:
+  - |
+    The functions ``sequence`` and  ``schedule`` from the :mod:`.compiler` 
+    module have been removed following their deprecation in Qiskit 1.3.
+    These functions depend on and relate to the Pulse package which is 
+    also being removed in Qiskit 2.0.
```
mtreinish, line 10: I might expand on this a bit more to explain there isn't an alternative to this because it depends on being able to translate circuits to pulse from backend definitions that is not present. But we can do this in the release round-up PR.

mtreinish (dismissed):

releasenotes/notes/remove-schedule-sequence-a6249577da8d1c86.yaml
ElePT: Expanded in [c6f1366](https://github.com/Qiskit/qiskit/pull/13809/commits/c6f1366c68a9d5914217f42420763146a064899e), but we should probably improve it in the release PR (as most of the removal renos that relate to these large removal epics).

test/python/transpiler/test_calibrationbuilder.py
ElePT: This test has actually just been removed in `main`, so problem solved :)

mtreinish (approved):

PR #13811, opened 2025-02-07
[Stretch] Support `const` classical expressions.

kevinhartman:
### Summary
Adds support for `const` classical expressions. This is necessary to properly represent `Duration` and `Stretch` data types, which are always constant expressions.

Everything pretty much works the same, except `expr.Expr` has gained a property called `const` which indicates its const-ness. An expression is constant if all of its operands are constant. The `Value` expression is always constant, while `Var` is (for now) non-const.

### Details and comments

Originally, this PR had introduced const-ness as a property of the type system, itself. This was done to mirror what we do in the QASM3 importer, and was what we (@jakelishman and I) thought was the best approach at the time. However, after the initial review, it became clear that const-ness as a property of the expression (`Expr`) is a much more natural fit for Qiskit's classical expression system.

With const-ness as a property of `Expr`, we can get away with tracking const-ness entirely in the initializers of our expression classes, which simply set their own const-ness based on the const-ness of their operands. The `Value` and `Var` nodes act as our base-case since their const-ness is known, statically. This also means that no special handling is needed for QPY.

The original design notes of the type-based const-ness approach is below for posterity, but it is no longer true:
 
> * `expr.lift` now supports lifting non-classical resource scalars into values with const types, via parameter `try_const=True`.
> * Type ordering now considers a const type to be less than its non-const version.
>  * An implicit `Cast` node is generated when a const expression must be coerced into a non-const context.
>* ~An expression with a const type is never an l-value, even if it is a `Var`.~
>  * EDIT: I decided not to make this decision right now, and to instead block const variables from being added to the circuit via `add_var` and `add_input`.
>  * Binary operations will attempt to lift inputs with matching const-ness.
>    * A `Cast` node is inserted if both operands are `Expr` and only one of them is const.
>  * Index and shift operations will lift their RHS to match const-ness of the LHS.

#### To-do
- [x] Needs QPY tests of const types used with old versions and with current version.


kevinhartman:
> Unlike binary bitwise, a Cast node is currently inserted if both operands are Expr and only one of them is const. This is an effect of the coercion to Bool. Perhaps this needs revisiting.

~If the consensus is that it's fine for two binary operands to differ in const-ness, I can also introduce a `CastKind.NON_CONST` and allow the caller to apply a cast node based on the context. My impression would be that we wouldn't need one anywhere for our current use cases if binary operands don't need the same const-ness.~

EDIT: In offline conversations, we decided that having a cast node is the most consistent choice for our expression system, so there is always an implicit cast node to take a const-typed expression down to its non-const counterpart.

qiskit/circuit/classical/expr/constructors.py
```diff
@@ -113,27 +117,42 @@ def lift(value: typing.Any, /, type: types.Type | None = None) -> Expr:
...
             >>> expr.lift(5, types.Uint(4))
             Value(5, Uint(4))
+
+        Lifting non-classical resource scalars to const values::
+
+            >>> from qiskit.circuit.classical import expr, types
+            >>> expr.lift(7)
+            Value(7, Uint(3, const=False))
+            >>> expr.lift(7, try_const=True)
+            Value(7, Uint(3, const=True))
```
kevinhartman: Should also have `expr.lift(7, Uint(3, const=True))` as an example.

qiskit/circuit/classical/expr/visitors.py
```diff
@@ -276,13 +276,15 @@ def is_lvalue(node: expr.Expr, /) -> bool:
...
-        :class:`~.expr.Var` nodes are always l-values, because they always have some associated
-        memory location::
+        :class:`~.expr.Var` nodes are l-values (unless their resolution type is `const`!), because
+        they have some associated memory location::
 
             >>> from qiskit.circuit.classical import types
             >>> from qiskit.circuit import Clbit
             >>> expr.is_lvalue(expr.Var.new("a", types.Bool()))
             True
+            >>> expr.is_lvalue(expr.Var.new("a", types.Bool(const=False)))
```
kevinhartman:
```suggestion
            >>> expr.is_lvalue(expr.Var.new("a", types.Bool(const=True)))
```

qiskit/circuit/classical/expr/constructors.py
```diff
@@ -91,20 +91,24 @@ def lift_legacy_condition(
...
-def lift(value: typing.Any, /, type: types.Type | None = None) -> Expr:
+def lift(value: typing.Any, /, type: types.Type | None = None, *, try_const: bool = False) -> Expr:
     """Lift the given Python ``value`` to a :class:`~.expr.Value` or :class:`~.expr.Var`.
 
-    If an explicit ``type`` is given, the typing in the output will reflect that.
+    By default, lifted scalars are not const.  To lift supported scalars to const-typed
+    expressions, specify `try_const=True`.
+
+    If an explicit ``type`` is given, the typing in the output will reflect that,
+    including its const-ness. The ``try_const`` parameter is ignored when this is specified.
```
jakelishman: Is there a concern that literals shouldn't always be `const` (also, I think we mean "literal", not "scalar")?

qiskit/circuit/classical/expr/expr.py
```diff
@@ -151,8 +151,9 @@ def new(cls, name: str, type: types.Type) -> typing.Self:
 
     @property
     def standalone(self) -> bool:
-        """Whether this :class:`Var` is a standalone variable that owns its storage location.  If
-        false, this is a wrapper :class:`Var` around a pre-existing circuit object."""
+        """Whether this :class:`Var` is a standalone variable that owns its storage
+        location, if applicable. If false, this is a wrapper :class:`Var` around a
+        pre-existing circuit object."""
```
jakelishman, line 160: is this just a line-wrapping change?  You might have an errant 80-character thing set somewhere haha

qiskit/circuit/classical/expr/visitors.py
```diff
@@ -276,13 +276,15 @@ def is_lvalue(node: expr.Expr, /) -> bool:
             >>> expr.is_lvalue(expr.lift(2))
             False
 
-        :class:`~.expr.Var` nodes are always l-values, because they always have some associated
-        memory location::
+        :class:`~.expr.Var` nodes are l-values (unless their resolution type is `const`!), because
+        they have some associated memory location::
```
jakelishman: Possibly this is clearer as

> non-`const` `` :class:`~.expr.Var` `` nodes are l-values ...

since the `const` bit is pretty fundamental here.

qiskit/circuit/classical/types/ordering.py
```diff
@@ -83,14 +83,39 @@ def order(left: Type, right: Type, /) -> Ordering:
...
+    # but the types differ in const-ness, the non-const variant is greater.
+    # If one type is greater (and thus is the only type that can represent
+    # both) an ordering is only defined if that type is non-const or both
+    # types are const.
+    if left.const is True and right.const is False:
+        if order_ is Ordering.EQUAL:
+            return Ordering.LESS
+        if order_ is Ordering.GREATER:
+            return Ordering.NONE
+    if right.const is True and left.const is False:
```
jakelishman: minor style: seeing `is True` and `is False` is slightly weird to me, given that they're regular booleans.  I'd just have expected `if left.const and not right.const` etc - actually I'm kind of surprised that pylint doesn't complain about it.

qiskit/circuit/classical/types/ordering.py
```diff
@@ -215,11 +244,20 @@ def cast_kind(from_: Type, to_: Type, /) -> CastKind:
...
             >>> types.cast_kind(types.Uint(8), types.Bool())
             <CastKind.IMPLICIT: 2>
+            >>> types.cast_kind(types.Uint(8, const=True), types.Uint(8))
+            <CastKind.IMPLICIT: 2>
             >>> types.cast_kind(types.Bool(), types.Uint(8))
             <CastKind.LOSSLESS: 3>
             >>> types.cast_kind(types.Uint(16), types.Uint(8))
             <CastKind.DANGEROUS: 4>
     """
+    if to_.const is True and from_.const is False:
```
jakelishman: same (very minor) comment about `to_.const ==/is True` and the like

qiskit/circuit/quantumcircuit.py
```diff
@@ -2938,8 +2943,14 @@ def add_input(  # pylint: disable=missing-raises-doc
...
             raise CircuitError("circuits to be enclosed with captures cannot have input variables")
-        if isinstance(name_or_var, expr.Var) and type_ is not None:
-            raise ValueError("cannot give an explicit type with an existing Var")
+        if isinstance(name_or_var, expr.Var):
+            if type_ is not None:
+                raise ValueError("cannot give an explicit type with an existing Var")
+            if name_or_var.type.const:
+                raise CircuitError("const variables are not supported")
+        elif type_ is not None and type_.const:
+            raise CircuitError("const variables are not supported")
```
jakelishman: More specifically, a `const` variable can't possibly be a runtime input.

qiskit/qpy/__init__.py
```diff
@@ -370,6 +370,30 @@ def open(*args):
...
+A :class:`~.types.Type` is encoded by a single-byte ASCII ``char`` that encodes the kind of type,
+followed by a payload that varies depending on the type.  The defined codes are:
+
+======================  =========  =================================================================
+Qiskit class            Type code  Payload
+======================  =========  =================================================================
+:class:`~.types.Bool`   ``b``      One `_Bool const`.
+
+:class:`~.types.Uint`   ``u``      One ``uint32_t width``, followed by one ``_Bool const``.
+======================  =========  =================================================================
```
jakelishman: Maybe want to be slightly clearer in this that this modifies the meanings of the type keys compared to version 9.  Line 392 needs double backticks around ``_Bool const``.

qiskit/qpy/binary_io/value.py
```diff
@@ -378,6 +389,25 @@ def _write_expr_type(file_obj, type_: types.Type):
...
+def _write_expr_type_v14(file_obj, type_: types.Type):
+    if type_.kind is types.Bool:
+        file_obj.write(type_keys.ExprType.BOOL)
+        file_obj.write(
+            struct.pack(formats.EXPR_TYPE_BOOL_PACK_V14, *formats.EXPR_TYPE_BOOL_V14(type_.const))
+        )
+    elif type_.kind is types.Uint:
+        file_obj.write(type_keys.ExprType.UINT)
+        file_obj.write(
+            # TODO: make sure you're calling this correctly
```
jakelishman: lol

qiskit/circuit/classical/types/ordering.py
jakelishman, line 1: I feel like the module-level documentation of `qiskit.circuit.classical.types` might need expanding a little with a description of `const`, and how it refers to compile-time constant values?  I don't think we've described it anywhere user-facing in the docs at the moment at the top level.

releasenotes/notes/const-expr-397ff09042942b81.yaml
```diff
@@ -0,0 +1,21 @@
...
+
+      expr.lift(5, try_const=True)
+      # >>> Value(5, Uint(3, const=True))
+
+    The result type of an operation applied to const types is also const::
+
+      from qiskit.circuit.classical import expr
+
+      expr.bit_and(expr.lift(5, try_const=True), expr.lift(6, try_const=True)).type.const
+      # >>> True
```
jakelishman: Super minor: it's weird to me to see the `>>>` used as the _output_ prompts, since that's the standard PS1 _input_ prompt for the Python interpreter haha

test/python/circuit/test_circuit_vars.py
jakelishman, line 1: I seem to have 🇬🇧'd all over this file and now infected you haha

jakelishman (commented):
Minor comments aside, this looks solid to me, thanks.  I wouldn't say I feel 100% confident on whether `const` is part of the type or part of the expression (I know we talked about it offline), but I think the way you've done it is the better way to go for us - I think it's easier on us later to have it all very explicit in the structure.

qiskit/circuit/classical/expr/constructors.py
kevinhartman: There are two things I was thinking about here:

1) In OpenQASM, the operands of a binary expression are always first cast to the greater of the two types. Following that pattern, if literal types are _always_ const, then that would mean we'd always need a cast node in a binary operation between a `Var` and a literal, which felt weird.
2) Lifting literals to non-const by default was the more API compatible choice (e.g. we can still write expressions that don't use const types with pre-14 QPY).

The way we've got it in this PR, `expr.lift` will lift literals to non-const by default. But, any of our API surface that we consider to be accepting an expression in a "const context" can lift its argument as const explicitly with `try_const=True`. All binary operations where one side is already a const-typed expression do this, and perhaps we'd do this in an eventual `QuantumCircuit.add_const_var(...)` as well.

qiskit/circuit/classical/expr/expr.py
kevinhartman: I probably edited this and then "unedited" it, but left the line-wrap change. Sorry for the unnecessarily confusing diff here lol

qiskit/circuit/classical/expr/visitors.py
kevinhartman: I believe this comment is actually no longer valid, since I decided not to "decide" const-typed expressions can never be lvalues. Originally, I'd changed the implementation of `is_lvalue` to return `False` for any const-typed expression.

The reason for my deferral was that it seemed reasonable to me to use a `store` instruction to initialize a const variable, even though we block it at the moment. I'm not sure if it's valid in QASM to initialize a constant without an initializer, e.g.:

```qasm
duration d;
d = 1200dt;
```

qiskit/qpy/__init__.py
kevinhartman: I've changed this around a bit in the later PRs. If it's alright, it might be easier for us to keep this as is and then make sure that the final PR reflects everything in a cohesive and complete manner.

qiskit/circuit/classical/types/ordering.py
kevinhartman: Addressed in f1dc1a150ed8a1875bca3497cd225483461bd501.

qiskit/circuit/classical/types/ordering.py
kevinhartman: Addressed in f1dc1a150ed8a1875bca3497cd225483461bd501.

qiskit/circuit/classical/expr/visitors.py
jakelishman: Yeah, I don't know if that's valid or not.  But tbh, even if it _is_ valid, I'd just rather not support it in Qiskit - a few languages let you forward-declare storage, and initialise it in some scope, as long as the set of initialising statements dominates the set of statements using the variable.  Rust is one of those - it's valid to have `let x;` and then assign to `x` in some scope, provided the structure of the program guarantees that `x` is assigned if it's used.

I'd rather just not support that in Qiskit, just to reduce complexity.  If we gain more complex handling for classical control flow analysis later, we can always revisit.

qiskit/qpy/__init__.py
jakelishman: I'm fine to defer, but can you open an issue tagged for the milestone about what you mean, just to remind us to do it as part of the documentation in the final PR?

qiskit/circuit/classical/expr/constructors.py
jakelishman: tbh, I don't think OpenQASM's type system (especially around `const`) is that well defined.  I guess my question is whether we think "literals may be either type `const uint[n]` _or_ `uint[n]`" is more or less weird than "we have implicit `Cast` nodes that only throw away `const`-ness".

I'm fine to leave it like this - it's probably the best choice already, and _if_ there's a problem, there's probably a lot bigger demons lurking in our type system elsewhere, since the initial thing was kind of just put together on vibes.

qiskit/qpy/__init__.py
kevinhartman: Added issue https://github.com/Qiskit/qiskit/issues/13927

qiskit/circuit/classical/types/ordering.py
```diff
@@ -83,14 +83,39 @@ def order(left: Type, right: Type, /) -> Ordering:
...
+        if order_ is Ordering.EQUAL:
+            return Ordering.LESS
+        if order_ is Ordering.GREATER:
+            return Ordering.NONE
+    if right.const and not left.const:
+        if order_ is Ordering.EQUAL:
+            return Ordering.GREATER
+        if order_ is Ordering.LESS:
+            return Ordering.NONE
+    return order_
```
jakelishman: The ordering relations are correct here, but I'm just thinking that the introduction of `const` means that implicit promotion rules for standard binary operations would no longer follow the tacit assumption that both left and right types coerce to either `left.type` or `right.type`.  For example, `const uint[32]` and `uint[16]` _should_ have compatible promotion behaviour - we should join on `uint[32]`, but `uint[32]` isn't either of the two types.  It _is_ the join of the two types, though - `const uint[32] <= uint[32]` and `uint[16] <= uint[32]`, and for any type `t` where `t >= uint[32]`, `t` is also `>=` `const uint[32]` and `uint[16]`.

So the type ordering as modified here is correct, I'm just wondering if we're breaking assumptions elsewhere.  I _think_ it's all fine, because the only time we do implicit promotion in Qiskit is when lifting the operands, and I don't think there's any valid way to both invoke the type inference _and_ for one variable to be `const uint[16]` and the other to be `uint[8]` - in any situation where that would happen (e.g. `expr.bit_and(1000, expr.Var("b", types.Uint(8)))`), the type promoter rejects the input because the integer literal (the only way to get the `const`) requires more width than the `Var`.

I don't think there's anything to do here.  Just commenting for posterity.

qiskit/circuit/classical/types/__init__.py
```diff
@@ -40,6 +40,10 @@
 .. autoclass:: Bool
 .. autoclass:: Uint
 
+All types have a :attr:`~Type.const` field to indicate their const-ness. When the result type of
+an expression is constant, the expression is considered a constant expression. Constant
+expressions can be used in certain contexts that aren't valid for runtime-initialized variables.
```
jakelishman: Thanks for the text.

Can you comment on briefly on what we mean by "const-ness", just to make clear we mean something more like the Rust `const` or C++ `constexpr` (can be compile-time evaluated) than the C `const` (pinky promise not to write to it at runtime through this identifier, unless I change my mind later)?

test/python/circuit/classical/test_expr_constructors.py
```diff
@@ -68,15 +68,27 @@ def test_value_lifts_qiskit_scalars(self):
 
     def test_value_lifts_python_builtins(self):
         self.assertEqual(expr.lift(True), expr.Value(True, types.Bool()))
+        self.assertEqual(expr.lift(True, try_const=True), expr.Value(True, types.Bool(const=True)))
```
jakelishman: Just looking through the tests, I'm wondering if a future extension will be to add thread-local context variables for the `expr` constructors to allow contextual control over `try_const`, just because I'm vaguely thinking that it's slightly tricksy that the default user-facing `lift` is `try_const=False`, but the binary constructors use `try_const=True` in more places.

I think that's all fine for now, just trying to record for posterity, since I imagine we'll come back to look at the context for the `const` inclusion at some point in the future.

qiskit/circuit/classical/expr/constructors.py
kevinhartman: > literals may be either type const uint[n] or uint[n]

This might just be an artifact of having the const-ness as part of the type rather than the expression, since it allows a `Value` expression to have a non-const type. If const-ness were instead an attribute of expression, then `Value` could be made to always report itself as const.

Just a thought. It's _probably_ too late to change this, but I'm happy to try if you think it's the right idea (and it's permitted by the feature freeze lol).

qiskit/circuit/classical/expr/visitors.py
kevinhartman: Sounds good to me. I thought of the `let x;` pattern in Rust too, but then I realized that Rust doesn't allow `const x;`, which is closer to the semantics I'm alluding to here—it requires an inline initializer.

test/python/circuit/classical/test_expr_constructors.py
```diff
@@ -168,6 +204,31 @@ def test_binary_functions_lift_scalars(self, function, left, right):
...
+        Note that logical operators (e.g. logic_and, logic_or) are excluded since these lift operands
+        independently."""
+        self.assertEqual(
+            function(expr.lift(left, try_const=True), right),
+            function(expr.lift(left, try_const=True), expr.lift(right, try_const=True)),
+        )
+        self.assertEqual(
+            function(left, expr.lift(right, try_const=True)),
+            function(expr.lift(left, try_const=True), expr.lift(right, try_const=True)),
+        )
```
jakelishman: I think perhaps this is indicating missing logic in `logic_and` and `logic_or` - it always lifts to `Bool(const=False)`, but `logic_and(False, True)` should have type `Bool(const=True)`, as should `logic_and(False, expr.Var.new("a", types.Bool(const=True))`.  `logic_not` does the same thing, I think - `logic_not(False)` should be `Bool(const=True)`.

jakelishman (commented):
Thanks for this - I think it's about ready to land, bar the merge conflict and the couple of minor comments.  Two of them are just musings about the effects of the type system and non-actionable, one is super minor documentation, and the other I think is a missing bit of logic in the logical-operation type inference.

I feel like there's an easy chance I could have missed things in the tests, but that's what bugfixes are for haha, and it _seems_ correct.

test/python/circuit/classical/test_expr_constructors.py
kevinhartman: You're right, and that's actually a bug I think from when I switched the default of `lift`'s `try_const` to `False`.

kevinhartman: Hmm, actually, this was intended. It's supposed to be consistent with `lift` going for non-const by default. Since the operands are lifted independently, this is the consistent behavior. We could instead do what the other operations do, which is to use `_lift_binary_operands` to agree on operand const-ness before ultimately coercing both the `bool`. But this would also widen integers to the same width, which isn't the current behavior. Thoughts?

jakelishman: > the operands are lifted independently

This is just an implementation detail of the pre-`const` type system that there was no influence of one binary operand on the other - I don't think there's any indication that it's an API contract of the inference?  I'd expect the inference signature of `&&` to be `const bool * const bool -> const bool` as a user, and similarly for `! : const bool -> const bool`.

If you're concerned about the consistency of `lift`, I think it might be indicative that we might want to get rid of `try_const`, or default it to `True`?

I think the QPY concern could in theory be handled separately (I agree it's a valid one), such as threading a keyword arg through `dump` like `allow_lossy_expr_types` or something?

qiskit/circuit/classical/types/__init__.py
kevinhartman: Done in 166d7b2

qiskit/circuit/classical/types/ordering.py
kevinhartman: This should no longer be a concern after making const-ness a property of `Expr` instead of `Type` 🙂.

test/python/circuit/classical/test_expr_constructors.py
kevinhartman: This is also all resolved with the change to `Expr.const`

test/python/circuit/classical/test_expr_constructors.py
kevinhartman: No longer applicable with `Expr.const`.

kevinhartman:
Thanks for the review, @jakelishman.

As we've discussed offline, your comments got me thinking that tracking const-ness as a property of `Expr` rather than as part of our type system would be a more natural fit for Qiskit. 

For others, my reasoning was:

The main idea is that we move `const` to `Expr` and let the constructors of `Expr` subclasses initialize it. `Value` initializes `const` to `True`, always. `Var` initializes `const` to `False`, always. Operation expressions initialize it based on their operands. E.g. for  `Binary`, we set `self.const = left.const and right.const`.

The benefits are:
1. We don’t need to change the logic of `lift` at all, and there’s no fussing with const-ness needed (no need to decide a default for `try_const`). And, there’s no future need to contextually lift something with const-ness in mind. An expression is const if all of its parts are, simple as that.
2. We don’t have the issue with ordering where the join of two types can be some third type.
3. Literals are always const, since `Value` is always const.
4. No QPY changes are needed whatsoever, since the expression tree is put back together from the bottom up, starting with the leaf nodes `Var` and `Value` whose const-ness we know statically.
5. Zero changes are needed to the expression constructors.
6. We avoid a potential weirdness that you’ll see in the timing types PR, which is that `Stretch` and `Duration` are always const. If some kind of runtime duration gets added later, durations will be const by default, but opt-in non-const, which is backwards from other types.

The branch has now been updated with these changes. This _seems_ like a big change, but really we've just moved an attribute from one class to another, and everything else is largely removals.

qiskit/circuit/classical/expr/constructors.py
jakelishman: For posterity: having the join of the two types not be equal to one of the two isn't a problem for a type system (for example, C has this property, and anything with ad-hoc sum types like Python's `Union` can do in certain contexts too), I was more meaning I worried we had assumptions elsewhere that it held.

But since [8021e00](https://github.com/Qiskit/qiskit/pull/13811/commits/8021e00850116efd16aa6b3688fac46f9d289afe), it's no longer relevant anyway, so no worries.

qiskit/qpy/__init__.py
jakelishman: Issue #13927 is obsoleted by 8021e00850116efd16aa6b3688fac46f9d289afe, I think?

qiskit/circuit/classical/expr/__init__.py
```diff
@@ -39,14 +39,18 @@
 
 These objects are mutable and should not be reused in a different location without a copy.
 
+All :class`Expr` instances define a boolean :attr:`~Expr.const` attribute, which indicates
+whether the expression can be evaluated at compile-time. Most expression classes infer this
```
jakelishman:
```suggestion
whether the expression can be evaluated at compile time. Most expression classes infer this
```

releasenotes/notes/const-expr-397ff09042942b81.yaml
```diff
@@ -0,0 +1,24 @@
+---
+features_circuits:
+  - |
+    The classical realtime-expressions module :mod:`qiskit.circuit.classical` can now represent
+    constant expressions. The :class:`~.expr.Expr` class now has a bool
+    :attr:`~.expr.Expr.const` attribute which indicates the expression's const-ness. This allows
+    us to enforce that expressions in certain contexts must be possible to evaluate at compile-time.
```
jakelishman:
```suggestion
    us to enforce that expressions in certain contexts must be possible to evaluate at compile time.
```

releasenotes/notes/const-expr-397ff09042942b81.yaml
```diff
@@ -0,0 +1,24 @@
...
+
+    All :class:`~.expr.Var` expressions are considered to be non-const, while all :class:`~.expr.Value`
+    expressions are const.
+
+    An expression comprised only of other const expressions is also const::
+
+      from qiskit.circuit.classical import expr
+
+      expr.bit_and(5, 6).const
+      # True
```
jakelishman:
```suggestion
      assert expr.bit_and(5, 6).const
```

releasenotes/notes/const-expr-397ff09042942b81.yaml
```diff
@@ -0,0 +1,24 @@
...
+
+    An expression that contains any non-const expression is non-const::
+
+      from qiskit.circuit.classical import expr, types
+
+      expr.bit_and(5, expr.Var.new("a", types.Uint(5)).const
+      # False
```
jakelishman:
```suggestion
      assert not expr.bit_and(5, expr.Var.new("a", types.Uint(5)).const
```

jakelishman (commented):
I think you're right, this will end up a lot cleaner, and an expression being `const` doesn't _oblige_ us to evaluate it at compile time.  I think this somewhat formalises `Value` to mean "immediate value", which I don't think I connected the dots on in the first iteration of the `expr` system, but I think it makes a lot of sense: `Var` is associated with memory locations; `Value` is immediate values; and then we've got a nice framework to fit stretches and durations into.

Super minor typographical comments only.

releasenotes/notes/const-expr-397ff09042942b81.yaml
jakelishman: This is stylistic choice, I just suggested it like this because I think the `expr` documentation does things like this, and it would (in theory) then immediately play nicely with automated running of the code snippets.

Feel free to ignore if you prefer anyway.

qiskit/qpy/__init__.py
jakelishman: Well, if not obsoleted, then at least the changes to QPY should now be localised to a later PR I think.

jakelishman (approved):

PR #13813, opened 2025-02-09
Port HighLevelSynthesis pass to rust

### LINKED ISSUE #12211: Port `HighLevelSynthesis` to rust
This issue tracks porting the `HighLevelSynthesis` pass to rust as part of the https://github.com/Qiskit/qiskit/issues/12208 epic. This pass in particular will still always require a fairly large python component, as the high level synthesis plugin interface necessitates a Python execution mode. One option is we can call Python from rust to pass a unitary array to the plugins and then operate exclusively from rust otherwise as well. The other thing worth investigating is for all the built-in synthesis methods we can have a rust only execution mode if the plugins are selected and only use python to call the external plugins if the `HLSConfig` contains any plugins outside the set of what's included in Qiskit.

alexanderivrii:
### Summary

This PR ports [most of] the ``HighLevelSynthesis`` transpiler pass (or HLS, for short) to Rust, addressing #12211. 

The `test_utility_scale.py` script provides several examples where HLS does not need to do anything. Running this script on my laptop, reduces the total HLS runtime from about 25% of the total transpile time to just 0.3%.

This PR is based on #13777 (a small fix for calling the ``definition`` methods from within the Rust space) and #13605 (reimplementing HLS in Python). Technically, we can skip merging #13605  and merge PR directly, I don't have a strong preference for what is the better option.

### Details and comments

The part of the pass that still remains in Python is the code for synthesizing operations using plugins. As an optimization, we precompute the set of (names of) operations with available plugins and we only call this Python code when an operation is in this set. Note that annotated operations are also handled via a plugin. 

The reason why porting the Python code that handles plugins is very tricky is because we allow passing arbitrary python classes in the ``HLSConfig``, something as follows: ``hls_config = HLSConfig("clifford"=[MyPythonCliffordSynthesisPlugin()])``.

In the case that the HighLevelSynthesis pass needs to do something (and in particular call synthesis plugins implemented in python), the runtime improvements are on the scale of 20%--50%.


crates/circuit/src/operations.rs
```diff
@@ -2860,8 +2848,21 @@ impl Operation for UnitaryGate {
...
+                .unwrap();
+            match py_op.getattr(py, intern!(py, "definition")) {
+                Ok(definition) => definition
+                    .getattr(py, intern!(py, "_data"))
+                    .ok()?
+                    .extract::<CircuitData>(py)
+                    .ok(),
+                Err(_) => None,
+            }
+        })
```
mtreinish, line 2863: Do we need a definition for `UnitaryGate` at all? I purposely didn't define it here because the python space one is to call unitary synthesis. In rust it's a bit of a layer violation to do that since it lives in the accelerate crate. While going through python avoids that issue, I'm wondering if it's really necessary because it's not a definition in the strict sense like it is for a concrete gate.

crates/circuit/src/circuit_data.rs
```diff
@@ -1135,6 +1135,29 @@ impl CircuitData {
...
+            qubits,
+            clbits,
+            params,
+            extra_attrs: ExtraInstructionAttributes::default(),
+            #[cfg(feature = "cache_pygates")]
+            py_op: OnceLock::new(),
+        });
+        Ok(())
+    }
+
```
alexanderivrii: I needed this function in addition to the already existing function `push_standard_gate`, but the code is literally identical. I am wondering if these two versions can be somehow combined?

mtreinish: I think the difference is that `push_standard_gate is optimized a bit for gates. We know that a gate has no clbits so we don't bother calling `cargs_interner.insert()`. The idea was to save the overhead of having to repeatedly query the interner when we know ahead of time that it's going to be the same default value for an empty list. 

There is probably a path we could add to unify the functions, maybe change the method internals to match based on the type of  `operation`  so we avoid the hash lookup if it's a `StandardGate`, `PyGate`, or `UnitaryGate`. But maybe look into that in a separate PR to avoid too much complexity here since this is already pretty large.

raynelfss:
Just took a quick glance at your code here and it looks promising. Just wanted to ask if you've been able to benchmark this change and whether performance is impacted in any way.

alexanderivrii:
Thanks @raynelfss, I have some performance data in this PR's summary. Essentially, we get a very nice speed-up when `HighLevelSynthesis` does not need to do anything - with the logic for determining that this is the case now being fully in rust. In the other extreme case, when the circuit contains a huge number of high-level-objects that need to be synthesized using the plugin mechanism, the speedup is only about 20% - 50%, and this is because we still need to call the Python code to use the plugins. I could also create a few interim examples, if you wish.

raynelfss:
That's pretty good! I was a little concerned about us storing a `HighLevelSynthesisData` object in rust space and how it could affect serialization performance. However, it seems that most of the objects stored there are still in Python after all, so we skip those conversions. Well done!

crates/accelerate/src/high_level_synthesis.rs
```diff
@@ -171,82 +213,807 @@ impl QubitTracker {
...
+        self.min_qubits
+    }
+
+    fn get_unroll_definitions(&self) -> bool {
+        self.unroll_definitions
+    }
+
+    fn set_device_insts(&mut self, device_insts: HashSet<String>) {
+        self.device_insts = device_insts;
+    }
```
raynelfss: You could add the `#[pyo3(get, set)]` specifier in the definition of the struct and these methods would be automatically generated for you.

For example:
```rust
#[derive(Clone, Debug)]
struct HighLevelSynthesisData {
    // The high-level-synthesis config that specifies the synthesis methods
    // to use for high-level-objects in the circuit.
    // This is only accessed from the Python space.
    #[pyo3(get)] // If the item will only get retrieved and not modified.
    hls_config: Py<PyAny>,
    ...
    // Supported instructions in case that target is not specified.
    #[pyo3(get, set)] // If the attribute is to ever be modified from python.
    device_insts: HashSet<String>,
    ...
}

File: crates/accelerate/src/high_level_synthesis.rs [hunk 4]
```diff
@@ -171,82 +213,807 @@ impl QubitTracker {
...
+    qubits: &[Qubit],
+) -> bool {
+    match &data.borrow().target {
+        Some(target) => {
+            let target = target.borrow(py);
+            if target.num_qubits.is_some() {
+                if data.borrow().use_qubit_indices {
+                    let physical_qubits = qubits
+                        .iter()
+                        .map(|q| PhysicalQubit(q.index() as u32))
```
raynelfss: To avoid casting to a u32, you could just directly use `Qubit.0` which already stores a u32.

crates/accelerate/src/high_level_synthesis.rs
```diff
@@ -171,82 +213,807 @@ impl QubitTracker {
...
+}
+
+/// Check whether an operation does not need to be synthesized.
+fn definitely_skip_op(
+    py: Python,
+    data: &Bound<HighLevelSynthesisData>,
+    op: &PackedOperation,
+    qubits: &[Qubit],
+) -> bool {
+    if qubits.len() < data.borrow().min_qubits {
```
raynelfss: If you believe you will call the `borrow()` method more than once, make sure to just do it once and store the reference in a variable.
```suggestion
    let borrowed_data = data.borrow();
    if qubits.len() < borrowed_data.min_qubits {
```

crates/accelerate/src/high_level_synthesis.rs
```diff
@@ -171,82 +213,807 @@ impl QubitTracker {
...
+/// over more qubits than the input circuit.
+///
+/// The function also updates in-place the qubit tracker, which keeps track of the
+/// state of each global qubits (whether it's clean, dirty, or cannot be used).
+fn run_on_circuitdata(
+    py: Python,
+    input_circuit: &CircuitData,
+    input_qubits: &[usize],
+    data: &Bound<HighLevelSynthesisData>,
+    tracker: &Bound<QubitTracker>,
```
raynelfss: You could technicaly still accept a `&mut QubitTracker` since it is a `pyclass`. Is there a reason why we accept a `Bound<'_, QubitTracker>` instead?

crates/accelerate/src/high_level_synthesis.rs
```diff
@@ -171,82 +213,807 @@ impl QubitTracker {
...
+    fn set_device_insts(&mut self, device_insts: HashSet<String>) {
+        self.device_insts = device_insts;
+    }
+
+    fn __str__(&self) -> String {
+        format!(
+            "HighLevelSynthesisData(hls_config: {:?}, hls_plugin_manager: {:?}, hls_op_names: {:?}, coupling_map: {:?}, target: {:?},  equivalence_library: {:?}, device_insts: {:?}, use_qubit_indices: {:?}, min_qubits: {:?}, unroll_definitions: {:?})",
+            self.hls_config, self.hls_plugin_manager, self.hls_op_names, self.coupling_map, self.target, self.equivalence_library, self.device_insts,  self.use_qubit_indices, self.min_qubits, self.unroll_definitions
         )
     }
```
raynelfss: You could have this function accept a `Bound` instance of `self`, use `getattr` to get the attributes, and use `.repr()?` to get the string representations of the python objects correctly. I also think `Bound` objects do implement `Display` so you you could skip the `repr` call. But that's up to you.

raynelfss (commented):
I've done a quick pass through the rust-level code here. Most of it looks good, I just had a couple of questions and suggestions.

crates/accelerate/src/high_level_synthesis.rs
```diff
@@ -171,82 +213,807 @@ impl QubitTracker {
...
+                let old_blocks_as_bound_obj = py_inst.instruction.bind(py);
+
+                // old_blocks_py keeps the orignal QuantumCircuit's appearing within control-flow ops
+                // new_blocks_py keeps the recursively synthesized circuits
+                let old_blocks_py = old_blocks_as_bound_obj.getattr(intern!(py, "blocks"))?;
+                let old_blocks_py = old_blocks_py.downcast::<PyTuple>()?;
+                let mut new_blocks_py: Vec<Bound<PyAny>> = Vec::with_capacity(old_blocks_py.len());
+
+                // We do not allow using any additional qubits outside of the block.
+                let block_tracker = tracker.clone();
```
raynelfss: I'm not sure this will work as intended. When you call clone on a `Bound` instance, what it does is increast the reference count in Python. But it is still pointing towards the same object. If you intended to modify the original one, then this should be fine. But if your intention was to make a clone of the original object and then modify it, this might not be doing that.

raynelfss (commented):
Here's one more thing I found during my review. I'm not sure it's causing the intended effects. Other than this, the rust code looks good and it will benefit greaty from #13860. :rocket:

crates/accelerate/src/high_level_synthesis.rs
```diff
@@ -171,82 +213,807 @@ impl QubitTracker {
...
+    if op.control_flow() {
+        return false;
+    }
+
+    // If the operation is natively supported, we can skip it.
+    if instruction_supported(py, data, op.name(), qubits) {
+        return true;
+    }
+
+    // If there are avilable plugins for this operation, we should try them
```
eliarbel:
```suggestion
    // If there are available plugins for this operation, we should try them
```

crates/accelerate/src/high_level_synthesis.rs
```diff
@@ -171,82 +213,807 @@ impl QubitTracker {
...
+        // and mark all the usable qubits as dirty. This is done in order to avoid complications
+        // that different subcircuits may choose to use different auxiliary global qubits, and to
+        // avoid complications related to tracking qubit status for while- loops.
+        // In the future, this handling can potentially be improved.
+        if inst.op.control_flow() {
+            let quantum_circuit_cls = QUANTUM_CIRCUIT.get_bound(py);
+            if let OperationRef::Instruction(py_inst) = inst.op.view() {
+                let old_blocks_as_bound_obj = py_inst.instruction.bind(py);
+
+                // old_blocks_py keeps the orignal QuantumCircuit's appearing within control-flow ops
```
eliarbel:
```suggestion
                // old_blocks_py keeps the original QuantumCircuit's appearing within control-flow ops
```

crates/accelerate/src/high_level_synthesis.rs
```diff
@@ -171,82 +213,807 @@ impl QubitTracker {
...
+                // new_blocks_py keeps the recursively synthesized circuits
+                let old_blocks_py = old_blocks_as_bound_obj.getattr(intern!(py, "blocks"))?;
```
eliarbel, line 519: Why not using `Instruction::blocks` here?

crates/accelerate/src/high_level_synthesis.rs
```diff
@@ -171,82 +213,807 @@ impl QubitTracker {
...
+
+    Ok(output_circuit_and_qubits)
+}
+
+/// Synthesizes an operation.
+///
+/// This function is currently called by the default plugin for annotated operations to
+/// synthesize the base operation.
+#[pyfunction]
+#[pyo3(signature = (py_op, input_qubits, data, tracker))]
```
eliarbel:
```suggestion
#[pyo3(name = "synthesize_operation", signature = (py_op, input_qubits, data, tracker))]
```

The `py_` prefix is a Rust-space convention that should be omitted from the Python name.

crates/accelerate/src/high_level_synthesis.rs
```diff
@@ -171,82 +213,807 @@ impl QubitTracker {
...
+    )
+}
+
+/// Runs HighLevelSynthesis transpiler pass.
+///
+/// This is the main function called from the Python space. If the pass does not need
+/// to do anything, it returns None, meaning that the DAG should remain unchanged.
+/// Otherwise, the new DAG is returned.
+#[pyfunction]
+#[pyo3(signature = (dag, data, qubits_initially_zero))]
```
eliarbel:
```suggestion
#[pyo3(name = "run_on_dag", signature = (dag, data, qubits_initially_zero))]
```

As with `py_synthesize_operation`.

qiskit/transpiler/passes/synthesis/high_level_synthesis.py
```diff
@@ -270,585 +271,157 @@ def run(self, dag: DAGCircuit) -> DAGCircuit:
...
+            plugin_method = plugin_specifier
 
-        The arguments ``num_clean_ancillas`` and ``num_dirty_ancillas`` specify
-        the number of clean and dirty qubits available to synthesize the given
-        operation. A synthesis method does not need to use these additional qubits.
+        # The additional arguments we pass to every plugin include the list of global
+        # qubits over which the operation is defined, high-level-synthesis data and options,
+        # and the tracker that tracks the state for global qubits.
+        #
+        # Note: the difference between the argument "qubits" passed explicitely to "run"
```
eliarbel:
```suggestion
        # Note: the difference between the argument "qubits" passed explicitly to "run"
```

qiskit/transpiler/passes/synthesis/hls_plugins.py
```diff
@@ -1764,6 +1776,153 @@ def run(self, high_level_object, coupling_map=None, target=None, qubits=None, **
...
+                raise TranspilerError(
+                    "HighLevelSynthesis: problem with the default plugin for annotated operations."
+                )
+
+            return synthesized
+
+        return None
+
+
+def _apply_annotations(circuit: QuantumCircuit, modifiers: list[Modifier]) -> QuantumCircuit:
```
eliarbel: Although this function does not rely on instance state, I think it would be better to make it part of the `AnnotatedSynthesisDefault` class as it's only used by it. This keeps the code a bit more organized for things like capturing design intent, refactoring,  deprecating, collapsing code in IDEs an the like.

qiskit/transpiler/passes/synthesis/hls_plugins.py
```diff
@@ -1764,6 +1776,153 @@ def run(self, high_level_object, coupling_map=None, target=None, qubits=None, **
...
+        else:
+            raise TranspilerError(f"HighLevelSynthesis: Unknown modifier {modifier}.")
+
+    if not isinstance(circuit, QuantumCircuit):
+        raise TranspilerError("HighLevelSynthesis: incorrect output of 'apply_annotations'.")
+
+    return circuit
+
+
+def _instruction_to_circuit(op: Operation) -> QuantumCircuit:
```
eliarbel: Like `_apply_annotations`, I would make this function part of the class.

releasenotes/notes/improve-hls-pass-620389dde24e9707.yaml
```diff
@@ -0,0 +1,6 @@
+---
+features_transpiler:
+  - |
+    The :class:`.HighLevelSynthesis` transpiler pass now synthesizes
+    objects of type :class:`~.AnnotatedOperation` objects via the
```
eliarbel:
```suggestion
    objects of type :class:`~.AnnotatedOperation` via the
```

eliarbel (changes requested):
Nice work @alexanderivrii, the speedup of the fast-path is really cool!

Overall this looks good to me given that the main motivation here is the performance improvement, so I think we should go with this PR as is for 2.0, after the comments raised by @raynelfss and the minor changes I commented in line here are addressed. 

When it comes to the overall workflow of running HLS plugins from within Rust, I have a slight discomfort with how this interaction is implemented currently, namely with `HighLevelSynthesisData` carrying some Python side only data and the fact that `_methods_to_try` and `_synthesize_op_using_plugins` are not part of the `HighLevelSynthesis` class anymore. I don't think we should address this aspect at this point of the release cycle and in this PR, but for the sake of recording my thoughts I'm dropping a design option here: if we end up having the final workflow of the plugin mechanism as in this PR, i.e. it remains in Python and called from Rust, I would make `HighLevelSynthesisData` contain only data which is needed in the Rust-side and keep the rest (like `HLSConfig`) as instance attributes of `HighLevelSynthesis`. Then `_methods_to_try` and `_synthesize_op_using_plugins` would become part of the class and would be called from the Rust side using `call_method1` (this requires passing `HighLevelSynthesis`'s `self` to `run_on_dag` for `call_method1` to call on). Another option suggested by Jake is to pass Python closures of `_methods_to_try` and `_synthesize_op_using_plugins` with the Python-side data, and pass those to Rust as callbacks. But again, this might be a nice design story to think about going forward, but **certainly** not in this PR.

crates/accelerate/src/high_level_synthesis.rs
alexanderivrii: Thanks! This makes the code so much cleaner and easier to use from the Python side. Done in [8d20d8c](https://github.com/Qiskit/qiskit/pull/13813/commits/8d20d8c1f5be0be033611fe59792b3b661a8b22a).

crates/accelerate/src/high_level_synthesis.rs
alexanderivrii: Nice catch! At some point in the past this was indeed `&mut QubitTracker`, but then I thought that I have to convert it to `Bound<...>`. But now I see that I don't, so making it `&mut QubitTracker` as you are suggesting.

crates/accelerate/src/high_level_synthesis.rs
alexanderivrii: Great catch! This could have been a problem indeed, but I think it's gone now, now that `tracker` is a `QubitTracker` and not a `Bound<QubitTracker>`.

qiskit/transpiler/passes/synthesis/hls_plugins.py
alexanderivrii: Nice idea, I am doing this! (In the follow-up PR #13916 I have also moved the annotated synthesis plugin to its own file, but this is orthogonal to the current suggestion).

crates/accelerate/src/high_level_synthesis.rs
alexanderivrii: The main reason is that I was learning by looking at the code in `gate_direction.rs` which also uses `getattr("blocks")` 😄.

But thinking about this now, `Instruction.blocks` seems to return `Vec<CircuitData>` which unfortunately loses important attributes of `QuantumCircuit` (on the Python side), like register names and variables. My current approach (after struggling with these issues for a really long time) to this is to call the Python's `copy_empty_like` to make sure these attributes propagate to the new `QuantumCircuit`s. Bottom-line, I would very happy to simplify this overly complex code, but currently I don't know how.

alexanderivrii:
Thanks @raynelfss and @eliarbel (and transitively @jakelishman) for the feedback! I have applied all of Ray's and Eli's review suggestions, with the exception of modifying `HighLevelSynthesisData.__str()__` (since this functionality is pure development) and using `Instruction::blocks` (since I believe this is not sufficient for me). I really like the ideas on how to further refactor code in the future.

crates/accelerate/src/high_level_synthesis.rs
```diff
@@ -171,82 +213,767 @@ impl QubitTracker {
...
     }
+}
 
-    /// Returns the local-to-global mapping
-    fn to_global_mapping(&self) -> Vec<usize> {
-        self.local_to_global.clone()
+/// Check whether an operation is natively supported.
+fn instruction_supported(
+    py: Python,
+    data: &Bound<HighLevelSynthesisData>,
```
Cryoris, line 365: With a `Bound` object in the signature you could skip the explicit `py: Python` argument here and just have 
```rust
let py = data.py();
```
on the first line -- but this is just optional, I'm also fine as is 🙂

crates/accelerate/src/high_level_synthesis.rs
```diff
@@ -171,82 +213,767 @@ impl QubitTracker {
...
+    equivalence_library: Option<Py<EquivalenceLibrary>>,
+
+    // Supported instructions in case that target is not specified.
+    #[pyo3(get)]
+    device_insts: HashSet<String>,
+
+    // A flag indicating whether the qubit indices of high-level-objects in the
+    // circuit correspond to qubit indices on the target backend.
+    #[pyo3(get)]
+    use_qubit_indices: bool,
```
Cryoris: Would this be a good moment to rename this to something more specific, like `use_physical_indices` or `is_routed`? Especially since now we use global indices everywhere this argument name is a bit confusing 🙂

crates/accelerate/src/high_level_synthesis.rs
```diff
@@ -171,82 +213,767 @@ impl QubitTracker {
...
+    py: Python,
+    dag: &DAGCircuit,
+    data: &Bound<HighLevelSynthesisData>,
+    qubits_initially_zero: bool,
+) -> PyResult<Option<DAGCircuit>> {
+    // Fast-path: check if HighLevelSynthesis can be skipped altogether. This is only
+    // done at the top-level since this does not track the qubit states.
+    let mut fast_path: bool = true;
+
+    for (_, inst) in dag.op_nodes(false) {
```
Cryoris, line 963: This originally comes from Matt, but what about adding an even simpler check before this to see whether all operations in `count_ops()` are in the target basis or equivalence library? Since the op count is already pre-computed this should be very fast -- here we instead have to iterate over all nodes and do additional checks.

crates/accelerate/src/high_level_synthesis.rs
```diff
@@ -171,82 +213,767 @@ impl QubitTracker {
...
+        let qubits = dag.get_qargs(inst.qubits);
+        if !definitely_skip_op(py, data, &inst.op, qubits) {
+            fast_path = false;
```
Cryoris, line 966: Could we avoid the `fast_path` variable if we just directly `return Ok(None)` here?

crates/accelerate/src/high_level_synthesis.rs
```diff
@@ -171,82 +213,767 @@ impl QubitTracker {
...
+fn run_on_circuitdata(
+    py: Python,
+    input_circuit: &CircuitData,
+    input_qubits: &[usize],
+    data: &Bound<HighLevelSynthesisData>,
+    tracker: &mut QubitTracker,
+) -> PyResult<(CircuitData, Vec<usize>)> {
+    if input_circuit.num_qubits() != input_qubits.len() {
+        return Err(TranspilerError::new_err(
+            "HighLevelSynthesis: the input to 'run_on_circuitdata' is incorrect.",
```
Cryoris: It would be nice if we'd raise more explicit errors here (and in other places), especially now that the pass is in Rust. Otherwise it's hard to guess what went wrong 🙂 E.g. something like
```suggestion
           format!(
               "HighLevelSynthesis: number of input qubits ({}) does not match the circuit size ({})",
                input_circuit.num_qubits(),
                input_qubits.len()
            ),
```

crates/accelerate/src/high_level_synthesis.rs
```diff
@@ -171,82 +213,767 @@ impl QubitTracker {
...
+        }
+
+        if inst.op.name() == "reset" {
+            output_circuit.push(py, inst.clone())?;
+            tracker.set_clean(op_qubits);
+            continue;
+        }
+
+        // Check if synthesis for this operation can be skipped
+        let op_qargs: Vec<Qubit> = op_qubits.iter().map(|q| Qubit::new(*q)).collect();
```
Cryoris, line 499: Would it help to change the input type of the qubit argument in `definitely_skip_op` from `&[Qubit]` to `&[u32]`? I may have missed something, but in the two places this is being called, the `u32` version is readily available without needing a clone

crates/accelerate/src/high_level_synthesis.rs
```diff
@@ -171,82 +213,767 @@ impl QubitTracker {
...
+    py: Python,
+    op: &PackedOperation,
+    params: &[Param],
+) -> PyResult<Option<CircuitData>> {
+    match op.view() {
+        OperationRef::Unitary(unitary) => {
+            let unitary: Array<Complex<f64>, Dim<[usize; 2]>> = match unitary.matrix(&[]) {
+                Some(unitary) => unitary,
+                None => return Err(TranspilerError::new_err("Unitary not found")),
+            };
```
Cryoris, line 685: In case you'd like to play with Rust syntax I think you could also write this something like
```rust
let Some(unitary) = unitary.matrix(&[]) else { // return error }
```

crates/accelerate/src/high_level_synthesis.rs
```diff
@@ -171,82 +213,767 @@ impl QubitTracker {
...
+            match unitary.shape() {
```
Cryoris, line 686: Isn't this something that should be taken care of by unitary synthesis at the moment? Otherwise we're duplicating the logic and run the risk of hardcoding the target basis here, whereas the unitary synthesis can handle more general cases, no?

crates/accelerate/src/high_level_synthesis.rs
```diff
@@ -171,82 +213,767 @@ impl QubitTracker {
...
+    // available for various high-level-objects.
+    // This is only accessed from the Python space.
+    #[pyo3(get)]
+    hls_plugin_manager: Py<PyAny>,
+
+    // The names of high-level objects with available synthesis plugins.
+    // This is an optimization to avoid calling python when an object has no
+    // synthesis plugins.
+    #[pyo3(get)]
+    hls_op_names: Vec<String>,
```
Cryoris: Would it make sense to use a HashSet here too, given that we're always checking whether some op name is in this vector?

crates/accelerate/src/high_level_synthesis.rs
```diff
@@ -171,82 +213,767 @@ impl QubitTracker {
...
-    /// Returns the global qubits corresponding to the given local qubit
-    fn to_globals(&self, qubits: Vec<usize>) -> Vec<usize> {
-        qubits.iter().map(|q| self.local_to_global[*q]).collect()
+    Ok(output_circuit_and_qubits)
+}
+
+/// Synthesizes an operation.
+///
+/// This function is currently called by the default plugin for annotated operations to
+/// synthesize the base operation.
```
Cryoris: It might be nice to add what the expected type of `py_op` is, on the Python side 🙂

crates/accelerate/src/high_level_synthesis.rs
Cryoris: Given that there's a lot of Rust-Python back and forth, would it be simpler to write a Python function to call here?

Cryoris (commented):
I'm not quite through yet, but here's a first round of comments 🙂 

Aside from the comments this looks like a good start to me. It would be great to set up a `Trait` for the Rust side of our plugins so that we can avoid going through Python to call them -- which I assume is what you already started on with #13929 😄

crates/accelerate/src/high_level_synthesis.rs
alexanderivrii: I like `use_physical_indices` the most -- will rename. But note that we still use `use_qubit_indices` in the `__init__` to `HighLevelSynthesis` in Python.

crates/accelerate/src/high_level_synthesis.rs
alexanderivrii: This is good to know, thanks, but I will not change it since it's fine as is 😄.

crates/accelerate/src/high_level_synthesis.rs
alexanderivrii: Hmm, we really want to improve the performance in the fast-path case, with the code being:
```
let qubits = dag.get_qargs(inst.qubits);
if !definitely_skip_op(py, data, &inst.op, qubits) {
    fast_path = false;
    break;
}
```
Here `qubits` is of type `&[Qubit]`, and I think it would be nice to avoid casting it to `&[u32]`.

crates/accelerate/src/high_level_synthesis.rs
alexanderivrii: I agree, I would have also liked to use some readily available unitary synthesis code :).

crates/accelerate/src/high_level_synthesis.rs
Cryoris: I misunderstood this, thanks for the explanation!

crates/accelerate/src/high_level_synthesis.rs
alexanderivrii: Adding the "super-fast" check in [f818709](https://github.com/Qiskit/qiskit/pull/13813/commits/f8187096252541ff0cd9466828043a6c1a70c0db). The logic is as follows: we use `dag.count_ops()` to find the set of operation names present in the circuit. If we have the target and the synthesis is called pre-routing (so that we do not need to examine actual qubits) we check whether every name in this set is supported by the target. If we do not have the target and are using the basis set instead, we check whether every name in the set is contained in the basis set.

crates/accelerate/src/high_level_synthesis.rs
alexanderivrii: Here and other places: improved transpiler errors, see [373c4ad](https://github.com/Qiskit/qiskit/pull/13813/commits/373c4ad4ad95a8046b4cf0085405ec4446e9002d).

crates/accelerate/src/high_level_synthesis.rs
alexanderivrii: Nice!

crates/accelerate/src/high_level_synthesis.rs
alexanderivrii: Done in [51e0662](https://github.com/Qiskit/qiskit/pull/13813/commits/51e0662c127f9b9f34858ea244660e8c302365c9).

crates/accelerate/src/high_level_synthesis.rs
alexanderivrii: Done in [8a135a3](https://github.com/Qiskit/qiskit/pull/13813/commits/8a135a32a1fc7a39040922a96e357c1600d744c5)

crates/accelerate/src/high_level_synthesis.rs
alexanderivrii: As discussed with @Cryoris in private, in the future we want to synthesize unitary gates within `HighLevelSynthesis` using a plugin mechanism. For now, the current PR basically ports the definition method for unitary gates to Rust, without changing how they are handled.

alexanderivrii:
Related to:
> Given that there's a lot of Rust-Python back and forth, would it be simpler to write a Python function to call here?

After a discussion with @Cryoris. Long-term it's better to have the code handling control-flow ops in rust (as done right now). There will be opportunities to simplify this code once other python attributes of the quantum circuit are fully ported to rust.

Regarding:

> It would be great to set up a Trait for the Rust side of our plugins so that we can avoid going through Python to call them -- which I assume is what you already started on with https://github.com/Qiskit/qiskit/pull/13929 😄

Also after a discussion with @Cryoris. The "Trait" idea is really cool, and will allow to call the main synthesis methods for various high-level objects without going to the Python space. However, #13929 is orthogonal to this, it only aims to port more of the synthesis methods to rust.

crates/accelerate/src/high_level_synthesis.rs
```diff
@@ -907,6 +942,17 @@ fn py_run_on_dag(
 ) -> PyResult<Option<DAGCircuit>> {
     // Fast-path: check if HighLevelSynthesis can be skipped altogether. This is only
     // done at the top-level since this does not track the qubit states.
+
+    // First, we apply a super-fast (but incomplete) check to see if all the operations
+    // present in the circuit are suported by the target / are in the basis. This check
+    // is based only on the names of the operations in the circuit.
+    let op_names: HashSet<String> = dag.count_ops(py, false)?.keys().cloned().collect();
```
Cryoris: Would it be possible to directly iterate over the `Keys` object instead of cloning and collecting it? The Keys should already be unique 🤔

qiskit/transpiler/passes/synthesis/hls_plugins.py
```diff
@@ -1764,6 +1776,153 @@ def run(self, high_level_object, coupling_map=None, target=None, qubits=None, **
...
+
+        operation = high_level_object
+        modifiers = high_level_object.modifiers
+
+        # The plugin needs additional information that is not yet passed via the run's method
+        # arguments: namely high-level-synthesis data and options, the global qubits over which
+        # the operation is defined, and the initial state of each global qubit.
+        tracker = options.get("qubit_tracker", None)
+        data = options.get("hls_data", None)
+        input_qubits = options.get("input_qubits", None)
```
Cryoris, line 1798: I've probably asked this before, but couldn't we use `qubits` here instead?

qiskit/transpiler/passes/synthesis/hls_plugins.py
```diff
@@ -1764,6 +1776,153 @@ def run(self, high_level_object, coupling_map=None, target=None, qubits=None, **
...
+
+        if data is None or input_qubits is None:
+            raise TranspilerError(
+                "HighLevelSynthesis: problem with the default plugin for annotated operations."
```
Cryoris: It would be nice if this was more specific 🙂

qiskit/transpiler/passes/synthesis/hls_plugins.py
```diff
@@ -1764,6 +1776,153 @@ def run(self, high_level_object, coupling_map=None, target=None, qubits=None, **
...
+
+            # The base operation cannot use control qubits as auxiliary qubits.
+            # In addition, when we have power or inverse modifiers, we need to set all of
+            # the operation's qubits to dirty. Note that synthesizing the base operation we
+            # can use additional auxiliary qubits, however they would always be returned to
+            # their previous state, so clean qubits remain clean after each for- or while- loop.
+            annotated_tracker = tracker.copy()
+            annotated_tracker.disable(input_qubits[:num_ctrl])  # do not access control qubits
+            if total_power != 0 or is_inverted:
+                annotated_tracker.set_dirty(input_qubits)
```
Cryoris: Why do all qubits need to be set to dirty? And why don't the "target" qubits need to be dirty in general?

qiskit/transpiler/passes/synthesis/hls_plugins.py
```diff
@@ -1764,6 +1776,153 @@ def run(self, high_level_object, coupling_map=None, target=None, qubits=None, **
...
+        if data is None or input_qubits is None:
+            raise TranspilerError(
+                "HighLevelSynthesis: problem with the default plugin for annotated operations."
+            )
+
+        if len(modifiers) > 0:
+            num_ctrl = sum(
+                mod.num_ctrl_qubits for mod in modifiers if isinstance(mod, ControlModifier)
+            )
+            total_power = sum(mod.power for mod in modifiers if isinstance(mod, PowerModifier))
```
Cryoris: `total_power` is a bit misleading: the "total" power is always 1 per default, since `U^1 = U` and `U^0` would be the identity. Maybe... `repeats` or `additional_powers` or just `power`?

qiskit/transpiler/passes/synthesis/hls_plugins.py
```diff
@@ -1764,6 +1776,153 @@ def run(self, high_level_object, coupling_map=None, target=None, qubits=None, **
...
+                synthesized_base_op = QuantumCircuit._from_circuit_data(
+                    synthesized_base_op_result[0]
+                )
+            tracker.set_dirty(input_qubits[num_ctrl:])
+
+            # This step currently does not introduce ancilla qubits. However it makes
+            # a lot of sense to allow this in the future.
+            synthesized = self._apply_annotations(synthesized_base_op, operation.modifiers)
+
+            if not isinstance(synthesized, QuantumCircuit):
```
Cryoris: Could this be removed? We're sure that the object is a circuit at this point, no?

qiskit/transpiler/passes/synthesis/hls_plugins.py
alexanderivrii: Unfortunately, when the `HighLevelSynthesis` transpiler pass was first introduced (i.e. way before we started to support auxiliary qubits), we made the decision to pass `qubits=None` to plugins meaning that the pass runs pre-routing. But now that we handle recursion by explicitly reasoning about the qubits in the original circuit (aka the global qubits), we also need to pass the actual qubits as well. Ideally, I would like to have `qubits` always be the set of global qubits, and then we wouldn't need `input_qubits`, but this would be a breaking change. We do need to think of a path forward to implement this and other changes to the synthesis plugins API.

qiskit/transpiler/passes/synthesis/hls_plugins.py
```diff
@@ -1764,6 +1776,153 @@ def run(self, high_level_object, coupling_map=None, target=None, qubits=None, **
...
+                    controlled_qubits = list(range(0, modifier.num_ctrl_qubits)) + [
+                        modifier.num_ctrl_qubits + circuit.find_bit(q).index for q in inst_qubits
+                    ]
+                    controlled_circuit.append(controlled_op, controlled_qubits)
+
+                circuit = controlled_circuit
+
+                if isinstance(circuit, AnnotatedOperation):
+                    raise TranspilerError(
+                        "HighLevelSynthesis: failed to synthesize the control modifier."
```
Cryoris: Technically this is not inside `HLS` anymore, maybe it would be good to remove the `HighLevelSynthesis:` from the error messages 🙂

qiskit/transpiler/passes/synthesis/hls_plugins.py
```diff
@@ -1764,6 +1776,153 @@ def run(self, high_level_object, coupling_map=None, target=None, qubits=None, **
...
+                    )
+
+            elif isinstance(modifier, PowerModifier):
+                circuit = circuit.power(modifier.power)
+
+            else:
+                raise TranspilerError(f"HighLevelSynthesis: Unknown modifier {modifier}.")
+
+        if not isinstance(circuit, QuantumCircuit):
```
Cryoris: Could we also remove this one?

qiskit/transpiler/passes/synthesis/hls_plugins.py
```diff
@@ -1764,6 +1776,153 @@ def run(self, high_level_object, coupling_map=None, target=None, qubits=None, **
...
+            raise TranspilerError("HighLevelSynthesis: incorrect input to 'apply_annotations'.")
+
+        for modifier in modifiers:
+            if isinstance(modifier, InverseModifier):
+                circuit = circuit.inverse()
+
+            elif isinstance(modifier, ControlModifier):
+                if circuit.num_clbits > 0:
```
Cryoris, line 1877: Let's revisit this in the future, but wouldn't the logic be exactly the same even with control flow logic? Since the quantum part would only be applied if the control are satisfied, the output should be correct, no?

releasenotes/notes/improve-hls-pass-620389dde24e9707.yaml
```diff
@@ -0,0 +1,6 @@
+---
+features_transpiler:
```
Cryoris, line 2: From the tests it seemed that the output changed as well, we should add an `upgrade` releasenote here too. Though there is no way to still achieve the old behavior, or is there?

Cryoris (commented):
Some more comments! 🙂

crates/accelerate/src/high_level_synthesis.rs
alexanderivrii: Done in [d25aaa0](https://github.com/Qiskit/qiskit/pull/13813/commits/d25aaa022e465d5e39852e26eedf1fec53ad0daf). 

Answering the offline question (I hope I understand this correctly): when we call
```
let ops = dag.count_ops(py, false)?;
let mut op_keys = ops.keys();
```
and later call
```
op_keys.all(|name| borrowed_data.device_insts.contains(name))
```
the `op_keys` need to be mutable: while the iterator does not change the elements, it needs to be consumed mutably because the `all` method needs to adjust its internal state.

qiskit/transpiler/passes/synthesis/hls_plugins.py
alexanderivrii: Since `power` might be a floating-point number, I prefer `power` to `repeats`.

qiskit/transpiler/passes/synthesis/hls_plugins.py
alexanderivrii: Yes, removing -- this is a remnant of previous refactoring (making sure that `apply_annotations` no longer returns a `DAGCircuit` or a standalone `Operation`).

qiskit/transpiler/passes/synthesis/hls_plugins.py
alexanderivrii: Done (or rather replaced by `AnnotatedSynthesisDefault`).

releasenotes/notes/improve-hls-pass-620389dde24e9707.yaml
alexanderivrii: I don't understand this comment: in the past we synthesized AnnotatedOperations as a part of the HighLevelSynthesis transpiler pass itself, and now we are doing it as a part of the default plugin for AnnotatedOperations- but these details should not impact users in any way, right?

Of course, with general improvements to HighLevelSynthesis + plugins, the circuits produced by `transpile` may change, but I think that's expected.

Cryoris: > Of course, with general improvements to HighLevelSynthesis + plugins, the circuits produced by transpile may change, but I think that's expected.

Such a change in the default behavior should be documented though, I think, since users could rely on the previous behavior. If it's a more efficient decomposition using the same set of gates it is maybe less important, but by looking at the tests it seems that here the wrapping of the circuit has changed, or no? 🤔 

image

In this case I think we should leave a note, as users could've relied on the previous wrapping in some way they used "decompose" or counted the operations

qiskit/transpiler/passes/synthesis/hls_plugins.py
alexanderivrii: The `input_qubits` are the qubits over which our annotated operation is defined: it consists of the qubits of the base operation (what you called "target") and the control qubits (if any). We may also have other global qubits at our disposal, these are not part of `input_qubits`.

The selected lines set `input_qubits` to dirty, this includes the target qubits but not the "other" global qubits. In addition, on the line just above the selected lines we call `annotated_tracker.disable(control_qubit)` to make sure these cannot be used for synthesis of the base operation).

We don't need to set the "other" qubits to dirty because we are currently assuming that all synthesis methods preserve the state of the ancilla qubits. On the other hand, we probably need to set "target" qubits to dirty when we have "repeats" (or more generally some floating-point power) or "inverse". For repeats: even if the target is initially at $|0\rangle>$, after one iteration of the loop this may no longer be the case. For inverse: we flip the order of operations, so the clean qubits "from the front" and not the same as the clean qubits "from the back".

qiskit/transpiler/passes/synthesis/hls_plugins.py
alexanderivrii: Right, so currently we are not allowing to control-annotate circuits with classical bits, and this PR does not change this behavior. I think you might be correct, and we should indeed revisit this in the future.

crates/accelerate/src/high_level_synthesis.rs
Cryoris: Thanks for the explanation!

crates/accelerate/src/high_level_synthesis.rs
eliarbel: OK, we can address this in a later PR, not worth blocking this one over since you already have a working path.

eliarbel (dismissed):
Changes related to my comments LGTM.

releasenotes/notes/improve-hls-pass-620389dde24e9707.yaml
alexanderivrii: Ah, right, done in [e6dffe8](https://github.com/Qiskit/qiskit/pull/13813/commits/e6dffe8e5736d70c15df153f7f68edce42ea2c45).

qiskit/transpiler/passes/synthesis/hls_plugins.py
alexanderivrii: As we discussed offline, I have slightly improved the code and added a detailed in-code comment explaining the above.

Cryoris (commented):
Modulo the failing test this LGTM, thanks for all the effort @alexanderivrii! There are some follow-ups we should not lose track of here, it might be good to have issues for them. The main ones I think are:
* Improve collaboration of `UnitarySynthesis` and `HighLevelSynthesis` -- this needs some discussion to see if we want to combine them or, if not, how we can have HLS respect the unitary synthesis plugin selection
* Use Rust-versions of HLS plugins from Rust directly (cf. `Trait` conversation above)
* Improve the control flow handling 

I can approve once the tests are fixed 👍🏻

Cryoris (approved):

PR #13814, opened 2025-02-09
Remove pulse support in QPY in 2.0

eliarbel:
### Summary
This PR removes support for saving `ScheduleBlock` programs in `qpy.dump`. It also includes support for loading payloads with `ScheduleBlock`s or pulse gates while ignoring the actual pulse information, returning partially specified circuits. 

Part of #13662 

### Details and comments
Some details to note and give feedback on:
1. Loading a `ScheduleBlock` program via `qpy.load` results with a `QuantumCircuit` object containing only the name and possibly metadata of that `ScheduleBlock`. Loading a `QuantumCircuit` payload with pulse gates leads to getting circuit object with undefined custom instructions (representing pulse gates without calibrations). In both cases warnings of type `QPYLoadingDeprecatedFeatureWarning` are being issued. Is this the type of warning we want to use here? Maybe we should add something like `QPYLoadingRemovedFeatureWarning` and update the Compatibility section in the QPY doc page accordingly?
2. This PR does not change the QPY format version, i.e. it stays on 13. This means that we still write CALIBRATION header (with 0 cals) so that `load` continues to work (since it still includes code for reading calibrations, for graceful backward compatibility). 
3. When loading payloads with calibrations, the code will issue a message for each calibration, warning the user that if there exists a gate with this name, it will be undefined. Theoretically, one can add calibrations to a circuit without any pulse gates. We could check this and have the warnings accurately relate to existing pulse gates, but it think it'll just impose extra runtime overhead. I'm open to other opinions though. 

Open Tasks: 
- [x] Complete handling of `test/qpy_compat`


qiskit/qpy/binary_io/circuits.py
```diff
@@ -649,21 +648,21 @@ def _read_calibrations(file_obj, version, vectors, metadata_deserializer):
...
-            for _ in range(defheader.num_qubits)
-        )
-        params = tuple(
-            value.read_value(file_obj, version, vectors) for _ in range(defheader.num_params)
-        )
-        schedule = schedules.read_schedule_block(file_obj, version, metadata_deserializer)
+        if name:
+            warnings.warn(
+                category=exceptions.QPYLoadingDeprecatedFeatureWarning,
+                message="Support for loading dulse gates has been removed in Qiskit 2.0. "
```
raynelfss:
```suggestion
                message="Support for loading pulse gates has been removed in Qiskit 2.0. "
```

qiskit/qpy/type_keys.py
```diff
@@ -353,92 +211,27 @@ class ScheduleOperand(TypeKeyBase):
...
-
-        raise exceptions.QpyError(
-            f"A class corresponding to type key '{type_key}' is not found in {cls.__name__} namespace."
-        )
+    def retrieve(cls, _):
+        raise NotImplementedError
 
 
 class Program(TypeKeyBase):
-    """Typle key enum for program that QPY supports."""
```
raynelfss, line 431: I find it funny that this typo has been there for so long

raynelfss (commented):
Just a typo I found :)

eliarbel:
I think it's better to just remove the `generate_*` functions which generate pulse circuits from `test/qpy_compat/test_qpy.py`. Will commit that soon.

qiskit/qpy/__init__.py
```diff
@@ -18,12 +18,11 @@
 .. currentmodule:: qiskit.qpy
 
 QPY is a binary serialization format for :class:`~.QuantumCircuit` and
-:class:`~.ScheduleBlock` objects that is designed to be cross-platform,
-Python version agnostic, and backwards compatible moving forward. QPY should
-be used if you need a mechanism to save or copy between systems a
-:class:`~.QuantumCircuit` or :class:`~.ScheduleBlock` that preserves the full
-Qiskit object structure (except for custom attributes defined outside of
-Qiskit code). This differs from other serialization formats like
+objects that is designed to be cross-platform,  Python version agnostic,
```
mtreinish, line 21: There is an extra `and` here on the previous line (so I can't leave an inline suggestion).

qiskit/qpy/__init__.py
```diff
@@ -170,6 +169,11 @@ def open(*args):
     it to QPY setting ``use_symengine=False``.  The resulting file can then be loaded by any later
     version of Qiskit.
 
+    With the removal of Pulse in Qiskit 2.0, QPY provides limited support for loading
+    payloads with pulse data. Loading a ``ScheduleBlock`` payload, a :class:`.QpyError` exception
+    will be raised. Loading a circuit with pulse gates, the circuit will contain custom
+    instructions without calibration data attached, leaving them undefined.
```
mtreinish: I would make this a second note because it's logically different from the previous note about symengine.

```suggestion
.. note::

    Starting with Qiskit version 2.0.0, which removed the Pulse module from the library, QPY provides limited support for loading
    payloads that includes pulse data. Loading a ``ScheduleBlock`` payload, a :class:`.QpyError` exception
    will be raised. Loading a payload for a circuit that contained pulse gates, the output circuit will contain
    custom instructions **without** calibration data attached for each pulse gate, leaving them undefined.
```

qiskit/qpy/binary_io/circuits.py
```diff
@@ -649,21 +648,21 @@ def _read_calibrations(file_obj, version, vectors, metadata_deserializer):
...
-            struct.unpack("!q", file_obj.read(struct.calcsize("!q")))[0]
-            for _ in range(defheader.num_qubits)
-        )
-        params = tuple(
-            value.read_value(file_obj, version, vectors) for _ in range(defheader.num_params)
-        )
-        schedule = schedules.read_schedule_block(file_obj, version, metadata_deserializer)
+        if name:
+            warnings.warn(
+                category=exceptions.QPYLoadingDeprecatedFeatureWarning,
```
mtreinish: I'd make this a `UserWarning`, that's what I did in #13506 for the equivalent around `c_if()` usage. I read the description of `QPYLoadingDeprecatedFeatureWarning` as the warning prior to the API change, not after.

qiskit/qpy/binary_io/schedules.py
```diff
@@ -10,53 +10,43 @@
 # copyright notice, and modified files need to carry a notice indicating
 # that they have been altered from the originals.
 
-"""Read and write schedule and schedule instructions."""
+"""Read schedule and schedule instructions.
+
+This module is kep post pulse-removal to allow reading legacy
```
mtreinish:
```suggestion
This module is kept post pulse-removal to allow reading legacy
```

qiskit/qpy/binary_io/schedules.py
```diff
@@ -547,89 +293,19 @@ def read_schedule_block(file_obj, version, metadata_deserializer=None, use_symen
...
-        key_str = instructions.Reference.key_delimiter.join(ref_keys)
-        flat_key_refdict[key_str] = schedule
-    common.write_mapping(
-        file_obj=file_obj,
-        mapping=flat_key_refdict,
-        serializer=_dumps_reference_item,
-        metadata_serializer=metadata_serializer,
-        version=version,
-    )
+    return QuantumCircuit(name=name, metadata=metadata)
```
mtreinish: Is this return type still right. This PR raises an error now if a schedule block is received so we shouldn't need this anymore.

releasenotes/notes/remove-pulse-qpy-07a96673c8f10e38.yaml
```diff
@@ -0,0 +1,8 @@
+---
+upgrade_qpy:
+  - |
+    With the removal of Pulse in Qiskit 2.0, support for serializing ``ScheduleBlock`` programs
+    via the :func:`qiskit.qpy.dump` function has been removed. Users can still load payloads
+    containing pulse gates using the :func:`qiskit.qpy.load` function, however those will be
+    treated as undefined custom instructions. Loading ``ScheduleBlock`` payloads is not supported
```
mtreinish: The term we normally use for this is opaque instructions. For an instruction that contains no definition or other details.

```suggestion
    treated as opaque custom instructions. Loading ``ScheduleBlock`` payloads is not supported
```

test/qpy_compat/test_qpy.py
```diff
@@ -973,6 +1003,30 @@ def load_qpy(qpy_files, version_parts):
...
 
+    while pulse_files:
+        path, version = pulse_files.popitem()
+
+        if version_parts < version or version_parts >= (2, 0):
+            continue
+
+        if path == "pulse_gates.qpy":
+            try:
+                load(open(path, "rb"))
```
mtreinish: Do we need to use a context manager here to avoid forcing python to close the file when the file object is dropped? I think at one point python was emitting warnings about that.

```suggestion
                with open(path, "rb") as fd:
                    load(fd)
```

test/qpy_compat/test_qpy.py
```diff
@@ -973,6 +1003,30 @@ def load_qpy(qpy_files, version_parts):
...
+            except:
+                msg = f"Loading circuit with pulse gates should not raise"
+                sys.stderr.write(msg)
+                sys.exit(1)
+        else:
+            try:
+                # A ScheduleBlock payload, should raise QpyError
+                load(open(path, "rb"))
```
mtreinish: Same question here?
```suggestion
                with open(path, "rb") as fd:
                    load(fd)
```

mtreinish (commented):
Overall this looks good. I left a few comments and suggestions inline, but nothing major. 

I was confused at first by the changes in the `binary_io/schedules.py` flle because I skipped over the module docs. But it's clear now that it's basically a bunch of `seek()` methods now. It's obvious now they're needed because the size to read for the pulse parts isn't known ahead of time, so this is needed to skip the pulse portion of a pulse gates circuit payload.

qiskit/qpy/binary_io/circuits.py
eliarbel: I wasn't sure about the right type to use here, thanks for clarifying.

test/qpy_compat/test_qpy.py
eliarbel: Agreed, although it's only a handful of files opened for reading but it's better practice anyway with a context manager.

qiskit/qpy/binary_io/schedules.py
eliarbel: Good catch, indeed a left over from the previous version of the PR.

qiskit/qpy/__init__.py
eliarbel: Got you, OK.

eliarbel:
Thanks for the review @mtreinish, all comments have been addressed.

test/qpy_compat/test_qpy.py
```diff
@@ -26,6 +27,7 @@
 from qiskit.circuit.quantumregister import Qubit
 from qiskit.circuit.parameter import Parameter
 from qiskit.circuit.parametervector import ParameterVector
+from qiskit.qpy.exceptions import QpyError
```
mtreinish: This import path doesn't exist in all the historical qiskit version that execute this file. For the beginning of qpy's life it lived in `qiskit.circuit`. We made it a standalone module in preparation for adding pulse in 0.20.0 (see: https://github.com/Qiskit/qiskit/pull/7582).

This is causing the test failures you're hitting here. You can probably just make this a runtime import right before you start catching the exception to fix the failure.

eliarbel: Ughh... this wasn't caught in my dev environment. I guess Python's import system managed somehow. Giving it another try now after it passed locally in a new env for qpy. Should we have `tox -e qpy_compat`?

test/qpy_compat/test_qpy.py
```diff
@@ -983,6 +1012,34 @@ def load_qpy(qpy_files, version_parts):
...
+            try:
+                # A ScheduleBlock payload, should raise QpyError
+                with open(path, "rb") as fd:
+                    load(fd)
+            except QpyError:
+                continue
+
+            msg = f"Loading payload {path} didn't raise QpyError"
+            sys.stderr.write(msg)
+            sys.exit(1)
```
mtreinish, line 1041: I probably would have used a for loop for this, but it's fine like this.

qiskit/qpy/__init__.py
```diff
@@ -17,13 +17,12 @@
...
 
-QPY is a binary serialization format for :class:`~.QuantumCircuit` and
-:class:`~.ScheduleBlock` objects that is designed to be cross-platform,
-Python version agnostic, and backwards compatible moving forward. QPY should
-be used if you need a mechanism to save or copy between systems a
-:class:`~.QuantumCircuit` or :class:`~.ScheduleBlock` that preserves the full
-Qiskit object structure (except for custom attributes defined outside of
-Qiskit code). This differs from other serialization formats like
+QPY is a binary serialization format for :class:`~.QuantumCircuit`
+objects that is designed to be cross-platform,  Python version agnostic,
```
mtreinish, line 21: Super tiny nit, not worth blocking over:
```suggestion
objects that is designed to be cross-platform, Python version agnostic,
```

test/qpy_compat/test_qpy.py
```diff
@@ -983,6 +1012,34 @@ def load_qpy(qpy_files, version_parts):
                 circuit, qpy_circuits[i], i, version_parts, bind=bind, equivalent=equivalent
             )
 
+    from qiskit.qpy.exceptions import QpyError
+
+    while pulse_files:
+        path, version = pulse_files.popitem()
+
+        if version_parts < version or version_parts >= (2, 0):
+            continue
```
mtreinish, line 1021: Just for any onlookers in the future, `version_parts` here is the version of qiskit used to generate the payloads being loaded.

mtreinish (approved):
This LGTM, I had two small inline comments but nothing worth blocking over. Mostly my idle musings and one extra space in a docstring. Thanks for doing this and the quick updates.

qiskit/qpy/__init__.py
eliarbel: Will apply this in #13662

test/qpy_compat/test_qpy.py
eliarbel: Will add a comment in #13662

PR #13815, opened 2025-02-10
Remove the `classical_function` module and all `tweedledum`-related code

### LINKED ISSUE #13755: Remove tweedledum from qiskit-sdk

gadial:
### Summary
Removes from qiskit the `classical_function` module and all code that relies on the `tweedledum` library.
Closes #13755 

### Details and comments
The `tweedledum` library is not compatible with new versions of Python, so reliance on it has to be dropped in Qiskit 2.0. Since the `classical_function` module basically servers as a qiskit wrapper for `tweedledum` functionality, we remove it from qiskit, after its deprecation in #13786.

To preserve the main functionality of `classical_function`, i.e. the ability to generate phase and bit-flip oracles from boolean expressions, we add a simple, `tweedledum`-independent implementation in #13769, which is placed in `qiskit.synthesis.boolean` instead of the `classical_function` module. This PR relies on #13769 being merged, otherwise the existing tests for `PhaseOracle` will not pass.


test/python/circuit/test_extensions_standard.py
```diff
@@ -1371,105 +1370,5 @@ def test_cswap_reg_reg_inv(self):
         self.assertEqual(instruction_set[1].qubits, (self.qr[1], self.qr2[1], self.qr3[1]))
         self.assertEqual(instruction_set[2].operation.params, [])
 
-
-class TestStandardMethods(QiskitTestCase):
```
ShellyGarion, line 1353: can we do these tests with PhaseOracle w/o tweedledum ?

test/python/visualization/test_circuit_text_drawer.py
```diff
@@ -1407,32 +1401,6 @@ def test_text_spacing_2378(self):
...
-                "           └───┘",
-            ]
-        )
-
-        @classical_function
-        def grover_oracle(a: Int1, b: Int1, c: Int1) -> Int1:
-            return a and b and not c
-
-        circuit = grover_oracle.synth(registerless=False)
-        self.assertEqual(str(circuit_drawer(circuit, output="text", initial_state=True)), expected)
```
ShellyGarion, line 1366: can we still do this tests with grover or phase oracle w/o tweedledum ?

ShellyGarion (commented):
looks good. I only wonder if we should keep some of the removed tests?

1ucian0:
`on hold` until https://github.com/Qiskit/qiskit/pull/13769 merges

test/python/circuit/test_extensions_standard.py
gadial: Yes, actually tweedledum was not really needed there. I returned the test.

test/python/visualization/test_circuit_text_drawer.py
gadial: It doesn't make sense, since this is a test for registerless synthesis, which is currently not supported by the oracles.

releasenotes/notes/remove_classical_function-7d9057bc5c77d268.yaml
```diff
@@ -0,0 +1,7 @@
+---
+upgrade_circuits:
+  - |
+    The :mod:`.classical_function` was removed. This module was dependent on the
+    `tweedledum` library which is no longer compatible with newer versions of Python.
+    As an alternative, the :class:`.PhaseOracle` and :class:`.BitFlipOracle` classes
```
Cryoris:
```suggestion
    As an alternative, the :class:`.PhaseOracleGate` and :class:`.BitFlipOracleGate` classes
```

Cryoris (commented):
LGTM, just one note on the reno 🙂 (also black is failing)

ShellyGarion (approved):
LGTM. I'll let @Cryoris or @1ucian0 merge it. Thanks @gadial !

1ucian0 (approved):

PR #13816, opened 2025-02-10
Consider `unit` in `Delay` comparisons

### LINKED ISSUE #13812: `Delay.__eq__` does not compare the delay unit
### Environment

- **Qiskit version**: 1.3.2 or 9ae0a808b45249d6f185baf54f078cb1ed00648a
- **Python version**: 3.11
- **Operating system**: Fedora Linux 41

### What is happening?

When comparing two `Delay` instructions, only the `duration` argument to `__init__` matters, not the `unit` argument.

```python
from qiskit.circuit.delay import Delay

qc1 = QuantumCircuit(1)
qc1.delay(1, 0, unit="s")

qc2 = QuantumCircuit(1)
qc2.delay(1, 0, unit="dt")

print("Compare circuits:", qc1 == qc2)
print("Compare delays:", Delay(1, "dt") == Delay(1, "s"))
```

### What should happen?

The code above should print False/False but it prints False/True.

### Any suggestions?

Before https://github.com/Qiskit/qiskit/pull/13486, both the circuit and instruction comparisons above were true. Now the circuit one is false. This came up in Qiskit Experiments because of [this other bug](https://github.com/Qiskit/qiskit/issues/10662) with delay units and qpy. A test that ran a circuit with a delay roundtrip through qpy that used to pass even though the delay unit was lost started to fail.

jakelishman:
The `unit` field was previously ignored, allowing delays in `dt` units to compare equal to those in `s`. This commit does not add additional on-the-fly unit conversion to the comparison: if the user specified durations in different time multiples, they may have had a reason to consider them non-equal.  This could be revisiting after the new system for duration handling lands (i.e. the new functionality for `stretch` and other delayed scheduling).

### Details and comments

Fix #13812.

### DIFF
--- a/qiskit/circuit/delay.py
+++ b/qiskit/circuit/delay.py
@@ -84,6 +84,11 @@ class Delay(Instruction):
         """
         return self.__array__(dtype=complex)
 
+    def __eq__(self, other):
+        return (
+            isinstance(other, Delay) and self.unit == other.unit and self._compare_parameters(other)
+        )
+
     def __repr__(self):
         """Return the official string representing the delay."""
         return f"{self.__class__.__name__}(duration={self.params[0]}[unit={self.unit}])"
--- a/releasenotes/notes/delay-compare-b7ecb26b94ff0cd3.yaml
+++ b/releasenotes/notes/delay-compare-b7ecb26b94ff0cd3.yaml
@@ -1,0 +1,5 @@
+---
+fixes:
+  - |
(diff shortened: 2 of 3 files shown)


kevinhartman (approved):
LGTM.

kevinhartman:
It's probably also worth noting here that the `PackedInstruction` version of `StandardInstruction::Delay` should already compare in the same way as what you've done here.

jakelishman:
Kevin: it does - the discrepancy between it and the Python-space version is what prompted #13812.

PR #13818, opened 2025-02-10
Implement `CircuitData::copy_empty_like()`

raynelfss:
### Summary
The following commits add some quality-of-life changes to `CircuitData` and `QuantumCircuit` discovered while working on #13686:
- Implement a separate `copy_empty_like()` method for `CircuitData`.
- Separate use of `copy()` and `copy_empty_like()` for `QuantumCircuit` to avoid creating repeated instances of `CircuitData`.

### Comments
Feel free to ask any questions down below or on a review.

### DIFF
--- a/crates/circuit/src/circuit_data.rs
+++ b/crates/circuit/src/circuit_data.rs
@@ -284,8 +284,12 @@ impl CircuitData {
     ///     CircuitData: The shallow copy.
     #[pyo3(signature = (copy_instructions=true, deepcopy=false))]
     pub fn copy(&self, py: Python<'_>, copy_instructions: bool, deepcopy: bool) -> PyResult<Self> {
-        let mut res = self.copy_empty_like(py, Some(self.data().len()))?;
+        let mut res = self.copy_empty_like(py)?;
+        res.qargs_interner = self.qargs_interner.clone();
+        res.cargs_interner = self.cargs_interner.clone();
+        res.reserve(py, self.data().len());
         res.param_table.clone_from(&self.param_table);
+
         if deepcopy {
             let memo = PyDict::new(py);
             for inst in &self.data {
@@ -317,22 +321,19 @@ impl CircuitData {
         Ok(res)
     }
 
-    /// Performs an empty-like shallow copy.
+    /// Performs a copy with no instruction.
     ///
     /// Returns:
     ///     CircuitData: The shallow copy.
-    #[pyo3(signature = (reserve = None,))]
-    pub fn copy_empty_like(&self, py: Python<'_>, reserve: Option<usize>) -> PyResult<Self> {
-        let mut res = CircuitData::new(
+    pub fn copy_empty_like(&self, py: Python<'_>) -> PyResult<Self> {
+        let res = CircuitData::new(
             py,
             Some(self.qubits.cached().bind(py)),
             Some(self.clbits.cached().bind(py)),
             None,
-            reserve.unwrap_or_default(),
+            0,
             self.global_phase.clone(),
         )?;
-        res.qargs_interner = self.qargs_interner.clone();
-        res.cargs_interner = self.cargs_interner.clone();
 
         Ok(res)
     }
--- a/crates/circuit/src/circuit_data.rs
+++ b/crates/circuit/src/circuit_data.rs
@@ -284,18 +284,8 @@ impl CircuitData {
     ///     CircuitData: The shallow copy.
     #[pyo3(signature = (copy_instructions=true, deepcopy=false))]
     pub fn copy(&self, py: Python<'_>, copy_instructions: bool, deepcopy: bool) -> PyResult<Self> {
-        let mut res = CircuitData::new(
-            py,
-            Some(self.qubits.cached().bind(py)),
-            Some(self.clbits.cached().bind(py)),
-            None,
-            self.data.len(),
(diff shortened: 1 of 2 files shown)


qiskit/circuit/quantumcircuit.py
```diff
@@ -3698,7 +3698,14 @@ def copy(self, name: str | None = None) -> typing.Self:
...
+        if not (name is None or isinstance(name, str)):
+            raise TypeError(
+                f"invalid name for a circuit: '{name}'. The name must be a string or 'None'."
+            )
+        cpy = _copy.copy(self)
+
+        _copy_metadata(self, cpy, "alike")
+
         cpy._data = self._data.copy()
         return cpy
```
jakelishman: How come this needed to change? It makes me a bit worried that it will be easy to get out-of-sync with `copy_empty_like` if it doesn't call it (a problem I'm relatively sure we had in the past) - the additional cost of making a new empty `CircuitData` should (in theory) be dwarfed by the Python-space `copy.copy` (though the current behaviour of always cloning the interners would change that).

crates/circuit/src/circuit_data.rs
```diff
@@ -327,6 +317,25 @@ impl CircuitData {
...
+            Some(self.clbits.cached().bind(py)),
+            None,
+            reserve.unwrap_or_default(),
+            self.global_phase.clone(),
+        )?;
+        res.qargs_interner = self.qargs_interner.clone();
+        res.cargs_interner = self.cargs_interner.clone();
+
+        Ok(res)
+    }
```
jakelishman: A couple of notes:

- Since it's empty of data, there's not really a distinction between a "shallow" and "deep" copy
- As the return value is an owned value of `Self`, the `reserve` argument might be better as a separate function on `CircuitData` (if it isn't already).  It's not really any more expensive to do the `reserve` call after the creation of the internal vec.

crates/circuit/src/circuit_data.rs
```diff
@@ -327,6 +317,25 @@ impl CircuitData {
...
+        let mut res = CircuitData::new(
+            py,
+            Some(self.qubits.cached().bind(py)),
+            Some(self.clbits.cached().bind(py)),
+            None,
+            reserve.unwrap_or_default(),
+            self.global_phase.clone(),
+        )?;
+        res.qargs_interner = self.qargs_interner.clone();
+        res.cargs_interner = self.cargs_interner.clone();
```
jakelishman: It's not clear to me that this is a good idea for a general `copy_empty_like` function.  It makes sense for `copy` to do it, but in general, this function isn't going to know what instructions are about to be pushed to it, so cloning the whole interner may well be quite wasteful.  Perhaps better to make that a separate method - I think the functionality is already public via the interner-merge stuff.

jakelishman (commented):
This seems like a sensible addition overall, though I have a couple of questions about the particular implementation here.

qiskit/circuit/quantumcircuit.py
raynelfss: Right now the change isn't as needed. If we go through with #13686, we might need to separate them because of how adding registers to `CircuitData` might work. I'm okay with leaving it as it was for this one and then modifying it if we go with the aforementioned approach.

crates/circuit/src/circuit_data.rs
raynelfss: Not needed, I will make those corrections soon.

raynelfss: Addressed in dc555c8

crates/circuit/src/circuit_data.rs
raynelfss: Addressed in dc555c8

qiskit/circuit/quantumcircuit.py
raynelfss: Addressed in dc555c8

alexanderivrii (commented):
I have two general questions (mainly for self-education):

Under which circumstances, would I need to set `deepcopy` to `true`?

Why do our copy-empty-like functions copy ``global_phase``? In my mind there is some discrepancy: if I were represent the global phase of a circuit using ``global_phase``, then it would get copied, but if were to represent it using ``PhaseGate``, then it would not.  

I am especially worried about the situation where a transpiler pass takes the original circuit data with ``global_phase``, replaces this ``global_phase`` by ``PhaseGate``, creates a new circuit data by calling ``copy_empty_like`` on the original data, and adds a sequence of transpiled gates -- this would result in the ``global_phase`` now being counted twice.

jakelishman:
Sasha: I think it was just historical, with how it came from an old private `DAGCircuit` method that was intended for doing circuit rebuilds.  I can see the argument for it being discrepant with `GlobalPhaseGate`, but the method precedes the gate.  If we were doing it again, I imagine we might not copy over the phase, but it is what it is now.  In your example, the transpiler pass is simply mistaken - it's using two separate interpretations of the circuit at the same time.  If we didn't copy global phase, perhaps an even more common error would be to forget to propagate it over.

`deepcopy=true` is typically the Python-space behaviour; it deep-copies all Python-space gate objects so there's no shared state between two circuits.  `deepcopy=false` is more Rust native, or when we're just making a shallow copy to hold onto, or something like that.

jakelishman (approved):
This looks a lot cleaner now, thanks!

PR #13820, opened 2025-02-10
Delegate `BasePass.__call__` to `PassManager.run`

### LINKED ISSUE #13444: Running ALAPScheduleAnalysis directly on a circuit fails
### Environment

- **Qiskit version**: 1.2.4
- **Python version**: 3.12

### What is happening?

Running the `ALAPScheduleAnalysis` pass directly on a circuit fails. Running it via a passmanager works.

A minimal example
```
import numpy as np
from qiskit.transpiler import PassManager, InstructionDurations
from qiskit.transpiler.passes import ALAPScheduleAnalysis

durations = InstructionDurations( [("reset", None, 10),
     ("cz", None, 120), ("cx", None, 120), ("crx", None, 160),
     ("rx", None, 80), ('ry', None, 80), ('rz', None, 0),  ("measure", None, 30_000)]
)

circ = QuantumCircuit(2)
circ.rx(np.pi/2, 0)
circ.rz(-np.pi/2, 0)

pm = PassManager([ALAPScheduleAnalysis(durations)])
circ = pm.run(circ) # works

circ = QuantumCircuit(2)
circ.rx(np.pi/2, 0)
circ.rz(-np.pi/2, 0)

p=ALAPScheduleAnalysis(durations)
p(circ) # fails
```

### What should happen?

Usually a qiskit pass can be applied directly to a quantum circuit, so it seems reasonable to expect the same from schedule passes.

### Any suggestions?

_No response_

jakelishman:
This ensures that calling a pass directly will follow the same execution logic as a regular pass-manager construction.  This particularly matters for passes that have `requires`.  It also ensures that the conversion from a circuit and back to a DAG will follow the same rules, which is expected since it is a shorthand notation, but liable to get out of sync as they are complex.

### Details and comments

Fix #13444.

Requires #13821 to effect the warning on the future behaviour change.


test/python/transpiler/test_elide_permutations.py
```diff
@@ -235,8 +235,7 @@ def test_partial_permutation_in_middle(self):
 
         # Make sure that the transpiled circuit *with* the final permutation
         # is equivalent to the original circuit
-        perm = pass_.property_set["virtual_permutation_layout"].to_permutation(qc.qubits)
-        res.append(PermutationGate(perm), [0, 1, 2, 3, 4])
+        res.append(PermutationGate(res.layout.routing_permutation()), [0, 1, 2, 3, 4])
```
jakelishman, line 228: `PassManager.run` explicitly clears out `virtual_permutation_layout` from a `property_set` after execution, since it updates the layout fields.  This new version of the `PermutationGate` is more properly the behaviour we _want_ to test, though.

jakelishman:
Oh whoops, I forgot I hadn't fixed that test error locally.

qiskit/transpiler/passmanager.py
```diff
@@ -462,6 +472,9 @@ def _replace_error(meth):
     def wrapper(*meth_args, **meth_kwargs):
         try:
             return meth(*meth_args, **meth_kwargs)
+        except TranspilerError:
+            # If it's already a `TranspilerError` subclass, don't erase the extra information.
```
mtreinish, line 476: I remember discussing this offline, but `TranspilerError` being a subclass of `PassManagerError` seems so weird. But it's hard to change this now so this makes sense.

mtreinish (approved):
This LGTM, it seems pretty straightforward and it'll be good to unify the logic between `__call__` and `PassManager.run()`

qiskit/transpiler/passmanager.py
jakelishman: I assume it was intended as a backwards-compatibility shim for a move to `PassManagerError` during the genericisation of `PassManager`, but I don't really know.

PR #13821, opened 2025-02-10
Warn on use of `property_set` kwarg to `BasePassManger.run`

jakelishman:
The transpiler's `BasePass.__call__` implementation accepts a `property_set` kwarg that can be used to seed the property set for the run of that pass.  Starting in Qiskit 2.0, `PassManager.run` will also have that behaviour, so this commit begins emitting a warning on the potential change.

The impact of this is expected to be minimal, as we are not aware of much use of the subclassing of the generic `BasePassManager`, and even less of passing arbitrary keyword arguments to the conversion functions.

### Details and comments

This is the warning to enable to API change of #13820.

### DIFF
--- a/qiskit/passmanager/passmanager.py
+++ b/qiskit/passmanager/passmanager.py
@@ -30,7 +30,7 @@ from .compilation_status import PropertySet, WorkflowStatus, PassManagerState
 
 logger = logging.getLogger(__name__)
 
-_NOT_PRESENT = object()
+_MISSING = object()
 
 
 class BasePassManager(ABC):
@@ -178,7 +178,7 @@ class BasePassManager(ABC):
         callback: Callable = None,
         num_processes: int = None,
         *,
-        property_set: object = _NOT_PRESENT,
+        property_set: object = _MISSING,
         **kwargs,
     ) -> Any:
         """Run all the passes on the specified ``in_programs``.
@@ -224,7 +224,7 @@ class BasePassManager(ABC):
         Returns:
             The transformed program(s).
         """
-        if property_set is not _NOT_PRESENT:
+        if property_set is not _MISSING:
             warnings.warn(
                 "From Qiskit 2.0, 'property_set' will be a reserved keyword argument of"
                 " 'BasePassManager.run', and not passed on to the conversion functions."
--- a/qiskit/passmanager/passmanager.py
+++ b/qiskit/passmanager/passmanager.py
@@ -14,6 +14,7 @@
 from __future__ import annotations
 
 import logging
+import warnings
 from abc import ABC, abstractmethod
 from collections.abc import Callable, Iterable
 from itertools import chain
@@ -28,6 +29,8 @@ from .flow_controllers import FlowControllerLinear
 from .compilation_status import PropertySet, WorkflowStatus, PassManagerState
 
 logger = logging.getLogger(__name__)
+
+_NOT_PRESENT = object()
 
 
 class BasePassManager(ABC):
@@ -174,6 +177,8 @@ class BasePassManager(ABC):
         in_programs: Any | list[Any],
         callback: Callable = None,
         num_processes: int = None,
+        *,
+        property_set: object = _NOT_PRESENT,
         **kwargs,
     ) -> Any:
         """Run all the passes on the specified ``in_programs``.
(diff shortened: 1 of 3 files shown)


releasenotes/notes/passmanager-run-property_set-cb19ce448ab1e93a.yaml
```diff
@@ -0,0 +1,13 @@
+---
+upgrade_transpiler:
+  - |
+    Passing ``property_set`` as an arbitrary keyword argument to the :meth:`~.BasePassManager.run`
+    method of a subclass of :class:`.BasePassManager` will change behavior in Qiskit 2.0.  It is
+    currently forwarded to the internal representation converting functions of the pass maanger, as
```
ElePT:
```suggestion
    currently forwarded to the internal representation converting functions of the pass manager, as
```

qiskit/passmanager/passmanager.py
```diff
@@ -211,12 +216,25 @@ def callback_func(**kwargs):
...
+        if property_set is not _NOT_PRESENT:
+            warnings.warn(
+                "From Qiskit 2.0, 'property_set' will be a reserved keyword argument of"
+                " 'BasePassManager.run', and not passed on to the conversion functions."
+                " This subclass of 'BasePassManager' needs to use a different keyword argument"
+                " for passing on this data.",
+                FutureWarning,
+                stacklevel=2,
+            )
+            kwargs["property_set"] = property_set
```
ElePT, line 236: I wonder if this indentation was an oversight? Shouldn't we populate the kwargs with `property_set` when it's present too? At least this is what I understood from the docstring. I might be wrong.

```suggestion
        kwargs["property_set"] = property_set
```

ElePT (commented):
LGTM, I just had a small comment to make sure I am understanding the changes.

qiskit/passmanager/passmanager.py
jakelishman: If the `property_set` parameter has the sentinel value `_NOT_PRESENT`, it means that the function was called without setting that kwarg - there's no valid way for a user to have passed the special sentinel object other than by not typing the argument.  So if the conditional fails (i.e. `property_set is _NOT_PRESENT`) then it shouldn't be passed on, because the user didn't give us one.

ElePT: hah, so I was wrong. Glad I asked :)

ElePT: It took me 3 re-reads but I got why, I was not reading the `not` (it was "not not present", not "not present"), which was what confused me in the first place. Anyways, reading comprehension, I get it now.

ElePT (dismissed):
Alright, I think that the PR looks good.

qiskit/passmanager/passmanager.py
jakelishman: oh yeah, the naming isn't super ideal.  Do you think `_MISSING` or `_ABSENT` would make it clearer?  There's still _kind of_ a double negative in there (asserting lack of absence), but might be easier to read?

qiskit/passmanager/passmanager.py
```diff
@@ -29,6 +30,8 @@
 
 logger = logging.getLogger(__name__)
 
+_NOT_PRESENT = object()
```
ElePT:
```suggestion
_MISSING = object()
```

qiskit/passmanager/passmanager.py
```diff
@@ -174,6 +177,8 @@ def run(
         in_programs: Any | list[Any],
         callback: Callable = None,
         num_processes: int = None,
+        *,
+        property_set: object = _NOT_PRESENT,
```
ElePT:
```suggestion
        property_set: object = _MISSING,
```

qiskit/passmanager/passmanager.py
```diff
@@ -211,12 +216,25 @@ def callback_func(**kwargs):
...
-
+            property_set: Currently a key that will be interpreted as all other arbitrary
+                ``kwargs``.  In Qiskit 2.0, this will instead seed the :class:`.PropertySet` of the
+                compilation (but does not do so in this version).
             kwargs: Arbitrary arguments passed to the compiler frontend and backend.
 
         Returns:
             The transformed program(s).
         """
+        if property_set is not _NOT_PRESENT:
```
ElePT:
```suggestion
        if property_set is not _MISSING:
```

ElePT (commented):
That is a good idea, I prefer "_MISSING" to "_NOT_PRESENT". I wouldn't say that the issue is the double negation as much as having two "nots" one directly after the other and accidentally skipping one.

ElePT (approved):
Nice!

PR #13822, opened 2025-02-10
Optimize Matplotlib Circuit Drawing: Auto-Close Figures to Prevent Memory Leak

miles0428:
### Summary

**Description**  
When drawing quantum circuits using Matplotlib (`circ.draw(output='mpl')`), the figures are not automatically closed after being created. This leads to a memory leak and results in the following warning when creating many circuits:

```
RuntimeWarning: More than 20 figures have been opened. Figures created through the pyplot interface (matplotlib.pyplot.figure) are retained until explicitly closed and may consume too much memory. (To control this warning, see the rcParam figure.max_open_warning). Consider using matplotlib.pyplot.close().
mpl_figure = plt.figure()
```

**Steps to Reproduce**  
1. Create and draw a quantum circuit in a loop: 
```python 
for i in range(101):
    circ = random_circuit(2, 2, max_operands=2)
    circ.draw(output='mpl')
```

2. Run the code via `python xxx.py`.

**Expected Behaviour**  
- Figures should be automatically closed after being drawn to avoid excessive memory usage.
- No warnings should be raised about too many open figures.

### Details and comments

To solve this problem, we could either add a generator **autodeleter** to fix this problem

### DIFF
--- a/qiskit/visualization/circuit/matplotlib.py
+++ b/qiskit/visualization/circuit/matplotlib.py
@@ -54,8 +54,6 @@ from qiskit.qasm3.printer import BasicPrinter
 from qiskit.circuit.tools.pi_check import pi_check
 from qiskit.utils import optionals as _optionals
 
-from matplotlib import pyplot as plt
-
 
 from .qcstyle import load_style
 from ._utils import (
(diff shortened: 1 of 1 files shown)


jakelishman:
It's very common to want to have more than one figure open at once. Even in your example for the functionality, it's _necessary_ to have more than one figure open at once in order to even see the intermediate circuits drawn.

Matplotlib gives you the tools to manage figures itself (e.g. `pyplot.close("all")`) - it's not clear to me that this is something that should be added to Qiskit.

PR #13801, opened 2025-02-06
Better decomposition for multi-controlled 1-qubit gates

ShellyGarion:
### Summary
close #13514

Reduce the CX count of the multi-controlled 1-qubit unitary gates for the gates: H, SX, SXdg and U,
based on the efficient synthesis of MCRX, MCRY, MCRZ, MCX and MCPhase gates.

### Details and comments

<html xmlns:v="urn:schemas-microsoft-com:vml"
xmlns:o="urn:schemas-microsoft-com:office:office"
xmlns:x="urn:schemas-microsoft-com:office:excel"
xmlns="http://www.w3.org/TR/REC-html40">

<head>

<meta name=ProgId content=Excel.Sheet>
<meta name=Generator content="Microsoft Excel 15">
<link id=Main-File rel=Main-File
href="file:///C:/Users/SHELLY~1/AppData/Local/Temp/msohtmlclip1/01/clip.htm">
<link rel=File-List
href="file:///C:/Users/SHELLY~1/AppData/Local/Temp/msohtmlclip1/01/clip_filelist.xml">
<style>

</style>
</head>

<body link="#0563C1" vlink="#954F72">

gate name | num controls | cx count (main) | cx count (new) | reduction
-- | -- | -- | -- | --
y | 1 | 1 | 1 | 1
y | 2 | 24 | 6 | 0.25
y | 3 | 60 | 14 | 0.233333
y | 4 | 112 | 36 | 0.321429
y | 5 | 206 | 84 | 0.407767
y | 6 | 298 | 128 | 0.42953
y | 7 | 496 | 208 | 0.419355
y | 8 | 728 | 312 | 0.428571
y | 9 | 984 | 432 | 0.439024
  |   |   |   |  
h | 1 | 1 | 1 | 1
h | 2 | 18 | 18 | 1
h | 3 | 54 | 34 | 0.62963
h | 4 | 106 | 60 | 0.566038
h | 5 | 200 | 124 | 0.62
h | 6 | 292 | 172 | 0.589041
h | 7 | 490 | 288 | 0.587755
h | 8 | 722 | 416 | 0.576177
h | 9 | 978 | 552 | 0.564417
  |   |   |   |  
sx | 1 | 2 | 2 | 1
sx | 2 | 28 | 6 | 0.214286
sx | 3 | 100 | 20 | 0.2
sx | 4 | 214 | 44 | 0.205607
sx | 5 | 408 | 84 | 0.205882
sx | 6 | 626 | 128 | 0.204473
[...]

PR #13824, opened 2025-02-11
Small improvement for multi-controlled 1-qubit gates

ShellyGarion:
### Summary
As [ACE07-Sev](https://github.com/ACE07-Sev) pointed out in https://github.com/Qiskit/qiskit/pull/13801#issuecomment-2648721053, we could further improve the decomposition of multi-controlled `HGate`, based on the definition of the [`CHGate`](https://docs.quantum.ibm.com/api/qiskit/qiskit.circuit.library.CHGate):

```
        gate ch a,b {
            s b;
            h b;
            t b;
            cx a, b;
            tdg b;
            h b;
            sdg b;
        }
```

### DIFF
--- a/qiskit/circuit/add_control.py
+++ b/qiskit/circuit/add_control.py
@@ -266,8 +266,13 @@ def apply_basic_controlled_gate(circuit, gate, controls, target):
         circuit.s(target)
 
     elif gate.name == "h":
-        circuit.mcry(pi / 2, controls, target, use_basis_gates=False)
+        circuit.s(target)
+        circuit.h(target)
+        circuit.t(target)
         circuit.mcx(controls, target)
+        circuit.tdg(target)
+        circuit.h(target)
+        circuit.sdg(target)
 
     elif gate.name == "sx":
         circuit.h(target)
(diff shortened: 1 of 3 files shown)


Cryoris (commented):
This LGTM, but one question, are there any tests on the multi-controlled H gate that we could use to verify the CX count is below some threshold (and that the unitary is correct)?

ShellyGarion:
> This LGTM, but one question, are there any tests on the multi-controlled H gate that we could use to verify the CX count is below some threshold (and that the unitary is correct)?

Since `test_controlled_gate.py` already contains (too) many tests, I don't want to add extra unnecessary tests.
I've looked at the coverage reports and checked that the new cases are covered, namely that the unitary is correct.
Moreover, I added some printouts to check that controlled h / y / sx / sxdg are checked for 2 and 3 controls.

Cryoris:
Hmm I guess it is implicitly tested here
https://github.com/Qiskit/qiskit/blob/288d2f41b295e93114290a8c0c7e69cb18283aad/test/python/circuit/test_controlled_gate.py#L299

But this is only a fairly imprecise way to test the change in functionality... the previous path was better tested with the explicit tests of `mcry` here https://github.com/Qiskit/qiskit/blob/288d2f41b295e93114290a8c0c7e69cb18283aad/test/python/circuit/test_controlled_gate.py#L622

Given that, I would think that explicit tests for the multi-controlled gates with new decompositions would be good to have. They don't have to be extensive (i.e. only a single test case would suffice), but I think we should have a place that checks that at least the operators are correct 🤔

ShellyGarion:
tests added in [66222ad](https://github.com/Qiskit/qiskit/pull/13824/commits/66222ad4b254e076d98f47f42c7e619f00b27ab5)

Cryoris (approved):
LGTM, thanks for the additions!

PR #13829, opened 2025-02-12
Remove pulse from GenericBackendV2

eliarbel:
This commit removes pulse-related functionality from GenericBackendV2, as part of Pulse removal in Qiskit 2.0. This includes the ability to initialize the backend with custom calibrations and query it for channel information. Also, various clean ups where made to accommodate for the updated API of GenericBackendV2.

Part of #13662


qiskit/pulse/builder.py
```diff
@@ -104,14 +104,12 @@
 
 Methods to return the correct channels for the respective qubit indices.
 
-.. plot::
-   :include-source:
-   :nofigs:
+.. code-block:: python
```
ElePT, line 107: I understand that you skipped the execution of the examples here because they won't work without calibrations. It is a bit strange to keep an example we know is incorrect, but I don't feel too strongly about it given that the class will be removed shortly anyways.

releasenotes/notes/remove-pulse-generic-backendv2-738ad9f7ab64b8fd.yaml
```diff
@@ -0,0 +1,6 @@
+---
+upgrade_providers:
+  - |
+    As part of Pulse removal in Qiskit 2.0, pulse support has been removed from
+    :class:`.GenericBackendV2`. This includes the ability to initialize the backend
+    with custom calibrations and pulse channels information.
```
ElePT: I would be a bit more specific and name the argument explicitly (at least I navigate the release notes using CTRL+F so having the keywords makes them easer to find)

```suggestion
    As part of Pulse removal in Qiskit 2.0, pulse support has been removed from
    :class:`.GenericBackendV2`. This includes the ability to initialize the backend
    with custom calibrations (``calibrate_instructions`` argument) and pulse channels information.
```

ElePT (dismissed):
LGTM! I just left a minor suggestion in the reno.

qiskit/pulse/builder.py
eliarbel: Exactly. The main motivation to do it like this is to keep the changes here as minimal as possible w.r.t the scope of this PR

ElePT:
Sorry for the conflict, I hope it's not too painful to address.

ElePT (approved):
LGTM! I am aware that the broken coverage tests are due to external causes, so I would move ahead with this PR regardless of the state of those tests.

PR #13830, opened 2025-02-12
Add `dt` to `GenericBackendV2`

ElePT:
### Summary
This PR allows to configure `dt` from the `GenericBackendV2` init, a useful feature for some scheduling tests. This should have been added from the start but was likely left out as an oversight.


mtreinish (approved):
LGTM, thanks for doing this

PR #13832, opened 2025-02-12
[Stretch] Support `Float` type in classical expressions.

kevinhartman:
### Summary
Adds a new `Float` type to the classical expression system. This is necessary to express timing relationships between durations and stretches in expressions.

In a follow-up(s), I'll add `Duration` and `Stretch` types as well as binary operators for `+`, `-`, `*` and `/`, and then plumb all of this into the existing circuit variable tracking infra, integrate with the `Delay` instruction, support serialization, etc.

### Details and comments

~Currently based on #13811. Here's [a readable diff](https://github.com/kevinhartman/qiskit/compare/const-expr...kevinhartman:qiskit:expr-float) in the meantime.~

* The `Float` type is considered of an "unspecified width", in accordance with the machine-precision `float` in OpenQASM3. If we add a fixed-width float (similar to what we've got for `Uint`), then it shouldn't be difficult to add an optional constructor argument to `Float` for `width`, defaulted to `None`. I didn't opt to do this since I imagine fixed-width floats aren't desperately needed for stretches, and they would've taken too long to stamp out.
* `expr.lift` now supports Python floats.
  * Floats can indeed be const.
* Qiskit is stricter than QASM, in that its type system considers casts from `Uint` or `Bool` to `Float` to be unsafe.
  * You can still do it, but it requires an explicit cast node.
* For this reason, `Float` cannot be mixed with `Uint` in binary operations. You must cast one of them, first.
* Floats are only directly usable with the existing binary relation operations. The rest forbid direct use of `Float` for one reason or another (e.g. not losslessly coercable to `Bool`). The arithmetic operations to come (`+`, `-`, `*` and `/`) will support them.

#### To-do
- [x] QPY support.
- [x] QASM support.


qiskit/circuit/classical/expr/constructors.py
```diff
@@ -32,6 +33,9 @@
     "less_equal",
     "greater",
     "greater_equal",
+    "shift_left",
+    "shift_right",
+    "index",
```
jakelishman, line 38: whoops, thanks

qiskit/circuit/classical/expr/constructors.py
```diff
@@ -181,11 +188,16 @@ def logic_not(operand: typing.Any, /) -> Expr:
             >>> expr.logic_not(ClassicalRegister(3, "c"))
             Unary(\
 Unary.Op.LOGIC_NOT, \
-Cast(Var(ClassicalRegister(3, 'c'), Uint(3)), Bool(), implicit=True), \
+Cast(Var(ClassicalRegister(3, 'c'), Uint(3)), \
+Bool(), implicit=True), \
```
jakelishman, line 190: super minor: are you able to stop whatever tool it is you have from modifying the line-wrapping when it's not necessary?  Just adds a bit more noise in review

qiskit/circuit/classical/expr/constructors.py
```diff
@@ -181,11 +188,16 @@ def logic_not(operand: typing.Any, /) -> Expr:
...
 Bool())
     """
-    operand = _coerce_lossless(lift(operand), types.Bool())
-    return Unary(Unary.Op.LOGIC_NOT, operand, operand.type)
+    operand = lift(operand)
+    try:
+        operand = _coerce_lossless(operand, types.Bool())
+        return Unary(Unary.Op.LOGIC_NOT, operand, operand.type)
+    except TypeError as ex:
+        raise TypeError(f"cannot apply '{Unary.Op.BIT_NOT}' to type '{operand.type}'") from ex
```
jakelishman: only the `_coerce_lossless` can fail, so if you want to modify its exception
```python
try:
    operand = _coerce_lossless(operand, types.Bool())
except TypeError:
    raise TypeError(...) from None
return Unary(..., operand, operand.type)
```
might be a bit clearer about the scope of the `try`.

As minor nitpicks too:
- the chained exception is an internal implementation detail, so it's potentially cleaner for the user to suppress chaining with `raise TypeError(...) from None`, otherwise it kind of looks like an internal mistake (imo)
- `Unary.Op.LOGIC_NOT` is probably too low-level detail for a user of this function (also you have a typo in the operand) - it's maybe clearer to say "cannot apply 'logic_not' to type ..." so the link is clearer

qiskit/circuit/classical/expr/constructors.py
```diff
@@ -300,9 +312,14 @@ def bit_xor(left: typing.Any, right: typing.Any, /) -> Expr:
...
-    right = _coerce_lossless(lift(right), bool_)
-    return Binary(op, left, right, bool_)
+    left = lift(left)
+    right = lift(right)
+    try:
+        left = _coerce_lossless(left, bool_)
+        right = _coerce_lossless(right, bool_)
+        return Binary(op, left, right, bool_)
+    except TypeError as ex:
+        raise TypeError(f"invalid types for '{op}': '{left.type}' and '{right.type}'") from ex
```
jakelishman: Same comments here about exception handling as before.

That said, `_coerce_lossless` is purely an internal detail, so if you want to modify how  it reports errors (such as `return None` on error and require checking, etc), that's fine by me too.

qiskit/circuit/classical/types/types.py
```diff
@@ -19,11 +19,7 @@
 
 from __future__ import annotations
 
-__all__ = [
-    "Type",
-    "Bool",
-    "Uint",
-]
+__all__ = ["Type", "Bool", "Float", "Uint"]
```
jakelishman: No need to change it back, but I _think_ I deliberately put in the trailing comma because this is a list that's expected to grow.

qiskit/circuit/classical/types/types.py
```diff
@@ -115,3 +111,21 @@ def __hash__(self):
 
     def __eq__(self, other):
         return isinstance(other, Uint) and self.width == other.width
+
+
+@typing.final
+class Float(Type, metaclass=_Singleton):
+    """A floating point number of unspecified width.
+    In the future, this may also be used to represent a fixed-width float.
+    """
```
jakelishman: In retrospect: I think we can say that this produces a double-precision IEE-754 by default, since that's what we'll actually be storing.  Nicer to be explicit, I think.

Any nastiness with quads, singles or halves can be handled some other time, but defaulting to doubles is fine, I think.

qiskit/qasm3/ast.py
```diff
@@ -136,6 +136,7 @@ class ClassicalType(ASTNode):
 class FloatType(ClassicalType, enum.Enum):
     """Allowed values for the width of floating-point types."""
 
+    UNSPECIFIED = 0
```
jakelishman: Similar to the type story - let's just be explicit and always have Qiskit output double precision for now.  (Fwiw, in a sane AST - which this is absolutely not an example of lol -, I think we would have called this enum `FloatWidth`, and had `FloatType` be a record type with an optional `FloatWidth`.)

qiskit/qpy/__init__.py
```diff
@@ -377,6 +377,37 @@ def open(*args):
...
+Changes to EXPR_VALUE
+~~~~~~~~~~~~~~~~~~~~~
+
+The classical expression's type system now supports new encoding types for value literals, in
+addition to the existing encodings for int and bool. The new value type encodings are below:
+
+===========  =========  ============================================================================
+Python type  Type code  Payload
+===========  =========  ============================================================================
+``float``    ``f``      One ``double value``.
```
jakelishman, line 407: I feel a great disturbance in the force, as if millions C programmers suddenly cried out in terror and were suddenly silenced.

(This is the right thing, no worries, I just have PTSD from `%f` vs `%lf` when learning to program in early uni haha.)

qiskit/qpy/formats.py
```diff
@@ -360,6 +360,10 @@
 
 # EXPR_TYPE
 
+EXPR_TYPE_FLOAT = namedtuple("EXPR_TYPE_FLOAT", [])
+EXPR_TYPE_FLOAT_PACK = "!"
+EXPR_TYPE_FLOAT_SIZE = struct.calcsize(EXPR_TYPE_FLOAT_PACK)
+
 EXPR_TYPE_DISCRIMINATOR_SIZE = 1
```
jakelishman: super super minor: can you put the discriminator first, since that's common to all things of `EXPR_TYPE`?

test/python/circuit/classical/test_expr_constructors.py
```diff
@@ -133,6 +139,11 @@ def test_bit_not_explicit(self):
             expr.Unary(expr.Unary.Op.BIT_NOT, expr.Var(clbit, types.Bool()), types.Bool()),
         )
 
+    @ddt.data(expr.bit_not)
+    def test_urnary_bitwise_forbidden(self, function):
```
jakelishman: `urnary` lol

test/python/circuit/classical/test_expr_constructors.py
```diff
@@ -149,6 +160,11 @@ def test_logic_not_explicit(self):
             expr.Unary(expr.Unary.Op.LOGIC_NOT, expr.Var(clbit, types.Bool()), types.Bool()),
         )
 
+    @ddt.data(expr.logic_not)
+    def test_urnary_logical_forbidden(self, function):
+        with self.assertRaisesRegex(TypeError, "cannot apply"):
+            function(7.0)
```
jakelishman: not sure the ddt parametrisation is buying anything here haha

(also: urnary again)

test/python/circuit/classical/test_expr_constructors.py
```diff
@@ -388,16 +482,29 @@ def test_index_explicit(self):
...
 
     def test_index_forbidden(self):
         with self.assertRaisesRegex(TypeError, "invalid types"):
             expr.index(Clbit(), 3)
         with self.assertRaisesRegex(TypeError, "invalid types"):
             expr.index(ClassicalRegister(3, "a"), False)
+        with self.assertRaisesRegex(TypeError, "invalid types"):
+            expr.index(ClassicalRegister(3, "a"), 1.0)
+        with self.assertRaisesRegex(TypeError, "invalid types"):
+            expr.index(0xFFFF, 1.0)
```
jakelishman, line 509: Can  you add one that asserts it's invalid to index into a float with an integer too?

qiskit/circuit/classical/types/__init__.py
```diff
@@ -95,6 +95,7 @@
 __all__ = [
     "Type",
     "Bool",
+    "Float",
```
jakelishman, line 102: Need to add `Float` to this module's docstring above too.

jakelishman (commented):
github says I have 12 comments, but I think most of them are either ludicrously minor or just jokes

I think there was only two real ones (though they appear in a couple of places each) - questions about the exception handling and one about maybe making the float width explicit.

qiskit/circuit/classical/expr/constructors.py
kevinhartman: It's actually me, and I won't typically do this 🤣. It's an artifact of moving const-ness from `Type` to `Expr` and having to manually update the comments (they needed wrapping when the types included their const-ness in their initializers).

qiskit/circuit/classical/expr/constructors.py
kevinhartman: I'll make these changes, but I think we ought to use `Unary.Op.LOGIC_NOT` in the message since that's what the other constructors do.

qiskit/circuit/classical/expr/constructors.py
kevinhartman: I like that better, I'll go ahead and change it to return `None`.

qiskit/circuit/classical/types/types.py
kevinhartman: Should we also emit a `float[64]` in QASM, though? I guess this would happen if a user were to create a `Cast` expression to a `Float`, e.g. QASM would emit `(Float[64])my_int` I think.

qiskit/circuit/classical/expr/constructors.py
jakelishman: Ok sounds fine if there's a consistency argument.  Now you mention it, I think I used the `Binary.Op` thing mostly just out of ease.

qiskit/circuit/classical/types/types.py
jakelishman: Yeah, I'm saying let's treat default-constructed `Float()` in Qiskit land as if it were `float[64]` in OpenQASM land?  If we later want to specify some `width=None` override in Qiskit land, we could make `width=64` the default.

I'm mostly thinking that the constants available to users in Python land _are_ `float[64]`, so we may as well just represent that reality verbatim.

kevinhartman: That sounds good, done in cbd55dbfe05aa0fd6e752cecc334c3b2a9f0c067.

releasenotes/notes/float-expr-02b01d9ea89ad47a.yaml
```diff
@@ -0,0 +1,17 @@
+---
+features_circuits:
+  - |
+    The classical realtime-expressions module :mod:`qiskit.circuit.classical` can now represent
+    IEE-754 double-precision floating point values using the new type :class:`~.types.Float`.
```
jakelishman: IEEE with 3 Es

qiskit/circuit/classical/expr/constructors.py
```diff
@@ -45,19 +49,17 @@
...
     if kind is CastKind.EQUAL:
         return expr
     if kind is CastKind.IMPLICIT:
         return Cast(expr, type, implicit=True)
     if kind is CastKind.LOSSLESS:
         return Cast(expr, type, implicit=False)
-    if kind is CastKind.DANGEROUS:
-        raise TypeError(f"cannot cast '{expr}' to '{type}' without loss of precision")
-    raise TypeError(f"no cast is defined to take '{expr}' to '{type}'")
+    return None
```
jakelishman, line 62: Checking this branch out locally, I can see several uses of `_coerce_lossless` in this file that appear to have unchecked return values after this change.

In particular: `_equal_like`, `_binary_relation`, and `_shift_like` (but I might have missed one or two).

I think the `_equal_like` and `_binary_relation` ones shake out ok because of `type.kind` checks higher up, but it would be good to put a one-liner comment into each explaining that.

The `_shift_like` one has a potential bug:
```python
from qiskit.circuit.classical import expr, types

expr.shift_left(expr.lift(5.0), 3, types.Uint(8))
```
is an `AttributeError`.

I don't know _why_ I gave `shift_left` and `shift_right` a cast field within them - I think I must have been attempting to save some typing with inserting an extra `expr.lift` on the left operand when passing an integer literal and trying to fix the width (since the left- and right operands don't need to have equal widths in shifts, so there's no inference) - but at any rate, now it's possible to fail the cast in there.

jakelishman (commented):
Thanks for this.  I think I shook out a bug in an esoteric combination of arguments, but that's it here.

test/qpy_compat/test_qpy.py
```diff
@@ -833,13 +833,19 @@ def generate_v12_expr():
...
+        pass
+
+    return [float_expr]
 
-    If load_context is True, avoid generating Pulse-based reference
-    circuits. For those circuits, load_qpy only checks that the cached
-    circuits can be loaded without erroring."""
 
+def generate_circuits(version_parts):
+    """Generate reference circuits."""
```
jakelishman: You've got a borked merge here - that's what's breaking the QPY tests.

kevinhartman: whoops!

releasenotes/notes/float-expr-02b01d9ea89ad47a.yaml
jakelishman: also appears in the `types.Float`, I think - I saw it in another PR.

qiskit/circuit/classical/expr/constructors.py
kevinhartman: Good catch. I've fixed this and added a test in 0f14cca4e604c025fe1b1c09c6d85743c6b45014

test/qpy_compat/test_qpy.py
kevinhartman: Resolved in 0f14cca4e604c025fe1b1c09c6d85743c6b45014

jakelishman (approved):
ace, assuming this passes the tests, I think this one's good to go too, thanks.

Sorry in advance for the large question left on the next one...

PR #9802, opened 2023-03-15
Add support for disjoint coupling maps to SabreLayout

mtreinish:
### Summary

This commit adds support to the SabreLayout pass for targeting disjoint
CouplingMap objects. The SABRE algorithm is written assuming a connected
graph, so to enable targeting a disjoint coupling graph this commit
decomposes the circuit into it's weakly connected components and then
maps those to the connected components of the coupling map and runs
the rust portion of the sabre code on each connected component of the
coupling graph. The results for each subgraph of the target backend is
then combined the pass is building the output dag.
    
In general the biggest potential issue for output quality right now is
the mapping function does not take into account the relative connectivity
of the dag component and the coupling map component. The mapping is just
done in order and it tries every component until it finds one that has
enough qubits. In the future we should use a heuristic to try and pack
the components based on what we expect will require the least amount of
swaps, but we can attempt that in a future PR.

### Details and comments

~This is still a **very** rough draft of this feature. Right now it does not handle barriers correctly at all
because an input circuit can have a multi-qubit barrier could correctly span more qubits than in a single
connected component and this will incorrectly treat that as a single dag component resulting in the circuit
being incorrectly mapped (or not being mappable at all).  This prevents this from being used in the preset
pass managers at all because we need to insert a multiqubit barrier before the final measurement to prevent
routing past a measurement.~ This should be fixed by: https://github.com/Qiskit/qiskit-terra/pull/9802/commits/b485d6a00eecf73fdad52f81bdcc87670348458e

~I pushed this up primarily as a way to save my work between systems and share the basic approach while I'm
still developing the feature to fix these edge cases. But in isolation for circuits without barriers (or circuits where
the barrier width is less than or equal to the width of the coupling graph's connected components) this seems to
work correctly.~

TODO:

- [x] Rebase after #9710 merges
- [x] Fix handling of barriers
- [x] Fix DAG splitting to ignore classical component
- [x] Expand testing
- [x] Add release note

This PR is based on top of #9710 in its current form, to see the contents of what will be this PR you can look at this diff: https://github.com/mtreinish/qiskit-core/compare/disjoint-coupling-map...mtreinish:qiskit-core:layout_and_route_disjoint_coupling_maps

PR #13833, opened 2025-02-12
Fix tracking of routing permutation in Sabre with disjoint backends

### LINKED ISSUE #13732: Error when running `final_index_layout(filter_ancillas=False)`
### Environment

- **Qiskit version**: 1.3.2
- **Python version**:  3.11
- **Operating system**: macos/linux

### What is happening?

On certain circuits transpiled with a backend and a coupling map, this call fails
```
transpiled_circuit.layout.final_index_layout(filter_ancillas=False)
```
raising `KeyError: "The item Qubit(QuantumRegister(5, 'q'), 3) does not exist in the Layout"`.

This worked fine until qiskit 1.3.0

Running the following code:

```
from qiskit import QuantumCircuit, transpile, qasm2
from qiskit_ibm_runtime import QiskitRuntimeService

service = QiskitRuntimeService(channel="ibm_quantum")
backend = service.backend("ibmq_manila")

qc = qasm2.loads('OPENQASM 2.0;include "qelib1.inc";qreg q[3];rccx q[2],q[1],q[0];t q[2];u(0.28905709212845454,5.626877140821112,5.481851443455051) q[1];h q[0];cx q[1],q[2];s q[0];', custom_instructions=qasm2.LEGACY_CUSTOM_INSTRUCTIONS)

transpiled_circuit = transpile(qc,backend=backend,coupling_map=[[0,1],[1,2],[2,3]],optimization_level=2,seed_transpiler=42)
transpiled_circuit.layout.final_index_layout(filter_ancillas=False)
```

### What should happen?

There should not be any error in the call. The output to that particular example in `qiskit 1.2.4` is `[3, 1, 2, 0]`

### Any suggestions?

_No response_

jakelishman:
### Summary

If the backing coupling graph is disjoint, and unused components of the coupling graph would not be considered when constructing the complete routing permutation.  In practice, Sabre aborts immediately after layout without attempting to route, if it needed to split the DAG across more than one disjoint component, because it can't guarantee correctness of the final routing in the presence of component-spanning barriers or classical communication, so the only way for a component to be forgotten is if the backend is disjoint, but the DAG fits into a single component.

### Details and comments

Fix #13732, though there's a _whole_ cornucopia of bugs split between very old versions of Qiskit, this commit's parent, and the example in the issue itself that lead from that issue seeing a regression in Qiskit 1.3 to this one.

In actuality, this bug has been around since the disjoint-coupling handling was first added in #9802 (qiskit-terra 0.24), and the test case in this PR would fail that version too (the fact that `TranspileLayout.initial_index_layout` wasn't added til later notwithstanding - the first assertion fails).

### DIFF
--- a/test/python/transpiler/test_sabre_layout.py
+++ b/test/python/transpiler/test_sabre_layout.py
@@ -470,6 +470,21 @@ class TestDisjointDeviceSabreLayout(QiskitTestCase):
         layout = pm.property_set["layout"]
         self.assertEqual([layout[q] for q in qc.qubits], [3, 1, 2, 5, 4, 6, 7, 8])
 
+    def test_dag_fits_in_one_component(self):
+        """Test that the output is valid if the DAG all fits in a single component of a disjoint
+        coupling map.."""
+        qc = QuantumCircuit(3)
+        qc.cx(0, 1)
+        qc.cx(1, 2)
+        qc.cx(2, 0)
+
+        disjoint = CouplingMap([(0, 1), (1, 2), (3, 4), (4, 5)])
+        layout_routing_pass = SabreLayout(disjoint, seed=2025_02_12, swap_trials=1, layout_trials=1)
+        out = layout_routing_pass(qc)
+        self.assertEqual(len(out.layout.initial_layout), len(out.layout.final_layout))
+        self.assertEqual(out.layout.initial_index_layout(filter_ancillas=False), [1, 0, 2, 3, 4, 5])
+        self.assertEqual(out.layout.final_index_layout(filter_ancillas=False), [2, 0, 1, 3, 4, 5])
+
 
 class TestSabrePreLayout(QiskitTestCase):
     """Tests the SabreLayout pass with starting layout created by SabrePreLayout."""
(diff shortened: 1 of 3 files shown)


test/python/transpiler/test_sabre_layout.py
```diff
@@ -470,6 +470,21 @@ def test_with_partial_layout(self):
...
+    def test_dag_fits_in_one_component(self):
+        """Test that the output is valid if the DAG all fits in a single component of a disjoint
+        coupling map.."""
+        qc = QuantumCircuit(3)
+        qc.cx(0, 1)
+        qc.cx(1, 2)
+        qc.cx(2, 0)
+
+        disjoint = CouplingMap([(0, 1), (1, 2), (3, 4), (4, 5)])
+        layout_routing_pass = SabreLayout(disjoint, seed=2025_02_12, swap_trials=1, layout_trials=1)
```
ElePT, line 482: Very precise seed

ElePT (approved):
LGTM!! Thanks @jakelishman for diving right into the rabbit hole!

ElePT:
Hmm I removed the PR from the queue because I saw the failed neko test in the merge CI run, but I don't think it's related to this PR. I think it's related to changes in qiskit-aer.

PR #13836, opened 2025-02-13
`SparseObservable` evolution

### LINKED ISSUE #13498: Support `SparseObservable` in `PauliEvolutionGate`
### What should we add?

What the title says 🙂

### LINKED ISSUE #13764: Add `SparseObservable` support for `PauliEvolutionGate`
### What should we add?

Enable building evolution gates from `SparseObservable`. This needs some extra care to handle the projectors efficiently.

Cryoris:
### Summary

Add support for `SparseObservable` in `PauliEvolution`.

### Details and comments

Evolution under projectors is implemented by controlling the target rotation with open controls (for +1 eigenstate projectors) or closed controls (for -1 eigenstate projectors). This is significantly more efficient than decomposing projectors into Paulis, shown here for the implementation of $exp(-i t Z \otimes |1\rangle\langle 1|^{\otimes n})$ for different numbers of projectors $n$.

image (1)


qiskit/circuit/library/pauli_evolution.py
```diff
@@ -159,32 +175,43 @@ def validate_parameter(self, parameter: ParameterValueType) -> ParameterValueTyp
...
-        # for just a single Pauli don't add brackets around the sum
-        else:
-            label = f"exp(-it ({' + '.join(operator.paulis.to_labels())}))"
+        return f"exp(-it ({[_operator_label(op) for op in operator]}))"
+    return f"exp(-it {_operator_label(operator)})"
+
 
-    return label
+def _sparse_term_label(term):
+    return "".join(BIT_LABELS[bit] for bit in reversed(term.bit_terms))
```
mtreinish: I feel like we should do this in Rust as a pyfunction because it'd be faster and we could use the enum variants for `BitTerm` directly. Not that it really matters, but it feels more natural in rust.

crates/accelerate/src/circuit_library/pauli_feature_map.rs
```diff
@@ -178,17 +178,17 @@ fn _get_evolution_layer<'a>(
...
-            )
-            .map(|(gate, params, qargs)| {
-                (gate.into(), params, qargs.to_vec(), vec![] as Vec<Clbit>)
-            })
-            .collect::<Vec<Instruction>>();
+            );
+            // .map(|(gate, params, qargs)| {
+            //     (gate.into(), params, qargs.to_vec(), vec![] as Vec<Clbit>)
+            // })
+            // .collect::<Vec<Instruction>>();
```
mtreinish:
```suggestion
```

mtreinish (commented):
This is really cool, I've only skimmed the PR so far, but I had a couple small nits/comments.

qiskit/circuit/library/pauli_evolution.py
Cryoris: I was thinking of adding a method to `SparseTerm`, maybe something like `bit_labels()` or so. But a standalone function would ofc also work

Cryoris: I added a `SparseTerm.bit_labels()` method in 4881b5f that allows to get the labels Rust side. They're then ordered in the same fashion as `SparseTerm.indices`

crates/accelerate/src/circuit_library/pauli_evolution.rs
```diff
@@ -10,48 +10,48 @@
...
 
-// custom types for a more readable code
-type StandardInstruction = (StandardGate, SmallVec<[Param; 3]>, SmallVec<[Qubit; 2]>);
+// custom type for a more readable code
 type Instruction = (
     PackedOperation,
     SmallVec<[Param; 3]>,
     Vec<Qubit>,
     Vec<Clbit>,
 );
```
alexanderivrii, line 29: A single `Instruction` is starting to look more and more like `PackedInstruction` and a vector/iterator over these is starting to look more and more like `CircuitData`.

crates/accelerate/src/circuit_library/pauli_evolution.rs
```diff
@@ -101,88 +123,192 @@ fn single_qubit_evolution(
...
+            cx_chain(pauli_qubits.clone()).rev(),
         ),
     };
 
-    // get the RZ gate on the first qubit
-    let first_qubit = active_paulis.first().unwrap().1;
-    let z_rotation = std::iter::once((
-        if phase_gate {
+    // Get the Z/phase rotation. If we have more than a single rotation qubit, each projector
+    // is implemented as open/closed control of the rotation.
```
alexanderivrii, line 252: To make sure I understand the logic correctly:

- If we have a "pure" multi-qubit Pauli rotation, then using the "basis change" gates and the CX-gates, we can reduce it to implementing a single-qubit Phase (or a single-qubit RZ)-gate.
- if we also have projectors, then using the "basis change gates" (over both pauli and projector qubits) and CX-gates (over Pauli qubits), we can reduce it to implementing a controlled-RZ rotation.
- and there is a special case when we have no pauli qubits, only projector qubits

qiskit/circuit/library/pauli_evolution.py
```diff
@@ -20,11 +20,23 @@
...
+    0b0001: "Z",
+    0b1001: "0",
+    0b0101: "1",
+    0b0010: "X",
+    0b1010: "+",
+    0b0110: "-",
+    0b0011: "Y",
+    0b1011: "r",
+    0b0111: "l",
+}
```
alexanderivrii: Why these specific binary values?

qiskit/synthesis/evolution/lie_trotter.py
```diff
@@ -87,6 +94,7 @@ def __init__(
             atomic_evolution,
             wrap,
             preserve_order=preserve_order,
+            atomic_evolution_sparse_observable=atomic_evolution_sparse_observable,
```
alexanderivrii, line 97: Lol, you used to like short names 😄

qiskit/synthesis/evolution/suzuki_trotter.py
```diff
@@ -88,15 +90,21 @@ def __init__(
             preserve_order: If ``False``, allows reordering the terms of the operator to
                 potentially yield a shallower evolution circuit. Not relevant
                 when synthesizing operator with a single term.
+            atomic_evolution_sparse_observable: If a custom ``atomic_evolution`` is passed,
+                which does not yet support :class:`.SparseObservable`\ s as input, set this
+                argument to ``False`` to automatically apply a conversion to :class:`.SparsePauliOp`.
+                This argument is supported until Qiskit 2.2, at which point all atomic evolutions
+                are required to support :class:`.SparseObservable`\ s as input.
+
```
alexanderivrii, line 98: Does it make sense to enforce this requirement in 2.2? Automatically converting to a `SparsePauliOp` seems like a nice default.

test/python/circuit/library/test_evolution_gate.py
```diff
@@ -519,12 +508,156 @@ def test_sympify_is_real(self):
...
+        evo = PauliEvolutionGate(op, time=1).definition
+
+        reference = QuantumCircuit(*evo.qregs)
+        reference.sx([2, 3])
+        reference.h([4, 5])
+
+        reference.append(PhaseGate(-1.0).control(5, ctrl_state="00110"), reference.qubits)
+
+        reference.sxdg([2, 3])
+        reference.h([4, 5])
```
alexanderivrii, line 611: As discussed offline, could you add a comment explaining why this is the reference circuit?

alexanderivrii (commented):
This is a really cool PR indeed! I have left a few minor questions and comments.

qiskit/circuit/library/pauli_evolution.py
ShellyGarion: @alexanderivrii - see the table in [`SparseObservable`](https://docs.quantum.ibm.com/api/qiskit/qiskit.quantum_info.SparseObservable) doc

qiskit/synthesis/evolution/suzuki_trotter.py
ShellyGarion: I also don't understand this parameter. It appears in the docs, but not in the release notes, and there are no tests for it

crates/accelerate/src/circuit_library/pauli_evolution.rs
Cryoris: Yes to all of the above 😄

qiskit/synthesis/evolution/lie_trotter.py
Cryoris: hahah you caught me! It seems my brain got lazy!

qiskit/synthesis/evolution/suzuki_trotter.py
Cryoris: Users can just do the conversion themselves if there's no custom logic for `SparseObservable`s, so I thought removing it asap would be good 🙂 

@ShellyGarion currently, an `atomic_evolution` provided by the user must only be able to handle `SparsePauliOp`, but with this PR we'd also like them to support `SparseObservable`. If this flag is `True` (current default) we'll automatically convert `SparseObservable`s to `SparsePauliOp` before calling the `atomic_evolution`, in order not to break existing code. But to transition the case where both are supported, we'll change this flag in the future.

I'll add a test where this flag is explicitly set.

test/python/circuit/library/test_evolution_gate.py
```diff
@@ -519,12 +508,156 @@ def test_sympify_is_real(self):
...
+    )
+    def test_projectors(self, projector):
+        """Test a SparseObservable with projectors."""
+        op = SparseObservable(projector)
+        evo = PauliEvolutionGate(op, time=1).definition
+        pauli = SparsePauliOp.from_sparse_observable(op)
+        ref = PauliEvolutionGate(pauli, time=1).definition
+        self.assertEqual(Operator(ref), Operator(evo))
+
+    def test_projector_custom_index(self):
```
ShellyGarion: this test and the next one are the same but with different inputs, right?
you may consider merging them

ShellyGarion (commented):
LGTM. very nice improvement! 
Perhaps you can add some more tests in the spirit of the graphs presented in the PR, showing that you get the same result, and providing some bounds on the CX cost or depth.
but anyway it's a great PR!

test/python/circuit/library/test_evolution_gate.py
Cryoris: done in 4073279

test/python/circuit/library/test_evolution_gate.py
Cryoris: done in 4073279

qiskit/synthesis/evolution/suzuki_trotter.py
Cryoris: done in 4073279

qiskit/circuit/library/pauli_evolution.py
Cryoris: this dict is actually no longer needed, removed in 4073279

alexanderivrii:
LGTM on my side.

ShellyGarion (approved):

PR #13837, opened 2025-02-13
delete erroneous note in XXMinusYY documentation

kevinsung:
### Summary

Actually, it doesn't matter which qubit the phase is applied to because the gate only affects the symmetric subspace containing the 00 and 11 states.


jakelishman (approved):
Thanks for this.

There's an effort (on the back burner at the moment) to look over and improve _all_ the circuit-library documentation, and personally I'd hope we can make these notes a lot better - I very much understand the bit-labelling conventions, and I find these notes super confusing, so I think we can do a better job.

That's unrelated to your PR of course, just me musing.

PR #13838, opened 2025-02-13
Deprecate backend_props arg in generate_translation_pass_manager

### LINKED ISSUE #13831: Breaking change in `generate_translation_passmanager`
> This was a breaking change to a public API - it breaks `qiskit-ibm-runtime`'s translation plugins.  The option wasn't deprecated previously, so we ought to at a minimum issue a deprecation warning.  Since it's necessary to pass for the entirety of Qiskit 1.x, I wouldn't be opposed to us leaving it as a deprecated no-op argument for the duration of Qiskit 2.x.

_Originally posted by @jakelishman in https://github.com/Qiskit/qiskit/pull/13706#discussion_r1952868621_

Given that there is no alternative in the 1.x series (you _must_ pass this option on), I think making it a deprecated no-op argument in Qiskit 2.0 is possibly the neater choice.

mtreinish:
### Summary

This commit deprecates the backend_props argument for the generate translation pass manager helper function. This argument was removed as part of cleaning up the transpiler interface in 2.0 to remove loose constraints and any tie to the BackendV1 models objects. However, the fact that this function was public was overlooked and a deprecation was not added ahead of time.

### Details and comments

Fixes #13831

### DIFF
--- a/qiskit/transpiler/preset_passmanagers/common.py
+++ b/qiskit/transpiler/preset_passmanagers/common.py
@@ -15,6 +15,7 @@
 
 import collections
 from typing import Optional
+import warnings
 
 from qiskit.circuit.equivalence_library import SessionEquivalenceLibrary as sel
 from qiskit.circuit.controlflow import CONTROL_FLOW_OP_NAMES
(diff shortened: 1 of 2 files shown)


ElePT:
I was going to extend a bit the message in https://github.com/Qiskit/qiskit/pull/13719, the PR that originally deprecated the argument (the one that triggered  https://github.com/Qiskit/qiskit/issues/13831). Would you be ok with that? That would allow to have everything contained in a single reno.

PR #13841, opened 2025-02-13
Deprecate subclassing of `Register` and `Bit`

jakelishman:
The data represented by these classes will move to Rust space in Qiskit 2.0, and the remaining Python objects will be thin wrappers around the Rust-space objects.  Rust space will not support additional bits beyond the existing triple `Qubit/AncillaQubit/Clbit`, and so Python-space will be unable to represent anything beyond this in the circuit data model.

It's not clear what, if any, meaning was intended by subclassing these objects already, and Qiskit did not consider them subclass safe.  We don't expect that anybody will have been doing this, so the impact should be minimal.

### Details and comments

This is prep work for #13686, or an in-the-works PR by Ray that will supersede that one.

### DIFF
--- a/qiskit/circuit/bit.py
+++ b/qiskit/circuit/bit.py
@@ -14,6 +14,7 @@
 Quantum bit and Classical bit objects.
 """
 import copy
+import warnings
 
 from qiskit.circuit.exceptions import CircuitError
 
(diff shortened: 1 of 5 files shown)


raynelfss (approved):
LGTM! At first, I was confused about the check but checking the destination module sounds logical.

jakelishman:
Yeah, I needed some way of marking that `ClassicalRegister`, `QuantumRegister` and `AncillaRegister` are "fine", and ended up with the somewhat Nixon-esque "when the library does it, that means it's not a crime" form you see here...

PR #13844, opened 2025-02-14
[Stretch] Support `Duration` type in classical expressions.

kevinhartman:
### Summary
Adds new type `Duration` to the classical expression type system, as well as a new (and separate) core `Duration` type to `qiskit.circuit`.

### Details and comments
~Based on #13832. Here's [a readable diff](https://github.com/kevinhartman/qiskit/compare/expr-float...kevinhartman:qiskit:expr-timing) in the meantime.~ **Rebased on main.**

#### Core type
The new core type `Duration` is a `pyclass` implemented in Rust as an `enum` type. For now, it's got variants for each of the supported timing kinds defined in OpenQASM (`ns`, `us`, `ms`, `s`, `dt`). All of these are `f64`s, except for `dt` which is a `u64`.

Equality of `Duration` currently follows the same choices made by #13816, in that durations are not first brought to the same unit and compared for magnitude. Perhaps in the future, we can consider equality between walltimes in different units if that feels useful.

#### Expression types

Like in QASM, the `Duration` type represents a length of time, possibly negative. Expressions of type `Duration` may be resolvable immediately, or may contain `stretch` expressions (to be added in #13852).

#### To-do

- [x] Decide on `Duration` representation and expression type (`Value` or `Var`).
- [x] Decide on `Duration` equality.
- [x] Support QPY and QASM.
- [x] QPY testing.
- [x] QASM testing.
- [x] Update QPY description.
- [x] Release note.


qiskit/circuit/classical/expr/expr.py
```diff
@@ -144,7 +144,14 @@ def __init__(
...
         super().__setattr__("type", type)
-        super().__setattr__("const", False)
+        # For now, Stretch is the only kind of const variable we allow.
+        # In the future, we may want to add a 'const' constructor arg here
+        # to let users create other kinds of constants, or perhaps introduce
+        # a separate expr.Const that requires a const expr initializer for this
+        # purpose. `QuantumCircuit.add_stretch` is the official way to create
+        # stretches, and makes no promise that we will track stretches using
+        # `Var` (it accepts just a name and returns just _some_ `Expr`).
+        super().__setattr__("const", type.kind is types.Stretch)
```
jakelishman: Given what we talked about offline, how plausible is it to instead have `stretch` be represented by a new node in the `Expr` tree called `Const`?

I would like to keep the type-system association we'd newly made with the move of the `const` system to `Expr` where `Var` is associated with a mutable l-value, `Value` was an immediate value, and now `Const` is some compile-time resolvable object.  I know it might involve relatively major changes throughout the control-flow builders, but the DAG wires I think should automatically work, because we weren't adding them anyway.

I can live with it if is too much work - `Var(const=True)` and `Var(const=False)` is still distinguishable, if not as pretty as having it encoded in the types, not just the values.

qiskit/circuit/classical/types/ordering.py
```diff
@@ -68,8 +64,13 @@ def _order_uint_uint(left: Uint, right: Uint, /) -> Ordering:
...
 
 _ORDERERS = {
-    (Bool, Bool): _order_bool_bool,
+    (Bool, Bool): lambda _a, _b, /: Ordering.EQUAL,
     (Uint, Uint): _order_uint_uint,
+    (Float, Float): lambda _a, _b, /: Ordering.EQUAL,
+    (Duration, Duration): lambda _a, _b, /: Ordering.EQUAL,
+    (Duration, Stretch): lambda _a, _b, /: Ordering.LESS,
+    (Stretch, Stretch): lambda _a, _b, /: Ordering.EQUAL,
+    (Stretch, Duration): lambda _a, _b, /: Ordering.GREATER,
```
jakelishman: I'm not dead certain that this is right.  The introduction of the classes calls `Duration` "a length of time, possibly negative", and `Stretch` "some not-yet-known non-negative duration".  By those terms, let's associate `Duration` with `int` and `Stretch` with `uint`.

By comparison to integers: no `int` is ever less than any `uint` (because no `uint` can ever represent all the values of an `int`), but `int` _can_ be greater than `uint` if its width is enough.

I think what we might actually be after here is to make `Duration` the "duration expression" type, and make `Stretch` an explicit subtype of `Duration` that is non-negative.  Then the ordering should be `Stretch < Duration`, because `Stretch` is a strict subtype of it.  So that's backwards to what we have here, but that's expected - `Stretch` should be stricter than `Duration`.

(Note that I don't mean that `Stretch` inherits from `Duration` in the Python-space class system - OO inheritance is not a good model for type systems - just that the type-theory type `Stretch` is a subtype of `Duration`.)

qiskit/circuit/__init__.py
```diff
@@ -1282,6 +1282,8 @@ def __array__(self, dtype=None, copy=None):
         \end{pmatrix}
 """
 
+from qiskit._accelerate.circuit import Duration  # pylint: disable=unused-import
+
```
jakelishman, line 1286: This is de facto public interface because of its use in the public `expr.Value` field for `types.Duration`, so we it likely needs to be documented.  Is that something that's coming in a later PR?

jakelishman (commented):
Hard to review cleanly until we're more merged, but I think the couple of comments I've got here are important to get out sooner rather than later.

I'm also thinking that `circuit.Duration` ought to become a valid input type to `Delay`, but if that happens (and if you've already done it, just ignore me) it'd be a follow-up anyway.

qiskit/circuit/classical/types/ordering.py
kevinhartman: Starting out with this comment, since I think it's the most fundamental for us to make sure we get right!

> The introduction of the classes calls Duration "a length of time, possibly negative", and Stretch "some not-yet-known non-negative duration".

I want to be clear that this is _not_ something I've come up with, but rather something I pulled directly from the OpenQASM docs here:

> Stretchable durations have variable _**non-negative**_ duration that are permitted to grow as necessary to satisfy constraints.
> Source: https://openqasm.com/language/delays.html#duration-and-stretch-types

and later

> Negative durations are allowed, however passing a negative duration to a gate[duration] or box[duration] expression will result in an error.
> Source: https://openqasm.com/language/delays.html#operations-on-durations

My assumption based on that last bit is that this must be some kind of compile-time check, separate from QASM's type system, e.g. "evaluate all compile-time expressions, and if a stretch resolves to a negative duration which gets fed into a timing-aware instruction, blow up."

In the second link, we also see this example:

```qasm
duration a = 300ns;
duration b = durationof({x $0;});
stretch c;
// stretchy duration with min=300ns
stretch d = a + 2 * c;
// stretchy duration with backtracking by up to half b
stretch e = -0.5 * b + c;
```

Particularly, `stretch d = a + 2 * c;` implies that `Duration` + `Stretch` => `Stretch`, which certainly makes it _seem_ like `Stretch` is greater than `Duration` in the typing.

Stepping back from that, I think we have 3 separate things going on in all of this:

1. We have uninitialized stretch _variables_, which are considered constants.
2. We have stretch _expressions_, which can (must?) include uninitialized stretch _variables_, constrained by duration literals.
3. We need a type that represents _just_ a literal duration (e.g. `1000ns`), since there are transformations only applicable to these, like `Duration / Duration => Float`. From what I understand `Stretch / Stretch` is not defined in QASM.

On point 2., my assumption has been that `stretch s = 1000dt;` is valid and implies a stretch-i-ness of 0, hence the ordering relationship in this PR. If that's not the case, then I think `Stretch` and `Duration` are unordered. We can't put a `Stretch` into a `Duration`, since we'll lose point 3.

qiskit/circuit/__init__.py
kevinhartman: I forgot to do this. I'll add documentation in this PR.

Can I get away with putting the docstring on the `pyclass` in Rust?

qiskit/circuit/classical/expr/expr.py
kevinhartman: Honestly, I'm not sure that a `Const` node would even be a good fit for uninitialized `stretch` variables. I think we will want a `Const` node eventually to represent const definitions with initializers, and I'd expect such a node to be constructed via `expr.Const.new(name: str, type: types.Type, initializer: expr.Expr)` where `initializer` gets validated to be a constant expression coercable to `type`.

> I can live with it if is too much work - Var(const=True) and Var(const=False) is still distinguishable

I think that `Var` is not a good candidate for representing initialized constants, even with a potential `const=True` constructor argument, since all initialized constants **require** an initializer. Perhaps `Var` will take an initializer in Qiskit eventually, but it'd be optional and not required to be a constant expression.

That said, `Var` can be declared uninitialized, which IMO makes it the more natural fit for representing uninitialized `stretch` variables (between `Var` and an imagined `Const`).

What we may want eventually is a third expression kind, `StretchVar`. I didn't want to attempt to introduce that this late in the development. I think we can get away with piggybacking off of `Var` for now as an implementation detail. In #13852 I was planning to make `QuantumCircuit.add_stretch` take only a name (no var) and return just an `expr.Expr`, and then add separate methods for iterating and getting `stretch` variables + filtering them out of the normal var methods. It's unclear to me if they should still show up in the same set of captures, though.

qiskit/circuit/classical/types/ordering.py
jakelishman: Oh yeah, I wasn't thinking you were making it up - I just think it's not quite the cleanest way of representing the OQ3 spec.  I see where you're coming from, I just think there's a combination of typing restrictions and value restrictions in play here, and we're not in the right one.

For your points one and two: the statement `stretch d (= Expr)?;` introduces a new degree of freedom into the system.  If the initialisation expression is present, it's just a lower bound on the constrained value, clipped to 0 (because stretches can never be negative).  That doesn't necessarily mean that the type of `Expr` needs to be `<= stretch` - it's actually kind of the other way round; as long as you can embed `stretch` in the valid total ordering of the type of the initialisation expression, you're fine.  The statement `stretch d = a + 2*c;` might be mathematically clearer as `stretch d >= a + 2*c;`.  So for:

> Particularly, `stretch d = a + 2 * c;` implies that `Duration + Stretch => Stretch`, which certainly makes it seem like `Stretch` is greater than `Duration` in the typing.

I think that's a mistaken inference: `a + 2*c` need not be positive at all, and `stretch d = -500dt;` is still a valid initialiser and doesn't require `-500dt` to be a valid value in `stretch`.  (In my example, there's two immediate restrictions on `d`: `d >= -500dt` from the explicit restriction, and `d >= 0dt` from the typing.

> On point 2., my assumption has been that `stretch s = 1000dt;` is valid and implies a stretch-i-ness of 0,

I think this isn't right: `stretch s` introduces a new degree of freedom, and `stretch s = 1000dt;` is saying `s >= 1000dt`.  Otherwise your `a + 2*c` _also_ wouldn't have introduced a new degree of freedom - it'd be constrained entirely by `a` and `c`.


Just below your linked section, there's the line

> OpenQASM and OpenPulse have a delay instruction, whose duration is defined by a `duration`.

So in the statement `delay[<expr>]`, `<expr>` must be a valid `duration`.  There's then a value restriction on that (as opposed to a typing one) that it must resolve (after stretch resolution and constant folding, etc) to a concrete non-negative value.  That's not unprecedented in type systems: Python indexes with `int`, but negative values have a totally different meaning (you could consider them a different operation, also defined in terms of non-negative integers); and array indexing has an upper bound on the valid values too to make it a valid instruction, regardless of the type width.

I would say that a `stretch` always resolves to a valid non-negative `duration`.  That a `stretch` always resolves to a `duration` at all is already a statement that `stretch <= duration` in the typing, really.  `duration` in OQ3 is more properly a "time delta", which can also be negative.  A `delay` takes a delta as argument, but is value-constrained to be a positive delta.

---

For point 3: I get what you're saying, but I don't think you necessarily _need_ this to be in the type system, so it's a trade-off between this property or the partial ordering in terms of representable values.  There's already plenty of operations that are ill-defined for certain values of a type - `/` of course isn't defined for 0 values on the RHS for integers.  I think it's fine to admit all of:
- `a / b` is valid for `duration`
- `stretch < duration`
- and `a / b` is undefined if `b` resolves to have degrees of freedom.

I'd accept the argument that `stretch` and `duration` could be unordered because the division operation isn't defined, but if we do that, then we have to special-case all the "cast" logic, which most naturally wants to use the partial ordering of representable values.

Using the type system to avoid unresolvable operations would be nice, but it's not universally possible, and I'd call division just one of these cases.

---

Final bits / clarifying:
- it's already possible to use our type system to create unresolvable stretches - have `box { delay [a] $0; delay [a + 10dt] $1; }` and there's no valid resolution of `a`.  I'd argue we just put "division" of stretches into the same vein, and not all divisions are necessarily unresolvable: if `stretch a = 10dt;`, then `a / a` is mathematically well-defined to be 1 (because `a` can't be zero - that's the only thing that would be a problem).
- I think keeping the partial ordering to just mean "in terms of representable values is what we want".  For example, `uint[16]` usefully defined as being `< int[32]` for the purposes of safe casting, but `int[32]` supports only a subset of the operations of `uint[16]` (no bitwise stuff).
- We _could_ argue that `stretch` isn't a "type" at all, it's just an unknown `duration` (though I probably wouldn't - it's harder to track them, I think).
- I'd say that the only objects of type `stretch` are the exact names declared `stretch <id>`.  I'd call the initialiser expression (if present) a `duration`, and interpret it as a limit `duration(<stretch id>) >= <duration expr>`.

qiskit/circuit/__init__.py
jakelishman: Yeah, then a line or two in the  `qiskit.circuit` module

kevinhartman:
After chatting a bit offline, me and @jakelishman came to the conclusion that it'd likely be more appropriate to represent `stretch` as an expression node of type `Duration`, and to forego having a separate `Stretch` type altogether. More details to follow in the later PRs of this series!

qiskit/circuit/classical/types/ordering.py
jakelishman: From offline conversations: we actually decided that `stretch` needn't be a separate type from `duration` at all.  Most of the rest of this conversation stands, though.

crates/circuit/src/duration.rs
```diff
@@ -0,0 +1,60 @@
...
+///
+/// And in Python 3.9, you can use ``isinstance`` to determine which variant
+/// is populated::
+///
+///   if isinstance(duration, Duration.dt):
+///       return duration[0]
+///   elif isinstance(duration, Duration.s):
+///       return duration[0] / 5e-7
+///   else:
+///       raise ValueError("expected dt or seconds")
```
jakelishman: Minor nitpicking: I think it's easier to read if the `#[]` proc macros are immediately above the struct definition, rather than above the documentation - they're typically super important to know, and related to the definition.

jakelishman: This is a bit of a nuisance in Python 3.9.  Perhaps cleaner to add a `unit` pyfunction that returns the string name, so you can dispatch on it?  E.g.
```rust
#[pymethods]
impl Duration {
  /// The corresponding ``unit`` of the duration.
  fn unit(&self) -> &'static str {
    match Self {
      Duration::ns(_) => "ns",
      // ...
    }
  }
}
```
then Python-space has a cheaper way of doing the dispatch programmatically, and it maybe provides a slightly easier link with the legacy `(value, unit)` 2-tuple.

I'd say that ^that particular one is a candidate for doing in this PR (if you think it's worthwhile), but ones that would certainly be follow-ups if we wanted them to be: this class is a sensible place to add methods like `as_seconds(self) -> f64`, `as_dt(self, dt_in_seconds) -> u64`, etc, which would be useful if/when we swap to using this more in the internals.

qiskit/circuit/classical/expr/visitors.py
```diff
@@ -297,4 +299,7 @@ def is_lvalue(node: expr.Expr, /) -> bool:
             >>> expr.is_lvalue(expr.bit_and(a, b))
             False
     """
+    if node.const:
+        # A constant expression is _never_ an lvalue.
+        return False
     return node.accept(_IS_LVALUE)
```
jakelishman: Might be cleaner just to have the `_IS_VALUE` visitor handle this - adding it here feels like splitting the logic a little bit, and it shouldn't really save anything the short-circuit, since it'd happen immediately on the first visit anyway.

Maybe indicative that the visitor is missing some logic: should it be handling `visit_var` as `return not var.const`?

qiskit/qasm3/ast.py
```diff
@@ -173,6 +173,18 @@ class BitType(ClassicalType):
...
+
+    __slots__ = ()
+
+
+class StretchType(ClassicalType):
+    """Type information for a stretch."""
+
+    __slots__ = ()
+
+
```
jakelishman, line 187: Maybe just semantics: my reading was we give-or-take decided to interpret the OQ3 spec as saying that `stretch x;` didn't involve a _type_ `stretch` so much as just being a special keyword for a `duration`-typed declaration.  I'm also fine to leave it the way we have (heaven knows this AST is a jumble), but in another world: would it make more sense to handle `stretch x (= Expr)?;` as a
```python
class StretchDeclaration(Statement):
    identifier: Identifier
    bound: Optional[Expr]
```
rather than a type?

qiskit/qpy/__init__.py
```diff
@@ -382,31 +382,58 @@ def open(*args):
...
-``float``    ``f``      One ``double value``.
+===========================  =========  ============================================================
+Python type / class          Type code  Payload
+===========================  =========  ============================================================
+``float``                    ``f``      One ``double value``.
 
-===========  =========  ============================================================================
+:class:`~.circuit.Duration`  ``t``      One ``DURATION``.
+
+===========================  =========  ============================================================
```
jakelishman: `circuit.Duration` is still a Python type, tbf

I think there's a risk with using `t`: I think that type code is already used elsewhere to represent "tuple", and there's a risk of a clash (despite the organisational structure on `type_keys.py`, the enums in there are _not_ necessarily separate namespaces in QPY land).  That said, I think `EXPR_VALUE` is localised enough that it can never hit the clash, but `PARAM_TYPE` would.  Up to you if you think it's worth changing.

crates/circuit/src/duration.rs
```diff
@@ -0,0 +1,60 @@
...
+pub enum Duration {
+    dt(u64),
+    ns(f64),
+    us(f64),
+    ms(f64),
+    s(f64),
+}
```
jakelishman: Shall we make `dt` a `i64`, just so we're not restricting ourselves from using this same `Duration` to represent deltas in the future - I think we _might_ get away with in in current Qiskit, but we saw it come up in handling for OQ3 duration expressions, and we're already in the signed-value boat with the floating-point ones.

jakelishman (commented):
Thanks for the changes - small comments only, most we can just ignore if we don't _want_ to make the changes in this PR.

qiskit/circuit/classical/types/ordering.py
kevinhartman: Yeah, and thanks for this comment, it was quite helpful.

> the statement stretch d (= Expr)?; introduces a new degree of freedom into the system. If the initialisation expression is present, it's just a lower bound on the constrained value, clipped to 0 (because stretches can never be negative).

This is what I wasn't getting—the `=` operator in QASM is more or less overloaded for a stretch declaration. I'd assumed we weren't introducing a new degree of freedom at all. As you pointed out, really the `stretch` declaration always introduces a new degree of freedom, and optionally adds a lower bound if there's an "initializer".

All of this should be more or less handled now in the later PRs of this series.

crates/circuit/src/duration.rs
kevinhartman: I wonder if there's a lint rule or formatting transform we could enable to catch this. I agree, and I still made the mistake lol

jakelishman: I can't see one: https://rust-lang.github.io/rust-clippy/master/index.html.

I think it might be slightly tricky, since there's funniness around how the documentation might itself be generated by a called proc macro, and you can intermix `/// literal doc` and `#[doc = <something that results to a string>]`, but it would be nice if there was one.

qiskit/circuit/classical/expr/visitors.py
kevinhartman: I actually don't think we depend on this at all, anyway. I can just remove it for now. I think we decided that `Var` cannot be `const`, correct?

It could eventually be short-circuiting though, like if we ever allow durations inside arrays (I think QASM allows this).

qiskit/qasm3/ast.py
kevinhartman: Ah yeah, that makes a ton more sense. We don't use this in this PR anyway, so if it's alright, I'll leave it for now and make the switch in the PR that actually does.

qiskit/qpy/__init__.py
kevinhartman: I might not be understanding. Are you saying that we should try to have unique characters across all of `type_keys.py` in case we introduce a bug that reads/writes a byte in the wrong place? My understanding of these is that we're manually reading / writing a byte that can only be one of e.g. `TypeKeys.ExprType.<KEYS>` in the place that we're reading / writing.

crates/circuit/src/duration.rs
kevinhartman: Done in 78b2951

crates/circuit/src/duration.rs
kevinhartman: Done in 78b2951.

I also added a `py_value` method to get the int or float (exposed in Python as `Duration.value()`)

qiskit/circuit/classical/expr/visitors.py
jakelishman: If it's not used, we can take it out.  Any short-circuit would also need to be implemented by the visitor anyway, I think is my point, to handle cases where you cast away the `const` somewhere in the tree, for example - you wouldn't want the visitor to iterate all the way from that point on either, but to eagerly just return "not an l-value".  So if the logic's already going to be in the visitor, feels better not to bother duplicating it out here, since if this short-circuit triggered, then it'd just short-circuit on the first double-dispatched call anyway, so the benefit of the split logic is marginal

kevinhartman: Ah alright, I see what you're getting at. Removed it in 49f2af835254b1b51cd18166fbefa773c46f56c0

qiskit/qpy/__init__.py
jakelishman: I can't remember the exact circumstances, but I remember having trouble with QPY at one point because there were locations where type keys from more than one `TypeKeys.<Thing>` were valid choices - I think the abstraction is leaky.

That said, given that I added the `ExprType` ones after that, I'm relatively sure that it's not a problem here, so it's ok to leave.

qiskit/qpy/type_keys.py
```diff
@@ -367,20 +367,21 @@ class CircuitDuration(TypeKeyBase):
...
+            if unit == "dt":
+                return cls.DT
+            if unit == "ns":
+                return cls.NS
+            if unit == "us":
+                return cls.US
+            if unit == "ms":
+                return cls.MS
+            if unit == "s":
+                return cls.S
```
jakelishman, line 381: tbf, we _could_ have done this stuff (and elsewhere in the file) with static-dict lookups, but it's not at all important

jakelishman (approved):
Thanks, this looks good to go now!

PR #13847, opened 2025-02-14
Fix Y-eigenstate convention in `SparsePauliOp.as_paulis`

jakelishman:
`|r>` is the positive eigenstate of Y, not the negative one.  The tests also asserted this backwards.

### Details and comments

No reno because the bug isn't released yet.


jakelishman:
I see now that this duplicates some work from #13836, but since I need it in #13766 as well and both those two PRs are large, seems cleaner to split the bugfix off into a separate patch.

Cryoris (approved):
Thanks!

PR #13850, opened 2025-02-15
[Stretch] Add `+`, `-`, `*`, `/`  operators for classical expressions.

kevinhartman:
### Summary
Adds new arithmetic operators for use with numeric and a duration expressions. These are needed to express relationships between stretch variables and durations.

### Details and comments

~Based on #13844. [More readable diff here.](https://github.com/kevinhartman/qiskit/compare/expr-timing...kevinhartman:qiskit:math-expr).~

* All new operations support `Uint` and `Float`, but these cannot be intermixed. You must first cast one to the other explicitly, since `Float => Uint` and `Uint => Float` are both considered dangerous cast kinds. If two `Uint`s of different width are used, the wider one is chosen for the expression's output type.
* Addition and subtraction support for `Duration`.
* Multiplication is supported `Duration` and `Float` operands. Currently, a `Uint` must first be cast to a `Float` to be used as a timing multiplier.
  * The type system does not prevent us from representing non-const expressions of type `Duration`, but the `expr` constructors will raise a `ValueError` if a user tries to produce such an expression (this was not possible in the original design of const-types, but what we have now feels more future-proof than baking this as a hard requirement into the type system).
  * A user would also be able to create a `Var` with a `Duration` type, which would not be a constant expression. If desired, we can block users from adding such variables to a circuit.
* Division is supported between `Duration` and `Float` operands, but only where the duration is in the `left` position. It is also supported between two `Duration` types, in which case the result is a `Float`. Dividing a `Duration` by a non-const expression will raise a `ValueError`.

#### To-do

- [x] Release note.


qiskit/circuit/classical/expr/expr.py
```diff
@@ -345,6 +345,14 @@ class Op(enum.Enum):
...
         SHIFT_RIGHT = 13
         """Zero-padding bitshift to the right.  ``lhs >> rhs``."""
+        ADD = 14
+        """Addition. ``lhs + rhs``."""
+        SUB = 15
+        """Subtraction. ``lhs - rhs``."""
+        MUL = 16
+        """Multiplication. ``lhs * rhs``."""
+        DIV = 17
+        """Division. ``lhs / rhs``."""
```
jakelishman, line 364: Can you add a paragraph of prose at the end of the class-level docstring mentioning these and the allowed types?

qiskit/circuit/classical/expr/constructors.py
```diff
@@ -561,3 +568,210 @@ def index(target: typing.Any, index: typing.Any, /) -> Expr:
...
+
+def _binary_sum(op: Binary.Op, left: typing.Any, right: typing.Any) -> Expr:
+    left, right = _lift_binary_operands(left, right)
+    if (
+        left.type.kind is types.Bool
+        or right.type.kind is types.Bool
+        or left.type.kind is not right.type.kind
+        and types.order(left.type, right.type) is types.Ordering.NONE
+    ):
+        raise TypeError(f"invalid types for '{op}': '{left.type}' and '{right.type}'")
```
jakelishman: I might be overthinking this, but the "negative" type assertion makes me slightly nervous - it'd be easy to forget to update this.  What I mean is: we assert that the type _isn't_ `Bool`, as opposed to asserting that the kind _is_ in `(Uint, Float, Duration)` (well, the assertion is the negation of that, but my point is "allowlist" vs "blocklist").

I'm just thinking that the allowlist form fails safer under expansion of the type system (we might loudly [falsely] error, rather than silently [falsely] produce nonsense) - maybe worth doing it that way round?

qiskit/circuit/classical/expr/constructors.py
```diff
@@ -561,3 +568,210 @@ def index(target: typing.Any, index: typing.Any, /) -> Expr:
...
+
+            >>> from qiskit.circuit.classical import expr
+            >>> expr.sub(5.0, 2.0)
+            Binary(\
+Binary.Op.SUB, \
+Value(5.0, Float()), \
+Value(2.0, Float()), \
+Float())
+
+    Subtraction of two durations::
```
jakelishman: this is too far unindented

qiskit/circuit/classical/expr/constructors.py
```diff
@@ -561,3 +568,210 @@ def index(target: typing.Any, index: typing.Any, /) -> Expr:
...
+        """
+    return _binary_sum(Binary.Op.ADD, left, right)
+
+
+def sub(left: typing.Any, right: typing.Any, /) -> Expr:
+    """Create a subtraction expression node from the given values, resolving any implicit casts and
+    lifting the values into :class:`Value` nodes if required.
+
+    Examples:
+        Subtration of two floating point numbers::
```
jakelishman: missing a c in "subtraction"

qiskit/circuit/classical/expr/constructors.py
```diff
@@ -561,3 +568,210 @@ def index(target: typing.Any, index: typing.Any, /) -> Expr:
...
+        """
+    return _binary_sum(Binary.Op.SUB, left, right)
+
+
+def mul(left: typing.Any, right: typing.Any) -> Expr:
+    """Create a multiplication expression node from the given values, resolving any implicit casts and
+    lifting the values into :class:`Value` nodes if required.
+
+    This can be used to multiply numeric operands of the same type kind, or to multiply a duration
+    operand by a :class:`~.types.Float`.
```
jakelishman: we should be able to multiply by an integer too, I think?  That's especially important for `dt`-based ones, which are inherently integral manipulations.

qiskit/circuit/classical/expr/constructors.py
```diff
@@ -561,3 +568,210 @@ def index(target: typing.Any, index: typing.Any, /) -> Expr:
...
+
+
+def div(left: typing.Any, right: typing.Any) -> Expr:
+    """Create a division expression node from the given values, resolving any implicit casts and
+    lifting the values into :class:`Value` nodes if required.
+
+    This can be used to divide numeric operands of the same type kind, to divide a
+    :class`~.types.Duration` operand by a :class:`~.types.Float`, or to divide two
+    :class`~.types.Duration` operands which yields an expression of type
+    :class:`~.types.Float`.
```
jakelishman: we should be ok to divide `duration` by `uint` too, I think?

qiskit/qasm3/printer.py
```diff
@@ -39,8 +39,12 @@
...
     ast.Unary.Op.BIT_NOT: _BindingPower(right=22),
     #
-    # Multiplication/division/modulo: (19, 20)
-    # Addition/subtraction: (17, 18)
+    # Modulo: (19, 20)
+    ast.Binary.Op.MUL: _BindingPower(19, 20),
+    ast.Binary.Op.DIV: _BindingPower(19, 20),
+    #
+    ast.Binary.Op.ADD: _BindingPower(17, 18),
+    ast.Binary.Op.SUB: _BindingPower(17, 18),
```
jakelishman, line 47: I'm helping, lol

jakelishman (commented):
I haven't looked at the test files (need to go to dinner), but here's a super quick scan over the implementation.

qiskit/circuit/classical/expr/constructors.py
kevinhartman: That's entirely fair. I was actually just copying what you'd done previously with `Bool` :3 

But I can change this to an allow list!

qiskit/circuit/classical/expr/constructors.py
kevinhartman: That sounds fine, added. How about division of a `Duration` by a `Uint`? I know that in QASM, `Duration` / `Duration` is a `Float`. Should `Duration` / `Uint` be `Uint`, `Float`, or not permitted? Right now, it isn't allowed.

jakelishman: `duration/uint` (or `int`) should be `duration`, right?

kevinhartman: Oh, yup! (...sorry, I need more sleep, evidently!)

qiskit/circuit/classical/expr/expr.py
kevinhartman: Done in f780d9a

qiskit/circuit/classical/expr/constructors.py
kevinhartman: Done in f780d9a. I was also able to remove the order stuff, since that was a holdover from when we had `Stretch` and `Duration` types.

qiskit/circuit/classical/expr/constructors.py
kevinhartman: Multiplication by `Uint` added in e7ca9c4b6529b012c9da3ffbe19ad663131860e1.

qiskit/circuit/classical/expr/constructors.py
kevinhartman: Added in 60bcfdbd6a3238fd8cff1613f4da1db44f796303.

qiskit/circuit/classical/expr/constructors.py
```diff
@@ -564,3 +568,209 @@ def index(target: typing.Any, index: typing.Any, /) -> Expr:
...
+            >>> expr.mul(Duration.dt(1000), 0.5)
+            Binary(\
+Binary.Op.MUL, \
+Value(Duration.dt(1000), Duration()), \
+Value(0.5, Float()), \
+Duration())
+    """
+    left, right = _lift_binary_operands(left, right)
+    type: types.Type
+    if left.type.kind is right.type.kind is types.Duration:
```
jakelishman, line 679: ooh, chained `is` comparisons.  Not a Python feature you see every day haha

qiskit/circuit/classical/expr/expr.py
```diff
@@ -314,6 +314,15 @@ class Op(enum.Enum):
         container types (e.g. unsigned integers) as the left operand, and any integer type as the
         right-hand operand.  In all cases, the output bit width is the same as the input, and zeros
         fill in the "exposed" spaces.
+
+        The binary arithmetic operators :data:`ADD`, :data:`SUB:, :data:`MUL`, and :data:`DIV`
```
jakelishman, line 318:
```suggestion
        The binary arithmetic operators :data:`ADD`, :data:`SUB`, :data:`MUL`, and :data:`DIV`
```

test/python/circuit/classical/test_expr_constructors.py
```diff
@@ -610,3 +621,311 @@ def test_shift_forbidden(self, function):
...
+    def test_binary_sum_explicit(self, function, opcode):
+        cr = ClassicalRegister(8, "c")
+
+        self.assertEqual(
+            function(cr, 200),
+            expr.Binary(
+                opcode, expr.Var(cr, types.Uint(8)), expr.Value(200, types.Uint(8)), types.Uint(8)
+            ),
+        )
+        self.assertFalse(function(cr, 200).const)
```
jakelishman, line 639: Super minor, but the `expr.Binary` equality check should already be doing the `const` checks.  Could we have written the tests such that the `const` is set explicitly in the `expr.Binary` constructor?

(No need to change all the tests now.)

test/python/circuit/classical/test_expr_constructors.py
```diff
@@ -610,3 +621,311 @@ def test_shift_forbidden(self, function):
...
+            expr.mul(255.0, 1)
+        with self.assertRaisesRegex(TypeError, "cannot multiply two durations"):
+            expr.mul(Duration.dt(1000), Duration.dt(1000))
+
+        # Multiply timing expressions by non-const floats:
+        non_const_float = expr.Var.new("a", types.Float())
+        with self.assertRaisesRegex(ValueError, "would result in a non-const"):
+            expr.mul(Duration.dt(1000), non_const_float)
+        with self.assertRaisesRegex(ValueError, "would result in a non-const"):
+            expr.mul(non_const_float, Duration.dt(1000))
```
jakelishman, line 816: In a world where we have a use and a way of functioning with runtime-specified `delay`, I can imagine us relaxing this restriction, but I think it's good to assert right now.

jakelishman (approved):
Thanks for the changes!  I think everything here is all right (though I don't trust that I read every test exactly correctly haha).

qiskit/circuit/classical/expr/expr.py
jakelishman: Maybe apply this in the next PR, just so I can get this one in the merge queue already?

PR #13852, opened 2025-02-16
[Stretch] Add `stretch` variable support to `QuantumCircuit`.

kevinhartman:
### Summary
Adds support for adding new `stretch` variables to a circuit via `QuantumCircuit::add_stretch`. The returned variable follows the same scoping rules as other classical circuit variables. For now, we don't support introducing a lower bound when declaring new `stretch` variables, but this can be added in the future.

### Details and comments
~Based on #13850. See [the readable diff here](https://github.com/kevinhartman/qiskit/compare/math-expr...kevinhartman:qiskit:circuit-stretch).~

* Stretch variables are represented with a new expression node, `Stretch`. It's type is always `Duration`, and it is always a constant expression.
* Stretch variables are tracked separately from classical run-time variables.
* Stretches are not allowed as `QuantumCircuit` "input" variables.
* We can now declare stretch variables and have those be emitted to QASM, but we won't be able to do anything with them until `Delay` and `Box` start using them in their duration expressions.

#### To-do
- [x] Release note.


crates/circuit/src/dag_circuit.rs
```diff
@@ -2244,6 +2267,32 @@ impl DAGCircuit {
...
+        for (our_stretch, their_stretch) in self
+            .declared_stretches
+            .values()
+            .zip(other.declared_stretches.values())
+        {
+            if !our_stretch.bind(py).eq(their_stretch)? {
+                return Ok(false);
+            }
+        }
+
```
kevinhartman, line 2295: Looking over this myself, I'm realizing that this isn't actually right since we should be doing set equivalence. Will update.

crates/circuit/src/dag_circuit.rs
```diff
@@ -4265,6 +4314,47 @@ impl DAGCircuit {
...
+            ));
+        }
+        if let Some(previous) = self.captured_stretches.get(&var_name) {
+            if var.eq(previous)? {
+                return Err(DAGCircuitError::new_err("already present in the circuit"));
+            }
+            return Err(DAGCircuitError::new_err(
+                "cannot add var as its name shadows an existing var",
+            ));
+        }
```
jakelishman: I think the "input var" should be referred to as `stretch`, and the "existing var" would now be an "existing identifier"?  Should `add_var` have its terminology in the error messages updated too?

crates/circuit/src/dag_circuit.rs
jakelishman: is this `__eq__`?  If so: the order is important for resolution, so I think what you've written is good?

kevinhartman: Actually, my comment was a bit hasty. The order does matter for stretch declaration resolution, so it makes sense to keep this as it is.

crates/circuit/src/dag_circuit.rs
```diff
@@ -4265,6 +4314,47 @@ impl DAGCircuit {
         Ok(())
     }
 
+    /// Add a captured stretch variable to the circuit.
+    ///
+    /// Args:
+    ///     var: the stretch variable to add.
+    fn add_captured_stretch(&mut self, py: Python, var: &Bound<PyAny>) -> PyResult<()> {
```
jakelishman: There's no need to take `Python` explicitly if you've got a `Bound`. I thought we'd roughly standardised on not taking it unnecessarily, but just flicking through now, I guess we didn't.  So no need to change it.

qiskit/circuit/classical/expr/expr.py
```diff
@@ -203,6 +205,75 @@ def __deepcopy__(self, memo):
...
+
+    __slots__ = (
+        "var",
+        "name",
+    )
+
+    var: uuid.UUID
+    """A :class:`~uuid.UUID` to uniquely identify this stretch."""
+    name: str | None
+    """The name of the stretch variable."""
```
jakelishman: Two minor points:
- would `var` be better named as `uid`?  `Var` originally called its inner field `var` because the inner object was a `Clbit` or `ClassicalRegister`, then the field got co-opted to mean something more like "l-value ID".  `Stretch` isn't wrapping anything else, so we might be better jumping straight to `uid`-based naming?  (Strictly, we don't need the uid to be _universally_ unique, just unique among stretches, so we might not want to hard commit to saying it's universally unique.)
- `name` must always be set and be a `str` for `Stretch`, I think?  `name: None` was for implicit `Var` nodes that wrap `Clbit`, etc.

qiskit/circuit/classical/expr/expr.py
```diff
@@ -203,6 +205,75 @@ def __deepcopy__(self, memo):
...
+
+    def __getstate__(self):
+        return (self.var, self.name)
+
+    def __setstate__(self, state):
+        var, name = state
+        super().__setattr__("type", types.Duration())
+        super().__setattr__("const", True)
+        super().__setattr__("var", var)
+        super().__setattr__("name", name)
```
jakelishman, line 266: Potentially we should just swap to a `__getnewargs__` form, but I'm assuming that's pre-existing based on how `Var` was written?  Let's do both at once (in a different PR post 2.0), if so.

qiskit/circuit/classical/expr/visitors.py
```diff
@@ -106,6 +142,26 @@ def iter_vars(node: expr.Expr) -> typing.Iterator[expr.Var]:
...
+            from qiskit.circuit.classical import expr
+
+            cr1 = ClassicalRegister(3, "a")
+            cr2 = ClassicalRegister(3, "b")
+
+            for node in expr.iter_vars(expr.bit_and(expr.bit_not(cr1), cr2)):
+                if isinstance(node.var, ClassicalRegister):
+                    print(node.var.name)
+    """
+    yield from node.accept(_IDENT_WALKER)
```
jakelishman: Can you throw a "See also" into both this and `iter_vars` to link to each other?

qiskit/circuit/controlflow/builder.py
```diff
@@ -121,6 +121,18 @@ def add_uninitialized_var(self, var: expr.Var):
...
+    @abc.abstractmethod
+    def add_stretch(self, var: expr.Stretch):
+        """Add a stretch variable to the circuit scope.
+
+        Args:
+            var: the variable to add, if valid.
+
+        Raises:
+            CircuitError: if the variable cannot be added, such as because it invalidly shadows or
+                redefines an existing name.
```
jakelishman: "variable" in the docs should probably be "stretch"?  (Comment repeats elsewhere in this file.)

qiskit/circuit/controlflow/control_flow.py
```diff
@@ -82,3 +82,13 @@ def iter_captured_vars(self) -> typing.Iterable[expr.Var]:
...
+
+    def iter_captured_stretches(self) -> typing.Iterable[expr.Stretch]:
+        """Get an iterator over the unique captured stretch variables in all blocks of this
+        construct."""
+        seen = set()
+        for block in self.blocks:
+            for stretch in block.iter_captured_stretches():
+                if stretch not in seen:
+                    seen.add(stretch)
+                    yield stretch
```
jakelishman, line 94: Is there a use-case for `ControlFlowOp.iter_captured_identifiers`?

qiskit/circuit/quantumcircuit.py
```diff
@@ -6757,11 +7062,14 @@ def _validate_expr(circuit_scope: CircuitScopeInterface, node: expr.Expr) -> exp
     # This takes the `circuit_scope` object as an argument rather than being a circuit method and
     # inferring it because we may want to call this several times, and we almost invariably already
     # need the interface implementation for something else anyway.
-    for var in set(expr.iter_vars(node)):
-        if var.standalone:
-            circuit_scope.use_var(var)
+    for ident in set(expr.iter_identifiers(node)):
+        if isinstance(ident, expr.Stretch):
+            circuit_scope.use_stretch(ident)
```
jakelishman, line 7186: I briefly tripped up on this because we're iterating through stretches in an undefined order and calling `use` on them, where declaration order is important.  Can you expand the comment to mention that if we're not in a capturing scope (i.e. if we're in the root scope), then the `use` is a no-op?

qiskit/circuit/quantumcircuit.py
```diff
@@ -2129,6 +2155,22 @@ def num_vars(self) -> int:
         This is the length of the :meth:`iter_vars` iterable."""
         return self.num_input_vars + self.num_captured_vars + self.num_declared_vars
 
+    @property
+    def num_stretches(self) -> int:
+        """The number of stretch variables in the circuit.
```
jakelishman: I guess they're not "stretch variables" so much as "stretch degrees of freedom"?

qiskit/circuit/quantumcircuit.py
```diff
@@ -2630,6 +2728,142 @@ def has_var(self, name_or_var: str | expr.Var, /) -> bool:
...
+
+            Get a variable from a circuit by name, returning some default if it is not present::
+
+                assert qc.get_stretch("my_stretch", None) is my_stretch
+                assert qc.get_stretch("unknown_stretch", None) is None
+        """
+        if (out := self._current_scope().get_stretch(name)) is not None:
+            return out
+        if default is Ellipsis:
+            raise KeyError(f"no variable named '{name}' is present")
```
jakelishman: no stretch

qiskit/qpy/formats.py
```diff
@@ -357,6 +357,10 @@
 EXPRESSION_BINARY_PACK = "!B"
 EXPRESSION_BINARY_SIZE = struct.calcsize(EXPRESSION_BINARY_PACK)
 
+EXPRESSION_STRETCH = namedtuple("EXPRESSION_STRETCH", ["var_index"])
+EXPRESSION_STRETCH_PACK = "!H"
+EXPRESSION_STRETCH_SIZE = struct.calcsize(EXPRESSION_STRETCH_PACK)
+
```
jakelishman, line 363: Do we need some more edits to the QPY module docstring for QPY version 14 to account for this and the type keys?

releasenotes/notes/stretch-variables-076070c3f57cfa09.yaml
```diff
@@ -0,0 +1,43 @@
+---
+prelude: >
```
jakelishman, line 2: We don't usually put stuff in the prelude from non-version-defining release notes, just because it's harder to manage.  Is your intention just to have it here so it's pre-written and can be copied into the right place?

releasenotes/notes/stretch-variables-076070c3f57cfa09.yaml
```diff
@@ -0,0 +1,43 @@
...
+
+      a = qc.add_stretch("a")
+      b = qc.add_stretch("b")
+      c = qc.add_stretch("c")
+
+      # Use the stretches as Delay duration.
+      qc.delay(a, [0, 1])
+      qc.delay(b, 2)
+      qc.delay(c, [3, 4])
+      qc.barrier()
```
jakelishman, line 25: Strictly I think this PR doesn't add this bit, right?

releasenotes/notes/stretch-variables-076070c3f57cfa09.yaml
```diff
@@ -0,0 +1,43 @@
...
+
+    For additional context and examples, refer to the 
+    `QASM language specification. <https://openqasm.com/language/delays.html#duration-and-stretch-types>`__
```
jakelishman:
```suggestion
    `OpenQASM 3 language specification. <https://openqasm.com/language/delays.html#duration-and-stretch-types>`__
```

releasenotes/notes/stretch-variables-076070c3f57cfa09.yaml
```diff
@@ -0,0 +1,43 @@
...
+
+features_circuits:
+  - |
+    A new expression node :class:`~.expr.Stretch` has been added to the classical expression
+    system to represent ``stretch`` variables. To create a new ``stretch` variable, you can
+    use :meth:`.QuantumCircuit.add_stretch`. The resulting expression is a constant
+    expression of type :class:`~.types.Duration`, which can currently be used as the ``duration``
+    argument of a :meth:`~.QuantumCircuit.delay` or :meth:`~.QuantumCircuit.box` instruction.
```
jakelishman: It probably won't be able to be used by `box` in time for 2.0, so likely best drop that line.

test/python/circuit/test_circuit_operations.py
```diff
@@ -411,17 +411,22 @@ def test_copy_empty_like_circuit(self):
         copied = qc.copy_empty_like("copy")
         self.assertEqual(copied.name, "copy")
 
+    # pylint: disable=invalid-name
```
jakelishman: I think this comment was supposed to go inside the function so it scopes to only that, but I also don't care because I don't like this rule haha.

test/python/circuit/test_circuit_operations.py
```diff
@@ -432,8 +437,10 @@ def test_copy_variables(self):
         copied.add_var(d, 0xFF)
         self.assertEqual({a, b}, set(copied.iter_input_vars()))
         self.assertEqual({c, d}, set(copied.iter_declared_vars()))
+        self.assertEqual({e}, set(copied.iter_declared_stretches()))
         self.assertEqual({a}, set(qc.iter_input_vars()))
         self.assertEqual({c}, set(qc.iter_declared_vars()))
+        self.assertEqual({e}, set(qc.iter_declared_stretches()))
```
jakelishman: iirc, the spirit of this test is to `copied.add_stretch(f)` and to assert that `{e, f}` is in `copied`, while `qc` is still `{e}` (to catch the case of a shared reference to a Python object).

That said, we then seem to repeat functionally the same test another 20ish times, so probably not a big deal lol

jakelishman (commented):
Thanks for all this work - I know quite a lot of it is tedious duplication of `Var`-based stuff.

I'm not convinced I'd have spotted if there's `stretch` handling _missing_ from anywhere, but I think that's par for the course for reviews like this.

One more point I didn't spot in the review (but maybe overlooked): could you add a test that `QuantumCircuit` equality requires the stretches to be declared in the same order?

qiskit/circuit/controlflow/control_flow.py
kevinhartman: I'm actually not even sure if `iter_captured_vars` is used anywhere. I just added this for consistency.

qiskit/circuit/quantumcircuit.py
kevinhartman: I'm going through and changing "stretch variable" to "stretch" and "stretch variables" to "stretches".

The only place I'm not happy with is the `var` argument to `add_capture`, since it can now be either a `Stretch` or a `Var`. I imagine I can't rename that to `identifier` without it being a breaking change.

qiskit/circuit/controlflow/control_flow.py
jakelishman: fair enough, let's just leave it as-is

releasenotes/notes/stretch-variables-076070c3f57cfa09.yaml
kevinhartman: It certainly doesn't have to go here, I was just under the impression it was a headline feature.

releasenotes/notes/stretch-variables-076070c3f57cfa09.yaml
kevinhartman: Correct. It'll come in the next PR.

releasenotes/notes/stretch-variables-076070c3f57cfa09.yaml
jakelishman: I'm sure we will put it in the prelude yeah - it's definitely a headline - just that we usually only write the prelude in a single release note, because it's written in prose not bullets (hence the `>` YAML intro - `prelude` is a single string, unlike the others).

qiskit/circuit/quantumcircuit.py
jakelishman: I was mostly just meaning the documentation - we should leave parameters that can be keywords as-is. I think in quite a few places in those functions, we use the positional-argument-only syntax, though (`(arg, /)`), so in those cases technically the name is free game to update, but equally it's also not public so doesn't really matter.

test/python/circuit/test_circuit_vars.py
```diff
@@ -184,6 +184,28 @@ def test_add_stretch_returns_input(self):
...
         self.assertIs(a, a_other)
 
+    def test_stretch_circuit_equality(self):
+        a = expr.Stretch.new("a")
+        b = expr.Stretch.new("b")
+        c = expr.Stretch.new("c")
+
+        qc1 = QuantumCircuit(captures=[a, b, c])
+        self.assertEqual(qc1, QuantumCircuit(captures=[a, b, c]))
+        self.assertNotEqual(qc1, QuantumCircuit(captures=[c, b, a]))
```
jakelishman, line 194: Actually this particular one should compare equal in both cases - captures _are_ orderless, it's just declarations for which the order matters (because captures don't introduce a new degree of freedom).

So in this test they should both be equal, but
```python
qc1 = QuantumCircuit()
qc1.add_stretch(a)
qc1.add_stretch(b)

qc2 = QuantumCircuit()
qc2.add_stretch(b)
qc2.add_stretch(a)
```
would have `qc1 != qc2`.

That said, at this point I'm fine leaving this til a follow-up anyway, though let's note to do it.

jakelishman (dismissed):
I'm fine leaving `Stretch.var` as `var` and not changing to `uid` - if nothing else it helps duck-typing over `Stretch` and `Var`, and they are used in similar contexts sometimes.

Last comment is about the equality assertions, but it's fine - let's just leave it til a follow up.  At the moment we'll false negative some assertions, but we false negative loads of things in `QuantumCircuit` comparisons, so it's not the biggest deal ever (might matter for control-flow builders one day, so we _ought_ to fix it, but still).

jakelishman (dismissed):
Re-approving, assuming I didn't bork the merge conflict.  (Just trying to get this in the queue before I got to bed.)

test/python/circuit/test_circuit_vars.py
kevinhartman: Added to my list, I'll sort it out in #13853

jakelishman (dismissed):
Thanks for the QPY docs, and for pointing out I'd missed them haha!

test/python/circuit/test_circuit_vars.py
jakelishman: If there's no time, it can wait til post 2.0 - it's a valid bugfix.

ElePT (approved):
LGTM!

PR #13853, opened 2025-02-17
[Stretch] Support stretch and duration expressions for `Delay` instruction.

kevinhartman:
### Summary
Adds support for using classical expressions of type `Stretch` and `Duration` as the `duration` of a `Delay` instruction.

This is probably the final PR in the series 🎉. All the missing QPY stuff I'll be doing on the earlier PR branches, and will merge all of that plus fixes forward through to this.

### Details and comments
~Based on #13852. [More readable diff here](https://github.com/kevinhartman/qiskit/compare/circuit-stretch...kevinhartman:qiskit:stretchy-delay).~

#### To-do
- ~[ ] Release note.~ edit: I'd say this is already captured in previous release notes of this series
- [x] Support transpile end-to-end.
- [x] Testing.
- [x] QPY.


test/python/compiler/test_transpiler.py
```diff
@@ -1329,6 +1362,206 @@ def test_delay_converts_to_dt(self):
...
+        with self.assertRaisesRegex(TranspilerError, ".*negative duration"):
+            transpile(
+                qc,
+                backend=GenericBackendV2(num_qubits=2),
+                seed_transpiler=42,
+            )
+
+    def test_rejects_mixed_units_delay_expr_without_target_dt(self):
+        """Test that a delay instruction with an expression duration is rejected
+        when the duration resolves to a negative number."""
```
jakelishman: Completely unimportant, but this docstring doesn't match.

qiskit/transpiler/passes/scheduling/time_unit_conversion.py
```diff
@@ -135,3 +175,74 @@ def _unified(unit_set: Set[str]) -> str:
...
+
+
+class _EvalDurationImpl(expr.ExprVisitor[float]):
+    """Evaluates the expression to a single float result.
+
+    If ``dt`` is provided or all durations are already in ``dt``, the result is in ``dt``.
+    Otherwise, the result will be in seconds, and all durations MUST be in wall-time (SI).
+    """
+
+    __slots__ = ("dt", "has_dt", "has_si")
```
jakelishman, line 181: Absolutely applaud the effort here haha.  I'm not certain who's going to write an expression using only duration literals in a way that is valid through this pass via constant folding, but I love that it'll work if they do.

qiskit/transpiler/passes/scheduling/time_unit_conversion.py
```diff
@@ -135,3 +175,74 @@ def _unified(unit_set: Set[str]) -> str:
...
+
+    def __init__(self, dt=None):
+        self.dt = dt if dt is not None else 1
+        self.has_dt = False
+        self.has_si = False
```
jakelishman: Would leaving `self.dt = None` be unsafe here?  Feels like `1` is a technically valid value for `dt`.  Can't imagine that there are many QCs out there that have a cycle time of 1s, but still.

qiskit/transpiler/passes/scheduling/time_unit_conversion.py
```diff
@@ -135,3 +175,74 @@ def _unified(unit_set: Set[str]) -> str:
...
+
+    def visit_value(self, node, /) -> float:
+        if isinstance(node.value, float):
+            return node.value
+        if isinstance(node.value, Duration.dt):
+            if self.has_si and self.dt == 1:
+                raise TranspilerError(
+                    "Fail to unify time units in delays. SI units "
+                    "and dt unit must not be mixed when dt is not supplied."
+                )
```
jakelishman: Technically I think the `self.dt == 1` check is unnecessary, since we both have SI and dt - we're going to be failing either way, right, even if dt isn't set?  (Not important to change.)

qiskit/transpiler/passes/scheduling/time_unit_conversion.py
```diff
@@ -135,3 +175,74 @@ def _unified(unit_set: Set[str]) -> str:
...
+            self.has_si = True
+            if isinstance(node.value, Duration.s):
+                return node.value[0] / self.dt
+            from_unit = _DURATION_KIND_NAME.get(type(node.value))
+            return apply_prefix(node.value[0], from_unit) / self.dt
+        raise TranspilerError(f"invalid duration expression: {node}")
+
+    def visit_binary(self, node, /) -> float:
```
jakelishman, line 219: This makes me realise that we added the float type but didn't add unary negation as an op.  No need to do it in this PR, just noting that we should add it for Qiskit 2.1.

qiskit/transpiler/passes/scheduling/time_unit_conversion.py
```diff
@@ -135,3 +175,74 @@ def _unified(unit_set: Set[str]) -> str:
...
+        if isinstance(node.value, Duration):
+            if self.has_dt and self.dt == 1:
+                raise TranspilerError(
+                    "Fail to unify time units in delays. SI units "
+                    "and dt unit must not be mixed when dt is not supplied."
+                )
+            self.has_si = True
+            if isinstance(node.value, Duration.s):
+                return node.value[0] / self.dt
+            from_unit = _DURATION_KIND_NAME.get(type(node.value))
```
jakelishman: I think we added `Duration.unit` that does this now?  (Not important to change, really.)

qiskit/transpiler/passes/scheduling/time_unit_conversion.py
```diff
@@ -135,3 +175,74 @@ def _unified(unit_set: Set[str]) -> str:
...
+            return left * right
+        if node.op == expr.Binary.Op.DIV:
+            return left / right
+        raise TranspilerError(f"invalid duration expression: {node}")
+
+    def visit_cast(self, node, /) -> float:
+        return node.operand.accept(self)
+
+    def visit_index(self, node, /) -> float:
+        return node.target.accept(self)
```
jakelishman: Probably we ought to leave this undefined and/or let it error out - it can't currently appear in valid expressions of type `Duration`, and this implementation wouldn't be valid (doesn't examine the index) for most hypothetical working implementations.

qiskit/transpiler/passes/scheduling/time_unit_conversion.py
```diff
@@ -71,52 +73,90 @@ def run(self, dag: DAGCircuit):
...
+                            f"Duration is rounded to {rounded_duration:d} [dt] from {duration:f}",
+                            UserWarning,
+                        )
+                    duration = rounded_duration
+                else:
+                    has_si = True
+                if duration < 0:
+                    raise TranspilerError(
+                        f"Expression '{node.op.duration}' resolves to a negative duration!"
+                    )
```
jakelishman, line 112: I don't think we do this for legacy durations, even though they're representable.  Do we do the check somewhere else (and so should let that somewhere-else handle it), or is it a failure path we should handle for other things?

Fwiw, `TimeUnitConversion` _itself_ probably doesn't care if a duration is negative (the conversions are well defined anyway), but other places would.

qiskit/transpiler/passes/scheduling/time_unit_conversion.py
```diff
@@ -71,52 +73,90 @@ def run(self, dag: DAGCircuit):
...
+                if visitor.in_cycles():
+                    has_dt = True
+                    # We need to round in case the expression evaluated to a non-integral 'dt'.
+                    rounded_duration = round(duration)
+                    rounding_error = abs(duration - rounded_duration)
+                    if rounding_error > 1e-15:
+                        warnings.warn(
+                            f"Duration is rounded to {rounded_duration:d} [dt] from {duration:f}",
+                            UserWarning,
+                        )
```
jakelishman: Do you reckon it'd be more or less gross to call `duration_in_dt(duration, 1.0)` to get this effect?  I don't love my thing, but I also feel weird about duplicating the logic.

jakelishman (commented):
I think this is a complete review of the last PR.  I promised myself I wasn't going to comment significantly on the `TimeUnitConversion` as I think it's a bonus feature for 2.0, but then I went ahead and just did it anyway because I have no self control.

qiskit/transpiler/passes/scheduling/time_unit_conversion.py
kevinhartman: This would normally blow up in the parameter validation for `Delay`. I can delete this check if we'd rather push down the error. I just figured it might be weird to get a `CircuitError` during transpilation, but I have no real reason to think that.

Edit: I think this probably _is_ `TimeUnitConversion`'s responsibility, since it'd be the thing creating the invalid `Delay` (we assign to `Delay.duration`, which end up calling parameter validation).

qiskit/transpiler/passes/scheduling/time_unit_conversion.py
kevinhartman: Yeah, this was sort of gross for me to have done lol.

Another "divide by 1" for less code trick, which I'll change to just be local rather than at the instance level.

60e6b5cc12db9a89599f3cd3e16dc3f51291a5fb

qiskit/transpiler/passes/scheduling/time_unit_conversion.py
kevinhartman: If we have both SI and cycles, we're fine since we can convert, unless we don't have a `self.dt`.

This might just be hard to read. I changed `self.dt` to be `None` in this case, so now this reads `if self.has_si and self.dt is None:`

qiskit/transpiler/passes/scheduling/time_unit_conversion.py
kevinhartman: I actually thought to do this originally, but then I read the code inside that method wrong and thought this wouldn't work. But it does, so thanks! 😄

qiskit/transpiler/passes/scheduling/time_unit_conversion.py
kevinhartman: Done in 60e6b5cc12db9a89599f3cd3e16dc3f51291a5fb

qiskit/transpiler/passes/scheduling/time_unit_conversion.py
kevinhartman: Very important, it's much cooler and makes that `unit` method feel even more useful 😉

60e6b5cc12db9a89599f3cd3e16dc3f51291a5fb

qiskit/transpiler/passes/scheduling/time_unit_conversion.py
kevinhartman: Ah that's a good point. Created issue #13966

qiskit/transpiler/passes/scheduling/time_unit_conversion.py
jakelishman: oh of course yeah - that's the whole purpose of the pass lol

qiskit/transpiler/passes/scheduling/time_unit_conversion.py
jakelishman: Oh, if literals would blow up because `Delay` wouldn't accept it, it's fine.

kevinhartman: Yup, yeah they would.

jakelishman (dismissed):
Thanks for all this work - best as I could see here, this is ready to go!

qiskit/circuit/delay.py
```diff
@@ -107,6 +132,12 @@ def validate_parameter(self, parameter):
                     raise CircuitError("Integer duration is expected for 'dt' unit.")
                 return parameter_int
             return parameter
+        elif isinstance(parameter, expr.Expr):
+            if parameter.type.kind is not types.Duration:
+                raise CircuitError(f"Expression duration of type '{parameter.type}' is not valid.")
+            if not parameter.const:
+                raise CircuitError("Duration expressions must be constant.")
+            return parameter
         elif isinstance(parameter, ParameterExpression):
```
mtreinish, line 141: It always surprises me that we allow this. Who is going to make their delay  `acos(A) / pi + e^(pi*B)`

qiskit/circuit/quantumcircuit.py
```diff
@@ -4553,17 +4553,20 @@ def barrier(self, *qargs: QubitSpecifier, label=None) -> InstructionSet:
...
         qarg: QubitSpecifier | None = None,
-        unit: str = "dt",
+        unit: str | None = None,
     ) -> InstructionSet:
         """Apply :class:`~.circuit.Delay`. If qarg is ``None``, applies to all qubits.
         When applying to multiple qubits, delays with the same duration will be created.
 
         Args:
-            duration (int or float or ParameterExpression): duration of the delay.
+            duration (Object):
```
mtreinish, line 4566: FWIW you can just drop `(Object)` here it's all superseded by the type hinting in the docs.

qiskit/transpiler/passes/scheduling/time_unit_conversion.py
```diff
@@ -71,52 +74,83 @@ def run(self, dag: DAGCircuit):
...
+                    has_dt = True
+                    # We need to round in case the expression evaluated to a non-integral 'dt'.
+                    duration = duration_in_dt(duration, 1.0)
+                else:
+                    has_si = True
+                if duration < 0:
+                    raise TranspilerError(
+                        f"Expression '{node.op.duration}' resolves to a negative duration!"
+                    )
+                expression_durations[node._node_id] = duration
```
mtreinish, line 113: I feel like we had a potential gotcha around using node ids as indexes/keys like this that triggered a bug in the 1.2, 1.3 timeframe. It might have just been that they ids are reused after removals and aren't unique between DAGs. But nothing here would change the node index so we should be fine.

qiskit/transpiler/passes/scheduling/time_unit_conversion.py
```diff
@@ -71,52 +74,83 @@ def run(self, dag: DAGCircuit):
...
+        for node in dag.op_nodes(op=Delay):
+            if isinstance(node.op.duration, expr.Expr):
+                if any(
+                    isinstance(x, expr.Stretch) for x in expr.iter_identifiers(node.op.duration)
+                ):
+                    # If any of the delays use a stretch expression, we can't run scheduling
+                    # passes anyway, so we bail out. In theory, we _could_ still traverse
+                    # through the stretch expression and replace any Duration value nodes it may
+                    # contain with ones of the same units, but it'd be complex and probably unuseful.
+                    self.property_set["time_unit"] = "stretch"
```
mtreinish, line 98: Does anything use this property set value? If so how do they handle the string stretch here? I would expect our current scheduling passes to just skip or fail in the presence of stretches. 

Is this what raises the error when transpilation is called with a stretch duration?

test/python/compiler/test_transpiler.py
```diff
@@ -1316,6 +1317,38 @@ def test_circuit_with_delay(self, optimization_level):
...
+
+        qc = QuantumCircuit(2)
+        qc.h(0)
+        qc.delay(delay_expr, 1)
+        qc.cx(0, 1)
+
+        with self.assertWarnsRegex(
+            DeprecationWarning,
+            expected_regex="The `target` parameter should be used instead",
+        ):
```
mtreinish: This was removed in the most recent merge I think so you'll probably need to update again to account for the changes (and use a target).

mtreinish (commented):
Overall this looks great, thanks for doing this! I left a few inline comments but none of them need fixing (most are just idle musings on my part) except for the tests which will cause a conflict because main has changed on you and you can't use the deprecated arguments to transpile you were using before.

test/python/compiler/test_transpiler.py
mtreinish: Oh yep, ci is failing now on the jobs that triggered after the PR merged: https://dev.azure.com/qiskit-ci/qiskit/_build/results?buildId=65343&view=logs&j=8ec877a4-4914-5e33-8fb1-060b8cba69e2&t=414c1271-fa57-5a5a-9f2d-85b6cf71e8eb&l=18275

qiskit/transpiler/passes/scheduling/time_unit_conversion.py
kevinhartman: Yeah, the scheduling passes look at this and raise if it's `"stretch"`. I think some of the other uses of `time_unit` in the property set have gone away since I started this work, but I think we still need something like this to let the scheduling passes know they shouldn't run, unless we want to examine all of the delays again.

qiskit/transpiler/passes/scheduling/time_unit_conversion.py
kevinhartman: Yeah, I think this is actually the _only_ way to do this sort of thing, so I was very happy when I remembered that the node ID was accessible from Python. 

I know we had a "gotcha" of using the nodes directly in a map/set since they're ephemeral.

mtreinish: Ah that was it, putting the node object itself in the set was what I was thinking of

mtreinish (approved):

PR #13861, opened 2025-02-17
Remove calibrations APIs and related functionality 

eliarbel:
### Summary
This PR removes the calibration functionality from everywhere in Qiskit, outside of Pulse itself. This includes things like the `calibrations` property in `QuantumCircuit` and `DAGCircuit`, and `Target`'s support for calibrations. It also removes transitive dependencies on the calibrations functionality, e.g. within passes and the scheduler.

### Details and comments
The following user-facing parts have been removed (alongside any internal dependencies):
- Calibration-related APIs (e.g. `calibration`, `has_calibration_for` etc.) from `QuantumCircuit`, `DAGCircuit`, `Target`, `DAGDependency` and `InstructionProperties`.
- Logic related to pulse-gate handling in passes.
- The `inst_map` argument from `generate_preset_pass_manager`, `transpile`, `PassManagerConfig` and `Target.from_configuration`.
- `qiskit.scheduler` module has been removed.

~~This PR is based on changes introduced in: https://github.com/Qiskit/qiskit/pull/13814~~

Part of #13662


crates/accelerate/src/basis/basis_translator/mod.rs
```diff
@@ -351,8 +345,8 @@ fn extract_basis_target(
                 unreachable!("Control flow op is not a control flow op. But control_flow is `true`")
             };
             let bound_inst = op.instruction.bind(py);
-            // Use python side extraction instead of the Rust method `op.blocks` due to
-            // required usage of a python-space method `QuantumCircuit.has_calibration_for`.
+            // TODO: Use Rust method `op.blocks` instead of Python side extraction now that
+            // the usage of a python-space method `QuantumCircuit.has_calibration_for` is not needed anymore
```
ElePT: It's a minor suggestion but I wasn't sure from the wording if the method had been removed or not:

```suggestion
            // TODO: Use Rust method `op.blocks` instead of Python side extraction now that
            // the python-space method `QuantumCircuit.has_calibration_for`
            // has been removed and we don't need to account for it.
```

qiskit/transpiler/passes/optimization/optimize_1q_decomposition.py
```diff
@@ -160,33 +160,24 @@ def _gate_sequence_to_dag(self, best_synth_circuit):
...
         """
         if new_circ is None:
             return False
 
-        # do we even have calibrations?
-        has_cals_p = dag._calibrations_prop is not None and len(dag._calibrations_prop) > 0
-        # does this run have uncalibrated gates?
-        uncalibrated_p = not has_cals_p or any(not dag._has_calibration_for(g) for g in old_run)
-        # does this run have gates not in the image of ._decomposers _and_ uncalibrated?
+        # does this run have gates not in the image of ._decomposers?
```
ElePT, line 170: I find this comment a bit cryptic but probably not the time to pick on it either.

releasenotes/notes/remove-pulse-calibrations-4486dc101b76ec51.yaml
```diff
@@ -0,0 +1,37 @@
...
+
+    Calibration support has been removed:
+
+      * ``calibration`` has been removed from the :class:`.InstructionProperties` 's constructor and is no longer a property of that class.
+      * The ``has_calibration``, ``get_calibration``, ``instruction_schedule_map`` and ``update_from_instruction_schedule_map`` methods have been removed from the :class:`.Target` class.
+
+
+upgrade_misc:
+  - |
+    As part of Pulse removal in Qiskit 2.0, the ``sequence`` and ``schedule_circuit`` functions
```
ElePT: I think it's better to specify the module here to avoid confusions with other ``sequence`` and ``schedule`` (they are all being removed anyways):

```suggestion
    As part of Pulse removal in Qiskit 2.0, the ``sequence`` and ``schedule_circuit`` functions from :mod:`.qiskit.scheduler`
```

ElePT (dismissed):
LGTM!! Thanks for this thorough PR. I have a couple of comments but they are tiny and given the queue of removals I would merge this PR now regardless of them.

qiskit/transpiler/passes/optimization/optimize_1q_decomposition.py
eliarbel: This is the original comment for this execution path. I agree it's cryptic especially since what is `.decomposers`. Also agree we should not try to address this here and now.

releasenotes/notes/remove-pulse-calibrations-4486dc101b76ec51.yaml
eliarbel: You're right.

eliarbel:
Thanks @ElePT for the review and comments, all applied. There was a stuck job from yesterday before your review, hopefully this time it'll go through.

ElePT (approved):

PR #13862, opened 2025-02-17
The unused  parameter `dag` in `DAGNode` constructor is deprecated

### LINKED ISSUE #13022: Consider deprecating use of `dag` keyword in `DAGNode` constructors
Shall we take a note to issue a deprecation warning on the use of the `dag` keyword argument in Qiskit 1.4, and document + release note that the argument is unused as of 1.3?

_Originally posted by @jakelishman in https://github.com/Qiskit/qiskit/pull/12550#discussion_r1713997777_

The `dag` argument to `DAGNode` subclasses is irrelevant since the move of the `DAGCircuit` to Rust space, so the argument no longer has a meaning.  It's a candidate for deprecation in 1.4 and removal in 2.0.

1ucian0:
### Summary

Fixes #13022 

### Details and comments

Because `qiskit._accelerate.circuit.DAGNode` (and family) are rust implementations, I had to write a custom decorator wrapper for them.

### DIFF
--- a/crates/circuit/src/dag_node.rs
+++ b/crates/circuit/src/dag_node.rs
@@ -123,7 +123,7 @@ impl DAGOpNode {
         op: Bound<PyAny>,
         qargs: Option<TupleLikeArg>,
         cargs: Option<TupleLikeArg>,
-        #[allow(unused_variables)] dag: Option<Bound<PyAny>>,
+        dag: Option<Bound<PyAny>>,
     ) -> PyResult<Py<Self>> {
         let py_op = op.extract::<OperationFromPython>()?;
         let qargs = qargs.map_or_else(|| PyTuple::empty_bound(py), |q| q.value);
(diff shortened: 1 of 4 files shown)


mtreinish (commented):
I'm pretty sure this approach won't work. It changes the semantics around the DAG node types and will break a lot of the usage around the classes. It should be pretty easy to emit the deprecation warning from rust though, here is an example of calling python's warning from rust:

https://github.com/Qiskit/qiskit/blob/581dd3d84c9517df0011f0a5a2305102527d6c79/crates/circuit/src/dag_circuit.rs#L407-L419

We can just wrap that in an `if dag.is_some()` for the dag node constructor and that should cover it.

1ucian0:
I saw your message about raising from rust too late and I implemented in python overwriting `__init__` in https://github.com/Qiskit/qiskit/pull/13862/commits/e51cf2c4d711ba60d548f00d5f4d3549a12dabb4. Not very elegant. What do you think?

mtreinish:
> I saw your message about raising from rust too late and I implemented in python overwriting `__init__` in [e51cf2c](https://github.com/Qiskit/qiskit/commit/e51cf2c4d711ba60d548f00d5f4d3549a12dabb4). Not very elegant. What do you think?

I think we should probably add it in rust. It's a lot simpler and less error prone..

1ucian0:
done in https://github.com/Qiskit/qiskit/pull/13862/commits/b6426fcc0c0e94421c22ff030097d94a1f06ddae . Easier than expected and much more correct. Thanks for the rusty push :)

ElePT (approved):
LGTM!

PR #13863, opened 2025-02-17
The unused parameter dag in DAGNode constructor is removed

1ucian0:
### Summary

Removal of the deprecation from #13862 . Therefore, `on hold` until the deprecation is merged in 1.4

### DIFF
--- a/crates/circuit/src/dag_node.rs
+++ b/crates/circuit/src/dag_node.rs
@@ -119,13 +119,12 @@ pub struct DAGOpNode {
 #[pymethods]
 impl DAGOpNode {
     #[new]
-    #[pyo3(signature = (op, qargs=None, cargs=None, *, dag=None))]
+    #[pyo3(signature = (op, qargs=None, cargs=None, *))]
     pub fn py_new(
         py: Python,
         op: Bound<PyAny>,
         qargs: Option<TupleLikeArg>,
         cargs: Option<TupleLikeArg>,
-        #[allow(unused_variables)] dag: Option<Bound<PyAny>>,
     ) -> PyResult<Py<Self>> {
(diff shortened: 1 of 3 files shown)


crates/circuit/src/dag_node.rs
```diff
@@ -119,13 +119,12 @@ pub struct DAGOpNode {
 #[pymethods]
 impl DAGOpNode {
     #[new]
-    #[pyo3(signature = (op, qargs=None, cargs=None, *, dag=None))]
+    #[pyo3(signature = (op, qargs=None, cargs=None, *))]
```
mtreinish: You don't need the `*` anymore since there are no keyword only args:

```suggestion
    #[pyo3(signature = (op, qargs=None, cargs=None))]
```

releasenotes/notes/removal_13022-13f237b337e3d5fb.yaml
```diff
@@ -0,0 +1,4 @@
+---
+upgrade_circuits:
+  - |
+    The ``DAGNode`` subclasses take a ``dag`` optional parameter when constructed that it was unused and ignored. The parameter was deprecated in 1.4 and now it is removed.
```
mtreinish:
```suggestion
    The deprecated argument ``.dag`` for :class:`.DAGNode` and its subclasses :class:`.DAGOpNode`,
    :class:`.DAGOutNode`, and :class:`.DAGInNode`.  ``dag`` was optional parameter when constructing
    these objects that was unused and ignored since the 1.3.0 release. The parameter was deprecated
    in 1.4 and has now been is removed.
```

mtreinish (approved):

PR #13865, opened 2025-02-17
The parameter ``idle_wires`` default flipped to False in all circuit drawers

### LINKED ISSUE #12361: Default to idle_wires=False in drawers
### What should we add?

Now that we are in the utility era, it would be nice to have `idle_wires=False` by default.  I end up having to set it manually every time, and it is getting old fast.

1ucian0:
### Summary

Closes #12361 

### Details and comments

As QPUs are getting bigger, also the circuits are growing in terms of qubits.


releasenotes/notes/closes_12361-d3ea2c442a4a74a7.yaml
```diff
@@ -0,0 +1,4 @@
+---
+upgrade_visualization:
+  - |
+    The default for the parameter ``idle_wires`` changed to ``False`` in all the circuit drawers. If you still want to see wires without instructions, set ``idle_wires=True`` explicitly.
```
jakelishman: It might help to either explain a bit more, or put in a couple of text-mode drawings just to illustrate the difference?  I'm not sure users necessarily know what the option is.

jakelishman (commented):
I was going to say that this is a better default, but then I thought: actually I use the drawers quite a lot for small-scale drawing, and I think suppressing idle wires there by default is quite possibly surprising behaviour.

I wonder if it's worth having `idle_wires` take a number to use as a cut-off size: if there are `n` or more idle wires (possibly only count qubit ones), hide them, and if there's fewer, show them?  So `idle_wires=0` is the same as the current `idle_wires=False`.  We'd have to catch `idle_wires=True` and turn it into `idle_wires=math.inf`, potentially.

Alternatively, we ignore the int stuff, and add a third option to `idle_wires`: `"layout"`.  I think this might be the happiest middle-ground: if the circuit has been transpiled to a layout (so the left margin will indicate qubit mapping), _then_ we set `idle_wires=False`, and if it hasn't, we set `idle_wires=True`.  (Could also call the option `"auto"` or something.)

I'm not 100% convinced of that, I'm just slightly concerned about breaking usability of drawing small-scale circuits - I would think there's a pretty sizeable number of people who almost invariably draw those, and it'd be confusing if wires are sometimes hidden on them.

test/visual/mpl/circuit/test_circuit_matplotlib_drawer.py
```diff
@@ -1117,7 +1117,9 @@ def test_idle_wires_barrier(self):
         circuit.barrier()
 
         fname = "idle_wires_barrier.png"
-        self.circuit_drawer(circuit, output="mpl", cregbundle=False, filename=fname)
+        self.circuit_drawer(
+            circuit, output="mpl", cregbundle=False, filename=fname, idle_wires=False
```
1ucian0, line 1121: I accidentally found that a test with the description "Test that idle_wires False works with barrier" was not doing `idle_wires=False`

test/visual/mpl/circuit/references/idle_wires_barrier.png
1ucian0, line 1: see https://github.com/Qiskit/qiskit/pull/13865/files#r1977946484

qiskit/visualization/circuit/_utils.py
```diff
@@ -387,7 +387,7 @@ def _get_valid_justify_arg(justify):
 
 
 def _get_layered_instructions(
-    circuit, reverse_bits=False, justify=None, idle_wires=True, wire_order=None, wire_map=None
+    circuit, reverse_bits=False, justify=None, idle_wires="auto", wire_order=None, wire_map=None
 ):
```
jakelishman: Can we leave this as `idle_wires: bool`, and resolve the `"auto"` at the top level of the drawers?

qiskit/visualization/circuit/circuit_visualization.py
```diff
@@ -205,7 +207,7 @@ def circuit_drawer(
     # Get default from config file else use text
     default_output = "text"
     default_reverse_bits = False
-    default_idle_wires = config.get("circuit_idle_wires", False)
+    default_idle_wires = config.get("circuit_idle_wires", "auto")
```
jakelishman: Can we now resolve `"auto"` into either `True` or `False` immediately after line 227, so the rest of the visualisers don't need to change?

qiskit/circuit/quantumcircuit.py
```diff
@@ -3192,10 +3192,12 @@ def draw(
...
                 the ``text`` output, will be silently ignored otherwise.
-            idle_wires: Include idle wires (wires with no circuit elements)
-                in output visualization. Default is ``False`` unless the
+            idle_wires: Include (or not) idle wires (wires with no circuit elements)
+                in output visualization. The string ``"auto"`` is also possible, in which
+                case idle wires are show except that the circuit has a layout attached.
+                Default is ``"auto"`` unless the
                 user config file (usually ``~/.qiskit/settings.conf``) has an
-                alternative value set. For example, ``circuit_idle_wires = True``.
+                alternative value set. For example, ``circuit_idle_wires = False``.
```
jakelishman: Do we need to update the configuration-file logic to make it possible to specify `"auto"` in there too?

jakelishman (commented):
Thanks Luciano, this does feel like a good default now.

releasenotes/notes/closes_12361-d3ea2c442a4a74a7.yaml
1ucian0: good point. done in 4f83dc702caebb3cb04c909b909d2b7034afd256

qiskit/visualization/circuit/circuit_visualization.py
```diff
@@ -225,6 +225,11 @@ def circuit_drawer(
 
     if idle_wires is None:
         idle_wires = default_idle_wires
+    elif isinstance(idle_wires, str):
```
jakelishman:
```suggestion
    if isinstance(idle_wires, str):
```

still need to handle `auto` if `idle_wires` was input as `None` - line 227 would set it to `"auto"` and then it'd skip this block that resolves it

Cryoris: whoops my bad!

jakelishman (approved):
I tried this locally and it seems to do what I expect - thanks for the changes to the default, Luciano, and for getting it over the line, Julien.

PR #13868, opened 2025-02-18
Address BackendV1 deprecation warning oversights

ElePT:
### Summary
This PR takes care of some oversights in the BackendV1 deprecation.


qiskit/providers/backend_compat.py
```diff
@@ -60,9 +60,19 @@ def convert_to_target(
     Returns:
         A ``Target`` instance.
     """
-    return _convert_to_target(
-        configuration, properties, defaults, custom_name_mapping, add_delay, filter_faulty
-    )
+    # If a deprecated error is raised during the conversion, we should not return the
+    # deprecation warning to the user,as it is not actionable for them.
+    with warnings.catch_warnings():
```
mtreinish, line 65: Shouldn't this whole method be deprecated since all the input to it are deprecated?

qiskit/result/result.py
```diff
@@ -38,34 +49,112 @@ class Result:
...
 
     def __init__(
         self,
-        backend_name,
-        backend_version,
-        qobj_id,
-        job_id,
-        success,
-        results,
+        *args,
```
mtreinish, line 61: This might be confusing for people because the signature went from saying `backend_name`, `backend_version`, `qobj_id`, `job_id`, `success`, and `results` were all required arguments to no longer specifying that anywhere. The condition hasn't changed, but nothing is indicating that anymore.

qiskit/providers/backend_compat.py
ElePT: It is, here: https://github.com/Qiskit/qiskit/pull/12915. It is one of the other PRs in the queue for 1.4.

qiskit/result/result.py
ElePT: I know it's worse, but it's the only path I found to be able to inspect the args for what I needed. I can add a better description of what is requried in the docstring...

ElePT: I think it's also important to note that 2.0 will get back the named arguments, this is really just necessary to be able to raise the deprecation warning.

mtreinish: Yeah I would add it to the docstring, I think that's all we can do in this situation

qiskit/result/result.py
```diff
@@ -25,47 +26,135 @@
...
+        be used in Qiskit 2.0. It will, however, still be possible to set ``qobj_id`` as a
+        generic kwarg, which will land in the metadata field with the other generic kwargs.
+
     Attributes:
-        backend_name (str): backend name.
-        backend_version (str): backend version, in the form X.Y.Z.
-        qobj_id (str): user-generated Qobj id.
-        job_id (str): unique execution id from the backend.
-        success (bool): True if complete input qobj executed correctly. (Implies
+        backend_name (str): (REQUIRED) backend name.
```
mtreinish, line 44: I'm wondering if we want this under `Args` instead of attributes? They're both really because the constructor just does `self.backend_name = backend_name`. But right now this is under attributes.

mtreinish (commented):
One small question about the docs updates. But I think otherwise it looks good after the merge conflict is resolved.

qiskit/result/result.py
ElePT: Yes, args would help covey the message that... well, these are the args. Will do.

ElePT: done [b9d9fee](https://github.com/Qiskit/qiskit/pull/13868/commits/b9d9fee0472faf519a4b1306825e44194161050a)

mtreinish (approved):
LGTM, thanks for the updates

PR #13869, opened 2025-02-18
Add representation of `box`

### LINKED ISSUE #13772: Add `box` scope
### What should we add?

We need to implement the `box` operation, as defined in Qiskit/RFCs#76.

We will also need to add support for custom annotations, and pass-manager handling for these, in a separate issue.

This issue is tracking just for the `box` operation itself.  Related issues:
- #13767
- #13279

jakelishman:
This adds in the base `box` control-flow construction, with support for containing instructions and having a literal delay, like the `Delay` instruction.

This supports basic output to OpenQASM 3, QPY and some rudimentary support in the text and mpl drawers.  The transpiler largely handles things already, since control flow is handled generically in most places.

Known issues:

- We expect this to be able to accept stretches in its duration, just as `Delay` can, which will need a follow-up.
- We expect `Box` to support "annotations" in a future release of Qiskit.
- There is currently no way in OpenQASM 3 to represent a qubit that is idle during a `box` without inserting a magic instruction on it.
- IBM backends don't claim support for `box` yet, so `transpile` against a backend will fail, though you can modify the `Target` to add the instruction manually.

### Details and comments

~No tests yet, but I tested various things casually - this is just the WIP to show what's going on.~

Close #13772.


qiskit/circuit/controlflow/box.py
```diff
@@ -0,0 +1,153 @@
...
+
+        if body.num_qubits != self.num_qubits or body.num_clbits != self.num_clbits:
+            raise CircuitError(
+                "Attempted to assign a body parameter with a num_qubits or "
+                "num_clbits different than that of the BoxOp. "
+                f"BoxOp num_qubits/clbits: {self.num_qubits}/{self.num_clbits} "
+                f"Supplied body num_qubits/clbits: {body.num_qubits}/{body.num_clbits}."
+            )
+
+        self._params = [body]
```
mtreinish, line 85: Do we need to allow inplace mutation like this? I guess it's a required part of the data model, it just seems like playing with fire to support it given how much trouble we had with this on gates.

But this also potentially problematic when we migrate control flow to rust.

test/python/circuit/test_control_flow.py
```diff
@@ -553,6 +554,72 @@ def test_switch_rejects_input_vars(self):
...
+    def test_box_invalid_params_setter(self):
+        """Verify we catch invalid param settings for BoxOp."""
+        body = QuantumCircuit(3, 1)
+
+        bad_body = QuantumCircuit(2, 1)
+        op = BoxOp(body)
+        with self.assertRaisesRegex(CircuitError, r"different than that of the BoxOp"):
+            op.params = [bad_body]
+        with self.assertRaisesRegex(CircuitError, "body parameter of type QuantumCircuit"):
+            op.params = [None]
```
mtreinish, line 608: Do you want to have a test of the `setter` working?

mtreinish (commented):
This lgtm thanks for doing this. It's all pretty straightforward. I just had a small comment/question inline about having a setter for params. But other than that I think this is good to merge.

qiskit/circuit/controlflow/box.py
jakelishman: Yeah, unfortunately I just copied it from the other control-flow ops - we're already in trouble.

test/python/circuit/test_control_flow.py
jakelishman: I don't _want_ the setter to work, but I can add it haha.

qiskit/circuit/controlflow/box.py
mtreinish: I guess I was thinking we don't need to continue the trend here. Or raise a `NotImplementedError` if it is required. But this is a reminder that we need to add the docs about not supporting mutating `Instruction` subclasses after inserting them into a circuit. We've done it piecemeal as we moved things into rust, but we meant to document it in 2.0 as not allowed for anything.

jakelishman: I'd rather be consistent with all control-flow ops, and not add new restrictions now - I wouldn't want to _bet_ it's not used by the `Instruction` init or even some other library code.

qiskit/circuit/controlflow/box.py
```diff
@@ -0,0 +1,153 @@
...
+                expected number of qubit and clbits for the resulting instruction are inferred from
+                the number in the circuit, even if they are idle.
+            duration: an optional duration for the box as a whole.
+            unit: the unit of the ``duration``.
+            label: an optional string label for the instruction.
+        """
+        super().__init__("box", body.num_qubits, body.num_clbits, [body], label=label)
+        self.duration = duration
+        self.unit = unit
+
```
ihincks, line 59: Would you be opposed to yet another alias to the blocks? But specifically, to get the only element.

```python
@property
def content(self):
    return self.params[0]
```

qiskit/circuit/controlflow/box.py
jakelishman: In retrospect: this needs to be settable so `QuantumCircuit.assign_parameters` works (or maybe we only mutate the list in-place?  Can't remember.)

qiskit/circuit/controlflow/box.py
jakelishman: Done in 1367570070370a4978b5fcd3089a8bb393d4678b.  I called it `body`, though, if not only to match the variable terminology used in the class.

test/python/circuit/test_control_flow.py
jakelishman: Done in 1367570070370a4978b5fcd3089a8bb393d4678b.

mtreinish (approved):

Issue #13566, opened 2024-12-16
Improve synthesis of controlled circuits with controlled gates within


alexanderivrii:
### What should we add?

Consider the following example (inspired by https://arxiv.org/pdf/2412.07372):

```
inner = QuantumCircuit(5)
inner.append(MCXGate(4), [0, 1, 2, 3, 4])
controlled_inner_gate = inner.to_gate().control(2)
qc1 = QuantumCircuit(15)
qc1.append(controlled_inner_gate, [0, 1, 2, 3, 4, 5, 6])
qct1 = transpile(qc1, basis_gates=["cx", "u"])
print(qct1.count_ops())

qc2 = QuantumCircuit(15)
qc2.append(MCXGate(6), [0, 1, 2, 3, 4, 5, 6])
qct2 = transpile(qc2, basis_gates=["cx", "u"])
print(qct2.count_ops())
```

Conceptually, both ``qc1`` and ``qc2`` are exactly the same, each containing a single MCX gate with 6 control qubits. However, the output is very different: for ``qc1`` we have ``OrderedDict({'u': 1272, 'cx': 1142})``, and for ``qc2`` we have ``OrderedDict({'u': 40, 'cx': 30})``. The reason is that for ``qc2`` we directly synthesis the controlled-MCX gate over 6 qubits and internally use the synthesis algorithm that uses available idle qubits. On the other hand, for ``qc1``, we first synthesize the MCX gate over 4 qubits, obtain a decomposition, control each gate with two controls, and decompose each of these controlled gates. 

We should either improve the synthesis of controlled/annotated gates directly, or do a reduction inside the ``OptimizeAnnotated`` pass and include the latter in the transpiler flow.

PR #13870, opened 2025-02-18
[WIP] control methods produce annotated operations by default

alexanderivrii:
### Summary

`AnnotatedOperation`s allow to define a "controlled" version of a gate when a circuit is being constructed, without having to eagerly synthesize the gate in the process. The actual synthesis is left to the transpiler (and specifically to the `HighLevelSynthesis` transpiler pass). This offers at least two benefits:
* We can first define a circuit with _controlled parameterized gates_, and later bind the parameters and transpile the circuit.
* We can use additional optimizations when transpiling a circuit and in particular efficiently handle the circuit in #13566. 

The `control` method for `Gate`s and `QuantumCircuit`s already accepts the argument `annotated` which can be either `False`, `True` or `None`, with roughly the following semantics:
* If `False`, the annotated version of the gate is not constructed, with the only exception that an `annotated=False`-controlled version of an `AnnotatedOperation` is an `AnnotatedOperation` as well.
* If `True`, the annotated operation is constructed, except for gates that implement their own control method, thus an `annotated=True`-controlled version of an `XGate` will produce either a `CX-Gate`, a `CCX-Gate`, or an `MCX-Gate`. 
* If `None`, we attempt to construct an `annotated=False`-version, but if this fails, such as when the gate is parameterized, we create the `annotated=True`-version. 

In the above, `None` is the default.

In addition, for backward compatibility (at the time), when a `QuantumCircuit` is converted to a `Gate` using the `to_gate` method, the argument `annotated` is set to `False`. Unfortunately, this means that when constructing hierarchical circuits that involve controlled subcircuits, the subcircuits are typically not annotated, preventing potential transpiler optimizations. For example this is the case for the older version of circuits in the circuit library, though this is somewhat improved for the new gate classes when `control(annotated=True)` is explicitly specified. On the other hand, most of the user circuit are not constructed using `annotated=True.`

The suggestion here is to change the default behavior to construct annotated versions of a gate (unless the gate implements a custom `control` method). This is a breaking API change, and in particular if someone now writes `my_controlled_custom_gate = my_custom_gate.control(3)`, the `my_controlled_custom_gate` will no longer be an instance of `ControlledGate`. I don't if this can break a lot of user code potentially depend on this `isinstance` check. In addition, the `my_controlled_custom_gate` will not have a `definition` available, one would need to use `transpile`/`HighLevelSynthesis` to obtain one. 

I would like to discuss if we can add this breaking change to 2.0 or possibly to a later release, and what the migration path for the users should be. Comments are highly welcome.

This PR shows that the number of factual changes within Qiskit is very small: the main difference comes up in visualization, where the name of a `ControlledGate`-version of a gate is drawn with all letters lowercased, while the name of an `AnnotatedOperation`-version of a gate is drawn with first letter uppercased (compare "ghz" and "Ghz"); and also the annotated-control SWAP gate is drawn differently.


alexanderivrii:
The plan to have this ready for 3.0 is the following:

* Have #15209 ready for 2.3. In particular, this makes sure that the default value of `annotated` in all `control` methods is `None`.
* Mark-for-deprecation the argument `annotated=None` in a follow-up PR. Can we also do this for 2.3?
* Change `annotated` from `None` to `True` for 3.0. 

Some possible concerns / things to figure out:

* The `AnnotatedOperation` class has no `label` parameter but general gates do. This would lead to problems when asking for `label` from an annotated-control gate. Would a better solution be to add `label` to `AnnotatedOperation` (in the past we were reluctant to do so) or leave this as is?

* The gates representing annotated-controlled quantum circuits are visualized a bit differently (in the ControlledGate-case the name of the circuit is preserved, in the AnnotatedOperation-case, it is capitalized). Is this better to adapt the visualization for control-annotated operations or to change the tests.

Issue #13871, opened 2025-02-18
Make azimuthal angle consistent for; `RZ(phi)`, `U1(theta)`, `CRZ(theta)` and their respective docs


GuillermoAbadLopez:
1st Issue (code consistency, harder):
---
The azimuthal angle of rotation is defined inconsistently around `qiskit`:
-  `U1`, `RZ` should have the same angle name, since they are basically the same gate (it even says so in the docu), 
- same with `CRZ`, since it's just an `RZ` controlled, it's confusing to have one with `phi`, and the other with `theta`, etc..

Of course, I know this is a decision that can't be taken lightly, due to backwards compatibility, etc..

2n Issue (documentation consistency, easier)
---
But at least, what should 100% be done quickly, is make sure their documentation is consistent, for each gate, since there are cases where `lambdas` appear instead of `thetas` or `phis`... [check attached images]:

<img width="387" alt="Image" src="https://github.com/user-attachments/assets/e96cf780-0dab-4fe5-8422-7b436540b448" />
<img width="294" alt="Image" src="https://github.com/user-attachments/assets/fc1860b9-7862-4fce-85cb-7227eac8d3df" />

---

What should we add:
---

1. Address 2nd Issue, editing the docstrings and making sure all the angles are correct in the documentation (maybe I can even do this PR myself)
2. Think about 1st Issue, if it makes sense to you, and take the actions you consider 🙌

jakelishman:
Thanks for the comment!

Changing the name of the argument in the gate would risk breaking users who are passing that argument by keyword. I wouldn't expect that to be _common_, but I think changing this would be an unnecessary risk as a breaking change.  I don't like the inconsistencies, but I'd rather not break potential users, or cause them to change their code for a superficial documentation reason.

In the `RZ` documentation, we should make the mathematics match the argument name for sure - I don't know why we use $\lambda$ and `phi`, though maybe it's because `lambda` is a keyword in Python.

GuillermoAbadLopez:
> Changing the name of the argument in the gate would risk breaking users who are passing that argument by keyword. I wouldn't expect that to be common, but I think changing this would be an unnecessary risk as a breaking change. I don't like the inconsistencies, but I'd rather not break potential users, or cause them to change their code for a superficial documentation reason.

Yeah, I was expecting this as a counterpoint.. I understand it, even though it doesn't feel very nice, your call of course 👍 

[...]

PR #13873, opened 2025-02-18
Consitency of azimuthal angle name in `RZ` docstring

GuillermoAbadLopez:
- [x] I have added the tests to cover my changes. (Doesn't apply :x:)
---

### Solves the 2nd part of Issue: https://github.com/Qiskit/qiskit/issues/13871

> But at least, what should 100% be done quickly, is make sure their documentation is consistent, for each gate, since there are cases where `lambdas` appear instead of `thetas` or `phis`... [check attached images]:
> 
> Image
> Image
>

### DIFF
--- a/qiskit/circuit/library/standard_gates/rz.py
+++ b/qiskit/circuit/library/standard_gates/rz.py
@@ -37,17 +37,17 @@ class RZGate(Gate):
     .. code-block:: text
 
              ┌───────┐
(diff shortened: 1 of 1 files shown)


jakelishman (approved):
Thanks, this looks like a very clean way to handle things, and to handle the inconsistency with `U1`.  I'm happy to accept the change to `RZ` like this.

PR #13874, opened 2025-02-18
Use average gate fidelity in the commutation checker

### LINKED ISSUE #13547: Incorrect transpilation with RY gate with a small angle
### Environment

- **Qiskit version**: 1.3.0, main
- **Python version**: 3.12.8
- **Operating system**: macOS 15.1.1

### What is happening?

Transpilation with optimization level 2 and 3 results in a incorrect circuit for circuits with RY gate with a small angle.

Note that the following code does not have issue with 1.2.4.

```python
from qiskit import QuantumCircuit, generate_preset_pass_manager, __version__

qc = QuantumCircuit(2)
qc.h(0)
qc.cx(0, 1)
qc.ry(1e-6, 1)
qc.cx(0, 1)

pm = generate_preset_pass_manager(optimization_level=2)
tqc = pm.run(qc)

print(__version__)
print(qc)
print(tqc)

print("equivalent?", Operator(qc).equiv(tqc))
```

```
2.0.0.dev0+841cb29
     ┌───┐
q_0: ┤ H ├──■─────────────────■──
     └───┘┌─┴─┐┌───────────┐┌─┴─┐
q_1: ─────┤ X ├┤ Ry(1e-06) ├┤ X ├
          └───┘└───────────┘└───┘
         ┌───┐
q_0: ────┤ H ├────
     ┌───┴───┴───┐
q_1: ┤ Ry(1e-06) ├
     └───────────┘
equivalent? False
```

```
1.3.0
     ┌───┐
q_0: ┤ H ├──■─────────────────■──
     └───┘┌─┴─┐┌───────────┐┌─┴─┐
q_1: ─────┤ X ├┤ Ry(1e-06) ├┤ X ├
          └───┘└───────────┘└───┘
         ┌───┐
q_0: ────┤ H ├────
     ┌───┴───┴───┐
q_1: ┤ Ry(1e-06) ├
     └───────────┘
equivalent? False
```

```
1.2.4
     ┌───┐
q_0: ┤ H ├──■─────────────────■──
     └───┘┌─┴─┐┌───────────┐┌─┴─┐
q_1: ─────┤ X ├┤ Ry(1e-06) ├┤ X ├
          └───┘└───────────┘└───┘
     ┌───┐
q_0: ┤ H ├──■─────────────────■──
     └───┘┌─┴─┐┌───────────┐┌─┴─┐
q_1: ─────┤ X ├┤ Ry(1e-06) ├┤ X ├
          └───┘└───────────┘└───┘
equivalent? True
```

### What should happen?

The transpiled circuit should be same as the input circuit as 1.2.4 does.

### Another example

The following example fails with 1.2.4, 1.3.0, and main branch probably due to a similar reason.
```python
from qiskit import QuantumCircuit, generate_preset_pass_manager, __version__
from qiskit.quantum_info import SparsePauliOp, Operator

qc = QuantumCircuit(2)
op = SparsePauliOp(["XX"], coeffs=[1])
evo = PauliEvolutionGate(op, 1e-6)
qc.append(evo, qargs=qc.qregs[0])

pm = generate_preset_pass_manager(optimization_level=2, basis_gates=["rz", "cx", "sx", "x"])
tqc = pm.run(qc)

print(__version__)
print(qc)
print(tqc)

print("equivalent?", Operator(qc).equiv(tqc))
```

```
2.0.0.dev0+841cb29
     ┌─────────────────────┐
q_0: ┤0                    ├
     │  exp(-it XX)(1e-06) │
q_1: ┤1                    ├
     └─────────────────────┘
global phase: π/2

q_0: ─────────────────────────────────────────────────
     ┌─────────┐┌────┐┌─────────────┐┌────┐┌─────────┐
q_1: ┤ Rz(π/2) ├┤ √X ├┤ Rz(-3.1416) ├┤ √X ├┤ Rz(π/2) ├
     └─────────┘└────┘└─────────────┘└────┘└─────────┘
equivalent? False
```

```
1.2.4
     ┌─────────────────────┐
q_0: ┤0                    ├
     │  exp(-it XX)(1e-06) │
q_1: ┤1                    ├
     └─────────────────────┘
global phase: π/2

q_0: ─────────────────────────────────────────────────
     ┌─────────┐┌────┐┌─────────────┐┌────┐┌─────────┐
q_1: ┤ Rz(π/2) ├┤ √X ├┤ Rz(-3.1416) ├┤ √X ├┤ Rz(π/2) ├
     └─────────┘└────┘└─────────────┘└────┘└─────────┘
equivalent? False
```

### Any suggestions?

_No response_

Cryoris:
### Summary

The commutation checker currently uses a different metric to check when gates commute than other places in Qiskit, such as the `RemoveIdentityEquivalent`. This leads to strange behavior, where one gate is treated as identity in the commutation, but is subsequently not removed as other passes do not treat it as identity. This is what causes #13547.

This PR updates the commutation checker to use the same average gate fidelity and thus the transpiler to handle almost-identity gates more consistently.

Closes #13547.

### Details and comments

* Still needs a reno. 
* This has one small caveat: the definition of the average gate fidelity is global phase agnostic, but our commutation rules are not. I.e. we currently _do not_ want to commute two gates, if doing that introduces a global phase. We could extend this in the future, but this is outside the scope of this PR.
* This PR increases the threshold for gates to commute (== rotation angles must now be _smaller_ than before in order to commute). This means the #gates we transpile a QFT to increases, see the benchmark below. However,  @ElePT ran benchpress and this didn't seem to cause a noticable regression overall.

Here's a small benchmark comparing transpile time, CX depth and count for the (1) the QFT, (2) an efficient SU2 circuit with log(num qubits) depth and (3) a Heisenberg Trotter evolution on a line.

image


Cryoris:
This currently fails because the error tolerance for the matrix-based checks is too low. This leads e.g. to the matrix-based check of `[H, H]` to return `False`. For this to be more robust we'd have to increase the check threshold, the question is whether to just use some `const * eps` or whether to base it on the matrix dimension.

crates/accelerate/src/commutation_checker.rs
```diff
@@ -294,20 +281,20 @@ impl CommutationChecker {
...
     ) -> PyResult<bool> {
-        // relative and absolute tolerance used to (1) check whether rotation gates commute
-        // trivially (i.e. the rotation angle is so small we assume it commutes) and (2) define
-        // comparison for the matrix-based commutation checks
-        let rtol = 1e-5;
-        let atol = 1e-8;
+        // If the average gate infidelity is below this tolerance, they commute. The tolerance
+        // is set to max(1e-12, 1 - approximation_degree), to account for roundoffs and for
+        // consistency with other places in Qiskit.
+        let tol = 1e-12_f64.max(1. - approximation_degree);
```
mtreinish, line 284: This does mean we can't approximate more than what a 1e-12  tolerance provides. But it's the same logic we use in other places so I think that's fine.

crates/accelerate/src/commutation_checker.rs
```diff
@@ -689,6 +644,26 @@ fn map_rotation<'a>(
...
+    };
+
+    let trace_over_dim = match rotation {
+        "rx" | "ry" | "rz" | "rxx" | "ryy" | "rzx" | "rzz" => {
+            Complex64::new((angle / 2.).cos(), 0.)
+        }
+        "crx" | "cry" | "crz" => Complex64::new(0.5 + 0.5 * (angle / 2.).cos(), 0.),
+        "p" | "u1" => (1. + Complex64::new(0., angle).exp()) / 2.,
+        "cp" => (3. + Complex64::new(0., angle).exp()) / 4.,
+        _ => return None,
```
mtreinish: I feel like we could do this without the strings and used the `StandardGate` enum variants from standard gates. It would be faster to do the enum match because it's a u8 vs a string comparison.

crates/accelerate/src/remove_identity_equiv.rs
```diff
@@ -14,6 +14,7 @@ use num_complex::ComplexFloat;
 use pyo3::prelude::*;
 use rustworkx_core::petgraph::stable_graph::NodeIndex;
 
+use crate::commutation_checker::rotation_trace_and_dim;
```
mtreinish: Do you think we should put this somewhere more central?

crates/accelerate/src/commutation_checker.rs
mtreinish: I just saw this is used elsewhere in accelerate. I feel like this comment is especially true if we're going to use this for standardized trace computation for standard gates.

crates/accelerate/src/remove_identity_equiv.rs
```diff
@@ -33,12 +34,15 @@ fn remove_identity_equiv(
 ) {
     let mut remove_list: Vec<NodeIndex> = Vec::new();
     let mut global_phase_update: f64 = 0.;
+    // Minimum threshold to compare average gate fidelity to 1. This is chosen to account
+    // for roundoff errors and to be consistent with other places.
+    let minimum_tol = 1e-12_f64;
```
mtreinish: I'd make this a `const`:

```suggestion
    const MINIMUM_TOL: f64 = 1e-12;
```

Although you might have to move it outside the function definition to do this.

crates/accelerate/src/remove_identity_equiv.rs
```diff
@@ -83,42 +87,44 @@ fn remove_identity_equiv(
...
-                            let trace = Complex64::new((theta / 2.).cos() * 4., 0.);
-                            (4., trace)
+                    | StandardGate::CRXGate
+                    | StandardGate::CRYGate
+                    | StandardGate::CRZGate
+                    | StandardGate::CPhaseGate => {
+                        let name = gate.name();
+                        if let Param::Float(angle) = inst.params_view()[0] {
+                            let (tr_over_dim, dim) =
+                                rotation_trace_and_dim(name, angle).expect("All rotation covered");
```
mtreinish: Since this becomes a panic message maybe something a bit more descriptive/clear. We should never encounter it because the match set is complete but since you went to the trouble of adding a message instead of just `unwrap()` we should provide a bit more detail.

crates/accelerate/src/unitary_compose.rs
```diff
@@ -217,24 +197,25 @@ fn _ind(i: usize, reversed: bool) -> usize {
...
-            if (left[(i, j)] - right[(i, j)]).abs() > atol + rtol * right[(i, j)].abs() {
-                return false;
+    qargs: &[Qubit],
+) -> Array2<Complex64> {
+    let mut out = arr2(&[
+        [C_ZERO, C_ZERO, C_ZERO, C_ZERO],
+        [C_ZERO, C_ZERO, C_ZERO, C_ZERO],
+        [C_ZERO, C_ZERO, C_ZERO, C_ZERO],
+        [C_ZERO, C_ZERO, C_ZERO, C_ZERO],
+    ]);
```
mtreinish: Is there a reason to not do?:
```suggestion
    let mut out = Array2::zeros((4, 4));
```

crates/accelerate/src/unitary_compose.rs
```diff
@@ -217,24 +197,25 @@ fn _ind(i: usize, reversed: bool) -> usize {
...
 }
 
-/// For equally sized matrices, ``left`` and ``right``, check whether all entries are close
-/// by the criterion
-///
-///     |left_ij - right_ij| <= atol + rtol * right_ij
-///
-/// This is analogous to NumPy's ``allclose`` function.
-pub fn allclose(
+pub fn mm2q(
```
mtreinish: Maybe write this out as `matmul_2q` or something a bit more descriptive? It took me a sec to understand what this meant at first. Or just a docstring to say what it is.

crates/accelerate/src/unitary_compose.rs
```diff
@@ -153,58 +153,38 @@ fn _einsum_matmul_index(qubits: &[u32], num_qubits: usize) -> String {
...
-                ba += right[[i, k]] * left[[_ind(k, rev), _ind(j, rev)]];
-            }
-            let sum = ab - ba;
-            if sum.abs() > atol + ba.abs() * rtol {
-                return false;
-            }
-        }
-    }
-    true
+fn mm1q(left: &ArrayView2<Complex64>, right: &ArrayView2<Complex64>) -> Array2<Complex64> {
```
mtreinish: Maybe write this out as matmul_1q or something a bit more descriptive? It took me a sec to understand what this meant at first then I looked at the code. Or just a docstring to say what it is.

There is also an existing function for this here: https://github.com/Qiskit/qiskit/blob/main/crates/accelerate/src/euler_one_qubit_decomposer.rs#L1233-L1245 but it works with slightly different types.

qiskit/circuit/__init__.py
```diff
@@ -806,14 +806,8 @@
     q_1: ───────────┤ X ├───┤ X ├───┤ X ├     q_1: ───┤ X ├───
                     └───┘   └───┘   └───┘             └───┘
 
-Performing these optimizations are part of the transpiler, but the tools to investigate commutations
-are available in the :class:`CommutationChecker`.
-
-.. autosummary::
-   :toctree: ../stubs/
-
-   CommutationChecker
```
mtreinish: Isn't the `CommutationChecker` part of the public API we probably shouldn't remove this. (we had this exact problem as part of 1.1 or something where somebody removed it from the docs)

releasenotes/notes/cc-gate-fidelity-bde7a01dc8c56e29.yaml
```diff
@@ -0,0 +1,24 @@
+---
+upgrade:
```
mtreinish, line 2: You need a feature note about the new argument on the commutative pass constructors.

mtreinish (commented):
Overall this LGTM, it'll be good to have everything consistent. Just a few inline comments.

qiskit/circuit/__init__.py
Cryoris: huh the page doesn't exist on the 1.4 docs (which is probably why I removed it here), but it does exist on the 2.0 dev doc -- I'll add it again, thanks for the check!

crates/accelerate/src/commutation_checker.rs
Cryoris: Yeah it is unfortunately. I don't like this very much either, but currently `approximation_degree` is `1.` everywhere per default which doesn't really allow us to set another default like `1-1e-12`. The other option would've been to use `None` as indicator, but that already means that the `target` error rates should be used.

crates/accelerate/src/remove_identity_equiv.rs
Cryoris: Yeah I was looking for a more central place but didn't find an existing one that suited well.. and I didn't want to create a new one just with this file. Maybe we can just create a `gate_metrics.rs` or something that also includes the `gate_fidelity` function

crates/accelerate/src/unitary_compose.rs
Cryoris: There's zero reason not to 😉

releasenotes/notes/cc-gate-fidelity-bde7a01dc8c56e29.yaml
Cryoris: in my mind it was not API documented 😛  I'll add it!

crates/accelerate/src/gate_metrics.rs
```diff
@@ -0,0 +1,76 @@
...
+    Some((trace_over_dim, dim))
+}
+
+pub fn gate_fidelity(
+    left: &ArrayView2<Complex64>,
+    right: &ArrayView2<Complex64>,
+    qargs: Option<&[Qubit]>,
+) -> (f64, f64) {
+    let dim = left.nrows(); // == left.ncols() == right.nrows() == right.ncols()
+                            // let trace = left.t().mapv(|el| el.conj()).dot(right).diag().sum();
```
mtreinish: Are these code comments intentional?

crates/accelerate/src/unitary_compose.rs
```diff
@@ -153,33 +152,8 @@ fn _einsum_matmul_index(qubits: &[u32], num_qubits: usize) -> String {
...
-    let phase = normalized_trace.arg(); // compute phase difference
-
-    let process_fidelity = normalized_trace.abs().powi(2);
-    let gate_fidelity = (dim * process_fidelity + 1.) / (dim + 1.);
-    (gate_fidelity, phase)
-}
-
-fn mm1q(left: &ArrayView2<Complex64>, right: &ArrayView2<Complex64>) -> Array2<Complex64> {
-    let mut out = arr2(&[[C_ZERO, C_ZERO], [C_ZERO, C_ZERO]]);
+pub fn matmul_1q(left: &ArrayView2<Complex64>, right: &ArrayView2<Complex64>) -> Array2<Complex64> {
```
mtreinish, line 156:
```suggestion
#[inline]
pub fn matmul_1q(left: &ArrayView2<Complex64>, right: &ArrayView2<Complex64>) -> Array2<Complex64> {
```

crates/accelerate/src/unitary_compose.rs
```diff
@@ -197,17 +171,12 @@ fn _ind(i: usize, reversed: bool) -> usize {
     }
 }
 
-pub fn mm2q(
+pub fn matmul_2q(
```
mtreinish, line 176:
```suggestion
#[inline]
pub fn matmul_2q(
```

crates/accelerate/src/commutation_checker.rs
```diff
@@ -652,13 +605,19 @@ fn map_rotation<'a>(
 ) -> (&'a OperationRef<'a>, &'a [Param], bool) {
     let name = op.name();
 
-    if let Some((pi_multiple, generator)) = SUPPORTED_ROTATIONS.get(name) {
+    if let Some(generator) = SUPPORTED_ROTATIONS.get(name) {
```
mtreinish, line 608: Maybe not for this PR but we can drop the hashset here all together and have a static lookup table based on the standard gates. Something like:

```rust
const fn builld_lut() -> [Option<StandardGate>; STANDARD_GATE_SIZE]  {
    ...
}
let static supported_rotations: [Option<StandardGate>; STANDARD_GATE_SIZE] = build_lut()
```

Alternatively you could just do a couple of if `matches!()` on `.standard_gate()`. 

When in rust space and working solely with standard gates we really never have a need for strings, so this just sticks out to me every time I see it. But it was pre-existing so we can do this in a follow up.

mtreinish (commented):
Thanks for the quick updates. There are 3 small nits inline otherwise I think this is ready to go. I left one larger comment inline about to rewrite the rotation gate handling without using a string. But we can save that for a follow up, it's just a pet peeve of mine and is arguably scope creep for this PR.

crates/accelerate/src/unitary_compose.rs
mtreinish: Actually you can ignore these `inline` comments These are all using heap allocated view objects so there is no real benefit here probably. If this was nalgebra or fixed sized arrays it would be more important. It was just a gut reaction when seeing the `pub` and how these functions were used.

crates/accelerate/src/gate_metrics.rs
Cryoris: The first lines yes, the second no 😄 but also the first line is not really necessary, I removed them both

crates/accelerate/src/commutation_checker.rs
Cryoris: Yeah that sounds like a good idea, I'll put it on the follow-up list if that's good for you 🙂

mtreinish (approved):

PR #13805, opened 2025-02-06
Remove fake backends based on `BackendV1` and related tools

ElePT:
### Summary
This PR branches off #13793 and contains the changes related to the fake backend removal. This PR is built on top of #13722 and currently BLOCKED by it.

### Details and comments

TODOs:

- [x] remove stray prints
- [x] fix benchmarking test
- [x] fix visualization tests
- [x] reno

PR #13876, opened 2025-02-18
Remove deprecated legacy scheduling passes

mtreinish:
### Summary

This commit removes the deprecated legacy scheduling passes. These passes were replaced by the new scheduling passes a long time ago and finally marked as deprecated in 1.1.0. This commit follows through on the promise of that deprecation and removes the passes marked as deprecated.

### Details and comments


ElePT (approved):
LGTM. I wish I had remembered the incoming removal before working on #13805, but well, the conflict resolution should be easy.

PR #13882, opened 2025-02-19
Applying a slight perturbation for ill-conditioned matrices

alexanderivrii:
### Summary

Fix two failing tests for the 1.3.3 release: 
`test.python.quantum_info.operators.channel.test_kraus.TestKraus.test_circuit_init` + one other very similar test.

The problem only occurs on aarch64 (arm linux), however the relevant matrix is ill-conditioned on my mac as well: the choi matrix has many eigenvalues that are negative and close to 0 (the "negative" is likely due to numerical imprecision errors), with the `numpy.linalg.cond` returning `np.inf`. 

By doing a search on the internet, one of the recommended techniques is to apply regularization, in this case adding a small perturbation to the matrix. Update: as per @jakelishman's suggestion, we can in fact perfectly recover the eigenvalues of the original matrix from the matrix with the perturbation.

### DIFF
--- a/qiskit/quantum_info/operators/channel/transformations.py
+++ b/qiskit/quantum_info/operators/channel/transformations.py
@@ -234,13 +234,23 @@ def _choi_to_kraus(data, input_dim, output_dim, atol=ATOL_DEFAULT):
         #
         # So the eigenvalues are on the diagonal, therefore the basis-transformation matrix must be
         # a spanning set of the eigenspace.
+        #
+        # In addition, to prevent `numpy.linalg` errors when the matrix A is ill-conditioned,
+        # we apply a small perturbation, replacing A by A + eI. Since (A + eI)x = kx is
+        # equivalent to Ax = (k-e)x, it means that the eigenvectors of A + eI and A are the same,
+        # and we can perfectly recover the eigenvalues of A from the eigenvalues of A + eI by
+        # subtracting e.
+        apply_perturbation = np.linalg.cond(data) >= 1e10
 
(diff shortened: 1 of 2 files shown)


qiskit/quantum_info/operators/channel/transformations.py
```diff
@@ -234,13 +234,23 @@ def _choi_to_kraus(data, input_dim, output_dim, atol=ATOL_DEFAULT):
...
+        apply_perturbation = np.linalg.cond(data) >= 1e10
+
+        if apply_perturbation:
             data += 1e-10 * np.eye(data.shape[0])
 
         triangular, vecs = scipy.linalg.schur(data)
         values = triangular.diagonal().real
+
+        if apply_perturbation:
+            values = [v - 1e-10 for v in values]
```
jakelishman: Possibly we want to do `values = values - 1e-10` to use the Numpy vectorised subtraction, and to keep `values` as an `ndarray`?

alexanderivrii: Indeed! Done in [59a7b3c](https://github.com/Qiskit/qiskit/pull/13882/commits/59a7b3cb3e623f2f5b1bc4cd624390e39d56918f).

jakelishman (approved):
This seems fine to me now, thanks for digging in to find it.

I've retagged with the wheel test, just to make sure, but assuming that passes on AArch64, let's merge and move on.

jakelishman:
@mergifyio backport main

jakelishman:
AArch64 passed its tests, so I'll merge.

PR #13883, opened 2025-02-19
Deprecate NormalizeRXAngle and rzx_templates

eliarbel:
### Summary
`NormalizeRXAngle` and `rzx_templates` should have been deprecated as well in 1.3 together with Pulse. They are not used anymore will be removed in Qiskit 2.0. 

This is an oversight of #13164


ElePT (approved):
LGTM. I will wait until the CI fix is through to get the PR into the queue

PR #13884, opened 2025-02-19
Add 2q fractional gates to the `ConsolidateBlocks` transpiler pass

### LINKED ISSUE #13428: Optimize consecutive RZZ gates into a single RZZ gate

ShellyGarion:
### Summary
With the introduction of new 2q-fractional gates, such as `RZZGate` as part of the QPU, we add them to the `ConsolidateBlocks` transpiler pass.
https://www.ibm.com/quantum/blog/fractional-gates

This PR should close #13428

### Details and comments
Note that this PR may be eventually superseded by #13419

### DIFF
--- a/crates/accelerate/src/consolidate_blocks.rs
+++ b/crates/accelerate/src/consolidate_blocks.rs
@@ -67,7 +67,7 @@ const MAX_2Q_DEPTH: usize = 20;
 pub(crate) fn consolidate_blocks(
     py: Python,
     dag: &mut DAGCircuit,
-    decomposer: &DecomposerType,
+    decomposer: DecomposerType,
     basis_gate_name: &str,
     force_consolidate: bool,
     target: Option<&Target>,
@@ -218,10 +218,10 @@ pub(crate) fn consolidate_blocks(
             let matrix = blocks_to_matrix(py, dag, &block, block_index_map).ok();
             if let Some(matrix) = matrix {
                 let num_basis_gates = match decomposer {
-                    DecomposerType::TwoQubitBasis(decomp) => {
+                    DecomposerType::TwoQubitBasis(ref decomp) => {
                         decomp.num_basis_gates_inner(matrix.view())
                     }
-                    DecomposerType::TwoQubitControlledU(decomp) => {
+                    DecomposerType::TwoQubitControlledU(ref decomp) => {
                         decomp.num_basis_gates_inner(matrix.view())?
                     }
                 };
--- a/crates/accelerate/src/consolidate_blocks.rs
+++ b/crates/accelerate/src/consolidate_blocks.rs
@@ -17,9 +17,6 @@ use num_complex::Complex64;
 use numpy::PyReadonlyArray2;
 use pyo3::intern;
 use pyo3::prelude::*;
-use rustworkx_core::petgraph::stable_graph::NodeIndex;
-use smallvec::smallvec;
-
 use qiskit_circuit::circuit_data::CircuitData;
 use qiskit_circuit::circuit_instruction::ExtraInstructionAttributes;
 use qiskit_circuit::dag_circuit::DAGCircuit;
@@ -28,12 +25,21 @@ use qiskit_circuit::imports::{QI_OPERATOR, QUANTUM_CIRCUIT};
 use qiskit_circuit::operations::{ArrayType, Operation, Param, UnitaryGate};
 use qiskit_circuit::packed_instruction::PackedOperation;
 use qiskit_circuit::Qubit;
+use rustworkx_core::petgraph::stable_graph::NodeIndex;
+use smallvec::smallvec;
 
 use crate::convert_2q_block_matrix::{blocks_to_matrix, get_matrix_from_inst};
 use crate::euler_one_qubit_decomposer::matmul_1q;
 use crate::nlayout::PhysicalQubit;
 use crate::target_transpiler::Target;
-use crate::two_qubit_decompose::TwoQubitBasisDecomposer;
+use crate::two_qubit_decompose::{TwoQubitBasisDecomposer, TwoQubitControlledUDecomposer};
+
+#[derive(Clone, Debug)]
+#[pyclass]
+pub enum DecomposerType {
+    TwoQubitBasis(TwoQubitBasisDecomposer),
+    TwoQubitControlledU(TwoQubitControlledUDecomposer),
+}
 
 fn is_supported(
     target: Option<&Target>,
@@ -62,7 +68,7 @@ const MAX_2Q_DEPTH: usize = 20;
 pub(crate) fn consolidate_blocks(
     py: Python,
     dag: &mut DAGCircuit,
-    decomposer: &TwoQubitBasisDecomposer,
+    decomposer: &DecomposerType,
     basis_gate_name: &str,
     force_consolidate: bool,
     target: Option<&Target>,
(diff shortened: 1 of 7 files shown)


crates/accelerate/src/consolidate_blocks.rs
```diff
@@ -28,12 +25,21 @@ use qiskit_circuit::imports::{QI_OPERATOR, QUANTUM_CIRCUIT};
...
 use crate::convert_2q_block_matrix::{blocks_to_matrix, get_matrix_from_inst};
 use crate::euler_one_qubit_decomposer::matmul_1q;
 use crate::nlayout::PhysicalQubit;
 use crate::target_transpiler::Target;
-use crate::two_qubit_decompose::TwoQubitBasisDecomposer;
+use crate::two_qubit_decompose::{TwoQubitBasisDecomposer, TwoQubitControlledUDecomposer};
+
+#[derive(Clone, Debug)]
+#[pyclass]
+pub enum DecomposerType {
```
mtreinish: This doesn't actually need to be a pyclass you just need to add `FromPyObject` to the derive line.

```suggestion
#[derive(Clone, Debug, FromPyObject)]
pub enum DecomposerType {
```

That tells pyo3 how to wrap either a `TwoQubitBasisDecomposer` or `TwoQubitControlledUDecomposer` pyclass object the function receives in the enum type.

ShellyGarion:
The only test currently failing is:
https://github.com/Qiskit/qiskit/blob/682e2d31133268f8a2e6bc06a5f24455416c90cd/test/python/transpiler/test_consolidate_blocks.py#L567
This test was added in #13450 to handle issue https://github.com/Qiskit/qiskit/issues/13438, and I wonder if we should remove it.
@mtreinish @ElePT - what's your opinion?

the original circuit is:
```
        ┌────┐   
q_0: ─■─┤ √X ├─■─
      │ ├────┤ │ 
q_1: ─■─┤ √X ├─■─
        └────┘   
```
the circuit after basis translation is:
```
global phase: 7π/4
               ┌──────────┐   ┌────┐            ┌──────────┐
q_0: ─■────────┤ Rz(-π/2) ├───┤ √X ├───■────────┤ Rz(-π/2) ├
      │ZZ(π/2) ├──────────┤┌──┴────┴─┐ │ZZ(π/2) ├──────────┤
q_1: ─■────────┤ Rz(-π/2) ├┤ Rx(π/2) ├─■────────┤ Rz(-π/2) ├
               └──────────┘└─────────┘          └──────────┘
```
The circuit after optimization_level 2 or 3 transpilation is:

```
global phase: 3π/2
     ┌──────────┐┌───────┐           ┌──────────┐┌────────┐           »
q_0: ┤ Rz(-π/2) ├┤ Rx(π) ├─■─────────┤ Rx(-π/2) ├┤ Rz(-π) ├─■─────────»
     ├─────────┬┘└───────┘ │ZZ(-π/2) ├──────────┤├────────┤ │ZZ(-π/2) »
q_1: ┤ Rz(π/2) ├───────────■─────────┤ Rx(-π/2) ├┤ Rz(-π) ├─■─────────»
     └─────────┘                     └──────────┘└────────┘           »
«     ┌──────────┐
«q_0: ┤ Rz(-π/2) ├
«     ├─────────┬┘
«q_1: ┤ Rz(π/2) ├─
«     └─────────┘ 
```

The reason that this circuit is being consolidated is that it originally contains CZ gates. If it had contained only RZZ gates, then it would not had been consolidated. 

I think that it would be better handled in #13419 (which should take into account the 1-qubit gates).

test/python/transpiler/test_consolidate_blocks.py
```diff
@@ -666,6 +666,16 @@ def test_non_cx_target(self):
...
 
+    def test_collect_rzz(self):
+        """Collect blocks with RZZ gates."""
+        qc = QuantumCircuit(2)
+        qc.rzz(0.1, 0, 1)
+        qc.rzz(0.2, 0, 1)
+        consolidate_pass = ConsolidateBlocks(basis_gates=["rzz", "rx", "rz"])
+        res = consolidate_pass(qc)
+        self.assertEqual({"unitary": 1}, res.count_ops())
+        self.assertEqual(Operator.from_circuit(qc), Operator(res.data[0].operation.params[0]))
```
ShellyGarion, line 681: Note that due to the `TwoQubitControlledUDecomposer` synthesis algorithm, the synthesized circuit is:

```
q_0: ──────────■──────────────────
     ┌───────┐ │ZZ(-0.3) ┌───────┐
q_1: ┤ Rx(π) ├─■─────────┤ Rx(π) ├
     └───────┘           └───────┘
```

and not just `RZZGate(0.3)`

crates/accelerate/src/consolidate_blocks.rs
```diff
@@ -28,12 +25,21 @@ use qiskit_circuit::imports::{QI_OPERATOR, QUANTUM_CIRCUIT};
...
 
 use crate::convert_2q_block_matrix::{blocks_to_matrix, get_matrix_from_inst};
 use crate::euler_one_qubit_decomposer::matmul_1q;
 use crate::nlayout::PhysicalQubit;
 use crate::target_transpiler::Target;
-use crate::two_qubit_decompose::TwoQubitBasisDecomposer;
+use crate::two_qubit_decompose::{TwoQubitBasisDecomposer, TwoQubitControlledUDecomposer};
+
+#[allow(clippy::large_enum_variant)]
+#[derive(Clone, Debug, FromPyObject)]
```
ElePT, line 38: I don't see any public uses of `DecomposerType`, so the enum can probably be kept private, right? Similarly, I didn't find where is the `FromPyObject` trait used, but I might have missed it.

qiskit/transpiler/passes/optimization/consolidate_blocks.py
```diff
@@ -77,13 +99,14 @@ def __init__(
...
             if kak_gates:
                 self.decomposer = TwoQubitBasisDecomposer(
                     KAK_GATE_NAMES[kak_gates.pop()], basis_fidelity=approximation_degree or 1.0
                 )
-            elif "rzx" in basis_gates:
-                self.decomposer = TwoQubitBasisDecomposer(
-                    CXGate(), basis_fidelity=approximation_degree or 1.0
+            elif kak_param_gates:
+                self.decomposer = TwoQubitControlledUDecomposer(
+                    KAK_GATE_PARAM_NAMES[kak_param_gates.pop()]
```
ElePT: I remember a discussion we had in unitary synthesis where we agreed not to use `pop` because of result reproducibility. I think the same logic would apply in this case.

ElePT (commented):
Thanks Shelly! The PR looks good, I just left two small comments that should be easy to look into.

crates/accelerate/src/consolidate_blocks.rs
ShellyGarion: `FromPyObject` is used to call either the `TwoQubitBasisDecomposer` or the `TwoQubitControlledUDecomposer` from python, see this comment:
https://github.com/Qiskit/qiskit/pull/13884#discussion_r1967929870

ShellyGarion: removing `pub` from the `enum` produces this warning:
`warning: type consolidate_blocks::DecomposerType is more private than the item consolidate_blocks`

qiskit/transpiler/passes/optimization/consolidate_blocks.py
ShellyGarion: done in [9095358](https://github.com/Qiskit/qiskit/pull/13884/commits/9095358f5f771e51d4e2013dca9b9f6d947eea95)

ElePT (approved):
LGTM! thanks for answering my questions.

mtreinish:
For those worried that this bounced off the merge queue twice, looking at the log those failures were caused by issues with the latest aer release and not issues in this pr itself.

PR #13885, opened 2025-02-19
Pin symengine in backwards compatibility tests

ElePT:
### Summary
The recent symengine 1.14 release breaks the qpy backwards compatibility tests in the `stable/1.4` branch with a `qiskit.qpy.exceptions.QpyError: 'Incompatible symengine version 0.14 used to generate the QPY payload'` error.

The symengine version was already pinned to <1.14 in the following locations:

- `requirements.txt`
- `test/qpy_compat/run_tests.sh`

The symengine version is not directly modified in `test/qpy_compat/process_version.sh`, but seems to pull the version constraints from `test/qpy_compat/qpy_test_constraints.txt`, which didn't include symengine. 

For this reason, this PR pins symengine to <1.14 in `test/qpy_compat/qpy_test_constraints.txt`. 

### Details and comments
The `main` branch doesn't seem to share this issue even though the inspected files look the same. There might be an additional constraint in `main` that I am missing, but for the moment decided to open this PR directly against 1.4. [Edit: I think that the tests in main are used the cached qpy files and not generating them from scratch. The bugfix should still probably be ported to 2.0 and 1.3]

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -7,6 +7,7 @@ python-dateutil>=2.8.0
 stevedore>=3.0.0
 typing-extensions
 
-# If updating the version range here, consider updating 'test/qpy_compat/run_tests.sh' to update the
-# list of symengine dependencies used in the cross-version tests.
+# If updating the version range here, consider updating the 
+# list of symengine dependencies used in the cross-version tests
+# in 'test/qpy_compat/run_tests.sh' and 'test/qpy_compat/qpy_test_constraints.txt'
 symengine>=0.11,<0.14
--- a/test/qpy_compat/qpy_test_constraints.txt
+++ b/test/qpy_compat/qpy_test_constraints.txt
@@ -1,2 +1,3 @@
 numpy===1.24.4
 scipy===1.10.1
+symengine>=0.11,<0.14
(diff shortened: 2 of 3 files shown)


mtreinish (approved):
This was definitely an oversight when we added the fixes for the symengine version dependency in 1.2.4. We fixed compatibility between symengine 0.13.0 and 0.11.0 and added the version cap to prevent this on new releases, and tested the range of supported symengine versions to handle the interop question. But forgot about future releases when installing historical qiskit releases. This should correct that oversight. I think your comment about it being related to the cache makes sense. I'm not sure why we're not hitting this on main though, because there are open PRs that should be invalidating the cache, but that doesn't change this PR as it is clearly correct.

ElePT:
@Mergifyio backport stable/1.3 main

jakelishman:
> I'm not sure why we're not hitting this on main though, because there are open PRs that should be invalidating the cache, but that doesn't change this PR as it is clearly correct.

For future posterity: the reason the existing PRs that invalidated the cache on `main` haven't yet failed is because each PR has its own cache context, which starts of as inheriting from the base branch.  So new PRs against `main` inherit its cache, unless they invalidate it.  Once they _have_ invalidated it, they have their own cache, so subsequent non-invalidating pushes on a PR won't cause a new QPY rebuild until the cache is invalidated again.  Each open PR that invalidates the QPY cache did so before the release of symengine 0.14, and hasn't broken it since, which is why it hasn't appeared yet.

PR #13888, opened 2025-02-19
Fix arithmetic bug in estimate_duration method

mtreinish:
### Summary

In the recently merged #13783 a bug slipped in an unexercised code path where we were dividing by dt instead of multiplying. This was fixed in the common code path where dt is an int, but nothing in the data model disallows float durations for dt units despite it not being real. Since the data model allowed it the rust code needed to support it and that path did the wrong operation. This commit fixes that oversight in the rust code and adds a test for it.

### Details and comments


ElePT (approved):
Whoops

PR #13891, opened 2025-02-19
Add `ContractIdleWiresInControlFlow` to default pipelines

jakelishman:
### Summary

This is cheap, and can be done both in the `init` stage (where it may help routing) and the the general optimisation loop (where it may allow chains of optimisations to collapse).

### Details and comments

Built on top of #13779.


kevinhartman (commented):
Looks good, I'll sign off once what we're based on here merges.

jakelishman:
Rebased now.

kevinhartman (approved):

PR #13892, opened 2025-02-19
Simplify return type of `DAGCircuit::control_flow_op_nodes`

jakelishman:
Technically this is a breaking change, not because of the type change (we're returning a subset of what we previously did), but because the docstring previously _explicitly_ said that `None` was used for the empty list.  In practice, this is a bit of a footgun (as I found), and inconsistent with the other `DAGCircuit` methods.  It's not clear that it has a performance advantage; technically it can save a single Python empty-list allocation (`None` is a singleton), but this cost is pretty trivial.

### Details and comments

I'll not be distraught if we choose not to do this because it's technically a breaking change, but I can say from experience that I managed to write a buggy transpiler pass because of the unexpected behaviour here, and `if not op_nodes` is naturally supported on both 1.x and 2.x.

### DIFF
--- a/crates/circuit/src/dag_circuit.rs
+++ b/crates/circuit/src/dag_circuit.rs
@@ -3628,28 +3628,24 @@ def _format(operand):
     /// Get a list of "op" nodes in the dag that contain control flow instructions.
     ///
     /// Returns:
-    ///     list[DAGOpNode] | None: The list of dag nodes containing control flow ops. If there
-    ///         are no control flow nodes None is returned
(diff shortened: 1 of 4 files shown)


kevinhartman (dismissed):
I'm good with this change. Even though it's technically "breaking", it seems like something we can do reasonably in 2.0.

jakelishman:
Just rebased over #13506, since we touched similar bits around `propagate_condition`.

kevinhartman (approved):

Issue #13699, opened 2025-01-21
Remove in 2.0 all the deprecated code in 1.*


1ucian0:
All the deprecations in 1.* (including not-yet-released 1.4) need to be removed in 2.0. This task is for tracking the big-picture.

| title | issue/epic/PR  | status |
| ---- | -------------- | ------ |
| Pulse | #13662  | not-started  |
| Provider ABC | #13435 | wip |
| BackendV1 | #13708 | not-started  |
| `.condition`/`c_if`| #13506 | under-review |
| `.duration` and `.unit` | #13506 | under-review |
| qobj | #13701  | not-started  |
| primitivesV1 |  | not-started  |
| pending deprecations | #13589 | not-started  |
| stochastic swap| #13791 | under-review |

**Statuses:**
 * not-started
 * assigned/draft/wip
 * under-review
 * at-risk
 * done

Other deprecations:

**1.1**:
 - `qiskit/transpiler/passes/scheduling/alap.py:29` (`__init__`)
 - `qiskit/transpiler/passes/scheduling/asap.py:35` (`__init__`)
 - `qiskit/transpiler/passes/scheduling/dynamical_decoupling.py:109` (`__init__`)
 - `qiskit/transpiler/passes/scheduling/alignments/align_measures.py:93` (`__init__`)
 - `qiskit/synthesis/two_qubit/two_qubit_decompose.py:194` (`specialize`)

---

**1.2**:
 - `qiskit/visualization/transition_visualization.py:125` (`visualize_transition`)
 - `qiskit/circuit/quantumcircuit.py:1468` (`cls_instances`)
 - `qiskit/circuit/quantumcircuit.py:1486` (`cls_prefix`)
 - `qiskit/circuit/quantumcircuit.py:2358` (`cast`)
 - `qiskit/circuit/quantumcircuit.py:2379` (`qbit_argument_conversion`) (remove the public API, not the code)
 - `qiskit/circuit/quantumcircuit.py:2414` (`cbit_argument_conversion`) (remove the public API, not the code)

---

**1.3**:
 - `qiskit/transpiler/passes/routing/stochastic_swap.py:63` (`__init__`)
 - `qiskit/transpiler/preset_passmanagers/generate_preset_pass_manager.py:47` (`generate_preset_pass_manager`)
 - `qiskit/transpiler/preset_passmanagers/generate_preset_pass_manager.py:56` (`generate_preset_pass_manager`)
[...]

PR #13893, opened 2025-02-19
Remove accidentally-public methods from `QuantumCircuit`

jakelishman:
These were never really meant to be public at all, and we didn't even notice they existed publicly til we rewrote the `QuantumCircuit` documentation.

### Details and comments

Part of #13699.


alexanderivrii (approved):
Thanks, this looks pretty straightforward. I guess the other two ``QuantumCircuit`` functions ``unit`` and ``duration`` that should be removed for 2.0 would be removed as a part of a different PR.

PR #13894, opened 2025-02-19
Remove deprecated QuantumCircuit.duration and DAGCircuit.duration

### LINKED ISSUE #13849: Remove deprecated `QuantumCircuit.duration` and `DAGCircuit.duration`
### What should we add?

The duration attribute for `QuantumCircuit` and `DAGCircuit` were deprecated in Qiskit 1.3. We should remember to remove them for Qiskit 2.0. We should only do this after: https://github.com/Qiskit/qiskit/pull/13783 merges though.

mtreinish:
### Summary

This commit removes the deprecated duration attributes for circuits.
This was deprecated in the 1.3.0 release and replaced by the
QuantumCircuit.estimate_duration() method in 1.4.0.

### Details and comments

This commit is based on top of #13506 and is blocked on #13708 and #13662
(as many tests using this functionality are of these deprecated
components)

### DIFF
--- a/crates/circuit/src/dag_circuit.rs
+++ b/crates/circuit/src/dag_circuit.rs
@@ -205,12 +205,6 @@ pub struct DAGCircuit {
     vars: BitData<Var>,
     /// Global phase.
     global_phase: Param,
-    /// Duration.
-    #[pyo3(set)]
-    duration: Option<PyObject>,
-    /// Unit of duration.
-    #[pyo3(set)]
-    unit: String,
 
     // Note: these are tracked separately from `qubits` and `clbits`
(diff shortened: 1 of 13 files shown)


mtreinish:
I'm thinking we should defer this to 3.0 we're very close to the release and this removal is not critical for 2.0 it was just a nice to have. But since we're still waiting on https://github.com/Qiskit/qiskit/issues/13708 and https://github.com/Qiskit/qiskit/issues/13662 before we can finalize this I think it's probably prudent to just wait. Especially given the whole issue with `estimate_duration` being forgotten about in 1.3.0. We can revisit this in 3.0, I'll push up a small PR to adjust the deprecation message to say 3.0 instead of 2.0.

PR #10977, opened 2023-10-05
Track runtime variables in control-flow builders

jakelishman:
### Summary

This adds support for all the new classical runtime variables through
the control-flow builder interface.  In particular, most usefully it
automatically manages the scoping rules for new declarations and inner
variable accesses, and ensures that its built scopes automatically close
over any variables used within them (including making sure nested scopes
do the same thing).

The builder API is factored out a little into an explicit interface
object, with `QuantumCircuit` getting an explicit implementation of
that.  This is done because the number of separate API methods we would
have needed to pass around / infer was getting overly large, and this
allows us to just use standard virtual dispatch to automatically do the
right thing.

Python doesn't have a way to have an object implement an
interface other than by structural (duck) typing, so to avoid name
leakage and collisions, we instead make `QuantumCircuit`'s
implementation a friend class that handles the inner state on its
behalf.

Not everything control-flow-builder related is factored out into the API
because it didn't seem overly useful to do this, especially when the
overridden behaviour would just have been to throw exceptions.

### Details and comments

Depends on #10963 and #10974

Close #10938

PR #13899, opened 2025-02-20
Support stretchy delays in `box`

jakelishman:
### Summary
Mostly this just needed to re-use the existing handling from within `Delay`, and to teach the control-flow builders to look in another special place (`BoxOp.duration`) for potential `Expr` instances to close over.

<details><summary>Old PR message</summary>
This is currently built on top of a roll-up merge of Kevin's `stretchy-delay` branch, and just unifies the handling between `Delay` and `Box` so that they both support the `Expr` stuff.  There's no need for Rust-space handling for `box`, because the entirety of the control-flow system isn't handled in Python-space yet.

This might need further modification based on Kevin's work on transpiler support.

This commit will be rebased and amended along with the rest of the `stretch` patch series.
</details>

### Details and comments

Built on top of both #13869 and #13853.

### DIFF
--- a/qiskit/circuit/quantumcircuit.py
+++ b/qiskit/circuit/quantumcircuit.py
@@ -2724,6 +2724,8 @@ class QuantumCircuit:
             if copy and is_parameter:
                 operation = _copy.deepcopy(operation)
         if isinstance(operation, ControlFlowOp):
+            if operation.name == "box" and operation.unit == "expr":
+                _validate_expr(circuit_scope, operation.duration)
             # Verify that any variable bindings are valid.  Control-flow ops are already enforced
             # by the class not to contain 'input' variables.
             if bad_captures := {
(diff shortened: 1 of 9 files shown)


1ucian0:
`on hold` until https://github.com/Qiskit/qiskit/pull/13869 and https://github.com/Qiskit/qiskit/pull/13853.

kevinhartman:
I'm sure you're probably quite busy at the moment, but depending on when 2.0 is going out, I can review this tomorrow if you want to rebase it.

qiskit/circuit/quantumcircuit.py
```diff
@@ -2724,6 +2724,8 @@ def append(
             if copy and is_parameter:
```
kevinhartman, line 2724: Above this line (beyond the scope of your changes), I see that we assign to `param`. My eyes might be playing tricks on me, but is there a point to that?

kevinhartman (approved):
Simple enough—looks great. Thanks for enabling this!

Feel free to merge.

qiskit/circuit/quantumcircuit.py
jakelishman: I don't know why we bother with the assignment in the lines above - the commit that introduced the assignment (#10977) doesn't say why - before, we didn't have the assignment -, and we drop the variable immediately after.  I suspect `_validate_expr` returns the input just for chaining convenience - there's at least one place that does `node = _validate_expr(expr.lift(node))`, so that's probably the base.

Either that, or some earlier version of the code (as in, during local development) had more of a call for it.

PR #13903, opened 2025-02-21
Bump to ndarray 0.16 and vendor/fork ndarray_einsum_beta

mtreinish:
### Summary

This commit bumps the ndarray version to ndarray 0.16 which was released some time ago. We have been blocked from upgrading this because of our use of ndarray_einsum_beta, which is functionally inactive at this point with no development activity in > 4 years. To facilitate the upgrade this commit forks ndarray einsum and vendors it as an internal crate in Qiskit. It's not a very large library only ~1k lines and this is the fastest path to getting us on a new version of ndarray. If it's preferred I can create an fork as a separate repository outside the qiskit workspace and update the dependency that way instead.

### Details and comments

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -43,15 +43,6 @@ checksum = "683d7910e743518b0e34f1186f92494becacb047c7b6bf616c96772180fef923"
 
 [[package]]
 name = "approx"
-version = "0.4.0"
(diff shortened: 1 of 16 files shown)


mtreinish:
Actually looking at the code in this PR, I'm going to refactor this to be a separate fork repo. For some reason I thought it was smaller than it actually is.

PR #13904, opened 2025-02-21
Fix release note config for 1.4

Eric-Arellano:
Otherwise, 1.3 shows up in /api/qiskit/release-notes/1.4.

### DIFF
--- a/docs/release_notes.rst
+++ b/docs/release_notes.rst
@@ -10,4 +10,4 @@ Qiskit |version| release notes
     `:earliest-version:` should be set to the rc1 release for the current minor release series. For example, the stable/1.1 branch should set it to 1.1.0rc1. If on `main`, set to the prior minor version's rc1, like `1.0.0rc1`.
 
 .. release-notes::
-   :earliest-version: 1.3.0b1
+   :earliest-version: 1.4.0


docs/release_notes.rst
```diff
@@ -10,4 +10,4 @@ Qiskit |version| release notes
     `:earliest-version:` should be set to the rc1 release for the current minor release series. For example, the stable/1.1 branch should set it to 1.1.0rc1. If on `main`, set to the prior minor version's rc1, like `1.0.0rc1`.
 
 .. release-notes::
-   :earliest-version: 1.3.0b1
+   :earliest-version: 1.4.0
```
Eric-Arellano, line 13: According to [PyPI](https://pypi.org/project/qiskit/#history), there was no rc, so 1.4.0 is the earliest version in the minor series.

ElePT (approved):
Thanks @Eric-Arellano! Not having rc1 didn't allow me to make that change at the same time as the release preparation PR, the reno build broke because of the lack of 1.4 tag. I was planning to open a follow-up but you beat me to it (ok, I forgot to do it this morning). It should work now that the tag is in place.

jakelishman:
Do we need to update `main` too?  I forgot the process for the release notes.

Eric-Arellano:
```rst
 `:earliest-version:` should be set to the rc1 release for the current minor release series.
For example, the stable/1.1 branch should set it to 1.1.0rc1. If on `main`,
set to the prior minor version's rc1, like `1.0.0rc1`.
```

Technically, yes. But I don't think it actually matters because we currently ignore the release notes for `/api/qiskit/dev`. I proposed in https://github.com/Qiskit/documentation/issues/1296 that we may want to change that, but we won't have bandwidth to get to it soon.

jakelishman:
@mergifyio backport main

PR #13909, opened 2025-02-21
Bump to ndarray 0.16

mtreinish:
### Summary

This commit bumps the ndarray version to ndarray 0.16 which was released some time ago. We have been blocked from upgrading this because of our use of ndarray_einsum_beta, which is functionally inactive at this point with no development activity in > 4 years. After waiting a month after pushing a PR to update the dependency I created a fork of the library to facilitate the ndarray upgrade here. This commit switches to using the latest version of ndarray and the new crate from the fork I created.

### Details and comments


jakelishman (approved):
Glad to see the bump to the new `ndarray`.

I flicked through the commits on the new fork, and other than being mildly surprised to see the `serde` feature ripped out (not that we care, here on Qiskit), the changes seem like the minimal amount necessary to support modern `clippy` and `ndarray 0.16`.  The previous licence was Apache 2.0, which is also fine for us (though we'd already be having problems if it weren't!).

mtreinish:
I don't remember why I removed the serde component I forked it locally about a month ago. There was a reason for it. If a user finds the fork and needs the serde support I can always add it back easily enough.

PR #13910, opened 2025-02-21
Propagate `DAGCircuit.name` in `ApplyLayout`

jakelishman:
This has always been absent, but `transpile` masked the problem by assigning the output circuit names by overwriting the `output_names` input kwarg if it wasn't set.  `generate_preset_pass_manager` didn't have the same logic, so it was observable that `ApplyLayout` didn't propagate the name.

We could have added similar name-propagation logic to `generate_preset_pass_manager`, but really the bug is in `ApplyLayout`; passes _should_ propagate the metadata.


releasenotes/notes/fix-pm-name-36baa550ce5d9c2c.yaml
```diff
@@ -0,0 +1,6 @@
+---
+fixes:
+  - |
+    Circuits compiled used a preset passmanager constructed by :func:`.generate_preset_pass_manager`
```
jlapeyre:
```suggestion
    Circuits compiled using a preset passmanager constructed by :func:`.generate_preset_pass_manager`
```

test/python/compiler/test_transpiler.py
```diff
@@ -1646,7 +1648,10 @@ def test_no_infinite_loop(self, optimization_level):
     @data(0, 1, 2, 3)
     def test_transpile_preserves_circuit_metadata(self, optimization_level):
         """Verify that transpile preserves circuit metadata in the output."""
-        circuit = QuantumCircuit(2, metadata={"experiment_id": "1234", "execution_number": 4})
+        metadata = {"experiment_id": "1234", "execution_number": 4}
+        name = "my circuit"
+        circuit = QuantumCircuit(2, metadata=metadata.copy(), name=name)
+        circuit.name = "my circuit"
```
jlapeyre: Isn't `circuit.name = "my circuit"` redundant? Passing `name=name` to the `__init__` function sets the attribute `name` already.

jlapeyre (dismissed):
I made one minor correction and a question on a minor point. Other than that it looks good.

test/python/compiler/test_transpiler.py
jakelishman: Yeah, thanks - I must have left that line in when I modified how I originally wrote the test.  Done in e2d9967ae2.

kevinhartman (approved):

PR #13912, opened 2025-02-22
13911 | Add ignoring of RuntimeWarning of aer in GenericBackendV2 test

### LINKED ISSUE #13911: Filter Aer not found warning in tests
### What should we add?

The `GenericBackend.run()` method emits a `RuntimeWarning` when it's called and qiskit-aer isn't installed. This is to inform users that the simulation will not include noise. This is useful for users, because it's not otherwise obvious that the simulation is ideal if the fallback backend of `BasicSimulator` is being used. However, when running unit tests we often don't install aer in some environments to save time or to explicitly test the `BasicSimulator` path. This ends up spamming the test logs with the warning message:
```
qiskit/providers/fake_provider/generic_backend_v2.py:348: RuntimeWarning: Aer not found using BasicSimulator and no noise
```
which provides no value. Normally the warning would only be shown once, but our test config changes the default filters so we see it on every call to run. We should adjust the base testing harness's filters so that we filter this warning from being displayed during the run.

LeanderCS:
### Summary
Because the `aer` isn't available on all envs, the test logs get flooded with sometimes meaningless `DeprecationWarnings`.

This PR adds that the affected class's error gets ignored in the tests.

Fixes #13911

### DIFF
--- a/test/python/providers/fake_provider/test_generic_backend_v2.py
+++ b/test/python/providers/fake_provider/test_generic_backend_v2.py
@@ -13,7 +13,6 @@
 """ Test of GenericBackendV2 backend"""
 
 import math
-import warnings
 
 from qiskit import ClassicalRegister, QuantumCircuit, QuantumRegister, transpile
 from qiskit.providers.fake_provider import GenericBackendV2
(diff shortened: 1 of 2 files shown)


ElePT (commented):
Hi [LeanderCS](https://github.com/LeanderCS), thanks for your contribution. If you read the issue carefully, you will notice that the proposed course of action is not stopping the issue of warnings inside the `GenericBackendV2` class, but modifying the base test configuration to filter the warnings only during our unit test process. If you'd like to contribute to that, please revert the current change and apply the filter in the unit test base class.

LeanderCS:
@ElePT Do I understand it correctly, that you want the ignoring to be added inside of `test/utils/base.py`?

Why would the ignoring in the affected test be not suitable?

ElePT (approved):
My bad, I read the changes too quickly and assumed the filter was in the class not the test, so my previous comment didn't fully make sense. Let me clarify: the approach was not incorrect, but because the generic backend is widely used in other unit tests (and those were the ones causing the flooding of the logs), it's more efficient to have the filter in the base test class. LGTM, and thanks again.

ElePT:

Issue #6740, opened 2021-07-13
Utilizing the symmetry of MCX/Toffoli gates for optimization


msciwoj1:
### What is the current behavior?

I don't have much experience with Github, I hope this is the right place.

A Toffoli gate, or a double-controlled NOT, or a CCXGate, is its own inverse. However, its standard decomposition into CX and single unitary gates in the Qiskit library is not itself symmetric. Qiskit recognizes that the CCX gate is its own inverse, that is, applying the` .inverse()` method to a circuit containing a single CCX returns back the same circuit. However, if two Toffoli gates are applied to the same qubits back-to-back, the simple transpiler pass with `optimization_level=3` or any lower will first decompose the gates, and then attempt to optimize the circuit and it will fail to recognize the circuit does nothing, i.e. implements an identity operation. 

If we have a look at the decomposition of the Toffoli gate: 
![singletoffoli](https://user-images.githubusercontent.com/23357449/125536056-86e3f868-b57d-4154-af4c-db091197b652.png)

We can see that there are 2 CNOTs and 3 single qubit rotations at the end that do not involve the target qubit.

If we use the following code:
```
qc = QuantumCircuit(4) 
qc.ccx(0,1,2)
qc.cx(2,3)
qc.ccx(0,1,2)
```
That generates the following circuit:
![twotoffoli](https://user-images.githubusercontent.com/23357449/125536396-d0ed2247-900a-4661-9189-f86363c25939.png)

We can see by inspection that nothing is happening to the first two qubits between the two Toffolis and the third qubit only acts as a control. If the second Toffoli is decomposed into the **inverse** of its standard decomposition, the standard transpiler will cancel the 5 operations mentioned above and produce the following result:
![twotoffoli_good](https://user-images.githubusercontent.com/23357449/125536827-c265b578-194a-4937-b215-85a068622db2.png)
However, if that is not done, the transpiler won't understand the need to do it by itself and produce the following circuit:
![twotoffoli_bad](https://user-images.githubusercontent.com/23357449/125536880-87a3110f-052d-423d-ab21-585c142d1516.png)
Those two circuits are identical in their effect, but the first one uses 8 CNOTs and 8 T-gates, and the second one uses 12 CNOTs and 14 T-gates. Also, the second one requires the underlying layout to have a connection between qubits 1 and 2, while the first one does not. 

[...]

PR #13916, opened 2025-02-24
Improve annotated plugin

### LINKED ISSUE #13566: Improve synthesis of controlled circuits with controlled gates within

alexanderivrii:
### Summary

This PR significantly improves the high-level-synthesis plugin for synthesizing ``AnnotatedOperation`` objects and in particular the synthesis of hierarchical circuits involving controlled subcircuits. Addresses #13566. ~Should be rebased on top of #13813.~ Does not need #13870, but serves as a basis to that PR explaining why we want to change the default handling of controlled gates to annotated gates.

In addition, the PR improves synthesis of multi-controlled CZ-gates and synthesis of controlled QFT-adders.

#### Multi-controlled X-gates

Here are experiments for the following 4 simple quantum circuits. All of these circuits are equivalent to a 6-controlled X-gate, however they are constructed in somewhat different ways: `circuit1` is a 6-controlled X-gate; `circuit2` is a 5-controlled CX-gate; `circuit3` is a 2-controlled quantum circuit that contains a 3-controlled CX-gate; `circuit4` is the same as `circuit3`, however the controls are added as *annotations* rather than being eagerly synthesized.

```
# circuit1: 6-controlled X
qc = QuantumCircuit(15)
qc.append(XGate().control(6), [0, 1, 2, 3, 4, 5, 6])

# circuit2: 5-controlled CX
qc = QuantumCircuit(15)
qc.append(CXGate().control(5), [0, 1, 2, 3, 4, 5, 6])

# circuit3: hierarchical circuit morally equivalent to 5-controlled CX
inner = QuantumCircuit(5)
inner.append(CXGate().control(3, annotated=False), [0, 1, 2, 3, 4])
controlled_inner_gate = inner.to_gate().control(2, annotated= False)
qc = QuantumCircuit(15)
qc.append(controlled_inner_gate, [0, 1, 2, 3, 4, 5, 6])

# circuit4: hierarchical circuit morally equivalent to 5-controlled CX
inner = QuantumCircuit(5)
inner.append(CXGate().control(3, annotated=True), [0, 1, 2, 3, 4])
controlled_inner_gate = inner.to_gate().control(2, annotated=True)
qc = QuantumCircuit(15)
qc.append(controlled_inner_gate, [0, 1, 2, 3, 4, 5, 6])
```
We are going to transpile these circuits with ``HighLevelSynthesis(basis_gates=["cx", "u"])``.

MAIN:
```
CIRCUIT1: {'u': 57, 'cx': 30}
CIRCUIT2: {'u': 57, 'cx': 30}
CIRCUIT3: {'u': 1575, 'cx': 974}
CIRCUIT4: {'u': 849, 'cx': 574}
```
THIS PR:
```
CIRCUIT1: {'u': 57, 'cx': 30}
CIRCUIT2: ({'u': 57, 'cx': 30}
CIRCUIT3: {'u': 1543, 'cx': 990}
CIRCUIT4: {'u': 57, 'cx': 30}
```
As we can see, both `circuit1` and `circuit2` are efficiently synthesized before and after this PR (note that the synthesis algorithm uses available ancillary qubits). However, the hierarchical `circuit3` is synthesized very poorly: due to eagerly synthesizing control logic, we first construct the circuit for a 3-controlled CX-gate, and then control every gate in this circuit using 2 additional controls. To improve the above, we need to do two things: create controlled gates as annotated gates (using `annotated=True`) and improving the `HighLevelSynthesis` pass to handle annotated operations, which is done in this PR by improving the `AnnotatedSynthesisDefault` plugin. With both of these changes, `circuit4` is synthesized just as efficiently as `circuit1` and `circuit2`.

#### Multi-controlled Z-gates

We repeat experiment with multi-controlled Z-gates instead of multi-controlled X-gates. Let circuits 5, 6, 7, 8 be the versions of circuits 1, 2, 3, 4 with Z/CZ gates instead of X/CX. One important difference is that we don't have native `MCZ` gates in Qiskit as compared to native `MCX`-gates, and in particular we don't have dedicated plugins for `MCZ`-gates. In theory, a multi-controlled Z-gate is equivalent to a multi-controlled X-gate with just two additional H-gates, and so we should get the same quality of results as above.

MAIN:
```
CIRCUIT5: {'u': 59, 'cx': 30}
CIRCUIT6: {'u': 159, 'cx': 78}
CIRCUIT7: {'u': 2549, 'cx': 1586}
CIRCUIT8: {'u': 5719, 'cx': 3626}
```
THIS PR:
```
CIRCUIT5: {'u': 59, 'cx': 30}
CIRCUIT6: {'u': 59, 'cx': 30}
CIRCUIT7: {'u': 1573, 'cx': 1002}
CIRCUIT8: {'u': 59, 'cx': 30}
```
However, in practice `circuit6` is worse than `circuit2`: this is because a `CZ`-gate is translated to a `CX-gate` with 2 Hadamards, and adding 5-controls creates two 5-controlled Hadamard gates instead of their non-controlled versions. This PR improves this behavior by providing an explicit decomposition for multiply-controlled `CZ`-gates in `_add_control.py`, similarly to how other controlled gates are currently handled. As before, creating a hierarchical circuit using `annotated=True` and using the improved `AnnotatedSynthesisDefault` plugin allows to synthesize `circuit8` using `30` CX-gates as compared to `1586`.

#### Controlled QFT-adders

A long discussed optimization (actually available as a part of unused by default `OptimizeAnnotated` transpiler pass) is to simplify `control-[U -- V -- U^{-1}]` to `U -- [control-V] -- U^{-1}`. This immediately applies to controlled QFT-adders, where the logic is precisely of this form: `control-[QFT -- V -- QFT^{-1}]`. With `AnnotatedSynthesisDefault` adapting two simple optimizations from `OptimizeAnnotated` pass and after allowing in the construction of the QFT-based adder to create the inverse `QFT^{-1}` as annotated (hence a new argument to `adder_qft_d00`), a controlled QFT-adder can be synthesized much more efficiently than before:

```
#circuit 9: our old friend controlled QFT-based adder
gate = HalfAdderGate(3).control(2, annotated=True)
qc = QuantumCircuit(gate.num_qubits)
qc.append(gate, qc.qubits)
```
MAIN:
```
CIRCUIT9: {'u': 1607, 'cx': 1310}
```
THIS PR:
```
CIRCUIT9: {'u': 539, 'cx': 450}
```


qiskit/circuit/add_control.py
```diff
@@ -109,10 +109,10 @@ def control(
 
     global_phase = 0
 
-    basis = ["p", "u", "x", "z", "y", "h", "sx", "sxdg", "rx", "ry", "rz", "cx"]
+    basis = ["p", "u", "x", "z", "y", "h", "sx", "sxdg", "rx", "ry", "rz", "cx", "cz"]
```
ShellyGarion: we should check (in a follow-up PR) what happens with other controlled gates, such as:
cy, ch, cp, crx, cry, crz, cs, csdg, cswap, csx, csxdg, cu, 
as well as ccx, ccz, c3x, c4x, c3sx.
Will adding them here improve the total CX cost?

ShellyGarion: btw, I really don't like the fact that we're copying this array in different part of the code.
I would suggest to set it somewhere and use it, see e.g. the `CollectClifford` transpiler pass:
https://github.com/Qiskit/qiskit/blob/f6608c23ed2f0e672179cc130dfffb9efa806515/qiskit/transpiler/passes/optimization/collect_cliffords.py#L77

qiskit/circuit/add_control.py
```diff
@@ -187,7 +187,7 @@ def apply_basic_controlled_gate(circuit, gate, controls, target):
 
     This implements multi-control operations for the following basis gates:
 
-        ["p", "u", "x", "z", "y", "h", "sx", "sxdg", "rx", "ry", "rz", "cx"]
+        ["p", "u", "x", "z", "y", "h", "sx", "sxdg", "rx", "ry", "rz", "cx", "cz"]
```
ShellyGarion: see my comment about this array above.

qiskit/circuit/add_control.py
```diff
@@ -230,6 +230,14 @@ def apply_basic_controlled_gate(circuit, gate, controls, target):
             controls[:] + [target[0]],  # CX has two targets
             target[1],
         )
+    elif gate.name == "cz":
```
ShellyGarion: btw, what is the improvement in cx-cost of adding `cz` here?

qiskit/transpiler/passes/synthesis/annotated_synthesis.py
```diff
@@ -0,0 +1,337 @@
...
+    an :class:`~.HLSConfig` object to use this method with :class:`~.HighLevelSynthesis`.
+    """
+
+    def run(self, high_level_object, coupling_map=None, target=None, qubits=None, **options):
+        # The plugin is triggered based on the name (i.e. for operations called "annotated").
+        # However, we should only do something when the operation is truthfully an AnnotatedOperation.
+        if not isinstance(high_level_object, AnnotatedOperation):
+            return None
+
+        # Combine the modifiers. If the were no modifiers, or the modifiers magically canceled out,
```
ShellyGarion: the --> there

qiskit/transpiler/passes/synthesis/annotated_synthesis.py
```diff
@@ -0,0 +1,337 @@
...
+        # the base operation with respect to a larger set of "basis" gates, to which the control
+        # logic can be added more efficiently. In particular, we add annotated operations to be
+        # in this larger set, exploting the fact that adding control to annotated operations
+        # returns a new annotated operation with an extended list of modifiers.
+        #
+        # Note that it is fine for this function to return a circuit with high-level objects
+        # (including annotated operations) as the HighLevelSynthesis transpiler pass will
+        # recursively re-synthesize this circuit, However, we should always guarantee that some
+        # progress is made.
+        basis = {
```
ShellyGarion: see my comment on the array above

qiskit/transpiler/passes/synthesis/annotated_synthesis.py
```diff
@@ -0,0 +1,337 @@
...
+
+
+def _are_inverse_ops(inst1: "CircuitInstruction", inst2: "CircuitInstruction"):
+    """A very naive function that checks whether two circuit instructions are inverse of
+    each other. The main use-case covered is a ``QFTGate`` and its inverse, represented as
+    an ``AnnotatedOperation`` with a single ``InverseModifier``.
+    """
+    res = False
+
+    if inst1.qubits != inst2.qubits or inst1.clbits != inst2.clbits or inst1.params != inst2.params:
```
ShellyGarion: in standard gates, usually the params of the inverse would be different than the params of the original gate,
for example: the inverse of rz(t) is rz(-t).
it does make sense however to check that both params have the same number of params.

qiskit/transpiler/passes/synthesis/annotated_synthesis.py
```diff
@@ -0,0 +1,337 @@
...
+    return res
+
+
+def _conjugate_decomposition(
+    circuit: QuantumCircuit,
+) -> tuple[QuantumCircuit, QuantumCircuit, QuantumCircuit] | None:
+    """
+    Decomposes a circuit ``A`` into 3 sub-circuits ``P``, ``Q``, ``R`` such that
+    ``A = P -- Q -- R`` and ``R = P^{-1}``.
+
```
ShellyGarion: can you do it more efficiently if you generate the circuit "P - R" and check if some transpiler passes return it to identity?
e.g. [InverseCancellation](https://docs.quantum.ibm.com/api/qiskit/qiskit.transpiler.passes.InverseCancellation), [CommutativeInverseCancellation](https://docs.quantum.ibm.com/api/qiskit/qiskit.transpiler.passes.CommutativeInverseCancellation), [CommutativeCancellation](https://docs.quantum.ibm.com/api/qiskit/qiskit.transpiler.passes.CommutativeCancellation), [RemoveIdentityEquivalent](https://docs.quantum.ibm.com/api/qiskit/qiskit.transpiler.passes.RemoveIdentityEquivalent)

test/python/transpiler/test_high_level_synthesis.py
```diff
@@ -677,6 +695,114 @@ def test_both_basis_gates_and_plugin_specified(self):
...
+            transpiled = hls_pass(qc3)
+            transpiled_block = transpiled[0].operation.blocks[0]
+            expected_block = QuantumCircuit(2, global_phase=0.8)
+            expected_block.cx(0, 1)
+            self.assertEqual(transpiled_block, expected_block)
+
+
+class TestHighLevelSynthesisQuality(QiskitTestCase):
+    """Test the "quality" of circuits produced by HighLevelSynthesis."""
+
```
ShellyGarion, line 744: Are there tests that use several modifiers? the QFT-adder should use both conjuage and control, right?

test/python/transpiler/test_high_level_synthesis.py
```diff
@@ -677,6 +695,114 @@ def test_both_basis_gates_and_plugin_specified(self):
...
+        qct = HighLevelSynthesis(basis_gates=["cx", "u"])(qc)
+        self.assertLessEqual(qct.count_ops()["cx"], 30)
+
+    def test_controlled_qft_adder(self):
+        """Test QFT-based synthesis of half-adder gate."""
+        gate = HalfAdderGate(3).control(2, annotated=True)
+        qc = QuantumCircuit(gate.num_qubits)
+        qc.append(gate, qc.qubits)
+        qct = HighLevelSynthesis(basis_gates=["cx", "u"], qubits_initially_zero=False)(qc)
+        self.assertLessEqual(qct.count_ops()["cx"], 450)
```
ShellyGarion, line 801: is this bound optimal? (assuming that the QFT requires 198). what is the formula for the CX count?
is the bound of 198 for QFT optimal? is there a formula for the CX count?

test/python/transpiler/test_high_level_synthesis.py
ShellyGarion: perhaps this example is also relevant for the conjugate modifier (since ccx gate is self-conjugate):
https://github.com/Qiskit/qiskit/issues/6740

ShellyGarion (commented):
That's a great PR! see some comments and suggestions below.

qiskit/circuit/add_control.py
alexanderivrii: Done in [86db649](https://github.com/Qiskit/qiskit/pull/13916/commits/86db649bbf91d6e915829972eec1bec5ba4406c3)

qiskit/circuit/add_control.py
alexanderivrii: Done in [86db649](https://github.com/Qiskit/qiskit/pull/13916/commits/86db649bbf91d6e915829972eec1bec5ba4406c3)

qiskit/circuit/add_control.py
alexanderivrii: Without the explicit treatment of `cz` (i.e. without this PR), we take the definition of `cz`: `cz(0, 1) = h(1) cx(0, 1) h(1)` and then add controls, leading to a circuit with a multi-controlled Hadamard gate followed by a multi-controlled CX gate, followed by another multi-controlled Hadamard gate. The current optimization allows to use Hadamard gates instead of multi-controlled Hadamard gates.

qiskit/transpiler/passes/synthesis/annotated_synthesis.py
alexanderivrii: Done in [50902ea](https://github.com/Qiskit/qiskit/pull/13916/commits/50902ead2c4703e9948cdcbbb21e39a5c92597f1)

qiskit/transpiler/passes/synthesis/annotated_synthesis.py
alexanderivrii: Thanks @ShellyGarion! I have fixed the above + added tests in [a5cb4a7](https://github.com/Qiskit/qiskit/pull/13916/commits/a5cb4a726c45730d865aa034b172a32728454693).

qiskit/transpiler/passes/synthesis/annotated_synthesis.py
alexanderivrii: Hmm, this is an interesting suggestion. However, none of the passes above would detect that the annotated-controlled QFT gate is the inverse of the QFT-gate (for the same numbers of control qubits), unless a matrix-based comparison would be used (but that would be slow and only work for small numbers of qubits). In principle we could run the full `transpile`, but that would be very slow. So I do think that we need a dedicated function that checks whether one gate is an inverse of another gate, while supporting either or both to be of type `AnnotatedOperation`.

test/python/transpiler/test_high_level_synthesis.py
alexanderivrii: For the first comment, the `test_controlled_qft_adder` test should be an example of a circuit with an annotated-control gate that has an annonated-inverse gate in its definition (I have made the more explicit about this).

For the second comment, I don't see how this PR helps #6740.

test/python/transpiler/test_high_level_synthesis.py
alexanderivrii: I don't think any of these bounds are optimal. This is what we can get currently _by explicitly reasoning about controls and inverses_ via annotated operations, but there is no guarantee that the same circuit cannot be simplified or synthesized in a different way. The point of these tests is that we accidentally don't make the synthesis of certain high-level-objects worse than they are now.

alexanderivrii:
Thanks @ShellyGarion for the review. I have addressed your review comments, please take another look! Update: I have also improved/updated this PR's summary and release notes.

qiskit/circuit/_add_control.py
```diff
@@ -24,21 +24,47 @@
...
+    "z",
+    "y",
+    "h",
+    "sx",
+    "sxdg",
+    "rx",
+    "ry",
+    "rz",
+    "cx",
+    "cz",
```
ShellyGarion, line 46: I think that some gates should be added here:
- s and sdg are missing (since we have cs and csdg).
- swap (since we have cswap)
- u1 and u3 (since we have cu1 and cu3)
- should we also add here ccx, c3x ?

qiskit/circuit/_add_control.py
```diff
@@ -24,21 +24,47 @@
 from ._utils import _ctrl_state_to_int
 
 
+# The list of gates whose controlled versions have efficient synthesis algorithms.
```
ShellyGarion, line 27: do you mean controlled or multi-controlled?

releasenotes/notes/improve-annotated-plugin-39afc1ede65fb9fb.yaml
```diff
@@ -0,0 +1,33 @@
+---
+features:
+  - |
+    Improved synthesis of multi-controlled :class:`.CZGate` gates.
```
ShellyGarion, line 4: is there a test for it? (in test_multi_controlled_synthesis)?

qiskit/transpiler/passes/synthesis/hls_plugins.py
```diff
@@ -1803,67 +1813,110 @@ def run(self, high_level_object, coupling_map=None, target=None, qubits=None, **
...
+        # the base operation with respect to a larger set of "basis" gates, to which the control
+        # logic can be added more efficiently. In addition, we add annotated operations to be
+        # in this larger set, exploiting the fact that adding control to annotated operations
+        # returns a new annotated operation with an extended list of modifiers.
+        #
+        # Note that it is fine for this function to return a circuit with high-level objects
+        # (including annotated operations) as the HighLevelSynthesis transpiler pass will
+        # recursively re-synthesize this circuit, However, we should always guarantee that some
+        # progress is made.
+        basis = set(EFFICIENTLY_CONTROLLED_GATES + ["annotated", "mcx", "qft"])
```
ShellyGarion, line 1830: are these the only relevant gates? how about mcphase? or mcrz/mcry/mcrx?

ShellyGarion (commented):
I still need to look more carefully into the code and tests. 
How do we decide which gates to include in EFFICIENTLY_CONTROLLED_GATES and in the HLS plugin?

qiskit/circuit/_add_control.py
alexanderivrii: By "controlled" I mean "controlled with an arbitrary number of qubits". And this is what we generally mean when we say "controlled" in Qiskit, as reflected in the names ``Gate.control``, ``_add_control``, etc.

releasenotes/notes/improve-annotated-plugin-39afc1ede65fb9fb.yaml
alexanderivrii: Nice catch, I have added explicit tests in [26cf007](https://github.com/Qiskit/qiskit/pull/13916/commits/26cf007b3180b3a77825a7c1a54644ffa85377d7).

qiskit/circuit/_add_control.py
alexanderivrii: This is a good question, also related to your question on how to decide which gates to include in this list.

Let's try to think through this together, taking the ``SGate`` as a perfect example. What happens when we call ``SGate().control(num_ctrl_qubits)``? We can check that ``SGate`` implements its own ``control`` method, which returns ``CSGate`` when ``num_ctrl_qubits==1`` and falls back on the general ``Gate.control`` mechanism otherwise. The latter calls this code in the ``_add_control`` module. Since ``SGate`` is currently not in the ``EFFICIENTLY_CONTROLLED_GATES``, we are going to unroll it into this set using ``UnrollCustomDefinitions`` and ``BasisTranslator`` passes, which will produce (a circuit containing) a single ``PhaseGate``. Now, ``PhaseGate`` is in ``EFFICIENTLY_CONTROLLED_GATES`` and the method ``apply_basic_controlled_gate`` will be used, producing an ``MCPhase`` gate. I would claim that this is the best we can do from the efficiency perspective (modulo to how then an ``MCPhase`` gets synthesized), so technically we do not need to add an ``SGate`` to the list above. On the other hand, adding it to the list would avoid the need to run the unroller pass, and since ``SGate`` is a very common gate it might make sense to add it to the list from the perspective of efficiency.

Summarizing, we probably want to consider the gates above + other standard gates and see if we should add them to the list. We should also decide which part should be handled by a gate's native control mechanism vs. ``apply_basic_controlled_gate``, hopefully avoiding any kind of duplication. I suggest to do this as a follow up.

qiskit/transpiler/passes/synthesis/hls_plugins.py
alexanderivrii: This is also a great question. For ``MCPhaseGate``, we don't have a plugin mechanism (correct?), so the definition circuit will be used. Can we do something better using ancillas? Is an ``MCPhaseGate`` used when constructing more complex circuits? If yes, then it could make sense to first unroll a more complex circuit to a circuit with ``MCPhaseGate``, then apply the best known algorithm for ``MCPhaseGate``. There is also a question on whether we need a native ``MCPhaseGate `` at all, as opposed to a controlled phase gate. Bottom-line: I agree that we should think about the above, but I would not like to add more gates here without concrete use cases.

ShellyGarion: what about adders? maybe in the future? say someone wants to add a controlled adder (w/o specifying the method)

test/python/circuit/library/test_multipliers.py
```diff
@@ -139,10 +139,13 @@ def test_modular_cumulative_multiplier_custom_adder(self):
             _ = HRSCumulativeMultiplier(3, 3, adder=VBERippleCarryAdder(3))
 
     def test_plugins(self):
-        """Test setting the HLS plugins for the modular adder."""
+        """Test setting HLS plugins for the multiplier."""
 
-        # all gates with the plugins we check, including an expected operation
-        plugins = [("cumulative_h18", "ch"), ("qft_r17", "mcphase")]
+        # For each pliugin, we check the presence of an expected operation after
```
ShellyGarion: pliugin --> plugin

ShellyGarion (commented):
LGTM! there is a minor misprint, but this PR is ready to go.

test/python/circuit/library/test_multipliers.py
alexanderivrii: thanks for spotting this :) fixed in [29cd6d7](https://github.com/Qiskit/qiskit/pull/13916/commits/29cd6d776fffcdb2601419de69595f6b31daffb3)

alexanderivrii:
> what about adders? maybe in the future? say someone wants to add a controlled adder (w/o specifying the method)

Hmm, that's an interesting point: while we try to make sure that the default plugins use the best synthesis algorithm for each high-level-gate in the circuit, these "locally best" choices might not be the overall best, taking the rest of the circuit and/or the following transpilation into account. I don't see an immediate solution to this, and I am afraid that there is only so much we can do automatically...

ShellyGarion (approved):

PR #13278, opened 2024-10-04
Add ParameterExpression in Rust

doichanj:
### Summary
This PR is a new implementation of  ParameterExpression written in Rust for issue #13267 

### Details and comments

This PR implements our own symbolic expression library written in Rust, this replaces symengine from `ParameterExpression` class.

`crates/circuit/symbol_expr.rs` is a symbolic expression module. Equations with symbols and values are expressed by using AST structure with nodes of operations. This module contains all the operators defined in Python's ParameterExpression class.

To separate Rust's implementation from Python we prepared Python interface class `crates/circuit/symbol_expr_py.rs` that is called from ParameterExpression in Python, instead of symengine.

`crates/circuit/parameterexpression.rs` is the ParameterExpression class for Rust. It provides similar functionalities to Python version. (but maybe we have to add more)

PR #13917, opened 2025-02-24
Handle an optional symengine and sympy in qpy

mtreinish:
### Summary

This commit changes the API for qpy.load() to make it explicit that symengine may not be installed in the future but is a requirement for deserializing certain payloads that were generated using symengine. We're expecting to drop symengine as a requirement with #13278 which should improve Qiskit's installation story and also will facilitate adding a C api for circuits. But the qpy v10, v11, and v13 have a hard requirement on symengine (specifically 0.11.0 and 0.13.0) to be able to deserialize `ParameterExpression` if the symbolic encoding was set to use it. Since the function needs to support being able to deserialize these payloads symengine will need to be installed. This commit adds the document and release note to indicate this change in behavior.

Similarly, `qpy.dump()` will have issues if symengine is not used anymore for the internal representation for QPY 10, 11, and 12 of an expression relies on symengine. So we won't necessarily be able to generate a symengine expression losslessly or without any overhead so trying to support this in a model without symengine installed isn't a tenable. This commit raises the minimum compatibility version so symengine (or sympy) isn't needed to generate the payload anymore.

### Details and comments


jakelishman (commented):
Do you potentially want to put in the necessary `HAS_SYMENGINE` optional and the `@require_in_call` decorators with this PR, so that the whole principle is in place?

test/python/qpy/test_circuit_load_from_qpy.py
```diff
@@ -209,52 +209,6 @@ def test_no_register(self, opt_level):
 class TestVersionArg(QpyCircuitTestCase):
     """Test explicitly setting a qpy version in dump()."""
 
-    def test_custom_gate_name_overlap_persists_with_minimum_version(self):
```
mtreinish, line 212: This test was removed because it was explicitly testing the bug fixed in QPY 11 was present if we specifically set the QPY version to 10. Since the bug is still fixed in QPY 13, there isn't anything to test anymore since we can't generate a buggy payload.

qiskit/qpy/interface.py
```diff
@@ -134,11 +134,9 @@ def dump(
...
             will be passed the ``.metadata`` attribute for each program in ``programs`` and will be
             used as the ``cls`` kwarg on the `json.dump()`` call to JSON serialize that dictionary.
-        use_symengine: If True, all objects containing symbolic expressions will be serialized
-            using symengine's native mechanism. This is a faster serialization alternative,
-            but not supported in all platforms. This flag only has an effect if the emitted QPY format
-            version is 10, 11, or 12. For QPY format version >= 13 (which is the default starting in
-            Qiskit 1.3.0) this flag is no longer used.
+        use_symengine: This flag is no longer used by QPY versions supported by this function and
+            will have no impact on the generated QPY payload except to set a field in a QPY v13 file
+            header which is unused.
```
jakelishman, line 130: It was potentially a mistake for us not to update `symbolic_encoding` in version 13 to be some typecode that's independent of sympy and symengine.  I guess there's not much to be done about it now.

qiskit/qpy/common.py
```diff
@@ -311,13 +308,19 @@ def mapping_from_binary(binary_data, deserializer, **kwargs):
     return mapping
 
 
-def load_symengine_payload(payload: bytes) -> symengine.Expr:
+@HAS_SYMENGINE.require_in_call
```
jakelishman: Since this is an internal function, we might want to give `require_in_call` a string argument like `require_in_call("QPY version 10--12 payloads with symengine parameters")`  to explain what the failure is?  Otherwise it'll be pretty esoteric for the user, like "symengine is required to use 'load_symengine_payload'" or something.

qiskit/qpy/binary_io/circuits.py
```diff
@@ -755,13 +755,12 @@ def _write_instruction(
...
-            # Assign a uuid to each instance of a custom operation
-            gate_class_name = f"{gate_class_name}_{uuid.uuid4().hex}"
+        # Assign a uuid to each instance of a custom operation
+        gate_class_name = f"{gate_class_name}_{uuid.uuid4().hex}"
         # ucr*_dg gates can have different numbers of parameters,
         # the uuid is appended to avoid storing a single definition
         # in circuits with multiple ucr*_dg gates.
-        elif instruction.operation.name in {"ucrx_dg", "ucry_dg", "ucrz_dg"}:
+        if instruction.operation.name in {"ucrx_dg", "ucry_dg", "ucrz_dg"}:
             gate_class_name = f"{gate_class_name}_{uuid.uuid4()}"
```
jakelishman: I think we're duplicating the uuid append to the ucr* gates now?

jakelishman (commented):
For most of this PR I'm happy, but I'm a bit concerned that we didn't issue deprecation warnings in Qiskit 1.4 about the withdrawal of support for writing QPY versions 10--12.  How painful would it be to continue to support them, and document that the optional is required to support write out of those QPY versions with symengine/sympy?

qiskit/qpy/interface.py
mtreinish: Yeah, it was an oversight, we were a bit rushed trying to get QPY13 in for 1.3. It's something we could potentially fix in v14, but it's a bit tight timewise to slip that in to v14. But we also missed the deprecation of the kwarg here so we'd be stuck with an arg that has no impact either way.

qiskit/qpy/binary_io/circuits.py
mtreinish: Ah good catch, I was just mechanically updating the logic to remove the version branch.

qiskit/qpy/interface.py
jakelishman: I'm not all that beaten up about the kwarg tbh, it's more the payload.  Right now, with QPY 13 already having shipped with a meaningless field there, I'm not sure how much purpose there is in changing it in QPY 14.

mtreinish:
> For most of this PR I'm happy, but I'm a bit concerned that we didn't issue deprecation warnings in Qiskit 1.4 about the  withdrawal of support for writing QPY versions 10--12. How painful would it be to continue to support them, and document that the optional is required to support write out of those QPY versions with symengine/sympy?

The way this is implemented in the #13278 currently is that it basically stringifies the inner rust expression object, runs it through a regex find replace to update the syntax, then passes that string to sympy to create a new expression object. If the sympy object is then either used verbatim if sympy encoding is used or converted to a symengine expression and that is encoded. It seems to work for all our current tests that use an explicit version, but it seems overly fragile for a core piece of qiskit.

FWIW, we did document this case in the [stability guidelines](https://docs.quantum.ibm.com/open-source/qiskit-sdk-version-strategy#upgrade-strategy) and the dump docstring to be only guaranteed to work for the n-1 release, so I  think we have a leg to stand on here. Obviously it would have been much better if we did warn about it in <= 1.4.0, but that ship has sailed.

qiskit/qpy/binary_io/circuits.py
mtreinish: Fixed in: https://github.com/Qiskit/qiskit/pull/13917/commits/36c2623d8abb98bec5b347668d2f0aa92748cad9

qiskit/qpy/common.py
mtreinish: Added in: https://github.com/Qiskit/qiskit/pull/13917/commits/8b9ec52f47498bc901abf91991ea6b12775a8590

qiskit/qpy/interface.py
```diff
@@ -252,9 +250,13 @@ def load(
         A list is always returned, even if there is only 1 program in the QPY data.
 
     Raises:
-        QiskitError: if ``file_obj`` is not a valid QPY file.
+        QiskitError: if ``file_obj`` is not a valid QPY file
+        TypeError: When invalid data type is loaded.
+        MissingOptionalLibraryError: If the ``symengine`` engine library is
```
kevinhartman, line 255: 🚂

kevinhartman (approved):
This looks good to me, though I'll avoid adding it to the queue in case anyone else is still looking.

Personally, I don't see much issue with dropping support for writing V12 and older, since this is squarely allowed in our stability guidelines.

mtreinish:
Given the time crunch I've enqueued this for merge

PR #13919, opened 2025-02-24
Remove giant static QV file from benchmarks

mtreinish:
### Summary

The qasm2 import benchmarks included a qasm file that was a transpiled 100 qubit quantum volume circuit that was taken from benchpress. This qasm file had ~300,000 lines and was only being used as a benchmark of the qasm2 importer. However, this file is unnecessarily large in the repo when we can just generate a large qasm file ourselves for the benchmark. This commit removes the file and rewrites the benchmark to generate a qasm file on the fly.

### Details and comments


test/benchmarks/circuit_construction.py
```diff
@@ -159,16 +163,16 @@ def time_param_circSU2_100_build(self, num_qubits):
 
 
 class QasmImport:
+    def setup(self):
+        qv_circuit = quantum_volume(100, seed=2025_12345)
+        qv_circuit.measure_all()
+        backend = GenericBackendV2(100, coupling_map=CouplingMap.from_line(100), seed=2025_12345)
+        t_qv = transpile(qv_circuit, backend, optimization_level=0)
+        self.qasm = dumps(t_qv)
```
jakelishman: A couple of comments:
- how long does this take to generate, transpile, and dump?  I think setup time still counts towards the "timeout" for a test, so I think it needs to reliably fall under 30s
- is `optimization_level=0` deterministic?  It still runs Sabre as the routing algorithm, so I'm assuming it needs a seed
- (Very minor: personally I'd always rather we did `from qiskit import qasm2; qasm2.dumps(...)` instead of importing `dumps`, since the name isn't super meaningful out of the context of its namespace.)

jakelishman (commented):
In general this seems like a sensible thing to do, though we should be super confident that we're not blowing out the timeout in the `setup` - I know the `quantum_volume` function referred to here is still some super old one that's vendored into the benchmark suite to avoid invalidating the benchmarks when the Qiskit version updates.

test/benchmarks/circuit_construction.py
mtreinish: A full `asv run` on the latest commit takes < 1 min (with a prebuilt env so no compilation) when I run it locally. I was thinking we should be quick enough now, but I can scale it back if you'd prefer. We're actually using the new rust generator in this module already since cf5ef0337d519c9123ce1aa828f0c1df051daf84 so construction time is very fast.

Oops, I missed the seed on the transpile call. I thought I added it. I'll push that up now and update the import path.

jakelishman: Oh my bad, I was thinking of `random_circuit`, not `quantum_volume`.

mtreinish: Done in: https://github.com/Qiskit/qiskit/pull/13919/commits/e437f248d88d4b927b35f2fdd56b139287237da8

ElePT (approved):
LGTM

PR #13928, opened 2025-02-26
Force LF style line endings in all text files

mtreinish:
### Summary

As a repository we need to have consistent line endings so that we can resolve diffs sanely. Right now we haven't had an opinion on this. But some CRLF line endings caused major merge conflicts on #13278 and was preventing unraveling an otherwise simple merge conflict resolution. This commit sets a global default to use LF line endings for all text files to prevent such issues in the future.

### Details and comments

The two release note files were normalized by git after adding the .gitattributes file.


1ucian0:
Somehow, it breaks the visual testing?

jakelishman:
The LaTeX and text-drawer tests use the same `filecmp.cmp` function as the image tests, which ofc opens the files in binary read mode, so does no line conversions.  On Windows, we're writing out the files with `\r\n` line endings from Python space, but the references always have `\n`.  We can either modify the `.gitattributes` so that `test/python/visualization/references/*.{txt,tex}` retains platform-native linefeeds (it's like `eol=native` or something), or modify the test function to pass a `binary=False`/`binary=True` flag.

.gitattributes
```diff
@@ -0,0 +1,6 @@
+# Default all text files to use LF line endings
+* text eol=lf
+
+visualization/references/*.{txt,tex} text=auto
```
jakelishman: I _think_ it's `eol=auto` rather than `text=auto`?  The `txt` and `tex` formats should be detected as `text`, and it's the EOL we care about.

mtreinish: Done in: https://github.com/Qiskit/qiskit/pull/13928/commits/d65c3e9ca624dd7c03d1f92fe1eb5cfc62e2fad8

PR #13929, opened 2025-02-27
Porting some of the MCX synthesis methods to Rust

alexanderivrii:
### Summary

This PR ports some of the MCX synthesis algorithms from Python to Rust:
* `c3x` - synthesis of MCX(3)
* `c4x` - synthesis of MCX(4)
* `synth_mcx_n_dirty_i15` - synthesis of MCX(k) using k-2 ancilla qubits
* `synth_mcx_noaux_v24` - synthesis of MCX(k) using no ancilla qubits

One important change is that the Rust circuits are "flat" (contain no nested subcircuits), so for instance the 4-controlled X-gate now gets expanded into 69 "basic" operations instead of a hierarchical circuit with nested subcircuits.

To support hierarchical synthesis algorithms, all the new internal synthesis functions return `CircuitData` (or rather `PyResult<CircuitData>`, instead of vectors or iterators over custom structs as used until now. In particular, this would immediately allow to use MCX-gates as building blocks within larger circuits.

In the process, I have extended `CircuitData` with multiple convenience functions:
* `push_standard_instruction` and `push_py_gate` are analogues of the existing `push_standard_gate`
* `from_standard_gate_definition` - returns the definition circuit of a particular gate
* `compose` - composes one quantum circuit onto another while optionally remapping qubits, this is a simpler variant of the Python `QuantumCircuit.compose` method
* `inverse` - constructs the inverse circuit, similar to the Python `QuantumCircuit.inverse` method
* `h`, `cx`, `cp`, etc. - similar to the Python `QuantumCircuit` functions.

The 4 ported MCX synthesis methods above each have a slightly different flavour:
* `c3x` - this is the definition circuit for the standard `C3X` gate
* `c4x` - hierarchical circuit
* `synth_mcx_n_dirty_i15` - another hierarchical circuit
* `synth_mcx_noaux_v24` - uses Python-side `MCPhase`-gate.


crates/accelerate/src/synthesis/multi_controlled/mcx.rs
```diff
@@ -0,0 +1,248 @@
...
+    CircuitData::from_standard_gate_definition(StandardGate::CCX, &[])
+}
+
+/// Efficient synthesis for 3-controlled X-gate.
+pub fn c3x() -> PyResult<CircuitData> {
+    CircuitData::from_standard_gate_definition(StandardGate::C3X, &[])
+}
+
+/// Efficient synthesis for 4-controlled X-gate.
+pub fn c4x() -> PyResult<CircuitData> {
```
ShellyGarion: would be nice to add a picture of the circuit here

crates/accelerate/src/synthesis/multi_controlled/mcx.rs
```diff
@@ -0,0 +1,248 @@
...
+        Some(&[Qubit(0), Qubit(1), Qubit(2), Qubit(3)]),
+    )?;
+    circuit.compose(
+        &CircuitData::from_standard_gate_definition(StandardGate::C3SX, &[])?,
+        Some(&[Qubit(0), Qubit(1), Qubit(2), Qubit(4)]),
+    )?;
+    Ok(circuit)
+}
+
+/// A block in the `action part`, see Iten et al.
```
ShellyGarion: which lemma or figure?

crates/accelerate/src/synthesis/multi_controlled/mcx.rs
```diff
@@ -0,0 +1,248 @@
...
+    let mut circuit = CircuitData::with_capacity(3, 0, 5, Param::Float(0.0))?;
+    circuit.h(2);
+    circuit.t(2);
+    circuit.cx(0, 2);
+    circuit.tdg(2);
+    circuit.cx(1, 2);
+    Ok(circuit)
+}
+
+/// A block in the `reset part`, see Iten et al.
```
ShellyGarion: which lemma or figure?

test/python/circuit/test_circuit_qasm.py
```diff
@@ -410,7 +410,7 @@ def test_circuit_qasm_with_mcx_gate_variants(self):
...
...,q3; u1(pi/4) q3; cx q1,q3; u1(-pi/4) q3; u2(0,pi) q3; cx q3,q6; t q6; cx q2,q6; tdg q6; h q6; u2(0,pi) q3; u1(pi/4) q3; cx q1,q3; u1(-pi/4) q3; cx q0,q3; u1(pi/4) q3; cx q1,q3; u1(-pi/4) q3; u2(0,pi) q3; h q5; p(pi/8) q3; p(pi/8) q4; p(pi/8) q6; p(pi/8) q5; cx q3,q4; p(-pi/8) q4; cx q3,q4; cx q4,q6; p(-pi/8) q6; cx q3,q6; p(pi/8) q6; cx q4,q6; p(-pi/8) q6; cx q3,q6; cx q6,q5; p(-pi/8) q5; cx q4,q5; p(pi/8) q5; cx q6,q5; p(-pi/8) q5; cx q3,q5; p(pi/8) q5; cx q6,q5; p(-pi/8) q5; cx q4,q5; p(pi/8) q5; cx q6,q5; p(-pi/8) q5; cx q3,q5; h q5; u2(-2*pi,pi) q3; u1(pi/4) q3; cx q1,q3; u1(-pi/4) q3; cx q0,q3; u1(pi/4) q3; cx q1,q3; u1(-pi/4) q3; u2(-2*pi,pi) q3; h q6; t q6; cx q2,q6; tdg q6; cx q3,q6; u2(-2*pi,pi) q3; u1(pi/4) q3; cx q1,q3; u1(-pi/4) q3; cx q0,q3; u1(pi/4) q3; cx q1,q3; u1(-pi/4) q3; u2(-2*pi,pi) q3; cx q3,q6; t q6; cx q2,q6; tdg q6; h q6; h q5; p(pi/8) q3; p(pi/8) q4; p(pi/8) q6; p(pi/8) q5; cx q3,q4; p(-pi/8) q4; cx q3,q4; cx q4,q6; p(-pi/8) q6; cx q3,q6; p(pi/8) q6; cx q4,q6; p(-pi/8) q6; cx q3,q6; cx q6,q5; p(-pi/8) q5; cx q4,q5; p(pi/8) q5; cx q6,q5; p(-pi/8) q5; cx q3,q5; p(pi/8) q5; cx q6,q5; p(-pi/8) q5; cx q4,q5; p(pi/8) q5; cx q6,q5; p(-pi/8) q5; cx q3,q5; h q5; }
```
ShellyGarion: this should be updated after https://github.com/Qiskit/qiskit/pull/14180 is merged, right?

test/python/qasm2/test_export.py
```diff
@@ -400,7 +400,7 @@ def test_mcx_gate_variants(self):
...
...,q3; u1(pi/4) q3; cx q1,q3; u1(-pi/4) q3; u2(0,pi) q3; cx q3,q6; t q6; cx q2,q6; tdg q6; h q6; u2(0,pi) q3; u1(pi/4) q3; cx q1,q3; u1(-pi/4) q3; cx q0,q3; u1(pi/4) q3; cx q1,q3; u1(-pi/4) q3; u2(0,pi) q3; h q5; p(pi/8) q3; p(pi/8) q4; p(pi/8) q6; p(pi/8) q5; cx q3,q4; p(-pi/8) q4; cx q3,q4; cx q4,q6; p(-pi/8) q6; cx q3,q6; p(pi/8) q6; cx q4,q6; p(-pi/8) q6; cx q3,q6; cx q6,q5; p(-pi/8) q5; cx q4,q5; p(pi/8) q5; cx q6,q5; p(-pi/8) q5; cx q3,q5; p(pi/8) q5; cx q6,q5; p(-pi/8) q5; cx q4,q5; p(pi/8) q5; cx q6,q5; p(-pi/8) q5; cx q3,q5; h q5; u2(-2*pi,pi) q3; u1(pi/4) q3; cx q1,q3; u1(-pi/4) q3; cx q0,q3; u1(pi/4) q3; cx q1,q3; u1(-pi/4) q3; u2(-2*pi,pi) q3; h q6; t q6; cx q2,q6; tdg q6; cx q3,q6; u2(-2*pi,pi) q3; u1(pi/4) q3; cx q1,q3; u1(-pi/4) q3; cx q0,q3; u1(pi/4) q3; cx q1,q3; u1(-pi/4) q3; u2(-2*pi,pi) q3; cx q3,q6; t q6; cx q2,q6; tdg q6; h q6; h q5; p(pi/8) q3; p(pi/8) q4; p(pi/8) q6; p(pi/8) q5; cx q3,q4; p(-pi/8) q4; cx q3,q4; cx q4,q6; p(-pi/8) q6; cx q3,q6; p(pi/8) q6; cx q4,q6; p(-pi/8) q6; cx q3,q6; cx q6,q5; p(-pi/8) q5; cx q4,q5; p(pi/8) q5; cx q6,q5; p(-pi/8) q5; cx q3,q5; p(pi/8) q5; cx q6,q5; p(-pi/8) q5; cx q4,q5; p(pi/8) q5; cx q6,q5; p(-pi/8) q5; cx q3,q5; h q5; }
```
ShellyGarion: this should be updated after https://github.com/Qiskit/qiskit/pull/14180 is merged, right?

ShellyGarion (commented):
LGTM - very nice work! I have some minor comments. I would be happy to hear @Cryoris opinion on the circuit_data.rs updates.

crates/circuit/src/circuit_data.rs
```diff
@@ -1960,6 +2008,144 @@ impl CircuitData {
...
+    /// Appends CPhase to the circuit.
+    #[inline]
+    pub fn cp(&mut self, theta: f64, q1: u32, q2: u32) {
+        self.push_standard_gate(
+            StandardGate::CU1,
+            &[Param::Float(theta)],
+            &[Qubit(q1), Qubit(q2)],
+        )
+        .expect("Error addding a standard gate to the circuit data");
+    }
```
raynelfss: Hmm... I'm unsure about this change here. Though I'm not against adding simpler methods to add gates onto a circuit, I do feel like this might be falling outside of the scope of this PR and should maybe be considered for a follow-up.

I do understand it imitates a lot of what Python already does but remember that Rust doesn't have to behave the same way.

crates/circuit/src/circuit_data.rs
```diff
@@ -1960,6 +2008,144 @@ impl CircuitData {
...
+        standard_gate: StandardGate,
+        params: &[Param],
+    ) -> PyResult<Self> {
+        standard_gate
+            .definition(params)
+            .ok_or(CircuitError::new_err(format!(
+                "Error extracting the definition of a standard gate {:?}",
+                standard_gate
+            )))
+    }
```
raynelfss: This may be a bit redundant, the gate itself should already provide you with its standard gate definition based on `StandardGate::definiton`, I don't see why this needs to be part of the `CircuitData` API just yet.

crates/circuit/src/circuit_data.rs
```diff
@@ -1444,6 +1446,52 @@ impl CircuitData {
...
+            op: operation.into(),
+            qubits,
+            clbits,
+            params,
+            label: None,
+            #[cfg(feature = "cache_pygates")]
+            py_op: OnceLock::new(),
+        });
+        Ok(())
+    }
```
raynelfss: I think `CircuitData::push_packed_operation` does all of these things here too. Since you have either a `StandardGate` or a `PyGate`, all you'd need to do is call `operation.into()` as you send either as an argument.

raynelfss (commented):
Just took a quick glance at what is being added to `CircuitData`.

crates/circuit/src/circuit_data.rs
alexanderivrii: Thanks! At the time that I wrote the above, I did not realize that this is possible. I have removed these functions in [c6376f4](https://github.com/Qiskit/qiskit/pull/13929/commits/c6376f408739b1f23d16ccace75a1875d9d3b132).

crates/circuit/src/circuit_data.rs
alexanderivrii: This is a valid point, I have removed this function in [106559b](https://github.com/Qiskit/qiskit/pull/13929/commits/106559bc68948bb1e6959e3cb008602a980f26ee)

crates/circuit/src/circuit_data.rs
alexanderivrii: Hmm, on the one hand I understand this point, and from a brief discussion with @mtreinish he was not super-excited to have these method inside CircuitData either.

On the other hand, many synthesis algorithms usually involve adding tons of standard gates to a quantum circuit, and I do feel that the code
```rust
fn action_gadget() -> PyResult<CircuitData> {
    let mut circuit = CircuitData::with_capacity(3, 0, 5, Param::Float(0.0))?;
    circuit.h(2);
    circuit.t(2);
    circuit.cx(0, 2);
    circuit.tdg(2);
    circuit.cx(1, 2);
    Ok(circuit)
}
```
is simpler to read and type than
```rust
fn action_gadget() -> PyResult<CircuitData> {
    let mut circuit = CircuitData::with_capacity(3, 0, 5, Param::Float(0.0))?;
    circuit.push_standard_gate(StandardGate::H, &[], &[Qubit(2)]);
    circuit.push_standard_gate(StandardGate::T, &[], &[Qubit(2)]);
    circuit.push_standard_gate(StandardGate::CX, &[], &[Qubit(0), Qubit(2)]);
    circuit.push_standard_gate(StandardGate::Tdg, &[], &[Qubit(2)]);
    circuit.push_standard_gate(StandardGate::CX, &[], &[Qubit(1), Qubit(2)]);
    Ok(circuit)
}
```
Though actually the latter is not as bad as I thought it would be. 

I can see multiple approaches on how to proceed:

1. Keep simple functions in `CircuitData` (as is done now and as in my first example)
2. Implement some `CircuitData` trait in `synthesis` (but not in `CircuitData` itself) that allows using the code as in the first example
3. Implicitly add these functions to `CircuitData` using some clever macro
4. Implement function like `h(&mut CircuitData, q: u32)` in synthesis
5. Use the longer form (as in my second example)

@raynelfss: you can probably suggest other and even better approaches. I will go with whatever you think is best.

alexanderivrii: Ok, I went ahead and implemented the approach (2), see [096bbbc](https://github.com/Qiskit/qiskit/pull/13929/commits/096bbbcb14230d3ef192c61eb32f63e7d68bae32). Do you think this is good enough?

test/python/qasm2/test_export.py
alexanderivrii: Quite possibly.

test/python/circuit/test_circuit_qasm.py
alexanderivrii: Quite possibly.

crates/accelerate/src/synthesis/multi_controlled/mcx.rs
alexanderivrii: This is a rather big circuit with about 100 gates. I don't think the picture is very illustrative.

crates/accelerate/src/synthesis/multi_controlled/mcx.rs
alexanderivrii: That's a good question. Our Python implementation did not fully match the figures in the paper by Iten et al. either, it's probably something that we have optimized along the way. What I did was to start with our Python code
```python
    for j in range(2):
        for i in range(num_ctrl_qubits):  # action part
            if i < num_ctrl_qubits - 2:
                if targets[i] != q_target or relative_phase:
                    # gate cancelling

                    # cancel rightmost gates of action part
                    # with leftmost gates of reset part
                    if relative_phase and targets[i] == q_target and j == 1:
                        qc.cx(q_ancillas[num_ancillas - i - 1], targets[i])
                        qc.t(targets[i])
                        qc.cx(q_controls[num_ctrl_qubits - i - 1], targets[i])
                        qc.tdg(targets[i])
                        qc.h(targets[i])
                    else:
                        qc.h(targets[i])
                        qc.t(targets[i])
                        qc.cx(q_controls[num_ctrl_qubits - i - 1], targets[i])
                        qc.tdg(targets[i])
                        qc.cx(q_ancillas[num_ancillas - i - 1], targets[i])
```
and introduce the subcircuits for the `if` and `else` branches; imho, this makes the code easier to read and avoids some code duplication.

crates/synthesis/src/multi_controlled/mcx.rs
```diff
@@ -0,0 +1,360 @@
...
+    /// Appends CPhase to the circuit.
+    #[inline]
+    fn cp(&mut self, theta: f64, q1: u32, q2: u32) {
+        self.push_standard_gate(
+            StandardGate::CPhase,
+            &[Param::Float(theta)],
+            &[Qubit(q1), Qubit(q2)],
+        );
+    }
+}
```
raynelfss: I understand the purpose of these making it easier for us to add operations to the circuit, and although I find them useful, there are a couple things to note.

1. The PR does extend `CircuitData` by adding ``CircuitData::compose()`` and ``CircuitData::inverse``. But adding this change here still feels a bit out of scope being more of an aesthetic change rather than adding much functionality. If we were to still want to include these changes, it should not be part of this PR.
2. If we do end up going for these changes, we may want to consider using the defined rust types for both `Qubit` and `Param` to avoid confusion and to allow us to extend the usage here. With the way these changes are implemented, one may only add gates with fixed parameters rather than having the possibility of adding a `ParameterExpression` in the future. So I suggest maybe sticking to using `Param` instead of an `f64`.

raynelfss (commented):
Just had a couple more thoughts about the `CircuitData` methods for adding gates. I still feel like they shouldn't be included in this PR.

test/python/qasm2/test_export.py
alexanderivrii: As expected, we had to update these tests: done in [536f1ac](https://github.com/Qiskit/qiskit/pull/13929/commits/536f1ac80dd3b1d4d3cdb2aea603a0ddcce69abb)

crates/synthesis/src/multi_controlled/mcx.rs
alexanderivrii: Thanks Ray! As we have discussed offline, I have removed the "pub" keyword and added a comment that the trait is not supposed to be user-facing (though we may want to make it into `pub(crate)` is it becomes used by other method in the `synthesis` crate).

crates/circuit/src/circuit_data.rs
```diff
@@ -1965,6 +1965,79 @@ impl CircuitData {
...
+                #[cfg(feature = "cache_pygates")]
+                py_op: OnceLock::new(),
+            });
+        }
+
+        self.add_global_phase(other.global_phase())?;
+        Ok(())
+    }
+
+    /// Construct the inverse circuit
```
raynelfss: You should mention that this only works with circuits that only contain standard gates.

crates/circuit/src/circuit_data.rs
```diff
@@ -1965,6 +1965,79 @@ impl CircuitData {
...
+    pub fn inverse(&self) -> PyResult<CircuitData> {
+        let inverse_global_phase =
+            Python::with_gil(|py| -> Param { multiply_param(self.global_phase(), -1.0, py) });
+
+        let mut inverse_circuit = CircuitData::clone_empty_like(self, None)?;
+        inverse_circuit.set_global_phase(inverse_global_phase)?;
+
+        for inst in self.data.iter().rev() {
+            let inverse_inst: Option<(StandardGate, SmallVec<[Param; 3]>)> = match &inst.op.view() {
+                OperationRef::StandardGate(gate) => gate.inverse(inst.params_view()),
```
raynelfss: Is there a reason why we can't inverse none of `StandardInstruction` variants (`Delay`, `Reset`, `Measure`, `Barrier`)?

crates/circuit/src/circuit_data.rs
```diff
@@ -1965,6 +1965,79 @@ impl CircuitData {
...
+                params: inst.params.clone(),
+                label: None,
+                #[cfg(feature = "cache_pygates")]
+                py_op: OnceLock::new(),
+            });
+        }
+
+        self.add_global_phase(other.global_phase())?;
+        Ok(())
+    }
```
raynelfss: This method is a bit different from the one we use in Python, it might need more expansion once #14421 is added to `CircuitData` so that it takes cares of vars too (if necessary).

raynelfss (commented):
Okay, I took a brief look at the new methods added and had a couple of questions.

crates/circuit/src/circuit_data.rs
alexanderivrii: This is a good point. In addition, some of the standard gates (such as DCX or ISwap) have inverses that do not correspond to a _single_ standard gate, however none of the synthesis algorithm currently unroll to such gates, and I think it is extremely unlikely that this will change in the future (and if it does, this would be a valid reason to introduce DCXDg, ISwapdg etc.)

crates/circuit/src/circuit_data.rs
alexanderivrii: For whatever reason, the `inverse` function has so far been implemented only for `StandardGate`s and not for other variants. 

In theory, we should implement this method for `PyGate`s and `UnitaryGate`s. 

For `StandardInstruction`s: it does not make sense to "invert" a reset or a measure instruction as these are simply not unitary operators, and in fact our Python code also does not allow this either. To be consistent with the Python code, we could say that inverting does not do anything to delays or barriers.

crates/circuit/src/circuit_data.rs
alexanderivrii: This is a very good point. Considering your points together, I think that we are not yet ready to expose `compose` and `inverse` methods as a part of `CircuitData`, and in addition we should have a deeper discussion on the API for these methods.

On the other hand, these are very useful convenience methods for the current `synthesis` code, similar to the `.h`, `.cp` and other methods that we hid under a non-public trait. Note that for the current synthesis methods do not require variables or stretches.

What do you say if for the purpose of this PR we also move the current implementations of the `compose` and `inverse` methods to that same non-public `synthesis` trait, with the idea to eventually properly implement these methods in separate PRs?

alexanderivrii: I have made this change in [75a2dc9](https://github.com/Qiskit/qiskit/pull/13929/commits/75a2dc99e6f27ef2bfef34b43b742282bbd956de)

crates/synthesis/src/multi_controlled/mcx.rs
```diff
@@ -0,0 +1,437 @@
...
+        .ok_or(CircuitError::new_err(
+            "Error extracting the definition of C3SX",
+        ))
+}
+
+/// Convenience methods to add gates to the circuit.
+///
+/// This trait is **not** intended to be user-facing. It defines utility functions
+/// that make the code easier to read and that are used only for synthesis.
+trait CircuitDataAdder {
```
raynelfss: Now that the trait adds more functionality than just adding gates to the circuit, you might want to rename it to something that reflects that it's an extension aimed mainly at synthesis. Some suggestions may be `CircuitDataForSynthesis` or `CircuitDataExt`, I'll let you be the judge here.

crates/synthesis/src/multi_controlled/mcx.rs
```diff
@@ -0,0 +1,437 @@
...
+        let num_qubits = (num_controls + 1) as u32;
+        let target = num_controls as u32;
+
+        let mut circuit = CircuitData::with_capacity(num_qubits, 0, 0, Param::Float(0.0))?;
+        circuit.h(target);
+
+        let mcphase_cls = imports::MCPHASE_GATE.get_bound(py);
+        let mcphase_gate = mcphase_cls
+            .call1((PI, num_controls))
+            .expect("Could not create MCPhaseGate Python-side");
```
raynelfss: Is there a reason why we're not leveraging the `Result` nature of the function here and unwrapping the value? Remember that panicking is an error that you can't recover from. Since this here is only creating the gate instance, I don't see why we couldn't just return whatever error happens here in the Python side. So here I would just keep the result and handle it with the `?` operator.

```rust
        let mcphase_gate = mcphase_cls
            .call1((PI, num_controls))?;
```

or if you wanted a custom error message (something with `CircuitError` rather), you could do something like:

```rust
        let mcphase_gate = mcphase_cls
            .call1((PI, num_controls))
            .map_err(|_| CircuitError::new_err("Could not create MCPhaseGate Python-side"))?;
```
I'll leave that up to you :)

crates/synthesis/src/multi_controlled/mod.rs
```diff
@@ -10,11 +10,85 @@
...
+pub fn py_c3x() -> PyResult<CircuitData> {
+    c3x()
+}
+
+/// Efficient synthesis for 4-controlled X-gate.
+#[pyfunction]
+#[pyo3(name = "c4x")]
+pub fn py_c4x() -> PyResult<CircuitData> {
+    c4x()
+}
```
raynelfss: **This is mostly a nitpick** but for these two methods,  as well as the ones following it, I think you could just expose the original methods as `pyfunction`. The only reason I'd see you separate them would be that they'd possibly return a rust-only error type instead of a `PyError`, or that they use references, but neither seems to be the case.

I do like the fact that it keeps things cleaner and streamlined, so I have no problem keeping it this way either. Let me know if there is any other reasoning that I might be missing here.

crates/synthesis/src/multi_controlled/mcx.rs
raynelfss: You might want to move the trait to its own file within the `synthesis` crate in the future, if needed. (Not for this PR).

test/python/transpiler/test_decompose.py
```diff
@@ -199,45 +199,78 @@ def test_decompose_only_given_label(self):
 
     def test_decompose_only_given_name(self):
         """Test decomposition parameters so that only given name is decomposed."""
-        decom_circ = self.complex_circuit.decompose(["mcx"], reps=2)
-        dag = circuit_to_dag(decom_circ)
+        # same circuit but with barriers, otherwise the order of nodes before and
+        # after expanding the MCX gate might not be preserved
```
raynelfss: Would adding barriers to `self.complex_circuit` affect any of the other tests?

test/python/circuit/test_circuit_qasm.py
```diff
@@ -412,7 +412,7 @@ def test_circuit_qasm_with_mcx_gate_variants(self):
...
+gate mcx_recursive q0,q1,q2,q3,q4,q5,q6 { h q6; t q6; cx q2,q6; tdg q6; cx q3,q6; h q3; t q3; cx q1,q3; tdg q3; cx q0,q3; t q3; cx q1,q3; tdg q3; h q3; cx q3,q6; t q6; cx q2,q6; tdg q6; h q6; h q3; t q3; cx q1,q3; tdg q3; cx q0,q3; t q3; cx q1,q3; tdg q3; h q3; h q5; p(pi/8) q3; p(pi/8) q4; p(pi/8) q6; p(pi/8) q5; cx q3,q4; p(-pi/8) q4; cx q3,q4; cx q4,q6; p(-pi/8) q6; cx q3,q6; p(pi/8) q6; cx q4,q6; p(-pi/8) q6; cx q3,q6; cx q6,q5; p(-pi/8) q5; cx q4,q5; p(pi/8) q5; cx q6,q5; p(-pi/8) q5; cx q3,q5; p(pi/8) q5; cx q6,q5; p(-pi/8) q5; cx q4,q5; p(pi/8) q5; cx q6,q5; p(-pi/8) q5; cx q3,q5; h q5; h q3; t q3; cx q1,q3; tdg q3; cx q0,q3; t q3; cx q1,q3; tdg q3; h q3; h q6; t q6; cx q2,q6; tdg q6; cx q3,q6; h q3; t q3; cx q1,q3; tdg q3; cx q0,q3; t q3; cx q1,q3; tdg q3; h q3; cx q3,q6; t q6; cx q2,q6; tdg q6; h q6; h q5; p(pi/8) q3; p(pi/8) q4; p(pi/8) q6; p(pi/8) q5; cx q3,q4; p(-pi/8) q4; cx q3,q4; cx q4,q6; p(-pi/8) q6; cx q3,q6; p(pi/8) q6; cx q4,q6; p(-pi/8) q6; cx q3,q6; cx q6,q5; p(-pi/8) q5; cx q4,q5; p(pi/8) q5; cx q6,q5; p(-pi/8) q5; cx q3,q5; p(pi/8) q5; cx q6,q5; p(-pi/8) q5; cx q4,q5; p(pi/8) q5; cx q6,q5; p(-pi/8) q5; cx q3,q5; h q5; }
```
raynelfss, line 415: What triggered the changes here? From checking locally it seems to just be the order some gates appear for different qubits, but still what is causing this to happen?

raynelfss (commented):
This is much closer now, thank you for the quick changes. Here are some more comments.

crates/synthesis/src/multi_controlled/mcx.rs
alexanderivrii: Changing this to `CircuitDataForSynthesis`. Done in Done in [501c568](https://github.com/Qiskit/qiskit/pull/13929/commits/501c568d11ac88f731c4117755c0b67452c6b65d)..

crates/synthesis/src/multi_controlled/mod.rs
alexanderivrii: This is a good suggestion. I think I have originally made the split because the functions needed the `'py` token.

test/python/transpiler/test_decompose.py
alexanderivrii: Unfortunately it does. But I agree that it makes sense to minimize code duplication, for example by creating `self.complex_circuit_with_barriers` and using it in multiple tests.

test/python/circuit/test_circuit_qasm.py
alexanderivrii: Ah yes, there is indeed a subtle difference between the circuits produced by the old Python code and the new Rust code for `synth_mcx_n_dirty_i15`.

The Python code had the following lines
```
# implements an optimized toffoli operation
# up to a diagonal gate, akin to lemma 6 of arXiv:1501.06911
qc.h(targets[i])
qc.t(targets[i])
qc.cx(q_controls[num_ctrl_qubits - i - 2], targets[i])
qc.tdg(targets[i])
qc.cx(q_controls[num_ctrl_qubits - i - 1], targets[i])
qc.t(targets[i])
qc.cx(q_controls[num_ctrl_qubits - i - 2], targets[i])
qc.tdg(targets[i])
qc.h(targets[i])
```
which implement a relative Toffoli gate, and which got  replaced by the single Rust line
```
circuit.compose(
  &rccx()?,
  &[Qubit(controls[0]), Qubit(controls[1]), Qubit(ancillas[0])], 
  &[]
)?;
```
where `rccx` is a shortcut for extracting the definition circuit for `RCCX` (from `operations.rs`).

This is correct in the sense that both the Python code above and the definition for `RCCX` in `operations.rs` lead to correct decompositions for the original MCX gate (and we have lots of tests that check the correctness of circuits produced by different synthesis methods). However, the two definitions for `RCCX` are slightly different (I think you can get one from the other by swapping qubits 0 and 1).

crates/synthesis/src/multi_controlled/mcx.rs
alexanderivrii: Done in [501c568](https://github.com/Qiskit/qiskit/pull/13929/commits/501c568d11ac88f731c4117755c0b67452c6b65d).

crates/synthesis/src/multi_controlled/mod.rs
alexanderivrii: Done in [501c568](https://github.com/Qiskit/qiskit/pull/13929/commits/501c568d11ac88f731c4117755c0b67452c6b65d).

crates/synthesis/src/multi_controlled/mcx.rs
alexanderivrii: This is a good idea.

test/python/transpiler/test_decompose.py
alexanderivrii: Done in [501c568](https://github.com/Qiskit/qiskit/pull/13929/commits/501c568d11ac88f731c4117755c0b67452c6b65d).

crates/synthesis/src/multi_controlled/mod.rs
```diff
@@ -10,11 +10,71 @@
...
 // that they have been altered from the originals.
 
+use mcx::{c3x, c4x, synth_mcx_n_dirty_i15, synth_mcx_noaux_v24};
 use pyo3::prelude::*;
+use qiskit_circuit::circuit_data::CircuitData;
 
 mod mcmt;
+mod mcx;
+
+/// Synthesize a multi-controlled X gate with :math:`k` controls based on the paper
```
Cryoris: It might be better not to duplicate the docs here: the Rust function is `synth_mcx_n_dirty_i15`, which is documented already, and the Python function is also documented already separately in the Python space 🙂

crates/synthesis/src/multi_controlled/mcx.rs
```diff
@@ -0,0 +1,438 @@
...
+///
+/// A quantum circuit with :math:`k + 1` qubits. The number of CX-gates is
+/// quadratic in :math:`k`.
+///
+/// # References
+///
+/// 1. Vale et. al., *Circuit Decomposition of Multicontrolled Special Unitary
+/// Single-Qubit Gates*, IEEE TCAD 43(3) (2024),
+/// [arXiv:2302.06377] (https://arxiv.org/abs/2302.06377).
+pub fn synth_mcx_noaux_v24(py: Python, num_controls: usize) -> PyResult<CircuitData> {
```
Cryoris, line 382: Does moving this into Rust actually make it _slower_? Effectively we've added an additional detour to Rust, because this synthesis code is calling Python and basically has no Rust logic?

crates/synthesis/src/multi_controlled/mcx.rs
```diff
@@ -0,0 +1,438 @@
...
+        let controls: Vec<u32> = (0..num_controls).map(|q| q as u32).collect();
+        let target = num_controls as u32;
+        let ancillas: Vec<u32> = ((num_controls + 1)..num_qubits).map(|q| q as u32).collect();
+
+        for j in 0..2 {
+            if !relative_phase {
+                circuit.compose(
+                    &ccx()?,
+                    &[
+                        Qubit(controls[num_controls - 1]),
```
Cryoris: Since you are a fan of short lines, you could maybe introduce variables for the qubits here 🙂

crates/synthesis/src/multi_controlled/mcx.rs
```diff
@@ -0,0 +1,438 @@
...
+        let mut circuit = CircuitData::with_capacity(num_qubits as u32, 0, 0, Param::Float(0.0))?;
+
+        let controls: Vec<u32> = (0..num_controls).map(|q| q as u32).collect();
+        let target = num_controls as u32;
+        let ancillas: Vec<u32> = ((num_controls + 1)..num_qubits).map(|q| q as u32).collect();
+
+        for j in 0..2 {
+            if !relative_phase {
+                circuit.compose(
+                    &ccx()?,
```
Cryoris: This seems a bit overcomplicated, wouldn't it be nicer to just define `circuit.ccx` as method? This would also incur at less overhead since you don't have to fetch the definitions etc.

crates/synthesis/src/multi_controlled/mcx.rs
```diff
@@ -0,0 +1,438 @@
...
+                    &[
+                        Qubit(controls[num_controls - 1]),
+                        Qubit(ancillas[num_controls - 3]),
+                        Qubit(target),
+                    ],
+                    &[],
+                )?;
+            } else if j == 0 {
+                circuit.compose(
+                    &action_gadget()?,
```
Cryoris: Similar, would it be more efficient (and cleaner) to just have `action_gadget` be a function that appends the gates to the circuit? This way we don't have to define a new circuit data + compose, but can just add the gates in place.

crates/synthesis/src/multi_controlled/mcx.rs
alexanderivrii: That's a good question. Here are quick benchmarking results (on my laptop) for `synth_mcx_noaux_v24` and `synth_mcx_n_dirty_i15` with varying  numbers of controls, each run `1000` times (so for a single run you should divide the results by 1000).

Main branch:
```
synth_mcx_noaux_v24
num_controls = 1, 0.02614617
num_controls = 2, 0.02418780
num_controls = 3, 0.09712219
num_controls = 4, 0.07472587
num_controls = 5, 0.02642417
num_controls = 6, 0.02766180
num_controls = 7, 0.02871490
num_controls = 8, 0.02915502
num_controls = 9, 0.03014803

synth_mcx_n_dirty_i15
num_controls = 1, 0.00988078
num_controls = 2, 0.01988816
num_controls = 3, 0.09700394
num_controls = 4, 0.11953712
num_controls = 5, 0.17409587
num_controls = 6, 0.22715592
num_controls = 7, 0.28122020
num_controls = 8, 0.33953714
num_controls = 9, 0.39850092
```
This PR:
```
synth_mcx_noaux_v24
num_controls = 1, 0.01181006
num_controls = 2, 0.00946403
num_controls = 3, 0.00682592
num_controls = 4, 0.01892591
num_controls = 5, 0.00957179
num_controls = 6, 0.01008511
num_controls = 7, 0.00939298
num_controls = 8, 0.01013231
num_controls = 9, 0.00996304

synth_mcx_n_dirty_i15
num_controls = 1, 0.00527406
num_controls = 2, 0.00584817
num_controls = 3, 0.00750637
num_controls = 4, 0.02089381
num_controls = 5, 0.03613806
num_controls = 6, 0.03370810
num_controls = 7, 0.03848481
num_controls = 8, 0.04513884
num_controls = 9, 0.05158806
```
So the Rust code is faster even for `synth_mcx_noaux_v24`, I am guessing that we are saving the runtime by circuit manipulations in Rust (even if the MCPhase gate is in Python).

crates/synthesis/src/multi_controlled/mod.rs
alexanderivrii: That's a good point, I will remove the duplicate documentation. Update: done in [71870d2](https://github.com/Qiskit/qiskit/pull/13929/commits/71870d22c6276ba9cf7438f44d748c8c0dfb6e49).

crates/synthesis/src/multi_controlled/mcx.rs
alexanderivrii: That's a good point. Since `ccx` and `rccx` are very common subcircuits, it does make perfect sense to add the corresponding methods to the ever-growing `CircuitDataForSynthesis` trait. I do prefer however to add _the definitions_ for CCX and RCCX rather than the gates themselves; while both would be valid options, the current way avoids an extra recursion step in `HighLevelSynthesis`. Update: done in [71870d2](https://github.com/Qiskit/qiskit/pull/13929/commits/71870d22c6276ba9cf7438f44d748c8c0dfb6e49).

crates/synthesis/src/multi_controlled/mcx.rs
alexanderivrii: After replacing the line above by `circuit.ccx(...)` as you suggested later, the line is already quite short. I do like to see the qubit indices explicitly, so I am not sure I want to hide them behind variable names.

crates/synthesis/src/multi_controlled/mcx.rs
Cryoris: Huh interesting, I would not have expected that. But it looks like the runtime for the no-aux case is essentially constant so maybe the few checks we do are just sufficiently faster in Rust 🤔 All good then, thanks for checking!

crates/synthesis/src/multi_controlled/mcx.rs
Cryoris: Unless someone wants to translate to Toffolis... but without any proof that we need this, more speed is more good 🙂

crates/synthesis/src/multi_controlled/mcx.rs
alexanderivrii: I have simplified the code in [c00ea29](https://github.com/Qiskit/qiskit/pull/13929/commits/c00ea294ad9e26428dbf46809f99674197a961ea). I would still like to have the original functions `action_gadget` and `reset_gadget` as I believe I have seen these subcircuits in multiple papers (if I understand correctly, this is an even further way to optimize constructions involving pairs of relative Toffoli gates), and, as you have suggested, I have added functions `add_action_gadget` and `add_reset_gadget`. Alternatively we `could` these methods to the `CircuitDataForSynthesis` trait as we have done for `ccx` and `rccx` gates, however I don't think these subcircuits deserve their special methods (and in fact you have not suggested this either, presumably for the same reason).

crates/synthesis/src/multi_controlled/mcx.rs
```diff
@@ -0,0 +1,430 @@
...
+    }
+
+    /// Appends the decomposition of the CCX to the circuit.
+    fn ccx(&mut self, q1: u32, q2: u32, q3: u32) -> PyResult<()> {
+        self.compose(&ccx()?, &[Qubit(q1), Qubit(q2), Qubit(q3)], &[])
+    }
+
+    /// Appends RCCX to the circuit.
+    fn rccx(&mut self, q1: u32, q2: u32, q3: u32) -> PyResult<()> {
+        self.compose(&rccx()?, &[Qubit(q1), Qubit(q2), Qubit(q3)], &[])
```
Cryoris, line 169: This gives us concise syntax, but I actually meant that I expect `push_standard_gate` to be faster than creating+composing a new `CircuitData` object 🙂

crates/synthesis/src/multi_controlled/mcx.rs
```diff
@@ -0,0 +1,430 @@
...
+        &rc3x()?.inverse()?,
+        &[Qubit(0), Qubit(1), Qubit(2), Qubit(3)],
+        &[],
+    )?;
+    circuit.compose(&c3sx()?, &[Qubit(0), Qubit(1), Qubit(2), Qubit(4)], &[])?;
+    Ok(circuit)
+}
+
+/// A block in the `action part`, see Iten et al.
+fn action_gadget() -> PyResult<CircuitData> {
```
Cryoris: Could this be moved into `add_action_gadget`, since this function is not used as standalone? That would also skip the additional creation of a `CircuitData`

alexanderivrii: Fair enough, while I think that these functions might become useful in the future, since I don't really need them right away,I have removed them in [7074f5b](https://github.com/Qiskit/qiskit/pull/13929/commits/7074f5b75119d5172df4ba4f9dfb66454b411589).

crates/synthesis/src/multi_controlled/mcx.rs
alexanderivrii: I see what you mean. However, I do prefer to have a single source of truth for the definitions of standard gates including CCX and RCCX, without having to explicitly reimplement it.

crates/synthesis/src/multi_controlled/mcx.rs
```diff
@@ -0,0 +1,416 @@
...
+#[pyfunction]
+pub fn c3x() -> PyResult<CircuitData> {
+    StandardGate::C3X
+        .definition(&[])
+        .ok_or(CircuitError::new_err(
+            "Error extracting the definition of C3X",
+        ))
+}
+
+/// Definition circuit for RC3X.
```
ShellyGarion: in the comment it should be RCCX

crates/synthesis/src/multi_controlled/mcx.rs
```diff
@@ -0,0 +1,416 @@
...
+/// Adds gates of the "action gadget" to the circuit
+fn add_action_gadget(circuit: &mut CircuitData, q0: u32, q1: u32, q2: u32) {
+    circuit.h(q2);
+    circuit.t(q2);
+    circuit.cx(q0, q2);
+    circuit.tdg(q2);
+    circuit.cx(q1, q2);
+}
+
+/// Adds gates of the "rest gadget" to the circuit
```
ShellyGarion: "rest gadget" -> "reset gadget"

ShellyGarion (commented):
The code looks great, I only found a few misprints.

Cryoris (approved):
LGTM but Iet's wait to see if @ShellyGarion or @raynelfss have other comments 👍🏻

raynelfss (approved):
LGTM! Thank you for the additions

PR #13934, opened 2025-02-28
Embed C API into Python extension

mtreinish:
### Summary

This commit adds the C extension crate into the Python extension build.
This is done by adding the qiskit-cext crate as a dependency for
qiskit-pyext. Then the contents of cext are re-exported from pyext. This
enables linking against the _accelerate shared library object using the
C API. This facilitates writing compiled Python extensions that
interface with Qiskit directly via C. For example, building a
SparseObservable object in a compiled language via the C API and then
returning that to Python for use with the rest of Qiskit.

### Details and comments

~This PR is based on top of #13694 and will need to be rebased once that merges. To view the contents of just this PR you can look at the HEAD commit on this PR branch: https://github.com/Qiskit/qiskit/pull/13934/commits/f2aa4742a284e0312fcc74e3c2b4a2959bab4c84~


crates/cext/src/sparse_observable.rs
```diff
@@ -847,3 +851,22 @@ pub extern "C" fn qk_bitterm_label(bit_term: BitTerm) -> u8 {
...
+pub unsafe extern "C" fn qk_obs_to_python(obs: *const SparseObservable) -> *mut PyObject {
+    let obs = unsafe { const_ptr_as_ref(obs) };
+    let py_obs: PySparseObservable = obs.clone().into();
+
+    Python::with_gil(|py| {
+        Py::new(py, py_obs)
+            .expect("Unable to create a Python object")
+            .into_ptr()
+    })
+}
```
jakelishman: I think we should feature gate this called something like `c-python`, and then similarly in the header file, have it generated into the header file with import guards around it, like
```c
// Up top of the file:
#ifdef QISKIT_C_PYTHON_INTERFACE
#  include <Python.h>
#13928

// ... other stuff ...

#ifdef QISKIT_C_PYTHON_INTERFACE
PyObject *qk_obs_to_python(SparseObservable *);
#endif
```
because having the _library_ built with the Python components (which the feature gate controls) doesn't guarantee you have the Python header files to include at build time.

The library always has to have Python components available at build time (for now), so the library-space feature gate is maybe premature.  I think the header include guards are fairly critical, though - it should be possible to link against our C API without needing `Python.h` available to you.

crates/pyext/src/lib.rs
```diff
@@ -11,6 +11,8 @@
 // that they have been altered from the originals.
 
 use pyo3::prelude::*;
+pub use qiskit_cext::sparse_observable::*;
+pub use qiskit_cext::*;
```
jakelishman: _Especially_ with `no_mangle` everywhere, I don't think we should be reaching into internal modules of another crate like this.  With the short-term intent of `qiskit_cext` being that it will be imported from another Rust crate (i.e. this), it should sort out its namespace so everything it wants available to `pub use` is just in the root, or in something like `qiskit_cext::api::*` (the latter in the case that there might be public functions that _don't_ want to be exposed generally).

qiskit/__init__.py
```diff
@@ -131,6 +131,20 @@
...
+    """Get the path to the compiled library object.
+
+    When interfacing with Qiskit's C API for a Python extension the shared library object
+    returned from this function must be what is linked against. This will ensure that all
+    objects are built using the same library file.
+
+    Returns:
+        The path.
+    """
+    return _accelerate.__file__
```
jakelishman: I think we should include a lot more comment on the difficulties with the build (the compiled object is only valid with this exact build of Qiskit, etc).

I'm not clear that this is the best interface, and I'm a bit concerned that this makes it de facto _Python_ public interface. Typically we need the user to figure out the CFLAGS they need, and this just kind of wiggles its eyes in the right direction, right.  The main things the user needs to know are:
- directory to add to the system include search path
- directory to add to the library search path
- name of the library to link against.

In the model proposed by this PR, the `qiskit.h` header file isn't present in the Python distribution, so there's no option for 1.  `get_lib` doesn't return something that's easily given to an LDFLAG (or at least not in any portable way - GNU `ld` has the `-l:<path>` form, but that doesn't work on Mac).  That said, linking against `_accelerate.abi3.so` is a nuisance on Mac in the first place, because the linkers are all configured to look for files called `libX.dylib`, and `.so` doesn't fit the bill (not to mention the missing `lib` prefix).

If we're going to go as far as adding code to the Python tooling to support this, I think we also ought to couple that with distributing the `qiskit.h` header file in some `/include` directory somewhere in the installed file.  I'd perhaps rather we just document how to do this somewhere separately with the C API, though (e.g. `gcc -Iwherever -l:$(python -c 'import qiskit; print(qiskit._accelerate.__file__')`), rather than squeezing a Python-space change this close to 2.0.

jakelishman (commented):
I think building this into the Python-space object is fine, but I don't think we should modify `qiskit/__init__.py`, given that I think this way of suggesting the linkage is not our intended long-term solution, and in the short term it's really fragile.

There's _so much_ fiddling you need to do compared to a normal build process to build an executable against this current version, which is fine for our MVP, but I think we need a lot more guide on how to make it work, the caveats of it all, and I think it oughtn't to be visible to Python users.

jakelishman:
For posterity: I was able to use this PR to build the test executable in `crates/cext/README.md` on my Raspberry Pi (which is Linux with a GNU build chain).  The input file `obs.c`:

```c
#include <Python.h>
#include <complex.h>
#include <qiskit.h>
#include <stdint.h>
#include <stdio.h>

int main(int argc, char *argv[]) {
    // build a 100-qubit empty observable
    uint32_t num_qubits = 100;
    QkSparseObservable *obs = qk_obs_zero(num_qubits);

    // add the term 2 * (X0 Y1 Z2) to the observable
    complex double coeff = 2;
    QkBitTerm bit_terms[3] = {QkBitTerm_X, QkBitTerm_Y, QkBitTerm_Z};
    uint32_t indices[3] = {0, 1, 2};
    QkSparseTerm term = {coeff, 3, bit_terms, indices, num_qubits};
    qk_obs_add_term(obs, &term);

    // print some properties
    printf("num_qubits: %u\n", qk_obs_num_qubits(obs));
    printf("num_terms: %lu\n", qk_obs_num_terms(obs));

    // free the memory allocated for the observable
    qk_obs_free(obs);

    return 0;
}

```

Note that I've inserted `#include <Python.h>` up top, even though I don't use it in the source file, just to account for the problem in `qiskit.h`.

I built Qiskit just with `python setup.py build_rust --inplace`.

Then the build flags:
```bash
# Where the root of my Qiskit checkout is.
QISKITROOT="$HOME/code/qiskit/terra"
# Where the root of my Python dev components are.
PYROOT="$HOME/code/pyenv/versions/3.13.2"

CFLAGS="-I$QISKITROOT/dist/c/include -I$PYROOT/include/python3.13"
LDFLAGS="-Wl,-rpath,$QISKITROOT/qiskit,-rpath,$PYROOT/lib -L$QISKITROOT/qiskit -l:_accelerate.cpython-313-aarch64-linux-gnu.so -L$PYROOT/lib -lpython3.13"

gcc $CFLAGS obs.c -o obs $LDFLAGS
```

I inserted the libraries in to the rpath during linking, but you could equally choose to insert `libpython3.13.so` at runtime with `LD_LIBRARY_PATH`.

I haven't managed to work out the incantations you need to get the Apple linker to use a library file with an "incorrect" name.

qiskit/__init__.py
mtreinish: I'm fine reverting this, I was on the fence about adding it. I don't like the story about linking here either, I decided to add it because I figured I'd come up with something more clever based around `dlopen()` in 2.1 and would want an api to get a path. But I agree it's trivial enough to figure out if you need it.

crates/pyext/src/lib.rs
mtreinish: I think longer term that is definitely the direction we should move. For this first PR with the small API this seemed fine since there is only one thing in there now. I fully expect as we go into 2.1 we'll want more structure around the re-export between the crates.

qiskit/__init__.py
mtreinish: Removed in: https://github.com/Qiskit/qiskit/pull/13934/commits/fc211a869bd51e55bc2bf75f5dca6a359253ef2d

crates/cext/src/sparse_observable.rs
mtreinish: I added a feature gate and corresponding cbindgen config in: https://github.com/Qiskit/qiskit/pull/13934/commits/b27569347c82fd41caf8b9875540ed6a0aa0e00a I didn't do anything for adding the `Python.h` include though, since I still need to read up on cbindgen to see how best to do that.

crates/cext/cbindgen.toml
```diff
@@ -24,8 +25,9 @@ prefix_with_name = true
 
 [export]
 prefix = "Qk"
-renaming_overrides_prefixing = false
+renaming_overrides_prefixing = true
 
 [export.rename]
 "SparseObservable" = "Obs"
 "CSparseTerm" = "ObsTerm"
+"PyObject" = "PyObject"
```
jakelishman: I think you need to change `Obs` and `ObsTerm` if you change this?

crates/cext/src/sparse_observable.rs
jakelishman: Yeah, I was about to leave a review comment saying that it's still not generated.  We need to include it in the header somehow or another, even if it's just hardcoded in at the same place as the `QkComplex64` injections (behind the import guard).

crates/cext/cbindgen.toml
mtreinish: Reverted in: https://github.com/Qiskit/qiskit/pull/13934/commits/4938258dadf296019664cb2397bfecf2dd26b31b

crates/cext/src/sparse_observable.rs
mtreinish: Added in: https://github.com/Qiskit/qiskit/pull/13934/commits/4938258dadf296019664cb2397bfecf2dd26b31b

crates/cext/cbindgen.toml
```diff
@@ -1,19 +1,24 @@
 include_version = true
-style = "type" 
+style = "type"
 sys_includes = ["complex.h"]
 after_includes = """
+#ifdef QISKIT_C_PYTHON_INTERFACE
+    #include<Python.h>
+#endif
```
jakelishman: ludicrously minor: `#include <Python.h>` wants a space between the `include` and the `<` haha.

crates/cext/src/sparse_observable.rs
```diff
@@ -834,3 +838,27 @@ pub extern "C" fn qk_bitterm_label(bit_term: BitTerm) -> u8 {
...
```
jakelishman: A C caller must hold the GIL in order to call this function.  I think we could document it in the safety notes, and then use PyO3's `assume_gil` (or whatever it's called) - we need to make sure that C space understands that it must hold the GIL.

crates/pyext/Cargo.toml
```diff
@@ -28,3 +28,4 @@ qiskit-accelerate.workspace = true
 qiskit-circuit.workspace = true
 qiskit-qasm2.workspace = true
 qiskit-qasm3.workspace = true
+qiskit-cext = { workspace = true, features = ["python_binding"] }
```
jakelishman, line 31: This change made me look again at the Makefile, and I realised there's a couple of mistakes in it in the C build.  I can push them up to this PR, if it's easier?

crates/cext/src/sparse_observable.rs
```diff
@@ -14,7 +14,11 @@ use std::ffi::{c_char, CString};
 
 use crate::exit_codes::{CInputError, ExitCode};
 use num_complex::Complex64;
-use qiskit_accelerate::sparse_observable::{BitTerm, SparseObservable, SparseTermView};
+use pyo3::ffi::PyObject;
+use pyo3::{Py, Python};
+use qiskit_accelerate::sparse_observable::{
+    BitTerm, PySparseObservable, SparseObservable, SparseTermView,
+};
```
jakelishman: All the `Py` stuff here is only used in the `cfg("python_binding")` paths, but we don't see a lint failure because we only run the entire workspace lint, which forcibly enables that feature via `pyext`.  Can we protect the imports for now, so `cargo build -p qiskit-cext` passes build and lint alone, and maybe think of a story for checking that going forwards?

crates/cext/cbindgen.toml
```diff
@@ -29,3 +34,4 @@ renaming_overrides_prefixing = false
 [export.rename]
 "SparseObservable" = "Obs"
 "CSparseTerm" = "ObsTerm"
+"PyObject" = "PyObject"
```
jakelishman: I think `PyObject` is misnamed now, with `renaming_overrides_prefixing` back to `false`.  It would be nice if there was a cleaner way to ask `cbindgen` to just treat a type as "already defined", rather than us needing to hack it in.

crates/pyext/Cargo.toml
mtreinish: Yeah feel free to push those up to this PR directly that's fine.

crates/cext/src/sparse_observable.rs
mtreinish: You're right this should have the cfg around the python bits. But I think we do have coverage for `qiskit-cext` standalone builds via the c tests already we shouldn't be building via pyext there or doing a whole workspace build.

crates/pyext/Cargo.toml
jakelishman: Done in b25868cd7f

crates/cext/src/sparse_observable.rs
jakelishman: The action workflow uses `make`, which was building the entire workspace before now.  One of the changes I made to it was to only build the necessary components - I hadn't noticed before when I was looking at the file that it didn't do that.  Fixing it is how I noticed these unused imports - I get the warnings when I build now!

(Btw, in the future we should have the Makefile be configurable so you can choose to build the Python interface using the Makefile, because at the moment it just builds `qiskit-cext` with its default features.)

crates/pyext/src/lib.rs
Cryoris: Side note here, I'm not sure the first line here is necessary, I think I forgot to remove it once everything was up and running 😄

crates/cext/cbindgen.toml
Cryoris: Hm yes I'm not sure this is possible with `cbindgen`... but given that `PyObject` is available through `Python.h`, which we include if the Python binding is enabled, couldn't we just add `PyObject` to the exclude list? I.e. add
```toml
[export]
exclude = ["PyObject"]
```
and remove it from the renaming section. That should avoid cbindgen from doing any typedefs/parsings/renamings.

jakelishman: Julien: if that works, that's what I was looking for.

Cryoris: Oh fair

mtreinish: Done in https://github.com/Qiskit/qiskit/pull/13934/commits/b92f156d9293655df2757ff11e9c8696da544185 per @Cryoris's suggestion

crates/cext/src/sparse_observable.rs
mtreinish: Added in: https://github.com/Qiskit/qiskit/pull/13934/commits/547904c244229d1597e34a81334e8012da11a281

crates/cext/cbindgen.toml
mtreinish: Fixed in: https://github.com/Qiskit/qiskit/pull/13934/commits/b92f156d9293655df2757ff11e9c8696da544185

crates/cext/cbindgen.toml
Cryoris: Hm crap that doesn't seem to have worked as expected I'm still getting `QkPyObject` used in the header, it's just no longer defined

Cryoris: The manual, annoying solution would be to manually do the prefix by renaming all structs we introduce...

crates/cext/src/sparse_observable.rs
mtreinish: I did this in: https://github.com/Qiskit/qiskit/pull/13934/commits/fc274aa877e6176212d43027975714c1e9df2520

crates/pyext/src/lib.rs
mtreinish:
```suggestion
pub use qiskit_cext::*;
```

crates/cext/src/sparse_observable.rs
```diff
@@ -834,3 +842,32 @@ pub extern "C" fn qk_bitterm_label(bit_term: BitTerm) -> u8 {
...
+pub unsafe extern "C" fn qk_obs_to_python(obs: *const SparseObservable) -> *mut PyObject {
+    let obs = unsafe { const_ptr_as_ref(obs) };
+    let py_obs: PySparseObservable = obs.clone().into();
+
+    unsafe {
+        let py = Python::assume_gil_acquired();
+        Py::new(py, py_obs)
+            .expect("Unable to create a Python object")
+            .into_ptr()
+    }
```
jakelishman:
```suggestion
    // SAFETY: Per documentation, the pointer is non-null and aligned.
    let obs = unsafe { const_ptr_as_ref(obs) };
    let py_obs: PySparseObservable = obs.clone().into();

    // SAFETY: the C caller is required to hold the GIL.
    unsafe {
        let py = Python::assume_gil_acquired();
        Py::new(py, py_obs)
            .expect("Unable to create a Python object")
            .into_ptr()
    }
```

safety comments

crates/cext/cbindgen.toml
```diff
@@ -25,6 +30,7 @@ prefix_with_name = true
 [export]
 prefix = "Qk"
 renaming_overrides_prefixing = false
+exclude = ["PyObject"]
```
jakelishman: This doesn't seem to be working - the generated file still has the prefix.

I'm fine just using `renaming_overrides_prefixing = true` and manually adding `Qk` to the renames for now, while we figure out a nicer pattern.

Probably to do later (because of additional infrastructure), but I think we ought to consider adding a test of the Python-interfacing method.

Cryoris:
To add to the above: the current version with the manual renaming works (on Mac) and with the up-to-date `main` we can compile a Qiskit C Python extension that creates a `SparseObservable` and is recognized as Python `SparseObservable`. The test looked something like
```python
from qiskit import transpile
from qiskit.quantum_info import SparseObservable

import cextension  # C Python extension

if __name__ == "__main__":
    obs = cextension.build_observable()
    print("-- SparseObservable")
    print(obs)
    print(isinstance(obs, SparseObservable))  # True

    # build an evolution circuit
    evo = PauliEvolutionGate(obs, time=2)
    circuit = QuantumCircuit(obs.num_qubits)
    circuit.append(evo, circuit.qubits)

    tqc = transpile(circuit, basis_gates=["u", "cx"]) 
    print("-- Evolution circuit")
    print(tqc.draw(idle_wires=False))
```

The thing I had to change from Jake's workflow above it so symlink `_accelerate.<....>.so -> libqiskit.so`, since the Mac linker wasn't happy with the original library name.

jakelishman:
If you're actually building a completely custom C extension for Python, you shouldn't need to do the `-rpath` stuff I did to insert `libpython` at load time (or alternatively the `LD_LIBRARY_PATH` [Linux] or `DYLD_LIBRARY_PATH` [Mac] bits), because Python should arrange for those to be preloaded before loading an extension module.

On Mac (or more strictly, when using the LLVM linker toolchain), I couldn't work out a set of commands to use a library file with a bad name (i.e. other than `lib<library>.so` [Linux] or `lib<library>.dylib` [Mac]), so symlink seems to be the short-term way to go.

For `SparseObservable`, _strictly_, I think we'd see the same behaviour if we copied the `.so` file (not symlink), but I think that will be very fragile, because we'll have two address spaces.  That would break things like the uniqueness-tracking globals for `QuantumCircuit` (`Bit`, in particular) once they're referenced by C, and anyway, it'd be duplicating a load of stuff in the memory space.

crates/cext/Cargo.toml
```diff
@@ -17,10 +17,12 @@ workspace = true
 thiserror.workspace = true
 num-complex.workspace = true
 qiskit-accelerate.workspace = true
+pyo3.workspace = true
```
mtreinish: This needs to move to the feature so we only directly depend on pyo3 in pyyhon_binding mode

Cryoris: But implicitly we'd still depend on pyo3 through qiskit-accelerate, until we move the Rust-only code into a separate crate, right?

jakelishman: `cext` itself only depends on `pyo3` if the Python feature is active. It happens to have transitive dependencies that depend on `pyo3` anyway, but that's an implementation detail as far as it's concerned.

jakelishman (approved):
Thanks everyone. We need to add more documentation on this, but we can follow up after 2.0rc1.

PR #13921, opened 2025-02-24
Circuits with control flow now raise a proper error when `circuit_to_instruction` is called

1ucian0:
### Summary

Fixes #11379 

### Details and comments

Converting a quantum circuit to a gate with `circuit_to_instruction` now fails properly when given circuit contains control flow instructions.

PR #13936, opened 2025-02-28
Add method `has_control_flow_op()` to `QuantumCircuit`

raynelfss:
### Summary
This method was added to fix a performance regression introduced by #13921. This check should be quicker than the multiple calls to `isinstance()`.

### Details and comments

This new method for `QuantumCircuit` performs a fast path to determine if a circuit has a control flow operation present within. This is performed by using `QuantumCircuit.count_ops()` and checking if any of the names there belong to a `ControlFlowOp`.

Additionally, this method fixes an unintentional performance regression caused by #13921.


jakelishman:
I'm not sure we need to do this beyond the feature-freeze deadline. There's already no need to do all the `isinstance` checks - the original PR could have written
```python
any(instruction.is_control_flow() for instruction in circuit.data)
```

crates/circuit/src/circuit_data.rs
```diff
@@ -423,6 +423,15 @@ impl CircuitData {
         Ok(())
     }
 
+    /// Checks whether the circuit has an instance of :class:`.ControlFlowOp`
+    /// present amongst its operations.
+    pub fn has_control_flow_op(&self) -> bool {
+        let operations = self.count_ops();
+        CONTROL_FLOW_OP_NAMES
+            .iter()
+            .any(|name| operations.contains_key(name))
```
mtreinish: `count_ops()` will iterate over the circuit anyway. It'll be better to just do:

```suggestion
        self.data.iter().any(|x| x.control_flow())
```

crates/circuit/src/dag_circuit.rs
```diff
@@ -70,7 +70,8 @@ use std::f64::consts::PI;
 #[cfg(feature = "cache_pygates")]
 use std::sync::OnceLock;
 
-static CONTROL_FLOW_OP_NAMES: [&str; 4] = ["for_loop", "while_loop", "if_else", "switch_case"];
+pub(crate) static CONTROL_FLOW_OP_NAMES: [&str; 4] =
```
mtreinish: With the above comment we don't need this change.

```suggestion
static CONTROL_FLOW_OP_NAMES: [&str; 4] =
```

qiskit/circuit/quantumcircuit.py
```diff
@@ -4341,6 +4344,11 @@ def map_calibration(qubits, parameters, schedule):
         )
         return None if inplace else target
 
+    def has_control_flow_op(self) -> bool:
```
mtreinish, line 4826: @jakelishman's earlier comment is on point here this is past feature freeze so we shouldn't make this public if we want to add it for 2.0. There is a path to fix the regression in python as he suggested, but with my other comments here doing the iteration in rust is faster. We could also still iterate in rust if we did the set intersection with `count_ops()` python side (like is currently in the rust code).

mtreinish (commented):
Thanks for tackling this. I left a few comments inline. The biggest issue @jakelishman already pointed out that this shouldn't be a new public method for 2.0 so close to the release. There are paths to fixing the performance issue without adding a new method, we can add the public api in 2.1 though.

mtreinish:
In this interest of getting a fix for the regression included in 2.0 I pushed up: https://github.com/Qiskit/qiskit/pull/13942 and moved this to 2.1. We should definitely do this after 2.0 is released.

mtreinish (commented):
Overall this LGTM now, but could you add some tests to test this method explicitly?

test/python/circuit/test_circuit_operations.py
```diff
@@ -1550,6 +1551,82 @@ def instructions():
...
+                    )
+                num_qubits = true_body.num_qubits
+                num_clbits = true_body.num_clbits
+
+                super().__init__(
+                    "custom_ctrl_flow", num_qubits, num_clbits, [true_body, false_body], label=label
+                )
+                # We do not need to validate this custom condition
+                self.condition = condition
+                self._params = [true_body, false_body]
```
mtreinish: This is not supported. We don't allow for custom control flow operations to be defined like this. So we shouldn't have a test for this.

test/python/circuit/test_circuit_operations.py
```diff
@@ -1550,6 +1551,82 @@ def instructions():
         self.assertEqual(circuit, expected)
         self.assertEqual(circuit.name, "test")
 
+    def test_circuit_has_control_flow_op(self):
```
mtreinish, line 1584: I would expand this to cover a circuit with each control flow op type and make sure it's detected. The set of defined operations is here: https://github.com/Qiskit/qiskit/blob/main/qiskit/circuit/controlflow/__init__.py#L28

raynelfss: Addressed in 98c2e2c

test/python/circuit/test_circuit_operations.py
raynelfss: Addressed in 98c2e2c

test/python/circuit/test_circuit_operations.py
```diff
@@ -1576,6 +1581,55 @@ def instructions():
...
+        # Use combinatorics to test all combinations of control flow operations
+        # to see if we can detect all of them.
+        for op_combination in op_combinations:
+            # Build a circuit
+            circ = QuantumCircuit(2, 2)
+            circ.h(0)
+            circ.cx(0, 1)
+            circ.measure(0, 0)
+            circ.measure(1, 1)
+            self.assertTrue(not circ.has_control_flow_op())
```
mtreinish:
```suggestion
            self.assertFalse(circ.has_control_flow_op())
```

mtreinish (approved):

PR #13937, opened 2025-02-28
Fix: Add `legacy_cmaps` to `python.benchmarks`

raynelfss:
### Summary
In a previous oversight, a beyond-local-level relative import was preventing us from running the asv suite locally. Due to the nature of asv, we cannot import these coupling maps from where the file is located. The following commits add a file into the local level of ASV benchmarks which contains the necessary legacy coupling maps.


jakelishman (approved):
thanks!

PR #13938, opened 2025-02-28
Remove QPY generation venvs more eagerly

jakelishman:
Previously, the venv was only tidied up at the end of the script.  We now have enough published versions of Qiskit that if a script fails its QPY runs (and so exited before cleaning up its venv), we could overflow the disk space allocated to the VM, and get no output from the job.

### DIFF
--- a/test/qpy_compat/process_version.sh
+++ b/test/qpy_compat/process_version.sh
@@ -52,7 +52,9 @@ if [[ ! -d $cache_dir ]] ; then
     mkdir -p "$cache_dir"
     pushd "$cache_dir"
     echo "Generating QPY files with $package==$version"
-    "$venv_dir/bin/python" "${our_dir}/test_qpy.py" generate --version="$version"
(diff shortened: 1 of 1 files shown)


jakelishman:
Temporarily on hold while the DNM commit is in the branch, testing that the QPY run fails, but successfully reports the error state in generation, rather than getting killed for saturating the disk.

jakelishman:
Ah of course, this doesn't actually help if we error out during the environment generation.  Need to rethink slightly - I guess the logic needs tweaking a bit.

jakelishman:
Ok, all good.  You can see a run from a dummy commit that caused generation to fail: https://github.com/Qiskit/qiskit/actions/runs/13598863905/job/38021435278.  Unlike other recent PRs that were failing generation, we now no longer overflow the disk space, so we get the proper logs back.

ElePT (approved):
LGTM. Thanks!

PR #13940, opened 2025-03-01
Restore HAS_TWEEDLEDUM optional checker

mtreinish:
### Summary

In #13815 the usage of the tweedledum library was removed from Qiskit. The classes that used this functionality was deprecated in 1.4.0, but the optional library checker was not deprecated. These optionals checkers are part of the public API and would have needed a deprecation before removal as there may be downstream users of the checker. This commit restores the checker, but alters the docstring to make it clear that tweedledum isn't used by Qiskit anymore. If we want to remove the checker in 3.0 we can deprecate it in 2.1.0 or another release in the 2.x series. Although, there is basically no real cost for keeping it around so it's probably not worth the effort.

### DIFF
--- a/qiskit/utils/optionals.py
+++ b/qiskit/utils/optionals.py
@@ -183,6 +183,12 @@ External Python Libraries
     `testtools <https://pypi.org/project/testtools/>`__ library is installed.  This is generally
     only needed for Qiskit developers.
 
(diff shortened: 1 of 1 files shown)


jakelishman (approved):
we still have `HAS_IGNIS` haha - no hurry to get rid of them

PR #13945, opened 2025-03-03
Fix composing virtual permutation layouts for `ElidePermutations` and `StarPreRouting`

gadial:
### Summary [Updated]

This PR fixes how virtual permutation layouts are updated within the `ElidePermutations` and `StarPreRouting` transpiler passes.

### Details and comments [Updated]

We have several passes that apply to a virtual (logical) circuit and remove or insert swap or permutation gates while also updating the implicit "final" permutation at the end of the circuit. These are: `ElidePermutations`, `Split2QUnitaries` (with `split_swap=True`), and `StarPreRouting`.

At the time this PR was originally written by Gadi, we had problems both in how these implicit permutations are updated when running multiple passes from the list above, and more importantly how these implicit final permutation are combined with layout and routing passes. Fortunately, the second problem has now been long solved by Jake, in his overhaul of `SetLayout` and `ApplyLayout`. However, we still had some bugs related to incorrectly taking the inverse of a final layout when composing these within `ElidePermutations` and `StarPreRouting` passes. Interestingly, `Split2QUnitaries` composed the final layouts correctly. The problems did not really manifest because we do not run `StarPreRouting` in the default transpilation pipeline, and because we always run `ElidePermutations` _before_ `Split2QUnitaries`, in which case the two passes set/update virtual layouts correctly. The problems would manifest, however, by changing the order of these two passes or by throwing `StarPreRouting` into the mix.

Edit: there is also `Commuting2qGateRouter` that incorrectly updates "virtual_final_layout", however (from the pass description) it is used in a very special compilation flow consisting of `FindCommutingPauliEvolutions`, `Commuting2qGateRouter`, `SetLayout`,  which seems to have other problems. This PR does not attempt to fix problems in  `Commuting2qGateRouter` and friends.

---

As part of the testing, a small bug in `ElidePermutations` surfaced; namely, composing the inverse of the current layout on the new layout instead of simply composing the current layout. A small example reproducing the bug is

```python
from qiskit.transpiler.passes.optimization.split_2q_unitaries import Split2QUnitaries
from qiskit.transpiler.passes.optimization.elide_permutations import ElidePermutations

unitary_swap = QuantumCircuit(2)
unitary_swap.swap(0, 1)
qc = QuantumCircuit(5)
qc.swap(0,1)
qc.append(UnitaryGate(Operator(unitary_swap)), (0,2))
qc.swap(0,3)
qc.append(UnitaryGate(Operator(unitary_swap)), (0,4))

pm = PassManager()
pm.append(Split2QUnitaries(split_swap=True))
pm.append(ElidePermutations())

res = pm.run(qc)
print(res.layout.final_index_layout())
```

Which for the old version of the code results in `[3, 0, 4, 2, 1]`. However, when using `qc.swap(0,2)` and `qc.swap(0,4)` instead of unitary swaps, the result is `[4, 0, 1, 2, 3]`, as can be expected.

### AI/LLM disclosure

- [x] I didn't use LLM tooling, or only used it privately.
- [ ] I used the following tool to help write this PR description:
- [ ] I used the following tool to generate or modify code:


test/python/transpiler/test_split_2q_unitaries.py
```diff
@@ -377,3 +384,49 @@ def test_2q_swap_with_large_circuit(self):
...
+        else:
+            pm.append(Split2QUnitaries(split_swap=True))
+            pm.append(ElidePermutations())
+
+        res = pm.run(qc_split)
+        self.assertEqual(expected_layout, res.layout.final_index_layout())
+
+        res_op = Operator.from_circuit(res)
+        expected_op = Operator(qc_split)
+        self.assertTrue(expected_op.equiv(res_op))
```
ShellyGarion, line 431: why do you check here that they are equivalent (=up to a global phase)?
I think that this equality check should work:
`self.assertEqual(expected_op, res_op)`

ShellyGarion (commented):
Thanks @gadial for finding this bug and fixing it. See my comment.

test/python/transpiler/test_split_2q_unitaries.py
gadial: I think I followed suit with other tests, but if your line works I can switch to it.

ShellyGarion: Indeed, I see that many tests have these two lines:
```
        self.assertTrue(Operator(qc).equiv(qc_split))
        self.assertTrue(
            matrix_equal(Operator(qc).data, Operator(qc_split).data, ignore_phase=False)
        )
```
namely, they check that the operators are the same (up to a global phase) and then also check that the matrices are the same (including the global phase). This does not make sense to me.
I think these two lines can be replaced everywhere by this line that asserts that the matrices are the same:
`self.assertEqual(Operator(qc), Operator(qc_split))`

gadial: I'm ok with it, we can do it as part of this PR, if the others agree.

alexanderivrii:
Thanks @gadial for doing this. We have similar (and possibly also incorrect) update to `"virtual_permution_layout"` in `star-prerouting`: https://github.com/Qiskit/qiskit/blob/f82aa0ac17ae63b224a33e210e8f142be2081113/qiskit/transpiler/passes/routing/star_prerouting.py#L283-L288. Could you please check if a similar fix is needed there as well?

gadial:
Further discussions revealed that we currently lack in qiskit a canonical interpretation of `final_index_layout`, so we postpone this fix to 2.1 until the larger issue is resolved.

alexanderivrii:
With Gadi's permission, I have taken over this old PR, also including a fix for `StarPreLayout` and updating the tests, the name of the PR and the PR summary (see it for the updated description). As far as I can tell, composing virtual permutation layouts now works correctly for any combination of the 3 passes that modify these (Elide, Split and Star).

test/python/transpiler/test_split_2q_unitaries.py
alexanderivrii: Done as a part of [46f7540](https://github.com/Qiskit/qiskit/pull/13945/commits/46f75406db34cff4dfc01e0aefd7154357663823). Now we check equality of operators (or rather of `Operator.from_circuit`).

test/python/transpiler/test_split_2q_unitaries.py
```diff
@@ -18,6 +18,7 @@
 
 from math import pi
 from test import QiskitTestCase
+import ddt
```
ShellyGarion: why was ddt added here? I don't see any parallel tests

test/python/transpiler/test_split_2q_unitaries.py
```diff
@@ -32,6 +33,7 @@
 from qiskit import qpy
 
 
+@ddt.ddt
```
ShellyGarion: why was ddt added here? I don't see any parallel tests

test/python/transpiler/test_virtual_permution_layout.py
```diff
@@ -0,0 +1,118 @@
...
+
+        res = pm.run(qc_split)
+        self.assertEqual(expected_layout, res.layout.final_index_layout())
+
+        res_op = Operator.from_circuit(res)
+        expected_op = Operator(qc_split)
+        self.assertEqual(expected_op, res_op)
+
+    def test_all_combinations_of_split2q_swap_and_elide(self):
+        """Test all possible combinations of passes the set or update virtual_permutation_layout."""
```
ShellyGarion: this sentence should be rewritten

ShellyGarion (commented):
Thanks for picking this up and fixing it.
I like the tests.
I only have very minor comments.

test/python/transpiler/test_split_2q_unitaries.py
alexanderivrii: Good catch, this change is no longer needed since the test was moved to a separate file. Done in [9580861](https://github.com/Qiskit/qiskit/pull/13945/commits/9580861c5629fa406cd39e0f9fb08f6686a42e61).

test/python/transpiler/test_split_2q_unitaries.py
alexanderivrii: Good catch, this change is no longer needed since the test was moved to a separate file. Done in [9580861](https://github.com/Qiskit/qiskit/pull/13945/commits/9580861c5629fa406cd39e0f9fb08f6686a42e61).

test/python/transpiler/test_virtual_permution_layout.py
alexanderivrii: Typo: the -> that. Done in [58196bb](https://github.com/Qiskit/qiskit/pull/13945/commits/58196bba924543792a3df2f4eec80fab000d7904).

ShellyGarion (dismissed):

test/python/transpiler/test_virtual_permution_layout.py
```diff
@@ -0,0 +1,118 @@
...
+            [(0, 1), (1, 0)],
+            [(0, 1), (0, 2), (0, 3), (0, 4)],
+            [(0, 3), (5, 6), (2, 3), (1, 5), (2, 1)],
+            [(0, 5), (2, 1), (1, 2), (6, 0), (3, 2), (4, 2), (0, 1), (4, 2), (1, 3), (6, 4)],
+        ],
+        elide_before=[True, False],
+        seed=[42, 13, 55],
+    )
+    def test_split2q_and_elide(self, swap_list, elide_before, seed):
+        """Tests that  of 2q-swap with the elide permutations pass"""
```
ShellyGarion:
```suggestion
        """Tests that 2q-swap with the elide permutations pass"""
```

ShellyGarion (approved):

PR #13950, opened 2025-03-03
Add `qk_obs_compose(_map)`

Cryoris:
### Summary

Adds composition for the C-side `SparseObservable`. 

Goes on top of #13766.

### Details and comments

* No reno as C API not yet released
* Is the pointer-reading in `compose_map` legit?

### DIFF
--- a/crates/cext/src/sparse_observable.rs
+++ b/crates/cext/src/sparse_observable.rs
@@ -632,6 +632,84 @@ pub unsafe extern "C" fn qk_obs_add(
 }
 
 /// @ingroup QkObs
+/// Compose (multiply) two observables.
+///
+/// @param right One observable.
+/// @param left The other observable.
+///
+/// @return ``right.compose(left)`` which equals the observable ``result = left @ right``,
+///     in terms of the matrix multiplication ``@``.
+///
+/// # Example
+///
+///     QkObs *left = qk_obs_identity(100);
+///     QkObs *right = qk_obs_zero(100);
+///     QkObs *result = qk_obs_compose(right, left);
+///
+/// # Safety
+///
+/// Behavior is undefined if ``left`` or ``right`` are not valid, non-null pointers to
+/// ``QkObs``\ s.
+#[no_mangle]
+#[cfg(feature = "cbinding")]
+pub unsafe extern "C" fn qk_obs_compose(
+    right: *const SparseObservable,
+    left: *const SparseObservable,
+) -> *mut SparseObservable {
+    // SAFETY: Per documentation, the pointers are non-null and aligned.
+    let right = unsafe { const_ptr_as_ref(right) };
+    let left = unsafe { const_ptr_as_ref(left) };
+
+    let result = right.compose(left);
+    Box::into_raw(Box::new(result))
+}
+
+/// @ingroup QkObs
+/// Compose (multiply) two observables according to a custom qubit order.
+///
+/// Notably, this allows composing two observables of different size.
+///
+/// @param right One observable.
+/// @param left The other observable. The number of qubits must match the length of ``qargs``.
+/// @param qargs The qubit arguments specified which indices in ``right`` to associate with
+///     the ones in ``left``.
+///
+/// @return ``right.compose(left)`` which equals the observable ``result = left @ right``,
+///     in terms of the matrix multiplication ``@``.
+///
+/// # Example
+///
+///     QkObs *left = qk_obs_identity(100);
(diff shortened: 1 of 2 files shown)


crates/cext/src/sparse_observable.rs
```diff
@@ -631,6 +631,84 @@ pub unsafe extern "C" fn qk_obs_add(
...
+/// @param left The other observable.
+///
+/// @return ``right.compose(left)`` which equals the observable ``result = left @ right``,
+///     in terms of the matrix multiplication ``@``.
+///
+/// # Example
+///
+///     QkObs *left = qk_obs_identity(100);
+///     QkObs *right = qk_obs_zero(100);
+///     QkObs *result = qk_obs_compose(right, left);
```
jakelishman: Can we use some different names to `left` and `right`?  This is making an analogy to matrix multiplication, but then gets you into the weird situation where the signature is `f(right, left)`.  Since they're observables anyway, and it's pretty vital that `SparseObservable` is a matrix-free representation (!) perhaps `first` and `second` are better names: the result of `compose` is the observable equivalent to applying `first` followed by applying `second`.

If/when we add `dot` (which was a complete oversight in #13766), then `left` and `right` would be suitable names, since the entire method of `dot` only exists by analogy to matrix multiplication.

crates/cext/src/sparse_observable.rs
```diff
@@ -631,6 +631,84 @@ pub unsafe extern "C" fn qk_obs_add(
...
+///
+/// # Safety
+///
+/// To call this function safely
+///
+///   * ``left`` or ``right`` must be valid, non-null pointers to ``QkObs``\ s
+///   * ``qargs`` must point to an array of ``uint32_t``, readable for ``qk_obs_num_qubits(left)``
+///     elements (meaning the number of qubits in ``left``)
```
jakelishman: Let's comment that `qargs` can be null if `left.num_qubits` is 0 (this is useful behaviour for C and C++), and then make sure our own code handles that safely too.

crates/cext/src/sparse_observable.rs
```diff
@@ -631,6 +631,84 @@ pub unsafe extern "C" fn qk_obs_add(
...
+    left: *const SparseObservable,
+    qargs: *const u32,
+) -> *mut SparseObservable {
+    // SAFETY: Per documentation, the pointers are non-null and aligned.
+    let right = unsafe { const_ptr_as_ref(right) };
+    let left = unsafe { const_ptr_as_ref(left) };
+
+    // SAFETY: Per documentation, qargs is safe to read up to ``left.num_qubits()`` elements,
+    // which is the maximal value of ``index`` here.
+    let qargs_map = |index: u32| unsafe { *qargs.add(index as usize) };
```
jakelishman: The justification given in this SAFETY section isn't sufficient to make _calling_ the function safe: there's a second assumption that nothing will call `qargs_map` with an out-of-bounds index.  That means that `qargs_map` itself should be an `unsafe` function.

A nicer way to make this safe is to do something like:
```rust
let qargs = if qargs.is_null() {
  if left.num_qubits != 0 {
    panic!();
  }
  &[]
} else {
  if !qargs.is_aligned() {
    panic!();
  }
  // SAFETY: `qargs` is required to be valid for reads of up to `left.num_qubits` elements.
  unsafe { ::std::slice::from_raw_parts(qargs, left.num_qubits) }
};
let qargs_map = |index| qargs[index as usize];
```

jakelishman (commented):
Thanks for writing the tests to not rely on the particular factorisation, haha.

*edit*: this review is mostly just formalising conversations Julien and I had offline, for posterity.  They should already be resolved by 5d7a5a9.

jakelishman (approved):
Ace, the changes in the latest commit address everything I'd said (as expected), thanks.

We _might_ have included explicit documentation that `qargs` can be null if `second.num_qubits` was zero, but in C land that's usually the case, so it's no big deal at all.  I'll just merge as-is.

crates/cext/src/sparse_observable.rs
jakelishman: Done in 5d7a5a90ce707ff66b294ad2213fcaed0fd32fa6.

crates/cext/src/sparse_observable.rs
jakelishman: Done in 5d7a5a90ce707ff66b294ad2213fcaed0fd32fa6.

PR #13952, opened 2025-03-04
Fix is_parameterized() function of controlled gates

AlbertJP:
### Summary

Controlled gates are not recognised as parameterized by the `is_parameterized()` function. This causes issues in Aer and Algorithms when using circuits with controlled rotations: https://github.com/Qiskit/qiskit-aer/issues/2323, https://github.com/qiskit-community/qiskit-algorithms/issues/216

### Details and comments

The `is_parameterized()` function reads from the `_params` variable, which is not set by some classes that override the `params` property, such as `ControlledGate`, `CUGate`, `AnnotatedOperation` and `Initialize`.

This pull request fixes `is_parameterized()`, such that it reads the `params` variable instead, which corresponds to the parameters of the base gate for controlled gates.

This pull request also adds a unit test and fixes an existing one.

### DIFF
--- a/qiskit/circuit/instruction.py
+++ b/qiskit/circuit/instruction.py
@@ -287,7 +287,7 @@ class Instruction(Operation):
         """Return whether the :class:`Instruction` contains :ref:`compile-time parameters
         <circuit-compile-time-parameters>`."""
         return any(
-            isinstance(param, ParameterExpression) and param.parameters for param in self._params
+            isinstance(param, ParameterExpression) and param.parameters for param in self.params
         )
 
     @property
--- a/test/python/circuit/test_controlled_gate.py
+++ b/test/python/circuit/test_controlled_gate.py
@@ -1300,7 +1300,8 @@ class TestControlledGate(QiskitTestCase):
         """
         for gate_class in ControlledGate.__subclasses__():
             with self.subTest(i=repr(gate_class)):
-                # Singleton class isn't intended to be created directly, and its subclasses are not parameterized
+                # Singleton class isn't intended to be created directly, and its subclasses are not
+                # parameterized
                 if gate_class in {SingletonControlledGate, _SingletonControlledGateOverrides}:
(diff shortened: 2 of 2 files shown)


AlbertJP:
The unit tests found another issue, the controlled gates in `qiskit_aer.backends.name_mapping` cannot be initialised anymore. (these did not get loaded when running `tox` locally, so I did not spot the issue before creating the PR.)

This is caused by commit https://github.com/doichanj/qiskit-aer/commit/4587384b63e894a60dc57fbd790b189c6be2d709 (pull request Qiskit/qiskit-aer#1995). I think this should be fixed on the Aer end?

AlbertJP:
Some tests still fail as Qiskit/qiskit-aer#2326 is not yet merged.

I could change the unit test to test only a fixed list of subclasses, or read the `__module__` attribute of the subclasses to exclude those from Aer - which feels like a quite ugly solution.

mtreinish:
> Some tests still fail as [Qiskit/qiskit-aer#2326](https://github.com/Qiskit/qiskit-aer/pull/2326) is not yet merged.
> 
> I could change the unit test to test only a fixed list of subclasses, or read the `__module__` attribute of the subclasses to exclude those from Aer - which feels like a quite ugly solution.

To be honest it's not super hacky, in this case I would view the failures as a testing harness bug. The intent of that test is to validate that all the `ControlledGate` instances in `qiskit` are correct. It pulling in other class definitions from other libraries that use qiskit is not the intent there and is not something that the test is intended to validate. I'd say updating the test harness to check that the leaf class being tested originates in `qiskit` is totally correct in this case.

AlbertJP:
I have excluded the classes that start with `qiskit_aer` and now all tests pass.

It only tests classes that are direct subclasses of `ControlledGate`, but there are not so many interesting classes that are sub-subclasses, only non-parameterized and deprecated ones.

PR #13953, opened 2025-03-04
Set version number to 2.0.0rc1 for first release candidate

mtreinish:
### Summary

For the 2.0.0 release we're going to push release candidates prior to the final to enable testing before we tag the final release. In preparation for tagging the first release candidate this commit updates the version string to indicate it's a release candidate. This commit should be the final commit on main for 2.0.0rc1 and what gets tagged as 2.0.0rc1 post-merge. This will trigger the branch creation for stable/2.0 which is where 2.0.0 will be tagged from.


ElePT (commented):
LGTM. I just wanted to leave a reminder to move the release notes from pending PRs to `releasenotes/notes/2.0` just before the release (I failed to do this in 1.4 and now it's ugly with a few dangling renos, the rendered docs look fine though).

ElePT (approved):
LGTM

PR #13956, opened 2025-03-04
Stop using faer-rs for uc_gate.rs

mtreinish:
### Summary

This commit stops using faer and ndarray for the uc_gate.rs module and switches to use nalgebra instead. The uc_gate.rs module works all in terms of 2x2 matrices and moving to nalgebra brings the advantage of having everything be fixed size stack allocated for the arrays used. While this code path isn't performance critical this should improve performance, despite the goal here being to remove a dependency on faer.

The tradeoff here is that nalgebra doesn't seem to have a general eigensolver natively. There is functionality for this in the nalgebra-lapack crate, but as the name implies this depends on linking against lapack to work. The reason we originally started using faer is because it's pure rust and doesn't complicate our build system. In this specific case this is not a big deal because the matrices we need to compute the eigenvectors and eigenvalues for is a 2x2 so this commit includes a function to compute that directly. This lack of functionality is something to keep in mind for the future if we need this functionality more generally we will have to make a choice between faer and linking against lapack.

### Details and comments

Part of #13665

Co-authored-by: Gadi Aleksandrowicz <gadial@gmail.com>
Co-authored-by: Alexander Ivrii <alexi@il.ibm.com>


crates/accelerate/src/uc_gate.rs
```diff
@@ -86,11 +122,14 @@ pub fn dec_ucg_help(
     sq_gates: Vec<PyReadonlyArray2<Complex64>>,
     num_qubits: u32,
 ) -> (Vec<PyObject>, PyObject) {
-    let mut single_qubit_gates: Vec<Array2<Complex64>> = sq_gates
+    let mut single_qubit_gates: Vec<Matrix2<Complex64>> = sq_gates
         .into_iter()
-        .map(|x| x.as_array().to_owned())
+        .map(|x| {
+            let res: MatrixView2<Complex64> = x.try_as_matrix().unwrap();
```
alexanderivrii, line 128: As discussed offline, we should change this to ``let res: MatrixView2<Complex64, Dyn, Dyn> = x.try_as_matrix().unwrap();``, otherwise the conversion fails.

qiskit/circuit/library/generalized_gates/uc.py
```diff
@@ -286,7 +286,7 @@ def _dec_ucg_help(self):
         This method finds the single qubit gate arising in the decomposition of UCGates given in
         https://arxiv.org/pdf/quant-ph/0410066.pdf.
         """
-        single_qubit_gates = [gate.astype(complex) for gate in self.params]
+        single_qubit_gates = [np.asarray(gate, dtype=complex, order="f") for gate in self.params]
```
alexanderivrii, line 289: Or does this line make the rust code work without needing the `Dyn` trait?

crates/accelerate/src/uc_gate.rs
mtreinish: This isn't necessary because the python space change you commented on below is using fortran ordering which matches up with what nalgebra is using. We'd only need dynamic striding like this if we wanted to support an arbitrary 2x2 array. But since this function only is called in one place `uc.py` we don't need that level of flexibility.

ShellyGarion (commented):
1) I went over the code and the relevant reference, and it looks well to me.
However, I would be happy if there were some tests (even a few fixed 2x2 matrices).
Basically, one should take a matrix `mat` and check that:
`mat * eigenvectors[i] = eigvals[i] * eigenvectors[i] for i in [0,1]`.
in particular, when `mat[0,1]=0` or `mat[1,0]=0` or both. 
Take also matrices whose rows are linearly dependent.

2. For a better code organization, and in order to avoid code duplication, one may consider moving the `compute_2x2_eig` and `euler_one_qubit_decomposer::det_one_qubit` to a separate `util_matrices` file. This can also be done in a follow-up PR.

mtreinish:
> 1. I went over the code and the relevant reference, and it looks well to me.
>        However, I would be happy if there were some tests (even a few fixed 2x2 matrices).
>        Basically, one should take a matrix `mat` and check that:
>        `mat * eigenvectors[i] = eigvals[i] * eigenvectors[i] for i in [0,1]`.
>        in particular, when `mat[0,1]=0` or `mat[1,0]=0` or both.
>        Take also matrices whose rows are linearly dependent.
> 

I pushed up tests in: https://github.com/Qiskit/qiskit/pull/13956/commits/76d95846ba12e810c8e184fea6fd60cf6af2e5a5

> 
>     2. For a better code organization, and in order to avoid code duplication, one may consider moving the `compute_2x2_eig` and `euler_one_qubit_decomposer::det_one_qubit` to a separate `util_matrices` file. This can also be done in a follow-up PR.

Lets do this in a follow-up PR where we reorganize these utility functions into a standalone module.

crates/accelerate/src/uc_gate.rs
```diff
@@ -160,3 +205,79 @@ pub fn uc_gate(m: &Bound<PyModule>) -> PyResult<()> {
...
+    use nalgebra::Matrix2;
+    use num_complex::{Complex64, ComplexFloat};
+    use rand::prelude::*;
+    use rand_distr::StandardNormal;
+    use rand_pcg::Pcg64Mcg;
+
+    #[inline(always)]
+    fn random_complex(rng: &mut Pcg64Mcg) -> Complex64 {
+        Complex64::new(rng.sample(StandardNormal), rng.sample(StandardNormal))
+            * std::f64::consts::FRAC_1_SQRT_2
```
ShellyGarion: why do you need to divide in sqrt(2) ?

crates/accelerate/src/uc_gate.rs
```diff
@@ -160,3 +205,79 @@ pub fn uc_gate(m: &Bound<PyModule>) -> PyResult<()> {
...
+    }
+
+    fn random_unitaries(seed: u64, size: usize) -> impl Iterator<Item = Matrix2<Complex64>> {
```
ShellyGarion: why there is a need to choose random unitaries? random complex 2x2 matrices are good enough.
I'm also not sure that this code returns a unitary matrix.

crates/accelerate/src/uc_gate.rs
```diff
@@ -160,3 +205,79 @@ pub fn uc_gate(m: &Bound<PyModule>) -> PyResult<()> {
...
+        let mut rng = Pcg64Mcg::seed_from_u64(seed);
+
+        (0..size).map(move |_| {
+            let mat: Matrix2<Complex64> = [
+                [random_complex(&mut rng), random_complex(&mut rng)],
+                [random_complex(&mut rng), random_complex(&mut rng)],
+            ]
+            .into();
+            let (q, r) = mat.qr().unpack();
```
ShellyGarion: not sure that qr decomposition is needed. r is not necessarily diagonal (it's triangular), and can have zeroes on the diagonal.

ShellyGarion: if you remove these lines for the qr decomposition, you should get a random matrix which is enough for the test.

crates/accelerate/src/uc_gate.rs
```diff
@@ -160,3 +205,79 @@ pub fn uc_gate(m: &Bound<PyModule>) -> PyResult<()> {
...
+            let diag = r.map_diagonal(|x| x / x.abs());
+            q.map_with_location(|i, _j, val| val * diag[i])
+        })
+    }
+
+    fn check_eig(mat: Matrix2<Complex64>) {
```
ShellyGarion, line 224: this is a good check!

crates/accelerate/src/uc_gate.rs
```diff
@@ -160,3 +205,79 @@ pub fn uc_gate(m: &Bound<PyModule>) -> PyResult<()> {
...
+
+    #[test]
+    fn test_unitary_eig() {
+        for unitary in random_unitaries(42, 1024) {
+            check_eig(unitary);
+        }
+    }
+
+    #[test]
+    fn test_diagonal_eig() {
```
ShellyGarion, line 250: it's good to check the diagonal case

crates/accelerate/src/uc_gate.rs
```diff
@@ -160,3 +205,79 @@ pub fn uc_gate(m: &Bound<PyModule>) -> PyResult<()> {
...
+        for _ in 0..1024 {
+            let mat = Matrix2::from_diagonal(
+                &[random_complex(&mut rng), random_complex(&mut rng)].into(),
+            );
+            check_eig(mat);
+        }
+    }
+
+    #[test]
+    fn test_off_diagonal_eig() {
```
ShellyGarion, line 262: it's also a good check

ShellyGarion (commented):
Thanks for adding the tests. I have some comments about them.

crates/accelerate/src/uc_gate.rs
mtreinish: I'll remove this and just use a random complex 2x2 matrix instead. I just copied this code from elsewhere and didn't really think it through that we didn't need to create a random unitary for the test.

mtreinish: Simplified in: https://github.com/Qiskit/qiskit/pull/13956/commits/7d7b43866e255ac27c20324d80bf7bc0305cbc7d

crates/accelerate/src/uc_gate.rs
mtreinish: Removed in: https://github.com/Qiskit/qiskit/pull/13956/commits/7d7b43866e255ac27c20324d80bf7bc0305cbc7d

crates/accelerate/src/uc_gate.rs
mtreinish: I doesn't need to be I can remove this although it functionally doesn't matter. I copied this from another function for generating random unitaries (which was based on the scipy functionality for doing the same).

ShellyGarion: I think that if it's not important, then it's better to remove it since it's confusing. other than that, the code looks good.

mtreinish:
```suggestion
```

ShellyGarion (approved):
LGTM. I very much like this algorithm & code :)

PR #13957, opened 2025-03-04
Prepare 1.4.1

1ucian0:


mtreinish (commented):
You need to update the `Cargo.toml` and run `cargo build` afterwords to update the rust component's version too. But otherwise this looks fine. Was there any copy editing needed in the fix release notes?

1ucian0:
> You need to update the `Cargo.toml` and run `cargo build` afterwords to update the rust component's version too.

done in https://github.com/Qiskit/qiskit/pull/13957/commits/e329c283a5c8c5d090804ec534569a37cbfdadb5

releasenotes/notes/prepare-1.4.1-33ba454a4c35cccb.yaml
```diff
@@ -0,0 +1,7 @@
+---
+prelude: >
+    Qiskit 1.4.1 is a small patch release, fixing some bugs found in the 1.4 extended support series.
+fixes:
+  - |
+    The serialization of :class:`ParameterExpression` now does not raise an ``AttributeError`` exception
+    when serializing with QPY 13.
```
mtreinish: This one is actually a bit more nuanced. I'd write it as something like:

```suggestion
    Fixed an issue with QPY 13 when serializing or deserializing a :class:`ParameterExpression`  object that was defined
    by calling :meth:`.ParameterExpression.subs` to substitute a :class:`Parameter` with a
    :class:`.ParameterExpression`. For example::
    
        from qiskit.circuit import Parameter
        
        a = Parameter("A")
        b = Parameter("B")
        expr = a + b
        c = Parameter("C")
        new_expr = c ** 3.14
        final_expr.subs({b: new_expr})
        
    In previous releases if you went to serialize a :class:`.QuantumCircuit` that contained an expression like
    this with :func:`.qpy.dump` it would have raised an ``AttributeError`` with an error message
    `'ParameterExpression' object has no attribute 'name'` when using QPY 13 (the default). This has been
    fixed so that the exception is no longer raised and you can serialize and deserialize a payload containing
    these nested :class:`.ParameterExpression` objects.
```

releasenotes/notes/1.4/prepare-1.4.0-aee22d10333374c8.yaml
mtreinish, line 1: You didn't actually need to update this as part of this release. My comment before was just about the 1.4.1 pending release notes (of which there is only 1). But it's good to do this and it will reflect in the output.

mtreinish (approved):

PR #13960, opened 2025-03-05
Extend deprecation of QuantumCircuit.duration and DAGCircuit.duration

mtreinish:
### Summary

Since we're unlikely to be able to include these removals as part of Qiskit 2.0.0 given the dependencies on them and that the removals are not critical for any work we have in the 2.x series this commit opts to just update the deprecation message to say the removal will occur in Qiskit 3.0.0 not 2.0.0 like was originally planned.

### Details and comments

### DIFF
--- a/crates/circuit/src/dag_circuit.rs
+++ b/crates/circuit/src/dag_circuit.rs
@@ -397,7 +397,7 @@ impl DAGCircuit {
     /// The total duration of the circuit, set by a scheduling transpiler pass. Its unit is
     /// specified by :attr:`.unit`
     ///
-    /// DEPRECATED since Qiskit 1.3.0 and will be removed in Qiskit 2.0.0
+    /// DEPRECATED since Qiskit 1.3.0 and will be removed in Qiskit 3.0.0
     #[getter]
     fn get_duration(&self, py: Python) -> PyResult<Option<Py<PyAny>>> {
         imports::WARNINGS_WARN.get_bound(py).call1((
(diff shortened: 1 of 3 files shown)


alexanderivrii (commented):
LGTM! Just one minor question about the release notes: we are suggesting to use :meth:`.QuantumCircuit.estimate_duration` for `DAGCircuit`s as well?

mtreinish:
> LGTM! Just one minor question about the release notes: we are suggesting to use :meth:`.QuantumCircuit.estimate_duration` for `DAGCircuit`s as well?

We are as the method isn't publicly exposed for `DAGCircuit` we could add it because it's actually internally operating on a dag, but at the time we added `estimate_duration` it was only really thought of as being useful for a circuit post-transpilation. We can look at adding it to `DAGCircuit` in 2.1, I'll open an issue to discuss/track this.

mtreinish:
Actually I remember now it wasn't straightforward to add it to DAGCircuit. The issue is that `DAGCircuit` is in the `circuit` crate and `Target` is in the `accelerate` crate. We need both objects for `estimate_duration` and I couldn't add a dependency in `circuit` on `accelerate` because `accelerate` depends on `circuit`. So I added it as a standalone function in `accelerate`: https://github.com/Qiskit/qiskit/blob/main/crates/accelerate/src/circuit_duration.rs and the method is added in Python. This is not trivial to do for `DAGCircuit` because it is all written in rust right now so I can't just add a method in Python space.

alexanderivrii (approved):

PR #13922, opened 2025-02-25
Better MCX Decomposition with logarithmic Toffoli Depth

patelvyom:
### Summary
Implement four new multi-controlled-x synthesis functions: `synth_mcx_1_clean_kg24`, `synth_mcx_1_dirty_kg24`, `synth_mcx_2_clean_kg24`, `synth_mcx_2_dirty_kg24` as discussed in #13913.

### Details and comments
Once implementations are confirmed, these should be added to the HLS transpiler synthesis plugin.

PR #13604, opened 2024-12-24
Move PendingDeprecations to Deprecations in the circuit library

1ucian0:
### Summary

The circuit library modernization from #13046 introduced some pending deprecations that this PR is moving to deprecations.

~reno is still missing.~

PR #13961, opened 2025-03-05
Add explicit tests for MCX synthesis algorithms

alexanderivrii:
### Summary

There are several ongoing developments related to MCX synthesis algorithms. For example, #13922 adds new synthesis algorithms  and #13929 ports some of the existing algorithms to Rust. However, we currently lack a dedicated test module for these, see [this comment](https://github.com/Qiskit/qiskit/pull/13922#issuecomment-2684303882) that describes the mess we are in.

This PR adds a dedicated test module for MCX synthesis algorithms, taking into account their usage of clean and dirty auxiliary qubits.


Cryoris:
This is much nicer than what we currently have in 
https://github.com/Qiskit/qiskit/blob/fdd5e96d6fc9d13fc2cca3dd2e285287ddf472d7/test/python/circuit/test_controlled_gate.py#L529

If you're adding this detailed check here, maybe we can reduce the test I linked above to just try a single number of control qubits, so it only checks that the `.mcx` interface works rather than checking correctness.

test/python/synthesis/test_mcx_synthesis.py
```diff
@@ -0,0 +1,130 @@
...
+        self.check_mcx_synthesis(num_ctrl_qubits, synthesized_circuit, clean_ancillas=True)
+
+    @data(3, 4, 5, 6, 7, 8)
+    def test_mcx_gray_code(self, num_ctrl_qubits: int):
+        """Test synth_mcx_gray_code by comparing synthesized and expected matrices."""
+        # Note: the method requires at least 3 control qubits
+        synthesized_circuit = synth_mcx_gray_code(num_ctrl_qubits)
+        self.check_mcx_synthesis(num_ctrl_qubits, synthesized_circuit, clean_ancillas=False)
+
+    @data(1, 2, 3, 4, 5, 6, 7, 8)
```
Cryoris: How did you pick these numbers? I'm wondering if we can find a good way to select these that doesn't just run redundant checks to save some time 🤔 E.g. if nothing in the algorithm changes in between n1 and n2 qubits, maybe not all numbers need to be checked.. unless it's super fast

alexanderivrii: The lower bounds were selected by looking at the code / reading the documentation for the corresponding HLS plugins (we have already done the work in the past to understand for which numbers of control/ancilla qubits each synthesis method is applicable). 

I actually slightly don't like the fact that some synthesis methods do not work with $1$ or $2$ control qubits, but it's probably fine since it's currently the job of the synthesis plugin and not of the synthesis function to decide when each method is applicable; also the plugin (unlike the function) can return `None`. What do you think about this? In addition, should a synthesis function raise some kind of an error if it's called with an incorrect number of qubits?

The upper bounds were selected so that the number of qubits in the synthesized circuit does not exceed 9 or 10, otherwise the tests do become very slow. For smaller numbers of control qubits, the tests are reasonably fast since both the size of the matrix is smaller and the number of gates in the synthesized circuit is smaller.

I would prefer not to skip the smaller numbers - both because as mentioned just before these are not the most expensive tests, and because the tests should be independent of possible tweaks to the synthesis algorithms we are planning in the future (i.e. maybe right now nothing changes in a range of qubits, but it might with a small change to the algorithm -- and we want the tests to detect if something is wrong).

alexanderivrii:
Removing (or changing) the linked test seems like a good idea. There are probably some other tests that should be removed as well, Shelly is looking into it.

alexanderivrii:
@Cryoris, regarding `test_multi_control_toffoli_matrix_advanced_num_ancillas`: I think we should keep this test for now as it's checking correctness of adding custom MCX variants (`MCXVChain`, `MCXRecursive`, etc.) to a quantum circuit. But looking at it, we should deprecate the argument `mode` in `QuantumCircuit.mcx` (and every other place that may use this), what do you think?

test/python/synthesis/test_multi_controlled_synthesis.py
```diff
@@ -0,0 +1,347 @@
...
+        qc = QuantumCircuit(num_ctrl_qubits + 1)
+        qc.append(base_gate.control(num_ctrl_qubits), range(num_ctrl_qubits + 1))
+        base_mat = base_gate.to_matrix()
+        test_op = Operator(qc)
+        cop_mat = _compute_control_matrix(base_mat, num_ctrl_qubits)
+        self.assertTrue(matrix_equal(cop_mat, test_op.data))
+
+        # TODO: fix optimization_level=0 after updating CS and CSdg in the equivalence library
+        pm = generate_preset_pass_manager(
+            optimization_level=2, basis_gates=["u", "cx"], seed_transpiler=12345
```
ShellyGarion: this should be updated after #14001 is merged

alexanderivrii: Done.

Cryoris:
> @Cryoris, regarding test_multi_control_toffoli_matrix_advanced_num_ancillas: I think we should keep this test for now as it's checking correctness of adding custom MCX variants (MCXVChain, MCXRecursive, etc.) to a quantum circuit.

Since we already test the decompositions here, that test should really only need to check whether the variant is correctly added -- not that the decomposition is correct. But if it's fast and you'd like to keep it I'm also fine 🙂 

> But looking at it, we should deprecate the argument mode in QuantumCircuit.mcx (and every other place that may use this), what do you think?

There's an open PR for that: #13604 🙂

test/python/synthesis/test_multi_controlled_synthesis.py
```diff
@@ -0,0 +1,380 @@
...
+            SGate(),
+            SdgGate(),
+            TGate(),
+            TdgGate(),
+            SXGate(),
+            SXdgGate(),
+            RXGate(0.789),
+            RYGate(0.123),
+            RZGate(0.456),
+        ],
```
ShellyGarion, line 345: Ugate does not appear in the list. otherwise LGTM.

test/python/synthesis/test_multi_controlled_synthesis.py
```diff
@@ -0,0 +1,380 @@
...
+            # MCPhase gate and other locally equivalent multi-controlled gates
+            expected = {1: 2, 2: 6, 3: 20, 4: 44, 5: 84, 6: 140, 7: 220, 8: 324}
+        elif isinstance(base_gate, RZGate):
+            expected = {1: 2, 2: 4, 3: 14, 4: 24, 5: 40, 6: 56, 7: 80, 8: 104}
+        elif isinstance(base_gate, (RXGate, RYGate)):
+            expected = {1: 2, 2: 8, 3: 20, 4: 24, 5: 40, 6: 56, 7: 80, 8: 104}
+        elif isinstance(base_gate, UGate):
+            expected = {1: 2, 2: 22, 3: 54, 4: 92, 5: 164, 6: 252, 7: 380, 8: 532}
+
+        self.assertEqual(cx_count, expected[num_ctrl_qubits])
```
ShellyGarion: should it better be `self.assertLessEqual` ?

alexanderivrii: Yes, definitely, done in [e3e8dd6](https://github.com/Qiskit/qiskit/pull/13961/commits/e3e8dd6fa8d3fd98e994480076413279ce7ecabe).

test/python/synthesis/test_multi_controlled_synthesis.py
alexanderivrii: Nice catch, done in [e3e8dd6](https://github.com/Qiskit/qiskit/pull/13961/commits/e3e8dd6fa8d3fd98e994480076413279ce7ecabe).

test/python/synthesis/test_multi_controlled_synthesis.py
```diff
@@ -0,0 +1,384 @@
...
+            # MCPhase gate and other locally equivalent multi-controlled gates
+            expected = {1: 2, 2: 6, 3: 20, 4: 44, 5: 84, 6: 140, 7: 220, 8: 324}
+        elif isinstance(base_gate, RZGate):
+            expected = {1: 2, 2: 4, 3: 14, 4: 24, 5: 40, 6: 56, 7: 80, 8: 104}
+        elif isinstance(base_gate, (RXGate, RYGate)):
+            expected = {1: 2, 2: 8, 3: 20, 4: 24, 5: 40, 6: 56, 7: 80, 8: 104}
+        elif isinstance(base_gate, UGate):
+            expected = {1: 2, 2: 22, 3: 54, 4: 92, 5: 164, 6: 252, 7: 380, 8: 532}
+        else:
+            # make sure we don't miss any case (and make pytint happy)
```
Cryoris: Could we just raise an error instead here? Otherwise there could be a scenario where this silently passes (where some magic gate has 0 CX) -- in case you want to keep it as is, there's a small typo "pylint" couldn't catch 😛 
```suggestion
            # make sure we don't miss any case (and make pylint happy)
```

test/python/synthesis/test_multi_controlled_synthesis.py
```diff
@@ -0,0 +1,384 @@
...
+        transpiled_circuit = pm.run(qc)
+        cx_count = transpiled_circuit.count_ops()["cx"]
+        # The synthesis of MCX(n) uses two MCRZ(n), one MCRY(n), and one MCPhase(n-1).
+        # Thus the number of CX-gate should be upper-bounded by
+        # 3*(16 * (n + 1) - 40) + (8 * (n-1)^2 - 16 * (n-1))
+        self.assertLessEqual(cx_count, 8 * num_ctrl_qubits**2 + 16 * num_ctrl_qubits - 96)
+
+    @combine(
+        num_ctrl_qubits=[1, 2, 3, 4, 5, 6, 7, 8],
+        base_gate=[
```
Cryoris, line 326: What about the U1/2/3 gates?

test/python/synthesis/test_multi_controlled_synthesis.py
```diff
@@ -0,0 +1,384 @@
...
+            XGate(),
+            YGate(),
+            ZGate(),
+            HGate(),
+            PhaseGate(0.123),
```
Cryoris, line 331: There are no numeric dependencies in these algorithms, right? If that's so how about using parameterized gates here to show the generality?

test/python/synthesis/test_multi_controlled_synthesis.py
```diff
@@ -0,0 +1,384 @@
...
+        transpiled_circuit = pm.run(synthesized_circuit)
+        cx_count = transpiled_circuit.count_ops()["cx"]
+        # The bound from the default construction for C3X
+        self.assertLessEqual(cx_count, 14)
+
+    def test_c4x_cx_count(self):
+        """Test synth_c4x bound on CX count."""
+        synthesized_circuit = synth_c4x()
+        pm = generate_preset_pass_manager(
+            optimization_level=0, basis_gates=["u", "cx"], seed_transpiler=12345
```
Cryoris: Does the seed matter?

test/python/synthesis/test_multi_controlled_synthesis.py
```diff
@@ -0,0 +1,384 @@
...
+        )
+        transpiled_circuit = pm.run(synthesized_circuit)
+        cx_count = transpiled_circuit.count_ops()["cx"]
+        # The bound from the default constuction for C4X
+        self.assertLessEqual(cx_count, 36)
+
+    @combine(
+        num_ctrl_qubits=[4, 5, 8, 10, 13, 15],
```
Cryoris: Is there a reason these are different from the `[5, 10, 15]` you used above?

test/python/synthesis/test_multi_controlled_synthesis.py
```diff
@@ -0,0 +1,384 @@
...
+        # The bounds are given in arXiv:2302.06377, Theorem 3.
+        # In practice, we actually get better bounds for small values of num_ctrl_qubits.
+        expected_cx_count = 16 * (num_ctrl_qubits + 1) - 40
+        self.assertLessEqual(cx_count, expected_cx_count)
+
+    @data(5, 8, 10, 13, 15)
+    def test_mcx_noaux_v24_cx_count(self, num_ctrl_qubits: int):
+        """Test synth_mcx_noaux_v24 bound on CX count."""
+        synthesized_circuit = synth_mcx_noaux_v24(num_ctrl_qubits)
+        pm = generate_preset_pass_manager(
```
Cryoris: You could consider moving the definition of this pass manager into a `def setUp(self)`, since it's always the same definition 🙂 I.e. something like
```python
def setUp(self):
    super().setUp()
    self.pm = generate_preset_pass_manager(...)
```
and then just use `self.pm.run` in the tests

test/python/synthesis/test_multi_controlled_synthesis.py
```diff
@@ -0,0 +1,384 @@
...
+        qc = QuantumCircuit(num_ctrl_qubits + 1)
+        qc.append(base_gate.control(num_ctrl_qubits), range(num_ctrl_qubits + 1))
+        pm = generate_preset_pass_manager(
+            optimization_level=0, basis_gates=["u", "cx"], seed_transpiler=12345
+        )
+        transpiled_circuit = pm.run(qc)
+        cx_count = transpiled_circuit.count_ops()["cx"]
+        # The synthesis of mcrx/mcry/mcrz gates uses _mcsu2_real_diagonal.
+        # The bounds are given in arXiv:2302.06377, Theorem 3.
+        # In practice, we actually get better bounds for small values of num_ctrl_qubits.
```
Cryoris, line 263: So you reached a conclusion as to whether the official bound from the paper or the actual number should be tested?

test/python/synthesis/test_multi_controlled_synthesis.py
```diff
@@ -0,0 +1,384 @@
...
+class TestMCSynthesisCorrectness(QiskitTestCase):
+    """Test correctness of synthesis methods for multi-controlled gates."""
+
+    @staticmethod
+    def mc_matrix(base_gate: Gate, num_ctrl_qubits: int):
+        """Return matrix for the MC gate with the given base gate and the number of control qubits."""
+        base_mat = base_gate.to_matrix()
+        return _compute_control_matrix(base_mat, num_ctrl_qubits)
+
+    def check_mc_synthesis(
```
Cryoris: The test methods are typically called something like `assertSynthesisCorrect` or something, in case you'd like to be consistent. We use the same naming scheme in some other tests too

test/python/synthesis/test_multi_controlled_synthesis.py
alexanderivrii: Lol about me misspelling pylint and pylint not catching that. Raising an error is indeed better.

test/python/synthesis/test_multi_controlled_synthesis.py
alexanderivrii: I don't think it does, but I think it's best to force everything be fully deterministic just in case.

test/python/synthesis/test_multi_controlled_synthesis.py
alexanderivrii: Exactly! No numeric dependencies as far as I know.

test/python/synthesis/test_multi_controlled_synthesis.py
alexanderivrii: This is much cleaner indeed.

test/python/synthesis/test_multi_controlled_synthesis.py
ShellyGarion: I thought that U1/U2/U3 are going to be deprecated...
U1 should basically be as PhaseGate, U3 should basically be as UGate, and I'm not sure about U2.

test/python/synthesis/test_multi_controlled_synthesis.py
ShellyGarion: according to our experiments this bound is tight for `num_ctrl_qubits >= 8`. For smaller values it can be less (we added a test for a small number of controls).

alexanderivrii: Yes, in certain cases we get slightly better bounds for smal values of `num_ctrl_qubits` due to special optimizations: for example, using the special decomposition for the 3-controlled X-gate. For larger values of `num_ctrl_qubits` the bounds are usually tight. We have decided to test both the official bounds, and add explicit tests for smaller values.

test/python/synthesis/test_multi_controlled_synthesis.py
alexanderivrii: Good point, will do!

test/python/synthesis/test_multi_controlled_synthesis.py
alexanderivrii: I will add these. And now the tests for small numbers of control qubits show something quite interesting: the 8-controlled `U3Gate` gets synthesized with **less** CX-gates than the 8-controlled `U1Gate`. And it's not the specific numerical parameter values -- I have locally tried different values and I always got consistent results. So there is an opportunity to optimize `U1Gate`s.

test/python/synthesis/test_multi_controlled_synthesis.py
alexanderivrii: Good point! I believe we can reduce the number of tests by restricting to the same `[5, 10, 15]`.

alexanderivrii:
Thanks @Cryoris, I have addressed your comments, this is ready for re-review.

test/python/synthesis/test_multi_controlled_synthesis.py
ShellyGarion: I think that the U1Gate still calls the old exponential MCXGrayCode. I didn't optimize it since I thought it was going to get deprecated...

alexanderivrii: I have added the gates to the tests (as it probably makes sense to include these while the gates are still alive),

Cryoris (approved):
LGTM, thanks for this!

PR #13962, opened 2025-03-05
Missing release note in #13890

1ucian0:
### Summary

While working in 1.4.1, I noticed that the PR #13890 was missing a release note.


mtreinish (commented):
This actually isn't needed here. We would just delete the release not for 2.0.0 because it'll already get published as part of the 1.4.1 release. After 2.0 we would probably want this so we could document the fix on both 1.4.2 and 2.0.1, but beforehand it's assumed that 2.0.0 includes everything from 1.4.1 so we typically only want one release note.

PR #13967, opened 2025-03-06
Fix wheel builds for i686 Linux

mtreinish:
### Summary

In the 1.4.1 release the i686 linux wheel builds failed because the recent release of rustup 1.28.0 and 1.28.1 added a runtime dependency on libatomic to download and install the rust compiler. The i686 manylinux container we run the wheel builds in didn't contain this library so rustup errored and crashed the build. This commit fixes this issue by installing libatomic prior to running rustup. This fix was used in a throwaway branch to push out the 1.4.1 release on PyPI, but this fixes it for real for when we publish 1.4.2.

### Details and comments

### DIFF
--- a/.github/workflows/wheels-build.yml
+++ b/.github/workflows/wheels-build.yml
@@ -154,6 +154,7 @@ jobs:
       - name: Build wheels
         uses: pypa/cibuildwheel@v2.21.3
         env:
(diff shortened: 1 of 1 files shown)


Cryoris (approved):
LGTM, this fixed the wheels for the 1.4.1 release (see e5a7844722ab1f7f4e4b95fdc7f0f888a9553ca7)

PR #13860, opened 2025-02-17
Move `Bit` and `Register` to live in Rust space

raynelfss:
### Summary

The following commits finally bring `Bits` and `Registers` entirely to Rust by performing a couple of tasks:
- Create native representations of `Bit` (a.k.a. `ShareableQubit` and `ShareableClbit`) and `Register` (a.k.a. `QuantumRegister` and `ClassicalRegister`) and move their python counterparts to be managed by rust PyO3.
- Leverage the usage of native bits and registers in `DAGCircuit` and `CircuitData`.
- Add helper structs `RegisterData` and `BitLocator` to represent the layout of bits/registers in the circuit.

All these changes supersede #13686.

### Details and comments

As we move more of our core data model to Rust, the main drawback that we've still had is needing to use Python space to create bits and registers. Although simple on the outside, the `Bit` and `Register` structures in Python can be quite challenging due to cross referencing that can happen between each type to the other (`Bit._register` <-> `Register[i]`] which is not as easy to replicate in Rust space.

With the following commits we aim to move the representation of `Bits` and `Registers` to Rust in which the behavior is more similar to that of Python and allows us to share instances between circuits, rather than having the circuit have complete ownership of the bits it contains (which is the reason for preferring this over #13686).

#### Bit representation:

To represent a `Qubit` or `Clbit` in Rust, we have a couple of identifiers:
- Within the circuit, `Qubit` or `Clbit` which is an index that the circuit can locally refer to when using a bit.
- Globally, we would use one of two types of `Bit`:
    - `ShareableQubit` which represents a global bit.
    - `ShareableClbit` the classical counterpart to the `ShareableQubit`.

The base structure for a `Bit` in Rust, called `BitInfo` describes a global bit as being one of two things:
- `Owned` where the bit is owned by a register.
- `Anonymous` where the bit is independent, containing its own unique ID.

Each of these `Bits` also has an extra property called `BitExtraInfo` that acts as an identifier, and contains any extra information a `Bit` of a specific type should have. In the case of `ShareableQubit` whether the instance is an ancilla or not.

##### Python:
- The python counterparts( `PyBit`, `PyQubit`, `PyClbit` and `PyAncillaQubit` ) have similar behavior and inheritance models as the originals, but should not be subclassed any further as it is no longer supported. See #13841.
- The python counterparts may no longer be comparable via `is()` checks due to conversions in PyO3 not taking on the same addresses.
[...]

PR #13985, opened 2025-03-10
Refactor `BitData` to `ObjectRegistry`.

kevinhartman:
### Summary
This is a quick followup to address `BitData`'s name no longer being aligned with its function, esp. after #13860.

It's now called `ObjectRegistry` and functions the same way. The `VarAsKey` struct has been renamed to `PyObjectAsKey` since it was made public in #13860 and is thus now usable for other purposes. We will likely use it to track `stretch` variables as well.

### Details and comments
* I've removed the restriction that `BitData` (now `ObjectRegistry`) be used only with `ShareableBit`.
* I've moved the bound requirement `IntoPyObject` for `B` from the `ObjectRegistry` impl to its specific cache methods.
* I've also changed the `IntoPyObject` impl for `& PyObjectAsKey` to use `Borrowed` as its output type [as per the PyO3 migration guide](https://pyo3.rs/main/migration.html#intopyobject-manual-implementation).

### DIFF
--- a/crates/circuit/src/object_registry.rs
+++ b/crates/circuit/src/object_registry.rs
@@ -10,8 +10,6 @@
 // copyright notice, and modified files need to carry a notice indicating
 // that they have been altered from the originals.
 
-use crate::BitType;
-
 use hashbrown::HashMap;
 use pyo3::exceptions::{PyKeyError, PyRuntimeError, PyValueError};
 use pyo3::prelude::*;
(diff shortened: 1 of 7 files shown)


crates/circuit/src/object_registry.rs
```diff
@@ -107,140 +104,150 @@ impl PartialEq for VarAsKey {
...
 
     /// Gets a reference to the cached Python list, even if not initialized.
     #[inline]
-    pub fn cached_raw(&self) -> Option<&Py<PyList>> {
+    pub fn cached_raw(&self) -> Option<&Py<PyList>>
+    where
+        for<'a> B: IntoPyObject<'a>,
+    {
         self.cached.get()
     }
```
raynelfss: Would it make sense to separate these methods into a different impl block with `<B: IntoPyObject<'_>` as an extra trait bound? Since the `PyList` usage is now directly dependent of the object implementing said trait, I belive we could just move this to only work if said condition is met.

raynelfss (commented):
This looks like a sensible change! I have one question about the implementation, but other than that, this looks good!

crates/circuit/src/object_registry.rs
kevinhartman: That helps with organization, I'd say. Thanks!

Done in 39ed097

raynelfss (approved):
Looks good to me!

PR #13987, opened 2025-03-11
Minor fix to UnitarySynthesis with RZZ and CZ gates

ShellyGarion:
### Summary

This is a minor fix to https://github.com/Qiskit/qiskit/issues/13428#issuecomment-2712715129.
(so no release notes are needed, since it has not been released yet).

### Details and comments

A minor fix following https://github.com/Qiskit/qiskit/issues/13428#issuecomment-2699855434
when the backend has both CZ and RZZ gates, we need to handle first the parametric KAK gate (RZZ) and not the KAK gate (CZ).

### DIFF
--- a/test/python/transpiler/test_consolidate_blocks.py
+++ b/test/python/transpiler/test_consolidate_blocks.py
@@ -656,7 +656,8 @@ class TestConsolidateBlocks(QiskitTestCase):
 
     @data(["rzz", "rx", "rz"], ["rzz", "rx", "rz", "cz"])
     def test_collect_and_synthesize_rzz(self, basis_gates):
-        """Collect blocks with RZZ gates, and re-synthesizing it."""
+        """Collect blocks with RZZ gates, and re-synthesizing it.
+        Regression test for https://github.com/Qiskit/qiskit/issues/13428"""
         qc = QuantumCircuit(2)
         qc.rzz(0.1, 0, 1)
         qc.rzz(0.2, 0, 1)
--- a/test/python/transpiler/test_consolidate_blocks.py
+++ b/test/python/transpiler/test_consolidate_blocks.py
@@ -664,9 +664,7 @@ class TestConsolidateBlocks(QiskitTestCase):
         res = consolidate_pass(qc)
         self.assertEqual({"unitary": 1}, res.count_ops())
         self.assertEqual(Operator.from_circuit(qc), Operator(res.data[0].operation.params[0]))
-        pm = generate_preset_pass_manager(
-            optimization_level=2, basis_gates=["rz", "rzz", "sx", "x", "rx"]
-        )
+        pm = generate_preset_pass_manager(optimization_level=2, basis_gates=basis_gates)
         tqc = pm.run(qc)
         self.assertEqual(tqc.count_ops()["rzz"], 1)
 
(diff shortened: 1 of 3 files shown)


t-imamichi:
Thanks. I confirmed that it works.
```
2.1.0.dev0+de52c43

['rz', 'rzz', 'rx']
global phase: π

q_0: ──────────■──────────────────
     ┌───────┐ │ZZ(-0.3) ┌───────┐
q_1: ┤ Rx(π) ├─■─────────┤ Rx(π) ├
     └───────┘           └───────┘

['rz', 'rzz', 'rx', 'cz']
global phase: π

q_0: ──────────■──────────────────
     ┌───────┐ │ZZ(-0.3) ┌───────┐
q_1: ┤ Rx(π) ├─■─────────┤ Rx(π) ├
     └───────┘           └───────┘
```

test/python/transpiler/test_consolidate_blocks.py
```diff
@@ -655,15 +655,18 @@ def test_non_cx_target(self):
...
         qc = QuantumCircuit(2)
         qc.rzz(0.1, 0, 1)
         qc.rzz(0.2, 0, 1)
         consolidate_pass = ConsolidateBlocks(basis_gates=basis_gates)
         res = consolidate_pass(qc)
         self.assertEqual({"unitary": 1}, res.count_ops())
         self.assertEqual(Operator.from_circuit(qc), Operator(res.data[0].operation.params[0]))
+        pm = generate_preset_pass_manager(optimization_level=2, basis_gates=basis_gates)
+        tqc = pm.run(qc)
+        self.assertEqual(tqc.count_ops()["rzz"], 1)
```
ElePT, line 670:
```suggestion
        # regression test for https://github.com/Qiskit/qiskit/issues/13428#issuecomment-2699855434
        pm = generate_preset_pass_manager(optimization_level=2, basis_gates=basis_gates)
        tqc = pm.run(qc)
        self.assertEqual(tqc.count_ops()["rzz"], 1)
```

ElePT (commented):
Thanks Shelly! The fix looks good, it's just reordering the "target" (rust) path in unitary synthesis to apply the decomposers following the same priority order as the "basis_gates" (python) path. As @t-imamichi pointed out, this inconsistency surfaced when we modified the `generate_preset_pass_manager` function to use target internally (https://github.com/Qiskit/qiskit/pull/12850). Having the extra check in the consolidate blocks test is fine, but I am on the fence about also adding a regression test to the unitary synthesis unit tests. One that confirms that the same decomposers are applied for equivalent inputs given through `target` and `basis_gates`. I am not sure because the behavior won't always be the same, it will depend on the target properties. But maybe a test where the target has no errors, in that case both paths should give the same decomposition.

test/python/transpiler/test_consolidate_blocks.py
ShellyGarion: I added a reference to the issue (in line 660 above), since all the discussion there is relevant.
Following your suggestion, I also extended `test_unitary_synthesis.py` with some tests that have basis_gates and target that include both RZZ and CZ gates.

ElePT (approved):
LGTM. https://github.com/Qiskit/qiskit/issues/13428#issuecomment-2713812888 points out a different use-case that isn't addressed by this PR, were if a circuit contains both a CZ and RZZ gate and both are part of the basis gates, they don't get consolidated. I think this is fine, it is reasonable to expect that a circuit with hw-native gates doesn't get synthesized, even if that would provide an even further optimization.

PR #13991, opened 2025-03-11
Fix string and standard gate mismatch in commutation checker

### LINKED ISSUE #13988: Overloaded gate names cause panic in commutation checker
### Environment

- **Qiskit version**: 2.0.0rc1
- **Python version**: 3.13
- **Operating system**: Linux

### What is happening?

When calling the commutation checker (via `CommutativeCancellation` typically) with a custom gate that uses a standard gate name that is in the set of specially handled rotation gates names for the commutation checker, but is not a standard gate instance this results in a panic with the message: "Supported gates are standard gates". The root cause is the rust code of the commutation checker is assuming if it gets a standard gate name for a gate in the circuit than it must be a standard gate and when it isn't it panics. 

While it's true we make it explicit that the name is like the op code and the standard gates are reserved, this can come up not by user error necessarily, but because of #10737 when loading a qasm file for a gate that's in Qiskit's standard library but not qasm's. So we should be able to gracefully handle this case correctly by being resilient to this mismatch.

```python
from qiskit.transpiler import generate_preset_pass_manager, Target, CouplingMap

qasm_str = """OPENQASM 2.0;
include "qelib1.inc";
gate ryy(param0) q0,q1
{
 rx(pi/2) q0;
 rx(pi/2) q1;
 cx q0,q1;
 rz(0.37801308) q1;
 cx q0,q1;
 rx(-pi/2) q0;
 rx(-pi/2) q1;
}
qreg q0[2];
creg c0[2];
h q0[0];
ryy(1.2182379) q0[0],q0[1];
measure q0[0] -> c0[0];
measure q0[1] -> c0[1];
"""

qc = QuantumCircuit.from_qasm_str(qasm_str)
print(qc)


target = Target.from_configuration(["r", "cz", "measure"], num_qubits=5, coupling_map=CouplingMap.from_line(5))
pm = generate_preset_pass_manager(target)
print(pm.run(qc))
```

### What should happen?

The code should not panic, the custom ryy gate resulting from the qasm loading should be either consistently be interpreted as the standard gate or not.

### Any suggestions?

I actually caught this in my code review of the most recent change to the logic here: https://github.com/Qiskit/qiskit/pull/13874#pullrequestreview-2662626939 but didn't think it would result in a panic/error and was just inefficient, I didn't consider overloaded gate names. We can do as I suggested though and use the rust `StandardGate` type to avoid the panic and handle it consistently in the commutation checker.

mtreinish:
### Summary

This commit reworks the internals of the CommutationChecker to not rely on operation name except for where we do a lookup by name in the commutation library provided (which is the only key available to support custom gates). This fixes the case where a custom gate that overloads the standard gate name, previously the code would assume it to be a standard gate and internally panic when it wasn't. When working with standard gates (or standard instructions) we don't need to rely on string matching because we can rely on the rust data model to do the heavy lifting for us. This commit moves all the explicit handling of standard gates to use the StandardGate type directly and makes this logic more robust.

This also removes are usage of the once_cell library in qiskit-accelerate because it was used to create a lazy static hashsets of strings which are no longer needed because static lookup tables replace this when we stopped using string comparisons.

Fixes #13988


crates/accelerate/src/commutation_checker.rs
```diff
@@ -287,14 +296,24 @@ impl CommutationChecker {
...
         // if we have rotation gates, we attempt to map them to their generators, for example
         // RX -> X or CPhase -> CZ
-        let (op1, params1, trivial1) = map_rotation(op1, params1, tol);
+        let (op1_gate, params1, trivial1) = map_rotation(op1, params1, tol);
         if trivial1 {
             return Ok(true);
         }
-        let (op2, params2, trivial2) = map_rotation(op2, params2, tol);
+        let op1 = if let Some(gate) = op1_gate {
+            &OperationRef::StandardGate(gate)
```
Cryoris, line 304: Why not keep `map_rotation` return the `OperationRef` such that we don't have to re-wrap it here? 🙂

crates/accelerate/src/commutation_checker.rs
```diff
@@ -580,15 +614,10 @@ fn matrix_via_operator(py: Python, py_obj: &PyObject) -> PyResult<Array2<Complex
...
-fn is_commutation_skipped<T>(op: &T, params: &[Param]) -> bool
-where
-    T: Operation,
-{
-    op.directive()
-        || SKIPPED_NAMES.contains(&op.name())
-        || params
-            .iter()
-            .any(|x| matches!(x, Param::ParameterExpression(_)))
+fn is_commutation_skipped(params: &[Param]) -> bool {
```
Cryoris: Could we give this a more precise name now that it only checks whether the parameters contain parameter expressions? E.g. `is_parameterized` (analogous to the CircuitInstruction method and some others)

crates/accelerate/src/commutation_checker.rs
```diff
@@ -544,11 +564,25 @@ fn commutation_precheck(
...
-        return None;
+    if let Some(gate_1) = op1.standard_gate() {
+        if let Some(gate_2) = op2.standard_gate() {
+            if SUPPORTED_OP[gate_1 as usize] && SUPPORTED_OP[gate_2 as usize] {
+                return None;
+            }
+        }
+    }
+
+    if matches!(
```
Cryoris, line 575: This is now a logical change since it will eagerly return `false` on instructions, which is fine here since we don't check commutations between directives and these don't have defined matrices. Should we also check for `OperationRef::Operation` here, which also doesn't have a matrix and is not in the commutation library (it isn't right?)?

Cryoris (commented):
That's more elegant plus also correct 😛 I left some minor comments below.

Should we add an upgrade reno stating the change in behavior that custom gates matching standard gate names will no longer be fast-tracked through the commutation checker? I'm imaging a scenario where someone creates a custom gate named e.g. `"ryy"` and doesn't give it a matrix -- then it would've worked before 2.0 but doesn't anymore now, right?

crates/accelerate/src/commutation_checker.rs
mtreinish: It's a lifetime issue, the lifetime of an `OperationRef` we create in `map_rotation` wouldn't be long enough because it would be dropped when the function returns. The issue is we only have an `&OperationRef` passed in to that function, so we can't return an unconditional `&OperationRef` like in the previous signature since we have an owned `StandardGate` internally. I could probably change the signature to be `Option<OperationRef>` but it didn't seem to make a big difference to wrap it in the enum here vs there.

crates/accelerate/src/commutation_checker.rs
mtreinish: I didn't view it as a logical change, just a small efficiency change while I was removing string usage, because we were matching a bit too restrictively before here. We know that no `Instruction` either standard (like measure, reset, etc) or from python (like `Intialize`). Since as you say we can't have a commutation any instruction as they're explicitly non-unitary.

I left `Operation` because I thought they could have a matrix (like `Clifford`) but looking at the operation trait for `PyOperation` we have it hard coded to `None`  https://github.com/Qiskit/qiskit/blob/6f317f953363e076a998f4e0960196ec605b4fb2/crates/circuit/src/operations.rs#L2726-L2728 so you're correct we can add it here. It does make me wonder what we should do about `Clifford` though

mtreinish:
> Should we add an upgrade reno stating the change in behavior that custom gates matching standard gate names will no longer be fast-tracked through the commutation checker? I'm imaging a scenario where someone creates a custom gate named e.g. "ryy" and doesn't give it a matrix -- then it would've worked before 2.0 but doesn't anymore now, right?

I don't think it's necessary in this case, it's not really supported to have a non-standard gate reuse a name like this. Its only the qasm path which make me think we should fix this. There is no API guarantee that the commutation checker will fast track anything either, as long as it's correct output that's fine. In the case of a custom "ryy" gate we should fall back to matrix based approach (assuming it has a definition or explicit matrix set) so it should all work as documented.

crates/accelerate/src/commutation_checker.rs
mtreinish: Done in: https://github.com/Qiskit/qiskit/pull/13991/commits/822f8beffb4b0bfb6f6e329c994f3483c9bc228f

crates/accelerate/src/commutation_checker.rs
Cryoris: Hmm yes I based this off the Rust implementation for when operations always return `None`... but I'm fine keeping the code logically correct as is, since we can add operations that have matrices 👍🏻 

(Otherwise it would be very easy to introduce a bug if we forget to remove the `Operation` case here once we add e.g. Cliffords to Rust)

Cryoris (approved):

PR #13992, opened 2025-03-11
Attempt to fix MacOS PGO build

Cryoris:
### Summary

PGO fails when we build `qiskit-cext` both as `cdylib` in addition to `rlib`, due to missing Python symbols. This could be due to us not setting the `pyo3/extension-module` correctly; which we want enabled when we compile the crate for use in the Python extension crate `qiskit-pyext`, but don't want for the standalone `cdylib` usage. 

This PR changes to only building the `rlib` per default, meaning that building the `qiskit-pyext` crate (and running PGO) will not build the `cdylib`. To build the `cdylib` for C standalone mode, one has to request that crate type via `cargo rustc .... --crate-type cdylib`. We're investigating further solutions but this now fixes PGO and the C standalone tests pass as well.

### Details and comments

Failed previous attempt:
* Link Python library manually in the PGO build, which works out locally. This is done inside `build_pgo.sh` instead of the `wheels-build.yml` to ensure the correct Python library is found. This is more of a backup solution, it would be nicer if PyO3/Rust would be able to link the required libraries automatically (maybe we're missing some flag). Possibly this needs an additional `DYLD_LIBRARY_PATH` export, but let's see what the wheels say...

### DIFF
--- a/.github/workflows/wheels-build.yml
+++ b/.github/workflows/wheels-build.yml
@@ -107,11 +107,18 @@ jobs:
               INSTALL_RUST_PATH=tools/install_rust.sh
               RUST_TOOLCHAIN=stable
           fi
+          if [[ `uname -s` == "Darwin" ]] ; then
+              PYLIB=$(python -c "import sysconfig; print(sysconfig.get_config_var('LIBDIR'))")
+              PYNAME=$(find $PYLIB -maxdepth 1 -name "libpython*" | grep -oE "[^/]+$" | grep -oE "python[0-9]+\.[0-9]+" || echo "python")
+              LDFLAGS="-L $PYLIB -l $PYNAME"
+          else
+              LDFLAGS=""
+          fi
           cat >>"$GITHUB_ENV" <<EOF
           CIBW_BEFORE_ALL_LINUX=yum install -y wget && {package}/$INSTALL_RUST_PATH
           CIBW_BEFORE_BUILD=bash ./tools/build_pgo.sh $PGO_WORK_DIR $PGO_OUT_PATH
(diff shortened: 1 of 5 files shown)


ElePT:
May the wheels be with you

Cryoris:
They were not (quite)... the `build_pgo.sh` script now runs successfully but it's still not finding the libraries in the final build of the optimized binary.

Cryoris:
As side note: changing the libtype of `qiskit-cext` from `cdylib` to `staticlib` works without requiring any library setting.

mtreinish (commented):
This seems like a viable path, it also fixes the linux aarc64 build which is good. Did you test that all of the examples of both standalone mode and hybrid/embedded mode for the C API work correctly with this change?

Also, do we need to update any of the documentation around the cext building now with this change. `cargo build` no longer will create a c dyn lib so you can't just cargo build and get the artifact directly. Even if we say use the makefile it might be good to add something to the readme on `cext` documenting this.

Cryoris:
Yeah I've tested the standalone (which should also be covered by the tests, right?) and the embedded mode 🙂 Maybe we can create a small repo for this kind of integration test in the embedded mode.

Re: documentation -- you mean the installation instructions and any documentation we want to add to the READMEs, but not any currently existing documentation, right?

mtreinish (approved):

kevinhartman (approved):
This seems like a good workaround.

I'm curious though to see the exact `rustc` invocation that's happening via PGO that causes the original issue. It'd seem to me that some kind of program image (either an executable or a dynamic library) is being built with `pyo3/extension-module` enabled (and thus, without `libpython` on the link line).

mtreinish:
For the record I backported this PR to `stable/2.0` as part of https://github.com/Qiskit/qiskit/pull/14023 which is embedded in that PR in the interest of time to expedite tagging 2.0.0rc2.

PR #13993, opened 2025-03-11
Remove: Unnecessary `py` tokens in `InstructionProperties`.

raynelfss:
### Summary
The following PR removes some leftover `py` tokens from  a couple of `InstructionProperties` methods in Rust.


mtreinish (approved):
Seems straightforward enough

PR #13995, opened 2025-03-11
Use `OnceLock` to cache a `NormalOperation` in `Target`

raynelfss:
### Summary
The following commits replace the `op_object` attribute of `NormalOperation` with a `OnceLock` to allow the python version to be either built dynamically or cached from its original source.

### Details and comments

As we plan to use `PackedOperation` to add instructions to a `Target` in Rust. We should not require having a python instance to exist within. The following commits fix that by ensuring that a python instruction is not required, and if so it can be built based on a `PackedOperation`'s internal data.

This should not have any visible effects on the current implementations, but can be proven to work via the added unit-tests.


crates/accelerate/src/target_transpiler/mod.rs
```diff
@@ -92,7 +92,47 @@ impl TargetOperation {
...
+}
+
+impl From<(PackedOperation, SmallVec<[Param; 3]>)> for NormalOperation {
+    fn from(value: (PackedOperation, SmallVec<[Param; 3]>)) -> Self {
+        Self {
+            operation: value.0,
+            params: value.1,
+            op_object: OnceLock::new(),
+        }
+    }
```
kevinhartman: I think it's usually better to only define `From` for a 1:1 sort of conversion. What you've got here looks more like multiple constructors for a `NormalOperation` to be used depending on which ingredients you want in the result.

Could you make these an explicit `from_` method on `NormalOperation` instead? E.g.

```rust
impl NormalOperation {
    fn from_packed(value: PackedOperation, params: Option<SmallVec<...>>) -> Self {
        todo!()
    }
}

File: crates/accelerate/src/target_transpiler/mod.rs [hunk 2]
```diff
@@ -1307,3 +1376,118 @@ pub fn target(m: &Bound<PyModule>) -> PyResult<()> {
...
+    use qiskit_circuit::{imports::UNITARY_GATE, packed_instruction::PackedOperation};
+    use smallvec::smallvec;
+
+    static PY_RXGATE: ImportOnceCell = ImportOnceCell::new("qiskit.circuit.library", "RXGate");
+    static BARRIER: ImportOnceCell = ImportOnceCell::new("qiskit.circuit.library", "Barrier");
+    static IMPORT_LOCK: Mutex<()> = Mutex::new(());
+
+    #[test]
+    fn add_standard_gate_without_python() -> PyResult<()> {
+        let _lock = IMPORT_LOCK
```
kevinhartman: There was some discussion a while back about having `qiskit` get imported up front to avoid this sort of thing:

https://github.com/Qiskit/qiskit/pull/13776#issuecomment-2637618226

I didn't bother submitting a PR to do that since there's also the issue of `PyObject`s not being extractable to Rust `pyclass` structs due to ABI compat issues (i.e. in the Rust tests, our `pyclass` structs come from the `rlib`, but the structs that back `PyObject`s coming from the actual Qiskit code originate from the `cdylib` linked into the Python process).

My assumption was that folks probably would just not really use PyO3 Rust unit testing for anything complex. But, it looks like you have a use case here. Should I make a PR to avoid the deadlock issue?

kevinhartman (commented):
This looks good to me.

My only issue really is with the `IMPORT_LOCK` mutex in the unit tests. See my comment there. I think we can likely avoid needing it, but it'd require a tweak to our test harness.

crates/accelerate/src/target_transpiler/mod.rs
raynelfss: If it does solve the issue, I would love to have a PR that does that. It would be pretty helpful.

kevinhartman: Done in #14200.

I tried it with this PR removing the mutex and it seemed to work 🙂/

jakelishman: I commented on #14200, but just to centralise: if we do this, I'd agree that #14200 is a better strategy. That said, I'm not convinced that we ought to use Python in this test. I'm not entirely sure what the use of Python is buying us - if we test the Rust objects from Rust, is there a really serious risk that "object created from Rust" might not produce the correct Python object later in a way that would not be caught by regular tests of the creation of Python-free `StandardGate` objects?

Even if this test can in theory work at this moment in time, I'm not crazy about doing more testing where `_accelerate` is imported from apython in Rust, because we know that that's not ABI compatible and feels like an accident waiting to happen.

kevinhartman: IMO, #14200 is a nice-to-have if we're going to keep PyO3 unit testing around for the time being. The main use-case I'd see for it is as a stop-gap measure for cases where our internal Rust code relies on importing parts of our data model that aren't yet native to Rust.

But in this particular case, I'd agree that we're not getting much by testing that the Rust-backed operations from the target are equivalent to gates constructed via Python. I'd rather expect any testing here to be poking at the `Target` Rust instance to make sure the instructions were added successfully (no Python involved), if you're to have any testing here at all.

crates/accelerate/src/target_transpiler/mod.rs
```diff
@@ -84,6 +84,17 @@ impl TargetOperation {
             }
         }
     }
+
+    /// Creates a of [TargetOperation] from an instance of [PackedOperation]
```
kevinhartman:
```suggestion
    /// Creates a [TargetOperation] from an instance of [PackedOperation]
```

crates/accelerate/src/target_transpiler/mod.rs
```diff
@@ -101,7 +141,11 @@ impl<'py> IntoPyObject<'py> for NormalOperation {
     type Error = PyErr;
 
     fn into_pyobject(self, py: Python<'py>) -> Result<Self::Output, Self::Error> {
-        Ok(self.op_object.bind(py).clone())
+        Ok(self
+            .op_object
+            .get_or_init(|| self.create_py_op(py, None).unwrap())
```
kevinhartman: Should we store `op_object` as a `PyResult<PyObject>` instead so we can propagate the error if `create_py_op` fails?

test/python/transpiler/test_target.py
```diff
@@ -1201,6 +1202,26 @@ def test_target_no_num_qubits_qubit_properties(self):
...
+        unitary = UnitaryGate([[0, 1], [1, 0]])
+
+        # Create Target and add rest of instructions.
+        target = Target()
+        target.add_instruction(standard_gate)
+        target.add_instruction(barrier)
+        target.add_instruction(unitary)
+
+        # Check the gate instances are working as expected
+        self.assertEqual(target.operation_from_name("rx"), target._raw_operation_from_name("rx"))
```
kevinhartman, line 1217: You might also want to make sure these aren't `None` or something like that.

raynelfss: As per documentation, if `target.operation_from_name()` doesn't find that operation within the `Target` it will throw an exception. So this test would fail if that was the case.

crates/accelerate/src/target_transpiler/mod.rs
raynelfss: I think it might not be necessary, I could just extract the operation before setting it instead.

raynelfss: Fixed in 3547921

raynelfss: After an internal conversation, we have gone with storing a `PyResult`. Fixed in cd8e63b.

kevinhartman (approved):
LGTM, thanks for the changes!

PR #13997, opened 2025-03-11
Add missing  `__getitem__` and other attributes to `BitLocations`

### LINKED ISSUE #13996: BitLocation is no longer indexable
### Environment

- **Qiskit version**: 

RC1

### What is happening?

In <=1.4, `BitLocation`, being a `namedtuple`, is indexable. With no deprecation period, it is now not indexable. 

You can reproduce by running this snippet on main
```python

circuit = QuantumCircuit(1)
circuit.find_bit(circuit.qubits[0])[0]
```

### What should happen?

The snippet should not raise an IndexError.

### Any suggestions?

Implement the getitem dunder.

raynelfss:
Fixes #13996

### Summary
We accidentaly removed the capability of indexing a `BitLocation` object. These commits add that functionality back!

### Details and comments
In an oversight after #13860, when switching to use the pyo3 friendly `BitLocations` which is not an instance of `namedtuple` we did not extend it to include sequence methods `__getitem__` and `__len__`, which results in breaking api changes.

The following commits add those methods to `BitLocations` and also labels the class correctly as a `sequence` in PyO3.

### DIFF
--- a/crates/circuit/src/dag_circuit.rs
+++ b/crates/circuit/src/dag_circuit.rs
@@ -399,7 +399,7 @@ impl PyBitLocations {
             match index {
                 0 => self.index.into_py_any(py),
                 1 => Ok(self.registers.clone_ref(py).into_any()),
-                _ => Err(PyIndexError::new_err("tuple index out of range")),
+                _ => Err(PyIndexError::new_err("index out of range")),
             }
         };
         if let Ok(index) = index.with_len(2) {
@@ -409,7 +409,7 @@ impl PyBitLocations {
                     .map(|obj| obj.into_any().unbind()),
             }
         } else {
-            Err(PyIndexError::new_err("tuple index out of range"))
+            Err(PyIndexError::new_err("index out of range"))
         }
     }
 
--- a/crates/circuit/src/dag_circuit.rs
+++ b/crates/circuit/src/dag_circuit.rs
@@ -34,6 +34,7 @@ use crate::operations::{ArrayType, Operation, OperationRef, Param, PyInstruction
 use crate::packed_instruction::{PackedInstruction, PackedOperation};
 use crate::register_data::RegisterData;
 use crate::rustworkx_core_vnext::isomorphism;
+use crate::slice::PySequenceIndex;
 use crate::{BitType, Clbit, Qubit, TupleLikeArg};
 
 use hashbrown::{HashMap, HashSet};
@@ -343,7 +344,7 @@ fn reject_new_register(reg: &Bound<PyAny>) -> PyResult<()> {
     )))
 }
 
-#[pyclass(name = "BitLocations", module = "qiskit._accelerate.circuit")]
+#[pyclass(name = "BitLocations", module = "qiskit._accelerate.circuit", sequence)]
 #[derive(Clone, Debug)]
 pub struct PyBitLocations {
     #[pyo3(get)]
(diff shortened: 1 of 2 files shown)


crates/circuit/src/dag_circuit.rs
```diff
@@ -392,6 +393,30 @@ impl PyBitLocations {
...
+                crate::slice::SequenceIndex::Int(index) => slf.getattr(names[index]),
+                _ => PyTuple::new(
+                    slf.py(),
+                    index.iter().map(|idx| slf.getattr(names[idx]).unwrap()),
+                )
+                .map(|obj| obj.into_any()),
+            }
+        } else {
+            Err(PyIndexError::new_err("tuple index out of range"))
+        }
```
mtreinish: This is a bit more complicated than it needs to be. You don't need to access everything via python attr interface like this, can't you just do the equivalent of something like:

```rust
 if index == 0 {
     return self.index.into_py_any(py);
 } else {
     return self.registers.clone_ref(py).into_any();
  }
```

That also lets you simplify the function signature and just use `&self` and probably elide the lifetimes too.

raynelfss: Oh I deliberately chose this over using `&self` to avoid a conversion step. If you believe that the conversion cost is worth paying for clearer code, then I'm down to change this.

mtreinish: I'm not sure what conversion cost you're worried about here. Using &self is the equivalent of basically just doing `.borrow()` to get `&Self`. The way this is written write now has a lot of extra overhead because you go back to python for every attribute access when it's faster to just get the fields from the rust struct directly. There is a conversion from the rust integer to a python integer but you have that either way.

raynelfss: Got it, I have addressed your comments in c315871

test/python/circuit/test_circuit_find_bit.py
```diff
@@ -165,3 +165,53 @@ def test_multiple_register_from_reg(self, reg_type):
                 self.assertEqual(
                     qc.find_bit(bit), (circ_idx, [(reg2, idx), (even_reg, circ_idx // 2)])
                 )
+
+
+@ddt
```
mtreinish: You're not using ddt to build parameterized tests in this class so you don't need this decorator.

```suggestion
```

crates/circuit/src/dag_circuit.rs
```diff
@@ -392,6 +393,30 @@ impl PyBitLocations {
...
+            }
+        };
+        if let Ok(index) = index.with_len(2) {
+            match index {
+                crate::slice::SequenceIndex::Int(index) => getter(index),
+                _ => PyTuple::new(py, index.iter().map(|idx| getter(idx).unwrap()))
+                    .map(|obj| obj.into_any().unbind()),
+            }
+        } else {
+            Err(PyIndexError::new_err("tuple index out of range"))
```
mtreinish: You don't need to say `tuple` here anymore. It's not typed as a tuple so it's a bit weird to say this. We don't need the error message to be exactly the same as when it was a named tuple. I'd just drop it and say:

```suggestion
            Err(PyIndexError::new_err("index out of range"))
```

crates/circuit/src/dag_circuit.rs
```diff
@@ -392,6 +393,30 @@ impl PyBitLocations {
     fn __getnewargs__(slf: Bound<Self>) -> PyResult<(Bound<PyAny>, Bound<PyAny>)> {
         Ok((slf.getattr("index")?, slf.getattr("registers")?))
     }
+
+    fn __getitem__(&self, py: Python, index: PySequenceIndex<'_>) -> PyResult<PyObject> {
+        let getter = |index: usize| -> PyResult<PyObject> {
+            match index {
+                0 => self.index.into_py_any(py),
+                1 => Ok(self.registers.clone_ref(py).into_any()),
+                _ => Err(PyIndexError::new_err("tuple index out of range")),
```
```suggestion
                _ => Err(PyIndexError::new_err("index out of range")),
```

mtreinish (commented):
This looks better to me, thanks for the updates. Just a few small inline comments

mtreinish (approved):

PR #13980, opened 2025-03-10
Preserve bit locations through pickle

mtreinish:
### Summary

Previously when we were pickling a DAGCircuit the bit locations fields were forgotten about. So when we loaded a DAGCircuit from a pickle the bit locations were empty. This would cause any call to find_bit() to raise an error because there was no entry for any of the bits in the dag. This commit fixes this by reconstructing the bit locations on __setstate__ by iterating over the dag's qubits and clbits.

### Details and comments

Fixes #13976

PR #14000, opened 2025-03-11
Unify identifer handling of `Var` and `Stretch` in `DAGCircuit`

### LINKED ISSUE #13998: Stretch capture order should not matter in circuit equality
### Environment

- **Qiskit version**: 2.0.0rc
- **Python version**: Any
- **Operating system**: Any

### What is happening?

When circuits with stretch captures are compared, the order of the captures is part of the equality. This should not be the case.

Create two circuits with the same captures, but added in a different order. Observe that they are not equal.

### What should happen?

The two circuit should be equal.

### Any suggestions?

_No response_

### LINKED ISSUE #13999: `DAGCircuit` does not properly pickle stretches
### Environment

- **Qiskit version**: 2.0.0rc
- **Python version**: Any
- **Operating system**:  Any

### What is happening?

The implementations of `DAGCircuit::__getstate__` and `DAGCircuit::__setstate__` are missing logic to for stretch identifiers.

Pickling and unpickling a `DAGCircuit` with stretches should result in the stretches getting dropped.

### What should happen?

The stretches should be preserved after unpickling.

### Any suggestions?

_No response_

kevinhartman:
### Summary
During testing, I've found two bugs with `DAGCircuit`'s handling of stretches:

1. #13998 
2. #13999 

They are fixed here.

### Details and comments
 I've also reworked `DAGCircuit`'s tracking of stretches to be the same as what we do for vars. Namely, there's now just an `identifier_info` field (which replaces `var_info`) which makes it easier to detect name collisions between all identifiers in tracking, and also should make it easier to support native handling of stretches in the future. 

To make things a bit cleaner, I've moved `Var` to `lib.rs` and defined a macro `impl_circuit_identifier` to reduce boilerplate between our `Qubit`, `Clbit`, `Var`, and now `Stretch` identifier types.

I've also moved the logic of pickling identifier info tracking from `__getstate__` and `__setstate__` to `FromPyObject` and `IntoPyObject` implementations.

#### To-do

- [x] It looks to me like we don't have any testing of pickling / unpickling a `DAGCircuit`. Can that possibly be true?


mtreinish:
>  It looks to me like we don't have any testing of pickling / unpickling a DAGCircuit. Can that possibly be true?

If we have any we get it indirectly through tests that are using multiprocessing (like transpiling > 1 circuit) because pickle will be used for IPC or deep copy which also internally uses pickle. That being said see #13980 for some explicit tests I had to add for another pickle bug that's been around since 1.3 (when we ported DAGCircuit to rust).

kevinhartman:
> see https://github.com/Qiskit/qiskit/pull/13980 for some explicit tests I had to add for another pickle bug that's been around since 1.3 (when we ported DAGCircuit to rust).

Great, thanks! I've added a few tests based on those here to validate that stretches are preserved after pickling and deepcopy.

I also tested vars locally, but I found that there's a bug there that predates 2.0 so those tests aren't added in this PR (we ought to address that separately):

- #14007

crates/circuit/src/dag_circuit.rs
```diff
@@ -239,11 +199,14 @@ pub struct DAGCircuit {
...
+
+    identifier_info: HashMap<String, DAGIdentifierInfo>,
+
+    // Indexed by DAGVarType as usize
     vars_by_type: [Py<PySet>; 3],
 
-    captured_stretches: IndexMap<String, Py<PyAny>, RandomState>,
-    declared_stretches: IndexMap<String, Py<PyAny>, RandomState>,
+    captured_stretches: Py<PySet>,
+    declared_stretches: Py<PyList>,
```
mtreinish: After #14069 we're not storing bare python objects in the dag they're all wrapped in either an `Option<>` or some other struct so we can defer the creation to a context where we have the py token to ensure we have a path to create a dag without the pytoken. So I don't think this will work when you rebase to main. We can wrap it in an `Option`, `OnceLock` or something else and maybe just make a `ContainedStretches` struct, or something similar to manage the lazy creation of them.

crates/circuit/src/dag_circuit.rs
```diff
@@ -434,14 +501,14 @@ impl DAGCircuit {
...
+            identifier_info: HashMap::new(),
             vars_by_type: [
                 PySet::empty(py)?.unbind(),
                 PySet::empty(py)?.unbind(),
                 PySet::empty(py)?.unbind(),
             ],
-            captured_stretches: IndexMap::default(),
-            declared_stretches: IndexMap::default(),
+            captured_stretches: PySet::empty(py)?.unbind(),
+            declared_stretches: PyList::empty(py).unbind(),
```
mtreinish: We don't have a py token here anymore so this will fail on rebase.

crates/circuit/src/dag_circuit.rs
```diff
@@ -2206,8 +2269,8 @@ impl DAGCircuit {
     /// but was changed by issue #2564 to return number of qubits + clbits
     /// with the new function DAGCircuit.num_qubits replacing the former
     /// semantic of DAGCircuit.width().
-    fn width(&self) -> usize {
-        self.qubits.len() + self.clbits.len() + self.vars_info.len()
+    fn width(&self, py: Python) -> usize {
```
mtreinish, line 2190: I assume this is something that won't be required after #14068 is done?

crates/circuit/src/dag_circuit.rs
```diff
@@ -4383,8 +4439,8 @@ impl DAGCircuit {
     /// Add a captured stretch to the circuit.
     ///
     /// Args:
-    ///     var: the stretch to add.
-    fn add_captured_stretch(&mut self, py: Python, var: &Bound<PyAny>) -> PyResult<()> {
+    ///     stretch: the stretch to add.
+    fn add_captured_stretch(&mut self, py: Python, stretch: &Bound<PyAny>) -> PyResult<()> {
```
mtreinish: This is actually a breaking api change right? Python's arguments by default are both positional and by name, and we didn't declare the signature of this pymethod to just deal with this positionally the name can be used to do `dag.add_captured_stretch(var=stretch)` which would be broken by this change. So I don't think we can do this change until 3.0 since we released 2.0.0rc1 and rc2 with a fixed API surface we're stuck with this name for now.

crates/circuit/src/dag_circuit.rs
```diff
@@ -4433,40 +4487,38 @@ impl DAGCircuit {
...
-        let var_name: String = var.getattr("name")?.extract::<String>()?;
-        if self.vars_info.contains_key(&var_name) {
-            return Err(DAGCircuitError::new_err(
-                "cannot add stretch as its name shadows an existing identifier",
-            ));
-        }
-        if let Some(previous) = self.declared_stretches.get(&var_name) {
-            if var.eq(previous)? {
+    ///     stretch: the stretch to add.
+    fn add_declared_stretch(&mut self, py: Python, stretch: &Bound<PyAny>) -> PyResult<()> {
```
mtreinish: This is the same problem with this being a breaking change

crates/circuit/src/dag_circuit.rs
```diff
@@ -4491,40 +4543,61 @@ impl DAGCircuit {
...
+    #[getter]
+    fn num_captured_stretches(&self, py: Python) -> usize {
+        self.captured_stretches.bind(py).len()
+    }
+
+    /// Number of declared local stretches tracked by the circuit.
+    #[getter]
+    fn num_declared_stretches(&self, py: Python) -> usize {
+        self.declared_stretches.bind(py).len()
+    }
```
mtreinish: If this is for 2.0 I don't think we can backport this since these are new attributes for the dag. This would only be for 2.1. They make total sense to add, but we might want to split this PR out into multiple parts.

mtreinish (commented):
I know you've got a pending rebase coming so let me just leave my comments from part way through the review and I'll look again after the rebase is up. That being said I'm wondering if we should concentrate on a dedicated 2.0 fix in a direct PR on stable/2.0 and leave the larger changes in this PR for just 2.1?

crates/circuit/src/dag_circuit.rs
kevinhartman: Should be possible, though it's not clear to me if it should happen in that PR or a follow-up. That PR (so far) ports the Python `Expr` and `Type` hierarchies to Rust, but does not yet change over how we track things in the circuit.

crates/circuit/src/dag_circuit.rs
kevinhartman: I was originally hoping this PR would land earlier, but I've renamed `stretch` to `var` in both signatures for now: 7bba59064d854bf3f115f0186ca5430380b27656

crates/circuit/src/dag_circuit.rs
kevinhartman: Makes sense. I've removed the backport label and will add a release note for this PR for 2.1.

kevinhartman:
@mtreinish

> That being said I'm wondering if we should concentrate on a dedicated 2.0 fix in a direct PR on stable/2.0 and leave the larger changes in this PR for just 2.1?

That sounds good. I've removed the backport label from this PR and will open a new one against `stable/2.0` that takes the tests from here and fixes things via other means.

crates/circuit/src/dag_circuit.rs
```diff
@@ -4463,40 +4550,38 @@ impl DAGCircuit {
...
-    fn add_declared_stretch(&mut self, var: &Bound<PyAny>) -> PyResult<()> {
-        let var_name: String = var.getattr("name")?.extract::<String>()?;
-        if self.vars_info.contains_key(&var_name) {
-            return Err(DAGCircuitError::new_err(
-                "cannot add stretch as its name shadows an existing identifier",
-            ));
-        }
-        if let Some(previous) = self.declared_stretches.get(&var_name) {
-            if var.eq(previous)? {
+    ///     stretch: the stretch to add.
```
```suggestion
    ///     var: the stretch to add.
```

releasenotes/notes/dag-stretch-counters-c26f39e3ea9b85a5.yaml
```diff
@@ -0,0 +1,6 @@
+---
+features_transpiler:
+  - |
+    Adds :class:`~.dagcircuit.DAGCircuit` methods :meth:`~.dagcircuit.DAGCircuit.num_stretches`,
+    :meth:`~.dagcircuit.DAGCircuit.num_captured_stretches`,
+    and :meth:`~.dagcircuit.DAGCircuit.num_declared_stretches`.
```
mtreinish:
```suggestion
    Adds :class:`~.dagcircuit.DAGCircuit` attributes :attr:`~.DAGCircuit.num_stretches`,
    :attr:`~.DAGCircuit.num_captured_stretches`,
    and :attr:`~.DAGCircuit.num_declared_stretches`.
```

crates/circuit/src/dag_circuit.rs
```diff
@@ -4416,38 +4505,36 @@ impl DAGCircuit {
     /// Add a captured stretch to the circuit.
     ///
     /// Args:
-    ///     var: the stretch to add.
+    ///     stretch: the stretch to add.
```
```suggestion
    ///     var: the stretch to add.
```

crates/circuit/src/dag_circuit.rs
```diff
@@ -415,6 +380,103 @@ struct DAGVarInfo {
...
+                0 => DAGVarType::Input,
+                1 => DAGVarType::Capture,
+                2 => DAGVarType::Declare,
+                _ => return Err(PyValueError::new_err("Invalid var type")),
+            },
+            in_node: NodeIndex::new(val_tuple.get_item(2)?.extract()?),
+            out_node: NodeIndex::new(val_tuple.get_item(3)?.extract()?),
+        })
+    }
+}
```
mtreinish: Are these just for pickling? If so I think it might be better to have them as a dedicate method. The reason is I'm worried these might get confused with the actual python object being represented here. I did this pattern for `Wire` with `to_pickle` and `from_pickle`.

crates/circuit/src/dag_circuit.rs
```diff
@@ -467,6 +529,45 @@ impl<'py> VarsByType {
     }
 }
 
+/// A container struct for the Stretches in a dag grouped by their types
+#[derive(Clone, Debug)]
+struct StretchesByType {
+    capture: OnceLock<Py<PySet>>,
+    declare: OnceLock<Py<PyList>>,
```
mtreinish, line 516: One is a set the other is a list?

crates/circuit/src/dag_circuit.rs
```diff
@@ -4517,40 +4602,61 @@ impl DAGCircuit {
...
     /// Is this stretch in the DAG?
     ///
     /// Args:
-    ///     var: the stretch or name to check.
-    fn has_stretch(&self, var: &Bound<PyAny>) -> PyResult<bool> {
-        match var.extract::<String>() {
-            Ok(name) => Ok(self.declared_stretches.contains_key(&name)
-                || self.captured_stretches.contains_key(&name)),
+    ///     stretch: the stretch or name to check.
+    fn has_stretch(&self, stretch: &Bound<PyAny>) -> PyResult<bool> {
```
```suggestion
    ///     var: the stretch or name to check.
    fn has_stretch(&self, var: &Bound<PyAny>) -> PyResult<bool> {
```

crates/circuit/src/dag_circuit.rs
mtreinish: Yeah as we discussed offline I think it's fine to do this in a follow up after #14068. 

I'm just mildly concerned that this is adding a deeper dependency on Python for something we should be able to figure out structurally from the dag without Python.

kevinhartman: Agreed. We _do_ have this information in a way that doesn't need Python as it is, but it'd be inefficient since we'd have to walk over all of the `DAGIdentifierInfo`s and count which ones are variables.

I will find some faster solution in a follow-up that ditches use of the `Python` token.

crates/circuit/src/dag_circuit.rs
kevinhartman: Correct. That's because stretch declaration order matters, whereas when you capture stretches, it doesn't matter what order they were originally defined in (we use this in `__eq__`).

crates/circuit/src/dag_circuit.rs
kevinhartman: Done in 02cbc96f8ce8ff2880c98668b00bf9e34e14b8fc

crates/circuit/src/dag_circuit.rs
```diff
@@ -413,6 +378,85 @@ struct DAGVarInfo {
...
+impl DAGIdentifierInfo {
+    fn to_pickle(&self, py: Python) -> PyResult<PyObject> {
+        match self {
+            DAGIdentifierInfo::Stretch(info) => (0, info.to_pickle(py)?).into_py_any(py),
+            DAGIdentifierInfo::Var(info) => (1, info.to_pickle(py)?).into_py_any(py),
+        }
+    }
+
+    fn from_pickle(ob: &Bound<PyAny>) -> PyResult<Self> {
+        let val_tuple = ob.downcast::<PyTuple>()?;
```
mtreinish, line 447: Is there a reason to not just do?:
```suggestion
    fn from_pickle(val_tuple: &Bound<PyTuple>) -> PyResult<Self> {
```

crates/circuit/src/lib.rs
```diff
@@ -37,53 +37,60 @@ use pyo3::prelude::*;
...
-    }
-}
+#[derive(Copy, Clone, Debug, Hash, Ord, PartialOrd, Eq, PartialEq, FromPyObject)]
+pub struct Clbit(pub u32);
+
+#[derive(Copy, Clone, Debug, Hash, Eq, PartialEq)]
+pub struct Var(u32);
+
+#[derive(Copy, Clone, Debug, Hash, Eq, PartialEq)]
+pub struct Stretch(u32);
```
mtreinish, line 50: Is there a reason to not make the inner u32 pub here?

crates/circuit/src/lib.rs
```diff
@@ -37,53 +37,60 @@ use pyo3::prelude::*;
 use pyo3::types::{PySequence, PyTuple};
 use pyo3::PyTypeInfo;
 
-pub type BitType = u32;
```
mtreinish, line 40: Is the reason to remove this because it's used for more than bits now?

test/python/dagcircuit/test_dagcircuit.py
```diff
@@ -10,6 +10,8 @@
 # copyright notice, and modified files need to carry a notice indicating
 # that they have been altered from the originals.
 
+# pylint: disable=invalid-name
```
mtreinish, line 13: What's the invalid name that pylint was complaining about?

crates/circuit/src/lib.rs
kevinhartman: I didn't make it public since it doesn't have real meaning, at least compared to the int of a `Qubit` or `Clbit`.

crates/circuit/src/lib.rs
kevinhartman: Pretty much, yeah.

I'd introduced it originally, but it doesn't feel very Rust-y. I think it wasn't doing much for us.

test/python/dagcircuit/test_dagcircuit.py
kevinhartman: This is, I believe, because of the additional letter-only names I used. Like "c" and "d".

There were already letter-only names, so I figured this should be allowed. Let me know if you have other suggestions :)

crates/circuit/src/dag_circuit.rs
kevinhartman: I think just to be consistent with the other `from_pickle` methods, really. There isn't any auto-extraction possible here anyway, since we store this all in a `PyDict` which, when iterated over, yields `PyAny` pairs.

mtreinish (approved):

PR #14001, opened 2025-03-12
Update CS and CSdg in the equivalence library

ShellyGarion:
### Summary
Update the synthesis of `CS` and `CSdg` in the equivalence library, so that the basis translator will use 2 CX gates for their synthesis and not 3 (a relevant test with `CX` counts should be updated as part of #13961 )

### Details and comments
The current translation (optimization_level=0) of `CS` looks like:
```
     ┌────────────┐                                                     »
q_0: ┤ U(0,0,π/4) ├────────────────■───────────────────■────────────────»
     ├────────────┤┌────────────┐┌─┴─┐┌─────────────┐┌─┴─┐┌────────────┐»
q_1: ┤ U(π/2,0,π) ├┤ U(π/2,0,π) ├┤ X ├┤ U(0,0,-π/4) ├┤ X ├┤ U(0,0,π/4) ├»
     └────────────┘└────────────┘└───┘└─────────────┘└───┘└────────────┘»
«                                 
«q_0: ────────────────────────────
«     ┌────────────┐┌────────────┐
«q_1: ┤ U(π/2,0,π) ├┤ U(π/2,0,π) ├
«     └────────────┘└────────────┘
```

and for `CSdg` it looks like (with 3 `CX` gates):
```
                        ┌────────────┐                                       »
q_0: ────────────────■──┤ U(0,0,π/4) ├──■───────────────────■────────────────»
     ┌────────────┐┌─┴─┐├────────────┤┌─┴─┐┌─────────────┐┌─┴─┐┌────────────┐»
q_1: ┤ U(π/2,0,π) ├┤ X ├┤ U(π/2,0,π) ├┤ X ├┤ U(0,0,-π/4) ├┤ X ├┤ U(0,0,π/4) ├»
     └────────────┘└───┘└────────────┘└───┘└─────────────┘└───┘└────────────┘»
«                                 
«q_0: ────────────────────────────
«     ┌────────────┐┌────────────┐
«q_1: ┤ U(π/2,0,π) ├┤ U(π/2,0,π) ├
«     └────────────┘└────────────┘
```

After this PR, the translation (optimization_level=0) of `CS` should look like:
```
     ┌────────────┐                         
q_0: ┤ U(0,0,π/4) ├──■───────────────────■──
     ├────────────┤┌─┴─┐┌─────────────┐┌─┴─┐
q_1: ┤ U(0,0,π/4) ├┤ X ├┤ U(0,0,-π/4) ├┤ X ├
     └────────────┘└───┘└─────────────┘└───┘
```
and for `CSdg` it should look like (with only 2 `CX` gates):
 ```
                              ┌─────────────┐
q_0: ──■──────────────────■──┤ U(0,0,-π/4) ├
      ┌─┴─┐┌────────────┐┌─┴─┐├─────────────┤
q_1: ┤ X ├┤ U(0,0,π/4) ├┤ X ├┤ U(0,0,-π/4) ├
      └───┘└────────────┘└───┘└─────────────┘
```

### DIFF
--- a/qiskit/circuit/library/standard_gates/equivalence_library.py
+++ b/qiskit/circuit/library/standard_gates/equivalence_library.py
@@ -807,38 +807,6 @@ def_sdg = QuantumCircuit(q)
 def_sdg.append(U1Gate(-pi / 2), [q[0]], [])
 _sel.add_equivalence(SdgGate(), def_sdg)
 
-# CSGate
-#
-#                    ┌───┐
-# q_0: ──■──    q_0: ┤ T ├──■───────────■──
-#      ┌─┴─┐         ├───┤┌─┴─┐┌─────┐┌─┴─┐
-# q_1: ┤ S ├ =  q_1: ┤ T ├┤ X ├┤ Tdg ├┤ X ├
-#      └───┘         └───┘└───┘└─────┘└───┘
-q = QuantumRegister(2, "q")
-def_cs = QuantumCircuit(q)
-def_cs.append(TGate(), [q[0]], [])
-def_cs.append(TGate(), [q[1]], [])
-def_cs.append(CXGate(), [q[0], q[1]], [])
-def_cs.append(TdgGate(), [q[1]], [])
-def_cs.append(CXGate(), [q[0], q[1]], [])
-_sel.add_equivalence(CSGate(), def_cs)
-
-# CSdgGate
-#
-#                                     ┌─────┐
-# q_0: ───■───    q_0: ──■─────────■──┤ Tdg ├
-#      ┌──┴──┐         ┌─┴─┐┌───┐┌─┴─┐├─────┤
-# q_1: ┤ Sdg ├ =  q_1: ┤ X ├┤ T ├┤ X ├┤ Tdg ├
-#      └─────┘         └───┘└───┘└───┘└─────┘
-q = QuantumRegister(2, "q")
-def_csdg = QuantumCircuit(q)
-def_csdg.append(CXGate(), [q[0], q[1]], [])
-def_csdg.append(TGate(), [q[1]], [])
-def_csdg.append(CXGate(), [q[0], q[1]], [])
-def_csdg.append(TdgGate(), [q[0]], [])
-def_csdg.append(TdgGate(), [q[1]], [])
-_sel.add_equivalence(CSdgGate(), def_csdg)
-
 # SdgGate
 #
 #    ┌─────┐        ┌───┐┌───┐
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: releasenotes/notes/update-cs-csdg-in-equivalence-library-c1e70d3246f4aa6d.yaml [hunk 1]
```diff
@@ -0,0 +1,5 @@
+---
+fixes:
```
Cryoris: Isn't this more of an `upgrade` or `features`? It wasn't wrong before, it's just better now 🙂

releasenotes/notes/update-cs-csdg-in-equivalence-library-c1e70d3246f4aa6d.yaml
```diff
@@ -0,0 +1,5 @@
...
+  - |
+    Update the synthesis of :class:`.CSGate` and :class:`.CSdgGate` in the equivalence library
+    so that the basis translator will use only two :class:`.CXGate`s for their synthesis and not 3.
```
Cryoris: `CSGate` was already using 2 only right?

Cryoris (commented):
Nice!

alexanderivrii (commented):
Thanks @ShellyGarion, this is a nice PR. The new equivalences look correct to me and in any case we check all of the equivalences automatically in the CI. Note that technically you will also get two CX-gates if you run transpile with `optimization_level=2`, but you will not necessarily get the nice "T" single-qubit gates. I do think it's worth to add a short test as part of this PR. Also, I am not sure whether this should be considered as a "bugfix" or an "upgrade synthesis" note.

releasenotes/notes/update-cs-csdg-in-equivalence-library-c1e70d3246f4aa6d.yaml
ShellyGarion: yes, but now it uses less 1-qubit gates, I'll fix this

ShellyGarion:
> Thanks @ShellyGarion, this is a nice PR. The new equivalences look correct to me and in any case we check all of the equivalences automatically in the CI. Note that technically you will also get two CX-gates if you run transpile with `optimization_level=2`, but you will not necessarily get the nice "T" single-qubit gates. I do think it's worth to add a short test as part of this PR. Also, I am not sure whether this should be considered as a "bugfix" or an "upgrade synthesis" note.

Note that with `optimization_level=2` it consolidates and re-synthesize the 2-qubit gates, so now this effort is no longer needed.
I don't know where to put a relevant test. There are tests that automatically check the equivalences in the equivalence library, and we'll add tests for CX counts in #13961.

ShellyGarion:
As `CS` and `CSdg` are elements of the [`CNOTDihedral group`](https://docs.quantum.ibm.com/api/qiskit/qiskit.quantum_info.CNOTDihedral), it's useful that their synthesis is based on the generators of this group, which are `T`, `X` and `CX` gates.

qiskit/circuit/library/standard_gates/equivalence_library.py
```diff
@@ -809,29 +809,34 @@ def _cnot_rxx_decompose(plus_ry: bool = True, plus_rxx: bool = True):
...
-#      ┌─┴─┐        ┌───┐┌─┴──┐┌───┐
-# q_1: ┤ S ├ = q_1: ┤ H ├┤ Sx ├┤ H ├
-#      └───┘        └───┘└────┘└───┘
+#                    ┌───┐
+# q_0: ──■──    q_0: ┤ T ├──■───────────■──
+#      ┌─┴─┐         ├───┤┌─┴─┐┌─────┐┌─┴─┐
+# q_1: ┤ S ├ =  q_1: ┤ T ├┤ X ├┤ Tdg ├┤ X ├
+#      └───┘         └───┘└───┘└─────┘└───┘
 q = QuantumRegister(2, "q")
 def_cs = QuantumCircuit(q)
```
Cryoris: This change will break the current behavior that we can do
```python
qc = QuantumCircuit(2)
qc.cs(0, 1)
tqc = transpile(qc, basis_gates=["h", "csx"])
```
Could you instead add your new relation as additional equivalence instead of replacing the old one? The same applies to CSdg below 🙂

releasenotes/notes/update-cs-csdg-in-equivalence-library-c1e70d3246f4aa6d.yaml
```diff
@@ -0,0 +1,7 @@
+---
+upgrade_synthesis:
```
Cryoris: If we add additional equivalences (see comment above) then I think this is now actually a new feature, instead of an upgrade, or what do you think?

Cryoris (commented):
Aside from the comments below, I'm wondering whether `CSGate` and `CSdgGate` should also include the definitions from their `_define` method here, like all other gates. Let's discuss this offline.

releasenotes/notes/update-cs-csdg-in-equivalence-library-c1e70d3246f4aa6d.yaml
```diff
@@ -0,0 +1,7 @@
+---
+features_synthesis:
+  - |
+    Add more equivalences to :class:`.CSGate` and :class:`.CSdgGate` in the equivalence library.
+    Previously, the basis translator used 3 :class:`.CXGate` gates for the synthesis of
+    :class:`.CSdgGate` and now it uses only two :class:`.CXGate`.
+    For both gates we reduce the number of single-qubit gates in their basis translation.
```
alexanderivrii: A nitpick: this paragraph implicitly assumes that you are transpiling to the `["cx", "u"]` basis set, without explicitly mentioning this. Also using both "synthesis" and "translation" within the same sentence might be a bit confusing. Maybe something like this:

```suggestion
    The standard equivalence library has additional equivalences for :class:`.CSGate` and :class:`.CSdgGate`.
    When transpiling to the `["cx", "u"]` basis set, the :class:`.BasisTranslator` transpiler pass now uses
    2  :class:`.CXGate` gates instead of 3 for :class:`.CSdgGate`, and in addition reduces the number of single-qubit
    gates in both cases.
```

alexanderivrii (approved):
Thanks for doing all these updates.

PR #14005, opened 2025-03-12
Create a secure internal path for custom prefixes in registers.

### LINKED ISSUE #14003: Bug in `QuantumCircuit` accessing `ClassicalRegister` attribute
### Environment

- **Qiskit version**: 2.0.0rc1
- **Python version**: 3.9
- **Operating system**: linux

### What is happening?

The Aer unit test suite shows an error trace that points to `quantumcircuit.py`:

```
File "/opt/hostedtoolcache/Python/3.10.16/x64/lib/python3.10/site-packages/qiskit/circuit/quantumcircuit.py", line 3892, in _create_creg
new_creg = ClassicalRegister(length, name=f"{name}{ClassicalRegister.instance_count}")
AttributeError: type object 'qiskit.circuit.ClassicalRegister' has no attribute 'instance_count'
```

This is apparently a typo in the attribute name, which is `instances_count` instead of `'instance_count'`.

Issue found when running `qiskit-aer/test/terra/primitives/test_estimator.py`

### What should happen?

No typo

### Any suggestions?

_No response_

raynelfss:
### Summary
The following commits add the internal method `_new_with_prefix`, which allows for the one-time replacement of the set prefix name and use of the same instance counter.

### Details and comments
Prior implementations would replace the `Register`'s prefix attribute in place which is an unsafe operation. The following commits add a secure path for a provisional replacement of a register's prefix name to fix changed unsafe behavior from #13860 and attempts to fix #14003.

### DIFF
--- a/crates/circuit/src/bit.rs
+++ b/crates/circuit/src/bit.rs
@@ -771,6 +771,25 @@ macro_rules! create_bit_object {
                 })
             }
 
+            /// Allows for the creation of a new register with a provisional prefix and the
+            /// same instance counter.
+            #[pyo3(signature=(size=None, name=None, bits=None))]
+            #[staticmethod]
+            fn _new_with_prefix(
+                py: Python,
+                size: Option<isize>,
+                name: Option<String>,
+                bits: Option<Vec<$bit_struct>>,
+            ) -> PyResult<Py<Self>> {
+                let name =
+                    format!(
+                        "{}{}",
+                        name.unwrap_or(Self::prefix().to_string()),
+                        $reg_struct::anonymous_instance_count().fetch_add(1, Ordering::Relaxed)
+                    );
+                Py::new(py, Self::py_new(size, Some(name), bits)?)
+            }
+
             #[classattr]
             fn prefix() -> &'static str {
                 $pyreg_prefix
--- a/crates/circuit/src/bit.rs
+++ b/crates/circuit/src/bit.rs
@@ -771,7 +771,7 @@ macro_rules! create_bit_object {
                 })
             }
 
-            /// Allows for the creation of a new register with a provisional prefix and the
+            /// Allows for the creation of a new register with a temporary prefix and the
             /// same instance counter.
             #[pyo3(signature=(size=None, name=None, bits=None))]
             #[staticmethod]
--- a/qiskit/circuit/quantumcircuit.py
+++ b/qiskit/circuit/quantumcircuit.py
@@ -3889,7 +3889,7 @@ class QuantumCircuit:
     def _create_creg(self, length: int, name: str) -> ClassicalRegister:
         """Creates a creg, checking if ClassicalRegister with same name exists"""
         if name in [creg.name for creg in self.cregs]:
-            new_creg = ClassicalRegister(length, name=f"{name}{ClassicalRegister.instance_count}")
+            new_creg = ClassicalRegister._new_with_prefix(length, name)
         else:
             new_creg = ClassicalRegister(length, name)
         return new_creg
@@ -3897,7 +3897,7 @@ class QuantumCircuit:
     def _create_qreg(self, length: int, name: str) -> QuantumRegister:
         """Creates a qreg, checking if QuantumRegister with same name exists"""
         if name in [qreg.name for qreg in self.qregs]:
-            new_qreg = QuantumRegister(length, name=f"{name}{QuantumRegister.instance_count}")
+            new_qreg = QuantumRegister._new_with_prefix(length, name)
         else:
             new_qreg = QuantumRegister(length, name)
         return new_qreg
--- a/test/python/circuit/test_circuit_operations.py
+++ b/test/python/circuit/test_circuit_operations.py
@@ -867,7 +867,7 @@ class TestCircuitOperations(QiskitTestCase):
         # Second call should also create a new register
         circuit.measure_all()
         self.assertEqual(len(circuit.cregs), 2)  # Now two cregs
-        self.assertTrue(all(circuit.cregs), lambda reg: len(reg) == 1)  # All of length 1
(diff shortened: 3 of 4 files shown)


ElePT (commented):
It would be nice to have a unit test for this path. The reproducing code that I found looks quite hacky, and this is probably why the issue hadn't surfaced until now. It's a test for the Aer estimator primitive that builds a circuit, adds measurements (via `circuit.measure_all()`), and then calls a part of the aer estimator code that does:

```
  circuit = self._circuits[i].copy()
  circuit.measure_all()
  .... # stuff
  circuit.remove_final_measurements()
```
Calling `measure_all()` twice after a shallow copy doesn't look like a great thing to do, but I guess we shouldn't hit this error if someone does it. I can see if I can isolate this into a more reasonable example.

qiskit/circuit/quantumcircuit.py
```diff
@@ -3889,15 +3889,15 @@ def clear(self) -> None:
     def _create_creg(self, length: int, name: str) -> ClassicalRegister:
         """Creates a creg, checking if ClassicalRegister with same name exists"""
         if name in [creg.name for creg in self.cregs]:
-            new_creg = ClassicalRegister(length, name=f"{name}{ClassicalRegister.instance_count}")
```
kevinhartman, line 3892: It looks to me like there is a behavior change here, in that the old version doesn't increment the instance count, but the new version does. Is that a concern?

kevinhartman: Ah, scratch that, I see now that this was happening before in `new`.

kevinhartman: Or actually (again lol), we're calling `py_new` from `_new_with_prefix` but also incrementing the instance count before we call it. So I _think_ that would double-increment, unless I'm missing something.

raynelfss: 1. The behavior change was actually in #13860 where we just retreived the counter and never incremented it, whereas before it would be incremented.
2. Calling `py_new` only increments the counter if no name is provided. Since we're already parsing the name beforehand, it should only count as one increment.

kevinhartman: Ah okay, I missed the subtlety of point 2. 🙂

crates/circuit/src/bit.rs
```diff
@@ -771,6 +771,25 @@ macro_rules! create_bit_object {
                 })
             }
 
+            /// Allows for the creation of a new register with a provisional prefix and the
```
kevinhartman: Nitpicking, but the word "provisional" here feels a bit unclear to me, since it makes me think the name is temporary.

kevinhartman (commented):
Looks good to me after @ElePT's comment is addressed!

ElePT (commented):
Thanks a lot! The PR looks good to me too. Here is the minimal reproducible code for the issue, I can't add it as a direct suggestion but I think it's enough to test this fix:

```

qc = QuantumCircuit(1)
qc.measure_all()
qc.measure_all()
```

test/python/circuit/test_circuit_operations.py
```diff
@@ -852,6 +852,32 @@ def test_measure_all_repetition(self):
...
+
+        # First call should create a new register
+        circuit.measure_all()
+        self.assertEqual(len(circuit.cregs), 1)  # One creg
+        self.assertEqual(len(circuit.cregs[0]), 1)  # Of length 1
+
+        # Second call should also create a new register
+        circuit.measure_all()
+        self.assertEqual(len(circuit.cregs), 2)  # Now two cregs
+        self.assertTrue(all(circuit.cregs), lambda reg: len(reg) == 1)  # All of length 1
```
kevinhartman: I think you probably want: 

```suggestion
        self.assertTrue(all(len(reg) == 1 for reg in circuit.cregs))  # All of length 1
```

test/python/circuit/test_circuit_operations.py
```diff
@@ -852,6 +852,32 @@ def test_measure_all_repetition(self):
...
+        # Check that no name is the same
+        self.assertEqual(len(set([reg.name for reg in circuit.cregs])), 2)
+
+        # Third call should also create a new register
+        circuit.measure_all()
+        self.assertEqual(len(circuit.cregs), 3)  # Now two cregs
+        self.assertTrue(all(circuit.cregs), lambda reg: len(reg) == 1)  # All of length 1
```
kevinhartman: 👀

kevinhartman (commented):
Looks like you've got a linter error in CI as well 🙂

test/python/circuit/test_circuit_operations.py
raynelfss: Just an extra check, since there's only one qubit, all cregs should be of length 1. Should I also rewrite it to use an iterable instead of a lambda function?

test/python/circuit/test_circuit_operations.py
```diff
@@ -852,6 +852,32 @@ def test_measure_all_repetition(self):
...
+        # Second call should also create a new register
+        circuit.measure_all()
+        self.assertEqual(len(circuit.cregs), 2)  # Now two cregs
+        self.assertTrue(all(len(reg) == 1 for reg in circuit.cregs))  # All of length 1
+        # Check that no name is the same
+        self.assertEqual(len({reg.name for reg in circuit.cregs}), 2)
+
+        # Third call should also create a new register
+        circuit.measure_all()
+        self.assertEqual(len(circuit.cregs), 3)  # Now two cregs
```
ElePT:
```suggestion
        self.assertEqual(len(circuit.cregs), 3)  # Now three cregs
```

ElePT (dismissed):
LGTM, I found a tiny typo in the test comment but it's not important.

ElePT (approved):

PR #13986, opened 2025-03-10
Remove Python dependency from circuit construction

mtreinish:
### Summary

In preparation for providing a C API for building and interacting with circuits this commit removes the required Python component from the circuit construction APIs on CircuitData. After #13860 the only use case of Python in the circuit constructor is to manipulate a ParameterExpression global phase as ParameterExpression is still a Python only construct. This moves to using Python::with_gil() in those places only when it's needed. This gives us a path for building a circuit without the Python interpreter for the C API.

Issue #13276, opened 2024-10-03
Add C FFI for Circuit Construction


mtreinish:
### What should we add?

After we have the full rust-native core data model in #13264 we can then expose a compiled language interface directly to that data model. Having a compiled language interface will let users from other programming languages besides Python leverage Qiskit for quantum computation. But also for Python users having this interface would let Python compiled extensions interface with Qiskit at a lower level directly from the extension. This can be exceedingly powerful for accelerating applications of Qiskit in python, the best example is numpy which [exposes a C API](https://numpy.org/doc/stable/reference/c-api/index.html) which is how Qiskit interfaces with numpy arrays from Rust and has been key for our performance as we've been doing more in Rust.

A C FFI is a natural choice for this as most languages can interface with C, including Python and Rust, and defining a C API via rust is very simple, see: https://doc.rust-lang.org/nomicon/ffi.html#calling-rust-code-from-c.

To start this interface should be very low level and only express a basic interface for building a `QuantumCircuit` with standard gates (or other operations defined solely by Qiskit in rust). Also, I think to start we should consider this API experimental and not commit to it being a fully stable interface as Qiskit has never had an API like this before, but we should also strive to keep the interface public and stable as defined by the [stability policy](https://docs.quantum.ibm.com/open-source/qiskit-sdk-version-strategy). Basically say it'll be stable but not guarantee it at least for the first few releases while we figure out the best patterns here.

mtreinish:
The other tool which might be useful in this effort is [cbindgen](https://crates.io/crates/cbindgen), the docs for this are here: https://github.com/mozilla/cbindgen/blob/master/docs.md

DisantHoridnt:
I am looking into this issue

Cryoris:
Thank you for your interest @DisantHoridnt. Due to the scope of this issue, we’re currently handling it in the maintainer team. However, once we've advanced on this, contributions on concrete, smaller-scoped tasks would be welcome!

DisantHoridnt:
sounds good, I currently am employed as a rust engineer who is responsible for translating c++ code. I might be able to help

mtreinish:
This issue originally was a single epic for the creation of a C API for circuits and the transpiler. But that scope was too large, so I created a second epic in #14245 which is tracking transpilation separately.

mtreinish:
[...]

PR #14006, opened 2025-03-12
Add initial C API for circuit construction

mtreinish:
### Summary

Building off the infrastructure for the sparse observable added in
PR #13445 this commit adds a C FFI for building Quantum Circuits. Right
now there is a function to create an empty circuit with n qubits and m
clbits and then a function to add standard gates to a circuit and then
also get the op counts out of a circuit. This is a start of the
functionality for a C API around interacting with circuits, later PRs
will expand this so we can have a more fully featured C API in the
future.

### Details and comments

Part of #13276

~This PR is based on top of #13986 and will need to rebased after that merges. To see the contents of just this PR you can look at: https://github.com/mtreinish/qiskit-core/compare/no-py-token-constructor-path...mtreinish:qiskit-core:c-api-circuit-rebased~

TODO:

- [x] Add release notes


crates/cext/src/circuit.rs
```diff
@@ -0,0 +1,264 @@
...
+use qiskit_circuit::operations::{Operation, Param, StandardGate};
+use qiskit_circuit::Qubit;
+
+/// @ingroup QkCircuit
+/// Construct a new circuit with the given number of qubits and clbits.
+///
+/// @param num_qubits The number of qubits the circuit contains.
+/// @param num_clbits The number of clbits the circuit contains.
+///
+/// @return A pointer to the created Circuit
```
Cryoris: I'll leave some nits since it'll be the public docs 😛 
```suggestion
/// @return A pointer to the created circuit.
```

crates/cext/src/circuit.rs
```diff
@@ -0,0 +1,264 @@
...
+/// # Safety
+///
+/// Behavior is undefined ``circuit`` is not a valid, non-null pointer to a ``QkCircuit``.
+#[no_mangle]
+#[cfg(feature = "cbinding")]
+pub unsafe extern "C" fn qk_circuit_num_qubits(circuit: *const CircuitData) -> u32 {
+    // SAFETY: Per documentation, the pointer is non-null and aligned.
+    let circuit = unsafe { const_ptr_as_ref(circuit) };
+
+    circuit.num_qubits() as u32
```
Cryoris, line 83: This is correct, but makes me think of another question: if we assume num_qubits to be `u32` everywhere, shouldn't `CircuitData::num_qubits` not return a `u32`?

crates/cext/src/circuit.rs
```diff
@@ -0,0 +1,264 @@
...
+/// Get the number of clbits the circuit contains.
+///
+/// @param circuit A pointer to the circuit.
+///
+/// @return The number of qubits the circuit is defined on.
+///
+/// # Example
+///
+///     QkCircuit *qc = qk_circuit_new(100);
+///     uint32_t num_qubits = qk_circuit_num_clbits(qc);  // num_qubits==100
```
Cryoris:
```suggestion
///     QkCircuit *qc = qk_circuit_new(100, 50);
///     uint32_t num_clbits = qk_circuit_num_clbits(qc);  // num_clbits==50
```

crates/cext/src/circuit.rs
```diff
@@ -0,0 +1,264 @@
...
+/// @param circuit A pointer to the circuit to free.
+///
+/// # Example
+///
+///     QkCircuit *qc = qk_circuit_new(100, 100);
+///     qk_circuit_free(qc);
+///
+/// # Safety
+///
+/// Behavior is undefined if ``circt`` is not either null or a valid pointer to a
```
Cryoris:
```suggestion
/// Behavior is undefined if ``circuit`` is not either null or a valid pointer to a
```

crates/cext/src/circuit.rs
```diff
@@ -0,0 +1,264 @@
...
+        // SAFETY: We have verified the pointer is non-null and aligned, so it should be
+        // readable by Box.
+        unsafe {
+            let _ = Box::from_raw(circuit);
+        }
+    }
+}
+
+/// @ingroup QkCircuit
+/// Append a standard gate to the circuit
```
Cryoris: Nit no. 2
```suggestion
/// Append a standard gate to the circuit.
```

crates/cext/src/circuit.rs
```diff
@@ -0,0 +1,264 @@
...
+///
+/// @param circuit A pointer to the circuit to add the gate to
+/// @param gate The StandardGate to add to the circuit
+/// @param qubits The pointer to the array of ``uint32_t`` qubit indices to add the gate on
+/// @param params The pointer to the array of ``double`` values to use for the gate parameters
```
Cryoris: I'm annoyed at myself for commenting this
```suggestion
/// @param circuit A pointer to the circuit to add the gate to.
/// @param gate The StandardGate to add to the circuit.
/// @param qubits The pointer to the array of ``uint32_t`` qubit indices to add the gate on.
/// @param params The pointer to the array of ``double`` values to use for the gate parameters.
```

crates/cext/src/circuit.rs
```diff
@@ -0,0 +1,264 @@
...
+///
+/// # Example
+///
+///     QkCircuit *qc = qk_circuit_new(100);
+///     qk_circuit_append_standard_gate(qc, HGate, *[0], *[]);
+///
+///
+/// # Safety
```
Cryoris, line 155: Not actually sure the spacing here needs to be exact, feel free to ignore 
```suggestion
/// # Safety
///
```

crates/cext/src/circuit.rs
```diff
@@ -0,0 +1,264 @@
...
+/// The ``qubits`` and ``params`` types are expected to be a non-null pointer to an array of
+/// ``uint32_t`` and ``double`` respectively where the length is matching the expectations for
+/// the standard gate. If the array is insufficently long the behavior of this function is
```
Cryoris: We should have a way of documenting how many parameters and qubits are needed, or to figure this out using functions/lookups. Especially if people (or we ourselves) would like to write automated tests which requires iterating over all gates or so.

What do you think of either e.g. adding a function that takes the stdgate enum? I assume we can't really expose a static array like we do in Rust

crates/cext/src/circuit.rs
```diff
@@ -0,0 +1,264 @@
...
+#[no_mangle]
+#[cfg(feature = "cbinding")]
+pub unsafe extern "C" fn qk_circuit_append_standard_gate(
+    circuit: *mut CircuitData,
+    gate: StandardGate,
+    qubits: *const u32,
+    params: *const f64,
+) -> ExitCode {
+    let circuit = unsafe { mut_ptr_as_ref(circuit) };
+    unsafe {
```
Cryoris, line 177: I'll channel my inner Jake: could you add a safety comment here? 😛

crates/cext/src/circuit.rs
```diff
@@ -0,0 +1,264 @@
...
+///
+/// @param circuit A pointer to the circuit to get the counts for
+///
+/// # Example
+///
+///     QkCircuit *qc = qk_circuit_new(100);
+///     qk_circuit_append_standard_gate(qc, HGate, *[0], *[]);
+///     qk_circuit_count_ops(qc);
+///
+/// # Safety
```
Cryoris:
```suggestion
/// Return a list of string names for instructions in a circuit and their counts.
///
/// @param circuit A pointer to the circuit to get the counts for.
///
/// # Example
///
///     QkCircuit *qc = qk_circuit_new(100);
///     qk_circuit_append_standard_gate(qc, HGate, *[0], *[]);
///     qk_circuit_count_ops(qc);
///
/// # Safety
///
```

crates/cext/src/circuit.rs
```diff
@@ -0,0 +1,264 @@
...
+}
+
+/// @ingroup QkCircuit
+/// Free a circuit op count list.
+///
+/// @param op_counts The returned op count list from ``qk_circuit_count_ops``
+///
+/// # Safety
+///
+/// Behavior is undefined if ``op_counts`` is not the object returned by ``qk_circuit_count_ops``
```
Cryoris:
```suggestion
/// @param op_counts The returned op count list from ``qk_circuit_count_ops``.
///
/// # Safety
///
/// Behavior is undefined if ``op_counts`` is not the object returned by ``qk_circuit_count_ops``.
```

crates/cext/src/circuit.rs
```diff
@@ -0,0 +1,264 @@
...
+            _ => unreachable!(),
+        };
+        circuit.push_standard_gate(gate, params, qargs);
+    }
+    ExitCode::Success
+}
+
+#[repr(C)]
+pub struct OpCount {
+    name: *mut c_char,
```
Cryoris: Why does the `name` and below the `data` pointers need to be `mut`? Could they also be `const`, since it seems we don't change them Rust-side? Not that it would matter once we expose them, but at least that would let the rust compiler verify they aren't changed here 😄

crates/cext/src/circuit.rs
```diff
@@ -0,0 +1,264 @@
...
+///
+///     QkCircuit *qc = qk_circuit_new(100);
+///     qk_circuit_append_standard_gate(qc, HGate, *[0], *[]);
+///     qk_circuit_count_ops(qc);
+///
+/// # Safety
+/// Behavior is undefined ``circuit`` is not a valid, non-null pointer to a ``QkCircuit``.
+#[no_mangle]
+#[cfg(feature = "cbinding")]
+pub unsafe extern "C" fn qk_circuit_count_ops(circuit: *const CircuitData) -> OpCounts {
```
Cryoris, line 388: This is a bit a different flow that we have compared to the `QkObsTerm/CSparseTerm` in the observable, where we ask the C user to allocate the struct and take it by reference to populate it. But after thinking about it I think this makes sense here since `OpCounts` needs to own the data, whereas `CSparseTerm` can just point to the data in the exisiting observable.

crates/cext/src/circuit.rs
```diff
@@ -0,0 +1,264 @@
...
+/// Free a circuit op count list.
+///
+/// @param op_counts The returned op count list from ``qk_circuit_count_ops``
+///
+/// # Safety
+///
+/// Behavior is undefined if ``op_counts`` is not the object returned by ``qk_circuit_count_ops``
+#[no_mangle]
+#[cfg(feature = "cbinding")]
+pub unsafe extern "C" fn qk_circuit_free_count_ops(op_counts: OpCounts) {
```
Cryoris: Maybe 
```suggestion
pub unsafe extern "C" fn qk_opcounts_free(op_counts: OpCounts) {
```
since the scheme is `qk_<object>_free`?

test/c/test_circuit.c
```diff
@@ -0,0 +1,99 @@
...
+
+int test_get_gate_counts_bv_no_measure() {
+    QkCircuit *qc = qk_circuit_new(1000, 1000);
+    double *params = NULL;
+    uint32_t i = 0;
+    uint32_t qubits[1] = {999};
+    qk_circuit_append_standard_gate(qc, QkStandardGate_XGate, qubits, params);
+    for (i = 0; i < 1000; i++) {
+        uint32_t qubits[1] = {i};
+        qk_circuit_append_standard_gate(qc, QkStandardGate_HGate, qubits, params);
```
Cryoris: ``QkStandardGate_HGate`` is a lot of typing, but this what we get since we use the Rust enum, right? We definitely need to check out `cbindgen` via the Rust interface to see whether we can rename this to something shorter (same still holds for the observable bit terms). We can do this in a follow up/parallel work 🙂

crates/cext/src/circuit.rs
```diff
@@ -0,0 +1,264 @@
...
+/// @param params The pointer to the array of ``double`` values to use for the gate parameters
+///
+/// # Example
+///
+///     QkCircuit *qc = qk_circuit_new(100);
+///     qk_circuit_append_standard_gate(qc, HGate, *[0], *[]);
+///
+///
+/// # Safety
+/// The ``qubits`` and ``params`` types are expected to be a non-null pointer to an array of
```
Cryoris: In the tests you're passing ``params`` as ``NULL``, which is not allowed per documentation. Maybe we can just rephrase this stating that `params` and `qubits` can be NULL if the gate has no qubits (GlobalPhaseGate?) or no params?

test/c/test_circuit.c
```diff
@@ -0,0 +1,99 @@
...
+    for (i = 0; i < 1000; i += 2) {
+        uint32_t qubits[2] = {i, 999};
+        qk_circuit_append_standard_gate(qc, QkStandardGate_CXGate, qubits, params);
+    }
+    for (i = 0; i < 999; i++) {
+        uint32_t qubits[1] = {i};
+        qk_circuit_append_standard_gate(qc, QkStandardGate_HGate, qubits, params);
+    }
+    QkOpCounts op_counts = qk_circuit_count_ops(qc);
+    if (op_counts.len != 3) {
```
Cryoris, line 160: op_counts and qc are not freed at failure 🙂

test/c/test_circuit.c
```diff
@@ -0,0 +1,99 @@
...
+    }
+    result = strcmp(op_counts.data[1].name, "cx");
+    if (result != 0) {
+        return result;
+    }
+    if (op_counts.data[1].count != 500) {
+        return EqualityError;
+    }
+    qk_circuit_free(qc);
+    return 0;
```
Cryoris: You could use `Ok` for consistency but it doesn't matter really 
```suggestion
    return Ok;
```

Cryoris (commented):
This looks really cool! I left some comments below 🙂

test/c/test_circuit.c
mtreinish: The test doesn't fail :D

test/c/test_circuit.c
mtreinish: I fixed this sort of in: https://github.com/Qiskit/qiskit/pull/14006/commits/3e62ec6d8d4b68bb1521de0660659c214b7afa83

test/c/test_circuit.c
mtreinish: Fixed in: https://github.com/Qiskit/qiskit/pull/14006/commits/3e62ec6d8d4b68bb1521de0660659c214b7afa83 so it always cleans up after itself now.

crates/cext/src/circuit.rs
mtreinish: Renamed in: https://github.com/Qiskit/qiskit/pull/14006/commits/3e62ec6d8d4b68bb1521de0660659c214b7afa83

crates/cext/src/circuit.rs
mtreinish: Looking at the code I expect it returns a usize because the method is also defined for python (and that was likely the original intent as the only consumer). It was less typing and simpler to just return a usize because it's doing `self.qubits.len()` and returning that to python which is going to convert it to a python int regardless of whether its a u32 or a usize. We can change it outside of this PR, it's a simple fix there but outside the scope of this PR.

test/c/test_circuit.c
Cryoris: > The test doesn't fail :D

famous last words 😂 it's only been a few weeks writing C and I already have trust issues

crates/cext/src/circuit.rs
mtreinish: We probably could expose the static array somehow, but having functions to look this up is a good idea. That's much simpler, and under the covers it will just look it up from the static array.

crates/cext/src/circuit.rs
```diff
@@ -0,0 +1,483 @@
...
+///
+///     QkCircuit *qc = qk_circuit_new(100);
+///     qk_circuit_append_measure(qc, 0, 0);
+///
+/// # Safety
+///
+/// Behavior is undefined ``circuit`` is not a valid, non-null pointer to a ``QkCircuit``.
+#[no_mangle]
+#[cfg(feature = "cbinding")]
+pub unsafe extern "C" fn qk_circuit_append_measure(
```
Cryoris: Should we just call this `qk_circuit_measure` (same for the other directives)? The `append` doesn't really add any information, or what do you think?

crates/cext/src/circuit.rs
```diff
@@ -0,0 +1,483 @@
...
+/// Append a measurement to the circuit
+///
+/// @param circuit A pointer to the circuit to add the gate to
+/// @param qubits The ``uint32_t`` for the qubit to measure
+/// @param clbits The ``uint32_t`` for the clbit to store the measurement outcome in
+///
+/// # Example
+///
+///     QkCircuit *qc = qk_circuit_new(100);
+///     qk_circuit_append_measure(qc, 0, 0);
```
Cryoris: The circuit doesn't have any classical bits, do we need to add one?
```suggestion
///     QkCircuit *qc = qk_circuit_new(100, 1);
///     qk_circuit_append_measure(qc, 0, 0);
```

crates/cext/src/circuit.rs
```diff
@@ -0,0 +1,483 @@
...
+        };
+        circuit.push_standard_gate(gate, params, qargs);
+    }
+    ExitCode::Success
+}
+
+/// @ingroup QkCircuit
+/// Append a measurement to the circuit
+///
+/// @param circuit A pointer to the circuit to add the gate to
```
Cryoris:
```suggestion
/// @param circuit A pointer to the circuit to add the measurement to
```

crates/cext/src/circuit.rs
```diff
@@ -0,0 +1,483 @@
...
+/// # Safety
+///
+/// Behavior is undefined ``circuit`` is not a valid, non-null pointer to a ``QkCircuit``.
+#[no_mangle]
+#[cfg(feature = "cbinding")]
+pub unsafe extern "C" fn qk_circuit_append_reset(
+    circuit: *mut CircuitData,
+    qubit: u32,
+) -> ExitCode {
+    let circuit = unsafe { mut_ptr_as_ref(circuit) };
```
Cryoris, line 307: I'm surprised cargo doesn't complain (maybe because of unsafe?) but the `circuit` should be declared `mut` if we push onto it, no? The same holds for the other methods 🙂

crates/cext/src/circuit.rs
```diff
@@ -0,0 +1,483 @@
...
+/// @param num_qubits The number of qubits wide the barrier is
+/// @param qubits The pointer to the array of ``uint32_t`` qubit indices to add the barrier on.
+///
+/// # Example
+///
+///     QkCircuit *qc = qk_circuit_new(100);
+///     qk_circuit_append_reset(qc, 0);
+///
+///
+/// # Safety
```
Cryoris:
```suggestion
///     QkCircuit *qc = qk_circuit_new(100);
///     uint32_t qubits[5] = {0, 1, 2, 3, 4};
///     qk_circuit_append_barrier(qc, 5, qubits);
///
/// # Safety
/// 
```

Cryoris:
```suggestion
///     QkCircuit *qc = qk_circuit_new(100);
///     uint32_t qubits[5] = {0, 1, 2, 3, 4};
///     qk_circuit_append_barrier(qc, 5, qubits);
///
/// # Safety
/// 
```

crates/cext/src/circuit.rs
```diff
@@ -0,0 +1,483 @@
...
+/// @ingroup QKCircuit
+/// Return the number of instructions in the circuit
+///
+/// @param m circuit A pointer to the circuit to get the counts for.
+///
+///   /// # Example
+///
+///     QkCircuit *qc = qk_circuit_new(100);
+///     qk_circuit_append_standard_gate(qc, HGate, *[0], *[]);
+///     qk_circuit_num_instructions(qc); // 1
```
Cryoris:
```suggestion
/// @ingroup QkCircuit
/// Return the number of instructions in the circuit
///
/// @param circuit A pointer to the circuit to get the counts for.
///
/// # Example
///
///     QkCircuit *qc = qk_circuit_new(100);
///     qk_circuit_append_standard_gate(qc, QkGate_H, *[0], *[]);
///     qk_circuit_num_instructions(qc); // 1
```

crates/cext/src/circuit.rs
```diff
@@ -0,0 +1,483 @@
...
+/// @ingroup QKCircuit
+/// Return the number of instructions in the circuit
+///
+/// @param m circuit A pointer to the circuit to get the counts for.
+///
+///   /// # Example
+///
+///     QkCircuit *qc = qk_circuit_new(100);
+///     qk_circuit_append_standard_gate(qc, HGate, *[0], *[]);
+///     qk_circuit_get_instruction(qc, 0); //
```
Cryoris:
```suggestion
/// @ingroup QkCircuit
/// Return the number of instructions in the circuit
///
/// @param circuit A pointer to the circuit to get the counts for.
///
/// # Example
///
///     QkCircuit *qc = qk_circuit_new(100);
///     qk_circuit_append_standard_gate(qc, QkGate_H, *[0], *[]);
///     QkCircuitInstruction inst = qk_circuit_get_instruction(qc, 0);
```

crates/cext/src/circuit.rs
```diff
@@ -0,0 +1,483 @@
...
+/// @ingroup QkCircuit
+/// Free a circuit instruction object
+///
+/// @param inst The instruction to free
+///
+/// # Safety
+/// Behavior is undefined if ``inst`` is not an object returned by ``qk_circuit_get_instruction``.
+#[no_mangle]
+#[cfg(feature = "cbinding")]
+pub unsafe extern "C" fn qk_free_circuit_instruction(inst: CInstruction) {
```
Cryoris: This should be `qk_circuit_instruction_free` for consistency -- and should it take a pointer to a `CInstruction` otherwise the struct would get passed by value? Same question actually below for `OpCounts` 🙂

crates/cext/src/circuit.rs
mtreinish: I can drop append if you prefer. I named it this way because of the gate function I used append.

crates/cext/src/circuit.rs
mtreinish: It does take it by value, but that's because we're freeing the pointers that are contained in the struct not the struct itself. Each `CInstruction` has 4 pointers to a heap allocated arrays that we need to free. If we just drop the instruction those would be leaked. We can take a pointer to `CInstruction` but it functionally doesn't matter in this case because the struct itself isn't what we're freeing.

crates/cext/src/circuit.rs
mtreinish: Added in: https://github.com/Qiskit/qiskit/pull/14006/commits/c992f7ad0ee41c04e06d28258329c44623bbe0cf

crates/cext/src/circuit.rs
mtreinish: I added functions to look this up in: https://github.com/Qiskit/qiskit/pull/14006/commits/06665c05c3c8a14e9ac6b0fb20fd70ae31714477

crates/cext/src/circuit.rs
mtreinish: I updated the docs for this in: https://github.com/Qiskit/qiskit/pull/14006/commits/659c7196c236e84ce5c77d1fa7a1ab6685e42b00

test/c/test_circuit.c
mtreinish: You fixed this so it's now `H` instead of `HGate`, and I renamed `QkStandardGate` to `QkGate` in https://github.com/Qiskit/qiskit/pull/14006/commits/66a989c819a7a86fc8d1362ec6fc422b9586fa53. So now it's `QkGate_H`.

crates/cext/src/circuit.rs
```diff
@@ -0,0 +1,544 @@
...
+///
+/// # Example
+///
+///     QkCircuit *qc = qk_circuit_new(100);
+///     qk_circuit_append_standard_gate(qc, QkGate_H, *[0], *[]);
+///     QkCircuitInstruction inst = qk_circuit_get_instruction(qc, 0);
+///
+/// # Safety
+///
+/// Behavior is undefined ``circuit`` is not a valid, non-null pointer to a ``QkCircuit``.
```
raynelfss: I realized this far into reading that this docstring is missing an "if"? There's also many instances of that so I'll just give you this one first 😄 
```suggestion
/// Behavior is undefined if ``circuit`` is not a valid, non-null pointer to a ``QkCircuit``.
```

releasenotes/notes/c-api-circuit-3ecb43d396425bed.yaml
```diff
@@ -0,0 +1,5 @@
+---
+features_circuits:
+  - |
+    The C API for Qiskit has been extended with support for building and
+    interacting with quantum circuits.
```
raynelfss, line 5: Maybe we should mention that the support of this API is limited ?
```suggestion
    The C API for Qiskit has been extended with limited support for building and
    interacting with quantum circuits.
```

test/c/test_circuit.c
```diff
@@ -0,0 +1,554 @@
...
+                return EqualityError;
+            }
+        } else {
+            if (qk_gate_num_qubits(i) != 4) {
+                return EqualityError;
+            }
+        }
+    }
+    return Ok;
+}
```
raynelfss: Would it be possible for us to iterate through the values of the Enum itself? AFAIK we could do something like:
```c
for (QkGate gate = X; i < RC3X; i++) {
    ...
}
```
But then again there's no perfect way of knowing the end value of an enum unless added explicitly. But in case we were to hypothetically add a new standard gate, we wouldn't necessarily have to update this test manually.

test/c/test_circuit.c
```diff
@@ -0,0 +1,554 @@
...
+        } else {
+            if (qk_gate_num_params(i) != 3) {
+                return EqualityError;
+            }
+        }
+    }
+    return Ok;
+}
+
+int test_get_gate_counts_bv_no_measure() {
```
raynelfss: Someone really likes Bernstein-Vazirani :)

raynelfss (commented):
I gave a quick review to what you've worked on so far. I feel like it makes sense to me even if the API is still very limited. I would suggest adding any additional support for other operations within the circuit in subsequent PRs.

Most of the observations here are just docstring comments :)

releasenotes/notes/c-api-circuit-3ecb43d396425bed.yaml
mtreinish: I'm not actually sure what limited would mean here. The API lets you build and interact with a quantum circuit. It's not the same as the python API but there will be continued incremental improvements to the API over time. We're also 3 months out from the final 2.1.0 release so it's hard to say how limited the API will be when we finish.

test/c/test_circuit.c
mtreinish: I can change the integers to use the gate enum variants, but in practice I think trying to make this robust against future gates won't really be feasible. It would depend on semantics being applied to the enum ordering which isn't a guarantee, we have a convention now. But if we add gates that definitely won't hold because we're unlikely to ever change the existing order of gates in the future because it would require very large refactoring, more than just these tests.

crates/cext/src/circuit.rs
mtreinish: Done in: https://github.com/Qiskit/qiskit/pull/14006/commits/92a4a8f5d4cf08600b9d875534948f41fd693ce5

crates/cext/src/circuit.rs
Cryoris: But don't we also need to free the struct if it was allocated by `qk_circuit_get_instruction`? 🤔

mtreinish: It's a directly returned stack allocated struct so we don't have to worry about freeing it, the memory will be reclaimed automatically like any local variable (unless the caller puts it on the heap but then it's their responsibility).

Cryoris: Ah right because we don't Box::free anything we don't have to consume it again -- like the 1:1 mapping of `malloc` and `free` in C

test/c/CMakeLists.txt
```diff
@@ -19,12 +19,13 @@ file (
 # test file.
 create_test_sourcelist (source_files test_driver.c ${discovered_tests})
 
+add_library(common common.c)
```
Cryoris: So I'm not sure if this fixes it on windows (locally it still works on my mac) but it seems like windows is not finding the library (again), maybe this fixes it?
```suggestion
add_library(common common.c)
install(TARGETS common RUNTIME DESTINATION ${CMAKE_BINARY_DIR})
```
which should install the dll into the path of the executable (see also [CMake docs](https://cmake.org/cmake/help/book/mastering-cmake/chapter/Install.html)). We could also consider putting this inside a `if (MSVC)` block, though it didn't seem to affect running tests for me.

test/c/CMakeLists.txt
```diff
@@ -19,12 +19,14 @@ file (
...
+install(TARGETS common RUNTIME DESTINATION ${CMAKE_BINARY_DIR})
```
Cryoris: With a static lib we shouldn't need to worry about placing the DLLs in the right location on windows
```suggestion
add_library(common STATIC common.c)
```

crates/cext/src/circuit.rs
mtreinish: Actually rust emits a warning saying the variable does not to be `mut` if you do `let mut circuit` here. It's likely because the return from `&mut` so it's explicitly mutable already.

crates/cext/src/circuit.rs
mtreinish: `data` and the other array types need to be mut because of the free implementation works. To coerce the pointer back into a `Box` that we then free we it's dropped we need the pointer to be mutable as the APIs all work with `* mut`. But for the string types I can make them `const` pointers and did in https://github.com/Qiskit/qiskit/pull/14006/commits/50b54087b37b42d8c958b775621b18dd4c81a168

crates/cext/src/circuit.rs
```diff
@@ -0,0 +1,544 @@
...
+    let data = output.as_mut_ptr();
+    let len = output.len();
+    std::mem::forget(output);
+    OpCounts { data, len }
+}
+
+/// @ingroup QkCircuit
+/// Return the number of instructions in the circuit
+///
+/// @param circuit A pointer to the circuit to get the counts for.
```
raynelfss:
```suggestion
/// @param circuit A pointer to the circuit to get the number of operations of.
```

crates/cext/src/circuit.rs
```diff
@@ -0,0 +1,544 @@
...
+    clbits: *mut u32,
+    num_params: u32,
+    params: *mut f64,
+}
+
+/// @ingroup QkCircuit
+/// Return the number of instructions in the circuit
+///
+/// @param circuit A pointer to the circuit to get the counts for.
+///
```
raynelfss: This docstring doesn't seem to be correct. If I understood correctly, this function access an instruction by index.

raynelfss: Also:
```suggestion
/// @ingroup QkCircuit
/// Return the number of instructions in the circuit.
///
/// @param circuit A pointer to the circuit to be accessed.
```

crates/cext/src/circuit.rs
```diff
@@ -0,0 +1,544 @@
...
+        PackedOperation::from_standard_instruction(StandardInstruction::Measure),
+        &[],
+        &[Qubit(qubit)],
+        &[Clbit(clbit)],
+    );
+    ExitCode::Success
+}
+
+/// @ingroup QkCircuit
+/// Append a reset to the circuit
```
raynelfss, line 289: Might be better to specify that we're adding a reset gate?
```suggestion
/// Append a reset gate to the circuit
```

test/c/common.c
```diff
@@ -0,0 +1,44 @@
...
+    char *msg;
+    if (result == Ok) {
+        did_fail = 0;
+        msg = "Ok";
+    } else if (result == EqualityError) {
+        msg = "FAILED with an EqualityError";
+    } else {
+        msg = "FAILED with unknown error";
+    }
+    fprintf(stderr, "--- %-30s: %s\n", name, msg);
```
raynelfss, line 40: Would it make sense to increase the padding to a little higher than 30 since some of the methods have longer names?

test/c/test_circuit.c
```diff
@@ -0,0 +1,551 @@
...
+ * Test the zero constructor.
+ */
+int test_empty(void) {
+    QkCircuit *qc = qk_circuit_new(0, 0);
+    uint32_t num_qubits = qk_circuit_num_qubits(qc);
+    uint32_t num_clbits = qk_circuit_num_clbits(qc);
+    size_t num_instructions = qk_circuit_num_instructions(qc);
+    qk_circuit_free(qc);
+
+    return (num_qubits != 0 || num_clbits != 0 || num_instructions != 0) ? EqualityError : Ok;
```
raynelfss: I think that separating these might be a bit better. Although I do understand we don't necessarily have tracebacks here. We could add a `print` statement in the future that says which of these might have failed.

test/c/test_circuit.c
```diff
@@ -0,0 +1,551 @@
...
+}
+
+int test_no_gate_1000_bits(void) {
+    QkCircuit *qc = qk_circuit_new(1000, 1000);
+    uint32_t num_qubits = qk_circuit_num_qubits(qc);
+    uint32_t num_clbits = qk_circuit_num_clbits(qc);
+    size_t num_instructions = qk_circuit_num_instructions(qc);
+    qk_circuit_free(qc);
```
raynelfss, line 52: See https://github.com/Qiskit/qiskit/pull/14006#discussion_r2047487020

raynelfss (commented):
I think this looks almost ready to merge. I have a couple docstring comments and a question about tests :)

crates/cext/src/circuit.rs
mtreinish: A reset is not a gate since it's not a unitary operation.

crates/cext/src/circuit.rs
mtreinish: You're correct this docstring is incorrect it's a copy and paste of a different function and I forgot to update it.

test/c/test_circuit.c
mtreinish: Updated in: https://github.com/Qiskit/qiskit/pull/14006/commits/296190d87fc99f097f33a057a4f1a02307654054

test/c/test_circuit.c
mtreinish: https://github.com/Qiskit/qiskit/pull/14006/commits/296190d87fc99f097f33a057a4f1a02307654054

crates/cext/src/circuit.rs
mtreinish: Fixed in: https://github.com/Qiskit/qiskit/pull/14006/commits/4f0e9660cb561bf765aa289b47b858eb2c1aea4a

crates/cext/src/circuit.rs
mtreinish: Fixed in: https://github.com/Qiskit/qiskit/pull/14006/commits/4f0e9660cb561bf765aa289b47b858eb2c1aea4a

test/c/common.c
mtreinish: It doesn't bother me too much that the Ok's are not aligned for the longer function names. Like: https://github.com/Qiskit/qiskit/actions/runs/14495258916/job/40661239698#step:4:370 looks fine to me

We can expand it, but I'd rather do that in a separate PR. I didn't really edit the code in this function I only split it out to support multiple test files.

raynelfss: That's okay then, it doesn't bother me too much either, but I wanted to bring it up here :)

crates/cext/src/circuit.rs
jakelishman: Adding to the bikeshedding, but generally I'm with Julien - the `append` is just noise (and we don't have it in Python).  Similarly, I'd suggest `qk_circuit_append_standard_gate` just becomes `qk_circuit_gate`, since `StandardGate` should be _by far_ the most common thing to add. Once we support custom instructions, we can call that something like `qk_circuit_custom`.

crates/cext/src/circuit.rs
```diff
@@ -0,0 +1,566 @@
...
+/// @ingroup QkCircuit
+/// Free a circuit instruction object
+///
+/// @param inst The instruction to free
+///
+/// # Safety
+/// Behavior is undefined if ``inst`` is not an object returned by ``qk_circuit_get_instruction``.
+#[no_mangle]
+#[cfg(feature = "cbinding")]
+pub unsafe extern "C" fn qk_free_circuit_instruction(inst: CInstruction) {
```
jakelishman: The name here should be `qk_circuit_instruction_free` for consistency, I think?

jakelishman (commented):
I didn't actually review the implementations, I just joined in with the bikeshedding.

crates/cext/src/circuit.rs
mtreinish: Fixed in: https://github.com/Qiskit/qiskit/pull/14006/commits/1240467ab1fba3a9ae3c57a514d9f272c235fe56

crates/cext/src/circuit.rs
mtreinish: Renamed in: https://github.com/Qiskit/qiskit/pull/14006/commits/1240467ab1fba3a9ae3c57a514d9f272c235fe56

crates/cext/src/pointers.rs
```diff
@@ -0,0 +1,48 @@
+// This code is part of Qiskit.
+//
+// (C) Copyright IBM 2024
```
raynelfss, line 3:
```suggestion
// (C) Copyright IBM 2025
```

test/c/common.c
```diff
@@ -0,0 +1,44 @@
+// This code is part of Qiskit.
+//
+// (C) Copyright IBM 2024.
```
raynelfss, line 3:
```suggestion
// (C) Copyright IBM 2025.
```

raynelfss (approved):
LGTM! (except for the minor dates detail)

crates/cext/src/pointers.rs
mtreinish: The contents of this file were just moved from the original unchanged and they had a 2024 copyright header so I just preserved that.

test/c/test_circuit.c
```diff
@@ -22,27 +22,52 @@
...
+int test_no_gate_1000_bits(void) {
     QkCircuit *qc = qk_circuit_new(1000, 1000);
     uint32_t num_qubits = qk_circuit_num_qubits(qc);
     uint32_t num_clbits = qk_circuit_num_clbits(qc);
     size_t num_instructions = qk_circuit_num_instructions(qc);
     qk_circuit_free(qc);
 
-    return (num_qubits != 1000 || num_clbits != 1000 || num_instructions != 0) ? EqualityError : Ok;
+    if (num_qubits != 1000) {
+        printf("The number of qubits %d is not 1000", num_qubits);
```
Cryoris, line 55: It would be nice to standardize this in the future. I'm wondering if this is printed somehow in order with CTest's output or if it is shown at another point and the origin is hard to pin down.

test/c/CMakeLists.txt
```diff
@@ -32,12 +31,7 @@ target_link_libraries (test_driver ${qiskit} common)
 if (MSVC)
     # Store the directory where the qiskit_cext.dll is located.
     get_filename_component(qiskit_dll_dir ${qiskit} DIRECTORY)
-
-    # Get the Python runtime library, which has been stored into a config file during the 
-    # crate build process. This is to ensure we link against the same library that PyO3 is using.
-    file(READ "${CMAKE_SOURCE_DIR}/target/pyo3_python.config" pyo3_config)
```
Cryoris, line 36: Does the `pyo3_python.config` file still have value if we don't use it here? We could also remove it (fine to do as follow up) if we don't need it anymore

Cryoris (approved):
Two small comments (that can be skipped) -- LGTM!

test/c/test_circuit.c
mtreinish: IIRC, I used print debugging for developing the tests in the first place and ctest put the output somewhere I could find. But if we need to make improvements here we can definitely do it in a follow up.

test/c/CMakeLists.txt
mtreinish: Probably not, lets look to drop it in a follow-up

mtreinish: I removed this in: https://github.com/Qiskit/qiskit/pull/14239

PR #14009, opened 2025-03-12
[Stretch] Don't reject non-const duration in constructors.

kevinhartman:
### Summary
Originally, I thought it'd make the most sense to reject construction of operations that'd have a duration type but also be non-const expressions. But, a user can currently create a `Var` with a duration type, so this seems like an incomplete restriction.

### Details and comments

Consider this PR more of a suggestion—perhaps it makes more sense to also block construction of `Var` when the type is `Duration`. Either way, I think we should decide on this before 2.0.0.


jakelishman (approved):
I'm fine with this; I don't think there's anything _necessarily_ inherent to the `duration` type that requires it to be a compile-time constant forever.  We can reject it in circuit-level type-checking if we want to.

PR #14013, opened 2025-03-13
Use `cbindgen` via Rust API

Cryoris:
### Summary

We currently generate the header using `cbindgen` as command line tool. This works, but 
* doesn't expose all features `cbindgen` has (ok almost all)
* requires users to additionally install the command line tool (and run it)
* emits a lot of warnings we (currently?) don't know how to silence

This PR moves to using the Rust API of `cbindgen` and builds the header as part of the `qiskit-cext` compilation process. This fixes all the above points. The one down side is that the `cbindgen.toml` was a bit nicer to read IMO, but that's a small price to pay.

It would be nice to backport this to `stable/2.0` for a smoother build process for users.

### Details

* The `Makefile` is slightly simpler now since we don't need to separately call the command line tool to generate the header file.
* The resulting header file is exactly the same as before, up to an additional include guard I added in this PR


crates/cext/build_header.rs
```diff
@@ -0,0 +1,87 @@
...
+    config.export = export;
+
+    // Define Rust dependencies to parse.
+    let mut parse = cbindgen::ParseConfig::default();
+    parse.parse_deps = true;
+    parse.include = Some(vec!["qiskit-accelerate".to_string()]);
+    config.parse = parse;
+
+    // Ensure the include directory exists, and then set the full header path.
+    let mut path = "../../dist/c/include".to_string();
```
mtreinish: I'm wondering if this is always what we want. Like I get that's what we had in the makefile, but it seems overly prescriptive on the use case. Like when I'm working on the C API I tend to have a cwd of `crates/cext` and just call `cargo build` and `cbindgen` directly.

I'm thinking that we might want the build file to only save it to the cwd and rely on the makefile to move it to the final destination.

mtreinish (commented):
I like the concept of this integrating the header generation into the build process, it makes a lot of sense, especially when building the standalone c dynamic library.

My concern is for the case where people are trying to get the header file without building the library. For example, if a user of the precompiled wheels is trying to generate the header and build a custom python extension the normal workflow is to run cbindgen manually or run `make cheader`. So I'm wondering what the workflow is for that use case here because you removed the cbindgen.toml so the standalone header generation doesn't work anymore does it?

crates/cext/build_header.rs
Cryoris: Yeah that sounds more stable 👍🏼

Cryoris: Actually, I'll revise: this path seems to be relative to the `cext` crate -- whether running 
```
cd $QISKIT_ROOT
cargo build -p qiskit-cext
```
or 
```
cd crates/cext
cargo build
```
The header file is correctly placed in `$QISKIT_ROOT/dist/c/include` 🙂 In this case this is more elegant than having an additional Makefile move the file, since we just need to cargo, right?

Cryoris:
> My concern is for the case where people are trying to get the header file without building the library. 

Yes that's a good point. It seems we can actually have both (until we would need features only exposed in the Rust API): we can have the settings in a `cbindgen.toml` and allow using the command line tool, but also pull that config file in the crate build process. I'll push a commit with this in a bit 🙂

crates/cext/build_header.rs
mtreinish: I was actually arguing that this was not desireable behavior. If you do:
```
cd crates/cext
cargo build
```

(or now `cargo rustc --crate-type cdylib -p qiskit-cext`)

I would expect the header to be in the CWD rather than `../../dist/c/include`.

Cryoris: Updated in 8a81b06! I updated the Makefile to enable using `cbindgen` from the command line as well. `make c` will now first try and compile the Rust crate, which might build the header, but if it doesn't (because Rust only builds lazily) `cbindgen` is invoked directly to build it. There might be nicer ways of doing this but I didn't find any 😄

jakelishman:
I thought a point of this PR was that the build-script API is more featureful than the stand-alone command-line build.  If so, then surely we must be removing the stand-alone build as a supported option, or we're setting ourselves up for pain when the two build methods start producing different results.

I don't see a huge problem with "to build the header file you must build the library", especially at this early stage, but as a minimal workaround at least: `cargo check` necessarily executes build scripts, but doesn't fully compile the library.

crates/cext/build_header.rs
```diff
@@ -0,0 +1,27 @@
...
+    let config = cbindgen::Config::from_file("cbindgen.toml").unwrap();
+
+    // Build the header.
+    cbindgen::Builder::new()
+        .with_crate(".")
+        .with_config(config)
+        .generate()
+        .expect("Unable to generate C bindings.")
+        .write_to_file("qiskit.h");
+}
```
jakelishman: Personally I would say that it's not good behaviour to have a build script write a dirty built file into the source repo as a matter of habit.  Artifacts built by `cargo` should probably go in Cargo's output directory `/target/.../`, and the `Makefile` version can "install" that into the distribution location.

Imo, only a manual call to the command-line tool should place dirty files in the CWD (if we even keep the command-line tool as a stand-alone thing).

crates/cext/cbindgen.toml
```diff
@@ -1,5 +1,8 @@
+language = "C"
 include_version = true
+include_guard = "QISKIT_H_INCLUDED"
```
jakelishman: Very minor: The conventional name for include-guard variables is just the name of the file, e.g. `QISKIT_H`, not `QISKIT_H_INCLUDED`.

Makefile
```diff
@@ -120,22 +119,29 @@ fix_cformat:
...
+
+.PHONY: c cheader
 cheader: $(C_QISKIT_H)
-clib: $(C_LIBQISKIT)
-c: clib cheader
+c: $(C_LIBQISKIT) $(C_QISKIT_H)
 
 # Use ctest to run C API tests
-ctest: $(C_QISKIT_H)
+ctest: $(C_LIB_CARGO_PATH)
```
jakelishman: If `ctest` does depend on `C_QISKIT_H` being installed into the `dist` directory, it should depend on that as well as the library being in the `dist` directory.

Makefile
```diff
@@ -120,22 +119,29 @@ fix_cformat:
...
-$(C_LIBQISKIT): $(C_DIR_LIB) $(C_LIB_CARGO_PATH)
+$(C_DIR_INCLUDE):
+	mkdir -p $(C_DIR_INCLUDE)
+
+$(C_LIBQISKIT): $(C_DIR_LIB)  $(C_LIB_CARGO_PATH)
 	cp $(C_LIB_CARGO_PATH) $(C_DIR_LIB)/$(subst _cext,,$(C_LIB_CARGO_FILENAME))
 
-.PHONY: cheader clib c
+$(C_QISKIT_H): $(C_LIB_CARGO_PATH)
+	@if [ ! -f "$(C_QISKIT_H)" ]; then cbindgen --crate qiskit-cext --output $(C_DIR_INCLUDE)/qiskit.h; fi
```
jakelishman: There's quite a lot of duplication going on here, and recipes building more than they're specifically asked.  Recipes with specific filenames should build that filename _only_.  Copying the file from one place to another is a separate recipe.

I don't think we should have `cbindgen` as a command-line tool anywhere in the Makefile anymore, since we're now doing this by build script?

crates/cext/build_header.rs
Cryoris: Updated to write `qiskit.h` into `target/qiskit.h` as per out offline discussion

crates/cext/cbindgen.toml
Cryoris: Changed to `QISKIT_H` 👍🏻

Makefile
Cryoris: The header handling is not analogous to how you did the lib handling before: there's separate targets for creating the include dir and copying the header into the include dir. Also `cbindgen` is now no longer needed as command line tool since I added `target/qiskit.h` as tracked file in the build script (to ensure it is rebuilt if it was removed/touched). I tested locally that deleting the headers manually retriggers the build process.

crates/cext/build_header.rs
```diff
@@ -0,0 +1,38 @@
...
+fn main() {
+    // Trigger this script if the header was changed/removed.
+    #![allow(clippy::print_stdout)]
+    println!("cargo:rerun-if-changed=../../target/qiskit.h");
+
+    // Pull the config from the cbindgen.toml file.
+    let config = cbindgen::Config::from_file("cbindgen.toml").unwrap();
+
+    // Ensure target path exists and then set the full filename of qiskit.h.
+    let mut path = ::std::path::PathBuf::from_str("../../target/").unwrap();
```
mtreinish: I always worry when I see a relative path unqualified like this, since it's assuming the cwd is `crates/cext/build_header` but I checked and there is not official way to get the target dir from cargo. The `file!()` macro might be useful here, but I'm not sure what that returns, if it's the absolute path to `build_header.rs` on the local filesystem that might be a better option, but if it just returns `build_header.rs` that's obviously not useful.

mtreinish (commented):
This is looking good to me. Thanks for the updates on this, I just have one small inline question/comment about the `build_header.rs` but otherwise I think this ready to merge.

crates/cext/build_header.rs
Cryoris: An absolute path would indeed be nicer, but I didn't really find a way to get it. `file!()` returned just `crates/cext/build_header.rs` as path, not the absolute path to reach the `target` dir... I tested building both from Qiskit root and `cargo/cext` and it seems to work, if that is consolidation enough 😬

mtreinish: Well then lets go with it, I couldn't come up with a better solution either and there is a tracking issue in cargo already: https://github.com/rust-lang/cargo/issues/9661

So I'm fine with doing this and if/when we come up with an alternative we can adjust it. But, do you know if this works on windows? Maybe we wait until #14030 merges just to make sure.

Cryoris: Done, let's see if CI passes

mtreinish (approved):

PR #14014, opened 2025-03-13
Fix dagger symbol in `RGate` API documentation

AlbertJP:
### Summary

The dagger symbol in the equation at `inverse()` is displayed as a "dagger" text with a superscript "d", because of a missing backslash.

### Details and comments

Screenshot showing the problem:
grafik

### DIFF
--- a/qiskit/circuit/library/standard_gates/r.py
+++ b/qiskit/circuit/library/standard_gates/r.py
@@ -79,7 +79,7 @@ class RGate(Gate):
         self.definition = qc
 
     def inverse(self, annotated: bool = False):
-        """Invert this gate as: :math:`R(θ, φ)^{\dagger} = R(-θ, φ)`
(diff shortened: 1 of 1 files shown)


qiskit/circuit/library/standard_gates/r.py
```diff
@@ -79,7 +79,7 @@ def _define(self):
         self.definition = qc
 
     def inverse(self, annotated: bool = False):
-        """Invert this gate as: :math:`r(θ, φ)^dagger = r(-θ, φ)`
+        """Invert this gate as: :math:`r(θ, φ)^{\dagger} = r(-θ, φ)`
```
Cryoris: It's also referred to as uppercase `R` in the docs, could you update that as well? 🙂 
```suggestion
        """Invert this gate as: :math:`R(θ, φ)^{\dagger} = R(-θ, φ)`
```

Cryoris (dismissed):
Thanks!

Cryoris (approved):

Issue #13983, opened 2025-03-10
Fix PGO on MacOS for 2.0.0 final


mtreinish:
### What should we add?

In the 2.0.0rc1 release jobs building on macOS with PGO was failing with linker errors. This seems to be caused by something related to #13934 although the exact cause is unknown still. For 2.0.0rc1 to unblock the release we disabled pgo on macos temporarily, but before we tag the final release we should understand the root cause and try to fix it so we can build with PGO on macOS for the final release.

mtreinish:
We also had failures on the aarch64 linux builds which look to be hitting https://github.com/llvm/llvm-project/issues/57501

@ElePT found similar issues reported on rust: https://github.com/rust-lang/rust/issues/117220 and https://github.com/rust-lang/rust/issues/115344 which should hopefully be fixed in a future rust release. We're using rust 1.79 for aarch64 because the llvm tools from rustup didn't work on the latest rust releases but if upgrading (probably to rust 1.86 when it's released) will fix this we should find a path to do that.

mtreinish:
@Cryoris was able to confirm that Rust 1.85 seems to fix the llvm failure we hit on aarch64 linux builds. He was able to reproduce the failure on macOS with 1.79 and then using 1.85 fixed it. So if we can get a working llvm tools build on rust 1.85 that should fix the issue there with the combined LTO and PGO builds.

mtreinish:
Fixed by: https://github.com/Qiskit/qiskit/pull/13992

PR #14015, opened 2025-03-13
Try building aarch64 wheels with Rust 1.83

mtreinish:
### Summary

This commit attempts to build the aarch64 linux wheels with rust 1.83. When we made aarch64 linux a tier 1 supported platform #13737 there were packaging issues with the official distrubtions for Rust 1.85 and 1.84 on the platform which were preventing PGO builds from succeeding. At the time we we determined that our MSRV of 1.79 was able to build successfully. However, during the 2.0.0rc1 pre-release the wheel build job failed because of reported LLVM bug when combining LTO and PGO which we do for the release jobs. Local debugging determined these errors should be been fixed if built with Rust 1.85. Since we know we're not able to build with Rust 1.85 or 1.84 this commit tries building with 1.83 to see if this can work. If this does not work, follow up commits will walk backwards to see if 1.82, 1.81, or 1.80 can build successfully with LTO+PGO.

### Details and comments

Related to: #13983

### DIFF
--- a/.github/workflows/wheels-build.yml
+++ b/.github/workflows/wheels-build.yml
@@ -102,7 +102,7 @@ jobs:
           mkdir -p "$PGO_WORK_DIR"
           if [[ `uname -m` == "aarch64" ]] ; then
               INSTALL_RUST_PATH=tools/install_rust_msrv.sh
-              RUST_TOOLCHAIN=1.79
+              RUST_TOOLCHAIN=1.83
           else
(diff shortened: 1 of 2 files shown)


mtreinish:
Ok it looks like this will not work, we'll just have to disable PGO on arm64 for now. Rust >1.81 has non-portable symbols in the llvm libs used for pgo, and rust <=1.80 is hitting the LLVM bug discussed in #13983.

PR #14020, opened 2025-03-14
`SingleQubitOperation` trait and speedup to `Optimize1qGatesDecomposition` pass 

alexanderivrii:
### Summary

This PR implements a new trait `SingleQubitOperation` for `StandardGate`, `PyGate`, `UnitaryGate` and `PackedOperation`. The trait currently defines a single method `fn get_mat_static_1q(&self, params: &[Param]) -> Option<[[Complex64; 2]; 2]>` which returns a matrix for single-qubit gates (and `None` otherwise). This is analogous to the `matrix` method in `Operation` except that it's tuned for single-qubit matrices, allowing to represent these as `[[Complex64; 2]; 2]` instead of `Array2<Complex64>`.

This is used to speed up matrix multiplication inside `optimize_1q_gates_decomposition`, leading to an about 10% performance improvement for the `Optimize1qGatesDecomposition` transpiler pass.

Update: after a discussion, it seemed best to implement the new method(s) as part of the ``Operation`` trait itself, but only as a _short-term measure_.


Cryoris (commented):
I'm no Rust expert, so it's very possible I'm wrong, but I'm not sure this is how `Trait` should be used here. I would expect something like a `SingleQubitOperation` would only be implemented for single-qubit gates, but here it is implemented for _all_ standard gates. Typically I expect a trait to allow me to write something like
```rust
fn my_special_function_for_1q(gate: impl SingleQubitOperation) {
  // now I am certain that ``gate`` is indeed a 1q operation
}
```
but this is not the case, since any `StandardGate` variant fulfills this trait (and just returns a `None` matrix).

If I understand correctly, the thing you want to enable is for `matrix` to return a slice instead of an owned `Array`, right? If that's so, would another solution be to add something like a `matrix_view` or `matrix_as_slice` to the `Operation` interface instead?

alexanderivrii:
> If I understand correctly, the thing you want to enable is for matrix to return a slice instead of an owned Array, right? If that's so, would another solution be to add something like a matrix_view or matrix_as_slice to the Operation interface instead?

Thanks @Cryoris. As we have discussed offline, there are indeed several different ways to implement the required functionality, with some decisions to be made about (1) should this be a part of the Operation trait or should this be a part of a different  trait, (2) should this new trait be called ``SingleQubitOperation`` or something like ``MatrixRepresentation``, and (3) do we want to have a method that returns ``Some([[Complex64; 2]; 2])`` for a single-qubit gate and ``None`` otherwise (as implemented right now), or do we want to have a method that returns a matrix in all possible cases.

I don't think we can implement different traits for different variants of the same `enum`, i.e. `SingleQubitOperation` for `StandardGate::X` and `TwoQubitOperation` for `StandardGate:CX`, and in particular I don't think we have a good alternative for (3): for performance considerations it's probably not worth to wrap the returned matrix into some other enum. The current decisions for (1) and (2) are based on an offline discussion with @mtreinish and transitively with @jakelishman; Matthew or Jake, could you please comment on this, especially how do you see the bigger picture of how the `SingleQubitOperation` trait might evolve in the future.

jakelishman:
A bit on naming / responsibilities first:

Using an "extension" trait is a fine pattern in Rust, but typically the reason to have a supertrait is because you either don't want it to always available (e.g. the supertrait might be access-restricted), or because you want to implement it for a subset of the types that implement the base trait.  This sort of fulfills the last category, but in a lot of ways there's nothing specific to "single qubit" in the implementers, only that they're "gate" extension methods.

So on naming: is this trait something that is specific to single-qubit matrices, or would it just be "gate"?

---

Next, on typing / usability:

Traits operate only at the level of _types_, but our implementers of `Operation` include several stratifications based on the _value_, or the particular implementer of `Operation`.  "Gate" or "not gate" is a distinction we make at the type level (except for `PyOperation`, which might need more thought), but "single qubit" is one we make at the value level.  Sasha's comment

> I don't think we can implement different traits for different variants of the same enum

is another way of saying that same thing.  That typically means that the "single qubit" restriction isn't buying us a lot to encode it in the trait name itself, because it's not enforceable by the typing; just knowing you have an `impl SingleQubitOperation` is not sufficient to know that you actually have a `SingleQubitOperation`.

That can motivate two different approaches:

1. leave the trait methods as-is, but either rename the trait to something that implies it's just about "gates", or just inline the methods into `Operation` (since they're fallible anyway).
2. define a _new_ type or types (say `SingleQubitGate`) with a private fallible constructor, and wraps `StandardGate` and/or `PyGate` and/or `OperationRef`, etc etc, and rejects attempts to construct itself if the base doesn't hold the invariants of "single qubit gate".  For example:
   ```rust
   #[repr(transparent)]
   pub struct StandardGate1q(StandardGate);
   
   // ... in the impl of `StandardGate`:
   impl StandardGate {
       pub fn as_1q(self) -> Option<StandardGate1q> {
           (self.num_qubits() == 1).then_some(StandardGate1q(self)))
       }
   }
   ```
   
   You can have the new wrapper types implement `Deref` of their inner type to allow calling other methods directly, and you can give them a `as_parent` method if we need to get the parent back (or they can just wrap a reference to the inner object instead of taking ownership, etc.

   In this world, you now _can_ have a `SingleQubitGate` trait, and the gate-matrix methods can then be infallible.

In terms of usability: new traits always have to be imported before you can use their methods.  It might be the case that we want to have a `qiskit::prelude::*` import that imports all the operations traits, if we start having several of them.

---

On fallibility:

We currently have two major ways in which things like `SingleQubitGate` could have fallible methods:
1. as of this review comment, the implementer of the trait might not actually be a 1q gate (might be 2q, etc).
2. the implementer of the trait might be a custom Python gate that didn't provide a matrix.

My suggestions above can deal with the problem of 1, but not the problem of 2.  It would be nice if we didn't need to include fallibility just on the off-chance of needing to handle Python gates, but it's not super obvious to me if there's a clean way to do that.

alexanderivrii:
Thanks @jakelishman for the very detailed and informative reply!

As further discussed with @jakelishman offline, since we don't immediately see a perfect design decision, *as a short-term measure* it makes most sense to move the two new methods `get_mat_static_1q` and `get_mat_nalgebra_1q` into the `Operation` trait, and brainstorm on how to best reorganize the code in a followup.

Done in  [7f3c83e](https://github.com/Qiskit/qiskit/pull/14020/commits/7f3c83ec1faf82f56ea3ff648d2e8f9ca3d2b86e).

crates/circuit/src/operations.rs
```diff
@@ -2627,6 +2789,33 @@ impl Operation for PyGate {
...
+                        }
+                    }
+                    Err(_) => None,
+                },
+                _ => None,
+            }
+        })
+    }
+
+    fn get_mat_nalgebra_1q(&self, params: &[Param]) -> Option<Matrix2<Complex64>> {
```
Cryoris: Could this just be implemented in `Operation` itself to avoid repeating it in every enum? Something like
```
fn get_mat_nalgebra_1q(...) -> Option< ... > {
    self.get_mat_static_1q(...).map(... your map to Matrix2 ...)
```
Since `map` will leave `None` as `None` that should also work for PyOperation etc, I think

crates/circuit/src/operations.rs
```diff
@@ -286,6 +288,27 @@ impl Operation for OperationRef<'_> {
             Self::Unitary(unitary) => unitary.directive(),
         }
     }
+    #[inline]
+    fn get_mat_static_1q(&self, params: &[Param]) -> Option<[[Complex64; 2]; 2]> {
```
Cryoris: Given that the current method to get the matrix is called `matrix`, how about calling these `matrix_as_static` and `matrix_as_nalgebra` for a consistent naming?

crates/circuit/src/operations.rs
```diff
@@ -2396,6 +2426,132 @@ impl Operation for StandardGate {
...
+                [Param::Float(phi), Param::Float(lam)] => Some(gate_matrix::u2_gate(*phi, *lam)),
+                _ => None,
+            },
+            Self::U3 => match params {
+                [Param::Float(theta), Param::Float(phi), Param::Float(lam)] => {
+                    Some(gate_matrix::u3_gate(*theta, *phi, *lam))
+                }
+                _ => None,
+            },
+            Self::CH => None,
```
Cryoris, line 2510: Instead of typing all of them out maybe just `_ => None`? 🙂 (incl. the GlobalPhaseGate)

crates/circuit/src/operations.rs
```diff
@@ -286,6 +288,27 @@ impl Operation for OperationRef<'_> {
             Self::Unitary(unitary) => unitary.directive(),
         }
     }
+    #[inline]
```
Cryoris, line 297: Could you add a docstring that clearly states that this will return `None` if the operation is not 1q? Calling something like `StandardGate::CXGate.get_mat_static_1q(&[])` still seems quite strange to me, so a comment would help 🙂

test/python/transpiler/test_optimize_1q_decomposition.py
```diff
@@ -730,6 +732,52 @@ def test_prefer_no_substitution_if_all_ideal(self):
...
+        expected.u(0, 0, 0.5, [0])
+
+        self.assertEqual(result, expected)
+
+    def test_unitary_gate_ndarray(self):
+        """Test that pass handles unitary single-qubit gates."""
+
+        qc = QuantumCircuit(1)
+        # The Rust code is not able to interpret mat as nalgebra matrix
+        mat = np.asarray(SGate().to_matrix(), dtype=complex)
```
Cryoris, line 763: No need to change but I think you can also just do
```suggestion
        mat = np.array(SGate())
```
since `SGate` implements `__array__`

test/python/transpiler/test_optimize_1q_decomposition.py
```diff
@@ -730,6 +732,52 @@ def test_prefer_no_substitution_if_all_ideal(self):
...
+        result = Optimize1qGatesDecomposition(["cx", "u"])(qc)
+
+        expected = QuantumCircuit(1)
+        expected.u(0, 0, np.pi / 2, [0])
+        self.assertEqual(result, expected)
+
+    def test_unitary_gate_oneq(self):
+        """Test that pass handles unitary single-qubit gates."""
+        qc = QuantumCircuit(1)
+        # The Rust code is able to interpret mat as nalgebra matrix
```
Cryoris: Why does this path use `nalgebra` but the path in the other test doesn't?

crates/circuit/src/operations.rs
```diff
@@ -286,6 +288,27 @@ impl Operation for OperationRef<'_> {
...
+        match self {
+            Self::StandardGate(standard) => standard.get_mat_static_1q(params),
+            Self::StandardInstruction(instruction) => instruction.get_mat_static_1q(params),
+            Self::Gate(gate) => gate.get_mat_static_1q(params),
+            Self::Instruction(instruction) => instruction.get_mat_static_1q(params),
+            Self::Operation(operation) => operation.get_mat_static_1q(params),
+            Self::Unitary(unitary) => unitary.get_mat_static_1q(params),
+        }
+    }
+    fn get_mat_nalgebra_1q(&self, params: &[Param]) -> Option<Matrix2<Complex64>> {
```
Cryoris: I don't see the `nalgebra` function being used anywhere -- is there a plan for using this somewhere? Otherwise I'm usually in favor of not adding this, as it's also a path that is not tested 🤔

Cryoris (commented):
I left some comments below 🙂 

It still seems very strange to me to have a fixed-size-method on objects that vary in size. What Jake outlined above with the `StandardGate1q` would seem like a cleaner solution to me (or maybe a related solution with `TryInto<StandardGate1q> for StandardGate` which fails if the gate is not 1q). But I'm also happy to accept this approach to be pragmatic and if everyone else is fine with it (I'm still learning the ways of Rust so maybe this is a common scheme 🙂).

crates/circuit/src/operations.rs
jakelishman: I'm generally in favour of not using the wildcard match - we've already had a few bugs because of wildcard matching as "generically not handled" and then _whoops_ the data model expands and we forgot to update a bunch of places.

Cryoris: Yeah I agree in general, but here I thought we won't add standard gates anyways and it would save a lot of code 😄 but I'm also fine always expanding if the wildcard match passes silently 🙂

crates/circuit/src/operations.rs
alexanderivrii: Great suggestion, done in [7db5f09](https://github.com/Qiskit/qiskit/pull/14020/commits/7db5f09582584939c0d61a29aae5e36deee7fee3).

crates/circuit/src/operations.rs
alexanderivrii: This makes sense. The new names as `matrix_as_static_1q` and `matrix_as_nalgebra_1q` to emphasize that we only expect 1q matrices. Done in [d70aa09](https://github.com/Qiskit/qiskit/pull/14020/commits/d70aa099be85af8a6534341bfa61598eb1000194).

crates/circuit/src/operations.rs
alexanderivrii: I am fine either way, with slightly leaning on the side of explicitly mentioning all gates, for the reason that Jake mentioned. And you never know, we may want to add another standard gate in the future.

crates/circuit/src/operations.rs
alexanderivrii: Done in [d70aa09](https://github.com/Qiskit/qiskit/pull/14020/commits/d70aa099be85af8a6534341bfa61598eb1000194).

Technically the same "strangeness" applies when asking for ``matrix`` from a non-gate (such as from a ``StandardInstruction`` or a ``PyOperation``), so many of the methods within the ``Operation`` traits are fallible and return an ``Option<...>``.

test/python/transpiler/test_optimize_1q_decomposition.py
alexanderivrii: Great to know, thanks (when you give me a choice, I will take the "no need to change option", haha).

test/python/transpiler/test_optimize_1q_decomposition.py
alexanderivrii: That's because of explicitly creating the numpy array with column-major ordering:  see the ``order="f"``. Since ``nalgebra`` also uses column-major ordering, it will be able to interpret this as nalgebra matrix, but not when the row-major ordering is used.

crates/circuit/src/operations.rs
alexanderivrii: I didn't put it in initially, but Matthew has asked for it :). Anyhow, I have locally double-checked that the function works (modifying the ``euler_one_qubit_decomposer`` to work with nalgebra matrices instead). In general, we probably need to start adding more pure rust tests. 

Yes, the function would be used when fully switching from faer to nalgebra.

crates/circuit/src/operations.rs
Cryoris: Hmmm I can't find the docstring, did you push it or did I just not find it?

alexanderivrii: Oops, indeed, I forgot to push this, done :).

test/python/transpiler/test_optimize_1q_decomposition.py
mtreinish: This does raise the question of whether we should update the conversion from Python -> Rust to be order independent. We're going to make an owned copy anyway so we should just convert any 2x2 to a `Matrix2` in rust and do the copy manually.

alexanderivrii: To make sure I have understood the suggestion correctly, the idea is to always produce the more efficient ``OneQ`` and ``TwoQ`` representations (instead of the ``NDArray`` one) when converting a 1q or 2q Python unitary matrix to Rust, independently of the ordering. I have done this in [d39fcf5](https://github.com/Qiskit/qiskit/pull/14020/commits/d39fcf5586ad262da2758926a516d0a3d0901565).

crates/circuit/src/circuit_instruction.rs
```diff
@@ -580,7 +580,7 @@ impl<'py> FromPyObject<'py> for OperationFromPython {
             let params = extract_params()?;
             if let Some(Param::Obj(data)) = params.first() {
                 let py_matrix: PyReadonlyArray2<Complex64> = data.extract(py)?;
-                let matrix: Option<MatrixView2<Complex64>> = py_matrix.try_as_matrix();
+                let matrix: Option<MatrixView2<Complex64, Dyn, Dyn>> = py_matrix.try_as_matrix();
```
mtreinish, line 583: This is saying we will get a readonly view of a dynamic stride 2x2 array which is potentially mismatched from what our memory layout expectations are for the owned `Array2` in `ArrayType::OneQ`. I was worried if this was doing the right thing so I checked the nalgebra source and this works because `.to_owned()` below builds a new copy with the default allocator and layout by iterating over the original view: https://github.com/dimforge/nalgebra/blob/db2d242dd9032224bdaf4a38a072dfeae2de88c4/src/base/matrix_view.rs#L226-L231 so it always has the same memory layout we expect.

This should mean that we are storing all 2x2s in a `Matrix2` now which seems like a good choice.

crates/circuit/src/operations.rs
```diff
@@ -2759,6 +2932,20 @@ impl Operation for UnitaryGate {
...
+                    Some([[arr[(0, 0)], arr[(0, 1)]], [arr[(1, 0)], arr[(1, 1)]]])
+                }
+                ArrayType::OneQ(mat) => {
+                    Some([[mat[(0, 0)], mat[(0, 1)]], [mat[(1, 0)], mat[(1, 1)]]])
+                }
+                _ => None,
+            },
+            _ => None,
+        }
+    }
```
mtreinish: I'd be tempted to invert the logic here so it's:

```suggestion
        match &self.array {
            ArrayType::OneQ(mat) => {
                Some([[mat[(0, 0)], mat[(0, 1)]], [mat[(1, 0)], mat[(1, 1)]]])
            }
            ArrayType::NDArray(arr) => {
                if self.num_qubits() == 1 {
                    Some([[arr[(0, 0)], arr[(0, 1)]], [arr[(1, 0)], arr[(1, 1)]]])
                } else {
                    None
                }
            }                        
            ArrayType::TwoQ(_) => None,                        
        }
    }
```

It feels slightly more compact and we're avoiding inspecting two conditions in the case of `ArrayType::OneQ` although these checks are so fast that performance doesn't matter.

crates/circuit/src/operations.rs
```diff
@@ -2759,6 +2932,20 @@ impl Operation for UnitaryGate {
     fn directive(&self) -> bool {
         false
     }
+    fn matrix_as_static_1q(&self, _params: &[Param]) -> Option<[[Complex64; 2]; 2]> {
```
mtreinish, line 2929: I wonder if it's worth overloading `matrix_as_nalgebra_1q` here so we can just do a copy if it's an `ArrayType::OneQ` and avoid going through the extra layer of function indirection.

mtreinish (commented):
This LGTM I'm looking forward to having this. I just had a couple of questions inline about the impl for the `UnitaryGate` path.

crates/circuit/src/circuit_instruction.rs
alexanderivrii: Yes, I have also previously double-checked (using `println!` statements) that we enter the `ArrayType::OneQ` branch.

crates/circuit/src/operations.rs
alexanderivrii: I like the suggested way, done in [948c82e](https://github.com/Qiskit/qiskit/pull/14020/commits/948c82e1244e0a25191cc4a3e284b2c702563150)

crates/circuit/src/operations.rs
alexanderivrii: That's a good point, the code is converting from `Matrix2` to `[[Complex64; 2]; 2]` and back to `Matrix2`. Done in [29752b9](https://github.com/Qiskit/qiskit/pull/14020/commits/29752b923d8c9e2d35bde5c21bf852a32bc8ca52).

crates/circuit/src/operations.rs
```diff
@@ -2627,6 +2774,28 @@ impl Operation for PyGate {
...
+                                Some([[arr[[0, 0]], arr[[0, 1]]], [arr[[1, 0]], arr[[1, 1]]]])
+                            }
+                            None => None,
+                        }
+                    }
+                    Err(_) => None,
+                },
+                _ => None,
+            }
+        })
```
mtreinish: I would rewrite this as to make it a bit more compact:

```suggestion
        if self.num_qubits() != 1 {
            return None;
        }
        Python::with_gil(|py| -> Option<[[Complex64; 2]; 2]> {
            let array = self
                .gate
                .call_method0(py, intern!(py, "to_matrix"))
                .ok()?
                .extract::<PyReadonlyArray2<Complex64>>(py)
                .ok()?;
            let arr = array.as_array();
            Some([[arr[[0, 0]], arr[[0, 1]]], [arr[[1, 0]], arr[[1, 1]]]])
        })
```

mtreinish (commented):
I have one more small inline suggestion on the pygate implementation. After that I think this should be good to merge.

mtreinish (approved):

alexanderivrii:
Thanks Matthew for taking care of this, with your change the `PyGate` function is significantly better and easier to read!

PR #14022, opened 2025-03-14
 Introduce custom sympy srepr parser [stable/1.4]

mtreinish:
### Summary

This commit fixes this issue by writing a custom parser for the string repesentation in a QPY payload based on python's ast module. This parser walks the abstract syntax tree and builds the sympy expression object as it it goes. It is restricted to the operations that ParameterExpression supports and if any part of the string tries to use functionality outside that set it will error.

For ScheduleBlock objects which also serialize a symbolic expression there is no new parser to securely load the payload. This is because the data model of ScheduleBlock enables using any sympy expression and the limited parser is not sufficient to represent the schedule block. Since this functionality is deprecated and will be removed in Qiskit 2.0.0, potentially vulnerable payloads are rejected by default now. If the input is trusted by the user they can explicitly opt-in to using it by setting a new flag to trust the input. This is technically a breaking API change, but since the payload is potentially vulnerable this is a necessary change to secure QPY.


ElePT (approved):
LGTM, I think that it should be ready to merge if it passes CI.

PR #14023, opened 2025-03-14
 Introduce custom sympy srepr parser [stable/2.0]

mtreinish:
### Summary

This commit fixes this issue by writing a custom parser for the string representation in a QPY payload based on python's ast module. This parser walks the abstract syntax tree and builds the sympy expression object as it it goes. It is restricted to the operations that ParameterExpression supports and if any part of the string tries to use functionality outside that set it will error.

### Details and comments


test/python/qpy/test_circuit_load_from_qpy.py
```diff
@@ -289,6 +289,7 @@ def test_rdiv(self):
         self.assert_roundtrip_equal(qc)
 
 
+@ddt
```
ElePT:
```suggestion
```
oversight?

releasenotes/notes/fix-ace-qpyv1-609551d84531e016.yaml
```diff
@@ -0,0 +1,9 @@
+---
+security:
+  - |
+    Fixed a security vulnerability in :func:`.qpy.load` when it was loading
+    payloads that are using ``sympy`` to serialize
+    :class:`.ParameterExpression` objects and other symbolic expressions.
+    This potentially includes any QPY payload using QPY version < 10, and
+    optionally 10, 11, and 12 depending on the symbolic encoding used in the
+    payload.
```
ElePT: Same suggestion as in the 1.4.2 reno:

```suggestion
    Fixed a security vulnerability in :func:`.qpy.load` when loading
    payloads that use ``sympy`` to serialize
    :class:`.ParameterExpression` objects and other symbolic expressions.
    This potentially includes any QPY payload using QPY version < 10, and
    optionally 10, 11, and 12 depending on the symbolic encoding used in the
    serialization step (:func:`.qpy.dump`).
```

ElePT (commented):
LGTM! I just have two minor comments and then I think it's ready.

ElePT:
Is the plan to also add version changes to this PR and use it to release like in #14022?

mtreinish:
> Is the plan to also add version changes to this PR and use it to release like in #14022?

Let me do that now and apply your suggestions, that will save us some CI time.

test/python/qpy/test_circuit_load_from_qpy.py
mtreinish: Yeah this was a leftover from the 1.4.2 PR's testing which no longer applies to the 2.0.0 branch because we've dropped exporting qpy < 13.

ElePT (dismissed):
LGTM

raynelfss (approved):

PR #14024, opened 2025-03-14
Introduce custom sympy srepr parser

mtreinish:
### Summary

This commit fixes this issue by writing a custom parser for the string representation in a QPY payload based on python's ast module. This parser walks the abstract syntax tree and builds the sympy expression object as it it goes. It is restricted to the operations that ParameterExpression supports and if any part of the string tries to use functionality outside that set it will error.

### Details and comments


ElePT (approved):
LGTM!

PR #3695, opened 2020-01-07
Consistent naming of the standard gates

Cryoris:
### Summary

Ensure consistent naming of the standard gates according to #3633 

### Details and comments

This PR ensures that the names of the standard gates follow the same consistent principles. See #3633 for more detail and the discussion on gate naming. The old/current gates will be deprecated.

Up-to-date list of the names:
```
| file                            | class                    | name           | qc._           |
|---------------------------------+--------------------------+----------------+----------------|
  extensions/standard                 
| barrier.py                      | Barrier                  | barrier        | barrier        |
| ccx.py                          | CCXGate <- ToffoliGate   | ccx            | ccx, toffoli   |
| ch.py                           | CHGate                   | ch             | ch             |
| crx.py                          | CRXGate <- CrxGate       | crx            | crx            |
| cry.py                          | CRYGate <- CryGate       | cry            | cry            |
| crz.py                          | CRZGate <- CrzGate       | crz            | crz            |
| cswap.py                        | CSwapGate <- FredkinGate | cswap          | cswap, fredkin |
| cu1.py                          | CU1Gate <- Cu1Gate       | cu1            | cu1            |
| cu3.py                          | CU3Gate <- Cu3Gate       | cu3            | cu3            |
| cx.py                           | CXGate <- CnotGate       | cx             | cx, cnot       |
| cy.py                           | CYGate <- CyGate         | cy             | cy             |
| cz.py                           | CZGate <- CzGate         | cz             | cz             |
| h.py                            | HGate                    | h              | h              |
| iden.py                         | IGate <- IdGate          | id             | i/id <- iden   |
| ms.py                           | MSGate                   | ms             | ms             |
| r.py                            | RGate                    | r              | r              |
| rx.py                           | RXGate                   | rx             | rx             |
| rxx.py                          | RXXGate                  | rxx            | rxx            |
| ry.py                           | RYGate                   | ry             | ry             |
| rz.py                           | RZGate                   | rz             | rz             |
| rzz.py                          | RZZGate                  | rzz            | rzz            |
| s.py                            | SGate                    | s              | s              |
| s.py                            | SdgGate                  | sdg            | sdg            |
| swap.py                         | SwapGate                 | swap           | swap           |
| t.py                            | TGate                    | t              | t              |
| t.py                            | TdgGate                  | tdg            | tdg            |
| u1.py                           | U1Gate                   | u1             | u1             |
| u2.py                           | U2Gate                   | u2             | u2             |
[...]

PR #14026, opened 2025-03-14
Concise ``StandardGate::<ID>`` enums in Rust space

Cryoris:
### Summary

The Rust-side, internal `StandardGate` enum variants all repeat the fact that they are a gate. For example, the Hadamard is ``StandardGate::HGate`` instead of simply ``StandardGate::H``. Being an enum variant, the fact that it is a gate is evident and always attached to the object. I suspect this is mainly a legacy naming from moving over from Python space (or maybe @mtreinish comes remembers 🙂)? 

I'd suggest to shorten the enum variants to just the gate ID. This makes the code more concise and avoid redundant qualifiers. It also enables that gate enums in the C API, introduced in #14006, can be called `QkGate_H` instead of `QkGate_HGate`.


crates/accelerate/src/circuit_library/pauli_evolution.rs
```diff
@@ -37,8 +37,8 @@ type Instruction = (
 ///     time: The rotation angle. Note that this will directly be used as input of the
 ///         rotation gate and not be multiplied by a factor of 2 (that should be done before so
 ///         that this function can remain Rust-only).
-///     phase_gate_for_paulis: If ``true``, use the ``PhaseGate`` instead of ``RZGate`` as
-///         single-qubit Pauli rotations.
+///     phase_gate_for_paulis: If ``true``, use the :class:`.PhaseGate` instead of :class:`.RZGate`
```
kevinhartman: This is not a Python function, so you need Rust-style docs :)

Maybe that's `[StandardGate::Phase]` and `[StandardGate::RZ]`.

crates/accelerate/src/commutation_cancellation.rs
```diff
@@ -20,10 +20,7 @@ use rustworkx_core::petgraph::stable_graph::NodeIndex;
 use smallvec::{smallvec, SmallVec};
 
 use qiskit_circuit::dag_circuit::{DAGCircuit, NodeType, Wire};
-use qiskit_circuit::operations::StandardGate::{
-    CXGate, CYGate, CZGate, HGate, PhaseGate, RXGate, RZGate, SGate, TGate, U1Gate, XGate, YGate,
-    ZGate,
-};
+use qiskit_circuit::operations::StandardGate::{Phase, CX, CY, CZ, H, RX, RZ, S, T, U1, X, Y, Z};
```
kevinhartman: IMO, enums are always better fully-qualified. The `use` statement here is preexisting, but having single character identifiers floating around feels odd.

crates/accelerate/src/remove_diagonal_gates_before_measure.rs
```diff
@@ -26,25 +26,25 @@ use qiskit_circuit::operations::StandardGate;
...
-        StandardGate::SGate,
-        StandardGate::TdgGate,
-        StandardGate::SdgGate,
-        StandardGate::U1Gate,
-        StandardGate::PhaseGate,
+        StandardGate::RZ,
+        StandardGate::Z,
+        StandardGate::T,
+        StandardGate::S,
+        StandardGate::Tdg,
```
kevinhartman, line 33: Is there a good reason it's `Tdg` and not `TDG`, or is that just historical?

crates/accelerate/src/twirling.rs
```diff
@@ -33,7 +33,7 @@ use qiskit_circuit::converters::dag_to_circuit;
 use qiskit_circuit::dag_circuit::DAGCircuit;
 use qiskit_circuit::gate_matrix::ONE_QUBIT_IDENTITY;
 use qiskit_circuit::imports::QUANTUM_CIRCUIT;
-use qiskit_circuit::operations::StandardGate::{IGate, XGate, YGate, ZGate};
+use qiskit_circuit::operations::StandardGate::{I, X, Y, Z};
```
kevinhartman, line 36: ... maybe this is nice in this file ;)

kevinhartman (commented):
This seems like a nice improvement to me. Just a few minor comments.

(I didn't look too closely for typos--I assume you used an IDE to do symbolic renaming 😉)

crates/accelerate/src/circuit_library/pauli_evolution.rs
Cryoris: Whoops thanks -- IDE renaming only gets one so far ... 😂

crates/accelerate/src/remove_diagonal_gates_before_measure.rs
Cryoris: There's a good historical reason 😛 See https://github.com/Qiskit/qiskit/pull/3695 and the associated issue, we decided the naming norm is: a single uppercase letter per gate type (e.g. `T` for T-gate, `C` for controlled, `R` for a rotation, `S` for square root...) and `dg` being just an inverse qualifier remains lowercase. Ofc there are exceptions, like `SwapGate` (no single letter gate type), or `SGate` (which I guess would technically be `SZ` as square root of Z) but that's minimal 🙂

crates/accelerate/src/commutation_cancellation.rs
Cryoris: Done!

crates/accelerate/src/twirling.rs
Cryoris: nothing done here 😉

kevinhartman (approved):
This looks good to me.

I think it's important for consistent naming in our C-API, and @mtreinish could have arbitrarily chosen to do this in the original standard gates PR, so I feel confident to merge this without more consensus. We can always change this again if anyone comes up with a good reason.

PR #14027, opened 2025-03-14
Allow circuits using only SI units to pass through TimeUnitConversion

### LINKED ISSUE #14016: Transpiling circuit with SI unit delay raises exception with Qiskit 2.0
### Environment

- **Qiskit version**: 3a80a814379f3888fbe26512f07c2ae6a601d2a9
- **Python version**: 3.12
- **Operating system**: Fedora Linux 41

### What is happening?

Transpiling a circuit with a delay in SI units causes an exception when not specifying a dt.

This code:

```python
from qiskit import QuantumCircuit, generate_preset_pass_manager


circ = QuantumCircuit(1, 1)
circ.x(0)
circ.delay(10e-6, unit="s")
circ.measure(0, 0)


pm = generate_preset_pass_manager()
pm.run(circ)
```

causes

```
  File "/qiskit-terra/qiskit/transpiler/passes/scheduling/time_unit_conversion.py", line 134, in run
    raise TranspilerError(
qiskit.transpiler.exceptions.TranspilerError: 'Fail to unify time units. SI units and dt unit must not be mixed when dt is not supplied.'
```

If a backend with a `dt` is supplied like this:

```python
from qiskit import QuantumCircuit, generate_preset_pass_manager
from qiskit_ibm_runtime.fake_provider import FakeVigoV2


circ = QuantumCircuit(1, 1)
circ.x(0)
circ.delay(10e-6, unit="s")
circ.measure(0, 0)


backend = FakeVigoV2()
pm = generate_preset_pass_manager(backend=backend)
pm.run(circ)
```

then there is no error (note: this currently requires the main branch of qiskit-ibm-runtime since the last release will not import with Qiskit 2.0).

### What should happen?

I am not sure 🙂  The message suggests that "SI units and dt unit must not be mixed when dt is not supplied" but the circuit only uses SI. So either the message should be updated to say that SI should not be used without passing `dt` to the transpiler or this case when only SI delays are used without passing dt should be allowed to work.

### Any suggestions?

_No response_

wshanks:
In f0576fa64, the `TimeUnitConversion` pass was modified so that it raises an exception if a circuit has only SI unit delays with no `dt` value set, but the intent seemed to be catch cases where SI units were used alongside instructions with `dt` units. Here the conditions for raising an exception are modified to allow circuits with all SI unit delays (which usually only happens when working with a backend like `AerSimulator`) to pass through.

Closes https://github.com/Qiskit/qiskit/issues/14016


mtreinish (commented):
I think this change makes sense, do you want to add a test to ensure we don't regress here again?

wshanks:
I thought about that but I wasn't sure where the test would go. I didn't see much in the way of existing tests. I think a new `test/python/transpiler/test_time_unit_conversion.py` would be needed?

kevinhartman:
That is most probably a bug, thanks for finding it.

There is already a test that makes sure that SI-only can pass through when `dt` is not specified:
https://github.com/Qiskit/qiskit/blob/main/test/python/compiler/test_transpiler.py#L1439

But it tests this only for new expression-based delay durations. Perhaps you can add a test beside it for the old-style delay duration specification.

qiskit/transpiler/passes/scheduling/time_unit_conversion.py
```diff
@@ -126,9 +126,9 @@ def run(self, dag: DAGCircuit):
             # Check what units are used in other instructions: dt or SI or mixed
             units_other = inst_durations.units_used()
             unified_unit = self._unified(units_other)
-            if unified_unit == "SI" and not has_dt:
+            if unified_unit in ("SI", "none") and has_si and not has_dt:
```
kevinhartman: I _think_ we may just want this:

```py
if (has_si or unified_unit == "SI") and not has_dt:
   time_unit = "s"
elif (has_dt or unified_unit == "dt") and not has_si:
   time_unit = "dt"
```

If `_unified` returns `"dt"` because its input set is empty, then the `elif` gets executed, which is the behavior we want (defaulting to `dt`).

And wouldn't need the change to return `"none"`.

wshanks: Ugh, there are so many edge cases :slightly_smiling_face: With `(has_si or unified_unit == "SI") and not has_dt` is that it would be true for a circuit with all SI unit delays but with instructions in `InstructionDurations` that were in dt. Isn't that a problem? (You could also consider the edge case where the instructions with dt units in `InstructionDurations` are not used in the current circuit).

wshanks:
@kevinhartman I wrote some tests before I saw your comment about the existing tests. I pushed them for reference. We move or rework them if you think that is better.

qiskit/transpiler/passes/scheduling/time_unit_conversion.py
kevinhartman: Yup, you're right. Would something like this be more readable? This assumes the presence of your addition of returning `"none"` in `_unified`.

```py
if inst_durations.dt is None:
    # Check what units are used in other instructions: dt or SI or mixed
    units_other = inst_durations.units_used()
    unified_unit = self._unified(units_other)
    has_si = has_si or unified_unit in {"SI", "mixed"}
    has_dt = has_dt or unified_unit in {"dt", "mixed"}
    if has_dt and has_si:
        raise TranspilerError(
            "Fail to unify time units. SI units "
            "and dt unit must not be mixed when dt is not supplied."
        )
    if has_si:
        time_unit = "s"
    else:
        # has_dt may be false (i.e. 'unified_unit' is "none"), but we still default to 'dt'
        # in this case.
        time_unit = "dt"
else:
    time_unit = "dt"
```

wshanks: That looks good to me. I tried pushing a version like that.

test/python/transpiler/test_time_unit_conversion.py
```diff
@@ -0,0 +1,90 @@
...
+            "si": InstructionDurations([("x", None, self.x_dt * self.dt, None, "s")]),
+            "dt": InstructionDurations([("x", None, self.x_dt, None, "dt")]),
+            "none": None,
+        }
+
+    @idata(itertools.product(("s", "dt"), ("dt", "si", "none"), ("dt", "nodt", "none")))
+    @unpack
+    def test_gate_and_delay(self, delay_unit, inst_durations_unit, target_type):
+        """Test delays are converted, passed through or flagged as appropriate"""
+        if delay_unit == "s":
```
kevinhartman: I think we generally prefer to "unroll" test code so it's simple to debug when something goes wrong. The orange flag for me here is that the test contains logic to special-case itself based on the `ddt` inputs. If you don't mind, I think it'd be helpful if you could break this into three separate tests instead, even though it'll be more verbose.

test/python/transpiler/test_time_unit_conversion.py
```diff
@@ -0,0 +1,90 @@
...
+
+        self.durations = {
+            "si": InstructionDurations([("x", None, self.x_dt * self.dt, None, "s")]),
+            "dt": InstructionDurations([("x", None, self.x_dt, None, "dt")]),
+            "none": None,
+        }
+
+    @idata(itertools.product(("s", "dt"), ("dt", "si", "none"), ("dt", "nodt", "none")))
+    @unpack
+    def test_gate_and_delay(self, delay_unit, inst_durations_unit, target_type):
```
kevinhartman: Perhaps the name here could be a bit more specific?

test/python/transpiler/test_time_unit_conversion.py
```diff
@@ -0,0 +1,90 @@
...
+from qiskit.circuit.library import XGate
+from qiskit.transpiler import InstructionDurations, InstructionProperties, Target
+from qiskit.transpiler.passes import TimeUnitConversion
+from qiskit.converters import circuit_to_dag
+from qiskit.transpiler.exceptions import TranspilerError
+from test import QiskitTestCase  # pylint: disable=wrong-import-order
+
+
+@ddt
+class TestTimeUnitConversion(QiskitTestCase):
```
kevinhartman: Personally, I like having separate files for different transpiler tests, but in this case, I think it'd be best to place your new tests in `test_transpiler.py` beside the others. And, I'd just inline all of the target setup stuff (as you'll see, we do that for the existing tests), since it's fairly specific for what we're testing.

kevinhartman (commented):
Looking good, just a few minor comments with the testing.

test/python/transpiler/test_time_unit_conversion.py
wshanks: Tests moved to test_transpiler in [b749029](https://github.com/Qiskit/qiskit/pull/14027/commits/b7490299ec032dc501770e0a912175a89c20a90f).

test/python/transpiler/test_time_unit_conversion.py
wshanks: Most of the labeling was in the `TestTimeUnitConversion` and ddt-label parts of the test name, but the tests are moved into test_transpiler now in [b749029](https://github.com/Qiskit/qiskit/pull/14027/commits/b7490299ec032dc501770e0a912175a89c20a90f).

test/python/compiler/test_transpiler.py
```diff
@@ -90,6 +90,7 @@
 from qiskit.transpiler.passmanager_config import PassManagerConfig
 from qiskit.transpiler.preset_passmanagers import generate_preset_pass_manager, level_0_pass_manager
 from qiskit.transpiler.target import (
+    InstructionDurations,
```
kevinhartman: Lint is likely to yell at you here 🙂.

kevinhartman (dismissed):
Looks great, thanks a lot for the fix!

If it doesn't pass lint, I'll come back to approve again after you update.

test/python/transpiler/test_time_unit_conversion.py
wshanks: I moved the tests to test_transpiler in [b749029](https://github.com/Qiskit/qiskit/pull/14027/commits/b7490299ec032dc501770e0a912175a89c20a90f). To fit in there, I used `transpile()` instead of the `TimeUnitConversion` interface. That shifted the scope of what seemed like relevant tests. `TimeUnitConversion` takes instruction durations and a target and behaves differently depending on the units of the `InstructionDurations` object and whether or not it or the target have a dt. With `transpile`, the durations only come from the backend and that just comes from `backend.target.durations()`. In Qiskit, `InstructionProperties` specifies duration in seconds so that prevents dt units in the `InstructionDurations`. So that removes a lot of the variants I was covering with my tests before. The si->dt case already had coverage so I added the most common cases of si->si with no backend or a backend without dt and si->error with mixed units and no dt. I think the other nuances my previous tests were covering about instruction durations versus target precedence are not so important for `transpile` since the input is only backend or target and not instruction durations directly.

test/python/compiler/test_transpiler.py
wshanks: Oh, right, this was from when I was trying to cover `InstructionDurations` in the old test before I realized that it was not really directly relevant for `transpile` since `InstructionProperties` is in seconds and `transpile` does not take `instruction_durations` as an input.

test/python/transpiler/test_time_unit_conversion.py
kevinhartman: Makes sense. Presumably, the tests you've added wouldn't have passed before your change at least, yeah?

wshanks: `test.python.compiler.test_transpiler.TestTranspile.test_delay_converts_to_seconds` fails as expected:

```
test.python.compiler.test_transpiler.TestTranspile.test_delay_converts_to_seconds                                                                                                                                                            
---------------------------------------------------------------------------------                                                                                                                                                            
                                                                                                                                                                                                                                             
Captured traceback:                                                                                                                                                                                                                          
~~~~~~~~~~~~~~~~~~~                                                                                                                                                                                                                          
                                                                                                                                                                                                                                             
      File "/qiskit-terra/test/python/compiler/test_transpiler.py", line 1382, in test_delay_converts_to_seconds                                                                                                 
    out = transpile([qc, qc], seed_transpiler=42)                                                                                                                                                                                            
                                                                                                                                                                                                                                             
      File "/qiskit-terra/qiskit/compiler/transpiler.py", line 291, in transpile                                                                                                                                 
    out_circuits = pm.run(circuits, callback=callback, num_processes=num_processes)
...
      File "/qiskit-terra/qiskit/transpiler/passes/scheduling/time_unit_conversion.py", line 134, in run
    raise TranspilerError(
    ...<2 lines>...
    )

    qiskit.transpiler.exceptions.TranspilerError: 'Fail to unify time units. SI units and dt unit must not be mixed when dt is not supplied.'
```

`test_rejects_mixed_units_delay_without_target_dt` passes, but I added that one because I was trying to carry over my ddt tests and didn't see a test like it (there was an expression version) and not because that behavior was affected by my change.

wshanks:
Oops, there was a bad variable name left in from when I was debugging and some stray changes in tox.ini which I have removed now. Hopefully the tests are clean now.

(I did lint locally but just missed the warnings. I don't have a smooth setup for Qiskit right now which is part of why I was fiddling with tox.ini. I was getting extra linting warnings possibly because of some other pylintrc file I have or something like that).

kevinhartman (approved):
Thanks for the updates!

This looks good to me now.

PR #14029, opened 2025-03-15
Fix new clippy warnings from Rust 1.85

mtreinish:
### Summary

The recently release Rust 1.85 introduced some new clippy rules that are triggering warnings in the Qiskit Rust code. This commit fixes the warnings flagged by these new rules.

### Details and comments


Cryoris (approved):
I verified locally that these were the only clippy warnings on 1.85 👍🏻

PR #14030, opened 2025-03-15
Run C tests on all tier 1 platforms

mtreinish:
### Summary

This commit expands the testing matrix for the C API tests to cover all the supported tier 1 platforms. As tier 1 platforms we document that we test them all on every PR, historically that's just been the Python testing, but as we grow the capabilities of our C API in Qiskit we should expand that guarantee to testing on all the tier 1 platforms too.

There will need to be some iteration done to ensure they all work as expected, it's unlikely that this will work by default on Windows and we can tweak the configuration based on the results of CI runs.

### Details and comments


test/c/common.h
```diff
@@ -10,8 +10,16 @@
...
 
+#include <complex.h>
+#include <qiskit.h>
 #include <stdio.h>
 
+#ifdef _MSC_VER
+QkComplex64 make_complex_double(double real, double imag) { return (QkComplex64){real, imag}; }
+#else
+QkComplex64 make_complex_double(double real, double imag) { return real + I * imag; }
+#endif
```
jakelishman, line 21: Ah ok, this really goes back to what I was saying on #14037.  If we're going to do this (i.e. provide a direct constructor for `QkComplex64`, then we might want to do away with the fake type at all, and just clearly define the structure and potentially something like a `qk_complex_as_native` conditionally compiled method that returns either `complex double` or `_Dcomplex` as appropriate (using this conditional compilation trick).

Doing the false `typedef double complex QkComplex64` was meant to avoid needing the custom constructor, but if we have to abstract it out _anyway_, there's little benefit to not just saying the type is what it actually is, I think.

Fwiw, the `complex double` constructor should probably be `CMPLX(re, im)` (`CMPLX` is a macro defined in `complex.h`) - there's subtle floating-point signed-number differences with what you've got here.

Makefile
```diff
@@ -143,5 +143,13 @@ ctest: $(C_QISKIT_H)
 	# -V ensures we always produce a logging output to indicate the subtests
 	ctest -V --test-dir $(C_DIR_TEST_BUILD)
 
+ctest_win: $(C_QISKIT_H)
```
kevinhartman: The name here feels a bit odd. What's this really doing? Shouldn't we call it `ctest_debug` or something like that?

.github/workflows/ctests.yml
```diff
@@ -10,11 +10,36 @@ on:
...
+          - ubuntu-24.04-arm
     steps:
       - uses: actions/checkout@v4
       - name: Install dependencies
         run: cargo install cbindgen
       - name: Run tests
-        run: make ctest
\ No newline at end of file
+        run: make ctest
+  win_tests:
```
kevinhartman: It looks to me like the only difference with the Windows run is that we run `ctest` in Debug. This feels a bit obfuscated here. Could we add a comment to explain why?

test/c/CMakeLists.txt
```diff
@@ -26,6 +26,12 @@ target_include_directories (test_driver PRIVATE ${CMAKE_SOURCE_DIR})
 # ...and linked with the qiskit library.
 target_link_libraries (test_driver ${qiskit})
 
+# On MSVC we need to link the Python dll, we search it here and adjust the PATH in the tests below
+if (MSVC)
+    find_package(Python REQUIRED Interpreter Development)
```
kevinhartman: Do we know that this will always find the right DLL? It might be helpful to look at how PyO3 determines which Python to link against here: https://pyo3.rs/v0.24.0/building-and-distribution.html#configuring-the-python-version

Makefile
mtreinish: I thought this was actually a windows specific `Debug` flag. When I tried it at the time of writing setting `-C Debug` did not work locally for me. But I just checked this out again and ran `make ctest_win` and it seemed to work fine. I think it's probably fine to just reconsolidate the commands and job definitions if it does work fine

.github/workflows/ctests.yml
mtreinish: Removed in: https://github.com/Qiskit/qiskit/pull/14030/commits/e04ccbe0def5d73515a5e470f541f4ff0d194bf2

Makefile
mtreinish: I removed this in: https://github.com/Qiskit/qiskit/pull/14030/commits/e04ccbe0def5d73515a5e470f541f4ff0d194bf2

test/c/CMakeLists.txt
Cryoris: It looks like we can maybe run `find_package(Python ..)` first and then tell pyo3 to use that version using `PYO3_PYTHON=...`, but I'm not yet sure how to make pyo3 tell us which version of Python it is using and e.g. dump it into a config file during the build process.

The current solution however was cross checked locally by @mrossinek on his windows VM 🙂

Cryoris: Using `pyo3_build_config` to dump the Python libraries seems to work 🙂

test/c/test_sparse_observable.c
```diff
@@ -77,30 +77,30 @@ int test_add() {
...
  * Test composing two observables.
  */
-int test_compose() {
-    u_int32_t num_qubits = 100;
+int test_compose(void) {
+    uint32_t num_qubits = 100;
 
     QkObs *op1 = qk_obs_zero(num_qubits);
-    complex double coeff = 1;
+    QkComplex64 coeff1 = make_complex_double(1.0, 0.0);
```
eliarbel, line 84: Thinking about unit-testing as another type of documentation, I would refrain from exemplifying testing-only constructs like `make_complex_double` for something quite common like `QkCompex64` initialization.   Consider my comment here: https://github.com/Qiskit/qiskit/pull/14037/files#r2020149452.

test/c/common.h
eliarbel: Taking @jakelishman's point a bit further, I would delegate defining a standard complex var initializer to `qiskit.h`. See suggestion [here](https://github.com/Qiskit/qiskit/pull/14037/files#r2020149452) for details. 
If we adopt my suggestion in #14037, I would then remove the def here entirely.

eliarbel (approved):
For the mere purpose of enabling C tests on all tier 1 platforms, this seems good to me so will go ahead and approve. The question of how complex numbers on the C API side are handled in an ergonomic and compiler-agnostic way remained open, but at this stage we should probably just move forward with this PR.
We'll continue the discussion in #14037.

PR #14037, opened 2025-03-17
MSVC complex number support in `qiskit.h`

Cryoris:
### Summary

In MSVC complex numbers are structs called `_Dcomplex` (for `complex double`) and `_Fcomplex` (for `complex float`). This commit covers the MSVC compiler.

### Details and comments

Tested on @anedumla's 64-bit windows machine and MSVC version 19.34.31933. Maybe we can add this to #14030 in case this gets windows CI to pass? Though there might be some extra work required..

### DIFF
--- a/crates/cext/cbindgen.toml
+++ b/crates/cext/cbindgen.toml
@@ -7,9 +7,14 @@ after_includes = """
 #endif
 
 // Complex number typedefs -- note these are memory aligned but
-// not calling convention compatible.
+// not calling convention compatible. 
+#ifdef _MSC_VER
+typedef _Fcomplex QkComplex32;
+typedef _Dcomplex QkComplex64;
+#else
 typedef float complex QkComplex32;
 typedef double complex QkComplex64;
+#endif
 
 // Always expose [cfg(feature = "cbinding")] -- workaround for
 // https://github.com/mozilla/cbindgen/issues/995


Cryoris:
This is likely going to be superseded by #14030.

crates/cext/cbindgen.toml
```diff
@@ -7,9 +7,14 @@ after_includes = """
...
 // Complex number typedefs -- note these are memory aligned but
-// not calling convention compatible.
+// not calling convention compatible. 
+#ifdef _MSC_VER
+typedef _Fcomplex QkComplex32;
+typedef _Dcomplex QkComplex64;
+#else
 typedef float complex QkComplex32;
 typedef double complex QkComplex64;
+#endif
```
jakelishman, line 17: These MSVC types are struct-like, just like what `QkComplex64` _should_ be - it's a lie that we say in this file that it's equivalent to `double complex`, because actually the type is more properly `struct { double re, im }`.  We tell that lie because we only expose the type through ways where the lie doesn't matter (behind pointers), and so construction is easier using proper C99 tooling.

If you then `typedef _Dcomplex QkComplex64`, that's effectively a typedef to the _actual_ struct, and a user of it is going to hit all the problems we know about from having the actual struct.  If those problems are solvable, then we should probably do away with the `typedef double complex` in the first place and simply use the true definition.

eliarbel: If we do end up with this conditional definition (frankly I don't think we have a reasonable way currently, w/o e.g implementing our own complex type), we should provide a compiler-agnostic way to construct complex number variables, preferably in `qiskit.h` already. 

We can go with something like this: 

```c
#ifdef _MSC_VER
typedef _Fcomplex QkComplex32;
typedef _Dcomplex QkComplex64;
#define QK_CMPLXF(real, imag) (QkComplex32){real, imag}
#define QK_CMPLX(real, imag) (QkComplex64){real, imag}
#else
typedef float complex QkComplex32;
typedef double complex QkComplex64;
#define QK_CMPLXF CMPLXF
#define QK_CMPLX CMPLX
#endif
```

Then user code would just use the `QK_` complex macros e.g.: 
``` c
QkComplex64 coeff1 = QK_CMPLX(1.0, 0.0);
QkComplex32 coeff2 = QK_CMPLXF(1.0, -1.0);
```

Otherwise, users will have to use different initializations, depending on the platform: 
```c
QkComplex64 coeff = {1.0, 0.0};  // MSVC way: struct like, no I
QkComplex64 coeff = 2.0 + I * 1.0;  // C99 way

```

jakelishman: Eli: right, but in this situation, why don't we just use the true type right from the start? All the abstracted methods to handle cross-platform portability are the same as you'd need if we just used the true type (`struct { double re,im; }`).

Cryoris: At this point I'm a bit unsure about officially including a constructor that creates either `complex double` or `_DComplex` (but called the same thing, `QkComplex64`) in the header. We couldn't guarantee and maths/operations and wouldn't be able to use the convenience of C99 standards anywhere -- which basically goes to what Jake said; then we could just directly use `QkComplex64` as struct and remove the typedefs. 

Since this seems to need more discussion, we can maybe go ahead with #14030 and think more about the proper way of dealing with complex (e.g. custom struct vs. typedefs) for 2.1?

eliarbel: Yes, you're both right, I guess I haven't thought it through. While there are many common functions between C99 and MSVC (e.g. `cconj`, `cpow`, `ccos`), arithmetic operators are not supported natively in MSVC. So in order to write a fully compiler-agnostic way, one will have to e.g. use `creal` and `cimag` to calculate over the components directly and then assign back with a custom constructor like suggested here. 
We can define macros to abstract away the differences (e.g. something like `ADD_CMPLX`) so we can get around the gaps but it's not the most elegant way. The benefit is that we facilitate writing compiler-agnostic code over our C-API this way, in the cost of imposing kind of a hybrid, qiskit-specific way for using complex numbers in qiskit client code, and the theoretical risk of needing to maintain the abstraction (thought probably very low risk considering how often these areas in the language spec change or the probability of seeing native complex number support in MSVC :-) ).

